PR #1877, opened 2023-11-22
Clean up analysis post logic / CUPED settings

bryce-fitzsimons:
### Features and Changes

- remove stats engine and CUPED info from postSnapshot
- move CUPED metric setting logic to shared
- fix agenda bug
- fix some issues with metric overrides + fact metrics
- small UI cleanups

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -5,8 +5,10 @@ import cloneDeep from "lodash/cloneDeep";
 import { DEFAULT_SEQUENTIAL_TESTING_TUNING_PARAMETER } from "shared/constants";
 import { getValidDate } from "shared/dates";
 import { getAffectedEnvsForExperiment } from "shared/util";
+import { getAllMetricRegressionAdjustmentStatuses } from "shared/experiments";
 import { getScopedSettings } from "shared/settings";
 import { v4 as uuidv4 } from "uuid";
+import uniq from "lodash/uniq";
 import { AuthRequest, ResponseWithStatusAndError } from "../types/AuthRequest";
 import {
   createManualSnapshot,
@@ -18,7 +20,7 @@ import {
   getLinkedFeatureInfo,
   getManualSnapshotData,
 } from "../services/experiments";
-import { MetricStats } from "../../types/metric";
+import { MetricInterface, MetricStats } from "../../types/metric";
 import {
   createExperiment,
   deleteExperimentByIdForOrganization,
@@ -65,7 +67,7 @@ import {
   ExperimentTargetingData,
   Variation,
 } from "../../types/experiment";
-import { getMetricMap } from "../models/MetricModel";
+import { getMetricById, getMetricMap } from "../models/MetricModel";
 import { IdeaModel } from "../models/IdeasModel";
 import { IdeaInterface } from "../../types/idea";
 import { getDataSourceById } from "../models/DataSourceModel";
@@ -80,8 +82,6 @@ import {
   ExperimentSnapshotAnalysisSettings,
   ExperimentSnapshotInterface,
 } from "../../types/experiment-snapshot";
-import { StatsEngine } from "../../types/stats";
-import { MetricRegressionAdjustmentStatus } from "../../types/report";
 import { VisualChangesetInterface } from "../../types/visual-changeset";
 import { PrivateApiErrorResponse } from "../../types/api";
 import { EventAuditUserForResponseLocals } from "../events/event-types";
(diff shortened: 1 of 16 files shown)


packages/back-end/src/controllers/experiments.ts
```diff
@@ -1707,12 +1705,50 @@ export async function postSnapshot(
...
+  } = getAllMetricRegressionAdjustmentStatuses({
+    allExperimentMetrics,
+    denominatorMetrics,
+    orgSettings,
+    statsEngine,
+    experimentRegressionAdjustmentEnabled:
+      experiment.regressionAdjustmentEnabled,
+    experimentMetricOverrides: experiment.metricOverrides,
+    datasourceType: datasource?.type,
+    hasRegressionAdjustmentFeature: true,
```
msamper, line 1750: Is this hard coded because the front-end is already responsible for the check?

msamper (commented):
Nice refactor! From a first look the code looks good to me although I'll need to spend more time understanding the logic for the two shared `getRegressionAdjustment` functions. From what I can tell it seems like the logic was mostly just moved over and is now referenced from the API. I'll let someone else with a better grasp of this code give an approval. Going to do some QA and will report if I spot any issues.

packages/back-end/src/controllers/experiments.ts
bryce-fitzsimons: mostly yeah. Subjective UX decision, but I figure if they've somehow gotten to the state where CUPID could have been enabled and then they lose support somehow (license expired), we don't immediately want to change their backend analysis based on pre-existing frontend changes.

packages/shared/src/experiments.ts
```diff
@@ -94,3 +103,172 @@ export function getMetricLink(id: string): string {
...
+
+  // get RA settings from metric override
+  if (metricOverrides) {
+    const metricOverride = metricOverrides.find((mo) => mo.id === metric.id);
+    if (metricOverride?.regressionAdjustmentOverride) {
+      regressionAdjustmentEnabled = !!metricOverride?.regressionAdjustmentEnabled;
+      regressionAdjustmentDays =
+        metricOverride?.regressionAdjustmentDays ?? regressionAdjustmentDays;
+      if (!regressionAdjustmentEnabled) {
+        reason = "disabled by metric override";
```
lukesonnet, line 161: Is it possible to have it turned off at the metric and turned off at the override level? It would still work in that case but our `reason` might get mangled.

lukesonnet, line 161: Is it possible to have it turned off at the metric and turned off at the override level? It would still work in that case but our reason might get mangled.

packages/shared/src/experiments.ts
```diff
@@ -94,3 +103,172 @@ export function getMetricLink(id: string): string {
...
+      regressionAdjustmentEnabled = false;
+      reason = "ratio metrics not supported";
+    }
+    if (metric?.denominator) {
+      const denominator = denominatorMetrics.find(
+        (m) => m.id === metric?.denominator
+      );
+      if (denominator && !isBinomialMetric(denominator)) {
+        regressionAdjustmentEnabled = false;
+        reason = "denominator is binomial";
```
lukesonnet, line 180: This is incorrect. If denominator is binomial we allow CUPED. The original logic here was right. It should read `denominator is count` and should replicate the original logic.

packages/shared/src/experiments.ts
```diff
@@ -94,3 +103,172 @@ export function getMetricLink(id: string): string {
...
+    }
+  }
+
+  // final gatekeeping
+  if (regressionAdjustmentEnabled) {
+    if (metric && isFactMetric(metric) && isRatioMetric(metric)) {
+      regressionAdjustmentEnabled = false;
+      reason = "ratio metrics not supported";
+    }
+    if (metric?.denominator) {
```
lukesonnet, line 174: To help clarify what's going on, maybe the first check should say 

above the first if statement in L170
// disable CUPED for fact ratio metrics

above the second in L174
// disable CUPED for classical metrics with count denominators (e.g. old ratio metrics)

lukesonnet (changes requested):
Left comments in the shared logic file. Haven't had time to review the rest or QA but would like to later today if it would be helpful.

packages/shared/src/experiments.ts
bryce-fitzsimons: made the language/reason more clear in this case

packages/shared/src/experiments.ts
bryce-fitzsimons: does !binomial necessarily mean count?

lukesonnet (approved):
lgtm from QA and the little bit I reviewed

Issue #1844, opened 2023-11-13
[Bug] Incorrect "In Progress Conversions" behavior with ratio metrics that have different conversion windows


jdorn:
### Summary

If the numerator and denominator of a ratio metric have different conversion windows, we currently only use the numerator when determining the cutoff for "Exclude In Progress Conversions".  We should instead of use MAX of the numerator and denominator's conversion windows.

august-growthbook:
More info from the customer:

Hi ­@Jeremy Dorn,
I’m not sure what you wrote in the ticket describe what I meants.

I try to put it more clearly:
I want my user to fulfill the whole funnel to be counted when I am in exluded in progress conversions,

In my example if my ratio is 100 / 1000 but both events have different conversion windows, I am quite convinced that each metric is computed on its own conversion window and not further selection is performed where only users who completed both conversion windows should be part of the final computation.

currently If i compare 2 funnel metrics :
• metric A / metric B
• metric C / metric B
with conversion windows
metric A = 2 days
metric B = 1 day
metric C = 5 days

I might have :
• metric A / metric B = 10/100
• metric C / metric B = 5 / 100
because metric B is computed with the same conversion window (1day) in both cases

PR #1880, opened 2023-11-23
Use maximum conversion window of metric chain when 'Exclude in-progress conversions' is selected

kevinalh:
### Features and Changes

When using the 'Exclude in-progress conversions' option, if we have a metric with a denominator, currently the exclusion is done using the denominator's conversion window. We should be using the maximum of the numerator and denominator's conversion windows instead. More generally, the change in this PR takes the maximum conversion window 
of the numerator and the recursive chain of denominators.

- Closes **https://github.com/growthbook/growthbook/issues/1844**

### Testing

Following these steps:
1. Selecting 'Exclude in-progress conversions' option in the experiment
2. Choosing a metric with conversion window of 24 hours, with a denominator of 1 hour (as an example)

in the current behavior, the exclusion was done using a 1 hour period, with this change it's 24 hours instead.

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -1373,19 +1373,16 @@ export default abstract class SqlIntegration
       settings.experimentId
     );
 
-    const initialMetric =
-      denominatorMetrics.length > 0 ? denominatorMetrics[0] : metric;
-
     // Get date range for experiment and analysis
-    const initialConversionWindowHours = getConversionWindowHours(
-      initialMetric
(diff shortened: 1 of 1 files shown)


lukesonnet:
Thanks for contributing this, it lgtm and cleans up some unnecessarily verbose code!

I'll do some testing myself before approving.

lukesonnet:
Thank you for this contribution! I'm going to close this because it doesn't capture all of the complexity we have with defining date windows (e.g. funnel metrics are different from ratio metrics) but I definitely borrowed from this approach to patch the bug.

You can follow here: https://github.com/growthbook/growthbook/pull/1921

PR #1883, opened 2023-11-27
Move URL check higher up SDK _run chain

bttf:
This fixes a bug occurring with visual experiments and preview links when the experiment had multiple pages.

It effectively moves our URL targeting check up in the series of checks in `_run` so that we exit earlier for experiments that do not have applicable URL targeting.

Tested locally and works as expected


jdorn:
We should add a unit test for this - one that fails on `main` and passes on this branch.  We're going to make other changes near here soon and want to make sure we don't have a regression.

bttf:
> We should add a unit test for this - one that fails on `main` and passes on this branch. We're going to make other changes near here soon and want to make sure we don't have a regression.

Tests added for existing and new functionality re: preview links in the SDK

jdorn (approved):

PR #1893, opened 2023-11-29
SDK compatibility versioning

bryce-fitzsimons:
### Features and Changes

- SDK version compatibility methods. Given an SDK language and version, what capabilities do we have?
- SDK version setting in SDK Connection UI
- SDK payload scrubbing when using a language + version that does not support loose JSON unmarshalling.

todo:
- [x] connections with multiple languages: safely handle sdk payload capabilities instead of defaulting to "other" (compatible)
- [ ] more in-app messaging & gating?
- [x] migrate some capability info away from SDKLanguageLogo to the new sdk-versioning methods.

### Screenshots

image


packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -104,6 +113,17 @@ export default function SDKConnectionForm({
...
   });
 
+  const usingLatestVersion = !isSDKOutdated(
+    form.watch("languages")?.[0] || "other",
+    form.watch("sdkVersion")
+  );
+
+  const useLatestSdkVersion = () => {
+    const language = form.watch("languages")?.[0] || "other";
+    const latest = getCurrentSDKVersion(language);
```
msamper, line 123: Is `getCurrentSDKVersion` grabbing the SDK version for the user or the latest available version for the language? If it's the latter I think the function should be called `getLatestSDKVersion`

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -121,22 +141,18 @@ export default function SDKConnectionForm({
       ? "backend"
       : "hybrid";
 
-  const selectedLanguagesWithoutRemoteEvalSupport = languages.filter(
-    (l) => !languageMapping[l].supportsRemoteEval
-  );
-
-  const selectedLanguagesWithoutEncryptionSupport = languages.filter(
-    (l) => !languageMapping[l].supportsEncryption
+  const maxSdkCapabilities = getConnectionSDKCapabilities(
```
msamper, line 144: Not sure if I understand what max capabilities means here. Is it max capabilities for the user's SDK? If so what is the difference between this and `currentSDKCapabilities`? I'll take a look at the actual function after this but maybe worth making the name of the consts more clear

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -739,36 +786,33 @@ export default function SDKConnectionForm({
...
-                                {i <
-                                  selectedLanguagesWithoutRemoteEvalSupport.length -
-                                    1 && ", "}
-                              </span>
-                            )
-                          )}
-                        </div>
+                        <FaExclamationCircle /> Remote evaluation may not be
+                        available in your current SDK.
+                        {languages.length === 1 && (
```
msamper, line 796: I noticed in a comment in another file you added a comment that we prefer to have a single language in this array. Do we surface that as a callout to the user anywhere in the UI? Might be out of scope, but just curious if we do since it seems like we want to steer them in that direction

packages/front-end/components/Features/SDKConnections/SDKLanguageLogo.tsx
```diff
@@ -176,23 +124,50 @@ export default function SDKLanguageLogo({
   showLabel = false,
   size = 25,
   titlePrefix = "",
+  version,
 }: {
   language: SDKLanguage;
   showLabel?: boolean;
   size?: number;
   titlePrefix?: string;
+  version?: string;
```
msamper, line 133: Is this latest version, users version, or something else?

packages/front-end/components/Features/SDKConnections/SDKLanguageLogo.tsx
```diff
@@ -176,23 +124,50 @@ export default function SDKLanguageLogo({
...
 }) {
   const { Icon, color, label } =
     languageMapping[language] || languageMapping["other"];
 
+  const versionOutdated = isSDKOutdated(language, version);
+
+  const labelText = label;
```
msamper, line 140: maybe rename when you destructure the object instead?

packages/shared/src/sdk-versioning/index.ts
```diff
@@ -0,0 +1,171 @@
...
+
+const getSdkData = (language: SDKLanguage = "other"): SDKData => {
+  let sdkData: SDKData = sdks[language];
+  if (!sdkData) {
+    sdkData = sdks["other"];
+  }
+  return sdkData;
+};
+
+export const getCurrentSDKVersion = (
```
msamper, line 55: nit: `getLatestSDKVersion`?

packages/shared/src/sdk-versioning/index.ts
```diff
@@ -0,0 +1,171 @@
...
+  version: string = "0.0.0"
+): SDKCapability[] => {
+  language = language || "other";
+  version = version && version !== "0" ? version : "0.0.0";
+  const sdkData = getSdkData(language);
+  const versions = sdkData?.versions || [];
+  const matches = versions.filter(
+    (data) => paddedVersionString(data.version) <= paddedVersionString(version)
+  );
+  const capabilities = matches.reduce(
```
msamper, line 83: Are capabilities per version additive to previous versions or are they an exhaustive list of all the capabilities for a given version? If they're exhaustive we could probably just take the capabilities from the matching version instead of reducing, but I'm guessing they're not exhaustive

packages/shared/src/sdk-versioning/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    (acc, data) => [...acc, ...(data?.capabilities ?? [])],
+    []
+  );
+  return uniq(capabilities) as SDKCapability[];
+};
+
+// Typically works the same as getCapabilities. However, if the connection has multiple languages, assume the
+// minimal-allowed SDK Version (0.0.0), and return the intersection of capabilities between all languages.
+export const getConnectionSDKCapabilities = (
```
msamper, line 92: I don't think I understand this logic, but I may just need some more context. Why are we assuming the minimal allowed SDK version if theres multiple languages?

msamper (commented):
Mostly questions and naming suggestions on this initial review while I try to understand. Might be worth doing a quick huddle so I can understand some things a bit faster

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
bryce-fitzsimons: "max" is basically the latest SDK version's list of capabilities, whereas "current" are those present in the user's current SDK version

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
bryce-fitzsimons: we basically force them to have array size of 1 for new SDK Connections, but we try not to force their hand with pre-existing connections. I don't think it needs to be called out anywhere since the UX kinds of pulls them towards 1.

packages/front-end/components/Features/SDKConnections/SDKLanguageLogo.tsx
bryce-fitzsimons: whatever you want it to be. The language logo is a reusable visible component that you can plug in a version number for and it will render it.

packages/shared/src/sdk-versioning/index.ts
bryce-fitzsimons: they are not exhaustive, they are incremental

packages/shared/src/sdk-versioning/index.ts
bryce-fitzsimons: i can talk through this one later, but it's really the only safe way to make SDK assumptions. Case in point is the recent Python regression; this algorithm would have protected us.

packages/front-end/components/Features/SDKConnections/SDKLanguageLogo.tsx
```diff
@@ -176,23 +124,49 @@ export default function SDKLanguageLogo({
...
   language: SDKLanguage;
   showLabel?: boolean;
   size?: number;
   titlePrefix?: string;
+  version?: string;
 }) {
   const { Icon, color, label } =
     languageMapping[language] || languageMapping["other"];
 
+  const versionOutdated = isSDKOutdated(language, version);
```
msamper, line 138: Need to handle the case where version is undefined here

msamper (commented):
Screenshot 2023-12-05 at 1 20 23 PM
Ran into a state when editing an existing connection where the connection's specified version is blank. I'm assuming this has something to do with the default value of the dropdown, but maybe its worth either hiding the second sentence until the user selects a version or prompting them to select a version to view compatibility info. We can also default to 0.0.0 but that seems confusing. Aside from that, the UI lgtm. The dropdown is definitely a better experience. There seems to be a typecheck error with version being undefined that you need to handle and I left a comment where it is

packages/front-end/components/Features/SDKConnections/SDKLanguageLogo.tsx
bryce-fitzsimons: oops forgot the `?:`

bryce-fitzsimons:
> Ran into a state when editing an existing connection where the connection's specified version is blank.
Should be fixed now, can you try again please?

msamper:
> > Ran into a state when editing an existing connection where the connection's specified version is blank.
> 
> @msamper Should be fixed now, can you try again please?

the messaging in the bottom is fixed but I think the value in the dropdown needs to be adjusted
Screenshot 2023-12-05 at 1 33 46 PM

packages/back-end/src/controllers/features.ts
```diff
@@ -166,6 +172,7 @@ export async function getPayloadParamsFromApiKey(
 
     return {
       organization,
+      capabilities: [],
```
jdorn, line 175: Is this going to break payloads for legacy API keys?  I'm pretty sure those payloads include hashing v2 fields today.

bryce-fitzsimons: updated the scrubbing to be multi-staged (and passed in capabilities here) based on our convo...

packages/shared/src/sdk-versioning/sdk-versions/php.json
```diff
@@ -0,0 +1,8 @@
+{
+  "versions": [
+    {
+      "version": "1.2.0",
+      "capabilities": ["bucketingV2", "encryption"]
+    }
```
jdorn, line 6: Add `looseUnmarshalling` to either version `0` or `1.0.0` depending on how we interpret it.  Prior to 1.0.0, we would log a warning on unknown keys, but not stop execution.

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  const allowedFeatureKeys = [...strictFeatureKeys];
+  const allowedFeatureRuleKeys = [...strictFeatureRuleKeys];
+  if (capabilities.includes("bucketingV2")) {
+    allowedFeatureRuleKeys.push(...bucketingV2Keys);
+  }
+
+  for (const k in features) {
+    if (!capabilities.includes("looseUnmarshalling")) {
```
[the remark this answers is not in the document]
bryce-fitzsimons: oops, forgot to clean up from earlier

jdorn (approved):

Issue #1774, opened 2023-10-20
[Bug] Expired License Keys Cause API Error State


Auz:
When a user adds an enterprise or pro key through the GrowthBook UI, and then that key expires, the API will throw an error about the expired license and not run. This means users cannot access the UI to delete the expired license key.

newb1e:
I have encountered a similar but even weirder bug. We did have some license which expired Oct 31st. 
We get a message about "Your License Key trial expired on 2023-10-31", though we're not using this key anymore. 
Upgrading to version 2.4.0 works well. Upgrade to version 2.5.0 locks me out the UI with this message.

newb1e:
As a workaround for my issues I had to connect to the Mongodb. After connecting to the mongodb server and appropriate db (organizations) was able to run:
```db.organizations.updateOne( { _id: ObjectId("obj_id") }, { $set: { licenseKey: '' } } )```

smrdotgg:
I've replicated and fixed the bug, but the code change occurred in the `enterprise` package. 

Should I make a PR anyway? @Auz

PR #1895, opened 2023-11-30
Fix expired license error handling

smrdotgg:
### Features and Changes

 - Fixes the issue where an expired license key causes an API error, preventing them from logging in and accessing the UI for license management. This issue was limited to API keys who's trials expired. Paying enterprise members who's license expired did not face this issue.

- Fixes [Expired License Keys Cause API Error State (#1774)](https://github.com/growthbook/growthbook/issues/1774)

### Dependencies

None

#### Test 1 - Login
1. Add an enterprise or pro key on a new account through the GrowthBook UI. (Ensure it is a trial key)
2. Allow the key to expire. (Can be simulated by changing your machine's date/time settings)
3. Ensure the license data isn't cached in server memory from when you added the key. (Memory can be flushed by restarting the dev server)
4. Attempt to log in using an account associated with the expired trial key.

Before this change, the login would fail.
After this change, the login succeeds and displays a banner on the top bar.

#### Test 2 - Unauthorized requests
1. Add an enterprise or pro key on a new account through the GrowthBook UI. (Ensure it is a trial key)
2. Allow the key to expire. (Can be simulated by changing your machine's date/time settings)
3. Ensure the license data isn't cached in server memory from when you added the key. (Memory can be flushed by restarting the dev server)
4. Log in using an account associated with the expired trial key.
5. Find an enterprise only feature that has been disabled. We can take exporting logs as json, which can be found in the `/events` route of the frontend.
6. The button will be disabled. Force-enable the button. For the case of the events export feature, this can be done by setting the `disabled` flag on the button in the EventsPage component (`packages/front-end/components/Events/EventsPage/EventsPage.tsx`) to false. 
7. Once the button is enabled, press it.

Before this change, the backend would process the request and provide a valid json response eventhough the license has expired. The feature-blocking was only happening in the frontend logic. (This could have also been bypassed with `curl`)

After this change, the backend will do its own check to see if the license key has expired. If it has, it will not process the request for the premium feature. So in this case, it would not process the invalid request.

#### Before
Before
#### After
After

### DIFF
--- a/packages/enterprise/src/license.ts
+++ b/packages/enterprise/src/license.ts
@@ -206,16 +206,6 @@ export async function getVerifiedLicenseData(
   }
   delete decodedLicense.eat;
 
-  // The `trial` field used to be optional, force it to always be defined
-  decodedLicense.trial = !!decodedLicense.trial;
-
-  // If it's a trial license key, make sure it's not expired yet
-  // For real license keys, we show an "expired" banner in the app instead of throwing an error
-  // We want to be strict for trial keys, but lenient for real Enterprise customers
-  if (decodedLicense.trial && decodedLicense.exp < new Date().toISOString()) {
-    throw new Error(`Your License Key trial expired on ${decodedLicense.exp}.`);
-  }
-
   // We used to only offer license keys for Enterprise plans (not pro)
   if (!decodedLicense.plan) {
     decodedLicense.plan = "enterprise";
--- a/packages/enterprise/src/license.ts
+++ b/packages/enterprise/src/license.ts
@@ -169,7 +169,9 @@ export function orgHasPremiumFeature(
   org: MinimalOrganization,
   feature: CommercialFeature
 ): boolean {
-  return planHasPremiumFeature(getAccountPlan(org), feature);
+  return (
+    !licenseIsExpired() && planHasPremiumFeature(getAccountPlan(org), feature)
+  );
 }
 
 async function getPublicKey(): Promise<Buffer> {
--- a/packages/enterprise/src/license.ts
+++ b/packages/enterprise/src/license.ts
@@ -170,7 +170,8 @@ export function orgHasPremiumFeature(
   feature: CommercialFeature
 ): boolean {
   return (
-    !licenseIsExpired() && planHasPremiumFeature(getAccountPlan(org), feature)
+    !shouldLimitAccessDueToExpiredLicense() &&
+    planHasPremiumFeature(getAccountPlan(org), feature)
   );
 }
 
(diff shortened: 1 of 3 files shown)


august-growthbook:
Thanks for the PR! Someone from Engineering will review it.

packages/enterprise/src/license.ts
```diff
@@ -189,16 +191,6 @@ export async function getVerifiedLicenseData(key: string) {
   }
   delete decodedLicense.eat;
 
-  // The `trial` field used to be optional, force it to always be defined
-  decodedLicense.trial = !!decodedLicense.trial;
```
tzjames, line 193: Something else might still rely on this code.  Better to leave it in.

packages/enterprise/src/license.ts
```diff
@@ -275,3 +267,33 @@ export async function setLicense(l: LicenseData | null) {
...
+    // Add a buffer of 2 days to the expiration date
+    const expiresWithBuffer = new Date(expirationDate);
+    expiresWithBuffer.setDate(expirationDate.getDate() + 2);
+
+    // Check if the adjusted expiration date is in the past
+    if (expiresWithBuffer < new Date()) {
+      // The license is expired
+      return true;
+    }
+  }
```
tzjames, line 295: From my understanding before if it was a `.trial` license and it expired we errored immediately with no buffer.  It errored too harshly to the point that the app became unusable.  This change would give them the two day buffer as well, but I think we only want to have the two day grace period with enterprise customers.

tzjames (changes requested):
Thanks again for doing this Semere!  It looks almost perfect.  I left a few comments though.  Did you check out an api endpoint that has enterprise features using curl?  Do they throw good error messages?

packages/enterprise/src/license.ts
smrdotgg: Good call, I'll leave that in.

packages/enterprise/src/license.ts
smrdotgg: Yeah, as it is, it adds the buffer for all users. I'll make sure it only does so for non-trial users.

smrdotgg:
> Thanks again for doing this Semere! It looks almost perfect. I left a few comments though. Did you check out an api endpoint that has enterprise features using curl? Do they throw good error messages?

I have tested the api endpoints with curl, and they only execute if the license key is valid. If its invalid, they throw appropriate errors with correct codes and proper messages.

I've also added the changes you proposed. :smiley:

packages/enterprise/src/license.ts
```diff
@@ -275,3 +269,34 @@ export async function setLicense(l: LicenseData | null) {
...
+    return false;
+  }
+
+  // If expiration date is provided in the license data
+  if (licenseData.exp) {
+    // Create a date object for the expiration date
+    const expirationDate = new Date(licenseData.exp);
+
+    // Add a buffer of 2 days to the expiration date for non-trial users
+    if (licenseData.trial) {
```
tzjames: I think you are missing a !

tzjames (approved):
It looks good now.  Congrats!

tzjames (changes requested):
We chatted in person about making the non-trial licenses just show the warning but not impede functionality.

tzjames (approved):
This is looking good now.

PR #1896, opened 2023-11-30
🚧  Add 'No Access' Role 🚧

mknowlton89:
### Features and Changes

This PR introduces a new `No Access` role type. This `noaccess` role will be available to enterprise organizations, and can give organizations the ability to completely hide projects from users.

There are a few different use cases we're looking to support.

1. An organization has a singular project that they want to hide from a number of users. In this case, all the organization would need to do is add a project-level `noaccess` role for users they want to hide the project from. This will hide the project from them entirely, meaning, it won't be visible to the user from the Project Selector dropdown in the top-left corner of the SideNav. And, experiments and features that are in this project won't be visible to the user (either via the features/experiments index pages, or on the individual feature/experiment page.

2. An organization wants their users to only see certain projects they have access to. In this case, the organization could give all user's a global role of "noaccess", which would hide ALL projects from the user, and then, they could give the user project-level permissions only for the projects the user can see.

### How this new feature works
In addition to adding the new `noaccess` role, we've built two new functions in the `permission.utils.ts` file - **hasReadAccess** and **getReadAccessFilter**.

**getReadAccessFilter** takes in a user's permissions, and builds an object to communicate what the user has readAccess to. The shape of the `readAccessFilter` is:

```ts
{
   globalReadAccess: boolean // this is determined by the user's global role
   projects: { id: string; readAccess: boolean}[]
}
```
This `readAccessFilter` is generated on every API request and it's added to the `req` object (this is for both our internal API and the REST API).

This is then ultimately used alongside the `hasReadAccess` function at the DB model level. In most cases, we will get the resource(s) from the DB, and then use the `hasReadAccess` to filter the resources down to only those the user has access to.

### Testing

- [x] - Create a new user where their global role is `noaccess` and then give them a project-level role of `experimenter`. Ensure that when you log in as that user, you only see experiments/features that are part of the project they have access to, and you don't see features/experiments that are in other projects or no projects.
- [x] - With the same user as above, ensure that you can only see the 1 project in the `ProjectSelector` drop-down in the top-left corner of the SideBar.
- [x] - With the same user, attempt to go to a [fid] or [eid] page of a feature/project that is in a project you don't have access to, and ensure that it is blocked.
- [x] - Create a new user where their global role is `experimenter` and then give them a project-level role of `noaccess` for a single project. When you log in as that user, ensure you can see all experiments/features that don't have a project, as well as features/experiments in all projects BUT the project they have `noaccess` for.
- [x] - With the same user as above, ensure that you can see all projects in the `ProjectSelector` drop-down, except for the single project the user has `noaccess` for.
- [x] Log in as a non-enterprise organization, and when you go to invite a new user, ensure the `noaccess` role isn't an option from the role drop-down.


bttf (commented):
Took a look through a portion today - leaving comments that I've made so far. Can look more tomorrow and probably sync up at some point as well

packages/back-end/src/queryRunners/MetricAnalysisQueryRunner.ts
```diff
@@ -75,6 +75,7 @@ export class MetricAnalysisQueryRunner extends QueryRunner<
     const model = await getMetricById(
       this.model.id,
       this.model.organization,
```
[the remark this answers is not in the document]
mknowlton89: @jdorn This is one of the questions I had for you.

packages/back-end/src/jobs/expireOldQueries.ts
```diff
@@ -64,7 +64,8 @@ export default async function (agenda: Agenda) {
     // Look for matching metrics and update the status
     const metrics = await findRunningMetricsByQueryId(
       [...orgIds],
-      [...queryIds]
+      [...queryIds],
+      { globalReadAccess: true, projects: [] } // This agenda job shouldn't have any read access restrictions
```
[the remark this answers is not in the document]
mknowlton89: @jdorn This is one of the questions I had for you.

packages/enterprise/src/license.ts
```diff
@@ -33,6 +33,7 @@ export type CommercialFeature =
   | "remote-evaluation"
   | "multi-org"
   | "custom-launch-checklist"
+  | "no_access_role"
```
mknowlton89, line 36: Removing this and using `advanced-permissions` instead

packages/enterprise/src/license.ts
```diff
@@ -131,6 +132,7 @@ export const accountFeatures: CommercialFeaturesMap = {
     "multi-org",
     "teams",
     "custom-launch-checklist",
+    "no_access_role",
```
mknowlton89, line 135: Removing this and using advanced-permissions instead

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -649,15 +652,21 @@ export async function getOrganization(req: AuthRequest, res: Response) {
...
     status: 200,
     apiKeys,
     enterpriseSSO,
     accountPlan: getAccountPlan(org),
-    commercialFeatures: res.locals.licenseError
-      ? []
-      : [...accountFeatures[getAccountPlan(org)]],
-    roles: getRoles(org),
+    commercialFeatures,
+    roles: !commercialFeatures.includes("no_access_role")
```
mknowlton89, line 667: doing this on the front end instead. Removing...

mknowlton89:
Breaking this PR up into multiple PRs. Closing this.

PR #1899, opened 2023-11-30
Fix environment permission bugs

### LINKED ISSUE #1684: [Bug] Users with `Limit access by environment` can't create/edit in Visual Editor
### Summary

[Slack convo here](https://growthbookapp.slack.com/archives/C05QYS2UFRD/p1695227524687569)

#### Primary issue

Visual Editor Experiments apply to all Environments. So if a user's role has "Restrict access by environment" checked, we currently block them from editing in the Visual Editor, since they might be making changes via the Visual Editor to environments they're not supposed to have access to.

An Admin would need to uncheck the "Restrict access by environment" toggle in order for the user in question to be able to make changes (to all environments!) in the Visual Editor.

This is very confusing to users and we should change the permissions logic a bit.  If "Restrict access by environment" is checked, but all environments are selected (as in the Visual Editor use case), we should treat permissions the same as if the "Restrict access by environment" toggle **wasn't** enabled.

#### Expected Behavior

Users should be able to make edits in the Visual Editor as if they had access to all Environments even if their permissions are restricted using the "Restrict access to environments" toggle ([per Jeremy's comment in the Slack thread](https://growthbookapp.slack.com/archives/C05QYS2UFRD/p1695227524687569)).

#### Current Behavior

Users with the "Restrict access to environments" toggle enabled cannot make edits in the Visual Editor.

#### Secondary issue

Jeremy discovered another bug while investigating the issue reported above.

We aren't actually enforcing any permissions when someone adds their very first visual change to an experiment.  We only enforce it when they edit. So the user should never have been able to use the Visual Editor in the first place.  They only found out about it since they tried to edit an existing targeting rule.

jdorn:
### Features and Changes

This PR fixes the following 3 bugs with permissions:
1. It's possible to get into a state where a user with an "admin" role has `limitAccessByEnvironment` set to true. These environment settings should be ignored when the role doesn't support them, but they weren't.
2. If all environments in an organization are selected when limiting access by environment, it was not being treated the same as if the user was not limited by environment in the first place
3. The analyst role was missing the `createSegments` permission

In addition, it cleans up the permissions test suite to make it easier to read and maintain (resulted in finding bug 3 above)

Fixes #1684


packages/back-end/test/permissions.test.ts
```diff
@@ -35,39 +45,7 @@ describe("Build base user permissions", () => {
...
-          createIdeas: false,
-          createMetrics: false,
-          createDatasources: false,
-          editDatasourceSettings: false,
-          runQueries: false,
-          publishFeatures: false,
-          manageEnvironments: false,
-          runExperiments: false,
-        },
+        permissions: roleToPermissionMap("readonly", testOrg),
```
mknowlton89, line 48: Much cleaner.

aminland:
Any word on when this will be released? It's causing problems for us at the moment...

PR #1900, opened 2023-11-30
Update README.md for docs site

august-growthbook:
Change order of instructions for running the docs locally so they are more clear.


mknowlton89 (approved):
LGTM

PR #1907, opened 2023-12-01
(Bug) - Segment w/ Postgres Discover Metrics Requires Schema

mknowlton89:
### Features and Changes

An organization found an issue when trying to use the `Discover Metrics` feature for a Segment + Postgres Data Source.

[Link to User Group thread](https://growthbookusers.slack.com/archives/C01T6Q1SVFV/p1700705326222549).

The organization kept getting a message that the `tracks` table doesn't exist, however, they were able to confirm via the `Schema Browser` that it does exist within their Data Source.

After doing a little digging, it appears that the `FROM` clause of the query we run to get a list of unique events Segment is tracking is incorrect - in this case, the `FROM` clause was simply `FROM tracks`. However, after doing a little bit of testing, it appears that the `FROM` clause actually needs the schema AND the table name. EX: The query should read something like `FROM <schema>.tracks`.

However, the `PostgresIntegration` file has `requiresSchema` set to false, which is true for all other cases (from what I can tell).

To address this, I updated how we're calculating the trackedEventTableName to include the `schema` if the dialect is "postgressql". This felt the cleanest, and like it could expand to support other use-cases down the road.

The other option that I explored was using the `queryRequiresSchema` override property on the `generateTablePath` method. Ultimately, while this prop does exactly what we want it to do, I felt it made the code in the `getTrackedEvents` method a little harder to read, but I'm open to feedback.

### Testing
Ultimately, I don't have a real Segment with a Postgres destination set up in order to fully test this end-to-end, but I was able to run a test where I created a Segment Data Source and pointed it to a local Postgres database that I have running, and I was able to confirm that the query was being updated to include the schema, and the error message I received when trying to Discover Metrics was now more accurate (that my Postgres database doesn't have a `tracks` table).

Likewise, I was able to test a Segment + BigQuery data source that I have (this one is actually legitimate where Segment sends events to BigQuery) and in that, I was able to confirm that the updates made here didn't introduce any regressions.


mknowlton89:
Closing due to age.

PR #1909, opened 2023-12-01
Allow orgs to define role when creating and updating groups via SCIM

mknowlton89:
### Features and Changes

It's been requested now that we've added the ability for organizations using SCIM to define a `growthbookRole` when creating/updating users, that we do the same thing for `Teams`.

To do this, we've added a new optional property to the `/scim/v2/groups` `POST` and `PATCH` requests. This property, like the user property we just added is `growthbookRole`.

It's important to note that currently, Okta is the only Identity Provider that GrowthBook has SCIM support for officially, and, Okta doesn't offer the ability to include custom attributes when creating/updating groups. Which means that currently, no organizations using Okta's SCIM application can take advantage of this.

However, its possible other organizations are using SCIM with Identity Providers other than Okta that do support this, so we want to extend support. It'll be helpful for a few organizations now, and also make our SCIM integration more flexible as time goes on.

To take advantage of this new functionality, here are two sample API requests that have been tested.

#### Create Group
Request Type: `POST`
Endpoint: `https://SCIM_BASE_URL/scim/v2/groups`
Request Body:
```json
{
    "schemas": [
        "urn:ietf:params:scim:schemas:core:2.0:Group"
    ],
    "displayName": "Test Team",
    "growthbookRole": "readonly",
    "members": [
        {
            "value": "userid_123",
            "display": "email@address.com"
        }
    ]
}
```

#### Update Group
Request Type: `PATCH`
Endpoint: `https://SCIM_BASE_URL/scim/v2/groups/:teamId`
Request Body:
```json
{
    "schemas": [
        "urn:ietf:params:scim:schemas:core:2.0:PatchOp"
    ],
    "Operations": [
        {
            "op": "replace",
            "value": {
                "id": "teamId",
                "displayName": "Test Team Name",
                "growthbookRole": "admin"
            }
        }
    ]
}
```

### Testing

- [x] Create a new team via Postman with the request above, minus the `growthbookRole` property and ensure that it's created successfully and the team's global role is the organizations default role.
- [x] Now do the same thing, but instead include the `growthbookRole` property set to `"admin"` and ensure the team is created successfully and the team's global role is `admin`.
- [x] Now do the same thing, but give it an invalid `growthbookRole` and ensure that an error is thrown.
- [x]Now, update an existing team with a valid role, and ensure that the role is updated correctly
- [x] Now, update an existing team, but this time, only update the team name and ensure it works correctly
- [x] Now, update an existing team with a bogus role and ensure an error is thrown
- [x] Within a working Okta SCIM application, ensure you can still provision/de-provision users without issue, and can still push and un-sync groups without issue.


mknowlton89:
Quick note: I made a change in this PR that moves the `isValidRole` check to the shared folder. I also broke the function out into two methods in order to support the new SCIM role that the PR linked will introduce. Depending on if that PR lands before this, there'll be conflicts that I will need to resolve, but it'll just be changing the import location.

packages/back-end/src/scim/groups/patchGroup.ts
```diff
@@ -93,11 +95,26 @@ export async function patchGroup(
...
+              schemas: ["urn:ietf:params:scim:api:messages:2.0:Error"],
+              status: "400",
+              detail: `"${growthbookRole}" is not a valid GrowthBook role.`,
+            });
+          }
+
+          updatedTeam.role = growthbookRole;
+        }
+
+        await updateTeamMetadata(team.id, org.id, updatedTeam);
```
msamper, line 117: I think you can replace this with 
```
        await updateTeamMetadata(team.id, org.id, {
          ...team,
          name: (value as BasicScimGroup).displayName,
          managedByIdp: true,
          growthbookRole,
        });
```
and have the same effect without needing `cloneDeep`. If its undefined it just wont be added and if its not a valid role we'll return the error before we get to update

msamper (approved):
Just a minor comment about simplifying the update logic, but otherwise lgtm!

packages/back-end/src/scim/groups/patchGroup.ts
mknowlton89: Ah, yea - much cleaner. Thanks!

PR #1911, opened 2023-12-01
fix CUPED availability messaging on toggle

bryce-fitzsimons:
### Features and Changes


lukesonnet:
LGTM.

One weirdness is that if an experiment has no metrics you could still get this tooltip which, while *technically* correct still reads somewhat strangely.

But maybe we already suppress the toggle altogether in that case? But that's a tiny issue and otherwise the immediate problem is solved.

lukesonnet (approved):

PR #1917, opened 2023-12-04
Fixed date off by one in graph

### LINKED ISSUE #1065: [Bug] Date Dimension displaying incorrect dates on the chart compared to query detail
### Summary

When looking at the date dimension chart within an experiment analysis the dates are displaying incorrectly compared in the values shown by "view queries"
#### Expected Behavior

Dates in the chart should reflect the correct dates
#### Current Behavior

The dates display is 1 day less than it should be (March 6th is showing as March 6th)
chart

table

gazzdingo:
### Features and Changes

- Closes #1065

### DIFF
--- a/packages/front-end/components/Experiment/DateResults.tsx
+++ b/packages/front-end/components/Experiment/DateResults.tsx
@@ -3,7 +3,7 @@ import {
   ExperimentReportResultDimension,
   ExperimentReportVariation,
 } from "back-end/types/report";
-import { getValidDate } from "shared/dates";
+import { getValidDateUTC } from "shared/dates";
 import { StatsEngine } from "back-end/types/stats";
 import { ExperimentMetricInterface } from "shared/experiments";
 import { useDefinitions } from "@/services/DefinitionsContext";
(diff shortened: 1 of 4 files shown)


packages/front-end/components/Experiment/DateResults.tsx
```diff
@@ -95,7 +97,9 @@ const DateResults: FC<{
 
     const sortedResults = [...results];
     sortedResults.sort((a, b) => {
-      return getValidDate(a.name).getTime() - getValidDate(b.name).getTime();
+      return (
+        getValidDateUTC(a.name).getTime() - getValidDateUTC(b.name).getTime()
```
lukesonnet, line 101: It's probably actually not needed in these sort function (here and above in L70) since it'll just be doing extra offsetting to both dates. But it doesn't matter.

lukesonnet (changes requested):
This doesn't actually work for `DateGraph` because the metric analysis object actually stores the dates as `Date` in the backend, so you end up correcting it on the front-end in some ways but then you get things like the dots and the line not matching up (see where the dot is): 

Screenshot 2023-12-04 at 3 52 19 PM

So I think we also need to add the UTC method to L56 of MetricAnalysisQueryRunner.ts in the back-end. I think you could probably then remove it from L190 of DateGraph.tsx (although IDK if they'd stack).

packages/front-end/components/Experiment/DateResults.tsx
gazzdingo: cool i can remove it from here and l70

lukesonnet (approved):
lgtm!

PR #1925, opened 2023-12-07
Add health tab onboarding

lukesonnet:
This adds the health tab onboarding modal flow to the health tab.

Adds `HealthTabOnboardingModal`, a several step modal to define dimension data and turn on the health tab query. On final save does three things:
* Turns on health tab query in org settings
* Saves dimension metadata if it was specified
* Runs a snapshot update if the user asked for it

Adds a new endpoint to update just an exposure query as part of saving dimension metadata above. Note that the dimension metadata update from the data source page uses a different endpoint where it patches the whole data source object and updates that.

TODO: 
- [x] Resolve issue with analysis settings resetting
- [ ] Consider refactoring some of the code around triggering three endpoint calls onsave

Loom: https://www.loom.com/share/8d12463bf67d42e7b8be0529065450d6


packages/back-end/src/integrations/Mysql.ts
```diff
@@ -71,9 +71,6 @@ export default class Mysql extends SqlIntegration {
   ensureFloat(col: string): string {
     return `CAST(${col} AS DOUBLE)`;
   }
-  // From https://rpbouman.blogspot.com/2008/07/calculating-nth-percentile-in-mysql.html
-  // One pass, but builds a long string of all values and then cuts it at the right
-  // percentile
```
lukesonnet, line 76: I am intentionally deleting this comment.

packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
```diff
@@ -23,33 +24,32 @@ export default function HealthTab({
...
-  const { apiCall } = useAuth();
   const { refreshOrganization } = useUser();
   const permissions = usePermissions();
   const hasPermissionToEditOrgSettings = permissions.check(
     "organizationSettings"
   );
   const [healthIssues, setHealthIssues] = useState<IssueValue[]>([]);
   // Clean up notification counter & health issues before unmounting
+
+  const [uiMode, setUiMode] = useState<"open" | "setup">("open");
```
msamper, line 43: Is this just used for the open state of the modal? If so maybe name this something like `setupModalOpen` and make it a boolean instead of a string

packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
```diff
@@ -75,10 +75,21 @@ export default function HealthTab({
...
+                experiment={experiment}
+                phase={phase}
+                close={() => setUiMode("open")}
+                refreshOrganization={refreshOrganization}
+                mutateSnapshot={mutateSnapshot}
+                setAnalysisSettings={setAnalysisSettings}
+                setLoading={setLoading}
+              />
+            ) : null}
           </>
```
msamper, line 93: Expose `open` from the `Modal` component this is wrapping and pass in the state you're using (`uiMode`) so the `Modal` can handle hiding on its own

packages/front-end/components/Experiment/TabbedPage/HealthTabOnboardingModal.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+  if (step === 1) {
+    const saveEnabled =
+      id &&
+      status === "succeeded" &&
+      data?.dimensionSlices?.results &&
+      data.dimensionSlices.results.length > 0;
+    return (
+      <>
+        <Modal
```
msamper, line 165: I don't know that each step should be a separate modal. We should ideally be conditionally changing the `children` depending on the step as well as whatever else changes depending on the step (secondary CTA, etc)

packages/front-end/components/Settings/EditDataSource/DimensionMetadata/UpdateDimensionMetadata.tsx
```diff
@@ -231,64 +284,37 @@ export const DimensionSlicesRunner: FC<DimensionSlicesRunnerProps> = ({
                 </div>
               </div>
             ) : null}
+            <div className="flex-1" />
```
msamper, line 287: Is this needed? Seems like it might not be doing anything since its self closing

msamper (commented):
Left some comments. Looking into the bug you're trying to fix, but Bryce might be able to find a solution faster than me

packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
lukesonnet: done, ty.

packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
lukesonnet: done ty

packages/front-end/components/Settings/EditDataSource/DimensionMetadata/UpdateDimensionMetadata.tsx
lukesonnet: Yes, this is how fill space in a couple of places so that the "days to look back" div stays right aligned

packages/front-end/components/Experiment/TabbedPage/HealthTabOnboardingModal.tsx
lukesonnet: done ty

msamper (approved):

PR #1926, opened 2023-12-07
GitHub Integration - Authentication

bttf:
This PR implements the basic authentication flow for users who wish to integrate their GrowthBook instance with GitHub. The auth flow is just the first major piece of the integration, with webhook handlers and a code references UX to follow.

Note: This new feature is **feature flagged** on the frontend and will not be accessible on prod in its current state. Backend endpoints are not gated. The intention is to ship these set of changes once approved to deploy the feature incrementally and make the review process easier.

Changes:
- New models `GithubIntegration` and `GithubUserToken`
  - `GithubIntegration`
    - Maps one-to-one between an organization and a single github integration configuration object
    - Important fields
      - `tokenId` (FK for GithubUserToken)
      - `repositories` (array of simple objects representing a github repo)
  - `GithubUserToken`
    - A user-specific auth token (and refresh token) that is generated by Github upon authentication and is used by GB app to make subsequent requests to Github's API on behalf of the user
- New Github authentication UX flow
  - Can be initiated from our app (currently under Settings > Github) or from Github marketplace (not currently public)
    - **All routes are feature-gated**
  - Frontend structure for Github Integration page
    - New page under `/integrations/github`
    - Will conditionally render one of two components based on whether `GithubIntegration` for the org exists or not:
      - `GithubIntegrationConfig` when document exists
      - `GithubIntegrationConnect` when it does not
  - Auth steps are as follows:
    1. User installs Growthbook app via Github app installation page
    2. Redirected to Growthbook oauth callback URL
    3. `GithubUserToken` document created, creates `GithubIntegration` document (pairing token and organization)
    5. Upon creation, fetches repositories and presents UI for checking off which repositories should be 'watched' by Growthbook

## Screenshots
Before installing
Screenshot 2023-12-12 at 4 06 38 PM

Installation step
Screenshot 2023-12-12 at 4 06 50 PM

After installing - repos displayed
Screenshot 2023-12-12 at 4 05 46 PM


packages/back-end/src/routers/github-integration/github-integration.controller.ts
```diff
@@ -0,0 +1,98 @@
...
+import { AuthRequest } from "../../types/AuthRequest";
+import {
+  getGithubIntegrationByOrg,
+  createGithubIntegration,
+} from "../../models/GithubIntegration";
+import { createGithubUserToken } from "../../models/GithubUserTokenModel";
+import { APP_ORIGIN } from "../../util/secrets";
+
+const githubPrivateKey = fs.readFileSync(
+  process.env.GITHUB_PRIVATE_KEY_PATH || "",
```
bttf, line 14: We'll need to store the private key file (.pem) somewhere to read from it. This was taken from the examples provided by GitHub. It's possible we could just throw the entire contents of the file into an env var, just need to be wary about variable size constraints.

mknowlton89 (approved):

packages/back-end/src/models/GithubUserTokenModel.ts
```diff
@@ -0,0 +1,80 @@
...
+    updatedAt: new Date(),
+  });
+  return toInterface(doc);
+};
+
+export const doesTokenExist = async (tokenId: string) => {
+  return await GithubUserTokenModel.exists({
+    id: tokenId,
+    createdAt: {
+      $gte: new Date(new Date().getTime() - 1 * 60 * 1000), // 1 minute
```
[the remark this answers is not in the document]
bttf: Good q - this was sort of leftover from the previous implementation - it was meant to prevent using tokens that were generated a long time ago, which could have been a vulnerability but with this new approach it isn't. Will clean it up

packages/back-end/src/routers/github-integration/github-integration.controller.ts
jdorn: We don't have the ability to easily mount files with ECS, so we will need to store the entire contents in an env variable instead (or fetch from secrets manager etc. at runtime)

packages/back-end/src/models/GithubIntegration.ts
```diff
@@ -0,0 +1,96 @@
...
+} from "../../types/github";
+import { OrganizationInterface } from "../../types/organization";
+import { fetchRepositories } from "../services/github";
+import { doesTokenExist } from "./GithubUserTokenModel";
+
+type GithubIntegrationDocument = mongoose.Document & GithubIntegrationInterface;
+
+const githubIntegrationSchema = new mongoose.Schema({
+  id: String,
+  organization: String,
```
jdorn, line 16: Should this be project scoped?  Thinking about an org with 2 projects - `app1` and `app2`.  If they are changing a feature in `app1`, there's no need to scan the GitHub repo for `app2`

packages/back-end/src/models/GithubUserTokenModel.ts
```diff
@@ -0,0 +1,77 @@
...
+import { refreshToken } from "@octokit/oauth-methods";
+import {
+  GithubUserTokenInterface,
+  CreateGithubUserTokenInput,
+} from "../../types/github";
+
+type GithubUserTokenDocument = mongoose.Document & GithubUserTokenInterface;
+
+const githubUserTokenSchema = new mongoose.Schema({
+  id: String,
```
jdorn, line 13: We should have an `organization` field for every data model and use it in all queries.  It's the safest way to prevent critical cross-account access bugs.

packages/back-end/src/models/GithubIntegration.ts
bttf: Conceptually I don't think an **integration** should be project scoped - an org will have only one integration (its only set up once). When we start to implement the code references part of this feature, I think we'd probably introduce a new data model such as `GithubRepo` that can be project scoped and handle storing data related to code references

PR #1938, opened 2023-12-12
Add CTA for users with 2+ exposure queries

lukesonnet:
Adds CTA for when health tab == on but no experiment dimensions are configured:
* if they "skipped" during setup
* if they are on a health page for an experiment that is not using the same exposure query as the one they went through setup with

loom: https://www.loom.com/share/5ef2ea800b39405a854ff8ae0fd275f0

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
@@ -11,8 +11,12 @@ import TrafficCard from "@/components/HealthTab/TrafficCard";
 import { IssueTags, IssueValue } from "@/components/HealthTab/IssueTags";
 import LoadingSpinner from "@/components/LoadingSpinner";
 import { useDefinitions } from "@/services/DefinitionsContext";
+import track from "@/services/track";
 import { useSnapshot } from "../SnapshotProvider";
-import { HealthTabOnboardingModal } from "./HealthTabOnboardingModal";
+import {
+  HealthTabConfigParams,
+  HealthTabOnboardingModal,
+} from "./HealthTabOnboardingModal";
 
 export interface Props {
   experiment: ExperimentInterfaceStringDates;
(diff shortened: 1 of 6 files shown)


packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
```diff
@@ -49,6 +58,17 @@ export default function HealthTab({
...
+  const healthTabConfigParams: HealthTabConfigParams = {
+    experiment: experiment,
+    phase: phase,
+    exposureQueryDimensions: exposureQuery?.dimensions || [],
+    refreshOrganization: refreshOrganization,
+    mutateSnapshot: mutateSnapshot,
+    setAnalysisSettings: setAnalysisSettings,
+    setLoading: setLoading,
+    resetResultsSettings: resetResultsSettings,
+  };
```
msamper, line 70: Nit: for object properties whose values have the same name as the value expected by the type you can omit the assignment portion (i.e. setLoading: setloading -> setLoading)
```suggestion
  const healthTabConfigParams: HealthTabConfigParams = {
    experiment,
    phase,
    exposureQueryDimensions: exposureQuery?.dimensions || [],
    refreshOrganization,
    mutateSnapshot,
    setAnalysisSettings,
    setLoading,
    resetResultsSettings,
  };
```

packages/front-end/components/HealthTab/DimensionIssues.tsx
```diff
@@ -274,11 +284,43 @@ export const DimensionIssues = ({ dimensionData, variations }: Props) => {
...
+                  </a>
+                </div>
+                {setupModalOpen ? (
+                  <HealthTabOnboardingModal
+                    open={setupModalOpen}
+                    close={() => setSetupModalOpen(false)}
+                    healthTabConfigParams={healthTabConfigParams}
+                    healthTabOnboardingPurpose={"dimensions"}
+                  />
+                ) : null}
```
msamper, line 320: You shouldn't need to do a ternary with `setupModalOpen` since the `HealthTabOnboardingModal` has access to the `open` state and should hide if it's `false`

lukesonnet: I think leaving it in forces it to re-render if you click back into the modal which is good because it resets the modal state. I removed it on the main health onboarding CTA to test and it would instead remember the last step of the modal (which would be the exit modal), so I think we should leave the ternary.

msamper (approved):

PR #1946, opened 2023-12-13
Fix exposure query error in health onboarding modal

lukesonnet:
### Features and Changes

Move exposure query and datasource checks up a level. If you don't have that data in the experiment (a state you can get in to quite often) then the health tab setup and state handling isn't great, then running a health query is impossible so we should be prompting you to set it up here when your problems exist elsewhere

Why?
* More descriptive in the edge cases where this data is missing

How?
* Just handle the states in the parent component that loads the onboarding/dimension slices modal.
* Add messaging in the HealthTab to direct people to fix the problem. This language isn't great (missing exposure queries), but the root cause beyond the scope of this PR.

### Testing/Screenshots

Tested main health tab happy paths, and also when exposure query is missing (for this test, see screenshots below).

Issue before:
https://www.loom.com/share/9a307b250c3444de92fe91d20c9b4e54

After:
Screenshot 2023-12-18 at 9 52 54 AM
Then if you add the exposure query you can tab back to the health tab and set up

After, we also handle the case where your health is set up but you're missing an exposure query:
Screenshot 2023-12-18 at 9 52 43 AM


bryce-fitzsimons (approved):

packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
```diff
@@ -86,12 +85,24 @@ export default function HealthTab({
     [onDrawerNotify]
   );
 
-  // If org has not updated settings since the health tab was introduced, prompt the user
-  // to enable the traffic query setting
+  const noExposureQueryMessage = `The health tab only works when your experiment has an Exposure
```
msamper, line 88: nit: If this string is always the same for `noExposureQueryMessage` pull it out outside of the component (above line 28) You'll also want to use `"` instead of `` ` ``

msamper (approved):

PR #1948, opened 2023-12-13
Replace Papercups with Pylon

mknowlton89:
### Features and Changes

This PR updates our In-App chat widget to use the Pylon chat widget, rather than the Papercups widget. Accompanied by this, is a new Feature Flag in our GrowthBook org for `pylon-config`.

This PR doesn't change any of the logic of who gets this in-app chat widget, vs the other help widget.

### Screenshots
Screen Shot 2023-12-13 at 2 28 29 PM

Screen Shot 2023-12-14 at 6 37 08 AM

Screen Shot 2023-12-13 at 2 28 46 PM

### DIFF
--- a/packages/front-end/components/Auth/InAppHelp.tsx
+++ b/packages/front-end/components/Auth/InAppHelp.tsx
@@ -8,30 +8,29 @@ import { GBPremiumBadge } from "../Icons";
 import UpgradeModal from "../Settings/UpgradeModal";
 
 export default function InAppHelp() {
-  const { app_id, script_content } = useFeature("pylon-config").value;
+  const config = useFeature("pylon-config").value;
   const [showFreeHelpWidget, setShowFreeHelpWidget] = useState(false);
   const [upgradeModal, setUpgradeModal] = useState(false);
   const { name, email, hasCommercialFeature } = useUser();
-  const showUpgradeModal =
-    app_id && script_content && !hasCommercialFeature("livechat") && isCloud();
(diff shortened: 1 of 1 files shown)


august-growthbook:
Looks great! Do you know why the icon for the docs isn't showing up in the widget in the 2nd screenshot?

mknowlton89:
> Looks great! Do you know why the icon for the docs isn't showing up in the widget in the 2nd screenshot?

I don't - I looked through the Pylon app, and I see where I can set an icon for the Docs link, but I don't see how I can add a URL for an image. You can update the icon [here](https://app.usepylon.com/settings/bookmarks).

For now, I just selected a stack of books as a placeholder. You're welcome to change it via the link above.

Screen Shot 2023-12-14 at 6 37 08 AM

packages/front-end/components/Auth/InAppHelp.tsx
```diff
@@ -8,36 +8,33 @@ import { GBPremiumBadge } from "../Icons";
 import UpgradeModal from "../Settings/UpgradeModal";
 
 export default function InAppHelp() {
-  const config = useFeature("papercups-config").value;
+  const { app_id, script_content } = useFeature("pylon-config").value;
```
jdorn, line 11: The feature value will be `null` if there's a network problem or the feature is toggled off for some reason.  That will throw an exception here since you are using object destructuring on it.

mknowlton89: Ah, thank you!

jdorn (approved):

PR #1949, opened 2023-12-13
Fix UI for multiple exposures setting

lukesonnet:
Before: a setting that ranges from 0-1 was called a percent, when really it's a proportion that needs to be * 100 to be a percent
Screenshot 2023-12-13 at 2 01 48 PM

After: just be more explicit about what they're setting
image

It'd be great if we could just make the field a percent but I don't want to mess with all our data models and field names for this.


bryce-fitzsimons:
Why not *100 on the display and /100 when submitting? Kinda like how we do with metric risk thresholds
image

PR #1954, opened 2023-12-14
add existence check for settings

tzjames:
### Features and Changes

Stops an error being thrown if settings doesn't exist on a datasource.  In general that shouldn't happen as it is a required field, but it doesn't hurt being defensive here as this is not used for any core functionality.

### Testing

manually delete the settings from a datasource.
start the app and see it load.


mknowlton89 (approved):
LGTM - I tested in my previously broken env and this resolved the issue. I then removed all of the offending datasources and tested again and confirmed it's working as expected.

I honestly don't recall how I had datasources object without a `settings` property. Its very possible I manually did it while testing something way back when. So, ultimately up to you if we merge this in. Since it's a single character change that doesn't add any complexity, but potentially blocks an exception, it doesn't hurt to add it.

Issue #1812, opened 2023-11-01
[Feature] Specify the default assignment query


jdorn:
### Description of Feature

In Organization settings, we let you specify the default data source used for new experiments.  We should also let you specify the default assignment query.  Alternatively, on the data source page, add the ability to specify one of the assignment queries as the default.

alyssamichelle:
**Our audit of stale tickets suggests closing this as superseded.** @ahdriel assigning to you for a decision — does that match your read?

**Why:** No org-level 'default assignment query' setting or per-datasource default flag exists (searched for defaultExposureQuery/defaultAssignmentQuery in back-end types and front-end). Instead, Experiment Templates (PR #3442, Jan 2025; docs/running-experiments/experiment-templates.mdx; validator packages/shared/src/validators/experiment-template.ts includes datasource + exposureQueryId) let orgs pre-set the data source and assignment query for new experiments and can be enforced. Additionally packages/front-end/components/Experiment/NewExperimentForm.tsx getDefaultDatasourceAndExposureQuery() auto-picks the exposure query whose identifier type matches the experiment's hash attribute.

If you agree, please close with a short note for the reporter — our suggestion: Experiment Templates cover setting a default assignment query (and much more); also the new-experiment form now auto-selects the assignment query matching the hash attribute.

If it stays open instead, please update the labels to reflect current state (add `roadmap` if it's planned, keep `bug`/`enhancement` accurate, drop anything stale like `2024-tracked`) so the next sweep doesn't flag it again.

<sub>Part of a sweep of 223 long-quiet issues audited against `main` (Sept 2026). Ping @alyssamichelle if the evidence looks wrong.</sub>

Issue #1943, opened 2023-12-12
Experiment from Feature Flags should pick an Assignment Query based on hash attribute (or ask for one to be set)


lukesonnet:
### Summary

When you create an experiment rule in a feature flag, it spins up a linked Experiment, but it comes with no selected experiment assignment query. So every time you do this you need to select the experiment assignment query that maps to the `attribute` you used for hashing in the feature.

We should consider one of two options:

1. An improved UX to require users to select this assignment table when the experiment is created, or at least give them a better error warning when it is unset
2. 
   a. Have some mapping between hash attributes in features <-> identifier types
   b. Automatically use that mapping to set the exposure query in the experiment (or, if no mapping, prompt experimenters to pick an exposure query)

Specifically, in here: https://github.com/growthbook/growthbook/blob/27198a9fb28b5999bff7c44bdb4d94f3c1b200f5/packages/front-end/components/Features/RuleModal.tsx#L296-L297 
We should pass down the `values.hashAttribute` and hopefully have the experiment assignment queries know which `hashAttribute` values map to them so we can easily set this when the experiment is created.

lukesonnet:
The above issue #2105 has a good example of how the status quo manifests in unclear errors.

I've also written more about this in a community PR [here](https://github.com/growthbook/growthbook/pull/1962#pullrequestreview-1816300423), copied below:

> First of all, thanks a bunch for trying to close some issues on this project! Unfortunately, I have some qualms about setting a default assignment query. I think it will exacerbate some issues in our UX.
> 
> The root of the issue: The Experiment Assignment Table dropdown in the experiment form must be set by users, but it is somewhat hidden and not always clear what should be selected (users have to know: what identifier type and hash attribute does this table map to?). Setting it incorrectly will silently cause "no data found" errors when people run queries for a lot of people, or worse, it could coincidentally match IDs and lead to silent errors.
> 
> Therefore, I think we shouldn't be setting a default experiment assignment query unless we slightly retool the new experiment form to:
> * have a separate modal page for datasource and modal, like "Data Settings". so that you always have to "accept" this default in some kind of simple Modal that only has (datasource and experiment assignment query) as the options. This way, if we do want defaults, at least we're putting it in front of people's eyes. (edge case: if people want to start an experiment but don't want to set this, we can let them skip it with a warning that they'll have to set them later in "analysis settings")
[...]

PR #1962, opened 2023-12-18
Add Default Assignment Query for Datasources 

smrdotgg:
### Features and Changes

 - Added a `default` tag that will be used to show that datasources and assignment queries as set to default.
 - Add a "Set as default" button in the three-dot menu attached to assignment queries on the parent datasource page.
 - Add an optional `defaultQueryId` on the `settings` object of each datasource. This will be used to mark the default Asssignment Query, if set.

- Closes [#1812](https://github.com/growthbook/growthbook/issues/1812)

#### Test 1 (Testing new default UI flag)
1. Set a datasource as default.
2. Open the datasource page. 

 - There should be a teal-blue `default` tag at the top of the page.

#### Test 2 (Testing default assignment query)
1. Set a datasource as default.
2. Open the datasource page. 
3. Scroll to the "Assignment Queries" section.
4. Click the three-dot menu, and click "Set as Default".

 - A teal-blue `default` tag should show next to the Assignment Query name.
 - Create a new Experiment. The new experiment should use the default datasource along with the default assignment query, by default.

#### Default Datasource Tag
Default Datasource Tag
#### Default Assignment Query Tag
Default Assignment Query Tag


packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -92,11 +92,13 @@ export function getNewExperimentDatasourceDefaults(
   return {
     datasource: initialDatasource.id,
     exposureQueryId:
-      getExposureQuery(
+      initialDatasource.settings.defaultQueryId ??
```
tzjames, line 95: The way you have this now it will always use the default query, even if there is an initialValue set.  Are you sure that is the behavior that we need?

tzjames (changes requested):

packages/front-end/pages/datasources/[did].tsx
```diff
@@ -160,8 +160,15 @@ const DataSourcePage: FC = () => {
           <h1 className="mb-0">{d.name}</h1>
         </div>
         <div className="col-auto">
-          <span className="badge badge-secondary">{d.type}</span>{" "}
-          <span className="badge badge-success">connected</span>
+          <span className="badge badge-secondary">
```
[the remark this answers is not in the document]
Comment (smrdotgg) [reply packages/front-end/pages/datasources/[did].tsx:163]: The `default` tag on the experiment query was capital, so I then made the `default` tag for the data source capital, which then led me to capitalizing all the other tags.

We can keep as before, and switch the only tag on the experiment queries (`default`) to small letters.

packages/front-end/components/Experiment/NewExperimentForm.tsx
smrdotgg: You're right, this does prioritize the default ID over the current ID for when a query is being edited. 

I'll fix that.

smrdotgg: I've made the fix, so as it stands now, we call `getExposureQuery`, and if we pass in an `initialValue`, it will use that. (This would be the case for when the form is used for editing instead of creating). 
If the `initialValue` is `null`, it will use the `defaultId` if set, and will finally fallback to the default fallback that was always there if the defaultId has not been set.

smrdotgg: I've reverted to the original, simpler span display method. :+1:

lukesonnet (changes requested):
First of all, thanks a bunch for trying to close some issues on this project! Unfortunately, I have some qualms about setting a default assignment query. I think it will exacerbate some issues in our UX.

The root of the issue: The Experiment Assignment Table dropdown in the experiment form must be set by users, but it is somewhat hidden and not always clear what should be selected (users have to know: what identifier type and hash attribute does this table map to?). Setting it incorrectly will silently cause "no data found" errors when people run queries for a lot of people, or worse, it could coincidentally match IDs and lead to silent errors.

Therefore, I think we shouldn't be setting a default experiment assignment query unless we slightly retool the new experiment form to:
* have a separate modal page for datasource and modal, like "Data Settings". so that you always have to "accept" this default in some kind of simple Modal that only has (datasource and experiment assignment query) as the options. This way, if we do want defaults, at least we're putting it in front of people's eyes. (edge case: if people want to start an experiment but don't want to set this, we can let them skip it with a warning that they'll have to set them later in "analysis settings")
* furthermore, I think it'd be good to make a hash attribute <-> exposure query mapping https://github.com/growthbook/growthbook/issues/1943 that could easily set this default query in the modal proposed above to make it so it's more of a confirmation step rather than a "oh wait which id did I use again?" step.

Anyways, I think these are out of scope for this PR, but I'm not sure this PR is heading in the right direction as it may exacerbate an existing issue.

PR #1977, opened 2023-12-21
Allow users to refresh license data from server on demand

tzjames:
### Features and Changes
Allows users to refresh license data from server on demand.  This might be needed if they recently contacted sales and we converted their trial to a real license, or updated the number of seats.

- Added argument to force refresh license
- Added test to make sure that worked
- Added backend endpoint for refreshing the license
- Moved license data to its own component
- Added RefreshLicenseButton component
- Added ShowLicenseInfo component to admin page (for multi-org)

### Testing

Test non-multi-org:

1. Set IS_MULTI_ORG=false in frontend and backend .env.local
2. Go to general settings page
3. Click on the Upgrade button and see the modal.
4. Click on the Edit license and see the edit license modal
5. Enter in a valid license.
6. See the license data.
7. Change the license data from the central license server
8. Click update
9. See the updated license data on the settings page

Test multi-org:

1. Set IS_MULTI_ORG=true in front end and backend .env.local
2. Remove the licenseKey from the org in mongo.
3. Add the licenseKey to the .env.local on the back end
4. Go to the admin page.
5. See the license data.
6. Change the license data from the central license server
7. Click update
8. See the updated license data on the admin page

### Screenshots
Screenshot 2023-12-21 at 2 28 47 PM
Screenshot 2023-12-21 at 2 27 31 PM

### DIFF
--- a/packages/front-end/components/License/ShowLicenseInfo.tsx
+++ b/packages/front-end/components/License/ShowLicenseInfo.tsx
@@ -1,0 +1,141 @@
+import { FC, useState } from "react";
+import { FaPencilAlt } from "react-icons/fa";
+import { date } from "shared/dates";
+import usePermissions from "@/hooks/usePermissions";
+import { useUser } from "@/services/UserContext";
+import { isCloud } from "@/services/env";
+import EditLicenseModal from "../Settings/EditLicenseModal";
+import { GBPremiumBadge } from "../Icons";
+import UpgradeModal from "../Settings/UpgradeModal";
+import RefreshLicenseButton from "./RefreshLicenseButton";
+
+const ShowLicenseInfo: FC<{
+  showInput?: boolean;
+}> = ({ showInput = true }) => {
+  const { accountPlan, license, refreshOrganization } = useUser();
+  const permissions = usePermissions();
+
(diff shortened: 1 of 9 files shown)


packages/front-end/components/License/ShowLicenseInfo.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    </div>
+                    <div className="col-sm-2">
+                      <div>Expires:</div>
+                      <span className="text-muted">
+                        {date(license.dateExpires)}
+                      </span>
+                    </div>
+                    <div className="col-sm-2">
+                      <div>Seats:</div>
+                      <span className="text-muted">{license.seats}</span>
```
mknowlton89, line 125: Small nit: Given that we default all trial license keys generated via the app to 999 seats. I wonder if we want to add a tooltip here that adds some context about where 999 came from and how when the trial ends they won't necessarily be billed for 999 seats.

I suppose we'd only want to show the tooltip for orgs in a trial with 999 seats, since some orgs will start a trial via sales that will generate the license key for them and define a pre-determined amount of seats.

mknowlton89 (approved):

PR #1984, opened 2023-12-22
lock calls to the license server

tzjames:
### Features and Changes

When visiting a front-end page for the first time, we often make 3-6 simultaneous calls to the backend for /organization, /user, /organization/definitions etc.  Each of them had been firing off a call to the central-license-server in immediate succession.  By adding a lock it will only call it once and the remaining times it will use the cache. 

### Testing
`yarn test` -> see it pass
`yarn dev` -> hit a url and see the "Getting license data from server for " info line only once.


mknowlton89 (approved):
LGTM - Tested and it worked as intended.

PR #1993, opened 2024-01-03
Prerequisite flags UI

bryce-fitzsimons:
### Features and Changes

Adds prerequisite flag support:
- top level prerequisites (function similar to kill switches)
- rule-based prerequisite targeting in both features and experiments
- payload scrubbing and sdk version safety
- feature evaluation / archetypes support


packages/shared/src/util/features.ts
```diff
@@ -468,3 +468,207 @@ export function validateAndFixCondition(
...
+        // todo: consider returning info about missing feature
+        state = "deterministic";
+        value = null;
+        break;
+      }
+      const { state: prerequisiteState, value: prerequisiteValue } = visit(
+        prerequisiteFeature
+      );
+      if (prerequisiteState === "deterministic" && prerequisiteValue === null) {
+        // "off" by null. Any "off" prereq state overrides feature's default state (#2)
```
jdorn, line 624: Is this true in the general case?  What if you have a prerequisite on a rule with the condition: `{ value: {$exists: false} }`.  Then, the prereq being deterministic null should NOT override since the condition passes.

packages/shared/src/util/features.ts
```diff
@@ -468,3 +468,207 @@ export function validateAndFixCondition(
...
+        state = "deterministic";
+        value = null;
+        break;
+      } else if (
+        prerequisiteState === "deterministic" &&
+        prerequisiteFeature.valueType === "boolean" &&
+        prerequisiteValue === "false"
+      ) {
```
jdorn, line 632: Same here.  I think we need to actually evaluate the condition here, at least in a lightweight way.

packages/shared/src/util/features.ts
```diff
@@ -468,3 +468,207 @@ export function validateAndFixCondition(
...
+}
+
+export function isPrerequisiteConditionConditional(
+  condition: string,
+  prerequisiteState: PrerequisiteState
+) {
+  if (
+    [`{"value": {"$exists": false}}`, `{"value": {"$exists": true}}`].includes(
+      condition
+    )
```
jdorn, line 489: We should ideally be matching on the parsed condition, not the stringified version.  Otherwise, it's super fragile when it comes to whitespace.

packages/back-end/src/services/features.ts
```diff
@@ -1047,3 +1151,255 @@ export const updateInterfaceEnvSettingsFromApiEnvSettings = (
...
+        case "conditional":
+          // keep the feature and the prerequisite
+          newPrerequisites.push(prereq);
+          break;
+        case "cyclic":
+          removeFeature = true;
+          break;
+        case "deterministic":
+          // todo: this makes the assumption that top-level prereqs are always boolean / true checks
+          if (state.value === null || state.value === "false") {
```
jdorn, line 1204: This assumption does not seem safe.  I think we need the same logic as rules where we look at the specific condition and determine if it always passes, always fails, or is conditional.

packages/back-end/src/services/features.ts
```diff
@@ -1047,3 +1151,255 @@ export const updateInterfaceEnvSettingsFromApiEnvSettings = (
...
+
+    switch (state.state) {
+      case "conditional":
+        // keep the rule and prerequisite
+        break;
+      case "cyclic":
+        // remove the rule
+        removeRule = true;
+        continue;
+      case "deterministic":
```
jdorn, line 1334: The logic in this case seems like it should be its own function:

```ts
// Given a condition and value, determine if it always passes, always fails, or is conditional
evalPrereqDeterministicValue(condition, value): "pass" | "fail" | "conditional";
```

Would be easier to unit test and make the code easier to read.

packages/back-end/src/services/features.ts
```diff
@@ -1047,3 +1151,255 @@ export const updateInterfaceEnvSettingsFromApiEnvSettings = (
...
+        if (state.value === null) {
+          // try to reduce the rule or feature
+          if (pc.condition === `{"value": {"$exists": false}}`) {
```
jdorn, line 1337: Again, this is extremely fragile when it comes to whitespace.  Ideally we should parse the condition.  I'm ok with this for MVP though

packages/back-end/src/services/features.ts
```diff
@@ -1047,3 +1151,255 @@ export const updateInterfaceEnvSettingsFromApiEnvSettings = (
...
+    }
+
+    switch (state.state) {
+      case "conditional":
+        // keep the rule and prerequisite
+        break;
+      case "cyclic":
+        // remove the rule
+        removeRule = true;
+        continue;
```
jdorn, line 1333: Is there any point in continuing here?  If we know we're going to remove the rule, we should be able to immediately return.

packages/shared/src/util/features.ts
bryce-fitzsimons: this is only true for the current implementation of top-level prereqs (boolean, is true)

packages/shared/src/util/features.ts
bryce-fitzsimons: true if we relax the constraint that top level prereqs can be general. do we want to do that now?

packages/back-end/src/services/features.ts
bryce-fitzsimons: agree, should be `break`

bryce-fitzsimons: actually i believe this is correct as-is. `continue` means stop evaluating the for loop of prereqs and jump ahead to removing the rule

packages/shared/src/util/features.ts
bryce-fitzsimons: replaced with evalCondition()

packages/shared/src/util/features.ts
bryce-fitzsimons: replaced with evalCondition()

packages/back-end/src/services/features.ts
bryce-fitzsimons: replaced with evalCondition()

packages/back-end/src/services/features.ts
bryce-fitzsimons: implemented using evalCondition(). We can handle all cases now, not just simple.

packages/back-end/src/services/features.ts
bryce-fitzsimons: replaced with evalCondition()

packages/shared/src/util/features.ts
bryce-fitzsimons: replaced with evalCondition()

packages/shared/src/util/features.ts
```diff
@@ -468,3 +469,217 @@ export function validateAndFixCondition(
...
+      const { state: prerequisiteState, value: prerequisiteValue } = visit(
+        prerequisiteFeature
+      );
+      if (prerequisiteState === "deterministic") {
+        const evaled = evalDeterministicPrereqValue(
+          prerequisiteValue ?? null,
+          prerequisite.condition
+        );
+        if (evaled === "fail") {
+          value = null;
```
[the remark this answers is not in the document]
bryce-fitzsimons: missed a `break`. will fix

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -47,6 +46,36 @@ export const scrubFeatures = (
...
+  //   see: reduceFeaturesWithPrerequisites()
+  if (!capabilities.includes("prerequisites")) {
+    for (const k in features) {
+      // delete feature
+      if (
+        features[k]?.rules?.some((rule) =>
+          rule?.parentConditions?.some((pc) => !!pc.gate)
+        )
+      ) {
+        delete features[k];
```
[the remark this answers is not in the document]
bryce-fitzsimons: missed this, thanks

packages/shared/src/sdk-versioning/index.ts
```diff
@@ -172,17 +172,36 @@ export const getConnectionSDKCapabilities = (
...
+  for (let i = 0; i < filteredConnections.length; i++) {
+    const connection = filteredConnections[i];
+    const connectionCapabilities = getConnectionSDKCapabilities(
+      connection,
+      strategy
     );
+    if (!mustMatchAllConnections || i === 0) {
+      capabilities = capabilities.concat(connectionCapabilities);
+    }
+    if (mustMatchAllConnections && i > 0) {
```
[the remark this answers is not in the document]
bryce-fitzsimons: yep! thanks

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -47,6 +46,36 @@ export const scrubFeatures = (
...
+        continue;
+      }
+      // delete rules
+      features[k].rules = features[k].rules?.filter(
```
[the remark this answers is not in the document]
bryce-fitzsimons: even after hoisting `features = { ...features }` up?

packages/back-end/src/services/features.ts
```diff
@@ -589,14 +657,47 @@ export function evaluateFeature({
...
+          definition.rules = definition.rules
+            ? (definition?.rules
+                ?.map((rule) => {
+                  if (rule?.parentConditions?.length) {
+                    rulesWithPrereqs.push(rule);
+                    if (rule.parentConditions.some((pc) => !!pc.gate)) {
+                      return null;
+                    }
+                    if (skipRulesWithPrerequisites) {
+                      delete rule.force; // make rule invalid so it is skipped
```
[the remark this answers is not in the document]
bryce-fitzsimons: good catch. will `delete rule.variations` as well

packages/back-end/src/services/features.ts
```diff
@@ -1047,3 +1148,199 @@ export const updateInterfaceEnvSettingsFromApiEnvSettings = (
...
+
+  // block "always off" rules, or reduce "always on" rules
+  for (let i = 0; i < newFeatures.length; i++) {
+    const feature = newFeatures[i];
+
+    const newFeatureRules: FeatureRule[] = [];
+
+    for (
+      let i = 0;
+      i < feature.environmentSettings[environment].rules.length;
```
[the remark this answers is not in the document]
bryce-fitzsimons: good catch, will add early exit above this if it doesn't exist

packages/shared/src/sdk-versioning/sdk-payload.ts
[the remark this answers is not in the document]
jdorn: yes, doing `{...features}` is a shallow clone

jdorn (approved):

PR #1997, opened 2024-01-03
Introduce new no access role

mknowlton89:
### Features and Changes

This PR introduces a new role called `no access`. This is now the most limited role. In order to support this, we had to add a new permission called `readData` that all roles, excluding the `no access` role receives. 

Previously, the `readOnly` role was our least permissive role, and it didn't have any permissions. Now that we've added a less permissive role (`no access`) we had to make the `readData` permission.

This new `no access` role is an enterprise level feature, so it does require the organization to have the `advanced-permissions` commercial feature.

This allows us to not show the `no access` role option in the `RoleSelector` when inviting/updating a user's global or project-level roles. 

### NOTE: This is the first of many PRs that will be introduced to formally support the no-access role. The goal of this PR is to simply introduce the concept in our application. Pushing this PR will not change any organizations or user's experience.

I've created a new feature flag `no-access-role-type` that ensures even if an org has the `advanced-permissions` commercial feature, they won't be able to add/invite users with this new role in production.

### Testing
- [x] Ensure that if the organization is not enterprise, they do not see the `no access` role regardless of the feature flag's state.
- [x] Ensure that if the organization IS an enterprise org, and the feature flag is enabled for the environment, the `no access` role is an option when inviting or updating a user.

### DIFF
--- a/packages/back-end/src/scim/users/createUser.ts
+++ b/packages/back-end/src/scim/users/createUser.ts
@@ -14,6 +14,7 @@ import {
 
 export function isRoleValid(role: MemberRole) {
   const validRoles: Record<MemberRole, boolean> = {
+    noaccess: true,
     readonly: true,
     collaborator: true,
     designer: true,
--- a/packages/back-end/src/util/organization.util.ts
+++ b/packages/back-end/src/util/organization.util.ts
@@ -20,6 +20,7 @@ export const ENV_SCOPED_PERMISSIONS = [
 ] as const;
 
 export const PROJECT_SCOPED_PERMISSIONS = [
+  "readData",
   "addComments",
   "createFeatureDrafts",
   "manageFeatures",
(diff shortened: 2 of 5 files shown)


packages/back-end/src/scim/users/createUser.ts
```diff
@@ -14,6 +14,7 @@ import {
 
 export function isRoleValid(role: MemberRole) {
   const validRoles: Record<MemberRole, boolean> = {
+    noaccess: true,
```
mknowlton89, line 17: I'm torn here - I could either update the type to OMIT `noaccess` role, but I'd need to make a reminder to myself to remove the omission after this is all live. Or, we can just add it in. Currently, only a single organization is using this, and they are not the org interested in the `no access` role, so it's likely fine to just include it.

mknowlton89: If we keep it as-is, technically, someone could hit this scim endpoint and create a user with the noaccess role before we've built out all of the logic.

bttf: I think that's fine given how unlikely it is. And if it becomes an issue for an org, we can help them out of it ad-hoc

bttf (approved):

PR #1998, opened 2024-01-03
Introduce read access filter & update apiKey DB Model

mknowlton89:
### Features and Changes

This PR is the second of many PRs needed to support the new `no access` role type. This PR is dependent on [PR 1997](https://github.com/growthbook/growthbook/pull/1997/files#diff-eafc191e26bd4d271ef5c8179bbbfe0a3e993ef61b3cc22587dd0239016385db). And the branched is currently based off `mk/add-no-access-role-type`, not main.

This PR introduces a few new concepts.

Mostly significantly, it introduces a new object called `readAccessFilter`. This object has the following shape:

```ts
type ReadAccessFilter = {
  globalReadAccess: boolean;
  projects: { id: string; readAccess: boolean }[];
};
```

This new object is being built on each API request (currently only on our internal API, but in a separate PR, we'll do the same for the public rest API) and added to the request object.

With this flow, we can then access the `readAccessFilter` in each controller.

Additionally, now that we've introduced the `readAccessFilter` and added it to the request object for each API request, I've updated the `ApiKeyModel` to actually use the `readAccessFilter` as an example for how we'll eventually want to update all of the other DB models. The `ApiKeyModel` is one of the more self-contained models, so it was a good candidate to show how we intend to consume the `readAccessFilter`.

Inside of the `ApiKeyModel` we've updated two methods: `getApiKeyByIdOrKey` and `getAllApiKeysByOrganization`. These models now take in the `readAccessFilter` and use that to ultimately filter the Api Keys we return to the user.

ApiKeys have taken a few different shapes over time at GrowthBook - we currently have the concept of `Api Keys` in the application, and those do not have projects. However, we used to store SDK Connections as Api Keys, and those do have projects and these all live in the `apiKeys` collection in Mongo.

Now, when we load the app, we call `getOrganization` in the organization controller and in that, we get all of the org's ApiKeys. Now, we filter the apiKeys so we only return keys the user should be able to view.

> [!NOTE]
> Let's say we have a user, we'll call him Ross. Ross is a part of an organization's GrowthBook account and has a globalRole of "collaborator" and no project-level roles. 
In Ross' case, when he logs into GrowthBook, he should see every Api Key/ Legacy SDK Endpoint, since his global role gives him access.
However, Ross's colleague Chandler who is a part of the same organization has different roles. His global role is `no access`, but he's helping on one project, so on that project, he has a `collaborator` role.
When Chandler logs into GrowthBook, he should only see Legacy SDK Endpoints that are in the project he has a `collaborator` role for. All of the Legacy SDK Endpoints that don't have a project, or associated with other projects shouldn't be visible when he visits `/sdks#legacy`. 

And lastly, this PR introduces updates to the permissions test suite to ensure we're building the `readAccessFilter` correctly.


packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -31,3 +31,77 @@ export function hasPermission(
...
+    projects: [],
+  };
+
+  Object.keys(userPermissions.projects).forEach((project) => {
+    readAccess.projects.push({
+      id: project,
+      readAccess:
+        userPermissions.projects[project].permissions.readData || false,
+    });
+  });
```
jdorn, line 52: Small code style recommendation - if you need both keys and values of an object, use `Object.entries()` instead of `Object.keys()`.

```suggestion
  Object.entries(userPermissions.projects).forEach(([project, projectPermissions]) => {
    readAccess.projects.push({
      id: project,
      readAccess: projectPermissions.permissions.readData || false,
    });
  });
```

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -31,3 +31,77 @@ export function hasPermission(
...
+
+  return readAccess;
+}
+
+export function hasReadAccess(
+  readAccessFilter: ReadAccessFilter,
+  resourceProjects: string[] // feature.project or experiment.project or metric.projects
+): boolean {
```
jdorn, line 60: This function is really hard to follow.  This much nesting and branching is a little suspicious. Does the logic need to be this complicated?

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -31,3 +31,77 @@ export function hasPermission(
...
+    // Check if user has project specific access for each project the resources is in
+    for (let i = 0; i < resourceProjects.length; i++) {
+      const projectAccessIndex = readAccessFilter.projects.findIndex(
+        (projectAccess) => projectAccess.id === resourceProjects[i]
+      );
+      if (projectAccessIndex === -1) {
+        userHasProjectSpecificAccessForEachResourceProject = false;
+        break;
+      }
+    }
```
jdorn, line 77: Make use of `.some()` and `.every()` for arrays.  Tends to make the code much more readable.

```ts
const userHasProjectSpecificAccessForEachResourceProject = resourceProjects.every((id) => {
  return readAccessFilter.projects.some(p => p.id === id);
})
```

packages/back-end/src/models/ApiKeyModel.ts
```diff
@@ -284,7 +286,14 @@ export async function getApiKeyByIdOrKey(
   const doc = await ApiKeyModel.findOne(
     id ? { organization, id } : { organization, key }
   );
-  return doc ? toInterface(doc) : null;
+
+  if (!doc) return null;
+
+  const apiKey = toInterface(doc);
+
+  return hasReadAccess(readAccessFilter, [apiKey.project || ""])
```
jdorn, line 294: Is an empty array treated the same as `[""]` in the read access logic?  I see a special case at the top of `hasReadAccess` to handle an empty array, but not an array of empty strings.  We should be consistent about how we specify "no projects"

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -31,3 +31,77 @@ export function hasPermission(
...
+    // if user has project specific role for each of the resource's projects, only allow readaccess if they have read access for every project the resource is in.
+    // e.g. metric.projects = ["project1", "project2", "project3"] and readAccessFilter.projects = [{id: "project1", readAccess: true}, {id: "project2", readAccess: true}, {id: "project3", readAccess: true}]
```
jdorn, line 79: Is this comment correct?

If a metric is tied to project A and B, I would think you only need access to at least one of the projects to read the metric.  This comment suggests you would need read access to both.

packages/back-end/src/models/ApiKeyModel.ts
mknowlton89: I've updated this to read:

```ts
 return keys.filter((k) =>
    hasReadAccess(readAccessFilter, k.project ? [k.project] : [])
  );
  ```
  
Technically, a `[""]` isn't treated the same as a `[]` - a `[""]` would pass the first `if` block.

I can add a special clause to the first block, something like: 
```ts
resourceProjects.every((project) => project === "")
``` 

We can ensure an early exit if a `[""]` is passed in.

Currently though, if a `[""]` is passed in, it'll just act as if it's a project not in the `readAccessFilter.projects` array.

For now, I've updated the `ApiKeyModel` and added the additional clause in the first block.

bttf (commented):
Great tests - Will review hasReadAccess after updates are made

bttf:
For `ApiKeyModel`, do we need to consider adding `hasReadAccess` filtering to the following methods:
- createLegacySdkKey
- getFirstPublishableApiKey

I'm wondering if in general we should apply hasReadAccess filtering to all model methods that may require it at some point (i.e. methods that could involve project scoping). Basically audit every model for methods that require 'noaccess' filtering.

packages/shared/src/permissions/permissions.utils.ts
mknowlton89: Just took a pass at updating this to remove a lot of the nesting. Still not super easy to follow, but better than the previous iteration.

packages/back-end/src/models/ApiKeyModel.ts
jdorn: I might recommend a slightly different approach.  Change the method signature of `hasReadAccess` to accept a wider range of values for project:

```ts
function hasReadAccess(filter: ReadAccessFilter, projects: string | string[] | null | undefined);
```

Because this is a utility function that will be called from tons of places, it will be much simpler if you can always just pass in `model.project` or `model.projects` without thinking about handling undefined, empty strings, converting to arrays, etc.

mknowlton89:
> For `ApiKeyModel`, do we need to consider adding `hasReadAccess` filtering to the following methods:
> 
> * createLegacySdkKey
> * getFirstPublishableApiKey
> 
> I'm wondering if in general we should apply hasReadAccess filtering to all model methods that may require it at some point (i.e. methods that could involve project scoping), and not just for methods that we need today.

Hmm... good question.

I think this is actually slightly different than filtering by read access - this should (I think) be handled by the existing `req.checkPermission` method. If a user makes a request and they don't have read access for a project, req.checkPermission should handle blocking the action well before it gets to the DB model.

I did notice re: `createLegacySdkKey` we weren't correctly passing in the project in the req.checkPermissions check within `postApiKey` in `organization.controller.ts`. I've updated this, so now, if the user doesn't have read access to a project, it will throw an error before well before `createLegacySdkKey` is called. 

With `getFirstPublishableApiKey` that is called on pages that are only visible if the user has `manageEnvironments` which requires an `engineer`, `experimenter`, or `admin` role. So if they had that role, they have `readAccess`. 

Let me know if you think we should be more explicit, but, that feels similar, by slightly adjacent to this ticket.

packages/back-end/src/models/ApiKeyModel.ts
mknowlton89: 10-4. I just updated the `hasReadAccess` method to the proposed signature and updated the function as needed. Please let me know if this is more in-line with what you were thinking.

The changes are in commit https://github.com/growthbook/growthbook/pull/1998/commits/ca832017c1f52b8c738fa2bbfd05783dfd2a61b4

bttf (approved):
I left a comment regarding readability of hasReadAccess. Otherwise looks good to me. My approval is contingent on Jeremy's approval.

bttf:
> > For `ApiKeyModel`, do we need to consider adding `hasReadAccess` filtering to the following methods:
> > 
> > * createLegacySdkKey
> > * getFirstPublishableApiKey
> > 
> > I'm wondering if in general we should apply hasReadAccess filtering to all model methods that may require it at some point (i.e. methods that could involve project scoping), and not just for methods that we need today.
> 
> Hmm... good question.
> 
> I think this is actually slightly different than filtering by read access - this should (I think) be handled by the existing `req.checkPermission` method. If a user makes a request and they don't have read access for a project, req.checkPermission should handle blocking the action well before it gets to the DB model.
> 
> I did notice re: `createLegacySdkKey` we weren't correctly passing in the project in the req.checkPermissions check within `postApiKey` in `organization.controller.ts`. I've updated this, so now, if the user doesn't have read access to a project, it will throw an error before well before `createLegacySdkKey` is called.
> 
> With `getFirstPublishableApiKey` that is called on pages that are only visible if the user has `manageEnvironments` which requires an `engineer`, `experimenter`, or `admin` role. So if they had that role, they have `readAccess`.
> 
> Let me know if you think we should be more explicit, but, that feels similar, by slightly adjacent to this ticket.



> > For `ApiKeyModel`, do we need to consider adding `hasReadAccess` filtering to the following methods:
> > 
> > * createLegacySdkKey
> > * getFirstPublishableApiKey
> > 
> > I'm wondering if in general we should apply hasReadAccess filtering to all model methods that may require it at some point (i.e. methods that could involve project scoping), and not just for methods that we need today.
> 
> Hmm... good question.
> 
> I think this is actually slightly different than filtering by read access - this should (I think) be handled by the existing `req.checkPermission` method. If a user makes a request and they don't have read access for a project, req.checkPermission should handle blocking the action well before it gets to the DB model.
> 
> I did notice re: `createLegacySdkKey` we weren't correctly passing in the project in the req.checkPermissions check within `postApiKey` in `organization.controller.ts`. I've updated this, so now, if the user doesn't have read access to a project, it will throw an error before well before `createLegacySdkKey` is called.
> 
> With `getFirstPublishableApiKey` that is called on pages that are only visible if the user has `manageEnvironments` which requires an `engineer`, `experimenter`, or `admin` role. So if they had that role, they have `readAccess`.
> 
> Let me know if you think we should be more explicit, but, that feels similar, by slightly adjacent to this ticket.

Thanks for explaining, that gives me a better idea of how this is to be used. I think this approach is fine.

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -31,3 +31,73 @@ export function hasPermission(
...
+  if (!userHasProjectSpecificAccessForEachResourceProject) {
+    // if there are resource projects that don't have a matching user permission, we default to global read permission, which effectively grants read access
+    return hasGlobaReadAccess;
+  }
+
+  const everyProjectRestrictsReadAccess = resourceProjects.every((project) => {
+    const projectAccessIndex = filter.projects.findIndex(
+      (projectAccess) => projectAccess.id === project
+    );
+    return filter.projects[projectAccessIndex].readAccess === false;
```
jdorn, line 99: Same here.  It's very rare that you will actually need `findIndex`.  Especially if you are then just going to use it to get the value at that index.

```ts
return filter.projects.some((p) => p.id === project && !p.readAccess)
```

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -31,3 +31,73 @@ export function hasPermission(
...
+  const userHasProjectSpecificAccessForEachResourceProject = resourceProjects.every(
+    (id) => {
+      return filter.projects.some((p) => p.id === id);
+    }
+  );
+
+  if (!userHasProjectSpecificAccessForEachResourceProject) {
+    // if there are resource projects that don't have a matching user permission, we default to global read permission, which effectively grants read access
+    return hasGlobaReadAccess;
+  }
```
jdorn, line 93: I don't think this check is needed.  I believe the `everyProjectRestrictsReadAccess` below covers this already.

PR #2000, opened 2024-01-04
Fallback to env var for LICENSE_KEY

tzjames:
### Features and Changes

Some people tried to use both an env var and the license key within mongo and couldn't understand why the env var wasn't "taking": https://growthbookapp.slack.com/archives/C03M46VMGTB/p1704321222440419
This change falls back to using the env var if the license key attached to the org has expired.

### Testing

`yarn test`
Added a key to licenseKey.
Restarted App.
Saw the licenseExpired in banner.
Added a LICENSE_KEY of a non-expired license.
Restarted App.
Saw the license in banner without expired message.


mknowlton89 (approved):
LGTM

PR #2009, opened 2024-01-09
Make sure enterprise package respects boolean like values in env vars

tzjames:
# Features and Changes

- `IS_CLOUD=false` and `IS_MULTI_ORG=false` was evaluated as `true` before this change.  This fixes that.
- In order to import things from shared in enterprise we needed to add a typing file for dirty-json like we have in back-end package.

# Test Plan
Set `IS_CLOUD=false` and `IS_MULTI_ORG=false` without a LICENSE_KEY in  packages/back-end/.env.local 
Start the server see "Try Enterprise" - which only shows on self-hosted in the menu.
`git grep process.env` manually see that all the boolean env vars have been wrapped.

### DIFF
--- a/packages/enterprise/package.json
+++ b/packages/enterprise/package.json
@@ -16,7 +16,9 @@
     "clean": "rimraf node_modules",
     "type-check": "tsc --pretty --noEmit"
   },
-  "dependencies": {},
+  "dependencies": {
+    "shared": "0.0.1"
+  },
   "devDependencies": {
     "@swc/cli": "^0.1.57",
     "@swc/core": "^1.3.4",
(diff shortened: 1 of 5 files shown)


packages/enterprise/package.json
```diff
@@ -16,7 +16,9 @@
     "clean": "rimraf node_modules",
     "type-check": "tsc --pretty --noEmit"
   },
-  "dependencies": {},
+  "dependencies": {
```
tzjames, line 19: So `yarn build` was working without this.  But `yarn build:deps` in top level was not working without it.  With the error 
```
enterprise | ../back-end/types/feature.d.ts(3,55): error TS2307: Cannot find module '@growthbook/growthbook' or its corresponding type declarations.
```
shared module calls back-end/types/features.d.ts which then imports from @growthbook/growthbook.

tzjames: Another way that works is to change yarn build:deps to build enterprise only after shared finishes: `"build:deps": "wsrun -p '@growthbook/growthbook' -p '@growthbook/growthbook-react' -p 'shared' -t -c build && wsrun -p 'enterprise' -c build",`

bttf (approved):

PR #2011, opened 2024-01-09
Replace jStat with stdlib

jdorn:
### Features and Changes

Replace the `jStat` library, which is unmaintained and untyped, with the `@stdlib` library.  The functions are a 1-to-1 swap.


packages/back-end/src/util/stats.ts
```diff
@@ -1,4 +1,4 @@
-import { jStat } from "jstat";
+import cdf from "@stdlib/stats/base/dists/chisquare/cdf";
```
lukesonnet: I think we should have the distribution name in the source code. Can you import `chisquare` instead and have it be `chisquare.cdf(...)`?

I also might even prefer the more verbose `random.base.normal()` for the data generating methods, but definitely for quantile and cdf methods we should have them prefixed by their distribution in the code.

lukesonnet (approved):
lgtm, stdlib docs look decent and it is reasonably active!

Issue #2020, opened 2024-01-10
Disallow using metrics with non-matched and non-joinable identifier types


Auz:

Auz:
Fixed in #2021

PR #2021, opened 2024-01-10
Filter out un-joinable metrics

lukesonnet:
Loom: https://www.loom.com/share/7a628585e6ef46ecb0aca7bf270c7239

Outstanding question: is there some other way we can teach people about needing to create join tables if they really want them?

Closes Issue #2020


packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -14,17 +15,63 @@ type MetricOption = {
   tags: string[];
   projects: string[];
   factTables: string[];
+  userIdTypes: string[];
 };
 
+export function MetricsSelectorTooltipBody(onlyBinomial: boolean = false) {
```
bryce-fitzsimons, line 21: should probably be a JSX component so that you can embed it via `<MetricsSelectorTooltipBody onlyBinomial={true} />`

lukesonnet: yeah, should do. thanks.

bryce-fitzsimons (approved):

PR #1006, opened 2023-02-17
"Softer" UX for invalid or expired license keys

bryce-fitzsimons:
### Features and Changes

We had an issue where a self-serve user using a trial license key had mismatched orgs (license versus GB instance). This was due to them having multiple GB instances running with different orgs and not being aware that a license key was for a specific org Id. The app was (intentionally) crashing due to the org mismatch, but that's not a great UX.

Instead, we will (a) push a notification into the navbar, and (b) revert their features to the unlicensed state (rather than crashing the app).

### Testing

Generate an invalid or expired license key and plug it into your GB instance. Either fake it, or ask me how to generate using the license generation script. You can enter the generated license key from the Settings page.

You should see on-page premium features get the standard paywall treatment, as well as a navbar notification.

### Screenshots

image

PR #2022, opened 2024-01-10
Allow remote downgrade and messages

tzjames:
### Features and Changes
Allow remote downgrade of licenses and show a message override from the license server.  We need this in case our customers are not paying up.  The override message will be used in the following cases: 		
- They aren't responding to our emails so we're using this to get their attention and tell them to reply to us
- We agreed to some super custom deal terms and the dynamic message language won't make sense

While implementing there were a few cases in the back-end which were relying on accountPlan which didn’t take into account whether there was an error with the license, so I changed that.  We weren't also downgrading due to mismatching orgId - so I centralized the check in a new `getEffectiveAccountPlan` function and made sure `orgHasPremiumFeature` calls it.  There were a few other small cleanups and improvements made.  Here is the full list:

- No longer export planHasPremiumFeature since it is not used externally of license.ts
- Created a new getLicenseErrorMessage function
- Created a new getEffectiveAccountPlan that takes into account if is expired, wrong org or remotely downgraded.
- Removed license middleware.  According to https://github.com/growthbook/growthbook/pull/1006/files#r1110309335 and git grep the header was not used, and git grep says that req.locals.licenseError was only used in one place in organization.controller.ts, so the middleware is no longer needed.
- Changed organization.controller.ts to check the license error directly itself.
- Changed orgHasPremiumFeature to use new getEffectiveAccountPlan for non-cloud.
- Have backend permissions check all use orgHasPremiumFeature instead
- Added new CommercialFeature for scim (it had been directly checking whether the plan was enterprise before)
- Attach effectiveLicensePlan to the userContext.
- If license has a message property use that to construct an AccountPlanNotice.
- Format the date in the expiry tooltip message of AccountPlanNotice
- Got rid of accountPlans as an option in SidebarLink as we weren’t using it in the one place we call SidebarLink: Layout.
- Stopped showing the badge if they no longer have the license features available due to a license error, in order to make it clear that they no longer have enterprise features.
- Refresh the organization when we get new license data as that has the commercial features available and the accountPlan - this allows us to see the Badge disappear without a page refresh.
- Show any AccountPlanNotice in the ShowLicenseInfo to further make it clear when there is an issue.
- Remove an out of date comment in Layout.tsx

### Testing

Insert the following into your shell with the correct authorization bearer and x-organization header taken by looking at a request in dev tools: 
```
postChecklist() {
curl -X POST \
  -H "Authorization: Bearer YOUR_BEARER_TOKEN" \
  -H "X-Organization: org_YOUR_ORG_ID" \
  -d '{
    "tasks": [
      {
        "name": "Task 1",
        "completed": false
      },
      {
        "name": "Task 2",
        "completed": true
      }
    ],
    "projectId": "your-project-id"
  }' \
  http://localhost:3100/experiments/launch-checklist
}
```

Add a valid trial enterprise license key to backend/.env.local
Go to Settings->General page.
See the correct license data.
See no account notice
See the badge.
Go to Settings->Logs page
See the export button is enabled.
See that scim works.
run `postChecklist` and see it work.

On the central license server:
Set the license expiry to be in the past
Go to Settings->General page.
Click refresh license.
See the correct license data.
See the expiry account notice
See the badge disappear (when trials end we revert them to free version functionality)
Go to Settings->Logs page
See the export button is disabled and there is a message saying it is only available in enterprise.
See that scim fails.
run `postChecklist` and see it fail.

On the central license server:
Change the license to no longer be a trial.
Go to Settings->General page.
Click refresh license.
See the correct license data.
See the expiry account notice
See the badge re-appear (when enterprise accounts have expired we assume it is an oversight and let them keep full enterprise functionality until we remotely downgrade them.)
Go to Settings->Logs page
See the export button is enabled. 
See that scim works.
run `postChecklist` and see it work.

On the central license server:
click the remotely downgrade button, and confirm (the AccountPlanNotice text is currently hardcoded, the tooltip isn't).
Go to Settings->General page.
Click refresh license.
See the correct license data.
See the hardcoded message, with tooltip.
See the badge disappear (to make clear they no longer have enterprise features).
Go to Settings->Logs page
See the export button is disabled and there is a message saying it is only available in enterprise.
See that scim fails.
run `postChecklist` and see it fail.

### Screenshots

Screenshot 2024-01-10 at 11 49 21 PM


packages/enterprise/src/license.ts
```diff
@@ -461,18 +473,55 @@ export function resetInMemoryLicenseCache(): void {
...
+
+  return false;
+}
+
+export function getEffectiveAccountPlan(org: MinimalOrganization): AccountPlan {
+  if (process.env.IS_CLOUD) {
+    return getAccountPlan(org);
+  }
+  const hasError = shouldLimitAccess(org.id);
+  const basicPlan = "oss";
```
mknowlton89, line 501: @james How quickly could someone's plan get into a state where `hasError === true`? In the past, we've been very lenient, especially with high paying enterprise accounts where if there is an issue, especially a billing issue, we don't disable their account functionality immediately, and instead allow them to continue operating and reach out to the org directly.

I think I'm seeing that we control when an org is remotely downgraded, as opposed to it being done programatically. Is that correct?

Just want to make sure we stay consistent (unless this has already been discussed).

CC: @jdorn

mknowlton89 (approved):
LGTM - Added one comment, but approving it for now, so if the comment is not a blocker, you can merge right away.

packages/enterprise/src/license.ts
tzjames: Correct.  hasError will be true if:

1. A trial license has expired
2. A license has the wrong orgId
3. A remotelyDowngraded flag gets set manually by someone from Retool.

Hence a paying customer with a working license, will continue to have all functionality until someone manually says enough is enough and turns them off.

msamper (approved):
Tested SCIM works under enterprise license and not Pro or expired enterprise trial

PR #2039, opened 2024-01-16
Trim license key & Better error message for license errors.

tzjames:
### Features and Changes
We weren't trimming the license keys, so copying the key with an extra space would fail.  We also had a generic "Invalid License Key" error which made it hard to determine what was wrong.  Now if the error is due to not being able to reach the license server then we tell them to make sure to whitelist the ip address.

### Testing

Change the LICENSE_SERVER to a made up address.
`yarn build` in enterprise package
Try and add a new license and see it say it could not connect and make sure to whitelist.

Change the LICENSE_SERVER  back.
`yarn build` in enterprise package
Try and add a bad new license and see it error saying "Invalid License Key"

Try and add a new license with extra spaces.
See it succeed.

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -1809,19 +1809,21 @@ export async function putLicenseKey(
     );
   }
 
-  const { licenseKey } = req.body;
+  const licenseKey = req.body.licenseKey.trim();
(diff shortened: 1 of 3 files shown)


mknowlton89 (approved):
LGTM - It looks like we need to update the test suite to handle the new logic. Approving so once the test suite is updated you can merge it in.

PR #2040, opened 2024-01-16
fix(sdk-js): fixed timout

nodirnasirov:
### Features and Changes

there is an error in code, 
```    const finish = (data?: T) => {
      if (resolved) return;
      resolved = true;
      timer && clearTimeout(timer as NodeJS.Timer); //<-- This throws error, correct should be NodeJS.Timout
      resolve(data || null);
    };```


jdorn:
Ended up getting fixed as part of #2637

PR #2046, opened 2024-01-17
Introduce backend context object

mknowlton89:
### Features and Changes

This PR introduces a new context object that ultimately replaces `getOrgFromReq`. Now, we can call `getContextFromReq` which returns everything `getOrgFromReq` did, in addition to the user's `readAccessFilter`. 

This `context` object can then be passed down to service files and DB files as a single param, rather than splitting everything up.

Eventually, we'd like to move the generation of this context object to the middleware, but will do so in a later PR as there are some intricacies around `organizations` being undefined that we need to work through.

In addition to introducing the new concept, it refactors the `apiKeyModel` to utilize this new context object, rather than passing in the `organization` id and the user's `readAccessFilter`, and instead gets that information via the context object.

This is the beginning of a larger refactor that will reduce prop drilling, initially, for the `No Access Role`.

### Dependencies

N/A

### Testing

- [x] Ensure that navigating to `/settings/keys` as an admin returns all of the API keys for an organization.
- [x] Clicked through the app to ensure that we didn't introduce any regressions as a result of this refactor.


packages/back-end/src/services/organizations.ts
```diff
@@ -97,20 +99,34 @@ export function validateLoginMethod(
...
+
+export async function getContextFromReq(req: AuthRequest): Promise<ReqContext> {
   if (!req.organization) {
     throw new Error("Must be part of an organization to make that request");
   }
   if (!req.userId || !req.email) {
     throw new Error("Must be logged in");
   }
 
+  const teams = await getTeamsForOrganization(req.organization.id);
```
bttf, line 119: I looked into how we could keep `getContextFromReq` from being `async` and it looks like we already are loading `teams` in our auth middleware, just not attaching it to the req object.

https://github.com/growthbook/growthbook/blob/0b77e87f936c0a9fd319a4f4f4df6e52bee5dd8e/packages/back-end/src/services/auth/index.ts#L178

What do you think about adding this to `req` via `req.teams` and using that to generate the read access filter? That should allow us to keep this method synchronous

packages/back-end/src/services/auth/index.ts
```diff
@@ -190,9 +190,6 @@ export async function processJWT(
 
         // init license for org if it exists
         await initializeLicense(req.organization.licenseKey);
-        req.readAccessFilter = getReadAccessFilter(
```
bttf, line 193: CI warning above

packages/back-end/src/services/organizations.ts
mknowlton89: You bet - I went back and forth on this. I've added `teams` to the `req` object as requested!

bttf (approved):
LGTM 👍🏾

PR #2048, opened 2024-01-18
Filter projects based on user's read access level

mknowlton89:
### Features and Changes

This PR is a continuation of the `No Access Role` series of PRs and introduces the following functionality:
- Filters the list of projects a user can access based on their permission level
- Updates the project drop-down in the left-nav to hide projects the user doesn't have readAccess to
- Updates the project drop-down in the left-nav so that if the user only has access to 1 project, rather than showing the dropdown, we just show that project name

In order to support this, I did have to update a number of other DB models, and thus, service files, and controller files to consume the `readAccessFilter`.

Additionally, this PR introduces the first changes to the REST API directory. Now, the REST API request object has a context object that contains the necessary information about the user/api key (including the `readAccessFilter`).

And, its important to note that in some of the asynchronous jobs (e.g. updateScheduledFeature) those jobs are initiated based on date/time and not directly from a user's action, so in that case, we are unable to build a `readAccessFilter` from the request object because there is no request object. In that case, I've created a constant in `permissions.utils.ts` that generates a `readAccessFilter` that is equivalent to the `readAccessFilter` an admin user would have. This way, the job can complete. This is ok because in order for the job to have been scheduled in the first place, we've checked the user's `readAccessLevel`.

> [!NOTE]  
> Given how interconnected everything is, updating the `project` DB model and associated controller file required updates to other DB models/controllers/etc. I will update these models fully in future PRs.

### Testing

- [x] Log in as an admin User and ensure that all projects are visible in the project drop-down
Screen Shot 2024-01-16 at 9 27 51 AM

- [x] Log in as a user with access to all projects except 1, and ensure that it is not visible in the project drop-down in the left-nav

Screen Shot 2024-01-16 at 9 28 25 AM

- [x] As the same user, create a new SDK Connection, and ensure the project the user doesn't have access to, isn't in the projects dropdown

Screen Shot 2024-01-16 at 9 28 44 AM

- [x] Now, log in as a user that only has readAccess to a single project and ensure that the dropdown in the left nav isn't displayed, and instead, just the project the user has access to is visible.

Screen Shot 2024-01-16 at 9 23 11 AM

- [x] Now, create an SDK connection and ensure only the project the user has access to is visible in the dropdown.

Screen Shot 2024-01-16 at 9 23 35 AM


packages/back-end/src/jobs/updateExperimentResults.ts
```diff
@@ -109,18 +110,21 @@ export default async function (agenda: Agenda) {
 
 async function updateSingleExperiment(job: UpdateSingleExpJob) {
   const experimentId = job.attrs.data?.experimentId;
-  const orgId = job.attrs.data?.organization;
-  if (!experimentId || !orgId) return;
+  const context = job.attrs.data?.context;
```
mknowlton89, line 113: Calling this out as I'm not confident I've done enough to correctly add the `context` object as a job attribute. With other jobs, I can see where they're being called from the codebase, but I don't see anywhere in the codebase where this job can be initiated via a user's action.

packages/back-end/src/jobs/updateExperimentResults.ts
```diff
@@ -94,7 +96,14 @@ export default async function (agenda: Agenda) {
...
     const job = agenda.create(UPDATE_SINGLE_EXP, {
-      organization,
+      context: {
+        org: organization,
+        userId: "",
+        email: "",
+        environments: [],
+        userName: "",
+        readAccessFilter: FULL_ACCESS_PERMISSIONS,
+      },
```
mknowlton89, line 106: Calling out this change as I'm not completely confident in it.

packages/back-end/src/jobs/updateScheduledFeatures.ts
```diff
@@ -33,7 +32,14 @@ async function queueFeatureUpdate(
...
     featureId: feature.id,
-    organization: feature.organization,
+    context: {
+      org: feature.organization,
+      userId: "",
+      email: "",
+      environments: [],
+      userName: "",
+      readAccessFilter: FULL_ACCESS_PERMISSIONS,
+    },
```
mknowlton89, line 42: Same here. Calling out this change as I'm not completely confident in it.

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
```diff
@@ -116,6 +123,19 @@ export default function authenticateApiRequestMiddleware(
 
       const teams = await getTeamsForOrganization(org.id);
 
+      req.context = {
+        org,
+        userId: req.user?.id || "",
+        email: req.user?.email || "",
```
bttf, line 129: When dealing with API requests, it seems like we can't always guarantee there is an authenticated user behind the request. Is that correct? Do we have an idea what the exact cases are? (Is `req.user` actually ever set?) 

It might make sense to create a different Context type to reflect this, such as `ApiReqContext` where fields like userId, email, and username are optional. This will make it more obvious that req.user can be undefined when dealing with API requests. What do you think?

packages/back-end/types/organization.d.ts
```diff
@@ -274,3 +275,12 @@ export type NamespaceUsage = Record<
...
 >;
+
+export type ReqContext = {
+  org: OrganizationInterface;
+  userId: string;
+  email: string;
+  environments: string[];
+  userName: string;
+  readAccessFilter: ReadAccessFilter;
+};
```
bttf, line 286: To follow up on the 'ApiReqContext' comment, we could leverage utility types to ensure the two interfaces stay in sync:

```
// req.user is not always guaranteed within API requests
export type ApiReqContext = Omit<ReqContext, 'userId' | 'email' | 'userName'> & {
  userId?: string;
  email?: string;
  userName?: string;
};
```

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
mknowlton89: @bttf You're correct - there is a case where `req.user` isn't set.

This middleware is for our REST API, and the requested can either use a personal access token (which is tied to a user) or an API Key that isn't tied directly to a user, but instead is just tied to an organization, and has a role (currently either `admin`, or `readonly`. (Note: I do have a PR to add support for a `SCIM` type role, but that hasn't landed yet.)

After writing the above, I wonder if we should just make those fields optional on `ReqContext` since there are valid cases where those fields wouldn't be defined. Then, if we ever need to use those fields, we can run quick checks to ensure their set.

bttf: >This middleware is for our REST API, and the requested can either use a personal access token (which is tied to a user) or an API Key that isn't tied directly to a user, but instead is just tied to an organization, and has a role (currently either admin, or readonly. (Note: I do have a PR to add support for a SCIM type role, but that hasn't landed yet.)

Thank you for clarifying, this is good to know.

>After writing the above, I wonder if we should just make those fields optional on ReqContext since there are valid cases where those fields wouldn't be defined. Then, if we ever need to use those fields, we can run quick checks to ensure their set.

Can you identify the valid cases where userId, email would not be defined (in a non-REST api situation)? My understanding is that since we're generating the context after the middleware layer using `getContextFromReq`, user and org should always be defined (otherwise `getContextFromReq` would throw an error).

From my experience it would be beneficial to have two contexts defined - one for in-app requests and the other for REST API requests. It would help address above situation and it lets us more accurately define a context object for either situation, which tend to have different requirements.

mknowlton89: Hmm... looking at the non-REST Api situation, the `ReqContext` object would never have `userId` or `email` not defined. Though, `userName` could be undefined as we don't require a name field for a user. 

I can definitely create the two different types (1 for our internal API where all fields but `userName` are required, and 1 for the rest API where those fields (userName, userId, email) are optional.

My only hesitation (and happy to talk through this in a huddle) is it feels like it won't make a material change in how we consume these two objects. E.G. In ProjectModel, when we're defining the `context` arg for a function, wouldn't we have to declare that it could be either `ReqContext | ApiReqContext`? And in that case `ApiReqContext` would essentially override `ReqContext` so we'd still have to run checks to see if `userId` or `email` was defined before using them?

Perhaps I'm looking at this too narrowly, and the value will come if we use the `ReqContext` object derived from `getContextFromReq` in other places, and that's the point you're making?

bttf: > ProjectModel, when we're defining the context arg for a function, wouldn't we have to declare that it could be either ReqContext | ApiReqContext? And in that case ApiReqContext would essentially override ReqContext so we'd still have to run checks to see if userId or email was defined before using them?

This is a great point and yes, that is correct. Still, I think we would benefit from being able to use `ReqContext` in places like the router and controller for in-app requests, where we know it's a `ReqContext` (and not an `ApiReqContext`). 

Furthermore, having a method whose context parameter is typed as a union of both contexts (`ReqContext | `ApiRequestContext`)  gives the developer some signal that 'this method is used for both in-app and API requests, so be wary.'  In the future, we can refactor these methods when complexity grows into separate methods for either context.

Overall I think it will help to have two contexts to have that added benefit when we can. What do you think?

packages/back-end/src/jobs/updateExperimentResults.ts
```diff
@@ -93,8 +98,18 @@ export default async function (agenda: Agenda) {
...
+
+    const context: ApiReqContext = {
+      org,
+      environments: getEnvironmentIdsFromOrg(org),
+      readAccessFilter: FULL_ACCESS_PERMISSIONS,
+    };
+
     const job = agenda.create(UPDATE_SINGLE_EXP, {
-      organization,
+      context,
```
bttf, line 112: I think this change requires an accompanying change to the `job.unique` call below. (Might want to double-check with Jeremy on this.)

Something like:
```
job.unique({
  experimentId,
  context: {
    organization: { id: org.id }
  },
});
```

packages/back-end/src/jobs/updateExperimentResults.ts
```diff
@@ -109,18 +124,21 @@ export default async function (agenda: Agenda) {
...
+  const experiment = await getExperimentById(org.id, experimentId);
   if (!experiment) return;
 
   const organization = await findOrganizationById(experiment.organization);
   if (!organization) return;
 
   let project = null;
   if (experiment.project) {
-    project = await findProjectById(experiment.project, organization.id);
+    project = await findProjectById(context, organization.id);
```
bttf, line 141: Seems like we can refactor `findProjectById` to take in just the context

packages/back-end/src/jobs/updateExperimentResults.ts
```diff
@@ -109,18 +124,21 @@ export default async function (agenda: Agenda) {
...
-  const orgId = job.attrs.data?.organization;
-  if (!experimentId || !orgId) return;
+  const context = job.attrs.data?.context;
+
+  if (!experimentId || !context) return;
 
-  const experiment = await getExperimentById(orgId, experimentId);
+  const { org } = context;
+
+  const experiment = await getExperimentById(org.id, experimentId);
```
bttf, line 133: We can refactor this to take in a context or leave for later

packages/back-end/src/jobs/updateScheduledFeatures.ts
```diff
@@ -31,9 +31,19 @@ async function queueFeatureUpdate(
...
+  const context: ApiReqContext = {
+    org,
+    environments: getEnvironmentIdsFromOrg(org),
+    readAccessFilter: FULL_ACCESS_PERMISSIONS,
+  };
+
   const job = agenda.create(UPDATE_SINGLE_FEATURE, {
     featureId: feature.id,
-    organization: feature.organization,
+    context,
```
bttf, line 46: Similar comment RE: unique

packages/back-end/src/controllers/experiments.ts
```diff
@@ -126,15 +126,15 @@ export async function getExperimentsFrequencyMonth(
...
   let project = "";
   if (typeof req.query?.project === "string") {
     project = req.query.project;
   }
 
-  const allProjects = await findAllProjectsByOrganization(org.id);
+  const allProjects = await findAllProjectsByOrganization(context);
   const { num } = req.params;
-  const experiments = await getAllExperiments(org.id, project);
+  const experiments = await getAllExperiments(context.org.id, project);
```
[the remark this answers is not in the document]
mknowlton89: Definitely. I do have a separate PR that handles updating the `ExperimentModel`, so I'll save this for that [PR](https://github.com/growthbook/growthbook/pull/2055).

packages/back-end/src/controllers/features.ts
```diff
@@ -1749,16 +1765,16 @@ export async function toggleStaleFFDetectionForFeature(
   res: Response<{ status: 200 }, EventAuditUserForResponseLocals>
 ) {
   const { id } = req.params;
-  const { org } = getContextFromReq(req);
-  const feature = await getFeature(org.id, id);
+  const context = getContextFromReq(req);
+  const feature = await getFeature(context.org.id, id);
```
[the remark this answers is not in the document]
mknowlton89: Same as above - I'm going to handle that in the FeatureModel PR

packages/back-end/src/jobs/updateExperimentResults.ts
mknowlton89: Same as above - I've got a separate PR that updates the `ExperimentModel` and all of it's methods [here](https://github.com/growthbook/growthbook/pull/2055).

packages/back-end/src/jobs/updateExperimentResults.ts
mknowlton89: Looking at AgendaJS' documentation, it looks like this ensure that only one instance of this job exists with the specified properties, so I think this should be sufficient. But I will double check with Jeremy just to be safe.

mknowlton89: Jeremy had feedback to update these jobs so we continue just passing in the organization when queing, and when executing the job, we build the `context` object from the `orgId`.

bttf (approved):

PR #2049, opened 2024-01-18
Make id join finder more efficient in how it picks needed joins

lukesonnet:
Makes `getBaseIdTypeAndJoins` find more efficient joins.

### DIFF
--- a/packages/back-end/test/util/sql.test.ts
+++ b/packages/back-end/test/util/sql.test.ts
@@ -202,6 +202,23 @@ describe("backend", () => {
       });
     });
 
+    it("uses id frequency count to find more efficient joins", () => {
+      expect(
+        getBaseIdTypeAndJoins([
+          ["id1", "id2"],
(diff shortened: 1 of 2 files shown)


packages/back-end/test/util/sql.test.ts
```diff
@@ -202,6 +202,23 @@ describe("backend", () => {
...
+          ["id1", "id3"],
+          ["id2", "id3"],
+          ["id4", "id3"],
+          // to make id 1 most common
+          ["id1", "id8"],
+          ["id1", "id9"],
+        ])
+      ).toEqual({
+        baseIdType: "id1",
+        joinsRequired: ["id3"],
```
lukesonnet, line 218: Previously this would return `joinsRequired: ["id2", "id4"]` because it just used `types[0]` to pick a type to join.

jdorn (approved):

PR #2055, opened 2024-01-22
First pass at updating the ExperimentModel

mknowlton89:
### Features and Changes
This PR primarily updates the `ExperimentModel` file to filter resources based on the user's read access level. In doing so, I did have to update 42 other files that ultimately consume the functions that are defined in the ExperimentModel file.

### Testing

- [x] Log in as an admin user and ensure all experiments are returned on the `/experiments` page
- [x] Now, log in as a user with access to only a single project (global role is `noaccess` and project-level role grants experimenter access) and ensure only experiments in the project the user has access to is visible
- [x] Now, log in as a user who's base access gives them experimenter permissions to all projects, but they have a project-level `noaccess` role for 1 project and ensure experiments from that project are not visible
- [x] Ensure that trying to directly access an experiment via the experiment's id page, returns an error if the experiment is attached to a project the user has a `noaccess` role for


packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -534,14 +548,18 @@ export async function getPastExperimentsByDatasource(
     }
   );
 
-  return experiments.map((exp) => ({
+  const experimentsUserCanAccess = experiments.filter((exp) =>
+    hasReadAccess(context.readAccessFilter, exp.project)
```
jdorn, line 552: This won't work.  In the above db query, we're only returning the `id` and `trackingKey` field, not the `project` field so this will always be undefined.

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -626,7 +645,11 @@ export async function getExperimentsForActivityFeed(
     }
   );
 
-  return experiments.map((exp) => ({
+  const filteredExperiments = experiments.filter((exp) =>
+    hasReadAccess(context.readAccessFilter, exp.project)
```
jdorn, line 649: Same.  This won't work because the above db query does not return a `project` column

jdorn (approved):

PR #2057, opened 2024-01-22
REST API - Visual changeset updates

bttf:
Updates the following endpoints:
- `POST /api/v1/visual-changesets/:id/visual-change`
- `PUT /api/v1/visual-changesets/:id/visual-change/:vc-id`
- `PUT /api/v1/visual-changesets`


packages/back-end/src/models/VisualChangesetModel.ts
```diff
@@ -505,3 +505,17 @@ export const deleteVisualChangesetById = async ({
...
+  organization: string
+): Promise<ExperimentInterface | null> => {
+  const visualChangeset = await findVisualChangesetById(
+    visualChangesetId,
+    organization
+  );
+
+  if (!visualChangeset) return null;
+
+  return getExperimentById(organization, visualChangeset.experiment);
```
jdorn, line 520: The method signature for `getExperimentById` has changed to work with a Context object so this will need to be updated

bttf: Will update.

Interesting that there are no merge conflicts. I'm guessing if this were merged, it would have failed on main.

jdorn (approved):

PR #2060, opened 2024-01-24
Enable source maps when running as dev

itsgrimetime:
### Description

Enables source maps when running with `yarn dev`, `start` and `start:with-tracing` so stacktraces are usable.

I don't think there are performance impacts (other than the maps themselves taking up a few more MB in the compiled app), but I'm not a node/ts expert :)


packages/back-end/package.json
```diff
@@ -12,16 +12,16 @@
     "clean": "rimraf node_modules",
     "dev": "yarn build:clean && concurrently \"yarn watch-compile\" \"yarn watch-dev | yarn pino-pretty\"",
     "build": "yarn build:clean && yarn build:typescript && yarn build:emails",
-    "start": "node dist/server.js",
-    "start:with-tracing": "node --require @opentelemetry/auto-instrumentations-node/register dist/server.js",
+    "start": "node --enable-source-maps dist/server.js",
+    "start:with-tracing": "node --enable-source-maps --require @opentelemetry/auto-instrumentations-node/register dist/server.js",
```
romain-growthbook, line 16:
```suggestion
    "start": "node dist/server.js",
    "start:with-tracing": "node --require @opentelemetry/auto-instrumentations-node/register dist/server.js",
    "start:with-source-maps": "NODE_OPTIONS=--enable-source-maps yarn start",
    "start:with-tracing:with-source-maps": "NODE_OPTIONS=--enable-source-maps yarn start:with-tracing",
```

romain-growthbook (approved):
Thanks for this PR!

While the source-maps options seems useful to debug the application, there are several reports of large memory usage increase and, at least originally, performances degradations:
* Original report: https://github.com/nodejs/node/issues/41541
* 2024 comment: https://github.com/evanw/esbuild/issues/1685#issuecomment-2205059975

Therefore, it seems wise to consider not enabling that systematically, at least for now.

For dev stack, it should be safe to have it by default to help debugging local work but for production I suggest that we add `with-source-maps` namespaces to the start script.

jdorn:
I agree, let's leave the changes for `yarn dev`, but revert all of the changes for production.

In production, you can enable source maps via an environment variable without needing to modify the code here at all.
```
NODE_OPTIONS=--enable-source-maps
```

romain-growthbook:
Update: After talking with @jdorn we believe simply using `NODE_OPTIONS` before invoking the default start script would suffice. Gonna revert the start scripts and merge.

packages/back-end/package.json
```diff
@@ -14,14 +14,16 @@
     "build": "yarn build:clean && yarn build:typescript && yarn build:emails",
     "start": "node dist/server.js",
     "start:with-tracing": "node --require @opentelemetry/auto-instrumentations-node/register dist/server.js",
+    "start:with-source-maps": "NODE_OPTIONS=--enable-source-maps yarn start",
+    "start:with-tracing:with-source-maps": "NODE_OPTIONS=--enable-source-maps yarn start:with-tracing",
```
romain-growthbook, line 18:
```suggestion
```

PR #2062, opened 2024-01-24
[SDK][Back-end] URL Redirect Testing

msamper:
### Features and Changes

Adds a way to run URL redirect tests on GrowthBook. URL redirect tests are a new type of linked change to an experiment that use Visual Changes under the hood. SDK users can customize the navigate function depending on where they are using the SDK (i.e. a user can customize the navigate function to use the router in a SPA) 

### Dependencies

#2121 Redirect Testing UI

### Testing

1. Setup a basic Create React App
2. Implement the Growthbook SDK on the app
3. Build the JS SDK locally
4. Use yarn link to reference the local build from the Growthbook backend and your sample app
5. Create a URL redirect test and make sure you are redirected as expected
6. Once you confirm basic redirects work, play around with enabling `antiFlicker` on the SDK and making sure query strings are persisted as expected when requested.


packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1298,6 +1298,18 @@ export class GrowthBook<
         undo.push(mutate.declarative(mutation as DeclarativeMutation).revert);
       });
     }
+    if (changes.urlRedirect) {
+      if (this._ctx.navigate) {
+        this._ctx.navigate(changes.urlRedirect);
+      } else {
+        const script = document.createElement("script");
+        script.innerHTML = `setTimeout(() => { window.location.replace("${
```
msamper, line 1306: Do we need to do a white overlay over the page or something similar so the user isn't exposed to the control during the timeout period?

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +500,37 @@ export class GrowthBook<
...
+          experiment.urlPatterns &&
+          isURLTargeted(url, [experiment.urlPatterns[0]])
+        ) {
+          // log or throw error when redirect URL matches URL pattern?
+          return result;
+        }
+        this._redirectedUrl = url;
+        const navigate = this._getNavigateFunction();
+        if (navigate) {
+          this._setAntiFlicker();
```
msamper, line 517: Move to inside default navigate function (prob not needed for something like next js

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +500,37 @@ export class GrowthBook<
...
-        });
+      if (result.value.urlRedirect) {
+        const url = experiment.persistQueryString
+          ? result.value.urlRedirect + window.location.search
+          : result.value.urlRedirect;
+        if (
+          experiment.urlPatterns &&
+          isURLTargeted(url, [experiment.urlPatterns[0]])
+        ) {
+          // log or throw error when redirect URL matches URL pattern?
```
msamper, line 511: log when this happens (refer to this.log used in other places)

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +500,37 @@ export class GrowthBook<
...
-      const undo = this._applyDOMChanges(result.value);
-      if (undo) {
-        this._activeAutoExperiments.set(experiment, {
-          undo,
-          valueHash,
-        });
+      if (result.value.urlRedirect) {
+        const url = experiment.persistQueryString
+          ? result.value.urlRedirect + window.location.search
+          : result.value.urlRedirect;
```
msamper, line 506: This wont work. Need to merge curr page query params with any possible supplied query params in the redirect URL

msamper: Use url in the context to grab query params, not window.location.search

packages/back-end/src/services/features.ts
```diff
@@ -104,8 +104,9 @@ function generateVisualExperimentsPayload({
   const isValidSDKExperiment = (
     e: AutoExperimentWithProject | null
   ): e is AutoExperimentWithProject => !!e;
-  const sdkExperiments: Array<AutoExperimentWithProject | null> = visualExperiments.map(
-    ({ experiment: e, visualChangeset: v }) => {
+  const sdkExperiments: Array<AutoExperimentWithProject | null> = visualExperiments
+    .sort((_a, b) => (b.visualChangeset.urlRedirects.length ? 1 : -1))
```
msamper, line 108: Filter and concat (split into url redirect experiments and visual editor exps and place url redirects first)

msamper:
Recent changes seem to have broken the redirects. I'm not seeing the tracking callback firing either. Investigating what is going on

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -521,8 +568,10 @@ export class GrowthBook<
     });
 
     // Re-run all new/updated experiments
-    experiments.forEach((exp) => {
+    experiments.some((exp) => {
       this._runAutoExperiment(exp, forceRerun);
+      // Break if we encounter an experiment that is redirecting
+      return exp.variations.some((v) => !!v.urlRedirect);
     });
```
msamper, line 575: This is the culprit for the redirect not happening

msamper: Need to switch code to only break out if the current page will be redirecting

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +501,46 @@ export class GrowthBook<
...
     if (result.inExperiment) {
-      const undo = this._applyDOMChanges(result.value);
-      if (undo) {
-        this._activeAutoExperiments.set(experiment, {
-          undo,
-          valueHash,
-        });
+      if (result.value.urlRedirect) {
+        const currUrl = new URL(this._getContextUrl());
+        const redirectUrl = new URL(result.value.urlRedirect);
```
jdorn, line 506: This will throw an error if the URL is invalid, so we should catch this.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +501,46 @@ export class GrowthBook<
...
+        mergeUrlSearchParams(currUrl.searchParams, redirectUrl.searchParams);
+
+        const url = experiment.persistQueryString
+          ? redirectUrl.toString()
+          : result.value.urlRedirect;
```
jdorn, line 511: This is a little weird, since we're doing the whole merge stuff above, but then only using it if persistQueryString is set.

I would restructure this a bit and do:

```js
const url = experiment.persistQueryString
  // New function, takes in two full URL strings and returns a new merged URL string
  ? mergeQueryStrings(this._getContextUrl(), result.value.urlRedirect)
  : result.value.urlRedirect
```

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +501,46 @@ export class GrowthBook<
...
+        if (
+          experiment.urlPatterns &&
+          isURLTargeted(url, experiment.urlPatterns)
+        ) {
```
jdorn, line 515: This makes me thing - what if the experiment doesn't have `urlPatterns` at all?  That could be pretty bad.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +501,46 @@ export class GrowthBook<
...
+        const navigate = this._getNavigateFunction();
+        if (navigate) {
+          if (isBrowser) {
+            this._setAntiFlicker();
+            window.setTimeout(() => {
+              navigate(url);
+            }, this._ctx.navigateDelay ?? 100);
+          } else {
+            navigate(url);
+          }
```
jdorn, line 534: Any time we call a user-supplied function, we should wrap it in a try/catch

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1278,6 +1359,31 @@ export class GrowthBook<
...
+    }
+    return null;
+  }
+
+  private _setAntiFlicker() {
+    if (!this._ctx.antiFlicker || !isBrowser) return;
+    const styleTag = document.createElement("style");
+    styleTag.innerHTML = ".gb-anti-flicker { opacity: 0 !important; }";
+    document.head.appendChild(styleTag);
+    document.body.classList.add("gb-anti-flicker");
```
jdorn, line 1384: We should wrap this whole function body in a try/catch.  Also, many people add growthbook in the `<head>` and if that loads super quickly, `document.body` might be undefined.  Safer to use `document.documentElement` instead, which points to the `<html>` element and will always be defined.

packages/sdk-js/src/GrowthBook.ts
msamper: Hmm how do we handle that possibility with a visual changesets currently?

packages/back-end/src/services/features.ts
```diff
@@ -120,17 +133,21 @@ function generateVisualExperimentsPayload({
         phase?.savedGroups
       );
 
+      const urlRedirects = v.urlRedirects;
```
msamper, line 138: If no urlPatterns for a URL redirect, exclude from the visual experiments for the payload

packages/sdk-js/test/json.test.ts
```diff
@@ -258,4 +261,14 @@ describe("json test suite", () => {
       );
     });
   });
+
+  it.each((cases as Cases).urlRedirect)(
+    "urlRedirect[%#] %s",
+    (name, ctx, exp, result) => {
+      const growthbook = new GrowthBook(ctx);
+      const res = growthbook.run(exp);
```
msamper, line 269: Don't use run since we're testing auto experiments. Grab results via deferred tracking callbacks instead

packages/sdk-js/test/json.test.ts
```diff
@@ -67,6 +68,8 @@ type Cases = {
     gt: [string, string, boolean][];
     eq: [string, string, boolean][];
   };
+  // name, context, experiment, result
+  urlRedirect: [string, Context, AutoExperiment, string][];
```
msamper, line 72: Pass in experiments via the `Context` and make result an array of objects so we can check to make sure that tracking callbacks are only fired for one experiment rather than multiple if there is more than one experiment targeting the same url

packages/back-end/src/services/features.ts
```diff
@@ -104,7 +104,22 @@ function generateVisualExperimentsPayload({
   const isValidSDKExperiment = (
     e: AutoExperimentWithProject | null
   ): e is AutoExperimentWithProject => !!e;
-  const sdkExperiments: Array<AutoExperimentWithProject | null> = visualExperiments.map(
+
+  const urlRedirectExperiments = visualExperiments.filter(
```
msamper, line 108: Check for circular dependencies among URL redirect experiments

packages/sdk-js/src/util.ts
```diff
@@ -328,3 +328,23 @@ export function loadSDKVersion(): string {
   }
   return version;
 }
+
+export function mergeQueryStrings(oldUrl: string, newUrl: string): string {
```
msamper, line 332: Unit test this function and also test via urlRedirect cases when `persistQueryString` is `true`

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1278,6 +1353,40 @@ export class GrowthBook<
...
+
+      // Fallback if GrowthBook fails to load in specified time or 3.5 seconds
+      setTimeout(() => {
+        document.documentElement.classList.remove("gb-anti-flicker");
+      }, this._ctx.antiFlickerTimeout ?? 3500);
+    } catch (e) {
+      console.error(e);
+    }
+  }
+
```
msamper, line 1389: Test this to make sure antiFlicker gets removed as expected

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1278,6 +1353,40 @@ export class GrowthBook<
...
+      };
+    }
+    return null;
+  }
+
+  private _setAntiFlicker() {
+    if (!this._ctx.antiFlicker || !isBrowser) return;
+    try {
+      const styleTag = document.createElement("style");
+      styleTag.innerHTML = ".gb-anti-flicker { opacity: 0 !important; }";
```
msamper, line 1377: pointer-events: none

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1278,6 +1353,40 @@ export class GrowthBook<
...
+      return this._ctx.navigate;
+    } else if (isBrowser) {
+      return (url: string) => {
+        window.location.href = url;
+      };
+    }
+    return null;
+  }
+
+  private _setAntiFlicker() {
```
msamper, line 1373: antiFlicker might actually introduce more flicker. i.e. Growthbook takes a while to load, loads-> setAntiFlicker so user is actually getting more flicker

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -522,6 +568,9 @@ export class GrowthBook<
 
     // Re-run all new/updated experiments
     experiments.forEach((exp) => {
```
msamper, line 570: use traditional for loop so we can actually break

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +498,48 @@ export class GrowthBook<
...
+        if (navigate) {
+          if (isBrowser) {
+            this._setAntiFlicker();
+            window.setTimeout(() => {
+              try {
+                navigate(url);
+              } catch (e) {
+                console.error(e);
+              }
+            }, this._ctx.navigateDelay ?? 100);
```
msamper, line 526: Worth noting that tracking callbacks could fire and navigate could potentially fail

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1146,8 +1195,34 @@ export class GrowthBook<
...
+
+  public fireDeferredTrackingCalls(data: string) {
+    const calls = JSON.parse(atob(data));
+    if (!Array.isArray(calls)) throw new Error("Invalid tracking data");
+    calls.forEach((call: TrackingData) => {
+      if (!call || !call.experiment || !call.result) {
+        throw new Error("Invalid tracking data");
+      }
+      this._track(call.experiment, call.result);
+    });
```
msamper, line 1210: Could create a flush function for `deferredTrackingCalls` and call it here. Could also be used if we decided to leverage deferredTrackingCalls for the base case (inBrowser)

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1146,8 +1195,34 @@ export class GrowthBook<
     else console.log(msg, ctx);
   }
 
+  public getDeferredTrackingCalls() {
+    return btoa(JSON.stringify(this._deferredTrackingCalls));
```
msamper, line 1199: Consider moving encoding stuff out of the getters and setters. We shouldn't be opinionated at this level about network transport

msamper: If we could provide a service, we could be opinionated in instances of the service rather than the API here

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1146,8 +1195,34 @@ export class GrowthBook<
...
+  }
+
+  public fireDeferredTrackingCalls(data: string) {
```
msamper, line 1202: Would not take any arguments if we have a setter that is responsible for setting `deferredTracking`

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -106,6 +111,8 @@ export class GrowthBook<
     this._activeAutoExperiments = new Map();
     this._triggeredExpKeys = new Set();
     this._loadFeaturesCalled = false;
+    this._redirectedUrl = "";
+    this._deferredTrackingCalls = [];
```
msamper, line 115: Should be a context object rather than a private variable so we can hydrate

packages/sdk-js/src/util.ts
```diff
@@ -328,3 +328,23 @@ export function loadSDKVersion(): string {
...
+    currUrl = new URL(oldUrl);
+    redirectUrl = new URL(newUrl);
+  } catch (e) {
+    console.error(`Unable to merge query strings: ${e}`);
+    return newUrl;
+  }
+
+  redirectUrl.searchParams.forEach((value, key) => {
+    currUrl.searchParams.set(key, value);
+  });
```
msamper, line 345: `forEach` on `currUrl` instead. Conditionally set if key doesn't exist on redirectUrl Search params

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +498,48 @@ export class GrowthBook<
...
     if (result.inExperiment) {
-      const undo = this._applyDOMChanges(result.value);
-      if (undo) {
-        this._activeAutoExperiments.set(experiment, {
-          undo,
-          valueHash,
-        });
+      if (result.value.urlRedirect && experiment.urlPatterns) {
+        const url = experiment.persistQueryString
+          ? mergeQueryStrings(this._getContextUrl(), result.value.urlRedirect)
```
msamper, line 503: Worth calling out in the documentation that if SDK is not in a browser environment users will need to use setUrl() to set a context url. Only really necessary if users are looking to user persistQueryString on redirect experiments

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -488,12 +498,48 @@ export class GrowthBook<
...
+        this._redirectedUrl = url;
+        const navigate = this._getNavigateFunction();
+        if (navigate) {
+          if (isBrowser) {
+            this._setAntiFlicker();
+            window.setTimeout(() => {
+              try {
+                navigate(url);
+              } catch (e) {
+                console.error(e);
```
msamper, line 524: Could consider having an unset function for antiFlicker and not having a timeout for situations with a slow network. i.e. navigate takes a long time and antiflicker is removed because the 3.5 second default timeout passes

packages/sdk-js/src/GrowthBook.ts
msamper, line 1373: Could take in a timeout as a param

packages/back-end/src/services/features.ts
```diff
@@ -104,7 +104,22 @@ function generateVisualExperimentsPayload({
...
+    (e) =>
+      e.visualChangeset.urlRedirects &&
+      e.visualChangeset.urlRedirects.length > 0
+  );
+
+  const visualEditorExperiments = visualExperiments.filter(
+    (e) => !e.visualChangeset.urlRedirects
```
msamper, line 115: check for the inverse of the previous filter

packages/back-end/src/services/features.ts
```diff
@@ -120,17 +135,24 @@ function generateVisualExperimentsPayload({
...
+        variations: urlRedirects
+          ? (urlRedirects.map((r) => ({
+              urlRedirect: r.url,
+            })) as AutoExperimentWithProject["variations"])
+          : (v.visualChanges.map((vc) => ({
+              css: vc.css,
+              js: vc.js || "",
+              domMutations: vc.domMutations,
+            })) as AutoExperimentWithProject["variations"]),
+        persistQueryString: v.persistQueryString,
```
msamper, line 155: don't ship this down if undefined. Conditionally add this if a url redirect experiment

packages/sdk-js/src/GrowthBook.ts
msamper: Decided to not do this for MVP

packages/sdk-js/src/GrowthBook.ts
msamper: We decided this is okay as long as `antiFlicker` remains opt in

msamper: won't do

packages/sdk-js/src/util.ts
msamper: No need to unit test because it's being tested via cases.json

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2217,6 +2273,23 @@ export async function postVisualChangeset(
...
+      reqVariationIds.includes(v)
+    );
+    if (!areValidVariations) {
+      throw new Error("Invalid variation IDs for urlRedirects");
+    }
+
+    const origin = req.body.urlPatterns;
+    const destinations = req.body.urlRedirects;
+
+    await _validateRedirect(origin, destinations, context, experiment);
```
msamper, line 2290: Could consider using a param after this first errors if the user decides to continue anyway

packages/back-end/src/models/VisualChangesetModel.ts
```diff
@@ -271,7 +287,11 @@ export const createVisualChangeset = async ({
       urlPatterns,
       editorUrl,
       visualChanges:
-        visualChanges || experiment.variations.map(genNewVisualChange),
+        visualChanges || !urlRedirects
+          ? experiment.variations.map(genNewVisualChange)
+          : undefined,
```
jdorn, line 292: This doesn't look right.  Missing parentheses?

packages/back-end/src/services/experiments.ts
```diff
@@ -1972,6 +1972,13 @@ export function visualChangesetsHaveChanges({
     return true;
   }
 
+  // If there are redirect url differences
+  const oldRedirects = oldVisualChangeset.urlRedirects?.map((r) => r.url);
+  const newRedirects = newVisualChangeset.urlRedirects?.map((r) => r.url);
+  if (!isEqual(oldRedirects, newRedirects)) {
+    return true;
+  }
```
jdorn, line 1980: Should we also add a check to see if `persistQueryString` has changed?

packages/back-end/src/services/features.ts
```diff
@@ -441,10 +470,26 @@ async function getFeatureDefinitionsResponse({
   features = scrubFeatures(features, capabilities);
   experiments = scrubExperiments(experiments, capabilities);
 
+  const includeAutoExperiments =
+    !!includeRedirectExperiments && !!includeVisualExperiments;
```
jdorn, line 474: Shouldn't this be OR instead of AND?

packages/back-end/src/services/experiments.ts
msamper: Yes we should. I’ll add that check in

packages/back-end/src/services/features.ts
```diff
@@ -441,10 +470,26 @@ async function getFeatureDefinitionsResponse({
...
+
+  if (includeAutoExperiments) {
+    if (!includeRedirectExperiments) {
+      experiments = experiments.filter((e) =>
+        e.variations.some((v) => v.urlRedirect)
```
jdorn, line 479: This seems backwards.  Don't we want to return `false` in this filter call?  Also, there's the same issue of an array being used in a boolean check

msamper: You're right. It looks like I got the two filters mixed up here. I added `changeType` to the `VisualChangesetInterface` like you recommended in a previous comment, but should we add `changeType` to `AutoExperiment` as well to avoid using the `variations` array for this filter?

packages/back-end/src/models/VisualChangesetModel.ts
```diff
@@ -101,6 +101,18 @@ const visualChangesetSchema = new mongoose.Schema<VisualChangesetInterface>({
...
+    {
+      _id: false,
+      variation: String,
+      url: String,
+    },
+  ],
+  persistQueryString: Boolean,
+  changeType: {
+    type: String,
+    enum: ["visualEditor", "urlRedirect"],
```
jdorn, line 115: We need a migration for existing mongo documents

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1290,18 +1290,30 @@ export class GrowthBook<
...
     this._deferredTrackingCalls.forEach((call: TrackingData) => {
       if (!call || !call.experiment || !call.result) {
-        throw new Error("Invalid tracking data");
+        this.log("Invalid deferred tracking call", { call: call });
+        hasInvalidTrackingCall = true;
+      } else {
+        this._track(call.experiment, call.result);
       }
-      this._track(call.experiment, call.result);
+      this._deferredTrackingCalls.shift();
```
jdorn, line 1301: I would switch this line back to how you had it.  Just do a single `this._deferredTrackingCalls = [];` after the loop.  Now that you're not throwing inside the loop, it will always reach that line.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1290,18 +1296,31 @@ export class GrowthBook<
   }
 
   public fireDeferredTrackingCalls() {
+    let hasInvalidTrackingCall = false;
     this._deferredTrackingCalls.forEach((call: TrackingData) => {
       if (!call || !call.experiment || !call.result) {
-        throw new Error("Invalid tracking data");
+        this.log("Invalid deferred tracking call", { call: call });
```
jdorn, line 1302: Do a `console.error` directly here.  `this.log` is for debug statements only (`context.debug === true`), but I think this is something we want to make sure to surface to everyone in the browser console

jdorn (approved):

PR #2069, opened 2024-01-26
Update MetricModel & FactTableModel to filter by read access level

mknowlton89:
### Features and Changes

This PR updates Metrics and Fact Tables to consume the new backend `context` object, as well as the `hasReadAccess` filter method in order to filter metrics and fact tables by the user's read access level.

### Testing
Metrics

- [ ] Log in as an admin user and navigate to the `/metrics` page and ensure all metrics for an organization are returned and accessible
- [ ] Now, log in as a user who's global role is `noaccess` but the user has `experimenter` permissions for a single project and ensure that when you land on the `/metrics` route, only the metrics associated with that project are visible.
- [ ] Now, long in as a user who's global access gives readaccess, but they have a `noaccess` role for a single project. Ensure that all metrics except those associated with the project the user has the `noaccess` role are visible. In the event that a metric is a part of multiple projects, one of which is the project they have `noaccess` to, they should be able to see the metric, but the project tags that the user sees should have an `Unknown Project` tag for the project they have no access to.
- [ ] Ensure that if a metric is associated with a project they don't have access to (and it's not associated with any projects they do have access to) the user is unable to access the `metric/mid` page.

Fact Tables

- [ ] Log in as an admin user and navigate to the `/facttables` page and ensure all fact tables for an organization are returned and accessible
- [ ] Now, log in as a user who's global role is `noaccess` but the user has `experimenter` permissions for a single project and ensure that when you land on the `/facttables` route, only the fact tables associated with that project are visible.
- [ ] Now, long in as a user who's global access gives readaccess, but they have a `noaccess` role for a single project. Ensure that all fact tables except those associated with the project the user has the `noaccess` role are visible. In the event that a fact table is a part of multiple projects, one of which is the project they have `noaccess` to, they should be able to see the fact table, but the project tags that the user sees should have an `Unknown Project` tag for the project they have no access to.
- [ ] Ensure that if a fact table is associated with a project they don't have access to (and it's not associated with any projects they do have access to) the user is unable to access the `facttables/ftid` page.

### DIFF
--- a/packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
+++ b/packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
@@ -42,9 +42,7 @@ export const postBulkImportFacts = createApiRequestHandler(
 
     const factTableMap = await getFactTableMap(req.organization.id);
 
-    const allFactMetrics = await getAllFactMetricsForOrganization(
-      req.organization.id
-    );
+    const allFactMetrics = await getAllFactMetricsForOrganization(req.context);
     const factMetricMap = new Map<string, FactMetricInterface>(
       allFactMetrics.map((m) => [m.id, m])
     );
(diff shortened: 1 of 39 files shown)


mknowlton89:
Closing this as it was done in Jeremy's [Official Metric PR](https://github.com/growthbook/growthbook/pull/2007) and the ProjectBadges work has moved to the [DataSource](https://github.com/growthbook/growthbook/pull/2073) PR

PR #2072, opened 2024-01-28
Introduce query labels

aresowj:
### Features and Changes

For the sake of better monitoring resource usage, query label is needed to quickly sort queries by metric, experiment, organization etc...

Currently this is only supported in BigQuery integration. Other integrations can use the labels as desired.

  - `datasource`: the id of the datasource being used
  - `organization`: the id of the organization
  - `call`: the origin function of the query
  - `metric_id`: metric id
  - `metric_name`: metric name of the experiment metric query
  - `query_type`: identifier of the query type
  - `query_parent_id`: the id of the parent query if exists

### Testing

Going to run locally and see if labels are actually applied in BigQuery console.


aresowj:
@lukesonnet could you or the team take a look? Thanks.

jdorn:
General approach looks ok.  Will likely need to also make updates to the AWS Athena data source since Athena also uses `setExternalId` and that is now the 3rd argument instead of the 2nd.  I can do more testing of this tomorrow.

lukesonnet:
Hi, thanks for this. There's a few issues.

First of all, `labels` has to be added to the type AND the data model (some boilerplate we're working on reducing). I solved this issue, so now problem.

Second of all, your label values aren't working, from the BigQuery [docs](https://cloud.google.com/bigquery/docs/labels-intro):
>Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes. All characters must use UTF-8 encoding, and international characters are allowed. Keys must start with a lowercase letter or international character.

So you'll have to update some of these values to work. It might be good to test now that I think at least the GrowthBook side of things is working.

Third of all, those same docs discourage you from having key-value pairs with many values, and I wonder if this is the right way to identify these queries or not. I'm specifically thinking of the `metric.id` and `metric.name` labels you've added. It seems to me that query labels are not the right way to pass that information on to BQ. Here's the documentation again:
> We don't recommend creating large numbers of unique labels, such as for timestamps or individual values for every API call. The problem with this approach is that when the values change frequently or with keys that clutter the catalog, this makes it difficult to effectively filter and report on resources.

I wonder if we should remove the metric.id and metric.name labels and instead put them directly in the SQL, like we already do with name, I think, and in some cases with fact metrics.

aresowj:
> Hi, thanks for this. There's a few issues.
> 
> First of all, `labels` has to be added to the type AND the data model (some boilerplate we're working on reducing). I solved this issue, so now problem.
> 
> Second of all, your label values aren't working, from the BigQuery [docs](https://cloud.google.com/bigquery/docs/labels-intro):
> 
> > Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes. All characters must use UTF-8 encoding, and international characters are allowed. Keys must start with a lowercase letter or international character.
> 
> So you'll have to update some of these values to work. It might be good to test now that I think at least the GrowthBook side of things is working.
> 
> Third of all, those same docs discourage you from having key-value pairs with many values, and I wonder if this is the right way to identify these queries or not. I'm specifically thinking of the `metric.id` and `metric.name` labels you've added. It seems to me that query labels are not the right way to pass that information on to BQ. Here's the documentation again:
> 
> > We don't recommend creating large numbers of unique labels, such as for timestamps or individual values for every API call. The problem with this approach is that when the values change frequently or with keys that clutter the catalog, this makes it difficult to effectively filter and report on resources.
> 
> I wonder if we should remove the metric.id and metric.name labels and instead put them directly in the SQL, like we already do with name, I think, and in some cases with fact metrics. The query parent id is also very often unique per snapshot, so I wonder if it is also a good candidate to include or not.
> 
> I understand why y'all want to include them, but I'm wondering if this workaround is the wrong approach.

Hi Luke, thanks for the review and detailed feedback. We want to have this labeling for setting metric query monitoring. The approach to include metric name in the query should work, but not efficient enough for rendering histogram of execution time for a single metric. It will be slower if the monitor query has to search in the full SQL text.

I agree that the workaround may look questionable. I think at least the number of unique metric id/names is not too large, likely under 1000 for most users? Currently we have ~500 metrics in our catalog. It sounds like a reasonable of unique values to have in labels (consider you have hundreds of clusters which each has its own label in resource manager).

We should definitely remove parent query id as that don't tell us much information in most cases, just a way to group all queries in the same batch.

PR #2073, opened 2024-01-29
Updates datasource model to consume context.

mknowlton89:
### Features and Changes

This PR updates the `DataSourceModel` to ensure we're filtering data source's based on the user's `readAccess` level.

It also introduces a change to the `ProjectBadge` so that if a resource is a part of a multiple projects and the user doesn't have access to one (or more) of the projects, but they DO have access to atleast one of the resource's projects, we'll display an "Unknown Project" badge for the projects that the user doesn't have access to, in order to provide context that the resource is being used in projects other than the projects they have access to. Eventually, it'd be great to differentiate a project the user doesn't have access to vs a project that has been deleted. I've got a to-do item to look into that at a later date.

### Testing

- [x]  Log in as an admin user and navigate to the /datasources page and ensure all datasources for an organization are returned and accessible
- [x] Now, log in as a user who's global role is noaccess but the user has experimenter permissions for a single project and ensure that when you land on the /datasources route, only the datasources associated with that project are visible.
- [x] Now, long in as a user who's global access gives readaccess, but they have a noaccess role for a single project. Ensure that all datasources except those associated with the project the user has the noaccess role are visible. In the event that a datasources is a part of multiple projects, one of which is the project they have noaccess to, they should be able to see the datasource, but the project tags that the user sees should have an Unknown Project tag for the project they have no access to.
- [x] Ensure that if a datasources is associated with a project they don't have access to (and it's not associated with any projects they do have access to) the user is unable to access the datasources/did page.


packages/back-end/src/scripts/migrate-encryption-key.ts
```diff
@@ -35,6 +36,7 @@ async function run() {
   const allDatasources = await _dangerousGetAllDatasources();
   for (let i = 0; i < allDatasources.length; i++) {
     const ds = allDatasources[i];
+    const context = await getContextForAgendaJobByOrgId(ds.organization);
```
mknowlton89, line 39: In a separate PR, I'm going to update the name of this as it's no longer just being used by Agenda jobs. Thinking something like `dangerouslyGetContextObj`.

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -68,49 +71,47 @@ export async function getInstallationDatasources(): Promise<
...
 
   const doc: DataSourceDocument | null = await DataSourceModel.findOne({
     id,
-    organization,
+    organization: context.org.id,
   });
 
-  return doc ? toInterface(doc) : null;
-}
-export async function getDataSourcesByIds(ids: string[], organization: string) {
```
mknowlton89, line 100: This appears to be abandoned.

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -123,12 +124,6 @@ export async function removeProjectFromDatasources(
   );
 }
 
-export async function getOrganizationsWithDatasources(): Promise<string[]> {
```
mknowlton89, line 126: This appears to be abandoned.

packages/back-end/src/scripts/migrate-encryption-key.ts
bttf: Makes sense. Another naming suggestion `getFullAccessContextObj`

packages/back-end/test/models/dataSourceModel.test.ts
```diff
@@ -293,7 +294,24 @@ describe("dataSourceModel", () => {
...
+              members: [],
+              invites: [],
+              settings: {},
+            },
+            environments: [],
+            readAccessFilter: FULL_ACCESS_PERMISSIONS,
+          },
+          datasource,
+          updates
+        )
```
bttf, line 314: Might want to leave a TODO or make a note to create a test utility for generating a context object .. That way if it changes in the future we have one place to update and lessen chance of bugs slipping through

bttf (commented):
Code looks good. I'm not really sure on the context of `ProjectBadges`. Is it possible to include some screenshots for context?

mknowlton89:
> Code looks good. I'm not really sure on the context of `ProjectBadges`. Is it possible to include some screenshots for context?

Absolutely. Here are two that show what it'll look like after the changes.
Screen Shot 2024-01-29 at 4 01 40 PM

And when hovering...
Screen Shot 2024-01-29 at 4 01 45 PM

Without this change, we'd simply not display a badge for the project the user doesn't have access to (screenshot below). I added this as it can add some context about why a user may not be able to edit a resource. As, with metrics for example, in order to be able to edit a metric, the user must have edit permissions for every project the resource is associated with.

Screen Shot 2024-01-29 at 4 04 18 PM

bttf:
> > Code looks good. I'm not really sure on the context of `ProjectBadges`. Is it possible to include some screenshots for context?
> 
> Absolutely. Here are two that show what it'll look like after the changes. Screen Shot 2024-01-29 at 4 01 40 PM
> 
> And when hovering... Screen Shot 2024-01-29 at 4 01 45 PM
> 
> Without this change, we'd simply not display a badge for the project the user doesn't have access to (screenshot below). I added this as it can add some context about why a user may not be able to edit a resource. As, with metrics for example, in order to be able to edit a metric, the user must have edit permissions for every project the resource is associated with.
> 
> Screen Shot 2024-01-29 at 4 04 18 PM

Got it, makes sense. I wonder if instead we could show a single warning icon that indicates that this resource belongs to restricted projects, and show a tooltip that explains that their limited ability to make changes as a result. This might be clearer and will avoid multiple unknown projects tags in the case that there are many.

mknowlton89:
> > > Code looks good. I'm not really sure on the context of `ProjectBadges`. Is it possible to include some screenshots for context?
> > 
> > 
> > Absolutely. Here are two that show what it'll look like after the changes. Screen Shot 2024-01-29 at 4 01 40 PM
> > And when hovering... Screen Shot 2024-01-29 at 4 01 45 PM
> > Without this change, we'd simply not display a badge for the project the user doesn't have access to (screenshot below). I added this as it can add some context about why a user may not be able to edit a resource. As, with metrics for example, in order to be able to edit a metric, the user must have edit permissions for every project the resource is associated with.
> > Screen Shot 2024-01-29 at 4 04 18 PM
> 
> Got it, makes sense. I wonder if instead we could show a single warning icon that indicates that this resource belongs to restricted projects, and show a tooltip that explains their limited ability to make changes as a result. This might be clearer and will avoid multiple unknown projects tags in the case that there are many.

That's a great idea - exploring that now.

packages/front-end/components/ProjectBadges.tsx
```diff
@@ -1,18 +1,24 @@
...
   className?: string;
 }
 
 export default function ProjectBadges({
+  resourceType,
   projectIds,
   sort = true,
   className = "badge-ellipsis short",
 }: Props) {
+  const [showMissingProjectError, setShowMissingProjectError] = useState(false);
```
bttf, line 21: We don't really need a state variable for this, it could be calcculated on the fly:
```
const showMissingProjectErr = filteredProjects.some(p => !p);
```

bttf (approved):

PR #2074, opened 2024-01-29
Update SDK Connections model to consume context and filter by readAccess

mknowlton89:
### Features and Changes

This PR updates the `SDKConnectionModel` to ensure we're filtering SDK connections based on the user's `readAccess` level.

SDK Connections have taken on a few different forms over the years - they've been legacy SDK endpoints, and now, SDK Connections. Given that, when testing, we need to look at both the newer `SDK Connections` as well as the legacy SDK endpoints, and also SDK Webhooks

### Testing

- [x]  Log in as an admin user and navigate to the /sdks page and ensure all SDK Connections for an organization are returned and accessible.
- [x] Do the same for the legacy SDK Endpoints
- [x] Do the same for the legacy SDK webhooks
- [x] Now, log in as a user who's global role is noaccess but the user has experimenter permissions for a single project and ensure that when you land on the /sdks route, only the SDK Connections associated with that project (or SDK Connections that can be used for All Projects) are visible.
- [x] Do the same for the legacy SDK Endpoints
- [x] Do the same for the legacy SDK webhooks
- [x] Now, log in as a user who's global access gives readaccess, but they have a noaccess role for a single project. Ensure that all datasources except those associated with the project the user has the noaccess role are visible. In the event that a datasources is a part of multiple projects, one of which is the project they have noaccess to, they should be able to see the datasource, but the project tags that the user sees should have an Unknown Project tag for the project they have no access to.
- [x] Do the same for the legacy SDK Endpoints
- [x] Do the same for the legacy SDK webhooks


packages/back-end/src/jobs/proxyUpdate.ts
```diff
@@ -141,7 +150,9 @@ export async function queueProxyUpdate(
   if (!CRON_ENABLED) return;
   if (!payloadKeys.length) return;
 
-  const connections = await findSDKConnectionsByOrganization(orgId);
+  const context = await getContextForAgendaJobByOrgId(orgId);
+
+  const connections = await findSDKConnectionsByOrganization(context);
```
jdorn, line 155: It looks like the only place calling `queueProxyUpdate` is `triggerWebhookJobs`, which already has a context object.  So I think we can just change the method signature for this function to take context instead of orgId.

packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
```diff
@@ -113,34 +114,15 @@ export default function SDKConnectionsList() {
...
-                              </span>
-                            </Tooltip>
-                          );
-                        } else {
-                          return (
-                            <span className="badge badge-secondary mr-1">
-                              {proj.name}
-                            </span>
-                          );
+                    <td className="d-flex align-items-center">
```
mknowlton89, line 117: Updated this to utilize the `ProjectBadges` component to increase continuity.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -1489,14 +1490,17 @@ export async function postApiKeyReveal(
 }
 
 export async function getWebhooks(req: AuthRequest, res: Response) {
-  const { org } = getContextFromReq(req);
+  const context = getContextFromReq(req);
   const webhooks = await WebhookModel.find({
```
mknowlton89, line 1494: I've created a follow up [ticket](https://github.com/growthbook/growthbook/issues/2086) to refactor the Webhook model so we're not exporting the model.

packages/back-end/src/jobs/proxyUpdate.ts
```diff
@@ -116,12 +123,14 @@ export default function addProxyUpdateJob(ag: Agenda) {
...
 
 export async function queueSingleProxyUpdate(
+  context: ReqContext | ApiReqContext,
   connection: SDKConnectionInterface,
   useCloudProxy: boolean = false
 ) {
   if (!connectionSupportsProxyUpdate(connection, useCloudProxy)) return;
 
   const job = agenda.create(PROXY_UPDATE_JOB_NAME, {
+    context,
```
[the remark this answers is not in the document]
bttf: The more I think about this, the less it seems like a good idea. It seems like our context object shouldn't be persisted anywhere nor leave the scope of our express webserver. It can contain sensitive information (users' emails), or be potentially huge (e.g. the upcoming data models collection from Jeremy's PR). 

Putting the context inside of a message body sent to distributed services makes our context object more sensitive to changes i.e. we have to be more careful with what we put in there. Right now, I think we just want it to be a toolbox for all things needed within the lifetime of a request within our express server without worrying about concerns like size or security. Our previous approach with embedding the only org ID seems more robust with this in mind.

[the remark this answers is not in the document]
mknowlton89: That's a great call. I've updated this job so it no longer takes in the context object, and instead just takes in the orgId, and builds a `FULL_ACCESS_CONTEXT` object on the fly.

bttf (approved):

PR #2075, opened 2024-01-29
Coderefs UI

bttf:
Screenshot 2024-02-21 at 3 00 10 PM
Screenshot 2024-02-21 at 3 29 36 PM
Screenshot 2024-02-21 at 3 30 04 PM
Screenshot 2024-02-21 at 3 31 35 PM
Screenshot 2024-02-21 at 4 46 20 PM


packages/front-end/components/Features/FeaturesStats.tsx
```diff
@@ -0,0 +1,148 @@
+import { useRouter } from "next/router";
+import { useMemo } from "react";
+import { FeatureCodeRefsInterface } from "back-end/types/code-refs";
```
mknowlton89, line 3: Test Comment

packages/front-end/components/Features/FeaturesStats.tsx
```diff
@@ -0,0 +1,148 @@
...
+            choice.
+          </p>
+          {hasFeature ? (
+            <Button
+              onClick={async () => {
+                router.push("/settings#configure-code-refs");
+              }}
+            >
+              Go to settings
+            </Button>
```
mknowlton89, line 72: Small nit: I think we should disable the button if the user doesn't have the `organizationalSettings` permission - otherwise the link will take them to a dead end. Perhaps we can wrap it with a ToolTip to indicate they don't have permission?

packages/front-end/components/Features/FeaturesStats.tsx
```diff
@@ -0,0 +1,148 @@
...
+            <PremiumTooltip commercialFeature="code-references">
+              <Button
+                color="warning"
+                onClick={async () => {
+                  router.push("/settings");
+                }}
+              >
+                Upgrade to Pro
+              </Button>
+            </PremiumTooltip>
```
mknowlton89, line 83: Small nit: I don't think routing users to the `/settings` page is the right destination here - the user can't upgrade their plan from that page.

Instead, we have the `UpgradeModal` which has some permission logic built into it. We can also differentiate the button CTA based on the plan. E.G. if the org is `oss` we can prompt them to `Try Enterprise` vs `Upgrade to Pro`.

```
  <button
                    className="btn btn-premium font-weight-normal"
                    onClick={() => setUpgradeModal(true)}
                  >
                    {accountPlan === "oss" ? (
                      <>
                        Try Enterprise <GBPremiumBadge />
                      </>
                    ) : (
                      <>
                        Try Pro <GBPremiumBadge />
                      </>
                    )}
                  </button>
```

[Here is a reference](https://github.com/growthbook/growthbook/blob/main/packages/front-end/components/License/ShowLicenseInfo.tsx) for how we do this on the SideBar.

mknowlton89 (approved):
Left a few small nit picks - I'm comfortable with those being fast follows given the timeline.

bttf:
> Left a few small nit picks - I'm comfortable with those being fast follows given the timeline.

Thanks a bunch! Made those changes as requested.
Screenshot 2024-02-22 at 10 08 19 AM

PR #2077, opened 2024-01-30
Fix bug: Empty metric name rendering as 'id' on front-end

jdorn:
### Features and Changes

Small typo causing empty metrics to render as `id` on the front-end instead of an empty string.


lukesonnet (approved):

lukesonnet (changes requested):
still ignoring `initialOption` and adding a blank option as per message in slack

PR #2078, opened 2024-01-30
Added support for editing the owner email if you are an account admin

Auz:
Screenshot 2024-01-30 at 1 38 12 PM


packages/front-end/components/Settings/EditOrganizationModal.tsx
```diff
@@ -1,41 +1,65 @@
...
       })}
       cta="Save"
     >
       <Field label="Organization Name" required {...form.register("name")} />
+      <Field
+        label={"Owner Email"}
+        type="email"
+        {...form.register("ownerEmail")}
+        disabled={!canEditOwner}
+        title={canEditOwner ? "" : "Only admins can change this"}
```
msamper, line 61: May not need this because users without `organizationSettings` permission can't access the page, but probably better to be safe

msamper (approved):
Lgtm. Not sure about whether or not there's a reason editing the owner email wasn't possible before though. Wonder if this may cause a mismatch issue with the admin account and the owner email we persist

packages/front-end/components/Settings/EditOrganizationModal.tsx
Auz: True, I originally had this split out into a different permission, but decided to combine it with organizationSettings. In case we want to split this again, there is a state for 'can edit name but not owner'

PR #2087, opened 2024-01-31
Add No Access docs and remove FF

mknowlton89:
### Features and Changes

This PR introduces docs updates for the new `No Access` role, and removes the GB FF. Merging this PR will launch this feature to enterprise users, so don't merge until we're ready for that.

### Dependencies

This PR is dependent on all other no access PRs merging in. Currently that is only https://github.com/growthbook/growthbook/pull/2074.

I will do an audit before merging this to ensure we don't have any no access gaps.

### Testing

- [x] Log in as an Enterprise org and ensure the `No Access` role is visible when inviting/updating a team member's role
- [x] Do the same, but ensure you can update a `teams` role.
- [x] Ensure the docs read ok on the staging docs
- [x] Ensure that the `No Access` role is not visible for non-enterprise organizations


packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
```diff
@@ -23,12 +22,10 @@ export default function SingleRoleSelector({
...
-  const growthbook = useGrowthBook();
   const { roles, hasCommercialFeature } = useUser();
   const hasFeature = hasCommercialFeature("advanced-permissions");
-  const isNoAccessRoleEnabled =
-    growthbook?.isOn("no-access-role-type") &&
-    hasCommercialFeature("no-access-role");
+
+  const isNoAccessRoleEnabled = hasCommercialFeature("no-access-role");
 
   let roleOptions = [...roles];
```
mknowlton89, line 31: If we add additional roles that are only available for certain plans, or if we add support for custom roles, we should refactor the `getRoles` call to handle that logic on the backend. But for now, I think it's overkill for this.

docs/docs/account/user-permissions.mdx
```diff
@@ -60,6 +60,27 @@ If your organization has enabled the setting to allow verified users to automati
...
+
+If, however, you only want to hide a certain project from a user, you can assign their project-level role of `No Access` which will make it so the user isn't able to view the project.
+
+If you need to apply these rules to many users as once, you can create a Team with the permissions needed, and then add users to that team. But, keep in mind, the user's permissions will be a combination of their user permissions, and the permissions of any team their on, so after adding the user to a team, you may want to reduce their user-level permissions to ensure their previous user-level permissions don't override the permissions inherited by the team(s) they're on.
+
+:::note
+
+Within GrowthBook, not all resources are project-specific. For example, `Attributes`, `Saved Groups`, `Namespaces`, `Environments`, `Presentations`, and `Ideas` all live at the organization level. So, if a user has a global role of `No Access`, they will not be able to view any of these resources. However, if the user has a project-specific role that does
```
jdorn, line 77: >  So, if a user has a global role of `No Access`, they will not be able to view any of these resources.

Is this true?  I thought since these resources are not project-scoped, that means everyone can read them.  Did we add read filters to all of these data models too?

mknowlton89: You are correct - this was from the original docs draft before I wrote any code. This has been updated.

PR #2092, opened 2024-02-01
Data Model Classes

jdorn:
## Overview
New data model classes to enforce best practices and reduce boilerplate code required to create new models.  For an initial test, this PR converts the `FactMetricModel` to use the new class.

### Example

Everything starts from a single Zod schema, representing the whole object:

```ts
// File: back-end/src/validators/foo.ts
import { z } from "zod";

export fooSchema = z.object({
  id: z.string(),
  organization: z.string(),
  dateCreated: z.date(),
  dateUpdated: z.date(),
  name: z.string(),
  // Use default() for properties that are optional when creating
  projects: z.array(z.string()).default([])
}).strict();
```

From this, you can export types to use throughout the front-end and back-end like we do today.
```ts
// File: back-end/types/foo.d.ts
import { z } from "zod";
import type { fooSchema } from "../src/validators/foo";
import { CreateProps, UpdateProps } from "./models";

// Full interface
export type FooInterface = z.infer<fooSchema>;

// Helper types for the subset of fields that can be set when creating/updating
// For example, you can't update 'id' after it's been created
export type CreateFooProps = CreateProps<FooInterface>;
export type UpdateFooProps = UpdateProps<FooInterface>;
```

Create the data model class based on the schema.
```ts
// File: back-end/src/models/FooModel.ts
import { fooSchema } from "../validators/foo";

export class FooDataModel extends BaseModel<typeof fooSchema> {
  protected getConfig() {
    const config: ModelConfig<typeof fooSchema> = {
      schema: fooSchema,
      collectionName: "foos",
      idPrefix: "foo_",
      auditLog: {
        entity: "foo",
        createEvent: "foo.create",
        updateEvent: "foo.update",
        deleteEvent: "foo.delete",
      },
      // Either "single", "multiple", or "none"
      projectScoping: "multiple",
      // If true, `id` is unique.  If false (default), the `organization`/`id` combo is unique.
      globallyUniqueIds: false,
    };
    return config;
  }

  // CRUD permission checks
  protected canCreate(doc: FooInterface): boolean {
    return this.context.permissions.canCreateFoo(doc);
  }
  protected canRead(doc: FooInterface): boolean {
    return this.context.permissions.canReadFoo(doc);
  }
  protected canUpdate(existing: FooInterface, updates: UpdateProps<FooInterface>): boolean {
    return this.context.permissions.canUpdateFoo(existing, updates);
  }
  protected canDelete(doc: FooInterface): boolean {
    return this.context.permissions.canDeleteFoo(doc);
  }
}
```

Add it to the Request Context class:
```ts
export class ReqContextClass {
  public models!: {
    // ...
    foos: FooDataModel;
  };
  private initModels() {
    this.models = {
      // ...
      foos: new FooDataModel(this),
    };
  }
  // ...
}
```

And then, you can access it anywhere you have context and use the built-in methods:

```ts
const foo = await context.models.foos.getById("foo_123");
if (!foo) throw new Error("Could not find foo");

console.log("Old Name: ", foo.name);

const newFoo = await context.models.foos.update(foo, {name: "New Name"});
console.log("New Name: ", newFoo.name);
```
### Built-in Public Methods

The following public methods are created automatically:
- `getAll()`
- `getAllByProject(project)`
- `getById(id)`
- `getByIds(ids)`
- `create(data)`
- `update(existing, changes)`
- `updateById(id, changes)`
- `delete(existing)`
- `deleteById(id)`

The `create`, `update`, and `updateById` methods accept either strongly typed arguments or `unknown`.  Everything is validated before saving, so this is safe.  This is useful when you want to pass in `req.body` directly without manually doing validation.

### Custom Methods

You can add more tailored data fetching methods as needed by referencing the `_findOne` and `_find` methods.  There are similar protected methods for write operations, although those are rarely needed.

Here's an example:

```ts
export class FooDataModel extends BaseModel<typeof fooSchema> {
  // ...
  
  public getByNames(names: string[]) {
    return this._find({name: {$in: names}});
  }
}
```

Note: Permission checks, migrations, etc. are all done automatically within the `_find` method, so you don't need to repeat any of that in your custom methods.  Also, the `organization` field is automatically added to every query, so it will always be multi-tenant safe.

### Hooks

The following hooks are available, letting you add additional validation or perform trigger-like behavior without messing with data model internals. All of these besides `migrate` are async.  Define these in your child class and they will be called at appropriate times.

- `migrate(legacyObj): newObj`
- `customValidation(obj)` (called for both update/create flows)
- `beforeCreate(newObj)`
- `afterCreate(newObj)`
- `beforeUpdate(existing, updates, newObj)`
- `afterUpdate(existing, updates, newObj)`
- `beforeDelete(existing)`
- `afterDelete(existing)`

Here's an example:

```ts
export class FooDataModel extends BaseModel<typeof fooSchema> {
  // ...
  
  protected async beforeDelete(existing: FooInterface) {
    const barsUsingFoo = await getAllBarsUsingFoo(this.context, existing.id);
    if (bars.length > 0) {
      throw new Error("Cannot delete. One or more bars is linking to this foo");
    }
  }
}
```

### Built-In Best Practices

Includes the following best practices:
- Everything is strongly typed and validated before saving
- Mongo indexes for common access patterns (org + id) plus additional custom indexes
- Automatically check permissions on both read and write
- Multi-tenant safe (auto-adds `organization` to every query)
- Creates audit log entries for all write actions
- Automatically runs JIT migrations
- Keep `dateUpdated` and `dateCreated` up-to-date
- Registers any new tags that are being used
- Makes sure common foreign key references are valid (e.g. projectIds exist)

## Future TODOs
-  Ability to run real migrations (persists result to Mongo)
-  Easier built-in pagination options
-  Helper to generate JSON schemas from zod validators (to reduce REST API boilerplate)
-  Move validator / type exports to `shared` package
-  Remove mongoose completely (still being used to manage the connection)
-  More common foreign reference checks (e.g. `datasource`)
-  Utilities to help with foreign key cascades on delete

## Testing

- [x] Fact metric page and actions within the app
   - [x] List page (with project filtering)
   - [x] Fact metric page
   - [x] Create fact metric
   - [x] Edit fact metric
   - [x] Delete fact metric
   - [x] Refresh experiment results
- [x] Fact metric REST endpoints
   - [x] List
   - [x] Get
   - [x] Create
   - [x] Update
   - [x] Delete
   - [x] Bulk import
- [x] No Access read filter behavior
- [x] Audit log entries
- [x] Permission checks
- [x] Custom Validation
- [x] Migrations
- [x] User data validation
- [x] Mongo Indexes
- [x] dateUpdated / dateCreated
- [x] New tags registered
- [x] Ensure valid project id


bttf:
High-level but is there a restriction or constraint that requires us to have models defined within `req.context`? It would be cleaner IMO to have something like `req.models`. It would also avoid confusion / name-clobbering in certain cases (`req.context.organizations` vs. `req.context.organization`)

jdorn:
> High-level but is there a restriction or constraint that requires us to have models defined within `req.context`? It would be cleaner IMO to have something like `req.models`. It would also avoid confusion / name-clobbering in certain cases (`req.context.organizations` vs. `req.context.organization`)

Nice thing about it being in context is that we can use it from anywhere, not just within request handlers.  I originally put them under `req.context.models`, but it did get pretty verbose.  Open to ideas

bttf:
> Nice thing about it being in context is that we can use it from anywhere, not just within request handlers.

Yeah, that's pretty important.

> I originally put them under `req.context.models`, but it did get pretty verbose. Open to ideas

I actually don't think that's verbose at all. We will likely destructure the context object like so:

```typescript
const someBusinessLogicFn = ({ org, models }: ReqContext, exp: ExperimentInterface) => {
  ...
}
```
Or

```typescript
const someBusinessLogicFn = (context: ReqContext, exp: ExperimentInterface) => {
  const { models, org } = context;
}
```

packages/back-end/src/models/BaseModel.ts
```diff
@@ -0,0 +1,600 @@
...
+    }
+    if ("dateCreated" in props) {
+      throw new Error("Cannot set dateCreated field");
+    }
+    if ("dateUpdated" in props) {
+      throw new Error("Cannot set dateUpdated field");
+    }
+
+    // Add default owner if empty
+    if ("owner" in props && props.owner === "") {
```
bttf, line 269: Could `props.owner` be `undefined` here?

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -331,19 +322,13 @@ export const getFactMetrics = async (
 };
 
 export const postFactMetric = async (
-  req: AuthRequest<CreateFactMetricProps>,
+  req: AuthRequest<unknown>,
```
bttf, line 325: Why do we lose this (`CreateFactMetricProps`)? Couldn't it be obtained from the FactMetricModel class?

packages/back-end/src/services/context.ts
```diff
@@ -16,18 +16,28 @@ import { EventAuditUser } from "../events/event-types";
...
-import { ReqContextInterface } from "../../types/context";
-import { getEnvironmentIdsFromOrg } from "./organizations";
 
-export class ReqContextClass implements ReqContextInterface {
+export class ReqContextClass {
+  // Models
+  public models!: {
+    factMetrics: FactMetricDataModel;
+  };
+  private addModels() {
```
bttf, line 35: Nit - would call it `initModels` to emphasize it only happens once

packages/back-end/types/fact-table.d.ts
```diff
@@ -135,12 +105,8 @@ export type TestFactFilterProps = z.infer<typeof testFactFilterPropsValidator>;
 export type UpdateColumnProps = z.infer<typeof updateColumnPropsValidator>;
 export type CreateColumnProps = z.infer<typeof createColumnPropsValidator>;
 
-export type CreateFactMetricProps = z.infer<
-  typeof createFactMetricPropsValidator
->;
-export type UpdateFactMetricProps = z.infer<
-  typeof updateFactMetricPropsValidator
->;
+export type CreateFactMetricProps = CreateProps<FactMetricInterface>;
```
bttf, line 108: Seems like we just need to use these RE: above question

bttf (approved):
Non-blocking comments / questions

packages/back-end/src/routers/fact-table/fact-table.controller.ts
jdorn: `unknown` is more accurate here since this is raw input data coming directly from users.  We would need to validate it first before knowing that it conforms to the type we expect.  The data model class does this validation already, so everything works fine now.  Changing the type here just prevents people making mistakes in the future and assuming something is validated when it's not.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -0,0 +1,600 @@
...
+import {
+  auditDetailsCreate,
+  auditDetailsDelete,
+  auditDetailsUpdate,
+} from "../services/audit";
+
+export type BaseSchema = z.ZodObject<
+  {
+    id: z.ZodString;
+    organization: z.ZodString;
```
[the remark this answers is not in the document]
jdorn: Yeah, those won't work with the current design.  Maybe we can make a common base class that does core things like managing indexes and querying, but doesn't do any validation or have any assumptions about data structure.

PR #2095, opened 2024-02-02
Update New BigQuery DS Flow - Fetch list of available data sets

mknowlton89:
### Features and Changes

This PR introduces an update to the New Data Source flow, specifically for BigQuery data sources to address the `my_dataset` issue/confusion.

Now, when an organization adds a BigQuery datasource, we have introduced a "Test Connection" button. This button fetches a list of the data sets for a particular BigQuery project. If at least 1 data set is returned, we then allow the user to select from the list of datasets when choosing a default dataset.The select field does allow custom imports, to satisfy the use case where the data set where the experiment assignments will be stored has not been created yet.

If not datasets are returned, the user can manually type in the data set.

The default dataset is now a required field as a part of this change. I've updated the Tooltip text and the `helperText` to try and provide more context about what the user should enter for the `Default Dataset`.

In the event that the organization uses the Auto-Discovery authentication method, we are not surfacing the Test Connection button. 

### Testing

- [x] - Add a new BigQuery datasource and use the "Test Connection" button with a valid key file and ensure it connects successfully, and the dataset options returned are then seeded into the Default Dataset dropdown.
- [x] Do the same, but use a key file connected to a BigQuery project that doesn't have any datasets, and ensure the warning message is displayed, and the user can manually enter a data set.
- [x] Do the same, but use a bad key file, and ensure the error is thrown

### Screenshots
Screen Shot 2024-02-02 at 10 20 53 AM

Successful connection
Screen Shot 2024-02-02 at 10 20 40 AM

Connection error
Screen Shot 2024-02-02 at 10 20 02 AM

Connection successful, but no data sets returned warning
Screen Shot 2024-02-02 at 10 19 16 AM

Experiment Assignment query after selecting a 
Screen Shot 2024-02-02 at 10 21 15 AM

Updated Default Dataset input helper text (only visible when JSON key file authentication method is selected)
Screen Shot 2024-02-02 at 11 49 19 AM

Updated Default Dataset tooltip copy
Screen Shot 2024-02-02 at 11 49 24 AM


august-growthbook:
This is such a big improvement to the user experience! Thank you.

packages/front-end/components/Settings/BigQueryForm.tsx
```diff
@@ -1,15 +1,66 @@
...
+            "We were able to connect to BigQuery, but we weren't able to retreive any datasets in this project.",
+        });
+        return;
+      }
+      setTestConnectionResults({
+        status: "success",
+        datasetOptions: datasets,
+        message: `Connected to ${params.projectId} successfully!`,
+      });
+      //TODO: Intelligently select a default dataset based on the datasource type (manual, ga4, segment, etc)
```
jdorn, line 53: It's worth doing something really simple here now.  Better than nothing and we can improve it over time.

```ts
const analyticsDataset = datasets.find(d => d.match(/^analytics_/));
if (analyticsDataset) {
  setParams({ ["defaultDataset"]: analyticsDataset })
}
```

jdorn (approved):

PR #2097, opened 2024-02-02
Add hashing v1 notifications/selector when SDK doesn't support v2

bryce-fitzsimons:
### Features and Changes
Warn when using v1-only SDK, add hash version select field, default to v1 when v2 is unsupported

image

image


packages/front-end/components/Experiment/EditTargetingModal.tsx
```diff
@@ -80,7 +87,10 @@ export default function EditTargetingModal({
     coverage: lastPhase?.coverage ?? 1,
     hashAttribute: experiment.hashAttribute || "id",
     fallbackAttribute: experiment.fallbackAttribute || "",
-    hashVersion: experiment.hashVersion || 2,
+    hashVersion:
+      experiment.hashVersion || hasSDKWithNoBucketingV2
+        ? 1
+        : (2 as 1 | 2 | undefined),
```
jdorn, line 93: Will this work?  `useSDKConnections` starts out as null, makes an API request, and then gets the real list.  Is that too late when using as a form default value?

bryce-fitzsimons: good point

bryce-fitzsimons: turns out its fine(ish) because experiment pages trigger a fetch of sdk-connections on load which then gets cached via useSWR() and is ready once EditTargetingModal appears.

Could possibly make this more resilient with some mutate/useEffect logic, but not sure its worth it...

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -174,7 +184,7 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
       name: initialValue?.name || "",
       hypothesis: initialValue?.hypothesis || "",
       activationMetric: initialValue?.activationMetric || "",
-      hashVersion: initialValue?.hashVersion || 2,
+      hashVersion: initialValue?.hashVersion || hasSDKWithNoBucketingV2 ? 1 : 2,
```
jdorn, line 187: I think this is missing parentheses.
```suggestion
      hashVersion: initialValue?.hashVersion || (hasSDKWithNoBucketingV2 ? 1 : 2),
```

packages/front-end/components/Features/RuleModal.tsx
```diff
@@ -152,6 +155,13 @@ export default function RuleModal({
   const experimentId = form.watch("experimentId");
   const selectedExperiment = experimentsMap.get(experimentId) || null;
 
+  const { data: sdkConnectionsData } = useSDKConnections();
+  const hasSDKWithNoBucketingV2 = !getConnectionsSDKCapabilities({
+    connections: sdkConnectionsData?.connections ?? [],
+    mustMatchAllConnections: true,
+    project: feature.project,
+  }).includes("bucketingV2");
```
jdorn, line 163: This same code is copy/pasted 3 times.  Can we abstract this?

bryce-fitzsimons: sure

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -149,6 +153,12 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
   const settings = useOrgSettings();
   const { refreshWatching } = useWatching();
 
+  const { data: sdkConnectionsData } = useSDKConnections();
+  const hasSDKWithNoBucketingV2 = !allConnectionsSupportBucketingV2(
+    sdkConnectionsData?.connections,
+    project
+  );
```
jdorn, line 160: In the onboarding flow for experiments, we don't prompt them to create their first SDK connection until they are ready to start their experiment.  So `sdkConnectionsData?.connections` will be an empty array, the check below will fail, and it will set hashing to v1, which doesn't seem right.

packages/front-end/components/Experiment/HashVersionSelector.tsx
```diff
@@ -42,33 +49,66 @@ export function HashVersionTooltip({ children }: { children: ReactNode }) {
...
+        }
+      />
+
+      {hasSDKWithNoBucketingV2 && (
+        <div className="mt-2 alert alert-warning">
+          <FaExclamationTriangle className="mr-1" />
+          Some of your{" "}
+          <a href="/sdks" target="_blank">
+            SDK Connections <FaExternalLinkAlt />
+          </a>{" "}
```
jdorn, line 96: Is this link actionable?  How are they supposed to know what to do when they land on this page?  Would be better to maybe link to a docs page instead that explains the issue and tells people how to update.

bryce-fitzsimons: removed the link. discussed later maybe including a `?capabilities` query string that would highlight connections with issues

PR #2108, opened 2024-02-08
Removes separate user prop from function signature when context is also a prop

mknowlton89:
This is a general cleanup PR. When we first introduced the backend `context` object, it didn't contain the `auditUser`. However, it was later added in [PR 2007](https://github.com/growthbook/growthbook/pull/2007). This PR updates the function signatures whenever we were passing in both the `context` object and the `user: EventAuditUser`, as we can now get that from the `context` object.

This PR also updates the constructor function signature for the QueryRunner service so that the `context` object is the first param, which is a pattern we've also defined.

### Testing
- [x] - Ensure the build builds as expected
- [x] Create a feature flag
- [x] Create an experiment
- [x] Create a metric
- [x] Update an existing feature flag
- [x] Update an existing metric
- [x] Generally move around the app and ensure no regressions were introduced

### DIFF
--- a/packages/back-end/src/queryRunners/QueryRunner.ts
+++ b/packages/back-end/src/queryRunners/QueryRunner.ts
@@ -105,9 +105,9 @@ export abstract class QueryRunner<
   private useCache: boolean;
 
   public constructor(
+    context: ReqContext | ApiReqContext,
     model: Model,
     integration: SourceIntegrationInterface,
-    context: ReqContext | ApiReqContext,
     useCache = true
   ) {
     this.model = model;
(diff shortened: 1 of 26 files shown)


packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -105,9 +105,9 @@ export abstract class QueryRunner<
   private useCache: boolean;
 
   public constructor(
+    context: ReqContext | ApiReqContext,
```
mknowlton89, line 108: Small clean up - we've made the design decision to always ensure `context` is the first prop passed in.

bttf (approved):

PR #2114, opened 2024-02-12
New SDK Connection Languages: Shopify, Webflow, and Wordpress

jdorn:
### Features and Changes

Adds a new section of SDK Connection "Languages" for no/low code platforms.  Includes Shopify, Webflow, Wordpress, and a Generic No Code option.

image

The instructions all use the auto script tag for implementation and are tailored for the specific platform.  Here's an example of the instructions for Shopify:

image

We hide the attribute and tracking callback instructions unless the user has a non-standard setup.  This avoids adding any confusion for the majority of users.


august-growthbook:
Yes yes yes! This is going to significantly improve the UX.

I know the `theme.liquid` files will vary, but is it possible to include a screenshot of the GrowthBook code snippet in the correct spot in the `theme.liquid` file? It's a small nicety we can provide that will help customers who have never seen HTML/JS before feel more confident in setting up the integration.

PR #2117, opened 2024-02-13
Allow Super Admins to view organizations they are not members of

jdorn:
### Features and Changes

Super admins for multi-org deployments will now get `readonly` access to organizations where they have not explicitly been invited.  If the super admin is invited to an org, they will keep whatever role they are invited with.  The fallback role for super admins defaults to `readonly`, but can be configured via an environment variable.

This also fixes an issue where upon refreshing the page  when logged into an org a superAdmin is not part of it, it would jump back to the first org. 

### Test Plan
`yarn test`
`yarn dev`
In .env.local set
```
IS_MULTI_ORG=true
IS_CLOUD=true
SSO=valid SSO config
```

In Mongo set a user to have superAdmin who is not a member of every organization.
Login as that user
Go to localhost:3000/admin
Click on the org that they are not a member of
See that they can still see an experiment
See that they can't see General or other settings, since they should have read only.
Refresh the page, and see that it stays in the same org as was selected.


mknowlton89 (approved):

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
```diff
@@ -137,10 +137,6 @@ export default function authenticateApiRequestMiddleware(
         project?: string | (string | undefined)[] | undefined,
         envs?: string[] | Set<string>
       ) => {
-        // Super admins have full access to every organization
-        if (req.user?.superAdmin) {
-          return;
-        }
```
bttf, line 143: potential regression: it seems like superAdmins who are members no longer have super admin rights when accessing the API (instead, defaults to member permissions)

packages/back-end/test/middleware/authenticateApiRequestMiddleware.test.ts
```diff
@@ -201,8 +207,49 @@ describe("REST API auth middleware", () => {
...
+          ...userKey,
+          userId: "u_unknown_id",
+        },
+        permission,
+        organization,
+        environments,
+        project,
+        teams: [],
+        superAdmin: true,
+      });
```
bttf, line 252: add'l test: it should allow superAdmins full permissions/member permissions if they are a part of org

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
tzjames: I think it is fine to regress here. In general superAdmins probably already have admin role in the orgs they are part of.

packages/front-end/services/UserContext.tsx
```diff
@@ -268,23 +264,19 @@ export function UserContextProvider({ children }: { children: ReactNode }) {
       environments: [],
       limitAccessByEnvironment: false,
       name: data.userName,
-      role: data.superAdmin ? "admin" : "readonly",
+      role: "readonly",
```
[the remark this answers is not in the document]
tzjames: Yes you were correct.  Thanks for the catch.  This is only currently used for telemetry, but worth the fix.

PR #2120, opened 2024-02-14
Approval Flows

gazzdingo:
### Features and Changes


packages/front-end/pages/settings/index.tsx
```diff
@@ -1576,6 +1577,20 @@ const GeneralSettingsPage = (): React.ReactElement => {
...
+                </div>
+                <div>
+                  <Toggle
+                    id={"toggle-require-reviews"}
+                    value={!!form.watch("requireReviews")}
+                    setValue={(value) => {
+                      form.setValue("requireReviews", value);
+                    }}
+                  />
+                </div>
```
jdorn, line 1593: This setting should be enterprise-only.  Need to define a new commercial feature for "require-approvals" in `license.ts` and assign it to the enterprise plan.

packages/back-end/src/controllers/features.ts
```diff
@@ -513,6 +518,71 @@ export async function postFeatureRebase(
...
+  const feature = await getFeature(org.id, id);
+  if (!feature) {
+    throw new Error("Could not find feature");
+  }
+
+  const revision = await getRevision(org.id, feature.id, parseInt(version));
+  if (!revision) {
+    throw new Error("Could not find feature revision");
+  }
+  if (revision.status === "draft") {
```
jdorn, line 573: Aren't there other statuses like `published` or `discarded` that we also want to block here?   Might be easier to whitelist the statuses we do allow instead.

packages/back-end/src/controllers/features.ts
```diff
@@ -535,9 +605,17 @@ export async function postFeaturePublish(
   req.checkPermissions("manageFeatures", feature.project);
 
   const revision = await getRevision(org.id, feature.id, parseInt(version));
+  const user = org.members.find((member) => member.id === userId);
+  const isAdmin = user?.role === "admin";
```
jdorn, line 609: Instead of checking the role directly, we should make a new permission called `bypassApprovalChecks`, grant that to the admin role, and then check the permission here.  We are going to add support for custom roles soon, so want to avoid referencing roles directly.

Also, users can be part of teams, so the roles get complicated.  The logic here is only checking their individual global role, which won't work.  Switching to permission checks will fix this too.

packages/back-end/src/controllers/features.ts
```diff
@@ -535,9 +605,17 @@ export async function postFeaturePublish(
...
   if (!revision) {
     throw new Error("Could not find feature revision");
   }
+  if (
+    (org.settings?.requireReviews && !isAdmin) ||
+    revision.status === "approved"
```
jdorn, line 615: This seems backwards?  If the revision is approved, shouldn't publishing be allowed?

packages/back-end/src/controllers/features.ts
```diff
@@ -723,6 +801,7 @@ export async function postFeatureFork(
     baseVersion: revision.version,
     changes: revision,
     environments,
+    requiresReview: !!org.settings?.requireReviews && !!req.superAdmin,
```
jdorn, line 804: Don't check `superAdmin`.  That's a special setting only for multi-org deployments and not relevant here.

packages/back-end/src/controllers/features.ts
```diff
@@ -762,12 +841,22 @@ export async function postFeatureDiscard(
...
     });
   }
 
+  const hasPendingReview = await pendingReview(org.id, feature, [
+    revision.version,
+  ]);
+  if (!hasPendingReview) {
+    await updateFeature(org, res.locals.eventAudit, feature, {
+      pendingReview: false,
+    });
```
jdorn, line 857: Instead of a 2nd `updateFeature` call, can we combine this with the `hasDrafts` logic above so it only has to hit Mongo once?

packages/back-end/src/controllers/features.ts
```diff
@@ -974,7 +1063,13 @@ async function getDraftRevision(
   if (!revision) {
     throw new Error("Cannot find revision");
   }
-  if (revision.status !== "draft") {
+  if (
+    !(
+      revision.status === "draft" ||
+      revision.status === "pending-review" ||
+      revision.status === "reviewed"
```
jdorn, line 1070: Are these statuses correct?  From the typedef, it looks like `reviewed` is not one of the options, but `approved` and `changes-requested` are.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -210,18 +231,23 @@ export async function createRevision({
...
+    status = "pending-review";
+  }
   const doc = await FeatureRevisionModel.create({
     organization: feature.organization,
     featureId: feature.id,
     version: newVersion,
     dateCreated: new Date(),
     dateUpdated: new Date(),
-    datePublished: publish ? new Date() : null,
+    datePublished: publish && requiresReview ? new Date() : null,
```
jdorn, line 246: This seems backwards.  Should be `!requiredReview`, right?  Or perhaps more clearly:

```ts
status === "published" ? new Date() : null
```

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -210,18 +231,23 @@ export async function createRevision({
...
     createdBy: user,
     baseVersion: baseVersion || feature.version,
-    status: publish ? "published" : "draft",
-    publishedBy: publish ? user : null,
+    status,
+    publishedBy: publish && requiresReview ? user : null,
```
jdorn, line 250: Same here, switch to checking if the status is `published` so we can this "requiresReview" logic in one place.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -243,7 +269,13 @@ export async function updateRevision(
 ) {
   // If editing defaultValue or rules, require the revision to be a draft
   if ("defaultValue" in changes || changes.rules) {
-    if (revision.status !== "draft") {
+    if (
+      !(
+        revision.status === "draft" ||
+        revision.status === "pending-review" ||
+        revision.status === "reviewed"
```
jdorn, line 276: Same here with the status.  I don't think `reviewed` is one of the statuses - need to change to `approved` and/or `changes-requested`, right?

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -302,12 +334,98 @@ export async function markRevisionAsPublished(
...
+    {
+      organization: revision.organization,
+      featureId: revision.featureId,
+      version: revision.version,
+    },
+    {
+      $set: {
+        status: "pending-review",
+        publishedBy: user,
+        datePublished: new Date(),
```
jdorn, line 362: Should these publish settings be updated here if a review is just being requested?  Nothing is being published yet.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -302,12 +334,98 @@ export async function markRevisionAsPublished(
...
+  }
+  const log: RevisionLog = {
+    action,
+    subject: "",
+    timestamp: new Date(),
+    user,
+    value: JSON.stringify(comment ? { comment } : {}),
+  };
+  if (status === "approved") {
+    await migrateToApproved(revision.organization, revision.featureId);
```
jdorn, line 399: What if the feature has other drafts still in a `pending-review` state?

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -302,12 +334,98 @@ export async function markRevisionAsPublished(
...
+    {
+      organization: revision.organization,
+      featureId: revision.featureId,
+      version: revision.version,
+    },
+    {
+      $set: {
+        status,
+        publishedBy: user,
+        datePublished: new Date(),
```
jdorn, line 412: Same here - we should probably not set these fields since nothing is being published here, it's just being reviewed.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -302,12 +334,98 @@ export async function markRevisionAsPublished(
...
+      featureId: revision.featureId,
+      version: revision.version,
+    },
+    {
+      $set: {
+        status: "pending-review",
+        publishedBy: user,
+        datePublished: new Date(),
+        dateUpdated: new Date(),
+        comment: comment ?? revision.comment,
```
jdorn, line 364: For both of these new functions, I'm not sure if we should be updating the revision comment.  If someone reviews and says `lgtm` as the comment, we wouldn't want to override the top-level revision's comment that describes the changes in details.  The review comment already exists in the log

packages/front-end/components/Features/requestReviewModal.tsx
```diff
@@ -0,0 +1,371 @@
+import { FeatureInterface } from "back-end/types/feature";
```
jdorn, line 1: The filename for components should start with a capital letter

packages/front-end/components/Features/RevisionLog.tsx
```diff
@@ -27,24 +37,48 @@ function RevisionLogRow({ log, first }: { log: RevisionLog; first: boolean }) {
...
-          {log.action} {log.subject}
+          <>
+            {log.action} {log.subject}
+          </>
         </h3>
-        <div className="ml-auto">
-          {open ? <FaAngleDown /> : <FaAngleRight />}
-        </div>
+        {commentsOnly ? (
+          <div>{openContent()}</div>
```
jdorn, line 73: This is causing it to show the full JSON details for all revision changes when requesting a review, which takes up a ton of vertical space

packages/front-end/components/Features/requestReviewModal.tsx
```diff
@@ -0,0 +1,371 @@
...
+            {isAdmin && (
+              <Text as="label" size="2" className="mt-2">
+                <Flex gap="2">
+                  <Checkbox
+                    onCheckedChange={(checkedState) => {
+                      checkedState === "indeterminate"
+                        ? setAdminPublish(false)
+                        : setAdminPublish(checkedState);
+                    }}
+                  />
```
jdorn, line 235: This has some weird behavior.  When I first check this, it flashes for a second and then goes back to unchecked, but the  "publish" button shows up

packages/front-end/pages/features/[fid].tsx
```diff
@@ -166,7 +172,9 @@ export default function FeaturePage() {
     }
 
     // If there's an active draft, show that by default, otherwise show the live version
-    const draft = data.revisions.find((r) => r.status === "draft");
+    const draft = data.revisions.find(
+      (r) => r.status === "draft" || r.status === "approved"
```
Comment (jdorn) [inline packages/front-end/pages/features/[fid].tsx:176]: This causes some weird behavior.   Like if I request a review, and someone goes to the feature page, it will default to the live revision and it's not obvious that there's any action required.  Same if changes are requested

packages/back-end/src/models/FeatureRevisionModel.ts
gazzdingo: i feel like we should state that it was approved on the lastest one or do you think that we should just still have it in pending review

packages/back-end/src/controllers/features.ts
```diff
@@ -542,10 +635,28 @@ export async function postFeaturePublish(
...
-  if (revision.status !== "draft") {
+  if (
+    (!adminOverride && org.settings?.requireReviews) ||
+    revision.status !== "approved"
+  ) {
+    throw new Error("needs review before publishing");
+  }
+  if (
+    !org.settings?.requireReviews &&
+    !reviewStatuses.includes(revision.status)
```
gazzdingo, line 658: TS seemed to have a weird bug that was saying it was unintentional. because of checking approved above

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -118,6 +138,23 @@ export async function getRevision(
   return doc ? toInterface(doc) : null;
 }
 
+export async function getRevisionsByStatus(
+  context: ReqContext,
+  statuses: string[]
+) {
+  const revisions = await FeatureRevisionModel.find({
+    id: context.org.id,
+    status: { $in: statuses },
```
jdorn, line 147: Every single query we run must always filter by `organization`.  This is critical for security since it prevents us from leaking information between organizations.

packages/back-end/src/controllers/features.ts
```diff
@@ -1211,6 +1331,40 @@ export async function postFeatureMoveRule(
...
+    revision: FeatureRevisionInterface;
+    feature: FeatureInterface | null;
+  }[] = [];
+
+  for (const r of revisions) {
+    if (r) {
+      const feature = await getFeature(context, r.featureId);
+      // we need to filter out features that are created after the revision
+      // eg someone deletes a feature with revisions and creates a new one with the same id
+      if (feature && feature.dateCreated.getTime() < r.dateCreated.getTime()) {
```
jdorn, line 1355: Use `<=` instead of just `<`.  Otherwise, I think there's a chance the very first revision might be missing since that gets created at the same time as the feature.

packages/front-end/styles/global.scss
```diff
@@ -3125,6 +3125,28 @@ input:disabled {
...
+}
+
+.callout {
+  padding: 16px;
+  border-radius: 4px;
+  &.callout-color-gray {
+    background-color: rgba(245, 247, 250, 1);
+  }
+  &.callout-color-amber {
+    background: rgba(235, 166, 0, 0.17);
```
jdorn, line 3141: Are these colors safe for dark mode?  Typically we use CSS variables and change them depending on the theme

gazzdingo: they are based off the radix colors that they have and i believe that they are. i can double check that

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -104,6 +109,21 @@ export async function hasDraft(
   return doc ? true : false;
 }
 
+export async function pendingReview(
```
jdorn, line 112: Is this function still being used?  If not, remove it

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -293,6 +344,86 @@ export async function markRevisionAsPublished(
...
+    {
+      organization: revision.organization,
+      featureId: revision.featureId,
+      version: revision.version,
+    },
+    {
+      $set: {
+        status: "pending-review",
+        datePublished: null,
+        dateUpdated: null,
```
jdorn, line 379: Why set `dateUpdated` to null?

packages/back-end/src/controllers/features.ts
```diff
@@ -520,20 +527,106 @@ export async function postFeatureRebase(
...
+  >,
+  res: Response
+) {
+  const context = getContextFromReq(req);
+  const { id, version } = req.params;
+  const { comment } = req.body;
+  const feature = await getFeature(context, id);
+  if (!feature) {
+    throw new Error("Could not find feature");
+  }
```
jdorn, line 545: I think we want a permission check here.  Maybe the `createFeatureDrafts` permission?

packages/back-end/src/controllers/features.ts
```diff
@@ -520,20 +527,106 @@ export async function postFeatureRebase(
...
+  res: Response
+) {
+  const context = getContextFromReq(req);
+  const { id, version } = req.params;
+  const { comment, review = "Comment" } = req.body;
+  const feature = await getFeature(context, id);
+  if (!feature) {
+    throw new Error("Could not find feature");
+  }
+  req.checkPermissions("canReview");
```
jdorn, line 581: This permission check should be project-scoped depending on what project the feature is in.

gazzdingo: do you have an example of a project scoped?

packages/back-end/src/controllers/features.ts
```diff
@@ -543,10 +636,29 @@ export async function postFeaturePublish(
...
 
   const revision = await getRevision(org.id, feature.id, parseInt(version));
+  const reviewStatuses = [
+    "pending-review",
+    "changes-requested",
+    "draft",
+    "approved",
+  ];
+  if (adminOverride && org.settings?.requireReviews) {
+    req.checkPermissions("bypassApprovalChecks");
```
jdorn, line 646: Same here, this permission check should be project-scoped

packages/back-end/src/controllers/features.ts
```diff
@@ -1211,6 +1332,43 @@ export async function postFeatureMoveRule(
...
+    "pending-review",
+  ]);
+  const featuresAndRevisions: {
+    revision: FeatureRevisionInterface;
+    feature: FeatureInterface | null;
+  }[] = [];
+
+  for (const revision of revisions) {
+    if (revision) {
+      const feature = await getFeature(context, revision.featureId);
```
jdorn, line 1353: Don't run Mongo queries inside of loops.  In this particular case, it might be better to move this filtering logic to the front-end instead.  The front-end already has all of the feature objects in memory, so it won't require any additional db lookups or API calls.

packages/front-end/pages/features/FeaturesDraftTable.tsx
```diff
@@ -44,26 +46,37 @@ export default function FeaturesDraftTable() {
...
+      if (feature && feature?.dateCreated <= revision.dateCreated) {
+        result.push({
+          ...revision,
+          feature,
+        });
+      }
+      return result;
+    },
+    []
   );
```
jdorn, line 62: Move this inside of the `useAddComputedFields` function instead.  Will be much cleaner and memoized correctly.

```ts
const revisions = useAddComputedFields(data?.revisions, (revision) => {
  const feature = features.find(f => f.id === revision.featureId);
  return {
     // ...
  }
}, [features]);
```

Also, notice I'm also passing in `features` as a dependency to this hook

jdorn (approved):

PR #2126, opened 2024-02-15
Introduces the concept of read only type permissions

mknowlton89:
### Features and Changes

This PR introduces an update to our permission check logic (both on the front end and backend). 

Previously, when evaluating whether a user had permission to perform a certain action, we'd go through the following logic:
 - If the permission was a `GLOBAL_SCOPED_PERMISSION`, we'd look to see if the user's global role gave them permission
 - If the permission was a 'PROJECT_SCOPED_PERMISSION' we'd look to see if either the user's global role gave them permission. If not, we'd loop through all of the user's project-specific permissions and if the user had the permission for EVERY project, we'd grant them permission.
 
 This worked well, but there are a few cases where we want to give the user permission for a `PROJECT_SCOPED_PERMISSION` if atleast 1 of their project-specific permissions give them access.
 
 Notably, these are `readData`, `viewEvents`, and `runQueries`.
 
 So now, when evaluating these permissions, if the user's global role doesn't give them permission, but atleast 1 of their project specific permissions does, we'll grant them access.
 
 As part of this PR, I've also updated a lot of `permission.check()` and `req.checkPermission()` calls to ensure that when checking permissions for resources that contain a `projects` array (datasources, metrics, etc) we always pass in the resource's `projects` array. Passing a string into the second argument of `req.checkPermissions` means that we're only evaluating the user's permissions via their global role.

### Testing

- [x] Scenario 1:

User’s global role is no access, but they have a project specific role of experimenter for a single project

On a datasource id page, if the datasource is in “All Projects” the user should not have edit capabilities

On a datasource id page, if the datasource is ONLY in the project the user has the experimenter role for, they should have edit capabilities

On a datasource id page, if the datasource is in multiple projects, but they don’t have explicit `experimenter` role for all of the projects, they should not have edit capabilities

If this users goes to the /experiments page, and attempts to import a previous experiment, they should be able to do so as long as the datasource is in “ALL PROJECTS” or the datasource is associated with at least 1 project that the user has experimenter permissions for (note: for datasources that have a projects array and that array doesn’t overlap with any projects the user has “readData” permissions for, the user shouldn’t even see the datasource in the list 

- [x] Scenario 2: 

User’s global role is experimenter, but they have a project specific role of no access for 1 project

On a datasource id page, if the datasource is in “All Projects” the user should have edit capabilities

On a datasource id page, if the datasource is ONLY in the project they have a no access role for, they shouldn’t even see it

On a datasource id page, if the datasource is in multiple projects, one of which is the project they have no access for, they should see it, but not be able to edit

If this users goes to the /experiments page, and attempts to import a previous experiment, they should be able to run the query. If a datasource has a projects array that ONLY contains the project the user has the `no access` role for, they shouldn't even see the datasource in the dropdown

- [x] Scenario 3:

User’s global role is experimenter, and they have no project-specific roles

On a datasource id page, if the datasource is in “All Projects” the user should have edit capabilities

On a datasource id page, if the datasource is a part of a project the user should have edit capabilities

On a datasource id page, if the datasource is in multiple projects, the user should have edit capabilities

### DIFF
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -513,7 +513,7 @@ export async function putDataSource(
     });
     return;
   }
-
+  // Require higher permissions to change connection settings vs updating query settings
   const permissionLevel = params
     ? "createDatasources"
     : "editDatasourceSettings";
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -514,7 +514,10 @@ export async function putDataSource(
     return;
   }
 
-  req.checkPermissions("editDatasourceSettings", datasource.projects);
+  req.checkPermissions(
+    "editDatasourceSettings",
+    datasource?.projects?.length ? datasource.projects : []
+  );
 
   if (type && type !== datasource.type) {
     res.status(400).json({
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -514,9 +514,12 @@ export async function putDataSource(
     return;
   }
 
+  const permissionLevel = params
+    ? "createDatasources"
+    : "editDatasourceSettings";
   req.checkPermissions(
-    "editDatasourceSettings",
-    datasource?.projects?.length ? datasource.projects : []
+    permissionLevel,
+    datasource?.projects?.length ? datasource.projects : ""
   );
 
   if (type && type !== datasource.type) {
@@ -570,7 +573,7 @@ export async function putDataSource(
     }
 
     if (updates?.projects?.length) {
-      req.checkPermissions("editDatasourceSettings", updates.projects);
+      req.checkPermissions(permissionLevel, updates.projects);
     }
 
     // If the connection params changed, re-validate the connection
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -513,14 +513,8 @@ export async function putDataSource(
     });
     return;
   }
-  // Require higher permissions to change connection settings vs updating query settings
-  const permissionLevel = params
-    ? "createDatasources"
-    : "editDatasourceSettings";
-  req.checkPermissions(
-    permissionLevel,
-    datasource?.projects?.length ? datasource.projects : ""
-  );
+
+  req.checkPermissions("editDatasourceSettings", datasource.projects);
 
   if (type && type !== datasource.type) {
     res.status(400).json({
@@ -573,7 +567,7 @@ export async function putDataSource(
     }
 
     if (updates?.projects?.length) {
-      req.checkPermissions(permissionLevel, updates.projects);
+      req.checkPermissions("editDatasourceSettings", updates.projects);
     }
 
     // If the connection params changed, re-validate the connection
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -2130,7 +2130,10 @@ export async function postPastExperiments(
   if (!datasourceObj) {
     throw new Error("Could not find datasource");
   }
-  req.checkPermissions("runQueries", datasourceObj.projects);
+  req.checkPermissions(
+    "runQueries",
+    datasourceObj?.projects?.length ? datasourceObj.projects : []
+  );
 
   const integration = getSourceIntegrationObject(datasourceObj, true);
 
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -2130,10 +2130,7 @@ export async function postPastExperiments(
   if (!datasourceObj) {
     throw new Error("Could not find datasource");
   }
-  req.checkPermissions(
-    "runQueries",
-    datasourceObj?.projects?.length ? datasourceObj.projects : ""
-  );
+  req.checkPermissions("runQueries", datasourceObj.projects);
 
   const integration = getSourceIntegrationObject(datasourceObj, true);
 
--- a/packages/back-end/src/services/auth/index.ts
+++ b/packages/back-end/src/services/auth/index.ts
@@ -103,6 +103,14 @@ export async function processJWT(
     }
 
     if (READ_ONLY_PERMISSIONS.includes(permission)) {
+      if (
+        projects.length === 1 &&
+        projects[0] === undefined &&
+        Object.keys(userPermissions.projects).length
+      ) {
+        // check to see if the user has permission globally or via any of their project specific permissions
+        projects.push(...Object.keys(userPermissions.projects));
+      }
       // Read only type permissions grant permission if the user has the permission globally or in atleast 1 project
       return projects.some((p) =>
         hasPermission(userPermissions, permission, p, envs)
@@ -130,12 +138,7 @@ export async function processJWT(
     );
     let checkProjects: (string | undefined)[];
     if (Array.isArray(project)) {
-      checkProjects =
-        project.length > 0
-          ? project
-          : Object.keys(userPermissions.projects).length
-          ? Object.keys(userPermissions.projects)
-          : [undefined];
+      checkProjects = project.length > 0 ? project : [undefined];
     } else {
       checkProjects = [project];
     }
--- a/packages/back-end/src/services/auth/index.ts
+++ b/packages/back-end/src/services/auth/index.ts
@@ -25,7 +25,6 @@ import {
   EventAuditUserLoggedIn,
 } from "../../events/event-types";
 import { insertAudit } from "../../models/AuditModel";
-import { TeamInterface } from "../../../types/team";
 import { getTeamsForOrganization } from "../../models/TeamModel";
 import { initializeLicense } from "../licenseData";
 import { AuthConnection } from "./AuthConnection";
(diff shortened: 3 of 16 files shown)


packages/back-end/src/controllers/experiments.ts
```diff
@@ -2130,10 +2130,7 @@ export async function postPastExperiments(
   if (!datasourceObj) {
     throw new Error("Could not find datasource");
   }
-  req.checkPermissions(
-    "runQueries",
-    datasourceObj?.projects?.length ? datasourceObj.projects : ""
-  );
+  req.checkPermissions("runQueries", datasourceObj.projects);
```
mknowlton89, line 2133: For resources that have a projects array, passing in an empty array indicates that it's a part of "ALL PROJECTS". We should only pass in an empty string if we want to check the user's global permissions, or for resources that have a single `project` field like features and experiments.

packages/back-end/src/services/auth/index.ts
```diff
@@ -118,23 +123,31 @@ export async function processJWT(
...
+    const userPermissions = getUserPermissions(
+      req.userId,
+      req.organization,
+      req.teams
+    );
     let checkProjects: (string | undefined)[];
     if (Array.isArray(project)) {
-      checkProjects = project.length > 0 ? project : [undefined];
+      checkProjects =
+        project.length > 0 ? project : Object.keys(userPermissions.projects);
```
mknowlton89, line 136: Now, instead of passing in an `[undefined]` array, if `project` is an empty array, we pass in all projects the user has project-specific permissions for.

packages/back-end/src/services/auth/index.ts
```diff
@@ -101,15 +104,17 @@ export async function processJWT(
...
+      // Read only type permissions grant permission if the user has the permission globally or in atleast 1 project
+      return projects.some((p) =>
+        hasPermission(userPermissions, permission, p, envs)
+      );
+    } else {
+      // All other permissions require the user to have the permission globally or the user must have the permission in every project they have specific permissions for
+      return projects.every((p) =>
+        hasPermission(userPermissions, permission, p, envs)
+      );
+    }
```
mknowlton89, line 117: This is the meat of the change. We introduced the concept of `READ_ONLY_PERMISSIONS` - currently those are: readData, runQueries, and viewEvents. When evaluating whether or not the user has permission to do these events, they only have to have the permission in one of the projects they have specific permissions for (or they need to have it globally).

Whereas the remaining permissions require the user to have the permission either globally, or for EVERY project they have project specific permissions for

packages/back-end/src/services/auth/index.ts
mknowlton89, line 136: Now, instead of passing in an [undefined] array, if project is an empty array, we pass in all projects the user has project-specific permissions for.

packages/front-end/pages/metrics.tsx
```diff
@@ -371,122 +371,130 @@ const MetricsPage = (): React.ReactElement => {
...
-              </td>
-              <td
-                style={{ cursor: "initial" }}
+          {items.map((metric) => {
+            const showMoreMenu: boolean =
+              (metric.onDuplicate && editMetricsPermissions[metric.id]) ||
+              (!metric.managedBy &&
+                metric.onArchive &&
+                editMetricsPermissions[metric.id]) ||
+              false;
```
mknowlton89, line 380: Previously, we were always showing the `<MoreMenu/>` component, but in a lot of cases, the modal that opened as a result was empty. This resolves that, so we only show it if there is at least 1 item inside.

packages/back-end/src/services/auth/index.ts
mknowlton89: Currently working through 1 issue - for user's who have a global role of `Experimenter`, but have a project-specific role for atleast 1 project of `No Access` the permissions check is failing for `runQueries` when the resource is in `ALL PROJECTS`, as when the `projects` prop is an empty array, we're looping over the user's projects, and adding that to the array. 

And in this case, the permissions for that project is `No Access`.

packages/front-end/pages/datasources/[did].tsx
```diff
@@ -59,7 +58,7 @@ const DataSourcePage: FC = () => {
 
   const canEdit =
     (d &&
-      checkDatasourceProjectPermissions(d, permissions, "createDatasources") &&
+      permissions.check("editDatasourceSettings", d.projects) &&
```
Comment (jdorn) [inline packages/front-end/pages/datasources/[did].tsx:61]: Is this change intentional?  Changing the permission that's being checked

packages/front-end/pages/metrics.tsx
jdorn: This is going to be hard to maintain.  A better approach is to create an array of MoreMenu links.  If that array is empty, then skip rendering the menu.

```tsx
const moreMenuLinks: ReactElement[] = [];

if (metric.onDuplicate && editMetricsPermissions[metric.id]) {
  moreMenuLinks.push(<button key="duplicate" ...>)
}
```

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -28,6 +25,45 @@ export function hasPermission(
...
+  }
+
+  let checkProjects: (string | undefined)[];
+  if (Array.isArray(project)) {
+    checkProjects = project.length > 0 ? project : [undefined];
+  } else {
+    checkProjects = [project];
+  }
+
+  if (["readData", "viewEvents", "runQueries"].includes(permission)) {
```
jdorn, line 46: Don't hard code this list here.  Put it in the organization.util file where the permissions are currently defined.  When we need to add a new permission, we don't want to have to remember to edit 2 different files in different packages.

mknowlton89: @jdorn That's where I had it originally, but in doing so I got a compiler error when trying to import from that file in the `shared` directory.

packages/front-end/pages/datasources/[did].tsx
Comment (mknowlton89) [reply packages/front-end/pages/datasources/[did].tsx:61]: It was - to me this feels like the more appropriate permission to check.

Checking against `createDatasources` mean that only `admins` will be able to edit a data source.

Looking at all of the components of the `[did]` page (things like "ExperimentAssignmentQueries`, `DataSourceInlineEditIdentifierTypes`, etc - those all check for against `editDatasourceSettings`. So having this check against `createDatasources` means we never rended the `Edit` button, but if the user was able to open the components, they would then have the ability to edit.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -513,14 +513,8 @@ export async function putDataSource(
...
-  // Require higher permissions to change connection settings vs updating query settings
-  const permissionLevel = params
-    ? "createDatasources"
-    : "editDatasourceSettings";
-  req.checkPermissions(
-    permissionLevel,
-    datasource?.projects?.length ? datasource.projects : ""
-  );
+
+  req.checkPermissions("editDatasourceSettings", datasource.projects);
```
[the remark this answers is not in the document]
mknowlton89: It was - the frontend of `[did].tsx` was checking for `createDatasources` to determine whether or not to render the `Edit Buttons`.

Once that was updated to check against the `editDatasourceSettings` permission, this was causing trouble as `params` were always included in the payload, even if they didn't change, so this ternary never evaluated to `editDatasourceSettings`.

But, before the updates to `[did].tsx` were made, this issue never surfaced since we were checking for `createDatasource` permission there.

packages/front-end/pages/metrics.tsx
mknowlton89: oof - yes, much better pattern. Updating now.

packages/front-end/pages/datasources/[did].tsx
Comment (mknowlton89) [reply packages/front-end/pages/datasources/[did].tsx]: I've dug into this more, and I think this deserves its own PR. We've got two permissions around data sources `createDatasources` and `editDatasourceSettings`. 

If a user is to create a data source or update an existing datasource's connection settings, that should require a `createDatasources` permission. But updating the data source name, projects, and other settings (excluding the connection settings) should only require the `editDatasourceSettings`. 

Currently, the modal to edit a data sources name, projects, and description is in the same form as the connection settings, so even if the `experimenter` user isn't updating the connection settings, they get a permissions error because `params` are included in the body and thus, the permission being checked is `createDatasources`.

It seems like we need to update [did] to specifiy editing connection settings and deleting, vs editing connection info. Likewise, we should give the user the ability to edit projects/name/description outside of the "Edit Connection Info" modal

Comment (mknowlton89) [reply packages/front-end/pages/datasources/[did].tsx]: Another update - I updated [did].tsx to differentiate between if the user can edit or if they can create/delete. So now, if a user can edit a datasource's settings, they can do so, but we only show `admins` the "Edit Connection Info" and "Delete" buttons. 

We should probably have a convo about whether a `Experimenter` can edit the name, description, or projects for a data source. Editing the projects could affect other people's permissions, so I could see an argument that only admins can do that.

packages/front-end/pages/metrics.tsx
```diff
@@ -371,122 +371,137 @@ const MetricsPage = (): React.ReactElement => {
...
-              <td
-                style={{ cursor: "initial" }}
+            if (
+              metric.onDuplicate &&
+              editMetricsPermissions[metric.id] &&
+              permissions.check("createMetrics", project) &&
+              canCreateMetrics()
+            ) {
+              moreMenuLinks.push(
+                <button
```
jdorn, line 384: I would add a `key` prop to these since they are being added to an array

jdorn (approved):

PR #2133, opened 2024-02-16
Make owner optional in experiment POST endpoint.

### LINKED ISSUE #1928: [Feature] Update Experiment REST API POST endpoint to make owner optional
### Description of Feature

We want to support owner-less experiments. Currently, the `owner` property is required.

With this update, the `owner` field should be optional. If it's not provided, we can skip it, but if it's provided, we use the current flow where we look up the user by email address.

Additionally, if the auth was done via a personal access token, and an owner property wasn't provided, we should default to the owner of the personal access token.

romain-growthbook:
### Features and Changes

This PR addresses the changes described in https://github.com/growthbook/growthbook/issues/1928. It makes `owner` body argument optional and uses the request's `userId` for its value when available and no `owner` has been spericified.

- Closes #1928


packages/back-end/src/api/experiments/postExperiment.ts
```diff
@@ -47,20 +47,28 @@ export const postExperiment = createApiRequestHandler(postExperimentValidator)(
       );
     }
 
-    const user = await getUserByEmail(owner);
+    const ownerPayload = await (async () => {
+      if (!ownerParam) {
+        return req.context.userId ? { owner: req.context.userId } : {};
```
romain-growthbook, line 52: The original feature request specification is:
> if the auth was done via a personal access token, and an owner property wasn't provided, we should default to the owner of the personal access token.

Is that the right interpretation?

packages/back-end/src/api/experiments/postExperiment.ts
```diff
@@ -47,20 +47,26 @@ export const postExperiment = createApiRequestHandler(postExperimentValidator)(
       );
     }
 
-    const user = await getUserByEmail(owner);
+    const ownerId = await (async () => {
+      if (!onwerEmail) return req.context.userId;
```
romain-growthbook, line 51: The original feature request specification is:
> if the auth was done via a personal access token, and an owner property wasn't provided, we should default to the owner of the personal access token.

Is that the right interpretation?

jdorn: Yes, that's correct

packages/back-end/src/api/experiments/postExperiment.ts
```diff
@@ -17,7 +17,7 @@ export const postExperiment = createApiRequestHandler(postExperimentValidator)(
   async (req): Promise<PostExperimentResponse> => {
     req.checkPermissions("createAnalyses", req.body.project);
 
-    const { datasourceId, owner } = req.body;
+    const { datasourceId, owner: onwerEmail } = req.body;
```
jdorn, line 20: Typo: `onwer` instead of `owner`

romain-growthbook: thanks, fixed.

jdorn (approved):

PR #2140, opened 2024-02-20
Finish removing relative imports in the front-end

romain-growthbook:
This PR finishes removing relative imports from the front-end. It also turns the corresponding `eslint` rule on for these files.


.eslintrc
```diff
@@ -106,6 +107,14 @@
         "import/no-named-as-default": "off"
       }
     },
+    {
+      "files": [
+        "./packages/front-end/**/*.ts*"
+      ],
+      "rules": {
+        "no-restricted-imports": ["error", {"patterns": ["..*"]}]
```
bttf, line 115: I think this might be too restrictive but not sure what the team thinks. Would be great to bring up at the eng. roundtable

bttf (commented):
Looks good, one suggestion for discussion across team. Would lean on testing via preview environments generated for the PR to test (see comments for auto-generated link)

.eslintrc
romain-growthbook: Will do!

packages/front-end/tsconfig.json
```diff
@@ -17,7 +17,9 @@
     "incremental": true,
     "baseUrl": "./",
     "paths": {
-      "@/*": ["./*"]
+      "@/*": ["./*"],
+      "@back-end/*": ["../back-end/*"],
+      "@shared/*": ["../shared/*"]
```
bttf, line 22: I'm not sure if this is correct ... We want to import from `'shared/*'` and not `../shared/*` if I'm not mistaken. The former is package reference (see frontend package.json dependencies  for `shared`), which should reference the compiled TS in your local dev. The latter is relative, and I'm not sure if it works the same or not. (Correct me if I'm wrong please!)

romain-growthbook: Work. I think you're right. Have removed the ts path config and substituted all `@shared` for `shared`.

bttf (approved):

PR #2141, opened 2024-02-20
Add schema if specified in auto metrics SQL

lukesonnet:
Problem: Right now we don't properly pass the schema through in auto-metric generation if `requiresSchema` is false for a SQL integration.

Fix: If schema exists when generating tracked events SQL, then make sure it gets added to the SQL. This will affect any auto-metric Integration with `requiresSchema = false`, but it seems like without this it will fail and this fix generates the expected path for Snowflake.

Loom showing before and after: https://www.loom.com/share/0fda8d56d1e94983916e57b2b39a63f6

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -2806,7 +2806,12 @@ AND event_name = '${eventName}'`,
         COUNT (*) as count,
         MAX(${timestampColumn}) as lastTrackedAt
       FROM
-        ${this.generateTablePath(trackedEventTableName, schema)}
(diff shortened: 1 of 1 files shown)


mknowlton89:
Looking at the code, it looks like the Data Source types that support `auto metrics`, that also have `requiresSchema = false` are: Athena, Databricks, Postgres, MS Sql, Snowflake, and Clickhouse. 

I don't have any of those spun up with sample data at the moment, so I'm not able to test conclusively.

I do also agree that `generateTablePath` could use a refresh, but it feels like we may need to do that separately with a lot of thought (and ideally, with good sample data).

mknowlton89 (approved):

PR #2149, opened 2024-02-22
update stale flag detection to accommodate prereqs/dependencies

bryce-fitzsimons:
### Features and Changes

### DIFF
--- a/packages/shared/src/util/features.ts
+++ b/packages/shared/src/util/features.ts
@@ -177,60 +177,113 @@ const areRulesOneSided = (
 
 export function isFeatureStale(
   feature: FeatureInterface,
+  features: FeatureInterface[],
+  experiments: ExperimentInterfaceStringDates[],
+  environments: string[],
   linkedExperiments: ExperimentInterfaceStringDates[] | undefined = []
 ): { stale: boolean; reason?: StaleFeatureReason } {
-  try {
-    if (feature.neverStale) return { stale: false };
+  const visitedFeatures = new Set<string>();
 
-    if (feature.linkedExperiments?.length !== linkedExperiments.length) {
-      // eslint-disable-next-line no-console
-      console.error("isFeatureStale: linkedExperiments length mismatch");
-      return { stale: false, reason: "error" };
-    }
-
-    const linkedExperimentIds = linkedExperiments.map((e) => e.id);
-    if (
-      !linkedExperimentIds.every((id) =>
-        feature.linkedExperiments?.includes(id)
-      )
-    ) {
-      // eslint-disable-next-line no-console
-      console.error("isFeatureStale: linkedExperiments id mismatch");
-      return { stale: false, reason: "error" };
-    }
-
-    const twoWeeksAgo = subWeeks(new Date(), 2);
-    const dateUpdated = getValidDate(feature.dateUpdated);
-    const stale = dateUpdated < twoWeeksAgo;
-
-    if (!stale) return { stale };
-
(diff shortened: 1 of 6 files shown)


packages/shared/src/util/features.ts
```diff
@@ -175,62 +175,131 @@ const areRulesOneSided = (
...
+      if (features && features.length > 1) {
+        const dependentFeatures = getDependentFeatures(
+          feature,
+          features,
+          environments
+        );
+        const hasNonStaleDependentFeatures = dependentFeatures.some((id) => {
+          const f = features?.find((f) => f.id === id);
+          if (!f) return true;
+          return !visit({ feature: f }).stale;
```
bttf, line 260: Is the depth of a feature tree limited to 1?

bryce-fitzsimons: no, it's unbounded

bryce-fitzsimons:
> With this new change, I think we've tipped the scales where we'd probably want it to run serverside so it can more efficiently aggregate an org's entire list of features and experiments in order to determine staleness based on the dependency tree.

@bttf I agree. But probably not worth making that leap at this moment. Instead, just update the logic now so we don't erroneously show stale when a prereq is not actually stale since prereqs are live. In the future, BE would be the place to do this, and we'd probably also include signals from telemetry.

packages/shared/src/util/features.ts
bttf: Ok, I may be misunderstanding the flow here, but if it's unbounded would we not want to pass in `features`and `experiments` to this `visit` call?

bttf:
> > With this new change, I think we've tipped the scales where we'd probably want it to run serverside so it can more efficiently aggregate an org's entire list of features and experiments in order to determine staleness based on the dependency tree.
> 
> @bttf I agree. But probably not worth making that leap at this moment. Instead, just update the logic now so we don't erroneously show stale when a prereq is not actually stale since prereqs are live. In the future, BE would be the place to do this, and we'd probably also include signals from telemetry.

Sure, makes sense. My only suggestion would be to try to have these discussions before we make a new feature live so that we don't end up shipping stuff like this 'under the gun'.

RE: Tests, since there's already a full suite of tests, I do think we should add at least a minimum amount of unit tests to cover this new functionality. What do you think? I'm not sure if my recent comment has uncovered a bug or not - either way, having tests would help with bugs or documenting use case.

packages/shared/src/util/features.ts
bryce-fitzsimons: they're global to the `isFeatureStale()` scope so they can be accessed by each `visit()` iteration. They don't need to mutate for each visit.

packages/shared/src/util/features.ts
```diff
@@ -175,62 +175,131 @@ const areRulesOneSided = (
...
-  linkedExperiments: ExperimentInterfaceStringDates[] | undefined = []
-): { stale: boolean; reason?: StaleFeatureReason } {
-  try {
-    if (feature.neverStale) return { stale: false };
+interface IsFeatureStaleInterface {
+  feature: FeatureInterface;
+  features?: FeatureInterface[];
+  experiments?: ExperimentInterfaceStringDates[];
+  environments?: string[];
+  linkedExperiments?: ExperimentInterfaceStringDates[];
```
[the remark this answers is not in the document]
bryce-fitzsimons: removed it. seems like it was broken anyhow (keep getting length mismatch bugs)

bttf (approved):

PR #2150, opened 2024-02-22
Romain/webhook enhancements

romain-growthbook:
### Features and Changes


packages/front-end/components/EventWebHooks/EventWebHookAddEditModal/EventWebHookAddEditModal.tsx
```diff
@@ -108,6 +139,20 @@ export const EventWebHookAddEditModal: FC<EventWebHookAddEditModalProps> = ({
...
+        value={form.watch("method")}
+        placeholder="Choose HTTP method"
+        options={eventWebHookMethods.map((method) => ({
+          label: method,
+          value: method,
+        }))}
+        onChange={(value: EventWebHookMethod) => {
+          form.setValue("method", value);
+          handleFormValidation();
+        }}
```
bryce-fitzsimons, line 153: fyi: `sort={false}` is needed if you want to preserve the order of options

packages/front-end/components/EventWebHooks/EventWebHookAddEditModal/EventWebHookAddEditModal.tsx
```diff
@@ -122,6 +167,68 @@ export const EventWebHookAddEditModal: FC<EventWebHookAddEditModalProps> = ({
...
+        value={form.watch("payloadType")}
+        placeholder="Choose payload type"
+        options={eventWebHookPayloadTypes.map((key) => ({
+          label: eventWebHookPayloadValues[key],
+          value: key,
+        }))}
+        onChange={(value: EventWebHookPayloadType) => {
+          form.setValue("payloadType", value);
+          handleFormValidation();
+        }}
```
bryce-fitzsimons, line 181: maybe `sort={false}`

packages/front-end/components/EventWebHooks/EventWebHookAddEditModal/EventWebHookAddEditModal.tsx
romain-growthbook: Applied, thanks

packages/back-end/src/events/handlers/utils.ts
```diff
@@ -0,0 +1,182 @@
...
+    case "experiment.updated":
+    case "experiment.deleted":
+      return true;
+
+    case "feature.created":
+    case "feature.deleted":
+      return true;
+
+    case "feature.updated":
+      return filterFeatureUpdatedNotificationEventForEnvironments({
```
bryce-fitzsimons, line 115: might be good to leave a comment here, since it kinda threw me off why we usually just return "true"

```
// Will only notify for environments in which rules were modified. Other feature events (currently) apply to all environments
```

packages/back-end/src/events/handlers/utils.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  // Manual environment filtering
+  const changedEnvironments = new Set<string>();
+
+  // Some of the feature keys that change affect all enabled environments
+  const relevantKeysForAllEnvs: (keyof ApiFeature)[] = [
+    "archived",
+    "defaultValue",
+    "project",
+    "valueType",
```
bryce-fitzsimons, line 148: we probably need to add checks for "prerequisites". Changes here should definitely trigger an update event.

packages/front-end/components/EventWebHooks/EventWebHookAddEditModal/EventWebHookAddEditModal.tsx
```diff
@@ -122,6 +194,72 @@ export const EventWebHookAddEditModal: FC<EventWebHookAddEditModalProps> = ({
         }}
       />
 
+      {false && (
```
bryce-fitzsimons, line 197: add //todo so we know why this is soft-removed

packages/back-end/src/models/EventWebhookModel.ts
```diff
@@ -264,16 +368,38 @@ export const getAllEventWebHooks = async (
...
     organizationId,
     events: eventName,
     enabled,
   });
 
+  const docs = allDocs.filter((doc) => {
+    if (doc.tags?.length && tags?.length && !intersection(doc.tags, tags))
+      return false;
+
+    if (
```
bryce-fitzsimons, line 394: possibly yes, but not 100% sure if this captures the desired behavior. I *think* this is what we're looking for; we should double check these scenarios:
- doc has no projects, event has no projects: allow
- doc has projects, event has no projects: allow
- doc has no projects, event has projects: remove
- doc has projects, event has projects: only allow if intersecting

jdorn (commented):
I didn't do a deep review, but just want to highlight some things we should test before merging this, especially since we have production users currently using event webhooks that we don't want to break.

- Make sure existing event webhooks continue to fire for the same events as before even though they are missing all of the new Mongo fields.
- Make sure you're able to edit an old event webhook that's missing all of the new fields in Mongo and save it successfully.  Ideally, at this point it should have all of the new fields specified with default values.
- Make sure the payloads, headers, etc. being sent by default are 100% identical to before

romain-growthbook:
> I didn't do a deep review, but just want to highlight some things we should test before merging this, especially since we have production users currently using event webhooks that we don't want to break.
> 
> * Make sure existing event webhooks continue to fire for the same events as before even though they are missing all of the new Mongo fields.

See: https://www.loom.com/share/460472514f29482d9f1f330ae25d3ce6

> * Make sure you're able to edit an old event webhook that's missing all of the new fields in Mongo and save it successfully.  Ideally, at this point it should have all of the new fields specified with default values.

See: https://www.loom.com/share/d0a931d2e1a64ed8b5ecf1114bd610d4

> * Make sure the payloads, headers, etc. being sent by default are 100% identical to before

I added tests for that: https://github.com/growthbook/growthbook/pull/2150/files#diff-1983fa110451ea11e1df239ab57d069c0959f59925110e43fc10ec94d6c3de54

packages/back-end/src/models/EventWebhookModel.ts
romain-growthbook: I added a test for that: https://github.com/growthbook/growthbook/pull/2150/files#diff-b6d0babef2699f106a9f4ddb27df1aaee061c3090d237cf6a13b96e9a8d09d9d

romain-growthbook:
I've addressed the above comments with a combination of manual and automated testing!

bryce-fitzsimons (approved):
looks good. have not tested the logic, but appears correct

PR #2154, opened 2024-02-23
Changed datalayer call from array to object

Auz:


vexersa:
I was hoping for an update on this fix 🙏a

Auz:
We have another PR in the works that will fix this in a more complete way. It should be live this week. I'll link the PR once it's published.

vexersa:
Great news

Auz:
This is fixed in [2210](https://github.com/growthbook/growthbook/pull/2210)

PR #2158, opened 2024-02-26
Fix SDK Webhook Disabled Button Check

jdorn:
### Features and Changes

SDK Connection Webhooks had incorrect logic for when the Create button is disabled.  Also, when disabled, there was no indication about why.  This PR fixes both issues.


gazzdingo (approved):
Lgtm

PR #2160, opened 2024-02-26
[bugfix] Apply metric overrides which carry computed RA settings

lukesonnet:
Previously, if a metric had its own `RegressionAdjustedEnabled: False` set to off, but there are org or experiment level settings that override it, that override wasn't applied properly.

So we have to use `applyMetricOverride` when building the metric settings for the stats engine to ensure these computed fields get passed through (even if there aren't specifically Experiment Metric Overrides), because the Snapshot Settings contains some of these overrides from other sources.

Demonstrating the issue and that it is resolved: https://www.loom.com/share/4634c16254a348c9a9f3b782ebceb92b

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -25,7 +25,6 @@ import {
 import {
   ExperimentMetricInterface,
   getRegressionAdjustmentsForMetric,
-  isBinomialMetric,
   isFactMetric,
   isFactMetricId,
 } from "shared/experiments";
@@ -112,6 +111,7 @@ import {
   QueryResultsForStatsEngine,
   analyzeExperimentMetric,
   analyzeExperimentResults,
+  getMetricSettingsForStatsEngine,
 } from "./stats";
 import { getEnvironmentIdsFromOrg } from "./organizations";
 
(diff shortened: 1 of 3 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -241,8 +241,20 @@ export async function getManualSnapshotData(
     const metric = metricMap.get(m);
     if (!metric) return null;
 
+    const denominator =
+      metric.denominator && !isFactMetric(metric)
+        ? metricMap.get(metric.denominator)
+        : undefined;
     metricSettings[m] = {
-      ...getMetricSettingsForStatsEngine(metric, metricMap, false),
```
bryce-fitzsimons, line 245: seems like it would be better to make `getMetricSettingsForStatsEngine` reflect this nuance than to copy/paste most of the logic here

lukesonnet: We would have to build a snapshot settings object for manual snapshots which we never have. I was just lazy here because we don't care about manual snapshots, but I'll take another look to see if it's easy to build it.

lukesonnet: Done, drier now.

bryce-fitzsimons (approved):
looks good, did not test

PR #2161, opened 2024-02-26
fixed url in webhooks

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Settings/WebhooksModal.tsx
+++ b/packages/front-end/components/Settings/WebhooksModal.tsx
@@ -22,9 +22,7 @@ const WebhooksModal: FC<{
 }> = ({ close, onSave, current, showSDKMode, sdkid }) => {
   const { apiCall } = useAuth();
   const [validHeaders, setValidHeaders] = useState(true);
-  const [displayHttpValidationError, setDisplayHttpValidationError] = useState(
-    false
-  );
+
   showSDKMode = showSDKMode || false;
   const methodTypes: WebhookMethod[] = [
     "POST",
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Settings/WebhooksModal.tsx
```diff
@@ -61,13 +64,29 @@ const WebhooksModal: FC<{
...
+  };
+
   const onSubmit = form.handleSubmit(async (value) => {
+    setDisplayHttpValidationError(false);
     if (value.endpoint.match(/localhost/g)) {
       throw new Error("Invalid endpoint");
     }
+    if (!isValidHttp(value.endpoint)) {
+      setDisplayHttpValidationError(true);
+      return;
```
jdorn, line 84: You can just throw from here instead and it will show the message and manage the error state automatically.

packages/front-end/components/Settings/WebhooksModal.tsx
```diff
@@ -61,13 +64,29 @@ const WebhooksModal: FC<{
...
+    }
     await handleApiCall(value);
     track(current.id ? "Edit Webhook" : "Create Webhook");
     onSave();
+    close();
```
jdorn, line 89: If you make the above change, you can get rid of this and the `autoCloseOnSubmit` change

packages/front-end/components/Settings/WebhooksModal.tsx
gazzdingo: oh that is nice! good to know

jdorn (approved):

PR #2164, opened 2024-02-26
Simplify hasReadAccess to use req.checkPermissions

mknowlton89:
### Features and Changes

Now that we have the concept of READ_ONLY_PERMISSIONS the logic of hasReadAccess and checkPermissions is duplicated, so I've got a draft PR up that adds checkPermissions to the backend context object, which allows us to use checkPermissions when using the hasReadAccess filter function.

Currently, checkPermissions throws an error if the user doesn't have permission, but hasReadAccess is used to filter resources so we only return resources the user has readData permission for, so we don't want it to throw an error, but simply return false if the user doesn't have permission. As such, hasReadAccess still exists as a function, but it's a wrapper that calls checkPermissions.

### DIFF
--- a/packages/back-end/src/init/config.ts
+++ b/packages/back-end/src/init/config.ts
@@ -200,7 +200,7 @@ export function getConfigMetrics(
         managedBy: "config",
       });
     })
-    .filter((m) => hasReadAccess(context.readAccessFilter, m.projects || []));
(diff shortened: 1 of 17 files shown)


mknowlton89:
Spoke with Jeremy - we'll likely want to update our mongo queries to handle filtering based on read access, so while this will clean things up, its going to be removed anyways. As a result, closing this PR.

PR #2169, opened 2024-02-27
Compare Env Rules

mknowlton89:
### Features and Changes

This PR introduces a new modal on the Feature ID page that allows a user to compare rules across two environments. Additionally, it allows the user to copy entire rule sets from one environment to another, overwriting the target environment's rule set entirely.

This compliments the existing feature of being able to copy single rules from one env to another. Now, users can copy complete rule sets from one env to another.

### Testing

See screenshots below

### Screenshots

New modal location (this is the experience for users with permission to edit features.
Screen Shot 2024-02-28 at 2 04 33 PM

Modal if the user only has view permissions
Screen Shot 2024-02-28 at 2 07 32 PM

Modal empty state
Screen Shot 2024-02-28 at 2 04 51 PM

Modal with two environments selected - in this case, there is an existing feature draft created
Screen Shot 2024-03-05 at 6 54 13 AM

Modal with two environments selected - in this case, there is no existing feature draft
Screen Shot 2024-03-05 at 6 52 27 AM

We disable the "Copy" button if the version the user is viewing is locked
Screen Shot 2024-03-05 at 6 55 23 AM

Experience for user without permission to create/edit features
Screen Shot 2024-03-05 at 6 57 55 AM

Experience when the rule sets for both environments are equal (hard to tell, but Copy button is disabled)
Screen Shot 2024-03-05 at 6 58 25 AM


packages/back-end/src/models/FeatureModel.ts
```diff
@@ -687,6 +687,23 @@ export async function editFeatureRule(
   });
 }
 
+export async function editFeatureRuleSet(
+  revision: FeatureRevisionInterface,
+  environment: string,
+  rules: FeatureRule[],
+  user: EventAuditUser
+) {
```
romain-growthbook, line 695: nit: I'd suggest using an object payload here for readability and flexibility w.r.t. argument orders and such,

packages/front-end/components/Features/CompareRuleDiffModal.tsx
```diff
@@ -0,0 +1,230 @@
...
+        sourceEnvironment: sourceEnv,
+        targetEnv,
+        rules: newRules,
+      });
+      await mutate();
+      res.version && setVersion(res.version);
+    } catch (e) {
+      setError(e.message);
+    }
+    setCopyingRules(false);
```
romain-growthbook, line 78: Might want to wrap this into a `finally`:
```suggestion
    } finally {
      setCopyingRules(false);
    }
```

packages/front-end/components/Features/CompareRuleDiffModal.tsx
```diff
@@ -0,0 +1,230 @@
...
+  function getDiffString(environment: string): string {
+    const updatedRules: PartialRule[] = feature.environmentSettings[
+      environment
+    ]?.rules.map((rule: FeatureRule) => {
+      // We don't want to include rule.ids in the diff as they'll always be different
+      const partialRule: Omit<FeatureRule, "id"> & { id?: string } = {
+        ...rule,
+      };
+      delete partialRule.id;
+      return partialRule;
```
romain-growthbook, line 97: nit but more immutable-y
```suggestion
      const { id: _, ...partialRule } = rule;
      return partialRule;
```

romain-growthbook (commented):
Looking good! I'm not sure if I have enough background to approve but I added some minor remarks if that can help!

packages/front-end/components/Features/CompareRuleDiffModal.tsx
mknowlton89: Great suggestion. I had to add an `eslint-disable-next-line` comment here. Looking at our eslint configuration it looks like we should be ignoring this automatically, but that doesn't seem to be the case.

I'm not super familiar with eslint to know if there is something else at play here.

bryce-fitzsimons:
it might be worth thinking about how this influences prerequisite features (both for features which are prerequisite of others, and dependents features who reference prerequisites in their rules). You could get into a scenario where you're making a prereq state non-deterministic and might need to call that out somehow in the UI

bryce-fitzsimons:
the diff modal and UI seem a bit confusing to me.
- which column represents the source versus destination?
- why is one red and one green? (implies removing and adding)
- why do we need a "compare new selection" button? seems to do the same thing as just choosing something else
- consider putting the CTA on the bottom bar of the modal
- consider putting the 2 environment selectors on top of each diff column so you know which column belongs to which env

mknowlton89:
@bryce-fitzsimons Great feedback/questions.

> the diff modal and UI seem a bit confusing to me.
> 
> * which column represents the source versus destination?
> * why is one red and one green? (implies removing and adding)
> * why do we need a "compare new selection" button? seems to do the same thing as just choosing something else
- I don't think we do - it was in the UI mock so I added it, but I'll discuss with Kari.
> * consider putting the CTA on the bottom bar of the modal
- What do you think the CTA on the bottom bar should be?
> * consider putting the 2 environment selectors on top of each diff column so you know which column belongs to which env
- Makes sense - I'll share this with Kari when she's back in the office and adjust.
> 
> I guess taking a step back, what is the purpose of a diff. We're not actually doing any merging (like a github PR), we're just showing that something is going to completely overwrite something else

So, the purpose of this is to easily compare rule sets across environments - currently, someone needs to tab back and forth between environment tabs to compare rules.

The idea is that it could be used to debug - e.g. why is it the feature firing in this env but not this env - with this tool, they can compare the rule set side by side and see any differences easily.

It's also a tool that can be used to copy entire rule sets (rather than having to copy each rule one by one. Thinking an org might create a feature and build the rule set for a test environment, and then want to easily duplicate the entire rule set to production once its validated.

mknowlton89:
@bryce-fitzsimons Just a quick update - I met with Kari yesterday and we discussed your feedback. We've removed the "Compare New Selection" reset button, added titles to each column, updated the colors (away from red/green) and I was also able to remove the `+` and `-` markers next to each line since they aren't really contextually accurate in this case.

As a reminder, this feature is to enable users to compare entire rule sets side by side, and, if the user has permission, copy/overwrite entire rule sets from one environment to another. The likely use case is testing the targeting in a testing/staging env, and then applying the entire rule set to production, or to debug two rule sets to identify what is different between them.

mknowlton89:
Closing due to age and being now very out of date.

PR #2179, opened 2024-02-29
Add project level scoping to Attributes

mknowlton89:
### Features and Changes

This PR updates adds project scoping to `Attributes`. Previously, attributes all lived at the organization level, now, attributes can be associated with multiple projects.

While this seemed like a straight-forward change, it did require some refactoring. Attributes live in an organization's settings object, and previously, when creating/editing/deleting attributes, we'd build a new copy of the attributes array, and then pass that full array into the `organizations.controller.ts/putOrganization` function. 

This meant that the singular function had to handle the logic for adding, removing, and updating an attribute, and it got really unwieldy. In response, I've updated the `putOrganization` endpoint to now disallow updating organization attributes, and instead created a new router and controller file that has specific endpoints for adding, updating, and deleting attributes, ensuring that we can have a specific function to handle each request.

Additionally, I've done an audit of the app to ensure that anywhere we were calling `/organizations` with a PUT request type, is not passing in the full `org.settings` object (as that would throw an error as it would seem like the user was trying to update the orgs attributes array.).

### Testing

- [x] Ensure that as an admin user, the user can create, update, archive, and delete attributes
- [x] Ensure that a user without `manageTargetingAttributes` don't see the CRUD options in the app
- [x] Ensure that a user without `manageTargetingAttributes` can not perform any crud actions
- [x] Ensure that users with project-specific permissions see the correct permissions


packages/front-end/components/Features/AttributeModal.tsx
```diff
@@ -65,52 +68,51 @@ export default function AttributeModal({ close, attribute }: Props) {
...
             "That attribute name is already being used. Please choose another one."
           );
         }
 
-        const settings: Pick<OrganizationSettings, "attributeSchema"> = {
-          attributeSchema,
-        };
-        await apiCall(`/organization`, {
-          method: "PUT",
+        const url = attribute ? `/attribute/${attribute}` : `/attribute`;
```
jdorn, line 78: We don't have any validation on attribute name today, so someone could define an attribute with a slash in it or other non-URL safe characters.  It's not safe to pass this in the URL

packages/front-end/components/Features/AttributeModal.tsx
```diff
@@ -65,52 +68,51 @@ export default function AttributeModal({ close, attribute }: Props) {
...
-        else {
-          attributeSchema.push(value);
-        }
-
         // Make sure this attribute name doesn't conflict with any existing attributes
-        if (
-          attributeSchema.filter((s) => s.property === value.property).length >
-          1
-        ) {
+        if (!attribute && schema.some((s) => s.property === value.property)) {
```
jdorn, line 72: The form below allows changing this for existing attributes, so we also need to check that the name they are changing to does not conflict with another property.

packages/front-end/pages/saved-groups.tsx
```diff
@@ -88,13 +89,7 @@ export default function SavedGroupsPage() {
...
         });
     }
-  }, [
-    permissions.manageTargetingAttributes,
-    apiCall,
-    refreshOrganization,
-    attributeSchema,
-    savedGroups,
-  ]);
+  }, [apiCall, refreshOrganization, attributeSchema, savedGroups, permissions]);
```
jdorn, line 92: `permissions` is not memoized and will change every time the app renders.  We should update `UserContext.tsx` and memoize it there since this is likely not the only place doing this.

packages/back-end/src/routers/attributes/attributes.controller.ts
```diff
@@ -0,0 +1,196 @@
...
+  attributeSchema[index] = {
+    ...attributeSchema[index],
+    property,
+    datatype,
+    projects,
+    format,
+    enum: enumValue,
+    hashAttribute,
+    archived,
+  };
```
jdorn, line 119: If the `property` field has changed, need to check that the new value doesn't conflict with any other attributes.

packages/front-end/pages/attributes.tsx
```diff
@@ -69,29 +69,21 @@ const FeatureAttributesPage = (): React.ReactElement => {
...
                   method: "PUT",
                   body: JSON.stringify({
-                    settings: { attributeSchema },
+                    property: v.property,
+                    datatype: v.datatype,
+                    projects: v.projects,
+                    format: v.format,
+                    enum: v.enum,
+                    hashAttribute: v.hashAttribute,
+                    archived: !v.archived,
```
jdorn, line 83: This is prone to breaking in the future without being caught by CI - for example, if we add a new required property to the attribute schema and forget to update this.

Adding typing here will make it so this will be caught by Typescript at compile time.

```ts
const newAttr: SDKAttribute = {
  // ...
};

//...
body: JSON.stringify(newAttr)
```

jdorn (approved):

PR #2180, opened 2024-02-29
custom results info banner

itsgrimetime:
### Features and Changes

Adds an environment variable toggle-able "CustomInfoBanner" that can be used to provide a custom banner on the experiment results view for self-hosted deployments.

This is achieved via a "DynamicComponentLoader" class that takes a `componentName` property and looks at env var flags (currently only one), and based on that conditionally loads a mapped component.

We use this currently to inform users of new features, documentation, or changes in how we are using GrowthBook internally. Ideally this would be something that's supported on the cloud-hosted instance as well, but I'm not sure really how that'd be able to be implemented (unless it just allowed completely static HTML). This solution is nice as it allows for dynamically loading a pre-compiled component, which can import and access experiment/org/user models and dynamically change the content based on that.

To enable this info banner the environment variable `NEXT_PUBLIC_RESULTS_CUSTOM_INFO_BANNER` needs to be set to `true` (otherwise the DynamicComponentLoader just returns `null`, resulting in no additional component being added to the results view).

### Testing

I've attached screenshots showing an example of how we're using it, as well as what the results view looks like with `NEXT_PUBLIC_RESULTS_CUSTOM_INFO_BANNER` enabled (but no changes to `CustomInfoBanner.tsx`), as well as when the env var is unset/set to false.

### Screenshots

Screenshot 2024-02-29 at 10 39 02 AM

Screenshot 2024-02-29 at 10 54 52 AM

Screenshot 2024-02-29 at 11 04 46 AM

### DIFF
--- a/packages/front-end/.env.example
+++ b/packages/front-end/.env.example
@@ -3,3 +3,5 @@ API_HOST=http://localhost:3100
 # We collect anonymous telemetry data to help improve GrowthBook
 # Setting the below to "true" disables this behavior
 DISABLE_TELEMETRY=true
+
+NEXT_PUBLIC_RESULTS_CUSTOM_INFO_BANNER=false
--- a/packages/front-end/components/Experiment/CustomInfoBanner.tsx
+++ b/packages/front-end/components/Experiment/CustomInfoBanner.tsx
@@ -1,0 +1,19 @@
+import clsx from "clsx";
+
+const CustomInfoBanner = () => {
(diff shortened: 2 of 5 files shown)


jdorn:
We've had thoughts of allowing custom Markdown to be injected throughout the app.  I'm curious to hear more about use cases where having a full component is helpful.  I'm wondering if markdown + a templating language is good enough to handle dynamic use cases.  The nice thing about markdown + templating language is that we can persist that in Mongo and it doesn't require creating a custom build.

PR #2181, opened 2024-02-29
Fix metric window graphic and add second example

lukesonnet:
The labels on the metric window graphic were incorrect. I also added a second graphic showing that conversion windows extend beyond the experiment end date.


gazzdingo (changes requested):
LGTM

gazzdingo (approved):
LGTM

PR #2183, opened 2024-03-01
New unified ReqContext class

jdorn:
### Features and Changes

Make a dedicated class for the request Context.  It now includes:
- `context.hasPermission` (returns boolean) and `context.requirePermission` (throws error)
- `context.logger` - instance of a Pino logger
- `context.auditLog(data)` method to insert an audit log entry
- `getProjects()` to get all projects in the org with a cache layer
- `registerTags(tags)` to add any new tags to the org

Most of these new methods are not being used yet - they were just added for future use (e.g. the upcoming data models project)


packages/back-end/types/organization.d.ts
```diff
@@ -284,12 +283,4 @@ export type NamespaceUsage = Record<
...
-export type ReqContext = {
-  org: OrganizationInterface;
-  userId: string;
-  email: string;
-  environments: string[];
-  userName: string;
-  readAccessFilter: ReadAccessFilter;
-  auditUser: EventAuditUser;
-};
+export type ReqContext = ReqContextInterface;
```
mknowlton89, line 286: Is there being done in order to prevent having to update ~100 files, or was there another determining factor here?

mknowlton89 (approved):

packages/back-end/types/organization.d.ts
jdorn: Yes, exactly

PR #2187, opened 2024-03-01
Project scoped environments with feature UI filters

bryce-fitzsimons:
### Features and Changes

based on: https://github.com/growthbook/growthbook/pull/2166

- Introduces environment filtering on [fid] and feature listing pages.


packages/back-end/src/api/sdk-payload/getSdkPayload.ts
```diff
@@ -25,11 +26,20 @@ export const getSdkPayload = createApiRequestHandler()(
...
     } = await getPayloadParamsFromApiKey(key, req);
 
+    const environmentDoc = req.context.org?.settings?.environments?.find(
+      (e) => e.id === environment
+    );
+    const filteredProjects = filterProjectsByEnvironment(
+      projects,
+      environmentDoc,
+      true
+    );
```
jdorn, line 36: What if an SDK Connection is filtered to just project A and it's for dev.  You remove project A from the dev environment.

So before filtering, projects = `["A"]`, but after it's an empty array.  Won't that change the behavior here?  Instead of only including features in project A, it now includes all features (no project scoping at all).

bryce-fitzsimons: Yes, this is the implicit danger in removing a project scope. It affects tied entities. In this case, there is no way to have a "zero project" SDK connection, so we default to the next best thing, back to all projects.

We warn before removing the project
> You have made the projects filter more restrictive than before. 2 SDK Connections using this environment may be impacted.

However, it's not exactly accurate per your example. Removing project A from the sdk connection resets it to [], which here is "less restrictive". Maybe we should call this out in the UI?

bryce-fitzsimons: how about this?
image

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -1178,13 +1183,26 @@ export async function putOrganization(
...
         // Require permissions for any new environments that have been added
-        const oldIds = new Set(existing.map((env) => env.id) || []);
+        const oldIds = new Set(existingEnvironments.map((env) => env.id) || []);
         settings[k]?.forEach((env) => {
           if (!oldIds.has(env.id)) {
             affectedEnvs.add(env.id);
           }
         });
 
+        // Check if any environments' projects have been changed (may require SDK Connection updates)
```
bryce-fitzsimons, line 1193:
```suggestion
        // Check if any environments' projects have been changed (may require webhook triggers)
```

packages/front-end/components/Features/FeaturesOverview.tsx
```diff
@@ -145,9 +148,17 @@ export default function FeaturesOverview({
...
-  const environments = useEnvironments();
+  const allEnvironments = useEnvironments();
+  const environments = filterEnvironmentsByFeature(allEnvironments, feature);
   const envs = environments.map((e) => e.id);
 
+  // Make sure you can't access an invalid env tab, since active env tab is persisted via localStorage
+  useEffect(() => {
+    if (!envs.includes(env)) {
+      setEnv(envs[0]);
+    }
```
bryce-fitzsimons, line 159:
```suggestion
    if (!envs?.length) return;
    if (!envs.includes(env)) {
      setEnv(envs[0]);
    }
```

packages/front-end/pages/features/[fid].tsx
```diff
@@ -135,18 +135,23 @@ export default function FeaturePage() {
...
+  }, [revisions, version, allEnvironments, baseFeature]);
 
   const feature = useMemo(() => {
     if (!revision || !baseFeature) return null;
     return revision.version !== baseFeature.version
       ? mergeRevision(
           baseFeature,
           revision,
-          environments.map((e) => e.id)
+          allEnvironments.map((e) => e.id)
```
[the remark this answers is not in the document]
Comment (bryce-fitzsimons) [reply packages/front-end/pages/features/[fid].tsx]: fixed. it was due to variable declaration order, but was able to rearrange the logic

packages/front-end/components/ProjectBadges.tsx
```diff
@@ -56,16 +61,32 @@ export default function ProjectBadges({
...
         if (!p?.name) return;
         return (
           <Badge
-            content={p.name}
+            content={
+              invalidProjectIds.includes(p.id) ? (
+                <Tooltip
+                  popperClassName="text-left"
+                  popperStyle={{ lineHeight: 1.5 }}
+                  body="This project is not allowed in the selected environment and will not be included in the SDK payload."
```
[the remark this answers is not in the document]
bryce-fitzsimons: externalized it

packages/back-end/src/controllers/features.ts
```diff
@@ -399,7 +423,7 @@ export async function postFeatures(
     project: "",
     environmentSettings: Object.fromEntries(
       Object.entries(environmentSettings).filter(([env]) =>
-        environments.includes(env)
+        environmentIds.includes(env)
```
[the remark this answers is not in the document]
bryce-fitzsimons: good catch

jdorn (approved):

PR #2204, opened 2024-03-07
Wire slack and discord payload.

romain-growthbook:
### Features and Changes

This PR wires the slack and discord payload for webhook notification integration.

Setting up a slack app and sending notifications seems pretty easy to do with it!

Setting up a discord webhook is equally easy.

Messages are factored out from the existing slack integration implementation. 

Discord webhooks have a slack-compatible mode that seems to only render text. This also makes it trickier on the user as they need to append `/slack` to the url.

Instead, this PR simply extracts the `text` param from the slack message and uses it for the discord payload.

Future work:
* Allow user to customize the notification in particular add `@user` to notification messages.
* Revisit discord payloads.

### Screenshots

slack-webhook


romain-growthbook:
Added forced params handling:

discord

bryce-fitzsimons (approved):

PR #2205, opened 2024-03-07
Add a hard cap on inviting new users

tzjames:
### Features and Changes

Add the ability to have a hard cap on licenses, including air gapped licenses.
Enforced the hard cap both on the front-end and the back-end.

### Dependencies

Need to create a license with a new script.  
https://github.com/growthbook/scripts/pull/12

### Testing

Set LICENSE_KEY env var to an air-gapped license that has a hardCap: true property.
Go to invite modal.  
Try and invite more users than seats on the license.
See a message saying that they are over seat limit.
Invite exactly the number of users as there are seats on the license.
Add a user from the orphaned list (if one is not there, remove one from the org, add a new different user, then add the one you had removed and that is in the orphaned list)
See the message saying they are over the seat limit.
Remove another user from the org.
Add the user from the orphaned list.
See that there is now space and you can.

Comment out the front end checks to show the over the limit modal, and repeat the above steps.  See the error message from the back-end only when you hit add.


packages/front-end/components/Settings/Team/InviteModal.tsx
```diff
@@ -81,6 +85,19 @@ const InviteModal: FC<{ mutate: () => void; close: () => void }> = ({
...
   }
 
+  // Hit a hard cap and needs to contact sales to increase the number of seats on their license
+  if (showContactSupport) {
+    return (
+      <Modal open={true} close={close} size="md" header={"Reached seat limit"}>
+        <div className="my-3">
+          Whoops! You reached the seat limit on your license. To increase your
+          number of seats, please contact{" "}
+          <a href="mailto:sales@growthbook.io">sales@growthbook.io</a>.
```
mknowlton89, line 95: Small suggestion - set this to open in a new tab.

```suggestion
          <a href="mailto:sales@growthbook.io" target="_blank" rel="noreferrer">
            sales@growthbook.io
          </a>.
```

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -1783,6 +1795,15 @@ export async function addOrphanedUser(
     });
   }
 
+  const license = getLicense();
+  if (
+    license &&
+    license.hardCap &&
+    getNumberOfUniqueMembersAndInvites(org) >= (license.seats || 0)
```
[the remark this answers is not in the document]
tzjames: yes

packages/front-end/components/Settings/Team/InviteModal.tsx
tzjames: Good idea

PR #2206, opened 2024-03-07
Introduce permissions class

mknowlton89:
### Features and Changes

This PR introduces a new class with the goal of (eventually) simplifying how engineers can check a user's permission throughout the application. 

To kick this off, I've taken on refactoring the `createMetrics` permission as a first stab.

Currently, there are a lot of intracacies around how to check a user's permissions, specifically around `projects`.

Some resources can only be a part of 1 project, whereas others, can be a part of multiple. Likewise, in the event a resource is in multiple projects, if the user's global role doesn't grant them access, they need to have a project-level role that grants them permission in EVERY project the resource is a part of. However, there are some permissions, that only require the user to have the permission in atleast 1 of the projects the resource is apart of.

Given this complexity, this class will eventually create helper methods that take care of this detail automatically, so the engineer doesn't have to worry about it.

During the transition period, this will live in the context object on the backend, and within the UserContext on the frontend. Likewise, I've created a new `usePermissionsUtil` hook for the frontend. Once we migrate all permissions over, we can remove the old `usePermissions` hook - and we can rename the new one, the standard `usePermission` hook, if we'd like.


packages/back-end/src/services/context.ts
```diff
@@ -90,50 +96,26 @@ export class ReqContextClass implements ReqContextInterface {
...
 
-  // Helper if you want to throw an error if the user does not have permission
-  public requirePermission(
-    permission: Permission,
-    project?: string | (string | undefined)[] | undefined,
-    envs?: string[] | Set<string>
-  ) {
-    if (!this.hasPermission(permission, project, envs)) {
-      throw new Error("You do not have permission to complete that action.");
-    }
```
mknowlton89, line 136: Given that hasPermission and requirePermission will be phased out, and these are not yet being consumed, I think it makes sense to remove them.

packages/front-end/services/UserContext.tsx
```diff
@@ -147,6 +152,17 @@ export const UserContext = createContext<UserContextValue>({
...
+    {
+      global: {
+        permissions: {},
+        limitAccessByEnvironment: false,
+        environments: [],
+      },
+      projects: {},
+    },
+    false
+  ),
```
mknowlton89, line 165: Should this actually call the instructor?

I could instead just do:
```
permissionsUtil: {
   global: {
      permissions: {...DEFAULT_PERMISSIONS },
      environments: [],
      limitAccessByEnv: false
   }
   projects: {}
}

File: packages/shared/src/permissions/permissionsClass.ts [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+    super(message);
+    this.name = "PermissionError";
+  }
+}
+
+export type PermissionsUtil = {
+  canCreateMetrics: (metric: Pick<MetricInterface, "projects">) => boolean;
+  throwPermissionError: (permission: Permission) => void;
+};
+export class permissionsUtil {
```
bryce-fitzsimons, line 14: classes should be PascalCase

mknowlton89: Thank you! Just updated.

mknowlton89:
@jdorn I've addressed the feedback from last week and also added patterns for `canUpdate` and `canDelete`. I'd love to get some feedback now. Once we get the patterns dialed in, I can do a full self-review and open the PR up for a formal review.

packages/back-end/src/services/context.ts
jdorn: I would add these back in for now.  The DataModel classes we're going to add soon will consume these.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -0,0 +1,100 @@
...
+      );
+    }
+
+    return hasPermission;
+  }
+
+  public canDeleteMetric(
+    metric: Pick<MetricInterface, "projects" | "managedBy">
+  ): boolean {
+    if (metric.managedBy) return false;
```
mknowlton89, line 63: Remove this - this isn't really a permission thing

packages/back-end/test/permissions.test.ts
```diff
@@ -1807,3 +1811,636 @@ describe("hasReadAccess filter", () => {
     ]);
   });
 });
+
+describe("PermissionsUtilClass.canCreateMetric check", () => {
```
mknowlton89, line 1815: Given that `canCreateMetric` is a wrapper around `checkProjectFilterPermission`, it feels like what we actually want to test is `checkProjectFilterPermission`. But, given `checkProjectFilterPermission` is a private method, I'm not aware of a pattern that would allow us to do so without having to mock `checkProjectFilterPermission`.

Open to suggestions.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -0,0 +1,100 @@
...
+      "You do not have permission to perform this action"
+    );
+  }
+
+  private checkProjectFilterPermission(
+    obj: { projects?: string[] },
+    permission: Permission
+  ): boolean {
+    const projects = obj.projects?.length ? obj.projects : [""];
+
```
mknowlton89, line 53: We'll need to add logic here for `READ_ONLY_TYPE` permissions e.g. `readData`, `addComments`, etc, but will wait to do so until it's relevant.

packages/front-end/pages/metric/[mid].tsx
```diff
@@ -137,8 +138,9 @@ const MetricPage: FC = () => {
   }
 
   const metric = data.metric;
+  // canUpdateMetric currently takes the existing metric, and any updates - but in this case, there are no updates, so just pass in the existing twice
   const canEditMetric =
```
Comment (mknowlton89) [inline packages/front-end/pages/metric/[mid].tsx:142]: TODO: Add call here for canDeleteMetric

packages/front-end/pages/metric/[mid].tsx
```diff
@@ -404,21 +406,23 @@ const MetricPage: FC = () => {
         {canEditMetric && (
```
Comment (mknowlton89) [inline packages/front-end/pages/metric/[mid].tsx:406]: TODO canEditMetric || canDeleteMetric &&

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  public canCreateMetric = (
+    metric: Pick<MetricInterface, "projects">
+  ): boolean => {
+    return this.checkProjectFilterPermission(metric, "createMetrics");
+  };
+
+  public canUpdateMetric = (
+    existing: Pick<MetricInterface, "projects" | "managedBy">,
```
jdorn, line 25: Remove "managedBy" from this interface (same below for delete)

packages/back-end/src/controllers/metrics.ts
```diff
@@ -91,15 +91,9 @@ export async function deleteMetric(
     return;
   }
 
-  req.checkPermissions(
-    "createAnalyses",
-    metric?.projects?.length ? metric.projects : ""
-  );
```
jdorn, line 97: Is removing this intentional?

packages/back-end/src/controllers/metrics.ts
```diff
@@ -296,10 +290,11 @@ export async function getMetricsFromTrackedEvents(
     return;
   }
 
-  req.checkPermissions(
-    "createMetrics",
-    dataSourceObj.projects?.length ? dataSourceObj.projects : ""
-  );
+  // When we create auto metrics, they inherit the data source's projects, so we check if the user
+  // has permission to createMetrics for the data source's projects
+  if (!context.permissions.canCreateMetric(dataSourceObj)) {
```
jdorn, line 295: Change this so it's clearer that we're passing in a metric-to-be-created here and not a data source:

```ts
if (!context.permissions.canCreateMetric({ projects: dataSourceObj.projects })) {
```

packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
```diff
@@ -166,11 +166,15 @@ export const postDemoDatasourceProject = async (
...
   >
 ) => {
+  const context = getContextFromReq(req);
+
   req.checkPermissions("manageProjects", "");
   req.checkPermissions("createDatasources", "");
-  req.checkPermissions("createMetrics", "");
   req.checkPermissions("createAnalyses", "");
-  const context = getContextFromReq(req);
+  if (!context.permissions.canCreateMetric({})) {
```
jdorn, line 174: We have a `demoProjId` below that we should use here since that's where the metric will be created.

packages/back-end/src/routers/project/project.controller.ts
```diff
@@ -197,7 +197,9 @@ export const deleteProject = async (
   // Clean up metrics
   if (deleteMetrics) {
     try {
-      req.checkPermissions("createAnalyses", id);
+      if (!context.permissions.canDeleteMetric({ projects: [id] })) {
+        context.permissions.throwPermissionError();
+      }
```
jdorn, line 202: Is this change intentional?  Before it was checking `createAnalyses`, but now it's checking metric permissions instead.

packages/front-end/components/FactTables/FactMetricList.tsx
```diff
@@ -51,7 +51,7 @@ export default function FactMetricList({ factTable }: Props) {
     searchFields: ["name^3", "description"],
   });
 
-  const canEdit = permissions.check("createMetrics", factTable.projects || "");
+  const canCreateMetrics = permissionsUtil.canCreateMetric(factTable);
```
jdorn, line 54: Make this clear that you are passing in a metric-to-be-created and not a fact table:

```ts
permissionsUtil.canCreateMetric({ projects: factTable.projects });
```

packages/front-end/components/Settings/EditDataSource/DataSourceMetrics.tsx
```diff
@@ -49,14 +48,21 @@ export default function DataSourceMetrics({
...
+    [id: string]: { canCreate: boolean; canUpdate: boolean };
+  } = {};
   metrics?.forEach((m) => {
-    editMetricsPermissions[m.id] = checkMetricProjectPermissions(
-      m,
-      permissions
-    );
+    editMetricsPermissions[m.id] = {
+      canCreate: permissionsUtil.canCreateMetric(m),
+      canUpdate: permissionsUtil.canUpdateMetric(m, m),
```
jdorn, line 57: The 2nd argument here should probably be an empty object `{}` to indicate that no changes are being made

packages/front-end/components/Settings/EditDataSource/DataSourceMetrics.tsx
```diff
@@ -244,27 +252,31 @@ export default function DataSourceMetrics({
                               </Tooltip>
                             ) : null}
                           </div>
-                          {editMetricsPermissions[metric.id] ? (
+                          {!metric.managedBy &&
```
jdorn, line 255: This change doesn't seem right.  Just because this metric is managed externally, doesn't mean you can't create a copy.

packages/front-end/components/Settings/EditDataSource/DataSourceMetrics.tsx
```diff
@@ -49,14 +48,21 @@ export default function DataSourceMetrics({
 
   const metrics: MetricInterface[] | undefined = data?.metrics;
 
-  const editMetricsPermissions: { [id: string]: boolean } = {};
+  const editMetricsPermissions: {
+    [id: string]: { canCreate: boolean; canUpdate: boolean };
```
jdorn, line 52: We should call is `canDuplicate` instead of `canCreate` since it's confusing why you would create a metric that already exists.

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -194,9 +194,12 @@ export default function FactMetricPage() {
     );
   }
 
+  // canUpdateMetric takes in the current metrics + any updates - canEdit is used to determine if we should show the "Edit" buttons, so pass in the existing metric twice
```
Comment (jdorn) [inline packages/front-end/pages/fact-metrics/[fmid].tsx:197]: Same as before, might make sense to pass an empty object as the 2nd argument instead

packages/front-end/pages/metric/[mid].tsx
```diff
@@ -137,8 +138,11 @@ const MetricPage: FC = () => {
   }
 
   const metric = data.metric;
+  // canUpdateMetric currently takes the existing metric, and any updates - but in this case, there are no updates, so just pass in the existing twice
```
Comment (jdorn) [inline packages/front-end/pages/metric/[mid].tsx:141]: Same here - should we pass an empty object as the 2nd parameter instead?

packages/front-end/pages/metrics.tsx
```diff
@@ -167,12 +166,15 @@ const MetricsPage = (): React.ReactElement => {
     },
     [showArchived, recentlyArchived, tagsFilter.tags]
   );
-  const editMetricsPermissions: { [id: string]: boolean } = {};
+
+  const editMetricsPermissions: {
+    [id: string]: { canCreate: boolean; canUpdate: boolean };
```
jdorn, line 171: Same here, we should call this `canDuplicate` instead of `canCreate`

packages/front-end/pages/metrics.tsx
```diff
@@ -167,12 +166,15 @@ const MetricsPage = (): React.ReactElement => {
...
+  } = {};
   filteredMetrics.forEach((m) => {
-    editMetricsPermissions[m.id] = checkMetricProjectPermissions(
-      m,
-      permissions
-    );
+    editMetricsPermissions[m.id] = {
+      canCreate: permissionsUtil.canCreateMetric(m),
+      canUpdate: permissionsUtil.canUpdateMetric(m, m),
```
jdorn, line 176: Pass empty object as 2nd argument

packages/front-end/pages/metrics.tsx
```diff
@@ -376,9 +384,9 @@ const MetricsPage = (): React.ReactElement => {
 
             if (
               metric.onDuplicate &&
-              editMetricsPermissions[metric.id] &&
-              permissions.check("createMetrics", project) &&
-              canCreateMetrics()
+              editMetricsPermissions[metric.id].canCreate &&
+              permissionsUtil.canCreateMetric(metric) &&
```
jdorn, line 388: Do we need both of these checks here?

packages/back-end/src/controllers/metrics.ts
mknowlton89: It was - it wasn't clear to me why a user would need the `createAnalyses` permission in order to remove a metric.

If we do want this to be a requirement, I think it would be best to add it to the `canDeleteMetric` function so we don't have to do it twice.

Is this ultimately trying to protect from a user removing a metric that is associated with an experiment that they don't have permission to impact? If so, it feels like it would be better to check that against all of the experiments, since a metric could be in all projects, and be associated with an experiment that is in a project hidden to a user, which the check above wouldn't catch.

packages/back-end/src/routers/project/project.controller.ts
mknowlton89: It was, but it's possible I'm missing something.

`createAnalyses` seems to relate to `experiments` and `ideas`, not necessarily directly with `metrics`.

Maybe we want to update `canDeleteMetric` to get a list of experiments using the metric, and ensure the user has `createAnalyses` on all of those experiments to protect a user from deleting a metric associated with an experiment they don't have permission to impact.

If we do that check specifically, we could then customize the error message they receive. It might be confusing to get a simple "You don't have permission", and it might be more helpful to get a "This metric is associated with an experiment you don't have permission to access, so you can not delete it."

packages/front-end/components/Settings/EditDataSource/DataSourceMetrics.tsx
mknowlton89: Ah - I thought those were protected from duplicating. 

Do you think we should prevent a user from archiving an externally managed metric?

packages/back-end/src/controllers/metrics.ts
jdorn: Not saying your change is wrong, just flagging anything in my review that looked out of the ordinary.

I think this was there because we're doing `removeMetricFromExperiments` after we delete, but I agree it's probably fine to remove this check.  It's a super limited form of editing the experiment, so you shouldn't need full access in order to do that.

packages/back-end/src/routers/project/project.controller.ts
jdorn: I think your change is fine. Let's leave it

packages/front-end/components/Settings/EditDataSource/DataSourceMetrics.tsx
jdorn: Yes, I'm pretty sure archiving an official metric will fail in the back-end already, so we should block it in the UI as well

mknowlton89: Sounds good - just updated this logic to:

```ts
{editMetricsPermissions[metric.id].canDuplicate ||
                          (!metric.managedBy &&
                            editMetricsPermissions[metric.id].canUpdate) ? (
                            <MoreMenu className="px-2">
                              {editMetricsPermissions[metric.id]
                                .canDuplicate ? (
                                <button
                                  className="btn dropdown-item py-2"
                                  onClick={(e) => {
                                    e.stopPropagation();
                                    e.preventDefault();
                                    setModalData({
                                      current: {
                                        ...metric,
                                        managedBy: "",
                                        name: metric.name + " (copy)",
                                      },
                                      edit: false,
                                      duplicate: true,
                                    });
                                  }}
                                >
                                  <FaRegCopy /> Duplicate
                                </button>
                              ) : null}
                              {!metric.managedBy &&
                              editMetricsPermissions[metric.id].canUpdate ? (
                                <button
                                  className="btn dropdown-item py-2"
                                  color=""
                                  onClick={async () => {
                                    const newStatus =
                                      metric.status === "archived"
                                        ? "active"
                                        : "archived";
                                    await apiCall(`/metric/${metric.id}`, {
                                      method: "PUT",
                                      body: JSON.stringify({
                                        status: newStatus,
                                      }),
                                    });
                                    mutateDefinitions({});
                                    mutate();
                                  }}
                                >
                                  <FaArchive />{" "}
                                  {metric.status === "archived"
                                    ? "Unarchive"
                                    : "Archive"}
                                </button>
                              ) : null}
                            </MoreMenu>
                          ) : null}
```

jdorn (approved):

PR #2215, opened 2024-03-11
Update Experimenter & Analyst Permission Doc detail

mknowlton89:
### Features and Changes

Updates our docs to improve the accuracy around data source permissions for Anaylsts and Experimenters.

Previously, our docs indicated that `Experimenter` and `Analyst` roles could only view Data sources, but not edit. This is only partially true - they can't edit certain settings like name, description, projects, or connection params, but they can edit the queries associated with metrics (experiment assignments, etc).

This update reflects that discrepancy within the docs.

This is due to how the `DataSourceForm` submits the `PUT` request. It passes in the full data source object, including the connection params. In the future, we can either move the connection params to a different modal, or update the `PUT` call to only pass in the updated properties, and add some front-end logic to disable fields a user without `createDatasources` permissions shouldn't be able to update (connection parameters).

Then, we could allow those with `editDataSourceSettings` permissions to edit things like name, description, projects, etc.

Link to updated [docs page](https://docs-4vmjfwmsl-growthbook.vercel.app/account/user-permissions).


august-growthbook (approved):
LGTM

PR #2220, opened 2024-03-12
Deploy after CI has ran.

romain-growthbook:
This PR changes the deploy logic to run only after the CI has passed.

When merging PRs, it is always possible that anything can have changed between the last run of the CI on the PR and the actual code being deployed. Thus, running the CI before actually deploying can act as a last safeguard to prevent deploying faulty code to production.

Another good measure to add would be the merge queue which also always run the code against the latest `main` before merging.

### DIFF
--- a/.github/workflows/ci-nochanges.yml
+++ b/.github/workflows/ci-nochanges.yml
@@ -1,13 +1,5 @@
 name: CI
 on:
-  push:
-    branches: main
-    paths-ignore:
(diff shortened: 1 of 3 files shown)


romain-growthbook:
From slack discussion: CI runtime is prohibitive at the moment. Will revisit if it goes down and also look at merge queue when the activity on the repo picks up.

PR #2221, opened 2024-03-12
Minor report fixes

lukesonnet:
### Adds end date component even if empty

Fixes it so that end date always shows even if it is empty (which is a state you can currently get to).

Before
Screenshot 2024-03-12 at 9 25 51 AM

After
Screenshot 2024-03-12 at 9 02 42 AM

### Adds date range to report page

Before
Screenshot 2024-03-12 at 9 25 42 AM

After
Screenshot 2024-03-12 at 9 02 38 AM

### Adds crumbs to reports

Before (notice page header and lack of header crumb)
Screenshot 2024-03-12 at 9 25 26 AM

Going back is also painful
Screenshot 2024-03-12 at 9 25 33 AM

After
Screenshot 2024-03-12 at 9 28 41 AM

Going back is more informative
Screenshot 2024-03-12 at 9 23 52 AM

### DIFF
--- a/packages/front-end/components/Report/ConfigureReport.tsx
+++ b/packages/front-end/components/Report/ConfigureReport.tsx
@@ -304,7 +304,7 @@ export default function ConfigureReport({
             labelClassName="font-weight-bold"
             type="datetime-local"
             {...form.register("startDate")}
-            helpText="Only include users who entered the experiment on or after this date"
+            helpText="Only include users who entered the experiment between the start and end dates"
           />
         </div>
         <div className="col">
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Report/ConfigureReport.tsx
```diff
@@ -308,15 +308,13 @@ export default function ConfigureReport({
...
-          {form.watch("endDate") && (
-            <Field
-              label="End Date (UTC)"
-              labelClassName="font-weight-bold"
-              type="datetime-local"
-              {...form.register("endDate")}
-              helpText="Only include users who entered the experiment on or before this date"
-            />
-          )}
+          <Field
```
bryce-fitzsimons, line 311: Probably should add a button to clear input to include "today". See the Impact Analysis PR for an example of this

bryce-fitzsimons (changes requested):

packages/front-end/components/Report/ConfigureReport.tsx
lukesonnet: done, ty.

bryce-fitzsimons (approved):

PR #2223, opened 2024-03-12
quantile estimation v1

lukebrawleysmith:
This code has 3 primary classes:

Quantile Statistic - creates effect estimate and CI for quantile nu when data are iid (e.g., event-level data).  

Quantile Statistic - creates effect estimate and CI for quantile nu when data are clustered (e.g., user-level data). 

GaussianBayesianABTest - creates posterior distribution using prior on treatment effect and frequentist effect estimate.

### DIFF
--- a/packages/stats/gbstats/bayesian/tests.py
+++ b/packages/stats/gbstats/bayesian/tests.py
@@ -2,6 +2,7 @@ from abc import abstractmethod
 from dataclasses import field
 from typing import List, Optional, Tuple, Union
 
+import scipy.stats
 import numpy as np
 from pydantic.dataclasses import dataclass
 from scipy.stats import norm  # type: ignore
@@ -21,15 +22,25 @@ from gbstats.models.statistics import (
     RatioStatistic,
     SampleMeanStatistic,
     TestStatistic,
+    QuantileStatistic,
+    QuantileStatisticClustered,
 )
-
+from gbstats.frequentist.tests import mean_diff, absolute_var
 
 # Configs
+
+
 @dataclass
 class GaussianPrior:
     mean: float = 0
     variance: float = 1
     pseudo_n: float = 0
+
+
+@dataclass
+class GaussianFlatPrior(GaussianPrior):
+    variance: float = 1e6
+    pseudo_n: float = 1
 
 
 @dataclass
@@ -57,6 +68,20 @@ class GaussianBayesianConfig(BayesianConfig):
     epsilon: float = 1e-4
 
 
+@dataclass
+class EffectConfigFlatPrior(GaussianBayesianConfig):
+    prior_a: GaussianPrior = field(default_factory=GaussianFlatPrior)
+    prior_b: GaussianPrior = field(default_factory=GaussianFlatPrior)
+    prior_effect: GaussianFlatPrior = field(default_factory=GaussianFlatPrior)
+
+
+@dataclass
+class EffectConfigStandardPrior(GaussianBayesianConfig):
+    prior_a: GaussianPrior = field(default_factory=GaussianFlatPrior)
+    prior_b: GaussianPrior = field(default_factory=GaussianFlatPrior)
+    prior_effect: GaussianPrior = field(default_factory=GaussianPrior)
+
+
 # Results
 @dataclass
 class BayesianTestResult(TestResult):
@@ -79,12 +104,15 @@ class BayesianABTest(BaseABTest):
         self,
         stat_a: TestStatistic,
         stat_b: TestStatistic,
-        inverse: bool = False,
-        ccr: float = 0.05,
+        config: BayesianConfig = BayesianConfig(),
     ):
         super().__init__(stat_a, stat_b)
-        self.ccr = ccr
-        self.inverse = inverse
+        self.ccr = config.ccr
+        self.inverse = config.inverse
+        self.relative = config.difference_type == "relative"
+        self.scaled = config.difference_type == "scaled"
+        self.traffic_proportion_b = config.traffic_proportion_b
+        self.phase_length_days = config.phase_length_days
 
     @abstractmethod
     def compute_result(self) -> BayesianTestResult:
@@ -151,7 +179,7 @@ class BinomialBayesianABTest(BayesianABTest):
         stat_b: ProportionStatistic,
         config: BinomialBayesianConfig = BinomialBayesianConfig(),
     ):
-        super().__init__(stat_a, stat_b, config.inverse, config.ccr)
+        super().__init__(stat_a, stat_b, config)
         self.prior_a = config.prior_a
         self.prior_b = config.prior_b
         self.relative = config.difference_type == "relative"
@@ -210,11 +238,21 @@ class BinomialBayesianABTest(BayesianABTest):
 class GaussianBayesianABTest(BayesianABTest):
     def __init__(
         self,
-        stat_a: Union[SampleMeanStatistic, RatioStatistic],
-        stat_b: Union[SampleMeanStatistic, RatioStatistic],
+        stat_a: Union[
+            SampleMeanStatistic,
+            RatioStatistic,
+            QuantileStatistic,
+            QuantileStatisticClustered,
+        ],
+        stat_b: Union[
+            SampleMeanStatistic,
+            RatioStatistic,
+            QuantileStatistic,
+            QuantileStatisticClustered,
+        ],
         config: GaussianBayesianConfig = GaussianBayesianConfig(),
     ):
-        super().__init__(stat_a, stat_b, config.inverse, config.ccr)
+        super().__init__(stat_a, stat_b, config)
         self.prior_a = config.prior_a
         self.prior_b = config.prior_b
         self.epsilon = config.epsilon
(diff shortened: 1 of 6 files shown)


packages/stats/gbstats/bayesian/tests.py
```diff
@@ -313,3 +351,121 @@ def compute_result(self) -> BayesianTestResult:
                 result, self.traffic_proportion_b, self.phase_length_days
             )
         return result
+
+
+class GaussianEffectABTest(GaussianBayesianABTest):
```
lukesonnet, line 356: Should be `BayesianABTest` as the parent class.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -313,3 +351,121 @@ def compute_result(self) -> BayesianTestResult:
...
+            else:
+                self.mean_diff /= np.abs(self.stat_a.mean)
+                self.std_diff /= np.abs(self.stat_a.mean)
+        # probably better to tear these out of superclass, have this be standalone class
+        result = BayesianTestResult(
+            chance_to_win=ctw,
+            expected=self.mean_diff,
+            ci=ci,
+            uplift=Uplift(
+                dist="lognormal" if self.relative else "normal",
```
lukesonnet, line 400: I think it should always be "normal" for this class, right? We're returning the mean and stddev of a normal posterior.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -57,6 +68,20 @@ class GaussianBayesianConfig(BayesianConfig):
...
 
+@dataclass
+class EffectConfigFlatPrior(GaussianBayesianConfig):
+    prior_a: GaussianPrior = field(default_factory=GaussianFlatPrior)
+    prior_b: GaussianPrior = field(default_factory=GaussianFlatPrior)
+    prior_effect: GaussianFlatPrior = field(default_factory=GaussianFlatPrior)
+
+
+@dataclass
+class EffectConfigStandardPrior(GaussianBayesianConfig):
```
lukesonnet, line 79: These two should instead be one type, with BayesianConfig as the parent class e.g.

```
class EffectBayesianConfig(BayesianConfig):
    prior_a: GaussianPrior = field(default_factory=GaussianFlatPrior)
    prior_b: GaussianPrior = field(default_factory=GaussianFlatPrior)
    prior_effect: GaussianPrior = field(default_factory=GaussianFlatPrior)
```

You shouldn't need to create a second class to handle the config for the Effect Bayesian AB test. If you want a different config, you should instantiate `EffectBayesianConfig` with different arguments, e.g.: `EffectBayesianConfig(prior_effect=GaussianPrior(mean=0, variance=10, pseudo_n=100))`

Also update the class __init__ method to just take a dataclass of type `EffectBayesianConfig` since we are trying to stick to one config type per class type.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -313,3 +351,121 @@ def compute_result(self) -> BayesianTestResult:
...
+            QuantileStatistic,
+            QuantileStatisticClustered,
+        ],
+        stat_b: Union[
+            SampleMeanStatistic,
+            RatioStatistic,
+            QuantileStatistic,
+            QuantileStatisticClustered,
+        ],
+        config: Union[EffectConfigFlatPrior, EffectConfigStandardPrior],
```
lukesonnet, line 371: This is what i'm referring to above, this should just be `EffectBayesianConfig`.

lukesonnet (commented):
looking good! some dataclass config feedback, but otherwise it looks close.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -79,12 +104,15 @@ def __init__(
...
-        inverse: bool = False,
-        ccr: float = 0.05,
+        config: BayesianConfig = BayesianConfig(),
     ):
         super().__init__(stat_a, stat_b)
-        self.ccr = ccr
-        self.inverse = inverse
+        self.ccr = config.ccr
+        self.inverse = config.inverse
+        self.relative = config.difference_type == "relative"
```
lukesonnet, line 112: We should remove these 4 lines (relative, scaled, traffic_proportion_b, phase_length_days) from `Gaussian` and `Binomial` BayesianABTest classes. They are redundant now.

packages/stats/gbstats/bayesian/tests.py
lukesonnet: Actually, shouldn't it just be:

```
class EffectBayesianConfig(BayesianConfig):
    prior_effect: GaussianPrior = field(default_factory=GaussianFlatPrior)
```

We don't use prior_a or prior_b in the EffectBayesianABTest class as far as I can tell.

lukebrawleysmith: fixed

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: yes, fixed

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: fixed

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: done

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -2,6 +2,7 @@
 from dataclasses import field
 from typing import List, Optional, Tuple, Union
 
+import scipy.stats
 import numpy as np
 from pydantic.dataclasses import dataclass
 from scipy.stats import norm  # type: ignore
```
lukesonnet, line 8: Let's remove this import and just call `scipy.stats.norm` where I use `norm` below to be consistent with imports.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -57,6 +62,12 @@ class GaussianBayesianConfig(BayesianConfig):
     epsilon: float = 1e-4
 
 
+@dataclass
+class EffectBayesianConfig(BayesianConfig):
+    prior_effect: GaussianPrior = field(default_factory=GaussianPrior)
+    epsilon: float = 1e-4
```
lukesonnet, line 68: Epsilon is used in the lognormal approximation to deal with possibly negative control means, but I don't think we need or use it in the Effect test class, right? Why do we need it here?

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -210,18 +220,24 @@ def compute_result(self) -> BayesianTestResult:
 class GaussianBayesianABTest(BayesianABTest):
     def __init__(
         self,
-        stat_a: Union[SampleMeanStatistic, RatioStatistic],
-        stat_b: Union[SampleMeanStatistic, RatioStatistic],
+        stat_a: Union[
+            SampleMeanStatistic,
+            RatioStatistic,
+            QuantileStatistic,
+            QuantileStatisticClustered,
```
lukesonnet, line 227: Does this class need to take QuantileStatistics? I don't think we ever want to pass it these, right?

packages/stats/gbstats/models/statistics.py
```diff
@@ -194,9 +199,93 @@ def create_joint_statistic(
...
+    q_hat: float  # sample estimate
+    q_lower: float
+    q_upper: float
+    main_sum: float  # numerator sum
+    main_sum_squares: float
+    denominator_sum: float  # denominator sum
+    denominator_sum_squares: float
+    main_denominator_sum_product: float
+
+    # Don't we need to similarly handle cases where the sample size and quantile range would make us go < 0 or > 1?
```
lukesonnet, line 215: Yes, I think that's what the original intent of `n` in `_has_zero_variance()` would be, but I guess that requires a bit more context than you have now. maybe we can do that in a follow-up PR.

packages/stats/gbstats/models/statistics.py
```diff
@@ -194,9 +199,93 @@ def create_joint_statistic(
...
+    def mean(self) -> float:
+        return self.q_hat
+
+    @property
+    def unadjusted_mean(self) -> float:
+        return self.mean
+
+    @property
+    def variance_init(self) -> float:
+        multiplier = scipy.stats.norm.ppf(0.975, loc=0, scale=1)
```
lukesonnet, line 232: Doesn't 0.975 need to be computed from some `alpha` from the configuration? Maybe `QuantileStatistic` also needs to take `alpha` as an input and use it here and use it when computing the `n` for the issue of requesting a quantile < 0 or > 1 (the other comment above?)

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: done

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: removed it

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: good catch,  I modified it as this was the first way I implemented Bayesian quantile approach, and I forgot to remove QuantileStatistics later.

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: should be included now

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: I thought we agreed that we would always use .975 when estimating uncertainty.  Happy to chat more about this.

packages/stats/gbstats/bayesian/tests.py
lukesonnet: I think then we can remove this import that isn't used anymore.

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: right, i updated after our convo

lukesonnet:
A few nits, a few slightly bigger questions. I think once there's a few unit tests in place we can land this in anticipation of merging it with the SQL PR.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -313,3 +323,141 @@ def compute_result(self) -> BayesianTestResult:
...
+                self.stat_b.mean,
+                self.stat_b.stddev,
+                self.stat_b.n,
+            ],
+        )
+        self.mean_diff = mu_b - mu_a
+        self.std_diff = np.sqrt(sd_a**2 + sd_b**2)
+        risk = [self.risk]
+        self.var_diff = frequentist_variance(
+            sd_a**2, mu_a, 1, sd_b**2, mu_b, 1, self.relative
```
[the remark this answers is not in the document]
lukebrawleysmith: yes, because `var_a` in `frequentist_variance` refers to the distributional variance, and divides by `n_a` to get the variance of the mean.  `Norm.posterior` outputs the variance for the mean, so we don't want to divide it again.

packages/stats/gbstats/models/statistics.py
```diff
@@ -194,9 +199,91 @@ def create_joint_statistic(
     )
 
 
+@dataclass
+class QuantileStatistic(Statistic):
+    n: int  # number of events here
+    nu: float
+    ccr: float
+    q_hat: float  # sample estimate
```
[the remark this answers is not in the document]
lukebrawleysmith: done

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: Suppose n = 100, the distribution is continuous, and `Q(nu_high)` = `Q(nu_low)`.   That means there were no sample observations that fell in about 100*(`nu_high` - `nu_low`)% of the distribution, which is at least 4 percent of the distributional mass (it's smallest for extreme &#nu).  The probability of 0 observations falling in this bin that has roughly 4% of the mass is 0.016.  So this is a low event probability.  In fact, it's even lower, because the separate validation check you mention above would exclude inference for extreme quantile levels if n = 100.

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: updated to alpha

packages/stats/gbstats/models/statistics.py
lukesonnet: Sounds reasonable to me. Thanks for the other checks!

packages/stats/gbstats/models/statistics.py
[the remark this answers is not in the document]
lukesonnet: I don't think you made this change?

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -313,3 +323,141 @@ def compute_result(self) -> BayesianTestResult:
...
+            v = (
+                self.prior_effect.variance
+                * mu_0**2
+                / (1.0 + self.prior_effect.mean**2)
+            )
+        else:
+            mu_1 = mu_0 + self.prior_effect.mean
+            v = 0.5 * self.prior_effect.variance
+        self.prior_a = GaussianPrior(mean=mu_0, variance=v, pseudo_n=1)
+        self.prior_b = GaussianPrior(mean=mu_1, variance=v, pseudo_n=1)
```
lukesonnet, line 440: Should the pseudo_n from the `GaussianPrior` be passed through here?

lukesonnet: I only ask because it seems like there isn't a way to pass through a flat, improper prior to this test class, right? Should we set up a way to do that?

You can kind of hack it into the existing `GaussianPrior` where if `prior_effect.pseudo_n` = 0, then we just pass this through like I suggest. Or we can be more explicit, where you can also specify no prior and that modifies this method as well.

lukebrawleysmith: good idea, I implemented this.  When we use non-flat priors in the future, we still want to use pseudo-n = 1.

packages/stats/gbstats/models/statistics.py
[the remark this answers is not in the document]
lukebrawleysmith: apologies, forgot to save my file, will be in latest push.

packages/stats/tests/bayesian/test_tests.py
```diff
@@ -156,5 +161,145 @@ def test_inexact_log_approximation(self):
...
+
+    def test_missing_data(self):
+        result = GaussianBayesianABTest(
+            SampleMeanStatistic(sum=0, sum_squares=0, n=0),
+            SampleMeanStatistic(sum=0, sum_squares=0, n=0),
+        ).compute_result()
+        self.assertEqual(result.chance_to_win, 0.5)
+        self.assertEqual(result.expected, 0)
+
+    def test_inexact_log_approximation(self):
```
lukesonnet, line 284: Why do we need to replicate this test here? Maybe copy paste?

packages/stats/gbstats/gbstats.py
```diff
@@ -211,20 +216,40 @@ def get_configured_test(
                 ),
             )
     else:
+        assert isinstance(
+            stat_a, type(stat_b)
+        ), "stat_a and stat_b must be of same type."
```
lukesonnet, line 221: Sorry to nit, I should have pointed this out initially, but this will pass even so long as stat_a is the same class *or a subclass* of stat_b. e.g is stat_a is a SampleMeanStatistic and stat_b is a Statistic, this will pass.

So I think we should instead just do `assert type(stat_a) is type(stat_b)`

lukebrawleysmith: accepted

packages/stats/tests/bayesian/test_tests.py
lukebrawleysmith: good catch, i removed this

lukesonnet (approved):
Looks good to me! There were two further nits that I just encountered when reading your PR again (sorry for missing these earlier), but you can accommodate them by first merging in my PR to your branch: https://github.com/growthbook/growthbook/pull/2288

PR #2225, opened 2024-03-13
Remove use of relative imports in the backend.

romain-growthbook:
This PR removes relative imports from the backend following a similar work done in https://github.com/growthbook/growthbook/pull/2140

All compilation is done using `tsc` so this seems fine to use in production. 

The only complication this introduces is the requirement to be able to track `back-end` code from other packages. The pattern proposed here follows the convention used in the front-end to do `@back-end` and `@front-end`. 

It seems that in other places we also use `import ... from "shared/.."` to import from the shared package. We could either change this convention to `@shared` or else use `import ... from "back-end/.."` and the same for the `front-end`.


bttf (commented):
No glaring mistakes I can find ... Will let @jdorn give the approval

PR #2226, opened 2024-03-13
Improves Attribute filtering by project

mknowlton89:
### Features and Changes

With the recently introduced feature of project-scoping attributes, we weren't properly filtering the list of attributes all the way in a few different places, based on the resource's project.

Ex:
For a feature that is in project A, when creating a rule, we should automatically filter the list of attributes to only attributes available to Project A (which would be any attribute in project A, or any attribute in "All Projects".

Previously, we were filtering based on a user's readAccess, but not also by project.

### Testing

- [x] Attempt to add a rule (experiment, force, and rollout) to a feature that is a part of "All Projects" and ensure that every non-archived attributes is available in the dropdown.
- [x] Now, do the same, but for a feature in "Project A" and ensure that only attributes associated with Project A are available in the dropdown. (Attributes associated with project A are either attributes that have "Project A" in it's projects list, or attributes that are a part of "All Projects").
- [x] Now, select "All Projects" on the sidenav, and attempt to create a new experiment, ensure that the list of attributes available in the dropdown is the entire list of non-archived attributes.
- [x] Now, switch to "Project A" on the sidenav, and attempt to create a new experiment, ensure that the list of attributes is only those that are associated with Project A, or in all projects (and are not archived).
- [x] Now, select "All Projects" on the sidenav, and go to create a new Saved Group - ensure that the list of attributes available in the dropdown is the entire list of non-archived attributes.
- [x] Now, switch to "Project A" on the sidenav, and attempt to create a Saved Group, ensure that the list of attributes is only those that are associated with Project A, or in all projects (and are not archived).
- [x] Ensure that no new regressions where introduced around `no_access` roles, as a result of these changes


mknowlton89:
> Saved groups are not project-scoped right now.
> 
> Also, should add to your list of tests to make sure editing an existing rule that references an out-of-project attribute continues to work. It may switch them to advanced mode, which is fine, as long as it's still editable.

Great suggestion. I've removed the filtering from saved groups until we add project scoping to saved groups. I also updated the rule modal so if a user is editing an existing rule that references an attribute no longer associated with the feature's project, it automatically switches to advanced mode. Previously, it was displaying an empty `SelectField` since the `SelectField` component was trying to select the `option` via `value`, but it was not able to find a match, so it just rendered the empty state.

packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -173,6 +173,18 @@ export default function ConditionInput(props: Props) {
...
+            const attributeOptions = filteredAttributeSchema.map((s) => ({
+              label: s.property,
+              value: s.property,
+            }));
+
+            // if the field is not in the attributeOptions list, switch to advanced mode
+            if (
+              !attributeOptions.find((attribute) => attribute.value === field)
+            ) {
+              setAdvanced(true);
```
jdorn, line 185: Should not set state during rendering.  Would be better to put this logic in the initial useState value for `advanced`

```ts
  const [advanced, setAdvanced] = useState(
    () => jsonToConds(props.defaultValue, attributes) === null
  );
```

Easiest solution would be to filter `attributes` before passing this in (there's already logic in `jsonToConds` to return `null` when an attribute is missing there)

mknowlton89: @jdorn Ah - I see - that makes sense.

I just pushed an update in this commit that moves the filtering logic to the useAttributeSchema hook. Let me know what you think of that pattern.

After looking in the app, it looks like there are a few other places we might want to filter the attributes based on the project (experiment assignment selector, feature flag testing tool, etc). That can be done in a separate PR, I just wanted to call it out.

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -168,8 +168,16 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
 
   const attributeSchema = useAttributeSchema();
+
+  const filteredAttributeSchema = attributeSchema.filter((attribute) => {
+    return (
+      !project ||
+      !attribute.projects?.length ||
+      attribute.projects.includes(project)
+    );
+  });
```
jdorn, line 178: Should this just pass project into `useAttributeSchema` instead of recreating the filtering logic here?

jdorn (approved):

PR #2235, opened 2024-03-15
Updates addComment as a READ_ONLY type permission.

mknowlton89:
### Features and Changes

This PR makes the `addComment` permission a `READ_ONLY_TYPE`. Meaning, if the user has the permission either globally, or in atleast 1 project a resource is associated with, they have the permission.

This means, if a user doesn't have the `addComment` permission through their global role, but atleast 1 of their project-specific roles does grant them permission, they'd be able to, for instance, add a comment on a metric that is in `All Projects`. Before this, their global role would've prevent this from happening. The metric would've needed to have specific projects associated with it, and the user would've needed to have the permission for every one of those projects.

Now, let's take the same situation, the `createMetrics` permission does require the user to have permission in every project a metric is associated with, so the user wouldn't have permission to edit this metric. But now, they could make a comment, for instance, to request a slight change to the metric.

### DIFF
--- a/packages/back-end/src/controllers/discussions.ts
+++ b/packages/back-end/src/controllers/discussions.ts
@@ -15,7 +15,7 @@ export async function postDiscussions(
   >,
   res: Response
 ) {
-  req.checkPermissions("addComments", "");
+  req.checkPermissions("addComments", []);
 
   const { org, userId, email, userName } = getContextFromReq(req);
   const { parentId, parentType } = req.params;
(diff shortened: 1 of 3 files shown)


jdorn (approved):
I don't love the name "readonly" anymore for this, but we can change it later.  It's really a list of permissions with loose vs strict project checking.  It just so happens that most of the time, read permissions use loose checks and write permissions use strict checks.

Issue #2231, opened 2024-03-15
[Feature] Add possibility to set timeout for Data Source connections (Clickhouse)


ndubitsk:
### Description of Feature

Some of our queries are heavy and they can't get through 60 seconds limit (that is set by default in Clickhouse connection).
We get "504: 504 Gateway Time-out" when we calculate results of an experiment or adding new Fact Table.

We would like to have a possibility to change this setting in Data Sources -> Connection Info

<img width="1385" alt="Screenshot 2024-03-15 at 14 19 59" src="https://github.com/growthbook/growthbook/assets/79838299/a79f31f7-fd13-4166-8a97-9446aa1ee0f6">

lukesonnet:
Hi @ndubitsk . I think this is included as of #2236. Closing this but feel free to re-open if it isn't working for you.

ndubitsk:
@lukesonnet Hey!

Thank a lot!
Looks like this feature works totally fine for queries, that are run in the analytics section.
But new limits seems like doesn't work when we are trying to update SQL query of a Fact Table.

Shall I open a new Feature or a Bug request for this issue?

ndubitsk:
@lukesonnet 
Actually, I just checked and we are getting the Timeout error but query was running only for ~1 minute.
We are expecting that query can run up until 10 minutes as I understand.

<img width="812" alt="Screenshot 2024-03-26 at 10 53 17" src="https://github.com/growthbook/growthbook/assets/79838299/a8177262-2472-4
<img width="1387" alt="Screenshot 2024-03-26 at 10 55 42" src="https://github.com/growthbook/growthbook/assets/79838299/39685807-a67b-47cd-b98b-8a4352ef8ab1">
dfe-a765-bf5b3b61bfea">

lukesonnet:
Long internal thread, but it seems like the main culprit here is a load balancer on the experimenter's Clickhouse infrastructure timing out the request.

PR #2236, opened 2024-03-15
Migrate to Official Clickhouse Client

### LINKED ISSUE #2130: [Bug] Constantly getting 400 "Bad Request" error, when adding capping to the Ratio metric
### Summary
We are getting error while calculating ratio metrics in experiments results.

Maybe the problem is somehow related to capping (ratio-metrics + capping)?

Because without capping a simple ratio metric "ARPPU" works fine, and with capping turned on it fails with the same strange error "Browser sent an invalid request" (it happens with any browser)

There is an example below

image (2)
image (3)
image
image (1)

### LINKED ISSUE #2194: Move from unofficial ClickHouse client to official one
### Description of Feature

Now GrowthBook relies on unofficial ClickHouse client [clickhouse](https://www.npmjs.com/package/clickhouse), which hasn't been maintained for 2 years already!

But there is official client [@clickhouse/client](https://www.npmjs.com/package/@clickhouse/client) from the developers of ClickHouse

Most of the changes, that have to be made would be located in this file: [packages/back-end/src/integrations/ClickHouse.ts](https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/integrations/ClickHouse.ts)

lukesonnet:
### Features and Changes

Move to official clickhouse client.

Closes #2194 

May resolve #2130 error which seems to be emitted from the clickhouse client we were using before.

* Also adds statistics for clickhouse queries, which is a nice side-effect.
* Also adds "GrowthBook" as the application making the request
* Also adds timeout (min 0, max 1800, 30 min) (closing #2231)

### Testing

Ran test, experiment metric and fact metric queries, and traffic query for ClickHouse. All works as expected.

### Screenshots

Statistics show up:
Screenshot 2024-03-15 at 11 14 47 AM

Timeout setting:
Screenshot 2024-03-15 at 12 22 59 PM

Timeout works:
Screenshot 2024-03-15 at 12 21 20 PM

### DIFF
--- a/packages/back-end/src/integrations/ClickHouse.ts
+++ b/packages/back-end/src/integrations/ClickHouse.ts
@@ -1,4 +1,4 @@
-import { ClickHouse as ClickHouseClient } from "clickhouse";
+import { createClient } from "@clickhouse/client";
 import { decryptDataSourceParams } from "../services/datasource";
 import { ClickHouseConnectionParams } from "../../types/integrations/clickhouse";
 import { QueryResponse } from "../types/Integration";
@@ -28,28 +28,20 @@ export default class ClickHouse extends SqlIntegration {
     return ["password"];
   }
   async runQuery(sql: string): Promise<QueryResponse> {
-    const client = new ClickHouseClient({
-      url: this.params.url,
-      port: this.params.port,
-      basicAuth: this.params.username
-        ? {
-            username: this.params.username,
-            password: this.params.password,
-          }
-        : null,
-      format: "json",
-      debug: false,
-      raw: false,
-      config: {
-        database: this.params.database,
+    const client = createClient({
+      host: `${this.params.url}${
+        this.params.port ? `:${this.params.port}` : ""
+      }`,
+      username: this.params.username,
+      password: this.params.password,
+      database: this.params.database,
+      additional_headers: {
+        "x-clickhouse-format": "JSON",
       },
-      reqParams: {
-        headers: {
-          "x-clickhouse-format": "JSON",
-        },
-      },
+      application: "GrowthBook",
     });
-    return { rows: Array.from(await client.query(sql).toPromise()) };
+    const results = await client.query({ query: sql });
+    return { rows: Array.from(await results.json()) };
   }
   toTimestamp(date: Date) {
     return `toDateTime('${date
--- a/packages/back-end/src/integrations/ClickHouse.ts
+++ b/packages/back-end/src/integrations/ClickHouse.ts
@@ -36,6 +36,9 @@ export default class ClickHouse extends SqlIntegration {
       password: this.params.password,
       database: this.params.database,
       application: "GrowthBook",
+      clickhouse_settings: {
+        max_execution_time: Math.min(this.params.maxExecutionTime ?? 240, 1800),
+      },
     });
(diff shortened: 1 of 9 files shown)


packages/back-end/src/integrations/ClickHouse.ts
```diff
@@ -28,28 +28,31 @@ export default class ClickHouse extends SqlIntegration {
...
-        database: this.params.database,
-      },
-      reqParams: {
-        headers: {
-          "x-clickhouse-format": "JSON",
-        },
+    const client = createClient({
+      host: `${this.params.url}${
+        this.params.port ? `:${this.params.port}` : ""
+      }`,
```
jdorn, line 34: The old Clickhouse client supported specifying a port in the URL while also passing in a `port` setting.  For example, this would work fine before, but would break today:

```ts
{
  url: "http://localhost:1234",
  port: "1234"
}
```

There was also some weird behavior in the old client if you passed in different ports in both places, but I don't think we need to support that edge case.

jdorn: An easy solution to this is to use the URL class.

```ts
function getHost(url: string | undefined, port: number) {
  if (!url) return undefined;
  const host = new URL(url);
  if (!host.port && port) host.port = port + "";
  return host.origin;
}
// ...

const client = createClient({
  host: getHost(this.params.url, this.params.port),
```

lukesonnet: One issue with this is that `URL(url)` will throw an error if `url` isn't a proper URL. Should I just catch this case and do my pasting approach or should we consider some better validation when the form is saved (I don't think we do any right now).

lukesonnet: for example, do something like this to catch some edge cases

```
export function getHost(url: string | undefined, port: number): string | undefined {  
  if (!url) return undefined;  
  try {
    const host = new URL(url);  
    if (!host.port && port) host.port = port + "";  
    return host.origin;  
  } catch (e) {
    return `${url}:${port}`;
  }
}  
```

packages/back-end/src/integrations/ClickHouse.ts
```diff
@@ -28,28 +28,31 @@ export default class ClickHouse extends SqlIntegration {
...
+      username: this.params.username,
+      password: this.params.password,
+      database: this.params.database,
+      application: "GrowthBook",
+      clickhouse_settings: {
+        max_execution_time: Math.min(this.params.maxExecutionTime ?? 240, 1800),
```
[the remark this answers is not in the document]
lukesonnet: Yeah, you're right. Not sure what I was thinking. Set it to 30 minutes, increased max to 60 minutes.

packages/back-end/src/integrations/ClickHouse.ts
jdorn: I think it's fine to just let it throw.  I'm pretty sure if they have an invalid URL today it will be broken, so this shouldn't introduce any regressions.

lukesonnet:
new stats look good
Screenshot 2024-03-19 at 9 35 07 AM

default now showing in form
Screenshot 2024-03-19 at 9 34 04 AM

PR #2239, opened 2024-03-18
Webhook test event

romain-growthbook:
This PR adds a new "test" button to webhooks to test them.

The implementation creates a new event of type `webhook.test`. This event is created in a webhook-specific endpoint and routed through the regular event notifier but notification is explicitely disable for slack integration.

### Demo

https://www.loom.com/share/f85f8f20e3f84513b9452bf55b8432a9

### DIFF
--- a/packages/front-end/components/EventWebHooks/EventWebHookDetail/EventWebHookDetail.tsx
+++ b/packages/front-end/components/EventWebHooks/EventWebHookDetail/EventWebHookDetail.tsx
@@ -13,6 +13,7 @@ import { useAuth } from "@/services/auth";
 import Badge from "@/components/Badge";
 import { useCopyToClipboard } from "@/hooks/useCopyToClipboard";
 import useApi from "@/hooks/useApi";
+import { useEventWebhookLogs } from "@/hooks/useEventWebhookLogs";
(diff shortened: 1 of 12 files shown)


packages/front-end/components/EventWebHooks/EventWebHookDetail/EventWebHookDetail.tsx
```diff
@@ -72,6 +135,14 @@ export const EventWebHookDetail: FC<EventWebHookDetailProps> = ({
             Edit
           </button>
 
+          <button
+            onClick={onTestWebhook}
+            className="btn btn-sm btn-outline-secondary mr-1"
+          >
+            <FaPaperPlane className="mr-1" />
+            Test
+          </button>
```
jdorn, line 144: We should probably have a loading state and disable this button.  Posting to Discord/Slack will almost always be fast, but some custom endpoints might take awhile and it's good to have UI feedback during that.

You can either implement the loading/disable part yourself or use our `<Button>` instead component which takes an async `onClick` property and will switch to a loading/disabled state while it awaits the promise automatically

jdorn (approved):

PR #2246, opened 2024-03-19
Clear up `@ts-expect-error`

romain-growthbook:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Experiment/ResultsDownloadButton.tsx
+++ b/packages/front-end/components/Experiment/ResultsDownloadButton.tsx
@@ -79,16 +79,12 @@ export default function ResultsDownloadButton({
           };
           const stats = variation.metrics[m];
           if (!stats) return;
-          const { relativeRisk } = getRiskByVariation(
-            index,
-            row,
-            metricDefaults
-          );
+          const risk = getRiskByVariation(index, row, metricDefaults);
           csvRows.push({
             ...(dimensionName && { [dimensionName]: result.name }),
             metric: metric?.name,
             variation: variations[index].name,
-            riskOfChoosing: relativeRisk,
+            riskOfChoosing: risk.show ? risk.relativeRisk : undefined,
             users: stats.users,
             totalValue: stats.value,
             perUserValue: stats.cr,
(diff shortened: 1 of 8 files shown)


packages/front-end/components/Experiment/ResultsDownloadButton.tsx
```diff
@@ -79,16 +79,12 @@ export default function ResultsDownloadButton({
...
-            row,
-            metricDefaults
-          );
+          const risk = getRiskByVariation(index, row, metricDefaults);
           csvRows.push({
             ...(dimensionName && { [dimensionName]: result.name }),
             metric: metric?.name,
             variation: variations[index].name,
-            riskOfChoosing: relativeRisk,
+            riskOfChoosing: risk.show ? risk.relativeRisk : undefined,
```
romain-growthbook, line 87: Is that intended? Should the value be `risk`?

packages/front-end/services/experiments.ts
```diff
@@ -123,29 +124,37 @@ export function getRisk(
...
 
+type Risk = { risk: number } & (
+  | {
+      show: true;
+      relativeRisk: number;
+    }
+  | {
+      show: false;
+    }
+);
```
romain-growthbook, line 135: The code logic is to always have `relativeRisk` set when `showRisk` is `true` so I changed the payload to reflect that.

packages/front-end/services/experiments.ts
```diff
@@ -123,29 +124,37 @@ export function getRisk(
...
-  let relativeRisk: number;
   let showRisk = false;
   const baseline = row.variations[0];
 
   if (riskVariation > 0) {
     const stats = row.variations[riskVariation];
-    // @ts-expect-error TS(2322) If you come across this, please fix it!: Type 'number | undefined' is not assignable to typ... Remove this comment to see the full error message
-    risk = stats?.risk?.[row.metric.inverse ? 0 : 1];
-    riskCR = stats?.cr;
+    risk = ensureAndReturn(stats.risk?.[row.metric.inverse ? 0 : 1]);
```
romain-growthbook, line 149: Is that safe?

PR #2256, opened 2024-03-21
Remove designer page.

romain-growthbook:
This PR removes the designer page. This code was originally part of the `ts-expect-error` cleanup but it was advised that it should be removed altogether.


jdorn (commented):
There's a banner in VisualChangesetTable.tsx that links to this page we should remove too.

Also, there are 2 other references to this page, but they are minor.

Do a global search for `/designer` and you should see all 3 of these.

romain-growthbook:
> There's a banner in VisualChangesetTable.tsx that links to this page we should remove too.
> 
> Also, there are 2 other references to this page, but they are minor.
> 
> Do a global search for `/designer` and you should see all 3 of these.

I found two references and cleaned them up along with some orphan `.ts` files.

Let me know if you see anything else.

jdorn (approved):

PR #2263, opened 2024-03-22
Move addComments to permissionClass

mknowlton89:
### Features and Changes

This PR moves the `addComments` permission check to the new permissionClass. This is the first `READ_ONLY` type permission that we've migrated, so there are updates to the `checkProjectFilterPermissions` method.

If the resource is in "All Projects", we'll add the user's project-specific roles to the `projects` array, as the user only needs to have the permission globally, or in at least 1 project in order to have the permission.


packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -38,6 +39,10 @@ export class Permissions {
     return this.checkProjectFilterPermission(metric, "createMetrics");
   };
 
+  public canAddComment = (projects: string[]): boolean => {
```
mknowlton89, line 42: open to suggestions on how to type the arguments here - we check for this permission across metrics (with have `projects`, along with ideas, experiments, and features, which only have a `project` property. 

It'd be nice to standardize so the engineer doesn't have to have much context when calling `canAddComment`. 

Currently, it still seems like a step in the right direction, as before it could be a string, or undefined, or string[], so this at least simplifies it a bit.

jdorn: One option is to type the argument like this:

```ts
type ObjWithProject = {project?: string; projects?: string[]};
```

That way, you could call it directly with the thing you want to comment on:
```
canAddComment(feature);
canAddComment(metric);
```

Another option is to make separate `canCommentOnFeature`, `canCommentOnMetric`, etc. methods.  They could all check the same `addComments` permission for now, but it leaves the door open for us to make this finer grained in the future.

jdorn: I'm also fine with how you have it

jdorn (approved):

PR #2266, opened 2024-03-22
Quantile Testing - Data model + FE changes (feature flagged)

lukesonnet:
- [ ] Change [fmid] Metric Settings for quantile metrics

### DIFF
--- a/packages/back-end/src/models/FactMetricModel.ts
+++ b/packages/back-end/src/models/FactMetricModel.ts
@@ -178,7 +178,6 @@ export function toFactMetricApiInterface(
   factMetric: FactMetricInterface
 ): ApiFactMetric {
   const {
-    quantileSettings,
     cappingSettings,
     windowSettings,
     regressionAdjustmentDays,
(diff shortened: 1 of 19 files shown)


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -183,8 +189,13 @@ export function toFactMetricApiInterface(
     ...otherFields
   } = omit(factMetric, ["organization"]);
 
+  const metricType = otherFields.metricType;
+  if (metricType === "quantile") {
+    throw new Error("Quantile metrics not avaiable yet");
+  }
```
jdorn, line 195: Doesn't seem like we should throw here.  If someone does happen to create a quantile metric in Mongo, this code will cause the `GET /fact-metrics` endpoint to fail, so they won't even be able to get a list of metrics from the API anymore.

lukesonnet: Ok! Seems reasonable enough to remove this for now. DB messers beware.

lukesonnet: I just set metricType to "mean" if it is "quantile" for now!

jdorn (approved):

PR #2270, opened 2024-03-25
Remove double navigation in the experiments list

BarreiroT:
### Features and Changes

This PR removes the `onClick` handler from the table element in the experiment list. This handler overrode default behavior, which caused the current tab to navigate to the location when using `ctrl/cmd + click`, for example.

### Testing

You can go to the experiment list before and after the changes, `ctrl/cmd + click` on an experiment name, and notice you will open a new tab with the experiment and navigate to the experiment on the current tab.

### Screenshots
HTML before the changes:
image

HTML after the changes:
image

List before the changes:
image

List after the changes:
image

It's worth noting the clickable area is now smaller since the `td` element padding is no longer clickable.

### DIFF
--- a/packages/front-end/pages/experiments/index.tsx
+++ b/packages/front-end/pages/experiments/index.tsx
@@ -355,55 +355,46 @@ const ExperimentsPage = (): React.ReactElement => {
                         type="icon"
                       />
                     </td>
-                    <td
-                      onClick={() => {
-                        router.push(`/experiment/${e.id}`);
-                      }}
-                      className="cursor-pointer"
-                      data-title="Experiment name:"
-                    >
-                      <div className="d-flex flex-column">
-                        <div className="d-flex">
-                          <Link
-                            href={`/experiment/${e.id}`}
-                            className="testname"
-                          >
-                            {e.name}
-                          </Link>
-                          {e.hasVisualChangesets ? (
-                            <Tooltip
-                              className="d-flex align-items-center ml-2"
-                              body="Visual experiment"
+                    <td data-title="Experiment name:">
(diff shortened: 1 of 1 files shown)


packages/front-end/pages/experiments/index.tsx
```diff
@@ -355,55 +355,46 @@ const ExperimentsPage = (): React.ReactElement => {
...
-                            >
-                              <BsFlag className="text-blue" />
-                            </Tooltip>
-                          ) : null}
-                          {e.hasURLRedirects ? (
-                            <Tooltip
-                              className="d-flex align-items-center ml-2"
-                              body="URL Redirect experiment"
+                    <td data-title="Experiment name:">
+                      <Link href={`/experiment/${e.id}`}>
```
jdorn, line 359: I think we can get back the bigger click area pretty easily by moving the padding to the link.

```suggestion
                    <td data-title="Experiment name:" className="p-0">
                      <Link
                        href={`/experiment/${e.id}`}
                        className="d-block p-2"
                      >
```

BarreiroT: That's a good point. Thank you!
I committed your suggestion since it's better than having a considerably smaller clicking area. However, I noticed the `td` padding is `0.75rem`, [which is not supported by the bootstrap spacing classes](https://getbootstrap.com/docs/5.0/utilities/spacing/#notation). We could add the correct padding using
```
style={{ padding: '0.75rem' }}
```
Since I don't know if you would prefer an aligned but custom style vs a slightly smaller one that uses classes, I will add another comment with the suggestion that you can commit or ignore.

PR #2272, opened 2024-03-25
Correct SDK version in URL redirect modal SDK connection warning

msamper:
### Features and Changes


gazzdingo (approved):
LGTM is this also correct on the docs?

msamper:
> LGTM is this also correct on the docs?

The current docs don't reference the minimum versions to be able to use Redirects, but I'll add that info in a follow up PR after this along with JS SDK documentation and `build-your-own.mdx` updates

PR #2273, opened 2024-03-25
Parametrize data-generator, typecheck before run.

romain-growthbook:
This PR parametrizes the data generation script to be able to set paremeters as CLI arguments.

It also runs the script with the full typescript check, which revealed one weird spot highlighted in the comments.

### DIFF
--- a/packages/back-end/test/data-generator/new-generator.ts
+++ b/packages/back-end/test/data-generator/new-generator.ts
@@ -379,6 +379,10 @@ function purchase(
     ),
   });
   trackExperiment(data, res, "confirmation-email", sim);
+  sim.dataTables.userRetention[parseInt(data.userId)] += normalInt(
+    res.value - 10,
+    res.value + 10
+  );
 }
 
 function getTimestamp(currentDate: Date) {
@@ -591,7 +595,7 @@ const {
     "user-count",
     "csv-dir",
     "messy-data",
-    "key-prexix",
+    "key-prefix",
   ],
   default: {
     days: "60",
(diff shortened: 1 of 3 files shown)


packages/back-end/test/data-generator/new-generator.ts
```diff
@@ -378,7 +379,6 @@ function purchase(
     ),
   });
   trackExperiment(data, res, "confirmation-email", sim);
-  sim[parseInt(data.userId)] += normalInt(res.value - 10, res.value + 10);
```
romain-growthbook, line 381: This was raised by the type checker. `sim` is an object holding simulation params. Not sure what this line is/was supposed to achieve..

lukesonnet: Thanks for finding this, looks like a bug that I landed. The intent is to make it so that the `"confirmation-email"` experiment changes the likelihood that user returns, which is handled by the `userRetention` object.

The line that you've deleted does nothing, I would think, but we should find a way to restore this functionality. I think the following might be the right way to do this:

```
sim.dataTables.userRetention[parseInt(data.userId)] += normalInt(res.value - 10, res.value + 10);
```

See similar functionality here: https://github.com/growthbook/growthbook/pull/2273/files#diff-a7e9d8cb1781870b7a4cb2d0dd0b12ef8cb0eac22867638f9d037a15898605c1L490

packages/back-end/test/data-generator/new-generator.ts
```diff
@@ -570,5 +577,50 @@ function generateAndWriteData(
...
+  "messy-data": messyDataArg,
+  "key-prefix": keyPrefixArg,
+} = parseArgs(process.argv.slice(2), {
+  string: [
+    "days",
+    "start-date",
+    "user-count",
+    "csv-dir",
+    "messy-data",
+    "key-prexix",
```
lukesonnet, line 594: typo

lukesonnet (approved):
LGTM!

PR #2277, opened 2024-03-26
Adds permission helpers for canReview and bypassApprovalChecks

mknowlton89:
### Features and Changes

This PR adds a permission check helper for `canRevew` and `bypassApprovalChecks` permissions to the permission class, while updating all instances of checking this permission in the app.

### Testing

- See unit test suite


packages/back-end/src/controllers/features.ts
```diff
@@ -585,7 +585,10 @@ export async function postFeatureReviewOrComment(
   if (!feature) {
     throw new Error("Could not find feature");
   }
-  req.checkPermissions("canReview", feature.project);
+
+  if (!context.permissions.canReviewFeatureDrafts(feature)) {
+    context.permissions.throwPermissionError();
```
gazzdingo, line 590: nit: but should we add a message to permission error eg
`throwPermissionError("you dont have permissions to review this")`

gazzdingo (approved):
LGTM just one comment on the error throw

packages/back-end/src/controllers/features.ts
mknowlton89: @gazzdingo `throwPermissionError` is currently returning a message when it's called.

```ts
public throwPermissionError(): void {
    throw new PermissionError(
      "You do not have permission to perform this action"
    );
  }
  ```
  
  We are planning to add some in-mem caching within the `permissionClass` that will retain the last checked permission, and we'll eventually use that to provide a more contextual error. Something like: `"You don't have bypassApprovalChecks permission."`

PR #2279, opened 2024-03-26
Add canCreateFeatureDrafts helper

mknowlton89:
### Features and Changes

This PR adds a helper function to the `permissionClass` to handle checking permissions for `createFeatureDrafts`.

Additionally, this PR renames that permission to `manageFeatureDrafts` as it's more contextually accurate, as this permission is responsible for checking whether a user can create, update, or delete feature drafts.

I did explore breaking this permission out into separate create, update, and delete methods (e.g. canCreateFeatureDrafts, canUpdateFeatureDrafts, canDeleteFeatureDrafts), but it felt a bit too granular. I ultimately didn't go that route as I don't see a really strong use case for a user being able to update a draft, but not create a draft.

Especially because currently, the endpoints all handle that logic today - ex: if a user is adding/updating a rule, when we go to get the revision, we look to see if the current version is the live version, and if so, we create a new draft, otherwise, we return the correct draft. So if we want to get granular, we'll need to update that logic so check the correct permissions.

But ultimately, it feels like supporting that logic and supporting our organizations trying to do that will be really difficult, so I'm not sure the juice is worth the squeeze.

Open to feedback, of course, as there may be legitimate use-cases that I haven't explored yet. 

### Testing

- See test suite


packages/front-end/components/Features/FeatureModal/FeatureFromExperimentModal.tsx
```diff
@@ -172,7 +174,11 @@ export default function FeatureFromExperimentModal({
   let ctaEnabled = true;
   let disabledMessage: string | undefined;
 
-  if (!permissions.check("createFeatureDrafts", project)) {
+  if (
+    !permissionsUtil.canManageFeatureDrafts({
+      project: experiment.project || project,
```
mknowlton89, line 179: This fixes a bug.

When duplicating a feature, we were checking the current active project (via the drop-down in the left nav). Now, if the user is trying to duplicate a feature, we check the project the feature is in.

This was a bug if the user had a global `noaccess` role, and multiple project-specific roles - if they had `All Projects` selected in the left-nav, it was then checking the user's global role for permission, when the feature was trying to be created in the experiment's project.

packages/front-end/components/Features/FeatureModal/index.tsx
```diff
@@ -171,7 +173,11 @@ export default function FeatureModal({
   let ctaEnabled = true;
   let disabledMessage: string | undefined;
 
-  if (!permissions.check("createFeatureDrafts", project)) {
+  if (
+    !permissionsUtil.canManageFeatureDrafts({
+      project: featureToDuplicate?.project || project,
+    })
+  ) {
```
mknowlton89, line 180: Similar to the bug above - if the user's global role was `noaccess`, but they had multiple project-specific roles, and they tried duplicating a feature in a project they have permission to create a feature in, it was using the current project (none), and thus checking the user's global role

packages/front-end/pages/features/index.tsx
```diff
@@ -343,15 +345,18 @@ export default function FeaturesPage() {
...
-                          className="dropdown-item"
-                          onClick={() => {
-                            setFeatureToDuplicate(feature);
-                            setModalOpen(true);
-                          }}
-                        >
-                          Duplicate
-                        </button>
+                        {permissions.check("manageFeatures", project) &&
+                        canManageFeatureDrafts ? (
```
jdorn, line 349: We should be using the feature's project, not the current project (I think `projectId` is the feature's project).  When duplicating, it copies over the project as well.

packages/front-end/components/Features/FeatureModal/index.tsx
jdorn: I think we want `?? project` instead of `|| project` to match the logic used inside of FeatureModal when duplicating.

packages/front-end/components/Features/FeatureModal/FeatureFromExperimentModal.tsx
jdorn: Same here `?? project` instead of `|| project`

packages/front-end/pages/features/index.tsx
mknowlton89: Ah - yea, good catch.

jdorn (approved):

PR #2280, opened 2024-03-26
Add CI SKIP to github deploy action.

romain-growthbook:
This PR adds support for a case insensitive optional `CI SKIP` or `CI_SKIP` tag in commit titles. When this tag is found, production deploys are skipped.

### DIFF
--- a/.github/scripts/ci-skip.sh
+++ b/.github/scripts/ci-skip.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+shopt -s nocasematch
+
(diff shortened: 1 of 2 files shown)


jdorn:
I think this is already possible without any custom scripts.  See this PR for example: https://github.com/growthbook/growthbook/pull/1792/commits

Docs here: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

romain-growthbook:
Nice I didn't know thanks.

PR #2282, opened 2024-03-26
type issues

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/back-end/src/models/FeatureModel.ts
+++ b/packages/back-end/src/models/FeatureModel.ts
@@ -651,7 +651,6 @@ export async function addFeatureRule(
   };
   changes.rules[env] = changes.rules[env] || [];
   changes.rules[env].push(rule);
-
   await updateRevision(revision, changes, {
     user,
     action: "add rule",
@@ -678,7 +677,6 @@ export async function editFeatureRule(
     ...changes.rules[environment][i],
     ...updates,
   } as FeatureRule;
-
   await updateRevision(revision, changes, {
     user,
     action: "edit rule",
--- a/packages/back-end/src/models/FeatureRevisionModel.ts
+++ b/packages/back-end/src/models/FeatureRevisionModel.ts
@@ -6,7 +6,7 @@ import {
   RevisionLog,
 } from "../../types/feature-revision";
 import { EventAuditUser, EventAuditUserLoggedIn } from "../events/event-types";
-import { ReqContext } from "../../types/organization";
+import { OrganizationInterface, ReqContext } from "../../types/organization";
 
 export type ReviewSubmittedType = "Comment" | "Approved" | "Requested Changes";
 
@@ -24,6 +24,7 @@ const featureRevisionSchema = new mongoose.Schema({
   defaultValue: String,
   rules: {},
   status: String,
+  requiresReview: Boolean,
   log: [
     {
       _id: false,
@@ -186,6 +187,7 @@ export async function createRevision({
   changes,
   publish,
   comment,
+  org,
   requiresReview,
 }: {
   feature: FeatureInterface;
@@ -195,6 +197,7 @@ export async function createRevision({
   changes?: Partial<FeatureRevisionInterface>;
   publish?: boolean;
   comment?: string;
+  org: OrganizationInterface;
   requiresReview?: boolean;
 }) {
   // Get max version number
@@ -234,12 +237,28 @@ export async function createRevision({
       rules,
     }),
   };
+
+  const live = await getRevision(org.id, feature.id, feature.version);
+  if (!live) {
+    throw new Error("Could not lookup feature history");
+  }
+
+  const base =
+    baseVersion === live.version
+      ? live
+      : baseVersion
+      ? await getRevision(org.id, feature.id, baseVersion)
+      : live;
+  if (!base) {
+    throw new Error("Could not lookup feature history");
+  }
   let status = "draft";
   if (publish && !requiresReview) {
     status = "published";
   } else if (publish && requiresReview) {
     status = "pending-review";
   }
+
   const doc = await FeatureRevisionModel.create({
     organization: feature.organization,
     featureId: feature.id,
@@ -254,6 +273,7 @@ export async function createRevision({
     comment: comment || "",
     defaultValue,
     rules,
+    requiresReview,
     log: [log],
   });
 
@@ -265,7 +285,7 @@ export async function updateRevision(
   changes: Partial<
     Pick<
       FeatureRevisionInterface,
-      "comment" | "defaultValue" | "rules" | "baseVersion"
+      "comment" | "defaultValue" | "rules" | "baseVersion" | "requiresReview"
     >
   >,
   log: Omit<RevisionLog, "timestamp">
@@ -283,7 +303,6 @@ export async function updateRevision(
       throw new Error("Can only update draft revisions");
     }
   }
-
   await FeatureRevisionModel.updateOne(
     {
       organization: revision.organization,
--- a/packages/back-end/types/organization.d.ts
+++ b/packages/back-end/types/organization.d.ts
@@ -31,6 +31,13 @@ export type UserPermissions = {
   global: UserPermission;
   projects: { [key: string]: UserPermission };
 };
+export type RequireReview = {
+  requireReview: boolean;
+  resetReviewOnChange: boolean;
+  enviroments: string[];
+  projects: string[];
+  tags: string[];
+};
 
 export type MemberRole =
   | "noaccess"
@@ -180,7 +187,7 @@ export interface OrganizationSettings {
   displayCurrency?: string;
   secureAttributeSalt?: string;
   killswitchConfirmation?: boolean;
-  requireReviews?: boolean;
+  requireReviews?: boolean | RequireReview[];
   defaultDataSource?: string;
   disableMultiMetricQueries?: boolean;
   useStickyBucketing?: boolean;
--- a/packages/front-end/components/Features/FeaturesOverview.tsx
+++ b/packages/front-end/components/Features/FeaturesOverview.tsx
@@ -14,6 +14,7 @@ import {
 import { ago, date, datetime } from "shared/dates";
 import {
   autoMerge,
+  checkIfRevisionNeedsReview,
   evaluatePrerequisiteState,
   getValidation,
   mergeResultHasChanges,
@@ -115,7 +116,6 @@ export default function FeaturesOverview({
   const { fid } = router.query;
 
   const settings = useOrgSettings();
-
   const [edit, setEdit] = useState(false);
   const [editValidator, setEditValidator] = useState(false);
   const [showSchema, setShowSchema] = useState(false);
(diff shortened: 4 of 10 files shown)


packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
```diff
@@ -112,20 +128,69 @@ export default function FeaturesSettings() {
...
-              <Toggle
-                id={"toggle-require-reviews"}
-                value={!!form.watch("requireReviews")}
-                setValue={(value) => {
-                  form.setValue("requireReviews", value);
-                }}
-              />
-            </div>
+            <div className="d-inline-block h4 mt-5 mb-0">Approval Flow</div>
+            {form.watch("requireReviews")?.map?.((requireReviews, i) => (
```
gazzdingo, line 132: add a box around this to make it feel like one setting area

packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
```diff
@@ -112,20 +128,69 @@ export default function FeaturesSettings() {
...
+                    value={!!form.watch(`requireReviews.${i}.requireReview`)}
+                    setValue={(value) => {
+                      form.setValue(`requireReviews.${i}.requireReview`, value);
+                    }}
+                  />
+                </div>
+
+                {!!form.watch(`requireReviews.${i}.requireReview`) && (
+                  <div className="mt-3">
+                    <label htmlFor="enviroments" className="h5">
```
gazzdingo, line 152: wrong spelling

packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
```diff
@@ -112,20 +128,69 @@ export default function FeaturesSettings() {
...
+                      Enviroments
+                    </label>
+                    <MultiSelectField
+                      id="enviroments"
+                      value={form.watch(`requireReviews.${i}.enviroments`)}
```
gazzdingo, line 157: same here

packages/front-end/pages/settings/index.tsx
```diff
@@ -110,7 +110,15 @@ const GeneralSettingsPage = (): React.ReactElement => {
...
-      requireReviews: false,
+      requireReviews: [
+        {
+          requireReview: false,
+          resetReviewOnChange: false,
+          environments: [],
+          projects: [],
+          tags: [],
+        },
+      ],
```
jdorn, line 121: Should this default to an empty array instead?

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -234,28 +236,44 @@ export async function createRevision({
...
+    baseRevision,
+    revision,
+    settings: org.settings,
   });
+  if (publish && !requiresReview) {
+    status = "published";
+    revision.publishedBy = user;
+    revision.datePublished = new Date();
+  } else if (publish && requiresReview) {
+    status = "pending-review";
```
jdorn, line 273: Isn't this just updating a local variable?  Don't you want to update `revision.status` instead? Same above

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -678,7 +687,13 @@ export async function editFeatureRule(
...
     ...updates,
   } as FeatureRule;
-
+  if (
+    (revision.status === "approved" ||
+      revision.status === "changes-requested") &&
+    resetReview
+  ) {
+    changes.status = "pending-review";
+  }
```
jdorn, line 696: There are lots of ways to edit a draft that should reset a review.  I think we're missing a bunch, for example `setDefaultValue`.  I think it would be safer to just add a `resetReview` required argument to `updateRevision`.  That will do 2 things:

1.  Abstract this logic of checking statuses and resetting it so we don't have to repeat it throughout the codebase
2.  Leverage typescript to force us to update every instance

For things like `putRevisionComment` that shouldn't reset anything, we can just hardcode `false` when we update the revision.

packages/front-end/components/Features/FeaturesOverview.tsx
```diff
@@ -215,7 +215,14 @@ export default function FeaturesOverview({
   const { jsonSchema, validationEnabled, schemaDateUpdated } = getValidation(
     feature
   );
-  const requireReviews = !!settings?.requireReviews;
+  const requireReviews = checkIfRevisionNeedsReview({
+    feature,
+    baseRevision: revisions.find(
+      (r) => r.version === revision?.baseVersion
+    ) as FeatureRevisionInterface,
```
jdorn, line 222: Remove the `as` here, it's hiding a legitimate error.  This might be undefined if the base revision can't be found for some reason and we need to make sure we handle that throughout the code.

packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
```diff
@@ -44,6 +46,20 @@ export default function FeaturesSettings() {
...
+        {
+          requireReview,
+          resetReviewOnChange: false,
+          environments: [],
+          projects: [],
+          tags: [],
+        },
+      ]);
+    }
+  }, [form]);
```
jdorn, line 62: This isn't the right place to do a migration.  It should be done in the back-end `migrations.ts`.  There's an `upgradeOrganizationDoc` method where we have migrations for several other setting fields that have changed in the past.

packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
```diff
@@ -112,20 +128,74 @@ export default function FeaturesSettings() {
...
-                  form.setValue("requireReviews", value);
-                }}
-              />
-            </div>
+            <div className="d-inline-block h4 mt-5 mb-2">Approval Flow</div>
+            {form.watch("requireReviews")?.map?.((requireReviews, i) => (
+              <div className="appbox py-2 px-3" key={`approval-flow-${i}`}>
+                <label
+                  className="mr-1 mt-3 d-block"
+                  htmlFor="toggle-require-reviews"
```
jdorn, line 136: This is inside a loop, so the id here needs to include `i` or else it won't be unique.  Same for all the other labels in this section

packages/shared/src/util/features.ts
```diff
@@ -324,6 +328,21 @@ export function mergeResultHasChanges(mergeResult: AutoMergeResult): boolean {
 
   return false;
 }
+export function listChangedEnvironments(
+  base: RulesAndValues,
+  revision: RulesAndValues
+) {
+  const environmentsList: string[] = [];
+  Object.keys(revision.rules)?.forEach((env) => {
```
jdorn, line 336: Environments can be added and removed from an org, but it won't update revisions.  This function should take in a list of valid environments from the org instead and use that to loop through.

packages/shared/src/util/features.ts
```diff
@@ -726,3 +745,97 @@ export function getParsedPrereqCondition(condition: string) {
...
+    return !!requiresReviewSettings;
+  }
+  for (const reviewSetting of requiresReviewSettings) {
+    if (reviewSetting.resetReviewOnChange === true) {
+      return checkEnvironmentsMatch(
+        feature,
+        requestedEnvironments,
+        reviewSetting
+      );
+    }
```
jdorn, line 822: This logic doesn't seem right.  I think we'd want to find the first review setting that matches the feature and return the `resetReviewOnChange` setting from that.

packages/shared/src/util/features.ts
```diff
@@ -726,3 +745,97 @@ export function getParsedPrereqCondition(condition: string) {
...
+  }
+
+  for (const reviewSetting of requiresReviewSettings) {
+    if (reviewSetting.requireReviewOn === true) {
+      return checkEnvironmentsMatch(
+        feature,
+        requestedEnvironments,
+        reviewSetting
+      );
+    }
```
jdorn, line 795: Same here, this logic seems wrong.  First find a matching review setting and then return `requireReview`

packages/shared/src/util/features.ts
```diff
@@ -726,3 +745,97 @@ export function getParsedPrereqCondition(condition: string) {
...
+    if (feature?.tags?.includes(tag)) {
+      return true;
+    }
+  }
+  return (
+    reviewSetting.environments.length === 0 &&
+    reviewSetting.projects.length === 0 &&
+    reviewSetting.tags.length === 0
+  ); //if everything is empty is means it is on for all.
+}
```
jdorn, line 772: Add unit tests for all of these functions here.  The logic is tricky to follow and I'm noticing a lot of potential bugs

packages/shared/src/util/features.ts
gazzdingo: This is checking to see if required review is on first for that settings. then check if it is matching. this way it reduces the amount of times we have to check environments if you turn off requires review

packages/front-end/pages/settings/index.tsx
gazzdingo: we could but that would make it more confusing on the form as this would ensure that we have all the default values set.

packages/back-end/types/organization.d.ts
```diff
@@ -180,7 +187,7 @@ export interface OrganizationSettings {
   displayCurrency?: string;
   secureAttributeSalt?: string;
   killswitchConfirmation?: boolean;
-  requireReviews?: boolean;
+  requireReviews?: boolean | RequireReview[];
```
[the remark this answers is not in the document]
gazzdingo: i would rather migrate than have two settings. it seems like the logic would be super annoying to always have to check the old version and see if you need to move it over to the new one.

packages/shared/src/util/features.ts
gazzdingo: should we update the revisions so that we are not holding around this data or just mark it as deleted

jdorn: No need to update anything.  Doesn't matter if the revision has changes to an environment that doesn't exist anymore.  All we need to change here is to loop over the org's environments instead of assuming `Object.keys(revision.rules)` is up-to-date.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -265,11 +284,14 @@ export async function updateRevision(
   changes: Partial<
     Pick<
       FeatureRevisionInterface,
-      "comment" | "defaultValue" | "rules" | "baseVersion"
+      "comment" | "defaultValue" | "rules" | "baseVersion" | "status"
```
jdorn, line 287: Don't add `status` to changes.  Instead, this function should use the `resetReview` argument that's passed in to determine what status needs to change to (if any).  This will reduce the amount of copy/pasted code and reduce the chance of errors in the future.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -640,34 +640,48 @@ export async function addFeatureRule(
...
-    subject: `to ${env}`,
-    value: JSON.stringify(rule),
-  });
+  if (
+    (revision.status === "approved" ||
+      revision.status === "changes-requested") &&
+    resetReview
+  ) {
+    changes.status = "pending-review";
+  }
```
jdorn, line 662: Made a comment in FeatureRevisionModel too, but we can remove this logic here and have `updateRevision` determine this internally.

packages/shared/src/util/features.ts
```diff
@@ -736,6 +756,119 @@ export function getParsedPrereqCondition(condition: string) {
...
+    return true;
+  }
+  return checkEnvironmentsMatch(feature, requestedEnvironments, reviewSetting);
+}
+
+export function resetReviewOnChange(
+  feature: FeatureInterface,
+  requestedEnvironments: string[],
+  defaultValueChanged: boolean,
+  settings?: OrganizationSettings
```
jdorn, line 819: Change these to named arguments.  This function is called all over the place and I keep having to check back here to see what the arguments mean.

packages/back-end/src/api/features/updateFeature.ts
```diff
@@ -146,6 +146,7 @@ export const updateFeature = createApiRequestHandler(updateFeatureValidator)(
           environments: orgEnvs,
           publish: true,
           changes: revisionChanges,
+          org: req.organization,
```
jdorn, line 149: We will need some way to support a "bypass reviews" option for the REST API.  Otherwise, it's going to be really hard to automate things.  This is not a blocker for now, just something to keep in mind for the future.

packages/back-end/src/models/FeatureModel.ts
gazzdingo: same here just forgot to remove

jdorn (approved):

PR #2286, opened 2024-03-26
Fix CTA on health tab.

romain-growthbook:
Quick PR to fix the CTA on the first health tag modal:

**Before**:

Screenshot 2024-03-26 at 2 15 57 PM

**After:**

Screenshot 2024-03-26 at 2 22 53 PM

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/HealthTabOnboardingModal.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/HealthTabOnboardingModal.tsx
@@ -344,7 +344,7 @@ export const HealthTabOnboardingModal: FC<HealthTabOnboardingModalProps> = ({
         setLastStep(step);
         setStep(-1);
       }}
-      includeCloseCta={false}
+      includeCloseCta={true}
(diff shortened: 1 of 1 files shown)


bryce-fitzsimons (approved):

lukesonnet:
Honestly, how did this happen? Has this been like this for a long time? We tested this flow so many times and people have been successfully using the health tab. Did we change something in `Modal` since then?

lukesonnet (changes requested):
I prefer we land this: https://github.com/growthbook/growthbook/pull/2291 instead to resolve the issue.

Why?

Health tab: the only difference for the health tab onboarding is whether we include both the X in the top right and cancel to `close` the modal. Our initial design with Kari lead us to only have one exit point for the modal, relying on the X in the top right. It also prevents  the following ugly design (IMO) on page 2 that happens with this fix, with 3 CTAs in one modal.

Screenshot 2024-03-26 at 10 33 22 PM

Big picture: the change that the other PR fixes is more in line with how I'd expect the arguments to `Modal` to work anyways. See that PR for more info why.

PR #2287, opened 2024-03-26
Bugfix: Updating url patterns for visual changesets

### LINKED ISSUE #2283: [Bug] Error when changing the URL on visual experiments
Screenshot 2024-03-26 at 10 33 45 AM

bttf:
The code diff looks confusing but basically swapped the ordering of checks in `visualChangesetsHaveChanges`.

From: 
- check for visual changes
- check for urlPatterns changes

To:
- check for urlPatterns changes
- check for visual changes

Also, return early if `visualChanges` for either visualchangeset is undefined (since this doesn't signify a change, just a malformed object).

Fixes #2283

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1961,21 +1961,27 @@ export function visualChangesetsHaveChanges({
   oldVisualChangeset: VisualChangesetInterface;
   newVisualChangeset: VisualChangesetInterface;
 }): boolean {
-  // If there are visual change differences
-  const oldVisualChanges = oldVisualChangeset.visualChanges.map(
-    ({ css, js, domMutations }) => ({ css, js, domMutations })
-  );
-  const newVisualChanges = newVisualChangeset.visualChanges.map(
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -1961,21 +1961,27 @@ export function visualChangesetsHaveChanges({
...
   );
-  const newVisualChanges = newVisualChangeset.visualChanges.map(
+  const newVisualChanges = newVisualChangeset.visualChanges?.map(
     ({ css, js, domMutations }) => ({ css, js, domMutations })
   );
-  if (!isEqual(oldVisualChanges, newVisualChanges)) {
-    return true;
+
+  // If there are no visual changes in either set, there are no changes
+  if (!oldVisualChanges || !newVisualChanges) {
```
jdorn, line 1980: I don't have full context here, but this doesn't sound right. If the old object is undefined and the new one isn't, wouldn't you want to return true here? Otherwise, you'll never get out of that broken state where visual changes is undefined.

PR #2289, opened 2024-03-27
Quantile Testing - Release with FE and API creation

lukesonnet:
## Release Quantile Testing for Pro and Enterprise via REST API and FE

* Adds openapi types for fact metric schema, post and update fact metric
* Update FE modal to work for datasources that can handle quantile testing

TODO:
- [x] Connect results to python once #2223 lands
- [x] Gate FE and REST API to be pro only
- [x] Gate REST API for MySQL datasources (WONTDO)
- [x] Gate BE to be pro only
- [x] Gate REST API to be pro only
- [x] Pro notice in Modal
- [x] Pro gating on snapshot creation (WONTDO)
- [x] Change value column in results as numerator/denominator won't make sense anymore
- [x] Fix return stats to work with `rowResult` logic (e.g. enough data)
- [x] Finalize nstar values
- [x] Test Redshift
- [x] Validate query values

Dependencies: #2223 

# Next Steps

* Documentation & blog post
* Pricing and website changes
* Turn on feature flag to actually release it


packages/back-end/src/api/openapi/payload-schemas/UpdateFactMetricPayload.yaml
```diff
@@ -55,6 +55,26 @@ properties:
...
+        enum:
+          - "event"
+          - "unit"
+        description: Whether the quantile is over unit aggregations or raw event values
+      ignoreZeros:
+        type: boolean
+        description: If true, zero values will be ignored when calculating the quantile
+      quantile:
+        type: number
+        description: The quantile value (from 0.001 to 0.999)
```
jdorn, line 77: You can add some more validation to the schema here if you want.

```yaml
multipleOf: 0.001
minimum: 0.001
maximum: 0.999
```

For reference: https://json-schema.org/understanding-json-schema/reference/numeric

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -883,11 +923,13 @@ export default function FactMetricModal({
...
-                  datasourceType={selectedDataSource.type}
-                  metricType={type}
-                />
+                {type !== "quantile" ? (
+                  <MetricCappingSettingsForm
+                    form={form}
+                    datasourceType={selectedDataSource.type}
+                    metricType={type}
+                  />
+                ) : null}
```
jdorn, line 932: Are we also protecting against this in the back-end or when generating SQL?  If I start with a mean metric, enable capping, and then switch the type to quantile, the capping settings will be saved in Mongo.

packages/front-end/services/metrics.tsx
```diff
@@ -23,6 +23,13 @@ export function getMetricConversionTitle(type: MetricType): string {
   return "Conversion Rate";
 }
 
+export function getPercentileLabel(quantile: number): string {
+  if (quantile === 0.5) {
+    return "Median";
+  }
+  return `P${quantile * 100}`;
```
jdorn, line 30: Fun javascript trivia: Guess what `0.007 * 100` equals? 
 Wrong, it's `0.7000000000000001`.

Or another fun one: `0.974 * 100` equals `97.39999999999999`

packages/front-end/services/metrics.tsx
```diff
@@ -115,7 +122,7 @@ export function formatPercent(
     maximumSignificantDigits: 3,
     ...options,
   });
-  return percentFormatter.format(value);
+  return percentFormatter.format(Math.round(value * 100000) / 100000);
```
jdorn, line 125: Another fun javascript trivia.  What is `Math.round(1.255*100)`?  Hint: It's not 126.

(I think in this particular case you're ok, but only because 100000 as the multiplier doesn't happen to have this bug)

packages/front-end/components/FactTables/FactMetricModal.tsx
lukesonnet: I had a TODO that I will remove and action on!

packages/front-end/services/metrics.tsx
lukesonnet: lmao ok I can truncate to one decimal.

packages/front-end/services/metrics.tsx
lukesonnet: I wanted to use `roundingPriority` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#roundingpriority) but it seems like it wasn't available.

jdorn (approved):

PR #2292, opened 2024-03-27
Move manageProjects to permissionClass

mknowlton89:
### Features and Changes

This PR moves the `manageProjects` permission to the `permissionsClass` and adds 3 finer-grain helper methods - `canCreateProjects`, `canUpdateProject`, and `canDeleteProject`. Currently, these three helpers all rely on the same `manageProjects` permission that is currently only granted to `admin` users. However, as we add finer-grain permissions, it will be easy to update these three helper methods, rather than having to update 10+ files.

In addition to this change, this PR also introduces a new change to the `Layout.tsx` file - previously, we were passing in an array of permissions to the `SidebarLink.tsx` component - now, we pass in `permissionCallbacks` that evaluate whether we should show a nav link or nav sublink based on the user's permissions.

As we migrate permissions to the `permissionClass`, we'll update the `permissions.check()` calls to the newer `canDoXYZ()` helpers.

E.G.
```ts
 {
    name: "Import your data",
    href: "/importing",
    path: /^importing/,
    feature: "import-from-x",
    // old way permissions: ["manageFeatures", "manageEnvironments", "manageProjects"],
    permissionCallbacks: [
        () => permissions.check("manageFeatures", project),
        () => permissions.check("manageEnvironments" as GlobalPermission),
        () => permissionsUtils.canCreateProjects(),
     ],
 },
```

In this change, we also updated the logic within `SidebarLink.tsx` so that we now only check permissions at the sublink level. Since we've removed the logic to check permission at the parent link level, we've added logic so that if a parent link has sublinks, and the user doesn't have permission to view ANY of the sublinks, we don't show the parent link at all.

### Testing

- See screenshots below, and Jest test suite

### NOTE: Screen is zoomed out in order to show all SideBar link options without needing to scroll

Experience for admin user (should be no different from current production experience)
Screen Shot 2024-03-28 at 11 07 41 AM

Experience for experimenter user (should be no different from current production experience)
Screen Shot 2024-03-28 at 11 06 08 AM

FUTURE EXPERIENCE - NOTE: This is not possible today, this is only possible by me manually manipulating a user's permission level, but it shows what will be possible when custom roles are introduces.

Experience for a user who's global role is `noaccess`, but has two project-level permissions, and one of those project-level permissions grant them access to edit the project name/description, and to view logs.
Screen Shot 2024-03-28 at 11 08 58 AM


packages/front-end/components/Layout/SidebarLink.tsx
```diff
@@ -55,15 +54,6 @@ const SidebarLink: FC<SidebarLinkProps> = (props) => {
...
   if (props.superAdmin && !superAdmin) return null;
-  if (props.permissions) {
-    let allowed = false;
-    for (let i = 0; i < props.permissions.length; i++) {
-      if (permissions.check(props.permissions[i] as GlobalPermission)) {
-        allowed = true;
-      }
-    }
-    if (!allowed) return null;
-  }
```
mknowlton89, line 66: Now, if the link has sublinks, and the user doesn't have permission for any sublinks, we don't show the parent link at all.

This removes the need for us to check permission at both the parent and sublink level.

packages/front-end/components/Layout/SidebarLink.tsx
```diff
@@ -75,6 +65,35 @@ const SidebarLink: FC<SidebarLinkProps> = (props) => {
...
+        }
+      } else {
+        permittedSubLinks.push(l);
+      }
+    });
+  }
+
+  if (props.subLinks && !permittedSubLinks.length) {
+    return null;
+  }
```
mknowlton89, line 95: This is the logic referenced above - if the parent link contains sublinks, but the user doesn't have permission for any sublinks, we don't render anything (parent or sublinks)

jdorn (approved):

PR #2299, opened 2024-03-27
Fall back to Gaussian Effect model when log approx is inexact

lukesonnet:
wip


lukesonnet:
Abandoned in favor of fully swapping out the models #2406

PR #2306, opened 2024-03-28
Wire user name in feature slack notifications.

romain-growthbook:
This is in response to a requested user feature: https://growthbookapp.slack.com/archives/C05QYS2UFRD/p1710956702258349

Screenshot 2024-04-01 at 10 07 58 AM

Screenshot 2024-04-01 at 10 08 09 AM


packages/back-end/src/models/EventWebhookModel.ts
```diff
@@ -393,8 +393,8 @@ export const getAllEventWebHooksForEvent = async ({
   });
 
   const docs = allDocs.filter((doc) => {
-    if (!filterOptional(tags, doc.tags)) return false;
-    if (!filterOptional(projects, doc.projects)) return false;
+    if (!filterOptional(doc.tags, tags)) return false;
+    if (!filterOptional(doc.projects, projects)) return false;
```
romain-growthbook, line 397: The logic was inverted here: we want webhooks with no projects to match all events regardless of their projects and same for tags.

packages/back-end/src/events/handlers/slack/slack-event-handler-utils.ts
```diff
@@ -120,9 +120,9 @@ const buildSlackMessageForFeatureCreatedEvent = (
   featureEvent: FeatureCreatedNotificationEvent,
   eventId: string
 ): SlackMessage => {
-  const featureId = featureEvent.data.current.id;
+  const { id: featureId, owner: createdBy } = featureEvent.data.current;
 
-  const text = `The feature ${featureId} has been created`;
+  const text = `The feature ${featureId} has been created by ${createdBy}`;
```
dmkorb, line 125:
```suggestion
  const { user } = featureEvent;
  const text = `The feature ${featureId} has been created by ${user.name} (${user.email})`;
```

or
```
const { user } = featureEvent;
const text = `The feature ${featureId} has been created ${getUserFormatted(
    user
  )}.`;
...
const getUserFormatted = (user: EventAuditUser): string => {
  if (!user) return "unknown user";
  if (user.type === "dashboard") {
    return `by ${user.name} (${user.email})`;
  }
  return "by API key";
};
```

dmkorb (commented):
Thanks for that PR, @romain-growthbook! My team was the one asking for the feature.

Should we also add user name/email? From typings, I see we have both in the `featureEvent` context.

Skimming through the code here I thought we could add a helper function like the one I suggested in the comment. I haven't tested it to check if it's possible though.

packages/back-end/src/events/notification-events.d.ts
```diff
@@ -30,6 +30,7 @@ export type FeatureUpdatedNotificationEvent = NotificationEventPayload<
   {
     current: ApiFeature;
     previous: ApiFeature;
+    updatedBy: string;
```
jdorn, line 33: We don't need this field in the event payload.  Events already have the `user` who triggered the event.

romain-growthbook: Thanks!

romain-growthbook:
This should be ready for merge now!

bttf (approved):

PR #2309, opened 2024-03-28
Super admin - Delete organizations

bttf:
WIP


packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -406,7 +407,87 @@ export function toOrganizationApiInterface(
...
+  }
+
+  const orgDeleted = await mongoose.connection.db
+    .collection("organizations")
+    .deleteOne({ id: orgId });
+
+  if (orgDeleted.deletedCount > 0) {
+    logger.info("Deleted org", orgId, orgDeleted.deletedCount);
+  } else {
+    logger.info("No org deleted?");
```
bttf, line 491: Include org id here

PR #2311, opened 2024-03-29
Migrates createIdeas to permissionClass.

mknowlton89:
### Features and Changes

This PR migrates the `createIdeas` permission to the permissionClass and introduces 4 new helper methods - `canViewIdeaModal`, `canCreateIdea`, `canUpdateIdea`, and `canDeleteIdea`. 

Currently, all of these modals are leveraging the `createIdeas` permission, but as we introduce finer-grain permissions, we can do so in 1 place, rather than all throughout the app.

Additionally, this PR introduces a new pattern `canViewIdeaModal` that takes in an optional project string. This is to be used exclusively on the frontend as a UI check. If we like this pattern, I'll go in an update all of the previous permissions that've been migrated.

### Testing

- See test suite

### DIFF
--- a/packages/shared/src/permissions/permissionsClass.ts
+++ b/packages/shared/src/permissions/permissionsClass.ts
@@ -18,7 +18,8 @@ export class Permissions {
     this.superAdmin = superAdmin;
   }
 
-  public canShowCreateIdeasButton = (project?: string): boolean => {
+  // This is a helper method to use on the frontend to determine whether or not to show certain UI elements
+  public canViewIdeaModal = (project?: string): boolean => {
     return this.checkProjectFilterPermission(
       {
         projects: project ? [project] : [],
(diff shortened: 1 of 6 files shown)


packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -17,6 +18,43 @@ export class Permissions {
...
 
+  // This is a helper method to use on the frontend to determine whether or not to show certain UI elements
+  public canViewIdeaModal = (project?: string): boolean => {
+    return this.checkProjectFilterPermission(
+      {
+        projects: project ? [project] : [],
+      },
+      "createIdeas"
+    );
+  };
```
mknowlton89, line 29: @bryce-fitzsimons New pattern introduced based on your comment Friday. If we move forward with this pattern, I'll open a PR to add this pattern to all of the previously migrated permissions

jdorn (approved):

PR #2317, opened 2024-03-30
Move URL Redirects to use Data Model class

jdorn:
### Features and Changes

Migrate URL Redirects collection to use the new Data Model class.

This also led to several improvements in the base model and context:
- Built-in support for foreign references (experiments and data sources only for now)
- New `WriteOptions` type that can be passed into all create/update/delete methods. Also forwarded to all related hooks
- Validate that ids passed into query functions are actually strings
- Early return in `getById` and `getByIds` when input is empty
- New `bypassPermissionChecks` option for internal `_find` method
- Remove `projectScoping` and the `getAllByProject` method. Simplified a lot of code.
- Unit tests for base model methods

### Write Options

Data models can now specify a custom WriteOptions type.  This is an options object that can be passed to all write methods (`create`, `update`, `updateById`, `delete`, and `deleteById`).  The options are also forwarded to all hooks and validation methods.

This solves 2 use cases for URL Redirects.

1. Bypassing SDK refresh for certain changes.
2. Opt-in to additional validation (circular dependency checks)

### Foreign Keys

Support for foreign references in data models is huge and will simplify a lot of use cases.

The URL Redirect model needs access to the associated experiment in order to do permission checks and validation.  Now, it can do so in an easy and efficient way:

```ts
canRead(doc: UrlRedirectInterface): boolean {
    const { experiment } = this.getForeignRefs(doc);
    return this.context.hasPermission("readData", experiment?.project || "");
}
```

Foreign references are detected automatically as long as you follow common naming conventions.  For example, if there's an `experiment` or `experimentId` key in your data model, it will detect this and assume it's a foreign key to the experiment's collection.  You can always override this behavior in the `getForeignKeys(doc)` method if needed.

The database calls behind the scenes are highly optimized.  For example, if you are doing a `getAll()` operation, it will build a list of all foreign keys in all matched documents and fetch them in a single query.  Plus, there's a caching layer shared between all data model classes to avoid fetching the same thing multiple times within a single request.


packages/back-end/src/models/BaseModel.ts
```diff
@@ -84,94 +88,150 @@ export abstract class BaseModel<T extends BaseSchema> {
...
+    if (typeof id !== "string") {
+      throw new Error("Invalid id");
+    }
+    if (!id) return Promise.resolve(null);
+
     return this._findOne({ id });
   }
   public getByIds(ids: string[]) {
+    // Make sure ids is an array of strings
+    if (!Array.isArray(ids) || !ids.every((id) => typeof id === "string")) {
```
romain-growthbook, line 185: You don't trust TS?

romain-growthbook (approved):
The code looks good and now the hot paths are covered by tests. Next it'd be nice to add tests around foreign keys but if that ever breaks we'd hear about it. I'm glad the ACL stuff is tested, this could silently break any time.

packages/back-end/src/models/BaseModel.ts
jdorn: A lot of times, we just type the incoming `Request` body without doing proper validation.  These sanity checks are meant to protect against security vulnerabilities related to that.  For example, someone passing in an "id" as `{ "$exists": true }` and our code is assuming it's a string and naively querying Mongo with `{ id: idFromReq }`.

PR #2318, opened 2024-04-01
SH Pro has unlimited users

tzjames:
### Features and Changes
Since self-hosted free is unlimited users, it makes sense that self-hosted pro is also unlimited users?

### Testing
Start dev server with `isCloud: false` in front-end/.env.local
Remove any licenseKey in Mongo on the organization.
Click upgrade modal and see the new message.


tzjames:
Actually we want to keep sh pro at 100 users.

PR #2321, opened 2024-04-01
Migrate manageFactTables to permissionClass

mknowlton89:
### Features and Changes

This PR migrates the `manageFactTables` permission to the `permissionClass` and adds some front-end helper methods as well. In the previous PR, we introduce a helper method to know when to show the create button - and now, I've done the same, and also added a helper method of when to show the edit button.

In the case of the edit button, to truly be able to edit, the user must have permissions both in the Fact Table's current projects, as well as any projects the user is trying to update the Fact Tables with.

But, in the case of the edit button, no edits have been made.

That, paired with the fact that the canViewFactTableModal takes in a string for the current project, whereas we need to know all of a Fact Table's existing projects before we can determine if they should see the edit button meant I created two helper methods, but open to feedback.

### Testing

- See test suite


packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -214,7 +218,9 @@ export const postFactFilterTest = async (
     throw new Error("Could not find fact table with that id");
   }
 
-  req.checkPermissions("manageFactTables", factTable.projects);
+  if (!context.permissions.canUpdateFactTable(factTable, {})) {
+    context.permissions.throwPermissionError();
+  }
```
mknowlton89, line 223: Is this necessary here? Is the `runQueries` permission check not sufficient?

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -55,6 +59,46 @@ export class Permissions {
...
+      "manageFactTables"
+    );
+  };
+
+  // This is a helper method to use on the frontend to determine whether or not a user can edit a Fact Table
+  public canViewEditFactTableModal = (
+    factTable: Pick<FactTableInterface, "projects">
+  ): boolean => {
+    return this.checkProjectFilterPermission(factTable, "manageFactTables");
+  };
```
mknowlton89, line 77: Open to feedback on whether or not others think having these two separate helpers is worth it. I ultimately opted for both given the signatures are different. While having two separate helpers adds weight to the class, I think it'll improve the developer experience.

packages/back-end/test/permissions.test.ts
```diff
@@ -2878,6 +2878,633 @@ describe("PermissionsUtilClass.canDeleteMetric check", () => {
   });
 });
 
+describe("PermissionsUtilClass.canViewCreateFactTableModal check", () => {
```
mknowlton89, line 2881: These tests are a bit repetitive - but, I feel like they'll be really valuable as we explore custom roles. Though, we'll likely want to adjust/add more test cases when that happens.

Open to suggestions if anyone has a pattern they've seen before that can either automate some of these tests, or make this a bit more extensible.

jdorn (approved):

PR #2322, opened 2024-04-01
Approval flow ui updates

gazzdingo:
### Features and Changes
Updated the UI to call out the revision state a bit more

Screenshot 2024-04-01 at 1 31 31 PM

Screenshot 2024-04-01 at 1 31 20 PM

Screenshot 2024-04-01 at 1 23 10 PM
Screenshot 2024-04-01 at 1 20 10 PM
Screenshot 2024-04-01 at 1 19 59 PM
Screenshot 2024-04-01 at 1 19 51 PM
Screenshot 2024-04-01 at 1 19 31 PM
Screenshot 2024-04-01 at 1 18 14 PM

### DIFF
--- a/packages/front-end/components/Features/RequestReviewModal.tsx
+++ b/packages/front-end/components/Features/RequestReviewModal.tsx
@@ -8,6 +8,7 @@ import {
 } from "shared/util";
 import { useForm } from "react-hook-form";
 import { EventAuditUserLoggedIn } from "back-end/src/events/event-types";
+import { PiCheckCircleFill, PiCircleDuotone, PiFileX } from "react-icons/pi";
 import { getCurrentUser } from "@/services/UserContext";
 import { useAuth } from "@/services/auth";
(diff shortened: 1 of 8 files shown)


packages/front-end/components/Features/RequestReviewModal.tsx
```diff
@@ -163,6 +164,36 @@ export default function RequestReviewModal({
   } else if (canReview) {
     ctaCopy = "Next";
   }
+  const showRevisionStatus = () => {
```
bryce-fitzsimons, line 167: these should probably be larger / bolder compared to the detailed log view below

bryce-fitzsimons (changes requested):
testing the "request changes" UI... ignore

bryce-fitzsimons (approved):
I think the UI could use some fast follows, and other small things. But we can break it up into another PR later...

PR #2323, opened 2024-04-02
Add a queries page to view all queries run on a Data Source

msamper:
### Features and Changes

Introduces a new Data Source Queries page that lets you view all queries for a Data Source, sort them by different properties, and inspect individual queries. This PR introduces an MVP version of the page but in the future the plan is to:

- Allow linking to the query originator (will require some changes to the query runner)
- Allow re-runs of failed queries
- Add Search/Filter functionality
- Create a similar page for an entire organization

### Screenshots

Screenshot 2024-04-01 at 8 16 06 PM

Screenshot 2024-04-01 at 8 15 39 PM


packages/front-end/pages/datasources/queries/[did].tsx
```diff
@@ -0,0 +1,203 @@
...
+        <p>No queries have been run on this Data Source.</p>
+      </div>
+    );
+  }
+
+  return (
+    <div className="container pagecontents">
+      {modalData && (
+        <Modal
+          open={true}
```
Comment (bttf) [inline packages/front-end/pages/datasources/queries/[did].tsx:79]: Nit, possible to do in shorthand
```suggestion
          open
```

bttf (approved):
Code and screenshots look good

lukesonnet:
Looks great.

I wonder if for experiment queries we should add the Experiment name as some kind of breadcrumb to the query itself, or have a field that is like

`queryParent = "experiment" | "datasource" | "metric"` and then represent the parent by link in a column, something like:

Table title: "Query Origin"
"Experiment: ${experiment.name}"
"Metric Analysis: ${metric.name}"

Just a nice to have.

msamper:
> Looks great.
> 
> I wonder if for experiment queries we should add the Experiment name as some kind of breadcrumb to the query itself, or have a field that is like
> 
> `queryParent = "experiment" | "datasource" | "metric"` and then represent the parent by link in a column, something like:
> 
> Table title: "Query Origin" "Experiment: ${experiment.name}" "Metric Analysis: ${metric.name}"
> 
> Just a nice to have.

Yeah I think that would be great and make this extra useful. We'll have to add an ID for the parent to the query model to make it possible to show this link though, right? Or did you have something else in mind for linking with the existing data that we have

lukesonnet:
> We'll have to add an ID for the parent to the query model to make it possible to show this link though, right?

Ah, yeah, I guess you're right. I remember discussing this, the references only go one way.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -778,6 +778,21 @@ export async function getDataSourceMetrics(
   });
 }
 
+export async function getDataSourceQueries(
```
msamper, line 781: Check for the same permissions that we're checking for in the front-end

PR #2335, opened 2024-04-03
Migrates createDatasources/editDatasourceSettings to permissionClass

mknowlton89:
This PR migrates the `createDatasources` and `editDatasourceSettings` permissions to the `permissionsClass` and introduces 5 new permission class helper methods:
- `canViewCreateDataSourceModal`
- `canCreateDataSource`
- `canUpdateDataSourceParams`
- `canUpdateDataSourceSettings`
- `canDeleteDataSource`

You'll notice that we have two different update methods - with our current codebase, there are two things a user can edit - the datasource's connection params, and the datasource's queries and configuration (things like assignment queries, identifier types, etc). Given that, I've kept that same logic across the two update methods. **Open to feedback if you have better naming suggestions.**

### Testing

- [x] Ensure an admin user can still create, delete, and fully update a data source
- [x] Ensure an experimenter user can update a datasource's identifier types, assignment queries, and view/update the datasource's information schema
- [x] Ensure an engineer is not able to update a datasource's identifier types, assignment queries, or update the datasource's information schema


packages/front-end/components/Experiment/TabbedPage/HealthTab.tsx
```diff
@@ -52,9 +54,11 @@ export default function HealthTab({
...
 
   const hasPermissionToConfigHealthTag =
-    permissions.check("organizationSettings") &&
-    permissions.check("runQueries", datasource?.projects || []) &&
-    permissions.check("editDatasourceSettings", datasource?.projects || []);
+    (datasource &&
+      permissions.check("organizationSettings") &&
+      permissions.check("runQueries", datasource.projects) &&
+      permissionsUtil.canUpdateDataSourceSettings(datasource)) ||
+    false;
```
mknowlton89, line 61: @msamper - I updated this logic so if we `datasource` is undefined, it calculates to `false`.

It looks like we'll never show the Health Tab if the `datasource` is undefined, so this feels like a safe fix and makes the code a bit easier to read.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -18,6 +19,42 @@ export class Permissions {
...
+    datasource: Pick<DataSourceInterface, "projects">
+  ): boolean => {
+    return this.checkProjectFilterPermission(datasource, "createDatasources");
+  };
+
+  public canUpdateDataSourceParams = (
+    datasource: Pick<DataSourceInterface, "projects">
+  ): boolean => {
+    return this.checkProjectFilterPermission(datasource, "createDatasources");
+  };
```
mknowlton89, line 41: This deviates slightly from the other `canUpdateXYZ` methods in that it doesn't take in the `existing` and `updates`. The logic within the `putDatasource` method is pretty deep, so it feels safer to leave it as is and go this route, but open to opinions.

jdorn (approved):

PR #2337, opened 2024-04-03
Use connection key for global webhook key

bryce-fitzsimons:
### Features and Changes

Global webhooks were using a hardcoded key from the environment variable when they should be using a dynamic key from the SDK connection (to match how SDK webhooks work).

Also fixed a spelling error in one of the related types + migration

### DIFF
--- a/packages/back-end/src/models/SdkWebhookLogModel.ts
+++ b/packages/back-end/src/models/SdkWebhookLogModel.ts
@@ -1,6 +1,7 @@
 import { randomUUID } from "crypto";
 import omit from "lodash/omit";
 import mongoose from "mongoose";
+import { migrateSdkWebhookLogModel } from "@back-end/src/util/migrations";
 import { SdkWebHookLogInterface } from "../../types/sdk-webhook-log";
 
 const sdkWebHookLogSchema = new mongoose.Schema({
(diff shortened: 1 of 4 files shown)


gazzdingo (approved):
LGTM

packages/back-end/src/models/SdkWebhookLogModel.ts
```diff
@@ -1,7 +1,7 @@
 import { randomUUID } from "crypto";
 import omit from "lodash/omit";
 import mongoose from "mongoose";
-import { migrateSdkWebhookLogModel } from "@back-end/src/util/migrations";
+import { migrateSdkWebhookLogModel } from "../util/migrations";
```
gazzdingo, line 4: should that be @back-end/src/util/migrations
not ...

packages/back-end/src/util/migrations.ts
```diff
@@ -736,3 +737,13 @@ export function migrateSavedGroup(
 
   return group;
 }
+
+export function migrateSdkWebhookLogModel(
+  doc: SdkWebHookLogDocument
+): SdkWebHookLogDocument {
+  if (doc?.webhookRequestId) {
```
jdorn, line 744: Typo?

bryce-fitzsimons, line 744:
```suggestion
  if (doc?.webhookReduestId) {
```

bryce-fitzsimons: ha, yes! fixed

PR #2353, opened 2024-04-08
Migrates manageFeatures to permissionsClass

mknowlton89:
### Features and Changes

This PR migrates `manageFeatures` to the `PermissionsClass` and in doing so, introduces 4 new helper methods: `canViewFeatureModal`, `canCreateFeature`, `canDeleteFeature`, `canUpdateFeature`. This level of granularity will be helpful as we introduce more fine-grain permissions.

### Dependencies

~~This PR is dependent on [PR-2292](https://github.com/growthbook/growthbook/pull/2292) being merged first, given the changes that PR makes to `Layout.tsx` and `SidebarLink.tsx`.~~

### Testing

- See test suite


packages/back-end/src/controllers/features.ts
```diff
@@ -506,7 +508,9 @@ export async function postFeatureRebase(
   const environments = filterEnvironmentsByFeature(allEnvironments, feature);
   const environmentIds = environments.map((e) => e.id);
 
-  req.checkPermissions("manageFeatures", feature.project);
+  if (!context.permissions.canUpdateFeature(feature, {})) {
```
mknowlton89, line 511: We are using this pattern a lot in this file - where we're checking if they have permission to update the feature, but it's changing an element of the feature that isn't something that if changed, would change their permission level. Open to ideas if this pattern is confusing to others.

packages/back-end/src/api/features/updateFeature.ts
```diff
@@ -34,23 +34,8 @@ export const updateFeature = createApiRequestHandler(updateFeatureValidator)(
...
-        feature.project,
-        getEnabledEnvironments(feature, orgEnvs)
-      );
-      req.checkPermissions(
-        "publishFeatures",
-        project,
-        getEnabledEnvironments(feature, orgEnvs)
-      );
+    if (!req.context.permissions.canUpdateFeature(feature, req.body)) {
+      req.context.permissions.throwPermissionError();
```
jdorn, line 38: This is removing the checks above for `publishFeatures`.  Add those back in

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -211,8 +211,8 @@ const navlinks: SidebarLinkProps[] = [
         name: "Import your data",
         href: "/importing",
         path: /^importing/,
-        filter: ({ permissions, permissionsUtils, gb }) =>
-          permissions.check("manageFeatures", "") &&
+        filter: ({ permissions, permissionsUtils, gb, project }) =>
+          permissionsUtils.canViewFeatureModal(project) &&
```
jdorn, line 215: The importer doesn't use the current project at all.  It creates new projects based on the ones in LaunchDarkly and adds features there.  So we really do need to check for global permissions here.

jdorn (approved):
Added 2 small comments to fix.  Approving now so you don't need to wait for a re-review when you fix them.

PR #2356, opened 2024-04-08
Undo re-flipping risk for inverse metrics

lukesonnet:
### FE bugfix for risk for inverse metrics

Both our stats engine AND our FE after some recent refactors were flipping "risk" for inverse metrics, meaning we were flipping it, and then flipping it back.

This PR undoes this by having the FE receive risk as expected for inverse metrics from the stats engine (which is the same as how it deals with ChanceToWin, which also is returned from the stats engine depending on `inverse`). Users will have their FE reflect the right data as soon as they get this PR in their GrowthBook.

But note that the results data in the DB depend on `metric.inverse` for both chance to win and risk, which means if the metric def changes you have to refresh results to get the UI to show the right data.

### Rip out unused risk code

Also noticed we had some duplicated risk logic that was never used; we only used a duplication of the business logic computing relative risk depending on `inverse` to check if risk existed (in the old `useRiskVariations` method). I simplified this to just check `hasRisk` which was all that method was used for anyways (presumably after ripping out the old results UI).

### Screenshots

Bugged before
Screenshot 2024-04-08 at 10 09 23 AM

Fixed after (no need to re-run results).
Screenshot 2024-04-08 at 10 09 53 AM

### DIFF
--- a/packages/front-end/components/Experiment/BreakDownResults.tsx
+++ b/packages/front-end/components/Experiment/BreakDownResults.tsx
@@ -16,8 +16,8 @@ import {
   applyMetricOverrides,
   setAdjustedPValuesOnResults,
   ExperimentTableRow,
-  useRiskVariation,
   setAdjustedCIs,
+  hasRisk,
 } from "@/services/experiments";
 import ResultsTable from "@/components/Experiment/ResultsTable";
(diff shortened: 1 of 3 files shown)


bryce-fitzsimons (approved):
looks reasonable based on our convo. we'll need to tackle metric state in snapshots at some point

PR #2357, opened 2024-04-08
Track last login date for users

bttf:
Changes
- Adds check to auth middleware for `lastLoginDate` on member row of the organization
- If `lastLoginDate` is undefined or >12 hours old, we update it.
  - 12 hour interval is to minimize DB calls per request. Consequently, the last login date shown in the UI is only accurate to the date (and not the hour, minute).
- If incoming request has no organization or member row associated with it, we skip entirely

Screenshot
Screenshot 2024-04-08 at 4 41 59 PM

### DIFF
--- a/packages/back-end/src/models/OrganizationModel.ts
+++ b/packages/back-end/src/models/OrganizationModel.ts
@@ -412,6 +412,8 @@ export async function updateMember(
   userId: string,
   updates: Partial<Member>
 ) {
+  if (updates.id) throw new Error("Cannot update member id");
+
   const member = org.members.find((m) => m.id === userId);
 
   if (!member) throw new Error("Member not found");


packages/back-end/src/services/auth/index.ts
```diff
@@ -154,6 +155,31 @@ export async function processJWT(
           return;
         }
 
+        const memberRecord = req.organization.members.find(
+          (m) => m.id === req.userId
+        );
+        if (memberRecord) {
```
romain-growthbook, line 161: Could you optimize this by passing a constraint in the `find`? Mongo seems to support searching for date ranges.

packages/front-end/components/Settings/Team/MemberList.tsx
```diff
@@ -149,6 +150,9 @@ const MemberList: FC<{
                     <td>
                       {member.dateCreated && datetime(member.dateCreated)}
                     </td>
+                    <td>
+                      {member.lastLoginDate && date(member.lastLoginDate)}
```
romain-growthbook, line 154: I would consider rounding it to the day so as to avoid confusing the user. You could use `data-fns` to return the corresponding day in the user locale too.

romain-growthbook (approved):
This looks pretty straight forward to me, only minor stuff on optimization question and one user-friendliness.

packages/front-end/components/Settings/Team/MemberList.tsx
romain-growthbook: Ha sorry I read too fast.

packages/back-end/src/services/auth/index.ts
bttf: Ah, this is an array `find` method and not a mongoose method. The `req.org.members` list is typically populated at this point with all org members (see preceding lines).

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -405,3 +407,38 @@ export function toOrganizationApiInterface(
     dateCreated: dateCreated?.toISOString() || "",
   };
 }
+
+export async function updateMember(
+  orgId: string,
+  userId: string,
+  updates: Partial<Member>
+) {
```
jdorn, line 415: Add sanity check to make sure `updates.id` is not being set.

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -405,3 +407,38 @@ export function toOrganizationApiInterface(
     dateCreated: dateCreated?.toISOString() || "",
   };
 }
+
+export async function updateMember(
+  orgId: string,
```
jdorn, line 412: Take the full OrganizationInterface as an argument instead of orgId.  Avoids an unnecessary db lookup and simplifies this function a bit.

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -405,3 +407,38 @@ export function toOrganizationApiInterface(
...
+        }
+        return m;
+      }),
+    });
+  } catch (e) {
+    logger.error("error updating member", {
+      orgId,
+      userId,
+      error: e,
+    });
```
jdorn, line 441: This log seems unnecessary since we're logging again when this is being used.

PR #2359, opened 2024-04-09
Migrates createAnalyses to permissionClass

mknowlton89:
### Features and Changes

This PR migrates the `createAnalyses` permission to the `permissionsClass` and adds 4 new helper methods for experiments: `canViewExperimentModal`, `canCreateExperiment`, `canUpdateExperiment`, and `canDeleteExperiment`.

Additionally, I've added 4 new helpers specific to reports - currently, reports use the same `createAnalyses` permission, now, there are 4 helper methods for reports: `canViewReportModal`, `canCreateReport`, `canUpdateReport`, and `canDeleteReport`.

As you'll notice, the naming convention has changed slightly with this PR - going from `createAnalyses` to `createExperiment`.

When looking at the other permissions, this felt like it fit the pattern better, and adds a bit more context (IMO - open to differing opinions). Likewise, one of the ideas introduces for another future iteration of permissions is having the resource and then the action - e.g. `feature.create`, `experiment.update`, and this fits that future pattern a bit better.

That being said, experiments, like features, have some different elements - a user can update an experiment, or a user can updates various parts of an experiment, ex: updating an experiment's phase has a separate endpoint, so I went back and forth on if we should introduce 2 (or more) variations of the `canUpdateExperiment` helper - as some times, we know that the `updates` doesn't include any updates to the experiment's project, which is the only thing we're really checking in the `canUpdateExperiment`.

Ultimately, though, it felt like adding multiple `update` options could introduce confusion for other engineers.

### Testing

- See test suite


packages/back-end/src/controllers/reports.ts
```diff
@@ -180,6 +183,18 @@ export async function deleteReport(
...
   }
 
+  const connectedExperiment = await getExperimentById(
+    context,
+    report.experimentId || ""
+  );
+
+  if (!connectedExperiment) {
+    throw new Error("Could not find connected experiment");
+  }
```
jdorn, line 193: We don't want to block deleting a report if the experiment doesn't exist.  For example, if they create a report, then delete the experiment, they will be left in an orphaned state where they can't get rid of the report.

packages/back-end/src/controllers/reports.ts
```diff
@@ -253,8 +268,14 @@ export async function putReport(
...
-  // Reports don't have projects, but the experiment does, so check that
-  req.checkPermissions("createAnalyses", experiment?.project || "");
+  if (!experiment) {
+    throw new Error("Could not find connected experiment");
+  }
+
+  // Reports don't have projects, but the experiment does, so check the experiment's project for permission
+  if (!context.permissions.canUpdateReport(experiment)) {
+    context.permissions.throwPermissionError();
+  }
```
jdorn, line 278: Same here, don't block if the experiment doesn't exist.

packages/front-end/pages/report/[rid].tsx
```diff
@@ -63,21 +64,20 @@ export default function ReportPage() {
...
-    "createAnalyses",
-    experimentData?.experiment.project || ""
-  );
+  const canUpdateReport = experimentData
+    ? permissionsUtil.canViewReportModal(experimentData.experiment.project)
+    : false;
+
+  const canDeleteReport = experimentData
+    ? permissionsUtil.canDeleteReport(experimentData.experiment)
+    : false;
```
Comment (jdorn) [inline packages/front-end/pages/report/[rid].tsx:80]: Same as above, don't block this if the experiment doesn't exist.

jdorn (approved):
Left comments on report permissions when experiment doesn't exist

packages/back-end/src/controllers/reports.ts
mknowlton89: Ah, yea - good call.

I think there could be an edge case bug here.

Ex: User's global role is "collaborator" and in Project A they have an "experimenter" role.

If they're trying to delete a report connected to an experiment in Project A - if project A is deleted, they'll get a permission error because we'll be checking the user's global role (which doesn't provide permission to delete reports.

PR #2360, opened 2024-04-09
Fetch from license server in the background

tzjames:
### Features and Changes
The license server can take 2-7s to respond depending upon if it is a cold start, and if it is pro and hitting stripe endpoints, and if stripe is slow or not. Hence to make sure our users are never slowed down by the license server checks (except for the very initial time) we use the mongo cache first as long as the cache is less than 7 days old, and then fetch the license in the background.

This meant that any license errors won't be immediately available on the page load.  To make sure they still are able to see any error message we now save the `lastServerErrorMessage` onto the mongo cache.  We had been wanting to show the error message if the license server was down for more than 2 days, but didn't quite have that logic right since we started using the cache first.  If they fetched the license and then didn't use the license for 6 days then fetched and found the license server down we said they had 0 days to fix.  Now we also save `firstServerErrorDate` and `lastServerErrorDate` whenever we get a license server error as well.  As long as that firstServerErrorDate is within the original 7 days from the updateDate, we let them keep using the cache from 7 days from the firstServerErrorDate.  If they then don't use GB for 3 more days, then load a page again, it will again use the cache to begin with and load the license in the background.  So only then on the following load from /organization will it show that the lastServerErrorDate-firstServerError date is > 2 days and it will show the error message.

- Update the license in the background
- Save error messages into the cache
- Show error messages only if it is down for >= 2 days.
- Let them keep using the cache for 7 days, or for 7 days from the time the license server first went down, to a maximum of 14 days from the update.
- For env var pro license keys that don't have a stripeSubscription.status on them, don't show them the status field.
- Allow Refresh button to work for env var license keys
- Fixed a TS issue in controllers/stripe.ts (not sure why it didn't complain in the past)
- When the central license server throws an error, it ends up as json with "error" property as the message.  When it does a `this.response.status(500).send("active license exists");` it returns a text error message.  We now handle either format.

### Testing
`yarn test`

#### Never had a successful connect due to license server being unreachable
set LICENSE_SERVER_URL=localhost:Anything
Empty growthbook/licenses collection. 
Start server
See "Could not contact license server" in account plan notice
Hover over the notice and see "Please make sure that you have whitelisted 75.2..."

#### Never had a successful connect due to a license server error
set LICENSE_SERVER_URL=http://localhost:8080/api/v1/
start dev central license server
Go to General->Settings->License->Refresh.
See "License server error" as the AccountNotice message.
Hover over the Account Notice: see "License not found".

#### Never had a successful connect due to a license server text error
Temporarily change the central license server code to do
```
 this.response.status(500).send("License not found alt");
        return;
```
instead of throwing the error.
Go to General->Settings->License->Refresh.
See "License server error" as the AccountNotice message.
Hover over the Account Notice: see "License not found alt".
Reset Central license server code.

#### Never had a successful connect for a long time should not affect the error message show.
In MongoDb Compass growthbook.licenses collection set firstFailedFetchDate to a month earlier. (The earlier date should not affect this since it doesn't have a plan - it never had a successful connect).
Go to General->Settings->License->Refresh.
See "License server error" as the AccountNotice message.
Hover over the Account Notice: see "License not found".

#### Setting a bad license key
Go to General->Settings->License and click edit license key pencil icon.
Put in bad key and submit
See "Invalid license key" message

#### Setting a real license key then disconnecting
Put in a valid key in the licenses.newLicenses collection.
Save.
See the license plan in the AccountPlan component in the top nav.
Kill the dev central license server.
Go to General->Settings->License->Refresh.
See no change in the account plan notice (We don't want to show error message within 2 days of the server going down, to give us time to fix any problem with the server before they realize it)

#### Simulating the server being down for 2 days
In MongoDb Compass growthbook.licenses collection set `dateUpdated` to 5 days earlier, set `firstFailedFetchDate` to 3 days earlier.
Go to General->Settings->License->Refresh.
See "Could not connect to license server. Fix within 3 day." in the AccountPlanNotics
Hover over the notice and see the "Please make sure that you have whitelisted 75.2..." message.

#### Simulating the server being down for 8 days and automatically getting downgraded.
In MongoDb Compass growthbook.licenses collection set `dateUpdated` to 8 days earlier, set `firstFailedFetchDate` to 8 days earlier. 
Go to General->Settings->License->Refresh.
See "Could not connect to license server." in the AccountPlanNotice
Hover over the notice and see the "Please make sure that you have whitelisted 75.2..." message.
See no plan tag anymore in the AccountPlan component in the top nav.

### DIFF
--- a/packages/enterprise/src/license.ts
+++ b/packages/enterprise/src/license.ts
@@ -8,7 +8,7 @@ import { pick, sortBy } from "lodash";
 import AsyncLock from "async-lock";
 import { stringToBoolean } from "shared/util";
 import { ProxyAgent } from "proxy-agent";
-import { LicenseModel } from "./models/licenseModel";
+import { LicenseDocument, LicenseModel } from "./models/licenseModel";
 
 export const LICENSE_SERVER_URL =
   process.env.LICENSE_SERVER_URL ||
@@ -102,6 +102,9 @@ export interface LicenseInterface {
   archived: boolean; // True if this license has been deleted/archived
   dateUpdated: string; // Date the license was last updated
   usingMongoCache: boolean; // True if the license data was retrieved from the cache
+  firstFailedFetchDate?: Date; // Date of the first failed fetch
+  lastFailedFetchDate?: Date; // Date of the last failed fetch
+  lastServerErrorMessage?: string; // The last error message from a failed fetch
   signedChecksum: string; // Checksum of the license data signed with the private key
 }
 
@@ -330,6 +333,15 @@ function getVerifiedLicenseData(key: string): Partial<LicenseInterface> {
 }
 
 function verifyLicenseInterface(license: LicenseInterface) {
+  if (
+    license.usingMongoCache &&
+    license.firstFailedFetchDate &&
+    !license.signedChecksum
+  ) {
+    // If there has never been a successful fetch, don't verify the license, it just contains the server error message
+    return;
+  }
+
   const publicKey = getPublicKey();
 
   // In order to verify the license key, we need to strip out the fields that are not part of the signed license data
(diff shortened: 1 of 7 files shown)


packages/enterprise/src/license.ts
```diff
@@ -618,6 +684,8 @@ const lock = new AsyncLock();
 const keyToLicenseData: Record<string, Partial<LicenseInterface>> = {};
 const keyToCacheDate: Record<string, Date> = {};
 
+export let backgroundUpdateLicenseFromServerForTests: Promise<void | LicenseInterface>;
```
mknowlton89, line 687: Does this need to be exported? I don't see it being used elsewhere

mknowlton89 (approved):
Looks good - left one small comment, but approving otherwise

packages/enterprise/src/license.ts
tzjames: Yes, as the name applied it is just used in license.test.ts.

PR #2361, opened 2024-04-09
Romain/experiment notifications

romain-growthbook:
### Features and Changes


packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  if (
+    !context.org.settings?.experimentNotificationsEnabled &&
+    !experiment.metricsNotificationsEnabled
+  )
+    return;
+
+  const [
+    {
+      results: [lastResult],
```
romain-growthbook, line 217: Get first result

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,232 @@
...
+  experiment: ExperimentInterface;
+  lastResult: ExperimentReportResultDimension;
+  snapshot: ExperimentSnapshotDocument;
+}) =>
+  updateWrapper({
+    context,
+    experiment,
+    type: "multiple-exposures",
+    handler: async () => {
+      const totalsUsers = lastResult.variations.reduce(
```
lukesonnet, line 148: Should we consider moving `multiExposureCheck` from the front-end into the shared package so the business logic lives in one place and use that method here?

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,232 @@
...
+}) =>
+  updateWrapper({
+    context,
+    experiment,
+    type: "srm",
+    handler: async () => {
+      const srmThreshold =
+        context.org.settings?.srmThreshold ?? DEFAULT_SRM_THRESHOLD;
+
+      if (srmThreshold <= lastResult.srm) return;
```
lukesonnet, line 188: Similarly I wonder if this should be in a shared package and used in both the FE and BE as well, akin to the `multiExposureCheck` method. This one is just a < comparison so maybe its an even dumber suggestion here, but I love the idea of the business logic being in one place.

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,232 @@
...
+      const totalsUsers = lastResult.variations.reduce(
+        (totalUsersCount, { users }) => totalUsersCount + users,
+        0
+      );
+      const percent = snapshot.multipleExposures / totalsUsers;
+      const multipleExposureMinPercent =
+        context.org.settings?.multipleExposureMinPercent ??
+        MINIMUM_MULTIPLE_EXPOSURES_PERCENT;
+
+      if (snapshot.multipleExposures < multipleExposureMinPercent) return;
```
lukesonnet, line 157: As we discussed in the call, may want to delete the old notification from `pastNotifications` if it exists and there is no longer multiple exposures issues.

lukesonnet (commented):
minor comments on having business logic in `shared` package. Mostly just reviewed the ExperimentNotification.ts file.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: from first analysis, and we also want to only fire these notifications for a certain set of snapshots (has no dimension might be the only check we do for now).

PR #2192, opened 2024-03-05
Impact Analysis

lukesonnet:
## Feature

Adds Experiment Impact module to the Dashboard.

* Has filters to select experiments of interest
* Gets all experiments that used a metric
* Computes adjustment using all experiments that fit the filters
* Returns adjusted scaled impact for all experiments that fit the filters

Computes some quantities of interest and returns them in a tabbed UI.

Under the hood, it tries to:
1. get all snapshots for all experiments in the dashboard using a GET request
2. filter snapshots based on user filters; if there are filtered with missing Scaled Impact analyses, we add a button letting the user update those scaled impacts
3. User can kick off batched updates for those exps. missing scaled impact

Then we use this data to compute our shrinkage, compute annual scaled impact, and build data for the different tabs.

TODO:
- [ ] Docs
- [X] Error catching on adjustment (default to off if < 3 experiments in sample)

## Screenshots

Screenshot 2024-04-03 at 9 38 05 AM
Screenshot 2024-04-03 at 9 38 12 AM
Screenshot 2024-04-03 at 9 38 22 AM
Screenshot 2024-04-03 at 9 38 31 AM
Screenshot 2024-04-03 at 9 38 42 AM
Screenshot 2024-04-03 at 9 38 50 AM
Screenshot 2024-04-03 at 9 39 25 AM
Screenshot 2024-04-03 at 9 39 33 AM

PR #2367, opened 2024-04-09
Allow Stats Integration to take bundles of Experiments

lukesonnet:
### `getSnapshots` endpoint

This PR overhauls `getSnapshots` so that you can get the latest snapshot for a variety of experiments more efficiently relying on a Mongo `aggregate()` command. I also added an endpoint for updating those snapshots with scaled impact if possible (`postSnapshotsWithScaledImpactAnalysis`). This later endpoint relies on the new `createSnapshotAnalyses` method which can update multiple snapshots at a time. 

This works in testing using direct calls to the endpoint, but has not been robustly QAd. This will happen in the follow up #2192  PR.

### Batching snapshot calls

Adds `createSnapshotAnalyses` to run multiple experiment analyses at one time. Not used in prod yet as well, but I did some refactoring to limit some duplication of some squirrely code.

### Stats engine changes

One of the changes from `createSnapshotAnalyses` is allowing the stats engine to take multiple bundles of `DataForStatsEngine`s to be passed, each with an id, to the stats engine, to be returned with results alongside that id. The goal is to eventually allow for easier batch creation of multiple snapshots or snapshot analyses when needed (e.g. for impact analysis).

Previously the stats engine could take a `snapshot` and an array of `analyses` for that snapshot to be computed for each metric.

Now, the stats engine takes an an array of `{snapshots, analyses}` so that you can compute several analyses (each set of analyses potentially different from one another) for any number of snapshots (within or across experiments).

This **is used in prod** as even if an analysis only has one snapshot, the stats integration is fixed and we just wrap that one snapshot before passing it to the stats engine (see 

---

This PR also adds default running scaled and absolute difference types to each stats engine call, since this should add minimal python runtime and will help reduce future burden when loading the dashboard for Impact Analysis.

## Testing

* Tested the `getExperiments` and post with scaled impact analyses that work well on the happy path. More QA and edge cases can be tested when we go for impact analysis.
* Tested in app that single experiment snapshots and snapshot analyses continue to work.
* Manual and reports also still work.


packages/back-end/src/services/stats.ts
```diff
@@ -218,6 +201,64 @@ print(json.dumps({
...
+      "Failed to analyze manual snapshot; Stats Engine returned no results"
+    );
+  }
+  return result.results;
+}
+
+export async function analyzeMultipleExperiments(
+  experimentParams: {
+    experiment: string;
+    params: ExperimentMetricAnalysisParams;
```
bryce-fitzsimons, line 250: perhaps better to put `experimentId` into `ExperimentMetricAnalysisParams`

packages/stats/gbstats/gbstats.py
```diff
@@ -639,3 +640,14 @@ def process_experiment_results(data: Dict[str, Any]) -> list[Dict[str, Any]]:
...
                     )
     return results
+
+
+def process_multiple_experiment_results(
+    data: List[Dict[str, Any]]
+) -> list[Dict[str, Any]]:
+    results: List[Dict] = []
+    for exp_data in data:
+        exp_data_proc = ExperimentDataForStatsEngine(**exp_data)
```
bryce-fitzsimons, line 650: maybe wrap each iteration in a `try/except`?

lukesonnet: Done.

packages/back-end/src/services/stats.ts
lukesonnet: Done, ty.

packages/stats/gbstats/models/results.py
```diff
@@ -67,3 +67,9 @@ class ExperimentMetricAnalysisResult:
 class ExperimentMetricAnalysis:
     metric: str
     analyses: List[ExperimentMetricAnalysisResult]
+
+
+@dataclass
+class MultipleExperimentMetricAnalysis:
```
jdorn, line 73: This class doesn't look like it's being used anywhere

packages/back-end/src/services/experiments.ts
```diff
@@ -669,6 +666,207 @@ export async function createSnapshot({
...
+    string,
+    ExperimentMetricAnalysisParams & {
+      analysis: ExperimentSnapshotAnalysis;
+      unknownVariations: string[];
+      snapshotSettings: ExperimentSnapshotSettings;
+      organization: string;
+      snapshot: string;
+    }
+  >();
+  await params.forEach(
```
jdorn, line 691: Cannot await `forEach`.  The only async part is `getQueryMap`, which looks up queries by id.  Since we have all snapshots in an array, we could build an array of all queries in all snapshots and fetch them all in a single Mongo call.  Then, building the query map for each analysis will not be async

packages/back-end/src/services/experiments.ts
```diff
@@ -669,6 +666,207 @@ export async function createSnapshot({
...
+        analysis: analysis,
+      });
+    }
+  );
+
+  const results = await analyzeMultipleExperiments(
+    Array.from(analysisParamsMap.values())
+  );
+
+  results.map(async (result) => {
```
jdorn, line 771: This doesn't appear to be using `await` anywhere, so remove the `async` here.

packages/back-end/src/services/experiments.ts
```diff
@@ -669,6 +666,207 @@ export async function createSnapshot({
...
+      analysis.results = [];
+      analysis.status = "error";
+      analysis.error = result.error;
+    } else {
+      analysis.results = experimentReportResults[0]?.dimensions || [];
+      analysis.status = "success";
+      analysis.error = undefined;
+    }
+
+    updateSnapshotAnalysis({ organization, id: snapshot, analysis, context });
```
jdorn, line 866: This is running in the background and not being awaited.  If that's intentional, then add a `.catch` here so we won't have errors for unhandled promise rejections.  Otherwise, if you do want to await, then need to queue up the promises in an array and use `promiseAllChunks` to batch them to avoid sending too many updates to Mongo at once.

lukesonnet: Changed to an array of promises and chunked into 10 at a time.

packages/back-end/src/services/experiments.ts
lukesonnet: Done

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2084,6 +2089,82 @@ export async function postSnapshotAnalysis(
...
+    return;
+  }
+  const metricMap = await getMetricMap(context);
+
+  // get latest snapshot for latest phase without dimensions but with results
+  let snapshotsPromises: Promise<ExperimentSnapshotInterface | null>[] = [];
+  snapshotsPromises = ids.map(async (i) => {
+    return await _getSnapshot(context, i, undefined, undefined, true);
+  });
+  const snapshots = await Promise.all(snapshotsPromises);
```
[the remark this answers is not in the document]
lukesonnet: Created a new `_getSnapshots` method using the aggregation approach we suggested. Still testing it out.

packages/stats/gbstats/models/results.py
lukesonnet: Now it is used properly. The stats engine isn't aware it needs to return a json serializable object anymore, and this is happened in the string of text we execute in `stats.ts`.

packages/back-end/src/app.ts
```diff
@@ -429,6 +428,12 @@ app.get(
   experimentsController.getSnapshotWithDimension
 );
 app.post("/experiment/:id/snapshot", experimentsController.postSnapshot);
+
+app.get("/experiments/snapshots/", experimentsController.getSnapshots);
+app.post(
+  "/experiments/snapshots/scaled/",
+  experimentsController.postSnapshotsWithScaledImpactAnalysis
```
jdorn, line 435: Don't add trailing slashes to the routes.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -307,34 +309,57 @@ export async function getExperiment(
...
+): Promise<ExperimentSnapshotInterface[]> {
+  const experimentPhaseMap: Map<string, number> = new Map();
+  const experimentObjs = await getExperimentsByIds(context, experiments);
+  experimentObjs.forEach((e) => {
+    if (e.organization !== context.org.id) {
+      throw new Error("You do not have access to view this experiment");
+    }
+    // get the latest phase:
+    const phase = String(e.phases.length - 1);
+    experimentPhaseMap.set(e.id, parseInt(phase));
```
jdorn, line 354: Why convert length to a string and then parse it?  Can't you use the length directly?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2114,27 +2136,22 @@ export async function postSnapshotsWithScaledImpactAnalysis(
...
   // get latest snapshot for latest phase without dimensions but with results
-  let snapshotsPromises: Promise<ExperimentSnapshotInterface | null>[] = [];
-  snapshotsPromises = ids.map(async (i) => {
-    return await _getSnapshot(context, i, undefined, undefined, true);
-  });
-  const snapshots = await Promise.all(snapshotsPromises);
+  const snapshots = await _getSnapshots(context, experiments);
 
+  // Add snapshots missing scaled analysis to list to fetch
+  const experimentObjs = await getExperimentsByIds(context, experiments);
```
jdorn, line 2142: You are executing this same query inside _getSnapshots already. Instead of passing in experiment IDs, can we just pass in experiment objects instead so only this outer scope is executing that query?

jdorn (approved):

packages/back-end/src/app.ts
lukesonnet: done, ty.

packages/back-end/src/controllers/experiments.ts
lukesonnet: This was how it is done earlier in this page, but yes we should be able to. I'll make the change.

PR #2371, opened 2024-04-10
Add simulation class to gbstats

lukesonnet:
Adds a simulation base class to gbstats to aid in future simulations.

Only change to prod code is a base config input that gets ignored in the base AB test.

### DIFF
--- a/packages/stats/gbstats/devtools/simulation.py
+++ b/packages/stats/gbstats/devtools/simulation.py
@@ -27,6 +27,7 @@ class SimulationStudy(ABC):
         self.test_names = list(test_dict.keys())
         self.tests = list([v[0] for v in test_dict.values()])
         self.configs = list([v[1] for v in test_dict.values()])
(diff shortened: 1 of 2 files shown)


packages/stats/gbstats/devtools/simulation.py
```diff
@@ -1,98 +1,95 @@
...
-            float,
+        return [
             np.mean(
-                (self.lower_limit <= self.theta) * (self.upper_limit >= self.theta)
-            ),
-        )
+                (self.lower_limit[:, j] <= self.theta[:, j])
+                * (self.upper_limit[:, j] >= self.theta[:, j])
+            )
+            for j in range(len(self.tests))
```
lukebrawleysmith, line 71: I would create variable self.n_tests and use that throughout, rather than recalculating using len

lukebrawleysmith (approved):
lgtm

PR #2378, opened 2024-04-11
Implement experiment warning

romain-growthbook:
This PR implements support for experiment warnings and establishes the basis for experiment info.

Experiment warning are events that can be routed through the regular app event utilities (webhook/slack integration).

Events are triggered once per onset. When the warning disappears no event is triggered, except for `auto-update` failures. When the event happens again after having it disappeared, it is triggered again.


packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,208 @@
...
+  snapshot: ExperimentSnapshotDocument;
+}) => {
+  const experiment = await getExperimentById(context, snapshot.experiment);
+  if (!experiment) throw new Error("Error while fetching experiment!");
+
+  const [
+    {
+      results: [firstResult],
+    },
+  ] = snapshot.analyses.sort(
```
lukesonnet, line 195: I think this should work, but I meant literally that we should get the `0`th indexed analysis. e.g. `snapshot.analyses[0]`

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,208 @@
...
+  const totalsUsers = firstResult.variations.reduce(
+    (totalUsersCount, { users }) => totalUsersCount + users,
+    0
+  );
+  const percent = snapshot.multipleExposures / totalsUsers;
+  const multipleExposureMinPercent =
+    context.org.settings?.multipleExposureMinPercent ??
+    MINIMUM_MULTIPLE_EXPOSURES_PERCENT;
+
+  const triggered = multipleExposureMinPercent <= snapshot.multipleExposures;
```
lukesonnet, line 126: nit: See my other PR for a few suggestions here and in notifySRM about re-using shared methods here and in the front-end.

romain-growthbook: We're gonna do this after this PR.

packages/back-end/src/services/experimentNotifications.ts
romain-growthbook: Done and factored out. Thanks for the help!

bttf (approved):
Generally looks good, could use a sign-off from @jdorn though

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,205 @@
...
+  const totalsUsers = results.variations.reduce(
+    (totalUsersCount, { users }) => totalUsersCount + users,
+    0
+  );
+  const percent = snapshot.multipleExposures / totalsUsers;
+  const multipleExposureMinPercent =
+    context.org.settings?.multipleExposureMinPercent ??
+    MINIMUM_MULTIPLE_EXPOSURES_PERCENT;
+
+  const triggered = multipleExposureMinPercent <= snapshot.multipleExposures;
```
lukesonnet, line 129: Ah shoot. I missed this. This needs to be the following, no?

```
  const triggered = multipleExposureMinPercent <= percent;
```

I guess this is one of the advantages of using the existing code first :)

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,205 @@
...
+  results,
+}: {
+  context: Context;
+  experiment: ExperimentInterface;
+  results: ExperimentReportResultDimension;
+}) => {
+  const srmThreshold =
+    context.org.settings?.srmThreshold ?? DEFAULT_SRM_THRESHOLD;
+
+  const triggered = srmThreshold <= results.srm;
```
lukesonnet, line 164: Here we actually want to trigger when `results.srm < srmThreshold`.

lukesonnet (changes requested):
Given we aren't refactoring the business logic for multiple exposures and SRM for this PR, I took a closer look at the logic there and I think both ME and SRM tests need twaeks to their logic.

Also, not blocking, but just calling out that this PR still appears to trigger warnings on any snapshot update regardless of source and that we have discussed this a bit offline.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: It's a little convoluted but you can see this logic in `SRMWarning`

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: You can see this logic (or rather `multipleExposureMinPercent > percent`) in `multiExposureCheck`

romain-growthbook: Fixed!

packages/back-end/src/services/experimentNotifications.ts
romain-growthbook: Fixed!

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -0,0 +1,205 @@
...
+  results,
+}: {
+  context: Context;
+  experiment: ExperimentInterface;
+  results: ExperimentReportResultDimension;
+}) => {
+  const srmThreshold =
+    context.org.settings?.srmThreshold ?? DEFAULT_SRM_THRESHOLD;
+
+  const triggered = srmThreshold < results.srm;
```
lukesonnet, line 164:
```suggestion
  const triggered = results.srm < srmThreshold;
```

You removed the equality but didn't flip the sign :)

lukesonnet (changes requested):
The SRM check isn't fully updated.

lukesonnet (approved):
lgtm!

PR #2387, opened 2024-04-12
Bump OTEL auto-instrumentation library; Fix metrics

bttf:
This PR introduces a metrics exporter to our OTEL configuration. It requires implementing a tracing script that supersedes the default auto-instrumentations/node script. The original script has been copied/pasted into `tracing.js` with the additional configuration required support metrics.

Also bumped the version of our OTEL lib to 0.44.0 in order to support `OTEL_NODE_ENABLED_INSTRUMENTATIONS` env var, which will allow us to disable unnecessary instrumentations. See [this PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1953) for more details.

(There is also an improvement to our 'generate-api-types' CI job so that it reports which files are out of sync, which was needed in this PR for debugging.)

TODO
- [x] Set cloud env var ([terraform PR](https://github.com/growthbook/terraform/pull/23))


bttf:
Tested locally

APM traces

List of metrics

Success metric for `expireOldQueries`

Duration metric for `expireOldQueries`

packages/back-end/src/tracing.ts
```diff
@@ -0,0 +1,56 @@
...
+);
+
+const metricReader = new PeriodicExportingMetricReader({
+  exporter: new OTLPMetricExporter(),
+  exportIntervalMillis: 1000,
+});
+
+const sdk = new opentelemetry.NodeSDK({
+  instrumentations: getNodeAutoInstrumentations(),
+  resourceDetectors: getResourceDetectors(),
```
jdorn, line 37: Looks like this is using `getResourceDetectorsFromEnv` in the built-in register script.  Was changing this intentional?

bttf: [They're the same function](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/585b833dc8e911c17b483e871f752ad4988cee5a/metapackages/auto-instrumentations-node/src/index.ts#L19)

jdorn (approved):

PR #2399, opened 2024-04-17
Set cloud license keys via admin

tzjames:
### Features and Changes

On cloud, only super admins should be able to set licenseKeys when not going through regular upgrade process.

- Adds LicenseKey field to EditOrganization admin component
- Handles LicenseKey in Put organization endpoint
- Correctly throws an error when a license is set via general->settings in self-host, and in EditOrganization in cloud
- Adds test to make sure that when forceRefresh is true that errors are thrown.
- Moves check for showing externalId field to the EditOrganization component itself to be consistent with checking whether to show the licenseKey field

#### Testing Cloud
`yarn test`
in back-end and front-end/.env.local
```
IS_CLOUD=true
IS_MULTI_ORG=true
SSO_CONFIG=valid SSO config
```
restart the server
Go to settings->general - see no input to edit the license.
Go to settings->admin - click org and the pencil icon
Edit the license with a bad license
See invalid licenseKey message
Enter in a valid licenseKey
See it update successfully.
Refresh the page, and see the new license show up.

#### Testing Self-Hosted Multi-Org (license key should only set by env var)
in back-end and front-end/.env.local
```
IS_CLOUD=false
IS_MULTI_ORG=true
```
restart server
Go to Settings->admin page
Click pencil icon on an organization
See the externalId, but not the licenseKey field.
Go to Settings->General
See no license section at all.

#### Testing normal Self-Hosted
in back-end and front-end/.env.local
```
IS_CLOUD=false
IS_MULTI_ORG=false
```
restart server
Go to Settings->General
Click pencil icon by license.
Enter "license_bad"
See "Invalid license key" message upon submit
Put in a valid license key.
See it succeed.


packages/front-end/components/Admin/EditOrganization.tsx
```diff
@@ -57,7 +61,19 @@ const EditOrganization: FC<{
...
         />
-        {showExternalId && (
+        {isCloud() && (
+          <div className="mt-3">
+            License Key
+            <input
+              type="text"
+              className="form-control"
+              value={licenseKey}
+              minLength={3}
```
jdorn, line 71: Does this `minLength` make it so you can't edit an organization's name if they don't have a license key?  Maybe we should move this logic to the submit handler.  If licenseKey is not empty, then validate the length/format.

packages/enterprise/src/license.ts
```diff
@@ -664,6 +664,10 @@ async function updateLicenseFromServer(
       logger.error(`Error creating mongo cache: ${e}`);
       throw e;
     });
+    verifyLicenseInterface(license);
+    keyToLicenseData[licenseKey] = license;
+    keyToCacheDate[licenseKey] = new Date();
+    throw e;
```
jdorn, line 670: Is this change intentional?  Seems unrelated to setting cloud licenses via admin

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -1273,6 +1273,12 @@ export async function putOrganization(
       }
     }
 
+    if (licenseKey && licenseKey.trim() !== org.licenseKey) {
+      updates.licenseKey = licenseKey.trim();
+      orig.licenseKey = org.licenseKey;
+      await setLicenseKey(org, updates.licenseKey);
```
jdorn, line 1279: This is calling `updateOrganization` immediately before another `updateOrganization` call.   Doesn't quite seem like the right level of abstraction.

tzjames: setLicenseKey does not call updateOrganization, it just sets org.licenseKey to be the new key and verifies it is a valid key.

packages/enterprise/src/license.ts
tzjames: Well it is related in the sense that it was a bug I found when the test plan didn't work as expected.  It should have been its own PR though.  But now this code has a merge conflict with https://github.com/growthbook/growthbook/pull/2413.  So I think I would like to land that one first.  Then I'll make a new PR that fixes this bug, then I'll update this PR removing this code.

packages/front-end/components/Admin/EditOrganization.tsx
tzjames: No, since it is not required it doesn't complain about being blank.  I left it in thinking it can't hurt, but it also not needed at all.  We already have validation in the submit handler to see if the license is valid license, and an "Invalid license key" is a more accurate error message than "This field needs to be at least three characters you have entered 1", so I'll just remove it.

tzjames:
Split this into:
https://github.com/growthbook/growthbook/pull/2451
https://github.com/growthbook/growthbook/pull/2452

PR #2400, opened 2024-04-17
Fix feature draft base revision bug

jdorn:
### Features and Changes

If you create and discard a draft, new draft revisions you create will be based on top of that discarded draft instead of the published feature version.  This can cause some weird states, like a revision having "no changes" even though it's clearly different from the published version.


gazzdingo (approved):
tested local and it was all working

PR #2404, opened 2024-04-18
Add and AgendaJob that will refresh the license once a day

tzjames:
### Features and Changes

For self-hosted installations this job will update their license data once a day.
This will allow us to make sure that any installation that hasn't contacted us within a day may have had their connection to us blocked, and their license would then be void within a week if something is not done to unblock the connection.

- Creates the new updateLicense agenda job.
- Removes the manual licenseInit when starting the server as this is done already when the server starts up via the job
- Removes special code  in initializeLicenseForOrg that handled the special case of starting the server and calling it with no args as that was done only on the server startup call.

#### Testing a new installation
rename organizations collection in mongo.
Start server
see no error

#### Testing an org without a license key
Create an organization
Restart server
See no error

#### Testing a new style env var LICENSE_KEY
Set a LICENSE_KEY in back-end/.env.var to a valid `license_` new style license.
Restart the server
See in the logs that it loads the license
Got to localhost:3000 and see that the license info is displayed.

#### Testing a old style env var LICENSE_KEY
Set LICENSE_KEY to be an old style license key
Restart the server
See in the logs that it loads the license
Got to localhost:3000 and see that the license info is displayed.

#### Testing a license key on the org
Set licenseKey property in mongo to a new style license
Restart the server
See in the logs that it loads the license
Got to localhost:3000 and see that the license info is displayed.


packages/back-end/src/init/index.ts
```diff
@@ -9,7 +8,6 @@ export async function init() {
     initPromise = (async () => {
       await mongoInit();
       await queueInit();
-      await initializeLicenseForOrg();
```
jdorn, line 12: This is required for the server to function properly, right?  In a horizontally scaled prod environment, are we 100% sure the agenda job will always run immediately at startup?

I don't want this getting blocked by some other jobs for 10 minutes and the instance basically being downgraded to free during that time.

packages/back-end/src/services/licenseData.ts
```diff
@@ -64,24 +61,9 @@ export async function getLicenseMetaData() {
...
-    if (IS_CLOUD) {
-      // On cloud we will initialize license on an as need basis once we know the org of a request
-      return;
-    }
-
-    // For self-hosted there is either
-    // no orgs upon first initialization - in which case we get the key from the env var
-    // exactly one org normally - in which case we get the key from the org
-    // or multiple orgs if it is a MULTI_ORG site - in which case the org is not allowed to have a license key and we shall also fall back to the env var
-    org = (await getOrganization()) || undefined;
```
jdorn, line 82: I didn't know this `getOrganization()` method existed.  We should delete this ASAP as it is extremely dangerous to have in the codebase.  It doesn't check for self-hosted, so if this were to be called on cloud, it could potentially expose a random organization to other users.

packages/back-end/src/init/index.ts
tzjames: The only time this does anything is if the license is stored in env var and it is their very first time loading the server so that there is nothing in the mongo cache.  However there will be no other jobs the first time the server starts up so the job will run right away.  Even if it doesn't they won't be downgraded to free, they would just wait for a single license server call to return.

packages/back-end/src/services/licenseData.ts
tzjames: Good point

packages/back-end/src/jobs/updateLicense.ts
```diff
@@ -0,0 +1,37 @@
...
+const UPDATE_LICENSES_JOB_NAME = "updateLicenses";
+
+const updateLicense = trackJob(UPDATE_LICENSES_JOB_NAME, async () => {
+  if (IS_CLOUD) {
+    return;
+  }
+
+  const org = await getOrganization();
+  if (org) {
+    initializeLicenseForOrg(org, true);
```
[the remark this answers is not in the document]
tzjames: Another good point.

jdorn (approved):

PR #2405, opened 2024-04-18
Add extra settings to fact metric REST API

lukesonnet:
Adds control over risk thresholds, percent change thresholds, and min sample size to fact metric CREATE and UPDATE endpoints.

Does some validation to ensure values are correct.

I also reset the defaults for these metrics to be more in line with what happens if you create fact metrics in the app.

Tested with the following in Chrome DevTools console (after latest changes some of this error logic changes but I re-tested more or less the same flow)

```ts
// Fill these with actual values for your GrowthBook instance
const secret = `sec...`;
const datasource = `ds...`;

async function req(method, endpoint, data) {
  const res = await fetch(`http://localhost:3100/api/v1${endpoint}`, {
    method,
    headers: {
      "Content-Type": "application/json",
      Authorization: `Bearer ${secret}`,
    },
    ...(data ? { body: JSON.stringify(data) } : {}),
  });
  const json = await res.json();
  console.log(json);
  return json;
}

// create fact metric with some setting
const factMetric = (
  await req("POST", `/fact-metrics`, {
    name: "TEST",
    metricType: "mean",
    numerator: {
      factTableId: "FILL IN YOUR ID",
      column: "amount",
    },
    minPercentChange: 0.05,
  })
).factMetric;
if (!factMetric) {
  throw new Error("POST /fact-metrics error");
}
// errors as max is missing {message: 'Must specify both `maxPercentChange` and `minPercentChange` or neither'}

// also errors
const factMetric = (
  await req("POST", `/fact-metrics`, {
    name: "TEST",
    metricType: "mean",
    numerator: {
      factTableId: "ftb_2s1by1xdqlubxxpm2",
      column: "amount",
    },
    minPercentChange: 0.50,
    maxPercentChange: 0.01
  })
).factMetric;
if (!factMetric) {
  throw new Error("POST /fact-metrics error");
}
// errors as max < min {message: '`maxPercentChange` must be greater than `minPercentChange`'}

// a working example
const factMetric = (
  await req("POST", `/fact-metrics`, {
    name: "TEST",
    metricType: "mean",
    numerator: {
      factTableId: "ftb_2s1by1xdqlubxxpm2",
      column: "amount",
    },
    minPercentChange: 0.001,
    maxPercentChange: 0.5
  })
).factMetric;
if (!factMetric) {
  throw new Error("POST /fact-metrics error");
}

// Update a fact metric
const updatedFactMetric = (
  await req("POST", `/fact-metrics/${factMetric.id}`, {
    name: "Revenue (updated)",
    maxPercentChange: 0.1
  })
).factMetric;
if (!updatedFactMetric || updatedFactMetric.name !== "Revenue (updated)") {
  throw new Error("POST /fact-metrics/:id error");
}

// Update a fact metric
const updatedFactMetric = (
  await req("POST", `/fact-metrics/${factMetric.id}`, {
    name: "Revenue (updated)",
    riskThresholdSuccess: 0.0001,
    riskThresholdDanger: 0.05
  })
).factMetric;
if (!updatedFactMetric || updatedFactMetric.name !== "Revenue (updated)") {
  throw new Error("POST /fact-metrics/:id error");
}
```


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -175,6 +175,8 @@ export class FactMetricModel extends BaseClass {
 
     return {
       ...otherFields,
+      riskThresholdDanger: otherFields.loseRisk,
+      riskThresholdSuccess: otherFields.winRisk,
```
jdorn, line 179: This will result in both `loseRisk` and `riskThresholdDanger` being added to the object.

Instead, add `loseRisk` and `winRisk` to the destructure call above like we do for `metricType` for example.

packages/back-end/src/api/fact-metrics/postFactMetric.ts
```diff
@@ -36,9 +38,46 @@ export async function getCreateMetricPropsFromBody(
...
+      "Must specify both `maxPercentChange` and `minPercentChange` or neither"
+    );
+
+  if (maxPercentExists && minPercentExists) {
+    // Enforce max is greater than min
+    if (maxPercentChange <= minPercentChange)
+      throw new Error(
+        "`maxPercentChange` must be greater than `minPercentChange`"
+      );
+  }
```
jdorn, line 79: This validation logic is duplicated twice (creating and updating) and is also not present for the internal API endpoints.  Seems like it would be better to put this in the FactMetricModel's `customValidation` method instead so it lives one place and applies everywhere.

packages/back-end/src/models/FactMetricModel.ts
lukesonnet: Ah shoot. ty

packages/back-end/src/api/fact-metrics/postFactMetric.ts
lukesonnet: Will do. I was looking at custom validation and thought it wouldn't be the right place initially but I see the overall flow now.

jdorn (approved):

PR #2407, opened 2024-04-18
[stats] Add ability to specify effect priors

lukesonnet:
## New feature

Adds ability to add priors to bayesian analysis at:
* Organization level
* With metric-level overrides
* Then with experiment-metric level overrides

Reports do not have overrides, but you can "freeze" the settings by default in a report, or let them inherit the latest settings on re-run.

## Main back-end changes

The logic largely mimics how regression adjustment settings currently work, and indeed I've largely rewritten the way we re-evaluate regression adjustment settings to do both RA and priors, so their logic is largely coupled. For example I've replaced:

```
export type MetricRegressionAdjustmentStatus = {
  metric: string;
  regressionAdjustmentEnabled: boolean;
  regressionAdjustmentAvailable: boolean;
  regressionAdjustmentDays: number;
  reason: string;
};
```

with

```
export interface MetricSnapshotSettings {
  metric: string;
  properPrior: boolean;
  properPriorMean: number;
  properPriorStdDev: number;
  regressionAdjustmentReason: string;
  regressionAdjustmentEnabled: boolean;
  regressionAdjustmentAvailable: boolean;
  regressionAdjustmentDays: number;
}
```

All of the methods that generate and store the former now produce the latter instead, where the `proper*` fields are the resolved settings based on which setting has priority (experiment-metric override -> metric override -> organization default -> GrowthBook default). These settings are then stored in the snapshot `settings.metricSettings` object.

## Data model changes and migrations

This requires changes to many data models, but mostly snapshots (storing different/new settings), reports (storing new settings), metrics (storing the metric overrides). New migrations have been added with unit tests for these three to ensure that they accept reasonable defaults and are not missing any critical fields.

Also added REST API updates to allow priors to be set via the API.

## Stats engine

Stats engine changes are minimal, but it just accepts the metrics' settings via `MetricSettingsForStatsEngine`.

## Front-end changes

* additional settings components for organization, metric overrides, and experiment-metric overrides
* adds tooltip data explaining the priors to the metric name and results table tooltips

- [X] test rest API endpoints
- [ ] Add ability to config via config.yml

## Testing

Added migration unit tests

Lots of QA, still ongoing.

### Organization settings
Screenshot 2024-05-07 at 9 16 51 AM
Screenshot 2024-05-07 at 9 16 40 AM

### Metric settings and [mid]/[fmid]

Screenshot 2024-05-07 at 9 21 24 AM
Screenshot 2024-05-07 at 9 21 28 AM

Screenshot 2024-05-07 at 9 21 14 AM
Screenshot 2024-05-07 at 9 21 02 AM

### Experiment-metric overrides

Screenshot 2024-05-07 at 10 35 29 AM

Screenshot 2024-05-07 at 10 33 17 AM

### Results tooltips

Screenshot 2024-05-07 at 10 33 09 AM
Screenshot 2024-05-07 at 10 33 07 AM

Following only shows if prior is applied and is proper (e.g. affects results)
Screenshot 2024-05-07 at 10 33 01 AM
Screenshot 2024-05-07 at 10 33 04 AM


packages/front-end/components/Experiment/MetricsOverridesSelector.tsx
```diff
@@ -448,6 +463,128 @@ export default function MetricsOverridesSelector({
...
+                      <div className="form-group mt-1 mb-2 mr-2 form-inline">
+                        <label
+                          className="mr-1 small"
+                          htmlFor={`toggle-properPrior_${i}`}
+                        >
+                          Use proper prior for this metric
+                        </label>
+                        <Toggle
+                          id={`toggle-properPrior_${i}`}
+                          value={
```
lukesonnet, line 511: TODO ensure we the value in the form matches what we show in the UI when turning on the override.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -504,7 +504,7 @@ export class ExperimentResultsQueryRunner extends QueryRunner<
       )
     )
       .map((m) => metricMap.get(m))
-      .filter((m) => m) as ExperimentMetricInterface[];
+      .filter((m) => m) as [];
```
jdorn, line 507: Is this change intentional?

packages/back-end/src/services/experiments.ts
```diff
@@ -1930,19 +1951,15 @@ export function updateExperimentApiPayloadToInterface(
...
 ): Promise<{
   regressionAdjustmentEnabled: boolean;
-  metricRegressionAdjustmentStatuses: MetricRegressionAdjustmentStatus[];
+  settingsForSnapshotMetrics: MetricSnapshotSettings[];
 }> {
   let regressionAdjustmentEnabled = false;
-  const metricRegressionAdjustmentStatuses: MetricRegressionAdjustmentStatus[] = [];
-
-  if (!experiment.regressionAdjustmentEnabled) {
-    return { regressionAdjustmentEnabled, metricRegressionAdjustmentStatuses };
```
jdorn, line 1944: This change means when RA is disabled for an experiment, it will now return a non empty array of metric regression statuses. Probably fine, but just wanted to call out in case it wasn't intentional

packages/back-end/src/util/migrations.ts
```diff
@@ -691,6 +745,33 @@ export function migrateSnapshot(
...
+            ...m.computedSettings,
+            properPrior:
+              m.computedSettings.properPrior ??
+              defaultMetricPriorSettings.proper,
+            properPriorMean:
+              m.computedSettings.properPriorMean ??
+              defaultMetricPriorSettings.mean,
+            properPriorStdDev:
+              m.computedSettings.properPriorStdDev ??
+              defaultMetricPriorSettings.stddev,
```
jdorn, line 769: Minor code style. You can control default and override behavior based on where you put the spread operator.

```ts
{
  a: "default",
  ...obj,
  b: "override"
}
```

obj.a will take precedence over the default, but obj.b will be ignored in favor of the override.

jdorn (commented):
I looked through all of the back end and most of the front end. Did not look at stats changes or test the UI/functionality in a browser.

packages/back-end/src/services/experiments.ts
lukesonnet: Yeah I didn't see any issues in testing as this flag gets used again later when building the actual array of settings. I'll check more carefully though.

packages/front-end/components/Experiment/ResultsTableTooltip.tsx
```diff
@@ -479,6 +484,39 @@ export default function ResultsTableTooltip({
...
+                          data.metricSnapshotSettings?.properPriorMean ?? 0
+                        } and standard deviation ${
+                          data.metricSnapshotSettings?.properPriorStdDev ??
+                          DEFAULT_PROPER_PRIOR_STDDEV
+                        }.`}
+                      </div>
+                      <div>
+                        {
+                          "This affects all of the results for this metric (the lift, the chance to win, credible intervals, and more). It also may explain why the experiment lift is not exactly the percent difference between the variation and the baseline."
+                        }
```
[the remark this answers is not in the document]
lukesonnet: It always appears when there are proper bayesian priors.

packages/front-end/components/Experiment/ResultsTableTooltip.tsx
```diff
@@ -479,6 +484,39 @@ export default function ResultsTableTooltip({
...
+                  "results-prior text-muted rounded d-flex justify-content-center mt-2",
+                  data.rowResults.resultsStatus
+                )}
+              >
+                <Tooltip
+                  className="cursor-pointer"
+                  body={
+                    <>
+                      <div className="mb-1">
+                        {`This metric was analyzed with a proper Bayesian prior of mean ${
```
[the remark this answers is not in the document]
lukesonnet: Done, thanks.

packages/front-end/components/Experiment/ResultsTableTooltip.tsx
```diff
@@ -479,6 +484,39 @@ export default function ResultsTableTooltip({
...
+                          data.metricSnapshotSettings?.properPriorMean ?? 0
+                        } and standard deviation ${
+                          data.metricSnapshotSettings?.properPriorStdDev ??
+                          DEFAULT_PROPER_PRIOR_STDDEV
+                        }.`}
+                      </div>
+                      <div>
+                        {
```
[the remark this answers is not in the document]
lukesonnet: No but we should add it. I'll add that as well.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
lukesonnet: no, undid. thank you.

lukebrawleysmith (commented):
this looks good to me (looked mainly at science aspects)

PR #2413, opened 2024-04-19
Replace license on refetching

tzjames:
### Features and Changes
When a new license was fetched from the license server we were updating the cache, but we really meant to be replacing what was there.  When the license server goes down we set properties on the license in the cache detailing the error message, and if we just update the cache those properties will remain.

We also had been saving what the license server returned as long as it 200ed in the mongo cache and erroring if the signature didn't match.  Now we verify after all successful calls to get a new license and only save when it is verified successfully.  

### Testing
`yarn test`

Set `LICENSE_SERVER_URL=anything` in back-end/.env.local
Restart the dev server.
Hit a page.
See in mongo growthbook/license collection that lastServerErrorMessage has been set.
Remove LICENSE_SERVER_URL line in back-end/.env.local
Restart the dev server
Hit a page.
See in mongo growthbook/license collection that lastServerErrorMessage is no longer there.

### DIFF
--- a/packages/enterprise/src/license.ts
+++ b/packages/enterprise/src/license.ts
@@ -580,16 +580,20 @@ export async function postResendEmailVerificationEmailToLicenseServer(
   );
 }
 
-// Creates or updates the license in the MongoDB cache in case the license server goes down.
-async function createOrReplaceLicenseMongoCache(license: LicenseInterface) {
-  await LicenseModel.findOneAndReplace(
-    { id: license.id },
-    { $set: license },
-    { upsert: true }
-  );
+// Creates or replaces the license in the MongoDB cache in case the license server goes down.
+async function createOrReplaceLicenseMongoCache(
+  license: LicenseInterface | LicenseDocument
+) {
+  const licenseObject =
+    license.constructor.name === "model"
+      ? (license as LicenseDocument).toObject()
+      : license;
+  await LicenseModel.findOneAndReplace({ id: license.id }, licenseObject, {
+    upsert: true,
+  });
 }
 
-// Updates the local daily cache, the one week backup Mongo cache, and verifies the license.
+// Updates the in memory cache, the one week backup Mongo cache, and verifies the license.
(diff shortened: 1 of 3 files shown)


packages/enterprise/src/license.ts
```diff
@@ -580,21 +580,25 @@ export async function postResendEmailVerificationEmailToLicenseServer(
...
-// Creates or updates the license in the MongoDB cache in case the license server goes down.
-async function createOrUpdateLicenseMongoCache(license: LicenseInterface) {
-  await LicenseModel.findOneAndUpdate(
-    { id: license.id },
-    { $set: license },
-    { upsert: true }
-  );
+// Creates or replaces the license in the MongoDB cache in case the license server goes down.
+async function createOrReplaceLicenseMongoCache(
+  license: LicenseInterface | LicenseDocument
```
jdorn, line 585: A `LicenseDocument` object should never make it this far.  I find it best to never use Mongoose objects.  If you get one back from a `find` operation, just immediately convert it to raw JSON.  We have helper methods like this in the app for all of our data models:

```ts
function toInterface(doc: FactTableDocument): FactTableInterface {
  const ret = doc.toJSON<FactTableDocument>();
  return omit(ret, ["__v", "_id"]);
}
```

packages/enterprise/src/license.ts
```diff
@@ -739,6 +737,9 @@ export async function licenseInit(
             // Use the cache
             license = mongoCache.toJSON();
             license.usingMongoCache = true;
+            verifyLicenseInterface(license);
+            keyToLicenseData[key] = license;
+            keyToCacheDate[key] = new Date();
```
jdorn, line 742: This logic is now only running within the else block.  Before it was running after both if and else.  Is that intentional?

tzjames: It is done within `updateLicenseFromServer` when it calls `verifyAndSetServerLicenseData` so hence it is done in both branches.

jdorn (approved):

PR #2416, opened 2024-04-22
Show error if first license load fails

tzjames:
### Features and Changes

If there is a brand new installation and the server is not reachable, or throws an error such as the license not being found we want that error to display.  

### Testing

`yarn test`
Empty growthbook.licenses in mongo.
Remove any license_key on organization in mongo.
set `LICENSE_SERVER_URL=anything` in back-end/.env.local and run `yarn dev`
hit localhost:3000 and see the error message that the license server is not reachable

kill the back-end server
reset `LICENSE_SERVER_URL`
set `LICENSE_KEY=license_bad_key` in back-end/.env.local and run `yarn dev`
hit localhost:3000 and see the error message that there was a license server error and the license could not be found.

kill the back-end server
reset the `LICENSE_KEY` to a valid key  in back-end/.env.local and run `yarn dev`
hit localhost:3000 and see no error message and the license plan in the AccountPlan component in the topnav.

### DIFF
--- a/packages/enterprise/src/license.ts
+++ b/packages/enterprise/src/license.ts
@@ -727,7 +727,7 @@ export async function licenseInit(
           if (
             forceRefresh ||
             !mongoCache ||
-            !mongoCache.dateUpdated ||
+            !mongoCache.dateUpdated || // If the first call to get the license errors, the cache will be created with no dateUpdated
             new Date(mongoCache.dateUpdated) < oneWeekAgo
           ) {
             license = await updateLicenseFromServer(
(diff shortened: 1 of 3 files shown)


packages/enterprise/src/license.ts
```diff
@@ -818,9 +819,12 @@ export function getLicenseError(org: MinimalOrganization): string {
   const licenseData = getLicense(key);
 
   // If there is no license it can't have an error
-  // Licenses might not have a plan if sign up for pro, but abandon checkout
-  // Or it might not have a plan if the license is set in the env var but the license server wasn't whitelisted.
+  // Licenses might not have a plan if sign up for pro, but abandon checkout, in which case we don't want to show an error
+  // Or it might not have a plan if the license is set in the env var but the license server wasn't whitelisted, in which case we do want to show the error
   if (!licenseData || !licenseData.plan) {
+    if (licenseData?.lastServerErrorMessage) {
+      return "License server down for too long";
```
jdorn, line 826:
```suggestion
      return "License server unreachable for too long";
```

packages/enterprise/src/license.ts
```diff
@@ -727,6 +727,7 @@ export async function licenseInit(
           if (
             forceRefresh ||
             !mongoCache ||
+            !mongoCache.dateUpdated ||
```
jdorn, line 730: In what scenario would this ever happen?

tzjames: In exactly the situation where we never got a reply from the server but errored the very first time.  We save that error in the mongoCache. `dateUpdated` is really `dateUpdatedFromServer`. I'll add a comment.

packages/enterprise/src/license.ts
tzjames: This was more of a error code than an error message.  The front end in AccountNotices changed what it says to the user.  

But I went ahead and split the two cases on the back-end now.

jdorn (approved):

PR #2420, opened 2024-04-22
Mk/add elixir to supported sdks

mkurkov:
### Features and Changes

Adds instructions and docs for the new official Elixir SDK.

### Dependencies

No.

### Testing

See Elixir SDK in docs and instructions.

### Screenshots

image
image


packages/shared/src/sdk-versioning/sdk-versions/elixir.json
```diff
@@ -0,0 +1,8 @@
+{
+  "versions": [
+    {
+      "version": "0.2.0",
+      "capabilities": ["prerequisites", "semverTargeting", "bucketingV2"]
```
jdorn, line 5: The most important capability is `looseUnmarshalling`.  That means it will just ignore any new fields in the feature JSON that it doesn't recognize.  That allows us to add new functionality without breaking existing implementations.

If that is supported already, can we add it to this list?  If not, can we add it to the Elixir SDK soon?

mkurkov: Yes, Elixir ignores unknown fields, will add `looseUnmarshaling`.

mkurkov: Done

jdorn:
Looks like there is a CI error.  Need to add `elixir` to the type SDKLanguage in `packages/back-end/types/sdk-connection.d.ts`

PR #2435, opened 2024-04-26
REST API - Allow updating in-progress conversion, attribution model, and stats engine for experiment settings

### LINKED ISSUE #2184: [Feature] Allow setting Experiment Analysis Settings on createExperiment and updateExperiment
### Description of Feature

Right now the following analysis-relevant settings can't be modified via the API:

```
inProgressConversions
attributionModel
statsEngine
```

Adding `attributionModel` and `statsEngine` is pretty straightforward, but `inProgressConversions` is a little more complicated, as it is a relatively new field to replace the deprecated(?) `skipPartialData` (which also wasn't exposed in these APIs). `skipPartialData` is still used in a lot of places

bttf:
Closes #2184


lukesonnet (approved):
LGTM. Seems like we need to add the same fields to the PostExperimentPayload and the same method in `experiments.ts`. Feel free to do so if you want to.

bttf:
> LGTM. Seems like we need to add the same fields to the PostExperimentPayload and the same method in `experiments.ts`. Feel free to do so if you want to.

That makes sense. Let me add that

PR #2437, opened 2024-04-28
Improve past experiment imports

jdorn:
### Features and Changes

This PR fixes a number of UX and performance issues when importing existing experiments from a data source into GrowthBook.

Changes:
- [x] Improve labels, headers, and tooltips to be more intuitive
- [x] Perform incremental updates when refreshing the list (with option to do a full refresh)
- [x] On initial load, do not start the query immediately.  Tell the user we are about to scan X days of data and have them confirm before starting.
- [x] Move all filtering from the back-end to the UI and persist in localStorage  to make exploration and debugging easier.
- [x] Smarter logic for guessing variation weights, plus warn in UI that this is a guess and might be inaccurate
- [x] Right after importing an experiment, auto-update results (opt-out toggle)

### Screenshots

Note: My local `IMPORT_LIMIT_DAYS` variable is set to 2000 in the screenshots below.  The default value is 365.

image

image

image

image

image

image

New toggle when importing to auto-run result queries on save (default ON):

image


lukesonnet (approved):
Overall seems to work with some light QA and quick code review. Some nits were shared with you in comments.

PR #2375, opened 2024-04-10
Migrates organizationSettings to permissionsClass

mknowlton89:
Features and Changes
NOTE: This PR is currently based on mk/migrate-createSegments - before we merge, we'll want to make sure we merge in https://github.com/growthbook/growthbook/pull/2374, and change this to merge into main. I've got it set to merge into mk/migrate-createSegments in order to show a clean diff for review.

This PR migrates `organizationSettings` permission to the permissionClass and introduces a new helper method: `canManageOrgSettings`

Dependencies
This PR needs to be merged after https://github.com/growthbook/growthbook/pull/2374

~~We'll also need to make sure https://github.com/growthbook/growthbook/pull/2353 is merged as it introduces changes to `Layout.tsx` that this PR relies on.~~

Testing
See test suite

PR #2443, opened 2024-04-29
Migrates superDelete to permission class

mknowlton89:
### Features and Changes

This PR migrates the `superDelete` permission to the `PermissionsClass` and in doing so, renames it to `canSuperDeleteReport`. When looking at the use for `superDelete`, it was only being used to check if the user had permission to delete a report created by someone other than themselves.

The existing permission `superDelete` has no direct naming connection to reports, and if the only intended purpose for `superDelete` is around reports, I think we should update the permission name to `superDeleteReport` to prevent any unintended uses.

### Testing

- [x] Ensure no regressions were introduced - admin users can delete reports they didn't create, but non-admins don't have that option
- [ ] TODO: I will add test cases to `permissions.test.ts`, but I'd like to merge in https://github.com/growthbook/growthbook/pull/2374 and https://github.com/growthbook/growthbook/pull/2375 first to avoid having to later resolve merge conflicts.


jdorn (approved):
There are some merge artifacts since this is based on migrate-organizationSettings and not main.  The changes related to this branch look fine though.

PR #2445, opened 2024-04-29
projects UI for approval flows

gazzdingo:
### Features and Changes
added the frontend for Approval flows to do project scoping. based on the the way this works we are checking first for projects then we are checking for environments 
the front end changes are in settings > general > approval flows 

### Screenshots
image

### DIFF
--- a/packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
+++ b/packages/front-end/components/GeneralSettings/FeaturesSettings.tsx
@@ -155,7 +155,7 @@ export default function FeaturesSettings() {
                           label: e.name,
                         };
                       })}
-                      placeholder="All Environments"
+                      placeholder="All Projects"
                     />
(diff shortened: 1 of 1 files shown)


msamper:
Ran into a case while testing where I set approval flows to be enabled for dev in a specific project & the feature I was testing was only enabled for production but it still required an approval flow. May be a bug if I'm not misunderstanding how this should work. A small NIT I'd add that doesn't necessarily pertain to this PR's changes is to reduce the margin at the top of the approval flows settings container so `Require approval to publish changes` doesn't have as much white space above

Screenshot 2024-04-30 at 4 49 27 PM
Screenshot 2024-04-30 at 4 50 22 PM

msamper (approved):
Bug turned out to be a feature! The change I was making was to the default value which applies to all environments so the behavior was as expected. 🚢 it

PR #2449, opened 2024-04-29
Migrate manageTeams to permissionsClass

mknowlton89:
### Features and Changes

This PR migrates the `manageTeams` permission to the permission class and introduces a new helper function `canManageTeam`.

This is a relatively straightforward migration.

### Dependencies

- https://github.com/growthbook/growthbook/pull/2374 changes the name of the method this PR uses from `checkGlobalPermission` to `checkGlobalFilterPermission`, so I'll need to update the naming once 2374 is merged in.

### Testing

- [x] Ensure no regressions were added when viewing the app (ensure that admins only can still see the `Teams` link in the sidebar.
- [x] Ensure that admins only can add/remove team members
- [x] Ensure that the SCIM integration continues to add the same permission checks as it did before
- [x] Build test suite


packages/front-end/components/GuidedGetStarted/GuidedGetStarted.tsx
```diff
@@ -11,13 +11,13 @@ import { useUser } from "@/services/UserContext";
 import { useDefinitions } from "@/services/DefinitionsContext";
```
mknowlton89, line 11: It doesn't look like we're using this anymore - the code in `getstarted.tsx` has a hard-coded value that means this won't render anymore.

jdorn (approved):

PR #2450, opened 2024-04-29
SDK: payload hydration, experiment blocking, enhanced auto-wrapper

bryce-fitzsimons:
### Features and Changes

- v1.0.0 of JS and React SDKs
- Enhanced auto wrapper SDK

From the changelog:
- New `init` function as a replacement for `loadFeatures`.
- Support payload import/export via `init({ payload })`, `setPayload`, and `getPayload`. Better supports hybrid apps, as well as externalizing the SDK Payload.
- Ability to block or control experiments by various context properties
  - `blockedChangeIds` targets individual experiments
  - `disableVisualExperiments` prevents visual experiments from running
  - `disableJsInjection` blocks any visual experiments which write <script\> code
  - `disableUrlRedirectExperiments` prevents redirect experiments from running
  - `disableCrossOriginUrlRedirectExperiments` blocks redirects if the origin changes
  - `jsInjectionNonce` injects a nonce string onto any generated <sript\> tags
  - `disableExperimentsOnLoad` prevents AutoExperiments from running automatically; `triggerAutoExperiments` manually triggers them.
- Provide a custom DOM mutation method by setting context `applyDomChangesCallback`. Useful for server-side rendering visual experiments.
- Get a list of triggered experiments by their `changeId` (a new, more-specific AutoExperiment identifier) via `getCompletedChangeIds`.
- Disable in-memory cache via `configureCache({ disableLocalCache: true })`
- Easier streaming customization via `init({ streaming: true })`.
- Prefetch payloads on backend SDK instances via `prefetchPayload()`. Also allows feature repository SSE streaming without connection each SDK instance to streaming (useful for NodeJS).
- Many improvements to `auto.min.js` script
  - Payload hydration via `payload` context property
  - Support sticky bucketing via `useStickyBucketService` context property (accepts "cookie" and "localStorage"); override cookie or localStorage key via `stickyBucketPrefix` context property.
  - Hydrate sticky bucket docs by setting context `stickyBucketAssignmentDocs` (useful for SSR hydration without reliance on cookies)
  - Customize attributes/cookies via `uuidCookieName` (default "gbuuid") and `uuidKey` (default "id").
  - Force a uuid cookie write on load via `persistUuidOnLoad` (useful for SSR hydration)


packages/sdk-js/src/feature-repository.ts
```diff
@@ -428,20 +461,30 @@ async function fetchFeatures(
             error: e ? e.message : null,
           });
         activeFetches.delete(cacheKey);
-        return Promise.resolve({});
+        const existing = !cacheSettings.disableLocalCache
+          ? cache.get(cacheKey)
+          : undefined;
```
jdorn, line 466: Should we default to `instance.getPayload()` instead?  Thinking about Next14 SSR, we will disable local cache.  If you call `gb.refreshFeatures()` and the fetch fails, we wouldn't want to overwrite the instance's existing payload with an empty object.

packages/sdk-js/src/auto-wrapper.ts
```diff
@@ -1,12 +1,29 @@
...
+  StickyBucketService,
+} from "./index";
+
+type WindowContext = Context & {
+  uuidCookieName?: string;
+  uuidKey?: string;
+  uuid?: string;
+  persistUuidOnLoad?: boolean;
+  useStickyBucketService?: "cookie" | "localStorage";
+  stickyBucketPrefix?: string;
```
jdorn, line 17: We should support most of these on the data context as well.  Outside of edge, I think being able to add `sticky-buckets="localstorage"` to the script tag will be really useful to people.

packages/sdk-js/src/auto-wrapper.ts
```diff
@@ -198,8 +240,20 @@ gb.setRenderer(() => {
...
+    for (const key in windowContext.stickyBucketAssignmentDocs) {
+      const doc = windowContext.stickyBucketAssignmentDocs?.[key];
+      if (doc) {
+        await stickyBucketService.saveAssignments(doc);
+      }
+    }
+  }
+
+  // Load features/experiments
+  gb.init({ payload: windowContext.payload });
```
jdorn, line 255: To match existing behavior, we should default to `streaming: true` here, but give people a way to opt-out via window or data context (can re-use existing `backgroundSync` setting if we want for this)

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -114,9 +114,10 @@ export type Experiment<T> = {
   groups?: string[];
 };
 
+export type AutoExperimentChangeType = "redirect" | "visual" | "unknown";
```
bryce-fitzsimons, line 117: export this from the package as well

packages/sdk-js/src/feature-repository.ts
bryce-fitzsimons: would something like this work where it's always plugging in getPayload() by default?
```
!cacheSettings.disableLocalCache
          ? cache.get(cacheKey)
          : undefined
          ?? instance.getPayload();
```

jdorn (approved):

PR #2453, opened 2024-04-30
REST API - Include per-variation user count in experiment results

bttf:


lukesonnet (approved):
This is fine. But they should probably be using the `denominator` field inside of `analyses` given that this is nested within metric. But I'm all for just returning the data we have for now until we create a better results data model/API.

PR #2454, opened 2024-04-30
Migrates manageTags to permissionClass

mknowlton89:
### Features and Changes

This PR migrates the `manageTags` permission to the `PermissionClass` and introduces 3 new helper methods: `canCreateTag`, `canUpdateTag`, and `canDeleteTag`. Previously, we only had the single "manageTag" check, so this will set us up should we want to support more finer-grain controls down the road.

Likewise, this adds a new `PUT` endpoint to the tags router, and an accompanied controller function to handle different logic for updating vs creating a tag.

This PR also introduces more front-end controls to only show create, update, and delete CTAs if the user has permission to do so.

### Dependencies

Like the other global filter permissions, I'll need to update the signature of `checkGlobalPermission` to `checkGlobalFilterPermission` once https://github.com/growthbook/growthbook/pull/2374 is merged in

### Testing

- [x] Ensure no new regressions are introduced - only `anaylsts`, `engineers`, and `experimenters` can add tags.
- [x] See test suite


jdorn:
Because of how tags work, I would merge the update/create into a single permission.  Will keep the code much simpler and greatly reduce the size of this PR.  I can't think of a scenario where someone would want to differentiate between create/update for tags.

mknowlton89:
> Because of how tags work, I would merge the update/create into a single permission. Will keep the code much simpler and greatly reduce the size of this PR. I can't think of a scenario where someone would want to differentiate between create/update for tags.

@jdorn Just updated so instead of having separate `canCreateTag` and `canUpdateTag` methods - we not have a singular `canCreateAndUpdate` method.

jdorn:
Yes, I would revert that PUT endpoint change

jdorn (approved):

PR #2458, opened 2024-05-01
Quick fix to check for multiple values that are pasted into the invite modal

Auz:


tzjames (approved):
Looks good.  I also tried it out and it works.  The PR is missing a test plan though.  Here is the one I did:

- Go to Settings->Team
- Click "Invite Member"
- Copy and paste `a@a.com, b@b.com,  a@a.com c@c.com` and then click away
- See three emails since a@a.com was duplicated.
- Paste it a second time and click away and see nothing get added since they are all duplicates.

PR #2459, opened 2024-05-01
Some minor edits to the docs

Auz:


docs/docs/faq.mdx
```diff
@@ -154,6 +154,55 @@ When you create an SDK connection, it is linked to a specific environment. You c
...
+    if (parts.length === 2) return parts.pop().split(';').shift();
+  }
+  const setCookie = (name, value) => {
+    var d = new Date();
+    d.setTime(d.getTime() + 24*60*60*1000*COOKIE_DAYS);
+    document.cookie = name + "=" + value + ";path=/;expires=" + d.toGMTString();
+  }
+
+  // get the existing UUID from cookie if set, otherwise create one and store it in the cookie
+  if(getCookie(COOKIE_NAME)) return getCookie(COOKIE_NAME);
```
tzjames, line 192: For long time regular users this might be a problem.  On the 401st day they will be assigned a new UUID which might mess up experiments.  If we reset the cookie every time we get it, then as long as the don't have a 400 day break it should be fine.

```
// get the existing UUID from cookie if set, otherwise create one
  let uuid = getCookie(COOKIE_NAME);
  if(!uuid) {
    uuid = genUUID();
  }
  
  // Set the new cookie or refresh the cookie's expiration date
  setCookie(COOKIE_NAME, uuid);
  
  return uuid;
```

We should probably do that in our SDK then as well.

docs/docs/faq.mdx
```diff
@@ -154,6 +154,55 @@ When you create an SDK connection, it is linked to a specific environment. You c
 
 Throughout the GrowthBook application, we randomly celebrate key milestones like launching experiments with on-screen confetti. If you'd like to disable this, you can click on your avatar in the top right corner and select "Edit Profile". From there, you can disable the toggle for "Allow Celebrations". Please note this is persisted in your browser's local storage, so if you clear your browser's local storage, you will need to disable this again.
 
+## How do I make my own identifier?
+
+There are cases when using feature flags client side where the 3rd party identifiers used for assignment will be slow to
+load, and may cause flickering as some of the DOM rerenders. In these cases, generating your
+own identifier will make sure that features are assigned correctly when GrowthBook loads. This id that is generated will
+typically align one to one with the other identifiers, and does not need to be passed outside the SDK.
```
tzjames, line 162: Might we want to mention that this is for a logged-out identifier, and that if the user logs in to another device they will count as a different user.  Hence it would not match up one to one with a logged-in identifier.

docs/docs/faq.mdx
Auz: Should we have an option of passing in the experiment id to be used for this cookie name? That way they'd only hit issues with experiments that run for 400+ days.

tzjames: > Should we have an option of passing in the experiment id to be used for this cookie name? That way they'd only hit issues with experiments that run for 400+ days.
I would think not, since then they might end up having one cookie per experiment.  It would wreck namespaces, no? 

I would imagine refreshing the cookie's expiry is good enough as that would work for experiments < 400 days, and even work for those older than 400 days as long as it is safe to assume long absent users never return.  What is your concern with that approach?

Auz: Good point, updated.

PR #2461, opened 2024-05-01
Migrate manageWebhooks to permissionsClass

mknowlton89:
### Features and Changes

This PR migrates the `manageWebhooks` permission to the permission class and introduces 3 new helper methods: `canCreateWebhook`, `canUpdateWebhook`, and `canDeleteWebhook`.

I could see two additional helpers being useful, depending on the future direction of this feature - those helpers being `canViewWebhookLogs`, and `canTestWebhook`. For now, I'm not doing any permission checking on viewing the webhook logs as with other globally-scoped permissions, we allow them to be seen by everyone. I could see an argument for treating these differently, though, so happy to make any changes.

And when it comes to testing, if the user can create, update, or delete, I've enabled them to test the webhook.

### Testing
 - [x] Ensure admin users are still able to create, update, test, and delete SDK Webhooks
 - [x] Ensure non-admins aren't bale to create, update, test, or delete SDK Webhooks
 - [x] Ensure admin users are still able to create, update, test, and delete Event Webhooks
 - [x] Ensure non-admins aren't bale to create, update, test, or delete Event Webhooks
 - [x] Ensure admin users are still able to create, update, test, and delete SDK Endpoint Webhooks (from `/sdks#legacy`)
 - [x] Ensure non-admins aren't bale to create, update, test, or delete SDK Endpoint Webhooks (from `/sdks#legacy`)
 
 I know we discussed breaking this permission out into one for each type of webhook - happy to do that if you think its appropriate right now.


packages/back-end/src/routers/event-webhooks/event-webhooks.controller.ts
```diff
@@ -56,12 +55,14 @@ export const getEventWebHook = async (
   req: GetEventWebHookByIdRequest,
   res: Response<GetEventWebHookByIdResponse | PrivateApiErrorResponse>
 ) => {
-  req.checkPermissions("manageWebhooks");
```
mknowlton89, line 59: Same here - For all other globally-scoped resources, we don't filter them by user read access, so keeping that pattern here.

packages/front-end/components/EventWebHooks/EventWebHooksPage.tsx
```diff
@@ -1,23 +0,0 @@
...
-      </div>
-    );
-  }
-
-  return (
-    <div className="pagecontents">
-      <EventWebHookListContainer />
-    </div>
-  );
-};
```
mknowlton89, line 23: We were checking the user's permission level twice - once here, and one on the page this component's parent component. Removing it make this abstraction unnecessary, imo.

jdorn:
Yes, I would separate the permission functions for SDK and Event webhooks.  We are going to need to do that when we define policies

packages/back-end/src/routers/event-webhooks/event-webhooks.controller.ts
```diff
@@ -33,8 +34,6 @@ export const getEventWebHooks = async (
   req: GetEventWebHooksRequest,
   res: Response<GetEventWebHooks>
 ) => {
-  req.checkPermissions("manageWebhooks");
```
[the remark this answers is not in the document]
jdorn: That isn't quite true.  We do the same for secret API keys where only admins can view them.  Event Webhooks are similar since they contain a signing secret that we don't want to expose to all users in the org.

Let's create `canView` helper functions for webhooks that can just check for `manageWebhooks` under the hood for now.

packages/back-end/src/routers/event-webhooks/event-webhooks.controller.ts
```diff
@@ -148,8 +151,6 @@ export const getEventWebHookLogs = async (
   req: GetEventWebHookLogsRequest,
   res: Response<GetEventWebHookLogsResponse | PrivateApiErrorResponse>
 ) => {
-  req.checkPermissions("manageWebhooks");
```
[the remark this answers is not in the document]
jdorn: Just a note: we're not planning to deprecate event webhooks

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -188,7 +188,10 @@ const navlinks: SidebarLinkProps[] = [
         name: "Webhooks",
         href: "/settings/webhooks",
         path: /^settings\/webhooks/,
-        filter: ({ permissions }) => permissions.check("manageWebhooks"),
+        filter: ({ permissionsUtils }) =>
+          permissionsUtils.canCreateEventWebhook() ||
+          permissionsUtils.canUpdateEventWebhook() ||
+          permissionsUtils.canDeleteEventWebhook(),
```
jdorn, line 194: When we add a `canViewEventWebhook()` permission, I think that's the only one we need to check here.

packages/front-end/pages/sdks/SdkWebhooks.tsx
```diff
@@ -52,53 +54,54 @@ export default function SdkWebhooks({ sdkid }) {
...
-                  />
-                </MoreMenu>
-              </div>
-            </>
-          )}
+          <Button
+            color="link"
+            className="btn-sm"
+            disabled={
+              !canCreateWebhooks || !canUpdateWebhook || !canDeleteWebhook
```
jdorn, line 62: Let's just use `canUpdateWebhook` for this.  We could create a new permission for testing webhooks, but that's probably overkill.  If they can update a webhook, then they should be able to test it.

packages/front-end/pages/sdks/SdkWebhooks.tsx
```diff
@@ -52,53 +54,54 @@ export default function SdkWebhooks({ sdkid }) {
...
+                  }}
+                >
+                  Edit
+                </button>
+              ) : null}
+              {canDeleteWebhook ? (
+                <DeleteButton
+                  className="dropdown-item"
+                  displayName="SDK Connection"
+                  text="Delete from SDK Webhook"
```
jdorn, line 93: This copy change makes it more confusing to me what's being deleted.

jdorn (approved):
Approving, pending the fixes from comments and CI

PR #2464, opened 2024-05-01
Update json-shema-to-zod to 2.1.0

lukesonnet:
Updates `json-schema-to-zod` so that we can properly use `exclusiveMinimum` and because the version we used was super old.

Had one main side effect that I can tell (besides adding a bunch of whitespace):

Before, `format: date` on a string field and `format: date-time` didn't do anything as the json schema to vod package skipped it. It produced in `z.string()`. Now, it produced `z.string().date()`  or something similar for date-time.  Because we don't have any true `dates` but those fields are actually `datetime`, I changed the yaml to match the right datatype. However, now we are validating these inputs as date-time and this could break user flows that previously worked (e.g. passing epoch time as a string).


bttf (approved):
Since we're already using `date-time` in most places, this seems fine and necessary to correct. It might break for some users but have to rip the bandaid at some point.

PR #2465, opened 2024-05-01
Migrates manageNamespaces to permissions class

mknowlton89:
### Features and Changes

This PR migrates the `manageNamespaces` permission to the permissionsClass and introduces 3 new helper methods: `canCreateNamespace`, `canUpdateNamespace`, and `canDeleteNamespace`.

### Dependencies
- If https://github.com/growthbook/growthbook/pull/2463 is merged in before this PR, I will update the `permissionsClass.test.ts` file with tests for these methods. Otherwise, I will add those tests to PR 2463.

### Testing

- [x] Ensure that only users with global engineer, experimenter, or admin roles can manage namespaces.


packages/front-end/pages/namespaces.tsx
```diff
@@ -88,7 +88,7 @@ const NamespacesPage: FC = () => {
               <th>Description</th>
               <th>Active experiments</th>
               <th>Percent available</th>
-              {canEdit && <th style={{ width: 30 }}></th>}
+              <th style={{ width: 30 }}></th>
```
mknowlton89, line 91: Removing this simplified things quite a bit. Together with the change on line 120 of `NamespaceTableRow.tsx` ensures that the table is always full width, regardless of if the user has permission to edit or not.

jdorn (approved):

PR #2467, opened 2024-05-02
Migrates manageArchetypes to permissionClass

mknowlton89:
### Features and Changes

This PR migrates the `manageArchetypes` permission to the permissionsClass and introduces 3 new helper methods: `canCreateArchetype`, `canUpdateArchetype`, and `canDeleteArchetype`. 

This PR also introduces a change to two archetype controller functions - the `getArchetype` and `getArchetypeAndEval`. These were checking for `manageArchetype` permission - so if a user had a global `noaccess` role, but project-specific roles that granted them permission to manage features, they wouldn't be able to use the assignment tester (since archetypes are a global resource). Now, they behave similarly to other globally scoped resources (e.g. saved groups), in that the user can use them, but they can't create, update, or delete them.

### Dependencies

If https://github.com/growthbook/growthbook/pull/2463 is merged in before this PR, I will update the permissionsClass.test.ts file with tests for these methods. Otherwise, I will add those tests to PR 2463.

### Testing

- [x] Ensure that only engineers, experimenters, and admins can add/update/remove archetypes while everyone else can only view them


packages/back-end/src/routers/archetype/archetype.controller.ts
```diff
@@ -40,8 +40,6 @@ export const getArchetype = async (
 ) => {
   const { org, userId } = getContextFromReq(req);
 
-  req.checkPermissions("manageArchetype");
```
mknowlton89, line 43: I removed this here, and on line 73 below. For other globally scoped resources, we don't filter them out of view for a user even if the user has a global role of `noaccess`. 

In this scenario, if a user has a global `noaccess` role, and then project-specific roles that allow them to manage features, they wouldn't see any archetypes because we'd be filtering them all out. 

With this update, archetypes now behave similar to how resources like saved groups work.

packages/front-end/components/Archetype/ArchetypeResults.tsx
```diff
@@ -428,26 +433,30 @@ const ArchetypeResults: FC<{
...
-                      text="Delete"
-                      useIcon={false}
-                      onClick={async () => {
-                        await apiCall(`/archetype/${archetype.id}`, {
-                          method: "DELETE",
-                        });
-                        onChange();
-                      }}
-                    />
+                    {canEdit ? (
```
mknowlton89, line 436: In other areas of the app, we don't show edit or delete options if we know the user doesn't have permission - this update makes this table behave more consistently with other areas of the app.

jdorn (approved):

Issue #2442, opened 2024-04-29
AWS Athena - Assume Role


thatguyfig:
### Description of Feature

Hi,

Currently in GrowthBook, one can configure an Athena Datasource, however this is a problem for multi-tenant uses of GrowthBook.

Currently with Athena, you have the option of:

- Use `auto-detect` which leverages the instance's IAM instance profile to obtain credentials and interact with Athena, OR
- You can use an IAM user created for programmatic access (secret key + access key) and embed them in the connection.

The above two will work fine, but I suggest a rework of this to allow for much more complicated permission setups.

The reason I say this is because:

- with `auto-detect` this basically requires the IAM role to have access to everything it may need to query. Which as a shared resource could reveal datasets to users who should not have access to see those datasets. Currently you can only set a default database, which doesn't stop users querying other databases in Athena.
- on the flip-side, managing access keys and secrets keys specifically for each connection is a major pain, not to mention all the fun involved with revoking those keys if they should be leaked. As you can probably imagine, this gets very complicated the more Athena connections you need - and rotating them regularly would be hard.

## What would work much better?

I suggest a third option maybe named `Assume Role` which leverages the current IAM role (the same role found during `auto-detect`) and uses this role to assume another (the one supplied in connection setup) which performs an `aws sts assume-role` to get temporary credentials via STS. Once this is done, the temporary credentials can be used for the process to interact with Athena.

Ultimately, this means we can completely lock down the instance's role to talk with Athena, and only allow it to use Athena through specially provisioned roles, which restrict the resources it can operate on.

With this feature we could actually consider introducing the Athena datasource to our users.

cleaton:
This can be achieved with https://github.com/growthbook/growthbook/pull/2438

thatguyfig:
The in-review PR for this feature is [here](https://github.com/growthbook/growthbook/pull/2471)

PR #2438, opened 2024-04-28
Use aws-sdk v3 for athena connection

cleaton:
### Features and Changes

Update athena-client to use aws-sdk v3.
This enables GB to connect to athena running in another aws account using cross account assume role.

https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html

```
SDK	                Supported	Notes or more information
SDK for JavaScript 3.x	Yes	
SDK for JavaScript 2.x	Partial	        credential_source not supported.
```

aws-sdk v2 is also in maintenance mode & marked for deprecation:
https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/

### Dependencies

None.

### Testing

Updated existing athena connected GB instance.

### Screenshots

No UI-changes.

PR #2471, opened 2024-05-02
Added Support for performing Assume role with Athena Connections

thatguyfig:
### Features and Changes

Provides the UI and backend for performing an assume role before running Athena queries. Since the `getAthenaInstance` function is used to obtain a client, this can also be used to interact with sts beforehand before returning the client with embedded credentials - this also incorporates the JavaScript v3 SDK.

I'm not to sure how the errors are propagated up the UI but it seems to work pretty well - so that should be one area to improve. Also the session is currently defaults to 900 seconds which could be too short for some beefy queries.

### DIFF
--- a/packages/back-end/src/services/athena.ts
+++ b/packages/back-end/src/services/athena.ts
@@ -27,7 +27,7 @@ async function getAthenaInstance(params: AthenaConnectionParams) {
   }
 
   // handle assuming a role first
-  if (!IS_CLOUD && params.authType == "assumeRole") {
+  if (!IS_CLOUD && params.authType === "assumeRole") {
     // use client to assume another role
     const credentials = await assumeRole(params);
 
--- a/packages/back-end/src/services/athena.ts
+++ b/packages/back-end/src/services/athena.ts
@@ -1,20 +1,52 @@
-import { Athena } from "aws-sdk";
-import { ResultSet } from "aws-sdk/clients/athena";
+import { STSClient, AssumeRoleCommand } from "@aws-sdk/client-sts";
+import { Athena, ResultSet } from "@aws-sdk/client-athena";
 import { AthenaConnectionParams } from "../../types/integrations/athena";
 import { logger } from "../util/logger";
 import { IS_CLOUD } from "../util/secrets";
 import { ExternalIdCallback, QueryResponse } from "../types/Integration";
 
-function getAthenaInstance(params: AthenaConnectionParams) {
+async function assumeRole(params: AthenaConnectionParams) {
+  // build sts client
+  const client = new STSClient({ region: "us-east-1" });
+  const command = new AssumeRoleCommand({
+    RoleArn: params.assumeRoleARN,
+    RoleSessionName: params.roleSessionName,
+    ExternalId: params.externalId,
+    DurationSeconds: 900,
+  });
+
+  return await client.send(command);
+}
+
+async function getAthenaInstance(params: AthenaConnectionParams) {
+  // handle the instance profile
   if (!IS_CLOUD && params.authType === "auto") {
     return new Athena({
       region: params.region,
     });
   }
 
+  // handle assuming a role first
+  if (!IS_CLOUD && params.authType == "assumeRole") {
+    // use client to assume another role
+    const credentials = await assumeRole(params);
+
+    return new Athena({
+      credentials: {
+        accessKeyId: credentials?.Credentials?.AccessKeyId || "",
+        secretAccessKey: credentials?.Credentials?.SecretAccessKey || "",
+        sessionToken: credentials?.Credentials?.SessionToken || "",
+      },
+      region: params.region,
+    });
+  }
+
+  // handle access key + secret key
   return new Athena({
-    accessKeyId: params.accessKeyId,
-    secretAccessKey: params.secretAccessKey,
+    credentials: {
+      accessKeyId: params.accessKeyId || "",
+      secretAccessKey: params.secretAccessKey || "",
+    },
     region: params.region,
   });
 }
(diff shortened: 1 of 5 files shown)


thatguyfig:
There is another PR floating around that referenced my issue, but the code seems completely half-baked

https://github.com/growthbook/growthbook/pull/2438

thatguyfig:
https://github.com/growthbook/growthbook/issues/2442

cleaton:
> There is another PR floating around that referenced my issue, but the code seems completely half-baked
> 
> #2438

the upgraded SDK allows configuring credentials via aws cli config file, can add assume role configuration in there

thatguyfig:
> > There is another PR floating around that referenced my issue, but the code seems completely half-baked
> > 
> > #2438
> 
> the upgraded SDK allows configuring credentials via aws cli config file, can add assume role configuration in there and auto will pick it up.

Yes that is possible, but that doesn't offer a solution for when you have different connections in different accounts and need to be able to assume a specific role. 

With this approach the assume role is tied to the connection and can be specified when you set it up so it can be created by the users themselves.

cleaton:
> > > There is another PR floating around that referenced my issue, but the code seems completely half-baked
> > > #2438
> > 
> > 
> > the upgraded SDK allows configuring credentials via aws cli config file, can add assume role configuration in there and auto will pick it up.
> 
> Yes that is possible, but that doesn't offer a solution for when you have different connections in different accounts and need to be able to assume a specific role.
> 
> With this approach the assume role is tied to the connection and can be specified when you set it up so it can be created by the users themselves.

Sorry I think there is some misunderstanding here, I don’t mean our PRs are equal or anything I only referenced it as a possible workaround as there was no PR mentioned there yet. 


We want the same thing and this is a nice solution so I’m all for adding this instead, thank you for sharing it 👍

packages/back-end/src/services/athena.ts
```diff
@@ -1,20 +1,52 @@
...
+import { Athena, ResultSet } from "@aws-sdk/client-athena";
 import { AthenaConnectionParams } from "../../types/integrations/athena";
 import { logger } from "../util/logger";
 import { IS_CLOUD } from "../util/secrets";
 import { ExternalIdCallback, QueryResponse } from "../types/Integration";
 
-function getAthenaInstance(params: AthenaConnectionParams) {
+async function assumeRole(params: AthenaConnectionParams) {
+  // build sts client
+  const client = new STSClient({ region: "us-east-1" });
```
jdorn, line 10: Should this come from `params.region` instead of being hard-coded?

packages/back-end/src/services/athena.ts
```diff
@@ -1,20 +1,52 @@
...
+  const command = new AssumeRoleCommand({
+    RoleArn: params.assumeRoleARN,
+    RoleSessionName: params.roleSessionName,
+    ExternalId: params.externalId,
+    DurationSeconds: 900,
```
jdorn, line 15: It looks like `params.durationSeconds` was added, but is not being used here

packages/front-end/components/Settings/AthenaForm.tsx
```diff
@@ -24,9 +24,13 @@ const AthenaForm: FC<{
...
               },
+              {
+                value: "assumeRole",
+                display: "Assume IAM Role",
+              },
             ]}
-            helpText="'Auto-discovery' will look for credentials in environment variables and instance metadata."
-            value={params.authType || "accessKey"}
+            helpText="'Auto-discovery' will look for credentials in environment variables and instance metadata. 'Assume IAM Role' uses the current role to assume another role and execute Athena with temporary credentials."
+            value={params.authType || "auto"}
```
jdorn, line 33: Change this default back to `accessKey` since that was the default before we introduced the `authType` param.

packages/front-end/components/Settings/AthenaForm.tsx
```diff
@@ -86,6 +68,68 @@ const AthenaForm: FC<{
           </div>
         </>
       )}
+      {(isCloud() || params.authType == "assumeRole") && (
```
jdorn, line 71: I think this should be `!isCloud() && ` instead of `isCloud() || `

jdorn (commented):
Thanks for contributing!  Left a few comments

packages/back-end/src/services/athena.ts
thatguyfig: Ah, so with STS it's a global service but I wasn't sure if the client needs a region supplied, ideally not.

Since Athena can be in different regions it makes sense to feed that in as a connection parameter during setup.

packages/back-end/src/services/athena.ts
thatguyfig: I'll look at passing that through from the connection.

packages/front-end/components/Settings/AthenaForm.tsx
thatguyfig: Will do

packages/front-end/components/Settings/AthenaForm.tsx
thatguyfig: Good catch, will fix that

thatguyfig: fixed

packages/front-end/components/Settings/AthenaForm.tsx
thatguyfig: done

packages/back-end/src/services/athena.ts
thatguyfig: passed from connection

packages/back-end/src/services/athena.ts
thatguyfig: dropped the region as it's not required

thatguyfig:
Thanks @jdorn - issues have been addressed, please could re-review?

thatguyfig:
bump

PR #2475, opened 2024-05-03
Zero downtime deploys

tzjames:
### Features and Changes
On each deploy some frontends that were loaded from an old server might try and read the files from a new server and vice-versa.  This PR tries to fix that situation by:
- deploying our static files to s3 in a bucket with the git date and hash prepended.
- building a `latest-cloud` docker image where the front-end app has an assetPrefix for nextJS that will look for static files that are normally in `.next/static` locally from s3 instead
- automatically creates a `variables.scss` file with a `$public-asset-prefix` variable that can be imported in scss files and prepended to an urls that would normally be looking for the files in `front-end/public`
- if `IS_CLOUD=true` will set a new `getPublicAssetsPath()` function to return the s3 public url.
- Updates all the images in the front end that had been reading from `front-end/public` to prepend the `getPublicAssetsPath()` function

### Dependencies

- [ ] CDN must be pointing to correct s3 bucket
- [ ] Must update urls to use the cdn instead of the s3 bucket directly
- [ ] Remove the test-deploy-with-s3.yml file

### Testing
Have the github action run with test-deploy-with-s3 action which is a modified deploy workflow that overwrites the latest tags to instead push to only test tags, and it doesn't actually do the deploy to ECS step.

Then locally do:
`docker pull --platform linux/amd64 growthbook/growthbook:deploy-with-s3-test-cloud`

Then modify docker-compose.yml:
```
version: "3"
services:
  mongo:
    image: "mongo:latest"
    environment:
      - MONGO_INITDB_ROOT_USERNAME=root
      - MONGO_INITDB_ROOT_PASSWORD=password
    volumes:
      - mongodata:/data/db
  growthbook:
    image: "growthbook/growthbook:latest"
    ports:
      - "3000:3000"
      - "3100:3100"
    depends_on:
      - mongo
    environment:
      - MONGODB_URI=mongodb://root:password@mongo:27017/growthbook?authSource=admin
      - IS_CLOUD=true
      - IS_MULTI_ORG=true
      - NODE_ENV=production
      - DISABLE_TELEMETRY=debug
      - LICENSE_SERVER_URL=http://localhost:8080/api/v1/
      - ENCRYPTION_KEY=asdkfjdkasljnvjkcnqeirnfioernf90nfi34nkjrncoip24inf
      - SSO_CONFIG= <VALID_SSO_CONFIG>
      - USE_REMOTE_ASSETS=true
    volumes:
      - uploads:/usr/local/src/app/packages/back-end/uploads
volumes:
  uploads:
  mongodata:
```

Then run `docker-compose up -d`

Then go to localhost:3000, open up the dev console networking tab, add "url" to the columns.  See that all the urls are hitting the cdn.

### DIFF
--- a/packages/front-end/next.config.js
+++ b/packages/front-end/next.config.js
@@ -21,9 +21,6 @@ const cspHeader = `
 module.exports = {
   // We already run eslint and typescript in CI/CD
   // Disable here to speed up production builds
-  assetPrefix: process.env.IS_CLOUD
-    ? `https://growthbook-cloud-static-files.s3.amazonaws.com/${gitCommitDate}/${gitSha}/`
-    : "",
   eslint: {
     ignoreDuringBuilds: true,
   },
@@ -45,4 +42,7 @@ module.exports = {
       ],
     },
   ],
+  assetPrefix: process.env.USE_REMOTE_ASSETS
+    ? `https://growthbook-cloud-static-files.s3.amazonaws.com/${gitCommitDate}/${gitSha}/`
+    : "",
 };
--- a/packages/front-end/next.config.js
+++ b/packages/front-end/next.config.js
@@ -6,9 +6,10 @@ const rootPath = path.join(__dirname, "..", "..");
 let gitSha = "";
 let gitCommitDate = "";
 if (fs.existsSync(path.join(rootPath, "buildinfo", "SHA"))) {
-  gitSha = trim(
-    fs.readFileSync(path.join(rootPath, "buildinfo", "SHA")).toString().trim()
-  );
+  gitSha = fs
+    .readFileSync(path.join(rootPath, "buildinfo", "SHA"))
+    .toString()
+    .trim();
 }
 if (fs.existsSync(path.join(rootPath, "buildinfo", "DATE"))) {
   gitCommitDate = fs
--- a/packages/front-end/next.config.js
+++ b/packages/front-end/next.config.js
@@ -1,3 +1,19 @@
+const fs = require("fs");
+const path = require("path");
+
+const rootPath = path.join(__dirname, "..", "..");
+
+let gitSha = "";
+let gitCommitDate = "";
+if (fs.existsSync(path.join(rootPath, "buildinfo", "SHA"))) {
+  gitSha = fs.readFileSync(path.join(rootPath, "buildinfo", "SHA")).toString();
+}
+if (fs.existsSync(path.join(rootPath, "buildinfo", "DATE"))) {
+  gitCommitDate = fs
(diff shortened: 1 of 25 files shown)


packages/front-end/pages/dashboard.tsx
```diff
@@ -34,7 +35,7 @@ export default function Analysis(): React.ReactElement {
     <>
       <Head>
         <title>GrowthBook</title>
-        <link rel="icon" href="/favicon.ico" />
+        <link rel="icon" href={`${getPublicAssetsPath()}/favicon.ico`} />
```
romain-growthbook, line 38: Pushing this to every url that needs the treatment feels like it would create a heavy maintenance load and also an easy footgun with public assets that are not prefixed appropriately.

I'd suggest to encapsulate this into a callback to make it clear what is the boundary and abstract away the actual implementation. Something like:
```typescript
const assetUrl: (path: string) => `${getPublicAssetsPath()}${path}`
```

And, then:
```suggestion
        <link rel="icon" href={assertUrl("/favicon.ico")} />
```

Or perhaps even an explicit module with all public asset urls exported?

packages/front-end/next.config.js
```diff
@@ -1,3 +1,32 @@
...
+  gitCommitDate = fs
+    .readFileSync(path.join(rootPath, "buildinfo", "DATE"))
+    .toString()
+    .trim();
+}
+
+fs.writeFileSync(
+  path.join(__dirname, "styles", "variables.scss"),
+  `$public-asset-prefix: "${
+    process.env.USE_REMOTE_ASSETS
```
romain-growthbook, line 24: The logic in `api/init.ts` is different:
```typescript
  const publicAssetsPath =
    IS_CLOUD && USE_REMOTE_ASSETS
      ? `https://growthbook-cloud-static-files.s3.amazonaws.com/${build.date}/${build.sha}/public`
      : "";
```
Is it possible to have only one reference implementation?

packages/front-end/next.config.js
```diff
@@ -1,3 +1,32 @@
...
+      ? `https://growthbook-cloud-static-files.s3.amazonaws.com/${gitCommitDate}/${gitSha}/public`
```
romain-growthbook, line 25: Should `https://growthbook-cloud-static-files.s3.amazonaws.com` also be configurable/an env variable?

romain-growthbook (commented):
The approach looks good and robust. Left some comments. I think parametrizing the bucket URL would leave some room to later push the assets to different places, for instance CDN edges etc.

tzjames:
Going to rewrite from load balancer instead.

PR #2480, opened 2024-05-06
Migrate project to new data model class.

romain-growthbook:


packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,126 +1,82 @@
...
 
-const projectSchema = new mongoose.Schema({
-  id: {
-    type: String,
-    unique: true,
-  },
-  organization: {
-    type: String,
-    index: true,
+const engineTypes = tuple<StatsEngine>()(["bayesian", "frequentist"]);
```
jdorn, line 10: Why is this needed?  Can't we just use a const array?

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,126 +1,82 @@
...
+    return true;
+  }
+  protected canDelete() {
+    return true;
+  }
+}
 
-const ProjectModel = mongoose.model<ProjectInterface>("Project", projectSchema);
+export const findAllProjectsByOrganization = (context: Context) =>
+  new ProjectModel(context).getAll();
```
jdorn, line 52: I'm not the biggest fan of this pattern. I think I'd prefer not exporting helper methods here.  Instead, add this class to `context.models` and access the methods like this throughout the code.

```ts
await context.models.projects.getAll();
```

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,126 +1,82 @@
...
+  globallyUniqueIds: true,
 });
 
-type ProjectDocument = mongoose.Document & ProjectInterface;
+export class ProjectModel extends BaseClass {
+  protected canRead(doc: ProjectInterface) {
+    return hasReadAccess(this.context.readAccessFilter, doc.id);
+  }
+  protected canCreate() {
+    return true;
```
jdorn, line 41: We will want to add actual permission checks here instead of just assuming the upstream code is doing it.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,126 +1,82 @@
...
-  dateUpdated: Date,
-  settings: {},
+  projectScoping: "none",
+  globallyUniqueIds: true,
 });
 
-type ProjectDocument = mongoose.Document & ProjectInterface;
+export class ProjectModel extends BaseClass {
+  protected canRead(doc: ProjectInterface) {
+    return hasReadAccess(this.context.readAccessFilter, doc.id);
```
jdorn, line 38: There is a bug in BaseModel that will prevent this check from running.  This is fixed (along with other bugs) in #2317 . I recommend we bring that PR up-to-date and merge it first as a prerequisite before doing this one.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,126 +1,82 @@
...
+  }
+}
 
-const ProjectModel = mongoose.model<ProjectInterface>("Project", projectSchema);
+export const findAllProjectsByOrganization = (context: Context) =>
+  new ProjectModel(context).getAll();
 
-function toInterface(doc: ProjectDocument): ProjectInterface {
-  const ret = doc.toJSON<ProjectDocument>();
-  ret.settings = ret.settings || {};
```
jdorn, line 33: This is a migration that we'll want to copy over to the new class.  If `settings` is undefined for a project, default it to an empty object.  We have code that is relying on this field always existing.

packages/back-end/src/models/ProjectModel.ts
romain-growthbook: We can. I am just trying to have a safety check that the array items are kept in sync with `StatsEngine`. This is another instance where `StatsEngine` cannot be defined using the usual:

```typescript
const values = [...] as const;
type Values = typeof values[number];
```

Because the API in `back-end/types` is types only.

packages/back-end/src/models/ProjectModel.ts
romain-growthbook: Ok

packages/back-end/src/models/ProjectModel.ts
jdorn: I don't think this tuples approach is very readable.  I'd prefer if we either:

1. Use `[...] as const` like in your comment.  Can just define in a separate `.ts` file and import it to the `.d.ts` types file to export the type
2. Keep the type def as-is and change the line here to: `const engineTypes: StatsEngine[] = ["bayesian", "frequentist"]`.  I know it's not 100% the same, but we're not going to be adding new stats engines very often (if at all), so I'm not too worried about maintenance here.

packages/back-end/src/models/ProjectModel.ts
jdorn: The controller methods should be doing checks already for these actions.  I would copy those

packages/back-end/src/models/ProjectModel.ts
romain-growthbook: I got rid of it. I agree that it's not super intuitive but I'm not sure we can do better. `/types` is purposedly setup to prevent import of `.ts` files (`rootDir` is set to `/src`) which is good and sub-typing to `StatsEngine[]` breaks the assumptions needed for zod to make it a tuple in its validator.

romain-growthbook: My bad. Locating the definitions under `./src` made it work.

packages/back-end/src/models/ProjectModel.ts
romain-growthbook: Done

packages/back-end/src/models/ProjectModel.ts
romain-growthbook: Done!

PR #2482, opened 2024-05-06
WIP: Migrates readData to permissionsClass

mknowlton89:
## Note:
This is a work-in-progress. I've checked in some work thus far that models how we might handle migrating `readData` to the permissionsClass. Currently, I've modeled how it could work for `metrics` with a `canReadMetric` helper method.

This approach provides flexibility should we want orgs to be able to control read access level by resources, not just by project. This approach also lends itself to creating policies for each resources similar to `readMetricsPolicy` and `readWriteMetricsPolicy`.

Going this approach will allow us to get rid of the `readAccessFilter` and `hasReadAccess` filter, and further consolidate the permissioning logic.

Alternatively, if we don't want to support resource-level read access filtering, I can create 2 helper methods (canReadGlobalData and canReadProjectLevelData) - the former can be used to check read access level for any global resource, and the latter can be used to check read access level for any project-scoped resource.

This would reduce the amount of code written within the `permissionClass` object (as we could write 2 methods as oppposed to ~10), but it would still require updating anywhere we're currently filtering/checking read access permission with `hasReadAccess` and the user's `readAccessFilter`.

Additionally, if we introduce helper methods on the `Permissions` class, we can remove the `getApiKeyReadAccessFilter` method since we're already building the `userPermissions` for non-users (aka, those using secret api keys) with the rest API.

### Features and Changes

### DIFF
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -548,11 +548,6 @@ export async function getDataSourceQueries(
     throw new Error("Could not find datasource");
   }
 
-  req.checkPermissions(
-    "readData",
-    datasourceObj?.projects?.length ? datasourceObj.projects : []
-  );
-
   const queries = await getQueriesByDatasource(context.org.id, id);
 
   res.status(200).json({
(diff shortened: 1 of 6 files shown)


mknowlton89:
@jdorn Scaffolded out the approach I was thinking for migrating `readData` - would love your thoughts before I go further

PR #2484, opened 2024-05-07
Migrates manageEnvironments to permission Class.

mknowlton89:
### Features and Changes

This PR migrates the `manageEnvironments` permission to the `Permissions` class and introduces 2 new helpers: `canCreateOrUpdateEnvironment` and `canDeleteEnvironment`.

I grouped `create` and `update` together, given that most of this is happening in the `putOrganization` method, and not in the `environments.controller.ts` file. I've created an issue [here](https://github.com/growthbook/growthbook/issues/2494) that handles removing the ability to update environments via the `putOrganization` method.

That issue has details on how to separate the `canCreateOrUpdateEnvironment` into two separate helpers.

Doing that work separately to reduce the size and risk of this PR.

In addition to that, I have also introduces 3 additional helpers: `canCreateSDKConnection`, `canUpdateSDKConnection`, and `canDeleteSDKConnection` - currently the permission to do those is still lumped under the `manageEnvironments` permission, but it feels different enough to warrant it's own permission (eventually). For now, I've created separate helpers, but the use the same permission under the hood.

### Testing

- [x] Ensure that user with global `manageEnvironments` access can add/update/remove all SDK Connections
- [x] Ensure that user with global `noaccess` can't create SDKConnections unless they're in a project in which they do have `manageEnvironments` permission - aka an `engineer` or `experimenter` role.
- [x] Ensure that a user with the above credentials can update, duplicate, and delete an SDK connection if it's in a project they have `manageEnvironments` permission for
- [x] Ensure the logic above works the same for legacy SDK Endpoints
- [x] Ensure that we block creating, updating, and deleting for users that have permission based on their project/role, but the SDK Connection is in an environment that they don't have these permission for


packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -1265,13 +1274,18 @@ export async function putOrganization(
...
+
+        envsWithModifiedProjects.forEach((env) => {
+          context.permissions.canCreateOrUpdateEnvironment(env);
+        });
       } else if (k === "sdkInstructionsViewed" || k === "visualEditorEnabled") {
-        req.checkPermissions("manageEnvironments", "", []);
+        context.permissions.canCreateSDKConnection({
+          projects: [],
+          environment: "",
+        });
```
jdorn, line 1288: These 3 `canCreate...` calls are missing a throw

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -217,9 +217,13 @@ const navlinks: SidebarLinkProps[] = [
...
         href: "/importing",
         path: /^importing/,
-        filter: ({ permissions, permissionsUtils, gb }) =>
+        filter: ({ permissionsUtils, gb }) =>
           permissionsUtils.canViewFeatureModal() &&
-          permissions.check("manageEnvironments" as GlobalPermission) &&
+          // This is a little weird, since this is only checking a user's global role
+          permissionsUtils.canCreateOrUpdateEnvironment({
+            projects: [],
+            id: "FOO",
```
jdorn, line 224: Can we use an empty string here or would that not work?  I don't love have "FOO" in our code base for this.  Only checking global role makes sense here since this is the importer, which will create new projects

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -649,6 +651,62 @@ export class Permissions {
     );
   };
 
+  //TODO: Refactor this into two separate methods and eliminate updating envs from organizations.controller.putOrganization - Github Issue #2494
+  public canCreateOrUpdateEnvironment = (
+    environment: Pick<Environment, "projects" | "id">
+  ): boolean => {
```
jdorn, line 657: Is this properly handling the case when the list of projects changes during an update?  If that's being handled upstream, that's fine for now.  Eventually, we'll want to move that logic here.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -733,6 +791,48 @@ export class Permissions {
...
+      if (
+        !this.checkEnvFilterPermission(
+          existing,
+          updates.environment ? [updates.environment] : [],
+          permission
+        )
+      ) {
+        return false;
+      }
+    }
```
jdorn, line 831: I think this is missing the combo case when both environment and projects are changing.  We're not checking new environment / new projects.  Also, you could get into a weird situation where you have permission for both `existing` and the end result, but not one of these in-between states.  For example, going from `Project A / dev` to `Project B / prod`.  We shouldn't care whether or not you have permission for `Project A / prod` or `Project B / dev`.

Instead of 4 if blocks, I think we can simplify to just having 2.  First test permissions on `existing`.  Then, check permissions on `{...existing, ...updates}`.

packages/front-end/components/Layout/Layout.tsx
mknowlton89: We can - I've updated this here and within `environments.tsx`.

packages/shared/src/permissions/permissionsClass.ts
mknowlton89: It does - before we're showing the edit button, we check if the user can update the existing env, and then when we make the call to update, we're only passing in the updated `env`, and so if the projects change, it'll get caught here.

packages/shared/src/permissions/permissionsClass.ts
mknowlton89: Just updated!

jdorn (approved):

PR #2486, opened 2024-05-07
Migrates runExperiments to permissionClass

mknowlton89:
### Features and Changes

This PR migrates the `runExperiments` permission to the Permissions Class and adds a new helper method `canRunExperiments`.

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -833,17 +833,19 @@ export async function postExperiment(
   }
 
   // Only some fields affect production SDK payloads
-  const needsRunExperimentsPermission = ([
-    "phases",
-    "variations",
-    "project",
-    "name",
-    "trackingKey",
-    "archived",
-    "status",
-    "releasedVariationId",
-    "excludeFromPayload",
-  ] as (keyof ExperimentInterfaceStringDates)[]).some((key) => key in changes);
+  const needsRunExperimentsPermission = (
+    [
+      "phases",
+      "variations",
+      "project",
+      "name",
+      "trackingKey",
+      "archived",
+      "status",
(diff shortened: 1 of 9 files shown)


packages/back-end/src/controllers/experiments.ts
```diff
@@ -833,17 +833,19 @@ export async function postExperiment(
...
+      "variations",
+      "project",
+      "name",
+      "trackingKey",
+      "archived",
+      "status",
+      "releasedVariationId",
+      "excludeFromPayload",
+    ] as (keyof ExperimentInterfaceStringDates)[]
+  ).some((key) => key in changes);
```
mknowlton89, line 848: prettier change only

jdorn (approved):

packages/front-end/pages/experiment/[eid].tsx
```diff
@@ -84,7 +82,7 @@ const ExperimentPage = (): ReactElement => {
   let canRunExperiment = !experiment.archived;
   const envs = getAffectedEnvsForExperiment({ experiment });
   if (envs.length > 0) {
-    if (!permissions.check("runExperiments", experiment.project, envs)) {
+    if (!permissionsUtil.canRunExperiment(experiment, envs)) {
       canRunExperiment = false;
     }
   }
```
Comment (mknowlton89) [inline packages/front-end/pages/experiment/[eid].tsx:88]: @jdorn Just wanted to double check this logic. We're only checking for the `canRunExperiment` permission if the experiment has environments that are affected.

In my testing, I had an experiment that wasn't linked to a feature that an `analyst` was able to start.

Is that expected? Or should we update this logic to remove the `if(envs.length > 0)` block?

Comment (jdorn) [reply packages/front-end/pages/experiment/[eid].tsx:88]: yes, that is expected

PR #2488, opened 2024-05-07
Migrates readData to permissionClass

mknowlton89:
### Features and Changes

This PR migrates the `readData` permission to the permissionsClass, and in doing so, removes the need to build the user's `readAccessFilter` and also allows us to move the logic of `permissions.utils.ts.hasPermission` to the `Permissions` class.

I initially tried to re-use the existing helpers, but some of the logic for `readAccess` is materially different from other permissions, even other `READ_ACCESS_PERMISSIONS`, in that the logic is different based on whether or not the resource has a `project` property (string) or a `projects` property (array of strings).

As a result, I've introduced two helpers: `canReadSingleProjectResource` and `canReadMultiProjectResource` as that allows us to simplify the logic.

Some of the differences that necessitated this change are:

For resources that have a `project` property, if that is empty, it's in `no projects`, and in order for the user to have read access, their global role must give them permission.

For resources that have a `projects` array property, if that array is empty, it is in `All Projects` and everyone should have access. This makes it so user's with a global `no access` role can see a metric if it's in `All Projects`.

Additionally, `readData` has some additional logic where if a resource is in multiple projects (e.g. a metric is in project A, B, and C) if the user has explicit roles for all projects of a resource and those explicit roles all take away read access) then even if the user has a global role that provides read access, they do not receive read access.

E.G if a user has a global role of `engineer` and for projects A, B, and C, they have  a `noaccess` role, they would not be able to view a metric that is in project A, B, and C.

### Testing

- [x] See test suite


packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -643,6 +643,48 @@ export class Permissions {
     );
   }
 
+  public canReadData = (projects: string | string[] | undefined): boolean => {
```
mknowlton89, line 646: I think I've found an issue with this.

With a lot of our resources, the `projects` array property is optional, so in the event it's undefined,  the first code block in this function won't evaluate as expected, unless when calling `canReadData` manually add an empty array E.G. `canReadData(datasource.projects || [])`. Otherwise, a user with a global role of `noaccess` won't be able to see the data sources in `All Projects`.

mknowlton89: It feels like we almost need two different `canReadData` helpers - 1 for resources with a `project` property, and one for those with a `projects` property. Not sure how we'd name those right now.

Or, we could update this signature to accept a resource type.

E.G. 
`canReadData = (projects: string | string[] | undefined, resourceType: "datasource" | "experiment" | "metric" ....)`

Then the first block can be:
```ts
if (RESOURCES_WITH_PROJECTS.includes(resourceType) && (!projects.length || projects === undefined)) {
// return true
}
```

A third option, is have a `canReadRESOURCE` for each resource.

jdorn: I would go with `canReadSingleProjectResource(project: string | undefined)` and `canReadMultiProjectResource(projects: string[] | undefined)`

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -643,6 +643,96 @@ export class Permissions {
...
+
+  public canReadMultiProjectResource = (
+    projects: string[] | undefined
+  ): boolean => {
+    // If the resource is available to all projects (an empty array or no projects property), then everyone should have read access
+    if (!projects || !projects.length || this.superAdmin) {
+      return true;
+    }
+
+    const hasGlobalReadAccess =
```
jdorn, line 660: Can we simplify all of the following logic to this one-liner?

```ts
return projects.some(p=>this.hasPermission("readData", project))
```

jdorn (approved):

packages/shared/src/permissions/permissionsClass.ts
mknowlton89: It can - I didn't think that would handle the logic of the last `readData` test suite test - `User with global readonly role should not be able to see metrics if the user has noaccess permission for every one of the metrics projects` but after testing, the code above does handle it correctly.

PR #2490, opened 2024-05-08
Adds specific helper methods for Fact Metric permissions

mknowlton89:
### Features and Changes

This PR introduces 3 new helper methods that allow us to differentiate permissions for managing Fact Metrics. Previously, we were grouping Fact Metrics and Non Fact Metrics together under the `canXYZMetric` methods. 

Some orgs have mentioned they don't want users to be able to use Fact Metrics, so by breaking it out, we're able to give them that flexibility.

This is the first permission that updates the `BaseModel` class. As a result, we do have to deviate from our existing `canUpdateXYZ` pattern given that `Fact Metrics`, AFAIK, are the first resource were the `projects` array isn't optional. This means that we have to pass in a `projects` array as the second argument to the `canUpdateFactMetrics`. This is unlike others where the `project` or `projects` property is optional. In those cases, we assume the `project` or `projects` isn't changing, so we don't check the permission again. Now, I'm checking if the `updates` object includes any updates to the `projects` array, and if so, passing it along, otherwise, passing in the `existing.projects` array.

This results in a second `hasPermission` check, but I think thats ok.

Open to feedback on that approach.

### Testing

- [x] Ensure that only admins, experimenters, and analysts can create Fact Metrics

### Note
- This issue is still present at the time of writing this where if a datasource that a Fact Table is built upon has more projects than the Fact Table, and the user doesn't have `createMetrics` permission on all of the data source's project, a user can get a "You don't have permission" message, even though it looks like they should based on the Fact Table's projects.

### DIFF
--- a/packages/back-end/src/models/FactMetricModel.ts
+++ b/packages/back-end/src/models/FactMetricModel.ts
@@ -37,9 +37,7 @@ export class FactMetricModel extends BaseClass {
     existing: FactMetricInterface,
     updates: UpdateProps<FactMetricInterface>
   ): boolean {
-    return this.context.permissions.canUpdateFactMetric(existing, {
-      projects: updates.projects || existing.projects,
-    });
+    return this.context.permissions.canUpdateFactMetric(existing, updates);
   }
   protected canDelete(doc: FactMetricInterface): boolean {
     return this.context.permissions.canDeleteFactMetric(doc);
--- a/packages/back-end/src/models/FactMetricModel.ts
+++ b/packages/back-end/src/models/FactMetricModel.ts
@@ -38,8 +38,7 @@ export class FactMetricModel extends BaseClass {
     updates: UpdateProps<FactMetricInterface>
   ): boolean {
     return this.context.permissions.canUpdateFactMetric(existing, {
-      ...existing,
-      ...updates,
+      projects: updates.projects || existing.projects,
     });
   }
   protected canDelete(doc: FactMetricInterface): boolean {
--- a/packages/back-end/src/models/FactMetricModel.ts
+++ b/packages/back-end/src/models/FactMetricModel.ts
@@ -31,16 +31,19 @@ export class FactMetricModel extends BaseClass {
     return this.context.hasPermission("readData", doc.projects || []);
   }
   protected canCreate(doc: FactMetricInterface): boolean {
-    return this.context.permissions.canCreateMetric(doc);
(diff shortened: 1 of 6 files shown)


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -31,16 +31,18 @@ export class FactMetricModel extends BaseClass {
...
-    return this.context.permissions.canCreateMetric(doc);
+    return this.context.permissions.canCreateFactMetric(doc);
   }
   protected canUpdate(
     existing: FactMetricInterface,
     updates: UpdateProps<FactMetricInterface>
   ): boolean {
-    return this.context.permissions.canUpdateMetric(existing, updates);
+    return this.context.permissions.canUpdateFactMetric(existing, {
+      projects: updates.projects || existing.projects,
```
jdorn, line 41: Is this needed?  Doesn't the update check already do this logic?

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -216,9 +216,11 @@ export default function FactMetricPage() {
   }
 
   const canEdit =
-    permissionsUtil.canUpdateMetric(factMetric, {}) && !factMetric.managedBy;
+    permissionsUtil.canUpdateFactMetric(factMetric, {
+      projects: factMetric.projects,
```
Comment (jdorn) [inline packages/front-end/pages/fact-metrics/[fmid].tsx:220]: This also doesn't seem to be needed

jdorn (approved):

packages/back-end/src/models/FactMetricModel.ts
mknowlton89: Are you referring to defining the `projects` property for the second argument?

This was an attempt to be explicit since the `updates` in `canUpdate` might not contain the `projects` array property, but `canUpdateFactMetric` is expecting (currently) an object with the shape of Pick<FactMetricInterface, "projects">. And since `projects` is a required property on the `FactMetricInterface`, this made typescript happy.

I suppose, I could update the signature of `canUpdateFactMetric` to be something like:

```ts
public canUpdateFactMetric = (
    existing: Pick<FactMetricInterface, "projects">,
    updates: Partial<FactMetricInterface>
  ): boolean => {
    return this.checkProjectFilterUpdatePermission(
      existing,
      updates,
      "createMetrics"
    );
  };
```
And then I can just do: 
```ts
return this.context.permissions.canUpdateFactMetric(existing, updates);
```

PR #2492, opened 2024-05-08
Deprecate legacy Slack integrations

jdorn:
### Features and Changes

With recent improvements to Event Webhooks, there is no longer a need to maintain a separate Slack Integration.  Event Webhooks support all the same functionality in a more flexible way that is easier to set up.

This PR removes Slack from the left nav, but still provides a link on the Webhooks settings page if someone has an existing Slack integration set up.

TODO: Decide how far we want to take this deprecation:
1. Add deprecation warning on the Slack page pointing people to Event Webhooks as a replacement
2. Make page less prominent (remove from left nav, add link from other settings page instead)
3. Block creating new Slack integrations or editing existing ones (still allow deleting)
4. Migrate Slack integrations to Event Webhooks automatically and remove the old page completely


jdorn:
@greptileai

PR #2493, opened 2024-05-08
Added SQL Server request timeout settings

Auz:
SQL Server connections have a default timeout of 15 seconds. This has to be set when connecting. This PR adds a request timeout field that is user adjustable, and also increases the default time to 120 seconds. 

Screenshot 2024-05-08 at 1 22 33 AM

### DIFF
--- a/packages/back-end/src/integrations/Mssql.ts
+++ b/packages/back-end/src/integrations/Mssql.ts
@@ -28,6 +28,7 @@ export default class Mssql extends SqlIntegration {
       user: this.params.user,
       password: this.params.password,
       database: this.params.database,
+      requestTimeout: (this.params.requestTimeout ?? 0) * 1000,
       options: this.params.options,
     });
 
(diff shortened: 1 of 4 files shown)


packages/back-end/src/integrations/Mssql.ts
```diff
@@ -28,6 +28,7 @@ export default class Mssql extends SqlIntegration {
       user: this.params.user,
       password: this.params.password,
       database: this.params.database,
+      requestTimeout: (this.params.requestTimeout ?? 0) * 1000,
```
lukesonnet, line 31: Do you also need to set `connectionTimeout` to the same value, or let customers customize both?

Auz: I thought about this—I have no problem adding, but it seems reasonable for the client to connect to the database within 15 seconds but unreasonable for all queries to complete within 15 seconds. I have no problem adding it if you think it's important.

jdorn (approved):

PR #2495, opened 2024-05-08
Add policies layer in between roles and permissions

jdorn:
### Features and Changes

Add a "policies" layer in between roles and permissions.

TODO:
- [x] Default Role -> Policy -> Permission mappings
- [x] Use this new mapping in the existing permissions code
- [x] Support for custom organization roles (data model only, no API support or UI yet)


packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -2,8 +2,266 @@ import {
...
+  DataSourcesFullAccess: [
+    "readData",
+    "createDatasources",
+    "editDatasourceSettings",
+  ],
+  DataSourceConfiguration: ["readData", "editDatasourceSettings"],
+  RunQueries: ["readData", "runQueries"],
+  MetricsFullAccess: ["readData", "createMetrics"],
+  // TODO: Fix fact permissions
+  FactTablesFullAccess: ["readData", "manageFactTables", "createMetrics"],
```
mknowlton89, line 79: @jdorn For this - it looks like we'll need to create new permissions to handle letting users manage Fact Metrics and Fact Filters. Both of those currently use `createMetrics` under the hood.

Given that, what are your thoughts on adding the following permissions:
- createFactMetrics, updateFactMetrics, deleteFactMetrics
- createFactFilters, updateFactFilters, deleteFactFilters

Note: In doing so, we can update then update their associated Permissions Class method.

packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -2,8 +2,266 @@ import {
...
+  FactMetricsFiltersFullAccess: ["readData", "createMetrics"],
+  DimensionsFullAccess: ["readData", "createDimensions"],
+  SegmentsFullAccess: ["readData", "createSegments"],
+  IdeasFullAccess: ["readData", "createIdeas"],
+  PresentationsFullAccess: ["readData", "createPresentations"],
+  SDKPayloadPublish: ["readData", "publishFeatures", "runExperiments"],
+  // TODO: add permissions
+  SDKConnectionsFullAccess: ["readData"],
```
mknowlton89, line 87: Seems like we'd want to add the following env_scoped permissions: `createSDKConnections`, `updateSDKConnections`, and `deleteSDKConnections`.

mknowlton89: We have also introduced methods to split `SDKWebhooks` and `EventWebhooks`. 

What do you think about adding a new permission `manageSDKWebhooks` and then adding that to the `SDKConnectionsFullAccess` array?

If we do it, we can then always update the existing `manageWebhooks` to be `manageEventWebooks` to be more contextual

mknowlton89: The only complexity with breaking `manageWebhooks` into `manageSDKWebhooks` and `manageEventWebhooks` is that currently `manageWebhooks` is a `GLOBAL_PERMISSION` currently only given to admins. It really feels like `manageSDKWebhooks` should be a `PROJECT_SCOPED_PERMISSION`.

We could make build it in a way that wouldn't affect existing users - we'd just make sure that only admins get the permission, but we would need to refactor the `can{action}SDKWebhook` methods.

packages/shared/src/permissions/permissions.utils.ts
mknowlton89: E.G. We would then have FactTablesFullAccess, FactMetricsFullAccess, and FactFiltersFullAccess.

```ts
FactTablesFullAccess: ["readData", "manageFactTables", "manageFactMetrics", "manageFactFilters"],
FactMetricsFullAccess: ["readData", "manageFactMetrics"],
FactMetricsFullAccess: ["readData", "manageFactFilters"],
```

packages/shared/src/permissions/permissions.utils.ts
mknowlton89: I've got this work queued up in a PR [here](https://github.com/growthbook/growthbook/pull/2503).

packages/shared/src/permissions/permissions.utils.ts
jdorn: I think I want to keep metrics and filters together for now.  There are some upcoming projects that are going to blur the line between them

packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -0,0 +1,523 @@
...
+  DataSourceConfiguration: ["readData", "editDatasourceSettings"],
+  RunQueries: ["readData", "runQueries"],
+  MetricsFullAccess: ["readData", "createMetrics"],
+  FactTablesFullAccess: [
+    "readData",
+    "manageFactTables",
+    "manageFactMetrics",
+    "manageFactFilters",
+  ],
+  FactMetricsFullAccess: ["readData", "manageFactMetrics", "manageFactFilters"],
```
mknowlton89, line 83: @jdorn Added this and the `FactFiltersFullAccess` below. Wanted to call it out this is new.

packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -0,0 +1,523 @@
...
+  FactFiltersFullAccess: ["readData", "manageFactFilters"],
+  DimensionsFullAccess: ["readData", "createDimensions"],
+  SegmentsFullAccess: ["readData", "createSegments"],
+  IdeasFullAccess: ["readData", "createIdeas"],
+  PresentationsFullAccess: ["readData", "createPresentations"],
+  SDKPayloadPublish: ["readData", "publishFeatures", "runExperiments"],
+  SDKConnectionsFullAccess: [
```
mknowlton89, line 90: @jdorn Also calling out this change for `SDKConnectionsFullAccess`

packages/shared/src/permissions/permissions.constants.ts
jdorn: I would keep fact metrics and filters together in a single policy and not break them out.  They are very intertwined in the UI, so having them separate would be kind of weird.  We can always do it later if people ask for it.

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -434,3 +440,106 @@ export async function updateMember(
...
+    member.role === role ||
+    (member.projectRoles || []).some((pr) => pr.role === role)
+  );
+}
+
+export async function removeCustomRole(
+  org: OrganizationInterface,
+  teams: TeamInterface[],
+  id: string
+) {
```
mknowlton89, line 517: I think we'll want to add one more check - to prevent deleting a role that is the org's default.
```ts
  // Make sure the id isn't the org's default
  if (org.settings?.defaultRole?.role === id) {
    throw new Error(
      "Cannot delete role. This role is set as the organization's default role."
    );
  }
  ```

mknowlton89:
I've been testing this alongside the UI PR [here](https://github.com/growthbook/growthbook/pull/2512) and it's working great.

In the PR above I made two small tweaks.
- I added some logic within the `getOrganization` to ensure we're passing the org's `customRoles` to the frontend context so when we call `getDefaultRole` - the org object we pass in contains the org's customRoles.
- I also added some logic to the deleteCustomRole endpoint so if the role was being used as the org's default, we block deletion.

jdorn:
I can't approve since I made the PR, but this looks good to me.

mknowlton89 (approved):

Issue #318, opened 2022-04-10
Fix language for experiments in the future


Auz:
When there are not enough results, we tell the user that the experiment just started, so come back later. When the experiment dates are set to a future time, this language doesn't fit very well. We should adjust the language to take this future state into account when figuring out the message. 

<img width="875" alt="CleanShot 2022-04-10 at 21 23 22@2x" src="https://user-images.githubusercontent.com/46107/162642125-362a52c0-7949-418c-80d6-24e8361579fd.png">

It should say something like "This experiment hasn't started and has no data yet. Update the results in/at N minutes/days/date"

sathithyayogi:
shall i take this issue ?

Aashish-Upadhyay-101:
@Auz is this open to work? If yes I would like to work on this !

ptdatta:
Hey @Auz Can I give it a try?

PR #2498, opened 2024-05-08
Update copy on experiment without results

Kevin-Chant:
### Features and Changes

Change text when experiment hasn't started yet to fix awkward copy

Also added a new helper to the shared dates utils to grab the relative date compared to now, as `ago` doesn't work well for future dates.

- Closes https://github.com/growthbook/growthbook/issues/318

### Dependencies

None

### Testing

* Open a sample experiment and edit the main phase to start at a future date
image

* Also verify that using a timestamp in the last 2 hours still triggers the existing copy

* And check that setting the phase to start > 2 hours ago allows results to show again

### Screenshots

Existing copy still applies:
image

Examples of copy with future dates:
image
image
image
image

Experiment results are not hidden when the phase properly starts in the past:
image

### DIFF
--- a/packages/front-end/components/Experiment/Results.tsx
+++ b/packages/front-end/components/Experiment/Results.tsx
@@ -2,7 +2,7 @@ import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
 import React, { FC, useEffect } from "react";
 import dynamic from "next/dynamic";
 import { DifferenceType, StatsEngine } from "back-end/types/stats";
-import { getValidDate, ago } from "shared/dates";
+import { getValidDate, ago, relativeDate } from "shared/dates";
 import { DEFAULT_STATS_ENGINE } from "shared/constants";
 import { ExperimentMetricInterface } from "shared/experiments";
 import { ExperimentSnapshotInterface } from "@back-end/types/experiment-snapshot";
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Experiment/Results.tsx
```diff
@@ -1,24 +1,24 @@
+import { ExperimentSnapshotInterface } from "@back-end/types/experiment-snapshot";
 import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
-import React, { FC, useEffect } from "react";
-import dynamic from "next/dynamic";
 import { DifferenceType, StatsEngine } from "back-end/types/stats";
-import { getValidDate, ago } from "shared/dates";
+import dynamic from "next/dynamic";
+import { FC, useEffect } from "react";
 import { DEFAULT_STATS_ENGINE } from "shared/constants";
+import { ago, getValidDate, relativeDate } from "shared/dates";
```
Kevin-Chant, line 7: I'm not sure if this diff is because of my linter or if it's configured in the repo. Could use some configuration help to make sure I'm not adding unnecessary diffs along with my changes

bryce-fitzsimons: yeah, I believe this will break. Specifically the `@back-end` reference (these aliases only work currently for front-end)

bryce-fitzsimons: actually looks like that may no longer be the case, nevermind :)

bryce-fitzsimons: Not sure if you're using VSCode, but wondering if this is an IDE setting causing the reordering:
https://stackoverflow.com/questions/73302972/how-to-disable-sorting-import-order-by-prettier-in-vscode

bryce-fitzsimons (changes requested):
See if you can get your IDE or linter from reordering the imports. Otherwise looks good.

packages/front-end/components/Experiment/Results.tsx
Kevin-Chant: Yep that was exactly the problem, thanks for grabbing that for me. I guess this is what I get for installing an opinionated extension and not configuring it myself :stuck_out_tongue:

bryce-fitzsimons (approved):

PR #2499, opened 2024-05-09
Permission Roles Test Suite

jdorn:
### Features and Changes

Unit test suite that tests the entire matrix of default roles / permission functions.

This is going to be critical for ensuring 100% backwards compatibility when moving to a policy-based approach in #2495 

Will need to update this as the last few permission functions are added


packages/shared/test/permissions.test.ts
```diff
@@ -130,6 +132,20 @@ describe("Role permissions", () => {
...
+    expect(p.canViewCreateSDKConnectionModal(project)).toBe(false);
+    expect(p.canCreateSDKConnection({ projects, environment })).toBe(false);
+    expect(p.canUpdateSDKConnection({ projects, environment }, updates)).toBe(
+      false
+    );
+    expect(p.canDeleteSDKConnection({ projects, environment })).toBe(false);
+    expect(p.canReadSingleProjectResource(project)).toBe(false);
+    // This is a weird one - for resources that have multiple projects, if their projects array is empty
+    // that means it's in 'All Projects' so users with global noaccess can view them
+    expect(p.canReadMultiProjectResource(projects)).toBe(true);
```
jdorn, line 148: Given your comment, I would add another one here checking when projects is `["foo"]`

jdorn: I guess we could improve the permission logic for this case to require the user to have at least one project-scoped permission that has read access.

There's not a good reason to mark someone as no access globally without overriding it in a project, so it's a little artificial

mknowlton89: @jdorn Pushed up a commit [here](https://github.com/growthbook/growthbook/pull/2499/commits/9bb52295f268189ddda9e6622441b3d9fbc6aa62) that updates the logic based on the above convo. Could probably still use some optimization.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -728,11 +728,22 @@ export class Permissions {
...
+    // If the resource doesn't have a projects property or it's an empty array
+    // that means it's in all projects
+    if (!projects || !projects.length) {
+      // If the user has read access via their global role, they should be able to read
+      if (this.hasPermission("readData", "")) {
+        return true;
+      }
+      // if the user has read access globally, or in atleast 1 project they should have read access
+      return this.hasReadAccessForAtleast1Project();
+    }
```
jdorn, line 744: I think we can simplify this a lot.  No need for a separate `hasReadAccessForAtLeast1Project` method.

```ts
    if (!projects || !projects.length) {
      // Must have read access globally or in at least 1 project
      const projectsToCheck = [
        "",
        ...Object.keys(this.userPermissions.projects),
      ];
      return projectsToCheck.some((p) => this.hasPermission("readData", p));
    }
```

mknowlton89 (approved):

PR #2502, opened 2024-05-09
Add typeguard for filtering out undefined

Kevin-Chant:
### Features and Changes

Found an easy fix for many instances of TS(2532) across the codebase while working on a separate bug. Added a new shared typeguard `isDefined` which will help methods like `Array.prototype.filter` to correctly infer the type of the result.
Searched the codebase for both instances of 2532 and `filter(Boolean)` and updated the usage manually to reduce the likelihood of unintentional bugs. For cases where it wasn't obviously an easy fix, I left the original ts-expect-error and functionality in place.

### Dependencies

None?

### Testing

There's too many files to do manual testing, so instead I'm focusing on thoroughly reading the PR & letting the test suite check for any large-scale issues

### DIFF
--- a/packages/front-end/components/Experiment/CompactResults.tsx
+++ b/packages/front-end/components/Experiment/CompactResults.tsx
@@ -15,6 +15,7 @@ import Link from "next/link";
 import { FaAngleRight, FaTimes, FaUsers } from "react-icons/fa";
 import Collapsible from "react-collapsible";
 import { ExperimentMetricInterface, getMetricLink } from "shared/experiments";
+import { isDefined } from "shared/util";
 import { useDefinitions } from "@/services/DefinitionsContext";
 import {
   applyMetricOverrides,
@@ -168,7 +169,7 @@ const CompactResults: FC<{
 
     const metricDefs = metrics
       .map((metricId) => getExperimentMetricById(metricId))
-      .filter(Boolean) as ExperimentMetricInterface[];
+      .filter(isDefined);
     const sortedFilteredMetrics = sortAndFilterMetricsByTags(
       metricDefs,
       metricFilter
@@ -176,7 +177,7 @@ const CompactResults: FC<{
 
     const guardrailDefs = guardrails
       .map((metricId) => getExperimentMetricById(metricId))
-      .filter(Boolean) as ExperimentMetricInterface[];
+      .filter(isDefined);
     const sortedFilteredGuardrails = sortAndFilterMetricsByTags(
       guardrailDefs,
       metricFilter
@@ -184,10 +185,10 @@ const CompactResults: FC<{
 
     const retMetrics = sortedFilteredMetrics
       .map((metricId) => getRow(metricId, false))
-      .filter((row) => row?.metric) as ExperimentTableRow[];
+      .filter(isDefined);
     const retGuardrails = sortedFilteredGuardrails
       .map((metricId) => getRow(metricId, true))
-      .filter((row) => row?.metric) as ExperimentTableRow[];
+      .filter(isDefined);
     return [...retMetrics, ...retGuardrails];
   }, [
     results,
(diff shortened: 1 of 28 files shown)


packages/front-end/components/Experiment/CompactResults.tsx
```diff
@@ -170,26 +171,26 @@ const CompactResults: FC<{
...
-      .filter(Boolean) as ExperimentMetricInterface[];
+      .filter(isDefined);
     const sortedFilteredGuardrails = sortAndFilterMetricsByTags(
       guardrailDefs,
       metricFilter
     );
 
     const retMetrics = sortedFilteredMetrics
       .map((metricId) => getRow(metricId, false))
-      .filter((row) => row?.metric) as ExperimentTableRow[];
```
Kevin-Chant, line 189: I checked the type definition for what `row` could be (other than undefined) and it looks like `row.metric` is always present, so the definition check is just on `row` itself, hence `isDefined` should be a drop-in replacement

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
```diff
@@ -57,12 +57,11 @@ export default function SingleRoleSelector({
         sort={false}
         formatOptionLabel={(value) => {
           const r = roles.find((r) => r.id === value.label);
+          if (!r) return <></>;
```
Kevin-Chant, line 60: Since the function returns a `ReactNode` it seemed preferable to just early-return an empty fragment instead of needing the expect-errors.

packages/front-end/components/SyntaxHighlighting/Snippets/TargetingAttributeCodeSnippet.tsx
```diff
@@ -62,8 +62,7 @@ function getExampleAttributes({
         ? ["foo", "bar"].map((v) => sha256(v, secureAttributeSalt))
         : ["foo", "bar"];
     } else if (datatype === "enum") {
-      // @ts-expect-error TS(2532) If you come across this, please fix it!: Object is possibly 'undefined'.
-      value = enumList.split(",").map((v) => v.trim())[0] ?? null;
+      value = enumList?.split(",").map((v) => v.trim())[0] ?? null;
```
Kevin-Chant, line 65: Tested in `node -i` that this still behaves appropriately when enumList is either `string` or `undefined`:

```javascript
> "foo"?.split(",").map(v => v.trim())[0] ?? null
'foo'
> undefined?.split(",").map(v => v.trim())[0] ?? null
null
```

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
jdorn: In this case, I think it's better to render just the label (`value.label`) than nothing if the lookup fails.

So we can just change the code below to output `value.label` instead of `r.id` since they are identical.  And then conditionally render the description.

packages/front-end/components/Ideas/ImpactModal.tsx
```diff
@@ -23,12 +23,9 @@ const ImpactModal: FC<{
     defaultValues: {
       metric: estimate?.metric || metrics[0]?.id || "",
       segment: estimate?.segment || "",
-      // @ts-expect-error TS(2532) If you come across this, please fix it!: Object is possibly 'undefined'.
-      userAdjustment: idea.estimateParams?.userAdjustment || 100,
-      // @ts-expect-error TS(2532) If you come across this, please fix it!: Object is possibly 'undefined'.
-      numVariations: idea.estimateParams?.numVariations || 2,
-      // @ts-expect-error TS(2532) If you come across this, please fix it!: Object is possibly 'undefined'.
-      improvement: idea.estimateParams?.improvement || 10,
+      userAdjustment: idea?.estimateParams?.userAdjustment || 100,
```
jdorn, line 26: I think instead we can just change the `idea` prop for this component to not be optional.  This component is only used 1 place and `idea` is non-null there.

jdorn (approved):

PR #2503, opened 2024-05-10
More fine-grained permissions for Fact Tables and SDK Connections

mknowlton89:
### Features and Changes

This PR, currently based on [2499](https://github.com/growthbook/growthbook/pull/2499) introduces some finer grain permissions that will eventually be used to define the policies in [2495](https://github.com/growthbook/growthbook/pull/2495).

This includes net-new permissions `manageFactMetrics`, `manageFactFilters`, `manageSDKConnections`, and also replaces `manageWebhooks` with `manageEventWebhooks` and `manageSDKWebhooks`.

This was done in a way where it shouldn't impact existing organizations at all.

### Dependencies

- This is dependent on [2499](https://github.com/growthbook/growthbook/pull/2499)

### Testing

- [x] Ensure test suite passes
- [x] Ensure an admin (and only admins) can still create event webhooks and SDK webhooks
- [x] Ensure only engineers and experimenters can create SDK Connections
- [x] Ensure only analysts and experimenters can create Fact Metrics and Fact Filters
- [ ] Ensure engineers and experimenters can create SDK Webhooks (new only) as long as their project/env permissions grant them the ability


packages/shared/src/permissions/permissions.utils.ts
```diff
@@ -40,7 +43,8 @@ export const GLOBAL_PERMISSIONS = [
   "manageTags",
   "manageApiKeys",
   "manageIntegrations",
-  "manageWebhooks",
+  "manageEventWebhooks",
+  "manageSDKWebhooks",
```
jdorn, line 47: Thinking about this more, I think we should make a breaking change and grant `manageSDKWebhooks` permission to engineers and experimenters.  I think it was a bug to not do this originally and it will simplify the policies since we can just include this permission in the `SDKConnectionsFullAccess` policy instead of making a separate SDKWebhook one.

mknowlton89: Sounds good. 

So, just to clarify, we'll keep the change that removes `manageWebhooks` in lieu of the two different `manageEventWebhooks` and `manageSDKWebhooks`.

And we'll update engineer and experimenter roles to include `manageSDKWebhooks`, but we'll keep `manageEventWebhooks` as an admin-only permission.

In these changes, I think it also makes sense to change the scoping of `manageSDKWebhooks` so it's env scoped like `manageSDKConnections`.

packages/front-end/components/Settings/Webhooks.tsx
```diff
@@ -150,7 +159,10 @@ const Webhooks: FC = () => {
           </tbody>
         </table>
       )}
-      {permissionsUtil.canCreateSDKWebhook() ? (
+      {permissionsUtil.canCreateSDKWebhook({
+        projects: [project],
+        environment: "",
+      }) ? (
```
mknowlton89, line 165: This isn't technically correct - but given this [PR](https://github.com/growthbook/growthbook/pull/2483/files#diff-9b0f37869e5ee3a773f1f808d099594cf2d7c9eea5fe7a82f4478092074d1449) will remove this option, I haven't fixed it.

jdorn (approved):

packages/back-end/src/util/permission-constants.ts
```diff
@@ -31,7 +34,8 @@ export const GLOBAL_PERMISSIONS = [
   "manageTags",
   "manageApiKeys",
   "manageIntegrations",
-  "manageWebhooks",
+  "manageEventWebhooks",
+  "manageSDKWebhooks",
```
mknowlton89, line 38: This should be moved to `ENV_SCOPED_PERMISSIONS`?

jdorn: This file in back-end is not being used. I deleted it in the other PR.

PR #2505, opened 2024-05-10
Project post/put/delete API endpoints.

romain-growthbook:
New endpoints using the migrated project models.


packages/back-end/src/api/openapi/payload-schemas/PostProjectPayload.yaml
```diff
@@ -0,0 +1,8 @@
+type: object
+required:
+  - name
+properties:
+  name:
+    type: string
+  description:
+    type: string
```
jdorn, line 8: Should we allow setting the stats engine during creation?  Seems weird to require 2 API calls to do that.

packages/back-end/src/api/projects/deleteProject.ts
```diff
@@ -0,0 +1,31 @@
...
+import { deleteProjectValidator } from "../../validators/openapi";
+import { auditDetailsDelete } from "../../services/audit";
+
+export const deleteProject = createApiRequestHandler(deleteProjectValidator)(
+  async (req): Promise<DeleteProjectResponse> => {
+    let id = req.params.id;
+    // Add `prj__` prefix if it doesn't exist
+    if (!id.startsWith("prj__")) {
+      id = `prj__${id}`;
+    }
```
jdorn, line 12: Remove this.  I'm assuming this was copied from fact metrics, but that's a very unique case where the prefix is meaningful.

packages/back-end/src/routers/project/project.controller.ts
```diff
@@ -333,7 +325,7 @@ export const putProjectSettings = async (
 
   const { settings } = req.body;
 
-  await updateProjectSettings(id, project.organization, settings);
+  await context.models.projects.updateSettingsById(id, settings);
```
jdorn, line 328: Since we have the `project` already in memory, we should not call a `byId` method, since that will require an unnecessary db lookup.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,139 +1,79 @@
...
-  },
-  organization: {
-    type: String,
-    index: true,
+const projectValidator = baseSchema
+  .extend({
+    name: z.string(),
+    description: z.string(),
+    settings: z.object({
+      statsEngine: z.enum(statsEngines),
```
jdorn, line 13: This is optional in the current interface.   We could either mark it optional here or call the `default()` method and give it a default value if missing.

Also, I noticed that `back-end/types/project.d.ts` wasn't updated in this PR.  We should update that to reference the validator instead of redefining the interface.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,139 +1,79 @@
...
+    }),
+  })
+  .strict();
+
+const BaseClass = MakeModelClass({
+  schema: projectValidator,
+  collectionName: "projects",
+  idPrefix: "prj__",
```
jdorn, line 21: This should have a single underscore.  The double underscore is unique to fact metrics and not used anywhere else in the app.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -1,139 +1,79 @@
...
-  };
-  await ProjectModel.updateOne(
-    {
-      id,
-      organization,
-    },
-    update
-  );
-}
+  protected migrate(doc: z.infer<typeof projectValidator>) {
```
jdorn, line 55: The type here isn't quite accurate and is making Typescript hide potential errors.  Some old documents don't have `settings` at all, and some that do don't have `statsEngine` defined.

packages/back-end/src/api/projects/deleteProject.ts
romain-growthbook: Well glad I hadn't included it in the tests. I though this might be superfluous..

jdorn (approved):

PR #2507, opened 2024-05-10
Bayesian power analysis

lukebrawleysmith:
This pull request extends power analysis to our Bayesian engine.  

This PR permits Bayesian power analysis by updating the front end Power Calculation code.
A customer can now specify a prior just as they would during an experiment.
Power analysis defaults to default stats engine for an org, and frequentist vs Bayesian power analysis can now be toggled under Analysis Settings:

Screenshot 2024-05-30 at 10 14 45 AM

If the power is switched from frequentist to Bayesian, power is recalculated using the the org defaults for a metric. If desired, the customer may select Edit and update Prior Mean and Prior Standard Deviation under the Metrics modal:

Screenshot 2024-05-30 at 10 09 30 AM


packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +345,543 @@ export function powerMetricWeeks(
...
+  }
+  /*case where mde converged to nonzero value*/
+  const mdeResults: MDEResults = {
+    type: "success",
+    mde: mde,
+  };
+  return mdeResults;
+}
+
+export function powerMetricWeeksBayesian(
```
lukesonnet, line 780: I see why you asked about copying. There's a lot of complex logic here that it might benefit us to only write once. Take a look at this slight rewrite https://github.com/growthbook/growthbook/compare/ls/power-structure?expand=1

I've done two key things:
* Make some of the code a bit more typescript friendly (e.g. don't unpack a bunch of stuff, use .forEach)
* Ask you to change `powerEst` and `powerEstBayesian` to instead take an object with your parameters specified. Then build those parameters and specify which function to use in one single if-else block.

My code is psuedo-code, since I didn't create those other types, and can just serve as a reference.

lukesonnet (commented):
Some smaller comments and one bigger comment.

Broad question: is the unit test here set up to mirror the python notebook and we should take the python notebook as the source of power validity, right?

lukebrawleysmith:
> Some smaller comments and one bigger comment.
> 
> Broad question: is the unit test here set up to mirror the python notebook and we should take the python notebook as the source of power validity, right?

packages/front-end/components/PowerCalculation/types.ts
```diff
@@ -13,6 +13,27 @@ export type MetricParams =
...
     };
 
+export type MetricParamsBayesian =
+  | {
+      type: "mean";
+      name: string;
+      effectSize: number;
+      mean: number;
+      standardDeviation: number;
+      priorMean: number;
```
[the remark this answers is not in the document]
lukebrawleysmith: made both changes

packages/back-end/types/stats.d.ts
```diff
@@ -6,6 +6,7 @@ export type PValueCorrection = null | "benjamini-hochberg" | "holm-bonferroni";
 
 export type DifferenceType = "relative" | "absolute" | "scaled";
 
+export type RiskType = "relative" | "absolute";
```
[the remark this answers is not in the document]
lukebrawleysmith: done

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -53,7 +55,7 @@ export function powerStandardError(
       variance,
       mean * (1 + effectSize),
       nPerVariation,
-      true
+      relative
```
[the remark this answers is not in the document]
lukebrawleysmith: confirmed that relative is always True, and its in there if we need it for later.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +345,543 @@ export function powerMetricWeeks(
   };
   return results;
 }
+
+/*******************/
+export function calculatePriorMean(
```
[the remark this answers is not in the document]
lukebrawleysmith: agreed, should be resolved now

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +345,543 @@ export function powerMetricWeeks(
...
+    priorMeanRelSpecified,
+    priorVarianceRelSpecified,
+    proper,
+    mean,
+    variance,
+    nPerVariation,
+    true
+  );
+  if (maxPower < power) {
+    console.log(`failing at iteration j: %d`, j);
```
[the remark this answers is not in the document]
lukebrawleysmith: so far, this case has never happened.  i had this for testing purposes, happy to remove it if you want.  

i haven't rigorously resolved the upper bound issue yet, so i kept this in there for now.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +345,543 @@ export function powerMetricWeeks(
...
+    priorVarianceRelDGP,
+    priorMeanRelSpecified,
+    priorVarianceRelSpecified,
+    proper,
+    mean,
+    variance,
+    nPerVariation,
+    relative
+  );
+  while (currentPower > power) {
```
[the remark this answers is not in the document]
lukebrawleysmith: added iters counter

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +345,543 @@ export function powerMetricWeeks(
...
+    nPerVariation,
+    relative
+  );
+  const properInt = proper ? 1 : 0;
+  return 1 / tauHatVariance + properInt / priorVarianceSpecified;
+}
+
+// Function to calculate upper cutpoint
+export function getCutpoint(
+  alpha: number,
```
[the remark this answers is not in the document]
lukebrawleysmith: agreed and done

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: I incorporated your pseudo-code.  However, I didn't define new types as inputs for `powerEst` and `powerEstBayesian`, as the linter was fussing at me for having different types as inputs for a `powerFunction` function.  Also, the way I have it is simple.  However, if you have more elegant approach, I'm open to incorporating it.  Also, I have `statsEngineSettings` as attribute of `powerSettings`, rather than standalone object, because we always want the freq statsEngineSettings to be part of `PowerCalculationParams` and bayesian statsEngineSettings to be associated with PowerCalculationParamsBayesian.  Again, if you have better suggestion, I'm open.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -130,39 +160,34 @@ export function sequentialPowerStandardError(
...
-  effectSize: number,
-  mean: number,
-  variance: number,
+  metric: MetricParamsFrequentist,
   n: number,
   nVariations: number,
   alpha: number = 0.05,
   twoTailed: boolean = true,
-  sequentialTuningParameter = 0
+  sequentialTesting: false | number
```
lukesonnet, line 168: In general I think having something typed as either `false | number` isn't the best approach if we can avoid it. It's a bit confusing, but no need to change it right now.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -223,85 +251,81 @@ export function findMde(
...
-        sequentialTuningParameter
-      );
+      /*initialize block-scoped variables*/
+      let thisPower = 0.0;
+      let thisMde: MDEResults = {
+        type: "success",
+        mde: -999,
+      };
+      if (powerSettings.statsEngineSettings.type === "frequentist") {
+        thisPower = powerEst(
```
lukesonnet, line 294: Should we call these methods `powerEstFrequentist` and `findMdeFrequentist` for consistency?

docs/docs/statistics/power.mdx
```diff
@@ -116,14 +124,14 @@ Frequently asked questions:
 3. What if my experiment is low-powered? Should I still run it? The biggest cost to running a low-powered experiment is that your results will probably be noisy, and you will face ambiguity in your rollout/rollback decision. That being said, you will probably still have learnings from your experiment.
 4. What does "N/A" mean for my MDE result? It means there is no solution for the MDE, given the current number of weekly users, control mean, and control variance. Increase your number of weekly users or reduce your number of treatment variations.
 
-## GrowthBook implementation
+## GrowthBook implementation (frequentist)
```
lukesonnet, line 127: At this point, this documentation is really long and complicated and I'm worried about having it be public facing when so few people will find it relevant, and the few people who are technical and are curious about our approach cannot parse this information effectively.

Can we reformat and simplify as much as possible?

Some suggestions:
* Clear formulae for each object that we care about (Power = ...; MDE = ...)
* Plain language summary of our approach?
* Minimal recreation of concepts explained elsewhere (e.g. var(\Delta))
* Less explanation of the steps to solve the math, and a bit more reliance on english to explain how we ended up with final result. We can keep more documentation in an internal notion doc.
* More sign posting and better formatting. This document is a wall of uneven line breaks and interspersed equations and plain text. Can we consider using more sub-headers, bolded concepts like "Power" with their equation right below it?

This comment is asking for a pretty big overhaul but this section is a bit messy and I think doesn't do a lot to build trust in the hard work you've done to do this right.

This applies to both Freq and Bayesian power sections.

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: I agree, I can talk to Romain about this, he suggested this input when only freq power analysis was available.  Now that we are implementing Bayesian, I can see if we can simplify.

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: done

lukebrawleysmith: not sure why there is not an option to reply to the comment below ("If it could possibly happen, then we should have proper error handling for it, if it is impossible, we should remove it."), but I have revamped the bayesian mde code, and this does not exist anymore.

lukebrawleysmith:
Adding some context:

This PR permits Bayesian power analysis by updating the front end Power Calculation code.  
A customer can now specify a prior just as they would during an experiment.  
Power analysis defaults to default stats engine for an org, and frequentist vs Bayesian power analysis can now be toggled under `Analysis Settings`:

Screenshot 2024-05-30 at 10 14 45 AM

If the power is switched from frequentist to Bayesian, power is recalculated using the the org defaults for a metric.  If desired, the customer may select `Edit` and update `Prior Mean` and `Prior Standard Deviation` under the Metrics modal:

Screenshot 2024-05-30 at 10 09 30 AM

I updated the testing files to match output from python notebook, where simulations confirmed that Bayesian power is accurate.

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
```diff
@@ -256,17 +313,31 @@ const MetricParamsInput = ({
...
               key={`${name}-${entry}`}
               entry={entry}
               form={form}
               metricId={metricId}
             />
-          )
-        )}
+          ))}
       </div>
+      {engineType === "bayesian" && (
```
lukesonnet, line 327: IMO these settings belong in a dropdown on the stats engine settings page as overrides you can set up after the fact, or they should be hidden by default with a small button to "modify metric priors" rather than presenting users with 5(!) input boxes per metric. If you modify metric priors it can open the toggle and two prior fields. You can see how we do this in the MetricForm index.tsx file where we only open the metric prior settings when you turn the override on.

docs/static/Power.pdf
lukesonnet, line 1: Delete this file?

docs/docs/statistics/powerTechnical.mdx
```diff
@@ -0,0 +1,346 @@
+---
+title: Power Analysis Technical Details
+description: Technical details of power analysis
+sidebar_label: Power Technical
+slug: powerTechnical
```
lukesonnet, line 5: Follow the naming convention of all of the other docs. This should be `power-technical.mdx` and `power-technical` should be the slug.

docs/docs/statistics/power.mdx
```diff
@@ -116,108 +124,61 @@ Frequently asked questions:
...
-$\rho = \sqrt{\frac{-2\text{log}(\alpha) + \text{log}(-2 \text{log}(\alpha) + 1)}{N^*}}$
-
-and $N^{\star}$ is a tuning parameter. This approach relies upon asymptotic normality. For power analysis we rewrite the confidence interval as
-
-$\left(\hat{\tau} \pm \hat{\sigma}*\sqrt{N} * \sqrt{\frac{2(N\rho^2 + 1)}{N^2\rho^2}\log\left(\frac{\sqrt{N\rho^2 + 1}}{\alpha}\right)}\frac{Z_{1-\alpha/2}}{Z_{1-\alpha/2}}\right)$$=\left(\hat{\tau} \pm \tilde{\sigma}Z_{1-\alpha/2}\right)$
-
-where
-
-$\tilde{\sigma} = \hat{\sigma}*\sqrt{N}\sqrt{\frac{2(N\rho^2 + 1)}{N^2\rho^2}\log\left(\frac{\sqrt{N\rho^2 + 1}}{\alpha}\right)}\frac{1}{Z_{1-\alpha/2}}$.
+Bayesian MDE
```
lukesonnet, line 182: Can we format this either bold or something? Right now it just looks like random text in the middle of the documentation.

docs/docs/statistics/power.mdx
```diff
@@ -116,108 +124,61 @@ Frequently asked questions:
...
 
 1. the false positive rate as $\alpha$ (GrowthBook default is $\alpha=0.05$).
-2. the critical values $Z_{1-\alpha/2}= \Phi^{-1}(1-\alpha/2)$ and $Z_{1-\alpha}= \Phi^{-1}(1-\alpha)$ where $\Phi$ is the inverse CDF of the standard normal distribution.
-3. the true relative treatment effect as $\tau$, its estimate as $\hat{\tau}$ and its estimated standard error as $s$. Note that as the sample size $n$ increases, s decreases by a factor of $1/\sqrt{n}$.
+2. the critical values $Z_{1- \alpha / 2}= \Phi^{-1}(1-\alpha/2)$ and $Z_{1-\alpha}= \Phi^{-1}(1-\alpha)$ where $\Phi$ is the inverse CDF of the standard normal distribution.
 
 We make the following assumptions:
 
-1. equal sample sizes across control and treatment variations;
+1. equal sample sizes across control and treatment variations. If unequal sample sizes are used in the experiment, use the smaller of the two sample sizes. This will produce conservative power estimates.
```
lukesonnet, line 138: Nice clarification, thanks.

docs/docs/statistics/power.mdx
```diff
@@ -116,108 +124,61 @@ Frequently asked questions:
 3. What if my experiment is low-powered? Should I still run it? The biggest cost to running a low-powered experiment is that your results will probably be noisy, and you will face ambiguity in your rollout/rollback decision. That being said, you will probably still have learnings from your experiment.
 4. What does "N/A" mean for my MDE result? It means there is no solution for the MDE, given the current number of weekly users, control mean, and control variance. Increase your number of weekly users or reduce your number of treatment variations.
 
-## GrowthBook implementation
+## GrowthBook implementation (frequentist)
+
+Below we describe high-level technical details of our implementation. All technical details can be found [here](/statistics/powerTechnical).
```
lukesonnet, line 129: Thanks for the updates here. I still think we need a bit more signposting and clear language.

Can we say something like

```
## GrowthBook implementation

For both Bayesian and frequentist engines, we produce two key statistics:

1. Power - After running the experiment for a given number of weeks and for a hypothesized effect size, what is the probability of a statistically significant result?
2. Minimum Detectable Effect (MDE) - After running an experiment for a given number of weeks, what is the smallest effect that we can detect as statistically significant 80% of the time.

Each engine arrives at these values differently...tech details link... etc...

### Frequentist engine

...clear signposting for **Power** and **MDE**...

### Bayesian engine

...clear signposting for **Power** and **MDE**...

**Review by lukesonnet — COMMENTED** [review]
Left some main doc comments for @lukebrawleysmith and a few important FE comments for @romain-growthbook re: priors.

Specifically:
* We need a much slimmer way for users to input priors. It's so overwhelming right now. IMO they should not change this every time they do a power calculation so we need those prior fields to either be in the StatsEngine modal or we need to greatly de-emphasize them (for example with a small text link "Override metric priors" to hide all the prior input fields until you click it to open the prior input fields).
* We need to inherit metric specific priors that override the org defaults. I've left comments about this in the fiels.

File: docs/static/Power.pdf
Comment (lukebrawleysmith) [reply docs/static/Power.pdf:1]: done

File: packages/stats/gbstats/devtools/simulation.py [hunk 17]
```diff
@@ -56,6 +56,9 @@ def create_storage_arrays(self):
         self.lower_limit = np.empty(array_shape)
         self.upper_limit = np.empty(array_shape)
         self.results = np.empty(array_shape, dtype=object)
+        self.stat_a = np.empty(array_shape, dtype=object)
+        self.stat_b = np.empty(array_shape, dtype=object)
+        self.estimand = np.empty(array_shape)
```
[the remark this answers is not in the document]
lukebrawleysmith: I meant to delete this, sorry.  It should be part of its own PR (I think it would be helpful for debugging to store these).

docs/docs/statistics/powerTechnical.mdx
lukebrawleysmith: done

docs/docs/statistics/power.mdx
lukebrawleysmith: thanks for catching typo, fixed

docs/docs/statistics/power.mdx
lukebrawleysmith: thx

docs/docs/statistics/power.mdx
lukebrawleysmith: thanks, I agree it's helpful for the key equations to really pop.  Changes implemented.

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
```diff
@@ -324,14 +409,55 @@ const SetParamsStep = ({
...
 }: Props) {
   const [step, setStep] = useState<"select" | "set-params">("select");
+  const settings = useOrgSettings();
 
   const form = useForm<PartialPowerCalculationParams>({
     defaultValues: params,
   });
 
+  const metrics = form.watch("metrics");
+  const defaultValues = Object.keys(config).reduce(
```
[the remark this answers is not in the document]
romain-growthbook: That should be the case in the current code. `defaultValues` are only added when there's no existing values already configured.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+}
+
+export function calculatePriorVariance(
+  priorVarianceRel: number,
+  mean: number,
+  relative: boolean
+): number {
+  return relative
+    ? priorVarianceRel
+    : priorVarianceRel * Math.pow(Math.abs(mean), 2);
```
lukesonnet, line 376: nit: is there a reason why we do Math.abs() first?

docs/docs/statistics/power-technical.mdx
```diff
@@ -0,0 +1,346 @@
...
+
+We define rejection as the $100(1-\alpha)$\% confidence interval not containing zero.  
+For our posterior approximation, this occurs if the posterior mean for $\Delta|\hat{\Delta}$ (i.e., $\Omega^{-1}\omega$) divided by its posterior standard deviation$\left(\text{i.e.,  }\sqrt{\Omega^{-1}}\right)$ is beyond the the appropriate critical threshold $Z^{\star}$ (e.g., $\Phi^{-1}(0.975)$ for $\alpha=0.05$).
+
+Inside of a Bayesian framework, it can help to permit the case where the prior model is misspecified.  
+That is, the prior specified by the customer differs from the true prior that generates the treatment effect.  
+We permit misspecification of the prior for $\Delta$, as we assume that the true data generating process (DGP) is $\Delta \stackrel{}{\sim}\mathcal{N}\left(\mu_{\star}, \sigma_{\star}^{2}\right)$, while the specified DGP has $\Delta \stackrel{}{\sim}\mathcal{N}\left(\mu_{prior}, \sigma_{prior}^{2}\right)$.
```
lukesonnet, line 232: While "true DGP" makes sense, I feel that calling the two DGPs the true and the specified DGPs is a bit misleading. I imagine in most customer's eyes, they're thinking of the "true DGP" they write down in the modal as the desired effect size, while the prior is their belief about the effect size in general. They want to know the answer to the question: "Given my prior is X, what is the likelihood that if the true effect is actually Y, that I can reach statistical significance?"

So in that way, maybe we need to find another term besides "specified DGP". WDYT?

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+
+function calculatePriorMeanDGP(
+  metric: MetricParams,
+  relative: boolean
+): number {
+  const metricMean = getMetricMean(metric);
+  return calculatePriorMean(metric.effectSize, metricMean, relative);
+}
+
+function calculatePriorVarianceDGP(
```
lukesonnet, line 407: Maybe we can have a comment about why this always returns zero?

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+  const tauHatVariance = estimateTauHatVariance(
+    metric,
+    nPerVariation,
+    relative
+  );
+  const properInt = metric.proper ? 1 : 0;
+  return 1 / tauHatVariance + properInt / priorVarianceSpecified;
+}
+
+// Function to calculate upper cutpoint
```
lukesonnet, line 464: Comment is out of date as upper/lower is parameterized in this fn.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+/**
+ * Performs power calculation
+ *
+ * @param metric an object of class MetricParams that has info about prior mean and sd, metric mean and sd, and effect size.
+ * @param alpha false positive rate (default: 0.05).
+ * @param nPerVariation sample size per variation.
+ * @param relative boolean indicator if relative inference is desired.
+ * @returns Estimated power.
+ */
+export function powerEstBayesian(
```
lukesonnet, line 515: This method and getCutpoint are really clear. Thanks!

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+ */
+export function findMdeBayesian(
+  metric: MetricParams,
+  alpha: number,
+  power: number,
+  nPerVariation: number,
+  relative: boolean
+): MDEResults {
+  /*fixed effect size, so prior variance of data generating process is 0*/
+  const dummyMetric = { ...metric };
```
lukesonnet, line 598: I think a safer way to do this would to do :

`const dummyMetric = { ...metric, effectSize: effectSize}` so that you're not overwriting the shallow copy but just building it with the new value.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+  );
+  /*case where 0 effect size results in at least 80% power*/
+  if (currentPower >= power) {
+    const mdeResults: MDEResults = {
+      type: "success",
+      mde: effectSize,
+    };
+    return mdeResults;
+  }
+  /*maximum effect size we check is 500%*/
```
lukesonnet, line 616: Our FE should handle this case. If the MDE we return is 0, or the MDE we return is > 500, we just need to make sure we have tooltips or something in the FE to very lightly explain why this is.

packages/front-end/components/PowerCalculation/stats.ts
```diff
@@ -340,3 +356,294 @@ export function powerMetricWeeks(
...
+  const maxEffectSize = 5;
+  const stepSizeCoarse = 1e-3;
+  const maxError = normal.pdf(0, 0, 1) * stepSizeCoarse;
+  while (currentPower < power && effectSize <= maxEffectSize) {
```
lukesonnet, line 620: Again, open while loops scare me. Can we re-write this to be more obviously finite using a different loop construct? I'd rather we build a range of effect sizes in an array (unless the array is huge) and loop over that. If the array is huge, let's instead know the number of max iterations to get through the array and stop then.

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: removed

docs/docs/statistics/power-technical.mdx
lukebrawleysmith: I agree that the customer often wants to have answered, "Given my prior is X, what is the likelihood that if the true effect is actually Y, that I can reach statistical significance?"  I used your language in my doc at a later point to emphasize this.  

I like "specified" in that specifying a prior is more all-encompassing than the term "prior belief".  Sometimes when you specify a prior, it is just your belief.  But sometimes prior specification comes from a previous study or another data driven source.  So I like "specified".  But I'm open to other terms.

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: good call, I added:
  /*priorStandardDeviationDGP is 0 because we assume true fixed effect size*/

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: removed, thanks

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: thanks!

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: done, thanks for the tip

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: we can ask the customer to increase the sample size, as we currently do when mde doesn't exist.

packages/front-end/components/PowerCalculation/stats.ts
lukebrawleysmith: good call, and updated

packages/front-end/components/PowerCalculation/stats.ts
romain-growthbook: In general I try to avoid combinations of types that are not possible to make sure we tighten the type descriptions and don't have to make assumptions about the data. Typically here, I'd like to avoid something like:
```typescript
sequentialTestingEnabled: boolean;
sequentialTestingParam: number;
```
I'm open to anything that can combine the two. Classic JS is a `nullable` number. We could also have `"disabled" | number`.
```

**Comment by romain-growthbook** [comment]
Just pushed the following changes:
* Put the bayesian settings under a toggle:
Screenshot 2024-06-06 at 2 00 05 PM

Screenshot 2024-06-06 at 2 00 11 PM
* Made sure that new calculations pick up the org default stats engine, not the stats engine from the existing computation
* Switched prior lift mean to number, not percentage. Confirmed that values for this and priori standard deviation are picked up from the org default

I think that this should address most concerns. Let me know if there are more.

File: packages/front-end/components/PowerCalculation/types.ts [hunk 27]
```diff
@@ -45,51 +53,97 @@ export type PartialPowerCalculationParams = Partial<
...
+    type: "percent",
     minValue: 0,
     maxValue: 1,
   },
+  priorLiftMean: {
+    title: "Prior mean",
+    metricType: "all",
+    type: "percent",
+    tooltip: "Prior mean for the relative effect size.",
+    defaultSettingsValue: (s) => s.metricDefaults?.priorSettings?.mean,
```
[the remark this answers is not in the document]
romain-growthbook: oh I see.

romain-growthbook:
Ok sorry I finally understood the metric override thing. Should be in the code now too.

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
lukesonnet: @romain-growthbook Can we turn the toggle just into a link text that says "Modify metric priors" and make it much smaller? Or at least change the copy to "Modify metric priors" It's still really busy IMO.

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
```diff
@@ -213,60 +264,108 @@ const InputField = ({
...
+  engineType: "bayesian" | "frequentist";
 }) => {
   const metrics = form.watch("metrics");
   // eslint-disable-next-line
   const { name, type: _type, ...params } = ensureAndReturn(metrics[metricId]);
+  const [showBayesian, setShowBayesian] = useState(false);
 
   return (
     <div className="card gsbox mb-3 p-3 mb-2 power-analysis-params">
       <div className="card-title uppercase-title mb-3">{name}</div>
```
lukesonnet, line 327: Can we make the name a link to the metric?

packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
```diff
@@ -66,14 +66,15 @@ export default function PowerCalculationStatsEngineModal({
...
+                <div className="container">
+                  <div className="row">
+                    <span className="text-muted mr-1">Bayesian</span>
+                    {orgSettings.statsEngine === "bayesian"
+                      ? "(Org default)"
+                      : ""}
+                  </div>
+                </div>
               ),
-              disabled: true,
```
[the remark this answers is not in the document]
lukesonnet: @romain-growthbook thoughts on my second comment here?

packages/front-end/components/PowerCalculation/PowerCalculationContent.tsx
```diff
@@ -75,9 +85,16 @@ const AnalysisSettings = ({
           <div className="col-7">
             <h2>Analysis Settings</h2>
             <p>
-              {params.nVariations} Variations · Frequentist (Sequential Testing{" "}
-              {params.statsEngine.sequentialTesting ? "enabled" : "disabled"}) ·{" "}
-              <Link href="#" onClick={() => setShowStatsEngineModal(true)}>
+              {params.nVariations} Variations ·{" "}
+              {engineType[params.statsEngineSettings.type]} (Sequential Testing{" "}
+              {params.statsEngineSettings.sequentialTesting
```
lukesonnet, line 90: @romain-growthbook we should only show `(Sequential Testing ...)` for Frequentist engine, regardless of whether it is enabled or disabled for now.

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
romain-growthbook: Inside the modal? That seems like a weird user experience to go from a modal to a new page..

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
romain-growthbook: Text updated. I'm not sure exactly how to have a good UX that's smaller. Text only, I don't see how you materialize open/close state. Maybe something to ask Kari?

packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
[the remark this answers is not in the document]
romain-growthbook: I'm down to do that. Do we have the doc page ready already?

packages/front-end/components/PowerCalculation/PowerCalculationContent.tsx
lukebrawleysmith: @romain-growthbook , in the screenshot you shared, `Effect size` follows `Conversion rate` for the top metric, but precedes `Mean` for the bottom metric.   Can you please ensure `Effect size` always comes first?

packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal.tsx
lukebrawleysmith: I like the toggle.  Nit: can it say "modify metric prior"?

packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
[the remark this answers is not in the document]
lukebrawleysmith: yes Romain, the doc page is ready

lukebrawleysmith:
Nit: In `Calculated Sample Size & Runtime`, can we right align `Effect Size` numbers?
Screenshot 2024-06-07 at 11 43 38 AM

lukebrawleysmith:
Can we make number of digits rounded consistent across the power table (currently 2) and MDE table (currently 3)?  My personal preference is two.  Can we also have these percentages right aligned?  See first column of power table.  
Screenshot 2024-06-07 at 11 46 29 AM

lukebrawleysmith:
In `New Calculation`, after I pick 5 metrics, when I click the `x` to remove a metric, it is not removed.  
Screenshot 2024-06-07 at 12 12 53 PM

lukebrawleysmith:
Can we change default to be empty field for Mean and Standard Deviation?  NaN looks strange.  
Screenshot 2024-06-07 at 12 13 16 PM

lukebrawleysmith:
I'm not sure how easy this is to fix, but seeing "Week 8" with wrapped text and "Week 9" on same line looks strange.  
Screenshot 2024-06-07 at 12 16 34 PM

lukebrawleysmith:
Can we remove this tool tip under `Choose Statistical Engine`?  We will enable Bayesian power analysis.  
Screenshot 2024-06-07 at 12 20 47 PM

lukebrawleysmith:
Can we have `Use proper prior` precede inputs for `Prior mean` and `Prior standard deviation`?  `Prior mean` and `Prior standard deviation` only matter if prior is proper, so customer should toggle that first.  
Screenshot 2024-06-07 at 12 23 56 PM

romain-growthbook:
> Nit: In `Calculated Sample Size & Runtime`, can we right align `Effect Size` numbers? Also, can we right align power and MDE numbers in the tables? Screenshot 2024-06-07 at 12 30 03 PM

I'd like to defer to unblock this. I'd like to take the time to make a app-wide determination so that we can start right aligning numbers on all tables through the app.

romain-growthbook:
> In `New Calculation`, after I pick 5 metrics, when I click the `x` to remove a metric, it is not removed. Screenshot 2024-06-07 at 12 12 53 PM

Fixed, great catch thanks. Options get disabled past 5 choices. The design is not great but at the the functionality is as expected.

romain-growthbook:
> Can we change default to be empty field for Mean and Standard Deviation? NaN looks strange. Screenshot 2024-06-07 at 12 13 16 PM

Fixed.

romain-growthbook:
> I'm not sure how easy this is to fix, but seeing "Week 8" with wrapped text and "Week 9" on same line looks strange. Screenshot 2024-06-07 at 12 16 34 PM
> 
> To reproduce, here are metrics along with (Effect Size, Mean, Standard Deviation), or (Conversion Rate, Effect Size) PURCHASES - TOTAL REVENUE (72 HOUR WINDOW) (1, 1, 2) RETENTION - [1, 14) DAYS (12, 1) Revenue (1,1,3) Latency (1, 12, 4) Latency_Mean (1,2,1)

Should be fixed. I put an unbreakable space between week and its number.

romain-growthbook:
> Can we remove this tool tip under `Choose Statistical Engine`? We will enable Bayesian power analysis. Screenshot 2024-06-07 at 12 20 47 PM

Fixed.

romain-growthbook:
> Can we have `Use proper prior` precede inputs for `Prior mean` and `Prior standard deviation`? `Prior mean` and `Prior standard deviation` only matter if prior is proper, so customer should toggle that first. Screenshot 2024-06-07 at 12 23 56 PM

I don't like how it looks at all:

Screenshot 2024-07-03 at 3 43 39 PM

romain-growthbook:
Ok, UI/UX comment fixed or addressed. I think we should merge this as is and follow-up with details on UI/UX level.

lukesonnet (approved):

PR #2512, opened 2024-05-13
Introduce UI for Custom Roles

mknowlton89:
### Dependencies

- This PR is dependent on https://github.com/growthbook/growthbook/pull/2495 and needs to be merged after

### Fast Follows (Will complete as separate PRs after this is approved/merged)
- Add character counter to name/description fields
- Start using the `getConfirmationContent` on `RoleList`'s `<DeleteButton />`
- Build our deactivate/reactivate flows

### Changes from the mock
- I did increase the character count for a custom role description to be 100 characters (up from 56). Looking at our existing descriptions, the longest of those is 95, so it felt odd to limit it lower than that.

### Testing

- [x] Ensure that free/pro orgs can't create, edit, delete, duplicate, or apply custom roles to users or teams or set the org's default role
- [x] Ensure that enterprise orgs can create new custom roles
- [x] Ensure that enterprise orgs can delete custom roles
- [x] Ensure that enterprise orgs can edit custom roles
- [x] Ensure that enterprise orgs can duplicate custom roles
- [x] Ensure that enterprise org  can duplicate existing roles
- [x] Ensure that enterprise orgs can not delete standard roles
- [x] Ensure that only enterprise orgs can apply custom roles to their users & teams
- [x] Ensure an org can not create a custom role with an id equal to one of our standard roles
- [x] Ensure an org can not delete a custom role if there is a user or team using it
- [x] Ensure an org can not delete a custom role if it is the org's default role
- [x] Ensure that if an enterprise org downgrades to free or pro, they’re not longer able to add, edit, remove custom roles, nor are they able to add the custom role to any of their new or existing users or teams
- [x] Ensure that if an enterprise org downgrades to free or pro and the org's default role is a custom role, we reset their default role back to `collaborator`.
> [!WARNING]  
> On the test above, the code in `migrations.ts` doesn't actually update the org record in mongo. It looks like that isn't called downstream of anything that actually calls `updateOrganization`. But I can confirm that calling `getDefaultRole` returns the expected result.
- [x] Ensure that if a user has a custom role and an org downgrades to free or pro, the user's permissions are still based on the custom role


packages/enterprise/src/license.ts
```diff
@@ -51,6 +51,7 @@ export type CommercialFeature =
   | "multi-metric-queries"
   | "no-access-role"
   | "teams"
+  | "custom-roles"
```
jdorn, line 54: This already exists further down the list.  I'm surprised Typescript didn't complain about this.  Same with the change below in this file.

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
```diff
@@ -33,10 +34,42 @@ export default function SingleRoleSelector({
     roleOptions = roles.filter((r) => r.id !== "noaccess");
   }
 
+  if (!includeAdminRole) {
+    roleOptions = roles.filter((r) => r.id !== "admin");
```
jdorn, line 38: This should be `roleOptions.filter`.  Otherwise, it will overwrite the above `isNoAccessRoleEnabled` logic.

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
```diff
@@ -48,23 +81,18 @@ export default function SingleRoleSelector({
...
-            label: r.id,
-            value: r.id,
-          }))}
+        options={groupedOptions}
         sort={false}
+        formatGroupLabel={formatGroupLabel}
         formatOptionLabel={(value) => {
           const r = roles.find((r) => r.id === value.label);
+          if (!r) {
+            return;
```
jdorn, line 90: Kevin fixed this in a separate PR.  We want to do `return value.label` here instead of returning undefined.

jdorn (approved):

packages/enterprise/src/license.ts
mknowlton89: Oh yea - very weird.

PR #2513, opened 2024-05-13
Add concurrency limit to QueryRunner and DataSourceSettings

Kevin-Chant:
### Features and Changes

Creates a new field in `DataSourceSettings`: `maxConcurrentQueries`. This field is respected by `QueryRunner`'s  `startQuery` and `startReadyQueries` entrypoints, which now check the number of running queries before committing to starting a new query.

Added a helper to `QueryModel` to grab the count of running queries & a new section to the `ConnectionSettings` modal to allow for editing of this setting (and create a place for other future shared settings to live).

### Testing

1. Set up at least one experiment which makes multiple queries
2. Edit the settings for the datastore and provide a concurrency limit (probably 1)
3. Open the datastore queries list page in a new tab
4. Force update the results for the experiment(s) and check the queries list page to see that some have been queued.
  a. (If the delay isn't noticeable, you can change the `INITIAL_CONCURRENCY_TIMEOUT` to 8000.)

### Screenshots

image

image

Here's a screen recording of the UX with a concurrency limit of 1

queryconcurrency_ux

And here's a corresponding recording of what's happening on the backend with a few extra `console.log`s added for clarity.
queryconcurrency_backend

### DIFF
--- a/packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
+++ b/packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
@@ -274,7 +274,7 @@ export const startExperimentResultQueries = async (
     org
   );
 
-  const singlePromises = singles.map(async (m) => {
+  for (const m of singles) {
     const denominatorMetrics: MetricInterface[] = [];
     if (!isFactMetric(m) && m.denominator) {
       denominatorMetrics.push(
@@ -308,9 +308,9 @@ export const startExperimentResultQueries = async (
         queryType: "experimentMetric",
       })
     );
-  });
+  }
 
-  const groupPromises = groups.map(async (m, i) => {
+  for (const [i, m] of groups.entries()) {
     const queryParams: ExperimentFactMetricsQueryParams = {
       activationMetric,
       dimensions: dimensionObj ? [dimensionObj] : [],
@@ -343,9 +343,7 @@ export const startExperimentResultQueries = async (
         queryType: "experimentMultiMetric",
       })
     );
-  });
-
-  await Promise.all([...singlePromises, ...groupPromises]);
+  }
 
   let trafficQuery: QueryPointer | null = null;
   if (runTrafficQuery) {
--- a/packages/back-end/src/queryRunners/QueryRunner.ts
+++ b/packages/back-end/src/queryRunners/QueryRunner.ts
@@ -7,6 +7,7 @@ import {
   QueryType,
 } from "../../types/query";
 import {
+  countRunningQueries,
   createNewQuery,
   createNewQueryFromCached,
   getQueriesByIds,
@@ -60,6 +61,10 @@ export type StartQueryParams<Rows, ProcessedRows> = {
 };
 
 const FINISH_EVENT = "finish";
+// How long to wait before retrying a query that was queued due to concurrency limit.
+// Wait is doubled on subsequent retries, capped at the maximum
+const INITIAL_CONCURRENCY_TIMEOUT = 250;
+const MAX_CONCURRENCY_TIMEOUT = 4000;
 
 export async function getQueryMap(
   organization: string,
@@ -104,6 +109,7 @@ export abstract class QueryRunner<
     };
   } = {};
   private useCache: boolean;
+  private queuedQueryTimers: Record<string, NodeJS.Timeout> = {};
 
   public constructor(
     context: ReqContext | ApiReqContext,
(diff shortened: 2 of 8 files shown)


packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -443,6 +454,34 @@ export abstract class QueryRunner<
...
+    currentTimeout: number = 500
+  ): Promise<void> {
+    // If too many queries are running against the datastore, use capped exponential backoff to wait until they've finished
+    const concurrencyLimitReached = await this.concurrencyLimitReached();
+    if (concurrencyLimitReached) {
+      logger.debug(
+        `${doc.id}: Query concurrency limit reached, waiting ${currentTimeout} before retrying`
+      );
+      const nextTimeout = Math.min(currentTimeout * 2, 8000);
+      // TODO: ran into issue with orphaned queries here; is that dev-only or do we need to check for stale queries
```
jdorn, line 476: Much more common on dev since Node is constantly being killed and restarted, but can still happen on production, especially during deploys.  Stale queries should be cleaned up within 2 minutes max.  That's because we record a heart beat every 30 seconds and if a query misses 2 heart beats, we assume it's orphaned and kill it.  So 60 seconds + cron every minute = up to 2 minutes stale.

jdorn (commented):
In `startExperimentResultQueries`, we currently start all queries at once and do a `Promise.all`.  This is causing the race condition you mention in your PR description.

We can fix this fairly easily by changing it to await each promise before starting the next query.

It's really only the experiment query that has this issue since all of the other ones fire 1 (or maybe 2) queries total.

jdorn:
One thing I can see happening is someone's experiment being stuck in a queued state because all of the slots are being hogged by a few super long running queries.

What would be really cool is to pair this with https://github.com/growthbook/growthbook/pull/2323 so you can easily see a list of all of your running queries and where they are coming from and maybe even kill them directly from that list.

Kevin-Chant:
> One thing I can see happening is someone's experiment being stuck in a queued state because all of the slots are being hogged by a few super long running queries.
> 
> What would be really cool is to pair this with #2323 so you can easily see a list of all of your running queries and where they are coming from and maybe even kill them directly from that list.

This is one thing I was thinking about; didn't know that we have a page for that now! Do you think showing a link to that page while data is updating would do a good enough job at explaining what's happening?

jdorn:
Yeah, in `AsyncQueriesModal`, if there are any queries that are "queued", we can show an alert banner that links to the datasource queries list page.

The page isn't super useful right now (it doesn't link to the experiment that spawned the query and there's no way to cancel), but it's still helpful and we can improve it over time.

packages/front-end/components/Queries/AsyncQueriesModal.tsx
```diff
@@ -20,6 +20,7 @@ const AsyncQueriesModal: FC<{
 
   const [showStats, setShowStats] = useState(false);
   const hasStats = data?.queries?.some((q) => q.statistics !== undefined);
+  const datasourceId = data?.queries?.find((q) => q.datasource)?.datasource;
```
Kevin-Chant, line 23: My understanding from reading through the codepaths is that any grouped queries showing up in this modal will all be on the same datasource, so it's fine to just grab the first one. Does that sound right?

Kevin-Chant:
> In `startExperimentResultQueries`, we currently start all queries at once and do a `Promise.all`. This is causing the race condition you mention in your PR description.
> 
> We can fix this fairly easily by changing it to await each promise before starting the next query.
> 
> It's really only the experiment query that has this issue since all of the other ones fire 1 (or maybe 2) queries total.

Makes sense. I tried rewriting the promise logic in `ExperimentResultsQueryRunner` but ran into some problems where the queries weren't being set up properly (erroring out with `Run callbacks not found`). I'm leaning towards just shipping it as-is since it seems unlikely that a single experiment will overload the db so we should get most of the benefit even with this edge case exceeding the concurrency limit. Does that sound good to you?

packages/front-end/components/Queries/AsyncQueriesModal.tsx
lukesonnet: Yes, everywhere this modal is currently used it pulls from one datasource. I could imagine us changing that in the future but probably not worth worrying about right now.

packages/front-end/components/Settings/SharedConnectionSettings.tsx
```diff
@@ -0,0 +1,44 @@
...
+            type="number"
+            label={
+              <>
+                Maximum Concurrent Queries (Optional){" "}
+                <Tooltip
+                  body={
+                    "When executing queries against this datasource, if this many queries are already" +
+                    " running then new connections will wait for existing connections to finish. This" +
+                    " limit is not exact, e.g. if set to 100 it still might allow slightly over 100" +
+                    " queries to run simultaneously due to race conditions"
```
lukesonnet, line 30: Is it worth being more precise here and customer friendly and say something like the following?

```suggestion
                    " queries to run simultaneously if many are initiated by a single experiment update
```

Since that's basically the only update that would spawn more than 1-2 queries now and in the foreseeable future?

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -279,7 +279,7 @@ export const startExperimentResultQueries = async (
     organization
   );
 
-  const singlePromises = singles.map(async (m) => {
```
Kevin-Chant, line 282: The ES6 for loop format allows for awaiting within each iteration of the loop without going to the next, so we can drop the concurrency granularity to 1

packages/front-end/components/Queries/ExpandableQuery.tsx
```diff
@@ -162,6 +162,15 @@ const ExpandableQuery: FC<{
...
               </div>
+              <div className="col-auto mb-2">
+                <em>Time queued</em>:{" "}
+                <strong>
+                  {formatDistanceStrict(
+                    getValidDate(query.createdAt),
+                    getValidDate(query.startedAt)
+                  )}
+                </strong>
+              </div>
```
Kevin-Chant, line 173: This is definitely an optional addition but I found myself wanting to check the queued time after re-running the queries so I thought I'd toss it in here. Let me know whether this seems useful to keep or if I should get rid of it.

Kevin-Chant:
@jdorn I fixed the issue I had before & the queries now go one at a time. I updated the gifs with the new functionality and dropped the wait time by half because it's much more impactful now with a low concurrency limit. Should be ready for a review again once you're done with the 3.0 stuff

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
lukesonnet: I see. So the plan here is to be in line with Jeremy's comment and not kick off the queries all at once and await the query model to be created but rather do it sequentially.

It's kind of hard to tell, but is there any chance that `startQuery` gets stuck? In that method we call `setTimeout` a few times, as well as `executeQueryWhenReady` and I want to be sure those don't hold up the rest of the queries from being written to the DB asap. Is this a concern or am I just misunderstanding what will actually be `await`ed in `startQuery`?

packages/front-end/components/Queries/ExpandableQuery.tsx
lukesonnet: Oh no, thanks for adding this! It will be 0 for most queries, right? Do we only want to show it if the time queued is > some tiny number of `ms`? Fine either way, just a thought.

packages/front-end/components/Settings/ConnectionSettings.tsx
```diff
@@ -142,48 +155,57 @@ export default function ConnectionSettings({
...
-
+      break;
     default:
       invalidType = datasource.type;
       throw `Invalid type: ${invalidType}`;
   }
+  return (
+    <>
+      {datasourceComponent}
+      <SharedConnectionSettings
```
lukesonnet, line 205: I really like this solution, it won't trigger a re-test of the connection since it's affecting settings and not params.

However, you're adding it here to the "ConnectionSettings" which used to only affect those connection `params` which kicks off these test queries and stuff in the [putDataSource endpoint](https://github.com/growthbook/growthbook/blob/4b319072d263e77daef9bf9bd1eb17aba28cd359/packages/back-end/src/controllers/datasources.ts#L361).

There's 0 need to change your approach, but I feel like these could definitely live in `NewDataSourceForm` and `DataSourceForm`, where we already natively handle `settings` changes and keep them separate from `params` changes. Not really important at all, but wanted to call that out!

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
lukesonnet: My main concern is that some query gets stuck waiting to start and then that await doesn't let any of the other queries in the model get created and enter the `queued` state because this for loop has paused execution on the queued query. I could be totally off base about this.

lukesonnet (approved):
didn't test it myself, but it lgtm if you're confident about (a) how we handle orphaned queries and (b) being sure awaiting each query on creation is reasonable and won't be held up by the timeout stuff!

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
Kevin-Chant: Yep, that's the plan. For posterity, all of the async logic within `startQuery` is just waiting for reads & writes to Mongo (i.e. all bookkeeping), not anything related to the content of the query itself. So those awaits should be tiny delays & all the queries will probably be kicked off by the time the first one finishes (if the concurrency limit allows)

packages/front-end/components/Settings/ConnectionSettings.tsx
Kevin-Chant: I was looking at putting them there, but opted for this as it felt like the simplest option to avoid all of the conditional logic in the outer forms. I'm also hoping there are other shared settings we can use this for so it's not a one-off addition

PR #2518, opened 2024-05-14
Fix organization prior settings

lukesonnet:
Fix bug with setting priors at the org level

Previously, `settings` from the userContext (which pulls from the DB) would be clobbering the sub-fields of `metricDefaults` that were initialized in the form defaultValues. This isn't a problem for other fields that are not nested, but it is a problem for `metricDefaults.priorSettings` which will be missing in most databases.

This fixes that issue in the org settings and makes the FE a bit more robust to this issue.

A migration might be a better solution, to be honest.

### DIFF
--- a/packages/front-end/pages/settings/index.tsx
+++ b/packages/front-end/pages/settings/index.tsx
@@ -65,7 +65,8 @@ const GeneralSettingsPage = (): React.ReactElement => {
   const hasStickyBucketFeature = hasCommercialFeature("sticky-bucketing");
 
   const { metricDefaults } = useOrganizationMetricDefaults();
-
+  console.log(metricDefaults);
+  console.log(metricDefaults.priorSettings);
   const form = useForm<OrganizationSettingsWithMetricDefaults>({
     defaultValues: {
       visualEditorEnabled: false,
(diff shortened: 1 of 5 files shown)


packages/front-end/pages/settings/index.tsx
```diff
@@ -193,8 +191,15 @@ const GeneralSettingsPage = (): React.ReactElement => {
       const newVal = { ...form.getValues() };
       Object.keys(newVal).forEach((k) => {
         const hasExistingMetrics = typeof settings?.[k] !== "undefined";
-        newVal[k] = settings?.[k] || newVal[k];
-
+        if (k === "metricDefaults") {
+          // Metric defaults are nested, so take existing metric defaults only if
+          // they exist and are not empty
+          Object.keys(newVal[k]).forEach((kk) => {
+            newVal[k][kk] = settings?.[k]?.[k] ?? newVal[k][kk];
```
bttf, line 198: This is accomplishing the same as if you used the spread operator (`...`), with `settings` spread first. Any reason you wouldn't want to do it that way?

bttf (approved):
mostly minor comments

packages/front-end/pages/settings/index.tsx
lukesonnet: It needs to spread settings?.[k] which could work.

lukesonnet:
ty, landing in the interest of time, your feedback is useful and i can do it in a follow up.

PR #2520, opened 2024-05-14
Grant first user on a new installation super admin access

bttf:
RE: [Thread on a previous PR](https://github.com/growthbook/growthbook/pull/2182#discussion_r1595506259), we want to grant the first user to be created on a new installation with super admin access.

Changes
- Updated `createUser` method to take single object argument
  - Add optional `superAdmin` property, defaults to `false`
- We only grant super admin access in the `/auth/firsttime` endpoint
	- Endpoint used in situations where a 'new installation' is asserted (i.e., no existing orgs, no existing users)

Questions
- Briefly had similar changes in the `/auth/signup` endpoint but reverted since it seemed like an impossible scenario. Could use confirmation on this and any add'l insights on what could have been missed.

### DIFF
--- a/packages/back-end/src/controllers/auth.ts
+++ b/packages/back-end/src/controllers/auth.ts
@@ -225,7 +225,7 @@ export async function postRegister(
   }
 
   // Create new account
-  const user = await createUser(name, email, password);
+  const user = await createUser({ name, email, password });
   sendLocalSuccessResponse(req, res, user);
 }
 
@@ -276,7 +276,9 @@ export async function postFirstTimeRegister(
     });
   }
 
-  const user = await createUser(name, email, password);
+  // grant the first user on a new installation super admin access
+  const user = await createUser({ name, email, password, superAdmin: true });
+
   await createOrganization({
     email,
     userId: user.id,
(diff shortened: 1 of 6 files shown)


packages/back-end/src/controllers/auth.ts
```diff
@@ -276,7 +276,9 @@ export async function postFirstTimeRegister(
     });
   }
 
-  const user = await createUser(name, email, password);
+  // grant the first user on a new installation super admin access
+  const user = await createUser({ name, email, password, superAdmin: true });
```
jdorn, line 280: Digging into this endpoint more, `isNewInstallation` uses an in-memory cache, so it's possible in a horizontally scaled environment it could return `true` when there are actually already orgs/users.

Looking through the code, I don't think that in-memory cache is valuable.  `isNewInstallation` is not called that often and it's only doing 2 efficient mongo lookups.

So, I'd recommend removing the caching layer to reduce the chance of bugs here and elsewhere.

jdorn (approved):
I'll approve, but I do think we should make that change to `isNewInstallation` for extra safety

packages/back-end/src/controllers/auth.ts
bttf: Removed the cache layer from `isNewInstallation` and the use of `markInstalled` entirely since the former method will check every time.

PR #2523, opened 2024-05-15
New Get Started Flow

msamper:
### Features and Changes

Adds a brand new Get Started page with different checklists depending on what a user is looking to setup. The new get started page also includes helpful resources to help you get started with GrowthBook.

### Testing

Click through the UI and make sure everything works as expected. Mostly a self-contained UI PR.

### Screenshots

Screenshot 2024-05-21 at 4 14 13 PM

Screenshot 2024-05-21 at 4 14 55 PM


msamper:
Added:
- Dark mode support
- Guided step bar for all steps
- Some smaller fixes

Dark mode is just missing the new SVGs Kari generated, but I need to figure out how to dynamically swap those when we use dark mode since right now the SVGs are loaded through `` tags

packages/back-end/types/organization.d.ts
```diff
@@ -241,6 +241,17 @@ export type OrganizationMessage = {
...
+export type GetStartedChecklistItem = {
+  step: string;
+  isCompleted: boolean;
+};
+
+export type GetStartedChecklist = {
+  features: GetStartedChecklistItem[];
+  experiments: GetStartedChecklistItem[];
+  importedExperiments: GetStartedChecklistItem[];
+};
```
jdorn, line 253: Breaking this up by list isn't going to work well.  There are shared steps like setting up environments.  Better to just have a single array.

jdorn: Also, it's much simpler if we can just do an array of strings instead of an array of objects.  This will let us do really efficient Mongo queries to add/remove a string from the set without needing to worry about default states, merging, or race conditions.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -2093,6 +2096,46 @@ export async function putDefaultRole(
...
+  const index = newChecklist.getStartedChecklists[type].findIndex(
+    (c) => c.step === item.step
+  );
+  if (index === -1) {
+    newChecklist.getStartedChecklists[type].push({
+      step: item.step,
+      isCompleted: true,
+    });
+  }
+
```
jdorn, line 2131: All API endpoints that write to Mongo should have permission checks.  Ideally the check would be dependent on which step is being checked.  If someone doesn't have permission to create environments, they shouldn't be able to mark that step as completed.

packages/front-end/public/images/get-started/Feature Flag Bar - Active.png
jdorn, line 1: Why do we have both png and svg versions of these?

msamper: Ah I just need to delete the PNGs. Was using those originally

jdorn (approved):

PR #2529, opened 2024-05-15
API environment endpoints

romain-growthbook:


packages/back-end/src/api/openapi/openapi.yaml
```diff
@@ -190,6 +193,16 @@ paths:
...
+  /organizations/{id}/environments:
+    get:
+      $ref: "./paths/listEnvironments.yaml"
+    post:
+      $ref: "./paths/postEnvironment.yaml"
+  /organizations/{id}/environments/${environmentId}:
+    put:
+      $ref: "./paths/putEnvironment.yaml"
+    delete:
+      $ref: "./paths/deleteEnvironment.yaml"
```
jdorn, line 205: Don't put these under `/organizations/{id}`.  The routes should be `/environments/` and `/environments/${id}`

packages/back-end/src/api/openapi/parameters.yaml
```diff
@@ -48,6 +48,13 @@ factTableId:
   description: Specify a specific fact table
   schema:
     type: string
+environmentId:
+  name: environmentId
+  in: path
+  required: true
+  description: Specify an environment ID
+  schema:
+    type: string
```
jdorn, line 57: Remove this and use the existing `id` instead

packages/back-end/src/api/openapi/schemas/Environment.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - defaultState
+properties:
+  id:
+    type: string
+  description:
+    type: string
+  toggleOnList:
+    type: boolean
+  defaultState:
+    type: boolean
```
jdorn, line 15: Should also have `projects` as an array of strings

packages/back-end/src/api/openapi/paths/postEnvironment.yaml
```diff
@@ -0,0 +1,45 @@
...
+            description: The ID of the new environment
+          description:
+            type: string
+            description: The description of the new environment
+          toggleOnList:
+            type: bool
+            description: Show toggle on feature list
+          defaultState:
+            type: string
+            description: Default state for new features
```
jdorn, line 34: Add `projects` here (optional, array of strings)

packages/back-end/src/api/openapi/paths/putEnvironment.yaml
```diff
@@ -0,0 +1,40 @@
...
+        properties:
+          description:
+            type: string
+            description: The description of the new environment
+          toggleOnList:
+            type: bool
+            description: Show toggle on feature list
+          defaultState:
+            type: string
+            description: Default state for new features
```
jdorn, line 29: Add `projects` here (optional, array of strings)

packages/back-end/src/api/organizations/deleteEnvironment.ts
```diff
@@ -0,0 +1,49 @@
...
+  findOrganizationById,
+  updateOrganization,
+} from "../../models/OrganizationModel";
+import { OrganizationInterface } from "../../../types/organization";
+
+export const deleteEnvironment = createApiRequestHandler(
+  deleteEnvironmentValidator
+)(
+  async (req): Promise<DeleteEnvironmentResponse> => {
+    await validateIsSuperUserRequest(req);
```
jdorn, line 17: These endpoints are not for super admins, they are open to everyone

packages/back-end/src/api/organizations/deleteEnvironment.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const id = req.params.id;
+    const environmentId = req.params.environmentId;
+
+    const org = await findOrganizationById(id);
+    if (!org) {
+      throw Error("Organization not found");
+    }
```
jdorn, line 25: Use `req.context` or `req.organization` instead.

packages/back-end/src/api/organizations/deleteEnvironment.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const environments = org.settings?.environments || [];
+
+    const environment = environments.find((env) => env.id === environmentId);
+    if (!environment) {
+      throw Error(`Environment ${environmentId} does not exists!`);
+    }
```
jdorn, line 32: Missing permission checks.

```ts
if (!req.context.permissions.canDeleteEnvironment(environment)) {
  req.context.permissions.throwPermissionError();
}
```

There is an equivalent `canCreateOrUpdateEnvironment` method for the other endpoints as well

packages/back-end/src/api/organizations/deleteEnvironment.ts
```diff
@@ -0,0 +1,49 @@
...
+    const updates: Partial<OrganizationInterface> = {
+      settings: {
+        ...org.settings,
+        environments: [
+          ...environments.filter((env) => env.id !== environmentId),
+        ],
+      },
+    };
+
+    await updateOrganization(id, updates);
```
jdorn, line 43: Missing call to `req.audit` to store an audit log entry of this change.  Same with the other endpoints

packages/back-end/src/api/organizations/listEnvironments.ts
```diff
@@ -0,0 +1,28 @@
...
+export const listEnvironments = createApiRequestHandler(
+  listEnvironmentsValidator
+)(
+  async (req): Promise<ListEnvironmentsResponse> => {
+    const id = req.params.id;
+
+    const org = await findOrganizationById(id);
+    if (!org) {
+      throw Error("Organization not found");
+    }
```
jdorn, line 15: Same as the other endpoints.  This one is especially unsafe because anyone can list any other org's environments even without being part of the org.

packages/back-end/src/api/organizations/listEnvironments.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    return {
+      environments: (org.settings?.environments || []).map(
```
jdorn, line 18: We also need permission checks  when reading environments.  You can use this:

```ts
const environments = org.settings?.environments?.filter((environment) =>
  context.permissions.canReadMultiProjectResource(environment.projects)
);
```

jdorn (changes requested):

packages/back-end/src/api/environments/listEnvironments.ts
```diff
@@ -0,0 +1,41 @@
...
+export const listEnvironments = createApiRequestHandler(
+  listEnvironmentsValidator
+)(
+  async (req): Promise<ListEnvironmentsResponse> => {
+    const id = req.params.id;
+
+    const org = await findOrganizationById(id);
+    if (!org) {
+      throw Error("Organization not found");
+    }
```
jdorn, line 15: Do not let the user specify the org.  The org should always come from the request context.  With this code, any user could get the list of environments from any other org just by knowing the org's id.

romain-growthbook: That's a mistake thanks for spotting it.

packages/back-end/src/api/environments/putEnvironment.ts
```diff
@@ -0,0 +1,55 @@
...
+      !req.context.permissions.canCreateOrUpdateEnvironment(updatedEnvironment)
+    )
+      throw new Error("You don't have permission to update this environment!");
+
+    const updates: Partial<OrganizationInterface> = {
+      settings: {
+        ...org.settings,
+        environments: [
+          ...environments.filter((env) => env.id !== id),
+          updatedEnvironment,
```
jdorn, line 35: Array order is important for environments so we want to preserve that.

packages/back-end/src/api/openapi/paths/postEnvironment.yaml
```diff
@@ -0,0 +1,47 @@
...
+        -u secret_abc123DEF456:
+requestBody:
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        required:
+          - id
+          - description
```
jdorn, line 19: Default description to empty string instead of requiring it

packages/back-end/test/api/environments.test.ts
```diff
@@ -0,0 +1,775 @@
...
+    mongod = await MongoMemoryServer.create();
+    const uri = mongod.getUri();
+    process.env.MONGO_URL = uri;
+    getAuthConnection().middleware.mockImplementation((req, res, next) => {
+      next();
+    });
+  });
+
+  afterAll(async () => {
+    await mongoose.connection.dropDatabase();
```
jdorn, line 53: What is this for?  Seems dangerous to have this here.  Feels like we should not reference `mongoose` directly here at all and instead do all of our operations on the `mongod` object, which we know is an in-memory mock server.

packages/back-end/test/api/environments.test.ts
```diff
@@ -0,0 +1,775 @@
...
+                toggleOnList: true,
+                defaultState: true,
+                projects: ["bla"],
+              },
+              {
+                id: "env2",
+              },
+            ],
+          },
+        },
```
jdorn, line 138: Lots of copy/pasting in this test files makes it a little hard to follow and maintain.  For example, if we add a new method to req.context that is needed under-the-hood, we will have to update it 13 times in this file.

packages/back-end/test/api/environments.test.ts
romain-growthbook: Looks like you still have to also stop mongoose but the DB drop can be skipped of course.

romain-growthbook:
All PR comments have been addressed now!

packages/back-end/src/api/environments/putEnvironment.ts
```diff
@@ -25,15 +25,15 @@ export const putEnvironment = createApiRequestHandler(putEnvironmentValidator)(
...
       settings: {
         ...org.settings,
-        environments: [
-          ...environments.filter((env) => env.id !== id),
-          updatedEnvironment,
-        ],
+        environments: environments.reduce(
+          (ret, env) => [...ret, env.id === id ? updatedEnvironment : env],
+          []
+        ),
```
jdorn, line 36: Minor code style, but `map` makes a lot more sense here than `reduce`.  I tend to avoid reduce whenever possible since it's usually much harder to read and follow the logic compared to alternatives.

jdorn (approved):

packages/back-end/src/api/environments/putEnvironment.ts
romain-growthbook: 100%

PR #2532, opened 2024-05-15
Show full error when trying to set a new license

tzjames:
### Features and Changes

We had been catching all errors and giving a generic error when setting a license.  This makes it really hard to debug what is going on when people got an "Invalid License Key" message.

- Shows the error message as is
- Makes JSON parsing errors a bit nicer to read.

### Testing
set `LICENSE_SERVER_URL=http://localhost:8080/api/v1/` in .env.local
and start dev license server
Remove licenseKey on organization in mongodb
Set a license "bad" and see "Could not parse license"
Set a license "license_bad" and see "License server errored with: License not found"
Kill the dev license server
Set "license_bad" and see "Could not connect to license server. Make sure to whitelist 75.2.109.47."
Start the dev license server
Set an actual license and see it accept it.

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -1978,18 +1978,8 @@ export async function setLicenseKey(
     );
   }
 
-  try {
-    org.licenseKey = licenseKey;
-    await initializeLicenseForOrg(org, true);
-  } catch (error) {
-    // As we show this error on the front-end, show a more generic invalid license key error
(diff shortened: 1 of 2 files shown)


mknowlton89 (approved):
@tzjames Not sure how to test this, but I know when we throw errors it logs it to Sentry. If we don't explicitly throw the error - do they still get logged to Sentry?

If not, could we not just catch and throw the full error?

E.G.

```ts
try {
    org.licenseKey = licenseKey;
    await initializeLicenseForOrg(org, true);
  } catch (error) {
      throw new Error(error);
  }
```

PR #2534, opened 2024-05-16
[WIP] Extra functionality for queries list page

Kevin-Chant:
### Features and Changes

Adds a new column to the Queries page of a datasource for actions to be taken on that query.

Initially there are two possible actions:
- Visit the experiment or metric responsible for triggering that query
- Cancel a running or queued query (if the datasource supports cancellation and the user has appropriate permissions)

Currently, only BigQuery and Athena support cancellation, but we can implement the `cancelQuery` function on more integrations in future PRs.

### Testing

Run queries from both an experiment & a metric to make sure that the back-linking works correctly.

### Screenshots

image
image

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -70,7 +70,6 @@ import { formatInformationSchema } from "../util/informationSchemas";
 import { ExperimentSnapshotSettings } from "../../types/experiment-snapshot";
 import { SQLVars, TemplateVariables } from "../../types/sql";
 import { FactTableMap } from "../models/FactTableModel";
-import { logger } from "../util/logger";
 import {
   FactFilterInterface,
   FactMetricInterface,
@@ -117,9 +116,6 @@ export default abstract class SqlIntegration
     sql: string,
     setExternalId?: ExternalIdCallback
   ): Promise<QueryResponse>;
-  async cancelQuery(externalId: string): Promise<void> {
-    logger.debug(`Cancel query: ${externalId} - not implemented`);
-  }
   abstract getSensitiveParamKeys(): string[];
 
   constructor(context: ReqContextClass, datasource: DataSourceInterface) {
--- a/packages/back-end/src/queryRunners/DimensionSlicesQueryRunner.ts
+++ b/packages/back-end/src/queryRunners/DimensionSlicesQueryRunner.ts
@@ -59,6 +59,10 @@ export class DimensionSlicesQueryRunner extends QueryRunner<
           this.integration.runDimensionSlicesQuery(query, setExternalId),
         process: (rows) => rows,
         queryType: "dimensionSlices",
+        querySource: {
+          sourceType: "DimensionSlices",
+          id: this.model.id,
+        },
       }),
     ];
   }
--- a/packages/back-end/src/queryRunners/PastExperimentsQueryRunner.ts
+++ b/packages/back-end/src/queryRunners/PastExperimentsQueryRunner.ts
@@ -43,6 +43,10 @@ export class PastExperimentsQueryRunner extends QueryRunner<
         process: (rows) =>
           this.processPastExperimentQueryResponse(rows, merge, params.from),
         queryType: "pastExperiment",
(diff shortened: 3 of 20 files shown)


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -117,9 +116,6 @@ export default abstract class SqlIntegration
     sql: string,
     setExternalId?: ExternalIdCallback
   ): Promise<QueryResponse>;
-  async cancelQuery(externalId: string): Promise<void> {
-    logger.debug(`Cancel query: ${externalId} - not implemented`);
-  }
```
Kevin-Chant, line 122: Getting rid of the abstract definition so we can more easily check whether an integration has a `cancelQuery` defined or not

packages/back-end/src/queryRunners/DimensionSlicesQueryRunner.ts
```diff
@@ -59,6 +59,10 @@ export class DimensionSlicesQueryRunner extends QueryRunner<
           this.integration.runDimensionSlicesQuery(query, setExternalId),
         process: (rows) => rows,
         queryType: "dimensionSlices",
+        querySource: {
+          sourceType: "DimensionSlices",
+          id: this.model.id,
+        },
```
Kevin-Chant, line 65: Checking my understanding here - there's no page to link back to for DimensionSlices, right? I figured it'd be worth storing the source anyway but open to changing that

packages/back-end/src/queryRunners/PastExperimentsQueryRunner.ts
```diff
@@ -43,6 +43,10 @@ export class PastExperimentsQueryRunner extends QueryRunner<
         process: (rows) =>
           this.processPastExperimentQueryResponse(rows, merge, params.from),
         queryType: "pastExperiment",
+        querySource: {
+          sourceType: "PastExperiments",
+          id: this.model.id,
+        },
```
Kevin-Chant, line 49: Same question as for DimensionSlices. It seems like a PastExperiments import doesn't have any canonical point to link back to but it might still be worth keeping the source tagged

packages/front-end/pages/datasources/queries/[did].tsx
```diff
@@ -86,6 +92,8 @@ const DataSourceQueries = (): React.ReactElement => {
     );
   }
 
+  const supportsQueryCancellation = ["athena", "bigquery"].includes(d.type);
```
Comment (Kevin-Chant) [inline packages/front-end/pages/datasources/queries/[did].tsx:95]: Is there somewhere better that this can live? The source of truth is really whether the backend integration includes a `cancelQuery` function so that can't exactly be checked from the frontend code here

packages/front-end/pages/datasources/queries/[did].tsx
```diff
@@ -155,6 +164,36 @@ const DataSourceQueries = (): React.ReactElement => {
               }
             }
 
+            let LinkToQuerySource = <></>;
+            if (query.querySource) {
+              // Backlinks are supported for Metrics and Experiments as they have canonical pages to link to
```
Comment (Kevin-Chant) [inline packages/front-end/pages/datasources/queries/[did].tsx:169]: If there are any other major types I should add support for let me know

packages/front-end/pages/datasources/queries/[did].tsx
```diff
@@ -215,6 +254,29 @@ const DataSourceQueries = (): React.ReactElement => {
...
+                          className="text-danger mr-2"
+                          onClick={(e) => {
+                            e.preventDefault();
+                            e.stopPropagation();
+                            apiCall<Response>(
+                              `/datasource/${did}/cancel/${query.id}`,
+                              {
+                                method: "POST",
+                              }
+                            );
```
Comment (Kevin-Chant) [inline packages/front-end/pages/datasources/queries/[did].tsx:272]: I could use some input on what to do with the response to this api call. I wasn't sure what the best way to surface an error message or success status to the user would be, as the UI on this page is already getting somewhat cluttered.

packages/shared/src/permissions/permissions.constants.ts
Kevin-Chant, line 1: Does this stab at permissions seem right? Or any obvious changes I should make?

PR #2536, opened 2024-05-16
Bug - Adds permission check to event webhook id page before fetching details

mknowlton89:
### Features and Changes

Previously the webhook event details page `/settings/webhooks/event/:eventWebhookId` didn't have a permission check on the front-end in front of the api call to fetch the details.

As a result, the endpoint was throwing an error, when in reality, we should be checking the user's permission level before we fetch the details.

Its likely orgs were using the event webhooks and a user was clicking on the slack link or notification to see what changed, but their role doesn't provide access to view that info.

### Testing

- [x] Navigate to an event webhook id page directly as a non-admin and confirm that rather than us doing a network call that throws an error, we get a frontend error and the network call isn't made
- [x] Ensure admins can still view the page without issue


bttf (approved):

mknowlton89:
Closing this in lieu of this PR https://github.com/growthbook/growthbook/pull/2539

PR #2537, opened 2024-05-16
Introduce Deactivate/Reactivate role flow

mknowlton89:
### Features and Changes

This PR introduces the concept of `deactivating` roles. This feature will only be available to `Enterprise` organizations. This is being introduced now that we have `custom-roles`. With custom roles, organizations might want to exclusively use  the custom-roles they have defined. Or there may be a single role that doesn't work well within their organization.

Deactivating roles allows these orgs to hide them from the role selection UI so someone with `manageTeam` permissions is not able to add it to any new or existing users. If an org does need to, for some reason, undo it, they can reactivate the role.

When a role is deactivated, it doesn't change the in-app experience or permissions for any users who have that role applied to their user. It simply provides filtering on the front end.

### Testing

- [x] Ensure that you can deactivate a reserved role and upon doing so, the role gets a "deactivated" badge on `/team#roles`
- [x] Ensure that you can deactivate a custom role and upon doing so, the role gets a "deactivated" badge on `/team#roles`
- [x] Ensure that you can NOT deactivate a role if it is the org's default role.
- [x] Ensure that you can reactivate a reserved role and upon doing so, the badge is removed
- [x] Ensure that you can reactivate a custom role and upon doing so, the badge is removed
- [x] Ensure that if you delete a custom role and it's deactivated at the time of deletion, we also remove the id from the deactivatedRoles array
- [x] Ensure that if you deactivate a role, it is no longer accessible via the role selector when updating a user's role or setting the org's default role
- [x] Ensure that if a user has a deactivated role, their experience doesn't change at all
- [x] Ensure an enterprise is only able to choose a default role from their activated roles

Edge Cases:
- [x] Ensure that if an org deactivates ALL reserved roles to use only their custom roles, the role selector handles that gracefully
- [x] Ensure that if an org downgrades away from enterprise, we don't respect their deactivated roles (e.g. they act as if they're not deactivated)


packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -548,4 +549,42 @@ export async function removeCustomRole(
   }
 
   await updateOrganization(org.id, { customRoles: newCustomRoles });
+
+  // If the role we're deleting is deactivated, remove it from the deactivatedRoles array
+  if (org.deactivatedRoles?.includes(id)) {
+    const newDeactivatedRoles = org.deactivatedRoles.filter((r) => r !== id);
+
+    await updateOrganization(org.id, { deactivatedRoles: newDeactivatedRoles });
```
jdorn, line 557: We should combine this with the `updateOrganization` call above.

jdorn (approved):

PR #2538, opened 2024-05-16
Adds character counter to Field component

mknowlton89:
### Features and Changes

Within the mocks for the **Custom Roles UI** Kari included a live character counter on the name and description fields.

Under the hood, both of these are using the `Field` component. With this update, I've added a new supported prop (`currentLength`) to the `Field` component. And now, if you pass in both the `currentLength` and a `maxLength` we will render the live character counter like in the screenshot below.

### Screenshots
Mock up:
Screen Shot 2024-05-16 at 12 08 47 PM

Live example:
Screen Shot 2024-05-16 at 12 12 35 PM

### Notes
I did explore going down the route of calculating the `currentLength` within the `Field` component, but I abandoned that when I started thinking about the developer experience of using the `Field` component. If an engineer wanted to enable the character counter, they'd need to then pass in a new prop (maybe `enableCharacterCounter`) in addition to the `maxLength` and then in the code we'd need to check for that, and also make sure we had values for `currentLength` and `maxLength`. With this, now if you pass in both `currentLength` and `maxLength`, it automatically renders (assuming both of those props are not undefined).

**Open to suggestions/feedback if you have a different approach that you think would be better. This is a highly used component so I want to make sure we get it right.**

### Testing

-[x] Test out a handful of `Field` components in the app and ensure this doesn't introduce any regressions at popular screensizes.
-[x] Ensure that the character counters are accurate

### DIFF
--- a/packages/front-end/components/Forms/Field.tsx
+++ b/packages/front-end/components/Forms/Field.tsx
@@ -41,6 +41,7 @@ export type BaseFieldProps = {
   prepend?: ReactElement | string;
   append?: ReactElement | string;
   comboBox?: boolean;
+  currentLength?: number;
 };
 
 export type FieldProps = BaseFieldProps &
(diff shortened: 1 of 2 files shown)


msamper (approved):
Reviewed/tested and lgtm! I wonder if there's a way to check the character count from within the Field component rather than passing it in as a prop. Presumably there should be, but I'm not sure how you'd do it since we're using `useForm`

romain-growthbook:
Sorry late for the party here. Yeah, a combined prop would be good because the risk is to have say `currentLength` drop out  at some point but `maxLength` being there. Could could also condition the prop display on having both.

Having the field be responsible for currentLength by default could be cool. An escape hatch is also nice since some fields may have e.g. markdown formatting which may not accurately reflect the character count.

PR #2539, opened 2024-05-16
View events refactor

jdorn:
### Features and Changes

Today, only admins can view an audit log of events for their org.  This makes sense for events like `user.login` - we don't want to expose the login ip/userAgent of everyone in the org to non-admins.  But for other events like `feature.updated`, it's too restrictive - we want users to be able to click from a Slack notification and view the event details without being admins.

This PR changes the logic for viewing event details.  There are some events that contain sensitive info (like `user.login`) that we want to keep admin-only.  But for everything else, we now use the event subject (e.g. the feature or experiment) to determine view access.  If a user is able to view an experiment, they are now also able to view events about that experiment.

1. Add new top-level fields to events: `projects`, `tags`, `environments`, and `containsSecrets`
2. Simplified webhook filtering logic to use these new fields
3. Replaced permission `viewEvents`/`canViewEvents` with `viewAuditLogs`/`canViewAuditLogs`
4. Added new permission function `canViewEvent(eventPayload)` that checks read permission based on the new top-level event fields.

Testing
- [x] Test webhook filtering on project/tag/environment for features
- [x] Test webhook filtering on project/tag/environment for experiments
- [x] Able to view non-sensitive events with `readonly` role
- [x] Only able to view sensitive events with `admin` role

### DIFF
--- a/packages/back-end/src/routers/events/events.controller.ts
+++ b/packages/back-end/src/routers/events/events.controller.ts
@@ -1,4 +1,5 @@
 import type { Response } from "express";
+import { NotificationEvent } from "@back-end/src/events/notification-events";
 import * as Event from "../../models/EventModel";
 import { AuthRequest } from "../../types/AuthRequest";
 import { EventInterface } from "../../../types/event";
@@ -17,13 +18,13 @@ export const getEvents = async (
 ) => {
   const context = getContextFromReq(req);
 
-  if (!context.permissions.canViewEvents()) {
-    context.permissions.throwPermissionError();
-  }
-
   const events = await Event.getLatestEventsForOrganization(context.org.id, 50);
 
-  return res.json({ events });
+  return res.json({
+    events: events.filter((event) =>
+      context.permissions.canViewEvent(event.data as NotificationEvent)
+    ),
+  });
 };
 
 type GetEventRequest = AuthRequest<null, { id: string }>;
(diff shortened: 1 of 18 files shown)


packages/back-end/src/routers/events/events.controller.ts
```diff
@@ -17,13 +18,13 @@ export const getEvents = async (
...
-  }
-
   const events = await Event.getLatestEventsForOrganization(context.org.id, 50);
 
-  return res.json({ events });
+  return res.json({
+    events: events.filter((event) =>
+      context.permissions.canViewEvent(event.data as NotificationEvent)
+    ),
+  });
```
mknowlton89, line 27: Now that this endpoint is filtering the events based on the requester's permission level, it feels like we could loosen the permissions needed to view `app.growthbook.io/events`.

I could imagine if a user is in the habit of clicking on a Slack notification to view an event, they might find value in being able to view all events (that they have permission to view).

mknowlton89: If we do go this route - we will want to conditionally render the export button as we can keep that to admins only.

mknowlton89 (approved):
Looks great - simplified quite a bit.

Left one comment for consideration.

packages/back-end/src/routers/events/events.controller.ts
jdorn: I can definitely see the value in seeing a filtered event log for just the things you have access to / care about.  Might want to think some more about the best UX for that.  I'm not sure sending people to `Settings -> Logs` is the right solution.

romain-growthbook (commented):
This looks great. I love the separation of concerns. 

One TS remark to leverage types on checking permissions. I think this could be very useful when extending events coverage in the future.

packages/back-end/src/services/users.ts
```diff
@@ -203,6 +203,12 @@ export async function trackLoginForUser({
     data: {
       current: auditedData,
     },
+    projects: [],
+    tags: [],
+    environments: [],
+    // The event contains the ip, userAgent, etc. of users
+    // When marked as containing secrets, view access will be restricted to admins
+    containsSecrets: true,
```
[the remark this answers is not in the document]
jdorn: I like the idea of leaning on Typescript more.  I might save this for a follow up since this PR is already getting quite big

PR #2542, opened 2024-05-16
Sdk connection api endpoint

romain-growthbook:


packages/back-end/src/api/openapi/payload-schemas/PostSdkConnectionPayload.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - name
+  - languages
+  - environment
+properties:
+  name:
+    type: string
+  languages:
+    type: array
+    items:
+      type: string
```
jdorn, line 12: For all of these new endpoints, we should use singular `language` here, make it optional, and an enum of SDKLanguage options.  We deprecated the ability to have multiple languages and require a single language for all new SDK Connections going forward.

packages/back-end/src/api/sdk-connections/deleteSdkConnection.ts
```diff
@@ -0,0 +1,27 @@
...
+  deleteSdkConnectionValidator
+)(
+  async (req): Promise<DeleteSdkConnectionResponse> => {
+    const sdkConnection = await findSDKConnectionById(
+      req.context,
+      req.params.id
+    );
+    if (!sdkConnection) {
+      throw new Error("Could not find sdkConnection with that id");
+    }
```
jdorn, line 19: Missing permission checks in all of these endpoints

packages/back-end/src/api/sdk-connections/postSdkConnection.ts
```diff
@@ -0,0 +1,80 @@
...
+      projects = [],
+      encryptPayload = false,
+      includeVisualExperiments = false,
+      includeDraftExperiments = false,
+      includeExperimentNames = false,
+      includeRedirectExperiments = false,
+      proxyEnabled,
+      proxyHost,
+      hashSecureAttributes = false,
+      remoteEvalEnabled,
```
jdorn, line 30: Some of these settings are pro/enterprise only.  We should check that here and throw errors if the user doesn't have those commercial features.  Same for the PUT endpoint.

For example:
```ts
if (!req.context.hasPremiumFeature("encrypt-features-endpoint")) {
  // Return error
}
```

packages/back-end/src/api/sdk-connections/postSdkConnection.ts
```diff
@@ -0,0 +1,80 @@
...
+        );
+    }
+
+    if (!languages.length)
+      throw new Error("You need to specify some lanuages!");
+
+    languages.forEach((lang) => {
+      if (!(sdkLanguages as readonly string[]).includes(lang))
+        throw new Error(`Language ${lang} is not supported!`);
+    });
```
jdorn, line 56: We should default the `sdkVersion` to the latest version using `getLatestSDKVersion(language)`.

We also might want to check version compatibility here.  Like only some languages/versions support remoteEval.

```ts
const capabilities = getSDKCapabilities(language, sdkVersion);

if (remoteEvalEnabled && !capabilities.includes("remoteEval")) {
  // Throw error
}
```

This capability checking also applies to the PUT endpoint

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -293,12 +293,13 @@ export async function editSDKConnection(
     }
   });
 
-  await SDKConnectionModel.updateOne(
+  const updatedConnection = await SDKConnectionModel.findOneAndUpdate(
```
jdorn, line 296: I prefer not using `findOneAndUpdate`.  Instead, just return `{...connection, ...updates}`.

jdorn (changes requested):

packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -0,0 +1,116 @@
+import { getLatestSDKVersion, getSDKCapabilities } from "shared/sdk-versioning";
+import { findAllProjectsByOrganization } from "../../models/ProjectModel";
+import { ApiReqContext } from "../../../types/api";
+import { sdkLanguages } from "../../util/constants";
+import { getEnvironments } from "../../services/organizations";
+
+const capabilityParams = [
+  ["encryption", "encryptPayload"],
+  ["remoteEval", "remoteEvalEnabled"],
```
jdorn, line 9: Also add `visualEditor` and `redirects`.

packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -0,0 +1,116 @@
...
+] as const;
+
+type CababilitiesParamKey = typeof capabilityParams[number][1];
+type CababilitiesParams = { [k in CababilitiesParamKey]?: boolean };
+
+const premiumFeatures = [
+  ["encrypt-features-endpoint", "encryptPayload"],
```
jdorn, line 16: Also add `visual-editor`, `hash-secure-attributes`, `remote-evaluation`, and `redirects`.

In addition, if the user doesn't have `cloud-proxy` and IS_CLOUD, then they shouldn't be able to specify any of the proxy settings.

packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -0,0 +1,116 @@
...
+    proxyHost,
+    hashSecureAttributes,
+    ...otherParams,
+  };
+
+  capabilityParams.forEach(([capability, param]) => {
+    if (payload[param] && !capabilities.includes(capability))
+      throw new Error(
+        `SDK version ${sdkVersion} doesn not support ${capability}`
+      );
```
jdorn, line 107: We should differentiate between the version not supporting this feature and the language not supporting it.  Basically, check the capabilities of the latest version and then you can say "This isn't supported by this language" vs "You need to update your version"

packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -7,17 +7,34 @@ import { getEnvironments } from "../../services/organizations";
...
-type PremiumFeatures = { [k in PremiumFeatureKey]?: boolean };
+type PremiumFeatureName = typeof premiumFeatures[number][0];
+type PremiumFeatureParam = typeof premiumFeatures[number][1];
+type PremiumFeatures = { [k in PremiumFeatureParam]?: boolean };
+
+const premiumExtraChecks: {
+  [k in PremiumFeatureName]?: { check: boolean; reason: string };
+} = {
+  "cloud-proxy": {
+    check: !!process.env.IS_CLOUD,
```
jdorn, line 34: Import this from `secrets.ts` instead of access `process.env` directly.  We do parsing of raw environment vars and some additional checks, so `IS_CLOUD` will always be a boolean.

packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -102,14 +120,30 @@ export const validatePayload = async (
...
       throw new Error(`Feature ${feature} requires premium subscription!`);
+
+    const extraCheck = premiumExtraChecks[feature];
+
+    if (extraCheck === undefined) return;
+
+    if (!extraCheck.check)
+      throw new Error(
+        `Feature ${feature} is not available: ${extraCheck.reason}`
+      );
```
jdorn, line 146: I think there's some misunderstandings here.

On self-hosted, the proxy settings are always allowed.  On cloud, the proxy settings are only allowed if the user has the `cloud-proxy` premium feature.

romain-growthbook: Gotcha

PR #2545, opened 2024-05-17
Exp checklist permission mismatch bug.

mknowlton89:
### Features and Changes

Our sentry logs indicated a user received an error when attempting to edit an experiment's checklist.

When digging in, it appeared that this error was thrown for an `analyst` user trying to update a manual checklist item on an experiment's checklist. After some digging, I found that on the frontend, when determining whether the user should have edit access, we were checking `canViewExperimentModal` which checks the `createAnalyses` permission, but on the backend, we were checking `canRunExperiment` which checks the `runExperiments` permission.

Looking at the roles and policies, an `analyst` is described as being able to create, edit, and delete experiments, so my understanding is the backend check we were doing was checking too high of a permission level.

This PR updates the backend to check the same permission that we're checking on the frontend so no users will get UI that says they can update, but the backend says otherwise.

### Testing

- [x] Ensure that analysts, experiments, and admins can edit manual checklist items, but no other roles.

### DIFF
--- a/packages/back-end/src/controllers/experimentLaunchChecklist.ts
+++ b/packages/back-end/src/controllers/experimentLaunchChecklist.ts
@@ -1,5 +1,6 @@
 import { Response } from "express";
 import { orgHasPremiumFeature } from "enterprise";
+import { ExperimentInterface } from "@back-end/types/experiment";
 import { getContextFromReq } from "../services/organizations";
 import { AuthRequest } from "../types/AuthRequest";
 import {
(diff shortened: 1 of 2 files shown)


packages/back-end/src/controllers/experimentLaunchChecklist.ts
```diff
@@ -135,9 +134,7 @@ export async function putManualLaunchChecklist(
     throw new Error("Could not find experiment");
   }
 
-  const envs = experiment ? getAffectedEnvsForExperiment({ experiment }) : [];
-
-  if (!context.permissions.canRunExperiment(experiment, envs)) {
+  if (!context.permissions.canViewExperimentModal(experiment.project)) {
```
jdorn, line 137: This seems like it should be `canUpdateExperiment` instead, right?

mknowlton89: DOH. Yes. I've updated the both places to use the `canUpdateExperiment`.

packages/back-end/src/controllers/experimentLaunchChecklist.ts
```diff
@@ -135,9 +134,8 @@ export async function putManualLaunchChecklist(
     throw new Error("Could not find experiment");
   }
 
-  const envs = experiment ? getAffectedEnvsForExperiment({ experiment }) : [];
-
-  if (!context.permissions.canRunExperiment(experiment, envs)) {
+  const updatedExperiment = { ...experiment, manualLaunchChecklist: checklist };
+  if (!context.permissions.canUpdateExperiment(experiment, updatedExperiment)) {
```
jdorn, line 138: The 2nd argument should just be the fields that are changing, not the entire updated experiment.

jdorn (approved):
Added a comment, but it's minor so approving

PR #2550, opened 2024-05-20
Updates RunQueries permission checks on front end

mknowlton89:
### Features and Changes
While doing some testing, I found that we weren't doing any permission checks before trying to generate or re-generate an information schema for a particular datasource, resulting in not only permission errors, but permission errors that weren't being handled and causing the app to crash.

Now, we are checking for permission on the frontend before we attempt to generate/regenerate a datasource's information schema.

Additionally, we didn't have any protections around the `Test Queries` permissions - now, in the EditSQLModal we are disabling the `Test Query` button and adding a tooltip if the user doesn't have 
 `canRunTestQueries` permission.

Likewise, we are automatically unchecking the `Test queries before saving` if the user doesn't have test query permissions, and disabling the input + wrapping it in a tooltip.

Reproduction steps / Testing
- Create a custom role with the following policies - `DatasourcesFullAccess`, `DataSourceConfiguration`, and `MetricsFullAccess`.
- Apply this role to a user and then attempt to create a metric on a datasource that supports information_schemas
- Click the `Edit SQL` button and see that there is no information schema or the info schema for the connected datasource is out of date and needs to be refreshed.
- Ensure that the CTA is disabled with a tooltip.

### Screenshots
Screenshot of the info schema table refresh button when user doesn't have permission
Screen Shot 2024-05-20 at 9 13 16 AM

Screenshot of the info schema databaes refresh button when user doesn't have permission
Screen Shot 2024-05-20 at 9 13 09 AM

Screenshot of the disabled input when user doesn't have test queries permission
Screen Shot 2024-05-20 at 9 12 29 AM

Screenshot of the disabled Test Query button if the user doesn't have test queries permission
Screen Shot 2024-05-20 at 9 08 24 AM

Screenshot of the disabled retry button when the user doesn't have canRunSchemaQueries permission
Screen Shot 2024-05-20 at 9 08 28 AM


mknowlton89:
NOTE: I've also updated the underlying policies that resulted in this issue in [PR 2551](https://github.com/growthbook/growthbook/pull/2551) - That likely removes most of the need for this PR, but I still think adding this protection on the frontend makes sense, especially since these errors caused the app to crash.

jdorn:
If the app was crashing, it's not because we were missing permission checks. It's because we're making an async call in a click handler without catching errors. Adding permission checks is good, but we should also fix the underlying issue. Otherwise, the crashing bug will still happen if the network request fails for any other reason.

mknowlton89:
> If the app was crashing, it's not because we were missing permission checks. It's because we're making an async call in a click handler without catching errors. Adding permission checks is good, but we should also fix the underlying issue. Otherwise, the crashing bug will still happen if the network request fails for any other reason.

Good call - after looking into it further, we weren't awaiting the response in the `refreshOrCreateInfoSchema` call, resulting in the uncaught exception. This is now async and will await the response and catch the error correctly.

See the change in [this commit](https://github.com/growthbook/growthbook/pull/2550/commits/edfffebed4e9b093c1ce5654d6e4e8361953ae07).

jdorn (approved):

PR #2557, opened 2024-05-21
Add input validation to RoleForm

mknowlton89:
### Features and Changes

Adds basic input validation to the `RoleForm` to prevent the roundtrip API call when we know an exception will be thown.

### Testing

- [x] Ensure a user can't submit the form with no `id` value set
- [x] Ensure a user can't submit the form to create a new role if the role is a reserved role id
- [x] Ensure a user can't submit the form to create a new role if the role id contains anything other than letters, numbers, or underscores
- [x] Ensure a user can submit the form to update a custom role and we don't do any validation on the role id (since it's not editable)

### DIFF
--- a/packages/front-end/components/Teams/Roles/RoleForm.tsx
+++ b/packages/front-end/components/Teams/Roles/RoleForm.tsx
@@ -60,8 +60,18 @@ export default function RoleForm({
 
   const saveSettings = form.handleSubmit(async (currentValue) => {
     setError(null);
+
+    if (!currentValue.id.length) {
+      setError("Name field is required");
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Teams/Roles/RoleForm.tsx
```diff
@@ -29,6 +29,28 @@ export default function RoleForm({
...
+      return false;
+    }
+
+    if (RESERVED_ROLE_IDS.includes(input.id)) {
+      setError("That role id is reserved and cannot be used");
+      return false;
+    }
+
+    if (!/^[a-zA-Z0-9_]+$/.test(input.id)) {
+      setError("Name field can only contain letters, numbers, and hyphens.");
```
jdorn, line 48: underscores, not hyphens

Issue #2494, opened 2024-05-08
[Feature] Update environment controller and remove ability to update environments from putOrganization


mknowlton89:
### Description of Feature

As part of an on-going effort to reduce to capabilities of organizations.controller, we should not allow an org to update an org's environments via the `organization.controller`'s `putOrganization` method.

We should block that functionality and instead add it explicitly to the `environment.controller`. Currently, we're missing a `putEnvironment` method (currently only supports updating multiple environments).

It is also missing a ` deleteEnvironment` method.

When we do this work, we should also split out the `PermissionsClass.canCreateOrUpdateEnvironment` method into separate `canCreate` and `canUpdate` methods, with the `canUpdate` method taking the shape of other `canUpdate` methods where we take in the existing environment, check to make sure the user has permission to update it, as well as the `updates` and make sure that after the updates, the user still have permission to edit.

Sample code for the `canUpdate` method (has not been tested, so just use as a starting point)

```ts
  public canUpdateEnvironment = (
    existing: Pick<Environment, "projects" | "id">,
    updates: Pick<Environment, "projects" | "id">
  ): boolean => {
    const existingProjects = existing.projects?.length
      ? existing.projects
      : [""];
    if (
      !existingProjects.every((project) =>
        this.hasPermission("manageEnvironments", project, [existing.id])
      )
    ) {
      return false;
    }

    if ("projects" in updates) {
      const updatedProjects = updates.projects?.length
        ? updates.projects
        : [""];

      if (
        !updatedProjects.every((project) => {
          this.hasPermission("manageEnvironments", project, [updates.id]);
        })
      ) {
        return false;
      }
    }
    return true;
  };
  ```

### Related Issues

- PR where we introduces `canCreateOrUpdateEnvironment` - https://github.com/growthbook/growthbook/pull/2484

PR #2564, opened 2024-05-23
Remove update environments from putOrganization.

mknowlton89:
### Features and Changes

This PR updates how we handle CRUD actions for an organization's environments. Previously most of this logic was nestled within the `putOrganization` method, and we're making an effort to reduce the overall power of that method and instead, handle the logic in specific controllers.

This PR removes the ability for an organization to update their environments via this method, and instead, moves it to the `environments.controller.ts` file.

This PR introduces/moves the following logic:
- Adds specific delete route
- Adds specific put/:id route
- Adds specific updateOrder endpoint for move up/move down - prevents any backdoor to updating envs or any permission error but only accepting an array of ids
- Updates the /post endpoint to accept the optional projects property (previously, we ignored this)
- Updates the bulk update endpoint 
- Moves logic to fire any web hooks to specific controller methods
- Updates `EnvironmentModal` to consume the PUT and POST requests (previously was using /organization)
- Updates ‘environments.tsx` to consume the new `/environment/order` PUT request

- Closes - https://github.com/growthbook/growthbook/issues/2494

### Testing
- [x] Ensure that engineers, experimenters, and admins can still add, update, and remove environments without issue
- [x] Ensure that non-engineers, experimenters, and admins can't add, update, remove or change order of environments
- [x] Ensure that engineers, experimenters, and admins can move up/move down environments, even if its an environment you don't have permission to edit


packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -18,6 +26,50 @@ type CreateEnvironmentResponse = {
...
+  const existingEnvs = org.settings?.environments;
+
+  if (!existingEnvs) {
+    return res.status(400).json({
+      status: 400,
+      message: "Unable to find organization's environments",
+    });
+  }
+
+  const updatedEnvs: Environment[] = [];
```
mknowlton89, line 48: TODO: Should I check to see if the user has permission to update envs globally or in atleast 1 project first?

packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -18,6 +26,50 @@ type CreateEnvironmentResponse = {
...
+    updatedEnvs.push(existingEnvs[index]);
+  });
+
+  await updateOrganization(org.id, {
+    settings: {
+      ...org.settings,
+      environments: updatedEnvs,
+    },
+  });
+  res.json({ environments: updatedEnvs });
```
mknowlton89, line 70: TODO: Add audit log here

packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -18,18 +26,88 @@ type CreateEnvironmentResponse = {
...
+    const index = existingEnvs.findIndex((env) => env.id === environment);
+
+    if (index < 0) {
+      return res.status(400).json({
+        status: 400,
+        message: `Unable to find environment: ${environment}`,
+      });
+    }
+
+    updatedEnvs.push(existingEnvs[index]);
```
romain-growthbook, line 70: Any reason for now using the plain `find` method?

romain-growthbook: Looks like the same code as `putEnvironments`. Maybe it could be factored out?

packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -41,13 +119,118 @@ export const putEnvironments = async (
...
+  );
+
+  if (existingEnvIndex < 0) {
+    return res.status(400).json({
+      status: 400,
+      message: `Could not find environment: ${environment.id}`,
+    });
+  }
+
+  //MKTODO: Update this to canUpdate when I break canCreateOrUpdateEnvironment out into two methods
```
romain-growthbook, line 174: `canOrShouldAgree` 😆

packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -41,13 +119,118 @@ export const putEnvironments = async (
...
+        ...org.settings,
+        environments: updatedEnvs,
+      },
+    });
+
+    if ("projects" in environment) {
+      const existingProjects = envsArr[existingEnvIndex].projects || [];
+      const newProjects = environment.projects || [];
+
+      if (JSON.stringify(existingProjects) !== JSON.stringify(newProjects)) {
```
romain-growthbook, line 194: This comparison is bound to be runtime dependent on the order of keys in the object. `lodash` is used in the code already and has a comparison utility that seems better suited: https://lodash.com/docs/4.17.15#isEqual

packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -41,13 +119,118 @@ export const putEnvironments = async (
...
+    await updateOrganization(org.id, {
+      settings: {
+        ...org.settings,
+        environments: updatedEnvs,
+      },
+    });
+
+    if ("projects" in environment) {
+      const existingProjects = envsArr[existingEnvIndex].projects || [];
+      const newProjects = environment.projects || [];
```
romain-growthbook, line 192: nit: 
```suggestion
    const  newProjects = environment.projects;
    if (newProjects) {
      const existingProjects = envsArr[existingEnvIndex].projects || [];
```
The typechecker should then know that inside this conditional, `newProjects` is actually defined. Also I remember some noise around `key in object` with old school JS but I'm not sure what the status of this is nowadays.. 😆

packages/back-end/src/routers/environment/environment.router.ts
```diff
@@ -29,8 +30,40 @@ router.put(
...
+      .object({
+        environment: z
+          .object({
+            id: z.string(),
+            description: z.string(),
+            toggleOnList: z.boolean().optional(),
+            defaultState: z.any().optional(),
+            projects: z.array(z.string()).optional(),
+          })
+          .strict(),
```
romain-growthbook, line 58: Should this be factored out to make it more visible for potential future re-use?

packages/back-end/types/audit.d.ts
```diff
@@ -46,6 +46,10 @@ export type EventType =
   | "organization.create"
   | "organization.update"
   | "organization.delete"
+  | "environment.create"
+  | "environment.update"
+  | "environments.update"
+  | "environment.delete"
```
romain-growthbook, line 52: Do you also need an entry in `EntityType`?

packages/front-end/pages/environments.tsx
```diff
@@ -222,12 +222,10 @@ const EnvironmentsPage: FC = () => {
                                 const newEnvs = [...environments];
                                 newEnvs.splice(i, 1);
                                 newEnvs.splice(i - 1, 0, e);
-                                await apiCall(`/organization`, {
+                                await apiCall(`/environment/order`, {
```
romain-growthbook, line 225: Do we need a net new endpoint for this? Should we simply allow the plural endpoint to also update the order?

romain-growthbook (commented):
Mostly questions and nitpicking stuff. The only strong suggestion is to use `loadash` to compare projects.

packages/front-end/pages/environments.tsx
mknowlton89: I ultimately broke it out into a separate endpoint as the permission logic for each of these is different.

For re-ordering, the user only needs to have permission to create environments globally, or in at least 1 project. Whereas, if the user is updating the actual environments, they need create/update permission for every endpoint passed in.

If we added this to the plural endpoint, the logic would get really complicated. We'd need to do a deep comparison of the environments being passed in to see if the user was just trying to update the order, or update the actual environments. 

Ultimately, it felt like it would lead to bugs, and was easily handled by breaking the logic out into two separate endpoints.

packages/back-end/src/routers/environment/environment.controller.ts
mknowlton89: No good reason for using `find` vs `findIndex`. I'll update the `find` as it does make things a bit more concise.

mknowlton89: Actually, I think for the `putEnvironment` it makes sense to keep the `findIndex` as with that, we want to find the index, so we can later update that specific index. I think `find` is better for the `putEnvironmentOrder` and `deleteEnvironment`, though.

packages/back-end/types/audit.d.ts
mknowlton89: When doing the bulk updates - either via the `putEnvironmentOrder` or `putEnvironments` - there isn't a single env.id for the entity, so I've currently just got it set to the organization. Not sure if that's the right direction...

romain-growthbook: I have a PR changing those type definitions to take a constant object mapping entities to audit event names. This should fix these kind of gaps.

Issue #2474, opened 2024-05-03
[Bug] Users with global no-access role are unable to view their personal access tokens


mknowlton89:
### Summary

If a user has a global `noaccess` role, we are not returning any of their personal access tokens when landing on `/account/personal-access-tokens` and it feels like they should be able to view them, considering they could have other project-based roles.

#### Expected Behavior

When visiting `/account/personal-access-tokens` users with global `noaccess` role should see their personal access tokens.

#### Current Behavior

Currently today, the user gets an empty state message and has a cta to create a personal access token, but once created, it never renders on the page because the no access filter is filtering them out.

akash-ind:
Hey @mknowlton89, Can you please assign me this bug?

PR #2565, opened 2024-05-23
Added condition so owners of the personal token can see their persona…

akash-ind:
### Features and Changes

Fixed Personal Token not visible to the user with no-access role.

- Closes **https://github.com/growthbook/growthbook/issues/2474**

### Testing

1. Create a Personal access token with a user
2. Assign this user a no-access role.
3. Check if the user can see their Personal Access Token.

### DIFF
--- a/packages/back-end/src/models/ApiKeyModel.ts
+++ b/packages/back-end/src/models/ApiKeyModel.ts
@@ -351,7 +351,10 @@ export async function getAllApiKeysByOrganization(
   });
 
   return keys.filter((k) => {
-    return context.permissions.canReadSingleProjectResource(k.project);
+    return (
+      context.permissions.canReadSingleProjectResource(k.project) ||
+      k.userId == context.userId
+    );
   });
 }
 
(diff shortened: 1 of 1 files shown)


packages/back-end/src/models/ApiKeyModel.ts
```diff
@@ -351,7 +351,10 @@ export async function getAllApiKeysByOrganization(
   });
 
   return keys.filter((k) => {
-    return context.permissions.canReadSingleProjectResource(k.project);
+    return (
+      context.permissions.canReadSingleProjectResource(k.project) ||
+      k.userId == context.userId
```
mknowlton89, line 356: @akash-ind - To prevent any unintended bugs, we should use strict equality here (e.g. `k.userId === context.userId`).

akash-ind: Got it

akash-ind: @mknowlton89 Fixed this.

mknowlton89 (approved):

PR #2571, opened 2024-05-26
Fix bug reading datasource from the config file

tzjames:
### Features and Changes

There was bug where people using datasources from config file while using new licenses were getting an error.  There was an off by one error thinking pages start at 0 instead of 1.  This gets around it by using the newer `getSelfHostedOrganization` function, which makes sense here as the line before it throws an error if it is cloud.

For multi-org self-hosted using config file, we don't care about which org the datasource is assigned to, so getting any arbitrary one is fine.  All organizations share the same datasource config.

### Testing
Add `config/config.yml` file where the datasource is defined. 
Start the dev server and see it start fine.
Look in the audit logs of the license server to see that the data->metaData->dataSourceTypes matches the data source type defined in the `config/config.yml`

### DIFF
--- a/packages/back-end/src/models/DataSourceModel.ts
+++ b/packages/back-end/src/models/DataSourceModel.ts
@@ -22,7 +22,7 @@ import { queueCreateInformationSchema } from "../jobs/createInformationSchema";
 import { IS_CLOUD } from "../util/secrets";
 import { ReqContext } from "../../types/organization";
 import { ApiReqContext } from "../../types/api";
-import { findAllOrganizations } from "./OrganizationModel";
+import { getSelfHostedOrganization } from "./OrganizationModel";
 
 const dataSourceSchema = new mongoose.Schema<DataSourceDocument>({
   id: String,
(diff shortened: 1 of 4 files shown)


packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -61,9 +61,15 @@ export async function getInstallationDatasources(): Promise<
...
-    return getConfigDatasources(organizationId);
+    // For multi-org self-hosted using config file, we don't care about which org the
+    // datasource is assigned to, so getting any arbitrary one is fine.  All organizations
+    // share the same datasource config.
+    const organizationId = (await getSelfHostedOrganization())?.id;
+    if (organizationId) {
+      return getConfigDatasources(organizationId);
+    } else {
+      throw new Error("No self-hosted organization found");
+    }
```
jdorn, line 72: If you look at the `getConfigDatasources` method, it only uses the org id to populate that field in the returned data sources.  So you can just pass in an empty string and it will be fine.  No need to do a Mongo lookup here and throw, potentially causing even more license error edge cases (e.g. someone sets a license env var before creating their first org).

Also, we should change the `getLicenseMetaData` function so it never throws (wrap the contents in try/catch).  We never want to block getting a license key just because there's an error gathering metadata.

tzjames: Good points.

packages/back-end/src/services/licenseData.ts
```diff
@@ -119,7 +120,13 @@ export async function initializeLicenseForOrg(
       );
     }
 
-    const metaData = await getLicenseMetaData();
+    let metaData;
```
jdorn, line 123: This should be typed

PR #2578, opened 2024-05-29
Add explicit multiple exposures section to troubleshooting doc

lukesonnet:
Title + link to trouble shooting from A/A test doc.

### DIFF
--- a/docs/docs/kb/experiments/aa-tests.mdx
+++ b/docs/docs/kb/experiments/aa-tests.mdx
@@ -37,7 +37,7 @@ A/A tests are a low-risk approach to help ensure that your experiment is set up 
 
 ### Problem: No traffic, incorrect metric values, or SRM errors (imbalanced traffic)
 
-For these errors, check our generic troubleshooting guide.
+For these errors, check our [troubleshooting guide](/kb/experiments/troubleshooting-experiments).
 
 ### Problem: Metrics show statistically significant lifts in the A/A test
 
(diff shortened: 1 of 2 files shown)


docs/src/components/Anchor.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+// eslint-disable-next-line import/no-unresolved
+import useBrokenLinks from "@docusaurus/useBrokenLinks";
```
romain-growthbook, line 3: This looks great. Do you know why the module is reported as unresolved?

docs/docs/event-trackers/amplitude.mdx
```diff
@@ -6,7 +6,7 @@ slug: amplitude
 ---
 
 import DefaultInstructions from './_default-event-instructions.mdx'
-const { title, description, sidebar_label, slug } = frontMatter
+export const { title, description, sidebar_label, slug } = frontMatter
```
romain-growthbook, line 9: nit: Do those need to be exported? They don't seem to be consumed anywhere.

romain-growthbook (approved):
This looks good if you've tested it. Do we expose the broken link check somewhere? Could a check be integrated into the CI?

romain-growthbook:
My bad, just reading your initial description. Ideally, we should make sure that we error when new broken link are discovered. Not sure how practical this would be.

docs/src/components/Anchor.tsx
lukesonnet: No I don't but I see we've done this elsewhere in this folder so I didn't interrogate it at the time.

docs/docs/event-trackers/amplitude.mdx
lukesonnet: I think the mdx to js approach we use expects it to be exported (it was the library complaining about these not being exported), presumably because they do get used somewhere else once the files are compiled. I'm not sure. Here's their docs: https://mdxjs.com/docs/using-mdx/

lukesonnet:
> My bad, just reading your initial description. Ideally, we should make sure that we error when new broken link are discovered. Not sure how practical this would be.

I agree, I just don't know if we can "ignore" these false positives in the .mdx files (I couldn't find a way) so we have to set the level to warn until the bug is resolved.

PR #2579, opened 2024-05-29
Generated Hypothesis <> Auto-generated Experiment flow

bttf:
This PR adds functionality to support onboarding from the hyp. generator into GrowthBook. It will receive identifiers via query params and fetch the generated hypothesis from supabase (via REST API call), then import the hypothesis into mongo and pre-generate an experiment. It's built into the existing getting started flow UI with tweaks for consistency.

Dependent PR: https://github.com/growthbook/terraform/pull/46

Changes
- Add `GET /generated-hypothesis/:uuid` backend endpoint
  - Used by frontend when first loading generated hypothesis inside growthbook
- Add `GeneratedHypothesis` model
  - Contains `findOrCreateGeneratedHypothesis` method which does heavy lifting
- Tweak Getting Started flow to support use case

Screenshots

First landing in GrowthBook

Screenshot 2024-07-23 at 6 54 47 PM

Pre-generated experiment
Screenshot 2024-07-23 at 6 54 53 PM

After starting experiment
image


packages/back-end/src/models/GeneratedHypothesis.ts
```diff
@@ -0,0 +1,264 @@
...
+        key: "1",
+        screenshots: [],
+        id: uniqid("var_"),
+      },
+    ],
+    phases: [
+      {
+        coverage: 1,
+        dateStarted: new Date(),
+        dateEnded: new Date(),
```
bryce-fitzsimons, line 113: are we supposed to set the end date?

packages/back-end/src/models/GeneratedHypothesis.ts
```diff
@@ -0,0 +1,264 @@
...
+          domMutations: sdk_payload.variations[1].domMutations,
+          variation: createdExperiment.variations[1].id,
+        },
+      ],
+    });
+    // TODO enable pro trial if they are not already enabled
+  } else {
+    // linked feature flag
+    const featureId = `${slug}-feature-flag`;
+    await createFeature(context, {
```
bryce-fitzsimons, line 166: do we need to do this? won't it just run as a standalone exp?

packages/back-end/src/models/GeneratedHypothesis.ts
bttf: I wasn't sure - we create experiments elsewhere in the app like this

packages/back-end/src/models/GeneratedHypothesis.ts
bttf: no, you can't run an exp without either a FF, visual change, or URL redirect. it's part of what we discussed is necessary for a 'pregenerated experiment'. creating this gives us one less step towards a running exp.

bryce-fitzsimons: got it, ok

packages/back-end/src/models/GeneratedHypothesis.ts
bttf: I removed it - seems to work fine

bryce-fitzsimons (approved):
approved:
- seems to be low-risk to existing flows
- functionality looks correct to me, but we should test e2e before directing traffic at this flow

packages/back-end/src/models/GeneratedHypothesis.ts
```diff
@@ -0,0 +1,228 @@
...
+  createdAt: Date,
+  organization: String,
+  url: String,
+  hypothesis: String,
+  experiment: {
+    type: String,
+    index: true,
+    require: false,
+  },
+});
```
jdorn, line 27: We should be using the BaseModel for all new DB models. It enforces best practices that are missing from here. For example, this code has a security vulnerability. You can fetch by uuid and it doesn't verify that the organization field matches your request context. With BaseModel, this would be handled automatically.

PR #2580, opened 2024-05-29
Fix Member's Page Horizontal Scroll

mknowlton89:
### Features and Changes

When viewing the `MembersList` and `InviteList` on `/settings/team` if the org has a decent number of environments, it was causing a weird horizontal scroll for the Members and Invites table.

It looks like we had some `overflowY: "auto"` styling in place, but it was only being applied to the `MembersList` when there was a `maxHeight` set.

Now, this is always being set. This is definitely not the ideal permanent solution, but I know there is a mock of a redesigned Members page, so this is just an attempt to fix what is otherwise a visual bug until we do that redesign.

This page needs a little love - I've outlined 1 issue with it here for tracking purposes - https://github.com/growthbook/growthbook/issues/2581.

### Screenshots

Before:
Screen Shot 2024-05-29 at 2 22 51 PM

(And after scrolling)
Screen Shot 2024-05-29 at 2 22 59 PM

After:
Screen Shot 2024-05-29 at 2 22 12 PM

### DIFF
--- a/packages/front-end/components/Settings/Team/InviteList.tsx
+++ b/packages/front-end/components/Settings/Team/InviteList.tsx
@@ -148,109 +148,111 @@ const InviteList: FC<{
       />
       {resending && <LoadingOverlay />}
       {resendMessage}
-      <table className="table appbox gbtable table-hover">
-        <thead>
-          <tr>
-            <th>Email</th>
-            <th>Date Invited</th>
-            <th>{project ? "Project Role" : "Global Role"}</th>
-            {!project && <th>Project Roles</th>}
-            {environments.map((env) => (
-              <th key={env.id}>{env.id}</th>
-            ))}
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Settings/Team/InviteList.tsx
```diff
@@ -148,109 +148,111 @@ const InviteList: FC<{
...
+                        Remove
+                      </button>
+                    </MoreMenu>
+                  </td>
+                </tr>
+              );
+            })}
+          </tbody>
+        </table>
+      </div>
```
mknowlton89, line 255: All this is doing is wrapping the table in a `<div>` with the style tag.

romain-growthbook (approved):
I you've tested it this looks good to me.

PR #2587, opened 2024-05-30
Filter projects by permission

mknowlton89:
### Features and Changes

There are a number of forms and modals throughout the app where we're rendering the list of projects, and currently, the only permission filtering we do is based on the user's `readAccess`. What that means is, all of these places where we're rendering a list of projects, we only filter projects out that the user doesn't have `readAccess` to.

However, in these places where we're rendering the lists of projects, the user is trying to do something - they're trying to create a metric, or edit a metric, as an example. And in these cases, we know which projects a user has permission to do what they're trying to do, but we're not filtering the projects based on their permission level.

So what was happening was, a user would go to edit a metric's projects, and we're showing all projects the user has `readAccess` for. If the user chooses a project they don't have permission to edit metrics in, they'll get a permission error.

Now, we filtering the projects based on not only `readAccess`, but also based on whether they have permission to do the action in a project (e.g. when editing a metric, we only render projects where the user CAN edit metrics in).

To do this, I've simply added an array.filter method when rendering the dropdown options.

There are two components where I've abstracted this logic to - `EditProjectForm` and `EditprojectsForm`.

### Note
<blockquote>
I did explore creating a specific `ProjectDropdown` that wrapped the `MultiSelectField` or the `SelectField` depending on if the resource has a `projects` property or a `project` property, but I wasn't able to come up with a pattern that was easy to use. Likewise, some of the permissions are `ProjectScoped` and some are `EnvScoped`.

 Handling this many dimensions (single project vs projects array, ProjectScoped vs EnvScoped permissions, and some components like `SDKConnectionForm` have custom logic on what the initial `projectOptions` should be) felt like too much for a single component to handle without introducing a bunch of potential bugs.

As a result, I opted for the approach that you'll see here, where I'm simply chaining an array.filter method when building the options array.
</blockquote>

### Testing

- [x] Ensure that admin users still see all available projects

**Metrics**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when creating a metric, they only see the two projects they're permitted to create metrics in.
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when editing a metric, they only see the two projects they're permitted to edit metrics in.

**Data Sources**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them permission to create data sources, when creating a datasource, they only see the two projects they're permitted to create datasource in.
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them permission to update a data source params, when editing a datasource, they only see the two projects they're permitted to edit datasource params in.

**Fact Tables**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when editing a Fact Table, they only see the two projects they're permitted to editing Fact Table in.

**Fact Metrics**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when editing a Fact Metric, they only see the two projects they're permitted to edit Fact metric in.

**Attributes**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when creating an attribute, they only see the two projects they're permitted to create attribute in.
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when editing an attribute, they only see the two projects they're permitted to edit attribute in.

**SDK Connections**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when creating a SDK Connection, they only see the two projects they're permitted to create SDK Connection in.
- [x]  Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when editing a SDK Connection, they only see the two projects they're permitted to edit SDK Connection in.
- [x] Ensure that if an SDK Connection's environment is project-scoped, we use that subset of environments as the base array to then filter based on permission level.

**Features**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when creating an experiment, they only see the two projects they're permitted to create experiments in.
- [x] Ensure that if the user doesn't have global permission to edit features, the "None" option isn't shown in the edit project dropdown

**Experiments**
- [x] Ensure that if a user has a global `readOnly` role, and two project-specific roles that grant them `Experimenter` permissions, when editing an experiment, they only see the two projects they're permitted to edit experiments in.
- [x] Ensure that if the user doesn't have global permission to edit experiments, the "None" option isn't shown in the edit project dropdown


jdorn:
When filtering select options, especially on edit forms, there's often an edge case where existing values are missing.  
For edit forms, I think it's better to include all current options, even if the user doesn't have permission for some of them.

As an example, if a metric is tagged to projects A and B, a user only has write access to A.  They go to edit the metric's projects, B is filtered out so they only see a single project selected - A - and they are allowed to save the form.  This basically means someone could open the form, not change anything, click save, and the projects would be modified.

It might be better to create a helper function that encapsulates this logic:

```ts
useProjectOptions(
  permissionFilter: (project: string) => boolean,
  existingProjects: string[]
): string[];
```

romain-growthbook (commented):
The code looks nice and straightforward.

However, I am wondering how this is gonna look like from a UX perspective?

I believe that it might be super confusing to have some projects showing up in some places and not in others without any context.

Is there any chance the projects that are filtered out could be shown as disabled? Or at least a help text added, explaining that the projects in each list are filtered by such and such?

mknowlton89:
> The code looks nice and straightforward.
> 
> However, I am wondering how this is gonna look like from a UX perspective?
> 
> I believe that it might be super confusing to have some projects showing up in some places and not in others without any context.
> 
> Is there any chance the projects that are filtered out could be shown as disabled? Or at least a help text added, explaining that the projects in each list are filtered by such and such?

That's a great callout.

I'll sync w/ Kari to get her opinion on the direction we go. I can either do some sort of `helpText` or `Tooltip`, or we can explore the route of disabling projects if the user doesn't have permission to perform a specific action.

Quick mock of using an info icon with a Tooltip. 

Screen Shot 2024-07-19 at 1 53 44 PM
Screen Shot 2024-07-19 at 1 53 39 PM

Note: We can hide the tooltip/icon whenever the list of projects isn't being filtered down (aka, if a user has full permission on all projects) to avoid any confusion.

mknowlton89:
@romain-growthbook Thanks for the feedback - just added a tooltip that outlines how the list is filtered and shared with Kari and got the thumbs up

romain-growthbook (approved):
Looks good now. Thanks for taking the time to enhance the UX!

PR #2588, opened 2024-05-30
Show license info in cloud admin view

tzjames:
### Features and Changes
When we update a cloud license in Retool, we need a way to refresh the mongo cache of their license in the growthbook app.  It would also be good for superAdmin's to be able to see the license data that it took.  So now we do a force refresh of the license when an org is expanded in the admin view, and show that info.  

Since the license checksum can be very long and lived in a table, it ended up expanding the size of the table, so we had to change the `tableLayout` to `fixed`, which then caused all columns to be equal width, so had to narrow the action columns to the width of their icon's to look correct.

- Allows license get endpoint to be viewed by superAdmins.
- Calls the license get endpoint when expanding an organization.
- Displays the license data on cloud.
- Fixes the expanded column width, to be the correct number of columns that the table has.

### Testing

In the `.env.local` files set 
```
IS_MULTI_ORG=true
IS_CLOUD=true
```
Restart the server
Login as a super admin that is not a member of every organization.
Set a licenseKey on an org that the superAdmin is not a member of in Mongosh.
On the /admin page expand the org
Expand the license collapsible and see the license data.
Expand another org that has no licenseKey set.
Expand the license collapsible and see it say there is no license data.
See that the expanded `td` width is the entire table width.

In the `.env.local` files set 
```
IS_MULTI_ORG=true
IS_CLOUD=false
```
Restart the server
Login as a super admin 
Go to /admin page and expand any org
See no License section.
See that the expanded `td` width is the entire table width.

### Screenshots
Screenshot 2024-05-30 at 5 53 38 PM

### DIFF
--- a/packages/back-end/src/controllers/license.ts
+++ b/packages/back-end/src/controllers/license.ts
@@ -28,10 +28,11 @@ import { updateSubscriptionInDb } from "../services/stripe";
  * want to restart their servers.
  */
 export async function getLicenseData(req: AuthRequest, res: Response) {
-  const context = getContextFromReq(req);
-
-  if (!context.permissions.canManageBilling()) {
-    context.permissions.throwPermissionError();
+  if (!req.superAdmin) {
+    const context = getContextFromReq(req);
+    if (!context.permissions.canManageBilling()) {
+      context.permissions.throwPermissionError();
+    }
   }
 
   let licenseData;
--- a/packages/front-end/pages/admin.tsx
+++ b/packages/front-end/pages/admin.tsx
@@ -49,7 +49,7 @@ function OrganizationRow({
   const { apiCall } = useAuth();
 
   useEffect(() => {
-    if (expanded && !license) {
+    if (isCloud() && expanded && !license) {
       const fetchLicense = async () => {
         setLicenseLoading(true);
         const res = await apiCall<{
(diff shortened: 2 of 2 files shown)


packages/back-end/src/controllers/license.ts
```diff
@@ -28,10 +28,11 @@ import { updateSubscriptionInDb } from "../services/stripe";
  * want to restart their servers.
  */
 export async function getLicenseData(req: AuthRequest, res: Response) {
-  const context = getContextFromReq(req);
-
-  if (!context.permissions.canManageBilling()) {
-    context.permissions.throwPermissionError();
+  if (!req.superAdmin) {
```
gazzdingo, line 31: do we want all admins to view this or just superadmins?

tzjames: Only super admins, or those with billing permissions can see it (which is what normal admins have) will be able to use this endpoint.

mknowlton89: I don't think the superAdmin check is 100% necessary. `canManageBilling` ultimately calls `checkGlobalPermission` which automatically returns true if the user is a superAdmin.

mknowlton89 (approved):
Looks good - left 1 comment, but approving as its inconsequential.

packages/back-end/src/controllers/license.ts
tzjames: It is 100% necessary in the getContextFromReq is what fails if the superAdmin is not a member of the org.  If we are expecting superadmins to be able to call context then perhaps an exception needs to be made in that function.

mknowlton89: Ah - I see now. Disregard. :)

PR #2589, opened 2024-05-30
users agreed to terms

gazzdingo:
### Features and Changes
THIS NEEDS TO BE DEPLOYED AFTER CHANGES ARE PUSHED TO AUTH0

we know that users are agreeing to terms on auth0 since we require it so we are setting that in our DB for all new users on clould


bryce-fitzsimons (approved):
conditionally approved: logic looks fine, just fix the merge conflicts and test

PR #2590, opened 2024-05-31
Join/Create a new organization in multi-org after having already joined one

tzjames:
### Features and Changes
Adds a link to the create or join component that is now made into its own page.

### Testing
With the .env var settings:
```
IS_CLOUD=false
IS_MULTI_ORG=true
SSO_CONFIG=<A valid SSO config>
```

#### Test Login as a new user
- Login as a new user via SSO whose email domain matches the verifiedDomain of one or more organizations
- See the same component as before this change with the WelcomeFrame.
- Click on the Create a new organization button.
- See the "Create organization" title.

#### Test both env vars are true
- No need to set these env vars as they are both true by default
```
ALLOW_SELF_ORG_CREATION=true
SHOW_MULTI_ORG_SELF_SELECTOR=true
```
- See "Create or join a different organization" in the dropdown selector
- Click the link
- See that you land on the join mode.
- See that you can click the create button at the bottom of the screen
- See that it says "Create a new organization"
- Click an organization that doesn't have `autoApproveMembers` set
- See the green message saying that they are in a pending state
- See the button now says "Pending" for that organization and is disabled.
- Click an organization that does have `autoApproveMembers` set to `true`
- See that you are redirected to "/"
- See that the organization you clicked is now the currently logged in organization.

#### Test logging in as a new user not via SSO
- See only the create a new organization screen, as they have no joinable orgs.

#### Test only ALLOW_SELF_ORG_CREATION env var is true
- Set these env vars:
```
ALLOW_SELF_ORG_CREATION=true
SHOW_MULTI_ORG_SELF_SELECTOR=false
```
- See "Create a new organization" link in the dropdown selector
- See it go straight to the "Create a new organization screen"
- Don't see a link to join a different organization

#### Test only SHOW_MULTI_ORG_SELF_SELECTOR env var is true
- Set these env vars:
```
ALLOW_SELF_ORG_CREATION=false
SHOW_MULTI_ORG_SELF_SELECTOR=true
```
- See "Join a different organization" link in the dropdown selector
- See it go straight to the "Join a different organization" screen
- Don't see a link to create a new organization.

#### Test that neither env var is true
- Set these env vars:
```
ALLOW_SELF_ORG_CREATION=true
SHOW_MULTI_ORG_SELF_SELECTOR=false
```
- Don't see a link in the dropdown selector
- Navigate manually to "/settings/organizations"
- See a red error message that that page is only available with the correct env var settings.

### Screenshots
https://www.loom.com/share/3c7cb723370541adb0dc42bbf9521c48

The dropdown style has been updated since that loom was recorded. It now looks like:
Screenshot 2024-06-04 at 4 22 57 PM


packages/back-end/src/routers/users/users.controller.ts
```diff
@@ -229,11 +229,15 @@ export async function getRecommendedOrgs(req: AuthRequest, res: Response) {
         const currentUserIsPending = !!org?.pendingMembers?.find(
           (m) => m.id === user.id
         );
+        const currentUserIsMember = !!org?.members?.find(
```
tzjames, line 232: I could change this to just filter out members from the result here, then no need to do it client side.  It kind of makes sense that we wouldn't recommend an org that they are already a member of.  It's possible though the final design we want to show both the orgs that they are currently members of and the orgs that they can join, in which case we will need it like this.

packages/front-end/components/Auth/CreateOrganization.tsx
```diff
@@ -162,7 +180,11 @@ export default function CreateOrganization(): ReactElement {
                         onClick={() => {
                           joinOrgFormSubmit(org);
                         }}
-                        disabled={org.currentUserIsPending || false}
+                        disabled={
+                          org.currentUserIsPending ||
+                          org.currentUserIsMember ||
```
tzjames, line 185: Can remove this line as it wouldn't be in joinableOrg if they are already a member.

packages/front-end/components/Auth/CreateOrJoinOrganization.tsx
```diff
@@ -15,9 +16,13 @@ import Field from "@/components/Forms/Field";
...
 
-import style from "./CreateOrganization.module.scss";
+import style from "./CreateOrJoinOrganization.module.scss";
 
-export default function CreateOrganization(): ReactElement {
+const CreateOrJoinOrganization: FC<{
+  showFrame?: boolean;
+  title?: string;
+  subtitle?: string;
+}> = ({ showFrame = true, title = "", subtitle = "" }) => {
```
gazzdingo, line 25: its safer to leave these as undefined here for the title and subtitle for your check later on to see if it is false.

packages/front-end/components/Auth/CreateOrJoinOrganization.tsx
```diff
@@ -250,8 +268,33 @@ export default function CreateOrganization(): ReactElement {
...
+      </div>
+    </div>
   );
-}
+
+  if (showFrame) {
+    return (
+      <>
+        <WelcomeFrame leftside={leftside} loading={loading}>
+          <a
```
gazzdingo, line 280: i think we are meant to use <Link> here

gazzdingo: oh wait it shouldnt matter since it doesnt link anywhere

tzjames: > i think we are meant to use here

Use `<Link>`? Github I think ate your tag and doesn't display it.  At any rate, yeah since it doesn't go anywhere the linter doesn't complain about this.  If it went anywhere I think the linter would complain.

gazzdingo: yeah i relized after i made this comment it didnt go anywhere no need to update it :)

gazzdingo (approved):

PR #2598, opened 2024-05-31
Databricks Pipeline Mode

lukesonnet:
Adds pipeline mode to Databricks.

Databricks has no automatic notion of a temporary table that persists beyond a session, so we have to tear down the units table after running the query. So we now need:
* A new drop table query
* The addition of `prerequisites: string[]` to the query model which a query waits for all queries to succeed OR fail (while `dependencies` require all queries to succeed, or else it fails out the query.

Also added ability to specify project (BQ), database (Snowflake), or Catalog (Databricks) for where to write tables (`writeDatabase`) in case you want it to be in a totally different part of your data warehouse.

Testing:
- [X] Generates queries and runs correctly
- [X] Check query statistics
- [x] Ensure Snowflake and BigQuery db/schema naming works as expected
- [X] Check table deletion and prerequisites


packages/back-end/src/integrations/Databricks.ts
```diff
@@ -14,6 +15,17 @@ export default class Databricks extends SqlIntegration {
       encryptedParams
     );
   }
+  isWritingTablesSupported(): boolean {
+    return true;
+  }
+  dropUnitsTable(): boolean {
+    return true;
```
gazzdingo, line 22: super nit: do we want any permitions checks around  isWritingTablesSupported and dropUnitTable to make sure that nothing slips through the cracks

lukesonnet: I suggested in slack: if your org has pipeline mode on, to make it work for all users, they basically must have permission to write and drop if they have permission to `runQuery` or else pipeline mode will need to be enabled user-by-user.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -788,6 +796,24 @@ export default abstract class SqlIntegration
     return { results: results.rows, duration };
   }
 
+  getDropUnitsTableQuery(params: DropTableQueryParams): string {
+    // valdidate units table query follows expected name to help
+    // prevent dropping other tables
+    if (!params.tableName.includes(UNITS_TABLE_PREFIX)) {
```
jdorn, line 802: `startsWith` instead of `includes`

jdorn: Wait, nevermind.  I guess there could be a schema/db added as well.

lukesonnet: I'm not sure it will because it will because of the database and schema that can be prepended.

packages/back-end/types/datasource.d.ts
```diff
@@ -166,8 +167,10 @@ export type DataSourceEvents = {
 
 export type DataSourcePipelineSettings = {
   allowWriting?: boolean;
+  writeDatabase?: string;
```
jdorn, line 170: This is configurable on the front-end, but it doesn't appear to be used on the back-end at all.

lukesonnet: Hmm, I went to add the ability to change database/project(BQ)/catalog(databricks) but then never followed through I guess. Well, it's working now.

You can override the default database by setting this field and it passes through. If it is unset, it accepts the default you have set in your connection settings.

lukesonnet: Confirmed it works with old settings for snowflake and BigQuery (blank continues to accept default database) and works for databricks.

jdorn (approved):

PR #2616, opened 2024-06-06
[WIP] Saved groups improvements

Kevin-Chant:
### Features and Changes

Adds a new saved group edit page for list-type saved groups which renders a paginated list of the members of the group. Each member can be removed individually, and members can be added via the existing `SavedGroupForm` which now also supports importing a csv.
Adds the backend endpoints (both standard and api/v1) to support the new operations
Changes the SDK payload generation process to use new operators, `$ingroup` or `$ningroup`, which use shared group IDs rather than the current `$in: [list_of_values]` format.
Creates a new top-level entry in the sdk payload for the definitions of those id lists
Adds versioning logic to scrub the payload back to its current format for sdks that don't support the new functionality

Remaining TODOs
- [ ] Remove 1.1.0 from javascript sdk versions (functionality to be built in a follow-up PR)
- [ ] Verify that adding `$ingroup` and `$ningroup` doesn't necessitate any other changes
- [ ] Get designs/specs for the ux of modifying saved groups
- [ ] Figure out why /api/v1 endpoints are failing and either fix it or remove them if they're unnecessary
- [ ] Add limits and warnings around list sizes
- [ ] Add sdk payload encryption?
- [ ] Add warnings or prevent users from using saved groups of sensitive attributes in frontend sdks

### Testing

The saved group edit page can be tested by clicking the links on any of the ID List saved groups on the `/saved-groups` page. Manually entering an id in the url as `/saved-groups/[sgid]` should work if and only if the saved group is an ID list. Condition groups should render an error.
The create-group modal now renders a choice between importing from csv or using the existing manual entry flow. This applies both on the `/saved-groups` list view and the `/saved-groups/[sgid]` edit page
Clicking the delete button on a single entry within a saved group should delete it from the group and properly mutate the local data

For the SDK payload, while this PR is in progress (and the 1.1.0 js sdk version is defined) you can test the new behavior by setting an sdk to use that version and loading the endpoint. Setting it back to 1.0.0 or below should revert to the current behavior.

### Screenshots

List view:
image

Import from csv:
image

Edit view:
image

Feature setup:
image

Resulting sdk payload:
image

Sdk payload on old version:
image


packages/front-end/components/SavedGroups/IdLists.tsx
```diff
@@ -2,6 +2,7 @@ import { SavedGroupInterface } from "back-end/types/saved-group";
 import { useMemo, useState } from "react";
 import { ago } from "shared/dates";
 import { getMatchingRules } from "shared/util";
+import { Link } from "spectacle";
```
bryce-fitzsimons, line 5: 👀

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -50,6 +52,11 @@ export const scrubFeatures = (
   if (capabilities.includes("prerequisites")) {
     allowedFeatureRuleKeys.push(...prerequisiteKeys);
   }
+  if (!capabilities.includes("savedGroupReferences")) {
+    features = JSON.parse(
```
bryce-fitzsimons, line 56: try/catch

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -136,3 +150,25 @@ export const scrubExperiments = (
...
+  idLists: IdLists,
+  capabilities: SDKCapability[]
+): IdLists | undefined => {
+  if (!capabilities.includes("savedGroupReferences")) {
+    return undefined;
+  }
+  return idLists;
+};
+
+const replaceGroupOperationsInString = (s: string, idLists: IdLists) => {
```
bryce-fitzsimons, line 164: probably should recurse: see `evaluatePrerequisiteState`

packages/back-end/src/routers/saved-group/saved-group.validators.js
Kevin-Chant, line 1: Not sure where these files came from since the import using them is 5 months old, but the code blames to me. Were they previously ignored but accidentally got added to git?

packages/back-end/src/app.ts
```diff
@@ -197,8 +197,25 @@ app.post(
...
+    credentials: true,
+    origin: origins,
+  })
+);
+
+// increase default max payload json size to 1mb and allow for per-route specific payload sizes
+app.use((req, res, next) => {
+  // Can add conditional logic on req.path and req.method to set the body limit here if needed
+  return bodyParser.json({ limit: "1mb" })(req, res, next);
+});
```
Kevin-Chant, line 218: The ordering here is important as we need the cors definition to be higher up in the chain and let it be overridden later. 

On the other hand, bodyParser limits can't be overridden, so I added this section for conditional logic. We ended up deciding to use the same 1mb limit for saved group payloads so this isn't actually necessary right now, but it might save some time in the future if we decide to expand the limit on any particular endpoint.

packages/back-end/src/api/saved-groups/listSavedGroups.ts
```diff
@@ -10,7 +10,9 @@ export const listSavedGroups = createApiRequestHandler(
   listSavedGroupsValidator
 )(
   async (req): Promise<ListSavedGroupsResponse> => {
-    const savedGroups = await getAllSavedGroups(req.organization.id);
+    const savedGroups = await getAllSavedGroups(req.organization.id, {
+      includeValues: false,
+    });
```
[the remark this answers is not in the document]
jdorn: We usually don't make breaking changes like this.  What we could do is only exclude values when the saved group is using references.  That will make this 100% backwards compatible and only affect groups that someone opts into being by reference.

packages/back-end/src/app.ts
jdorn: Have you tested that overriding cors for individual routes works as expected?  Specifically thinking of the routes which only specify origin and not credentials:

```ts
  cors({
    origin: "*",
  }),
```

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -303,7 +306,7 @@ export async function editSDKConnection(
     "includeDraftExperiments",
     "includeExperimentNames",
     "includeRedirectExperiments",
-    "remoteEvalEnabled",
+    "savedGroupReferencesEnabled",
```
jdorn, line 309: Is deleting `remoteEvalEnabled` intentional?

packages/back-end/src/app.ts
Kevin-Chant: I'm fairly sure I did but it's been a month and looking back at these changes I'm thinking it'd be safer to not add this logic now. I've at least gone through the exercise of figuring out how to change the limit, so that should save us time if we need to in the future.

packages/front-end/components/SavedGroups/LargeSavedGroupSupportWarning.tsx
```diff
@@ -0,0 +1,85 @@
...
+  unsupportedConnections: SDKConnectionInterface[];
+  unversionedConnections: SDKConnectionInterface[];
+} {
+  const { data: sdkConnectionData } = useSDKConnections();
+  let connections = sdkConnectionData?.connections || [];
+  connections = connections.filter(
+    // TODO: check that projects array being empty does include all projects
+    (conn) =>
+      conn.projects.length === 0 || conn.projects.includes(project || "")
+  );
```
Kevin-Chant, line 22: @bryce-fitzsimons need to double check this with you

bryce-fitzsimons: try to use `getConnectionsSDKCapabilities`, append if necessary

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -35,9 +36,31 @@ const stickyBucketingKeys = [
...
 
+// eslint-disable-next-line
+type Node = [string, any];
+// eslint-disable-next-line
+export type NodeHandler = (node: Node, object: any) => void;
+
+// Recursively traverses the given object and calls onNode on each key/value pair.
+// If onNode modifies the object in place, it walks the new values as they're inserted, updated, or deleted
+// eslint-disable-next-line
+export const recursiveWalk = (object: any, onNode: NodeHandler) => {
```
bryce-fitzsimons, line 47: move to shared/utils

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -136,3 +193,28 @@ export const scrubExperiments = (
...
+};
+
+// Returns a handler which modifies the object in place, replacing saved group IDs with the contents of those groups
+const replaceSavedGroups: (
+  savedGroups: Record<string, SavedGroupInterface>
+) => NodeHandler = (savedGroups: Record<string, SavedGroupInterface>) => {
+  return ([key, value], object) => {
+    if (key === "$inGroup" || key === "$notInGroup") {
+      const group = savedGroups[value];
+      object[key.replace("Group", "")] = group?.passByReferenceOnly
```
bryce-fitzsimons, line 214: not in group?

packages/front-end/styles/global.scss
```diff
@@ -2543,9 +2581,6 @@ html {
     fill: var(--text-color-muted);
   }
 }
-.cursor-pointer {
-  cursor: pointer;
-}
```
Kevin-Chant, line 2548: Dupe with a bootstrap built-in

packages/front-end/components/Features/RuleModal.tsx
```diff
@@ -193,7 +193,28 @@ export default function RuleModal({
...
+  const [
+    savedGroupTargetingSdkIssues,
+    setSavedGroupTargetingSdkIssues,
+  ] = useState(false);
+  const [
+    attributeTargetingSdkIssues,
+    setAttributeTargetingSdkIssues,
+  ] = useState(false);
+  const canSubmit = useMemo(() => {
+    console.log("Computing canSubmit memo");
```
bryce-fitzsimons, line 205:
```suggestion
```

bryce-fitzsimons: also seems unnecessary to useMemo here, but 🤷

packages/front-end/components/SavedGroups/IdListItemInput.tsx
```diff
@@ -0,0 +1,289 @@
...
+              onChange={(values) => {
+                setValues(values);
+              }}
+              placeholder="Separate values using the 'Enter' key"
+              delimiters={["Enter", "Tab"]}
+            />
+          )}
+          <div className="row justify-content-end mr-0">
+            <a
+              href="#"
```
bryce-fitzsimons, line 275: personal nit:
- don't use `href="#"`
- don't use `e.preventDefault()`
- do use `<a role="button" ...` instead

packages/front-end/components/Tabs/TabButton.tsx
```diff
@@ -40,6 +40,7 @@ export default function TabButton({
           active,
           last,
           "nav-button-item": newStyle,
+          "nav-tab-item": !newStyle,
```
bryce-fitzsimons, line 43:
```suggestion
```

bryce-fitzsimons (approved):

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -303,7 +306,7 @@ export async function editSDKConnection(
     "includeDraftExperiments",
     "includeExperimentNames",
     "includeRedirectExperiments",
-    "remoteEvalEnabled",
```
bryce-fitzsimons, line 306: why is this removed? even if the underlying payload format isn't changing, we still signal a cache eviction / webhook to things like proxy, CDN, etc

Kevin-Chant: There were two instances of `remoteEvalEnabled` in the list so I got rid of one of them when adding the new feature

bryce-fitzsimons: nvm: duplicate field

PR #2621, opened 2024-06-07
Use PiPlusBold Icon in Organization Dropdown

tzjames:
### Features and Changes

The icon file was missing, but it was also the wrong way to load the icons.

### Testing

Set `IS_MULTI_ORG=true`
restart backend server
In the organization dropdown see the plus icon next to the AddOrganization.


gazzdingo (approved):
LGTM

PR #2623, opened 2024-06-07
Introduce Auto Fact Tables

mknowlton89:
### Features and Changes

This PR introduces the backend changes necessary to support automatically generated Fact Tables.

As with automatic metrics, this is currently only supported for `Segment`, `GA4`, `Rudderstack`, and `Amplitude` data sources.

Now that we're supporting generating metrics AND fact tables automatically, I did make some changes - namely, I reused some of the logic for auto metrics here, but in doing so made some of the existing methods a bit more generic (e.g. getEventsTrackedByDatasource now gets the events tracked, and doesn't do any sort of formatting or processing of the data, and `TrackedEventData` is now generic). 

### Call Outs
This PR was just intended to introduce backend changes to support Auto Fact Tables, but as I made some of the resources a bit more generic, there were some changes that required updating the UI for Auto Metrics just a bit. Those changes are included here. I do have a separate PR that will handle the UI of Auto Fact Tables.

### Testing
(If you'd like to do end-to-end testing) you can use this PR/branch as it has both the backend and frontend changes. Please note: the frontend changes for Auto Fact Tables are only the changes required to test and aren't the final designs.
Auto Metrics
- [x] Ensure no new regressions are added around Auto Fact Metrics for Segment, Rudderstack, Amplitude, and GA4 data sources (currently I can only test Segment & GA4 - I need to stand up new Rudderstack and Amplitude test envs to test those)
Segment Testing
- [x] Ensure that you can generate auto fact table for Segment data sources successfully, and each tracked event becomes a Fact Table
- [x] Ensure that each Fact Table created has the following columns: user_id, anonymous_id, timestamp, device, browser
- [x] Ensure that you can create a valid metric from these fact tables and apply filters to them as expected
GA4 Testing
- [x] Ensure that you can generate auto fact table for GA4 data sources successfully, and each tracked event becomes a Fact Table
- [x] Ensure that each Fact Table created has the following columns: user_id, anonymous_id, timestamp, county, source, medium, device, browser, and os
- [x] Ensure that you can create a valid metric from these fact tables and apply filters to them as expected


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -148,6 +152,7 @@ export default abstract class SqlIntegration
       pastExperiments: true,
       supportsInformationSchema: true,
       supportsAutoGeneratedMetrics: this.isAutoGeneratingMetricsSupported(),
+      supportsAutoGeneratedFactTables: this.isAutoGeneratingFactTablesSupported(),
```
mknowlton89, line 155: I could combine these two into a single `supportsAutoGeneratingResources` if we think these two will always be in sync.

packages/back-end/types/datasource.d.ts
```diff
@@ -58,7 +58,11 @@ export type SchemaFormat =
   | "clevertap"
   | "custom";
 
-export type AutoMetricSchemas = "segment" | "rudderstack" | "amplitude" | "ga4";
+export type SchemasThatSupportAutoFactTablesAndMetrics =
+  | "segment"
+  | "rudderstack"
+  | "amplitude"
+  | "ga4";
```
mknowlton89, line 65: Such a long name - how can I make this more concise?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3126,10 +3218,117 @@ AND event_name = '${eventName}'`,
...
-  async getEventsTrackedByDatasource(
-    schemaFormat: AutoMetricSchemas,
+
+  async getAutoFactTablesToCreate(
+    schemaFormat: SchemasThatSupportAutoFactTablesAndMetrics,
+    existingFactTables: FactTableInterface[],
+    datasourceId: string,
+    schema?: string
+  ): Promise<AutoFactTableToCreate[]> {
+    // GA4 stores all events in on the events_* table(s) - so we don't need to get a list of all events
```
mknowlton89, line 3228: The only downside of creating a single Fact Table per GA4 Data source (code wise) is that we have to have this messy block. Open to ideas on how to approach this differently.

mknowlton89:
@jdorn This is ready for an initial review (I still need to do a bit more testing, but wanted to get it in front of you early). Called out a few things specifically that I'd like opinions on.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2968,6 +3017,8 @@ export default abstract class SqlIntegration
             `WHERE ((_TABLE_SUFFIX BETWEEN '{{date startDateISO "yyyyMMdd"}}' AND '{{date endDateISO "yyyyMMdd"}}') OR
  (_TABLE_SUFFIX BETWEEN 'intraday_{{date startDateISO "yyyyMMdd"}}' AND 'intraday_{{date endDateISO "yyyyMMdd"}}')) 
 AND event_name = '${eventName}'`,
+          getFactTableWhereClause: (eventName: string) =>
+            `WHERE event_name = '${eventName}'`,
```
jdorn, line 3021: I don't really understand what's going on here.  There's ` getDateLimitClause`, but then the `getMetricWhereClause` duplicates all of that logic, but the `getFactTableWhereClause` does not

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2977,9 +3028,22 @@ AND event_name = '${eventName}'`,
...
+            ELSE 'Tablet/Desktop' END
+          ) as device,
+          (CASE 
+            WHEN context_user_agent LIKE '% Firefox%' THEN 'Firefox'
+            WHEN context_user_agent LIKE '% OPR%' THEN 'Opera'
+            WHEN context_user_agent LIKE '% Edg%' THEN ' Edge' 
+            WHEN context_user_agent LIKE '% Chrome%' THEN 'Chrome'
+            WHEN context_user_agent LIKE '% Safari%' THEN 'Safari'
+            ELSE 'Other' END
+          ) as browser`,
```
jdorn, line 3043: This is missing the following for segment, although it looks like Rudderstack does not support these.  So maybe we need to treat them separately here instead of in a single case block

```sql
  context_campaign_source as source,
  context_campaign_medium as medium,
```

packages/back-end/src/integrations/SqlIntegration.ts
mknowlton89: So... I think there are two things happening here.

1. I think the `getMetricWhereClause` for `GA4` that contains the date filtering is a bug. We only want the 7 day date limit when getting the tracked events for a data source, not when actually running the metric - we instead when running the metric query we want to use the `metricAnalysisDays`.

I think this isn't causing an issue because when we go to run the metric query, the date fields used in the `getMetricWhereClause` for GA4 are essentially duplicating the `metricAnalysisDays` logic. I took a screenshot below of the different in the queries when the `getMetricWhereClause` contains the date filtering, and when it doesn't.

Screen Shot 2024-06-25 at 8 36 38 AM

Given that, we should be able to remove the date filtering from GA4's `getMetricWhereClause` - and when doing that, that means for all of the supported schema types, the values of `getMetricWhereClause` and the `getFactTableWhereClause` should be the same, so we can combine them into a single property. Maybe `getEventFilterWhereClause`?

packages/back-end/src/integrations/SqlIntegration.ts
mknowlton89: @jdorn Hmm... neither of my Segment or Rudderstack sandboxes seem to be tracking these columns.

Do you recall where you saw documentation around this?

I'm looking at this article for [Segment](https://segment.com/docs/connections/storage/warehouses/schema/#event-tables) and this [article](https://www.rudderstack.com/docs/destinations/warehouse-destinations/warehouse-schema/) for Rudderstack.

Note: I do see this [article](https://www.rudderstack.com/docs/event-spec/standard-events/common-fields/) for Rudderstack showing the available contextual fields, but it doesn't appear that they're automatic.

mknowlton89: Per discussion, it appears that segment will only add the above columns if they're tracked, so we're not including them here.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2951,22 +2977,32 @@ export default abstract class SqlIntegration
           eventColumn: "event_type",
           timestampColumn: "event_time",
           userIdColumn: "user_id",
+          filterColumns: `device_family as device,
```
jdorn, line 2980: Should define this as an array of strings here

```ts
["device_family as device", "os_name as os", "country", "paying"]
```

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3065,55 +3129,86 @@ AND event_name = '${eventName}'`,
...
+      filterColumns,
+      getDateLimitClause,
+    } = this.getSchemaFormatConfig(schemaFormat);
+
+    const sqlQuery = `
+      SELECT
+        ${userIdColumn} as user_id,
+        ${anonymousIdColumn} as anonymous_id,
+        ${timestampColumn} as timestamp,
+        ${filterColumns}
```
jdorn, line 3152: This assumes `filterColumns` is not empty.  If it is empty, the trailing comma after `timestamp` will be a syntax error

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3065,55 +3129,86 @@ AND event_name = '${eventName}'`,
...
+      anonymousIdColumn,
+      getTrackedEventTablePath,
+      getEventFilterWhereClause,
+      filterColumns,
+      getDateLimitClause,
+    } = this.getSchemaFormatConfig(schemaFormat);
+
+    const sqlQuery = `
+      SELECT
+        ${userIdColumn} as user_id,
```
jdorn, line 3149: Didn't we have a `hasUserId` setting before?  I thought Rudderstack did not always add this column

mknowlton89: You're correct - I misread Rudderstack's documentation the first time. I've updated this to use the same logic as Auto Metrics does for the `user_id` column.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -187,23 +188,15 @@ export default abstract class SqlIntegration
...
-        this.datasource.settings.schemaFormat as AutoFactTableSchemas
-      ]
-    ) {
-      return true;
-    }
-    return false;
+  schemaFormatisAutoFactTablesSchemas(
+    schemaFormat: SchemaFormat
+  ): schemaFormat is AutoFactTableSchemas {
+    return supportedEventTrackers[schemaFormat as AutoFactTableSchemas];
```
jdorn, line 199: Nit. This return either true or undefined. Should make this return false instead of undefined.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3146,10 +3146,10 @@ export default abstract class SqlIntegration
 
     const sqlQuery = `
       SELECT
-        ${userIdColumn} as user_id,
+        ${hasUserId ? `${userIdColumn}, ` : ""}
```
jdorn, line 3149: Do we want to keep the alis (`as user_id`) here?

mknowlton89: ah - yes, good catch. Updated for metrics and fact tables.

jdorn (approved):

PR #2624, opened 2024-06-07
Bug - estimateParams can be undefined

mknowlton89:
### Features and Changes

A user reported an error when attempting to generate an impact score for an idea - when defining the default values for `userAdjustment`, `numVariations`, and `improvement` within the `ImpactModal.tsx` component, the `idea.estimateParams` was expected to be defined, but it's possible for an idea to have this be `undefined`. This seems to be somewhat expected given the fallback values. But, in the event of this happening, there was an error thrown.

Now, if you create a new idea (that doesn't have an `estimateParams` property, we correctly use the fallback values.

### DIFF
--- a/packages/back-end/types/idea.d.ts
+++ b/packages/back-end/types/idea.d.ts
@@ -19,7 +19,7 @@ export interface IdeaInterface {
   dateUpdated: Date;
   impactScore: number;
   experimentLength: number;
-  estimateParams: {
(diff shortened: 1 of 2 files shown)


Kevin-Chant (approved):
This LGTM; I checked the rest of the codebase to see if there are other places we were assuming `estimateParams` is defined but every other usage I saw is correctly checking for existence.

PR #2641, opened 2024-06-12
Introduce Auto Fact Tables UI

mknowlton89:
### Features and Changes

This PR introduces the UI for Auto Fact Tables and is dependent on https://github.com/growthbook/growthbook/pull/2623.

A few things to call out:
- `/fact-tables` now has a secondary CTA to `Auto-generate Fact Tables` - this button is disabled if the user doesn't have permission to create fact tables OR if the organization doesn't have any data sources that support auto generating fact tables (currently only available for GA4, Segment, Rudderstack, or Amplitude) data sources.
- We show tooltips on top of the CTA if it's disabled, and the context of the tooltip depends on why the user can't click it - e.g. if the user doesn't have permission, we show a permission warning, vs if the org doesn't have any data sources that support auto generating, the tooltip text explains that. If a user doesn't have permission AND the org doesnt have any supported data sources, we default to the permission warning.
- The form that's opened when this new CTA is clicked has a drop down where the user can select a data source - this list only contains data sources that support auto generating fact tables.

> [!NOTE]
> **This PR doesn't include an updated empty state for `/fact-tables` - that will be handled in a separate PR**

### Testing

- [x] Ensure that if the user doesn't have permission to create fact tables, the cta is disabled with the permission warning tooltip
- [x] Ensure that if the org doesn't have any data sources that support auto facts, the cta is disabled with the message stating they don't have any eligible data sources
- [x] Ensure that if the user doesn't have permission AND the org doesn't have any data sources that support auto facts, we show the permission warning
- [x] Ensure that if the user has permission AND the org has at least 1 data source that supports auto metric, the user can click the CTA and create 1 fact table per tracked event
- [x] Ensure that if a tracked event already has a fact table created for it, the row is disabled with a tooltip, and the checkbox is unchecked and can't be rechecked.
- [x] Ensure that if the user unchecks all of the auto fact tables to create, the submit cta is disabled
- [x] Ensure that if the user unchecks all but 1 of the auto fact tables to create, the submit cta is `Generate Fact Table` (singular), but if there is more than 1 checked, it's plural (`Generate Fact Tables`).
- [x] Ensure that once submitted, the fact tables are created successfully. 
- [x] Ensure that the checkbox on the table header row checks all/unchecks all
- [x] Ensure that if the user unchecks all manually, the header row check is automatically unchecked
- [x] Ensure that if the user checks all manually, the header row check is automatically checked
- [x] Ensure that clicking the checkbox on the table header row skips changing the state of the checkbox for tracked events already created


packages/front-end/pages/fact-tables/index.tsx
```diff
@@ -205,13 +220,28 @@ export default function FactTablesPage() {
...
+                : !dataSourcesThatSupportAutoFactTables.length
+                ? "Generating Fact Tables automatically is only available for GA4, Segment, Amplitude, and Rudderstack Data Sources"
+                : ""
             }
           >
+            <button
+              className="btn btn-outline-info mr-2"
+              onClick={(e) => {
+                e.preventDefault();
+                if (!canCreate) return;
```
mknowlton89, line 236: Is this necessary?

packages/front-end/components/AutoGenerateMetricsModal.tsx
```diff
@@ -413,7 +413,10 @@ export default function AutoGenerateMetricsModal({
                 Uncheck All
               </Button>
             </div>
-            <table className="appbox table experiment-table gbtable">
+            <table
+              className="appbox table gbtable"
+              style={{ tableLayout: "fixed" }}
+            >
```
mknowlton89, line 419: This fixes a bug to ensure the table always respects the modal's width

packages/front-end/pages/fact-tables/index.tsx
```diff
@@ -205,13 +220,28 @@ export default function FactTablesPage() {
...
+              !canCreate
+                ? `You don't have permission to create fact tables ${
                     project ? "in this project" : ""
                   }`
+                : !dataSourcesThatSupportAutoFactTables.length
+                ? "Generating Fact Tables automatically is only available for GA4, Segment, Amplitude, and Rudderstack Data Sources"
+                : ""
             }
           >
+            <button
```
jdorn, line 232: Give each button its own tooltip.  Messaging will be confusing otherwise.

packages/front-end/components/AutoGenerateFactTablesModal.tsx
```diff
@@ -0,0 +1,553 @@
...
+        informationSchema?: InformationSchemaInterface;
+      }>(`/datasource/${dataSourceId}/schema`, {});
+      if (informationSchema?.error?.message) {
+        setAutoFactTableError(informationSchema.error.message);
+      }
+      setSelectedDatasourceData({ informationSchema });
+    }
+
+    if (selectedDatasource?.id) {
+      getInformationSchema(selectedDatasource.id);
```
jdorn, line 265: Need to add a `.catch` here, otherwise, this will be an uncaught error if the API call fails

jdorn (approved):

PR #2643, opened 2024-06-13
Don't allow superAdmins to create extra organizations if the license doesn't allow it

tzjames:
### Features and Changes

If someone sets superAdmin in mongo, even when IS_MULTI_ORG=false, and then they navigate to /admin page directly they see the page.  Furthermore they were able to add a new organization as we were were not throwing an error if they were a superAdmin.  But if IS_MULTI_ORG is false then superAdmin's shouldn't even exist.

In separate PRs I will close down the other back-end routes, and then another one the front-end /admin page.  I'm doing it separately otherwise the testing would be too tricky.

### Testing
In .env.local set 
```
IS_MULTI_ORG=false
```
Set superAdmin=true on your user in mongo
Go to /admin page.
Click Add Organization.
Fill out the form and submit.
See the error that "An organization already exists".

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -8,6 +8,7 @@ import {
   getLicense,
   getLicenseError,
   licenseInit,
+  orgHasPremiumFeature,
 } from "enterprise";
 import { experimentHasLinkedChanges } from "shared/util";
 import {
@@ -1200,8 +1201,11 @@ export async function signup(req: AuthRequest<SignupBody>, res: Response) {
     // there are odd edge cases where a user can exist, but not an org,
     // so we want to allow org creation this way if there are no other orgs
     // on a local install.
-    if (orgs && !req.superAdmin) {
-      throw new Error("An organization already exists");
+    if (orgs) {
+      const { org } = getContextFromReq(req);
+      if (!req.superAdmin || !orgHasPremiumFeature(org, "multi-org")) {
+        throw new Error("An organization already exists");
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


mknowlton89 (approved):
Looks good! 

As a general comment - what do you think about updating how we define `IS_MULTI_ORG`.

Is there a reason we can't create a helper function to calculate this instead of relying on the environment variables?

We have some helpers in `license.ts` like `orgHasPremiumFeature` that we could build a wrapper around, which would allow us to define `IS_MULTI_ORG` from the orgs license rather than just their env variables.

Not something that needs to be addressed here, it was just a thought I had while reviewing this. Maybe there are protections in place to prevent an org from just setting the ENV variable.

tzjames:
> Looks good!
> 
> As a general comment - what do you think about updating how we define `IS_MULTI_ORG`.
> 
> Is there a reason we can't create a helper function to calculate this instead of relying on the environment variables?
> We have some helpers in `license.ts` like `orgHasPremiumFeature` that we could build a wrapper around, which would allow us to define `IS_MULTI_ORG` from the orgs license rather than just their env variables.

We `orgHasPermiumFeature` could really be called `orgIsAllowedToUsePremiumFeature`.  Some enterprise users might not want multi-org functionality from showing up. So we need to rely on IS_MULTI_ORG. We do in `license.ts` run `planHasPremiumFeature` which `orgHasPremiumFeature` calls to do the check whether they are allowed to set the env var.  The problem here was that even if the env var was false, they was nothing stopping people from using that route.

> Not something that needs to be addressed here, it was just a thought I had while reviewing this. Maybe there are protections in place to prevent an org from just setting the ENV variable.

Yes we do have protections in place from setting the ENV var, which I just made more robust and you approved in: https://github.com/growthbook/growthbook/pull/2642.

Issue #2116, opened 2024-02-13
[Feature] Cleanup old agenda documents in Mongo


jdorn:
### Description of Feature

The `eventCreated` agenda job inserts a new document in Mongo for every event that happens in GrowthBook.  Eventually, this can cause the agenda jobs collection to have a huge number of documents and cause queries to time out or fail.

We should add a cronjob to periodically clean up these old documents that aren't needed anymore.  Specifically, we should delete any docs with `nextRunAt` set to `null` that were last run more than 24 hours ago to give plenty of buffer for failure retries.

PR #2646, opened 2024-06-13
Add job to delete old AgendaJobs

tzjames:
### Features and Changes

Cloud currently has 5M AgendaJobs.  The sheer number on cloud and other self hosted installations might be causing queries to run slower than they normally would have.

I have run an aggregation which took about 5s before the index to find a count of the documents that would be deleted on Cloud.  After the index it took around 30ms.  We tried deleteing in the way we have in this PR and it took about 1ms per job.  1000 was taking about 1s.  

We have about 4912012 jobs. 
We create about 33226 a day.
Hence we do a write every 2-3s.
Hence blocking write for 1s every 5 minutes shouldn't have too many knock on effects.

We need to delete 4878786 jobs.
So the cron would need to run approx 5k times.
So it should catch up in about 4878 * 5 /60 / 24 = 17 days. 
Plus another 2 days or so for all the new jobs we will create in that time.

- Closes https://github.com/growthbook/growthbook/issues/2116

### Testing

Search in Mongo 
```
{
    lastFinishedAt: { $lt: new Date(Date.now() - 24 * 3600 * 1000) },
    nextRunAt: null,
  }
```
on AgendaJobs and saw there were 17 documents.
Started dev server and saw in the logs that it deleted 17 documents.
Looked in Mongo and saw that there were 17 fewer AgendaJob documents afterwards.

### DIFF
--- a/packages/back-end/src/init/queue.ts
+++ b/packages/back-end/src/init/queue.ts
@@ -32,6 +32,9 @@ export async function queueInit() {
   refreshFactTableColumns(agenda);
   addSdkWebhooksJob(agenda);
   updateLicenseJob(agenda);
+
+  // Make sure we have index needed to delete efficiently
+  agenda._collection.createIndex({ lastFinishedAt: -1, nextRunAt: -1 });
   deleteOldAgendaJobs(agenda);
 
   await agenda.start();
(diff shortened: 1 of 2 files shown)


packages/back-end/src/init/queue.ts
```diff
@@ -32,6 +33,10 @@ export async function queueInit() {
   addSdkWebhooksJob(agenda);
   updateLicenseJob(agenda);
 
+  // Make sure we have index needed to delete efficiently
+  agenda._collection.createIndex({ lastFinishedAt: 1, nextRunAt: 1 });
```
jdorn, line 37: Add a `.catch()` here

jdorn (approved):

PR #2651, opened 2024-06-13
Make `useFeatureIsOn` reactive to growthbook value

### LINKED ISSUE #2594: [Bug] `useFeatureIsOn` doesn't return updated value when called before SDK has loaded.
### Summary

I am running an A/B experiment on a homepage where the control (old page) is generated server-side and cached on [Varnish](https://varnish-cache.org/), ensuring very fast load times. On the client-side, the control or variant is rendered based on the experiment value.

Since the homepage loads quickly, the `useFeatureIsOn` hook is called before Growthbook's SDK has fully loaded. As a result, even users who should see the variant (new page) are shown the control.

Because this is happening in production, the SDK's response is cached after the first load. Consequently, only when the user refreshes the page are they shown the correct variant.

#### Expected Behaviour

Ideally, the response of `useFeatureIsOn` should update after Growthbook's SDK has loaded, i.e., it should be **_reactive_**. If this happens, the control or variant would automatically render based on the experiment's value.

#### Current Behaviour

I delved deeper into Growthbook's codebase and created a minimal reproduction of the SDK. I found that the `useFeatureIsOn` hook calls the `isOn` method of Growthbook's class and returns the value based on the context. Since the hook is called before the SDK has loaded, the context's value isn't present, resulting in a return value of false.

I am attaching a [link to a CodeSandbox](https://codesandbox.io/p/github/rishavbharti/sdk-js/) with a minimal reproduction of the SDK. Here, I simulate a delay in the SDK class to mimic the delay in fetching the SDK. By observing the logs in the console, one can notice that even though the value is set after 3 seconds, the updated value doesn't appear in the console/UI.
Screenshot 2024-06-01 at 12 22 43 AM

rishavbharti:
### Features and Changes
Makes `useFeatureIsOn` reactive to growthbook value. 

- `useFeatureIsOn` returned `false` when called before growthbook has loaded.
- This change introduces a state variable and a useEffect hook to listen for updates from growthbook and keep `useFeatureIsOn` reactive.

- Closes [https://github.com/growthbook/growthbook/issues/2594](https://github.com/growthbook/growthbook/issues/2594)

### DIFF
--- a/packages/sdk-react/package.json
+++ b/packages/sdk-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@growthbook/growthbook-react",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "license": "MIT",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/packages/sdk-react/src/GrowthBookReact.tsx
+++ b/packages/sdk-react/src/GrowthBookReact.tsx
@@ -79,7 +79,12 @@ export function useFeatureIsOn<
   AppFeatures extends Record<string, any> = Record<string, any>
 >(id: string & keyof AppFeatures): boolean {
   const growthbook = useGrowthBook<AppFeatures>();
-  return growthbook.isOn(id);
+  const [isOn, setIsOn] = React.useState(growthbook.isOn(id));
+  React.useEffect(() => {
+    setIsOn(growthbook.isOn(id));
+  }, [id, growthbook]);
+
+  return isOn;
 }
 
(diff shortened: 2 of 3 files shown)


jdorn:
Do you have a reproducible example using the GrowthBook React SDK?  The Codesandbox link you added to the issue seems to be using a custom SDK class and not the official npm library.

The GrowthBook React SDK already has code to re-render when features load, so I'm wondering if this is a bug in the SDK or with your specific implementation.

jdorn:

PR #2652, opened 2024-06-14
Bandit variation weights code

lukebrawleysmith:
### Features and Changes

This code updates gbstats to be able to receive calls for bandit variation weight updates, and pass the updated weights to our back end. 
### Dependencies

None
### Testing

Unit tested `get_bandit_weights`, which is the method called from `process_experiment_results`.  `get_bandit_weights` calls the `Bandits` class in `tests.py`.  
### Screenshots

None


packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +227,114 @@ def get_risk(mu, sigma) -> List[float]:
...
+    burn: int = 0
+    iteration: int = 0
+    top_two: bool = True
+    prior_distribution: GaussianPrior = field(default_factory=GaussianPrior)
+
+
+class Bandits(object):
+    def __init__(
+        self,
+        stat: Sequence[Union[ProportionStatistic, SampleMeanStatistic]],
```
lukesonnet, line 243: Could it not also be a ratio metric?

lukesonnet, line 243: 1. Could it not also be a ratio metric?
2. Should we call this `stats` since it is an array of statistics?

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +227,114 @@ def get_risk(mu, sigma) -> List[float]:
...
+        p_star /= sum(p_star)
+        n_arms = len(p_star)
+        p_mat = np.repeat(np.expand_dims(p_star, axis=1), n_arms, axis=1)
+        p_mat_t = p_mat.T
+        probs = p_mat * p_mat_t / (1 - p_mat) + p_mat_t * p_mat / (1 - p_mat_t)
+        np.fill_diagonal(probs, 0)
+        return 0.5 * np.sum(probs, axis=1)
+
+    # create config for AB testing from Thompson sampling prior
+    def compute_delta_config(self):
```
lukesonnet, line 332: The goal here is to eventually pass this a series of `EffectBayesianTest` objects to produce effect estimates for the results table?

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +227,114 @@ def get_risk(mu, sigma) -> List[float]:
...
+        self,
+        stat: Sequence[Union[ProportionStatistic, SampleMeanStatistic]],
+        config: BanditConfig,
+    ):
+        self.stat = stat
+        self.config = config
+        self.n_variations = len(stat)
+
+    @property
+    def variation_means(self):
```
lukesonnet, line 251: For these and other methods, is a return type hint feasible? I know numpy can be annoying with typing, but lmk if you can add type hinting.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +227,114 @@ def get_risk(mu, sigma) -> List[float]:
...
+            + self.data_precision * self.variation_means
+        )
+
+    # number of Monte Carlo samples to perform when sampling to estimate weights for the SDK
+    @property
+    def n_samples(self):
+        return int(1e4)
+
+    # function that computes thompson sampling variation weights
+    def compute_variation_weights(self):
```
lukesonnet, line 304: Eventually we'll need to do something like include `p` in some kind of `BanditResult` object that will be consumed by the back-end.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +227,114 @@ def get_risk(mu, sigma) -> List[float]:
         risk_ctrl = float((1.0 - prob_ctrl_is_better) * mn_pos)
         risk_trt = -float(prob_ctrl_is_better * mn_neg)
         return [risk_ctrl, risk_trt]
+
+
+@dataclass
+class BanditConfig(BayesianConfig):
+    burn: int = 0
```
lukesonnet, line 234: Why do we need `burn` and `iteration`?

I feel like this logic is going to be handled by the back-end. We don't need updated sample weights if we are not ready to update weights, so in those cases we won't even make a call out to `gbstats` so we shouldn't need these two arguments.

lukesonnet (commented):
Looks good in general.

Main two comments:

* Config doesn't seem to fit with how we plan on actually getting updated weights. See comment on `BanditConfig`
* We should think about how we are going to (1) package the variation weights for the back-end to consume and (2) how this relates to the effect estimates we're also going to produce for the back-end to consume.

Those have implications for the design here.

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: good point, removed

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: accepted both

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: Actually, I was thinking we would output a collection of `EffectBayesianABTest` objects to produce results.

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: done

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: I have created class `BanditWeights` that has the weights and a string indicating whether or not the weights are updated.  I incorporated a `bandit_weights` attribute of type BanditWeights into DimensionResponse.  I created a unit test that indicates how this can flow through `format_results`.

packages/stats/tests/test_gbstats.py
```diff
@@ -226,230 +260,19 @@
 )
 
 
-class TestDiffDailyTS(TestCase):
```
lukesonnet, line 229: Why are we deleting all of these tests?

packages/stats/gbstats/gbstats.py
```diff
@@ -562,6 +598,46 @@ def process_single_metric(
...
+    analysis: AnalysisSettingsForStatsEngine,
+) -> pd.DataFrame:
+    reduced = preprocess_analysis(rows, var_id_map, metric, analysis)
+    num_variations = reduced.at[0, "variations"]
+    bandit_prior = GaussianPrior(mean=0, variance=float(1e4), proper=True)
+    bandit_config = BanditConfig(prior_distribution=bandit_prior)
+
+    def get_bandit_weights_single(s: pd.Series) -> pd.Series:
+        s_final = pd.Series({"dimension": s["dimension"]})
+        # initialize weights as constant; then just return s if any update checks below fail
```
lukesonnet, line 614: Why return them as constant and equally split? That could be different from what they came in as. I would consider a data structure that allows us to return nothing if some error field is returned. It's better not to return broken data if we don't need to. 

We unfortunately return broken data elsewhere when a test fails (e.g. see the `_default_output` methods), but this is not a great solution we should avoid if we can.

packages/stats/gbstats/gbstats.py
```diff
@@ -562,6 +598,46 @@ def process_single_metric(
...
+        s_final["update_message"] = "did not update"
+        s_final["weights"] = np.full((num_variations,), 1 / num_variations).tolist()
+        # need proportion, samplemean, or ratio;
+        bandit_types = Union[ProportionStatistic, SampleMeanStatistic, RatioStatistic]
+        s0 = variation_statistic_from_metric_row(
+            row=s, prefix="baseline", metric=metric
+        )
+        if isinstance(s0, bandit_types):
```
lukesonnet, line 622: Feels like we should set the update/error message to be more informative if this fails.

packages/stats/gbstats/gbstats.py
```diff
@@ -562,6 +598,46 @@ def process_single_metric(
...
+    bandit_prior = GaussianPrior(mean=0, variance=float(1e4), proper=True)
+    bandit_config = BanditConfig(prior_distribution=bandit_prior)
+
+    def get_bandit_weights_single(s: pd.Series) -> pd.Series:
+        s_final = pd.Series({"dimension": s["dimension"]})
+        # initialize weights as constant; then just return s if any update checks below fail
+        s_final["update_message"] = "did not update"
+        s_final["weights"] = np.full((num_variations,), 1 / num_variations).tolist()
+        # need proportion, samplemean, or ratio;
+        bandit_types = Union[ProportionStatistic, SampleMeanStatistic, RatioStatistic]
```
lukesonnet, line 618: What about `RegressionAdjustedStatistic`? I would also consider defining this in `statistics.py` using our naming convention (`BanditStatistics`). You can then also use it in the `Bandit` init method, rather than defining it in two places.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +240,129 @@ def get_risk(mu, sigma) -> List[float]:
         risk_ctrl = float((1.0 - prob_ctrl_is_better) * mn_pos)
         risk_trt = -float(prob_ctrl_is_better * mn_neg)
         return [risk_ctrl, risk_trt]
+
+
+class Bandits(object):
```
lukesonnet, line 245: What does `object` do here?

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +240,129 @@ def get_risk(mu, sigma) -> List[float]:
...
+    def __init__(
+        self,
+        stats: Sequence[
+            Union[ProportionStatistic, SampleMeanStatistic, RatioStatistic]
```
lukesonnet, line 249: See comment about `bandit_types` in `gbstats.py`

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +240,129 @@ def get_risk(mu, sigma) -> List[float]:
...
+        return int(1e4)
+
+    # function that computes thompson sampling variation weights
+    def compute_variation_weights(self) -> BanditWeights:
+        min_n = 100
+        if any(self.variation_counts < min_n):
+            update_message = "some variation counts smaller than " + str(min_n)
+            p = np.full((self.n_variations,), 1 / self.n_variations).tolist()
+            return BanditWeights(update_message=update_message, weights=p)
+        y = np.random.multivariate_normal(
```
lukesonnet, line 317: The numpy docs says to use https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.multivariate_normal.html#numpy.random.Generator.multivariate_normal instead

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +240,129 @@ def get_risk(mu, sigma) -> List[float]:
...
+        # sample sizes per period
+        period_counts = np.expand_dims(np.sum(variation_counts, axis=1), axis=1)
+        n_variations = variation_counts.shape[1]
+        variation_counts_balanced = np.tile(
+            period_counts / n_variations, (1, n_variations)
+        )
+        counts_diff = variation_counts - variation_counts_balanced
+        return np.sum(counts_diff * variation_means)
+
+    # create config for AB testing from Thompson sampling prior
```
lukesonnet, line 359: Maybe we can add comments here about this not being used?

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +240,129 @@ def get_risk(mu, sigma) -> List[float]:
...
+        row_maxes = np.max(y, axis=1)
+        p = np.mean((y == row_maxes[:, np.newaxis]), axis=0)
+        if self.config.top_two:
+            p = self.top_two_weights(p)
+        update_message = "successfully updated"
+        return BanditWeights(update_message=update_message, weights=p.tolist())
+
+    # function that takes weights for largest realization and turns into top two weights
+    @staticmethod
+    def top_two_weights(p) -> np.ndarray:
```
lukesonnet, line 331: I'd love if there was a simpler way to do this. Maybe there isn't, but we're relying a lot here and in compute_variation_weights on a bunch of numpy recastings of our data that are hard to follow. I wonder if we can't just get the weights in one pass the first time around.

packages/stats/gbstats/gbstats.py
```diff
@@ -319,14 +323,28 @@ def analyze_row(s: pd.Series) -> pd.Series:
...
 # serialized and used to display results in the GrowthBook front-end
 def format_results(
-    df: pd.DataFrame, baseline_index: int = 0
+    df: pd.DataFrame, df_weights: pd.DataFrame, baseline_index: int = 0
 ) -> List[DimensionResponse]:
     num_variations = df.at[0, "variations"]
     results: List[DimensionResponse] = []
+    includes_bandit_weights = df_weights.shape[0] > 0
+    if includes_bandit_weights:
+        df = df.merge(df_weights, on="dimension")
```
lukesonnet, line 332: I'd really love for us to go for an approach that doesn't rely on a merge. This is definitely risky if we change stuff in the future. I think we should consider just calling `get_bandit_weights` inside of `analyze_metric_df` and adding some fields to the returned dataframe there. That way we get rid of:
* calling `preprocess_analysis` twice in both process_analysis and get_bandit_weights
* the unnecessary `var_id_map` and `analysis` arguments to the get_bandit_weights method
* this merge which could break if we change any of the key variables

packages/back-end/src/services/stats.ts
```diff
@@ -76,6 +76,7 @@ export interface MetricSettingsForStatsEngine {
   prior_proper?: boolean;
   prior_mean?: number;
   prior_stddev?: number;
+  decision_metric?: boolean;
```
lukesonnet, line 79: Question for consideration. Should this instead be in AnalysisSettingsForStatsEngine as `bandit_metric?: string` where its an id. Then that why there's no chance of getting two decision metrics that return weights that compete? I'm actually not sure which is better, but just a thought.

packages/stats/tests/test_gbstats.py
lukebrawleysmith: i removed tests on a dummy branch (i thought) to make scrolling back and forth in the file faster, all tests are now restored.

packages/stats/gbstats/gbstats.py
lukebrawleysmith: changed to return `None` if variation counts are too small or any update checks fail

packages/stats/gbstats/gbstats.py
lukebrawleysmith: good call, and done

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: updated, thx

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: I removed `compute_delta_config` from the code base, I see little value for keeping it in prod.

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: I incorporated an empirical approach to make the code more readable.  It gives a little bit different answers than the closed form solution I had before, but not enough to make a difference.  I may run my previous math by you just to see if I'm missing something, because the previous approach is a bit faster, and I can't figure out why empirical would differ from theoretical.

packages/stats/gbstats/gbstats.py
lukebrawleysmith: changed to "test statistic is not of type BanditStatistic"

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: it clutters the screen :)

removed `object`

packages/stats/gbstats/gbstats.py
lukebrawleysmith: good call, i removed the merge

packages/back-end/src/services/stats.ts
lukebrawleysmith: I accepted this change.  I also added test to ensure that for bandit analysis, the decision metric field is non-empty.

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: accepted

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -38,6 +41,19 @@ class EffectBayesianConfig(BayesianConfig):
     prior_effect: GaussianPrior = field(default_factory=GaussianPrior)
 
 
+@dataclass
+class BanditConfig(BayesianConfig):
+    bandit_weights_seed: int = 1
```
lukesonnet, line 46: nit: IMO a seed shouldn't have a fixed default. It should be unset perhaps as a fallback.

packages/stats/gbstats/bayesian/tests.py
lukesonnet: Looks reasonable.

packages/stats/gbstats/gbstats.py
```diff
@@ -258,6 +261,9 @@ def analyze_metric_df(
             df[f"v{i}_uplift"] = None
             df[f"v{i}_error_message"] = None
 
+    if analysis.bandit and analysis.decision_metric == metric.id:
+        df = get_bandit_weights(df, metric, analysis)
```
lukesonnet, line 265: Two notes:

1. Seems like we can do this within `analyze_row`. I'd consider just getting rid of `get_bandit_weights` that takes a `df` and instead just pass in a row (or rename `get_bandit_weights_single`) and then set the values right in the copied row.
2. This also avoids overwiting the `df` like this, which is a little scary.

packages/stats/gbstats/gbstats.py
lukesonnet: Thanks for simplifying this. I think we can go one more step. Left another comment, so resolving this one.

packages/back-end/src/services/stats.ts
lukesonnet: Thanks for making the change. I'm not sure it does, because it still has to loop over a bunch of `DimensionResponse` objects to find the bandit weights. I think we could consider having it live at the `ExperimentMetricAnalysisResult` level so that it's trivially easy to find it.

I think we can resolve this later as well. Consider making the shit and if it's easy to manage with `gbstats.py` then go for it. Otherwise, maybe keep `BanditResponse` but remove it from `DimensionResponse` and we can figure out where it goes in the final integration PR.

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: changed default seed to be 0.  changed `compute_variation_weights` so the seed is used only if positive, so unset is the default.

packages/stats/gbstats/gbstats.py
lukebrawleysmith: As discussed offline, I have now made `get_bandit_weights` standalone, and not nested inside of `analyze_row`.

packages/back-end/src/services/stats.ts
lukebrawleysmith: Thanks for your feedback.  As discussed offline, bandit_weights are now called in `process_experiment_results`.

lukebrawleysmith: Thanks for pointing this out.  As discussed offline, I now call `get_bandit_weights` directly from `process_experiment_results`.

lukesonnet (commented):
pls remove docs and do some more in-app QA (just try to do some analyses) and then I'll approve.

added 2 nits.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +241,234 @@ def get_risk(mu, sigma) -> List[float]:
         risk_ctrl = float((1.0 - prob_ctrl_is_better) * mn_pos)
         risk_trt = -float(prob_ctrl_is_better * mn_neg)
         return [risk_ctrl, risk_trt]
+
+
+class Bandits:
+    def __init__(
+        self,
+        stats: Dict,  # keys are 0, 1, 2, etc. mapping to periods; values are lists of length n_variations of summary_statistics
```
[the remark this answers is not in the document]
lukebrawleysmith: done

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +241,234 @@ def get_risk(mu, sigma) -> List[float]:
...
+    def weights_array(self) -> np.ndarray:
+        if self.config.weight_by_period:
+            return np.tile(
+                np.expand_dims(self.period_weights, axis=1), (1, self.num_variations)
+            )
+        else:
+            variation_sample_sizes = np.sum(self.counts_array, axis=0)
+            if any(variation_sample_sizes == 0):
+                error_string = (
+                    "Need at least 1 observation per variation to weight by period."
```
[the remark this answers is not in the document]
lukebrawleysmith: changed error message to "Need at least 1 observation per variation per period if not weighting by period."

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -223,3 +242,230 @@ def get_risk(mu, sigma) -> List[float]:
...
+        )
+        # construct the variance of the weighted mean
+        sample_mean_variances = np.sum(
+            sample_mean_variances_by_period * (self.weights_array) ** 2, axis=0
+        )
+        # scale by number of counts to get back to distributional variance
+        return sample_mean_variances * self.variation_counts
+
+    @property
+    def sample_mean_statistics(self) -> List[SampleMeanStatistic]:
```
lukesonnet, line 361: unused?

packages/back-end/src/services/stats.ts
```diff
@@ -67,6 +67,13 @@ export interface AnalysisSettingsForStatsEngine {
   max_dimensions: number;
 }
 
+export interface BanditSettingsForStatsEngine {
+  var_names: string[];
+  var_ids: string[];
```
bryce-fitzsimons, line 72: variations

packages/back-end/types/stats.d.ts
```diff
@@ -60,8 +60,15 @@ export type ExperimentMetricAnalysis = {
   }[];
 }[];
 
+export type BanditResponse = {
+  banditUpdateMessage?: string;
+  banditWeights?: number[];
```
bryce-fitzsimons, line 65: try to return weights at all times (even if 50/50), but keep it optional for cases when stats engine can't reliably make an assumption

packages/back-end/src/services/stats.ts
bryce-fitzsimons: do we need both of these? any of these? the response object doesn't reference these variation ids at all.

packages/back-end/src/services/stats.ts
```diff
@@ -67,6 +67,13 @@ export interface AnalysisSettingsForStatsEngine {
...
+  decision_metric: string;
```
bryce-fitzsimons, line 73: mid

packages/back-end/src/services/stats.ts
```diff
@@ -67,6 +67,13 @@ export interface AnalysisSettingsForStatsEngine {
...
+  bandit_weights_seed: number;
```
bryce-fitzsimons, line 74: will be set as a random number/hash on every "create analysis and update weights" event, will be stored in the snapshot

packages/back-end/types/stats.d.ts
```diff
@@ -60,8 +60,15 @@ export type ExperimentMetricAnalysis = {
   }[];
 }[];
 
+export type BanditResponse = {
```
bryce-fitzsimons, line 63: todo: figure out how this gets persisted / consumed by GB.
- immediately update experiment model's variation weights, purge SDK cache
- probably needs to get piped to an event log or similar for weight changes (srm, maybe for vis)

packages/back-end/src/services/stats.ts
```diff
@@ -67,6 +67,13 @@ export interface AnalysisSettingsForStatsEngine {
...
+}
```
bryce-fitzsimons, line 75: need to send up changelog of (per phase)
```
{
  dateRun,
  weights: []
}[]
```

bryce-fitzsimons: events where weights changed (but don't actually need the weights)

lukesonnet (approved):
looks good! I QA to make sure it doesn't affect common current analyses, and I was able to finally get some more understanding of how the simulations work. Thanks!

packages/back-end/src/services/stats.ts
lukebrawleysmith: We use these in the query for constructing results going into the stats engine.  
https://github.com/growthbook/growthbook/blob/4b319072d263e77daef9bf9bd1eb17aba28cd359/packages/stats/gbstats/gbstats.py#L106

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: removed

PR #2662, opened 2024-06-19
Migrate Segment Model to BaseClass

mknowlton89:
### Features and Changes

This PR migrates the Segment Model from being the one-off custom written model, to one that extends the BaseClass.

In doing so, we've updated all of the calls to this model as it now uses the `model.create` or `model.find` style of calls.

Additionally, this PR introduces changes to the `BaseModels` `_find()` and `_findOne()` calls as `Segments` can be defined via Mongo, or via the config.yml for sefl-hosted organizations.

Also changed in this PR is the fact that we're now using the `owner` field set within the `SegmentForm` when creating the segment. Previously, it was being set to the `context.userName` which currently always appears to be an empty string.

In my testing, I did find a bug with the `getSegmentUsage` that I've filed [here](https://github.com/growthbook/growthbook/issues/2695) and plan to address in a follow-up PR.

### Testing
- [x] Ensure that you can create, update, and delete segments via the UI if the org isn't using the config file and all fields are updated/created/deleted successfully
- [x] Ensure that when viewing the segments page, if the org isn't using a config file, it grabs segments from mongo
- [x] Ensure that when viewing the segments page, if the org IS using a config file, it grabs the segments from the config file and not mong
- [x] Ensure that if the org IS using a config file, the `create` CTA isn't visible on the segments page
- [x] Ensure that if the org IS using a config file, the dateUpdated column isn't displayed (to make this work, we're setting the dateUpdated to the current date when we grab the segments from the config file, so this would be confusing to show)
- [x] Ensure that if the org IS using a config file, the user can not edit or delete the segments
- [x] Ensure that if the org ISNT using a config file, they can update and delete the segments
- [x] Ensure that if the org IS using a config file, other config file defined resources (data sources, metrics, dimensions) are being pulled from the config file correctly. (This works, except for metrics, which is expected behavior)


packages/back-end/src/init/config.ts
```diff
@@ -239,8 +231,10 @@ export function getConfigSegments(organization: string): SegmentInterface[] {
       id,
       ...d,
       organization,
-      dateCreated: null,
-      dateUpdated: null,
+      dateCreated: new Date(), //temporary
+      dateUpdated: new Date(), //temporary
+      // dateCreated: null,
+      // dateUpdated: null,
```
mknowlton89, line 237: This is an issue - the `BaseModel` type requires dateCreated and dateUpdated, but config defined resources do not have these.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -344,16 +351,38 @@ export abstract class BaseModel<
       ...query,
       organization: this.context.org.id,
     };
-    const cursor = this._dangerousGetCollection().find(queryWithOrg);
+    let rawDocs;
+
+    if (this.useConfigFile()) {
+      const docs =
+        this.getConfigDocuments().filter((doc) =>
+          evalCondition(doc, queryWithOrg)
```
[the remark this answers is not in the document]
mknowlton89: Sounds great - thanks, @Kevin-Chant!

packages/back-end/src/services/organizations.ts
```diff
@@ -828,21 +823,18 @@ export async function importConfig(
...
-          const existing = await findSegmentById(k, organization.id);
+          const existing = await context.models.segments.getById(k);
           if (existing) {
             const updates: Partial<SegmentInterface> = {
               ...s,
             };
             delete updates.organization;
 
-            await updateSegment(k, organization.id, updates);
+            await context.models.segments.updateById(k, updates);
```
jdorn, line 833: Use `update` instead of `updateById` to avoid an unnecessary db lookup

packages/front-end/pages/segments/index.tsx
```diff
@@ -296,7 +296,9 @@ const SegmentPage: FC = () => {
                           expandable={true}
                         />
                       </td>
-                      <td>{s.dateUpdated ? ago(s.dateUpdated) : ""}</td>
+                      {canStoreSegmentsInMongo ? (
+                        <td>{s.dateUpdated ? ago(s.dateUpdated) : ""}</td>
```
jdorn, line 300: nit: can get rid of the ternary now since `dateUpdated` is no longer optional

packages/back-end/src/models/SegmentModel.ts
```diff
@@ -1,108 +1,76 @@
...
+    return getConfigSegments(this.context.org.id);
   }
+  public async getByDataSource(
+    datasourceId: string
+  ): Promise<SegmentInterface[]> {
+    const allSegments = await this.getAll();
 
-  await SegmentModel.deleteOne({ id, organization });
-}
+    return allSegments.filter((segment) => segment.datasource === datasourceId);
```
jdorn, line 50: Instead of filtering in javascript, we should call `this._find` and pass in a query.

packages/back-end/src/models/SegmentModel.ts
```diff
@@ -1,108 +1,76 @@
...
-  // If using config.yml & the org doesn't have the env variable STORE_SEGMENTS_IN_MONGO,
-  // immediately return the list from there
-  if (usingFileConfigForSegments()) {
-    throw new Error("Cannot update. Segments are being managed by config.yml");
+  protected async beforeCreate() {
+    if (this.useConfigFile()) {
+      throw new Error(
+        "Cannot create. Segments are being managed by config.yml"
+      );
+    }
```
jdorn, line 58: This logic of blocking writes when using config should probably live in BaseModel instead.

packages/back-end/src/models/BaseModel.ts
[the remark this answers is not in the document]
Kevin-Chant: Update here - after talking more about it with Bryce we opted to make it an optional argument defaulting to `{}`. So actually no changes necessary here anymore :)

packages/back-end/src/routers/segment/segment.controller.ts
```diff
@@ -276,13 +262,13 @@ export const deleteSegment = async (
...
   const { org } = context;
-  const segment = await findSegmentById(id, org.id);
+  const segment = await context.models.segments.getById(id);
 
   if (!segment) {
     throw new Error("Could not find segment");
   }
 
-  await deleteSegmentById(id, org.id);
+  await context.models.segments.deleteById(id);
```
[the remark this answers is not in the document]
mknowlton89: Sounds good - I like option 2.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -722,14 +755,15 @@ export abstract class BaseModel<
...
-      if (!projects.some((p) => p.id === obj.project)) {
+      if (!projects || !projects.some((p) => p.id === obj.project)) {
         throw new Error("Invalid project");
       }
     }
     // Resources with multiple projects
     else if ("projects" in obj && obj.projects && Array.isArray(obj.projects)) {
       const projects = await this.context.getProjects();
       if (
+        !projects ||
```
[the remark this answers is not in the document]
mknowlton89: So, I'm still a little fuzzy on it, but, my IDE was throwing an error here that `projects` could be null. [Link to error](https://github.com/growthbook/growthbook/actions/runs/9748833851/job/26904678074)

I think it may be because of how we're instantiating `rawDocs` as an undefined value within `_find()` and then later reassigning it, either via the config.yml file or the Mongo results.

If I update the code to something like:

```ts
  protected async _find(
    query: FilterQuery<Omit<z.infer<T>, "organization">> = {},
    {
      sort,
      limit,
      skip,
      bypassReadPermissionChecks,
    }: {
      sort?: Partial<
        {
          [key in keyof Omit<z.infer<T>, "organization">]: 1 | -1;
        }
      >;
      limit?: number;
      skip?: number;
      bypassReadPermissionChecks?: boolean;
    } = {}
  ) {
    const queryWithOrg = {
      ...query,
      organization: this.context.org.id,
    };

    if (this.useConfigFile()) {
      const docs =
        this.getConfigDocuments().filter((doc) =>
          evalCondition(doc, queryWithOrg)
        ) || [];

      if (!docs.length) return [];

      sort &&
        docs.sort((a, b) => {
          for (const key in sort) {
            const typedKey = key as keyof z.infer<T>;
            const sortDir = sort[typedKey] as 1 | -1;

            if (a[typedKey] < b[typedKey]) return -1 * sortDir;
            if (a[typedKey] > b[typedKey]) return 1 * sortDir;
          }
          return 0;
        });

      const migrated = docs.map((d) =>
        this.migrate(this._removeMongooseFields(d))
      );
      const filtered = bypassReadPermissionChecks
        ? migrated
        : await this.filterByReadPermissions(migrated);

      if (!skip && !limit) return filtered;

      return filtered.slice(skip || 0, limit ? (skip || 0) + limit : undefined);
    } else {
      const cursor = this._dangerousGetCollection().find(queryWithOrg);
      sort &&
        cursor.sort(
          sort as {
            [key: string]: 1 | -1;
          }
        );
      const docs = await cursor.toArray();

      if (!docs.length) return [];

      const migrated = docs.map((d) =>
        this.migrate(this._removeMongooseFields(d))
      );
      const filtered = bypassReadPermissionChecks
        ? migrated
        : await this.filterByReadPermissions(migrated);

      if (!skip && !limit) return filtered;

      return filtered.slice(skip || 0, limit ? (skip || 0) + limit : undefined);
    }
  }
  ```
  
  The error goes away.
  
  I haven't found the exact smoking gun yet as to why, reading through the code, it looks like we'd never return null, and the generated signature doesn't appear to change.

File: packages/back-end/src/models/BaseModel.ts [hunk 9]
```diff
@@ -198,7 +198,11 @@ export abstract class BaseModel<
     // Do nothing by default
   }
   protected async beforeCreate(doc: z.infer<T>, writeOptions?: WriteOptions) {
-    // Do nothing by default
+    if (this.useConfigFile()) {
+      throw new Error(
+        "Cannot create. Segments are being managed by config.yml"
+      );
+    }
```
jdorn, line 205: This should live outside `beforeCreate`.  Otherwise, when child classes override this method to add their own logic, they might forget to call `super()` and would unintentionally stop this check from running.

The same places in the code that call `beforeCreate` , `beforeUpdate`, and `beforeDelete` should also do this check.

jdorn (approved):

PR #2667, opened 2024-06-20
Add a new date format for String targeting attributes

msamper:
### Features and Changes

Adds a new Date format for string targeting attributes. Selecting the date format for an attribute changes the language for `less than` and `greater than` operators to `before` and `after` and provides a date picker rather than a free text field for providing a targeting value.

### Testing

Create a new string attribute and set its `String Format` to `Date string`
Try creating a feature flag and target by the new Date string attribute.
Equal to, after, after or on, before, and before or on should all trigger the new date picker input. The other operators should not.

### Screenshots

Screenshot 2024-07-15 at 1 38 38 PM
Screenshot 2024-07-15 at 1 39 17 PM


packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -425,6 +444,18 @@ export default function ConditionInput(props: Props) {
                       containerClassName="col-sm-12 col-md mb-2"
                       required
                     />
+                  ) : ["string", "secureString"].includes(attribute.datatype) &&
+                    attribute.format === "date" &&
+                    !["$regex", "$notRegex"].includes(operator) ? (
+                    <Field
+                      type="datetime-local"
+                      value={format(new Date(value), "yyyy-MM-dd'T'HH:mm")}
```
msamper, line 452: Jeremy recommends using the same field that we're using for regular strings. Determine field type depending on whether the format is date or not. Don't format the value

jdorn (approved):

PR #2669, opened 2024-06-20
Defines userIdTypes when creating auto metrics

mknowlton89:
### Features and Changes

@Auz Noted that when we're generating auto metrics, we're not defining the `userIdTypes`. This results in a bug in a few scenarios.

If, with GA4, GA4 isn't tracking a `user_id` field, just the `user_pseudo_id`, then when we're getting the Metric from the DB and upgrading it within `upgradeMetricDoc`, we're defaulting the `userIdTypes` array to `["user_id", "anonymous_id"]`.

And then, if a user attempts to edit the SQL of an auto metric, it's requiring both `user_id` and `anonymous_id`, even though we previously knew that the org wasn't tracking `user_id`'s. 

Now, when building the auto metric, we're building the `userIdTypes` array and passing that in when creating the metrics so we don't have to use the fall back.

^ NOTE
^ I've tested this with a GA4 instance, but I need to do additional testing.


mknowlton89:
Update as of 7/1 - I'm waiting on this [PR](https://github.com/growthbook/growthbook/pull/2623) to land and then I'll pull in the changes there to this branch and update accordingly.

jdorn (approved):

PR #2670, opened 2024-06-21
Added namespace ids and label

### LINKED ISSUE #2666: [Bug] Add separate namespace id and display fields.
### Summary

The namespace `name` is used like both an id and display.  This is bad for a few reasons:
1. Changing the display name breaks all references to the namespace
2. The name might not be URL-safe, so creating a namespace with `#` in it for example, will break our API endpoints that are expecting the namespace id in the URL

We should:
1. Add separate `id` and `display` fields for namespaces instead of having a single "name" which acts as both.  `id` would be auto-generated and URL-safe.
3. Add a workaround to allow editing/deleting existing namespaces with non-url safe names

Auz:
Splits label from the id for namespaces so that the label can be edited without reassigning users or causing reference errors. 

Screenshot 2024-06-20 at 11 18 39 PM

Fixes #2666


packages/front-end/components/Features/ExperimentRefSummary.tsx
```diff
@@ -183,9 +186,19 @@ export default function ExperimentRefSummary({
...
-              </span>
+              <Link href={`/namespaces`}>
+                <span className="mr-1 border px-2 py-1 bg-light rounded">
+                  {namespaces?.find((n) => n.name === phase.namespace.name)
+                    ?.label || (
+                    <span
+                      className="italic text-danger"
+                      title="this namespace is not found"
+                    >
+                      phase.namespace.name
```
jdorn, line 197: Wrap this in curly braces

jdorn:
The other part of that linked issue is that if someone created a namespace that has a special character in the name like `#`, the API calls to edit/delete it fail.  It might be as simple as using `encodeURIComponent` on the front-end when constructing the API calls.

packages/front-end/components/Features/ExperimentRefSummary.tsx
Auz: done

jdorn (approved):

PR #2672, opened 2024-06-21
setting a google font as the main font so it works across browsers/OS better

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Layout/Layout.module.scss
+++ b/packages/front-end/components/Layout/Layout.module.scss
@@ -15,7 +15,7 @@ $topbarheight: 60px;
   color: #fff;
   overflow: hidden;
   overflow-y: auto;
-  font-family: Poppins, sans-serif;
+  font-family: Inter, sans-serif;
   z-index: 1000;
 }
 .sidebarlogo {
--- a/packages/front-end/styles/global.scss
+++ b/packages/front-end/styles/global.scss
@@ -18,9 +18,9 @@ main {
 }
 
 body {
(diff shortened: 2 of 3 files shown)


packages/front-end/components/Layout/Layout.module.scss
```diff
@@ -15,7 +15,7 @@ $topbarheight: 60px;
   color: #fff;
   overflow: hidden;
   overflow-y: auto;
-  font-family: Poppins, sans-serif;
+  font-family: Inter, sans-serif;
```
jdorn, line 18: If we're no longer using Poppins anywhere, can we remove the import in `global.scss`?

gazzdingo: i think we can remove poppins i wanted to double check that the Kari
but yes we can remove it if we are not going to use it.

gazzdingo: it can be removed :)

packages/front-end/styles/global.scss
```diff
@@ -6,7 +6,7 @@
 
 @import "bootstrap/scss/bootstrap.scss";
 @import url("https://fonts.googleapis.com/css2?family=Poppins:ital,wght@0,300;0,400;0,500;0,600;0,700;1,400&display=swap");
```
gazzdingo, line 8:
```suggestion
```

jdorn (approved):

jdorn:
Fixed in #2692 in a different way (using `next/font`)

Issue #2661, opened 2024-06-18
[Feature] Make "Proxy Host URL" Optional on Cloud


jdorn:
### Description of Feature

When using a proxy server in front of GrowthBook Cloud, there are 2 types of network requests that happen:
1. The proxy **pulls** SDK payloads from GrowthBook using a stale-while-revalidate caching strategy
2. GrowthBook **pushes** SDK payloads to the proxy with a webhook whenever payloads change

The **push** method (2) is a nice-to-have that enables faster rollouts (by invalidating cache), but it's not required to function.  In some network setups, it's not possible to expose the proxy server to accept webhooks.

Currently, we require you to enter your proxy URL in GrowthBook and we always try to send **push** webhooks.  We should make this field optional and skip the webhook if it's empty.  We should also update our docs / tooltips in the app to make this setting and the implications clear.

![image](https://github.com/growthbook/growthbook/assets/1087514/3c1ae7a2-88a7-4d66-83fa-4b8f18773e9b)

PR #2679, opened 2024-06-25
Make proxy webhooks optional

bryce-fitzsimons:
### Features and Changes

- Closes https://github.com/growthbook/growthbook/issues/2661

Makes proxy webhooks optional for both cloud and self-hosted customers. The proxy webhook is not actually needed for the proxy to minimally function; its purpose is to make rollouts instantaneous (push-based). Without the webhook, the proxy still stays somewhat up-to-date via SWR.

image

image

### DIFF
--- a/packages/back-end/src/models/SdkConnectionModel.ts
+++ b/packages/back-end/src/models/SdkConnectionModel.ts
@@ -74,6 +74,7 @@ const SDKConnectionModel = mongoose.model<SDKConnectionInterface>(
 );
 
 function addEnvProxySettings(proxy: ProxyConnection): ProxyConnection {
+  // TOOD: remove when done testing!!!
   return proxy;
   if (IS_CLOUD) return proxy;
 
--- a/packages/back-end/src/models/SdkConnectionModel.ts
+++ b/packages/back-end/src/models/SdkConnectionModel.ts
@@ -74,8 +74,6 @@ const SDKConnectionModel = mongoose.model<SDKConnectionInterface>(
 );
 
 function addEnvProxySettings(proxy: ProxyConnection): ProxyConnection {
-  // TOOD: remove when done testing!!!
-  return proxy;
   if (IS_CLOUD) return proxy;
 
   return {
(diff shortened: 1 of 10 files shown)


packages/front-end/pages/sdks/[sdkid].tsx
```diff
@@ -400,15 +400,17 @@ export default function SDKConnectionPage() {
...
-            <Button
-              color="link"
-              className="btn-sm"
-              onClick={async () => {
-                await mutate();
-              }}
-            >
-              <BsArrowRepeat /> re-check
-            </Button>
+            connection.proxy.host ? (
```
Comment (jdorn) [inline packages/front-end/pages/sdks/[sdkid].tsx:403]: Do we need this change? When the proxy uses SWR to fetch the payload, it should set connection.connected properly, so this is still useful for making sure the proxy's fetch is still working, even if webhooks are disabled

Comment (bryce-fitzsimons) [reply packages/front-end/pages/sdks/[sdkid].tsx:403]: I think we still need this. This change hides the "re-check" button. Otherwise we'd be asking `/sdk-connections/${id}/check-proxy` to ping the proxy `/healthcheck` endpoint. But if host is missing, then it's a network error.

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -464,7 +480,10 @@ export async function testProxyConnection(
     return {
       status: statusCode || 0,
       body: body || "",
-      error: e.message || "Failed to connect to Proxy server",
+      error:
+        e?.code === "ECONNREFUSED"
+          ? "Failed to connect to proxy server (ECONNREFUSED)"
+          : e.message,
```
jdorn, line 486: Do we want a fallback here in case `e.message` is empty?  Not sure this would ever happen in practice, so maybe not worth worrying about.

jdorn (approved):

packages/back-end/src/models/SdkConnectionModel.ts
bryce-fitzsimons: added one just in case

PR #2683, opened 2024-06-25
fixed revision data

gazzdingo:
### Features and Changes
the API response for features was hardcoding the user and the comment to an empty string.
this PR fetches the feature revision based on the version and sets the data to the correct user, date created and comment 
Also fixed was a small bug where the comment on the revision was not set if the revision was not part of a review now


packages/back-end/src/services/features.ts
```diff
@@ -977,7 +978,15 @@ export function getApiFeatureObj({
...
   });
-
+  const revision = await getRevision(
+    feature.organization,
+    feature.id,
+    feature.version
+  );
+  const publishedBy =
+    revision?.publishedBy?.type === "api_key"
+      ? "API"
```
gazzdingo, line 988: @jdorn not sure if we can get this data but would we wait the API key name here?

packages/back-end/src/api/features/listFeatures.ts
```diff
@@ -21,15 +21,17 @@ export const listFeatures = createApiRequestHandler(listFeaturesValidator)(
...
-
     return {
-      features: filtered.map((feature) =>
-        getApiFeatureObj({
-          feature,
-          organization: req.organization,
-          groupMap,
-          experimentMap,
-        })
+      features: await Promise.all(
```
jdorn, line 25: I typically don't like doing an unbounded number of Mongo queries in parallel like this.  If someone has a lot of features, this could be doing dozens or hundreds of queries

gazzdingo: would you suggest that we fetch all at once then assign during code

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -380,18 +381,29 @@ async function logFeatureUpdatedEvent(
...
+    current.organization,
+    current.id,
+    current.version
+  );
+  const previousRevision = await getRevision(
+    previous.organization,
+    previous.id,
+    previous.version
+  );
+  const currentApiFeature = await getApiFeatureObj({
```
gazzdingo, line 394:
```suggestion
  const currentApiFeature = getApiFeatureObj({
```

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -380,18 +381,29 @@ async function logFeatureUpdatedEvent(
...
     feature: current,
     organization: context.org,
     groupMap,
     experimentMap,
+    revision: currentRevision,
   });
-  const previousApiFeature = getApiFeatureObj({
+  const previousApiFeature = await getApiFeatureObj({
```
gazzdingo, line 401:
```suggestion
  const previousApiFeature = getApiFeatureObj({
```

jdorn (approved):

PR #2685, opened 2024-06-25
Add support for big saved groups to JS sdk

Kevin-Chant:
### Features and Changes

Splits the JS sdk logic out of #2616 so that it can be merged first. The new logic is all backwards compatible, so this will allow us to release the new 1.0.2 version of the sdk ahead of time and prevent any issues with the proxy server.

### Testing
Setup:
1. Checkout the `kc/saved-groups-improvements` branch to grab the code that generates the new style of payload
2. Run `yarn link` in the `sdk-js` package and start running `yarn dev`
3. Clone the https://github.com/growthbook/kevin-sdk-testing/tree/main/sdk-js repo and run `yarn dev` there

Changing the saved group & feature definitions in your app should result in the correct feature evaluation in the next app.

### DIFF
--- a/packages/sdk-js/src/GrowthBook.ts
+++ b/packages/sdk-js/src/GrowthBook.ts
@@ -204,10 +204,10 @@ export class GrowthBook<
       this._ctx.experiments = data.experiments;
       this._updateAllAutoExperiments();
     }
-    if (data.idLists) {
-      this._ctx.idLists = {};
-      Object.entries(data.idLists).forEach(([key, valueList]) => {
-        this._ctx.idLists![key] = new Set(valueList);
+    if (data.savedGroups) {
+      this._ctx.savedGroups = {};
+      Object.entries(data.savedGroups).forEach(([key, valueList]) => {
+        this._ctx.savedGroups![key] = new Set(valueList);
       });
     }
     this.ready = true;
@@ -464,10 +464,10 @@ export class GrowthBook<
       );
       delete data.encryptedExperiments;
     }
-    if (data.encryptedIdLists) {
-      data.idLists = JSON.parse(
+    if (data.encryptedSavedGroups) {
+      data.savedGroups = JSON.parse(
         await decrypt(
-          data.encryptedIdLists,
+          data.encryptedSavedGroups,
           decryptionKey || this._ctx.decryptionKey,
           subtle
         )
@@ -1158,7 +1158,7 @@ export class GrowthBook<
     return evalCondition(
       this.getAttributes(),
       condition,
-      this._ctx.idLists || {}
+      this._ctx.savedGroups || {}
     );
   }
 
--- a/packages/sdk-js/src/GrowthBook.ts
+++ b/packages/sdk-js/src/GrowthBook.ts
@@ -977,7 +977,7 @@ export class GrowthBook<
             const evaled = evalCondition(
               evalObj,
               parentCondition.condition || {},
-              this._ctx.idLists || {}
+              {}
             );
             if (!evaled) {
               // blocking prerequisite eval failed: feature evaluation fails
@@ -1326,13 +1326,7 @@ export class GrowthBook<
           }
 
           const evalObj = { value: parentResult.value };
-          if (
-            !evalCondition(
-              evalObj,
-              parentCondition.condition || {},
-              this._ctx.idLists || {}
-            )
-          ) {
+          if (!evalCondition(evalObj, parentCondition.condition || {}, {})) {
             process.env.NODE_ENV !== "production" &&
               this.log("Skip because prerequisite evaluation fails", {
                 id: key,
--- a/packages/sdk-js/src/GrowthBook.ts
+++ b/packages/sdk-js/src/GrowthBook.ts
@@ -204,6 +204,9 @@ export class GrowthBook<
       this._ctx.experiments = data.experiments;
       this._updateAllAutoExperiments();
     }
+    if (data.idLists) {
+      this._ctx.idLists = data.idLists;
+    }
     this.ready = true;
     this._render();
   }
@@ -457,6 +460,15 @@ export class GrowthBook<
         )
       );
       delete data.encryptedExperiments;
+    }
+    if (data.encryptedIdLists) {
+      data.idLists = JSON.parse(
+        await decrypt(
+          data.encryptedIdLists,
+          decryptionKey || this._ctx.decryptionKey,
+          subtle
+        )
+      );
     }
     return data;
   }
@@ -964,7 +976,8 @@ export class GrowthBook<
             const evalObj = { value: parentResult.value };
             const evaled = evalCondition(
               evalObj,
-              parentCondition.condition || {}
+              parentCondition.condition || {},
+              this._ctx.idLists || {}
             );
             if (!evaled) {
               // blocking prerequisite eval failed: feature evaluation fails
@@ -1139,7 +1152,11 @@ export class GrowthBook<
   }
 
   private _conditionPasses(condition: ConditionInterface): boolean {
-    return evalCondition(this.getAttributes(), condition);
+    return evalCondition(
+      this.getAttributes(),
+      condition,
+      this._ctx.idLists || {}
+    );
   }
 
   private _isFilteredOut(filters: Filter[]): boolean {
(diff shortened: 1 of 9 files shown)


packages/sdk-js/src/GrowthBook.ts
```diff
@@ -964,7 +976,8 @@ export class GrowthBook<
             const evalObj = { value: parentResult.value };
             const evaled = evalCondition(
               evalObj,
-              parentCondition.condition || {}
+              parentCondition.condition || {},
+              this._ctx.idLists || {}
```
jdorn, line 980: Always pass in `{}` here.  Prerequisites should not use saved groups

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1309,7 +1326,13 @@ export class GrowthBook<
           }
 
           const evalObj = { value: parentResult.value };
-          if (!evalCondition(evalObj, parentCondition.condition || {})) {
+          if (
+            !evalCondition(
+              evalObj,
+              parentCondition.condition || {},
+              this._ctx.idLists || {}
```
jdorn, line 1333: Same here.  Always pass in `{}`

packages/sdk-js/src/mongrule.ts
```diff
@@ -15,26 +16,27 @@ const _regexCache: { [key: string]: RegExp } = {};
 // The top-level condition evaluation function
 export function evalCondition(
   obj: TestedObj,
-  condition: ConditionInterface
+  condition: ConditionInterface,
+  idLists: IdLists
```
jdorn, line 20: Even though this is marked as required here, it's worth doing a line at the top of this function: `idLists = idLists || {}`.

Since this function is being exported, others might be relying on the existing behavior and not everyone is using Typescript or even if they are, they might have a `@ts-ignore` comment or something here.

packages/sdk-js/src/mongrule.ts
```diff
@@ -176,22 +184,26 @@ function evalOperatorCondition(
     case "$in":
       if (!Array.isArray(expected)) return false;
       return isIn(actual, expected);
+    case "$inGroup":
+      return isIn(actual, idLists[expected] || []);
+    case "$notInGroup":
+      return !isIn(actual, idLists[expected] || []);
```
jdorn, line 190: As far as naming, it's a little weird to use "group" here, but call them "idLists" in the context settings.  I know why it's like this, but now might be a good opportunity to pick better language.  Once we commit to this in the SDK, we can't change it later easily.

packages/sdk-js/test/json.test.ts
```diff
@@ -99,11 +99,11 @@ describe("json test suite", () => {
 
   it.each((cases as Cases).evalCondition)(
     "evalCondition[%#] %s",
-    (name, condition, value, expected) => {
+    (name, condition, value, expected, idLists = {}) => {
```
jdorn, line 102: When we change `cases.json`, we need to change a few other things at the same time:

1. Update the `Cases` type above in this file
2. Bump the spec version at the top of the `cases.json` file
3. On `build-your-own.mdx`, bump the spec version at the top and add it to the changelog at the bottom
4. On `build-your-own.mdx`, update the cases.json documentation at the bottom where it lists out all inputs/outputs

packages/sdk-js/test/cases.json
```diff
@@ -2828,6 +2828,60 @@
...
+      { "group_id": [1, 2, 3] }
+    ],
+    [
+      "$notInGroup passes for unknown group id",
+      {
+        "id": { "$notInGroup": "unknowngroup_id" }
+      },
+      { "id": 1 },
+      true,
+      { "group_id": [1, 2, 3] }
```
jdorn, line 2884: Other test cases to add:
- Mix of strings/numbers for list values

Also, we should add test cases for `run` and `feature` to make sure idLists are being properly passed through from those functions.  Specifically, using it in a feature force rule, using it in an experiment directly, and using it in an experiment rule for a feature.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -964,7 +977,8 @@ export class GrowthBook<
             const evalObj = { value: parentResult.value };
             const evaled = evalCondition(
               evalObj,
-              parentCondition.condition || {}
+              parentCondition.condition || {},
+              {}
```
bryce-fitzsimons, line 981: Why do we pass an empty context (with no saved groups) here?

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1309,7 +1327,7 @@ export class GrowthBook<
           }
 
           const evalObj = { value: parentResult.value };
-          if (!evalCondition(evalObj, parentCondition.condition || {})) {
+          if (!evalCondition(evalObj, parentCondition.condition || {}, {})) {
```
bryce-fitzsimons, line 1330: same question

packages/sdk-js/src/mongrule.ts
```diff
@@ -15,26 +16,29 @@ const _regexCache: { [key: string]: RegExp } = {};
 // The top-level condition evaluation function
 export function evalCondition(
   obj: TestedObj,
-  condition: ConditionInterface
+  condition: ConditionInterface,
+  savedGroups: SavedGroupsValues
```
bryce-fitzsimons, line 20: should this be optional?
`savedGroups?: ...`

packages/shared/src/util/features.ts
```diff
@@ -744,7 +744,7 @@ export function evalDeterministicPrereqValue(
   const parsedCondition = getParsedPrereqCondition(condition);
   if (!parsedCondition) return "fail";
   const evalObj = { value: value };
-  const pass = evalCondition(evalObj, parsedCondition);
+  const pass = evalCondition(evalObj, parsedCondition, {});
```
bryce-fitzsimons, line 747: save for a different PR? but is this safe to pass {} for context?

packages/sdk-js/src/mongrule.ts
Kevin-Chant: I don't think so. It's required in all contexts where `condition` might include a group operation, so I think the prereqs and such were just special cases and the general case is that it should be required

packages/shared/src/util/features.ts
Kevin-Chant: Good point; this import is version locked so it won't get the new signature right away huh. I think I just changed it here bc I'd yarn linked the package and it was complaining about missing an arg

packages/back-end/package.json
```diff
@@ -34,7 +34,7 @@
     "@dqbd/tiktoken": "^1.0.7",
     "@google-cloud/bigquery": "5",
     "@google-cloud/storage": "^5.20.5",
-    "@growthbook/growthbook": "^1.0.1",
+    "@growthbook/growthbook": "1.0.1",
```
Kevin-Chant, line 37: Prevented all the packages from upgrading to 1.0.2 until we're ready. I'll swap these deps back to use `^` when I add the supporting code and bump the version req

packages/sdk-js/src/mongrule.ts
bryce-fitzsimons: you have this line: `savedGroups = savedGroups || {};`. Should that be removed then?

packages/shared/src/util/features.ts
bryce-fitzsimons: i guess it goes back to whether that arg should be optional or not. If not, then you probably need to do both at the same time.

packages/sdk-js/src/mongrule.ts
bryce-fitzsimons: also by not making it optional, there is a small but real risk of temporarily breaking libs (such as `shared`) which depend on evalCondition having 2 args.

Kevin-Chant: No, that was a suggestion from Jeremy that some users might not be using typescript or might have `// @ts-ignore` statements that prevent them from automatically catching the new signature.
Do you think there's a risk of breaking the libs even with the change to the `package.json`s which swaps `^1.0.1` to `1.0.1`? I would think that specifying an exact version will keep things in a safe intermediary state until the new code is in place.

bryce-fitzsimons (approved):

jdorn (approved):

PR #2689, opened 2024-06-26
Merge audit collection into notification event collection

romain-growthbook:
This PR merges the audit collection into the notification event collections. It is the first step toward merging the two notions into one and making it possible to tap into any audit event as part of the notification frame work.

### Goals
* No data migration
* Backward compatibility with the current backend

### Implementation assumptions

All the implementation assumptions below are supported by static typing (explained later)

The audits are stored as entries in the notification event collection.
* Event names are normalized, i.e. `feature.create` in audit is mapped to `feature.created`. Normalization happens for all 3 of: `create`, `update` and `delete` events. Other events are untouched.
* A `auditData` payload is added (and required) to all notification event payload that map to an existing audit event.

For instance, notification events for `feature.created` MUST have a `auditData`.

* Notification events that do not map to an audit event do not have a `auditData` payload.

### For insertion:
* All existing notification event creation call sites that map to an existing audit evevent are expected to provide a `auditData`.
* Explicit `auditLog` creation endpoint are only valid for audit events that do not have an existing notification event payload.
* In this case, the saved audit payload is converted into a notification event payload.

### For read/fetch:
Fetch are done both on the legacy audit collection and on the event collection based on event name and the presence of a `auditData` payload. This should make sure that legacy event notification without a `auditData` payload never surface into the system in this call chain. Audit entries stored in the notification event collection are converted back to a audit interface using the associated `auditData`.

### Details about the type checker implementation

All the above assumption are validated at type-check. This is possible because all event/resources mappings are done via static declaration.

The typing can be a little confusing because it relies on the following pattern:
```typescript
type Foo = "foo" | "gni"

// This is : { attribute: "foo" | "gni" }
type NotAMap = { attribute: Foo }

type MapTemplate<T> = T extends "foo" | "gni"  ? { attribute: T } : never;

// This is: { attribute: "foo" } | { attribute: "gni" }
type Map = MapTemplate<Foo>
```

The other tricky thing is to support mappings single events to structured resource/event, for instance:
```typescript
{ "feature.created": { object: "feature", event: "feature.create" } }
```
And make the mapping match e.g. between structured event and audit payloads. 

This is achieved via type mappings using the technique above and ad-hoc code mappers for which the type is forced. Care should be put on making sure that these implementations actually implement what they claim to do. See PR comments for those spots.

I have tried to keep the naming consistent and isolate the complex typescript into their own files, documenting the expected types as well.

Since this typing element will not be visible on the finished code, here are some screenshots:

### Missing `auditData` at a notification event call site:
Screenshot 2024-06-27 at 9 56 19 AM

### Invalid call to `auditLog` for an event that has a proper notification event payload:
Screenshot 2024-06-27 at 9 56 28 AM

### TODO:
- [ ] Add invalid typed expression with `// @ts-expect-error` to validate typing assumptions and make sure they never break.


packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+              };
+              details?: string;
+            };
+          };
+        }
+    : never
+  : never;
+
+// This is used to make audit interface to event interface.
+// Tt has type: { "experiment.update": { entity: "experiment", event: "experiment.updated" } & ...
```
romain-growthbook, line 40:
```suggestion
// It has type: { "experiment.update": { entity: "experiment", event: "experiment.updated" } & ...
```

packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+                id: string;
+              };
+              details?: string;
+            };
+          };
+        }
+    : never
+  : never;
+
+// This is used to make audit interface to event interface.
```
romain-growthbook, line 39:
```suggestion
// This is used to mapaudit interface to event interface.
```

packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+// It has type: { "experiment.update": { entity: "experiment", event: "experiment.updated" } & ...
```
romain-growthbook, line 40:
```suggestion
// It has type: { "experiment.update": { entity: "experiment", event: "experiment.updated", ... } & ...
```

packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+          entity: { object: R; id: string; name?: string };
+          event: `${R}.${EventAuditMap<R, E>}`;
+          parent?: { object: R; id: string };
+        };
+      }
+    : never
+  : never;
+
+// This is used to map notification event to audit event.
+// Tt has type: { "experiment.updated": experiment.update } & ...
```
romain-growthbook, line 133:
```suggestion
// It has type: { "experiment.updated": { event: "experiment.update", ... } } & ...
```

packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+    : never
+  : never;
+
+// This is used to map notification event to audit event.
+// It has type: { "experiment.updated": { event: "experiment.update", ... } } & ...
+type EventAuditMappings = UnionToIntersection<
+  EventAuditMappingTemplate<AuditEventResource>
+>;
+
+export const eventAuditMappings = ({
```
romain-growthbook, line 138: The type of this function is forced. It should be proof-read to make sure it actually implements what it says it does.

packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+  ? Event
+  : Event extends "created"
+  ? "create"
+  : Event extends "updated"
+  ? "update"
+  : Event extends "deleted"
+  ? "delete"
+  : Event;
+
+export const eventAudit = <
```
romain-growthbook, line 95: The type of this function is forced. It should be proof-read to make sure it actually implements what it says it does.

packages/back-end/src/util/legacyAudit/maps.ts
```diff
@@ -0,0 +1,169 @@
...
+  : never;
+
+// This is used to mapaudit interface to event interface.
+// It has type: { "experiment.update": { entity: "experiment", event: "experiment.updated", ... } & ...
+// and excludes the events already defined.
+type AuditEventMappings = UnionToIntersection<
+  AuditEventMappingTemplate<AuditEventResource>
+>;
+
+export const auditEventMappings = ({
```
romain-growthbook, line 46: The type of this function is forced. It should be proof-read to make sure it actually implements what it says it does.

packages/back-end/src/util/legacyAudit/base.ts
```diff
@@ -0,0 +1,93 @@
...
+  ? Event
+  : Event extends "create"
+  ? "created"
+  : Event extends "update"
+  ? "updated"
+  : Event extends "delete"
+  ? "deleted"
+  : Event;
+
+export const auditNotificationEvent = <
```
romain-growthbook, line 31: The type of this function is forced. It should be proof-read to make sure it actually implements what it says it does.

PR #2690, opened 2024-06-26
[hotfix] remove reliance on variation key to construct redirect API payload

msamper:
### Features and Changes


lukesonnet (approved):
approving because before I think it basically relied on v.key to === i to work, so this makes sense as a hotfix.

PR #2696, opened 2024-06-27
[docs] Add experiment.warning to event webhooks

lukesonnet:


romain-growthbook (approved):
Fine by me but could we do a live test before merging this?

lukesonnet:
I can't seem to get an `experiment.warning` message to fire in my testing, so I guess we should hold off until we figure out why.

PR #2697, opened 2024-06-27
Add select all environments box when copying feature rules to environments

bryce-fitzsimons:
### Features and Changes

Adds this checkbox:
image

### DIFF
--- a/packages/front-end/components/Features/CopyRuleModal.tsx
+++ b/packages/front-end/components/Features/CopyRuleModal.tsx
@@ -95,20 +95,24 @@ export default function CopyRuleModal({
         </span>{" "}
         from <span className="text-indigo h5">{environment}</span> to...
       </div>
-      <div className="mt-3">
-        <label className="cursor-pointer hover-underline py-1 px-1 mb-0">
-          <input
-            type="checkbox"
-            id="select_all"
-            className="position-relative mr-2"
-            style={{ top: "2px" }}
-            onChange={toggleAll}
-            checked={allToggled}
-          />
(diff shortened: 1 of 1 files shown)


AtakanColak:
Thanks 👍

packages/front-end/components/Features/CopyRuleModal.tsx
```diff
@@ -32,13 +32,29 @@ export default function CopyRuleModal({
...
+  const envs = filteredEnvironments.map((e) => e.id);
 
   const ruleTxt = rules.length === 1 ? "rule" : "rules";
 
   const [selectedEnvironments, setSelectedEnvironments] = useState<
     Record<string, boolean>
   >({});
 
+  const allToggled =
+    Object.values(selectedEnvironments).filter((v) => v).length === envs.length;
```
gazzdingo, line 44: the filter doesn't look like it is doing anything here i think we can remove that
```suggestion
    const allToggled = Object.values(selectedEnvironments).length === envs.length;
```

gazzdingo: wait do we even need object.values here?

bryce-fitzsimons: nah it's needed, otherwise it includes items like `prod: false`

gazzdingo: wait it does its a boolean value :)

gazzdingo (approved):
Nit: selected environments is a little misleading i thought that it would only contain environments that are selected not all environments with a boolean flag.

PR #2699, opened 2024-06-28
Added the ability to edit tag names after creation

### LINKED ISSUE #2694: [Feature] Add the ability to edit tag names
Currently tag names are not editable once created.

Auz:
Added a label field to the structure of tags internally which allows tags to be updated.  

Screenshot 2024-06-27 at 11 10 44 PM

Also update tag creation/editing/deleting to log to the audit logs

Fixes #2694


packages/back-end/src/models/TagModel.ts
```diff
@@ -27,16 +27,38 @@ function toTagInterface(doc: TagDocument | null): TagInterface[] {
...
+    organization,
+  });
+  if (!doc) return null;
+
+  if (!doc.tags.includes(tag)) return null;
+
+  const settings = doc.settings || {};
+  return {
+    id: tag,
+    color: settings[tag]?.color || "#029dd1",
```
romain-growthbook, line 49: This is duplicated above. It should this be factored out and placed as a constant on top of the file.

packages/back-end/src/models/TagModel.ts
```diff
@@ -107,6 +130,42 @@ export async function removeTag(organization: string, tag: string) {
       $pull: { tags: tag },
     }
   );
+  await TagModel.updateOne(
+    {
+      organization,
+    },
+    {
+      $unset: { [`settings.${tag}`]: 1 },
```
romain-growthbook, line 138: Can this be combined with the previous call?

packages/back-end/src/models/TagModel.ts
```diff
@@ -107,6 +130,42 @@ export async function removeTag(organization: string, tag: string) {
...
+export async function updateTag(
+  organization: string,
+  tag: string,
+  color: string,
+  description: string,
+  label: string
+) {
+  if (description.length > 256) {
+    description = description.substring(0, 256);
+  }
```
romain-growthbook, line 152: nit: You don't need the conditional here:
```
> "aabb".substring(0, 256);;
'aabb'
```

packages/back-end/src/routers/tag/tag.controller.ts
```diff
@@ -33,15 +44,92 @@ export const postTag = async (
...
+  const { id } = req.params;
+  const existing = await getAllTags(context.org.id);
+  const matchingId = existing.find((tag) => tag.id === id);
+  const matchingLabel = existing.find((tag) => tag.label === label);
+  if (!matchingId) {
+    throw new Error("Tag not found");
+  }
+  if (matchingLabel) {
+    throw new Error("A Tag with this name already exists");
+  }
```
romain-growthbook, line 106: What if we are updating the same tag with the same label but changing `color`? I don't think that this should fail.

packages/front-end/components/Tags/TagsModal.tsx
```diff
@@ -90,8 +97,6 @@ export default function TagsModal({
           </div>
         </div>
         <Field
-          // @ts-expect-error TS(2783) If you come across this, please fix it!: 'name' is specified more than once, so this usage ... Remove this comment to see the full error message
-          name="Name"
```
romain-growthbook, line 94: Nice!

romain-growthbook (commented):
Mostly looked at the backend. Mostly one issue when editing re: label uniqueness.

This could be split into separate back-end/front-end PRs with the band-end going first potentially.

Also, it would be great to add tests to validate the assumptions made in the checks. I started adding some over at: https://github.com/growthbook/growthbook/tree/main/packages/back-end/test/api

Happy to pair-up to adapt them to these endpoints!

PR #2701, opened 2024-06-28
Add support for Fact Segments on upgraded Segment Model

mknowlton89:
### Features and Changes

### Testing

**Creation**
- [x] Ensure you can create a simple SQL based Segment without errors
- [x] Ensure you can create a simple Fact Based Segment with no filters
- [x] Ensure you can create a simple Fact Based Segment WITH filters
- [x] Ensure that when creating a Fact Based Segment, the userIdTypes is a subset of the Fact Table's types, not the data sources
- [x] Ensure that you can create a SQL based Segment even if you start off building a Fact Based segment, and switch back to a SQL based Segment
- [x] Ensure that you can create a Fact based Segment even if you start off building a SQL Based segment, and switch back to a Fact based Segment

**Editing**
- [x] Ensure that you can edit a previously created SQL based Segment
- [x] Ensure that you can edit a previously created legacy SQL based Segment
- [x] Ensure that you can edit a previously created Fact based Segment

**Deleting**
- [x] Ensure that you can delete a previously created SQL based Segment
- [x] Ensure that you can delete a previously created legacy SQL based Segment
- [x] Ensure that you can delete a previously created Fact based Segment

**Usage**
- [x] Ensure that you can apply a legacy SQL Segment to an experiment without issue
- [x] Ensure that you can apply a new SQL Segment to an experiment without issue
- [x] Ensure that you can apply a Fact Segment to an experiment without issue
- [x] Ensure that the experiment results don't change for Segments whether its SQL or Fact (as long as the two Segments are identical)


packages/back-end/src/routers/segment/segment.validators.ts
```diff
@@ -0,0 +1,135 @@
...
+//   sql: z.string().optional(),
+//   factTableId: z.string(),
+//   filters: z.array(z.string()),
+// });
+
+// // Union of the create segment schemas
+// export const createSegmentValidator = z.union([
+//   createSqlSegmentSchema,
+//   createFactSegmentSchema,
+// ]);
```
mknowlton89, line 128: As seen, I've explored trying to create a `baseSegmentSchema` and then use that to create `sqlSegmentSchema` and `factSegmentSchema` via a union, but haven't been able to build it out successfully, so for now, I just added the new properties and made the sql, factTableId, and filter properties optional. Definitely not type-safe yet, but drawing a line in the sand here to move forward with the UI.

packages/back-end/src/api/segments/listSegments.ts
```diff
@@ -12,7 +9,7 @@ import { listSegmentsValidator } from "../../validators/openapi";
 
 export const listSegments = createApiRequestHandler(listSegmentsValidator)(
   async (req): Promise<ListSegmentsResponse> => {
-    const segments = await findSegmentsByOrganization(req.organization.id);
+    const segments = await req.context.models.segments.getAll();
 
     // TODO: Move sorting/limiting to the database query for better performance
```
mknowlton89, line 14: TODO: Now that we're using the BaseModel, do the filtering/limiting there via the use of the find() method, with the limit/sort query.

packages/back-end/src/routers/segment/segment.controller.ts
```diff
@@ -153,7 +167,11 @@ export const postSegment = async (
     userIdType,
     name,
     sql,
+    id: uniqid("seg_"),
```
mknowlton89, line 170: Does this need to be here?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3420,6 +3423,62 @@ AND event_name = '${eventName}'`,
     );
   }
 
+  //MKTODO: Explore how can I refactor getFactTableCTE to work for metrics or fact segments
+  private getFactTableCTEForSegments({
```
mknowlton89, line 3427: @jdorn Any ideas on how to clean this up a bit? This is very similar to `getFactTableCTE` that takes in metrics.

I did explore combining this with the function above, but it turned into a case where so many of the function props had to be optional, requiring a lot of `if` checks.

Likewise, it got a bit complicated when building the `SQLVars`.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3485,7 +3487,7 @@ export default abstract class SqlIntegration
     return filterValues;
   }
   // Get a Fact Table CTE for multiple fact metrics that all share the same fact table
-  private getFactTableCTE({
+  private getFactMetricCTE({
```
mknowlton89, line 3490: Renamed this, so it would be more specific to building the fact table CTE for metrics.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3611,6 +3613,62 @@ export default abstract class SqlIntegration
     );
   }
 
+  // Get a Fact Table CTE for segments based on fact tables
+  private getFactSegmentCTE({
```
mknowlton89, line 3617: This is very similar to `getFactMetricCTE`, but instead of taking in a metric, it takes in a segment.

I did explore combining these two into a single method `getFactTableCTE` and having it handle either metrics or a segment, but the logic got pretty twisty, and it seems like we handle dates a bit different for `metrics` vs `segments`.

This code definitely doesn't follow the DRY principles, but I don't feel like I have enough context to optimize these functions.

packages/back-end/src/api/openapi/schemas/Segment.yaml
```diff
@@ -24,4 +23,12 @@ properties:
   dateCreated:
     type: string
   dateUpdated:
-    type: string
\ No newline at end of file
+    type: string
+  type:
+    type: string
```
jdorn, line 28: This should have enum options here

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -188,7 +188,12 @@ export async function getDefinitions(req: AuthRequest, res: Response) {
       };
     }),
     dimensions,
-    segments,
+    segments: segments.map((segment) => {
+      return {
+        ...segment,
+        type: segment.type || "SQL", // if no type, its an old sql type segment
+      };
+    }),
```
jdorn, line 196: This logic belongs in a migration within the Segment model. That way, everywhere that segments are used, you can rely on this property existing.

packages/back-end/src/services/segments.ts
```diff
@@ -7,8 +7,11 @@ export function toSegmentApiInterface(segment: SegmentInterface): ApiSegment {
     name: segment.name,
     owner: segment.owner || "",
     identifierType: segment.userIdType || "user_id",
-    query: segment.sql,
+    query: segment.sql || "",
     datasourceId: segment.datasource || "",
+    type: segment.sql ? "SQL" : "FACT",
```
jdorn, line 12: Use the type field here instead of inferring.

packages/front-end/components/Segments/FactSegmentForm.tsx
```diff
@@ -0,0 +1,210 @@
...
+          helpText="This list has been filtered to only show data sources that have at least one Fact Table built on top of it"
+        />
+        <SelectField
+          label="Identifier"
+          required
+          value={form.watch("userIdType")}
+          onChange={(v) => form.setValue("userIdType", v)}
+          placeholder="Select an identifier"
+          options={
+            datasource?.settings.userIdTypes?.map((userIdType) => ({
```
jdorn, line 128: These options should actually come from the selected fact table, not the data source. It's possible a fact table doesn't select all userIdType columns.

As such, probably want to put the fact table select above this

packages/back-end/src/models/SegmentModel.ts
```diff
@@ -47,4 +47,10 @@ export class SegmentModel extends BaseClass {
   ): Promise<SegmentInterface[]> {
     return await this._find({ datasource: datasourceId });
   }
+
```
jdorn, line 50: We should add a bunch of validation to this model (customValidation hook). If type is FACT, make sure fact table isn't empty, etc.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3783,12 +3841,54 @@ export default abstract class SqlIntegration
...
+    factTableMap: FactTableMap,
     sqlVars?: SQLVars
   ) {
     // replace template variables
-    const segmentSql = sqlVars
-      ? compileSqlTemplate(segment.sql, sqlVars)
-      : segment.sql;
+    let segmentSql = "";
+
+    if (segment.sql) {
```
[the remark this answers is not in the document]
jdorn: Make type required instead of optional and add a migration to the segment model to set it if it's missing. Then, you can rely on that here.

The reason you want to use type instead of inferring is because someone could in theory, start with a SQL segment and then switch the type to FACT and the SQL field wouldn't get wiped out in Mongo. So it might be possible for SQL to be set, but it should still be treated like a fact table.

Yes, you'll still need additional empty checks for the SQL/fact table fields but that's ok.

packages/back-end/src/models/SegmentModel.ts
```diff
@@ -47,4 +51,49 @@ export class SegmentModel extends BaseClass {
...
+  }
+
+  protected async customValidation(segment: SegmentInterface): Promise<void> {
+    if (segment.type === "SQL") {
+      if (!segment.sql) {
+        throw new Error(
+          `${segment.name} is a SQL type Segment, but contains no SQL value`
+        );
+      }
+      if (segment.factTableId) {
```
jdorn, line 73: This should not be an error. A user might switch the type of a segment without clearing out previous fields. Also remove the check for filters below this and the check for SQL in the type=fact branch

jdorn (approved):

PR #2712, opened 2024-07-02
Fix power doc typos

lukesonnet:
We must escape dollar signs in our documentation.

before
Screenshot 2024-07-02 at 12 01 00 PM

after
Screenshot 2024-07-02 at 12 00 55 PM


royalfig (approved):
Is it because the `$` interferes with the katex rendering?

PR #2718, opened 2024-07-03
[docs] Add new language to sequential testing for multi-arm

lukesonnet:
Adds new language for multi-arm tests and rearranges order to put more important info first.


august-growthbook (approved):
LGTM

PR #2720, opened 2024-07-03
Added additional auth options for Presto

Auz:
Added Custom Authentication HTTP header option, and 'none', which lets users handle authentication outside of GrowthBook. 

Screenshot 2024-07-03 at 3 49 09 PM


lukesonnet:
Added a presto-client version bump to 1.1.0, updated some typing, and added query statistics.

lukesonnet (approved):

PR #2721, opened 2024-07-04
Add SDK Webhook payload formatters, update docs

bryce-fitzsimons:
### Features and Changes

- Provide some purpose-driven formatters for SDK webhooks. The current format is not conducive for key/val storage (for example when using an Edge SDK) or for setting caches.

- Add a migration (`sendPayload` field is deprecated)

- Update the docs (sdk webhooks and edge sdks)

image

image

### DIFF
--- a/packages/back-end/src/jobs/sdkWebhooks.ts
+++ b/packages/back-end/src/jobs/sdkWebhooks.ts
@@ -179,10 +179,19 @@ async function runWebhookFetch({
 
   const timestamp = Math.floor(date.getTime() / 1000);
 
-  let body: string | undefined = undefined;
+  let body: string | undefined;
+  const standardBody = JSON.stringify({
+    type: "payload.changed",
+    timestamp: date.toISOString(),
+    data: { payload },
+  });
+  let invalidValue: never;
 
   if (method !== "GET") {
     switch (payloadFormat) {
+      case "none":
+        body = undefined;
+        break;
       case "standard-no-payload":
         body = JSON.stringify({
           type: "payload.changed",
@@ -193,12 +202,12 @@ async function runWebhookFetch({
         body = payload;
         break;
       case "standard":
+        body = standardBody;
+        break;
       default:
-        body = JSON.stringify({
-          type: "payload.changed",
-          timestamp: date.toISOString(),
-          data: { payload },
-        });
+        body = standardBody;
+        invalidValue = payloadFormat;
+        logger.error(`Invalid webhook payload format: ${invalidValue}`);
     }
   }
 
--- a/packages/back-end/src/jobs/sdkWebhooks.ts
+++ b/packages/back-end/src/jobs/sdkWebhooks.ts
@@ -34,6 +34,7 @@ type SDKWebhookJob = Job<{
   webhookId: string;
   retryCount: number;
 }>;
+const sendPayloadFormats = ["standard", "sdkPayload"];
 
 const fireWebhooks = trackJob(
   SDK_WEBHOOKS_JOB_NAME,
@@ -161,14 +162,13 @@ async function runWebhookFetch({
   const signingKey = webhook.signingKey;
   const headers = webhook.headers || "";
   const method = webhook.httpMethod || "POST";
-  // todo: better inference based on sendPayload
   const payloadFormat = webhook.payloadFormat || "standard";
   const organizationId = webhook.organization;
   const requestTimeout = 30000;
   const maxContentSize = 1000;
 
   const sendPayload =
-    method !== "GET" && ["standard", "sdkPayload"].includes(payloadFormat);
+    method !== "GET" && sendPayloadFormats.includes(payloadFormat);
 
   const date = new Date();
   const signature = createHmac("sha256", signingKey)
@@ -292,7 +292,7 @@ export async function fireSdkWebhook(
     let payload = "";
     const sendPayload =
       webhook.httpMethod !== "GET" &&
-      ["standard", "sdkPayload"].includes(webhook.payloadFormat ?? "standard");
+      sendPayloadFormats.includes(webhook.payloadFormat ?? "standard");
     if (sendPayload) {
       const environmentDoc = webhookContext.org?.settings?.environments?.find(
         (e) => e.id === connection.environment
--- a/packages/back-end/src/jobs/sdkWebhooks.ts
+++ b/packages/back-end/src/jobs/sdkWebhooks.ts
@@ -161,10 +161,14 @@ async function runWebhookFetch({
   const signingKey = webhook.signingKey;
   const headers = webhook.headers || "";
   const method = webhook.httpMethod || "POST";
-  const sendPayload = webhook.sendPayload;
+  // todo: better inference based on sendPayload
+  const payloadFormat = webhook.payloadFormat || "standard";
   const organizationId = webhook.organization;
   const requestTimeout = 30000;
   const maxContentSize = 1000;
+
+  const sendPayload =
+    method !== "GET" && ["standard", "sdkPayload"].includes(payloadFormat);
 
   const date = new Date();
   const signature = createHmac("sha256", signingKey)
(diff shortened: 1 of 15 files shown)


packages/back-end/src/models/WebhookModel.ts
```diff
@@ -39,7 +41,18 @@ type WebhookDocument = mongoose.Document & WebhookInterface;
...
+  const oldDoc = doc.toJSON<WebhookDocument>();
+  const newDoc = omit(oldDoc, ["__v", "_id", "sendPayload"]);
+  if (!oldDoc.payloadFormat) {
+    if (oldDoc.httpMethod === "GET") {
+      newDoc.payloadFormat = "none";
+    } else if (oldDoc.sendPayload) {
+      newDoc.payloadFormat = "standard";
+    } else {
+      newDoc.payloadFormat = "standard-no-payload";
+    }
```
romain-growthbook, line 53: I would put this into its own stand-alone method so that it doesn't get erased by mistake. Also, could we save the migrated payload after applying it?

packages/back-end/src/models/WebhookModel.ts
```diff
@@ -39,7 +41,18 @@ type WebhookDocument = mongoose.Document & WebhookInterface;
 const WebhookModel = mongoose.model<WebhookInterface>("Webhook", webhookSchema);
 
 function toInterface(doc: WebhookDocument): WebhookInterface {
-  return omit(doc.toJSON<WebhookDocument>(), ["__v", "_id"]);
+  const oldDoc = doc.toJSON<WebhookDocument>();
+  const newDoc = omit(oldDoc, ["__v", "_id", "sendPayload"]);
+  if (!oldDoc.payloadFormat) {
```
romain-growthbook, line 46: Should you also check that `sendPayload` is not set?

packages/back-end/src/jobs/sdkWebhooks.ts
```diff
@@ -161,29 +161,42 @@ async function runWebhookFetch({
   const signingKey = webhook.signingKey;
   const headers = webhook.headers || "";
   const method = webhook.httpMethod || "POST";
-  const sendPayload = webhook.sendPayload;
+  // todo: better inference based on sendPayload
```
romain-growthbook, line 164: What do you mean?

packages/back-end/src/jobs/sdkWebhooks.ts
```diff
@@ -161,29 +161,42 @@ async function runWebhookFetch({
...
+  const payloadFormat = webhook.payloadFormat || "standard";
   const organizationId = webhook.organization;
   const requestTimeout = 30000;
   const maxContentSize = 1000;
 
+  const sendPayload =
+    method !== "GET" && ["standard", "sdkPayload"].includes(payloadFormat);
```
romain-growthbook, line 171: This should ensure that the keys in the array are actually payload formats, in case they change in the future etc.
```suggestion
    method !== "GET" && (["standard", "sdkPayload"] as const).includes(payloadFormat);
```

packages/back-end/src/jobs/sdkWebhooks.ts
```diff
@@ -161,29 +161,42 @@ async function runWebhookFetch({
...
+        data: { payload },
+      });
+    } else if (payloadFormat === "standard-no-payload") {
+      body = JSON.stringify({
+        type: "payload.changed",
+        timestamp: date.toISOString(),
+      });
+    } else if (payloadFormat === "sdkPayload") {
+      body = payload;
+    }
```
romain-growthbook, line 198: You could use the extensible switch form here to make sure the code is updated whenever a new format is added:
```typescript
let invalidFormat: never;
switch (payloadFormat) {
  case "standard"
      body = JSON.stringify({
        type: "payload.changed",
        timestamp: date.toISOString(),
        data: { payload },
      });
      break;
      ...
    
  // Should have exhausted all possibilities here:
  default:
    invalidFormat = format;
     // Should never happen:
   throw new Error(`Invalid format: ${invalidFormat}`);
}

File: packages/back-end/src/jobs/sdkWebhooks.ts [hunk 6]
```diff
@@ -277,7 +290,10 @@ export async function fireSdkWebhook(
     }
 
     let payload = "";
-    if (webhook.sendPayload) {
+    const sendPayload =
+      webhook.httpMethod !== "GET" &&
+      ["standard", "sdkPayload"].includes(webhook.payloadFormat ?? "standard");
```
romain-growthbook, line 295:
```suggestion
      (["standard", "sdkPayload"] as const).includes(webhook.payloadFormat ?? "standard");
```

packages/back-end/src/jobs/sdkWebhooks.ts
```diff
@@ -393,7 +409,24 @@ export async function fireGlobalSdkWebhooks(
...
+      } = webhook;
+      let format = payloadFormat;
+      if (!format) {
+        if (method === "GET") {
+          format = "none";
+        } else if (sendPayload) {
+          format = "standard";
+        } else {
+          format = "standard-no-payload";
+        }
```
romain-growthbook, line 428: Aren't we guaranteeing that this never happens b/c of the migration in the interface conversion function?

packages/back-end/types/webhook.d.ts
```diff
@@ -11,14 +11,22 @@ export interface WebhookInterface {
...
   httpMethod?: WebhookMethod;
 }
 
 export type WebhookMethod = "GET" | "PUT" | "POST" | "DELETE" | "PURGE";
 
+export type WebhookPayloadFormat =
+  | "standard"
+  | "standard-no-payload"
+  | "sdkPayload"
+  | "none";
```
romain-growthbook, line 28: It would be nice to derive this from a `const` array from the model so we have one single source of truth for all these values.

romain-growthbook (commented):
Some comments!

packages/back-end/src/jobs/sdkWebhooks.ts
gazzdingo: you could also make this a const above like it makes it easier to read compaired to the inline const 
```
const canSendPayloadFormats =  ["standard", "sdkPayload"];
const sendPayload =
    method !== "GET" && canSendPayloadFormats.includes(payloadFormat);
```

packages/back-end/src/models/WebhookModel.ts
bryce-fitzsimons: sure we can refactor. as for saving the model on migration, we typically don't do that unless the conversion is expensive

packages/back-end/src/models/WebhookModel.ts
bryce-fitzsimons: don't think so. If the payloadFormat exists, it should take precidence

packages/back-end/src/jobs/sdkWebhooks.ts
bryce-fitzsimons: old comment, forgot to remove

packages/back-end/src/jobs/sdkWebhooks.ts
bryce-fitzsimons: yeah, we can remove this

bryce-fitzsimons: actually no, we need this. This is for global webhooks which are set via env var. The interface conversion is only for standard webhook models.

packages/back-end/src/jobs/sdkWebhooks.ts
bryce-fitzsimons: I'm not comfortable throwing an error here. I'd much rather default to something sensible. But I can migrate this to a switch statement

packages/back-end/types/webhook.d.ts
bryce-fitzsimons: Going to leave this alone for now. I find the plain types much more readable, plus its in line with how we generally do it.

packages/back-end/src/jobs/sdkWebhooks.ts
romain-growthbook: The error is never sent. You can put a sensitive returned value if you prefer but if the code ever reach it, something is wrong in the underlying assumptions.

More details:

You have two situations:
* The `case` above have exhausted all possible types for `format`, there is no possible type for `format` so it is in fact of type `never` in this branch and it works but `default` is never executed
* The `case` above are missing one variant for `format` an the type-checker raises an error.

Here's a typescript playground with a smaller version for illustration: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAYg9nKBeKAiARgGwIaqgHzQHMA7AS1QChKB6GqAFQAsyBnKOE6NqAWzdZkSRKNigBjbKwgAuSuM6tgUIQt5hMEYBAAKAJzjiIrVgEFMmeIhQAKAB5QZsBAEpkAPigBvSlCiblIQA3bEwyABMANVCAV1koLiCIPQBuaj9WAHcyYHEmKHs3Hz8-SWk0LFw5EpK9LRi9EjQcqEypUShK1DSa8IgAM2wYzGBqmpUSELCo2OgUOx7x4CYDTISINYBRPQM9GwADAElJ0IioKbinABIvYNOZzDiAX32XRafKD9p6Zh5ObnYRAQ4TkChISigYAMRhM5ksCGQBQcTisbiQnmK-i0EymEWij3iiWSaV8UCyOTySKKpNKUmgGBwqDGNTqwAaTVQLTa7DEXUWNTK9NIFGZtXqjWaym5HT5NKgfUGw1Gcr8d2m+LiiIWKqgy1W6y2OzgeyOJ2m51m11uZrxsxeb1JHyeQA

The idea of this pattern is to make sure that the typechecker can detect unhandled cases in the future. Type for instance to add a case to the type `Foo` in the above playground example..

bryce-fitzsimons: We shouldn't throw an error here. Format can be something outside of the allowed types because one entrypoint is raw parsing of an environment variable (for global webhooks). They don't go through the standard zod validators and migration.

I added a missing case guard though, similar to your suggestion.

gazzdingo (approved):

PR #2724, opened 2024-07-08
Add setting to limit data used for test queries

### LINKED ISSUE #2723: [Feature] Add possibility to set up date range for Fact Table test query
### Description of Feature

When we try to save edited Fact Table, Growthbook tests query for **one year**, we would like to set custom date ranges for test queries. 
Now we often can't save updates queries, because our queries are too heave with data for one year :(

### Related Issues

None

lukesonnet:
When generating a test query, se use 365 days of data (or `IMPORT_LIMIT_DAYS` if the user has set that variable) as the lookback window.

This only applies:
* If we are running a test query that has a `{{startDate}}` template variable in it, OR
* Whenever we save a fact table, since we run a test query to get the fact table columns and such

This can be very expensive.

This PR:
* changes that default to 30
* allows customers to set an org level number of days to lookback.

30 is more than enough for most test queries. Why not go shorter? Well we use this test query to find all of the metadata for fact tables so we should err on the longer side by default and let those who have large tables change the setting themselves.

Closes #2723 . We could consider instead a setting right in the fact table form, but I'd rather pull it out of there for now and just have a single setting.

## Screenshots

Settings page
image

A test query with `{{startDate}}`
Screenshot 2024-07-08 at 2 53 17 PM


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -750,16 +752,24 @@ export default abstract class SqlIntegration
...
   getTestQuery(
     query: string,
     templateVariables?: TemplateVariables,
+    testDays?: number,
     limit: number = 5
   ): string {
     const startDate = new Date();
-    startDate.setDate(startDate.getDate() - IMPORT_LIMIT_DAYS);
+    startDate.setDate(
+      startDate.getDate() - (testDays ?? DEFAULT_TEST_QUERY_DAYS)
```
romain-growthbook, line 771: Should you validate that `testDays` is positive?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -750,16 +752,24 @@ export default abstract class SqlIntegration
...
+      DEFAULT_TEST_QUERY_DAYS,
+      1
+    );
   }
 
   getTestQuery(
     query: string,
     templateVariables?: TemplateVariables,
+    testDays?: number,
     limit: number = 5
```
romain-growthbook, line 767: Nit: It's not very developper friendly to have multiple optional arguments without switching to a single hash/record parameter. Otherwise, you have to pass `testDays` explicitly to be able to pass `limit` etc. Another benefit of a single hash/record parameter is to have named arguments.

packages/front-end/pages/settings/index.tsx
```diff
@@ -350,29 +350,7 @@ const GeneralSettingsPage = (): React.ReactElement => {
...
-                      form.setValue("defaultDataSource", v)
-                    }
-                    isClearable={true}
-                    placeholder="Select a data source..."
-                    helpText="The default data source is the default data source selected when creating metrics and experiments."
-                  />
-                </>
-              </div>
-            </div>
+            <DatasourceSettings />
```
romain-growthbook, line 353: Nice.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -60,7 +63,6 @@ import {
   DropTableQueryParams,
 } from "../types/Integration";
 import { DimensionInterface } from "../../types/dimension";
-import { IMPORT_LIMIT_DAYS } from "../util/secrets";
```
romain-growthbook, line 63: Does this need to be removed? I don't see it being deleted.

romain-growthbook (approved):
This looks pretty straight forward to me, only nitpicking remarks. Of course, maybe someone else should have a look.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: There's validation in the front-end which I think is good enough for now.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Moved to a params object.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: It gets deleted in the `getTestQuery()` method below.

PR #2730, opened 2024-07-11
Added option to only show active rules

Auz:
Screenshot 2024-07-10 at 6 05 44 PM

Screenshot 2024-07-10 at 6 05 31 PM

### DIFF
--- a/packages/front-end/components/Features/RuleList.tsx
+++ b/packages/front-end/components/Features/RuleList.tsx
@@ -17,7 +17,11 @@ import {
 } from "@dnd-kit/sortable";
 import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
 import { useAuth } from "@/services/auth";
-import { getRules, isRuleFullyCovered } from "@/services/features";
+import {
+  getRules,
+  isRuleDisabled,
+  isRuleFullyCovered,
+} from "@/services/features";
 import usePermissionsUtil from "@/hooks/usePermissionsUtils";
 import { useLocalStorage } from "@/hooks/useLocalStorage";
 import { Rule, SortableRule } from "./Rule";
@@ -66,7 +70,7 @@ export default function RuleList({
     })
   );
   const showInactiveToggle =
-    items.filter((r) => !r.enabled || r.scheduleRules?.length).length > 0;
+    items.filter((r) => isRuleDisabled(r, experimentsMap)).length > 0;
 
   if (!items.length) {
     return (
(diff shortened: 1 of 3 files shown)


packages/front-end/components/Features/RuleList.tsx
```diff
@@ -60,6 +65,8 @@ export default function RuleList({
       coordinateGetter: sortableKeyboardCoordinates,
     })
   );
+  const showInactiveToggle =
+    items.filter((r) => !r.enabled || r.scheduleRules?.length).length > 0;
```
tzjames, line 69: If the intention of this is to only show the toggle when there is at least one rule that is inactive then it could be made a bit better as sometimes all of the scheduled rules will be active.  We have code somewhere that determines if a scheduled rule is currently active or not as we gray out the row when it is not, so ideally we would use the same logic here.

tzjames (changes requested):
It makes sense to me to have this here.  It seems like a good solution and design wise I can't think of anything better.  

I would fix two things and then I think it is good to land:

1. Fix logic for whether to show the checkbox to make it more accurate
2. Renumber the rules once the rule is hidden.  Right now if you have 4 rules are #1 and #3 are hidden then the rules will be numbered (#2 and #4).

packages/front-end/components/Features/RuleList.tsx
Auz: I was thinking that the numbers might help the user understand that there are hidden rules there - but thinking this might not be an important distinction. I've refactored this a bit more to address these concerns, and moving the filtering up to the list level.

packages/front-end/components/Features/RuleList.tsx
```diff
@@ -46,6 +51,10 @@ export default function RuleList({
   experimentsMap: Map<string, ExperimentInterfaceStringDates>;
 }) {
   const { apiCall } = useAuth();
+  const [hideDisabled, setHideDisabled] = useLocalStorage(
+    `hide-disabled-rules-${environment}`,
```
bryce-fitzsimons, line 55: should this be per feature too? not sure if yes, just calling it out

bryce-fitzsimons (approved):
conditionally approved. but suggest a couple of things:
1. adjust margins / whitespace on checkbox + label
2. check if the localstorage key should be per feature+env or just per env

packages/front-end/components/Features/RuleList.tsx
Auz: No, specifically was asked to make this per environment so they didn't have to keep clicking it.

PR #2744, opened 2024-07-15
Add hypgen params to auth redirect

bryce-fitzsimons:
### Features and Changes

### DIFF
--- a/packages/back-end/src/services/auth/OpenIdAuthConnection.ts
+++ b/packages/back-end/src/services/auth/OpenIdAuthConnection.ts
@@ -37,6 +37,8 @@ if (USE_PROXY) {
   custom.setHttpOptionsDefaults(getHttpOptions());
 }
 
+const passthroughQueryParams = ["hypgen", "hypothesis"];
+
 // Micro-Cache with a TTL of 30 seconds, avoids hitting Mongo on every request
 const ssoConnectionCache = new MemoryCache(async (ssoConnectionId: string) => {
   const ssoConnection = await getSSOConnectionById(ssoConnectionId);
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/auth/OpenIdAuthConnection.ts
```diff
@@ -37,6 +37,8 @@ if (USE_PROXY) {
   custom.setHttpOptionsDefaults(getHttpOptions());
 }
 
+const passthroughQueryParams = ["hypgen", "hypothesis"];
```
bryce-fitzsimons, line 40: `hypgen=1` triggers the new layout
`hypothesis=foo+bar` renders the hypothesis text onto the page

bttf: Is it possible to consolidate into `hypothesis`? (Check if it's non-empty for new layout)

bttf (commented):
Like we talked, probably good to have @jdorn take a look

packages/back-end/src/services/auth/OpenIdAuthConnection.ts
bryce-fitzsimons: it is, but thought it might be good to keep things separate

bryce-fitzsimons: (also bc we'll eventually import multiple hypotheses/variants at once for bandits, in which case we likely wouldn't render hypothesis text)

bttf: Seems premature but I'll allow it 🤷🏾

bttf (approved):

PR #2748, opened 2024-07-16
Add analysis graphs to fact metric pages

lukesonnet:
### Features and Changes

Adds the ability to run metric queries on fact metrics, picking custom lookback windows, aggregation units, and populations (for Pro/Enterprise users). This adds histograms for mean metrics and the first way to do proper ratio metric analysis outside of the context of an experiment. **quantile metrics not yet supported**

Defaults to l30 days, first identifier type, and fact table as the population (similar to the classic metric analysis). On page load we just load the last analysis for this metric. If you change the settings, we hide results but let you reset to the last analysis. We do some date fudging to help caching work.

Main changes:
* Rename `MetricAnalysis` used on metric page to `LegacyMetricAnalysis` to eventually deprecate it and move this approach to classic metrics.
* Create `MetricAnalysis` model and query runner to manage data somewhat like snapshots.
  *  For the `BaseModel`, added ability to have "metric" be a foreign ref in data models.
* Add get, post, and cancel endpoints for new metric analysis model.
* Add new section on `[fmid]` that governs metric analysis.
* New `Histogram` visx component

### Testing

Lots of QA in dev instance in starter + enterprise plans. More ongoing.

### Screenshots

Location on [fmid] with no results
Screenshot 2024-08-06 at 11 24 05 AM

Population chooser with pro locked settings
Screenshot 2024-08-06 at 11 24 12 AM

Result with no rows/values
Screenshot 2024-08-06 at 11 24 26 AM

Mean metric result
Screenshot 2024-08-06 at 11 32 20 AM

Histogram tooltip
Screenshot 2024-08-09 at 8 34 11 AM

Proportion metric result
Screenshot 2024-08-06 at 11 32 41 AM

Ratio metric result
Screenshot 2024-08-06 at 11 24 42 AM

Quantile metric result (not enabled)
Screenshot 2024-08-09 at 8 35 37 AM

State when changing settings from results that are available
Screenshot 2024-08-06 at 11 34 19 AM

More menu
Screenshot 2024-08-06 at 11 34 31 AM

error state
Screenshot 2024-08-06 at 12 24 15 PM


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -580,6 +585,446 @@ export default abstract class SqlIntegration
...
+          }),
+          ...(parseFloat(row.units_bin_22) && {
+            units_bin_22: parseFloat(row.units_bin_22),
+          }),
+          ...(parseFloat(row.units_bin_23) && {
+            units_bin_23: parseFloat(row.units_bin_23),
+          }),
+          ...(parseFloat(row.units_bin_24) && {
+            units_bin_24: parseFloat(row.units_bin_24),
+          }),
```
lukesonnet, line 1020: Yeah, I know. Happy to do something else here instead.

packages/back-end/src/routers/metric-analysis/metric-analysis.controller.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const metric = await getExperimentMetricById(context, metricAnalysis.metric);
+
+  if (!metric?.datasource) {
+    throw new Error("Could not cancel query, datasource not found");
+  }
+  const integration = await getIntegrationFromDatasourceId(
+    context,
+    metric.datasource
+  );
```
jdorn, line 114: I feel like we should have the same "canRunQueries" permission check here as we do when we're starting the queries.

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
```diff
@@ -0,0 +1,260 @@
...
+  innerWidth: number,
+  yScale: ScaleLinear<unknown, unknown, never>
+) {
+  const index = data.indexOf(datapoint);
+  if (index === -1) {
+    return null;
+  }
+  const x =
+    (data.length > 0 ? (index + 0.5) / (data.length + 1) : 0) * innerWidth;
+  const y = (yScale(datapoint.units ?? 0) ?? 0) as number;
```
gazzdingo, line 42: can we use the `.undfined(0)` and get rid of the `?? 0` im pretty new to d3 graphs but from the docs it seems like that would work

gazzdingo: also at the moment Y could be any value other than `null` and `undefined`
again im not sure what values could come out of `yScale()`

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  const [marginTop, marginRight, marginBottom, marginLeft] = margin;
+
+  const width = (containerBounds?.width || 0) + marginRight + marginLeft;
+  const yMax = height - marginTop - marginBottom;
+  const xMax = containerBounds?.width || 0;
+  const graphHeight = yMax;
+  const binWidth = xMax / data.length;
+
+  const xScale = scaleLinear({
```
gazzdingo, line 88: do we care about setting `undefined()` here for any unexpeted behavior from data?

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
```diff
@@ -0,0 +1,260 @@
...
+      }
+
+      showTooltip({
+        tooltipLeft: tooltipData.x,
+        tooltipTop: tooltipData.y,
+        tooltipData: tooltipData,
+      });
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [hoverBin, data, marginLeft, marginRight, width, showTooltip, hideTooltip]
```
gazzdingo, line 143: i think that this can be pulled out of a use effect into a function that runs when hoverBin changes.

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
```diff
@@ -0,0 +1,260 @@
...
+  const xMax = containerBounds?.width || 0;
+  const graphHeight = yMax;
+  const binWidth = xMax / data.length;
+
+  const xScale = scaleLinear({
+    domain: [0, data.length],
+    range: [0, xMax],
+  });
+
+  const yScale = scaleLinear({
```
gazzdingo, line 93: same here

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  >
+                    {tooltipData && getTooltipContents(tooltipData, formatter)}
+                  </TooltipWithBounds>
+                </>
+              )}
+            </div>
+            <svg width={width} height={height}>
+              <Group top={marginTop} left={marginLeft}>
+                {binWidth
+                  ? data.map((d, i) => (
```
gazzdingo, line 207: its better to break looping functions outside of the main render. so it would be best to put this data.map into a funtion and then call it from here

gazzdingo: you can also do the binwidwith logic in that function to tidy up the main render

packages/front-end/components/MetricAnalysis/IdentifierChooser.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <div>
+      <div className="uppercase-title text-muted">Aggregation Unit</div>
+      <SelectField
+        disabled={factTable.userIdTypes.length < 2}
+        containerClassName={"select-dropdown-underline"}
+        options={factTable.userIdTypes.map((d) => {
+          return { label: d, value: d };
+        })}
+        formatOptionLabel={({ label }) => {
+          return <code>{`${label}`}</code>;
```
gazzdingo, line 32: you can just do `{label}`

packages/front-end/components/MetricAnalysis/MetricAnalysis.tsx
```diff
@@ -0,0 +1,784 @@
...
+  let numeratorText: string;
+  let denominatorText: JSX.Element;
+  let numeratorValue: JSX.Element;
+  let denominatorValue: JSX.Element;
+  if (metricType === "ratio" && numeratorFormatter && denominatorFormatter) {
+    numeratorText = "Numerator: ";
+    denominatorText = <>{"Denominator: "}</>;
+    numeratorValue = <>{numeratorFormatter(result.numerator ?? 0)}</>;
+    denominatorValue = <>{denominatorFormatter(result.denominator ?? 0)}</>;
+  } else {
```
gazzdingo, line 76: you could just set your inital values to this then override if meets the if satement above

gazzdingo: i feel like none of them need to be JSX elements as well you can just set them all as string so no need for `= <>{bla}</>` you can just do `= bla`

packages/front-end/components/MetricAnalysis/MetricAnalysis.tsx
```diff
@@ -0,0 +1,784 @@
...
+    populationType: values.populationType,
+    populationId: values.populationId ?? null,
+  };
+}
+
+function settingsMatch(
+  settings: MetricAnalysisSettings,
+  desiredSettings: CreateMetricAnalysisProps
+) {
+  return (
```
gazzdingo, line 163: not sure if this is going to extend but this is a way to make it so that if you add settings it will be more robust
```
  Object.entries(settings).every(
    ([key, value]) => desiredSettings[key] === value
  )
```

gazzdingo: that will just return a true or false if they match or not

lukesonnet: Thanks, safer now. I excluded a few fields that I don't want to check

packages/back-end/src/routers/metric-analysis/metric-analysis.controller.ts
lukesonnet: added one below, ty.

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
bryce-fitzsimons: We had some weird bug in the past where yScale would return something nonsensical, so i'd prefer to leave this guard in. The inner ?? 0 is unnecessary though.

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
bryce-fitzsimons: not sure, but these should be memoized (unrelated). will fix

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
bryce-fitzsimons: i think we should keep the deps. the tooltip (position, content, hovered bin) will need to recalc if any of these things change. Data, size, etc.

packages/front-end/components/MetricAnalysis/HistogramGraph.tsx
bryce-fitzsimons: Personal pref, but i disagree. For simple maps over data with simple render logic, this is more readable to me. Here's how AirBnB renders their histograms:
https://airbnb.io/visx/bars

gazzdingo (approved):
LGTM

PR #2749, opened 2024-07-16
Auto Fact Table Docs

mknowlton89:
### Features and Changes

Updates docs with some guidance on automatic fact table generation


royalfig:
@mknowlton89 I added some changes for clarity, tone, and consistency. 

Please take a peek and make sure the meaning is still intact. 

- The gist of the changes were to have a more positive, matter-of-fact tone.
- When referring to a button or place in the app, use **Bold** instead of `Code`. (Use `code` when talking about object properties, SQL queries, etc.)
- In the docs, the button is identified as **Auto Generate Fact Tables** but it has a different name in the UI (**Auto-generate Fact Tables...**). I think the way it is in the docs is preferable and would update the UI to match.

royalfig (approved):

PR #2752, opened 2024-07-16
Don't display secret for non raw webhook.

romain-growthbook:
Secret is only relevant for raw webhooks!

Screenshot 2024-07-15 at 11 43 54 AM

Screenshot 2024-07-16 at 11 41 27 AM


gazzdingo (approved):
LGTM

PR #2754, opened 2024-07-16
Add FAQ for experiment refresh setting

royalfig:
### Features and Changes
Add text and screenshot for experiment data refresh settings in FAQ

Ref: https://growthbookapp.slack.com/archives/C07BMHJKG2U/p1721144053881009


romain-growthbook (approved):

romain-growthbook:
At some point, we might want to look at some kind of automation to conver png images into compressed JPEG.

royalfig:
> At some point, we might want to look at some kind of automation to conver png images into compressed JPEG.

@romain-growthbook In my experience, the JPEGs never look as good as PNGs when it's an image of text (which is common in our docs). However, AVIF/WEBP might be a different story. I'll look into. Thanks for the review

PR #2768, opened 2024-07-19
Added request timeout for presto/trino

Auz:
Screenshot 2024-07-19 at 10 05 34 AM


lukesonnet (approved):
Screenshot 2024-07-22 at 9 33 06 AM

Tested and can get a timeout. Setting seems to work, hope that if it is null it is indeed not set, but hard to test. Should work though because test queries run if the field is unset.

lukesonnet:
One nit, the placeholder text and the help text are repetitive. Can we just do it in one place?

Issue #2765, opened 2024-07-19
[Feature] Remove .gb-anti-flicker when resolves


marceloformentao-hotmart:
### Description of Feature

Taking a look at how the antiFlicker works, when using a URL Redirect for example, It helps me when the user needs to be redirected, but, when the user is supposed to stay on that variation, the white page stay put for the whole 3500ms.

It should have something that sees the antiFlicker active and remove it when the growthbook sdk finish running.

https://github.com/growthbook/growthbook/blob/233cb2a79d6b202791dbe1194e2a488b81ae8b3d/packages/sdk-js/src/GrowthBook.ts#L1757-L1773

For now, I'm doing something like this when using the HTML Script Tag in a slow page:

```html
<script>
    // Configure GrowthBook settings
    window.growthbook_config = window.growthbook_config || {};
    window.growthbook_config.antiFlicker = true;  // Enable anti-flicker
    window.growthbook_config.antiFlickerTimeout = 3500;  // Set max anti-flicker timeout

    // Function to remove the anti-flicker class
    function removeAntiFlickerClass() {
      console.log('Removing anti-flicker class');
      document.documentElement.classList.remove('gb-anti-flicker');
    }

    // Check if GrowthBook is ready and handle anti-flicker class based on redirect
    function checkGrowthBookReady() {
      if (window._growthbook && window._growthbook.ready) {
        console.log('GrowthBook is fully initialized');

        // Get the redirect URL if exists
        const redirectUrl = window._growthbook.getRedirectUrl();
        if (redirectUrl) {
          console.log('Redirect URL found:', redirectUrl);
          // The redirect will be handled automatically by GrowthBook, no need to manually redirect
        } else {
          console.log('No redirect URL, removing anti-flicker class');
          removeAntiFlickerClass();
        }
      } else {
        console.log('GrowthBook not ready yet, retrying...');
        setTimeout(checkGrowthBookReady, 50);  // Retry every 50ms until ready
      }
    }

    // Start checking if GrowthBook is ready
    document.addEventListener("DOMContentLoaded", checkGrowthBookReady);

    // Fallback to remove the class if GrowthBook is not available
    setTimeout(() => {
      if (!window._growthbook || !window._growthbook.ready) {
        console.log('GrowthBook is not available, removing anti-flicker class');
        removeAntiFlickerClass();
      }
    }, 3500);  // Fallback timeout
  </script>
```

Issue #2735, opened 2024-07-12
URL Redirect sends mass requests if server takes its time to respond to initial request


AtakanColak:
I believe it's due to the lack of a scheduling/checking mechanism for existing navigate requests in the [code](https://github.com/growthbook/growthbook/blob/2ec26c406c642dcf17abf05706e796fe7ba53ab4/packages/sdk-js/src/GrowthBook.ts#L732C6-L750C10)

Issue #2676, opened 2024-06-24
[Bug] BrowserCookieStickyBucketService is exported as a type


AtakanColak:
### Summary

I attempt to create `BrowserCookieStickyBucketService` instance by importing it from ReactSDK. It causes the following error;

> 'BrowserCookieStickyBucketService' cannot be used as a value because it was exported using 'export type'.

srinivasarynh:
Hi @AtakanColak,
![Screenshot from 2024-07-09 15-58-11](https://github.com/growthbook/growthbook/assets/61593110/0abeaf47-149d-4bb3-afb7-4975f28d72c0)


In the packages/sdk-react/src folder, inside the index.ts file, BrowserCookieStickyBucketService is exported as a type. This is the reason for the issue "'BrowserCookieStickyBucketService' cannot be used as a value because it was exported using 'export type'."

AtakanColakSoftwire:
This can be closed now @bryce-fitzsimons

PR #2770, opened 2024-07-19
Update SDKs for 1.2.0

bryce-fitzsimons:
### Features and Changes

- [x]  fix StickyBucketService [exports (2676)](https://github.com/growthbook/growthbook/issues/2676) in React SDK
- [x] refresh `stickyBucketIdentifierAttributes` when recalculating sticky buckets so that we can pick up SB docs that were previously not picked up by the SDK
- [x] make the tracking callback async, await it when navigating for url redirects
- [x] resolve some [quirks (2765)](https://github.com/growthbook/growthbook/issues/2765) with the anti flicker logic
- [x] try to prevent [navigate stampedes (2735)](https://github.com/growthbook/growthbook/issues/2735)
- [x] change default maxAge to reduce multiple exposures on phase change. expose cache settings for auto wrapper.
- [x] version bumps: sdks
- [ ] version bumps / json: GB app
- [ ] docs (pending PR review)
- [x] qa

note: version will be bumped to 1.2.0 despite the branch name


packages/sdk-js/src/GrowthBook.ts
```diff
@@ -733,13 +736,22 @@ export class GrowthBook<
...
               try {
                 navigate(url);
               } catch (e) {
                 console.error(e);
               }
-            }, this._ctx.navigateDelay ?? 100);
+            });
           } else {
             try {
               navigate(url);
```
bryce-fitzsimons, line 757: wondering if we should also do something similar for non-browser contexts. I do not think we'd need this on edge fwiw.

packages/sdk-js/test/sticky-buckets.test.ts
```diff
@@ -209,6 +209,7 @@ describe("sticky-buckets", () => {
         country: "USA",
       },
     });
+    growthbook1a.debug = true;
```
jdorn, line 212: Do we need this?

packages/sdk-js/src/auto-wrapper.ts
```diff
@@ -233,20 +233,43 @@ if (
...
+        eventCallback: datalayerResolve,
+      });
+    }
 
     // Segment - analytics.js
-    window.analytics &&
-      window.analytics.track &&
-      window.analytics.track("Experiment Viewed", p);
+    if (window.analytics && window.analytics.track) {
+      window.analytics.track("Experiment Viewed", eventParams);
```
jdorn, line 269: If someone is only using Segment, previously, this trackingCallback would return void, which would cause it to use the `navigateDelay`.  Now, it will return `Promise<void>` which resolves immediately, which will cause it to immediately redirect before Segment could fire.

We should use a promise here instead that waits a hardcoded amount of time.  Segment's docs recommend using a value of `300ms` to give the tracking call enough time to finish.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -726,20 +734,33 @@ export class GrowthBook<
...
         if (navigate) {
           if (isBrowser) {
-            this._setAntiFlicker();
-            window.setTimeout(() => {
+            this._setAntiFlicker(true);
+            // Wait for the possibly-async tracking callback, bound by min and max delays
+            Promise.all([
+              promiseTimeout(
+                this._activeTrackingCall,
+                this._ctx.maxNavigateDelay ?? 10000
```
jdorn, line 749: 10 seconds seems way too long here.  I was thinking it would be more like 1 second

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -726,20 +734,33 @@ export class GrowthBook<
...
+              ),
+              new Promise((resolve) =>
+                window.setTimeout(resolve, this._ctx.navigateDelay ?? 100)
```
jdorn, line 752: Do we want this?  If the tracking call finishes fast, we shouldn't need to wait the full 100ms

packages/sdk-js/test/sticky-buckets.test.ts
bryce-fitzsimons: no, leftover from debugging :)

packages/sdk-js/src/auto-wrapper.ts
bryce-fitzsimons: The SDK will always wait for at least the MAX of:
- `context.navigateDelay ?? 100`
- the tracking promise

So we could just set navigateDelay to 300

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: sure. i'm looking at it thru the lens of the folks who have been experiencing major SRM issues due to a large number of tracking calls taking over 1-2 secs

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: I put it here to protect against false positives (tracking complete after 0ms) / misconfigured tracking callbacks. It's adjustable to 0. We can remove it if you prefer though, but it does seem useful for things like the script tag

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -398,11 +399,13 @@ export type InitOptions = {
   skipCache?: boolean;
   payload?: FeatureApiResponse;
   streaming?: boolean;
+  cacheSettings?: CacheSettings;
 };
 
 export type InitSyncOptions = {
   payload: FeatureApiResponse;
   streaming?: boolean;
+  cacheSettings?: CacheSettings;
```
jdorn, line 408: I don't think we need this for `initSync`, since that method doesn't use the feature repository at all.

packages/sdk-js/src/auto-wrapper.ts
jdorn: I'd prefer having `navigateDelay` set to something smaller like 100, and have this Segment call specifically wait 300ms.  So they only pay that extra penalty if they are using Segment.  Otherwise, if GA finishes sooner, it doesn't have to wait.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -214,6 +221,10 @@ export class GrowthBook<
   public initSync(options: InitSyncOptions): GrowthBook {
     this._initialized = true;
 
+    if (options.cacheSettings) {
+      configureCache(options.cacheSettings);
+    }
```
jdorn, line 226: Remove this from initSync

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -726,20 +743,33 @@ export class GrowthBook<
...
-            window.setTimeout(() => {
+            this._setAntiFlicker(true);
+            // Wait for the possibly-async tracking callback, bound by min and max delays
+            Promise.all([
+              promiseTimeout(
+                this._activeTrackingCall,
+                this._ctx.maxNavigateDelay ?? 3000
+              ),
+              new Promise((resolve) =>
+                window.setTimeout(resolve, this._ctx.navigateDelay ?? 300)
```
jdorn, line 761: Revert this to 100

packages/sdk-js/test/url-redirects.test.ts
```diff
@@ -45,7 +45,7 @@ describe("urlRedirects", () => {
     await sleep();
 
     gb.setURL("http://www.example.com/home");
-    await sleep(100);
+    await sleep(350);
```
jdorn, line 48: Revert the changes in this file

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -683,6 +698,7 @@ export class GrowthBook<
     if (isBlocked) {
       process.env.NODE_ENV !== "production" &&
         this.log("Auto experiment blocked", { id: experiment.key });
+      this._unsetAntiFlicker();
```
jdorn, line 701: This doesn't feel like it belongs in this method.  Imagine you have 2 redirect experiments.  The first one is blocked, so it hits this and removes the overlay.  The next one is not blocked so it redirects, but the overlay is already gone

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -726,20 +743,33 @@ export class GrowthBook<
...
+                this._ctx.maxNavigateDelay ?? 3000
+              ),
+              new Promise((resolve) =>
+                window.setTimeout(resolve, this._ctx.navigateDelay ?? 300)
+              ),
+            ]).then(() => {
               try {
-                navigate(url);
+                if (!this._isNavigating) {
+                  this._isNavigating = true;
```
jdorn, line 766: This needs to get unset after the navigation finishes.  Especially important for SPAs where the navigate callback doesn't do a full page load.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1483,7 +1513,8 @@ export class GrowthBook<
     }
 
     // 14. Fire the tracking callback
-    this._track(experiment, result);
+    // Store the promise in case we're awaiting it (ex: browser url redirects)
+    this._activeTrackingCall = this._track(experiment, result);
```
jdorn, line 1517: I don't like the idea of storing this globally.  Seems like it might open us up to race conditions.  Instead, we could include this promise in the return value of the function.

bryce-fitzsimons: pinged you, but I kind of disagree, unless I'm misunderstanding your suggestion:
- We'd need to add something like `result.trackingCall` which may not serialize well for people consuming/logging `result` directly. We might break existing integrations.
- the global `_activeTrackingCall` is set and consumed in the same exe cycle since `_runAutoExperiment()` is synchronous, so its hard to imagine a race condition occurring.

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: Do you mean something like
```
navigate(url).then(() => this._isNavigating = false);
```
One possible problem is that navigate() may not be async or at least not trackable (such as `window.location.replace(url)`), so it would get unset instantly. Maybe that's ok though, but then it doesn't do much to dedupe.

Or perhaps we make the default navigate function return a promise with an arbitrary 1sec timeout?

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: Sure we can remove. The flip side though is the anti-flicker stays up far too long (3000ms by default) when only a single experiment triggers but is blocked.

bryce-fitzsimons: there are a few other bail-out scenarios where i'm calling `_unsetAntiFlicker();`:
- redirect variant does not actually redirect (ex: control)
- invalid urlPatterns (rare)
- already on the target page (rare)

the same double experiment scenario you mentioned earlier could happen here too, right?

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: discussed. we decided to return the tracking call locally in the private _run() method without affecting public run()

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: move anti flicker to _updateAllAutoExperiments, outside of loop

bryce-fitzsimons: no anti-flicker in constructor, just as part of the (default) navigate function.

bryce-fitzsimons: no anti-flicker for dom changes. only for default navigate function. make sure hooks work...

packages/sdk-js/src/auto-wrapper.ts
```diff
@@ -266,6 +297,44 @@ gb.init({
     dataContext.noStreaming ||
     windowContext.backgroundSync === false
   ),
+  cacheSettings: windowContext.cacheSettings,
```
bryce-fitzsimons, line 300: if antiFlicker set, hide the screen before init(). Otherwise we're waiting for features to load

bryce-fitzsimons:
todo: add unit test for tracking call being async

bryce-fitzsimons:
unit test: navigate stampede. custom navigate function, make sure we aren't running exp multiple times / re-navigating

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1741,35 +1787,27 @@ export class GrowthBook<
...
+        delay: 100,
+      };
     } else if (isBrowser) {
-      return (url: string) => {
-        window.location.replace(url);
+      return {
+        navigate: (url: string) => {
+          window.location.replace(url);
+        },
+        delay: 0,
```
jdorn, line 1804: These delays are backwards?

packages/sdk-js/src/auto-wrapper.ts
```diff
@@ -205,6 +209,46 @@ function getAttributes() {
...
 
+function setAntiFlicker() {
+  window.clearTimeout(antiFlickerTimeout);
+
+  const timeoutMs =
+    windowContext.antiFlickerTimeout ??
+    (dataContext.antiFlickerTimeout
+      ? parseInt(dataContext.antiFlickerTimeout)
+      : null) ??
+    3500;
```
jdorn, line 220: If `dataContext.antiFlickerTimeout` is not parseable as an int (e.g. `"a"`), this will result in `NaN`.  This means, the delay will be treated as `0ms` instead of `3500ms`.

bryce-fitzsimons: that seems like a user error, but can add some guards

jdorn (approved):

PR #2775, opened 2024-07-23
Improved admin functionality

Auz:


tzjames (changes requested):
Looks good overall.  

I am a bit concerned about the disabled property, in that perhaps we are forgetting places where it should be filtered out/not work.  For instance should api endpoints that are access data within that organization still work?  Or the other find function I pointed out.  etc..

I am also not sure why, but the org list doesn't seem to update the background color of the org when you disable it until you refresh the page.

tzjames (changes requested):
Looks good overall.  I did leave some suggestions though.

I am a little concerned that we didn't catch everthing that needs to happen when an org is marked as "disabled".  For instance the extra find... function I pointed out.  Or perhaps organization api endpoints should filter it out, or a have a parameter to.  

Speaking of which do the api endpoints need to be updated so that they can handle the new prop?

Also when I tried to disable an org testing it out, it didn't correctly update the list without refreshing the page.  I don't immediately see what is missing.  Not necessarily a show stopper.

packages/back-end/src/controllers/admin.ts
```diff
@@ -80,6 +114,107 @@ export async function putOrganization(
...
+    });
+  }
+
+  updates.disabled = true;
+  orig.disabled = org.disabled;
+
+  await updateOrganization(org.id, updates);
+
+  await req.audit({
+    event: "organization.update",
```
tzjames, line 180: We might consider having a more explicit event type:
`event: "organization.disable",`

tzjames: And `organization.enable` below...

packages/back-end/src/controllers/admin.ts
```diff
@@ -96,3 +231,131 @@ export async function putOrganization(
...
+    });
+  }
+
+  const { page, search } = req.query;
+
+  const organizationInfo: Record<string, object> = {};
+  const allUsers = await getAllUsersFiltered(parseInt(page ?? "1"), search);
+  if (allUsers?.length > 0) {
+    for await (const user of allUsers) {
+      const orgs = await findOrganizationsByMemberId(user.id);
```
tzjames, line 251: This has the potential to be inefficient. I believe you are loading 50 members a page and hence this would make 50 db requests. It is probably worth it to make a new query that finds all organizations by a list of members and then create your array from the result of that single query.

packages/back-end/src/controllers/admin.ts
```diff
@@ -96,3 +231,131 @@ export async function putOrganization(
...
+  if (!org) {
+    return res.status(404).json({
+      status: 404,
+      message: "Organization not found",
+    });
+  }
+
+  const members: UserInterface[] = [];
+  for await (const member of org.members) {
+    const user = await getUserById(member.id);
```
tzjames, line 296: Again this could use a performance boost that is perhaps even worse than in getMembers. We have an org that has 217 members. Fortunately we already have a `getUsersByIds` function, you can just use that.

packages/back-end/src/controllers/admin.ts
```diff
@@ -96,3 +231,131 @@ export async function putOrganization(
...
+    verified: boolean;
+  }>,
+  res: Response
+) {
+  if (!req.superAdmin) {
+    return res.status(403).json({
+      status: 403,
+      message: "Only superAdmins can update members",
+    });
+  }
```
tzjames, line 322: We don't have to do it in this PR, but we could consider making an adminRouter and adding this as middleware to remove code duplication and make it less likely that someone forgets to add this check.

packages/back-end/src/controllers/admin.ts
```diff
@@ -96,3 +231,131 @@ export async function putOrganization(
...
+  if (verified !== member.verified) {
+    updates.verified = verified;
+    orig.verified = member.verified;
+  }
+
+  await updateUser(userId, updates);
+
+  await req.audit({
+    event: "team.update",
```
tzjames, line 350: copypasta?  You probably meant `"user.update"` here.

packages/back-end/src/models/UserModel.ts
```diff
@@ -52,6 +52,46 @@ export async function getAllUsers(): Promise<UserInterface[]> {
...
+    .find(query)
+    .sort([["dateCreated", -1]])
+    .skip((page - 1) * 50)
+    .limit(50)
+    .toArray();
+
+  // return the user interface but filter out password hash
+  return docs
+    .map((u) => toInterface(u))
+    .map((u) => ({ ...u, passwordHash: "" }));
```
tzjames, line 79: It is probably better to list the properties that you actually want. If another sensitive property gets added in the future then we might forget to update this.

packages/front-end/components/Admin/EditOrganization.tsx
```diff
@@ -96,6 +177,63 @@ const EditOrganization: FC<{
...
+                className="ml-2"
+                id="legacyEnterpriseToggle"
+                value={legacyEnterprise}
+                setValue={setLegacyEnterprise}
+              />
+              <div>
+                <span className="text-muted small">
+                  Organizations with enterprise enabled this way are not billed,
+                  and will not expire. This is a legacy feature. Be sure to also
+                  adjust the free seats to match org requirements.
```
tzjames, line 231: `free seats` does nothing in terms of limiting the number of users in the account.  I would just remove that language.  I would include deprecated language here.
```
<b>Deprecated:</b> This is an old way to enable enterprise features for an organization, which doesn't expire, and doesn't restrict seats. Please uncheck this and instead user Retool and set a licenseKey instead.
```

packages/front-end/components/Admin/EditOrganization.tsx
```diff
@@ -96,6 +177,63 @@ const EditOrganization: FC<{
...
+              <input
+                type="number"
+                min={0}
+                className="form-control"
+                value={freeSeats}
+                onChange={(e) => setFreeSeats(parseInt(e.target.value))}
+              />
+              <div>
+                <span className="text-muted small">
+                  Number of seats that will not be billed
```
tzjames, line 215: `Number of seats that can be added when on a free plan`

tzjames (changes requested):
Fixed the bad line numbers in the previous review.

packages/back-end/src/controllers/admin.ts
Auz: done

packages/back-end/src/controllers/admin.ts
Auz: done

packages/front-end/components/Admin/EditOrganization.tsx
Auz: done

packages/front-end/components/Admin/EditOrganization.tsx
Auz: done

packages/back-end/src/controllers/admin.ts
Auz: done

packages/back-end/src/controllers/admin.ts
Auz: done

packages/front-end/components/Admin/EditOrganization.tsx
```diff
@@ -2,28 +2,49 @@ import { useState, FC } from "react";
...
   close?: () => void;
   id: string;
+  disablable: boolean;
+  currentDisabled: boolean;
   currentName: string;
   currentExternalId: string;
   currentLicenseKey: string;
   currentOwner: string;
+  currentDomain: string;
+  currentAutoApproveMembers: boolean;
```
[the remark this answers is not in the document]
Auz: done

packages/front-end/components/Admin/EditOrganization.tsx
```diff
@@ -52,6 +58,59 @@ const EditOrganization: FC<{
...
       inline={!close}
+      secondaryCTA={
+        disablable ? (
+          <div className="flex-grow-1">
+            {currentOrg.disabled ? (
+              <button
+                className="btn btn-info"
+                onClick={(e) => {
+                  e.preventDefault();
+                  apiCall<{
```
tzjames, line 69: The onClick should be `async` and you should `await` this call to complete, otherwise when you call `onEdit` below it may return organizations endpoint before the enabling happens.  It did this fairly consistently for me.

packages/front-end/components/Admin/EditOrganization.tsx
```diff
@@ -52,6 +58,59 @@ const EditOrganization: FC<{
...
+              <button
+                className="btn btn-danger"
+                onClick={(e) => {
+                  e.preventDefault();
+                  if (
+                    confirm(
+                      "Are you sure you want to disable this organization? Users will not be able to use this organization"
+                    )
+                  ) {
+                    apiCall<{
```
tzjames, line 94: Similarly here it should also be `await`ed.

tzjames (approved):
I left a few more comments, but I'll approve as I'm not sure if I'll be able to get another chance to review before I leave and remaining comments are mostly minor. 

I think there is still a bit of a misunderstanding of how Free Seats currently work to compared to how you expect them to work that we might need to address.

PR #2779, opened 2024-07-23
Adjust validation on UrlRedirectModal to match SDK validation

### LINKED ISSUE #2760: [Bug] Redirect URL variant not rendering when URL contains params
### Summary

Variant not rendering with the URL redirect when the URL 

URL redirect variation is not rendering when URL is https://googletest.com/?lp=v3 but works when it is a different URL like https://googletest.com/pricing-plans. The devtools show experiment is running but the variation is never shown to the end user, only the control (https://googletest.com/). Is URL redirects supported when only the query params differs between the control and variant? This is for a HTTP script tag implementation

#### Expected Behavior

The URL redirect should render the variant when the redirection URL contains params

#### Current Behavior

The variant is not rendering in the URL redirection tests when the URL contains params such as googletest.com/?lp=v3

msamper:
### Features and Changes

- Closes #2760

### DIFF
--- a/docs/docs/running-experiments/url-redirects.mdx
+++ b/docs/docs/running-experiments/url-redirects.mdx
@@ -21,9 +21,9 @@ You then specify "Destination URLs" for each of your variations. You can also tu
 
 When a user visits the Original URL and is included in the experiment, they will be assigned a variation. Then, an "Experiment Viewed" event will be sent to your data warehouse. After a short delay (default `100ms`) to allow for the tracking event to finish, the user will be redirected to the destination URL.
 
-### Redirects to the same URL with the addition of query params
+### Redirecting when only query params change
 
-For redirect tests where you want to redirect a user to the same page with different query params you'll need to include the query params you plan to use in the original URL with no value for our targeting to work correctly. (i.e. if you want to redirect a user to `http://www.growthbook.io/test?abc=123` from `http://www.growthbook.io/test`, your original URL will need to be `http://www.growthbook.io/test?abc`)
+Setting up tests where the redirect URL only varies from the original by the query parameters requires special consideration. You'll need to include the query params you plan to use in the original URL with no value entered in order for our URL targeting to work correctly (i.e. if you want to redirect a user to `http://www.growthbook.io/test?abc=123` from `http://www.growthbook.io/test`, your original URL will need to be `http://www.growthbook.io/test?abc`).
 
 ## Implementation
 
--- a/docs/docs/running-experiments/url-redirects.mdx
+++ b/docs/docs/running-experiments/url-redirects.mdx
@@ -21,6 +21,10 @@ You then specify "Destination URLs" for each of your variations. You can also tu
 
 When a user visits the Original URL and is included in the experiment, they will be assigned a variation. Then, an "Experiment Viewed" event will be sent to your data warehouse. After a short delay (default `100ms`) to allow for the tracking event to finish, the user will be redirected to the destination URL.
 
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Experiment/UrlRedirectModal.tsx
```diff
@@ -231,26 +232,39 @@ const UrlRedirectModal: FC<{
...
               if (originUrl.protocol !== variantUrl.protocol) {
-                warning = `Destination URL is using "${variantUrl.protocol}" and the origin URL is using "${originUrl.protocol}"`;
+                warning = `Destination URL is using "${variantUrl.protocol}" and the original URL is using "${originUrl.protocol}"`;
               }
               if (originUrl.host !== variantUrl.host) {
-                warning = `Destination URL is "${variantUrl.host}" and the origin URL is "${originUrl.host}"`;
+                warning = `Destination URL is "${variantUrl.host}" and the original URL is "${originUrl.host}"`;
+              }
+              if (
+                !originUrl.searchParams.size &&
```
bryce-fitzsimons, line 257: I'm not sure this exactly sums up the gap. For example, you could be on `www.mysite.io/?unrelated=1` and redirect to `www.mysite.io/?unrelated=1&newPage=1`

bryce-fitzsimons: do we want to block saving when this is the case or just warn?

msamper: It will already block with the adjusted validation for `destinationMatchesOrigin`

docs/docs/running-experiments/url-redirects.mdx
```diff
@@ -21,6 +21,10 @@ You then specify "Destination URLs" for each of your variations. You can also tu
 
 When a user visits the Original URL and is included in the experiment, they will be assigned a variation. Then, an "Experiment Viewed" event will be sent to your data warehouse. After a short delay (default `100ms`) to allow for the tracking event to finish, the user will be redirected to the destination URL.
 
+### Redirects to the same URL with the addition of query params
+
+For redirect tests where you want to redirect a user to the same page with different query params you'll need to include the query params you plan to use in the original URL with no value for our targeting to work correctly. (i.e. if you want to redirect a user to `http://www.growthbook.io/test?abc=123` from `http://www.growthbook.io/test`, your original URL will need to be `http://www.growthbook.io/test?abc`)
```
bryce-fitzsimons, line 26:
```suggestion
### Redirecting where query params change

Setting up tests where the redirect URL only varies from the original by the query parameters requires special consideration. You'll need to include the query params you plan to use in the original URL with no value entered in order for our URL targeting to work correctly (i.e. if you want to redirect a user to `http://www.growthbook.io/test?abc=123` from `http://www.growthbook.io/test`, your original URL will need to be `http://www.growthbook.io/test?abc`).
```

packages/front-end/components/Experiment/UrlRedirectModal.tsx
```diff
@@ -231,26 +232,39 @@ const UrlRedirectModal: FC<{
...
+                variantUrl.searchParams.size
+              ) {
+                warning = (
+                  <>
+                    Destination URL has query parameters but the original URL
+                    does not. See{" "}
+                    <DocLink docSection="url_redirects">our docs</DocLink> for
```
bryce-fitzsimons, line 264: consider linking to the #subsection of the document. maybe not necessary if the doc itself is small though...

bryce-fitzsimons (approved):

PR #2781, opened 2024-07-23
Add `regressionAdjustmentEnabled` to experiment APIs

itsgrimetime:
### Features and Changes
Allows setting & reading `regressionAdjustmentEnabled` through experiment APIs. Also adds a default to `postExperiment` to set `regressionAdjustmentEnabled` to the org default.

### Testing
Tested locally & in our staging environment


lukesonnet:
I've made a few changes to be more in line with how we currently use org defaults and tested it out a bit. @itsgrimetime , would you mind testing out my latest version to make sure it achieves your goals?

lukesonnet (approved):

PR #2783, opened 2024-07-24
Add documentation for Custom Markdown

msamper:
### Features and Changes

Adds documentation for Custom Markdown

### DIFF
--- a/docs/docs/using/growthbook-best-practices.mdx
+++ b/docs/docs/using/growthbook-best-practices.mdx
@@ -136,7 +136,7 @@ before your team is able to start an experiment.
 
 ### Custom Markdown
 
-For larger organizations, it can helpful to provide some org-specific messaging to your team around processes for setting up experiments and features. GrowthBook gives enterprise users the capability to inject markdown into key pages in the app so that you can do just that. We've also surfaced certain Handlebars variables on each page so that you can leverage the power of [Handlebars templates](https://handlebarsjs.com/guide/#simple-expressions) to customize your messaging depending on the context.
+Enterprise users can inject custom markdown content into key pages of the GrowthBook app. This feature allows you to provide organization-specific guidance on processes for setting up experiments and features. Additionally, we've made certain Handlebars variables available on each page. This enables you to use [Handlebars templates](https://handlebarsjs.com/guide/#simple-expressions) to dynamically customize your messaging based on context.
 
 | Page             | Variables                                                            |
 | ---------------- | -------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


docs/docs/using/growthbook-best-practices.mdx
royalfig, line 1: Does it matter to the user that we're using Hbs to parse the variables? Are there other Hbs expressions available, too? 

If it's only the ability to use the variables in markdown, then I probably would omit reference to Handlebars because it's not important that the user know that. However, I'd add an example of how to use it or a screenshot. 

Writing `Welcome to {{orgName}}` in markdown becomes "Welcome to Acme Co." when rendered. You likely have a better use case to put as an example. 

---
At minimum, there's a "be" missing after "it can" in the first sentence. 

I'd likely rewrite like this, tho, to make the feature front and center:

> Enterprise users can inject custom markdown content into key pages of the GrowthBook app. This feature allows you to provide organization-specific guidance on processes for setting up experiments and features. 

> Additionally, we've made certain Handlebars variables available on each page. This enables you to use [Handlebars templates](https://handlebarsjs.com/guide/#simple-expressions) to dynamically customize your messaging based on context. For example, ...

royalfig (approved):
I made some suggestions for improvement, but this is good to go when you're ready!

docs/docs/using/growthbook-best-practices.mdx
jdorn: Handlebars is important if someone wants to use conditional rendering, not just outputing variables.  For example, only show a block of markdown if project = "X"

docs/docs/using/growthbook-best-practices.mdx
```diff
@@ -134,6 +134,19 @@ The first is to make it super easy to add documentation directly in the platform
 for the experiment. Secondly, we added launch checklists, which can require that certain files be filled
 before your team is able to start an experiment.
 
+### Custom Markdown
+
+For larger organizations, it can helpful to provide some org-specific messaging to your team around processes for setting up experiments and features. GrowthBook gives enterprise users the capability to inject markdown into key pages in the app so that you can do just that. We've also surfaced certain Handlebars variables on each page so that you can leverage the power of [Handlebars templates](https://handlebarsjs.com/guide/#simple-expressions) to customize your messaging depending on the context.
```
msamper, line 139:
```suggestion
Enterprise users can inject custom markdown content into key pages of the GrowthBook app. This feature allows you to provide organization-specific guidance on processes for setting up experiments and features. Additionally, we've made certain Handlebars variables available on each page. This enables you to use [Handlebars templates](https://handlebarsjs.com/guide/#simple-expressions) to dynamically customize your messaging based on context.
```

docs/docs/using/growthbook-best-practices.mdx
royalfig: Thanks. It wasn't clear from the docs alone to me whether that was the case. That's really cool, and I'd include an example showcasing that ability.

PR #2789, opened 2024-07-24
Allow min sample size to be 0 in results table

lukesonnet:
If metric min sample size was 0 and metric default min sample size was greater than 0, we weren't letting that 0 fall through.

Before
Screenshot 2024-07-24 at 10 35 52 AM

After
Screenshot 2024-07-24 at 10 36 09 AM


gazzdingo (approved):
LGTM

bryce-fitzsimons (approved):

PR #2794, opened 2024-07-24
Allow updating `windowSettings` on `PUT /metric/:id`

itsgrimetime:
### Features and Changes
Fixes: #2795

It looks like PUT metric was changed (or maybe it never allowed it, I didn't dig deep enough to find out for sure) and the `windowSettings` weren't being copied over from the payload to the updated metric.

### Testing
Tested locally and on our internal deployment

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1663,6 +1663,20 @@ export function putMetricApiPayloadToMetricInterface(
         metric.windowSettings.windowValue =
           behavior.conversionWindowEnd - behavior.conversionWindowStart;
       }
+    } else {
+      metric.windowSettings = {
+        type:
+          behavior.windowSettings?.type == "none"
+            ? ""
+            : behavior.windowSettings?.type ?? DEFAULT_METRIC_WINDOW,
+        delayHours:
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -1663,6 +1663,20 @@ export function putMetricApiPayloadToMetricInterface(
         metric.windowSettings.windowValue =
           behavior.conversionWindowEnd - behavior.conversionWindowStart;
       }
+    } else {
```
lukesonnet, line 1666: Because this is designed to work with only partial behavior updates, I think this check should happen first (e.g. `if (typeof behavior.windowSettings !== "undefined")` then do this code, then `else if (typeof behavior.conversionWindowStart !== "undefined")`.

Otherwise right now if you don't specify `windowSettings` but make other behavior updates, you'll just overwrite the window with the defaults.

Switching the order also helps us prefer the non-deprecated field.

lukesonnet (changes requested):
Thanks for this fix; one minor change needed in the order of the conditional logic execution.

lukesonnet (approved):
ty, looks good. tested it a bit as well.

PR #2798, opened 2024-07-25
Add padding to the very last element right under `main`

romain-growthbook:
This is a follow-up from https://github.com/growthbook/growthbook/pull/2790 which was too rough.

This time, we tap directly into the app's main Layout: https://github.com/growthbook/growthbook/blob/main/packages/front-end/pages/_app.tsx#L108

Assuming that `<main>...</main>` always contains the main content of the app, we ask CSS to add some padding to the very last immediate child, which should be reasonable and achieve what we intend to here, which is to give some scrolling leeway when needer to scroll past the help button.

This has also been confirmed to not add scrolling when the page does not reach the bottom of the screen.

### DIFF
--- a/packages/front-end/styles/global.scss
+++ b/packages/front-end/styles/global.scss
@@ -14,6 +14,10 @@ body,
 #__next,
 main {
   height: 100%;
+
+  > *:last-child {
+    padding-bottom: 70px !important;
+  }
 }
 
 body {


lukesonnet (approved):
Clicked around and this looks good, but now it introduces another visual oddity on experiments page. 

At the bottom of the Features page, our comments is a card, so the extra grey space looks fine:
Screenshot 2024-09-17 at 3 25 24 PM

However, at the bottom of experiments, the white block just ends and turns into the grey background. Can we either add the hline or whatever that we have at the top of the comments block on the experiments page, or turn it into the card that we have on the feature page? I prefer the former since it's probably similar.

Screenshot 2024-09-17 at 3 25 33 PM

You can see the missing border in dark mode more clearly:

Screenshot 2024-09-17 at 3 27 56 PM

PR #2802, opened 2024-07-26
fix data offset for dategraph

bryce-fitzsimons:
### Features and Changes

We were unintentionally mutating the underlying data, twice (one per each stacked graph), leading to some minor discrepancies.


jdorn (approved):

lukesonnet (approved):
fixes the bug on my local tysm

PR #2811, opened 2024-07-29
i think i got all the places

gazzdingo:
### Features and Changes
updated so we set the seed to UUID on creation. we default back to tracking key for older experiments
you can now edit the seed in the phases modal

### DIFF
--- a/packages/front-end/components/Experiment/EditPhaseModal.tsx
+++ b/packages/front-end/components/Experiment/EditPhaseModal.tsx
@@ -3,6 +3,8 @@ import {
   ExperimentInterfaceStringDates,
   ExperimentPhaseStringDates,
 } from "back-end/types/experiment";
+import { useState } from "react";
+import { PiCaretDown, PiCaretUp } from "react-icons/pi";
 import { useAuth } from "@/services/auth";
 import Field from "@/components/Forms/Field";
 import Modal from "@/components/Modal";
@@ -26,12 +28,15 @@ export default function EditPhaseModal({
   const form = useForm<ExperimentPhaseStringDates>({
     defaultValues: {
       ...experiment.phases[i],
+      seed: experiment.phases[i].seed ?? experiment.trackingKey,
       dateStarted: (experiment.phases[i].dateStarted ?? "").substr(0, 16),
       dateEnded: experiment.phases[i].dateEnded
         ? (experiment.phases[i].dateEnded ?? "").substr(0, 16)
         : "",
     },
   });
+  const [advancedOptionsOpen, setAdvancedOptionsOpen] = useState(false);
+
   const { apiCall } = useAuth();
 
   const isDraft = experiment.status === "draft";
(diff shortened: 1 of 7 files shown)


packages/front-end/components/Experiment/EditPhaseModal.tsx
```diff
@@ -110,6 +115,31 @@ export default function EditPhaseModal({
...
+      {advancedOptionsOpen && (
+        //edit seed
+        <Field
+          label="Seed"
+          type="input"
+          {...form.register("seed")}
+          helpText={
+            <>
+              <strong className="text-danger">Warning:</strong> Changing this
+              will change bucketing for you variations.
```
bryce-fitzsimons, line 128: maybe more like "Changing this will re-randomize experiment traffic" (check with the Lukes though)

packages/front-end/components/Experiment/EditPhasesModal.tsx
```diff
@@ -146,15 +147,17 @@ export default function EditPhasesModal({
...
-        <button
-          className="btn btn-primary"
-          onClick={(e) => {
-            e.preventDefault();
-            setEditPhase(-1);
-          }}
-        >
-          <GBAddCircle /> New Phase
-        </button>
+        <>
```
bryce-fitzsimons, line 150:
```suggestion
```

packages/front-end/components/Experiment/EditPhasesModal.tsx
```diff
@@ -146,15 +147,17 @@ export default function EditPhasesModal({
...
+          <button
+            className="btn btn-primary"
+            onClick={(e) => {
+              e.preventDefault();
+              setEditPhase(-1);
+            }}
+          >
+            <GBAddCircle /> New Phase
+          </button>
+        </>
```
bryce-fitzsimons, line 160:
```suggestion
```

packages/front-end/components/Experiment/EditPhaseModal.tsx
gazzdingo: yeah i will check with Luke

gazzdingo: @lukesonnet

gazzdingo: @bryce-fitzsimons this was the same language on the Docs BTW

bryce-fitzsimons (approved):
looks good. probably should run it by Luke before deploying just in case...

packages/front-end/components/Experiment/EditPhaseModal.tsx
lukesonnet: "Changing this will re-randomize experiment traffic" is good to me.

gazzdingo:
> This looks good to me!
> 
> Does this cover all experiment creation workflows?
> 
> e.g. API, create from feature, create from experiment?
> 
> I'm not 100% sure whether all of them (including the REST API), hit `postExperiments`.

Let me have a look at that I could just put it further down in the creation flow and that will cover everything

lukesonnet (approved):
There's one unfinished copy suggestion in a comment. Otherwise, lgtm!

PR #2821, opened 2024-07-30
Update SCIM API to support Azure

mknowlton89:
### Features and Changes

This PR does two things: 1, it updates GrowthBook's SCIM API to work with Azure AD / Microsoft Entra ID and it also adds a how-to guide on setting up SCIM with Azure AD. / Microsoft Entra ID

A quick note: Azure AD and Microsoft Entra ID are the same platform - Azure AD has just been renamed to Microsoft Entra ID.

One piece of functionality that is missing from the how-to guide is how to define a user's role when provisioning them - I'm still working out how to do that, and plan to release a follow-up PR to update the guide. That isn't required to get the integration working.

### Testing
- [x] Ensure that you can provision users and groups from Azure AD to GrowthBook successfully
- [x] Ensure that you can de-provision users and groups from Azure AD to GrowthBook successfully
- [x] Ensure that when you add members to a provisioned group in Azure - those members are added to the corresponding GB team correctly
- [x] Ensure that this doesn't introduce any regressions with Okta's SCIM integration


docs/static/images/integrations/azure/growthbook-entra-scim-group-mapping.png
royalfig, line 1: What's the focus here? See my other comments in Slack

msamper:
How did you set up a test Azure AD account for this? I'm trying use their free trial, but it keeps asking me for a credit card. Not sure if there's another way to go about it

mknowlton89:
> How did you set up a test Azure AD account for this? I'm trying use their free trial, but it keeps asking me for a credit card. Not sure if there's another way to go about it

Unfortunately, I wasn't able to find a way around having to put a CC on file. I also set up an account and I'm still within the free trial period. If you'd like, since this is just a dev sandbox, I can share my login credentials with you.

packages/back-end/src/scim/users/listUsers.ts
```diff
@@ -51,12 +51,12 @@ export async function listUsers(
 
   return res.status(200).json({
     schemas: ["urn:ietf:params:scim:api:messages:2.0:ListResponse"],
-    totalResults: filteredUsers.length,
+    totalResults: resources.length,
```
msamper, line 54: I think the reason this was `filteredUsers` instead of `resources` is that `totalResults` is supposed to be the number of results regardless of the `count` or `startIndex` requested, but I have to look at the spec for SCIM to make sure

msamper (commented):
Was able to verify that provisioning and de-provisioning users works as expected. Wasn't able to test groups because my Azure account apparently doesn't have the right plan for that. For the docs, my only recommendation would be to indicate what the action or focus point is so its more clear to someone following them to setup the GrowthBook SCIM application. Other than that, I was able to follow the docs fairly easily and set up my test.

mknowlton89:
> Was able to verify that provisioning and de-provisioning users works as expected. Wasn't able to test groups because my Azure account apparently doesn't have the right plan for that. For the docs, my only recommendation would be to indicate what the action or focus point is so its more clear to someone following them to setup the GrowthBook SCIM application. Other than that, I was able to follow the docs fairly easily and set up my test.

@msamper Do you mean in the images themselves? Like highlighting with arrows/squares, or via the copy?

msamper:
> > Was able to verify that provisioning and de-provisioning users works as expected. Wasn't able to test groups because my Azure account apparently doesn't have the right plan for that. For the docs, my only recommendation would be to indicate what the action or focus point is so its more clear to someone following them to setup the GrowthBook SCIM application. Other than that, I was able to follow the docs fairly easily and set up my test.
> 
> @msamper Do you mean in the images themselves? Like highlighting with arrows/squares, or via the copy?

I think in the images with arrows and circles but I defer to Ryan if he has different opinions!

packages/back-end/src/scim/users/listUsers.ts
msamper: I think the above is true so I think in this case `totalResults` should be `filteredUsers.length`

See https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2

packages/back-end/src/scim/users/listUsers.ts
```diff
@@ -51,12 +51,12 @@ export async function listUsers(
...
     Resources: resources,
     startIndex: queryOptions.startIndex,
     itemsPerPage:
-      resources.length < queryOptions.count
-        ? resources.length
-        : queryOptions.count,
+      !resources.length || resources.length > queryOptions.count
+        ? queryOptions.count
+        : resources.length,
```
msamper, line 60: Hmm do you need the `!resources.length || ` here?

mknowlton89: Looking at the documentation further, I think we can simplify this even more to just:

```ts
itemsPerPage: queryOptions.count,
```

The documentation states:

_"itemsPerPage  The number of resources returned in a list response
      page.  REQUIRED when partial results are returned due to
      pagination."_
      
So it seems like if we aren't returning partial results, the field is ignored. I'll do some testing around this today.

msamper (approved):
lgtm!

PR #2825, opened 2024-08-01
Move interface definitions to validators.

romain-growthbook:
### Features and Changes


packages/back-end/src/validators/event-webhook.ts
```diff
@@ -0,0 +1,35 @@
...
+    url: z.string().url(),
+    name: z.string().trim().min(2),
+    events: z.array(z.enum(zodNotificationEventNamesEnum)).min(1),
+    enabled: z.boolean(),
+    projects: z.array(z.string()),
+    tags: z.array(z.string()),
+    environments: z.array(z.string()),
+    payloadType: z.enum(eventWebHookPayloadTypes),
+    method: z.enum(eventWebHookMethods),
+    headers: z.object({}).catchall(z.string()),
```
jdorn, line 27: Instead of object and `catchall`, should we do this?
```ts
z.record(z.string())
```

romain-growthbook: Definitely.

romain-growthbook, line 27:
```suggestion
    headers: z.record(z.string(), z.string()),
```

packages/back-end/src/validators/experiments.ts
```diff
@@ -0,0 +1,183 @@
...
+    owner: z.string(),
+    datasource: z.string(),
+    exposureQueryId: z.string(),
+    /**
+     * @deprecated Always set to 'code'
+     */
+    implementation: z.enum(implementationType),
+    /**
+     * @deprecated
+     */
```
jdorn, line 123: The `@deprecated` comments when this was a Typescript interface were used by IDEs in code completion, but these are not.  I don't see a good alternative that works with Zod though, so I think it's ok, just a little unfortunate.

jdorn (approved):

PR #2826, opened 2024-08-01
Added the ability to edit visual experiment changes

Auz:
Screenshot 2024-07-31 at 10 48 17 PM

Previous version just let you delete rules

### DIFF
--- a/packages/front-end/components/Experiment/EditDOMMutationsModal.tsx
+++ b/packages/front-end/components/Experiment/EditDOMMutationsModal.tsx
@@ -1,8 +1,8 @@
 import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
 import { VisualChange } from "back-end/types/visual-changeset";
-import { FC, useCallback, useState } from "react";
-import Code from "@/components/SyntaxHighlighting/Code";
+import { FC, Fragment, useCallback, useState } from "react";
 import Modal from "@/components/Modal";
+import Field from "@/components/Forms/Field";
 
 const EditDOMMutatonsModal: FC<{
   experiment: ExperimentInterfaceStringDates;
@@ -13,12 +13,15 @@ const EditDOMMutatonsModal: FC<{
   const [newVisualChange, setNewVisualChange] = useState<VisualChange>(
     visualChange
   );
+  const [css, setCss] = useState(visualChange.css);
+  const [js, setJs] = useState(visualChange.js);
 
   const deleteCustomJS = useCallback(() => {
     setNewVisualChange({
       ...newVisualChange,
       js: "",
     });
+    setJs("");
   }, [newVisualChange, setNewVisualChange]);
 
   const deleteGlobalCSS = useCallback(() => {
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Experiment/EditDOMMutationsModal.tsx
```diff
@@ -13,19 +13,23 @@ const EditDOMMutatonsModal: FC<{
   const [newVisualChange, setNewVisualChange] = useState<VisualChange>(
     visualChange
   );
+  const [css, setCss] = useState(visualChange.css);
+  const [js, setJs] = useState(visualChange.js);
```
jdorn, line 17: Why have separate state for these?  Everywhere you're updating state, you are also having to call `setNewVisualChange` anyway

packages/front-end/components/Experiment/EditDOMMutationsModal.tsx
```diff
@@ -65,67 +81,82 @@ const EditDOMMutatonsModal: FC<{
...
-                  containerClassName="w-100"
+                <Field
+                  textarea
+                  minRows={2}
+                  value={JSON.stringify(m)}
+                  className="w-100 my-3"
+                  onChange={(e) => {
+                    try {
+                      const newMutation = JSON.parse(e.target.value);
+                      setDOMMutation(i, newMutation);
```
jdorn, line 153: We need more validation here.  Otherwise, someone could cause some serious bugs in the SDK
1. All required fields are set
2. There are no extra fields
3. The `action` is a valid value

Here's a type to work from for validation: https://github.com/growthbook/dom-mutator/blob/main/src/index.ts#L515

Auz: Done

packages/front-end/components/Experiment/EditDOMMutationsModal.tsx
Auz: done

packages/front-end/components/Experiment/EditDOMMutationsModal.tsx
```diff
@@ -40,7 +47,100 @@ const EditDOMMutatonsModal: FC<{
...
+    },
+    [setDOMMutation, setDOMMutationErrors]
+  );
+
   const onSubmit = () => {
+    // make sure all DOM mutations are valid
+    newDOMMutationStr.forEach((m, i) => {
+      validateDOMMutations(i, m);
+    });
+    if (newDOMMutationErrors.some((e) => e)) {
```
jdorn, line 141: This won't work as expected.  The `setDOMMutationErrors` call within `validateDOMMutations` doesn't take effect until the next render.  This will likely still work because you're also calling `validateDOMMutations` when the input changes, but it is a potential race condition.  Better to have the validate method return true or false and check the return value here.

jdorn (approved):
Added a comment about a race condition, but it's not a huge deal.  I'm fine if you want to merge without fixing.

PR #1286, opened 2023-05-15
Version string comparison

tinahollygb:
### Features and Changes

This is an alternative to this solution: https://github.com/growthbook/growthbook/pull/1282 

It was requested that the implementation be around 100 bytes and the initial implementation was around 610 bytes. The original version enforces a valid semantic version while this newer approach is more flexible, allowing any string to be compared, splitting the string into segments on the dot and dash characters, and comparing each segment alphanumerically, while padding digit-only segments so they can be compared as strings. This also means that a version comparison like `v1.2.3 == 1-2.3` is true since it treats `.` and `-` characters the same.

Co-authored by @jdorn 

- Closes https://github.com/growthbook/growthbook/issues/672

### Dependencies

n/a

### Testing

See the tests and the operators.

PR #2828, opened 2024-08-01
fix(type-error): fix misleading compatible version suggestions

Hephaest:
#### Backgrounds
Our mobile engineers created a new attribute and selected the Version string. Unfortunately, this didn't work even though we're in minimum SDK requirements(JavaScript & Node.js). **The worst thing is this bug can CRASH our app!**
image

image

I spent some time investigating the root cause and realized the tooltip hint is misleading... That's why I raised the PR and hope you guys can help merge it as soon as possible, to avoid any future damage to other businesses.

#### Context
Since our mobile product is based on React Native and uses Redux Saga under the hook, `@growthbook/growthbook` has become a more flexible SDK for us to integrate with.

### Dependencies

None

### Testing

- Step 1: install `@growthbook/growthbook` as `0.27.0`
- Step 2: create a new string attribute called 'mobileVersion' and select the `Version String`
- Step 3: Enable the rule on test environment
- Step 4: do not set this attribute on the codebase
- Step 5: load and evaluate your feature flag on the local server as normal

### Screenshots
My PR's UT test results:
image

Bugfix result:
image


packages/shared/src/sdk-versioning/sdk-versions/javascript.json
```diff
@@ -33,15 +33,15 @@
     },
     {
       "version": "0.30.0",
-      "capabilities": ["visualEditorDragDrop"]
+      "capabilities": ["semverTargeting", "visualEditorDragDrop"]
```
Hephaest, line 36: `semverTargeting` is only safe to use when the SDK version is `0.30.0` or above.

This is the original commit in `0.27.0` that crashes our app: https://github.com/growthbook/growthbook/pull/1286/files

And later it's fixed in `0.30.0`: https://github.com/growthbook/growthbook/pull/1286/files

Hephaest: Hi @jdorn Thanks to the reply! 

As you said "as long as the attribute is passed into the SDK", but old version app has been publish to app stores and they can't pass this new attribute so they crashed (as screenshot is shown), that's why I found this is quite serious. From my understanding, the minimum version should still meet **everyone**'s basic needs, not just most of the users. Otherwise, it's still misleading, right?

In addition, I don't think updating the changelog can make a difference. Not every engineer will read changelog, they probably just trust this UI tip instead! 😂

But I respect your decision, if you still think this is not a risk I will close the PR.

Hephaest: Dude, I think you miss the mobile case:
image

React Native use `@growthbook/growthbook` under the hook, so this is not a edge case.

PR #2829, opened 2024-08-01
Add DB migration facility

romain-growthbook:
This PR establishes the basic framework to add DB migrations. Migrations are statically declared for now and applied at startup. Applied migrations are stored in the database.

### DIFF
--- a/packages/back-end/src/migrations.ts
+++ b/packages/back-end/src/migrations.ts
@@ -1,0 +1,60 @@
+import { Collection } from "mongodb";
+import bluebird from "bluebird";
+import mongoose from "mongoose";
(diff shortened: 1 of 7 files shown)


packages/back-end/src/migrations.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (existingMigrations.includes(name)) {
+      console.log(
+        `Migration ${name} on collection ${collection} already applied!`
+      );
+      return;
+    }
+
+    console.log(`Applying migration ${name} on collection ${collection}..`);
+    await apply(c);
+    await appendMigration(collection, name);
```
romain-growthbook, line 112: Race condition.

PR #2835, opened 2024-08-02
Get actual list of envs affected by experiment

mknowlton89:
### Features and Changes

This PR updates the logic for who can start and experiment. Previously, if a user had an `experimenter` role and had any environment restrictions (e.g. `limitAccessByEnvironment === true`) they would not be able to start an experiment as `getAffectedEnvsForExperiment` would return `["__ALL__"]` instead of the environments actually affected.

Now, we get the linkedExperiments if there are any, and only return the environments where the experiment would be active.

So now, a user with a global `experimenter` role who could only affect the `STAGING` environment can start an experiment if it only affects the `STAGING` environment. If the experiment affected `STAGING` and `PRODUCTION` or only `PRODUCTION` the user would not have permission to start the experiment.

### Testing
Publishing
- [x] Ensure that a user with a global `experimenter` role and no env access limits can start any draft experiment
- [x] Ensure that a user with a global `experimenter` role and `STAGING` env access limits can start experiment that only affects `STAGING`
- [x] Ensure that a user with a global `experimenter` role and `STAGING` env access limits can NOT start experiment that only affects `PRODUCTION`
- [x] Ensure that a user with a global `experimenter` role and `STAGING` env access limits can NOT start experiment that affects `STAGING` AND `PRODUCTION`
- [x] Ensure that a user with a global `experimenter` role and `STAGING` env access limits can NOT start experiment that has a URL redirected test
- [x] Ensure that a user with a global `experimenter` role and no env access limits CAN start experiment that has a URL redirected test
- [x] Ensure that a user with a global `experimenter` role and no env access limits CAN start experiment that has a visual editor changeset
- [x] Ensure that a user with a global `experimenter` role and `STAGING` env access limits can start experiment that has no linked feature flag, no url redirect test, and no visual editor changeset

Editing - These are based on an experimenter user who’s experimenter permissions are limited to “staging” and “test-env”
- [x] Ensure that if an experiment doesn’t affect ANY environments, the user has permission to edit (e.g. an experiment’s linked feature doesn’t have any enabled rules)
- [x] Ensure that if an experiment only affects environments that the user has experimenter permissions for, they have permission to edit (e.g. an experiment’s linked feature has an enabled experiment rule on for the “staging” env (should also work if there are two rules enabled, one for “staging” and one for “test-env)
- [x] Ensure that if an experiment affects environments other than those a user has experimenter permissions for, they don’t have permission to edit - (e.g an experiment’s linked feature has enabled experiment rules for “staging” and “production” - this user doesn’t have experimenter permissions for “production”)


packages/shared/src/util/index.ts
```diff
@@ -19,16 +20,26 @@ export * from "./features";
...
+
+    linkedFeatures.forEach((linkedFeature) => {
+      for (const env in linkedFeature.environmentStates) {
+        if (linkedFeature.environmentStates[env] === "active") {
+          envs.push(env);
+        }
+      }
+    });
+    return envs;
   }
```
jdorn, line 43: Since the `linkedFeatures` argument is optional, should we have a fallback that returns `["__ALL__"]` if it's not passed in and `experiment.linkedFeatures` is non-empty?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -873,8 +873,10 @@ export async function postExperiment(
     "excludeFromPayload",
   ] as (keyof ExperimentInterfaceStringDates)[]).some((key) => key in changes);
   if (needsRunExperimentsPermission) {
+    const linkedFeatures = await getLinkedFeatureInfo(context, experiment);
```
jdorn, line 876: `getLinkedFeatureInfo` is a very expensive function and overkill for what we need.  It fetches all revisions of all linked features when we only care about the rules in the FeatureInterface itself.  There are some users with features that have thousands of revisions so if we can avoid querying them, that's ideal.

mknowlton89: @jdorn Given this, it feels like there could be a much simpler solution here.

We could update the signature of `getAffectedEnvsForExperiment` to take in the experiment and the linkedFeatures, but the shape of the linkedFeatures is just the `FeatureInterface[]`.

E.G.

```ts
export function getAffectedEnvsForExperiment({
  experiment,
  linkedFeatures,
}: {
  experiment: ExperimentInterface | ExperimentInterfaceStringDates;
  linkedFeatures?: FeatureInterface[];
}): string[] {
  // Visual changesets are not environment-scoped, so it affects all of them
  // Also fallback to all envs if linkedFeatures is undefined, but the experiment does actually have linked features
  if (
    experiment.hasVisualChangesets ||
    experiment.hasURLRedirects ||
    (!linkedFeatures && !!experiment.linkedFeatures?.length)
  )
    return ["__ALL__"];

  if (linkedFeatures?.length) {
    const envs: string[] = [];

    linkedFeatures.forEach((linkedFeature) => {
      const envSettings = linkedFeature.environmentSettings;
      if (!linkedFeature.archived) {
        for (const env in envSettings) {
          if (envSettings[env].enabled) {
            envs.push(env);
          }
        }
      }
    });
   return envs;
  }

  return [];
}
```

packages/back-end/src/controllers/experiments.ts
```diff
@@ -297,14 +298,24 @@ export async function getExperiment(
     experiment.id
   );
 
-  const linkedFeatures = await getLinkedFeatureInfo(context, experiment);
+  const linkedFeatureInfo = await getLinkedFeatureInfo(context, experiment);
+
+  const linkedFeatureIds = experiment.linkedFeatures || [];
+
+  const linkedFeatures = await getFeaturesByIds(context, linkedFeatureIds);
```
jdorn, line 305: We should add an early return to the `getFeaturesByIds` method if the array is empty (right now, it will still run the Mongo query).

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2591,7 +2658,17 @@ export async function deleteVisualChangeset(
     visualChangeset.experiment
   );
 
-  const envs = experiment ? getAffectedEnvsForExperiment({ experiment }) : [];
+  if (!experiment) {
+    throw new Error("Experiment not found");
+  }
```
jdorn, line 2663: I don't think we want to block deleting a visual changeset if the experiment doesn't exist.  For example, this may be used to clean up orphaned documents.

packages/shared/src/util/index.ts
```diff
@@ -20,16 +20,34 @@ export * from "./saved-groups";
...
-  // TODO: get actual environments for linked feature flags. We are being overly conservative here
-  if (experiment.linkedFeatures && experiment.linkedFeatures.length > 0) {
-    return ["__ALL__"];
+  if (linkedFeatures?.length) {
+    const envs: string[] = [];
+
+    linkedFeatures.forEach((linkedFeature) => {
+      const envSettings = linkedFeature.environmentSettings;
+      if (!linkedFeature.archived) {
+        for (const env in envSettings) {
```
jdorn, line 43: We only want to include environments where the experiment rule exists.  There's a helper method `getMatchingRules` we should use for this instead of trying to loop through envs manually.

Also, this method will need to know the list of environments from the organization. We can't rely on the keys in `environmentSettings` because some of those environments might have been deleted in the org.

packages/shared/src/util/index.ts
```diff
@@ -176,6 +208,8 @@ export function getMatchingRules(
       ([environmentId, settings]) => {
         if (!isValidEnvironment(environmentId, environments)) return;
 
+        if (omitDisabledEnvironments && !settings.enabled) return;
```
mknowlton89, line 211: @jdorn This is the risky part I mentioned. It feels like if an environment's feature-level toggle is disabled, we don't want to include the environment, even if the environment has an enabled ff on it.

packages/shared/src/util/index.ts
```diff
@@ -15,22 +15,68 @@ import { FeatureInterface, FeatureRule } from "back-end/types/feature";
...
+          false,
+        orgEnvIds,
+        undefined,
+        // the boolean below skips environments if they are disabled on the feature
+        true
+      );
+      if (matches.length) {
+        const featureProject = linkedFeature.project || "";
+        matches.forEach((matchRule) => {
+          const environmentProjects =
```
bryce-fitzsimons, line 63: Have a look at `featureHasEnvironment()` (shared/src/util/features.ts) which might help simplify and standardize

mknowlton89: Oooh - that's a handy little utility function. Just pushed up some changes that now leverage this [here](https://github.com/growthbook/growthbook/pull/2835/commits/2bd90c7b28b433f389fb32a78ca9da0c96119a7c).

bryce-fitzsimons (approved):
looks good from my perspective

PR #2836, opened 2024-08-02
New webhook framework.

romain-growthbook:
This PR establishes the basis for the new webhook framework:

### Declaration-driven

Events are declared using a constant object with type:

```typescript
{
  [resource: string]: {
    [event: string]: {
      schema: ZodType<any>;
      description: string;
      isDiff: boolean;
    }
}
```

Raw webook payloads are then of the form:
```typescript
{
  event: `${Resource}.${Event}`;
  object: Resource;
  api_version: `${number}-${number}-${number}`;
  created: number;
  data: {
    object: z.infer<typeof schema>,
    ...(isDiff ? { previous_attributes: Partial<z.infer<typeof schema> } : {})
  },
  user: EventUser;
  projects: string[];
  tags: string[];
  environments: string[];
  containsSecrets: boolean;
}
```

* `data` payload always has a top-level `object` with all attributes described by the event's `schema`
* If the event set `isDiff` to `true` then a `previous_attributes` top-level key is also added to the `data` payload. The object it points to contains all top-level fields that have changed.

This also makes it possible to automatically document events and webhooks using the `schema` and `description` for display on the documentation website.

### Backward compatibility

New event payload are currently converted back to legacy payloads before being passed to the `raw` webhoooks. In the future, existing raw webhook should be renamed `legacy` and new `raw` webhook should be expected to consume new event raw payloads.

### Cleanup up

Some API cleanups:
* Create and queue events at once in `createEvent` function.
* Simply arguments of `createEvent`, let the function handle converting existing and previous object into `object` and `previous_attributes`

### Future work

- [x] Add new event
- [x] Add migration to rename existing `raw` webhook to `legacy`
- [x] Wire automatic event and webhook documentation


packages/back-end/src/models/EventModel.ts
```diff
@@ -25,13 +36,18 @@ const eventSchema = new mongoose.Schema({
   object: {
     type: String,
     required: true,
-    enum: notificationEventResources,
+    enum: zodNotificationEventResources,
   },
```
romain-growthbook, line 40: objectId?: number,

romain-growthbook:
Add more unit testing

packages/back-end/src/models/EventWebhookModel.ts
```diff
@@ -165,8 +167,15 @@ type EventWebHookDocument = mongoose.Document & EventWebHookInterface;
  * @param doc
  * @returns
  */
-const toInterface = (doc: EventWebHookDocument): EventWebHookInterface =>
-  omit(doc.toJSON<EventWebHookDocument>(), ["__v", "_id"]);
+const toInterface = (doc: EventWebHookDocument): EventWebHookInterface => ({
+  ...omit(doc.toJSON<EventWebHookDocument>(), ["__v", "_id"]),
+  method: doc.method || "GET",
```
jdorn, line 172: Is this right?  I thought the default for old webhooks was `POST`

packages/back-end/src/routers/event-webhooks/event-webhooks.controller.ts
```diff
@@ -372,46 +370,11 @@ export const createTestEventWebHook = async (
...
-
-  const emittedEvent = await createEvent(organizationId, payload);
-
-  if (!emittedEvent) throw new Error("Error while creating event!");
-
-  new EventNotifier(emittedEvent.id).perform();
+  await sendEventWebhookTestEvent(context, webhookId);
 
-  return res.json({ eventId: emittedEvent.id });
+  return res.status(200);
```
jdorn, line 377: Should update the `apiCall` return type in `EventWebHookDetail.tsx` to reflect that `eventId` is no longer being returned.

packages/back-end/src/validators/experiment-warnings.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from "zod";
+
+export const autoUpdateFailed = z
+  .object({
+    type: z.literal("auto-update"),
+    success: z.boolean(),
```
jdorn, line 6: I should have caught this earlier, but we should not be firing a `warning` event when a snapshot update is successful.  Successful snapshots can happen hundreds of times a day for some orgs, so coupling that to the same event as actual warnings makes it impossible to use it for a Slack notification.  There will be so much noise in the Slack channel, you won't be able to see actual warnings

packages/front-end/components/Events/EventsPage/utils.ts
```diff
@@ -1,52 +1,76 @@
...
-
-export const getEventText = (
-  event: EventInterface<NotificationEvent>
-): string => {
+import { EventInterface } from "back-end/types/event";
+
+// region Feature
+
+const getTitleForFeatureCreated = ({ id }: { id: string }) =>
+  `The feature ${id} was created`;
```
jdorn, line 6: These descriptions should probably not be generated on the front-end like this.  It seems like this would be pretty much the same logic as building a Slack message, or if we add email notification in the future, a subject line.

Feels like we need a centralized way to convert an event into a short human-readable description (or maybe a few different versions with varying verbosity).  We'll also likely need slightly different formatting (e.g. Markdown for Slack, HTML for email, Plain Text for other things).

packages/back-end/src/models/EventModel.ts
```diff
@@ -90,62 +106,174 @@ const eventSchema = new mongoose.Schema({
...
+  containsSecrets,
+  projects,
+  tags,
+  environments,
+}: CreateEventParams<Resource, Event>) =>
+  createEventWithPayload<Resource, Event>({
+    payload: {
+      event: `${object}.${event}`,
+      object,
+      data: diffData(data),
```
jdorn, line 251: Right now, we expect people to pass in the full previous object as `previous_attributes` and then we rewrite it here to be a filtered version with only the changed keys.  I don't know if this is super intuitive and might be too much magic happening internally.

I think it might be clearer if we're a little more explicit, something like this:

```ts
createEvent({
   data: {
      object: feature,
      previous_attributes: getPreviousAttributesFromDiff(feature, prevFeature)
   },
   ...
})
```

Basically, just bring this diff logic up to where `createEvent` is being called instead of hiding it internally.

packages/back-end/src/models/EventWebhookModel.ts
romain-growthbook, line 172:
```suggestion
  method: doc.method || "POST",
```

packages/back-end/src/models/EventModel.ts
romain-growthbook: I usually prefer encapsulating everything to prevent crazy node dependencies and also make refactorization/reduce boilerplate. Happy to do that, tho, I understand the other side of this argument.

packages/back-end/src/events/base-types.ts
```diff
@@ -1,26 +1,104 @@
...
 import { EventUser } from "./event-types";
 
+type Webhook = {
+  readonly [key: string]: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    readonly schema: ZodType<any>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    readonly extra?: ZodType<any>;
+    readonly description: string;
+    readonly isDiff?: boolean;
```
romain-growthbook, line 18: Add version

packages/front-end/components/Events/EventsPage/utils.ts
romain-growthbook: I agree. This PR is mostly re-organizing code. Do you want to add that to the current work or as a follow-up?

packages/back-end/src/models/EventModel.ts
```diff
@@ -25,13 +37,22 @@ const eventSchema = new mongoose.Schema({
   object: {
     type: String,
     required: true,
-    enum: notificationEventResources,
+    enum: zodNotificationEventResources,
+  },
+  objectId: {
+    type: String,
+    require: false,
```
jdorn, line 44: Typo? `required` vs `require`

packages/back-end/src/services/users.ts
```diff
@@ -121,42 +119,40 @@ export async function trackLoginForUser({
...
   try {
     // Create a login event for all of a user's organizations
-    const eventCreatePromises = organizationIds.map(async (organizationId) => {
-      const emittedEvent = await createEvent(organizationId, event);
-      if (emittedEvent) new EventNotifier(emittedEvent.id).perform();
-    });
+    const eventCreatePromises = organizationIds.map((organizationId) =>
+      createEventWithPayload({
+        payload: {
+          object: "user",
```
jdorn, line 127: Is there a reason we aren't setting `objectId: user.id` here?

jdorn (approved):
Added 2 minor comments and approving

packages/back-end/src/models/EventModel.ts
romain-growthbook: Thanks!

packages/back-end/src/services/users.ts
romain-growthbook: Good catch thanks!

PR #2837, opened 2024-08-02
don't return SUM(value) when userAggregationSQL is empty

itsgrimetime:
### Features and Changes
using just `||` to check if `userAggregationSQL` should be returned as the default value (`'SUM(value)'`) results in returning `'SUM(value)'` when `userAggregationSQL` is unset (empty string). Using `??` will change this to only return the default when its null/undefined (which can't be set through the API, but is possible when creating a new binomial metric).

### Testing

Been using this for our deployment to be able to correctly diff metrics when updating via API.


jdorn:
Should we just always set it to an empty string for binomial metrics?

PR #2841, opened 2024-08-03
Project scoped Archetypes

Auz:
Screenshot 2024-08-03 at 1 19 08 PM


packages/back-end/src/models/ArchetypeModel.ts
```diff
@@ -54,24 +55,32 @@ export function parseArchetypeString(list: string) {
...
-    organization,
-  });
+  const query: FilterQuery<ArchetypeDocument> = {
+    organization: organization,
+    $or: [{ projects: { $exists: false } }, { projects: { $eq: [] } }], // returns archetypes that are not associated with a project (all projects)
+  };
+  // if project is set, return archetypes that are either all projects
+  // or are associated with the specified project
+  if (project && query["$or"]) {
+    query["$or"].push({ projects: { $eq: project } });
```
jdorn, line 81: Does this work? Checking if an array field equals a string?

packages/back-end/src/routers/archetype/archetype.controller.ts
```diff
@@ -147,6 +149,7 @@ type CreateArchetypeRequest = AuthRequest<{
   owner: string;
   isPublic: boolean;
   attributes: string;
+  projects?: string[];
```
jdorn, line 152: Also need to update all of the permission functions related to archetypes to be project scoped.

packages/back-end/src/models/ArchetypeModel.ts
Auz: Yes, I believe so

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -400,6 +389,33 @@ export class Permissions {
...
+  ): boolean => {
+    return this.checkProjectFilterPermission(
+      { projects: archetype?.projects ? archetype.projects : [] },
+      "manageArchetype"
+    );
+  };
+
+  public canUpdateArchetype = (
+    archetype: Pick<ArchetypeInterface, "projects">
+  ): boolean => {
```
jdorn, line 403: For update permission checks, you need to have two arguments: the old object and the new one (in case the list of projects is being changed).

Instead of `checkProjectFilterPermission`, you can use `checkProjectFilterUpdatePermission`.

packages/back-end/src/routers/archetype/archetype.controller.ts
```diff
@@ -235,7 +240,7 @@ export const putArchetype = async (
     });
   }
 
-  if (!context.permissions.canUpdateArchetype()) {
+  if (!context.permissions.canDeleteArchetype(req.body)) {
```
jdorn, line 243: This needs to change back to `canUpdateArchetype`

jdorn (approved):

PR #2851, opened 2024-08-07
Add User & Team Rest API Endpoints

mknowlton89:
### Features and Changes

This PR introduces 3 new REST API endpoints for managing an organization's members.

- listMembers - GET `/api/v1/members` - takes in optional filters for `userName`, `userEmail`, and `globalRole`. Also supports `limit` and `offset` filters
- updateMemberRole - POST `/api/v1/members/:id/role` - allows an organization to update a members role - can update their global role, as well as env restrictions (if the role supports it), can also update a member's projectRoles array. 
- removeMember - DELETE `/api/v1/members/:id` - removes a member from an organization

All three of these are available for all organizations, regardless of their plan level. There are validations in place for the `updateMemberRole` endpoint for various commercial features (e.g. if an org tries to set a user with the `noaccess` role, we check they have that commercial feature, and we check for the `advanced-permissions` commercial feature if trying to restrict an engineer/experimenters environment permissions, or specificying project-level permissions.

I do have a few follow-up PRs planned for `inviteMembers` that will allow an org to invite multiple users, along with an `addMembers` that will allow SSO orgs to skip the invite process and simply add members to their organization.

### Testing

**listMembers**
- [x] Ensure that you can get a list of all users for an organization
- [x] Ensure that the limit filter limits accordingly
- [x] Ensure that the filter query works accordingly
- [x] Ensure that you can return a single user by userEmail
- [x] Ensure that querying by userName returns all users with that userName
- [x] Ensure that querying by globalRole returns all users with that globalRole

**removeMember**
- [x] Ensure that you can remove a member from an organization successfully, and that it just removes the user from the org, it doesn't delete the user from Mongo

**updateMemberRole**
- [x] Ensure that you can update a member's role information with valid data
- [x] Ensure that you can not update a member's role with `collaborator` and pass in `environments` - ensure it throws an error
- [x] Ensure that you can not update a member's role with `experimenter` and pass in `environments` if the org doesn't have the `advanced-permissions` commercial feature
- [x] Ensure that you can not pass in any `projectRoles` if the org doesn't have the `advanced-permissions` commercial feature
- [x] Ensure that you can not give a user the `noaccess` role if the org doesn't have the `no-access-role` commercial feature
- [x] Ensure that you can not update a user's role if they have the `managedByIdp` property set to true
- [x] Ensure that if the role supports env restrictions, and you pass in an `environments` array, the `limitAccessByEnvironments` property is set to true
- [x] Ensure that if you pass in a `projectRoles` array, and the role and envs are valid, it updates the user's projectRoles accordingly
- [x] Ensure that if you pass in an invalid `environment` in the `environments` array, we throw an error and don't process the update

**I haven't built any sort of logic to handle superAdmins - the user I tested all of this out with is a superAdmin - I created a personal access token, but that token still only references the organization I created the PAT within. Not sure if there are any special cases I need to handle for superAdmins here.**


packages/back-end/src/api/members/updateMemberRole.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    if (environments?.length) {
+      if (!orgHasPremiumFeature(org, "advanced-permissions")) {
+        throw new Error(
+          "Must have a commercial License Key to restrict permissions by environment."
+        );
+      }
+      environments.forEach((env) => {
+        const environmentIds =
+          org.settings?.environments?.map((e) => e.id) || [];
```
jdorn, line 39: nit: Move this outside the forEach since it's the same for each iteration

packages/back-end/src/api/members/updateMemberRole.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    const { member } = req.body;
+
+    const updatedEnvironments = member.environments || orgUser.environments;
+
+    const updatedMember: Member = {
+      ...orgUser,
+      role: member.role || orgUser.role,
+      environments: updatedEnvironments,
+      limitAccessByEnvironment: !!updatedEnvironments.length,
```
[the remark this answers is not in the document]
jdorn: I think the payload design is fine and makes sense.  The specific situation I'm thinking of is if the current member in Mongo looks like this:

```ts
{
   role: "engineer",
   environments: ["staging"],
   limitAccessByEnvironments: false
}
```

And they call the API with just:

```ts
{
  role: "experimenter"
}
```

I would expect `limitAccessByEnvironments` to remain `false` (keep what's in Mongo because envs aren't being passed into the API).  But because we're spreading in the existing doc and the payload, it will change it to be true.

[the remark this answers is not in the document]
mknowlton89: Ah - I see. I thought we reset the `environments` array to an empty array if the `limitAccessByEnvironments` toggled was toggled to "off". 

I can update the logic to:

```ts
    const updatedMember: Member = {
      ...orgUser,
      role: member.role || orgUser.role,
      environments: member.environments || orgUser.environments,
      // If environments are passed in, recalculate limitAccessByEnvironment, otherwise, use existing value
      limitAccessByEnvironment: member.environments
        ? !!member.environments.length
        : orgUser.limitAccessByEnvironment,
    };
```

packages/back-end/src/api/members/updateMemberRole.ts
```diff
@@ -0,0 +1,185 @@
+import { isRoleValid, roleSupportsEnvLimit } from "shared/permissions";
+import { cloneDeep } from "lodash";
+import { orgHasPremiumFeature } from "enterprise";
+import { updateOrganization } from "../../models/OrganizationModel";
```
romain-growthbook, line 4: We should start using absolute paths:
```suggestion
import { updateOrganization } from "@back-end/src/models/OrganizationModel";
```

romain-growthbook (commented):
It would be nice to add tests. I have started some here: https://github.com/growthbook/growthbook/tree/main/packages/back-end/test/api Happy to help with adding more.

packages/back-end/src/api/members/updateMemberRole.ts
jdorn: Let's hold off on using absolute paths for now since there are some issues with our prod builds.

packages/back-end/src/api/members/updateMemberRole.ts
```diff
@@ -88,18 +88,14 @@ export const updateMemberRole = createApiRequestHandler(
...
     const updatedMember: Member = {
       ...orgUser,
       role: member.role || orgUser.role,
-      environments: updatedEnvironments,
-      limitAccessByEnvironment: updatedLimitAccessByEnv,
+      environments: member.environments || orgUser.environments,
+      // If environments are passed in, recalculate limitAccessByEnvironment, otherwise, use existing value
+      limitAccessByEnvironment: member.environments
+        ? !!member.environments.length
+        : orgUser.limitAccessByEnvironment,
```
jdorn, line 98: Original role:
```ts
{
  role: "engineer",
  environments: ["dev", "production"],
  limitAccessByEnvironment: true
}
```

API Request:
```ts
{
  role: "admin"
}
```

Right now, I believe this PR will just throw an error, which is fine, but it might be better to just always set `limitAccessByEnvironment` to `!!member.environments?.length` instead of having that ternary.

jdorn (approved):

PR #2856, opened 2024-08-08
try fixing oauth caching

itsgrimetime:
### Features and Changes

okta call duration proportion (avg, then sum):
Screenshot 2024-08-08 at 9 33 23 AM

Screenshot_2024-08-08_at_9 28 05_AM

I think the way that the middleware was being created caused it to actually never cache the keys from an OAuth provider (in our case okta)

### Testing
Still testing this out, but there's definitely a latency reduction, and reduced call volume to `*.okta.com/api/v1/keys`. Gonna test it out a bit longer to make sure nothing is broken with SSO/OAuth


packages/back-end/src/services/auth/OpenIdAuthConnection.ts
```diff
@@ -159,19 +161,22 @@ export class OpenIdAuthConnection implements AuthConnection {
...
-          rateLimit: true,
-          jwksRequestsPerMinute: 5,
-          jwksUri,
-        }),
-        audience: connection.clientId,
-        issuer,
-        algorithms,
-      });
-      middleware(req as Request, res, next);
+      if (!jwksClients[jwksUri]) {
```
jdorn, line 164: I think we'll want a compound key here to be safe that hashes together `jwksUri`, `connection.clientId`, `issuer`, and maybe also `algorithms`.  Some IdPs like Google have a single shared jwksUri for everyone and the clientId is the only way to differentiate between tenants.

jdorn:
The clientId is static and doesn't change, so the LRU cache is probably not necessary, but it doesn't hurt. Self-hosted deployments should have exactly 1 cache entry and that's it.

jdorn (approved):

PR #2863, opened 2024-08-12
Fix setting metrics bug on update experiment via API

itsgrimetime:
### Features and Changes
It looks like in #2620 `updateExperimentApiPayloadToInterface` wasn't updated to map `metrics` to `goalMetrics`, which was causing `updateExperiment` to succeed (and update the experiment's `metrics` field, but not `goalMetrics`) which is what's actually used now when generating a new results snapshot.

This change fixes this so experiment metrics can be updated via API.


jdorn:
Thanks.  I see a few other places in that file that also need updating.  I'll push up those additional changes here and get this merged

jdorn (approved):

PR #2876, opened 2024-08-14
Revert "Revert "Actually emit user.login event."" (fix underlying errors)

bryce-fitzsimons:
Reverts growthbook/growthbook#2875
- fix underlying import bugs


romain-growthbook (approved):
Thanks for doing this.

PR #2880, opened 2024-08-15
POC - Auto Fact Metrics

mknowlton89:
### Features and Changes

This PR introduces the concept of automatically generating fact metrics from a fact table. Like other auto-generated metrics and fact tables, this is only supported for `GA4`, `Rudderstack`, `Amplitude`, and `Segment` data sources. 

This PR is meant to only encompass the backend changes necessary to build out basic auto generated fact metrics. I've included 2 front-end files that are only to be used for testing purposes - they are to be removed from the PR before this is merged in.

Once this PR is merged in, there is still work to be done before this is released to GrowthBook organizations. We'll need to:
- Build the front end for this - this means building a new modal to handle fetching the list of auto metrics we can create. This modal will also need to allow users to preview and possibly edit the metrics (mainly thinking they'll want to customize the name, maybe add any filters). We should also explore if we want to limit this feature to only fact tables that we automatically created.
- We'll also want to update the logic of the auto-generated fact tables. We need to update that logic to capture all of the additional columns. Ex: For GA4 datasources, we'll want to see what columns are available via the unnested `event_params`. This will enable us to create more advanced auto fact metrics like `Average Order Value` or `Revenue Per Customer`. **More details are included in the Notion for GB employees.**
- Finally, we'll want to update the logic of the `getAutoFactMetricsToCreate` function to consume these new fields/columns and build more advanced metrics.


packages/front-end/components/AutoGenerateFactMetricsModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+              console.log(e.message);
+            }
+          }}
+        >
+          Get Auto Metrics To Create
+        </Button>
+      </>
+    </Modal>
+  );
+}
```
mknowlton89, line 64: This is just a shell - I think this PR will only be backend changes.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3563,6 +3564,83 @@ export default abstract class SqlIntegration
...
+    });
+  }
+
+  getAutoFactMetricsToCreate(
+    existingFactMetrics: FactMetricInterface[],
+    factTable: FactTableInterface
+  ): AutoFactMetricToCreate[] {
+    const event = factTable.eventName;
+    // based on these docs - https://support.google.com/analytics/answer/9267735?hl=en&ref_topic=13367566&sjid=16477269010846037969-NA
+    const ga4InvertedEvents: string[] = [
```
mknowlton89, line 3587: TODO: Make this more generic and include recommended events from Rudderstack/Segment/Amplitude

packages/front-end/components/AutoGenerateFactMetricsModal.tsx
```diff
@@ -0,0 +1,62 @@
...
```
mknowlton89, line 62: This file should be removed prior to merging this PR. Only meant for testing the backend flow.

mknowlton89:
@jdorn This is ready for an initial review. I expect there to be changes, so not marking it as "ready" just yet.

PR #2884, opened 2024-08-16
treat pending review linked FFs the same as draft

bryce-fitzsimons:
### Features and Changes

Hotfix: the linked features module on an experiment page was failing to include any newly added feature flag, because that change would put the flag into "pending-review" status, leading to a very confusing UX. The linkage only worked for "draft" status flags.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -2207,7 +2207,7 @@ export async function getLinkedFeatureInfo(
 
     const draftMatches =
       revisions
-        .filter((r) => r.status === "draft")
+        .filter((r) => r.status === "draft" || r.status === "pending-review")
         .map((r) => getMatchingRules(feature, filter, environments, r))
         .filter((matches) => matches.length > 0)[0] || [];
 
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -2207,7 +2207,7 @@ export async function getLinkedFeatureInfo(
 
     const draftMatches =
       revisions
-        .filter((r) => r.status === "draft")
+        .filter((r) => r.status === "draft" || r.status === "pending-review")
```
jdorn, line 2210: I think there are more draft-like statuses for approved and changes-requested.

bryce-fitzsimons: thanks, added

romain-growthbook (commented):
Is it possible to add a regression test for this?

bryce-fitzsimons:
> Is it possible to add a regression test for this?

I can't think of an obvious test for this. Something like an enum switch guard may be more helpful. I think we can look into this further after the hotfix lands.

PR #2890, opened 2024-08-16
Add Access URL to Snowflake data source connections

lukesonnet:
### Features and Changes

Adds optional Access URL specification to Snowflake Data Source form: https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-options#additional-connection-options. Uses a toggle because we want it to be undefined for sure if a user doesn't want to use it. If users leave some value in this field by accident it can cause unclear errors.

For advanced configs, this may be helpful; it overrides the `Account` connection method.

### Testing

Can connect to our GrowthBook "SAMPLE" database using this access URL; leaving it unset also works; as does setting and unsetting it.

### Screenshots

Screenshot 2024-08-16 at 3 19 57 PM

Screenshot 2024-08-16 at 3 20 02 PM

### DIFF
--- a/packages/front-end/components/Settings/SnowflakeForm.tsx
+++ b/packages/front-end/components/Settings/SnowflakeForm.tsx
@@ -9,7 +9,7 @@ const SnowflakeForm: FC<{
   onParamChange: ChangeEventHandler<HTMLInputElement>;
   onManualParamChange: (name: string, value: string) => void;
 }> = ({ params, existing, onParamChange, onManualParamChange }) => {
-  const [useAccessUrl, setUseAccessUrl] = useState(false);
+  const [useAccessUrl, setUseAccessUrl] = useState(!!params.accessUrl);
   return (
     <div className="row">
       <div className="form-group col-md-12">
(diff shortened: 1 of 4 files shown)


joornby-angel:
@lukesonnet I ran into a use case where I need this feature.  During my research, I found some other odd things related to snowflake proxy queries before I saw this PR.  The results of that research is in a [PR here](https://github.com/growthbook/growthbook/pull/2941).  What can I do to get this merged and out in the next growthbook release?

packages/front-end/components/Settings/SnowflakeForm.tsx
```diff
@@ -1,12 +1,15 @@
...
+import Toggle from "@/components/Forms/Toggle";
 
 const SnowflakeForm: FC<{
   params: Partial<SnowflakeConnectionParams>;
   existing: boolean;
   onParamChange: ChangeEventHandler<HTMLInputElement>;
-}> = ({ params, existing, onParamChange }) => {
+  onManualParamChange: (name: string, value: string) => void;
+}> = ({ params, existing, onParamChange, onManualParamChange }) => {
+  const [useAccessUrl, setUseAccessUrl] = useState(false);
```
jdorn, line 12: Should the default state value be set to `true` if the existing params already has a non-empty `accessUrl`?  Thinking about editing a data source - if they were using an accessUrl, we want that section to be expanded already when they go to edit.

jdorn (approved):

packages/front-end/components/Settings/SnowflakeForm.tsx
lukesonnet: correct, fixed.

PR #2893, opened 2024-08-19
First POC to generate doc from code.

romain-growthbook:
This PR adds the first POC to generate doc content from the code.

One thing that could be enhanced is the doc generation hook-up. I couldn't find a notion of dependency/generated files directly in the `docusaurus` build pipeline.


docs/package.json
```diff
@@ -6,6 +6,7 @@
     "dev": "docusaurus start --port 3200 --no-open",
     "docusaurus": "docusaurus",
     "start": "docusaurus start --port 3200",
+    "gen": "ts-node -r tsconfig-paths/register ./scripts/gen-event-webhook-doc.ts",
```
romain-growthbook, line 9: * Can this be generated each time a dependent file changes?
* Add result in the git repo/diff.

romain-growthbook: Add pre-commit hook to regenerate.

royalfig:
This is looking good!

It's possible (and probably preferable) to import the markdown without the plugin, `docusaurus-plugin-includes`.  

The docs make it seem like you could do this anywhere, but I got it to work like this:

- Add the generated md file to `/src/partials/` 
- Import the file in the Events file: `import EventWebhookList from "@site/src/partials/_event-webhook-list.md" and use it as a component: <EventWebhookList />

romain-growthbook:
> This is looking good!
> 
> It's possible (and probably preferable) to import the markdown without the plugin, `docusaurus-plugin-includes`.
> 
> The docs make it seem like you could do this anywhere, but I got it to work like this:
> 
> * Add the generated md file to `/src/partials/`
> * Import the file in the Events file and use it as a component:
> 
> ```js
> import EventWebhookList from "@site/src/partials/_event-webhook-list.md"
> 
> <EventWebhookList />
> ```

Thanks, this is done now!

bryce-fitzsimons (approved):

PR #2894, opened 2024-08-19
Add feature flag for opting orgs out of large saved groups

Kevin-Chant:
### Features and Changes

Adds a new feature flag for dogfooding the large saved group changes from #2616. The default behavior is to bypass the limits (and use the old, non-performant saved group implementation) if our dogfood feature isn't enabled. The feature will be default-enabled with specific exceptions for organizations as needed until we can add a better flow in the UI for users being able to opt out on their own.

[Dogfooding feature](https://app.growthbook.io/features/apply-saved-group-id-list-size-limit)

### Testing

Running the app locally with at least one unsupported SDK connection (old version or non-JS), the saved group support warning should pop when adding a new ID List. 
After adding a rule to exclude the local organization in the dogfooding feature, the warnings should disappear and >100 items can be imported without a list being marked as PBRO

### Screenshots

image
image

### DIFF
--- a/packages/front-end/components/SavedGroups/IdListItemInput.tsx
+++ b/packages/front-end/components/SavedGroups/IdListItemInput.tsx
@@ -9,6 +9,7 @@ import {
   FaRetweet,
 } from "react-icons/fa";
 import clsx from "clsx";
+import { useFeatureValue } from "@growthbook/growthbook-react";
 import Field from "@/components/Forms/Field";
 import StringArrayField from "@/components/Forms/StringArrayField";
 import LargeSavedGroupSupportWarning, {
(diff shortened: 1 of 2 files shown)


packages/front-end/components/SavedGroups/IdListItemInput.tsx
```diff
@@ -51,6 +52,10 @@ export const IdListItemInput: FC<{
   );
   const [fileName, setFileName] = useState("");
   const [fileErrorMessage, setFileErrorMessage] = useState("");
+  const orgBypassSmallListSizeLimit = !useFeatureValue(
+    "apply-saved-group-id-list-size-limit",
+    false
```
Kevin-Chant, line 57: I have a default-false here so if the instance doesn't get the GB features payload it won't block users. Then I have the dogfooding feature set up with default-true + opt-out rules for specific orgs. That sound right @jdorn?

PR #2895, opened 2024-08-19
Added a few tracking Calls

gazzdingo:
### Features and Changes
Lukes request For tracking

### DIFF
--- a/packages/front-end/pages/_app.tsx
+++ b/packages/front-end/pages/_app.tsx
@@ -89,6 +89,9 @@ function App({
     growthbook.loadFeatures().catch(() => {
       console.log("Failed to fetch GrowthBook feature definitions");
     });
+    track("page-load", {
+      pathname: router.pathname,
+    });
   }, [router.pathname]);
 
   const renderPreAuth = () => {
--- a/packages/front-end/pages/_app.tsx
+++ b/packages/front-end/pages/_app.tsx
@@ -90,7 +90,7 @@ function App({
       console.log("Failed to fetch GrowthBook feature definitions");
     });
     track("page-load", {
-      pathname: router.pathname,
+      pathName: router.pathname,
     });
   }, [router.pathname]);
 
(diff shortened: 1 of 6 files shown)


packages/front-end/pages/power-calculator/index.tsx
```diff
@@ -115,6 +115,16 @@ const PowerCalculationPage = (): React.ReactElement => {
         <PowerCalculationSettingsModal
           close={() => setShowModal(false)}
           onSuccess={(p) => {
+            track("power-calculation-settings-update", {
+              numberOfMetrics: p.metrics.length,
```
lukesonnet, line 119: There's already a `numMetrics` used for tracking snapshot analyses. Maybe lets change the name to that (you can ctrl-f for it in another `track()` call to confirm).

packages/front-end/pages/power-calculator/index.tsx
```diff
@@ -115,6 +115,16 @@ const PowerCalculationPage = (): React.ReactElement => {
...
+              metricsMetaData: Object.keys(p.metrics).map((m: string) => {
```
lukesonnet, line 120: Why loop over the keys and then call the object with p.metrics? I just don't know why that's the preferred approach.

In any case, looks good, we just need to ensure we can add `metricsMetaData` to our whitelist.

gazzdingo: p.metrics is an object of objects so the first is finding the object key `object.keys` the `map` is just going over the metric object to get the info we want out

lukesonnet (approved):

packages/front-end/pages/_app.tsx
```diff
@@ -89,6 +89,9 @@ function App({
     growthbook.loadFeatures().catch(() => {
       console.log("Failed to fetch GrowthBook feature definitions");
     });
+    track("page-load", {
+      pathname: router.pathname,
```
gazzdingo, line 93:
```suggestion
      pathName: router.pathname,
```

PR #2896, opened 2024-08-20
Fix secureString hashing in saved groups

Kevin-Chant:
### Features and Changes

Fixes a bug where saved groups of secureString attributes stopped hashing the contents sometime in the last week.

### Testing

Added a unit test which fails on main (`git checkout main && git cherry-pick 9d92b11a5 && yarn run test`) but passes with the fix
Manually tested with the sdk feature enabled and disabled

### Screenshots

Hashing in referenced groups (some plain strings as well)
image

Hashing in inline groups
image

### DIFF
--- a/packages/back-end/src/services/features.ts
+++ b/packages/back-end/src/services/features.ts
@@ -35,6 +35,7 @@ import {
   SavedGroupsValues,
   SavedGroupInterface,
 } from "shared/src/types";
+import { clone } from "lodash";
 import {
   ApiReqContext,
   AutoExperimentWithProject,
@@ -1222,7 +1223,8 @@ export function applySavedGroupHashing(
   attributes: SDKAttributeSchema,
   salt: string
 ): SavedGroupInterface[] {
-  return savedGroups.map((group) => {
+  const clonedGroups = clone(savedGroups);
+  clonedGroups.forEach((group) => {
     const attribute = attributes.find(
       (attr) => attr.property === group.attributeKey
     );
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/features.ts
```diff
@@ -1238,32 +1216,27 @@ export function applyExperimentHashing(
...
-        : [savedGroupId, savedGroupItems];
-    })
-  );
+): SavedGroupInterface[] {
+  return savedGroups.map((group) => {
+    const attribute = attributes.find(
+      (attr) => attr.property === group.attributeKey
+    );
+    if (attribute) {
+      group.values = hashStrings({
```
jdorn, line 1230: Instead of mutating the group, should we shallow clone first?

Kevin-Chant: Ah yeah that's in a map huh. I'm gonna add at least one unit test so I'll put that in with it as well.
Not sure how this bug got missed when I explicitly tested this behavior. Must've changed just before final merge or after some of the patching.

jdorn:
Can you add a unit test that reproduces this issue on main and passes in this PR?

jdorn (approved):

packages/back-end/test/features.test.ts
```diff
@@ -1443,4 +1462,100 @@ describe("SDK Payloads", () => {
...
+
+    it("Hashes secure attributes in inline saved groups", async () => {
+      const { features, savedGroups } = await getFeatureDefinitionsResponse({
+        features: { featureName: cloneDeep(featureDef) },
+        experiments: [],
+        dateUpdated: new Date(),
+        projects: [],
+        capabilities: [],
+        savedGroups: [cloneDeep(groupDef)],
+        organization: baseOrganization,
```
jdorn, line 1506: For a more accurate test, `organization.settings.attributeSchema` should also be set to `[secureStringAttr]`.

PR #2900, opened 2024-08-20
Reword large saved group setting

Kevin-Chant:
### Features and Changes

Changes the user-facing SDK setting to better reflect the behavior we'll have going forward. Might need another revision as we finish the code changes but this should be much closer to the current state than the original copy

### Screenshots

Without remote eval
image

With remote eval
image

### DIFF
--- a/packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
+++ b/packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
@@ -1141,7 +1141,7 @@ export default function SDKConnectionForm({
                           Reduce the size of your payload by moving ID List
                           Saved Groups from inline evaluation to a separate key
                           in the payload json. Re-using an ID List in multiple
-                          features/experiments will no longer meaningfully
+                          features or experiments will no longer meaningfully
                           increase the size of your payload.
                         </p>
                         <p>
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -1139,11 +1138,14 @@ export default function SDKConnectionForm({
...
-                          {SMALL_GROUP_SIZE_LIMIT} items, which are not
-                          supported by old SDK versions.
+                          Reduce the size of your payload by moving ID List
+                          Saved Groups from inline evaluation to a separate key
+                          in the payload json. Re-using an ID List in multiple
+                          features or experiments will no longer meaningfully
+                          increase the size of your payload.
                         </p>
                         <p>
+                          This feature is not supported by old SDK versions.
```
bryce-fitzsimons, line 1148: Do we need this text, given it only shows up for new SDKs right? Or is the concern for unversioned and/or related services (remote eval)?

Kevin-Chant: Good point, I think the concern there is for unversioned though (or people not realizing they need to bump their package version)

Kevin-Chant: That said, I think I'd like to get this merged now and refine the copy more when revisiting this

bryce-fitzsimons (approved):

PR #2901, opened 2024-08-20
Fix div by zero error from prior rescaling

lukesonnet:
### Features and Changes

### DIFF
--- a/packages/stats/gbstats/bayesian/tests.py
+++ b/packages/stats/gbstats/bayesian/tests.py
@@ -130,23 +130,9 @@ class EffectBayesianABTest(BayesianABTest):
         config: EffectBayesianConfig = EffectBayesianConfig(),
     ):
         super().__init__(stat_a, stat_b, config)
-        # rescale prior if needed
-        if self.relative and config.prior_type == "absolute":
-            self.prior_effect = GaussianPrior(
-                config.prior_effect.mean / abs(self.stat_a.unadjusted_mean),
-                config.prior_effect.variance / pow(self.stat_a.unadjusted_mean, 2),
-                config.prior_effect.proper,
-            )
-        elif not self.relative and config.prior_type == "relative":
-            self.prior_effect = GaussianPrior(
-                config.prior_effect.mean * abs(self.stat_a.unadjusted_mean),
-                config.prior_effect.variance * pow(self.stat_a.unadjusted_mean, 2),
-                config.prior_effect.proper,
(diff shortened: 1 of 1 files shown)


bryce-fitzsimons (approved):

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -158,6 +144,23 @@ def compute_result(self):
...
+        # rescale prior if needed
+        scaled_prior_effect = self.config.prior_effect
+        if self.relative and self.config == "absolute":
+            scaled_prior_effect = GaussianPrior(
+                self.config.prior_effect.mean / abs(self.stat_a.unadjusted_mean),
+                self.config.prior_effect.variance / pow(self.stat_a.unadjusted_mean, 2),
+                self.config.prior_effect.proper,
+            )
+        elif not self.relative and self.config.prior_type == "relative":
+            if self.config.prior_effect.proper and self.stat_a.unadjusted_mean <= 0:
```
lukebrawleysmith, line 156: couldn't we scale by absolute value here if unadjusted_mean < 0?  We take absolute value when scaling below for the positive case.  So we should either permit unajudsted_mean < 0, or remove the abs() call below.

packages/stats/gbstats/bayesian/tests.py
```diff
@@ -174,17 +177,16 @@ def compute_result(self):
             self.stat_a.unadjusted_mean,
         )
 
-        post_prec = (
-            1 / data_variance
-            + int(self.prior_effect.proper) / self.prior_effect.variance
+        post_prec = 1 / data_variance + (
+            0 if not scaled_prior_effect.proper else 1 / scaled_prior_effect.variance
```
lukebrawleysmith, line 181: nit: would be easier to read as 
```1 / scaled_prior_effect.variance if scaled_prior_effect.proper else 0```

lukebrawleysmith (approved):
2 minor comments

packages/stats/gbstats/bayesian/tests.py
lukesonnet: yes, much better. I was being hasty. thanks

packages/stats/gbstats/bayesian/tests.py
lukesonnet: Sure, we just have other problems with negative metrics but I think it should be fine here.

PR #2911, opened 2024-08-21
Make it possible to duplicate/archive fact metrics & archive fact tables

msamper:
### Features and Changes

Allows users to duplicate/archive fact metrics and archive fact tables. For duplicating fact metrics we're only allowing users to do it within its parent fact table and not in the overall metrics list. I felt like it would be a weird UX to allow duplication from that overall list.

### Testing

1. Create a few fact tables with metrics
2. Archive some of the fact tables and make sure their associated fact metrics don't show up as potential metrics for experiments.
3. Make sure the show archived toggles work as expected
4. Try archiving metrics within a fact table
5. Archive and unarchive a fact table with some archived metrics and make sure their archived state remains the same

### To-Do

- [x] Show which metrics have been archived within a Fact Table page

### DIFF
--- a/packages/front-end/components/FactTables/FactMetricList.tsx
+++ b/packages/front-end/components/FactTables/FactMetricList.tsx
@@ -2,7 +2,7 @@ import {
   FactMetricInterface,
   FactTableInterface,
 } from "back-end/types/fact-table";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import Link from "next/link";
 import { date } from "shared/dates";
 import MoreMenu from "@/components/Dropdown/MoreMenu";
@@ -36,13 +36,7 @@ export function getMetricsForFactTable(
 
 export default function FactMetricList({ factTable }: Props) {
   const [newOpen, setNewOpen] = useState(false);
-  const [showArchived, setShowArchived] = useState(
-    factTable.archived ? true : false
-  );
-
-  useEffect(() => {
-    setShowArchived(factTable.archived ? true : false);
-  }, [factTable]);
+  const [showArchived, setShowArchived] = useState(false);
 
   const { apiCall } = useAuth();
   const [isDeleting, setIsDeleting] = useState(false);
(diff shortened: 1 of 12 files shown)


packages/front-end/components/FactTables/FactMetricList.tsx
```diff
@@ -34,13 +34,15 @@ export function getMetricsForFactTable(
 }
 
 export default function FactMetricList({ factTable }: Props) {
-  const [editOpen, setEditOpen] = useState("");
```
msamper, line 37: This didn't seem like it was being used. Nothing was setting `editOpen` so I removed this state

packages/front-end/components/FactTables/FactMetricList.tsx
```diff
@@ -84,26 +98,37 @@ export default function FactMetricList({ factTable }: Props) {
...
-      {editOpen && (
+      {duplicateMetric && (
         <FactMetricModal
-          close={() => setEditOpen("")}
-          initialFactTable={factTable.id}
-          existing={metrics.find((m) => m.id === editOpen)}
+          close={() => setDuplicateMetric(undefined)}
+          existing={duplicateMetric}
+          duplicate
           source="fact-table"
```
jdorn, line 106: Change the source to `fact-table-duplicate`

packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -75,6 +76,7 @@ const defaultValue: DefinitionContextValue = {
   projects: [],
   factTables: [],
   factMetrics: [],
+  _factMetricsIncludingArchived: [],
```
jdorn, line 79: We should do a similar thing for fact tables.  There are a few places in the UI where we ask you to select a fact table.  If it's archived, it should not appear in the select box.  Easiest is to do what we do with metrics where `definitions.factTables` excludes them by default.

jdorn (approved):

PR #2921, opened 2024-08-22
Update JS SDK to be compatible with the JSR

royalfig:
### Features and Changes
The [JSR (JavaScript Registry)](https://jsr.io/) is an open-source package registry (it builds on npm). In order to publish our JS SDK, we need to make a few updates to our codebase:

- Define explicit return types
- Remove or refactor global type declarations

The reason is that JSR allows you to publish TS directly to the registry. They then handle building compatible versions for users. For this and features, JSR analyzes source code, and it turns out that certain[ types are slow](https://jsr.io/docs/about-slow-types). 

These changes enable us to publish to the registry and, in theory, should improve type analysis.

### Dependencies

None

### Testing
Passes all tests in the SDK test suite.

### DIFF
--- a/packages/sdk-js/jsr.json
+++ b/packages/sdk-js/jsr.json
@@ -1,0 +1,5 @@
+{
+  "name": "@growthbook/growthbook",
+  "version": "0.0.1",
+  "exports": "./src/index.ts"
+}
--- a/packages/sdk-js/jsr.json
+++ b/packages/sdk-js/jsr.json
@@ -1,5 +1,8 @@
 {
   "name": "@growthbook/growthbook",
   "version": "0.0.1",
-  "exports": "./src/index.ts"
+  "exports": "./src/index.ts",
+  "publish": {
+    "include": ["README.md", "CHANGELOG.md", "src/**/*.ts"]
+  }
 }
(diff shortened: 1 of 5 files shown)


.github/workflows/jsr.yml
```diff
@@ -0,0 +1,31 @@
+name: JSR
+
+on:
+  pull_request:
+    paths:
+      - "packages/sdk-js/**"
```
jdorn, line 6: This will publish before we merge to main, which I don't think we want.

royalfig: Good call! 

This should run when there's a push to main && when there are changes to `/sdk-js`. Does that seem right?
```
on:
  push:
    branches:
      - main
    paths:
      - "packages/sdk-js/**"
```

packages/sdk-js/jsr.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@growthbook/growthbook",
+  "version": "0.1.3",
```
jdorn, line 3: Our latest SDK version is `1.2.0`.  Does this need to match here?

royalfig: Do you think they should match? I wasn't sure either way, because the JSR is a different release process and non-compiled code. But I could see a case for matching, too

PR #2923, opened 2024-08-23
Add experiment.info.significance event.

romain-growthbook:
### Features and Changes


packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -170,16 +192,206 @@ export const notifySrm = async ({
...
+export const notifySignificance = async ({
+  context,
+  experiment,
+  snapshot: currentSnapshot,
+}: {
+  context: Context;
+  experiment: ExperimentInterface;
+  snapshot: ExperimentSnapshotDocument;
+}) => {
+  const lastSnapshot = await getLatestSnapshot(
```
romain-growthbook, line 270: Makse sure that this is executed _before_ saving a new snaopshot!

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -170,16 +192,206 @@ export const notifySrm = async ({
...
+    experiment.id,
+    experiment.phases.length - 1
+  );
+
+  if (!lastSnapshot) return;
```
romain-growthbook, line 275: Cases:
1. Something broken
2. First snapshot for the experiment.

Test case: make sure we fire on the first snapshot if it is significant on the first run.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: (instead we now use dateCreated of the snapshots)

packages/back-end/src/events/handlers/slack/slack-event-handler-utils.ts
```diff
@@ -371,6 +377,56 @@ const buildSlackMessageForExperimentDeletedEvent = (
...
+    metricName,
+    variationName,
+    experimentName,
+  }: {
+    metricName: string;
+    variationName: string;
+    experimentName: string;
+  }) => {
+    if (statsEngine === "frequentist") {
+      return `In experiment ${experimentName}: metric ${metricName} for variation ${variationName} is
```
lukesonnet, line 401: Make experimentName here and in the "bayesian" case below a link to the experiment with the #results anchor.

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -95,11 +103,18 @@ export const notifyAutoUpdate = ({
...
-        success,
-        experimentId: experiment.id,
-        experimentName: experiment.name,
+      dispatchEvent({
+        context,
+        experiment,
+        event: "warning",
+        data: {
+          object: {
+            type: "auto-update",
```
lukesonnet, line 112: auto-update isn't really a "warning" event type. Is this necessary to classify it as such?

lukesonnet (approved):
tested in local for email and slack, and unit tests look correct logic wise.

furthermore took a closer look at the more important parts (e.g. `getLatestSnapshot` changes) and it lgtm

packages/back-end/src/events/handlers/slack/slack-event-handler-utils.ts
```diff
@@ -288,6 +294,11 @@ const buildSlackMessageForFeatureDeletedEvent = async (
 export const getExperimentUrlFormatted = (experimentId: string): string =>
   `\n• <${APP_ORIGIN}/experiment/${experimentId}|View Experiment>`;
 
+export const getExperimentUrlAndNameFormatted = (
+  experimentId: string,
+  experimentName: string
+): string => `<${APP_ORIGIN}/experiment/${experimentId}|${experimentName}>`;
```
romain-growthbook, line 300: `APP_ORIGIN` was missing indeed! Also didn't realized this was not a `markdown` formatted string..

PR #2927, opened 2024-08-23
Allow for API creation of deprecated experiment rules

### LINKED ISSUE #2925: [Feature] Support `experiment` type rules in API PostFeaturePayload
### Description of Feature
Currently the `PostFeaturePayload` schema only supports `experiment_ref` type rules, not `experiment`. This makes it impossible to fetch a feature definition and then pass it back into a post body if it has an `experiment` rule configured. This should be added as a (deprecated) option to more easily tie a new feature to the existing experiment (likely with some checks that the configuration provided for the experiment matches the current state if the id is in use).

Kevin-Chant:
### Features and Changes
Adds the deprecated `experiment` rule type to the set of valid types when posting a new feature to the API. This is important as API access is the only real way to clone feature definitions across GB environments, and currently `experiment` rules will throw a validation error.
It's also a non-starter to create an experiment and use the proper `experiment-ref` rule type, as the existing feature rules don't have enough information to fully describe a valid experiment.

- Closes #2925 

### Testing

Given an example feature pulled from the getFeature api endpoint, use curl to post it and verify it's created successfully.

Curl script:
```
#!/bin/bash

curl --header "Content-Type: application/json" \
  --header "Authorization: Bearer <YOUR_API_TOKEN_HERE>" \
  --request POST \
  --data '{
"name": "name",
                    "description": "",
                    "hashAttribute": "id",
                    "trackingKey": "standalone-pii-form",
                    "variations": [
                        {
                            "key": "variation-lenglf0k",
                            "name": "",
                            "value": "standalone-page",
                            "weight": 0.5,
                        },
                        {
                            "key": "variation-lenglf0l",
                            "name": "",
                            "value": "modal",
                            "weight": 0.5
                        }
                    ]
}' \
  http://localhost:3100/api/v1/experiments
```


packages/back-end/src/api/openapi/payload-schemas/postFeature/FeatureExperimentRule.yaml
Kevin-Chant, line 1: Clone of the existing `FeatureExperimentRule` other than the `values` key being fixed to be plural (to match the model) where it's still `value` in the original. This removed the need to modify a number of unrelated files so it seemed like a worthwhile change

jdorn: Question: does this mean the output of GET feature cannot be passed into POST feature without a translation step to rename this field? If so, I think we should support both keys and just mark `value` as deprecated. Should just be a few lines changed in the controller to handle this.

Kevin-Chant: No, that change is necessary to make the output of GET be possible to pass into POST.
GET is correctly returning the key as `values` (and the validator responsible for typechecking is using `values`). The issue is that the old `FeatureExperimentRule` yaml schema accidentally specified `value` and is stuck that way.

That said, I'll check if it's easy enough to support either one without getting into a ton of typescript nonsense. That definitely seems safest.

Kevin-Chant: Update: this now supports `value` or `values` as the source of truth (prioritizing `values` if both are present). Should be good to go now

jdorn (approved):

PR #2932, opened 2024-08-27
support fallbackAttribute during experiment creation

bryce-fitzsimons:
The `fallbackAttribute` was not getting set on experiment create.

Also fixes a local hook for doc gen.


romain-growthbook (approved):
Thanks!

PR #2937, opened 2024-08-29
Add list of experiment results to metric pages

lukesonnet:
### Features and Changes

Adds a paginated, sortable list of Experiments and their results to Fact Metric and Metric pages.

Uses a new endpoint to retrieve all experiments for a metric (up to 1000 experiments) and attaches their snapshot to it to populate the list. This happens whenever the Experiments tab is opened.

Modifies `useSearch` to allow devs to specify `undefinedLast` which puts undefined values at the bottom of the table.

### Testing

Clicked around in the UI. Tested missing results, etc.

### Screenshots

Fact Metric
Screenshot 2024-08-29 at 8 17 55 AM

Missing data
Screenshot 2024-08-29 at 8 19 06 AM

Metric
Screenshot 2024-08-29 at 8 21 22 AM


packages/back-end/src/controllers/experiments.ts
```diff
@@ -339,7 +339,7 @@ async function _getSnapshot(
   );
 }
 
-async function _getSnapshots(
+export async function _getSnapshots(
```
jdorn, line 342: Move this outside of the controller.  We have some code that expects all exported methods of a controller file to be request handlers with the format `(req, res) => Promise<void>`

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -185,6 +197,22 @@ export default function FactMetricPage() {
...
+    const handler = () => {
+      const hash = window.location.hash.replace(/^#/, "") as MetricTab;
+      if (metricTabs.includes(hash)) {
+        setTab(hash);
+      }
+    };
+    handler();
+    window.addEventListener("hashchange", handler, false);
+    return () => window.removeEventListener("hashchange", handler, false);
+  }, [setTab]);
```
Comment (jdorn) [inline packages/front-end/pages/fact-metrics/[fmid].tsx:214]: We have a `Tabs` component that already includes this logic.  We should use that instead.

packages/front-end/styles/global.scss
```diff
@@ -128,6 +128,13 @@ pre {
   }
 }
 
+#metric-page-tabs {
+  font-size: 1.2em;
+  .active-tab {
+    color: var(--text-color-primary);
+  }
+}
```
jdorn, line 136: I would remove this.  We're going to switch tabs to use Radix soon, and any custom styling just makes that migration more difficult.

jdorn (approved):

PR #2941, opened 2024-09-03
Default protocol ports for proxy uri parsing

joornby-angel:
### Features and Changes

Our company uses the "Yuki" cost optimization tool on our snowflake instance.  I attempted to use the `SNOWFLAKE_PROXY` environment override on our self-hosted instance.  The yuki proxy is not a self hosted authenticated proxy, but instead a fully qualified https access point to our snowflake database.  The nodejs URL constructor drops port definitions when parsing a uri, and the snowflake sdk requires a port when you specify a proxy host.  Therefore if you specify a proxy uri such as `https://abcd-wxyz.yukicomputing.com:443`, node will drop the 443 port definition, and the snowflake sdk will throw an error saying that you must specify the `proxyPort`.  

Ultimately, our company actually needs the results of this [PR](https://github.com/growthbook/growthbook/pull/2890).  I didn't discover it until now, but figured I'd throw this up in case it's needed by someone else.

### Testing

- Included automated testing for the proxy settings parsing logic

### DIFF
--- a/packages/back-end/src/services/snowflake.ts
+++ b/packages/back-end/src/services/snowflake.ts
@@ -10,15 +10,31 @@ type ProxyOptions = {
   proxyPort?: number;
   proxyProtocol?: string;
 };
-function getProxySettings(): ProxyOptions {
-  const uri = process.env.SNOWFLAKE_PROXY;
+
+function proxyPort(url: URL): number | undefined {
+  if (url.port) {
+    return parseInt(url.port);
+  }
+
+  if (url.protocol === "http:") {
+    return 80;
+  }
+
+  if (url.protocol === "https:") {
+    return 443;
+  }
+
+  return undefined;
+}
+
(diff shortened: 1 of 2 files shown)


lukesonnet:
@joornby-angel 

Merging the Access URL PR now. Are you confident that will resolve your issue instead of this approach here? I'll close this if so. Thanks for your investigation into this issue.

joornby-angel:
> @joornby-angel
> 
> Merging the Access URL PR now. Are you confident that will resolve your issue instead of this approach here? I'll close this if so. Thanks for your investigation into this issue.

After I found the other PR, I tested it by hard coding the accessUrl field to ensure that approach would work.  Once that is released, I plan on using it.  I don't care what happens with this PR, but based off of the interplay between the node URL constructor and snowflake's validation, this is still a problem.  I believe the main use case that would be problematic is if someone tries to use an http snowflake proxy on port 80 (the default).

jdorn:

PR #2952, opened 2024-09-06
Added archetype listing page and way to simulate feature values

Auz:
Added new archetype page with options to create and edit archetypes. This page also lets you simulate what feature values users will get based on archetypes or user attributes. 

Screenshot 2024-09-05 at 7 29 31 PM

Screenshot 2024-09-05 at 7 29 47 PM

Screenshot 2024-09-05 at 7 30 06 PM

- [x] Need to explain issues around hash versions


packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1046,7 +1046,7 @@ export class GrowthBook<
           // If it's a conditional rule, skip if the condition doesn't pass
           if (rule.condition && !this._conditionPasses(rule.condition)) {
             process.env.NODE_ENV !== "production" &&
-              this.log("Skip rule because of condition ff", {
+              this.log("Skip rule because of force condition didn't match", {
```
bryce-fitzsimons, line 1049:
```suggestion
              this.log("Skip rule because force condition doesn't pass", {
```

packages/front-end/components/EventWebHooks/utils.tsx
```diff
@@ -49,6 +49,10 @@ export const notificationEventNames = [
   "experiment.deleted",
   "experiment.warning",
   "experiment.info.significance",
+  // metrics
+  "metric.created",
+  "metric.updated",
+  "metric.deleted",
```
bryce-fitzsimons, line 55: remove?

packages/front-end/components/Archetype/SimulateFeatureValues.tsx
```diff
@@ -0,0 +1,516 @@
...
+          }}
+          onSubmit={({ archetype, attributes }) => {
+            setArchetype(archetype);
+            setAttributes(attributes);
+            setEditAttributesModalOpen(false);
+            refreshResults();
+          }}
+        />
+      )}
+      <div className="container-fluid pagecontents pt-4">
```
bryce-fitzsimons, line 470: consistent margins between 2 tabs
```suggestion
      <div className="pagecontents">
```

packages/front-end/components/Archetype/ArchetypeList.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  return (
+    <>
+      <div className="row mb-3">
+        <div className="col">
+          <h1>Archetypes</h1>
+        </div>
+        {canCreateGlobal && (
+          <div className="col-auto">
+            <button
```
bryce-fitzsimons, line 50: use the new Radix button class with no "+" icon for consistency

packages/front-end/components/Archetype/AttributeForm.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
```
bryce-fitzsimons, line 1: line 146 or so:
We should include the <DatePicker> for attribute values that use a date format (i.e. string formatted as "date")

packages/front-end/components/Archetype/SimulateFeatureValues.tsx
```diff
@@ -0,0 +1,516 @@
...
+  if (showAllEnv) {
+    environmentOptions.unshift({ label: "All", value: "all" });
+  }
+
+  const featureTableResults = (
+    <>
+      <div className="mb-3">
+        <div className="row mb-3">
+          <div className="col">
+            <div
```
bryce-fitzsimons, line 209: something feels a bit clunky about this component:

1. No clear CTA or button... entire div is clickable but doesn't give the impression that it is
2. Attribute values are listed in plain text on the same line, which feels hard to parse / cluttered in many cases. I think either format it in some sort of table inline or provide a tooltip or accordion to conditionally show the formatted version

packages/front-end/components/Archetype/SimulateFeatureValues.tsx
```diff
@@ -0,0 +1,516 @@
...
+          onClick={(e) => {
+            e.preventDefault();
+            setOpenWarning(!openWarning);
+          }}
+        >
+          <div className="p-2 pr-3">
+            <FiAlertTriangle />
+          </div>
+          <div>
+            These results use the JS SDK, which supports the V2 hashing
```
bryce-fitzsimons, line 429: Do we need this text? It's confusing to me.
- Why is this a warning? Do I need to change something?
- What are the older/unsupported SDKs? Am I using one?
- What experiment? These are features... (presumedly linked experiments, but its not obvious)

And I don't think we're actually answering the important question here:
- What hash version was used to evaluate these features? Did we force v1 or v2? Or did we use the per experiment setting?

packages/front-end/components/Archetype/ArchetypeList.tsx
```diff
@@ -0,0 +1,212 @@
...
+              className="btn btn-primary float-right"
+              onClick={() => {
+                setEditArchetype({});
+                track("Viewed Add Archetype Modal", {
```
bryce-fitzsimons, line 54: all modals auto-track now, this is probably not needed

packages/back-end/src/controllers/features.ts
```diff
@@ -1855,18 +1856,53 @@ export async function postFeatureEvaluate(
...
+}) {
+  const feature = await getFeature(context, featureId);
+  const { org } = context;
   if (!feature) {
     throw new Error("Could not find feature");
   }
-
+  if (feature.organization !== org.id) {
+    throw new Error("Feature does not belong to this organization");
+  }
```
bryce-fitzsimons, line 1902: is this possible? getFeature() is scoped to the org id in the context obj

packages/back-end/src/controllers/features.ts
```diff
@@ -1886,10 +1922,58 @@ export async function postFeatureEvaluate(
...
+}
+
+export async function postFeaturesEvaluate(
+  req: AuthRequest<
+    {
+      attributes: Record<string, boolean | string | number | object>;
+      featureIds: string[];
+      environment: string;
+    },
+    { version: string }
```
bryce-fitzsimons, line 1934: what does this do?

packages/front-end/components/Archetype/ArchetypeList.tsx
Auz: Done, though I think the new button is way too big

packages/front-end/components/Archetype/ArchetypeList.tsx
Auz: They don't track entirely without modifying the modal file, but I made that change.

packages/front-end/components/Archetype/SimulateFeatureValues.tsx
Auz: done

packages/front-end/components/Archetype/AttributeForm.tsx
Auz: done

packages/front-end/components/Archetype/SimulateFeatureValues.tsx
Auz: The hash can be set per experiment/rollout. We don't force it one way or the other, but it can be v2, and the SDK they use might not support it.

packages/front-end/components/Archetype/SimulateFeatureValues.tsx
Auz: I adjusted the design to be more consistent with other sections.

Auz:
Screenshot 2024-11-26 at 2 28 49 PM

bryce-fitzsimons (approved):

packages/back-end/src/services/features.ts
```diff
@@ -979,6 +980,114 @@ export function evaluateFeature({
...
+    };
+    switchEnv = true;
+  }
+
+  // get all features definitions
+  const experimentMap = await getAllPayloadExperiments(context);
+  // I could loop through the feature's defined environments, but if environments change in the org,
+  // the values in the feature will be wrong.
+  if (!environments || environments.length === 0) {
+    return;
```
tzjames, line 1024: This  is an early returns in this function that would execute before the code to change it back to production. So all future requests would be stuck in development.

tzjames (changes requested):

packages/back-end/src/services/features.ts
```diff
@@ -979,6 +980,114 @@ export function evaluateFeature({
...
+  }
+  // change the NODE ENV so that we can get the debug log information:
+  let switchEnv = false;
+  if (process.env.NODE_ENV === "production") {
+    process.env = {
+      ...process.env,
+      NODE_ENV: "development",
+    };
+    switchEnv = true;
+  }
```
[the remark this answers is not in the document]
Auz: I added additional checks to make sure it's set back.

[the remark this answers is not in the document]
tzjames: Ok thats better, but it is still theoretically unsafe if we have other prod checks in the future (from what I can find all the prod checks we currently have are in secrets or other file that probably get loaded when the application boots). More robust ways of doing this would be to:

- change the sdk to have an envOverride, or debugInProd property
- use `worker_threads` npm package. I'll DM you how that would work.

packages/front-end/components/Archetype/AttributeForm.tsx
```diff
@@ -1,23 +1,34 @@
...
 
+const deepEqual = (x, y) => {
+  const ok = Object.keys,
+    tx = typeof x,
+    ty = typeof y;
+  return x && y && tx === "object" && tx === ty
+    ? ok(x).length === ok(y).length &&
+        ok(x).every((key) => deepEqual(x[key], y[key]))
+    : x === y;
+};
```
tzjames, line 30: I believe you can just use `import { isEqual } from 'lodash'` instead of this.

tzjames (approved):
Ideally how we change env thing should be fixed so it doesn't bite us in the future.  But I think it is safe to land for now, and I realize we are not in an ideal world.

PR #2953, opened 2024-09-06
Allow Saved Groups to be tied to projects

msamper:
### Features and Changes

Allows Saved Groups to be project scoped. Introduces new UI to add a project to Saved Groups and filters by project depending on the users current project.

### Screenshots

Screenshot 2024-09-13 at 11 10 05 AM
Screenshot 2024-09-13 at 11 10 26 AM


packages/back-end/src/routers/saved-group/saved-group.controller.ts
```diff
@@ -480,6 +495,8 @@ export const putSavedGroup = async (
     fieldsToUpdate.description = description;
   }
 
+  fieldsToUpdate.projects = projects;
```
msamper, line 498: Do we want to check here that the user has permission to create within all the projects in `fieldsToUpdate`?

packages/back-end/src/routers/saved-group/saved-group.controller.ts
```diff
@@ -53,7 +53,7 @@ export const getSavedGroups = async (
   const context = getContextFromReq(req);
   const { org } = context;
 
-  if (!context.permissions.canCreateSavedGroup()) {
+  if (!context.permissions.canCreateSavedGroup({ projects: [] })) {
     context.permissions.throwPermissionError();
   }
```
msamper, line 58: I can't find any use of this in our code. Seems like we just get all saved groups through `/organization/definitions`. Do we want to get rid of this unused endpoint?

packages/back-end/src/routers/saved-group/saved-group.controller.ts
```diff
@@ -425,19 +431,28 @@ export const putSavedGroup = async (
...
 
-  if (!context.permissions.canUpdateSavedGroup()) {
+  const savedGroup = await getSavedGroupById(id, org.id);
+
+  if (
+    !context.permissions.canUpdateSavedGroup({ projects: savedGroup?.projects })
+  ) {
     context.permissions.throwPermissionError();
   }
```
msamper, line 454: Move permission check below check to see if saved group exists in all files

packages/front-end/components/SavedGroups/ConditionGroups.tsx
```diff
@@ -150,12 +164,27 @@ export default function ConditionGroups({ groups, mutate }: Props) {
...
+                              resourceType="saved group"
+                              className="badge-ellipsis short align-middle"
+                            />
+                          )}
+                        </td>
                         <td>{s.owner}</td>
                         <td>{ago(s.dateUpdated)}</td>
-                        {canUpdate || canDelete ? (
+                        {canUpdate({ projects: s.projects }) ||
+                        canDelete({ projects: s.projects }) ? (
```
jdorn, line 184: For all of these permission methods, just pass in the saved group interface in its entirety instead of creating an object with just `projects`.

For example, do this:

```ts
canUpdate(s) || canDelete(s) ? ...
```

packages/front-end/components/SavedGroups/ConditionGroups.tsx
```diff
@@ -198,7 +227,12 @@ export default function ConditionGroups({ groups, mutate }: Props) {
                   {!items.length && isFiltered && (
                     <tr>
                       <td
-                        colSpan={canUpdate || canDelete ? 6 : 5}
+                        colSpan={
+                          canUpdate({ projects: [project] }) ||
+                          canDelete({ projects: [project] })
+                            ? 6
+                            : 5
```
jdorn, line 234: Are these column counts accurate still?  It looks like you added 1 additional column to the table, so will this have to go to 6/7 instead of 5/6?

Kevin-Chant (changes requested):
I did a super scuffed test for the `forEach` behavior but this implies to me that it's a real issue. I read through the rest of the PR and it looks good to me. Want to get a fix in for this then I'll try to break it one last time?

image

packages/back-end/src/api/saved-groups/postSavedGroup.ts
```diff
@@ -9,11 +9,24 @@ import { postSavedGroupValidator } from "../../validators/openapi";
...
+      // Ensure the project exists
+      const project = await req.context.models.projects.getById(projectId);
+      if (!project) {
+        throw new Error("Project does not exist");
+      }
+      // Ensure project is a part of the organization
+      if (project.organization !== req.organization.id) {
+        throw new Error("Project does not belong to this organization");
+      }
+    });
```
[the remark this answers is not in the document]
msamper: Good catch! You're definitely right. Switched to using a `map` and then wrapping that in a `Promise.all()`

packages/back-end/src/routers/saved-group/saved-group.controller.ts
```diff
@@ -96,11 +57,29 @@ export const postSavedGroup = async (
...
-  if (!context.permissions.canCreateSavedGroup()) {
+  if (!context.permissions.canCreateSavedGroup({ ...req.body })) {
     context.permissions.throwPermissionError();
   }
+
+  if (projects) {
+    await Promise.all(
+      projects.map(async (projectId) => {
+        // Ensure the project exists
+        const project = await context.models.projects.getById(projectId);
```
msamper, line 71: Hmm should our `getById` contextual methods filter by request org by default? Or they do already do that and then lines 76-78 are unnecessary? From what I can tell quickly looking at the code it looks like they don't

Kevin-Chant: I remember seeing "always filter by org" as one of the important query standards (for security), so if we could have that built in that seems ideal. Maybe Jeremy can weigh in on the best way to make that happen?

packages/front-end/components/SavedGroups/SavedGroupForm.tsx
```diff
@@ -57,9 +60,15 @@ const SavedGroupForm: FC<{
       type,
       values: current.values || [],
       description: current.description || "",
+      projects: current.projects || project ? [project] : [],
```
Kevin-Chant, line 63: Something about this isn't working and I haven't identified why yet. I added a project to a saved group then went to edit it and the current group isn't already in the multiselect. Hopefully these debug logs are helpful:

```
  console.log(
    "Rendering savedGroupForm. Current projects are",
    current.projects
  );
  console.log("Form projects are", form.watch("projects"));

```

image

Kevin-Chant (changes requested):

packages/front-end/components/SavedGroups/SavedGroupForm.tsx
Kevin-Chant: Oh yeah, found it: you just need parenthesis on the logic like this: `current.projects || (project ? [project] : [])`

Kevin-Chant, line 63:
```suggestion
      projects: current.projects || (project ? [project] : []),
```

Kevin-Chant (changes requested):

Kevin-Chant (commented):
That issue with the `SavedGroupForm` is the only bug I can find. The new promise mapping works correctly & I'm getting the right errors when hitting the api with garbage project ids

packages/back-end/src/routers/saved-group/saved-group.controller.ts
msamper: cc: @jdorn

jdorn: All of the `context.models` automatically filter by org id, so just doing `getById(projectId)` is fine.   It will actually do a query for `{organization: context.org.id, id: projectId}` under the hood

Kevin-Chant (approved):
LGTM. I'd say just merge https://github.com/growthbook/growthbook/pull/3000 into this as long as you're happy with the styling, and then this should be good to go (as long as Jeremy's comments around the db access are all addressed)

msamper:
Just realized, but we probably want to update `deleteProject` to include removing or deleting savedGroups associated with the deleted project right?

Kevin-Chant:
> Just realized, but we probably want to update `deleteProject` to include removing or deleting savedGroups associated with the deleted project right?

Nice catch. We at least need to remove the project from the list for all associated saved groups.@jdorn what's the expected behavior for removing the only project a group is associated with? Does it become an "all projects" group?

msamper:
> > Just realized, but we probably want to update `deleteProject` to include removing or deleting savedGroups associated with the deleted project right?
> 
> Nice catch. We at least need to remove the project from the list for all associated saved groups.@jdorn what's the expected behavior for removing the only project a group is associated with? Does it become an "all projects" group?

So I just did a test and it looks like for experiments we just remove the project from the experiment and it becomes an `all projects` experiment and then for sdk connections we break it and its tied to a non existent project. I feel like maybe we'd want to ask them what behavior they want? If a saved group was in a project for security reasons, it could be bad to suddenly expose it to all projects

jdorn:
Looks like elsewhere we just simply remove the project from the array.  So it could go from `["prj_being_removed"]` to `[]` and suddenly be available to "all projects", but I think that's ok for now.

This is what we do for data sources for example:

```ts
export async function removeProjectFromDatasources(
  project: string,
  organization: string
) {
  await DataSourceModel.updateMany(
    { organization, projects: project },
    { $pull: { projects: project } }
  );
}
```

packages/back-end/src/api/saved-groups/postSavedGroup.ts
```diff
@@ -9,11 +9,24 @@ import { postSavedGroupValidator } from "../../validators/openapi";
...
+    if (projects) {
+      await Promise.all(
+        projects.map(async (projectId) => {
+          // Ensure the project exists
+          const project = await req.context.models.projects.getById(projectId);
+          if (!project) {
+            throw new Error("Project does not exist");
+          }
+        })
+      );
```
jdorn, line 27: Avoid doing Mongo queries inside of loops since it can be very expensive.  A better approach is to use `getByIds` and pass in an array of ids.

```ts
if (projects && projects.length) {
  const projectObjects = await req.context.models.projects.getByIds(projects);
  if (projectObjects.length !== projects.length) {
    throw new Error("Invalid project ids");
  }
}
```

You can improve the above by having the error message list the specific project ids that are invalid.

And since this validation logic is generally useful (already used in a few places in this PR), it might make sense to add it to ProjectModel directly, so you could just do something like this:

```ts
await req.context.models.projects.ensureProjectsExist(projects);
```

packages/front-end/components/SavedGroups/ConditionGroups.tsx
```diff
@@ -133,9 +145,11 @@ export default function ConditionGroups({ groups, mutate }: Props) {
                     <SortableTH field="groupName">Name</SortableTH>
                     <SortableTH field="condition">Condition</SortableTH>
                     <th>Description</th>
+                    <th className="col-2">Projects</th>
                     <SortableTH field="owner">Owner</SortableTH>
                     <SortableTH field="dateUpdated">Date Updated</SortableTH>
-                    {(canUpdate || canDelete) && <th />}
+                    {(canUpdate({ projects: [project] }) ||
+                      canDelete({ projects: [project] })) && <th />}
```
jdorn, line 152: Conditionally rendering the `<th>` is going to cause some weird edge cases here.  If no project is selected in the left nav, but there is a saved group with projects set to `["prj_a"]`, it's possible for this `<th>` to not be rendered, but the associated `<td>` below to be rendered.

To fix, always render the `<th/>`/`<td>` no matter what and just rely on the conditional rendering inside of `<MoreMenu>`.  That should already properly handle things (if no children of MoreMenu are rendered, it should not output anything to the screen).

packages/front-end/components/SavedGroups/IdLists.tsx
```diff
@@ -166,11 +182,13 @@ export default function IdLists({ groups, mutate }: Props) {
...
                       <SortableTH field="attributeKey">Attribute</SortableTH>
                       <th>Description</th>
+                      <th className="col-2">Projects</th>
                       <SortableTH field={"owner"}>Owner</SortableTH>
                       <SortableTH field={"dateUpdated"}>
                         Date Updated
                       </SortableTH>
-                      {(canUpdate || canDelete) && <th></th>}
+                      {(canUpdate({ projects: [project] }) ||
+                        canDelete({ projects: [project] })) && <th></th>}
```
jdorn, line 191: Same here as on the condition list

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -79,6 +79,17 @@ export class ProjectModel extends BaseClass {
     return super.updateById(id, { settings });
   }
 
+  public async ensureProjectsExist(projectIds: string[]) {
+    const projects = await super.getByIds(projectIds);
```
jdorn, line 83: Use `this`, not `super`.

jdorn (approved):

samjarman:
Potentially silly question, but does the sidebar need a redesign at this point? Perhaps putting saved groups, attributes together under flags? I feel the UX is not quite working for these features which are both global and project-owned. Perhaps They need to be project owned only, or have a global project that applies to all projects?

msamper:
> Potentially silly question, but does the sidebar need a redesign at this point? Perhaps putting saved groups, attributes together under flags? I feel the UX is not quite working for these features which are both global and project-owned. Perhaps They need to be project owned only, or have a global project that applies to all projects?

Hey Sam! This is definitely on our radar. I shared your feedback with our designer for when we start rethinking how the sidebar is organized.

PR #2957, opened 2024-09-06
Add support for serving robots.txt in packages/back-end

### LINKED ISSUE #2865: Enable robots.txt Support for API Subdomains
We have observed that our Features API hosted on a subdomain is being indexed by Google. To prevent search engines from crawling and indexing this subdomain, we need to serve a robots.txt file that disallows bots.  We propose adding support for serving the robots.txt file from the GrowthBook backend.

One potential solution is to amend the api.routers.ts file within the packages/backend/api directory to include a new route specifically for robots.txt.

Kevin-Chant:
### Features and Changes

Adds a new route in the base Express app for `/robots.txt`. Reads an env var for where the file is stored, and falls back to a reasonable default (within `packages/back-end`).
I didn't create a default robots.txt with this PR, so it will be up to customers to configure their own if desired.

- Closes #2865

### Testing

- [x] Visit https://localhost:3100/robots.txt without creating a `robots.txt` file
- [x] Create the file in `packages/back-end` and reload the page
- [x] Move the file to another directory and specify an environment variable pointing to it, then reboot the server and reload


packages/back-end/src/app.ts
```diff
@@ -150,6 +152,26 @@ app.get("/favicon.ico", (req, res) => {
   res.status(404).send("");
 });
 
+let robotsTxt = "";
+app.get("/robots.txt", (_req, res) => {
+  if (!robotsTxt) {
+    const file =
+      process.env.ROBOTS_TXT_PATH || path.join(__dirname, "..", "robots.txt");
```
gazzdingo, line 159: do we want a default  .robots.txt in out repo?

Kevin-Chant: Maybe? I didn't feel like I knew the tradeoffs there well enough so I opted not to create it, just surface the route so customers can add it if they want to.

romain-growthbook: I like it this way. Meaningful default with and overridable option. Would be nice to document this somwehere as well.

Kevin-Chant: Yeah this definitely deserves docs. Was gonna toss that to Ryan after I merge this cause I still don't have a great mental map of what docs exist already/where to add it

romain-growthbook: They're in `/docs` in the repo! It would be nice to add it to the PR so that it self-contained.

romain-growthbook, line 159: Are you sure that this'll work in production? Files get transpiled and etc it's not clear if this file and the robot.txt file will be placed in the same relative location.

Since the file content is expected to be pretty small, I'd suggest to make the whole file content an env variable. This will also be much easier to override.

gazzdingo (approved):

romain-growthbook (commented):
One comment on file content in prod. I think doc should be added in this PR it shouldn't be too hard.

packages/back-end/src/app.ts
Kevin-Chant: Hmm, that's a good question. I'd assumed that in prod you'd end up with this code in the `dist` folder the way you do when you build it locally, so stepping out one directory would be good.

I disagree about putting the file contents in an env var personally, as I think the companies that would want to do this setup are more likely to want to version-track their files (incl. robots.txt) & env vars are relatively hard to discover depending on your deployment method.

romain-growthbook: Cool. I don't think env var content should be blocking but making sure that the file is located at the right place should! Also, how would this work in practice when overriding the file location in production for users?

Kevin-Chant: So in prod I would assume that a self-hosted org has some build/deploy process to provision their e.g. EC2 instance, and as part of that they'd add a step to put a `robots.txt` somewhere on the machine. Then if they set the `ROBOTS_TXT_PATH` env variable, either globally or as part of the command they use to start their instance, it'll pick that content up the first time the `/robots.txt` url is requested. Does that answer your question?

romain-growthbook: Yeah. For a docker image you would have to do the same but in the `Dockefile` file. 

Not sure how this would allow version tracking but I imagine this would be done on the user side of things?

Definitely something worth documenting!

One more thing: you could also allow both file location AND file content in env variable. As you explained above, file location can be a little more involved and complicated to deploy in production so have a fallback to content env var would be user friendly for those who do not care about tracking the file.

romain-growthbook: Also, using env variable with a default on our end would alleviate any production build concerns.

Kevin-Chant: Makes sense, but this is definitely getting out of my wheelhouse, and I don't want to let the perfect be the enemy of the good. I'm gonna merge this as-is now that I added the docs reference & double checked the default path in prod, but I do agree that there's improvements that can be made in future to make this more flexible.

I just want to keep this a lightweight fix for a customer request rather than making it a full-on project/feature that takes longer to go live.

packages/back-end/src/app.ts
```diff
@@ -150,6 +152,26 @@ app.get("/favicon.ico", (req, res) => {
...
+    } else {
+      res.status(404).json({
+        message: "Not found",
+      });
+      return;
+    }
+  }
+
+  res.setHeader("Cache-Control", "max-age=3600");
+  res.setHeader("Content-Type", "text/yaml");
```
fabianmarinog, line 171: @Kevin-Chant perhaps this should be `text/plain`?

Kevin-Chant: Good catch. That's what I get for copying existing code wholesale :sweat_smile:

PR #2958, opened 2024-09-06
Create CountrySelector and add isoCountryCode attribute type

Kevin-Chant:
### Features and Changes

Adds a new frontend component - `<CountrySelector>` which wraps the existing `<SelectField>` and `<MultiSelectField>` with a definition of the country codes to display & an option for whether to display the flags for each country. Uses the 2-character country code for the value but displays the full name in the dropdown. Type-ahead search works based on both the label & value already, so e.g. "Algeria" with country code "DZ" can be found by searching either "Alg" or "D".

The CountrySelector is used in attribute targeting for the new datatype of `isoCountryCode`. For list operations (`$in`/`$nin`), it renders the multiselect variant, otherwise it's a single dropdown (with the same typeahead behavior). The new datatype is treated as an enum, and collapsed to a string as its base datatype.

### Testing

First, create a new attribute to use the country codes
- [x] Creating a condition saved group renders working selectors for
  - [x] List-type operations
  - [x] Single value operations
- [x] A new feature targeting rule against that attribute renders working selectors for
  - [x] List-type operations
  - [x] Single value operations
  - [x] Saved group operations
- [x]  Existing SDK connections produce valid JSON with the 2-character strings for each country code
- [x] An existing JS SDK (v1.1.0) can pass in the country code attribute (as a plain string) without needing to be updated, and the user is included in the previously created feature targeting rule

### Screenshots

country_code_filtering

image

image

### DIFF
--- a/packages/back-end/src/util/organization.util.ts
+++ b/packages/back-end/src/util/organization.util.ts
@@ -307,4 +307,5 @@ export const attributeDataTypes = [
   "string[]",
   "number[]",
   "secureString[]",
+  "isoCountryCode",
 ] as const;
--- a/packages/front-end/components/Features/AttributeModal.tsx
+++ b/packages/front-end/components/Features/AttributeModal.tsx
@@ -185,6 +185,7 @@ export default function AttributeModal({ close, attribute }: Props) {
           { value: "string", label: "String" },
           { value: "enum", label: "Enum" },
           { value: "secureString", label: "Secure String" },
+          { value: "isoCountryCode", label: "ISO Country Code" },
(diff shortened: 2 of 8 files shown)


romain-growthbook (approved):
This looks great!

msamper:
This looks great! Thank you for working on this. Caught this a bit late, but we may want the option for ISO Country Code to be under `String Format` rather than `Data Type` within the create attribute modal if this is still a string data type under the hood. That's the pattern we follow for Data and version strings currently. (Similarly to this, Date format string use a different input vs regular strings)

PR #2966, opened 2024-09-10
Make trackingKey, datastore, and assignmentQuery optional when creating experiment in API

tzjames:
### Features and Changes
- Made trackingKey, datastore, assignmentQueryId optional, using default values if they are not set.
- Got rid of code duplication with getting the default datastore and assignment query id by moving it to shared package.
- Added auditing of creating the experiment via the api.
- Added tests for postExperiment endpoint to test most of the existing and all the new logic.
- Added @types/superagent to get rid of typescript warnings in some api tests

### Testing

Create a new experiment in the UI.
Create a new experiment via the API.
`yarn test`


tzjames:
We have suffered from the circular import issue before and @jdorn was able to find the culprit.  I'm pretty sure the errors are coming from me adding the datasource.d.ts file in shared package... but the root cause of that I am unsure about.

lukesonnet:
In our sync with Kari about the Exp creation flow in general, we tentatively aligned on making these optional fields when creating an experiment, but we likely will add other FE/steps to help resolve this case when these fields aren't filled. Maybe it's fine to make it optional for now, and we can separately try to add better error handling when we fall back to some system defaults.

PR #2972, opened 2024-09-11
Swap isoCountryCode to string format not attribute type

Kevin-Chant:
### Features and Changes

We realized after shipping #2958 it probably makes more sense for isoCountryCode to be a string format rather than a type of attribute on its own. This PR makes that adjustment as a fast-follow to the original launch.

Note: any customers who have already (in the past ~2 hours) created a country code attribute will likely need to delete their attribute and start over. This is likely to be a small enough population that it's not worth adding extra code to automatically make the conversion

### Testing
Coming soon

### Screenshots

Coming soon


msamper (approved):
Tested and looks good. Thank you!

PR #2973, opened 2024-09-11
Add attributes API

romain-growthbook:
This PR adds the attributes API. Pretty straight forward with these notes:
* An `id` is added on the fly when it is not present using `uuid`.
* `attributes` are soft-deleted. This seems to be what the `archived` value stands for.


packages/back-end/src/api/openapi/paths/postAttribute.yaml
```diff
@@ -0,0 +1,54 @@
...
+        required:
+          - property
+          - datatype
+        properties:
+          property:
+            type: string
+            description: The attribute property
+          datatype:
+            type: string
+            enum: ["boolean", "string", "number", "secureString", "enum", "string[]", "number[]", "secureString[]"]
```
Kevin-Chant, line 26: I haven't worked with our yaml API schemas much, is there a way to make these enums reference a definition elsewhere? (Or vice-versa, change the other definition to reference this instead of duplicating it?)

packages/back-end/src/api/attributes/listAttributes.ts
```diff
@@ -0,0 +1,45 @@
...
+      const org = req.context.org;
+      const updates: Partial<OrganizationInterface> = {
+        settings: {
+          ...org.settings,
+          attributeSchema: attributes,
+        },
+      };
+
+      await updateOrganization(org.id, updates);
+    }
```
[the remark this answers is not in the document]
Kevin-Chant: Also question, is this something we could do with a JIT migration instead of a write here? We'd still need to write back to mongo in the migration, but that seems a bit preferable to only having it on this list api endpoint

Kevin-Chant (commented):
Looking good, one minor question and I'm gonna play with it a bit before I approve. Test cases are probably rigorous enough to cover the important stuff.

Kevin-Chant (approved):
LGTM. My comments are non-blocking but it'd be great if you could weigh in before merging. In particular I want to prevent an issue where someone changes the definition of attributes elsewhere in the code and forgets to update the API spec.

packages/back-end/src/api/openapi/paths/postAttribute.yaml
romain-growthbook: I don't think so. But the typechecker will fail if this gets out of sync, I've had it happen with `isoCountryCode` below.

PR #2974, opened 2024-09-11
Add heartbeats to queued queries to allow for cleaning them with other stale queries

Kevin-Chant:
### Features and Changes

Moves the heartbeat `setInterval` to apply to all queries, whether running or queued. Then the `getStaleQueries` filter is expanded to included `queued` status as well.
Separately cleans up and refactors a few of the interval/timeout helpers in QueryRunner to make it easier to follow. 

### Testing

1. Set the query concurrency limit of a datasource to 1 to increase the likelihood that queries end up in the queued state
2. Refresh an experiment to trigger multiple queries
3. Watch that they still complete sequentially correctly if the server is undisturbed
4. Refresh again but immediately kill the node server
5. After rebooting the server, observe the queries are still in the queued state
6. Wait 30s-1min and observe that the queued queries are cleaned up for being stale

### Screenshots

After starting queries with concurrency limit of 1, killing and rebooting node:
image

After waiting ~1 minute (without pressing anything in the UI)
image

### DIFF
--- a/packages/back-end/src/queryRunners/QueryRunner.ts
+++ b/packages/back-end/src/queryRunners/QueryRunner.ts
@@ -110,6 +110,7 @@ export abstract class QueryRunner<
   } = {};
   private useCache: boolean;
   private queuedQueryTimers: Record<string, NodeJS.Timeout> = {};
+  private queryHeartbeatIntervals: Record<string, NodeJS.Timeout> = {};
 
   public constructor(
     context: ReqContext | ApiReqContext,
@@ -342,13 +343,11 @@ export abstract class QueryRunner<
           this.onQueryFinish();
         } else {
           if (await this.concurrencyLimitReached()) {
-            this.queuedQueryTimers[query.id] = setTimeout(() => {
-              this.executeQueryWhenReady(
-                query,
-                runCallbacks.run,
-                runCallbacks.process
-              );
-            }, INITIAL_CONCURRENCY_TIMEOUT);
+            this.queueQueryForExecution(
+              query,
+              runCallbacks.run,
+              runCallbacks.process
+            );
           } else {
             await this.executeQuery(
               query,
@@ -439,14 +438,17 @@ export abstract class QueryRunner<
         (q) => q.status === "running" || q.status === "queued"
       )
     ) {
-      const runningIds = this.model.queries
-        .filter((q) => q.status === "running")
+      const queryIds = this.model.queries
+        .filter((q) => q.status === "running" || q.status === "queued")
         .map((q) => q.query);
 
-      if (runningIds.length) {
+      if (queryIds.length) {
+        queryIds.forEach((qid) => {
+          this.clearQueuedQueryTimer(qid);
+        });
         const queryDocs = await getQueriesByIds(
           this.model.organization,
-          runningIds
+          queryIds
         );
 
         const externalIds = queryDocs.map((q) => q.externalId).filter(Boolean);
@@ -479,7 +481,7 @@ export abstract class QueryRunner<
     }
   }
 
-  public async executeQueryWhenReady<
+  public async executeQueryIfReady<
     Rows extends RowsType,
     ProcessedRows extends ProcessedRowsType
   >(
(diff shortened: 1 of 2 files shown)


packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -491,20 +493,22 @@ export abstract class QueryRunner<
...
   ): Promise<void> {
-    // If too many queries are running against the datastore, use capped exponential backoff to wait until they've finished
+    // If too many queries are running against the datastore
+    // use capped exponential backoff to wait until they've finished
     const concurrencyLimitReached = await this.concurrencyLimitReached();
     if (concurrencyLimitReached) {
       logger.debug(
         `${doc.id}: Query concurrency limit reached, waiting ${currentTimeout} before retrying`
       );
-      this.runCallbacks[doc.id] = { run, process };
```
Kevin-Chant, line 500: This doesn't need to be repeated each time the query is attempted so it's been moved to before the query is queued

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -656,29 +652,31 @@ export abstract class QueryRunner<
     logger.debug("Creating query for: " + name);
     const concurrencyLimitReached = await this.concurrencyLimitReached();
     const dependenciesComplete = dependencies.length === 0;
-    const readyToRun =
-      dependenciesComplete && !runAtEnd && !concurrencyLimitReached;
+    const readyToRun = dependenciesComplete && !runAtEnd;
```
Kevin-Chant, line 655: Changed the definition of `readyToRun` to be slightly clearer. A query hitting the concurrency limit is "ready to run"; the reason it's not running is external.

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -656,29 +652,31 @@ export abstract class QueryRunner<
...
       language: this.integration.getSourceProperties().queryLanguage,
-      dependencies: dependencies,
-      running: readyToRun,
-      runAtEnd: runAtEnd,
+      running: readyToRun && !concurrencyLimitReached,
     });
 
+    // Update heartbeat for the query once every 30 seconds
+    // This lets us detect orphaned queries where the thread died
+    this.updateHeartbeatOnInterval(doc);
```
Kevin-Chant, line 669: This happens before the queries are split to be handled separately based on queued/running status, so now all queries will have a heartbeat

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -439,14 +438,17 @@ export abstract class QueryRunner<
         (q) => q.status === "running" || q.status === "queued"
       )
     ) {
-      const runningIds = this.model.queries
-        .filter((q) => q.status === "running")
+      const queryIds = this.model.queries
+        .filter((q) => q.status === "running" || q.status === "queued")
         .map((q) => q.query);
```
jdorn, line 443: I don't think it makes sense to combine running and queued here since we're doing 2 entirely separate actions.  Create 2 separate arrays instead.

For queued queries, we just need to cancel the timer.  For running, we need to get the queries from Mongo and try to cancel them in the data warehouse.

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -656,29 +652,31 @@ export abstract class QueryRunner<
...
+    this.updateHeartbeatOnInterval(doc);
+
     logger.debug("Created new query " + doc.id + " for " + name);
-    if (readyToRun) {
+    if (readyToRun && !concurrencyLimitReached) {
       this.executeQuery(doc, run, process);
-    } else if (dependenciesComplete && !runAtEnd) {
-      this.queuedQueryTimers[doc.id] = setTimeout(() => {
-        this.executeQueryWhenReady(doc, run, process);
-      }, INITIAL_CONCURRENCY_TIMEOUT);
```
jdorn, line 678: Was this `else if` block removed on purpose?

jdorn:
I think this PR would have been better split into 2 or 3 smaller ones to isolate the refactor part from the new functionality.  No need to change anything now, just a note for the future.

packages/back-end/src/queryRunners/QueryRunner.ts
Kevin-Chant: Yes, this got folded into the existing else block with a slight shuffle of the condition to make it easier to follow the logic.

jdorn: I'm not 100% following the code here, so hard for me to see where the `dependenciesComplete && !runAtEnd` check has moved to.

The only time we use `runAtEnd` is for DataBricks when pipeline mode is enabled.  I believe the intention is that it's a cleanup query that we want to run at the end no matter what, even if the other queries failed.  It's pretty hard to get into this state to test, so just flagging this as something we should try and verify to make sure there's not a regression.

Kevin-Chant: Here's the relevant part of this code block for this if/else logic:

```
const readyToRun = dependenciesComplete && !runAtEnd;
```
defines `readyToRun` as to whether this query wants to be executed as soon as concurrency is available. Then once the query doc is instantiated we hit this if/else (which used to be if/elif/else) for

`if (readyToRun && !concurrencyLimitReached)`, i.e. if we're 100% good to run it now, and then the else.
Within the else, the new `if` replaces the `else if` that this comment thread is tagged on:
```
if (readyToRun && concurrencyLimitReached) {
```
which is equivalent to `dependenciesComplete && !runAtEnd && concurrencyLimitReached`, i.e. we want to start it but there's a concurrency limit.

The reason I folded it into the `else` rather than having an `else if` for it is that the previous line needs to be set for these queries as well:
```
this.runCallbacks[doc.id] = { run, process };
```

jdorn: Ah, ok.  I think I follow now.

jdorn (approved):

PR #2980, opened 2024-09-12
Add public API endpoints to kick off snapshot and check status

lukesonnet:
### Features and Changes

Add simple snapshot endpoints:

* A way to create a new experiment snapshot for running experiments (no ability to customize the snapshot settings): `POST` to `/experiments/{id}/snapshot`.
* A way to check the status of a snapshot, eventually expandable to more snapshot data if we want: `GET` at `/snapshots/{id}`.

Powered by some simple controllers that check for permissions and basic logic before executing a snapshot.

New basic snapshot model just returns the `id`, `experiment`, and `status` for a snapshot to help you see if the snapshot succeeded or not.

### Testing

Tested in my local and could successfully:

- [X] Can kick off snapshot
- [X] Can get snapshot
- [x] Permissions checks on create snapshot (engineer role cannot create snapshot)
- [x] Permissions checks on get snapshot (snapshot for an experiment some project cannot be accessed by noaccess global role)


packages/back-end/src/api/experiments/postExperimentSnapshot.ts
```diff
@@ -0,0 +1,85 @@
...
+      throw new Error(`Experiment has no phases`);
+    }
+    // use last phase by default
+    const phase = experiment.phases.length - 1;
+    const dimension = undefined;
+    const useCache = true;
+
+    // This is doing an expensive analytics SQL query, so may take a long time
+    // Set timeout to 30 minutes
+    req.setTimeout(30 * 60 * 1000);
```
romain-growthbook, line 53: It would be nice to move this to a constant at the top of the file for visibility.

packages/back-end/src/api/snapshots/getExperimentSnapshot.ts
```diff
@@ -0,0 +1,32 @@
...
+export const getExperimentSnapshot = createApiRequestHandler(
+  getExperimentSnapshotValidator
+)(
+  async (req): Promise<GetExperimentSnapshotResponse> => {
+    const snapshot = await findSnapshotById(req.context.org.id, req.params.id);
+    if (!snapshot) {
+      throw new Error("Snapshot not found");
+    }
+    // no permission check in above method, so have to make sure they can read
+    // experiment first, which will be thrown by getExperimentById method
```
romain-growthbook, line 16: Do we need to capture this for future work?

packages/back-end/src/api/openapi/paths/getExperimentSnapshot.yaml
```diff
@@ -0,0 +1,28 @@
...
+      type: string
+get:
+  tags:
+    - snapshots
+  summary: Get an experiment snapshot status
+  operationId: getExperimentSnapshot
+  x-codeSamples:
+    - lang: "cURL"
+      source: |
+        curl -X GET https://api.growthbook.io/api/v1/snapshots/snp_1234/ \
```
romain-growthbook, line 16: nit: 
```suggestion
        curl -X GET https://api.growthbook.io/api/v1/snapshots/snp_1234 \
```

packages/back-end/src/api/openapi/openapi.yaml
```diff
@@ -170,8 +173,14 @@ paths:
       $ref: "./paths/getExperiment.yaml"
     post:
       $ref: "./paths/updateExperiment.yaml"
+  /experiments/{id}/snapshot:
```
romain-growthbook, line 176: Should this be plural?

packages/back-end/src/api/experiments/postExperimentSnapshot.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const snapshot = await createExperimentSnapshot({
+      context,
+      experiment,
+      datasource,
+      dimension,
+      phase,
+      useCache: true,
```
romain-growthbook, line 61:
```suggestion
      useCache,
```

romain-growthbook: I see that you're trying to capture the creation payload for the creation call and the audit call. You could consider capturing the whole object at once, it would be less bug prone:

```ts
const createPayload = {
        phase: experiment.phases.length - 1
        // dimension not needed
        useCache: true
}

const snapshot = await createExperimentSnapshot({
      context,
      experiment,
      datasource,
      ...createPayload
});

    await req.audit({
      event: "experiment.refresh",
      entity: {
        object: "experiment",
        id: experiment.id,
      },
      details: auditDetailsCreate({
        ...createPayload,
        manual: false,
      }),
    });

**Review by romain-growthbook — COMMENTED** [review]
This looks pretty straight forward. Only minor comments.

Would you consider adding tests? There are a few already there: https://github.com/growthbook/growthbook/tree/main/packages/back-end/test/api

It would be very useful when doing future changes on endpoint and refactorizing code to make sure that some of the asumptions you're making here don't get broken in the future.

File: packages/back-end/src/api/snapshots/getExperimentSnapshot.ts (hunk 2 above)
Comment (lukesonnet) [reply packages/back-end/src/api/snapshots/getExperimentSnapshot.ts:16]: maybe, but it's just replicating work elsewhere and I don't want to fix this permissions issue so we can get this out in 3.3

File: packages/back-end/src/api/openapi/paths/getExperimentSnapshot.yaml (hunk 3 above)
Comment (lukesonnet) [reply packages/back-end/src/api/openapi/paths/getExperimentSnapshot.yaml]: ty

File: packages/back-end/src/api/openapi/openapi.yaml (hunk 4 above)
Comment (lukesonnet) [reply packages/back-end/src/api/openapi/openapi.yaml:176]: I'm not sure. It is for a single snapshot for the experiment, not all the experiments snapshots, so I feel like this is more grammatically correct. but idk if we have a strong preference.

**Comment by lukesonnet** [comment]
I added a few tests, hope to land this in 3.3!

File: packages/back-end/test/api/snapshots.test.ts [hunk 6]
```diff
@@ -0,0 +1,144 @@
...
+    console.log(response.body);
+
+    expect(response.status).toBe(400);
+    expect(response.body).toEqual({
+      message: "Snapshot not found or no permission to access",
+    });
+  });
+
+  // Cannot successfully mock createExperimentSnapshot from controllers/experiments
+  // it("can post a snapshot", async () => {
```
romain-growthbook, line 85: nit: you could `xit` it to mark it as skipped so it is visibile in the logs.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -116,6 +117,8 @@ import { OrganizationSettings, ReqContext } from "back-end/types/organization";
 import { CreateURLRedirectProps } from "back-end/types/url-redirect";
 import { logger } from "back-end/src/util/logger";
 
+export const SNAPSHOT_TIMEOUT = 30 * 60 * 1000;
```
romain-growthbook, line 120: Nice

packages/back-end/src/api/openapi/openapi.yaml
romain-growthbook: It should be plural no sure why I suggested otherwise.. This is the pattern for all other endpoints already.

romain-growthbook (approved):
Looks great! Thanks for taking the time to add tests.

PR #2982, opened 2024-09-13
Remove @back-end imports

tzjames:
### Features and Changes

Using `@back-end` within shared package causes both swc and tsc to barf with misleading error messages.  swc creates extra .js and .js.map files within the /back-end/src dir which then cause files in the dist dir to be mangled.  tsc complains about `'rootDir' is expected to contain all source files`.

With the `@back-end` alias there VS-code will suggest that missing imports get added from there which can then waste a lot of developers time.

There is no need for the `@back-end` alias at all though since it is a package name, and pack-names work as is. 

### Testing

`yarn dev`.  Poke around the app
`yarn build`.  See it complete with no errors.
`yarn test`. 
`git grep "@back-end"` and see no results.
Click around the preview app and see no errors.

### DIFF
--- a/docs/scripts/gen-event-webhook-doc.ts
+++ b/docs/scripts/gen-event-webhook-doc.ts
@@ -1,6 +1,6 @@
 import fs from "node:fs";
 import path from "node:path";
-import { notificationEvents } from "@back-end/src/events/base-types";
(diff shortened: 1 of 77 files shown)


jdorn (approved):

romain-growthbook (approved):
Thanks for taking care of this.

Issue #2954, opened 2024-09-06
[Bug] Silently Failing Queries


fabianmarinog:
**Expected behavior when a query fails:**

GrowthBook reports detailed queries in the report view and in the query details view.

**Actual Behavior when a query fails:**

GrowthBook does not report failing queries in the report view, but in the query details view.

**Steps to reproduce:**

1. Add a known failing metric to a custom report
2. Use a Dimension in the configuration of the report to break down results for metrics
3. Click Refresh Data
4. Wait for the queries to finish/fail and Observe

A warning in the report shows there is an error with a query: 
![Image](https://github.com/user-attachments/assets/901cb71d-b06d-4c7e-a9c8-f08ac76fba76)
No error warning shows up. Instead, the metric just shows "no data" in the report:
![Image](https://github.com/user-attachments/assets/72557d0f-c566-4b1f-8741-cfdc9c611723)

GrowthBook Build: add1cb2 (2024-07-24)

lukesonnet:
(closed this out, thanks for the bug report!)

PR #2984, opened 2024-09-13
Report minor fixes

### LINKED ISSUE #2975: [Bug] CSV export of report doesn't include dimensions
### Summary
When using the "Export to CSV" button on a custom report that's slicing by a specific dimension, the dimension column isn't included in the output.

#### Expected Behavior

There should be one row per dimension/variation pair for each metric

#### Current Behavior

There is only one row per variation for each metric

lukesonnet:
Fixes #2975 and #2954 .

Both one line changes where more data needed to be added to a prop used by Reports.

## Testing

Dimension Report CSV now has dimension column.

Query errors now render in dimension reports for partially-succeeded runs.

Screenshot 2024-09-13 at 11 22 56 AM


romain-growthbook (approved):
Looks pretty straight forward to me.

PR #2989, opened 2024-09-14
Updated the skipPartialData default value

### LINKED ISSUE #2962: [Feature] Allow org-wide default of "Exclude in Progress Conversions" for Metric Conversion Windows
Submitting on behalf of a customer.

It would be great if we could change the default experiment analysis settings to "Exclude in Progress Conversions" instead of "Include" for the "Metric Conversion Windows".

For most of our metrics (like retention) it makes more sense to ensure that the denominator excludes users that haven't been enrolled in the experiment for the full conversion window. This is how we've typically calculated experiment metrics so hoping we can have this set as the default if possible. Thanks!

Screen Shot 2024-09-09 at 12 55 27 PM

faiznazeer:
### Features and Changes

Updated the default value of skipPartialData from false to true. This variable is later used to set the value of **Metric Conversion Windows**. Hence the default value of **Metric Conversion Windows** is updated to **Exclude in Progress Conversions**

- Closes #2962

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -535,7 +535,7 @@ export async function postExperiments(
     activationMetric: data.activationMetric || "",
     segment: data.segment || "",
     queryFilter: data.queryFilter || "",
-    skipPartialData: !!data.skipPartialData,
+    skipPartialData: data.skipPartialData || true,
     attributionModel: data.attributionModel || "firstExposure",
     variations: data.variations || [],
     implementation: data.implementation || "code",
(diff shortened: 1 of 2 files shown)


lukesonnet (changes requested):
Hi, sorry, but this is just changing the default, not creating an org-level default setting that GrowthBook users can choose to change their default. It also has other ramifications for how we optimize queries for fact tables, so this may not be a great candidate for a community contribution.

lukesonnet:
I'm going to close this for now! Thanks for taking a stab at it. You're free to re-open it if you want to take a stab at organization settings, but it may be a bit in the weeds for a first time contribution.

PR #2991, opened 2024-09-16
Show ISO country codes in selection dropdown

Kevin-Chant:
### Features and Changes

One-liner to add the 2-digit country code to the dropdown so people can more easily see what value the country's name will correspond to. Especially helpful with odd cases like Algeria becoming DZ

### Testing

Played with the selector modal. Typeahead still works effectively & the value being selected is still just the 2-digit code

### Screenshots
image

image

image

image

### DIFF
--- a/packages/front-end/components/Forms/CountrySelector.tsx
+++ b/packages/front-end/components/Forms/CountrySelector.tsx
@@ -27,7 +27,7 @@ export default function CountrySelector(props: CountrySelectorProps) {
   }));
 
   const formatOptionLabel = ({ label, value }, { context }) => {
-    const text = context === "menu" ? label : value;
+    const text = context === "menu" ? label + ` (${value})` : value;
     const flag = getCountryFlagEmojiFromCountryCode(value);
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Forms/CountrySelector.tsx
```diff
@@ -27,7 +27,7 @@ export default function CountrySelector(props: CountrySelectorProps) {
   }));
 
   const formatOptionLabel = ({ label, value }, { context }) => {
-    const text = context === "menu" ? label : value;
+    const text = context === "menu" ? label + ` (${value})` : value;
```
gazzdingo, line 30: super nit 
```suggestion
    const text = context === "menu" ? `${label} (${value})` : value;
```

gazzdingo: just so you have a space between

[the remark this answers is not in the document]
Kevin-Chant: Thats what i get for coding while in the weekly meeting xD

gazzdingo (approved):
LGTM

PR #3000, opened 2024-09-17
Add display for saved group with no projects

Kevin-Chant:
image

### DIFF
--- a/packages/front-end/pages/saved-groups/[sgid].tsx
+++ b/packages/front-end/pages/saved-groups/[sgid].tsx
@@ -281,12 +281,16 @@ export default function EditSavedGroupPage() {
               <div className="mr-2">Projects:</div>
 
               <div>
-                <div className={"d-flex align-items-center"}>
-                  <ProjectBadges
-                    projectIds={savedGroup.projects}
-                    resourceType="saved group"
-                  />
-                </div>
+                {savedGroup.projects?.length ? (
+                  <div className={"d-flex align-items-center"}>
+                    <ProjectBadges
+                      projectIds={savedGroup.projects}
+                      resourceType="saved group"
+                    />
+                  </div>
+                ) : (
+                  <strong>None</strong>
+                )}
               </div>
             </div>
           )}


packages/front-end/pages/saved-groups/[sgid].tsx
```diff
@@ -281,12 +281,16 @@ export default function EditSavedGroupPage() {
...
-                </div>
+                {savedGroup.projects?.length ? (
+                  <div className={"d-flex align-items-center"}>
+                    <ProjectBadges
+                      projectIds={savedGroup.projects}
+                      resourceType="saved group"
+                    />
+                  </div>
+                ) : (
+                  <strong>None</strong>
```
Comment (msamper) [inline packages/front-end/pages/saved-groups/[sgid].tsx:292]: Hmm maybe the logic should be more like how we do it on the ID Lists list page on second thought
```
                            {(s?.projects?.length || 0) > 0 ? (
                              <ProjectBadges
                                resourceType="saved group"
                                projectIds={s.projects}
                                className="badge-ellipsis short align-middle"
                              />
                            ) : (
                              <ProjectBadges
                                resourceType="saved group"
                                className="badge-ellipsis short align-middle"
                              />
                            )}
```
I think we want it to render as `All Projects` instead of none

msamper (changes requested):

packages/front-end/pages/saved-groups/[sgid].tsx
Comment (Kevin-Chant) [reply packages/front-end/pages/saved-groups/[sgid].tsx:292]: yeah sounds right

Comment (Kevin-Chant) [reply packages/front-end/pages/saved-groups/[sgid].tsx:292]: Wait, why do we need this to be in a ternary? Shouldn't we just make `ProjectBadges` do that if `projectIds` is an empty list? (probably out of scope for right now cause I want to get your changes merged, but it's weird to me to have this behavior)

msamper (approved):

packages/front-end/pages/saved-groups/[sgid].tsx
Comment (Kevin-Chant) [reply packages/front-end/pages/saved-groups/[sgid].tsx]: I'm fairly sure that's from this line. Either way we can deal with that behavior separately
https://github.com/growthbook/growthbook/blob/msamper/project-save-groups/packages/back-end/src/models/SavedGroupModel.ts#L151

PR #3004, opened 2024-09-18
Smith/scaled impact update

lukebrawleysmith:
### Features and Changes

updates scaled impact calculation to permit bandit calculation.  The same calculation will be used for both bandits and A/B tests.  This approach reduces variance for the A/B test scaled impact calculation. 

### Dependencies

none
### Testing

tested against current approach in both freq and Bayesian engines, and results were appropriately similar


packages/stats/gbstats/bayesian/tests.py
```diff
@@ -126,14 +128,18 @@ def chance_to_win(self, mean_diff: float, std_diff: float) -> float:
...
+    def scale_result(self, result: BayesianTestResult) -> BayesianTestResult:
         if result.uplift.dist != "normal":
             raise ValueError("Cannot scale relative results.")
-        if p == 0:
+        if (
+            self.phase_length_days == 0
+            or self.traffic_percentage == 0
+            or not isinstance(self.stat_a, ScaledImpactStatistic)
+        ):
             return self._default_output(ZERO_SCALED_VARIATION_MESSAGE)
```
lukesonnet, line 139: This is the wrong error message for the `not is instance` check above. I know these are opaque/not used right now, but we should create the right message.

lukesonnet: Also, we should throw an error if the total_users is None.

packages/stats/gbstats/models/tests.py
```diff
@@ -15,8 +15,9 @@
 @dataclass
 class BaseConfig:
     difference_type: DifferenceType = "relative"
-    traffic_proportion_b: float = 1
+    traffic_percentage: float = 1
     phase_length_days: float = 1
+    total_users: int = 0
```
lukesonnet, line 20: Make the default here None and make it optional so we can properly throw errors.

lukesonnet (changes requested):
Left some requested changes inline. Can you also check what happens in the UI when you do scaled impact with a quantile metric before and after?

packages/stats/gbstats/bayesian/tests.py
lukebrawleysmith: done

packages/stats/gbstats/models/tests.py
lukebrawleysmith: done

lukebrawleysmith:
I made the requested changes, and compared scaled impact before and after changes using both revenue and quantile metrics.  Scaled impact now no longer shows for quantile metrics. 
before changes: 
scaled_impact_before_changes
after changes (pls note the top metric is a revenue metric): 
scaled_impact_after_changes

PR #3005, opened 2024-09-18
Add names for modal tracking events

Kevin-Chant:
### Features and Changes

Adds tracking event names to a smattering of existing modals

### Testing

I haven't tested this at all yet; we should do at least a spot check of a few random modals and make sure the tracking events are firing correctly to Jitsu.


Kevin-Chant:
@jdorn I was hoping to get this further along (ideally to 100% coverage) but I'll probably have to hand it off here. As I noted in the PR description this is just coming from editor changes, I didn't do any testing yet. I'll let you and @lukesonnet decide how to divvy it up from here

PR #3009, opened 2024-09-18
Modify slack notif message

lukesonnet:


lukesonnet:
Abandoning this, because frankly it's nice in some ways, but we should consider it potentially a different kind of update "digest" and we should leave these base change alone.

PR #3011, opened 2024-09-18
Add org date created as SDK attribute

lukesonnet:
### Features and Changes

Adds org date created as an attribute, useful for targeting to recent orgs, old orgs, etc. 

First use will be to hide the "Slack" sidebar button since it is currently not used in favor of Webhooks and clutters the UI for no reason. That sidebar button and subsequent flow is also undocumented.

### Screenshots

Targeting in `dev` if the date created is after today to remove Slack from sidebar.

You can see my old dev org gets the right date and I still get the slack sidebar:
Screenshot 2024-09-18 at 11 29 16 AM
Screenshot 2024-09-18 at 11 33 41 AM

When I modify the user attributes to be `""` it correctly is not greater than a date as well, and when I set the date to the future, the slack sidebar goes away:
Screenshot 2024-09-18 at 11 30 57 AM

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -776,6 +776,7 @@ export async function getOrganization(req: AuthRequest, res: Response) {
       messages: messages || [],
       pendingMembers: org.pendingMembers,
       getStartedChecklistItems: org.getStartedChecklistItems,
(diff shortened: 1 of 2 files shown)


lukesonnet:
If we land this, I would make the following change to disable slack in the sidebar for new orgs.

Screenshot 2024-09-18 at 11 35 29 AM

jdorn (approved):

PR #3024, opened 2024-09-20
Add service name/namespace/version to opentelemetry tracing.

romain-growthbook:
This should help our observability infra.

### DIFF
--- a/packages/back-end/src/tracing.ts
+++ b/packages/back-end/src/tracing.ts
@@ -19,8 +19,11 @@ import {
   getNodeAutoInstrumentations,
   getResourceDetectors,
 } from "@opentelemetry/auto-instrumentations-node";
+import { Resource } from "@opentelemetry/resources";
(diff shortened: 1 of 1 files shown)


packages/back-end/src/tracing.ts
```diff
@@ -36,6 +39,11 @@ const sdk = new opentelemetry.NodeSDK({
   instrumentations: getNodeAutoInstrumentations(),
   resourceDetectors: getResourceDetectors(),
   metricReader,
+  resource: new Resource({
+    [SemanticResourceAttributes.SERVICE_NAME]: "growthbook",
+    [SemanticResourceAttributes.SERVICE_NAMESPACE]: "backend",
+    [SemanticResourceAttributes.SERVICE_VERSION]: getBuild().sha,
+  }),
```
jdorn, line 46: These are good default values. If env vars are specified for any of these, they should take precedence. Specifically service name. That may already be happening (I didn't test it), but just noting it here.

jdorn (approved):

PR #3027, opened 2024-09-23
Generate sentry source maps at release.

romain-growthbook:
The changes in this PR makes it so that, when deploying a production release, we extract sourcemaps from the docker build, injects sentry debug IDs and create a release on sentry with the sourcemaps.

This should help debugging by being able to pinpoint the exact source location where an error occurs and correlating them with releeases.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -93,8 +93,8 @@ jobs:
           version: ${{ steps.metadata.outputs.sha }}
         env:
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
-          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
-          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+          SENTRY_ORG: growthbook
+          SENTRY_PROJECT: api
 
       - name: Deploy docker image to ECS for GrowthBook Cloud API
         run: aws ecs update-service --cluster prod-api --service prod-api --force-new-deployment --region us-east-1
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -75,26 +75,21 @@ jobs:
           docker rm temp-container
           aws s3 sync ./static s3://growthbook-cloud-static-files/_next/static
 
-      - name: Checkout code
-        uses: actions/checkout@v2
-
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
           node-version: '20'
-          cache: 'yarn'
 
       - name: Generate sentry sourcemaps
         run: |
-          mv ./back-end-dist ./packages/back-end/dist
-          yarn install
-          yarn sourcemap:inject
+          npm install -g @sentry/cli
+          npm exec sentry-cli -- sourcemaps inject ./back-end-dist
 
       - name: Create Sentry release
         uses: getsentry/action-release@v1
         with:
           environment: 'production'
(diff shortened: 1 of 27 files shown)


.github/workflows/deploy.yml
```diff
@@ -52,6 +52,31 @@ jobs:
...
+      - name: Create Sentry release
+        uses: getsentry/action-release@v1
+        with:
+          environment: 'production'
+          sourcemaps: './packages/back-end/dist'
+          version: ${{ steps.metadata.outputs.sha }}
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
```
jdorn, line 78: I would move these steps to the `prod` job below.  The `docker` job is just about uploading the build artifact to Docker Hub.  The `prod` job is about deploying this new image to our Cloud.

A nice side effect of this is that it avoids an unnecessary Docker download.  Both the "Generate sentry sourcemaps" and "Extract static assets" steps need the Docker image, so if they are in the same job, we can share the cost.

romain-growthbook: Excellent suggestion thanks.

.github/workflows/deploy.yml
```diff
@@ -70,8 +70,31 @@ jobs:
...
+      - name: Create Sentry release
+        uses: getsentry/action-release@v1
+        with:
+          environment: 'production'
+          sourcemaps: './back-end-dist'
+          version: ${{ steps.metadata.outputs.sha }}
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
```
tzjames, line 97: I don't think these need to be secrets.  We can just hard code them in.

.github/workflows/deploy.yml
```diff
@@ -70,8 +70,31 @@ jobs:
...
+          npm exec sentry-cli -- sourcemaps inject ./back-end-dist
+
+      - name: Create Sentry release
+        uses: getsentry/action-release@v1
+        with:
+          environment: 'production'
+          sourcemaps: './back-end-dist'
+          version: ${{ steps.metadata.outputs.sha }}
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
```
tzjames, line 95: I think @jdorn will need to add this one to our secrets. Here is where the token can be created https://growthbook.sentry.io/settings/auth-tokens/

tzjames (commented):
In order to have the individual error events tagged we need to add it. 
 I actually did it in https://github.com/growthbook/growthbook/pull/3050/files which I will abandon since I also did the sentry github action there which this PR takes care of.

.github/workflows/deploy.yml
romain-growthbook: Done

romain-growthbook:
Thanks for the review. I think that this is ready but will need to be prepped with the secret and monitored during deploy.

tzjames (approved):
Looks good.  Hope it works.  I like your suggestion to make sure that there is nothing else that is urgent to deploy as it might take a few tries.

jdorn (approved):

PR #3028, opened 2024-09-24
Added metric groups

Auz:
Added the ability to make metric groups, which can be sorted and reordered, and are passed by reference when added to experiments. 

Screenshot 2024-09-23 at 7 45 15 PM

Screenshot 2024-09-23 at 7 46 38 PM

metric-group-sorting


packages/back-end/src/routers/metric-group/metric-group.controller.ts
```diff
@@ -0,0 +1,187 @@
...
+    context.permissions.throwPermissionError();
+  }
+
+  const metricGroup = await context.models.metricGroups.getById(req.params.id);
+
+  if (!metricGroup) {
+    throw new Error("Could not find the metric group");
+  }
+
+  // should we delete all references to this metric group in the experiments?
```
bryce-fitzsimons, line 110: i think no. eventually we will have metric versioning, and could show better status messages

bryce-fitzsimons (approved):

PR #3031, opened 2024-09-25
Move generation of the key to store/retrieve a sticky bucket assignment into separate method

avokhmin:
### Move generation of the key to store/retrieve a sticky bucket assignment into separate method

In this PR we moved the generation of  the key to store/retrieve a sticky bucket assignment into separate method.
It should help to use custom logic to generate a keys with custom sticky bucket services. Also, it will make code more clear.

Example:

```js
class MyBrowserCookieStickyBucketService extends BrowserCookieStickyBucketService {
  getKey(attributeName, attributeValue) {
    // Some web servers can encode cookie keys with `||` into `%7C%7C`
    //  In this case, we can override the `getKey` and use other characters, e.g. `XX`
    return `${this.prefix}${attributeName}XX${attributeValue}`;
  }
}
```

### DIFF
--- a/packages/sdk-js/src/sticky-bucket-service.ts
+++ b/packages/sdk-js/src/sticky-bucket-service.ts
@@ -46,6 +46,13 @@ export interface ResponseCompat {
  * Responsible for reading and writing documents which describe sticky bucket assignments.
  */
 export abstract class StickyBucketService {
+  protected prefix: string;
+
+  constructor(opts?: { prefix?: string }) {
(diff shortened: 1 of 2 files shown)


bryce-fitzsimons:
Thanks for the PR. Rolling this change into a separate PR here:
https://github.com/growthbook/growthbook/pull/3040

avokhmin:
https://github.com/growthbook/growthbook/pull/3040 is merged 🎉

PR #3032, opened 2024-09-25
Remove relative imports in the back-end

romain-growthbook:
This is pass #2 at removing relative imports in the back-end now that @tzjames has cleaned up the import path!

The advantages of using absolute import paths over relative paths are:
* Imports are canonical and only related to the _target_ file.

This means that importing `./src/path/to/foo.ts` is always:
```typescript
import ... from "back-end/src/path/to/foo"
```
Regardless of where the importing file is located

This makes it easier to refactor and reorganize the code:
  * If the imported file is moved, all imports can be fixed in a single rewrite cal
  * If the importing file is move all imports remain valid!
  * When splitting up code, imports can be copied/paste into a new file regardless of this file's location.

This also greatly enhances code readibility, in particular when two import paths share a common suffix, for instance `back-end/types` and `back-end/src/types`. In this case, relative imports of the form `"../types/..."` are confusing. With absolute imports, it is very clear which import path is expected.


.eslintrc
```diff
@@ -109,6 +109,7 @@
     },
     {
       "files": [
+        "./packages/back-end/**/*.ts*",
```
tzjames, line 112: nice that you are linting for this now

tzjames (approved):
The link to pass #2 in your description is perhaps not the PR you meant to link to?

Could also provide a rationale for why you are removing the relative imports and a test plan as well as part of the commit message?  

The test plan in this case is probably good enough to say the lint passed, the tests passed, and the build threw no errors.  I'd also say that you poked around the preview environment and everything seemed to be loading as normal.  Though I just tested the preview environment now and the back-end doesn't seem to be working ... we should probably fix that.

tzjames (changes requested):
I don't see anything wrong in the code, but requesting changes until the preview env works for this change set.

tzjames (approved):
Thanks for making sure there is a way to get this to work with VScode before landing it.

PR #3036, opened 2024-09-26
create ConditionalWrapper components, use to fix modal footer

bryce-fitzsimons:
Helper HOC for wrapping things with things based on a condition. Apply it to modal footer to fix UI regression.

before:
image

after:
image


msamper (approved):
Tested and looks good! Thanks for fixing 🙏🏼

PR #3038, opened 2024-09-26
Support Inline Filters for Fact Metrics

jdorn:
### Features and Changes

Add support for quick inline filtering when creating a metric.  This is an alternative to a 2-step process of defining a filter and then referencing it.

These inline filters only support the form `{column} IN (value1, value2, ...)`.  For anything more complex, you can continue to use standalone filters which allow full SQL.

Because of this limitation, we can provide a much nicer UI.  We run a quick SQL query to get the top values for the column in the past 7 days and use that to power an autocomplete multi-select field.  The query results are persisted in Mongo and only refreshed manually, so the load on the db should be minimal.

This first iteration lets you opt a string column into inline filtering.  Only these opted-in columns show up in the metric form. Future iterations may allow inline filtering on any column without opting in first.

We also use these opted-in columns to auto-suggest metrics you should create with a UI to review and add them in bulk.

### Screenshots

Mark a column to always be available for inline filtering.

image

Tooltip:

image

These columns are identified on the fact table page with a filter icon:

image

We query the top 100 values in the past 7 days and use that to power an auto-complete.  We still allow typing custom values.

image

Fall back to text input if there are no events in the past 7 days.

image

Auto-suggest metrics based on the opted-in columns and their top values:

image

image

Filters show up on the fact metric page correctly (and get injected into the SQL correctly):

image

### Future Improvements

- Allow inline filtering on any column without opting-in first
- Relationships between columns.  For example, first prompt for `event_type`.  If they select "Page View", then prompt to filter on the `page_location` column.
- Support non-string columns
- Auto-refresh top values (or provide a nicer way to manually refresh them in the UI)
- Allow more customization for auto-suggested metric names
- Look for numeric columns and suggest `mean`, `ratio`, and/or `quantile` metrics based on some heuristics


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3891,19 +3893,57 @@ export default abstract class SqlIntegration
...
+  public async runColumnTopValuesQuery(
+    sql: string
+  ): Promise<ColumnTopValuesResponse> {
+    const { rows, statistics } = await this.runQuery(sql);
+
+    return {
+      statistics,
+      rows: rows.map((r) => ({
+        value: r.value + "",
+        count: 1 * r.count,
```
lukesonnet, line 3942: Should we prefer this approach going forward over parseFloat and other methods?

packages/front-end/components/FactTables/ColumnModal.tsx
```diff
@@ -56,7 +59,19 @@ export default function ColumnModal({ existing, factTable, close }: Props) {
             name: value.name,
             numberFormat: value.numberFormat,
             datatype: value.datatype,
+            alwaysInlineFilter: value.alwaysInlineFilter,
           };
+
+          if (
+            data.alwaysInlineFilter &&
```
lukesonnet, line 66: Should we warn if existing.alwaysInlineFilter and you change the datatype from string AND it's used on any metrics? I know it keeps working, but that is kind of good luck. If you actually change the column data type you probably don't *want* it to keep working as an inline filter!

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3891,19 +3893,57 @@ export default abstract class SqlIntegration
...
+      startDate: start,
+      templateVariables: {
+        eventName: factTable.eventName,
+      },
+    })}
+  ),
+  __topValues as (
+    SELECT
+      ${column.column} as value,
+      COUNT(*) as count
```
lukesonnet, line 3922: nit: I kind of want us to start being a bit more consistent and capitalize keywords. So can we write `AS` here and in the CTE expressions?

I'm personally also using leading commas since so much of what we do is composable and this leads to fewer errors once you get used to it, but I'm less convinced that makes sense since it can be less readable.

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -220,12 +281,46 @@ function ColumnRefSelector({
             required
           />
         </div>
+        {inlineFilterFields.map(({ label, key, options, error }) => (
+          <div className="col-auto" key={key}>
+            <MultiSelectField
```
lukesonnet, line 286: WDYT about formatting this so as soon as you select a second value, we write "is any of" after the column name (e.g. "browser is any of")? 

... or, put "any of" somewhere more attractive to signify that it's going to be an OR
... or, even write IN( multi-select field here ), but that's clunkier.

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -235,7 +330,7 @@ function ColumnRefSelector({
                 label: f.name,
                 value: f.id,
               }))}
-              placeholder="All Rows"
+              placeholder={inlineFilterFields.length > 0 ? "None" : "All Rows"}
```
lukesonnet, line 333: WDYT a "test" functionality on the inline filter fields to make sure you can get some data, especially if you're creating a field (since this is now harder)? Maybe it's off by default, or it's literally a "test" button that shows up when you Create a new field?

IDK may be overkill, but it could be easy to misspecify this field and I use the test function for Filters all the time.

lukesonnet (approved):
LGTM!
Mostly looked at the metric/column creation modals and the SQL stuff and did some testing in app.

packages/front-end/components/FactTables/FactMetricModal.tsx
jdorn: Yeah, a test button when creating a metric makes sense, even before this PR. Even if we just show a count of matched events in the past X days.

Will probably do this in a separate PR

packages/front-end/components/FactTables/FactMetricModal.tsx
jdorn: I can put a similar tooltip as the Additional Filters input which says how multiple values are treated.

packages/back-end/src/integrations/SqlIntegration.ts
jdorn: I'll change this to use `parseFloat`.  It's more clear that way.

packages/front-end/components/FactTables/ColumnModal.tsx
jdorn: It's such an edge case, I wouldn't worry about it.

PR #3043, opened 2024-09-27
Fix back-end dev compilation.

romain-growthbook:
This must have gotten lost during the final tweaks of https://github.com/growthbook/growthbook/pull/3032

This also revealed some issues when json import when using swc in the tests. Thus, the tests are switched to `ts-jest` to use plain `typescript` compilation options.


bryce-fitzsimons (approved):

bryce-fitzsimons (changes requested):
BE tests failing

bryce-fitzsimons (approved):
working after yarn install & yarn setup

bryce-fitzsimons (changes requested):
Please make sure the Okteto deploy failure isn't a problem. I have a hunch we're hitting some bumps because of the `frozen-lockfile` rule. This PR makes a lot of changes to `yarn.lock`

PR #3049, opened 2024-10-01
Minor changes

Auz:
Removed the word option from the SDK instructions for attributes as it's confusing. Added the ability for self-hosting instances to stop real time feature monitoring.

### DIFF
--- a/packages/front-end/pages/_app.tsx
+++ b/packages/front-end/pages/_app.tsx
@@ -45,7 +45,9 @@ export const growthbook = new GrowthBook<AppFeatures>({
       : "sdk-UmQ03OkUDAu7Aox",
   enableDevMode: true,
   subscribeToChanges: true,
-  realtimeKey: "key_prod_cb40dfcb0eb98e44",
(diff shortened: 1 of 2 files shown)


packages/front-end/pages/_app.tsx
```diff
@@ -45,7 +45,9 @@ export const growthbook = new GrowthBook<AppFeatures>({
       : "sdk-UmQ03OkUDAu7Aox",
   enableDevMode: true,
   subscribeToChanges: true,
-  realtimeKey: "key_prod_cb40dfcb0eb98e44",
+  realtimeKey: process?.env?.DISABLE_TELEMETRY
```
tzjames, line 48: process.env is not available on the front-end.  You need:
```
realtimeKey: isTelemetryEnabled() ? "key_prod_cb40dfcb0eb98e44" : "",
```

tzjames (changes requested):

tzjames (approved):
Ok looks good now.

PR #3052, opened 2024-10-01
Pass logs in opentelemetry.

romain-growthbook:
This PR enables logs output in opentelemetry setup.

I was able to confirm collection via a locally run collector:
```
2024-10-01T11:26:41.134-0500	info	MetricsExporter	{"kind": "exporter", "data_type": "metrics", "name": "debug", "resource metrics": 1, "metrics": 5, "data points": 11}
2024-10-01T11:26:41.335-0500	info	LogsExporter	{"kind": "exporter", "data_type": "logs", "name": "debug", "resource logs": 1, "log records": 1}
2024-10-01T11:26:42.139-0500	info	MetricsExporter	{"kind": "exporter", "data_type": "metrics", "name": "debug", "resource metrics": 1, "metrics": 5, "data points": 11}
```

Logs collection can be disabled by setting the `GROWTHBOOK_OTEL_DISABLE_LOGS_COLLECTION` environment variable.


packages/back-end/src/tracing.ts
```diff
@@ -36,8 +39,12 @@ const metricReader = new PeriodicExportingMetricReader({
 });
 
 const sdk = new opentelemetry.NodeSDK({
-  instrumentations: getNodeAutoInstrumentations(),
+  instrumentations: [
+    ...getNodeAutoInstrumentations(),
+    new PinoInstrumentation(),
+  ],
   resourceDetectors: getResourceDetectors(),
+  logRecordProcessor: new BatchLogRecordProcessor(new OTLPLogExporter()),
```
jdorn, line 47: For self-hosted users who are currently using OpenTelemetry, is this a backwards incompatible change?  Lets say someone is already piping logs into Datadog (e.g. through CloudWatch), is this going to start double logging?

Wondering if we should make this opt-in with an environment variable.

Also, it's worth testing this end-to-end with the Datadog Agent to make sure logs are tagged properly, etc. in the Datadog UI

romain-growthbook: Opt-in with an env variable is a great idea! 

I think that to test with a datadog agent, I'd need at least to know the exact configuration of ours in production otherwise there's no way to know if the test is relevant or not.

We could also test directly in production. Not sure if there's much risk, I imagine these mechanism must be designed to be robust when the collector is not available.

romain-growthbook: Added `GROWTHBOOK_OTEL_DISABLE_LOGS_COLLECTION`

jdorn (approved):

PR #3057, opened 2024-10-02
Base Radix badge component.

romain-growthbook:
This PR adds the base `Badge` component for Radix:

Screenshot 2024-10-02 at 3 31 13 PM
Screenshot 2024-10-02 at 3 31 21 PM

### DIFF
--- a/packages/front-end/components/Radix/Badge.tsx
+++ b/packages/front-end/components/Radix/Badge.tsx
@@ -1,0 +1,33 @@
+import { Flex, Badge as RadixBadge } from "@radix-ui/themes";
+import { MarginProps } from "@radix-ui/themes/dist/cjs/props/margin.props";
+import { RadixColor } from "@/components/Radix/HelperText";
+
+type Props = {
+  content: string;
+  color?: RadixColor;
+  variant?: "solid" | "soft" | "surface" | "outline";
+  highContrast?: boolean;
+  radius?: "none" | "small" | "medium" | "large" | "full";
+} & MarginProps;
+
+export default function Badge({
+  content,
+  color,
+  variant,
+  highContrast,
+  radius,
+  ...containerProps
+}: Props) {
+  return (
+    <Flex {...containerProps}>
+      <RadixBadge
+        color={color}
+        variant={variant}
+        highContrast={highContrast}
+        radius={radius}
+      >
+        {content}
+      </RadixBadge>
+    </Flex>
+  );
+}
--- a/packages/front-end/components/Radix/HelperText.tsx
+++ b/packages/front-end/components/Radix/HelperText.tsx
@@ -9,6 +9,8 @@ import {
 
 export type Status = "info" | "warning" | "error" | "success";
 
+export type RadixColor = TextProps["color"];
+
 export function getRadixColor(status: Status): TextProps["color"] {
   switch (status) {
     case "info":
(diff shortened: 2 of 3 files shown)


packages/front-end/components/Radix/Badge.tsx
```diff
@@ -0,0 +1,33 @@
+import { Flex, Badge as RadixBadge } from "@radix-ui/themes";
+import { MarginProps } from "@radix-ui/themes/dist/cjs/props/margin.props";
+import { RadixColor } from "@/components/Radix/HelperText";
+
+type Props = {
+  content: string;
+  color?: RadixColor;
+  variant?: "solid" | "soft" | "surface" | "outline";
+  highContrast?: boolean;
+  radius?: "none" | "small" | "medium" | "large" | "full";
```
jdorn, line 10: Looking at the Figma file, I think we can restrict these options a bit:
- Only support `solid` and `soft` variants
- Remove `radius` and `highContrast` props entirely

Also, for color, this is the one component where we want to be able to choose any Radix color, not just the limited palette used by the HelperText component.

packages/front-end/components/Radix/Badge.tsx
```diff
@@ -0,0 +1,33 @@
...
+}: Props) {
+  return (
+    <Flex {...containerProps}>
+      <RadixBadge
+        color={color}
+        variant={variant}
+        highContrast={highContrast}
+        radius={radius}
+      >
+        {content}
```
jdorn, line 29: Is there a reason we need to wrap this in `<Flex>`?  Also, I think I prefer using `children` instead of a custom `content` prop here.

romain-growthbook: Children is cool. For flex, I followed the examples here: https://www.radix-ui.com/themes/docs/components/badge

packages/front-end/components/Radix/Badge.tsx
romain-growthbook: I found a place to replace one and it looks like we're gonna need `radius` if we want to be able to keep a similar look. See screenshot in updated description.

jdorn: We're not trying to keep a similar look to what's in the product today, so I think it's ok to get rid of `radius` and just do what's in the design system Figma

packages/front-end/components/Radix/Badge.tsx
jdorn: Those examples are confusing, but I think they are just using Flex since they are showing multiple example badges next to each other and want to add spacing between them.  A single badge on its own doesn't need to be wrapped in Flex

packages/front-end/components/Radix/Badge.tsx
```diff
@@ -0,0 +1,13 @@
+import { Badge as RadixBadge } from "@radix-ui/themes";
+import { MarginProps } from "@radix-ui/themes/dist/cjs/props/margin.props";
+import { RadixColor } from "@/components/Radix/HelperText";
+
+type Props = {
+  children: string | JSX.Element;
```
jdorn, line 6: I've been trying to avoid accepting `JSX.Element` as a prop since I think it makes it too easy for engineers to deviate from the design system.

romain-growthbook: If we're not accepting `JSX.Element` then I think that we should use another name for the parameter. `children` is the convention for child elements so it will confuse people if you can't use it this way. Also, react does some caching/mangling of `children`.

I had it initially as a string and switched to a full element to be able to add bold/styling to the content. I can also imagine use case where we might want to pass an icon etc.

romain-growthbook:
PR updated with label back to regular string attribute. Should be ready to go.

jdorn (approved):

PR #3058, opened 2024-10-02
[Radix] Create wrapper around Radix Callout component

msamper:
### Features and Changes

Creates a wrapped Radix Callout component based on our design system. Replaced both uses of the Radix Callout component in the new setup flow with our new wrapped component.

I also went ahead and alphabetized the existing components on the design system page to make it easier to find them in the future.

### Screenshots

Screenshot 2024-10-02 at 1 48 50 PM

### DIFF
--- a/packages/front-end/components/Radix/Callout.tsx
+++ b/packages/front-end/components/Radix/Callout.tsx
@@ -1,20 +1,25 @@
 import { Callout as RadixCallout } from "@radix-ui/themes";
 import { ReactNode } from "react";
+import { MarginProps } from "@radix-ui/themes/dist/cjs/props/margin.props";
 import { RadixStatusIcon, Status, getRadixColor } from "./HelperText";
 import styles from "./RadixOverrides.module.scss";
+
+type Props = {
+  children: ReactNode;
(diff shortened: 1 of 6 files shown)


packages/front-end/components/Radix/Callout.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { ReactNode } from "react";
+import { RadixStatusIcon, Status, getRadixColor } from "./HelperText";
+import styles from "./RadixOverrides.module.scss";
+
+export default function Callout({
+  children,
+  type,
+}: {
+  children: ReactNode;
+  type: Status;
```
jdorn, line 11: Lets use the same `status` prop name as we do with HelperText to keep things consistent.

jdorn: Also, follow the pattern in Checkbox where you merge in `MarginProps` to the accepted props and pass them to the outer container.  This will allow getting rid of wrapping `<div className="mt-3">`... components and specifying margins directly.

jdorn (approved):

PR #3060, opened 2024-10-02
RadioGroup radix component

lukesonnet:
### Features and Changes

Adds the `RadioGroup` radix component. 

Did not:
* add standalone `Radio` component since it wasn't in the notion hit list, but could be easily pulled out of the meat of this component.
* add a way to set a title/label for the whole group

Replaces all instances `RadioSelector` with `RadioGroup`. There are still a few `"radio"` inputs but they are *all* horizontal right now.

### Testing

Clicked through most (not all!) options to ensure values were set correctly.

#### Design system page

Screenshot 2024-10-04 at 3 29 11 PM
Screenshot 2024-10-04 at 3 29 15 PM

#### Classic metric creation modal

Before
Screenshot 2024-10-02 at 3 14 46 PM

After
Screenshot 2024-10-03 at 8 53 50 AM

#### Experiment make changes flow
Now warning appears conditionally, no tooltip.

Before
Screenshot 2024-10-03 at 10 54 48 AM

After 1
Screenshot 2024-10-04 at 3 14 33 PM

After 2
Screenshot 2024-10-04 at 3 14 36 PM

####  `PowerCalculationStatsEngineSettingsModal`
Completely rewrote this one to be simpler and use component.

Before
Screenshot 2024-10-02 at 4 27 13 PM

After
Screenshot 2024-10-04 at 2 56 17 PM

#### Health tab onboarding

Before
Screenshot 2024-10-02 at 3 59 06 PM

After
Screenshot 2024-10-02 at 3 58 58 PM

#### Initial connection form

Before

Screenshot 2024-10-02 at 4 04 24 PM

After
Screenshot 2024-10-02 at 4 01 16 PM

#### Review approve changes

Before
Screenshot 2024-10-02 at 4 11 16 PM

After
Screenshot 2024-10-02 at 4 11 00 PM

### Open questions

* Font is still weird; options are grey, I think inheriting from Modal
* How to deal with some horizontal radios, all of which tend to move items below them around
Screenshot 2024-10-04 at 3 21 05 PM


packages/front-end/components/Radix/RadioGroup.tsx
```diff
@@ -0,0 +1,89 @@
+import { ReactElement } from "react";
+import { Flex, Text, RadioGroup as RadixRadioGroup } from "@radix-ui/themes";
+import { MarginProps } from "@radix-ui/themes/dist/cjs/props/margin.props";
+import HelperText, { getRadixColor } from "@/components/Radix/HelperText";
+
+export type RadioOptions = {
+  value: string;
+  label?: string | ReactElement;
+  description?: string | ReactElement;
```
jdorn, line 9: I've been trying to avoid using `ReactElement` since I think it encourages bad practices like including tooltips and extra text formatting that goes against the design system.

packages/front-end/components/Radix/RadioGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+  sub?: string;
```
jdorn, line 10: The design system does not support this additional level of text.  I know this is used today in the metric type selector.  We should check with Kari about how we want to handle that.

packages/front-end/components/Radix/RadioGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+              return (
+                <RadixRadioGroup.Item
+                  key={value}
+                  value={value}
+                  disabled={disabled}
+                >
+                  <Text color={disabled ? "gray" : undefined}>
+                    <Flex direction="column" gap="1">
+                      <Text weight="bold">{label || value}</Text>
+                      <Flex direction="column" gap="1">
```
jdorn, line 64: I don't think this nested Flex is doing anything here.

packages/front-end/components/Radix/RadioGroup.tsx
```diff
@@ -0,0 +1,89 @@
...
+                            {sub && (
+                              <Text size="1" weight="light">
+                                {sub}
+                              </Text>
+                            )}
+                          </>
+                        )}
+                        {error && (
+                          <HelperText status={errorLevel}>{error}</HelperText>
+                        )}
```
jdorn, line 77: It looks like in the Figma, the HelperText is only shown when the option is selected.  Also might need to tweak how you're specifying color.  Instead of doing it at the top-level, can we do it at the individual radio item level?

image

packages/front-end/components/Radix/RadioGroup.tsx
lukesonnet: Ok, I agree with this. I was just adding this to eventually accommodate what Romain did in the Power Calculator modal, but I'll remove it here.

packages/front-end/components/Radix/RadioGroup.tsx
lukesonnet: >can we do it at the individual radio item level?

I looked at this, and there isn't a prop to set it that way. I'll try and do it manually.

packages/front-end/components/Radix/RadioGroup.tsx
lukesonnet: Ah, it was before and I forgot to remove it, thnks.

packages/front-end/pages/design-system/index.tsx
```diff
@@ -96,6 +98,56 @@ export default function DesignSystemPage() {
         </Flex>
       </div>
       <div className="appbox p-3">
+        <h3>Radio Group</h3>
```
jdorn, line 101: nitpick: Wrap this in its own `appbox` instead of sharing the one for HelperText

packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
```diff
@@ -24,13 +23,16 @@ export default function PowerCalculationStatsEngineSettingsModal({
     orgSettings.sequentialTestingTuningParameter ||
     DEFAULT_SEQUENTIAL_TESTING_TUNING_PARAMETER;
   const [currentParams, setCurrentParams] = useState(params);
+  const [currentEngine, setCurrentEngine] = useState<StatsEngineWithSequential>(
+    params.type
+  );
```
jdorn, line 28: Instead of two states, which is error prone, you can derive `currentEngine` from `currentParams`.

```ts
const currentEngine = currentParams.type === "bayesian" ? "bayesian" 
  : currentParams.sequentialTesting ? "sequential" 
  : "frequentist";
```

jdorn (approved):
Left some nit code style comments, but looks good overall

packages/front-end/pages/design-system/index.tsx
lukesonnet: was a mistake in resolving the merge conflict, sorry. fixed.

PR #3061, opened 2024-10-02
Block dimension analyses from firing significance events

lukesonnet:
### Features and Changes

Do not fire `experiment.info.significance` or `experiment.warning` events for dimensional analyses.

### Testing

* On main, get an experiment with a stat sig metric
* Delete all snapshots so it will notify
* If I do a dimension analysis, it notifies

* Switch to this branch, delete that old snapshot from mongo
* Run dimension analysis, no notification
* Non-dimension analysis notifies!


romain-growthbook (approved):
Go for it!

PR #3063, opened 2024-10-03
Radix button component

bryce-fitzsimons:
Adds the `Button` wrapper component for Radix.

todo (fast follow):
- [x] Implement loading logic for async click handlers

Create feature button before:
image

After:
image

-----

Experiment list page before:
image

After:
image


packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+  {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onClick?: (e: any) => void;
+    href?: string;
+    color?: Color | ButtonProps["color"];
+    variant?: Variant;
+    size?: Size;
+    disabled?: boolean;
+    loading?: boolean;
+    children: ReactNode;
```
jdorn, line 24: I would restrict this to string only instead of `ReactNode` to discourage deviating from the design system.  We will definitely want to support left/right icons, but I would do those as dedicated props instead.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+import ConditionalWrapper from "@/components/ConditionalWrapper";
+
+type Overwrite<T, NewT> = Omit<T, keyof NewT> & NewT;
+
+export type Color = "primary" | "danger";
+export type Variant = "solid" | "soft" | "outline" | "ghost";
+export type Size = "xs" | "sm" | "md";
+
+export type Props = Overwrite<
+  ButtonProps,
```
jdorn, line 14: I would avoid exposing all of these raw Radix props.  In the other components (e.g. Checkbox), we only expose the MarginProps, not any of the others.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+  {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onClick?: (e: any) => void;
+    href?: string;
+    color?: Color | ButtonProps["color"];
```
jdorn, line 19: Only allow the colors from our design system, not all of the Radix colors.  That might mean we can't easily migrate as many buttons to the new component yet until they are redesigned, but that's fine.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,99 @@
+import { Button as RadixButton, ButtonProps, Text } from "@radix-ui/themes";
+import { CSSProperties, ReactNode } from "react";
+import { Responsive } from "@radix-ui/themes/dist/cjs/props";
+import Link from "next/link";
+import ConditionalWrapper from "@/components/ConditionalWrapper";
+
+type Overwrite<T, NewT> = Omit<T, keyof NewT> & NewT;
+
+export type Color = "primary" | "danger";
```
jdorn, line 9: I think we can keep the Radix color names of `violet` and `red`.  That's what will show up in Figma and it should be pretty intuitive which ones to use.  If we do need to change them later for any reason, it's an easy refactor with Typescript.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+      margin: 0,
+      height: "var(--base-button-height)",
+      paddingTop: 0,
+      paddingBottom: 0,
+      paddingLeft: `var(--space-${parseInt(radixSize as string) + 1})`,
+      paddingRight: `var(--space-${parseInt(radixSize as string) + 1})`,
+    };
+  }
+  if (otherProps.style) {
+    style = Object.assign(style, otherProps.style);
```
jdorn, line 77: We definitely don't want to support custom styles or class names being passed in.  That's just asking for trouble.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,99 @@
...
+export type Color = "primary" | "danger";
+export type Variant = "solid" | "soft" | "outline" | "ghost";
+export type Size = "xs" | "sm" | "md";
+
+export type Props = Overwrite<
+  ButtonProps,
+  {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onClick?: (e: any) => void;
+    href?: string;
```
jdorn, line 18: Wonder if we should split these into a `ButtonLink` component (with required `href` prop) and a `Button` component (with required `onClick` prop).  Possibly even a 3rd one for `SubmitButton` (with neither `href` or `onClick`).  They can share 99% of the code under the hood, but splitting them would help avoid mistakes.

Also, for the `onClick` prop, do we need to pass the event in as an argument?  In the other Button prop, it automatically calls `e.preventDefault()`, which I think we should do as well.  Then, there's no need to have the event.

bryce-fitzsimons: Re: ButtonLink and SubmitButton, I'm on the fence about this. As a dev, I think I would expect a button to behave at least largely like a native HTML button, and wouldn't think that I'd need to reference 3 different things. We'll likely have some combinatory issues as well if we decide to do <TextButton> or any other <StyledButton> supersets.

onClick: Yeah I guess we almost never need the underlying event. It does make me think though that perhaps we need to fully implement the async + auto-loading states, like in the old Button.

bryce-fitzsimons: Added the async loading stuff. lmk what we want to do about links and other button variants. I'm ok splitting them if you prefer.

bryce-fitzsimons: Decided to implement a LinkButton. Thinking we hold off on SubmitButton for now because its so similar to Button.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -0,0 +1,83 @@
...
+              }
+              setLoading(false);
+            }
+          : undefined
+      }
+      color={color}
+      variant={variant}
+      size={getRadixSize(size)}
+      disabled={disabled}
+      loading={loading}
```
jdorn, line 76: We should pass `type="button"` in here.  Otherwise, if we use this button inside of a modal or form, clicking it will submit automatically.

bryce-fitzsimons: ok, defaulted to "button"

packages/front-end/components/Radix/LinkButton.tsx
```diff
@@ -0,0 +1,18 @@
+import Link from "next/link";
+import Button, { Props as ButtonProps } from "@/components/Radix/Button";
+
+export type Color = "violet" | "red";
+export type Variant = "solid" | "soft" | "outline" | "ghost";
+export type Size = "xs" | "sm" | "md" | "lg";
```
jdorn, line 6: Remove these exports

packages/front-end/pages/settings/index.tsx
```diff
@@ -398,10 +398,8 @@ const GeneralSettingsPage = (): React.ReactElement => {
               </TempMessage>
             )}
           </div>
-          <div>
+          <div style={{ marginRight: "4rem" }}>
             <Button
-              style={{ marginRight: "4rem" }}
-              color={"primary"}
```
jdorn, line 404: The old button would show errors inline.  If we want to switch this to the new button, we should use the `setError` prop and manage state on this page.  Otherwise, errors while submitting will be swallowed.

jdorn (approved):
Approving. Left 2 minor comments.

PR #3065, opened 2024-10-03
Add snapshot type and configure notifications

lukesonnet:
### Features and Changes

Adds a snapshot type to each snapshot:
* "standard" <- default analysis settings for that snapshot
* "exploratory" <- some custom exploratory (old phase or dimension slice) analysis

Adds `triggeredBy` field to designate how the snapshot was created:
* "scheduled" <- triggered by agenda job
* "manual" <- triggered by user action in the UI (refresh, importing existing experiment)

Also tweaks significance emails to only send for "scheduled", "standard" updates.

Webhooks still fire for all "standard" updates (whether scheduled or manual).

Also stops checking significance when new analyses are added to an existing snapshot.

### Testing

TESTS BELOW ARE OLD, BUT WERE REPEATED WITH NEW FIELDS ABOVE

Manual refresh sets type manul
Screenshot 2024-10-03 at 10 20 57 AM

Dimension refresh sets type adhoc
Screenshot 2024-10-03 at 10 21 13 AM

Adding a phase1 and then going back to phase0 and refreshing sets type adhoc
Screenshot 2024-10-03 at 10 21 46 AM

Scheduled update sets type standard
Screenshot 2024-10-03 at 10 27 27 AM

confirmed events don't fire for ad-hoc analyses

still can't confirm emails send, waiting for auto-update

### DIFF
--- a/packages/back-end/src/services/experimentNotifications.ts
+++ b/packages/back-end/src/services/experimentNotifications.ts
@@ -390,8 +390,8 @@ export const notifySignificance = async ({
 
   if (!experimentChanges.length) return;
 
-  // send email if enabled and the snapshot is standard
-  if (isEmailEnabled() && snapshot.type === "standard")
+  // send email if enabled and the snapshot is scheduled standard analysis
+  if (isEmailEnabled() && snapshot.triggeredBy === "schedule" && snapshot.type === "standard")
     await sendSignificanceEmail(experiment, experimentChanges);
 
   await Promise.all(
@@ -418,10 +418,12 @@ export const notifyExperimentChange = async ({
   const experiment = await getExperimentById(context, snapshot.experiment);
   if (!experiment) throw new Error("Error while fetching experiment!");
 
-  // do not fire significance or error events for ad-hoc analyses
-  if (snapshot.type === "ad-hoc") return;
+  // do not fire significance or error events for exploratory analyses
+  if (snapshot.type === "exploratory") return;
   // do not fire significance events for old snapshots that have no type
   if (snapshot.type === undefined) return;
+  // do not fire for snapshots where statistics are manually entered in the UI
+  if (snapshot.manual) return;
 
   await notifySignificance({ context, experiment, snapshot });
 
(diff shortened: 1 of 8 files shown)


packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -418,8 +418,8 @@ export const notifyExperimentChange = async ({
   const experiment = await getExperimentById(context, snapshot.experiment);
   if (!experiment) throw new Error("Error while fetching experiment!");
 
-  // do not fire significance or error events for dimension analyses
-  if (snapshot.dimension) return;
+  // do not fire significance or error events for ad-hoc analyses
+  if ((snapshot.type ?? "manual") === "ad-hoc") return;
```
jdorn: If `snapshot.type` is undefined, it means it was stored in Mongo before this PR and may have a dimension.  It's safer to just make this return if it's undefined.

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -418,8 +418,8 @@ export const notifyExperimentChange = async ({
   const experiment = await getExperimentById(context, snapshot.experiment);
   if (!experiment) throw new Error("Error while fetching experiment!");
 
-  // do not fire significance or error events for dimension analyses
-  if (snapshot.dimension) return;
+  // do not fire significance or error events for ad-hoc analyses
+  if (snapshot.type === "ad-hoc") return;
```
jdorn, line 422: Also return if `snapshot.type` is not set (just in case this code runs for ad ad-hoc snapshot that is in Mongo from before the PR merges)

jdorn (approved):

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: done ty

romain-growthbook (approved):
This looks great. Abstracting away the logic is great, thanks!

bryce-fitzsimons (approved):
based on our convo, 2 fields (as you have it now) is the preferred solution. Better handles the nuances how snapshots are generated, updated, and scheduled, and prevents notification flooding during manual exploration.

PR #3068, opened 2024-10-04
Update mysql2 dependency

jdorn:
### Features and Changes

Fixes multiple vulnerabilities in the mysql2 dependency.  No known exploits, just staying on top of CVE reports.

According to release notes, this update should be 100% backwards compatible.


jdorn:
Fixes #2575
Fixes #2419
Fixes #2390
Fixes #2364

PR #3079, opened 2024-10-07
Componentize Stepper

msamper:
### Features and Changes
Adjusts styling for all uses of the stepper within the `PagedModal` component and creates a new separate `Stepper` component that can be used outside of `PagedModal`

### Screenshots

Screenshot 2024-10-08 at 1 57 14 PM
Screenshot 2024-10-08 at 1 57 22 PM


packages/front-end/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  setStep,
+  steps,
+  skipped,
+  setError,
+  validateSteps,
+}: Props) {
+  return (
+    <nav
+      className={
+        "nav mb-4 justify-content-start nav-default paged-modal-default"
```
gazzdingo, line 30: can we convert the spacing to radix spacing most of this is using bootstrap

packages/front-end/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,79 @@
...
+      }
+    >
+      {steps.map(({ label, enabled }, i) => {
+        return (
+          <div
+            className={clsx(
+              "step d-flex align-items-center justify-content-between",
```
gazzdingo, line 37: same here

packages/front-end/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,79 @@
...
+                completed: i < step && !skipped?.has(i),
+                disabled: !enabled,
+              }
+            )}
+            key={i}
+          >
+            <a
+              key={i}
+              role="button"
+              className={clsx("nav-link")}
```
gazzdingo, line 49: dont think we need clsx here. you should just be able to do `classname="nav-link"`

packages/front-end/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,79 @@
...
+                  setError(e.message);
+                }
+              }}
+            >
+              <span className="step-number rounded-circle">
+                {i < step ? (
+                  skipped?.has(i) ? (
+                    <PiCircleDashed />
+                  ) : (
+                    <MdCheck />
```
gazzdingo, line 66: we are moving away from Md to Pi so we should update this to the PiCheck

packages/front-end/components/Stepper/Stepper.tsx
msamper: Yeah I can, but this isn't a radix component so I don't think I can do it on the component itself. Is what you had in mind using the CSS variables they have available within the CSS for this?

packages/front-end/components/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,80 @@
...
+      as="div"
+      justify={"start"}
+      className={"nav nav-default paged-modal-default"}
+      mb="4"
+    >
+      {steps.map(({ label, enabled }, i) => {
+        return (
+          <Flex
+            as="div"
+            align={"center"}
```
gazzdingo, line 38: Super nit: are we able to make these just `"center"` instead of `{"center"}` there is a few places this is happening

gazzdingo (approved):
LGTM

PR #3080, opened 2024-10-07
Fix result group for dimension results

lukesonnet:
### Features and Changes

Corrected p-values are not showing in dimension results in the overall results and the tooltip for the pvalue column is broken. This fixes that problem.

How did this happen? Changing the TableRow type as part of creating primary and secondary metrics created a silent error due to the use of `... as ExperimentTableRow[]` which was failing to warn about missing fields.

I fixed this `as` call and resolving the actual TableRow data resolved the issue.

### Testing

See screenshots

### Screenshots

Before
Screenshot 2024-10-07 at 2 46 10 PM

Screenshot 2024-10-07 at 2 46 01 PM

After
Screenshot 2024-10-07 at 3 03 23 PM
Screenshot 2024-10-07 at 3 03 20 PM

### DIFF
--- a/packages/front-end/components/Experiment/BreakDownResults.tsx
+++ b/packages/front-end/components/Experiment/BreakDownResults.tsx
@@ -162,6 +162,11 @@ const BreakDownResults: FC<{
             (s) => s.metric === metricId
           );
         }
+        const resultGroup = getMetricResultGroup(
+          metricId,
+          goalMetrics,
+          secondaryMetrics
+        );
 
         const rows: ExperimentTableRow[] = results.map((d) => ({
           label: d.name,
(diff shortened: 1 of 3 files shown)


packages/front-end/components/Experiment/BreakDownResults.tsx
```diff
@@ -140,27 +152,37 @@ const BreakDownResults: FC<{
...
+          metric: newMetric,
+          variations: d.variations.map((variation) => {
+            return variation.metrics[metricId];
+          }),
+          metricSnapshotSettings,
+          resultGroup: getMetricResultGroup(
+            metricId,
+            goalMetrics,
+            secondaryMetrics
+          ),
```
jdorn, line 177: Nit: This can be done outside of the map.  Also, if we do that, we can use the result of this for `isGuardrail` below instead of duplicating the logic.

jdorn (approved):

PR #3081, opened 2024-10-07
Feature/radix slider

madhuchavva:
### Features and Changes

- Adding Radix `Slider` to percentage rollout rule and new experiment rule. Did not find a reason to create another wrapper on top of Radix's Slider component. 

- Percentage Rollout
  - Before:  
  image
  
  - After: 
  image

-New Experiement Rule
- Before
image

- After
image

- added a utility function to convert percent to decimal for `number`

### DIFF
--- a/packages/front-end/components/Features/FeatureVariationsInput.tsx
+++ b/packages/front-end/components/Features/FeatureVariationsInput.tsx
@@ -1,4 +1,5 @@
 import { FeatureInterface, FeatureValueType } from "back-end/types/feature";
+import { Slider } from "@radix-ui/themes";
 import React, { useState } from "react";
 import { FaInfoCircle } from "react-icons/fa";
 import {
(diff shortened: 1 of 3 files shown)


packages/front-end/components/Features/FeatureVariationsInput.tsx
```diff
@@ -90,20 +92,20 @@ export default function FeatureVariationsInput({
             </label>
             <div className="row align-items-center pb-3">
               <div className="col">
-                <input
-                  value={isNaN(coverage) ? 0 : decimalToPercent(coverage)}
-                  onChange={(e) => {
-                    let decimal = percentToDecimal(e.target.value);
+                <Slider
+                  name="RadixSlider"
+                  defaultValue={[50]}
```
jdorn, line 97: Specify either `defaultValue` or `value`, but  not both.  In this case, we want to use `value` since it's a controlled component.  Also, you do not need to specify a `name`.

Same comment for the other file.

madhuchavva: Sounds good!

jdorn (approved):

PR #3100, opened 2024-10-11
Add fact metrics to North Star display

lukesonnet:
### Features and Changes

Adds fact metrics as northstar metrics and allows updates of northstar metrics on dashboard page

Also allows updating time series right on dasboard.

Key change:
* New endpoint to just collect all data for both Metric (Legacy Metric Analysis) and Fact Metric (Metric Analysis) as a "Metric Analysis" like object to power the dashboard and make it more robust to future changes.
* Also moved where we get experiments for a metric to this endpoint to get all data in one call (and that other experiments look up was no longer being used on the [mid] page now that we have MetricExperiments)

https://www.loom.com/share/c35146871e734b87a789f1a8b7d605db

### Testing

Can add new fact metric, update all northstar metrics, and things look as they should.


packages/front-end/components/HomePage/NorthStarMetricDisplay.tsx
```diff
@@ -24,64 +37,166 @@ const NorthStarMetricDisplay = ({
...
+      experiments: ExperimentInterfaceStringDates[];
+      analysis: MetricAnalysisInterface | null;
+      metric: MetricInterface | null;
+    };
+  }>(`/metrics/${metricId}/northstar`);
 
-  // @todo: get the metric period in days from the 'window'.
+  if (!metric || !data) {
+    return <LoadingOverlay />;
+  }
```
jdorn, line 63: You typically want to check for `error` first before the loading state.  When there's an error, data is often empty, so with how it is here it would show a loading overlay forever.

packages/front-end/components/MetricAnalysis/MetricAnalysis.tsx
```diff
@@ -132,11 +134,17 @@ function getLookbackSelected(lookbackDays: number): string {
     : `custom`;
 }
 
-function getDesiredSettings(
-  id: string,
-  values: MetricAnalysisFormFields,
-  endOfToday: Date
-): CreateMetricAnalysisProps {
+export function getMetricAnalysisProps({
```
jdorn, line 137: Exporting multiple functions from a component file causes HMR (hot module reloading) to stop working on dev.  You might have noticed it causing the entire page to refresh on every change.  Not a big deal, but better to move this to a separate util/services file instead and import it both places it is used.

jdorn (approved):

packages/front-end/components/HomePage/NorthStarMetricDisplay.tsx
lukesonnet: done ty

packages/front-end/components/MetricAnalysis/MetricAnalysis.tsx
lukesonnet: Thanks, moved it to a util file in the same metric analysis folder.

PR #3112, opened 2024-10-15
Prevent empty experiment names

romain-growthbook:
This PR prevents empty experiment names:

* Error when saving experiment model with empty name, after trimming
* Validate non-empty experiment name in zod validators, after trimming.
* Prevent submitting empty experiment names in front end create and edit modals, also after trimming.


packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -358,14 +358,21 @@ export async function getSampleExperiment(
...
+  data: rawData,
   context,
 }: {
   data: Partial<ExperimentInterface>;
   context: ReqContext | ApiReqContext;
 }): Promise<ExperimentInterface> {
+  const data: Partial<ExperimentInterface> = {
+    ...rawData,
+    name: rawData.name?.trim(),
+  };
```
jdorn, line 370: Let's get rid of all of the trim stuff from the back end. The front end can do the trim before submitting to the backend.

packages/front-end/components/Experiment/EditExperimentNameForm.tsx
```diff
@@ -24,6 +24,7 @@ const EditExperimentNameForm: FC<{
       open={true}
       close={cancel}
       size="lg"
+      ctaEnabled={!!form.watch("name").trim()}
```
jdorn, line 27: I don't think we should disable submit buttons because it's hard for users to know why. Better to let them try and submit and throw an error

romain-growthbook: ok

packages/back-end/src/models/ExperimentModel.ts
romain-growthbook: How about API requests? Having the back-end double down on data cleanup and validation is always a good idea as last resource.

romain-growthbook:
I've remove the cta disabled state and confirmed that errors are surfaced to the user. Also added an error when updating with an empty name:

Screenshot 2024-10-16 at 9 40 42 AM

Screenshot 2024-10-16 at 9 40 21 AM

jdorn (approved):

PR #3122, opened 2024-10-16
Data warehouse enrichment endpoint

Kevin-Chant:
### Features and Changes

Creates a new API endpoint for serving information about all configured SdkConnections. This is only accessible by superAdmins as it provides cross-org information, and will be used to provide a mapping from SdkConnection ID to organization & datasource IDs for the data ingestion service to consume.

### Testing

1. Create a Personal Access Token for your superadmin user
2. Run the following script, replacing the placeholder with your PAT
```bash
#!/bin/bash
curl \
  --header "Authorization: Bearer <Personal_Access_Token>" \
  --request GET \
  http://localhost:3100/api/v1/ingestion/data-enrichment
```


packages/back-end/src/api/openapi/openapi.yaml
```diff
@@ -108,6 +108,9 @@ tags:
   - name: code-references
     x-displayName: Code References
     description: Intended for use with our code reference CI utility, [`gb-find-code-refs`](https://github.com/growthbook/gb-find-code-refs).
+  - name: ingestion
+    x-displayName: Event Tracking
+    description: For supporting tracking and enriching event data
```
romain-growthbook, line 113: I don't think that you need this to be part of the documented public API.

packages/back-end/src/api/openapi/openapi.yaml
```diff
@@ -286,6 +289,9 @@ paths:
   /code-refs:
     post:
       $ref: "./paths/postCodeRefs.yaml"
+  /ingestion/data-enrichment:
+    get:
+      $ref: "./paths/getDataEnrichment.yaml"
```
romain-growthbook, line 294: Same

packages/back-end/src/api/openapi/paths/getDataEnrichment.yaml
```diff
@@ -0,0 +1,20 @@
+tags:
```
romain-growthbook, line 1: Same

packages/back-end/src/api/openapi/openapi.yaml
Kevin-Chant: Yeah I was thinking about removing this & the tag on the new endpoint but wasn't sure what our exact stance is

romain-growthbook (commented):
I don't think that this should be part of the public API for now. You can define functions outside of the `/api` path. See for instance `src/controllers/slack.ts` for a simple example.

packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -0,0 +1,53 @@
...
+);
+
+function sdkInfo(conn: SDKConnectionInterface) {
+  return {
+    organization: conn.organization,
+    client_key: conn.id,
+    // TODO: pull datasource
+    datasource: "",
+  };
+}
```
romain-growthbook, line 43: nit: I know JS allows functions to be anywhere in the modules but I think keeping functions declarations before their call promotes much more readable code..

packages/back-end/src/controllers/ingestion.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (
+    !req.headers["x-gb-shared-secret"] ||
+    req.headers["x-gb-shared-secret"] !==
+      process.env.DATA_ENRICHMENT_SHARED_SECRET
+  ) {
+    return res.status(401).json({
+      message: "Invalid Secret Key",
+    });
+  }
+  const sdkConnections = await findAllSDKConnectionsAcrossAllOrgs();
```
romain-growthbook, line 32: Shall this be cached?

romain-growthbook (approved):
This looks good to me. We should make sure to validate the approach with another eng before merging.

packages/back-end/src/controllers/ingestion.ts
```diff
@@ -0,0 +1,56 @@
...
+    client_key: conn.id,
+    // TODO: pull datasource
+    datasource: "",
+  };
+}
+
+// Refresh in-mem cache every minute
+const REFRESH_INTERVAL = 60_000;
+let sdkData: Record<string, SdkInfo> = {};
+let lastUpdate = Date.now() - REFRESH_INTERVAL;
```
Kevin-Chant, line 24: There's also a cache on the ingestion endpoint which will ping this every time there's a cache miss but otherwise won't update its data

PR #3126, opened 2024-10-16
Athena enable query reuse

### LINKED ISSUE #2994: [Feature] Enable Athena Query Result Reuse
Hi, currently the Athena configuration doesn't allow the reuse of query results, this feature will help to reduce the costs associated with Athena query execution and the query execution times. 

[GrowthBook code](https://github.com/growthbook/growthbook/blob/2a5de0db34d2673ef3bc23bce8eb0e750da0cb44/packages/back-end/src/services/athena.ts#L76C45-L76C64)

[Athena Documentation](https://docs.aws.amazon.com/athena/latest/APIReference/API_StartQueryExecution.html#athena-StartQueryExecution-request-ResultReuseConfiguration)

Kevin-Chant:
### Features and Changes

Adds an option to Athena connection settings for enabling re-use of query results up to a certain age limit.
 
- Closes #2994 

### Testing

This hasn't been manually tested as we don't have access to an Athena DB to test against in dev. The code should prevent any changes for datasources that don't have the new setting enabled so it should be safe to ship as-is and iterate if needed

### Screenshots

image

### DIFF
--- a/packages/back-end/src/services/athena.ts
+++ b/packages/back-end/src/services/athena.ts
@@ -77,13 +77,14 @@ export async function runAthenaQuery(
     (parseInt(process.env.ATHENA_RETRY_WAIT_TIME || "60") || 60) * 1000;
 
   const resultReuseSettings =
-    typeof conn.resultReuseMaxAgeInMinutes === "undefined"
+    typeof conn.resultReuseMaxAgeInMinutes === "undefined" ||
+    conn.resultReuseMaxAgeInMinutes === "0"
       ? {
           Enabled: false,
         }
       : {
           Enabled: true,
-          MaxAgeInMinutes: conn.resultReuseMaxAgeInMinutes,
+          MaxAgeInMinutes: parseInt(conn.resultReuseMaxAgeInMinutes),
         };
 
   const { QueryExecutionId } = await athena.startQueryExecution({
--- a/packages/back-end/src/services/athena.ts
+++ b/packages/back-end/src/services/athena.ts
@@ -76,6 +76,16 @@ export async function runAthenaQuery(
   const retryWaitTime =
     (parseInt(process.env.ATHENA_RETRY_WAIT_TIME || "60") || 60) * 1000;
 
+  const resultReuseSettings =
+    typeof conn.resultReuseMaxAgeInMinutes === "undefined"
+      ? {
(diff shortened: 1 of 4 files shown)


packages/front-end/components/Settings/SharedConnectionSettings.tsx
```diff
@@ -21,7 +21,7 @@ export default function SharedConnectionSettings({
             type="number"
             label={
               <>
-                Maximum Concurrent Queries (Optional){" "}
+                Maximum Concurrent Queries (optional){" "}
```
Kevin-Chant, line 24: Unrelated change; I just noticed that the capitalization was inconsistent

packages/back-end/src/services/athena.ts
```diff
@@ -76,6 +76,17 @@ export async function runAthenaQuery(
...
+  const resultReuseSettings =
+    typeof conn.resultReuseMaxAgeInMinutes === "undefined" ||
+    conn.resultReuseMaxAgeInMinutes === "0"
+      ? {
+          Enabled: false,
+        }
+      : {
+          Enabled: true,
+          MaxAgeInMinutes: parseInt(conn.resultReuseMaxAgeInMinutes),
+        };
```
jdorn, line 88: This can be made a little less risky with some more typeguards.  Like the value might be an empty string or `parseInt` might fail.

Also, it's less risky if instead of always passing in `ResultReuseConfiguration`, we only pass it in when it's enabled.  That way, the baseline code remains completely unchanged for existing users and there's no risk of it breaking unintentionally.

fabianmarinog: @jdorn @Kevin-Chant Do you have any updates on the progress of this feature?

Kevin-Chant: Hi @fabianmarinog, right now the team is focused on finishing features for our 3.4 release, but after the code freeze (~25th) I planned to get a testing environment set up so I could verify these changes and get it merged. So no updates yet but hopefully it should be relatively quick once I can test that it works the way I expect it to :+1:

fabianmarinog: @Kevin-Chant cool. let us know when you start again to progress on it.

lukesonnet (approved):

PR #3131, opened 2024-10-17
Ji/create clickhouse datasource

tzjames:
### Features and Changes

Creates an inbuilt Growthbook Clickhouse option

### Testing

- Create new datasource
- Click on the create in built datasource button
- Click Finish
- Edit the anonymous_id assignment query on the datasource page 
- Click Test Query
- See 4 results (`test_enriched_events` needs to have events that match the org in your database)

### Screenshots

Screenshot 2024-10-17 at 6 50 22 PM
when the sample database is also showing:
Screenshot 2024-10-17 at 6 50 01 PM


packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  for (let i = 3; i < length; i++) {
+    password += allChars[Math.floor(Math.random() * allChars.length)];
+  }
+
+  return password
+    .split("")
+    .sort(() => 0.5 - Math.random())
+    .join("");
+}
```
toots, line 33: We should consider using a third party here like this: https://www.npmjs.com/package/generate-password. This package uses the node crypto module to generate random bytes.

packages/front-end/components/Settings/NewDataSourceForm.tsx
```diff
@@ -410,6 +410,26 @@ const NewDataSourceForm: FC<{
...
+      newDatasourceForm: true,
+    });
+
+    setCreatedDatasource(res.datasource);
+    createResources(res.datasource);
+    setStep("done");
+  };
+
+  // TODO: Make this a feature flag
+  const showInbuiltDatasource = true;
```
toots, line 431: This should be made conditional.

toots (commented):
This looks straight forward to me.

packages/back-end/src/services/clickhouse.ts
romain-growthbook: We should consider using a third party here like this: https://www.npmjs.com/package/generate-password. This package uses the node crypto module to generate random bytes.

packages/front-end/components/Settings/NewDataSourceForm.tsx
romain-growthbook: This should be made conditional.

romain-growthbook (commented):
This looks straight forward to me.

packages/back-end/src/app.ts
```diff
@@ -657,6 +657,13 @@ app.post(
   datasourcesController.fetchBigQueryDatasets
 );
 
+if (IS_CLOUD) {
```
romain-growthbook, line 660: I think long term we should find a way to be more dev-friendly with those.

packages/back-end/types/datasource.d.ts
```diff
@@ -12,6 +12,7 @@ import { MetricType } from "./metric";
 import { MssqlConnectionParams } from "./integrations/mssql";
 
 export type DataSourceType =
+  | "growthbook_clickhouse"
```
romain-growthbook, line 15: Aren't we here just overlooking the fact that this should be database type clickhouse, schema growthbook? Not blocking for now.

packages/front-end/pages/datasources/[did].tsx
```diff
@@ -61,7 +61,10 @@ const DataSourcePage: FC = () => {
     (d && permissionsUtil.canDeleteDataSource(d) && !hasFileConfig()) || false;
 
   const canUpdateConnectionParams =
-    (d && permissionsUtil.canUpdateDataSourceParams(d) && !hasFileConfig()) ||
+    (d &&
+      permissionsUtil.canUpdateDataSourceParams(d) &&
+      !hasFileConfig() &&
+      d.type != "growthbook_clickhouse") ||
```
Comment (romain-growthbook) [inline packages/front-end/pages/datasources/[did].tsx:67]: These should be extracted and encapsulated.

romain-growthbook (approved):
Nothing blocking for a merge into the feature branch. Couple of things we might want to clean up before eventually merging to `main`.

PR #3133, opened 2024-10-17
Redesign EnvironmentSelect

msamper:
### Features and Changes

Reworks the EnvironmentSelect to use radix checkboxes instead of toggles in a Grid, 4 column format that lists environments in order from left to right. Adds a select all button

### Dependencies

Update radix-ui/themes to 3.1.3 #3140 

### Testing

Add several environments to your org and test that overflow works correctly
Make sure new EnvironmentSelect still correctly persists selected values

### Screenshots

Before
Screenshot 2024-11-01 at 1 05 04 PM

After
Screenshot 2024-11-01 at 1 06 30 PM


packages/front-end/components/Features/FeatureModal/EnvironmentSelect.tsx
```diff
@@ -14,32 +21,45 @@ const EnvironmentSelect: FC<{
   const { project } = useDefinitions();
   return (
     <div className="form-group">
-      <label>Enabled Environments</label>
-      <div className="appbox px-3 pt-3 pb-2 bg-light">
-        <div className="row">
+      <Container p="5" style={{ background: "#FAF8FF" }}>
```
msamper, line 24: Create new theme variable for purple color background

msamper:
For some reason I can't get the indeterminate checkbox to render. This is what we use in the design as a Select All button. Given that it's meant to be used as an indeterminate checkbox we may want to consider using something else anyway and I left a comment on the design about it

packages/front-end/components/Features/FeatureModal/EnvironmentSelect.tsx
```diff
@@ -14,32 +21,45 @@ const EnvironmentSelect: FC<{
...
+          >
+            <Flex gap="2" mb="2">
+              <RadixCheckbox checked={"indeterminate"} color={"violet"} />
+              <Flex direction="column" gap="1">
+                <Text weight="bold" className="main-text">
+                  Select All
+                </Text>
+              </Flex>
+            </Flex>
+          </Text>
```
msamper, line 47: Used the radix component here directly for now because I wasn't sure if we wanted to support the indeterminate state on our Checkbox component. If we do, I may just need to make `setValue` optional

msamper:
> For some reason I can't get the indeterminate checkbox to render. This is what we use in the design as a Select All button. Given that it's meant to be used as an indeterminate checkbox we may want to consider using something else anyway and I left a comment on the design about it

We're just on an old version of radix that doesn't support indeterminate checkboxes. Once the update PR merges this will be unblocked.

packages/front-end/components/Features/FeatureModal/EnvironmentSelect.tsx
```diff
@@ -12,34 +13,66 @@ const EnvironmentSelect: FC<{
...
+              environments.forEach((env) => {
+                setValue(env, v === true);
+              })
+            }
+            label="Select All"
+            weight="bold"
+            mb="5"
+          />
+        </div>
+        <Flex
```
msamper, line 54: change direction to row

msamper: Achieved with Grid

ahdriel:
The layout breaks a bit when the environment name is long without hyphens -- should we add `word-break: break-all` to this and add a margin to the right of the text to ensure it won't ever touch the following checkbox?

Screenshot 2024-10-31 at 2 11 04 PM

packages/front-end/styles/theme-config.css
```diff
@@ -14,4 +14,9 @@
 .light,
 .light-theme {
   --color-page-background: #f5f7fa;
+  --color-background: #faf8ff;
+}
+.dark,
+.dark-theme {
+  --color-background: #10172e;
```
ahdriel, line 21: question: Do we need to define these new values or should we use `--color-panel-solid` which is provided by Radix?

packages/front-end/components/Features/FeatureModal/EnvironmentSelect.tsx
```diff
@@ -12,34 +13,66 @@ const EnvironmentSelect: FC<{
...
+          <Checkbox
+            value={
+              selectAllChecked
+                ? true
+                : selectAllIndeterminate
+                ? "indeterminate"
+                : false
+            }
+            setValue={(v) =>
+              environments.forEach((env) => {
```
ahdriel, line 45: issue: If I have environments that I don't have permission to create feature in, this toggle still updates it in the UI. I have not tested how the server would handle it though.

ahdriel (commented):
Awesome work! It looks so much better 🎉 

additional question is: do we plan on bringing this to the Feature page as well (not just the creation modal)?

packages/front-end/styles/theme-config.css
msamper: I think we need to because these are custom colors we came up with for our design system. I’ll double check that they’re not the same

msamper:
> Awesome work! It looks so much better 🎉
> 
> additional question is: do we plan on bringing this to the Feature page as well (not just the creation modal)?

Thanks! I think for the feature page we will retain toggles since those act as kill switches. We’ll still need to migrate them over to radix switches though and likely do a bit of redesign

packages/front-end/components/Features/FeatureModal/EnvironmentSelect.tsx
msamper: Let me know if this latest change fixes that. Permissions are not working correctly for my test user for some reason so I can't check myself

packages/front-end/styles/theme-config.css
msamper: Was able to fix my test user and seems to be working

packages/front-end/components/Features/FeatureModal/EnvironmentSelect.tsx
msamper: Was able to fix my test user and seems to be working

packages/front-end/styles/theme-config.css
msamper: Yeah they're different. The `--color-panel-solid` seems to be white in light mode and a lighter black in dark mode whereas we want this to be light violet in light mode and a dark indigo in dark mode

ahdriel (approved):
:shipit:

PR #3135, opened 2024-10-17
hide data with 0 users from the date graph for bandits

bryce-fitzsimons:
Bandits date graph:
- hide data with 0 users
- add user counts to the tooltip
- random cleanup


lukesonnet (changes requested):
I think we also need to remove the following case from the data points of the cumulative variation means, but not the rug plot with the update dates:

* If all cis and means are exactly 0
* or perhaps more safely, if error has some text (although in the future we could have all 0s with a message rather than a technical "error", but maybe that's semantics)

lukesonnet (approved):
Will deal with above issues in follow-up. For now this is a step in the right direction.

PR #3139, opened 2024-10-18
Initial fact table for clickhouse

romain-growthbook:
This PR adds the basic fact table for clickhouse


packages/back-end/src/services/clickhouse.ts
```diff
@@ -49,7 +49,7 @@ export async function createClickhouseUser(
   });
 
   const orgId = context.org.id;
-  const user = orgId;
+  const user = orgId + "ter";
```
tzjames: What does ter stand for?  I could see maybe prefixing it with "user_", but it would be easier to copy and paste if we just left it as orgId.

packages/back-end/src/integrations/ClickHouse.ts
```diff
@@ -102,4 +102,7 @@ export default class ClickHouse extends SqlIntegration {
       );
     return `table_schema IN ('${this.params.database}')`;
   }
+  isAutoGeneratingFactTablesSupported() {
+    return true;
```
tzjames: There are other Clickhouse where their schema's are not going to match up, right?  So we want this to only be true for the growthbook-clickhouse datasource types.

packages/front-end/services/datasources.ts
```diff
@@ -979,6 +1036,12 @@ export function getInitialDatasourceResources({
       userIdTypes: datasource.settings?.userIdTypes || [],
     });
 
+  if (datasource.type === "clickhouse")
```
tzjames: The type should be "growthbook-clickhouse" as other users using clickhouse are not going to have the same schema.

packages/front-end/services/datasources.ts
```diff
@@ -979,6 +1036,12 @@ export function getInitialDatasourceResources({
...
+    return getClickHouseInitialDatasourceResources({
+      table: "test_enriched_events",
```
tzjames, line 1076: I made this into an env var now `CLICKHOUSE_MAIN_TABLE` so we can easily swap it over to the real deal.

packages/back-end/src/integrations/ClickHouse.ts
```diff
@@ -102,4 +102,7 @@ export default class ClickHouse extends SqlIntegration {
...
+  }
```
tzjames: There are other Clickhouse where their schema's are not going to match up, right?  So we want this to only be true for the growthbook-clickhouse datasource types.  I don't think we pass the type into the integration, but theoretically you could determine it if `this.params.host` matches up with `CLICKHOUSE_HOST`.  The env var right now has the protocol and the port included which originally seemed more useful, but in one other place we break it apart, so we could consider making the port as a separate env var.

Alternatively we could do a secondary check elsewhere that is using this to make sure that the datasource.type is `growthbook-clickhouse` but then we should at least comment it here because the name of the function would be misleading.

packages/back-end/src/services/clickhouse.ts
tzjames: What does ter stand for? I could see maybe prefixing it with "user_", but it would be easier to copy and paste if we just left it as orgId. That might be useful in debugging or looking into an issue for a specific org.

tzjames (changes requested):
I'm not sure about whether `getClickhouseInitialDatasourceResources` is the right thing as I haven't checked out the fact metric code much yet, but I added a few comments about other things.

packages/front-end/services/datasources.ts
romain-growthbook: I'm gonna swap it but, since this is the front end, this would mean we have to maintain this env var in both env which is annoying.

romain-growthbook:
I just tested this locally and it is working. Gonna merge it to make sure that it is available for demo.

packages/back-end/src/services/clickhouse.ts
romain-growthbook: That was an old diff.

packages/back-end/src/integrations/ClickHouse.ts
romain-growthbook: This is also an old diff.

packages/front-end/services/datasources.ts
romain-growthbook: Also an old diff!

romain-growthbook:
Confirmed with @tzjames, merging!

PR #3141, opened 2024-10-18
Fix useFeaturesList changing object pointers every rerender

Kevin-Chant:
### Features and Changes

On Firefox there's been a general error "the operation is insecure" on the main `/features` page for a few weeks now.
The reason for this is we have a `useMemo` in `services/search.tsx` which does a shallow replace on the router's query params whenever an object changes. That object in turn is a `useMemo` based on the items being searched over, and those items (in the case of `/features`) are the `FeatureInterface[]` returned by `useFeaturesList`.

The problem is that each time the hook is called (due to re-rendering for other reasons) the default empty array that's returned is a new object, so this was causing us to change the query params hundreds of times, which Firefox blocked as a security issue.

### Testing

Using Firefox, load http://localhost:3000/features on `main` and on this branch and observe the difference.
Optionally, add a console log to the `if (updateSearchQueryOnChange)` clause in `services/search.tsx` to see the frequency at which we call `router.replace`

#### Before

image

#### After

image

### DIFF
--- a/packages/front-end/services/features.ts
+++ b/packages/front-end/services/features.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from "react";
+import { useEffect, useMemo, useState } from "react";
 import {
   Environment,
   NamespaceUsage,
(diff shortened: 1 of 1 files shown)


packages/front-end/services/features.ts
```diff
@@ -204,13 +209,21 @@ export function useFeaturesList(withProject = true, includeArchived = false) {
     hasArchived: boolean;
   }>(url);
 
+  useEffect(() => {
+    if (data) {
+      setFeatures(data.features);
+      setExperiments(data.linkedExperiments);
+      setHasArchived(data.hasArchived);
+    }
+  }, [data]);
```
Kevin-Chant, line 218: I don't think I need to have an `else` to reset the fields back to `[]` if the data changes back to empty, but could be wrong

bryce-fitzsimons (approved):

PR #3143, opened 2024-10-18
Fix savedGroups sometimes hashing non-securestring attributes

Kevin-Chant:
### Features and Changes

A customer reported an issue where when they updated GrowthBook their feature payload started hashing the contents of saved groups even if they were only plain string IDs.

This only happens on a subset of codepaths which is why we didn't notice it yet, but it seems the original logic for `doHash` was incorrectly set to hash (cipher) attributes which can be used for hashing (experiment assignment).

This PR fixes the issue by extracting the existing, correct logic into a helper and calling that helper instead.

### Testing

My repro steps on `main` which no longer reproduce the bug on this branch were:

#### Setup

1. Ensure you have an  `id` attribute of type String which is a Unique Identifier
2. Create an ID List saved group for that attribute and populate it with a few simple values
3. Create a feature and add a targeting rule referencing that ID List
4. Generate the feature payload for an SDK with "Hash secure attributes" on but "Pass Saved Groups by reference" off

#### Reproduction steps

Most of the time, this results in the correct payload (when it's leveraging the cache). To reproduce the issue, toggle the feature off and back on, then refresh the SDK endpoint without changing any code (this would send it down a working code branch).

### Screenshots

Following the steps on `main`
image

Following the steps on this branch
image

### DIFF
--- a/packages/back-end/src/services/features.ts
+++ b/packages/back-end/src/services/features.ts
@@ -1238,7 +1238,7 @@ export function applySavedGroupHashing(
         salt,
         attributes,
         attribute,
-        doHash: attribute.hashAttribute,
+        doHash: shouldHash(attribute),
       });
     }
   });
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/features.ts
```diff
@@ -1287,15 +1287,7 @@ any {
...
-      doHash = attribute
-        ? !!(
-            attribute?.datatype &&
-            ["secureString", "secureString[]"].includes(
-              attribute?.datatype ?? ""
-            ) &&
-            !["$inGroup", "$notInGroup"].includes(key)
-          )
-        : doHash;
+      doHash = attribute ? shouldHash(attribute, key) : doHash;
```
gazzdingo, line 1290: we should never do hash its bad for you.

gazzdingo (approved):
LGTM

PR #3148, opened 2024-10-21
Bump TS version.

romain-growthbook:
~~This seems pretty straight forward!~~


tzjames (changes requested):
Thanks for doing this.  However is there a good reason why we have imports that are calling dist at all?  My best guess is that people or their IDE added it there by incorrectly by accident.  I think we can just remove the /dist in each of the current imports that are currently doing it, and it should probably work.  Then we don't need to make the tsconfig.json change at all.

tzjames (approved):
Ok should be good.  It would probably be wise to sanity check a page that loads from the imported files, but I don't see why it wouldn't work.

packages/back-end/package.json
```diff
@@ -149,9 +149,8 @@
     "mongodb-memory-server": "^9.2.0",
     "rimraf": "^3.0.2",
     "supertest": "^7.0.0",
-    "ts-patch": "^3.2.1",
-    "typescript": "5.3.3",
-    "typescript-transform-paths": "^3.5.1"
+    "tsconfig-replace-paths": "^0.0.14",
+    "typescript": "5.3.3"
```
jdorn, line 153: We should also update the typescript version for all of the sub-packages

PR #3159, opened 2024-10-23
Factor out datasource creation/update permissions.

romain-growthbook:
This PR factors out the logic for data source creation/update to include internal data warehouse logic.

The `type` attribute has been made explitly present but with possible `undefined` value to make sure that it is taken into account when calling the permission API.

### DIFF
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -212,7 +212,7 @@ export async function postDataSources(
   const { name, description, type, params, projects } = req.body;
   const settings = req.body.settings || {};
 
-  if (!context.permissions.canCreateDataSource({ projects })) {
+  if (!context.permissions.canCreateDataSource({ projects, type })) {
     context.permissions.throwPermissionError();
   }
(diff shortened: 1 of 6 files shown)


packages/back-end/src/controllers/datasources.ts
```diff
@@ -625,7 +625,8 @@ export async function postGoogleOauthRedirect(
   const context = getContextFromReq(req);
   const { projects } = req.body;
 
-  if (!context.permissions.canCreateDataSource({ projects })) {
+  // Huh, why? -- rb
```
Kevin-Chant, line 628: Extraneous comment?

Kevin-Chant (commented):
LGTM

Kevin-Chant (approved):

packages/back-end/src/controllers/datasources.ts
romain-growthbook: Well I'm not sure why that call is inside the oauth code but I'm guessing it's a smokescreen type of thing.

PR #3165, opened 2024-10-24
Data warehouse foundations

Kevin-Chant:
### Features and Changes

Creates and updates necessary helpers for the `growthbook_clickhouse` datasource type as well as the data enrichment flow.

### DIFF
--- a/packages/back-end/src/api/ingestion/ingestion.router.ts
+++ b/packages/back-end/src/api/ingestion/ingestion.router.ts
@@ -1,0 +1,63 @@
+import { z } from "zod";
+import { Router } from "express";
+import {
+  createApiRequestHandler,
+  validateIsSuperUserRequest,
+} from "back-end/src/util/handler";
+import { SDKConnectionInterface } from "back-end/types/sdk-connection";
+import { findSdkConnectionsAcrossMultipleOrgs } from "back-end/src/models/SdkConnectionModel";
+import { _dangerousGetAllGrowthbookClickhouseDataSources } from "back-end/src/models/DataSourceModel";
+
+interface SdkInfo {
+  organization: string;
+  client_key: string;
+  datasource: string;
+}
+
+interface GetDataEnrichmentResponse {
+  sdkData: {
+    [key: string]: SdkInfo;
+  };
+}
+
+function sdkInfo(conn: SDKConnectionInterface, datasource: string): SdkInfo {
+  return {
+    organization: conn.organization,
+    client_key: conn.id,
+    datasource,
+  };
+}
+
+export const getDataEnrichment = createApiRequestHandler({
+  bodySchema: z.never(),
+  querySchema: z.never(),
+  paramsSchema: z.never(),
+})(
+  async (req): Promise<GetDataEnrichmentResponse> => {
+    // Must be a super-user to make cross-org mongo queries
+    await validateIsSuperUserRequest(req);
+
+    const dataSources = await _dangerousGetAllGrowthbookClickhouseDataSources();
+    const dataSourcesByOrgId = Object.fromEntries(
+      dataSources.map((ds) => [ds.organization, ds.id])
+    );
+    const sdkConnections = await findSdkConnectionsAcrossMultipleOrgs(
+      Object.keys(dataSourcesByOrgId)
+    );
+    const sdkData = Object.fromEntries(
+      sdkConnections.map((conn) => [
+        conn.id,
+        sdkInfo(conn, dataSourcesByOrgId[conn.organization] || ""),
+      ])
+    );
+
+    return { sdkData };
+  }
+);
+
+const router = Router();
+
+// Mounted at /api/v1/ingestion
+router.get("/data-enrichment", getDataEnrichment);
+
+export default router;
--- a/packages/back-end/src/api/ingestion/ingestion.router.ts
+++ b/packages/back-end/src/api/ingestion/ingestion.router.ts
@@ -5,7 +5,7 @@ import {
   validateIsSuperUserRequest,
 } from "back-end/src/util/handler";
 import { SDKConnectionInterface } from "back-end/types/sdk-connection";
-import { findSdkConnectionsAcrossMultipleOrgs } from "back-end/src/models/SdkConnectionModel";
+import { _dangerousGetSdkConnectionsAcrossMultipleOrgs } from "back-end/src/models/SdkConnectionModel";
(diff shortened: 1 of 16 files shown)


packages/back-end/src/controllers/datasources.ts
```diff
@@ -111,7 +111,7 @@ export async function deleteDataSource(
     );
   }
 
-  await deleteDatasourceById(datasource.id, org.id);
+  await deleteDatasourceById(datasource, org.id);
```
toots, line 114: This should either reverted or the name should be changed. I don't see why this change slipped through.

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -88,6 +88,13 @@ export async function getDataSourcesByOrganization(
     context.permissions.canReadMultiProjectResource(ds.projects)
   );
 }
+// WARNING: This does not restrict by organization
+export async function _dangerousGetAllGrowthbookClickhouseDataSources() {
```
toots, line 92: I thought we wanted to cache this? Maybe at least a in-code cache using a dangling `let` variable.

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -124,12 +131,15 @@ export async function removeProjectFromDatasources(
   );
 }
 
-export async function deleteDatasourceById(id: string, organization: string) {
+export async function deleteDatasourceById(
```
toots, line 134: ditto

toots (commented):
Could of changes I think we should apply before merging.

packages/back-end/src/models/DataSourceModel.ts
Kevin-Chant: Yeah that's what I had initially but Jeremy suggested letting the ingestion endpoint handle the caching rather than having two caching layers.

packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -0,0 +1,63 @@
...
+    const dataSourcesByOrgId = Object.fromEntries(
+      dataSources.map((ds) => [ds.organization, ds.id])
+    );
+    const sdkConnections = await findSdkConnectionsAcrossMultipleOrgs(
+      Object.keys(dataSourcesByOrgId)
+    );
+    const sdkData = Object.fromEntries(
+      sdkConnections.map((conn) => [
+        conn.id,
+        sdkInfo(conn, dataSourcesByOrgId[conn.organization] || ""),
```
jdorn, line 50: Instead of defaulting to an empty string, we should skip these entries and log an error

packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -0,0 +1,63 @@
...
+interface GetDataEnrichmentResponse {
+  sdkData: {
+    [key: string]: SdkInfo;
+  };
+}
+
+function sdkInfo(conn: SDKConnectionInterface, datasource: string): SdkInfo {
+  return {
+    organization: conn.organization,
+    client_key: conn.id,
```
jdorn, line 26: This should be `conn.key` instead I believe

packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -0,0 +1,63 @@
...
+    const dataSources = await _dangerousGetAllGrowthbookClickhouseDataSources();
+    const dataSourcesByOrgId = Object.fromEntries(
+      dataSources.map((ds) => [ds.organization, ds.id])
+    );
+    const sdkConnections = await findSdkConnectionsAcrossMultipleOrgs(
+      Object.keys(dataSourcesByOrgId)
+    );
+    const sdkData = Object.fromEntries(
+      sdkConnections.map((conn) => [
+        conn.id,
```
jdorn, line 49: Same here, this should be `conn.key` instead

packages/back-end/src/controllers/datasources.ts
```diff
@@ -491,7 +491,8 @@ export async function postGoogleOauthRedirect(
   const context = getContextFromReq(req);
   const { projects } = req.body;
 
-  if (!context.permissions.canCreateDataSource({ projects })) {
+  // Huh, why? -- rb
+  if (!context.permissions.canCreateDataSource({ projects, type: undefined })) {
```
jdorn, line 495: This function is super old and not actually reachable in the UI anymore, but if we want to be correct here, the type should be `"google_analytics"`

packages/back-end/src/models/DataSourceModel.ts
jdorn: Yeah, it's using an index so will be super fast.  Dealing with cache invalidation in a horizontally scaled deployment is not fun and not worth it here.

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -132,6 +132,16 @@ export async function findSDKConnectionsByOrganization(
   );
 }
 
+export async function findSdkConnectionsAcrossMultipleOrgs(
```
jdorn, line 135: Should we name this as "dangerous" as well?

packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
```diff
@@ -181,7 +181,10 @@ export const postDemoDatasourceProject = async (
 
   if (
     !context.permissions.canCreateMetric({ projects: [demoProjId] }) ||
-    !context.permissions.canCreateDataSource({ projects: [demoProjId] })
+    !context.permissions.canCreateDataSource({
+      projects: [demoProjId],
+      type: undefined,
```
jdorn, line 186: The type here should be `"postgres"` since that's what we use for the demo data

packages/back-end/src/api/ingestion/ingestion.router.ts
Kevin-Chant: I realized that with some of the changes we made this isn't actually necessary anymore. The query that pulls these `conn`s is by their organization which is just the keys of the object being indexed here, so this should actually be safe now.

packages/back-end/src/models/DataSourceModel.ts
romain-growthbook: Word.

romain-growthbook (approved):
Seems good to go to me.

PR #3166, opened 2024-10-24
Add dogfooded event tracking to frontend

Kevin-Chant:
### Features and Changes

Creates a new side effect of calling `track()` which sends the event to the configured ingestion endpoint. This will enable dogfooding of our automated event tracking

### Dependencies

1. Ingestion endpoint must be ready & configured for production
2. #3165 must be deployed for the data enrichment endpoint to exist


romain-growthbook (approved):
I think this is pretty straight forward and conditional on the env var so wouldn't be executed until we have the ingestor in place.

PR #3157, opened 2024-10-23
[DRAFT] Data warehouse

romain-growthbook:
This is DRAFT PR! To see the total amount of changes that need to be merged into `main` from our data warehouse topic branch.

PR #3170, opened 2024-10-24
Support cloud clickhouse warehouse

Kevin-Chant:
### Features and Changes

The remainder of the changes from #3157 now that #3165 and #3166 have been merged


packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  return params;
+}
+
+export async function deleteClickhouseUser(
+  datasourceId: string,
+  organization: string
+) {
+  const client = createClickhouseClient({
```
Kevin-Chant, line 95: Need a permissions check here

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,122 @@
...
+import { logger } from "back-end/src/util/logger";
+
+export async function createClickhouseUser(
+  context: ReqContext,
+  datasourceId: string
+): Promise<DataSourceParams> {
+  // Temporary protection to prevent users from manually changing feature flag to create Clickhouse users.
+  if (ENVIRONMENT === "production" && context.org.id != "org_24yyifrkf649iz6") {
+    throw new Error(
+      "Clickhouse user creation is only allowed for the Growthbook organization"
```
romain-growthbook, line 23: Great.

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,122 @@
...
+    clickhouse_settings: {
+      max_execution_time: 3600,
+    },
+  });
+
+  const orgId = context.org.id;
+  const user =
+    ENVIRONMENT === "production"
+      ? `${orgId}_${datasourceId}`
+      : `test_${orgId}_${datasourceId}`;
```
romain-growthbook, line 43: Should this be placed into its own `const` for visibility?

packages/back-end/src/services/clickhouse.ts
romain-growthbook, line 95: Same protection as above? Maybe factor it out and re-use in both calls?

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,122 @@
...
+    application: "GrowthBook",
+    request_timeout: 3620_000,
+    clickhouse_settings: {
+      max_execution_time: 3600,
+    },
+  });
+  const user =
+    ENVIRONMENT === "production"
+      ? `${organization}_${datasourceId}`
+      : `test_${organization}_${datasourceId}`;
```
romain-growthbook, line 109: Definitely something to factor out!

packages/back-end/src/util/secrets.ts
```diff
@@ -267,3 +267,10 @@ export const USE_PROXY =
 
 export const SUPERADMIN_DEFAULT_ROLE =
   process.env.SUPERADMIN_DEFAULT_ROLE ?? "readonly";
+
+export const CLICKHOUSE_HOST = process.env.CLICKHOUSE_HOST || "";
+export const CLICKHOUSE_ADMIN_USER = process.env.CLICKHOUSE_ADMIN_USER || "";
+export const CLICKHOUSE_ADMIN_PASSWORD =
+  process.env.CLICKHOUSE_ADMIN_PASSWORD || "";
+export const CLICKHOUSE_DATABASE = process.env.CLICKHOUSE_DATABASE || "";
+export const CLICKHOUSE_MAIN_TABLE = process.env.CLICKHOUSE_MAIN_TABLE || "";
```
romain-growthbook, line 276: Shall we do some sanitization check, leave these as optional and fail at the calling site when they're not defined? This would be cleaner than relying on the external call and API to return an error..

romain-growthbook (commented):
Looking good. Couple of cleanups that I think should happen but otherwise looks good to go for me.

packages/back-end/src/services/clickhouse.ts
Kevin-Chant: That's what I'd thought at first but this is only called by `DataSourceModel#deleteDatasource` with
```
  if (datasource.type === "growthbook_clickhouse") {
    await deleteClickhouseUser(datasource.id, organization);
  }
```
so it seems safe to leave open like this. Users won't have clickhouse datasources and if somehow they do then they should be able to delete them

toots: I think that it would be more clear centralize the checks in the same file. Otherwise we risk missing one and/or making the code harder to proofread.

packages/back-end/src/util/secrets.ts
Kevin-Chant: Good idea. I'll add a check to the clickhouse service helpers to error out if they're not defined

packages/back-end/src/services/clickhouse.ts
Kevin-Chant: I agree in general, but I'm thinking this flow doesn't need a check at all. We shouldn't let users create the datasources (hence the other check), but if we mess up and they're able to create one at some point, we should let them delete it (no check here).

romain-growthbook: My concern when reading this code is: "This code is supposed to only apply to our own internal consumption at the moment, how can I make sure that this function is not executed by people from other orgs"?

Having the check right here makes it more clear. Also, when we open-up the functionality to other orgs, it would be easier to change the logic.

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -223,6 +228,7 @@ export async function createDataSource(
     integration.getInformationSchema &&
     integration.getSourceProperties().supportsInformationSchema
   ) {
+    logger.info("queueCreateInformationSchema");
```
jdorn, line 231: Make this a debug instead of info log

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  logger.info(`Creating Clickhouse user ${user}`);
+  await client.command({
+    query: `CREATE USER ${user} IDENTIFIED WITH sha256_hash BY '${hashedPassword}'`,
+  });
+
+  logger.info(`Creating Clickhouse view ${viewName}`);
+  await client.command({
+    query: `CREATE VIEW ${viewName} DEFINER=CURRENT_USER SQL SECURITY DEFINER AS SELECT * FROM ${CLICKHOUSE_MAIN_TABLE} WHERE organization = '${orgId}'`,
```
jdorn, line 87: This is fine for now, but I think we'll eventually want `datasourceId` to also be in the WHERE clauses here

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,140 @@
...
+  const viewName = `${database}.events`;
+
+  logger.info(`creating Clickhouse database ${database}`);
+  await client.command({
+    query: `CREATE DATABASE ${database}`,
+  });
+
+  logger.info(`Creating Clickhouse user ${user}`);
+  await client.command({
+    query: `CREATE USER ${user} IDENTIFIED WITH sha256_hash BY '${hashedPassword}'`,
```
jdorn, line 82: Should also specify `DEFAULT DATABASE` here

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  logger.info(`Creating Clickhouse view ${viewName}`);
+  await client.command({
+    query: `CREATE VIEW ${viewName} DEFINER=CURRENT_USER SQL SECURITY DEFINER AS SELECT * FROM ${CLICKHOUSE_MAIN_TABLE} WHERE organization = '${orgId}'`,
+  });
+
+  logger.info(`Granting select permissions on ${viewName} to ${user}`);
+  await client.command({ query: `GRANT SELECT ON ${viewName} TO ${user}` });
```
jdorn, line 91: This is fine for now.  In the future, we might want to be more generous with permissions and give the user full read/write access to their entire database.

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  logger.info(`Clickhouse user ${user} created`);
+
+  const parts = CLICKHOUSE_HOST.split(":");
+  const port = parseInt(CLICKHOUSE_HOST.split(":").pop() || "9000");
```
jdorn, line 96: Is this a safe way to get the port?  Can we parse with the `URL()` class instead?

packages/back-end/src/services/clickhouse.ts
jdorn: I would lean towards leaving this open and just gating the creation side of things.

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,140 @@
...
+    host: CLICKHOUSE_HOST,
+    username: CLICKHOUSE_ADMIN_USER,
+    password: CLICKHOUSE_ADMIN_PASSWORD,
+    database: CLICKHOUSE_DATABASE,
+    application: "GrowthBook",
+    request_timeout: 3620_000,
+    clickhouse_settings: {
+      max_execution_time: 3600,
+    },
+  });
```
jdorn, line 126: I would abstract out creating the admin clickhouse client into a separate function since it's shared in 2 places (and likely more in the future).

packages/back-end/src/services/clickhouse.ts
```diff
@@ -0,0 +1,140 @@
...
+  const user = clickhouseUserId(organization, datasourceId);
+
+  logger.info(`Deleting Clickhouse user ${user}`);
+  await client.command({
+    query: `DROP USER ${user}`,
+  });
+
+  logger.info(`Deleting Clickhouse database ${user}`);
+  await client.command({
+    query: `DROP DATABASE ${user}`,
```
jdorn, line 136: I assume this also deletes the view?  The ClickHouse docs are a little vague here - it says it deletes all tables, but doesn't mention anything about views.

packages/front-end/components/Settings/NewDataSourceForm.tsx
```diff
@@ -565,6 +616,34 @@ const NewDataSourceForm: FC<{
...
+                      Use Growthbook&apos;s Warehouse
+                    </h3>
+                    <p>
+                      If you don&apos;t have your own datasource you can use
+                      Growthbook&apos;s own warehouse to house your event
+                      tracking data
+                    </p>
+                  </div>
+                </div>
+              )}
```
jdorn, line 646: Thinking about this more, I would revert this entire file.  The final UI will be completely different anyway.  

If we need a quick button for us to set it up on our Cloud org, we can add something to the `admin.tsx` page or even just make the POST call manually.

packages/front-end/services/datasources.ts
```diff
@@ -720,43 +722,125 @@ interface InitialDatasourceResources {
...
+  geo_lon,
+  ua_device_type,
+  ua_browser,
+  ua_os,
+  utm_source,
+  utm_medium,
+  utm_campaign,
+  url_path,
+  session_id
+FROM ${table} WHERE organization = '${organization}'`,
```
jdorn, line 755: Ideally table would always just be `events` here.  I'd like to avoid using the full qualified table path if possible since it's very opaque to the user (they don't know their org/datasource id).  As a user, I would want to just be able to run `select * from events` and have it work.

packages/front-end/services/datasources.ts
```diff
@@ -720,43 +722,125 @@ interface InitialDatasourceResources {
...
+}): InitialDatasourceResources => {
+  return {
+    factTables: [
+      {
+        factTable: {
+          name: "Clickhouse Events",
+          description: "",
+          sql: `SELECT
+  timestamp,
+  user_id,
```
jdorn, line 740: Add `device_id as anonymous_id,` (and the associated column below)

packages/front-end/services/datasources.ts
```diff
@@ -720,43 +722,125 @@ interface InitialDatasourceResources {
...
+  utm_medium,
+  utm_campaign,
+  url_path,
+  session_id
+FROM ${table} WHERE organization = '${organization}'`,
+          columns: generateColumns({
+            timestamp: { datatype: "date" },
+            user_id: { datatype: "string" },
+            user_attributes_json: { datatype: "string" },
+            event_name: { datatype: "string" },
```
jdorn, line 760: Set `alwaysInlineFilter: true` for event_name

packages/front-end/services/datasources.ts
```diff
@@ -720,43 +722,125 @@ interface InitialDatasourceResources {
...
+            ua_device_type: { datatype: "string" },
+            ua_browser: { datatype: "string" },
+            ua_os: { datatype: "string" },
+            utm_source: { datatype: "string" },
+            utm_medium: { datatype: "string" },
+            utm_campaign: { datatype: "string" },
+            url_path: { datatype: "string" },
+            session_id: { datatype: "string" },
+          }),
+          userIdTypes: [],
```
jdorn, line 774: This needs to be `["user_id", "anonymous_id"]`

packages/front-end/services/datasources.ts
```diff
@@ -939,44 +1022,64 @@ SELECT
...
+    datasource.settings?.schemaFormat === "ga4"
+  )
+    return getBigQueryWithGa4InitialDatasourceResources({
+      params: datasource.params,
+      userIdTypes: datasource.settings?.userIdTypes || [],
+    });
+
+  if (datasource.type === "growthbook_clickhouse")
+    return getClickHouseInitialDatasourceResources({
+      table: process.env.CLICKHOUSE_MAIN_TABLE || "test_enriched_events",
```
jdorn, line 1080: You can't access `process.env` on the front-end.  Better if we can lookup the table from the `datasource.settings` or `datasource.params` object instead.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -249,6 +251,138 @@ export async function postDataSources(
...
+export async function postInbuiltDataSource(
+  req: AuthRequest<
+    {
+      name?: string;
+      description?: string;
+      type?: DataSourceType;
+      params?: DataSourceParams;
+      settings?: DataSourceSettings;
+      projects?: string[];
+      metricsToCreate?: AutoMetricToCreate[];
```
jdorn, line 263: Remove this argument

packages/back-end/src/controllers/datasources.ts
```diff
@@ -249,6 +251,138 @@ export async function postDataSources(
...
+        userIdType: "user_id",
+      },
+    ],
+    queries: {
+      exposure: [
+        {
+          id: "anonymous_id",
+          dimensions: [
+            "country",
+            "city",
```
jdorn, line 296: Remove `city` as a dimension.  Cardinality is too high.

PR #2922, opened 2024-08-23
New JSON webhook type, new doc

romain-growthbook:
This PR introduces a new `JSON` event webhook type. This new type is used to map to the new event webhook. payloads.

Existing `raw` webhooks are then considered legacy and frozen:
* No new `raw` webhook can be created
* Existing `raw` webhook can be edited but their event list and type are also frozen and cannot be changed.

This makes it possible to switch to the new type of webhooks without having to migrate any data!

Also, the doc is updated to reflect the new webhook and their payload. See screenshot below.

### Screenshots

**App setup**

Screenshot 2024-08-23 at 9 39 45 AM

Screenshot 2024-08-23 at 9 39 33 AM

**Doc**

Screenshot 2024-09-06 at 11 56 18 AM

Screenshot 2024-09-05 at 10 25 55 AM

PR #3172, opened 2024-10-24
Assume a webhook is raw (legacy) when no payload types are specified.

romain-growthbook:
This PR is related to some users seeing new JSON payload being delivered to raw (legacy) webhook.

It seems that the previous PRs (https://github.com/growthbook/growthbook/pull/2922 and https://github.com/growthbook/growthbook/pull/2836) introducing the new payload format missed a case where very old webhook definitions don't have a `payloadType` at all.

On the front end, the interface exporter adds a default `"raw"` payload type when no payload exists so this was never an issue.

However, on the back-end, there were a couple of code path where no `payloadType` could have resulted in the wrong type being assumed.

This PR:
* Fixes at least one of those cases
* Adds a model migration that will add the missing `payloadType` on any webhook model that is exported via `toInterface`.

This should ensure that, should another code path have been missed, the chances of it being being triggered would be very low and limited to one occurence since, most likely, the corresponding model would have been migrated after the first occurence.

### DIFF
--- a/packages/back-end/src/models/EventWebhookModel.ts
+++ b/packages/back-end/src/models/EventWebhookModel.ts
@@ -178,6 +178,7 @@ const toInterface = (doc: EventWebHookDocument): EventWebHookInterface => {
     ...(payload.payloadType ? {} : { payloadType: "raw" }),
     ...(payload.headers ? {} : { headers: {} }),
     ...(payload.tags ? {} : { tags: [] }),
+    ...(payload.projects ? {} : { projects: [] }),
     ...(payload.environments ? {} : { environments: [] }),
   };
 


packages/back-end/src/models/EventWebhookModel.ts
```diff
@@ -167,17 +167,33 @@ type EventWebHookDocument = mongoose.Document & EventWebHookInterface;
...
+    // All webhook are created with a payloadType. This is here for antiquated ones
+    // which don't have one and should be considered raw.
+    ...(payload.payloadType ? {} : { payloadType: "raw" }),
+    ...(payload.headers ? {} : { headers: {} }),
+    ...(payload.tags ? {} : { tags: [] }),
+    ...(payload.environments ? {} : { environments: [] }),
+  };
+
+  if (Object.keys(defaults).length)
+    await EventWebHookModel.updateOne(
```
Kevin-Chant, line 187: Can we do this without an `await`? We can let the update be handled in the background so this doesn't need to become an async function. Might also be related to the timeouts in CI

Kevin-Chant (commented):
I think it'd be better to leave `toInterface` sync and just not wait on the mongo writes

packages/back-end/src/models/EventWebhookModel.ts
romain-growthbook: Excellent suggestion, thanks!

packages/back-end/src/models/EventWebhookModel.ts
```diff
@@ -170,14 +170,28 @@ type EventWebHookDocument = mongoose.Document & EventWebHookInterface;
...
+    ...(payload.environments ? {} : { environments: [] }),
+  };
+
+  if (Object.keys(defaults).length)
+    void EventWebHookModel.updateOne(
+      { id: doc.id },
+      {
+        $set: defaults,
+      }
+    );
```
jdorn, line 190: This needs a `catch` block

Kevin-Chant (approved):

PR #3175, opened 2024-10-24
changing small sample size error to warning

lukebrawleysmith:
### Features and Changes
when sample size is small for a bandit, return warning rather than error
previous: 
error
updated: 
warning

### DIFF
--- a/packages/stats/gbstats/bayesian/bandits.py
+++ b/packages/stats/gbstats/bayesian/bandits.py
@@ -38,7 +38,8 @@ class BanditResponse:
     bandit_weights: Optional[List[float]]
     best_arm_probabilities: Optional[List[float]]
     seed: int
-    bandit_update_message: Optional[str]
+    bandit_update_message: str
+    enough_units: Optional[bool]
 
 
 def get_error_bandit_result(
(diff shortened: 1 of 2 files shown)


packages/stats/gbstats/bayesian/bandits.py
```diff
@@ -42,7 +42,7 @@ class BanditResponse:
 
 
 def get_error_bandit_result(
-    error: str, reweight: bool, current_weights: List[float]
+    update_message: str, error: str, reweight: bool, current_weights: List[float]
```
lukesonnet, line 45: update_message isn't used anywhere in here.

bryce-fitzsimons (approved):

packages/stats/gbstats/gbstats.py
```diff
@@ -650,12 +650,25 @@ def get_bandit_result(
...
                 error="not all statistics are instance of type BanditStatistic",
                 reweight=bandit_settings.reweight,
                 current_weights=bandit_settings.current_weights,
             )
         srm_p_value = b.compute_srm()
         bandit_result = b.compute_result()
+        if (
+            bandit_result.bandit_update_message
+            and bandit_result.bandit_update_message.startswith(
+                "total sample size is only "
```
lukesonnet, line 663: Let's use a flag instead of this string matching.

lukesonnet (changes requested):

packages/stats/gbstats/bayesian/bandits.py
lukebrawleysmith: fixed

lukesonnet (approved):

PR #3177, opened 2024-10-25
Update and move ProjectTagBar to reflect design

msamper:
### Features and Changes

Updates the ProjectTagBar to reflect new designs for the upcoming updated experiment creation/editing flow. As a part of the changes with the design, the ProjectTagBar has been moved to live below the experiment header rather than in the Overview tab.

This PR also eliminates the hex color picker we have for tags and updates the color presets to reflect Radix colors available to use with the new Radix Badge component.  Old badges using hex colors are migrated over to Radix colors using the CIELAB Color Space (which accounts for human perception) at the `DefinitionsContext` level. We could consider migrating the colors within the collection once all components using tags have been migrated over to Radix. I also went ahead and created a new `UserAvatar` to encapsulate logic needed for rendering a users initials or an icon.

### Testing

Add/Remove tags to experiments and create new tags through the tags page. 
Check different experiments and make sure the ProjectTagBar is rendering correctly in the header.

### Screenshots

Screenshot 2024-10-24 at 8 01 05 PM

Screenshot 2024-10-24 at 8 01 55 PM

Existing tags with hex colors before
Screenshot 2024-10-30 at 7 10 05 PM

Existing tags with hex colors using the closest radix color
Screenshot 2024-10-30 at 7 11 41 PM


packages/front-end/components/Tags/Tag.tsx
```diff
@@ -1,32 +1,45 @@
...
 export default function Tag({ tag, color, description, skipMargin }: Props) {
   const { getTagById } = useDefinitions();
   const fullTag = getTagById(tag);
 
   const displayTitle = description ?? fullTag?.description ?? "";
-  const displayColor = color ?? fullTag?.color ?? "#029dd1";
+
+  const tagColor = color ?? fullTag?.color ?? "blue";
+  // If a tag is still using a hex code color, we'll default to blue
+  const displayColor = isTagColor(tagColor) ? tagColor : "blue";
```
jdorn, line 34: This is too aggressive of a change.  A lot of people rely on tag colors to differentiate between types of tags.

If we want to fully move to the Radix badge component, I think we need to:
1. Allow the full range of Radix colors, not just a few handpicked ones
2. Migrate existing hex colors to the "closest" Radix color

At the very least, we should make sure all of the old pre-selected `tagColors` values map to similar Radix colors.

Finding the "closest" color is not trivial.  There's a whole wikipedia page on this topic - https://en.wikipedia.org/wiki/Color_difference .  Basically need to compute the distance between the hex and each of the Radix colors and then take the one with the smallest distance.  Depending on how perceptually accurate we want the distance to be, there's a whole bunch of different formulas to choose from.

packages/front-end/components/Tags/SortedTags.tsx
```diff
@@ -37,11 +38,11 @@ export default function SortedTags({
...
-        <Tag
-          tag={tagCopy}
-          key="tag-ellipsis"
-          skipMargin={useFlex}
-          color="#ffffff"
+        <Badge
+          label={tagCopy}
+          color="violet"
+          variant="soft"
+          ml={useFlex ? undefined : "2"}
```
msamper, line 45: Replace with text to avoid having it look like a badge

packages/front-end/components/Avatar/InitialAvatar.tsx
```diff
@@ -0,0 +1,16 @@
+import Avatar, { Props as AvatarProps } from "@/components/Radix/Avatar";
+
+type Props = {
+  name: string;
+} & Omit<AvatarProps, "children">;
+
+export default function InitialAvatar({ name, ...otherProps }: Props) {
```
bryce-fitzsimons, line 7: not sure this should be a separate component. Imagine at some point we allow avatar customization. You'd need to conditionally choose the component type everywhere this is implemented. Maybe instead use a helper function `getUserAvatar()` that either returns the avatar icon or defaults to the user initials.

msamper: I originally did this inside of our Radix Avatar component, but I think I'd need to change its props. Right now that component takes in children for its content and that would be fine if the avatar was for the current user, but in this case it's the experiment owner so we'd likely need to pass in name as a prop. If we create a helper function like `getUserAvatar()` then presumably we don't want it to take in `children` anyway, but it seems like we're also using the Avatar component for an icon in `EventSourceList` right now. Maybe we keep the HOC if we don't want to change our Radix Avatar wrapper and give it the helper function or we just don't use Avatars for icons and I can remove the `children` prop

packages/front-end/components/Experiment/TabbedPage/ProjectTagBar.tsx
```diff
@@ -36,111 +35,124 @@ export default function ProjectTagBar({
...
+          )}{" "}
+        </div>
+      </div>
+      <div className="row">
+        <div className="col-auto">
+          <Text weight="medium">Tags: </Text>
+          {experiment.tags?.length > 0 && (
+            <SortedTags
+              tags={experiment.tags}
+              skipFirstMargin={true}
```
bryce-fitzsimons, line 126: thinking it might be better to have no margins at all (useFlex) and instead use flex gap. Allows for more graceful wrapping.

packages/front-end/components/Teams/Roles/RoleList.tsx
```diff
@@ -45,9 +45,21 @@ export default function RoleList() {
...
+                        ml="2"
+                      />
+                    ) : null}
                     {isDeactivated ? (
-                      <Tag color="#808080" tag="Deactivated" />
+                      <Badge
+                        label="Deactivated"
+                        color="gray"
+                        variant="soft"
+                        ml="2"
```
bryce-fitzsimons, line 61: same feedback about using flex + gap

packages/front-end/services/tags.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const l = 116 * y - 16;
+  const a = 500 * (x - y);
+  const b = 200 * (y - z);
+
+  return { l, a, b };
+}
+
+// Calculate Euclidean distance in CIELAB
+function labDistance(lab1, lab2) {
```
bryce-fitzsimons, line 114: fancy

packages/front-end/services/tags.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return Math.sqrt(
+    Math.pow(lab1.l - lab2.l, 2) +
+      Math.pow(lab1.a - lab2.a, 2) +
+      Math.pow(lab1.b - lab2.b, 2)
+  );
+}
+
+// Find the closest color in CIELAB space
+export function findClosestRadixColor(targetHex: string): RadixColor {
```
bryce-fitzsimons, line 123: kudos for pulling this off

packages/front-end/components/Avatar/InitialAvatar.tsx
bryce-fitzsimons: can you make a <UserAvatar> HOC that takes in necessary user info and then makes a decision about what to pass as `children` to <Avatar>?

bryce-fitzsimons: maybe props look like...
```
{
  name: string,
  icon?: ReactElement,
  color?: RadixColor,
}
```

msamper: Kept the `Avatar` props and excluded `children` and added `name` and `icon`. Not sure if we have a `radius` and `variant` that we plan to use for all user avatars.

bryce-fitzsimons: maybe I'm just hung up on the name? I'd expect this to be the definitive user avatar component, not just the "initial" component used when we don't have enough information to render a proper avatar

msamper: Oh sorry I renamed the component and not the file 🤦🏻‍♂️

msamper: Fixed

bryce-fitzsimons (approved):

PR #3192, opened 2024-10-28
Add null typeguard for window history

Kevin-Chant:
### Features and Changes

We have a few instances coming through of `window.history.state is null`, which can be easily fixed by allowing a default key for the initial pageload before the router has populated the history. It's possible that this can make two events on that initial pageview appear under different page ids depending on timing, but I think that should be an insignificant edge case.

### Testing

I haven't been able to reproduce a case where this codepath naturally runs on a null history, but you can check the behavior in the console on a new tab with no history 

### Screenshots

On a new tab before navigating
image

On that tab after navigating to a non-spa site
image

On a new tab after loading growthbook
image

### DIFF
--- a/packages/front-end/services/track.ts
+++ b/packages/front-end/services/track.ts
@@ -111,7 +111,7 @@ function getOrGenerateDeviceId() {
 function getOrGeneratePageId() {
   // On initial load if the router hasn't initialized a state change yet then history.state will be null.
   // Since this only happens on one pageload, using a hardcoded default key should still work as its own key
-  if (!((window.history.state?.key || "null") in pageIds)) {
+  if (!((window.history.state?.key || "") in pageIds)) {
     pageIds[window.history.state.key] = uuidv4();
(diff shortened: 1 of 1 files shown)


packages/front-end/services/track.ts
```diff
@@ -109,7 +109,9 @@ function getOrGenerateDeviceId() {
 }
 
 function getOrGeneratePageId() {
-  if (!(window.history.state.key in pageIds)) {
+  // On initial load if the router hasn't initialized a state change yet then history.state will be null.
+  // Since this only happens on one pageload, using a hardcoded default key should still work as its own key
+  if (!((window.history.state?.key || "null") in pageIds)) {
```
gazzdingo, line 114: could we just use an empty string here?
```suggestion
  if (!((window.history.state?.key || "") in pageIds)) {
```

gazzdingo (approved):
LGTM

PR #3204, opened 2024-10-29
BUGFIX: Do not close modal when 'Enter' is presser while an input is focused

ahdriel:
### Behavior

When a modal is open and we have a input selected, pressing Enter would immediately close the modal which is not the intended behavior.

### Fix

Now when Enter is pressed we will submit the form instead (same as clicking Submit on bottom right).

### Fix explanation

Because the button type was not specified, it defaults to `submit`.

When there are multiple submit buttons in the DOM, the browser will call the first one which in this case was the Cancel button on the bottom actions.

### DIFF
--- a/packages/front-end/components/Modal.tsx
+++ b/packages/front-end/components/Modal.tsx
@@ -255,6 +255,7 @@ const Modal: FC<ModalProps> = ({
           >
             {close && includeCloseCta ? (
               <button
+                type="button"
                 className={closeCtaClassName}
                 onClick={async (e) => {
                   e.preventDefault();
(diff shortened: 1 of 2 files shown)


bryce-fitzsimons:
Looks good. Could you also double check the injected CTA `<button>`s in `PagedModal.tsx`. You can see this type of modal implemented on the Experiments page > New Experiment

jdorn (approved):

ahdriel:
@bryce-fitzsimons Good catch! It worked for the first screen but it broke when we had a `Skip` or `Back` button on step 2+.

Fixed.

PR #3212, opened 2024-10-29
Add new layout for SDK connections empty state page

ahdriel:
### Features and Changes

Added a new empty state layout for the SDK Connections page under the flag `sdk-connections-new-empty-state`.

### Screenshots

| Before | After |
|--------|--------|
| GrowthBook _ SDK Connections | sdk-empty-after |


packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
```diff
@@ -40,17 +43,79 @@ export default function SDKConnectionsList() {
     project
   );
 
+  const useNewEmptyStateLayout = useFeatureIsOn(
+    "sdk-connections-new-empty-state"
+  );
+
+  const popularLanguages = getLanguagesByFilter("popular");
```
ahdriel, line 50: @jdorn I am using `popular` here to keep it in sync with the options shown when the modal is first opened. But we can also hardcode the list for this layout if that's preferred.

jdorn: I'm worried someone using Golang for example, would just abandon if they saw this because there's no indication that we support more than what's listed.

A few options:
1. Show all languages, but group by category so it's not one huge list
2. Add tabs for the different categories and let the user switch between them (like we do in the modal)
3. Add a "See more..." link that reveals the rest of the languages when clicked

packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
```diff
@@ -40,17 +43,79 @@ export default function SDKConnectionsList() {
     project
   );
 
+  const useNewEmptyStateLayout = useFeatureIsOn(
+    "sdk-connections-new-empty-state"
```
lukesonnet, line 47: @jdorn useFeatureIsOn doesn't seem to respect the list of AppFeatures. Is that a problem? It's certainly more convenient for the engineer but it's out of line with using, say, `growthbook.isOn`?

jdorn: `useFeatureIsOn` is a generic method that accepts AppFeatures as a type argument.  Exactly the same as `useGrowthBook`.  Neither is particularly good at enforcing type safety because it's easy to forget to pass it in.

lukesonnet: Ah, I see. Thanks.

packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
ahdriel: @jdorn Updated the layout to use option 3 & updated the gif in the description with the new experience.

packages/front-end/components/Features/SDKConnections/SDKLanguageSelector.tsx
```diff
@@ -16,44 +16,26 @@ const tabs: Record<LanguageFilter, string> = {
...
-  setValue,
-  multiple,
+  onClick,
 }: {
   language: SDKLanguage;
-  selected: Set<SDKLanguage>;
-  setValue: (languages: SDKLanguage[]) => void;
-  multiple: boolean;
+  selected: boolean;
+  onClick: (language: React.MouseEvent) => void;
```
jdorn, line 26: I would make this accept no arguments and just always do `e.preventDefault()` internally before calling the provided `onClick()` method.  I can't think of a reason why a consumer of this component would need access to the raw mouse event.

packages/front-end/components/Radix/Button.tsx
```diff
@@ -17,7 +17,7 @@ export type Props = {
   setError?: (error: string | null) => void;
   icon?: ReactNode;
   iconPosition?: "left" | "right";
-  children: string | string[];
+  children: ReactNode;
```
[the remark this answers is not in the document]
ahdriel: Updated -- I am concerned that in the future we will have a Button component with 30 props because we will need to surface all underlying children props to the consumer, but we are not there yet so I rolled this back and added `textWeight` as a new prop.

packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
```diff
@@ -40,17 +56,113 @@ export default function SDKConnectionsList() {
...
+        </Heading>
+        <Text size="3" align="center">
+          Select your SDK to connect your front-end, back-end or mobile app.
+        </Text>
+      </Flex>
+
+      <Separator size="4" mt="7" mb="5" />
+
+      <Heading as="h3" size="2" mb="3">
+        <Strong>Select your SDK:</Strong>
```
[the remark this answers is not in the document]
ahdriel: agreed -- removed

packages/front-end/components/Radix/Button.tsx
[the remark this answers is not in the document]
jdorn: I agree.  Looking at the figma, it looks like there's a mismatch between our design system and Radix.  Radix uses bold for all buttons except for ghost variants.  However, in our design system the ghost buttons also use bold.

So instead of a prop, we should either fix our design system to match Radix or add some custom CSS within the Button component to force ghost buttons to be bold.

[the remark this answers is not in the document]
ahdriel: I have reverted the changes and will keep all buttons with weight `medium` for now.

If we feel more strongly in the future that all ghost buttons should be `600` then we can easily add that rule -- but will default to keep all buttons the same at the moment.

packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
```diff
@@ -40,17 +56,109 @@ export default function SDKConnectionsList() {
     project
   );
 
+  const useNewEmptyStateLayout = useFeatureIsOn(
+    "sdk-connections-new-empty-state"
+  );
```
jdorn, line 61: Just realized we should change how we're evaluating this feature flag.  We only want to check the feature flag if the user is going to see the empty state.  If someone already has SDK connections, we don't want to include them in the experiment since it will just add noise.

We did something similar for the onboarding test we ran:

```ts
const gb = useGrowthBook();

let useNewEmptyStateLayout = false;
if (connections.length === 0 && canCreateSDKConnections) {
  useNewEmptyStateLayout = gb.isOn("sdk-connections-new-empty-state");
}
```

ahdriel: Oh that's a good catch! Updated following your recommendation.

jdorn (approved):

PR #3227, opened 2024-10-31
Update caniuse-lite version

ahdriel:
### Features and Changes

Updating caniuse-lite version, which is used internally by NextJS.


msamper (approved):
Thanks for updating! Tested and looks like the warning is gone

PR #3231, opened 2024-10-31
updating bandit unit tests

lukebrawleysmith:
adding unit tests for bandit variation weight updates and single variation results

### DIFF
--- a/packages/stats/tests/test_gbstats.py
+++ b/packages/stats/tests/test_gbstats.py
@@ -80,83 +80,6 @@ QUERY_OUTPUT = [
 
 MULTI_DIMENSION_STATISTICS_DF = pd.DataFrame(QUERY_OUTPUT)
 
-# used for testing bandits
-QUERY_OUTPUT_BANDITS = [
-    {
-        "dimension": "All",
-        "bandit_period": 0,
-        "variation": "zero",
-        "main_sum": 270,
-        "main_sum_squares": 848.79,
-        "users": 100,
-        "count": 100,
-    },
-    {
-        "dimension": "All",
-        "bandit_period": 0,
-        "variation": "one",
-        "main_sum": 300,
-        "main_sum_squares": 869,
-        "users": 120,
-        "count": 120,
-    },
-    {
-        "dimension": "All",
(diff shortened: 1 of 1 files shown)


packages/stats/tests/test_gbstats.py
```diff
@@ -832,70 +755,99 @@ def test_format_results_denominator(self):
...
+            best_probs[variation] = np.mean(y[:, variation] == best_rows)
+        best_probs[best_probs < 0.01] = 0.01
+        best_probs /= sum(best_probs)
+        best_probs = np.round(np.array(best_probs, dtype=float), decimals=4)
+        if self.r.updatedWeights:
+            result_rounded = np.array([round(num, 4) for num in self.r.updatedWeights])
+            self.assertTrue(np.array_equal(result_rounded, best_probs))
         else:
-            assert 1 > 2, "wrong class"
+            self.assertEqual(1, 2)
```
lukesonnet, line 850: Is the point to throw an error if self.r.updatedWeights is false? Let's do that explicitly instead?

packages/stats/tests/test_gbstats.py
```diff
@@ -832,70 +755,99 @@ def test_format_results_denominator(self):
...
+                round(single_variation_result.ci[1], 4)
+                if single_variation_result.ci
+                else 0
+                for single_variation_result in self.r.singleVariationResults
+            ]
+            self.assertEqual(cr_rounded, list(self.variation_results["cr"]))
+            self.assertEqual(lower_rounded, list(self.variation_results["lower"]))
+            self.assertEqual(upper_rounded, list(self.variation_results["upper"]))
+        else:
+            assert 1 > 2, "singleVariationResults do not exist"
```
lukesonnet, line 828: Same here.

packages/stats/tests/test_gbstats.py
lukebrawleysmith: good point, and done

packages/stats/tests/test_gbstats.py
lukebrawleysmith: done

PR #3234, opened 2024-11-01
Make sure targeting is copied when duplicating an experiment

bryce-fitzsimons:
Targeting conditions (which live in the `phases` array) were not being copied over on duplicate.

### DIFF
--- a/packages/front-end/components/Experiment/NewExperimentForm.tsx
+++ b/packages/front-end/components/Experiment/NewExperimentForm.tsx
@@ -6,6 +6,7 @@ import {
   Variation,
 } from "back-end/types/experiment";
 import { useRouter } from "next/router";
+import { getValidDate } from "shared/dates";
 import { DataSourceInterfaceWithParams } from "back-end/types/datasource";
 import { OrganizationSettings } from "back-end/types/organization";
 import {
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -227,33 +226,28 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
-              coverage: initialValue.phases?.[0].coverage || 1,
-              dateStarted: getValidDate(
-                initialValue.phases?.[0]?.dateStarted ?? ""
-              )
-                .toISOString()
-                .substr(0, 16),
-              dateEnded: getValidDate(initialValue.phases?.[0]?.dateEnded ?? "")
-                .toISOString()
-                .substr(0, 16),
-              name: initialValue.phases?.[0].name || "Main",
```
jdorn, line 241: I think this other code here is important to keep for importing past experiments.  I'm pretty sure the `getValidDate` stuff at the very least was added because of a bug.

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -227,33 +228,45 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
-              reason: "",
-              variationWeights:
-                initialValue.phases?.[0].variationWeights ||
-                getEqualWeights(
-                  initialValue.variations ? initialValue.variations.length : 2
+        ...(initialValue?.phases?.[initialValue?.phases?.length - 1]
```
jdorn, line 231: nit: Change to use `lastPhase` here

jdorn (approved):

packages/front-end/components/Experiment/NewExperimentForm.tsx
bryce-fitzsimons, line 231:
```suggestion
        ...(initialValue?.phases?.[lastPhase]
```

PR #3241, opened 2024-11-02
Metrics Empty State

jdorn:
### Features and Changes

First pass at a metrics / fact tables empty state.  This is not the final design, but aims to solve a few of the most glaring issues with the current page:

- [x] If you don't have a data source yet, we should direct you to add that first before adding a fact table / metric
- [x] If you don't have any metrics yet, don't show the Metric Groups tab
- [x] Remove "Manual" option when creating a new metric
- [x] Define "Initial Resources" for segment, rudderstack, and amplitude data sources (created automatically after connecting)
- [x] Change "Generate Fact Tables" to use "Initial Resources" approach instead

Metrics Empty State (no data source created yet):

image

Metrics Empty State (after data source created):

image

Fact Tables Empty State (no data source created yet):

image

Fact Tables Empty State (after data source created):

image


packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -255,7 +255,7 @@ const MetricsList = (): React.ReactElement => {
     (m) => ({
       datasourceName: m.datasource
         ? getDatasourceById(m.datasource)?.name || "Unknown"
-        : "Manual",
+        : "None",
```
lukesonnet, line 258: nit: Should we have a warning in this case? Maybe a fast follow to help people clean up orphaned metrics.

packages/front-end/pages/fact-tables/index.tsx
```diff
@@ -254,26 +256,28 @@ export default function FactTablesPage() {
...
             >
-              Add Fact Table
-            </Button>
-          </Tooltip>
+              <Button
+                onClick={() => {
+                  if (!canCreate) return;
+                  setCreateFactOpen(true);
+                }}
+                disabled={!canCreate}
```
lukesonnet, line 275: nit: Should we just have this disabled instead of invisible? So that way the "connect a datasource first" message and CTA are more just added to this page, but it always looks like it will in the future, minus this button being disabled? That's kind of what I expect from apps, but might not be worth it.

packages/front-end/components/Metrics/MetricsList.tsx
jdorn: Yeah, a fast follow would be good

packages/front-end/pages/fact-tables/index.tsx
jdorn: People tend not to read, so I think having a single CTA is better and will be less confusing to users.

PR #2368, opened 2024-04-09
Allow copying and pasting metrics in analysis settings

bttf:
Apr-09-2024 17-58-26

PR #3247, opened 2024-11-05
Fix copy/paste metrics in the Analysis settings screen

ahdriel:
### Features and Changes

Originally implemented in #2368 it likely broke when we added Fact Tables as a new metric source.

This fixes it.

### DIFF
--- a/packages/front-end/components/Experiment/MetricsSelector.tsx
+++ b/packages/front-end/components/Experiment/MetricsSelector.tsx
@@ -247,7 +247,14 @@ const MetricsSelector: FC<{
         try {
           const clipboard = e.clipboardData;
           const data = JSON.parse(clipboard.getData("Text"));
-          if (data.every((d) => d.startsWith("met_") || d.startsWith("mg_"))) {
+          if (
+            data.every(
+              (d) =>
+                d.startsWith("met_") ||
+                d.startsWith("mg_") ||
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -247,7 +247,14 @@ const MetricsSelector: FC<{
         try {
           const clipboard = e.clipboardData;
           const data = JSON.parse(clipboard.getData("Text"));
-          if (data.every((d) => d.startsWith("met_") || d.startsWith("mg_"))) {
+          if (
+            data.every(
```
ahdriel, line 251: Two questions I have (but not necessarily for this PR) are:
- Should this be `every` or should we also succeed partially if we can?
- Would we benefit of having these significant prefixes list somewhere and consume it here so this doesn't happen again?

jdorn (approved):

packages/front-end/components/Experiment/MetricsSelector.tsx
jdorn: I could go either way on every vs some.  The only real magic prefix is `fact__` (and we do have a helper method already that checks if something is a fact metric id). The rest are just conventions and not strictly enforced. So there are some metrics that don't start with `met_` for example.

PR #3265, opened 2024-11-07
Align tags with Radix Soft badges

lukesonnet:
### Features and Changes


lukesonnet:
Abandoning this until we land Autocomplete.

Note the color picker on the tags page is also probably showing the wrong colors.

PR #3268, opened 2024-11-07
fixed colors on head and add metadata component

gazzdingo:
### Features and Changes
updated the header to look more like the designs

### Screenshots
image

### DIFF
--- a/packages/front-end/styles/theme.scss
+++ b/packages/front-end/styles/theme.scss
@@ -122,6 +122,10 @@
   --step-completed-background: #218358;
   --step-completed-color: #fbfefc;
   --step-color: #6550b9;
+
+  // radix color-scheme;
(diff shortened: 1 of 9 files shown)


packages/front-end/styles/theme.scss
```diff
@@ -122,6 +122,10 @@
   --step-completed-background: #218358;
   --step-completed-color: #fbfefc;
   --step-color: #6550b9;
+
+  // radix color-scheme;
+  --color-text-high: var(--indigo-12);
+  --color-text-mid: rgba(31, 45, 92, 0.5);
```
gazzdingo, line 128: need to be 15%

lukesonnet (changes requested):
changes needed in DM

lukesonnet (approved):

PR #3269, opened 2024-11-07
Fix loading flicker on features, metrics, and other pages

jdorn:
### Features and Changes

The `/features` page was briefly flickering on page load.  It would show the empty state before immediately re-rendering with the features table.  This was due to using `useEffect` and `useState` instead of `useMemo`.

The `/metrics` page (and others) were also flickering on page load.  This was due to us rendering pages before `DefinitionsContext` was ready.  Instead of adding a `ready` check in every page, I just prevent the entire app from rendering until definitions are ready which simplifies a lot.

### DIFF
--- a/packages/front-end/services/DefinitionsContext.tsx
+++ b/packages/front-end/services/DefinitionsContext.tsx
@@ -10,6 +10,7 @@ import {
   FC,
   ReactNode,
   useCallback,
+  ReactElement,
 } from "react";
 import { TagInterface } from "back-end/types/tag";
 import {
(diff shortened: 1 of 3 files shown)


packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -308,3 +310,13 @@ export const DefinitionsProvider: FC<{ children: ReactNode }> = ({
     </DefinitionsContext.Provider>
   );
 };
+
+export function DefinitionsGuard({ children }: { children: ReactElement }) {
+  const { ready, error } = useDefinitions();
+
+  if (!error && !ready) {
+    return <LoadingOverlay />;
+  }
```
Kevin-Chant, line 319: Is it worth also abstracting out the error case here so it doesn't need to be handled by each child component? (Doesn't have to be in this PR)

Kevin-Chant (approved):
LGTM

PR #3276, opened 2024-11-08
Add Radix Select component and use Select for colors in TagsModal

msamper:
### Features and Changes

Adds the Radix Select component to our wrapped components and implements it within the `TagsModal` to pick colors

### Screenshots

Screenshot 2024-11-08 at 3 21 17 PM
Screenshot 2024-11-08 at 3 08 00 PM
Screenshot 2024-11-08 at 3 08 07 PM


packages/front-end/components/Radix/RadioGroup.tsx
```diff
@@ -64,7 +64,7 @@ export default function RadioGroup({
                   >
                     <Text className={disabled ? "rt-TextDisabled" : undefined}>
                       <Flex direction="column" gap="1">
-                        <Text style={{ fontWeight: 500 }} className="main-text">
+                        <Text weight="medium" className="main-text">
```
msamper, line 67: medium weight is 500

lukesonnet: ty! I think i landed this before the fonts were ready.

packages/front-end/components/Radix/Select.tsx
```diff
@@ -0,0 +1,75 @@
...
+      {error && (
+        <HelperText status={errorLevel} mt="1">
+          {error}
+        </HelperText>
+      )}
+    </Flex>
+  );
+}
+
+export function SelectItem({
```
lukesonnet, line 57: This is different from our current `SelectField` where you pass in `options: (SingleValue | GroupedValue)[];`. Is that our intent? Does this provide more control over styling or something but with more overhead when creating a Select component?

packages/front-end/pages/design-system/index.tsx
```diff
@@ -450,6 +452,52 @@ export default function DesignSystemPage() {
...
+          <Select
+            label="Select"
+            defaultValue="carrot"
+            value={selectValue}
+            setValue={setSelectValue}
+          >
+            <SelectItem value="apple">Apple</SelectItem>
+            <SelectItem value="carrot">Carrot</SelectItem>
+            <SelectSeparator />
+            <SelectItem value="apple-pie" disabled>
```
lukesonnet, line 467: Do you maybe want to add the text  "(disabled)" at the end? I couldn't figure out why I could select it when I picked it in the second dropdown (just because they use the same setter function). It also doesn't do the cursor thing to symbolize disabled like it does when the full selector is disabled, but maybe that's a Radix issue.

lukesonnet (approved):

packages/front-end/components/Radix/Select.tsx
msamper: Ah honestly maybe that's a better interface. I was just following the interface we created for the `DropdownMenu` radix component. This component allows you to disable individual items but that could a part of the `SingleValue` type. Only thing I'm not sure about with the interface we have for the current `SelectField` is how we would allow for separators.

lukesonnet: Yeah, you'd need to pass props like `disabled` from something like `SingleValue` to the Selector component.

Maybe `GroupedValue` is a way you can handle a selector, where you ask about selectors before or after or something in the grouped value interface.

IDK, making you compose directly when you implement the `Select` also seems reasonable, but it could lead to too many custom stylings. I'd get another opinion here if you want, tbh, but this can always be changed later before we use this component more widely.

msamper: I don't think we'd get any crazy styling because of the very basic interface I provide for the `SelectItem`, but maybe it's more verbose (?) I'm down to just merge this now just to get it in for `TagsModal` and revisit when we want to use it in more places

PR #3277, opened 2024-11-08
Add new Radix link component

ahdriel:
### Features and Changes

- Add a new `Link` component that merges `Radix/Link` and `next/link`.
- Add new `color` option `dark` which is used in the Get Started page.

### Screenshots

Note: I am hovering over the second item on both screenshots.

Screenshot 2024-11-08 at 3 36 05 PM
Screenshot 2024-11-08 at 3 35 57 PM

### DIFF
--- a/packages/front-end/components/Radix/Link.tsx
+++ b/packages/front-end/components/Radix/Link.tsx
@@ -1,0 +1,32 @@
+import clsx from "clsx";
+import NextLink from "next/link";
+import { LinkProps, Link as RadixLink } from "@radix-ui/themes";
+import styles from "./RadixOverrides.module.scss";
+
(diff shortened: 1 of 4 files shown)


bryce-fitzsimons:
It would be really helpful to either extend or clone this component to allow for text links with button functionality. Basically instead of wrapping a `<NextLink>`, wrap a `<a role="button" onClick=... >`

ahdriel:
@bryce-fitzsimons Good suggestion! I have updated to component to account for that. It'll use `button` if href is not provided.

I went with `button` as it has built-in support for accessibility features (like keyboard focusing and interactions) that `a role="button"` does not.

PR #3280, opened 2024-11-12
Disallow `parseInt`

romain-growthbook:
`parseInt` implementation is tricky and has multiple issues, in particular:

* It cannot parse numbers expressed in scientific notion such as `"5e3"`
* It converts anything that is given to it to string

Therefore:
```
> parseInt(0.000005);;
0
> parseInt(0.0000005);;
5
```

Because:
```
> String(0.000005);;
'0.000005'
> String(0.0000005);;
'5e-7'
```

And:
```
> parseInt('5e-7');;
5
> Number('5e-7');;
5e-7
```

You would think that typescript would prevent us from passing anything other than strings to it but:
* We don't check at runtime that the string is not a number expressed in scientific notation
* You'd be surprised about the amount of `<thing> + ""` that we blindly have to simply pass the type-check.

Thus, this PR mitigates the current code and prevents future issues by:
* Replacing all `parseInt` calls by calls to a safe wrapper `safeParseInt`
* Prohibits future use of `parseInt` via a custom `eslint` rule.


jdorn:
I'm generally in favor of this, but to get the equivalent of `parseInt`, we would need a 2nd step after `Number()` to make sure we end up with an integer and not a float.

toots:
> I'm generally in favor of this, but to get the equivalent of `parseInt`, we would need a 2nd step after `Number()` to make sure we end up with an integer and not a float.

Yep. I rushed to send for review and realized it.

jdorn (commented):
Revert all changes to the generated files.  There are 7000 lines changed there without any good reason.

PR #3285, opened 2024-11-12
Clean up orphaned feature rules on env deletion

Kevin-Chant:
### Features and Changes

Currently when an environment is deleted, any feature rules that were created for that environment are left intact. This adds a quick cleanup to delete them (to prepare for upcoming changes which will allow the creation of many more environments)

### Testing

Checked in a mongo shell to confirm deletion behavior:

1. Create a new environment `staging`
2. Create a rule on a feature in the `staging` environment
3. Run the following in a mongo shell
```
const environmentKey = `environmentSettings.staging`;
const query = {[environmentKey]: {$exists: true}}
db.features.find(query)
```
You should see the feature whose rule you just created.
4. Delete the environment in the UI or by API request, leaving `removeAssociatedFeatureRules` empty or false
5. Re-run `db.features.find(query)`; the feature should still be there
6. Re-create the environment `staging` in the UI
7. Run the delete again, this time setting `removeAssociatedFeatureRules` to true
8. Re-run `db.features.find(query)` and observe that there are no results

### Screenshots
image


packages/back-end/src/models/FeatureModel.ts
```diff
@@ -367,6 +367,24 @@ export async function deleteAllFeaturesForAProject({
...
+) {
+  const environmentKey = `environmentSettings.${envId}`;
+  const query = {
+    organization: context.org.id,
+    [environmentKey]: { $exists: true },
+  };
+
+  await FeatureModel.updateMany(query, {
+    $unset: { [environmentKey]: "" },
+  });
```
Kevin-Chant, line 385: @jdorn are we good with this sort of raw deletion? It doesn't leave an audit log or create a revision for the feature, but I was thinking it makes more sense to do it this way because it's a cascading delete from the environment itself. I'm also worried about performance if we do all that overhead work, but I figured I should run that past you

Kevin-Chant: Chatted w/ Jeremy about this - we're good with this being a direct delete (no audit, no revision) as long as it's an option (default false) for users to opt into.

gazzdingo (approved):
LGTM

PR #3286, opened 2024-11-12
Add event for new bandit rule

lukesonnet:


packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -443,12 +443,17 @@ export default function RuleModal({
...
         }
 
-        track("Create Experiment", {
-          source: "experiment-ref-new-rule-modal",
-          numTags: feature.tags?.length || 0,
-          numMetrics: 0,
-          numVariations: values.values.length || 0,
-        });
+        track(
+          values.experimentType === "multi-armed-bandit"
```
gazzdingo, line 447: could we just use the values.experimentType as the value since if we add a new Experiment type this will give false data?

packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -443,12 +443,17 @@ export default function RuleModal({
...
+            ? "CreateBandit"
```
bryce-fitzsimons, line 448:
```suggestion
            ? "Create Bandit"
```

lukesonnet: ok, so long as you fix it in the create modal as well

bryce-fitzsimons (approved):

packages/front-end/components/Features/RuleModal/index.tsx
lukesonnet: yes, standardizing this would be good, but it will require cleaning up old data. This is a hotfix to be consistent with the other usage.

gazzdingo: 😎  its a non blocking comment BTW just thought i would throw it out there

PR #3287, opened 2024-11-12
Remove getStartedPage class in favor of Radix Separator

ahdriel:
### Features and Changes

Small clean up in preparation for bigger changes in the Get Started page.
- Removed custom css class that customized `hr`
  - Using `Separator` now
  - Updated some `p` tags to `Box` to remove the default browser margin.

### DIFF
--- a/packages/front-end/components/GetStarted/GetStarted.module.scss
+++ b/packages/front-end/components/GetStarted/GetStarted.module.scss
@@ -10,12 +10,6 @@
   inherits: false;
 }
 
-.getStartedPage {
-  hr {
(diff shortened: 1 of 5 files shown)


gazzdingo (approved):
code Look good to me if you can add some screenshot that would be amazin

ahdriel:
@gazzdingo I skipped the screenshot on this case as there was literally no changes visually 😬  but here's the follow-up where bigger changes were made: https://github.com/growthbook/growthbook/pull/3289 and it has screenshots.

PR #3290, opened 2024-11-13
custom DatePicker component

bryce-fitzsimons:
image

image

image


packages/front-end/components/Events/EventsPage/EventsPage.tsx
```diff
@@ -224,44 +228,38 @@ export const EventsPageContainer = () => {
...
           label="To"
-          className="text-muted"
-          labelClassName="mr-2 mb-0"
-          containerClassName="ml-2 d-flex align-items-center mb-0"
-          value={toDate ? toDate.toISOString().split("T")[0] : ""}
-          onChange={(d) => {
-            setToDate(d.target.value ? new Date(d.target.value) : null);
-          }}
+          precision="date"
+          containerClassName=""
```
jdorn, line 247: Why not use the date range functionality here instead of 2 inputs?

jdorn (approved):

packages/front-end/components/Events/EventsPage/EventsPage.tsx
bryce-fitzsimons: The interval start/end dates are both optional, which doesn't play nicely with this widget. We could eventually add custom support for optional ranges

ahdriel (commented):
Clicking around I was able to make my end date be before my start date. Is this something we want to allow for a daterange or would it be better to say that a range always has the earlier date on the left?

Screenshot 2024-11-14 at 8 57 06 AM

packages/front-end/test/services/async-queue.test.ts
```diff
@@ -1,3 +1,4 @@
+import { vi } from "vitest";
```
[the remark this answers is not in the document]
bryce-fitzsimons: jest was unable to compile some of the ESM modules in the libraries that were updated (date-fns namely) as part of this change. it's a known issue with jest. vitest fixes it and is generally a better test engine.

PR #3299, opened 2024-11-15
Make test features work for scheduled rules

### LINKED ISSUE #1878: [Feature] Add a date and time to the "Test feature rules"
### Description of Feature

We love the "Test feature rule" function in GrowthBook. However, we often use scheduling to release features, especially for marketing. Currently, it is not possible to test the future behavior of a feature with "Test feature rule". It would be great if it would be possible to set the data and time in the simulation.

**Please use our custom dev time to work on this issue.**

Auz:
For features with scheduled rules, this PR adds a date picker to allow you to simulate how this feature would render at a particular date. 

Screenshot 2024-11-14 at 11 18 02 PM

Fixes #1878


packages/front-end/components/DatePicker/index.tsx
```diff
@@ -133,6 +135,7 @@ export default function DatePicker({
                 style={{ width: inputWidth, minHeight: 38, overflow: "clip" }}
               >
                 <Field
+                  id={id ?? ""}
```
bryce-fitzsimons, line 138: default should be `undefined`
```suggestion
                  id={id}
```

bryce-fitzsimons (approved):
approved with nit about id

packages/front-end/components/DatePicker/index.tsx
Auz: It was throwing a linting error without

PR #3305, opened 2024-11-15
added new radix colors

gazzdingo:
### Features and Changes
Super small PR to add Radix colors


packages/front-end/styles/theme.scss
```diff
@@ -257,6 +260,9 @@
   // radix color-scheme;
   --color-text-high: #ffffff;
```
ahdriel, line 261: We also have these variables in `theme-config.css`.

How should we merge/mix them all?
I believe the radix recommended way is that they provide all of the colors of the scale (`violet-1` through `violet-11`) via css variables.

bryce-fitzsimons: I had to move the light/dark mode variables from `theme-config.css` to `theme.scss` because the former was not reliably switching between light and dark mode when using "system default settings"

ahdriel: The Radix docs have a section on how NextJS does not preserve the import order which can create these issues with no specific root cause. I have [opened a PR](https://github.com/growthbook/growthbook/pull/3315) to update the NextJS version to one where this has been fixed.

After that lands I think we should consolidate all CSS variables in a single file, for both Radix overrides and also our own custom variables -- I can work on that.

gazzdingo: yeah that is a good idea i will wait for your PR then I will update this

PR #3312, opened 2024-11-18
Add button on superadmin page to create an inbuilt clickhouse datasource

tzjames:
### Features and Changes

- Adds button to create the clickhouse warehouse
- Adds tooltip
- Adds modal to ask if they are sure
- Got rid of check to make sure that it is only the main growthbook org on prod.
- Makes a superadmin check for adding the clickhouse warehouse

### Dependencies

- Might want to land fact table creation before using this.

### Testing

- Set `IS_CLOUD=true` and `SSO_CONFIG` in .env.local.
- Start server
- Go to /admin
- See the DB icons.
- Hover over one and see the tooltip message
- Click on the button and see the modal
- Click on Create
- See the button is grayed out
- Hover over and see the tooltip message

### Screenshots

Screenshot 2024-11-18 at 7 21 47 PM
Screenshot 2024-11-18 at 7 21 36 PM

### DIFF
--- a/packages/back-end/src/models/DataSourceModel.ts
+++ b/packages/back-end/src/models/DataSourceModel.ts
@@ -90,6 +90,22 @@ export async function getDataSourcesByOrganization(
     context.permissions.canReadMultiProjectResource(ds.projects)
   );
 }
+
+// WARNING: This does not restrict by organization
+export async function _dangerourslyGetAllDatasourcesByOrganizations(
+  organizations: string[]
+): Promise<DataSourceInterface[]> {
+  if (usingFileConfig()) {
+    return getConfigDatasources(organizations[0]);
+  }
+
+  const docs: DataSourceDocument[] = await DataSourceModel.find({
+    organization: { $in: organizations },
+  });
+
+  return docs.map(toInterface);
+}
+
 // WARNING: This does not restrict by organization
 export async function _dangerousGetAllGrowthbookClickhouseDataSources() {
   const docs: DataSourceDocument[] = await DataSourceModel.find({
--- a/packages/back-end/src/models/DataSourceModel.ts
+++ b/packages/back-end/src/models/DataSourceModel.ts
@@ -95,10 +95,6 @@ export async function getDataSourcesByOrganization(
 export async function _dangerourslyGetAllDatasourcesByOrganizations(
   organizations: string[]
 ): Promise<DataSourceInterface[]> {
-  if (usingFileConfig()) {
-    return getConfigDatasources(organizations[0]);
-  }
-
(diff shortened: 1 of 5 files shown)


packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -90,6 +90,22 @@ export async function getDataSourcesByOrganization(
     context.permissions.canReadMultiProjectResource(ds.projects)
   );
 }
+
+// WARNING: This does not restrict by organization
+export async function _dangerourslyGetAllDatasourcesByOrganizations(
+  organizations: string[]
+): Promise<DataSourceInterface[]> {
+  if (usingFileConfig()) {
+    return getConfigDatasources(organizations[0]);
```
Kevin-Chant, line 99: Is this a valid assumption that `organizations` will only have one org when using file config? Or does this need to loop through all the orgs in the list

packages/front-end/pages/admin.tsx
```diff
@@ -176,6 +207,35 @@ function OrganizationRow({
...
           </a>
         </td>
+        <td className="p-0 text-center">
+          <Tooltip
+            body={
+              hasGrowthbookClickhouse
+                ? "Already has an Inbuilt Growthbook Datasource"
+                : "Create Inbuilt Growthbook Clickhouse DataSource"
+            }
+          >
```
Kevin-Chant, line 217: Not for this PR, but have we considered adding these ternaries under an `enabled/disabled` prop to `Tooltip`? Seems like that could be useful in other places too

Kevin-Chant (commented):
One question, the rest looks good to me but want to have others weigh in before we ship it

packages/back-end/src/models/DataSourceModel.ts
tzjames: good catch.  This doesn't belong here as it doesn't apply to cloud, and this is only used on cloud, I've gotten rid of it.

packages/front-end/pages/admin.tsx
tzjames: Not sure if that is really needed.  In this case we want different messages if what it is wrapping is disabled or not.  Would we just ignore the `body` property in your suggestion?  The ternary seems simple enough.

Kevin-Chant (approved):
Lgtm

packages/front-end/pages/admin.tsx
Kevin-Chant: Maybe it's overkill, but i was thinking of it having two props for each of the items with a ternary. So an enabledBody, disabledBody etc. On second thought it probably clutters things more than it helps

PR #3317, opened 2024-11-19
fix experiment variation editor / variations.length mismatch

bryce-fitzsimons:
- Improve component UI/UX, especially when used in a "Edit Variations" context
- Ensure adding/removing variations does not corrupt phase's variationWeights
- hide editing "id" field unless necessary

https://www.loom.com/share/dccbdceb5e914521906783ce72e83bda?sid=c7b6dd4d-0239-444a-8cf0-0683c8c20617

image
Screenshot 2024-11-19 at 9 35 55 PM

todo:

- [x] check EditTargetingModal
  - [x] safe mode
  - [x] Make Changes
- [x] NewPhaseForm
- [x] BanditRefNewFields

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -777,6 +777,7 @@ export async function postExperiment(
       currentPhase?: number;
       phaseStartDate?: string;
       phaseEndDate?: string;
+      variationWeights?: number[];
     },
     { id: string }
   >,
@@ -1009,6 +1010,16 @@ export async function postExperiment(
         ...changes,
       } as ExperimentInterface);
     }
+  }
+
+  if (data.variationWeights) {
+    const phases = [...experiment.phases];
+    const lastIndex = phases.length - 1;
+    phases[lastIndex] = {
+      ...phases[lastIndex],
+      variationWeights: data.variationWeights,
+    };
+    changes.phases = phases;
   }
 
   // Only some fields affect production SDK payloads
--- a/packages/front-end/components/Experiment/EditVariationsForm.tsx
+++ b/packages/front-end/components/Experiment/EditVariationsForm.tsx
@@ -50,19 +50,35 @@ const EditVariationsForm: FC<{
         data.variations = [...data.variations];
 
         // fix some common bugs
-        const newWeights = [
-          ...data.variations.map((_, i) =>
-            Math.min(
-              Math.max(
-                data.variationWeights?.[i] ?? 1 / data.variations.length ?? 0.5,
-                0
-              ),
-              1
-            )
-          ),
-        ];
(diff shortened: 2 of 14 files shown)


packages/back-end/src/controllers/experiments.ts
```diff
@@ -1011,6 +1012,16 @@ export async function postExperiment(
...
 
+  if (data.variationWeights) {
+    const phases = [...experiment.phases];
+    const lastIndex = phases.length - 1;
+    phases[lastIndex] = {
+      ...phases[lastIndex],
+      variationWeights: data.variationWeights,
+    };
+    changes.phases = phases;
+  }
```
lukesonnet, line 1023: Should we do any validation in the controller to prevent this from happening to running experiments or something like that? Feels like this exposes us to some risk if someone just starts using `postExperiment` without knowing they could be affecting the prod weights? But maybe that's just the price we have to pay.

packages/front-end/components/Experiment/EditVariationsForm.tsx
```diff
@@ -54,6 +49,37 @@ const EditVariationsForm: FC<{
...
         data.variations = [...data.variations];
 
+        // fix some common bugs
+        if (!isBandit) {
+          const newWeights = [
+            ...data.variations.map((_, i) =>
+              Math.min(
+                Math.max(
+                  data.variationWeights?.[i] ??
+                    1 / data.variations.length ??
```
lukesonnet, line 59: What's the order of operations on `??` Seems like hopefully it's after division and such, in which case I think this makes sense? Might wnat to consider some parentheses for people like me.

lukesonnet (approved):

packages/back-end/src/controllers/experiments.ts
bryce-fitzsimons: I think its the price we pay for now at least. We have lots of these partial edit widgets that use the same endpoint.

packages/front-end/components/Experiment/EditVariationsForm.tsx
bryce-fitzsimons: good call

packages/back-end/src/controllers/experiments.ts
lukesonnet: Yeah, this just seems like a critical one that affects the actual `feature`, which in general we should strive to put more safety around, rather than less. It's already confusing when you make an experiment change and it auto-affects a feature people thought was approval flow controlled.

bryce-fitzsimons: i think we should put effort into locking the controller logic (not just for this widget). i can take a look at that in a follow up pr

PR #3318, opened 2024-11-19
Clean up warehouse track event props

jdorn:
### Features and Changes

- Add new GrowthBook attributes:
  - `role`
  - `cloudOrgId`
  - `configFile`
  - `usingSSO`
  - `buildSHA`
  - `buildDate`
  - `buildVersion`
- Removed GrowthBook attributes with potential PII
  - `userAgent`
  - `name`
  - `company`
- Fixed value of `accountPlan` GrowthBook attribute
- Made warehouse track send GrowthBook attributes as `user_attributes_json`.  `properties_json` now only contains explicit event properties.
- Get actual SDK version instead of hard-coding
- Hard-code ingestor endpoint with ability to override via env var
- Enable ingestor endpoint for self-hosted (and respect DISABLE_TELEMETRY setting)
- Change event name `page-load` to `Page View` (keep old name for Jitsu for backwards compatibility)
- Remove `Organization Loaded` and `App Load` events
- New `Session Start` event
- Fix/improve default exposure queries and fact table definitions for built-in clickhouse


packages/front-end/services/track.ts
```diff
@@ -199,22 +205,20 @@ export default function track(
...
     sdk_language: "react",
-    // TODO: programmatically get sdk version. Importing from _app breaks tests
-    sdk_version: "1.2.0",
+    sdk_version: growthbook.version,
     url: trackProps.url,
     user_id: id,
-    user_attributes_json: "{}",
+    user_attributes_json: JSON.stringify({
+      ...growthbook.getAttributes(),
+      cloudOrgId: isCloud() ? org : "",
```
lukesonnet, line 218: Why does this go in user_attributes_json when it isn't an attribute?

lukesonnet (approved):

packages/front-end/services/track.ts
jdorn: I know for debugging purposes, it's nice to have the actual org id on Cloud instead of just a hashed version.  I want us to continue using the hashed id for targeting, so was a little hesitant to add this as an attribute.

packages/front-end/services/track.ts
```diff
@@ -81,15 +81,21 @@ const SESSION_ID_COOKIE = "gb_session_id";
 const pageIds: Record<string, string> = {};
 
 const dataWareHouseTrack = async (event: DataWarehouseTrackedEvent) => {
-  if (!dataWarehouseUrl) return;
+  if (inTelemetryDebugMode()) {
+    console.log("Telemetry Event - ", event);
+  }
+  if (!isTelemetryEnabled()) return;
```
tzjames, line 87: This line means we no longer send the event when in debug mode. I think we still want to send the telemetry as we are debugging, at we are trying to output failed requests in debug mode below on line 101.

PR #3320, opened 2024-11-19
Hide additional tabs when Experiment is in draft state

ahdriel:
### Features and Changes

When a experiment is in draft state and there are not previous analysis done, the `Results` and `Health` tab do not contain any data. So we can simplify the user experience by showing them only the `Overview` tab while the Experiment is being setup.

| Before | After |
|--------|--------|
| Screenshot 2024-11-19 at 11 16 03 AM | Screenshot 2024-11-19 at 11 15 47 AM |


packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -541,103 +541,105 @@ export default function ExperimentHeader({
...
-                    </>
-                  }
-                  anchor="results"
-                  onClick={() => setTab("results")}
-                  newStyle={false}
-                  activeClassName="active-tab"
-                  last={false}
-                />
-                {isBandit && (
+      {experiment.status !== "draft" ? (
```
jdorn, line 544: It's possible to revert a running experiment back to a draft.  In that case, there would be results and we would want to show the tabs.  We should change this conditional to check for draft AND no results yet.

bryce-fitzsimons: also its possible there could be results in previous phases but not the current phase, and they're wanting to look back at previous results

ahdriel:
@bryce-fitzsimons I think that's okay for now so we don't add more complexity. There's nothing harmful if users access the page via the URL directly.

bryce-fitzsimons (approved):

PR #3322, opened 2024-11-19
Environment inheritance/forking

Kevin-Chant:
### Features and Changes

Updates the `Environment` interface to include a new field: `parent`. An environment with a parent will inherit (recursively if needed) all of its Feature and FeatureRevision rules when there are no definitions for that environment. This is handled via our JIT migrations so it always appears like the Features have the correct `environmentSettings` and is only persisted to the db on a write.

### Testing

Here's a few curl requests I'm using to test

```
curl --request POST \
  --url http://localhost:3100/api/v1/environments \
  --header 'content-type: application/json' \
  --header 'Authorization: Bearer <token>' \
  --data '{
  "parent": "production",
  "id": "staging",
  "synchronous": true
}'
```

```
curl --request DELETE \
  --url http://localhost:3100/api/v1/environments/staging \
  --header 'content-type: application/json' \
  --header 'Authorization: Bearer <token>' \
```


packages/back-end/src/api/openapi/paths/postForkEnvironment.yaml
```diff
@@ -0,0 +1,54 @@
...
+        properties:
+          id:
+            type: string
+            description: The ID of the new environment
+          forkBase:
+            type: string
+            description: The ID of the environment to fork from
+          synchronous:
+            type: boolean
+            description: Whether this fork should be run synchronously. If true, the server may take a long time to respond, particularly if there are many features with rules in the forkBase environment.
```
Kevin-Chant, line 21: I'd like some input on this param. I opted to let the user decide whether they want this to be sync or async because the performance will depend heavily on how many features they have with rules to be forked. I'd like to default it to sync but that could lead to unacceptably slow performance for this endpoint. Ideally when running it async there'd be a way to check the progress but I also felt like adding that would be overcomplicating this feature.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -640,6 +658,20 @@ export async function archiveFeature(
...
+  context: ReqContext | ApiReqContext,
+  feature: FeatureInterface,
+  sourceEnvironment: string,
+  destinationEnvironment: string
+) {
+  return await updateFeature(context, feature, {
+    environmentSettings: {
+      ...feature.environmentSettings,
+      [destinationEnvironment]: feature.environmentSettings[sourceEnvironment],
+    },
```
Kevin-Chant, line 671: This doesn't copy any rules in a current draft state but publishing that draft also doesn't clobber or otherwise affect the forked environment. That seems like reasonable and safe behavior to me but it's worth mentioning.

Kevin-Chant: One TODO for me before this PR is ready though - this leaves the settings in a weird state where we hit a sanity check error when fetching the feature later
```
`Published revision rules.${env} does not match feature ${org.id}.${feature.id}`
```
it still doesn't affect behavior as sdk connections and the feature page both show the correct values, but it's still worth fixing

Kevin-Chant: Man this was a convoluted behavior to untangle. I ended up actually making a new draft based on the current draft and then automatically publishing it so this won't happen anymore.

packages/back-end/src/api/features/postFeature.ts
```diff
@@ -175,6 +175,7 @@ export const postFeature = createApiRequestHandler(postFeatureValidator)(
       feature.id
     );
     const revision = await getRevision(
+      req.context,
```
gazzdingo, line 178: since we are updating the Params i feel like we should move it over to an Object like
```getRevision({
    context: req.context,
    org: feature.....
})```

File: packages/back-end/src/models/FeatureRevisionModel.ts [hunk 4]
```diff
@@ -90,10 +100,12 @@ export async function getRevisions(
     .limit(25);
```
gazzdingo, line 100: when moving to another env do we want to remove the limit so that all revisions are passed over? or we might want to add something in the UI that will state only the previous 25 will be moved

gazzdingo (approved):
LGTM :shipit:

PR #3324, opened 2024-11-20
Update Projects list to use plain text instead of Badges

ahdriel:
Using badges for listing the Projects can add a lot of noise to tables and other places where we render them, so we want to reduce its impact by using plain text instead.

### Changes

* `ProjectBadges` renamed to `ProjectNames`
* The component now renders either `All Project` or a list of project names separated by comma
* It does not use Badges anymore

| Before | After |
|--------|--------|
| Screenshot 2024-11-25 at 2 18 20 PM | Screenshot 2024-11-25 at 2 18 04 PM |


mknowlton89 (approved):
Tested locally and looks great!

PR #3325, opened 2024-11-20
Update global link colors to violet

ahdriel:
### Features and Changes

Updating all links that are using Bootstrap blue to use Radix violet.

### DIFF
--- a/packages/front-end/styles/_bootstrap-theme-overrides.scss
+++ b/packages/front-end/styles/_bootstrap-theme-overrides.scss
@@ -4,10 +4,11 @@
 
 a,
 .btn-link {
-  color: var(--text-link-color);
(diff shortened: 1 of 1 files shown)


packages/front-end/styles/_bootstrap-theme-overrides.scss
```diff
@@ -4,10 +4,11 @@
 
 a,
 .btn-link {
```
bryce-fitzsimons, line 6: should probably also change `link-purple` and `text-purple`

ahdriel: Got caught up on some other work, but I checked and now we are good.

bryce-fitzsimons (approved):

PR #3330, opened 2024-11-20
Make schema browser robust to switching databases

lukesonnet:
### Features and Changes

If you changed the dataset for a datasource that already had a schema browser, it would struggle to update the old schemas since they no longer exist in the new datasource. This makes that process a bit safer by only trying to do the update logic on an old schema if it can find it

Before (when I switched datasources and the information schema had stale schemas that were not updating)
Screenshot 2024-11-20 at 11 08 59 AM

After (shows only the new, appropriate schema)
Screenshot 2024-11-20 at 11 29 10 AM

### DIFF
--- a/packages/back-end/src/services/informationSchema.ts
+++ b/packages/back-end/src/services/informationSchema.ts
@@ -26,10 +26,11 @@ export function getRecentlyDeletedTables(
       (updatedInformationSchemaRecord) =>
         updatedInformationSchemaRecord.databaseName === database.databaseName
     );
+    if (!database.schemas) return;
     database.schemas.forEach((schema) => {
       const correspondingSchemaIndex = updatedInformationSchema[
         correspondingIndex
-      ].schemas.findIndex(
+      ]?.schemas.findIndex(
         (updatedSchemaRecord) =>
           updatedSchemaRecord.schemaName === schema.schemaName
       );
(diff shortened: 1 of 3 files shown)


packages/back-end/src/services/informationSchema.ts
```diff
@@ -26,10 +26,11 @@ export function getRecentlyDeletedTables(
       (updatedInformationSchemaRecord) =>
         updatedInformationSchemaRecord.databaseName === database.databaseName
     );
+    if (!database.schemas) return;
```
lukesonnet, line 29: Seems unnecessary but... whatever

packages/front-end/components/SchemaBrowser/SchemaBrowser.tsx
```diff
@@ -299,12 +298,14 @@ export default function SchemaBrowser({
         </>
       </SchemaBrowserWrapper>
       {error && <div className="alert alert-danger mt-2 mb-0">{error}</div>}
-      <DatasourceTableData
-        canRunQueries={canRunQueries}
-        tableId={currentTable}
-        datasourceId={datasource.id}
-        setError={setError}
-      />
+      {currentTable ? (
```
lukesonnet, line 301: prevents 404 in DatasourceTaleData when currentTable is ""

packages/back-end/src/services/informationSchema.ts
```diff
@@ -88,21 +89,21 @@ export async function mergeStaleInformationSchemaWithUpdate(
     database.schemas.forEach((schema) => {
       const correspondingSchemaIndex = staleInformationSchema[
         correspondingIndex
-      ].schemas.findIndex(
+      ]?.schemas.findIndex(
```
mknowlton89, line 92: I'm wondering if in addition to adding these optional chaining operators, we update the early return logic so it also early returns if correspondingIndex === -1.

It shouldn't change any behavior, but I think it could make for a better dev experience since it'd be more clear that if there isn't a correspondingIndex, we don't need to look any further.

E.G. Update the early return on line 88 from:

```ts
if (!database.schemas) return;
```
to 

```ts
if (!database.schemas || correspondingIndex === -1) return;
```

lukesonnet: Yeah, let's do both.

mknowlton89 (approved):

packages/back-end/src/services/informationSchema.ts
```diff
@@ -84,25 +85,25 @@ export async function mergeStaleInformationSchemaWithUpdate(
...
       );
 
       if (correspondingSchemaIndex > -1) {
         schema.dateCreated =
-          staleInformationSchema[correspondingIndex].schemas[
+          staleInformationSchema[correspondingIndex]?.schemas[
             correspondingSchemaIndex
           ].dateCreated;
       }
       if (!schema.tables) return;
```
mknowlton89, line 103: I think it might make sense to update the early return here as well if `correspondingSchemaIndex === -1`.

PR #3338, opened 2024-11-21
cleanup experiment variation editor value

bryce-fitzsimons:
- don't send "value" to endpoint (unused field)
- validate key on save

### DIFF
--- a/packages/front-end/components/Experiment/EditVariationsForm.tsx
+++ b/packages/front-end/components/Experiment/EditVariationsForm.tsx
@@ -115,16 +115,14 @@ const EditVariationsForm: FC<{
           form.setValue(
             "variations",
             v.map((data) => {
-              // eslint-disable-next-line @typescript-eslint/no-explicit-any
-              const newData: any = { ...data };
-              newData.key = data.value;
-              delete newData.value;
+              const { value, ...newData } = data;
               return {
                 // default values
                 name: "",
                 description: "",
                 screenshots: [],
                 ...newData,
+                key: value,
               };
             })
           );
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/EditVariationsForm.tsx
```diff
@@ -111,14 +114,17 @@ const EditVariationsForm: FC<{
         setVariations={(v) => {
           form.setValue(
             "variations",
-            v.map((data, i) => {
+            v.map((data) => {
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ahdriel, line 118: nit: We can do this to make TypeScript happy

```
const { value, ...newData } = data;
return {
  // default values
  name: "",
  description: "",
  screenshots: [],
  ...newData,
  key: value,
};
```

packages/front-end/components/Experiment/EditVariationsForm.tsx
```diff
@@ -111,14 +114,17 @@ const EditVariationsForm: FC<{
...
+              const newData: any = { ...data };
+              newData.key = data.value;
```
ahdriel, line 120: question: Why does the key come named as `value` here? Is it possible to change that to be `key` before here so we don't have to do this renaming? Unsure about the level of effort tho

ahdriel (approved):

packages/front-end/components/Experiment/EditVariationsForm.tsx
bryce-fitzsimons: it's because the widget is also used for feature rules which do not have a key and instead use value to achieve the same thing.

packages/front-end/components/Experiment/EditVariationsForm.tsx
bryce-fitzsimons: good suggestion, thanks. fixed

PR #3341, opened 2024-11-21
Added sortable description to experiment listing page

Auz:
Screenshot 2024-11-21 at 3 50 46 PM


packages/front-end/pages/experiments/index.tsx
```diff
@@ -294,6 +295,7 @@ const ExperimentsPage = (): React.ReactElement => {
   }
 
   const hasExperiments = experiments.length > 0;
+  const hasDescription = experiments.some((e) => e.description);
```
mknowlton89, line 298: I like this logic - I think we should apply it to some of the other columns (tags and project, namely).

mknowlton89 (approved):

packages/front-end/pages/experiments/index.tsx
```diff
@@ -453,9 +455,14 @@ const ExperimentsPage = (): React.ReactElement => {
...
                     <th></th>
-                    <SortableTH field="name" className="w-100">
+                    <SortableTH field="name" className="nowrap">
                       Experiment
                     </SortableTH>
+                    {hasDescription && (
+                      <SortableTH field="description" className="">
+                        Description
+                      </SortableTH>
+                    )}
```
jdorn, line 465: I don't think this column should be shown by default.  The table is already really busy.  What if we add a cog icon to configure the table so users can add in additional columns.  Would get us pretty close to a full advanced search without cluttering the UI for others.

PR #3344, opened 2024-11-22
Project-scope segments

mknowlton89:
### Features and Changes

This PR adds project scoping to GrowthBook `Segments`, which were formally globally scoped only.

Now, when creating (or editing existing Segments) a user within an organization with permission to do so, can add Segments to various projects.

### A few highlights of the logic introduced with this PR:
- Given that project-scoping Segments is new, we've got slightly different validation rules for creating Segments vs editing an existing Segment.

When creating a Segment, we have strict rules about which projects can/cannot be set on the Segment. If a Segment is built upon a Data Source that is in at least 1 projects, the Segment being created cannot be in "All Projects" (aka have no projects). Only if the connected data source is in "All Projects" can the Segment be in "All Projects". Furthermore, if the connected Data Source is in Projects A & B, for example, the Segment must be placed in either A or B, or A & B. The Segment can't be in Project C.

However, the rules for editing are less strict. Since project-scoping Segments is new, all existing Segments will be placed in "All Projects" when this PR is merged, since an empty (or undefined) projects array is evaluated to "All Projects". As such, when editing an existing Segment, we won't apply the same strict validation rules as outlined above. This means an org will be able to update an existing Segment and if the connected Data Source is in Projects A & B, they org will be able to save a Segment if it's still in "All Projects".

Additionally, there is the case where an organization creates a Data Source, then creates a Segment, and later updates the Data Source's project list. In that case, we don't apply automatic updates to the child Segment's project list, so it is possible for the Data Source's project list to be a mismatch of the Segments. Ex: A Data Source and a child Segment is originally in Project A & B, but later down the road the Data Source is removed from Project B. We won't automatically update the child Segment's projects to remove it from Project B. The Segment will stay in Project B - and the organization can continue to update the Segment as needed without strict validation. However, if the organization ever removes the Segment from Project B, they will then lose the ability to ever add it back to Project B, unless the parent Data Source is first added back to Project B.

#### Happy Paths (Creating):
- [x] Ensure that you can create a new SQL Segment via a Data Source in "All Projects" and set the new SQL Segment in "All Projects".
- [x] Ensure that you can create a new SQL Segment via a Data Source in "All Projects" and set the new SQL Segment in multiple projects (Ex: Project A and B) - so long as the user has permission to `createSegments` in both of those projects.
- [x] Ensure that you can create a new Fact Segment via a Data Source in "All Projects" and set the new SQL Segment in "All Projects".
- [x] Ensure that you can create a new Fact Segment via a Data Source in "All Projects" and set the new SQL Segment in multiple projects (Ex: Project A and B) - so long as the user has permission to `createSegments` in both of those projects.
- [x] Ensure that the "Projects" drop-down for new Segments (both Fact and SQL) are a subset of the parent data source's projects AND only projects where the user has `createSegments` permission.

#### Error/Edge Paths (Creating):
- [x] Ensure that you cannot create a SQL or a Fact Segment in "All Projects" if the connected Data Source isn't also in "All Projects" - ensure that an error is thrown on submit (**Screenshot A below**)
- [x] Ensure that you cannot create a SQL or a Fact Segment in Project C if the connected Data Source isn't also in Project C - Project C shouldn't be in the project drop-down list at all
- [x] Ensure that if a Segment is in Projects A and B, and the data source was recently updated removing it from Project B, next time the user edits the Segment, it'll still be in Projects A and B, and updating the Segment (ex: updating the description) doesn't change that. But, if the user removes Project B from the Segment's project list, they can not add it back unless the connected data source is first added back to Project B.

#### Happy Paths (Updating):
- [x] Ensure that you can update an existing SQL Segment that is in "All Projects", regardless of the parent data source's project list
- [x] Ensure that you can update an existing Fact Segment that is in "All Projects" regardless of the parent data source's project list
- [x] Ensure that you can update an existing SQL Segment that is in "All Projects", and add it to a subset of the connected data source's project list
- [x] Ensure that you can update an existing Fact Segment that is in "All Projects", and add it to a subset of the connected data source's project list

#### Permissions Testing:
- [x] Ensure that we filter out projects on the segments list view if a user doesn't have read-access to at least 1 of the segment's project
- [x] Ensure that if a segment is in multiple projects, one of which the user doesn't have read access to, we display the tooltip next to the project pills on the list view that outlines this segment is in a project that the user doesn't have access to.
- [x] Ensure that the `Create Segment` CTA is only displayed if the org isn't using the config file and the user has permission to create segments in the currently selected project (if no project is selected, the user must have the permission globally)

#### Screenshots
Screenshot A:
Screenshot 2024-11-22 at 9 42 21 AM


packages/front-end/components/Segments/FactSegmentForm.tsx
```diff
@@ -55,12 +59,32 @@ export default function FactSegmentForm({
...
+        : filteredDatasources[0]?.projects || [],
     },
   });
 
   const datasource = getDatasourceById(form.watch("datasource"));
   const factTable = getFactTableById(form.watch("factTableId"));
 
+  // Projects must be a subset of a data source's projects
+  const filteredProjects = projects.filter((project) => {
+    // only filter projects is the data source isn't in All Projects (aka, projects is an empty array)
```
Kevin-Chant, line 73:
```suggestion
    // only filter projects if the data source isn't in All Projects (aka, projects is an empty array)
```

packages/front-end/components/Segments/SegmentForm.tsx
```diff
@@ -60,6 +69,22 @@ const SegmentForm: FC<{
...
+    if (datasource?.projects && datasource.projects.length) {
+      return (
+        datasource.projects.includes(project.id) ||
+        form.watch("projects").includes(project.id)
+      );
+    }
+  });
+
+  const projectOptions = useProjectOptions(
+    (project) => permissionsUtil.canCreateMetric({ projects: [project] }), //TODO: Do we need to check here for the create vs update permission?
```
Kevin-Chant, line 83: Did you get an answer for this question? From my reading of things it's correct to use canCreate but maybe you found a reason that isn't right

Kevin-Chant (commented):
Gonna do some manual testing now, one question about a lingering todo otherwise the code looks good

Kevin-Chant (approved):
LGTM

PR #3351, opened 2024-11-23
Shareable reports

bryce-fitzsimons:
New snapshot-based reports with share capabilities.
- All new reports will be snapshot-based instead of custom model (with legacy report support)
- New public report route (`/r/:tinyId`) that can optionally show a public, server-rendered version of a report
- Share controls: public (anonymous), organization, private [default: organization]
- View/listing controls: public (published/listed), private
- OpenGraph support for pasting links into Slack / Discord / social media

private link:
image

public link [anonymonus]:
image

public link [logged in]:
image

share controls:
image

todo:

- [x] listing page logic
- [x] tracking
- [x] exp page buttons?
- [x] OG preview (images?)


lukesonnet:
Hey Bryce! Took it for a test drive and just clicked around and tried to do as much as I could. Here are some notes, roughly in order of my perceived priority of fixing/changing:

- [ ] Existing reports are not appearing below experiments. I definitely have reports attached to this experiment that I was able to find by clicking my user name in the top right and going to "My Reports" but it wasn't below the experiment. My new reports end up down here, but old ones are missing.
image

- [ ] I think Activation Metric belongs near the In-Progress conversions or the Segment on the analysis page, not on the metrics page. Let's leave the metrics page about what you want to measure. It also shouldn't necessarily live right next to conversion window override, so it's another reason to move activation metric.
Screenshot 2024-12-17 at 7 42 16 AM

- [ ] It would be cool to get conversion overrides in here as it is reports also serve as a place to play with settings, but agree it's not critical for the use case we're considering.

- [ ] This All Metrics -> Goal Metrics -> Metric Name flow for dimension breakdowns (here and in the main experiment results) is super clunky. It's probably out of scope, but folks will use reports to share dimensions so it may be worth fixing now. Just removing some text somehow would be sufficient. I'd even just get rid of "All Metrics" here.
Screenshot 2024-12-17 at 7 55 32 AM


- [ ] This "updated" status being ephemeral is a bit... odd of a pattern. It flickered away before I could see what it was. I personally think we should consider using the gsuite pattern that is very familiar and a bit safer. If setting changes, say it is pending and make them save. A secondary CTA lets you copy the link, and we should remove `Close`. They already have the X in the top right and can just click away to close. Copy Link being the main CTA here is very counter intuitive.
Screenshot 2024-12-17 at 8 00 27 AM
Screenshot 2024-12-17 at 8 02 01 AM


- [ ] This cancel button doesn't cancel queries (expected for many DBs that we can't actually send cancel signals to), but it doesn't even respect my request to stop refreshing and keep the existing report.
Screenshot 2024-12-17 at 7 54 59 AM

- [ ] Notebooks are a bit screwed up, I'll take a look at this myself.

- [ ] I personally find "published" a little bit unclear as it could mean public unless i see it side-by-side with public. Should we add tooltips to the badges? "Only visible to you", "Visible to anyone in your organization", "Visible to anyone with the url". I feel that kind of language is standard for sharing. You learn this if you click share but a tooltip might be reasonable.
Screenshot 2024-12-17 at 7 34 47 AM

- [ ] Wanna put the published etc. badges down here below the experiment? and replace the "Status" field with "Visibility" or something on the Reports page and use the badge?
image
Screenshot 2024-12-17 at 7 59 20 AM

- [ ] Also, the pages/reports.tsx doesn't seem to have a a header so it shows up as `GrowthBook > `. Let's call it Reports!

- [ ] The three dot menu doesn't appear to use the radix style? it is grey all across when you hover over an item, not purple and with a margin all around?
Screenshot 2024-12-17 at 7 37 07 AM

- [ ] nit: I feel this is too prominent and one more different stylistic thing to process in the top right of the new report page. Can it just be combined with the "last updated" text? IDK, this is super minor. Maybe it should live next to the difference type stuff?
Screenshot 2024-12-17 at 7 36 33 AM

packages/back-end/src/controllers/reports.ts
```diff
@@ -166,6 +232,147 @@ export async function getReport(
...
+  const report = await getReportByTinyid(tinyid);
+  if (!report) {
+    throw new Error("Unknown report id");
+  }
+  if (report.type !== "experiment-snapshot") {
+    throw new Error("Invalid report");
+  }
+  const context = await getContextForAgendaJobByOrgId(report.organization);
+
+  if (report.shareLevel === "private") {
```
bryce-fitzsimons, line 249: also block org for non-auth users

packages/back-end/src/controllers/reports.ts
```diff
@@ -166,6 +232,147 @@ export async function getReport(
   });
 }
 
+export async function getReportPublic(
+  req: Request<{ tinyid: string }>,
+  res: Response
+) {
+  const { tinyid } = req.params;
```
bryce-fitzsimons, line 239: use long uuid

packages/back-end/src/jobs/expireOldQueries.ts
```diff
@@ -63,6 +63,7 @@ const expireOldQueries = trackJob(JOB_NAME, async () => {
   const reports = await findReportsByQueryId([...queryIds]);
   for (let i = 0; i < reports.length; i++) {
     const report = reports[i];
+    if (report.type !== "experiment") continue;
```
bryce-fitzsimons, line 66: double check canceling queries

packages/front-end/components/Radix/Badge.tsx
```diff
@@ -8,6 +9,7 @@ type Props = {
   color?: RadixColor;
   variant?: "solid" | "soft";
   radius?: "none" | "small" | "medium" | "large" | "full";
+  style?: CSSProperties;
```
bryce-fitzsimons, line 12: remove

packages/front-end/components/Radix/DropdownMenu.tsx
```diff
@@ -8,34 +8,51 @@ type AllowedChildren = string | React.ReactNode;
...
+  disabled?: boolean;
 } & MarginProps;
 
 export function DropdownMenu({
   trigger,
+  triggerClassName,
   menuPlacement = "start",
   menuWidth,
   children,
+  color,
```
bryce-fitzsimons, line 28: needed?

packages/front-end/components/Radix/Select.tsx
```diff
@@ -60,7 +61,7 @@ export function SelectItem({
   disabled = false,
 }: {
   value: string;
-  children: string;
+  children: string | string[] | ReactNode;
```
bryce-fitzsimons, line 64: undo

lukesonnet (approved):
by proxy

PR #3355, opened 2024-11-25
Support legacy "inline experiment" rule editing

bryce-fitzsimons:
As part of redesign work, we removed the ability to edit existing legacy experiment rules (which users haven't been able to create for about 2 years). This adds back some minimal editing and display capabilities.

image
image

### DIFF
--- a/packages/front-end/components/Features/RuleModal/index.tsx
+++ b/packages/front-end/components/Features/RuleModal/index.tsx
@@ -728,7 +728,7 @@ export default function RuleModal({
         }
         step={step}
         setStep={setStep}
-        hideNav={ruleType !== "experiment-ref-new"}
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -795,7 +795,9 @@ export default function RuleModal({
           />
         ) : null}
 
-        {ruleType === "experiment-ref-new" && experimentType === "experiment"
+        {(ruleType === "experiment-ref-new" &&
+          experimentType === "experiment") ||
+        ruleType === "experiment"
```
ahdriel, line 800: question: what's the different between `experiment`, `experiment-ref` and `experiment-ref-new`? For my learning in understanding how the codebase has evolved

ahdriel (approved):

PR #3357, opened 2024-11-26
JS SDK 1.3.1: New scoped instances for Node.js

jdorn:
### Features and Changes

- Rename `GrowthBookMultiUser` to `GrowthBookClient`
- New `GrowthBookClient.getScopedInstance(userContext)` method to make it easier to use in express-like back-end frameworks and client-side environments
- Updated code snippets in-app to use GrowthBookClient for Node.js by default
- Brand new NodeJS docs using GrowthBookClient by default


docs/docs/lib/node.mdx
```diff
@@ -21,20 +21,86 @@ npm install --save @growthbook/growthbook
 
 ## Quick Usage
 
-GrowthBook instances are scoped to a single incoming request. The easiest way to do this is with Middleware:
+First, create and instantiate a singleton GrowthBook client. You can do this as part of your server startup script.
+
+```js
+const client = new GrowthBookMultiUser({
```
royalfig, line 27: Would it be better to call `client` something more explicit like `gbClient`? I'm thinking of copying/pasting this code into a large app where `client` might already be defined and where it might be clearer what this client does.

docs/docs/lib/node.mdx
```diff
@@ -21,20 +21,86 @@ npm install --save @growthbook/growthbook
...
+  }
+
+  // Create a scoped GrowthBook instance and store in req
+  req.growthbook = client.getScopedInstance(userContext);
+
+  next();
+});
+```
+
+Now, you can use this scoped instance from any route without needing to pass in userContext again.
```
royalfig, line 75: userContext formatting -> `userContext`

docs/docs/lib/node.mdx
```diff
@@ -21,20 +21,86 @@ npm install --save @growthbook/growthbook
...
+}
+
+// String, Number, or JSON flags
+const value = client.getFeatureValue("my-string-feature", "fallback", userContext);
+console.log(value);
+```
+
+### Express Middleware
+
+Instead of constantly defining and passing `userContext` around your app, we recommend creating a "scoped" instance of GrowthBook in a middleware instead.
```
royalfig, line 56: It'd be helpful to have another sentence on why you need to create a scoped instance/what it is.

docs/docs/lib/node.mdx
```diff
@@ -45,59 +111,45 @@ app.use(function(req, res, next) {
...
-  if (gb.isOn("my-feature")) {
-    // Do something
-  }
-
-  // String/Number/JSON flag
-  const value = gb.getFeatureValue("my-string-feature", "fallback");
-  console.log(value);
-})
-```
+The `GrowthBook` class is identical to the one used in client-side environments. Check out the [/lib/js](Client-side JavaScript SDK docs) for more info on all of the settings and methods available.
```
royalfig, line 114: Syntax is reversed for the link:
`[/lib/js](Client-side)` -> `[Client-side JS SDK docs](/lib/js)`

docs/docs/lib/node.mdx
```diff
@@ -45,59 +111,45 @@ app.use(function(req, res, next) {
...
 
 ## Loading Features and Experiments
 
 In order for the GrowthBook SDK to work, it needs to have feature and experiment definitions from the GrowthBook API. There are a few ways to get this data into the SDK.
 
 ### Built-in Fetching and Caching
 
-If you pass an `apiHost` and `clientKey` into the GrowthBook constructor, it will handle the network requests, caching, retry logic, etc. for you automatically.
+If you pass an `apiHost` and `clientKey` into the GrowthBookMultiUser constructor, it will handle the network requests, caching, retry logic, etc. for you automatically.
```
royalfig, line 122: Add `` to GrowthBookMultiUser

docs/docs/lib/node.mdx
```diff
@@ -210,17 +269,40 @@ configureCache({
 
 In order to run A/B tests, you need to set up a tracking callback function. This is called every time a user is put into an experiment and can be used to track the exposure event in your analytics system (Segment, Mixpanel, GA, etc.).
 
+You can specify this globally in your GrowthBookMultiUser class and/or per-request in your user context.
+
+We recommend using a global callback if you plan to track events directly from Node and using a user context callback if you plan to send events to the front-end to fire.
```
royalfig, line 274: I think this can be a bit clearer. I understand the difference approaches (maybe add link to the section explaining the diff between global/user context earlier in the docs), and I know you say more with deferred tracking below. However, I don't think it's entirely clear when the tracking cb will fire. It'd be helpful if you could give more guidance on when/under what conditions that happens.

docs/docs/lib/node.mdx
```diff
@@ -489,34 +515,31 @@ const userEmail = sha256(salt + user.email);
...
 
 ### Feature Usage Callback
 
 GrowthBook can fire a callback whenever a feature is evaluated for a user. This can be useful to update 3rd party tools like NewRelic or DataDog.
 
+Like with the `trackingCallback`, this can be defined on either the GrowthBookMultiUser instance or as part of the user context. When defined on the GrowthBookMultiUser instance, a 3rd argument with the user context will be sent so you can identify which user evaluated a feature.
+
 ```ts
-new GrowthBook({
+const userContext ={
```
royalfig, line 535: It'd be helpful to have an example of the multiuser version, too.

docs/docs/lib/node.mdx
```diff
@@ -547,14 +570,14 @@ console.log(result.experimentResult);
 
 ### Inline Experiments
```
royalfig, line 571: Do you think it'd be helpful to explicitly say these experiments aren't communicated back to GB? Or is it clear enough?

jdorn: Inline experiments fire a tracking callback just like ones that are used via feature flags.  The difference is that inline experiments are defined in code vs in the GrowthBook UI.  Inline experiments are only good for a few niche use cases.

PR #3359, opened 2024-11-26
Add project validation to various REST API endpoints.

mknowlton89:
### Features and Changes

As a continuation of https://github.com/growthbook/growthbook/pull/3354 - This PR adds project validation to the Metrics `CREATE` and `UPDATE` Rest API Endpoints, as well as the `CREATE` and `UPDATE` Environment Rest API endpoints.

Now, if an invalid/non-existent project is passed in, an error will be thrown with a contextual message.

### Testing

Experiments
- [x] Pass in a valid project when creating a new experiment and ensure it works as expected
- [x] Pass in a valid project when updating an existing experiment and ensure it works as expected
- [x] Pass in an invalid project id when creating a new experiment and ensure an error is thrown and the experiment is not created
- [x] Pass in an invalid project id when updating an existing experiment and ensure an error is thrown and the experiment is not updated

Metrics
- [x] Pass in a valid projects array when creating a new metric and ensure it works as expected
- [x] Pass in a valid projects array when updating an existing metric and ensure it works as expected
- [x] Pass in an invalid projects array when creating a new metric and ensure an error is thrown and the metric is not created
- [x] Pass in an invalid projects array when updating an existing metric and ensure an error is thrown and the metric is not updated


packages/back-end/src/api/experiments/updateExperiment.ts
```diff
@@ -21,6 +21,16 @@ export const updateExperiment = createApiRequestHandler(
...
 
+    // Validate projects - We can remove this validation when FeatureModel is migrated to BaseModel
+    if (req.body.project) {
+      const projects = await req.context.getProjects();
+      if (!projects.some((p) => p.id === req.body.project)) {
+        throw new Error(
+          `Project id ${req.body.project} is not a valid project.`
+        );
+      }
+    }
```
jdorn, line 32: Why not use `ensureProjectsExist` here too?  I know that function is meant for an array of projects, but it works fine for a single one too.  Just pass in `[req.body.project]`.  The only thing is the error message language is a little different, but I think that's fine.

jdorn (approved):

PR #3369, opened 2024-11-27
add page_title to tracked events sent to datawarehouse

tzjames:
### Features and Changes
Adds the page title to the datawarehouse.


jdorn:
I think page titles may contain sensitive info like experiment or feature names.  We don't want tracked telemetry events to contain any sensitive info, especially for self-hosted instances.

tzjames:
Not landing due to possible PII.

PR #3381, opened 2024-12-03
Add an enable-debug mode for ingestor calls

tzjames:
### Features and Changes
In "debug" mode we don't actually send the telemetry data to the ingestor, but just output the event in the console.  So there was no way to actually see if the request failed to send, or had an error response since it was impossible to be both `isTelemetryEnabled()` and `inTelemetryDebugMode()`.  This PR adds an "enable-debug" mode.

### Dependencies
Needs to first land
https://github.com/growthbook/growthbook-ingestor/pull/28

### Testing

Set `DISABLE_TELEMETRY=enable-debug` in .env.local
In .env.local set: `INGESTOR_HOST=http://localhost:3003`
Load a page.
See in the dev console that only a single call gets sent to the /track endpoint (no option call) and that it returns a 200 response.
See the console.log "Telemetry Event - Successfully recorded event"
Edit track.ts to comment out one of the required properties in `dataWareHouseTrack()` call.
Reload the page.
See in the dev console that only a single call gets sent to the /track endpoint (no option call) and that it returns a 400 response.
See the console.log show "Tracking event - Ingestor returned a 400" with the body of the response.

### DIFF
--- a/packages/front-end/pages/api/init.ts
+++ b/packages/front-end/pages/api/init.ts
@@ -4,7 +4,7 @@ import { NextApiRequest, NextApiResponse } from "next";
 import { stringToBoolean } from "shared/util";
 
 export interface EnvironmentInitValue {
-  telemetry: "debug" | "enable" | "disable" | "enable-debug";
+  telemetry: "debug" | "enable" | "disable" | "enable-with-debug";
   cloud: boolean;
   isMultiOrg?: boolean;
   allowSelfOrgCreation: boolean;
@@ -115,8 +115,8 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
     telemetry:
       DISABLE_TELEMETRY === "debug"
         ? "debug"
-        : DISABLE_TELEMETRY === "enable-debug"
-        ? "enable-debug"
+        : DISABLE_TELEMETRY === "enable-with-debug"
+        ? "enable-with-debug"
         : DISABLE_TELEMETRY
         ? "disable"
         : "enable",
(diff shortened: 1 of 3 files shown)


packages/front-end/pages/api/init.ts
```diff
@@ -4,7 +4,7 @@ import { NextApiRequest, NextApiResponse } from "next";
 import { stringToBoolean } from "shared/util";
 
 export interface EnvironmentInitValue {
-  telemetry: "debug" | "enable" | "disable";
+  telemetry: "debug" | "enable" | "disable" | "enable-debug";
```
romain-growthbook, line 7: How about `enable-with-debug`? That would be more clear I think.

romain-growthbook (approved):
This is functional so ✅ but I think the naming could be clearer. Thanks!

packages/front-end/services/track.ts
```diff
@@ -87,17 +87,28 @@ const dataWareHouseTrack = async (event: DataWarehouseTrackedEvent) => {
...
+    if (inTelemetryDebugMode()) {
+      if (!result.ok) {
+        console.error(
+          `Telemetry - Ingestor returned a ${result.status}`,
+          await result.json()
+        );
+      } else {
+        console.log("Telemetry - Successfully recorded event");
+      }
+    }
```
jdorn, line 111: I don't think we need to log successes.  And since this is wrapped in a try/catch already with logging, we can simply do:

```ts
const result = await fetch(...);
if (!result.ok) {
  throw new Error(...)
}
```

tzjames: done

PR #3384, opened 2024-12-03
Adds MAX and COUNT DISTINCT aggregations to mean Fact Metrics

lukesonnet:
### Features and Changes

Adds COUNT DISTINCT and MAX to mean fact metrics.

COUNT DISTINCT uses HLL estimation methods (or ClickHouse's `uniq` default estimation method) exclusively and is not available for: non-SQL warehouses (Mixpanel), Postgres, Vertica, MySQL, and MS SQL (SQL SERVER).

Does so by adding a new `aggregation` field to column ref, and default is `SUM` if undefined, otherwise one of defined aggregations.

TODO:
- [X] API endpoint updates
- [ ] Documentation updates
- [X] Testing

### Testing

Tested all but Presto/Athena with dummy queries.

- [ ] Test in experiment with
  - [X] Clickhouse
  - [X] BigQuery
  - [X] Snowflake
  - [X] Databricks
  - [x] Redshift
  - [x] Presto
  - [ ] Athena (won't do, only Presto)
- [ ] Test in metric analysis with
  - [X] Clickhouse
  - [X] BigQuery
  - [X] Snowflake
  - [X] Databricks
  - [x] Redshift
  - [x] Presto
  - [ ] Athena (won't do, only Presto)

Tested Create and Update API endpoints, as well as errors when trying to use the wrong aggregation.

### Screenshots

https://www.loom.com/share/dc78491dbeff475da17d0d197d4eb989


packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -577,14 +591,22 @@ function getPreviewSQL({
       ? "COUNT(*)"
       : numerator.column === "$$distinctUsers"
       ? "1"
-      : `SUM(${numerator.column})`;
+      : numerator.aggregation === "count distinct"
+      ? `-- HyperLogLog estimation used instead of COUNT DISTINCT\n  COUNT(DISTINCT ${numerator.column})`
```
jdorn, line 595: I'm sure for a certain type of user this comment would be interesting, but I wonder if it will just confuse most people.  Seems like this should probably live in the docs instead where we can explain it more.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -4779,6 +4818,14 @@ ${this.selectStarLimit("__topValues ORDER BY count DESC", limit)}
...
       ) {
         return `SUM(${valueColumn})`;
+      } else if (columnRef?.aggregation === "count distinct") {
+        if (willReaggregate) {
+          return this.hllAggregate(valueColumn);
+        }
+        // TODO test null
+        return this.hllCardinality(this.hllAggregate(valueColumn));
+      } else if (columnRef?.aggregation === "max") {
+        return `MAX(COALESCE(${valueColumn}, 0))`;
```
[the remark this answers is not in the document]
lukesonnet: Oh, that is interesting. I think the idea that NULL is 0 always made sense until now. I'll think about this some more.

packages/front-end/components/FactTables/FactMetricModal.tsx
lukesonnet: Yeah, I waffled on this. I'll remove it.

PR #3386, opened 2024-12-03
Fix cumulative ratio + quantile date cohort graph

lukesonnet:
### Features and Changes

Removes non-sense cumulative date cohort graph for quantiles.

Fixes cumulative denominator for ratio metric cumulative date cohort graph.

### Testing

Quantile metrics hide when toggling on Cumulative.

Ratio metrics change, haven't confirmed exact values.

### Screenshots

Quantile new state
Screenshot 2024-12-03 at 11 10 55 AM

### DIFF
--- a/packages/front-end/components/Experiment/DateResults.tsx
+++ b/packages/front-end/components/Experiment/DateResults.tsx
@@ -9,6 +9,7 @@ import {
   ExperimentMetricInterface,
   isExpectedDirection,
   isStatSig,
+  quantileMetricType,
   shouldHighlight,
 } from "shared/experiments";
 import { DifferenceType, StatsEngine } from "back-end/types/stats";
@@ -141,6 +142,7 @@ const DateResults: FC<{
           // Keep track of cumulative users and value for each variation
           const totalUsers: number[] = [];
           const totalValue: number[] = [];
+          const totalDenominator: number[] = [];
 
           const datapoints: ExperimentDateGraphDataPoint[] = sortedResults.map(
             (d) => {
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Experiment/DateResults.tsx
```diff
@@ -177,10 +182,10 @@ const DateResults: FC<{
                   // For non-baseline variations and cumulative turned ON, calculate uplift from cumulative data
                   else if (i) {
                     const crA = totalUsers[0]
-                      ? totalValue[0] / totalUsers[0]
+                      ? totalValue[0] / totalDenominator[0]
```
bryce-fitzsimons, line 185: might need a divide-by-zero check

packages/front-end/components/Experiment/DateResults.tsx
```diff
@@ -341,20 +344,24 @@ const DateResults: FC<{
...
-              differenceType === "relative"
-                ? formatPercent
-                : getExperimentMetricFormatter(metric, getFactTableById, true)
-            }
-            formatterOptions={metricFormatterOptions}
-            variationNames={variations.map((v) => v.name)}
-            statsEngine={statsEngine}
-            hasStats={!cumulative}
-          />
+          {!quantileMetricType(metric) || !cumulative ? (
```
bryce-fitzsimons, line 347: seems like it would be better to disable the toggle and force the non-cumulative view to render

lukesonnet: The toggle is for all graphs at once. So I don't want to disable it if they have one quantile metric.

packages/front-end/components/Experiment/DateResults.tsx
lukesonnet: oh shoot, thanks, the ifelse logic needs to use totalDenominator

bryce-fitzsimons (approved):

PR #3394, opened 2024-12-04
created a Portal made with radix so that we can keep the style

gazzdingo:
### Features and Changes


mknowlton89 (approved):
Looks good! I see the `Modal` component is the only place we're using the old `Portal` component ( located at `/components/Modal/Portal`) - Does it make sense to update the Modal to use this portal so we can remove the legacy portal so no one accidentally uses it?)

PR #3397, opened 2024-12-05
Make uncontrolled tabs use URL hash by default

ahdriel:
### Features and Changes

When using uncontrolled Tabs now we will store the hash in the URL by default.
If we want a different behavior we can use a controlled Tabs with `useURLHash` hook directly and any other option (like useLocalStorage).

Additionally had to do some updates to the `useSearch` hook to not modify the URL when it is not needed as React was complaining about it.


Kevin-Chant (approved):
LGTM

PR #3400, opened 2024-12-06
Updates experiment hypothesis to use modal

mknowlton89:
### Features and Changes

This PR updates the experiment ID page - specifically the `Hypothesis` section. Previously, we were using the `MarkdownInlineEdit` component, that allowed users to edit the hypothesis inline. Now, we've converted this section to using a modal, with an `Edit` link in the top-right corner of the div.

### Testing

- [x] Ensure the empty state matches the mock
- [x] Ensure the completed state matches the mock
- [x] Ensure the modal matches the mock
- [x] Ensure that a user with permission to edit an experiment can add/edit an experiment's hypothesis
- [x] Ensure that a user without permission to edit can not add/edit the hypothesis
- [x] Ensure that the design looks good on both light and dark modes
- [x] Ensure the design is usable on mobile

### Screenshots

Walkthrough video - https://www.loom.com/share/914b1b6c213f4731ae2eb6ce7e68a5cc


packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -43,6 +46,18 @@ export default function SetupTabOverview({
 
   const isBandit = experiment.type === "multi-armed-bandit";
 
+  if (showHypothesisModal) {
+    return (
```
ahdriel, line 50: question: Should we conditionally render this instead of returning only the modal?

Currently when you open the modal the content on the page shifts as we are not rendering the `Overview` section anymore.

packages/front-end/components/Experiment/EditHypothesisModal.tsx
```diff
@@ -0,0 +1,56 @@
+import { useForm } from "react-hook-form";
+import { useAuth } from "@/services/auth";
+import Modal from "../Modal";
+import Field from "../Forms/Field";
+
+interface Props {
+  source: string;
+  close: () => void;
+  experimentId: string;
+  hypothesis?: string;
```
ahdriel, line 10: nit: should it be called `initialHypothesis` or `initialValue`?

ahdriel (approved):

packages/front-end/components/Experiment/EditHypothesisModal.tsx
mknowlton89: I think that's a good improvement. Thanks!

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
mknowlton89: Ooof - great catch! This has been updated so there isn't any shift. Thanks!

PR #3401, opened 2024-12-06
Introduces UI updates to URL Redirect section & modal

mknowlton89:
### Features and Changes

This PR updates the design of the URL Redirect section within an experiment's ID page, along with the modal for creating and editing URL Redirects for an experiment

### Dependencies

- n/a

### Testing

- [x] Ensure that you can view existing URL Redirects on the page
- [x] Ensure that you can create a new URL Redirect successfully
- [x] Ensure that you can edit an existing URL Redirect successfully
- [x] Ensure that you can delete a URL Redirect successfully

### Screenshots

URL Redirect Section (Before)
Screenshot 2024-12-09 at 2 58 10 PM

URL Redirect Section (After)
Screenshot 2024-12-09 at 2 57 36 PM

URL Redirect Create Modal (Before)
Screenshot 2024-12-09 at 3 00 27 PM

URL Redirect Create Modal (After)
Screenshot 2024-12-10 at 8 51 19 AM

URL Redirect Edit Modal (Before)
Screenshot 2024-12-09 at 2 58 24 PM

URL Redirect Edit Modal (After)
Screenshot 2024-12-10 at 8 51 12 AM


packages/front-end/components/Experiment/UrlRedirectModal.tsx
```diff
@@ -365,36 +318,34 @@ const UrlRedirectModal: FC<{
...
-          </label>
-          <input
-            type="checkbox"
-            {...form.register("circularDependencyCheck")}
-            id={"toggle-circularDependencyCheck"}
+          <Checkbox
+            label={
+              <>
+                <label>Circular Dependency Check</label>
+                <p style={{ color: "var(--color-text-mid)" }}>
```
mknowlton89, line 338: @bryce-fitzsimons @ahdriel - Still getting up to speed on the radix transition - is there a cleaner way to utilize our custom colors via Radix? 

(As an aside - I've got a question out to Kari to clarify if we always want the `description` of the `Checkbox` component to be this custom color. If so, I'll update the `Checkbox` component accordingly, but the question about is still valid).

ahdriel: I think Radix has a `<Text>` component which we can use -- but I am unsure if we have updated its default style to use `color-text-mid`.

I do agree this should be a more broad change and it looks like our default text color should be `mid` (which is in line with your thoughts around updating the Checkbox component default text color).

packages/front-end/components/Modal.tsx
```diff
@@ -72,7 +72,7 @@ const Modal: FC<ModalProps> = ({
   fullWidthSubmit = false,
   submitColor = "primary",
   open = true,
-  cta = "Submit",
+  cta = "Save",
```
ahdriel, line 75: question: This seems like a very broad change -- are we okay with this potentially affecting all Modals?

ahdriel (approved):

packages/front-end/components/Experiment/LinkedChanges/RedirectLinkedChanges.tsx
```diff
@@ -2,11 +2,12 @@ import { useState } from "react";
 import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
 import { diffChars } from "diff";
 import { URLRedirectInterface } from "back-end/types/url-redirect";
+import { FaExternalLinkAlt } from "react-icons/fa";
```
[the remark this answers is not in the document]
mknowlton89: Thank you! I think I missed that update when I was on leave. I'll update accordingly

packages/front-end/components/Modal.tsx
mknowlton89: I'll double check with Kari just to be safe, but I recall she mentioned wanting to get rid of the `Submit` language in exchange for `Save`. Thanks!

packages/front-end/components/Experiment/LinkedChanges/constants.ts
```diff
@@ -12,7 +11,7 @@ export const ICON_PROPERTIES = {
   },
   redirects: {
     color: "#11B081",
-    component: PiShuffle,
+    component: FaLink,
```
msamper, line 14: Do you know why we decided to change this logo? We originally considered the Link icon, but decided it didn't make sense for redirects. We also use this icon in other places in the app to copy a link to a page

mknowlton89: I don't know the specifics - I was just updating this based on Kari's new designs.

PR #3413, opened 2024-12-10
Updates modal scroll on SchemaBrowser on [did].tsx

mknowlton89:
### Features and Changes


tzjames (approved):
Looks great.  I like the fact that wasted space in the datasource is gone, though we might want SchemaBrowserWrapper py-2 from pt-2 so that there is a just a little bit of space seperating the two sections.

PR #3417, opened 2024-12-10
[bugfix] Re-allow activation metric to be joined to query

lukesonnet:
### Features and Changes

### Testing

Confirmed that I can't use joinable activation metric before, but now I can.


Kevin-Chant (approved):
LGTM

PR #3419, opened 2024-12-11
Introduces new exp description UI

mknowlton89:
### Features and Changes

This PR introduces a new pattern for the Experiment Overview tab based on Kari's updated designs. The description now has an edit modal rather than supporting in-line editing. Additionally, this PR also introduces an update where an organization can show and hide the description (in case it's super long).

This PR doesn't change any of the functionality of the description field - it still supports markup, and still has a maxHeight property with a scroll overflow.

Given that a user can now show/hide a description, rather than the user having to hide the description everytime they land on the experiment page, we're now persisting the state of the description box (open/closed), however this only persists in localStorage, not at the experiment level. This lets users decide if they want it opened or closed. As it's in localStorage, the state does not sync across devices/browers.

If the experiment doesn't have a description, we disable the show/hide functionality, and default to showing the preview text.

### Testing

- [x] If an experiment does not have a description, ensure we show the previewText and don't allow the user to collapse the div
- [x] If an experiment does have a description, ensure that we show it (if it's expanded) and ensure the user, regardless of permission level can collapse it, and the state persists so long as they keep their browser open
- [x] Ensure that a user is able to add text and markup to the description and it works successfully
- [x] Ensure that the designs look good on both light mode and dark mode
- [x] Ensure that if a user edits the description, we reset the expand/collapse state back to expanded
- [x] Ensure that if a user removes a description, we reset the expand/collapse state back to expanded, disable the ability to collapse it (until there is a description) and show the previewText
- [x] Ensure that we show a scroll bar if the description is long


packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -37,8 +40,13 @@ export default function SetupTabOverview({
   setChecklistItemsRemaining,
   envs,
 }: Props) {
-  const { apiCall } = useAuth();
   const [showHypothesisModal, setShowHypothesisModal] = useState(false);
+  const [showDescriptionModal, setShowDescriptionModal] = useState(false);
+
+  const expandDescription =
+    localStorage.getItem(`collapse-${experiment.id}-description`) === "true"
```
lukesonnet, line 47: Is there a reason to do this instead of `useLocalStorage`?

lukesonnet (approved):
lgtm.

May be out of scope, but the tabs on MarkdownInput should probably be stylistically updated.

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -75,30 +92,76 @@ export default function SetupTabOverview({
...
-              containerClassName="mb-1"
-            />
-          </div>
-        </div>
+            {experiment.description ? (
+              <Box
+                as="div"
+                px="5"
+                py="2"
+                maxHeight="491px"
```
lukesonnet, line 151: I kind of feel like this should be larger.... IDK if you have a long-ish description, that's really what this overview page is for.

lukesonnet: You can also get unlucky and the scroll nature of the window can be really hard to see:

Screenshot 2025-01-06 at 8 59 27 AM

Michael samper had to resolve this on the dimension issues card on the health tab with dark shading:
Screenshot 2025-01-06 at 9 00 10 AM

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
mknowlton89: I simply forgot about this hook. I'll refactor to use it for consistency.

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
mknowlton89: Thanks! I'll leverage the same pattern Michael Samper used

mknowlton89: I'll share the feedback above re: length with Kari and get her thoughts.

mknowlton89:
Thanks for the review, @lukesonnet! I've created a follow-up ticket for the `MarkdownInput` tabs so the feedback doesn't get lost. I'll tackle this as a downtime project. 

https://github.com/growthbook/growthbook/issues/3496

PR #3421, opened 2024-12-11
Decision Framework: Power calculation for metrics

ahdriel:
### Features and Changes


ahdriel:
Closing in favor of #3557 which builds on the foundation set by #3540.

PR #3423, opened 2024-12-11
api(teams): Add check for the org

ahdriel:
### Features and Changes

While the `teams` feature is available only for Enterprise licenses, we were not doing the proper check in the API.

This fixes it by validating the license when creating a team.

### DIFF
--- a/packages/back-end/src/routers/teams/teams.controller.ts
+++ b/packages/back-end/src/routers/teams/teams.controller.ts
@@ -51,9 +51,7 @@ export const postTeam = async (
   const { name, description, permissions } = req.body;
 
   if (!orgHasPremiumFeature(org, "teams")) {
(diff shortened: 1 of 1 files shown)


packages/back-end/src/routers/teams/teams.controller.ts
```diff
@@ -49,6 +50,12 @@ export const postTeam = async (
   const { org, userName } = context;
   const { name, description, permissions } = req.body;
 
+  if (!orgHasPremiumFeature(org, "teams")) {
+    throw new Error(
+      "Must have a commercial License Key to restrict permissions by environment."
```
jdorn, line 55: Change error message to reference teams

jdorn (approved):

PR #3432, opened 2024-12-13
Updates Experiment MoreMenu

mknowlton89:
### Features and Changes

This PR updates the `MoreMenu` on the Experiment ID page to use Radix's `DropDownMenu`. Additionally, it updates the UI and UX based on the new 3.4 designs in Figma.

Note: In the 3.4 designs, there is an "Edit Info" option not included in this PR - the work to build that modal is being done [here](https://github.com/growthbook/growthbook/pull/3462), but will be merged in after this PR.

### Testing

Note: Contact me for login credentials to the staging environment if you'd like to test these changes - in that environment I have an experiment in draft, running, and stopped state.

- [x] Ensure all of the buttons work as expected for a draft experiment
- [x] Ensure all of the buttons work as expected for a running experiment
- [x] Ensure all of the buttons work as expected for a stopped experiment
- [x] Ensure that async dropdown items work as expected
- [x] Ensure that async dropdown items that return an error display the error and don't crash the app

Draft Experiment
Screenshot 2025-01-03 at 8 47 07 AM

Draft Bandit Experiment
Screenshot 2025-01-03 at 8 48 05 AM

Running Experiment
Screenshot 2025-01-03 at 8 47 22 AM

Running Bandit Experiment
Screenshot 2025-01-03 at 8 54 18 AM

Stopped Experiment
Screenshot 2025-01-03 at 8 47 34 AM

Stopped Bandit Expe
Screenshot 2025-01-03 at 8 54 28 AM
riment

Error state of clicking `Start Watching`
Screenshot 2025-01-03 at 8 58 21 AM

Error state of clicking `Create Shareable Report`
Screenshot 2025-01-03 at 8 58 56 AM

Light Mode
Screenshot 2025-01-03 at 8 57 42 AM


packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -419,144 +622,131 @@ export default function ExperimentHeader({
...
-                            method: "POST",
-                          }
-                        );
-                        mutate();
-                      } catch (e) {
-                        console.error(e);
-                      }
-                    }}
+                    <DropdownMenuItem
+                      onClick={async () => await handleWatchUpdates(true)}
```
mknowlton89, line 666: The DropdopMenuItem component's `onClick` prop doesn't support asychronous callbacks explicitly. This is working (and the error states work as well in that it doesn't cause an app crash). 

If anyone has a better pattern for this, let me know. We may just need to open a modal like we do everywhere else.

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -555,203 +755,148 @@ export default function ExperimentHeader({
...
-                  <button
-                    className="dropdown-item"
-                    onClick={() => setShareModalOpen(true)}
+                    <DropdownMenuItem
+                      onClick={async () =>
+                        await handleWatchUpdates(!isWatching)
+                      }
+                    >
+                      {isWatching ? "Stop" : "Start"} watching
+                    </DropdownMenuItem>
```
mknowlton89, line 805: Would love feedback on this approach. It works, but `DropdownMenuItem` doesn't formally accept an async callback function, so while it works, there isn't a great way to surface any errors that may come up. I've got it wrapped in a try/catch so errors won't cause the app the crash, and I'm logging the errors to the console, but I feel like we should improve this experience.

This also applies to the `Create shareable report` if that fails, the error is just quietly logged to the console.

We could create an error banner that lives on the `Experiment Header` that we could display if/when there is an error with one of the `MoreMenu` items. It'd need to be clearable, I guess.

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -256,18 +256,14 @@ export default function ExperimentHeader({
...
-          method: "POST",
-        }
-      );
-      refreshWatching();
-      mutateWatchers();
-    } catch (e) {
-      console.error(e);
-    }
+    await apiCall(
+      `/user-bad/${watch ? "watch" : "unwatch"}/experiment/${experiment.id}`,
```
mknowlton89, line 260: This is to force an error

packages/front-end/components/Radix/DropdownMenu.tsx
```diff
@@ -110,15 +110,34 @@ export function DropdownMenuItem({
...
-      color={color}
+      onSelect={async (event) => {
+        if (error) {
+          setError("");
+        }
+        event.preventDefault();
+        if (onClick) {
+          try {
+            await onClick(event);
+            // If this promise is resolved without an error, we need to close
```
mknowlton89, line 125: This is where we need to be able to hoist some method to close the dropdown.

packages/front-end/components/Radix/DropdownMenu.tsx
```diff
@@ -110,15 +118,43 @@ export function DropdownMenuItem({
...
+        setLoading(true);
+        if (onClick) {
+          try {
+            await onClick(event);
+            // If this promise is resolved without an error, we need to close
+          } catch (e) {
+            setError(e.message);
+            console.error(e);
+          }
+          setLoading(false);
```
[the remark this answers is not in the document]
msamper: Revisiting this -- I guess it's not really better if you have to pass in a callback to close the dropdown to each `DropdownMenuItem`. Same amount of code. Something you could consider doing is injecting the `() => setDropdownOpen(false)` callback into each child's props via the parent (`DropdownMenu`) and giving the parent a new prop to be able to close itself

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -695,92 +906,69 @@ export default function ExperimentHeader({
...
-                      } catch (e) {
-                        console.error(e);
-                      }
-                    }}
-                    cta="Archive"
-                  >
-                    <button className="dropdown-item" type="button">
+                {canRunExperiment &&
+                  growthbook.isOn("bandits") &&
+                  experiment.status === "draft" && (
```
msamper, line 913: Does this need a check for whether or not a user has the bandits feature?

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -358,6 +387,174 @@ export default function ExperimentHeader({
...
+            <div>
+              <p>Are you sure you want to delete this experiment?</p>
+              {!safeToEdit ? (
+                <div className="alert alert-danger">
+                  This will immediately stop all linked Feature Flags and Visual
+                  Changes from running
+                </div>
+              ) : null}
+            </div>
+          </Modal>
```
msamper, line 517: We might want to componentize this since I imagine we'll be replacing our dropdowns slowly with the Radix dropdown. Maybe something like `DropdownDeleteButton`. Just a note though, doesn't need to happen here

msamper: Same with the confirmation dialog above

ahdriel: Radix does have [Alert Dialog](https://www.radix-ui.com/themes/docs/components/alert-dialog) component, unsure if it is helpful/we should use it or not though.

Agreed on making this pattern easier to access.

packages/front-end/components/Radix/DropdownMenu.tsx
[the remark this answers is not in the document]
mknowlton89: Re: "Should setLoading happen inside the conditional instead to avoid state changes if there's no onClick callback passed in?" - absolutely, updating this.

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -576,105 +773,119 @@ export default function ExperimentHeader({
...
+                    </Flex>
+                  </DropdownMenuItem>
+                </DropdownMenuGroup>
+                <DropdownMenuSeparator />
                 {canCreateTemplate && !isBandit && (
-                  <button
-                    className="dropdown-item"
-                    onClick={() => setShowTemplateForm(true)}
-                  >
+                  <DropdownMenuItem onClick={() => setShowTemplateForm(true)}>
```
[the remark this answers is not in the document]
mknowlton89: Yup - thanks!

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
mknowlton89: So, this was a design choice that I wasn't originally involved with, but it looks like we always want to show the "Convert to Bandit" CTA, regardless of plan. The modal that is opened then highlights that this is a premium feature and prompts the user to upgrade.

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
mknowlton89: @ahdriel The `Alert Dialog` is interesting - I'll share that with Kari and get her opinion on it. I explored a quick MVP of it and it looks like we can use the DropdownMenuItem as a trigger for it. 

I think we could then build a wrapper around it like we've done for the modal component to handle things like event tracking and such. 

Though, I feel like it may make sense to do that work in a separate PR to give it the attention it deserves.

mknowlton89: As an aside, I know Kari has updated designs for our `Modal` component, so it might not make sense to use two different solutions (our Modal and Radix's Alert Dialog) as we probably want a consistent experience. I've let Kari know and will keep you all posted.

msamper:
Screenshot 2025-01-09 at 12 04 42 PM
Tested and everything seems to work as expected. Only nit is that for running experiments it seems like we're missing a divider under `Share experiment`

packages/front-end/components/Radix/DropdownMenu.tsx
```diff
@@ -110,15 +118,43 @@ export function DropdownMenuItem({
...
     >
-      {children}
+      <Flex as="div" justify="between" align="center">
+        <Box as="span" className={`mr-4 ${loading ? "font-italic" : ""}`}>
+          {children}
+        </Box>
+        <Box width="14px">
+          {loading ? <LoadingSpinner /> : null}
+          {error ? (
+            <Tooltip body={`Error: ${error}. Exit menu and try again.`}>
```
msamper, line 152: I think the designs have this as a Radix Tooltip but not sure if we want to create a wrapper around it before we actually start using it

msamper (approved):

PR #3435, opened 2024-12-14
Added automatic Snowplow tracking to the html SDK

Auz:
Blocked by https://github.com/snowplow/iglu-central/pull/1423


jdorn (approved):

packages/sdk-js/src/plugins/third-party-tracking.ts
```diff
@@ -68,6 +68,25 @@ export function thirdPartyTrackingPlugin({
...
+              experimentId: e.key,
+              variationId: r.key,
+              hashAttribute: r.hashAttribute,
+              hashValue: r.hashValue,
+            },
+          },
+        });
+        // snowplow uses local storage so we don't need to wait long
+        const snowplowPromise = new Promise((resolve) =>
+          window.setTimeout(resolve, 80)
```
jethron, line 85: You can `window.snowplow(resolve)` and it will call the function once it has successfully processed the above event and committed it to localStorage.

But if the tracker hasn't loaded for some reason the Promise will never resolve (not sure what awaits this promise).

packages/sdk-js/src/plugins/third-party-tracking.ts
```diff
@@ -68,6 +68,25 @@ export function thirdPartyTrackingPlugin({
         promises.push(segmentPromise);
       }
 
+      if (trackers.includes("snowplow") && window.snowplow) {
```
jethron, line 71: It's not super common but we do support names other than `snowplow`; not sure if that should be configurable or you just want to recommend using the generic `trackingCallback` in that case.

We also have the [Browser Tracker](https://docs.snowplow.io/docs/sources/trackers/javascript-trackers/web-tracker/tracker-setup/?platform=browser) which is an imported package so usually isn't exposed globally; that would likely also need a custom function to be able to call the equivalent `trackSelfDescribingEvent` method and may be worth documenting similar to your [Mixpanel example](https://docs.growthbook.io/lib/script-tag#mixpanel).

PR #3439, opened 2024-12-16
Add minutes as option for windows and reduce SQL coarsening

lukesonnet:
### Features and Changes

Allow minutes for windows since they are supported by our SQL engines.

### Testing

Created BQ, Trino, and Postgres delays and windows with minutes and logic worked as expected, rounding to hours when possible but allowing minutes to pass through.

### Screenshots

Screenshot 2024-12-16 at 9 35 14 AM
Screenshot 2024-12-16 at 9 35 00 AM
Screenshot 2024-12-16 at 9 33 48 AM
Screenshot 2024-12-16 at 9 30 01 AM
Screenshot 2024-12-16 at 9 29 24 AM
Screenshot 2024-12-16 at 9 28 58 AM
Screenshot 2024-12-16 at 9 17 02 AM


packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -645,7 +645,7 @@ export default function FactMetricPage() {
               open={() => setEditOpen(true)}
               canOpen={canEdit}
             >
-              {factMetric.windowSettings.delayValue > 0 && (
+              {factMetric.windowSettings.delayValue && (
```
Comment (jdorn) [inline packages/front-end/pages/fact-metrics/[fmid].tsx:648]: Don't do this.  `0 && ...` will be evaluated to `0` and React will print that to the screen.  If you force it to be a boolean (e.g. by doing `> 0`), then it will evaluate to `false` instead, which React will NOT print to the screen.  Alternatively, turn this into a ternary operator and do `delayValue ? (...) : null` to be explicit about what to output when it is zero.

jdorn (approved):

PR #3442, opened 2024-12-17
Add Experiment Templates

msamper:
### Features and Changes

Adds templates for experiments that allow users to create templates that can speed up experiment creation. Templates are created in a flow similar to the create experiment flow and can be used during experiment creation to pre-fill fields set by the template. Users are still able to modify selections derived from the template. Also adds a new org setting that can enforce template usage for experiment creation. A detail page is available for each template with either a list of experiments using the template or a CTA to use the template to create a new experiment.

### Testing

Try testing the entire lifecycle of a template. 

1. Creation
2. Editing
3. Duplicating
4. Deleting
5. Using a template to create an experiment
6. Converting an experiment to a template
7. Requiring templates for an org
8. Viewing template usage

#### Templates Page Empty State
Screenshot 2024-12-23 at 1 24 47 PM
#### Add experiment/Add Template dropdown
Screenshot 2024-12-23 at 1 24 55 PM
#### Create Template flow
Screenshot 2024-12-23 at 1 25 06 PM
#### Templates Select in New Experiment Form
Screenshot 2024-12-23 at 1 26 58 PM
#### Templates Page with a template
Screenshot 2024-12-23 at 1 27 40 PM
#### Template Detail page
Screenshot 2024-12-23 at 1 27 48 PM
#### Experiment more menu with new Save as template... option
Screenshot 2024-12-23 at 1 28 06 PM
#### New org setting to require experiment template for experiment creation
Screenshot 2024-12-23 at 1 28 36 PM


packages/front-end/components/Experiment/Templates/TemplateForm.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+        {["Overview", "Traffic", "Targeting", "Metrics"].map((p, i) => {
+          // skip, custom overview page above
+          if (i === 0) return null;
+          return (
+            <Page display={p} key={i}>
+              <ExperimentRefNewFields
+                step={i}
+                source="experiment"
+                project={project}
```
msamper, line 323: Experiment fields use a single project to determine options for certain fields that might change depending on the project, but templates currently can be created for multiple projects. Do we need to make templates limited to one project like we do for experiments for compatibility with `ExperimentRefNewFields`?

msamper: Right now I'm passing in the current project which is definitely not what we want to do

packages/back-end/src/models/ExperimentTemplateModel.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+  protected canDelete(doc: ExperimentTemplateInterface): boolean {
+    return this.context.permissions.canDeleteExperimentTemplate(doc);
+  }
+
+  public getByProject(project: string) {
+    return this._find({ project: project });
+  }
+
+  // TODO: Implement this for OpenAPI
```
msamper, line 45: Will add OpenAPI support in a fast follow PR

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -472,6 +501,53 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+                        <Text>{value.label}</Text>
+                        <Text size="1" className="text-muted" ml="auto">
+                          Created {date(t.dateCreated)}
+                        </Text>
+                      </Flex>
+                    );
+                  }}
+                  disabled={!hasCommercialFeature("templates")}
+                  required={
+                    templateRequired && hasCommercialFeature("templates")
```
msamper, line 546: remove `hasCommercialFeature`. Already checked in `templateRequired`

packages/front-end/components/Experiment/Templates/TemplateForm.tsx
```diff
@@ -0,0 +1,354 @@
...
+          body,
+        });
+
+    track("Create Experiment Template", {
+      source,
+      numTags: value.tags?.length || 0,
+      numMetrics:
+        (value.goalMetrics?.length || 0) +
+        (value.secondaryMetrics?.length || 0),
+    });
```
msamper, line 174: See what else might be useful to include here. Add whether or not an experiment was created from a template to new experiment tracking event

packages/front-end/components/Modal/PagedModal.tsx
```diff
@@ -175,6 +175,7 @@ const PagedModal: FC<Props> = (props) => {
   }, [step]);
 
   return (
+    // TODO: Pass nav as a prop so that body can be styled differently
```
msamper, line 178: delete

packages/front-end/services/experiments.ts
```diff
@@ -654,3 +661,67 @@ export function getEffectLabel(differenceType: DifferenceType): string {
...
+
+export function convertExperimentToTemplate(
+  experiment: ExperimentInterfaceStringDates
+): Partial<ExperimentTemplateInterface> {
+  const latestPhase = experiment.phases[experiment.phases.length - 1];
+  const template = {
+    templateMetadata: {
+      name: "",
+      description: `Template based on ${experiment.name}`,
+      tags: [],
```
msamper, line 702: delete

packages/back-end/src/routers/experiment-template/template.controller.ts
```diff
@@ -0,0 +1,143 @@
...
+  req: AuthRequest<CreateTemplateProps>,
+  res: Response<
+    CreateTemplateResponse | PrivateApiErrorResponse,
+    EventUserForResponseLocals
+  >
+) => {
+  const context = getContextFromReq(req);
+  const { userId } = context;
+  const template = req.body;
+
```
mknowlton89, line 59: Should we check that the org has the premium feature here and return an error if they don't?

packages/front-end/pages/experiments/index.tsx
```diff
@@ -327,13 +350,33 @@ const ExperimentsPage = (): React.ReactElement => {
...
+      {canAddExperiment && (
+        <DropdownMenuItem onClick={() => setOpenNewExperimentModal(true)}>
+          Create New Experiment
+        </DropdownMenuItem>
+      )}
+      {canAddTemplate && (
+        <DropdownMenuItem onClick={() => setOpenTemplateModal({})}>
+          Create Template
+        </DropdownMenuItem>
+      )}
```
mknowlton89, line 371: Should we disable or hide this option if the org doesn't have the premium feature?

packages/front-end/components/Experiment/Templates/TemplatesPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    {canCreate ? (
+                      <button
+                        className="dropdown-item"
+                        onClick={() => {
+                          setOpenDuplicateTemplateModal(t);
+                        }}
+                      >
+                        Duplicate
+                      </button>
+                    ) : null}
```
mknowlton89, line 148: This also needs to take into consideration if the org has the premium feature

packages/front-end/components/Experiment/Templates/TemplateForm.tsx
```diff
@@ -0,0 +1,354 @@
...
+          `/templates/${initialValue.id}`,
+          {
+            method: "PUT",
+            body,
+          }
+        )
+      : await apiCall<{ template: ExperimentTemplateInterface }>("/templates", {
+          method: "POST",
+          body,
+        });
```
mknowlton89, line 166: It looks like when trying to duplicate an existing template, the template.id is passed in, so this code block ends up updating the template a user is trying to duplicate, rather than creating a new template.

packages/front-end/components/Experiment/Templates/TemplatesPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+                </td>
+                <td data-title="Tags">
+                  <SortedTags
+                    tags={Object.values(t.tags ?? [])}
+                    useFlex={true}
+                  />
+                </td>
+                <td className="text-gray col-2">
+                  {t.project ? getProjectById(t.project)?.name : ""}
+                </td>
```
mknowlton89, line 122: We're conditionally rendering this column only if `showProjectColumn` is true, so we should do the same here.

packages/front-end/styles/global-radix-overrides.scss
```diff
@@ -315,6 +315,10 @@
   }
 }
 
+.rt-Text {
+  margin-bottom: 0 !important;
+}
+
```
mknowlton89, line 321: This seems like a pretty widespread change - is this necessary?

packages/front-end/components/Experiment/Templates/TemplatesPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+        />
+      )}
+      <div className="appbox p-5 text-center">
+        <h1>Create Reusable Experiment Templates</h1>
+        <Text size="3">
+          Save time configuring experiment details, and ensure consistency
+          across your team and projects.
+        </Text>
+        <div className="mt-3">
+          {/* TODO: Fix docs link once docs are ready */}
```
bryce-fitzsimons, line 188: reminder

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -605,6 +620,14 @@ export default function ExperimentHeader({
                     mutate={mutate}
                   />
                 )}
+                {canCreateTemplate && !isBandit && (
+                  <button
+                    className="dropdown-item"
+                    onClick={() => setShowTemplateForm(true)}
+                  >
+                    Save as template...
```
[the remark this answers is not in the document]
msamper: We’re doing this for another feature (reports, I think) it’s a design choice to indicate that clicking “save as” will require the user to go through a flow before they can save the experiment as a template

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -472,6 +502,51 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
                 datasource project is deleted.
               </div>
             )}
+            {availableTemplates.length >= 1 && !isBandit && (
```
bryce-fitzsimons, line 505: I don't think templates should be here for "Import existing experiment", since they can lead to some weird conflicts.

bryce-fitzsimons: Same for "Duplicate Experiment"

packages/front-end/components/Experiment/Templates/TemplateForm.tsx
```diff
@@ -0,0 +1,354 @@
...
+              <div className="alert alert-warning">
+                You are creating a template under the demo datasource project.
+                This template will be deleted when the demo datasource project
+                is deleted.
+              </div>
+            )}
+
+            <h4 className="mb-3">Template Details</h4>
+
+            <Field
```
bryce-fitzsimons, line 249: can we apply a default name based on the experiment name?
`${experiment.name} Template` or something

packages/front-end/styles/global-radix-overrides.scss
bryce-fitzsimons: +1

bryce-fitzsimons:
I don't know if I like this button tbh. It kind of looks like a battery or something
image

msamper:
> I don't know if I like this button tbh. It kind of looks like a battery or something image

The idea was to make it work for both adding experiments and templates. I guess we could also do something like "Add..." which we currently do in a tooltip, but let me know if you have any suggestions

bryce-fitzsimons:
> > I don't know if I like this button tbh. It kind of looks like a battery or something image
> 
> The idea was to make it work for both adding experiments and templates. I guess we could also do something like "Add..." which we currently do in a tooltip, but let me know if you have any suggestions

I think I'd go with Add... or Create... personally

packages/front-end/components/Experiment/Templates/TemplateForm.tsx
msamper: Done. Added name in `convertExperimentToTemplate`

packages/front-end/services/experiments.ts
```diff
@@ -654,3 +661,66 @@ export function getEffectLabel(differenceType: DifferenceType): string {
...
+    templateMetadata: {
+      name: "",
+      description: `Template based on ${experiment.name}`,
+    },
+    project: experiment.project,
+    type: "standard" as const,
+    hypothesis: experiment.hypothesis,
+    tags: experiment.tags,
+    datasource: experiment.datasource,
+    userIdType: experiment.userIdType, // TODO: Check if this needs to be removed from ExperimentTemplateInterface
```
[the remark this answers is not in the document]
msamper: Good catch! Removing this because it's deprecated. `exposureQueryId` is what we use now

packages/front-end/styles/global-radix-overrides.scss
msamper: I think I did this to override the styling coming from the bootstrap stylesheet that messes with the way a few radix components render (i.e. radix tooltip & radix callout), but I can just do the override for the specific components to avoid any unintended widespread changes

packages/front-end/pages/experiments/index.tsx
msamper: Do we have a pattern for what we do with premium feature actions that show up in a dropdown? Do we just hide them or do we disable them and indicate they're premium somehow

bryce-fitzsimons: usually disable and wrap in a PremiumTooltip

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -472,6 +502,51 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+                      <Flex as="div" align="baseline">
+                        <Text>{value.label}</Text>
+                        <Text size="1" className="text-muted" ml="auto">
+                          Created {date(t.dateCreated)}
+                        </Text>
+                      </Flex>
+                    );
+                  }}
+                  disabled={!hasCommercialFeature("templates")}
+                  required={templateRequired}
```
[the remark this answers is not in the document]
msamper: There doesn't seem to be a straightforward way to customize the default required message for `ReactSelect`. Let me chat with Kari about how to make this a bit more clear

bryce-fitzsimons:
minor nit: Bit of a disconnect when clicking the Add... button and seeing "create" as the next options. Seems like they should be consistent. Maybe a Kari question.
image

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -312,6 +322,11 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
       throw new Error("Name must not be empty");
     }
 
+    if (!value.templateId && templateRequired) {
```
bryce-fitzsimons, line 325: this breaks importing and duplicating

bryce-fitzsimons (changes requested):

msamper:
> minor nit: Bit of a disconnect when clicking the Add... button and seeing "create" as the next options. Seems like they should be consistent. Maybe a Kari question. image

Let me ask her for her thoughts here. I’ll change it to “Create…” for now

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -420,6 +445,11 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
   const type = form.watch("type");
   const isBandit = type === "multi-armed-bandit";
 
+  const templateRequired =
```
bryce-fitzsimons, line 448: possibly should also check whether there are any valid templates first before forcing them to use it.

bryce-fitzsimons:
a couple of nits from the template listing page
image

1. The "Power Calculator" button seems out of place / unnecessary when viewing templates. Maybe out of scope?
2. I keep wanting to see a quick "create experiment from template" button in the kebab menu and possibly also in the "list of experiments using this template" page.

bryce-fitzsimons:
UI nit: these dropdown items should be full width
image

bryce-fitzsimons:
> > minor nit: Bit of a disconnect when clicking the Add... button and seeing "create" as the next options. Seems like they should be consistent. Maybe a Kari question. image
> 
> Let me ask her for her thoughts here. I’ll change it to “Create…” for now

Wondering if we should also use consistent language on the Bandits listing page.
image

bryce-fitzsimons:
Is there a way to use a template when creating an experiment from feature rule?

msamper:
> Is there a way to use a template when creating an experiment from feature rule?

I didn't include that in this PR, but if its a similar create experiment flow it should be super easy to add

bryce-fitzsimons:
> > Is there a way to use a template when creating an experiment from feature rule?
> 
> I didn't include that in this PR, but if its a similar create experiment flow it should be super easy to add

Your call, but I'd think that should be part of MVP if possible. Otherwise you can't really enforce mandatory templates. The flow to adding an experiment from rule is similar but not exactly the same. A lot of the form field names have subtly changed and require some translation.

msamper:
> > > Is there a way to use a template when creating an experiment from feature rule?
> > 
> > 
> > I didn't include that in this PR, but if its a similar create experiment flow it should be super easy to add
> 
> Your call, but I'd think that should be part of MVP if possible. Otherwise you can't really enforce mandatory templates. The flow to adding an experiment from rule is similar but not exactly the same. A lot of the form field names have subtly changed and require some translation.

Gotcha. I can probably just add another helper function to do the translation for this form. Let me take a look

bryce-fitzsimons (changes requested):
requesting we also add templates to feature rules, otherwise the enforceability of these templates is lost

msamper:
> minor nit: Bit of a disconnect when clicking the Add... button and seeing "create" as the next options. Seems like they should be consistent. Maybe a Kari question. image

Keeping this as "Add..." for now per Kari's recommendation.

> After further review, I think "Add" would be a good one to try. Since we have Import as an option here still, "Create" doesn't apply. "Add new" also doesn't apply. "Add" seems intuitive enough for now.

lukesonnet:
A few comments, feel free to ignore if design/you have decided against it in the past...

1. Delete should be red to be consistent with other kebab menus
Screenshot 2025-01-02 at 9 21 15 AM

2. Should the detail page exist and be linkeed even if there are 0 linked experiments? It was a bit odd for me to learn about it's existence only through the PR description. If it's clickable right away, then I'm more likely to learn the detail page exists after I create my first template.

3. In the templates list, the row is not clickable to the detail page even when you have 1+ experiments. I feel like normally the whole row is clickable in these tables.

4. I feel like the text and labels here could be a little confusing. I think most people would use these fields to create some kind of structured markdown like thing that someone fills in like a GH PR template. IDK if there's a good way to have some default text here nudge people in that direction. It probably only works if you can do multi-line default text and that might not be a good UX.
Screenshot 2025-01-02 at 9 20 19 AM

Some broader comments:

1. It's very odd to not see the template explicitly applied when you select it. Maybe this is for a follow-up but it's really hard in the creation flow to know what has changed when you apply a template.

2. This is out of scope, but QAing this laid bare how awkward our current division of settings is across flows (analysis settings, metrics, and the fields covered by the template now live in 3 different places when an experiment is in draft, and the template affects all 3).

bryce-fitzsimons (commented):
For creating/editing a template, we need a way to select nothing on many of these Select fields. Otherwise we force apply a field that maybe shouldn't have been templatized:
Screenshot 2025-01-02 at 10 22 37 AM


Compounding the issue, some of these fields start off as unset (which is fine), but once you've set them, you can never un-set them:
Screenshot 2025-01-02 at 10 23 37 AM


Things like "activation metric" should distinguish between "none" and "unset"

Some of this is quite tricky and might require larger UI changes, so perhaps we need to officially slate some of it for a fast follow, but I don't want it to get lost.
Screenshot 2025-01-02 at 10 24 38 AM

bryce-fitzsimons:
I agree with Luke's points 1, 2, and 3 as high priority:
> 1. Delete should be red to be consistent with other kebab menus
>    Screenshot 2025-01-02 at 9 21 15 AM
> 2. Should the detail page exist and be linkeed even if there are 0 linked experiments? It was a bit odd for me to learn about it's existence only through the PR description. If it's clickable right away, then I'm more likely to learn the detail page exists after I create my first template.
> 3. In the templates list, the row is not clickable to the detail page even when you have 1+ experiments. I feel like normally the whole row is clickable in these tables.

bryce-fitzsimons:
The templates table is styled differently from the experiments list table and the template detail table. Mostly highlight colors etc

packages/front-end/hooks/useTemplates.ts
```diff
@@ -0,0 +1,38 @@
...
+  ExperimentTemplateInterface,
+} from "back-end/types/experiment";
+import { useMemo } from "react";
+import useApi from "./useApi";
+import { useExperiments } from "./useExperiments";
+
+export function useTemplates(project?: string) {
+  const { data, error, mutate } = useApi<{
+    templates: ExperimentTemplateInterface[];
+  }>(`/templates?project=${project || ""}`);
```
bryce-fitzsimons, line 12: When I have a project selected, I am not able to use any templates that were set to "all projects"

bryce-fitzsimons (changes requested):

packages/front-end/hooks/useTemplates.ts
msamper: Should be fixed now. Was filtering the options for the selector incorrectly

msamper:
@lukesonnet Addressed 1,2, and 3 from your list since they seem like the highest priority. Agree with #4 and I'll need to chat with Kari about how we can make it more obvious to folks that they can use those fields for something like a Github PR template so folks just have to fill in some blanks. For the broader comments, we made a decision for now that for MVP it was fine if we don't make it clear what has been applied from the template, but I agree it can be awkward especially when the template doesn't include fields from the first step in the modal. I want to do some more thinking on that for the next iteration of this. 

@bryce-fitzsimons Yeah I agree that selecting nothing for templates could be helpful. However, given that this is using a flow very similar to the experiment creation flow where org defaults are used, I didn't think it was too pressing given that if the user doesn't modify the selection for the template it should be the same value they'd see when creating the experiment anyway. If we want to change that, I think we need to think more broadly about how that would affect the way a template is applied to the new experiment form. Unsetting I think might be more important because I can see a user selecting something and then realizing they don't want to provide a preset for the template and having to start over. I think we can chat a bit more about this and handle it in some fast follows

bryce-fitzsimons:
add an edit template button on the [tid] page?
image

packages/front-end/hooks/useTemplates.ts
bryce-fitzsimons: Still not working for me:
1. For feature rules, I get no "all project" templates if my feature flag has a project.
2. Similar problem when I've selected a project in the left-nav and am visiting the "templates" listing page. The "all projects" templates do not show here.

packages/front-end/components/Experiment/Templates/TemplatesPage.tsx
```diff
@@ -0,0 +1,236 @@
...
+          <tr>
+            <SortableTH field="templateName">Template Name</SortableTH>
+            <SortableTH field="templateDescription">Description</SortableTH>
+            <SortableTH field="tags">Tags</SortableTH>
+            {showProjectColumn && (
+              <SortableTH field="project">Project</SortableTH>
+            )}
+            <SortableTH field="dateCreated">Created</SortableTH>
+            <SortableTH field="usage">Usage</SortableTH>
+            <th />
```
bryce-fitzsimons, line 105:
```suggestion
            <th style={{ width: 50 }} />
```

packages/front-end/hooks/useTemplates.ts
msamper: Fixed 1, I forgot to modify the filteredTemplates list for the overview page feature flags use.

For 2, I thought that was expected behavior because of how the Experiments list works, but I see that we follow that pattern for other things like environments and SDK connections so makes sense that we'd want that for this too

msamper: Fixed 2

packages/front-end/pages/experiments/index.tsx
```diff
@@ -327,13 +353,38 @@ const ExperimentsPage = (): React.ReactElement => {
...
+      {canAddTemplate && (
+        <PremiumTooltip commercialFeature="templates">
+          <DropdownMenuItem
+            onClick={() => setOpenTemplateModal({})}
+            disabled={!hasTemplatesFeature}
+            className="w-100"
+          >
+            Create Template
```
bryce-fitzsimons, line 376: creating a template from the "templates" tab did not properly mutate() and show the newly-created template on the list of templates

msamper: Hmm I'm not able to replicate this

msamper: Can you record a quick loom if you're still seeing it so I can try to replicate and debug?

bryce-fitzsimons: https://www.loom.com/share/8a1306dc23424d088f8b43fe25c1098a?sid=f36a45dc-5e3f-4e6f-aab2-28a134161c22

msamper: Thank you! Should be fixed now

packages/back-end/src/routers/experiment-template/template.controller.ts
```diff
@@ -0,0 +1,152 @@
...
+  ExperimentTemplateInterface,
+  UpdateTemplateProps,
+} from "./template.validators";
+
+export const getTemplates = async (
+  req: AuthRequest<unknown, unknown, { project?: string }>,
+  res: Response<{ status: 200; templates: ExperimentTemplateInterface[] }>
+) => {
+  const context = getContextFromReq(req);
+  const { project } = req.query;
```
bryce-fitzsimons, line 19: remove project? clean up model methods (remove getByProject)?

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -420,6 +448,25 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+      const templateAsExperiment = convertTemplateToExperiment(template);
+      form.reset(templateAsExperiment, {
+        keepDefaultValues: true,
+      });
+    }
+  }, [initialValue, isNewExperiment, isImport, isBandit]);
+
+  const templateRequired =
+    hasCommercialFeature("templates") &&
+    !isBandit &&
```
msamper, line 466: Add `!isImport`

bryce-fitzsimons:
are all of these templates supposed to be visible, including the ones inside of projects? I currently have "all projects" selected:
image

msamper:
> are all of these templates supposed to be visible, including the ones inside of projects? I currently have "all projects" selected: image

I'd think so! As long as you have read access to the project resource. I do that filtering on the controller for getting all templates

bryce-fitzsimons (approved):

PR #3444, opened 2024-12-17
rf/docs-strapi

royalfig:
Adds a Strapi tutorial to our docs


docs/docs/guide/strapi.mdx
```diff
@@ -0,0 +1,532 @@
...
+    ? growthbookExperiment.data.find((el) => el.featureFlagId === featureFlagId)
+    : growthbookExperiment.data;
+
+  if (!experimentData || !isExperimentData(experimentData)) {
+    return null;
+  }
+
+  // Evaluate the feature flag in the current context to get the variation index
+  const result = gb.evalFeature(featureFlagId).experimentResult;
+  const variationIndex = result?.variationId;
```
jdorn, line 285: I don't like this.  It's using a feature flag, but then only works if you add a specific experiment rule.  Also, the feature flag values are ignored.  I don't think we want to encourage this.

Some alternatives:

1. Skip feature flags entirely and use inline experiments instead with `gb.run(...)`.  https://docs.growthbook.io/lib/js#inline-experiments
2. Use a `number` feature flag and store the array index as the value.
3. Instead of an array of children in Stripe, store an array of objects, each with an id.  Use those ids as the feature flag values.

royalfig: Makes sense! What do you think about using option 3:

## Strapi
In Strapi, create a GrowthBook container that has a `featureFlagId` that connects to GrowthBook and a `variations` slot, in which you can put whatever components you want, as long as they have a `variationId`. 

CleanShot 2024-12-21 at 17 45 23

## GrowthBook
Use a feature flag with `number` that maps to the `variationId` and `featureFlagId` above.

CleanShot 2024-12-21 at 17 50 41

## Next
The `getVariation` function returns `null` if the feature flag isn't live, otherwise it maps the component `variationId` from Strapi to the flag evaluation result from GrowthBook. This lets users employ feature flagging and experiments.
CleanShot 2024-12-21 at 17 51 49

docs/docs/guide/strapi.mdx
```diff
@@ -0,0 +1,230 @@
...
+3. `featureFlagKey`: The feature flag key from GrowthBook that [we created above](#create-feature-flags-and-ab-tests-in-growthbook).
+
+In your Next.js app, add the following code:
+
+```tsx
+import { type GrowthBook } from "@growthbook/growthbook";
+
+type Props = {
+  gb: GrowthBook;
+  variations: unknown[] | null;
```
jdorn, line 124: Should we make Props generic? 

```ts
type Props<T> = {
  gb: GrowthBook;
  variations: T[] | null;
  featureFlagKey: string;
}
```

jdorn (approved):

PR #3445, opened 2024-12-17
Shareable experiments

bryce-fitzsimons:
Makes entire experiments/bandits shareable (not shared by default)

share link visible when experiment is in a "stopped" state:
image

pops a modal, nearly identical to shareable reports:
image

if user does not have permissions to change share status, they just get a "copy link" button instead (or if the report is not shared, no button is rendered):
image

You can also share (2 ways!) from the dropdown menu:
image

Public shared report page mirrors the actual experiment page, but scrubs a bunch of extraneous or sensitive information away:
image

image

Also works for Bandits:
image

### DIFF
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -291,6 +291,15 @@ app.get(
   }),
   reportsController.getReportPublic
 );
+// public shareable experiments
+app.get(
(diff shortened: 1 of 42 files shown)


Kevin-Chant (approved):
L G T M, :ship: 'er public

PR #3452, opened 2024-12-18
Ensure validity query respects test query length setting

lukesonnet:
### Features and Changes

When editing an EAQ, even if you uncheck "test query before saving" in the SQL edit modal, it does run a validity check when you finally save it. This last test query doesn't respect the test days window, and I believe it should. 

This PR enables that.

We should also consider letting users disable this validity query if they want.

### Testing

Tested on my dev instance that when saving edits to an EAQ, it now uses the org setting for number of test days rather than the default of 30.


lukesonnet:
ok, let me see if I can add that here

lukesonnet:
It's non-trivial. So I'm skipping it for now. Merging this change, however.

PR #3453, opened 2024-12-18
Send feature usage data to Clickhouse and show feature usage graphs.

tzjames:
### Features and Changes

- Sends feature-usage events to the ingestor
- Modifies the materialized view for events to not copy in feature-usage events
- Creates a materialized view for feature-usage events
- Creates backend code to query for feature usage
- Adds Feature Usage Graphs behind a feature flag

### Dependencies
Need to either destroy and recreate the growthbook datasource, or manually modify and create the feature-usage materialized view

### Testing
- yarn dev in growthbook-ingestor/packages/ingestor
- in .env.local set: `INGESTOR_HOST=http://localhost:3003`
- in .env.local set: `CLICKHOUSE_MAIN_TABLE=test_enriched_events`
- Create a feature flag with the same id as prod such as "bandits" that gets called when loading pages
- Go to that feature and see the feature usage graphs.

### Screenshots
Screenshot 2024-12-18 at 6 31 56 PM
Screenshot 2024-12-18 at 6 31 44 PM


packages/front-end/components/Features/FeatureUsageGraph.tsx
```diff
@@ -0,0 +1,163 @@
...
+    "featureUsageLookback",
+    "15minute"
+  );
+
+  const useFeatureUsage = growthbook.isOn("feature-usage");
+
+  const { data: featureUsage, mutate: mutateFeatureUsage } = useApi<{
+    usage: FeatureUsageData;
+  }>(`/feature/${featureId}/usage?lookback=${lookback}`, {
+    shouldRun: () => useFeatureUsage,
```
jdorn, line 49: We should also only run this if the org has a valid growthbook data source.  Otherwise, the API calls are just going to 4xx immediately.

packages/front-end/components/Features/FeatureUsageGraph.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  useEffect(() => {
+    if (!featureUsage) return;
+    if (lookback !== "15minute") return;
+    const timer = setInterval(
+      () => {
+        mutateFeatureUsage();
+      },
+      // Update slower when there's no data yet
+      featureUsage.usage?.overall?.total ? 2000 : 5000
```
jdorn, line 60: Since we're hitting ClickHouse directly without a cache layer, I'm a little worried about auto-refreshing right now.  Maybe we can add another feature flag with the interval time so we can control this remotely.  If we notice high load on ClickHouse, we can set the interval higher or set it to `0` to disable auto-refresh entirely.

That would also let us have a fast refresh on our demo account so we can show off the graphs better.

packages/front-end/components/Features/BarChart100.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  <div
+                    style={{
+                      width: 10,
+                      height: 10,
+                      background: colorScale(key),
+                    }}
+                  ></div>
+                </td>
+                <td>{key ? key : <em>unknown</em>}</td>
+                <td>{formatter.format(data[key])}</td>
```
jdorn, line 108: Might want to truncate this to avoid huge JSON blobs pushing the rest of the page content down too far.  We can show the full value in a tooltip on hover (using the HTML title attribute) so it's there if needed.

packages/front-end/services/track.ts
```diff
@@ -80,17 +80,57 @@ const DEVICE_ID_COOKIE = "gb_device_id";
...
+  events: DataWarehouseTrackedEvent | DataWarehouseTrackedEvent[]
+) => {
   if (inTelemetryDebugMode()) {
-    console.log("Telemetry Event - ", event);
+    console.log(
+      "Telemetry Event - ",
+      Array.isArray(events)
+        ? events.map((e) => e.event_name)
+        : events.event_name
+    );
```
jdorn, line 125: We should just log `events` here.  That way we can easily inspect the full array and objects in dev tools.

packages/front-end/pages/_app.tsx
```diff
@@ -71,34 +75,30 @@ function App({
...
-
-                {
-                  cache: "no-cache",
-                  mode: "no-cors",
-                }
-              )
-              .catch(() => {
-                // TODO: retry in case of network errors?
-              });
+            dataWareHouseTrackMultipleEventsWithDefaults(q[0]);
```
jdorn, line 101: Should this be `q` instead of `q[0]`?

packages/front-end/pages/_app.tsx
```diff
@@ -71,34 +75,30 @@ function App({
...
+          source: res.source,
+          value: res.value,
+          ruleId: res.source === "defaultValue" ? "$default" : res.ruleId || "",
+          variationId: res.experimentResult?.key || "",
+        };
+
         _rtQueue.push({
-          key,
-          on: res.on,
+          eventName: "feature_usage",
```
jdorn, line 92: We use `Experiment Viewed` for the other event.  I think we should follow the same pattern and do `Feature Evaluated` here as the event name.

packages/front-end/pages/_app.tsx
tzjames: No.  We set `const q = [_rtQueue];` as a cheap way to copy the queued events. `_rtQueue` itself is the list of events, so we need to pass in `q[0]`

jdorn (approved):

Issue #3456, opened 2024-12-19
[Bug] Test SQL: Scroll not working 


natasha-growthbook:
### Summary

When testing an SQL query the scroll does not work in the popup, instead the background UI will scroll.

This can be replicated when testing the SQL for the Fact Tables, Metrics and Experiment assignment queries

https://github.com/user-attachments/assets/e4c19f5b-c582-466b-845f-2cd1fa2ee252

#### Expected Behavior

Using the scroll should scroll the popup UI and not the background UI so should be consistent behaviour as shown when testing the SQL query under fact tables > saved filters

#### Current Behavior

When testing an SQL query the scroll does not work in the popup, instead the background UI will scroll.

This can be replicated when testing the SQL for the Fact Tables, Metrics and Experiment assignment queries

PR #3457, opened 2024-12-19
Test SQL: Fix scroll not working

mknowlton89:
### Features and Changes

When we made the switch to Radix Tabs, I think a UI issue was introduced that broke the scrolling within the `DisplayTestQueryResults`. With this change, we're back to the experience we previously had where any overflow is visible via a scroll. Additionally, when rendering the results, I've made it so the table header row is sticky so the user doesn't have to scroll up to remember which column they're looking at.

- Closes **(https://github.com/growthbook/growthbook/issues/3456)**

### Testing
 
- [x] Ensure that when viewing the results and the rendered SQL, there is both a horizontal and vertical scroll when necessary
- [x] Ensure that everything works on both large and small screens

### Loom Walkthough
https://www.loom.com/share/5e6ce23655814051b5ecaae71294ac01

### DIFF
--- a/packages/front-end/components/SchemaBrowser/EditSqlModal.tsx
+++ b/packages/front-end/components/SchemaBrowser/EditSqlModal.tsx
@@ -312,15 +312,13 @@ export default function EditSqlModal({
               />
             </div>
             {testQueryResults && (
-              <div className="" style={{ flex: 1, maxHeight: "45%" }}>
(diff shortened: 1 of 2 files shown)


bryce-fitzsimons (approved):

ahdriel:
@mknowlton89 Thank you so much for the fix! And my bad for missing it when doing the update 🙇

PR #3459, opened 2024-12-20
Improve order when sorting by Status

ahdriel:
### Features and Changes

- Improve the order of the experiments when sorting by Status
- Add `min-width` to the Status column to avoid content jumping around when changing sort order

### Screenshots

| Before | After |
|--------|--------|
| descending-before | descending-after |
| ascending-before | ascending-after |


gazzdingo (approved):
LGTM

PR #3461, opened 2024-12-20
hotfix for "stopped" sub-statuses on metric pages

bryce-fitzsimons:
On metric pages, we list experiments and also show those experiments' statuses. These statuses were not picking up the correct stopped state and were defaulting to "Stopped: Awaiting decision" even for won/lost/etc.


gazzdingo (approved):
Approving to unblock code looks good but I don't have much context

ahdriel (approved):
Thanks for the fix!

PR #3462, opened 2024-12-20
Introduce Edit Info Modal for Experiments

mknowlton89:
### Features and Changes

This PR introduces a new "Edit Info" modal for Experiments - this modal is now where organizations can edit an Experiment's project, tags, owner, and experimentKey (_if the experiment is in draft mode_). 

Given that this new edit path exists, we've now removed a user's ability to edit an experiment's project or tags (if any are previously set) in-line. We have protected a user's ability to add a project or tags inline if none exist. However, unlike previously where this opened specific modals for adding a project or adding tags, this now simply opens the Edit Info modal.

### Testing

- [x] Ensure that you can add a project to an experiment if one doesn't already exist inline
- [x] Ensure that you can add tags to an experiment if none exist
- [x] Ensure that you can not update a project inline if one already exists
- [x] Ensure that you can not update tags inline if one already exists
- [x] Ensure that you can not edit an experiment's name in line, you have to go to MoreMenu > Edit Info in order to do so
- [x] Ensure that from the Edit Info modal you can edit an experiments name
- [x] Ensure that from the Edit Info modal you can edit an experiment's key only if it is in draft state
- [x] Ensure that from the Edit Info modal you can edit an experiment's tags
- [x] Ensure that from the Edit Info modal you can edit an experiment's project
- [x] If you do change an experiment's project, ensure the warning pops up about why this could cause issues
- [x] Ensure this works on both light mode and dark mode

### Screenshots
Light mode menu when no project or tags are set (note the "+ Add") cta
Screenshot 2025-01-14 at 10 14 18 AM

Light mode menu when project and tags are set (note that there is no longer an "edit" CTA)
Screenshot 2025-01-14 at 10 14 38 AM

Light mode for draft experiment - note that the `experimentKey` field is enabled. This screenshot was also taken as the user was trying to change the experiment's project, so note the warning
Screenshot 2025-01-14 at 10 14 31 AM

Dark mode for running experiment - not that the `experimentKey` field is disabled
Screenshot 2025-01-14 at 10 15 01 AM


packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -724,300 +706,266 @@ export default function ExperimentHeader({
...
+                    setDropdownOpen(false);
+                  }}
+                >
+                  Share {isBandit ? "Bandit" : "Experiment"}
+                </DropdownMenuItem>
+              )}
+              {showShareableReportButton ? (
+                <DropdownMenuItem
+                  onClick={async () => {
+                    await handleWatchUpdates(!isWatching);
```
mknowlton89, line 888: This is incorrect after merging in main. Need to fix

packages/front-end/components/Tags/TagsInput.tsx
```diff
@@ -87,25 +84,35 @@ const TagsInput: FC<{
         },
       };
     },
+    control: (styles, { isFocused }) => {
+      return {
+        ...styles,
+        boxShadow: `0px 0px 0px 1px ${isFocused ? "var(--violet-8)" : "000"}`,
+      };
+    },
```
mknowlton89, line 92: Need to remove this as I'm handling this in a separate PR [here](https://github.com/growthbook/growthbook/pull/3509).

packages/front-end/services/tags.tsx
```diff
@@ -57,6 +57,38 @@ export const TAG_COLORS_MAP: Record<NonNullable<RadixColor>, string> = {
...
+  cyan: cyan.cyan5,
+  teal: teal.teal5,
+  jade: jade.jade5,
+  green: green.green5,
+  grass: grass.grass5,
+  lime: lime.lime5,
+  mint: mint.mint5,
+  sky: sky.sky5,
+};
+
```
mknowlton89, line 91: Need to remove this as I'm handling this in a separate PR [here](https://github.com/growthbook/growthbook/pull/3509).

packages/front-end/styles/global.scss
```diff
@@ -3173,6 +3173,10 @@ td.disabled {
   cursor: not-allowed;
   background-color: var(--disabled-background) !important;
 }
+.form-control:focus {
+  border-color: #{colors.$gray-400};
+  box-shadow: 0px 0px 0px 1px var(--violet-8);
+}
```
mknowlton89, line 3179: Need to remove this as I'm handling this in a separate PR [here](https://github.com/growthbook/growthbook/pull/3509).

packages/front-end/components/Experiment/EditExperimentNameForm.tsx
```diff
@@ -1,44 +0,0 @@
-import { FC } from "react";
```
mknowlton89, line 1: This is now handled by the EditExperimentInfoModal exclusively

packages/front-end/services/tags.tsx
```diff
@@ -57,6 +57,38 @@ export const TAG_COLORS_MAP: Record<NonNullable<RadixColor>, string> = {
   sky: sky.sky11,
 };
 
+export const TAG_COLORS_BACKGROUND_MAP: Record<
```
bryce-fitzsimons, line 60: These big arrays look like a code smell to me. Naive question, but could we just do something like this...

```
import * as radixColors from "@radix-ui/colors";
...
const color = `${radixColors[c]}11`;
const bgcolor = `${radixColors[c]}5`;
```

bryce-fitzsimons (approved):
didn't QA closely, but looks good from a code standpoint. minor nits about hardcoded color arrays

packages/front-end/services/tags.tsx
mknowlton89: Oooh yea - that's slick. Thanks for the suggestion!

ahdriel: Another approach (more for knowledge sharing, not necessarily a request for change) would be to use `background-color: var(--accent-5)` in CSS.

And the accent color can be controlled by the parent `data-accent-color` attribute like `data-accent-color="amber"`.

This is how the Radix components can update their colors dynamically based on the `color` prop that we pass in.

mknowlton89: Thanks, @ahdriel - that feels a bit more readable. I ultimately landed on a hybrid, the syntax below maps to the collect accent color.

```ts
const color = findClosestRadixColor(data.color) || "#029dd1";
backgroundColor: `var(--${color}-a3)`
```

PR #3467, opened 2024-12-20
Apply enterprise license check for environment inheritance

Kevin-Chant:
### Features and Changes

The new logic for handling environment parents is meant to be only for enterprise customers, but initially didn't have a license check. This PR adds the check to the validation helper on the only entrypoint for environment inheritance.
Also adds a description to the API documentation for `parent` to clarify its usage and that it is only for enterprise users.

### Testing

- [ ] Enterprise org can create environment with a parent and the behavior still works as expected
- [ ] Free org gets a permission error trying to create an environment with a parent
- [ ] Pro org gets a permission error trying to create an environment with a parent
- [ ] Updating an existing environment to try to add a parent ignores the parent key from the payload (inheritance is only relevant when creating an environment)

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -3359,6 +3359,7 @@ paths:
                     type: string
                 parent:
                   type: string
+                  description: An environment that the new environment should inherit feature rules from. Requires an enterprise license
       responses:
         '200':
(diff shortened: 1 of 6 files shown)


gazzdingo (approved):
LGTM

PR #3468, opened 2024-12-20
Log (don't throw) error when loading an invalid/old-legacy report

bryce-fitzsimons:
...otherwise "list all" type endpoints will cancel.

### DIFF
--- a/packages/back-end/src/models/ReportModel.ts
+++ b/packages/back-end/src/models/ReportModel.ts
@@ -10,6 +10,7 @@ import {
 } from "back-end/types/report";
 import { ReqContext } from "back-end/types/organization";
 import { ApiReqContext } from "back-end/types/api";
+import { logger } from "back-end/src/util/logger";
 import { getAllExperiments } from "./ExperimentModel";
 import { queriesSchema } from "./QueryModel";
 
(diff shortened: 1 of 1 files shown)


packages/back-end/src/models/ReportModel.ts
```diff
@@ -57,7 +58,17 @@ const toInterface = (doc: ReportDocument): ReportInterface => {
...
+      logger.error(
+        `Invalid report type: [${
+          (doc as ExperimentReportDocument)?.type ?? ""
+        }] (id: ${(doc as ExperimentReportDocument)?.id ?? "unknown"})`
+      );
+      return migrateExperimentReport(
+        omit(
+          (doc as ExperimentReportDocument).toJSON<ExperimentReportDocument>(),
+          ["__v", "_id"]
+        )
```
msamper, line 70: Is migrating the report safe if it's an invalid report?

bryce-fitzsimons: yes, worst case scenario you can't load it depending on how broken the model is. If it's only a little bit broken, the migration should mostly work.

msamper (approved):

PR #3469, opened 2024-12-20
update text-purple and link-purple to more closely match our radix styles

bryce-fitzsimons:


mknowlton89 (approved):
Should we also add the `link-purple` class on the `<a>` tags for the `items`, e.g. on line 238.

PR #3473, opened 2024-12-24
Added the ability to make a metric custom field

Auz:
Allows for custom fields to be of type "metrics" and let the user select from existing metrics. This functionality allows for users to add, for example, a field for OEC, and then let the creator of the experiment specify what metric(s) that is.

### DIFF
--- a/packages/back-end/src/routers/custom-fields/custom-fields.validators.ts
+++ b/packages/back-end/src/routers/custom-fields/custom-fields.validators.ts
@@ -13,6 +13,7 @@ export const customFieldTypes = z.enum([
   "boolean",
   "date",
   "datetime",
+  "metrics",
 ]);
 
 export const customFieldsPropsValidator = z.object({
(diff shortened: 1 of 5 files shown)


packages/back-end/src/routers/custom-fields/custom-fields.validators.ts
```diff
@@ -13,6 +13,7 @@ export const customFieldTypes = z.enum([
   "boolean",
   "date",
   "datetime",
+  "metrics",
```
mknowlton89, line 16: Should this be singular like all of the others?

packages/front-end/components/CustomFields/CustomFieldInput.tsx
```diff
@@ -81,6 +82,7 @@ const CustomFieldInput: FC<{
         ) : (
           <>
             {availableFields.map((v, i) => {
+              console.log("current value", currentCustomFields?.[v.id]);
```
mknowlton89, line 85: Reminder to remove this before merging.

mknowlton89 (approved):
Left a few small comments. Probably makes sense to update the docs in the PR as well as the docs currently outline what `customFieldTypes` we support, so we probably want to add `metrics` to that list as well.

PR #3479, opened 2024-12-26
Let metrics tag filter show all potential metric tags in project

### LINKED ISSUE #3460: [Bug] Filter by tags on metrics is limited
When on the metric page, the only options in the tag dropdown are the ones that happen to be applied to metrics on the current page. If there are multiple pages of metrics, the tags applied to these metrics are not available for filtering. Allowing to choose from all tags regardless of if they have metrics tagged with them is probably the best solution.

lukesonnet:
### Features and Changes

- Closes #3460 

### Screenshots

Screenshot 2024-12-26 at 9 05 36 AM

Screenshot 2024-12-26 at 9 05 39 AM

### DIFF
--- a/packages/front-end/components/Metrics/MetricsList.tsx
+++ b/packages/front-end/components/Metrics/MetricsList.tsx
@@ -394,7 +394,7 @@ const MetricsList = (): React.ReactElement => {
           </div>
         )}
         <div className="col-auto">
-          <TagsFilter filter={tagsFilter} items={items} />
+          <TagsFilter filter={tagsFilter} items={filteredMetrics} />
         </div>
       </div>
       <table className="table appbox gbtable table-hover">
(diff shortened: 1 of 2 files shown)


jdorn (commented):
I would expect the tag list to be filtered to just the metrics that match the search term entered. The bug here is that it is only looking at the first page. A better solution would be to have the useSearch hook return both the paginated list as well as an unpainted list.

lukesonnet:
> A better solution would be to have the useSearch hook return both the paginated list as well as an unpaginated list.

I had this before but opted for this solution as it fits @Auz's issue solution better: "Allowing to choose from all tags regardless of if they have metrics tagged with them is probably the best solution."

lukesonnet:
Updated to use filtered items. Works as expected.

Auz (approved):

PR #3486, opened 2024-12-28
Fix report permission checks

bryce-fitzsimons:
- was not respecting `editLevel` for edit permissions
- more BE gating
- prevent useEffect from PUTting share updates when permissions are invalid

also fixes a small UI bug with floating table headers having too much top offset


Kevin-Chant (approved):
LGTM

PR #3487, opened 2024-12-30
Add frontend entrypoint for environment inheritance

Kevin-Chant:
### Features and Changes

* Defines a set of default environment ids: production, dev, staging, and test
* Changes the Create Environment modal's ID input to a select field which allows creating custom entries
* Adds a new field to the Create Environment modal for specifying a parent environment
* Disables manual parent inheritance for custom environment ids; still allows API to use arbitrary environment IDs (with enterprise license)

### Testing

On the [environments page](http://localhost:3000/environments) click to Add Environment and test the behavior of the ID and Parent fields. Try creating a custom ID that follows the required pattern and one that doesn't, and see that the parent field is only enabled for the 4 default environments

### Screenshots

Empty state when opening modal

Allowed inheritance for default environment

Controller enforces inheritance only working for default environments

### DIFF
--- a/packages/front-end/components/Settings/EnvironmentModal.tsx
+++ b/packages/front-end/components/Settings/EnvironmentModal.tsx
@@ -177,11 +177,8 @@ export default function EnvironmentModal({
               form.setValue("parent", value || undefined);
             }}
             options={environments.map((e) => ({ label: e.id, value: e.id }))}
-            isClearable={true}
-            disabled={
-              !!existing.id ||
-              !DEFAULT_ENVIRONMENT_IDS.includes(form.watch("id") || "")
-            }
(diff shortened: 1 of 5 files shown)


packages/front-end/components/Settings/EnvironmentModal.tsx
```diff
@@ -147,6 +168,34 @@ export default function EnvironmentModal({
         placeholder=""
         textarea
       />
+      {!existing.id && (
+        <div className="mb-3">
```
gazzdingo, line 172: Nit: we should stop using Bootstrap for spacing

gazzdingo (approved):
LGTM

PR #3485, opened 2024-12-27
Abstract Feature Rules to separate component

Kevin-Chant:
### Features and Changes

Moves the existing jsx for the Feature Rules panel to its own component to make it easier to change without affecting the rest of the file. This is a separate PR to make it easier to review the changes for the new Compare Environments option without the clutter of this extra refactor

The only behavioral change is swapping our existing ControlledTabs to Radix's Tabs to match designs

Note: This PR is much easier to review when looking at each commit separately

### Testing
Manually tested the frontend to look for any behavioral issues with switching tabs or any of the actions within a tab

Working based on the designs in figma [here](https://www.figma.com/design/c650Wgk7wRBkDyJ5lryzta/Compare-Environments?node-id=107-3300&p=f&t=fSY2Gdt7BJpuHwsP-0)

#### Before
image
image

#### After
image
image

PR #3490, opened 2025-01-02
Compare Feature Environments

Kevin-Chant:
### Features and Changes

Adds a button to the feature page to compare the rules between two environments and overwrite one with the other.
Exposes a new endpoint to simplify that functionality + a new helper for the FeatureModel.

### Dependencies

- [x] Should merge #3485 first to make reviewing the diff for these changes much cleaner

### Testing
On the [fid] feature page, use the "Compare environments" button to pop the modal and inspect the behavior when comparing various envs + overwriting the rules

#### Less than 6 environments - tabs
image

#### 6+ environments - dropdown
image

#### Modal without target selected
image

#### Modal with diff
image

#### Environments are still scrollable when not in a dropdown
image


gazzdingo (approved):
LGTM

PR #3491, opened 2025-01-03
Update event webhook doc to include guardrail metric

royalfig:
Add guardrail metrics to webhook definition


docs/src/partials/event-webhook/_event-webhook-list.md
```diff
@@ -1299,7 +1299,7 @@ Triggered when a warning condition is detected on an experiment
 
 ### experiment.info.significance
 
-Triggered when a goal metric reaches significance in an experiment (e.g. either above 95% or below 5% chance to win). Be careful using this without Sequential Testing as it can lead to peeking problems.
+Triggered when a goal or guardrail metric reaches significance in an experiment (e.g. either above 95% or below 5% chance to win). Be careful using this without Sequential Testing as it can lead to peeking problems.
```
romain-growthbook, line 1302: These are auto-generated: https://github.com/growthbook/growthbook/blob/974ff5914b7fbef177be21f1a788610bfff4ffb2/packages/back-end/src/events/base-types.ts#L71

Let me know if you need help with it. I believe all you should have to do is edit the code and let the `husky` commit hook to the job.

romain-growthbook (changes requested):
These docs are auto-generated! We should edit the source or else the change will be overwritten.

romain-growthbook (approved):

PR #3502, opened 2025-01-07
chore: Wrap all our Radix-wrappers in forwardRef

ahdriel:
### Features and Changes

As the [Radix documentation](https://www.radix-ui.com/primitives/docs/guides/composition#your-component-must-forward-ref) states, for composition with Radix elements we need to forwardRef, so we are updating all components to do it.

This happened to me when trying to use Tooltip, which was not working before forwarding the ref.


packages/front-end/components/Radix/Select.tsx
```diff
@@ -53,24 +56,25 @@ export function Select({
...
+  HTMLDivElement,
+  {
+    value: string;
+    children: string | string[] | ReactNode;
+    disabled?: boolean;
+  }
+>(function SelectItem({ value, children, disabled = false }, ref) {
   return (
-    <RadixSelect.Item value={value} disabled={disabled}>
+    <RadixSelect.Item value={value} disabled={disabled} ref={ref}>
```
mknowlton89, line 70: Should we also be spreading in the props here?

packages/front-end/components/Radix/Metadata.tsx
```diff
@@ -27,10 +32,9 @@ const Metadata: FC<{
     }
   };
   return (
-    <Flex gap="1">
+    <Flex gap="1" ref={ref}>
```
mknowlton89, line 35: Should we spread in the props here?

packages/front-end/components/Radix/HelperText.tsx
```diff
@@ -63,24 +63,22 @@ export function RadixStatusIcon({
...
+  {
+    children: string | string[];
+    status: Status;
+    size?: "sm" | "md";
+  } & MarginProps
+>(function HelperText({ children, status, size = "md", ...otherProps }, ref) {
   return (
     <Text color={getRadixColor(status)} size={getRadixSize(size)}>
-      <Flex gap="1" {...otherProps}>
+      <Flex gap="1" {...otherProps} ref={ref}>
```
mknowlton89, line 76: Less feedback and more of a question - it's not initially clear to me how it's decided which element gets the ref passed to it and the props spread into it - do you have any insight on this?

mknowlton89 (approved):
LGTM - Tested all of the components and didn't see any regressions introduced. Left 1 questions for myself.

packages/front-end/components/Radix/HelperText.tsx
ahdriel: I think it usually is the one that 'makes more sense'. So if we are doing a Input that has a text for a helper and other components, we would expect the ref to be the input. But I also see an argument for it being the whole container in the scenarios where what we want to do is related to calculating its size/positioning.

So yeah, I am also not super sure. Right now I decided to mainly follow where the overall props are being spread, or adding the ref to the internal Radix component.

PR #3503, opened 2025-01-08
chore: Move Power calculation code to shared

ahdriel:
### Features and Changes

As we are working on adding new features to power calculation (see #3421), a pre-req is to have this code available for both front-end and back-end, therefore this PR.

### Testing

Tests pass


ahdriel:
The first commit is just moving files around (and creating the necessary files for imports) and the [second commit](https://github.com/growthbook/growthbook/pull/3503/commits/0b99ab517c1e0fc9349c521648744aa6da260720) has some changes to how the code is structured which should still have the same behavior, but it is a code change.

Kevin-Chant (approved):
LGTM

PR #3507, opened 2025-01-09
Update Streaming section for React Native apps in react.mdx

valeriobelli:
### Features and Changes

I updated the React Native streaming section by redirecting it to the React Native documentation. The current documentation tells us to use `eventsource`, which is incompatible with React Native apps.

I also considered removing it, given that the React Native section can be accessed from the sidebar, but I think this change can be a compromise solution to help users discover the related document.


jdorn:

PR #3511, opened 2025-01-09
Updates Dockerfile's Poetry Install

mknowlton89:
On 1/5/25, Poetry released [version 2.0](https://github.com/python-poetry/poetry/releases/tag/2.0.0) which, according to the docs formally remove support for the previously deprecated `--no-dev` install command option which was causing our preview environment builds to fail. [Link to a failed build](https://github.com/growthbook/growthbook/actions/runs/12692299174/job/35387324944?pr=3509).

The error message we received was:
> 16.51 The option "--no-dev" does not exist

This PR locks the Dockerfile's Poetry version to `1.8.5` while also switching from using the `--no-dev` install command option to the `--without dev` option.

Poetry version 2.0 made some updates to how you can [manage dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups) - I did explore not locking the version, but it seems the way we're currently defining the dependency groups within `pyproject.toml` isn't compatible with 2.0 and specifically the `--without dev` install command option.

If we don't want to lock the version to 1.8.5 and instead want to leverage 2.0, let me know and I can explore updating the pyproject.toml further so we can ensure we're not installing and dev group dependencies in prod.

> [!CAUTION]
> I am a bit confused as to why this is only failing for the preview environment - it doesn't seem to be affecting the actual Dockerfile packaging (from what I can tell)


ahdriel (approved):
Not familiar with the stats package but approving as I think this change is low-risk and will unblock our Preview env.

lukesonnet (approved):

PR #3512, opened 2025-01-09
Update custom reports docs

royalfig:
Update custom reports/sharing docs to reflect the improvements recently made to the app


bryce-fitzsimons (approved):
LGTM. Wondering if we should call out that this also works for Bandits, or is it fairly obvious already?

PR #3518, opened 2025-01-10
Change demo exp. to have single goal metric

lukesonnet:
### Features and Changes

To be more in line with our recommendations now that we have secondary metrics, this changes our demo experiment to have only 1 goal metric.

### Screenshots

Adding the sample experiment works fine, runs well, and has the right metrics in the right place.

Screenshot 2025-01-10 at 10 00 16 AM

### DIFF
--- a/packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
+++ b/packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
@@ -269,6 +269,7 @@ export const postDemoDatasourceProject = async (
       | "description"
       | "datasource"
       | "goalMetrics"
+      | "secondaryMetrics"
       | "project"
       | "hypothesis"
       | "exposureQueryId"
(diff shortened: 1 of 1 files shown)


packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
```diff
@@ -289,7 +290,9 @@ spacing and headings.`,
       owner: ASSET_OWNER,
       datasource: datasource.id,
       project: project.id,
-      goalMetrics: metrics
+      goalMetrics: metrics.slice(0, 1).map((m) => m.id),
```
ahdriel, line 293: oh this is actually quite easy to do as you said 😂  -- nice!

ahdriel (approved):

packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
lukesonnet: The first one made sense as the goal, and this was the simplest way to do it.

PR #3526, opened 2025-01-14
Fix useEffect for loading a template into NewExperimentForm

msamper:
### Features and Changes

Fixes a bug where clicking `Create Experiment` from a template detail page doesn't properly fill out the `NewExperimentForm`. This PR also eliminates `useTemplates` and moves the logic to `useDefinitions` to leverage the fact that we can guarantee that `getTemplateById` will be loaded in on the initial render of the `NewExperimentForm`. Thanks to this we can avoid needing to have the template Map in the dependency array for the useEffect which can be dangerous since we're doing a form reset inside of that.

### Testing

Should functionally be the same to existing template behavior aside from the bug fix


packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -451,13 +451,7 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
 
   // If a template id is provided as an initial value, load the template and convert it to an experiment
   useEffect(() => {
-    if (
-      initialValue?.templateId &&
-      initialValue?.templateId !== form.watch("templateId") &&
```
msamper, line 456: This line was stopping the conditional from passing

packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -303,6 +340,7 @@ export const DefinitionsProvider: FC<{ children: ReactNode }> = ({
       },
       mutateDefinitions: async (changes) => {
         await mutate(Object.assign({ ...data }, changes), true);
+        await mutateExperiments();
```
msamper, line 343: This feels like a bit of a strange behavior for `mutateDefinitions`, but I need to mutate experiments to get `getTemplateExperimentsById` to update. Should I just manually `mutateExperiments` as well when needed from the component using this hook?

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -417,7 +414,7 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
     refreshWatching();
 
     data.tags && refreshTags(data.tags);
-    data.templateId && refreshTemplates();
+    data.templateId && mutate();
```
jdorn, line 417: Why do we need to mutate here? We aren't changing the template itself at all.

packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -21,9 +21,14 @@ import { ExperimentMetricInterface, isFactMetricId } from "shared/experiments";
...
 import { MetricGroupInterface } from "back-end/types/metric-groups";
 import { CustomField } from "back-end/types/custom-fields";
+import {
+  ExperimentInterfaceStringDates,
+  ExperimentTemplateInterface,
+} from "back-end/types/experiment";
 import useApi from "@/hooks/useApi";
 import { useLocalStorage } from "@/hooks/useLocalStorage";
 import LoadingOverlay from "@/components/LoadingOverlay";
+import { useExperiments } from "@/hooks/useExperiments";
```
jdorn, line 31: Do not use this. Fetching all experiments is very expensive and should not be done like this on every page load in the app.

packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -61,6 +67,10 @@ type DefinitionContextValue = Definitions & {
   getFactMetricById: (id: string) => null | FactMetricInterface;
   getExperimentMetricById: (id: string) => null | ExperimentMetricInterface;
   getMetricGroupById: (id: string) => null | MetricGroupInterface;
+  getTemplateById: (id: string) => null | ExperimentTemplateInterface;
+  getTemplateExperimentsById: (
```
jdorn, line 71: Remove this from definitions. The only thing in definitions should be the templates themselves, not the experiment mappings.

jdorn (changes requested):

PR #3527, opened 2025-01-14
Update Pre-Launch Checklist with Radix Components

mknowlton89:
### Features and Changes

This PR updates the `Pre-Launch Checklist` with Radix UI components (Checkboxes, Badge) and updates the links to purple so they match the rest of the application.

While I was in there, I added a few quality of life improvements including:
- The open/closed state of the pre-launch checklist collapse is now stored in localStorage like we do for the experiment's description
- I added a loading spinner that displays at the end of the checklist item to indicate we're updating the checked state
- I also updated the function that makes the updateChecklist API call so that we're awaiting the `mutate()` call before we update the loading state. There was a weird lag with slow connections where the checkbox's state would take a second or two to update after the API call returned successfully making it feel like the update didn't take. This is now resolved.

I did have to make a small change to our `Checkbox` component - previously, it would accept a "disabled" boolean prop and if set, would disable both the checkbox, and any label. In this particular case, we want to disable only the checkbox and not the label, so I changed the `disabled` prop to `disable` and rather than it being a boolean, it's an optional prop that takes an enum of `"all"` or `"checkbox` and the `Checkbox` component will then only disable the `checkbox` if `disable="checkbox"`.

It's an interesting pattern, so open to feedback on if you have ideas that'll lead to a better developer experience.

### Approvals
- [x] Get approval from design

### Testing

- [x] Ensure no new regressions where introduced
- [x] Ensure the open/closed state is sticky as intended
- [x] Ensure that everything looks good on light and dark modes
- [x] Ensure that everything works good on fast and slow internet

### Screenshots

Light Mode (Before)
Screenshot 2025-01-14 at 3 56 57 PM

Light Mode (After)
Screenshot 2025-01-14 at 3 56 17 PM

Dark Mode 
Screenshot 2025-01-14 at 3 56 47 PM
(Before)

Dark Mode (After)
Screenshot 2025-01-14 at 3 56 29 PM


packages/front-end/components/Experiment/PreLaunchChecklist.tsx
```diff
@@ -373,11 +392,11 @@ export function PreLaunchChecklist({
           checklist: updatedManualChecklistStatus,
         }),
       });
+      await mutateExperiment();
```
mknowlton89, line 395: Moving this here resolves a weird experience on slow connections where the `Loading` state resolves, but the checklist item hasn't yet been updated (because we weren't awaiting the `mutateExperiment` call.

mknowlton89:
Closing as this is really far out of date. Will start a fresh PR

PR #3528, opened 2025-01-14
Improve saved group reference checks

Kevin-Chant:
### Features and Changes

Currently when deleting saved groups we check for any referencing features before allowing the deletion. This PR refactors the existing checks into a shared helper for consistency and adds a second helper to also check for experiment references.

### Testing

Manually tested the delete flows from the three entrypoints: condition groups view, id lists view, and [sgid] per-group page
* Delete button should be disabled if at least one reference exists
* Delete button should be enabled if no references exist
* Links should appear for all referencing features and experiments
* Links should direct to the proper pages for those features/experiments

### DIFF
--- a/packages/shared/src/util/index.ts
+++ b/packages/shared/src/util/index.ts
@@ -16,6 +16,7 @@ import { ExperimentReportVariation } from "back-end/types/report";
 import { VisualChange } from "back-end/types/visual-changeset";
 import { FeatureRevisionInterface } from "back-end/types/feature-revision";
 import { Environment } from "back-end/types/organization";
+import { SavedGroupInterface } from "../types";
 import { featureHasEnvironment } from "./features";
 
 export * from "./features";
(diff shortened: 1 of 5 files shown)


packages/shared/src/util/index.ts
```diff
@@ -336,3 +337,62 @@ export function formatByteSizeString(numBytes: number, decimalPlaces = 1) {
...
+  features: FeatureInterface[];
+  environments: Environment[];
+}): Record<string, FeatureInterface[]> {
+  const referenceMap: Record<string, FeatureInterface[]> = {};
+  features.forEach((feature) => {
+    savedGroups.forEach((savedGroup) => {
+      const matches = getMatchingRules(
+        feature,
+        (rule) =>
+          rule.condition?.includes(savedGroup.id) ||
```
bryce-fitzsimons, line 356: consider using `recursiveWalk()` instead, here and for experiment conditions

bryce-fitzsimons (approved):
lgtm, small nit about recursiveWalk() but not a blocker. Would also like to see usage on the page somewhere, but can be a separate PR

PR #3540, opened 2025-01-16
Add Unhealthy status for Experiments (SRM & Multiple Exposures)

ahdriel:
### Features and Changes

- Added a new Unhealthy status for Experiments
  - It does not modify `experiment.status`, it uses that in conjunction with other data points
  - If there's an alert on the Health tab, the Experiment status should be `Unhealthy`
    - Currently we have 2 health checks: SRM & Multiple Exposures
  - To support showing the Unhealthy status on the Experiments List page we have added a new `analysisSummary` field to `Experiment`
    - It includes the calculated SRM, total users and users exposed to multiple variations, as those are derived from the latest snapshot.

### Testing

- Run an update for an Experiment that has Multiple Exposures above the threshold set at the org-level
  - You can also hardcode the value [here](https://github.com/growthbook/growthbook/blob/av/unhealthy-status-multiple-exposures/packages/back-end/src/services/stats.ts#L523) and Update the experiment results
- Update the SRM check so the test is considered unhealthy
  - Can be hardcoded [here for testing](https://github.com/growthbook/growthbook/blob/av/unhealthy-status-multiple-exposures/packages/back-end/src/services/stats.ts#L773) if getting the data in a broken state is hard
- Observe the new Unhealthy status on the Experiments List & Experiment Details pages

### Screenshots

|  | Before | After |
|--------|--------|--------|
| Experiments List | Screenshot 2025-01-20 at 3 26 37 PM | Screenshot 2025-01-20 at 3 25 28 PM |
| Experiment Details | Screenshot 2025-01-20 at 3 26 46 PM | Screenshot 2025-01-20 at 3 25 40 PM |


lukesonnet (commented):
I really like this PR! Thanks Adriel!

I left a comment I think needs to be resolved/discussed re: when to trigger an update.

Otherwise, lgtm. One question though, is your screenshot out of date? I'd think based on the code that the badge would read `Running: Unhealthy` right now, right?

ahdriel:
> One question though, is your screenshot out of date? I'd think based on the code that the badge would read Running: Unhealthy right now, right?

Good point -- That's on me passing the wrong configuration and overriding `Running` with `Unhealthy`. Will update accordingly for the Experiment Details page badge.

When I synced with Kari last we decided on keeping the status on the Experiment Lists page simplified and say only `Unhealthy` in this case.

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -219,6 +221,25 @@ export async function updateSnapshot({
   });
   if (!experimentSnapshotModel) throw "Internal error";
 
+  const shouldUpdateExperimentAnalysisSummary =
+    experimentSnapshotModel.type === "standard" &&
+    experimentSnapshotModel.status !== "running";
```
bryce-fitzsimons, line 226: Should we have additional checks for latest phase (I think you can refresh a previous phase's snapshot)?

bryce-fitzsimons, line 226: do these statuses work for bandits as well? SRM is similar but computed slightly differently, ME should be identical, but the other power-calculator-related statuses perhaps not...

packages/front-end/components/Experiment/TabbedPage/ExperimentStatusIndicator.tsx
```diff
@@ -1,48 +1,81 @@
...
-function getBadgeProps(
+function getStatusIndicatorData(
   experimentData: ExperimentData,
-  skipArchived: boolean
+  skipArchived: boolean,
+  healthSettings: {
+    srmThreshold: number;
+    multipleExposureMinPercent: number;
+  }
 ): [
```
bryce-fitzsimons, line 73: probably better to use an object

packages/back-end/src/models/ExperimentSnapshotModel.ts
lukesonnet: Oh, great call. Yeah I think you can.

lukesonnet: I think it's computed differently but we still only return one p-value and one status, so I think the changes here should work.

ahdriel: @lukebrawleysmith recommended to skip the mid-experiment power calculation for Bandits, so `health.power` will be optional.

@lukesonnet So it sounds that this implementation works for SRM & ME for Bandit Experiments?

lukesonnet: I think it should.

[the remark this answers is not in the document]
ahdriel: Agreed! Given that we are not going to surface the query errors in the Experiment List at this moment, running it in case of error does not provide a lot of value.

So maybe we bring this back when we add a new `Error` status for the Experiment or something along those lines.

Will update.

lukesonnet (approved):
lgtm!

Are we going to link right to the health tab later for unhealthy statuses or is that in this PR? I couldn't quite tell from reading it over.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -401,7 +401,10 @@ export default function AnalysisSettingsSummary({
                     <RunQueriesButton
                       cta="Update"
                       cancelEndpoint={`/snapshot/${latest.id}/cancel`}
-                      mutate={mutateSnapshot}
+                      mutate={() => {
```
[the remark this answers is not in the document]
bryce-fitzsimons: I think it's fine as you have it

PR #3542, opened 2025-01-16
sdk: Prevent overly-eager early exit when evaluating prereqs

bryce-fitzsimons:
We have a bit of code in the SDK that checks for cycles when evaluating prereq. It is overly-eager in its bailout behavior. For example, this will return "null" because the evaluator encounters the `prereq` flag more than once:
image

The cycle check now uses a DFS stack to check for duplicate (cyclic) evaluation instead of a global Set.

TODO:

- [ ] bump versions
- [ ] publish


Kevin-Chant (approved):
LGTM

jdorn (approved):

PR #3545, opened 2025-01-16
Add plan level to premium tooltip

lukesonnet:
### Features and Changes

Passes the `accountFeatures` map via userContext to the front-end to easily look-up commercial feature <-> account plan lookup.

Use this to improve the tooltip language and be more precise. When you add `commercialFeature` to a `PremiumTooltip` it always tries to say "Pro" if that's the lowest plan needed or "Enterprise" if enterprise is needed.

Also fixes some premium language elsewhere.

### Screenshots
**Before** it just said "This is a premium feature." 
Screenshot 2025-01-16 at 12 27 08 PM

**Now** it says...

Screenshot 2025-01-16 at 12 25 07 PM
Screenshot 2025-01-16 at 12 24 52 PM

Other fix
Screenshot 2025-01-16 at 12 21 38 PM

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -750,7 +750,10 @@ export async function getOrganization(req: AuthRequest, res: Response) {
     licenseError: getLicenseError(org),
     commercialFeatures: [...accountFeatures[getEffectiveAccountPlan(org)]],
     fullCommercialFeaturesMap: Object.fromEntries(
-      Object.entries(accountFeatures).map(([plan, features]) => [plan, Array.from(features)])
+      Object.entries(accountFeatures).map(([plan, features]) => [
+        plan,
+        Array.from(features),
+      ])
     ),
     roles: getRoles(org),
     members: expandedMembers,
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -7,6 +7,7 @@ import {
   getEffectiveAccountPlan,
   getLicense,
   getLicenseError,
+  getLowestPlanPerFeature,
   licenseInit,
 } from "enterprise";
 import { experimentHasLinkedChanges } from "shared/util";
@@ -741,6 +742,8 @@ export async function getOrganization(req: AuthRequest, res: Response) {
 
   const watch = await getWatchedByUser(org.id, userId);
 
+  const commercialFeatureLowestPlan = getLowestPlanPerFeature(accountFeatures);
+
   return res.status(200).json({
     status: 200,
     apiKeys,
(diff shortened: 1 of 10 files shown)


packages/front-end/components/Marketing/PremiumTooltip.tsx
```diff
@@ -31,11 +31,21 @@ export default function PremiumTooltip({
   usePortal,
   ...otherProps
 }: Props) {
-  const { hasCommercialFeature } = useUser();
+  const { hasCommercialFeature, fullCommercialFeaturesMap } = useUser();
   const hasFeature = commercialFeature
     ? hasCommercialFeature(commercialFeature)
     : true;
+  console.log(fullCommercialFeaturesMap);
```
jdorn, line 38: Remove console log

packages/front-end/components/Archetype/ArchetypeList.tsx
```diff
@@ -52,7 +52,7 @@ export const ArchetypeList: FC<{
             <h2>Create Reusable Archetypes</h2>
             <p>
               Archetypes are named sets of attributes that help you test your
-              features. Archetypes are a premium feature.
+              features. Archetypes are a Pro feature.
```
jdorn, line 55: We should not hardcode this since it will become out of date if we change the plan for this feature later.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -749,6 +749,12 @@ export async function getOrganization(req: AuthRequest, res: Response) {
     effectiveAccountPlan: getEffectiveAccountPlan(org),
     licenseError: getLicenseError(org),
     commercialFeatures: [...accountFeatures[getEffectiveAccountPlan(org)]],
+    fullCommercialFeaturesMap: Object.fromEntries(
+      Object.entries(accountFeatures).map(([plan, features]) => [
+        plan,
+        Array.from(features),
+      ])
+    ),
```
jdorn, line 757: I might recommend doing the pre-processing on the back-end so we just send down a map from feature to the minimum plan required.  Will make it easier to manage if that logic is in one place.

packages/front-end/components/Marketing/PremiumTooltip.tsx
```diff
@@ -31,11 +31,21 @@ export default function PremiumTooltip({
...
 
+  const planLevelText =
+    !commercialFeature || !fullCommercialFeaturesMap
+      ? undefined
+      : fullCommercialFeaturesMap["pro"].includes(commercialFeature)
+      ? "a Pro"
+      : fullCommercialFeaturesMap["enterprise"].includes(commercialFeature)
+      ? "an Enterprise"
+      : "a Premium";
+  const tooltipText = premiumText ?? `This is ${planLevelText} feature`;
```
jdorn, line 48: `planLevelText` can be undefined, which produces a weird string here.

packages/back-end/src/routers/organizations/organizations.controller.ts
lukesonnet: Yeah I thought about doing it but wasn't sure I wanted to make "lowest plan per feature" a canonical mapping outside of a component since "lowest plan" is debatable.

Auz (approved):

PR #3548, opened 2025-01-17
Lots of UI tweaks. Cleaned up no access states for metric groups, templates, and bandits

Auz:
Added component for NoAccess state. 
Screenshot 2025-01-16 at 5 04 54 PM


packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -281,116 +283,120 @@ export default function TabbedPage({
...
+                </div>
+              </div>
+            )}
           <div
-            className={
-              // todo: standardize explore & results tabs across experiment types
-              isBandit && tab === "results" ? "d-block" : "d-none d-print-block"
-            }
+            className={clsx(
+              "pt-3 where",
```
jdorn, line 341: Is this intentional?  I don't see any `where` class name in our styles.

packages/front-end/components/Experiment/Templates/TemplatesPage.tsx
```diff
@@ -96,6 +97,20 @@ export const TemplatesPage = ({
     return <div className="alert alert-danger">{error.message}</div>;
   }
 
+  if (!hasTemplatesFeature) {
+    return (
+      <>
+        <NoAccessState
```
jdorn, line 103: I would rename this component.  Something like `PremiumEmptyState`.  We use "no access" to mean something different in the code base (a user role without read access).

packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -199,6 +199,7 @@ const MetricsSelector: FC<{
 
   const selector = !forceSingleMetric ? (
     <MultiSelectField
+      style={{ color: "red" }}
```
jdorn, line 202: Is this intentional?

packages/front-end/pages/settings/customfields.tsx
```diff
@@ -1,56 +1,24 @@
...
-            {upgradeModal && (
-              <UpgradeModal
-                close={() => setUpgradeModal(false)}
-                reason="Add custom metadata,"
-                source="custom-fields"
-              />
-            )}
-          </div>
-        </div>
+                  optional. Custom fields are part of our enterprise plan."
```
jdorn, line 17: Feels like we should not encode the plan level in the description.  If we change the feature level later this will be inaccurate.

jdorn (approved):

packages/front-end/components/Experiment/TabbedPage/index.tsx
Auz: ah, debugging I forgot about... good catch

packages/front-end/components/Experiment/Templates/TemplatesPage.tsx
Auz: Fixed

packages/front-end/components/Experiment/MetricsSelector.tsx
Auz: debugging, fixed

packages/front-end/pages/settings/customfields.tsx
Auz: done

bryce-fitzsimons (approved):

PR #3549, opened 2025-01-17
Style fixes for feature rules component

Kevin-Chant:
### Features and Changes

Minor fixes for two weird visual behaviors
1. On large screens, the environment tab selection was width-constrained and adding unwanted padding
2. The empty state for rules was missing a top border and blended into the tab row

### Testing

Load the features page and then zoom way out to ensure the environment selector stays full width (both tabs for <6 envs and dropdown for 6+)
Use an environment with no rules to check the top border

### Screenshots

image

image

### DIFF
--- a/packages/front-end/components/Features/FeatureRules.tsx
+++ b/packages/front-end/components/Features/FeatureRules.tsx
@@ -89,7 +89,7 @@ export default function FeatureRules({
     <>
       <Tabs value={env} onValueChange={setEnv}>
         {environments.length < 6 ? (
-          <Container>
+          <Container maxWidth="100%">
             <Flex align="center" justify="between">
               <TabsList>
                 <Flex overflow="scroll">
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Features/FeatureRules.tsx
```diff
@@ -160,7 +160,7 @@ export default function FeatureRules({
                     experimentsMap={experimentsMap}
                   />
                 ) : (
-                  <div className="p-3 bg-white border-bottom">
+                  <div className="p-3 bg-white border-bottom border-top">
```
Kevin-Chant, line 163: I know this is bootstrap but it's not making our bootstrap usage any worse (this component already has bootstrap) and I didn't see a trivial way to replicate this border behavior. Would rather fix it now and deal with removing bootstrap later

bryce-fitzsimons (approved):

PR #3555, opened 2025-01-19
JS SDK: Add plugins

jdorn:
### Features and Changes

Adds new `plugins` option for JS SDK with several built-in plugins.  These plugins are not included in the main package export, so should not increase bundle size unless explicitly imported.

Two plugins that were part of the HTML script tag, but are now available to all JS implementations:

- `third-party-tracking` - Configures the trackingCallback to send "Experiment Viewed" events to Segment (analytics.js), GA4 (gtag), and GTM (dataLayer).
- `auto-attributes` - Detects common user attributes such as browser and utm parameters, and generates a random `id` attribute to persist in a cookie.

```ts
import {
  thirdPartyTrackingPlugin,
  autoAttributesPlugin
} from "@growthbook/growthbook/plugins";

const gb = new GrowthBook({
  clientKey: "sdk-test",
  plugins: [
    thirdPartyTrackingPlugin({ trackers: ["segment", "gtm"] }),
    autoAttributesPlugin()
  ]
});
await gb.init();
```

Plus a new plugin for sending events to GrowthBook's built-in ingestor endpoint (in private beta):

```ts
import {growthbookTrackingPlugin} from "@growthbook/growthbook/plugins";

const gb = new GrowthBook({
  clientKey: "sdk-test",
  plugins: [growthbookTrackingPlugin()]
});
await gb.init();

// Experiment Viewed and Feature Usage events sent automatically
gb.isOn("my-feature");

// Can also fire arbitrary custom events
gb.logEvent("purchase", { amount: 99.99 });
```

For easier integration with tools like Segment and GTM, you can also log events directly on the `window` in browser environments:

```ts
window.gbEvents = window.gbEvents || [];
window.gbEvents.push("event with no properties");
window.gbEvents.push({
  eventName: "event with properties",
  properties: {
    prop: 123
  }
})
```

### Bundle Size (in KB)

ESM with tree shaking:  12.0 -> 12.2
Full ESM: 13.2 -> 13.6
auto.min.js (no growthbook-tracking plugin): 14.3 -> 14.9
Full auto.min.js: 14.3 -> 15.8


packages/sdk-js/src/plugins/growthbook-tracking.ts
```diff
@@ -0,0 +1,289 @@
...
+  if (!events.length) return;
+  const data: EventPayload[] = events.map(
+    ({ eventName, properties, attributes, url }) => {
+      const { nested, topLevel } = parseAttributes(attributes || {});
+
+      return {
+        event_name: eventName,
+        properties_json: properties || {},
+        ...topLevel,
+        sdk_language: "js",
```
bryce-fitzsimons, line 123: wondering if we should have a Context var for sdk type as well to differentiate "js", "react", "script tag", etc

packages/front-end/services/utils.tsx
```diff
@@ -22,14 +27,23 @@ export const gbContext: Context = {
   clientKey: GB_SDK_ID,
   enableDevMode: true,
   trackingCallback: (experiment, result) => {
```
bryce-fitzsimons, line 29: is this the time to also consider tracking the user's attributes?

packages/sdk-js/src/core.ts
```diff
@@ -58,6 +61,49 @@ function getForcedVariations(ctx: EvalContext) {
...
+
+  if (ctx.global.trackingCallback) {
+    const cb = ctx.global.trackingCallback;
+    calls.push(safeCall(() => cb(experiment, result, ctx.user)));
+  }
+  if (ctx.user.trackingCallback) {
+    const cb = ctx.user.trackingCallback;
+    calls.push(safeCall(() => cb(experiment, result)));
+  }
+  if (ctx.global.eventLogger) {
```
bryce-fitzsimons, line 87: can this potentially replace the trackingCallback? If so, do we need to do async event logging judo for redirect tests?

Also, track the user attributes, not just hash attribute?

packages/sdk-js/src/plugins/growthbook-tracking.ts
```diff
@@ -0,0 +1,289 @@
...
+        sdk_version: SDK_VERSION,
+        url: url,
+        context_json: nested,
+      };
+    }
+  );
+
+  const endpoint = `${
+    ingestorHost || "https://us1.gb-ingest.com"
```
bryce-fitzsimons, line 132: Probably out of scope, but do we need to think about DNS level killswitches here for DDOS?

packages/sdk-js/src/plugins/growthbook-tracking.ts
```diff
@@ -0,0 +1,289 @@
...
+  }/track?client_key=${clientKey}`;
+  const body = JSON.stringify(data);
+
+  try {
+    if (typeof navigator !== "undefined" && "sendBeacon" in navigator) {
```
bryce-fitzsimons, line 137: should we rig up any way to defer and persist these calls in a stack across pages?

bryce-fitzsimons (approved):
preemptively approving, but have a look at my suggestions and see if any are actionable now or soon

packages/sdk-js/src/plugins/growthbook-tracking.ts
jdorn: sendBeacon should reliably send the event, even while the page is unloading.

packages/sdk-js/src/core.ts
jdorn: Yes, this could replace trackingCallback.  It's more opinionated about the shape of the data and also includes feature usage events.

This only includes properties specific to the event.  Within the event logger, you can call `gb.getAttributes()` to get all of the attributes to log.

packages/sdk-js/src/plugins/auto-attributes.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export type AutoAttributeSettings = {
+  uuidCookieName?: string;
+  uuidKey?: string;
+  uuid?: string;
+  uuidAutoPersist?: boolean;
+};
+
+export function autoAttributesPlugin(settings: AutoAttributeSettings = {}) {
+  // Browser only
```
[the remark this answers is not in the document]
jdorn: Yeah, let's do that as a follow-up

packages/sdk-js/src/plugins/growthbook-tracking.ts
jdorn: won't work with our current architecture.  The React SDK has our JS SDK as a dependency, it doesn't build its own copy.

packages/sdk-js/src/plugins/growthbook-tracking.ts
jdorn: Since this will be tied to billing on day one, I'm less worried about it.
This ingestion domain will probably end up on adblock lists eventually.  So we will likely offer a custom CNAME option for bigger enterprises.

PR #3569, opened 2025-01-25
fix experiment creation flow bugs, consistently include metrics page on creation flow

bryce-fitzsimons:
- consistently adds the "metrics" tab when creating an experiment (new experiment, new bandit, new experiment from template, new experiment/bandit feature rule)
- fixes bug where metric/analysis settings were not being set when creating an experiment from a feature rule
- fixes bug where `skipPartialData` was not importing from template nor saving correctly (mangled types). Still messy, could use a re-write.
- fixes bug where you could not create a bandit from feature flag if mandatory templates were turned on
- random UI tweaks

image


bryce-fitzsimons:
@lukesonnet One consequence of always showing the Metrics tab for new experiments is that the user can (generally) not leave "datasource" blank; it comes pre-filled with the org default ds or the first valid one. I think this is desired, just FYI...

lukesonnet:
> it comes pre-filled with the org default ds or the first valid one. I think this is desired, just FYI...

This is fine as it makes it easy to proceed and is no worse than the status quo.

The biggest change here is that we require you to set an exposure table, whereas before we didn't and just picked one by default. I think we should force people to select this, but I also wonder if there are users out there who are starting experiments and then letting someone else set up the data stuff for whom we are making things harder.

I think it's fine as is, though, especially since we have Identifiers linked to the exposure queries in the UI.

What happens if you don't have a datasource though? In status quo, you can still start an experiment and set it up later, is there any way to do that here?

Should we:
* allow for submission if no datasource exists?
* allow for submission even without datasource and EAQ fields if you pick a secondary CTA of "Add metrics later"?

lukesonnet (approved):
cursory overview, but lgtm and thanks for testing what happens with limited data

PR #3570, opened 2025-01-26
Design Updates 

Auz:
Update designs for the features pages, data source pages, and settings. Added assignment testing to its own tab. Migrated many components to Radix.


packages/front-end/components/Features/EditFeatureInfoModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+    defaultValues: {
+      tags: feature.tags || [],
+      owner: feature.owner,
+      project: feature.project || "",
+    },
+  });
+  const permissionsUtil = usePermissionsUtil();
+
+  const permissionRequired = (project) =>
+    permissionsUtil.canUpdateFeature({ project }, {});
```
jdorn, line 38: This isn't quite right.  It should be `canUpdateFeature(feature, {project})`.  We want to make sure they have permission to edit in both the original and new project.

packages/front-end/components/Features/EditFeatureInfoModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+          ) : (
+            <Callout status="warning">
+              Changing the project may prevent this Feature Flag and any linked
+              Experiments from being sent to users.{" "}
+              <Tooltip
+                body={
+                  "SDK endpoints are linked to specific environments and (optionally) projects. Changing the project of this feature may result in this feature returning in a different payload."
+                }
+              />
+            </Callout>
```
jdorn, line 108: Should we only show this if they have made changes to the project?  Otherwise, it makes other innocuous changes like adding a tag seem dangerous because of the big yellow warning box.

packages/front-end/components/ProjectBadges.tsx
```diff
@@ -14,14 +13,14 @@ export interface Props {
     | "member"
     | "team"
     | "fact table"
+    | "feature"
```
jdorn, line 16: This component was meant for resources with an array of projects.  Feature only has a single project.  This is mostly fine, except we want the empty state to show "None" instead of "All" for features

packages/front-end/components/Radix/Tabs.tsx
```diff
@@ -100,6 +101,34 @@ export const TabsList = forwardRef<HTMLDivElement, TabsListProps>(
...
+    const TABS_HEADER_HEIGHT_PX = 55;
+    const tabsRef = useRef<HTMLDivElement>(null);
+    const [headerPinned, setHeaderPinned] = useState(false);
+    const { scrollY } = useScrollPosition();
+    useEffect(() => {
+      if (!tabsRef.current) return;
+      const isHeaderSticky =
+        tabsRef.current.getBoundingClientRect().top <= TABS_HEADER_HEIGHT_PX;
+      setHeaderPinned(isHeaderSticky);
+    }, [scrollY]);
```
jdorn, line 118: Running this on every scroll change is probably expensive.  Might be better to use an IntersectionObserver.  Even better if we can get rid of the need for a `pinnedClass` entirely and have one set of styles that works for both pinned and unpinned.

jdorn (approved):

packages/front-end/components/DemoDataSourcePage/DemoDataSourcePage.tsx
```diff
@@ -119,6 +119,7 @@ export function DeleteDemoDatasourceButton({
       displayName="Sample Data"
       title="Sample Data"
       text="Delete Sample Data"
+      useRadix={true}
       outline={false}
```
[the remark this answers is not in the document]
Auz: fixed

packages/front-end/components/Features/EditFeatureInfoModal.tsx
Auz: Fixed

packages/front-end/components/Features/EditFeatureInfoModal.tsx
Auz: Fixed

packages/front-end/components/ProjectBadges.tsx
Auz: Fixed

packages/front-end/components/Radix/Tabs.tsx
Auz: I used IntersectionObserver, there was no good way to remove the sticky class entirely given our design.

PR #3575, opened 2025-01-27
Fix experiment header margins

bryce-fitzsimons:
- accidentally removed margin for exp header
- fix top padding on health tab

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
@@ -723,7 +723,11 @@ export default function ExperimentHeader({
       )}
 
       <div
-        className={clsx("position-relative experiment-header", "px-3", "pt-3")}
+        className={clsx(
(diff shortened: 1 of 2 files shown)


lukesonnet (approved):

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -723,7 +723,11 @@ export default function ExperimentHeader({
       )}
 
       <div
-        className={clsx("position-relative experiment-header", "px-3", "pt-3")}
+        className={clsx(
```
msamper, line 726: nit: no need for `clsx` here since we're not applying classnames conditionally

msamper (approved):

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
bryce-fitzsimons: just noticed this, removed

PR #3576, opened 2025-01-27
Billing Portal

mknowlton89:
### Features and Changes

Adds billing portal support for new billing platform.

### Dependencies

This PR will depend on the work others are doing to support the new billing platform - specifically. this PR needs the `subscription.type` field added to each subscription and then to update the useStripeSubscription hook to surface this.


packages/front-end/components/Settings/OrbSubscriptionInfo.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <div>
+      {portalUrl ? (
+        <iframe
+          src={portalUrl}
+          onLoad={() => setLoading(false)}
+          style={{
+            width: "100%",
+            height: "80vh",
```
mknowlton89, line 62: Hacky, but heigh: 100% wasn't being respected. This does introduce a potential scroll in a scroll.

PR #3577, opened 2025-01-28
Add API endpoints for Archetypes

Kevin-Chant:
### Features and Changes

Adds CRUD endpoints for archetypes to the API, along with an endpoint for finding an sdkConnection by the sdk key.
These endpoints will be consumed by the new devtools but could also be used for general programmatic control of Archetypes

### Testing
Hit the various endpoints to check behavior & see the changes reflected on the Archetypes page

### Screenshots
image

### DIFF
--- a/packages/front-end/components/Archetype/ArchetypeList.tsx
+++ b/packages/front-end/components/Archetype/ArchetypeList.tsx
@@ -112,6 +112,15 @@ export const ArchetypeList: FC<{
                   archetype,
                   {}
                 );
+                let parsedAttributes = {};
+                try {
+                  parsedAttributes = JSON.parse(archetype.attributes);
+                } catch {
+                  console.error(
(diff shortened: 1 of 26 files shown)


packages/front-end/components/Archetype/ArchetypeList.tsx
```diff
@@ -121,7 +121,7 @@ export const ArchetypeList: FC<{
                           <>
                             <Code
                               code={JSON.stringify(
-                                JSON.parse(archetype.attributes),
+                                JSON.parse(archetype.attributes || "{}"),
```
Kevin-Chant, line 124: Currently you can create an empty archetype in the UI which completely prevents the page from rendering. This is just a quick bandaid and the API enforces that the string is present & valid in the validations

bryce-fitzsimons (approved):

PR #3578, opened 2025-01-28
New billing framework - Step 1

jdorn:
### Features and Changes

On GrowthBook Cloud, migrate away from Stripe Billing.  This will be done in several steps.

Step 1 (this PR):
- [x] New `SubscriptionInfo` object that is an abstraction over `StripeSubscription`
- [x] Update code throughout app to not reference Stripe directly except in a few specific places
- [x] New `billingPlatform` field for licenses that is always set to `"stripe"` today, but can be used later for migration

Future steps:
- Ability to manage payment methods on `/billing` directly instead of redirecting to Stripe Customer Portal - https://github.com/growthbook/growthbook/pull/3593
- Ability to enter payment method in-app when upgrading instead of redirecting to Stripe Checkout - https://github.com/growthbook/growthbook/pull/3666
- Add an `orb` billingPlatform type to the license model (and required meta info)
- Sync seat usage data to Orb to support our seat-based pricing model
- On `/billing`, support both Stripe and Orb customer portals - https://github.com/growthbook/growthbook/pull/3576
- New trials/subscriptions get created in Orb instead of Stripe
- Migrate existing Stripe subscriptions over to Orb


packages/front-end/services/UserContext.tsx
```diff
@@ -462,6 +430,37 @@ export function UserContextProvider({ children }: { children: ReactNode }) {
     if (data) setReady(true);
   }, [data]);
 
+  const organization: Partial<OrganizationInterface> | undefined =
+    currentOrg?.organization;
+  const subscription = currentOrg?.subscription || null;
```
mknowlton89, line 435: @jdorn Is this going to break for orgs with Stripe subscriptions that started before we introduced the license server? 

It looks like the subscription can either live at `currentOrg.subscription`, while we also have metadata that lives at `currentOrg.organization.subscription`

PR #3584, opened 2025-01-29
Move Impact on Dashboard (pseudo a/a test)

lukesonnet:
### Features and Changes

This PR introduces a feature flag to move the experiment impact just below the top-line graphs on the Dashboard. This change is a test on a low risk surface that acts as a pseudo a/a test of our Cluster Experiment recommended flow.

Rather than create a totally artificial test, I prefer to do something that has a reasonable goal, even though our power will be very limited.

FF off
Screenshot 2025-01-28 at 4 19 13 PM

FF on
Screenshot 2025-01-28 at 4 19 17 PM

### DIFF
--- a/packages/front-end/components/HomePage/Dashboard.tsx
+++ b/packages/front-end/components/HomePage/Dashboard.tsx
@@ -26,7 +26,7 @@ export default function Dashboard({ experiments }: Props) {
     nameMap.set(e.id, e.name);
   });
 
-  const experimentImpactWidget = () => (
+  const experimentImpactWidget = (
     <div className="col-xl-13 mb-4">
       <div className="list-group activity-box overflow-auto pt-1">
         {hasCommercialFeature("experiment-impact") ? (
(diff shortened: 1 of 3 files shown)


packages/front-end/components/HomePage/Dashboard.tsx
```diff
@@ -17,12 +19,37 @@ export interface Props {
...
 export default function Dashboard({ experiments }: Props) {
   const { name, hasCommercialFeature } = useUser();
+  const growthbook = useGrowthBook<AppFeatures>();
 
   const nameMap = new Map<string, string>();
   experiments.forEach((e) => {
     nameMap.set(e.id, e.name);
   });
 
+  const experimentImpactWidget = () => (
```
jdorn, line 29: nit pick, you don't need to wrap this in an arrow function.

```ts
const someHTML = (
  <div>...</div>
);

return (
  <div>
    {someHTML}
  </div>
)

```

jdorn (approved):

PR #3591, opened 2025-01-29
fix: Experiment Status Indicator says SRM is unhealthy for Bandit but it is not

ahdriel:
### Features and Changes

An issue was reported where the Experiment had a badge saying `Unhealthy (SRM)` but when opening the `Health Tab` no error was reported.

This happened because we were using `health.traffic.overall.srm` for Bandit experiments, which is not the correct source. So now we are getting it from `snapshot.banditResult.srm` instead for Bandit experiments.

This means that `experiment.analysisSummary` will have the appropriate SRM value, which then we will use to check against the defined threshold to consider healthy/unhealthy.

### Future Improvements

The logic from BanditSRMCard should be the shared with the one we are using to set AnalysisSummary, so we can ensure ExperimentStatusIndicator & Health Tab are looking at the same set of data, to prevent inconsistencies.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -25,6 +25,7 @@ import {
   MatchingRule,
   validateCondition,
 } from "shared/util";
+import { getSRMValue } from "shared/health";
 import {
   expandMetricGroups,
   ExperimentMetricInterface,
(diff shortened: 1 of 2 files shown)


bryce-fitzsimons (approved):
wondering if we should log the default: error instead of throwing. Sometimes wild stuff can happen in the wild, and the preference is to not crash the app.

PR #3593, opened 2025-01-30
Mk/cc management

mknowlton89:
### Features and Changes

This PR adds front end functionality that lets cloud organizations with Stripe subscriptions via a license key manage what payment methods they have on file with Stripe via the `/settings/billings` page.

Screenshot 2025-02-07 at 11 45 15 AM
Screenshot 2025-02-07 at 11 45 05 AM

> [!IMPORTANT]  
> It doesn't seem possible to test Google Pay or Apple Pay via test mode - not only do you have to use a real card, you have to provide a domain that you want to accept these payment methods on, and Stripe doesn't accept a localhost url, nor can we port it to a public url with a tool like ngrok, so this hasn't been pressure tested, but built to spec per the docs.

> Additionally, there may be some payment methods that organizations have like `SEPA` or `Sofort` that don't return much information about the payment method - so in those cases, we're just returning the payment type to the list. Not ideal, but Stripe just don't give us much info on them. Just an id and a type. An example of this is in the screenshot above.

### Dependencies

This PR must be merged after https://github.com/growthbook/central-license-server/pull/32. (And PR-32 needs to be merged after PR-29)

### Testing

- [x] Ensure that as a cloud organization with a stripe subscription via a license key, I can view all of payment methods
- [x] Ensure that as a cloud organization with a stripe subscription via a license key, I can add up to 3 payment methods
- [x] Ensure that as a cloud organization with a stripe subscription via a license key, I can remove a payment method from their account, so long as it's not their only payment method. (If it's their only payment method, and their subscription is cancelled, only then can they remove their payment method)
- [x] Ensure that as a cloud organization with a stripe subscription via a license key, I can mark a payment method as the default payment method
- [x] Ensure that 3DS cards work
- [x] Ensure that a user can't add a card when the cards expiration date is in the past


packages/front-end/components/Settings/PaymentInfo.tsx
```diff
@@ -0,0 +1,198 @@
...
+export default function PaymentInfo() {
+  const [cardModal, setCardModal] = useState(false);
+  const [defaultCard, setDefaultCard] = useState<string | undefined>(undefined);
+  const [error, setError] = useState<string | undefined>(undefined);
+  const [cardData, setCardData] = useState<Card[]>([]);
+  const [loadingCards, setLoadingCards] = useState(false);
+  const { subscription } = useUser();
+  const { apiCall } = useAuth();
+
+  const customerId = "cus_Rg3aee6F7wi9EH"; //MKTODO: Fix this so the customer is from the subscription
```
mknowlton89, line 26: This is here as I need to make a change to the parent branch `jd/orb-billing` in order to persist the `customerId`

packages/front-end/components/Settings/PaymentInfo.tsx
```diff
@@ -0,0 +1,198 @@
...
+    if (!defaultCard) throw new Error("Must specify card id");
+    try {
+      await apiCall("/subscription/payment-methods/set-default", {
+        method: "POST",
+        body: JSON.stringify({
+          paymentMethodId: defaultCard,
+          customerId,
+        }),
+      });
+      fetchCardData();
```
mknowlton89, line 55: These `fetchCardData()` calls are the expensive way to do it (it refetches the card data from Stripe) but enables us to not have to worry about managing the state and having a `clean` and `dirty` state.

packages/front-end/components/Settings/PaymentInfo.tsx
```diff
@@ -0,0 +1,198 @@
...
+                                    e.preventDefault();
+                                    setDefaultCard(card.id);
+                                  }}
+                                >
+                                  Set as Default Card
+                                </button>
+                                <DeleteButton
+                                  onClick={
+                                    async () => await detachCard(card.id)
+                                    // console.log("delete card")
```
mknowlton89, line 176: I'll remove this during the PR cleanup step.

packages/back-end/src/app.ts
```diff
@@ -400,6 +400,22 @@ app.post(
...
+  stripeController.listPaymentMethods
+);
+app.post(
+  "/subscription/payment-methods/detach",
+  stripeController.deletePaymentMethod
+);
+app.post(
+  "/subscription/payment-methods/set-default",
+  stripeController.postStripeCustomerDefaultCard
+);
```
jdorn, line 418: These new endpoints only work on cloud.  They should be wrapped in an `IS_CLOUD` check so they aren't even added to the app when self-hosting.  Same with the front-end code.

packages/back-end/src/controllers/stripe.ts
```diff
@@ -261,3 +270,118 @@ export async function postWebhook(req: Request, res: Response) {
...
+}
+
+export async function deletePaymentMethod(
+  req: AuthRequest<{ paymentMethodId: string }>,
+  res: Response
+) {
+  const { paymentMethodId } = req.body;
+
+  try {
+    await stripe.paymentMethods.detach(paymentMethodId);
```
jdorn, line 380: We should do an extra sanity check here to make sure the payment method is owned by the customer.  Otherwise, any user could delete any other user's credit card just by guessing the paymentMethodId.

packages/front-end/pages/_app.tsx
```diff
@@ -51,6 +53,10 @@ type ModAppProps = AppProps & {
   };
 };
 
+export const stripePromise = loadStripe(
+  process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY || ""
+);
```
jdorn, line 58: Revert all changes to the `_app.tsx` file.  The logic of instantiating stripe and adding the wrapper should only be done on the specific billing components.  We don't want any stripe-related API calls to happen unless someone visits `/billing` on cloud (or in the future opens the upgrade modal on cloud).

packages/back-end/src/controllers/stripe.ts
```diff
@@ -261,3 +270,118 @@ export async function postWebhook(req: Request, res: Response) {
...
+      customer,
+      payment_method_configuration: STRIPE_PAYMENT_METHOD_CONFIG_ID,
+      automatic_payment_methods: {
+        enabled: true,
+      },
+    });
+    return res.status(200).json({ clientSecret: setupIntent.client_secret });
+  } catch (e) {
+    throw new Error(e.message);
+  }
```
jdorn, line 291: We will want to move all of the stripe logic to the license server.  We can keep the API routes here, but they will just call out to the license server instead of stripe directly.  This will allow it to function for self-hosted users as well as cloud.

packages/back-end/src/app.ts
```diff
@@ -400,6 +400,22 @@ app.post(
   "/subscription/new-pro-trial",
   stripeController.postNewProTrialSubscription
 );
+app.post(
+  "/subscription/payment-methods/setup-intent",
+  stripeController.postSetupIntent
+);
+app.get(
+  "/subscription/payment-methods/:subscriptionId",
```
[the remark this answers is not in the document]
mknowlton89: @jdorn Would you suggest using the `organization.id` here instead, and then doing the `newLicense` db lookup based on that? 

I could be wrong, but I feel like its possible for there to be multiple records on the `newLicenses` collection for a single org, so we'll need to get the record that was added the latest, no?

packages/back-end/src/services/organizations.ts
```diff
@@ -1149,6 +1149,7 @@ export function getSubscriptionFromOrg(
       trialEnd: organization.subscription.trialEnd,
       status: getStripeSubscriptionStatus(organization.subscription.status),
       hasPaymentMethod: !!organization.subscription.hasPaymentMethod,
+      hasLicense: false,
```
mknowlton89, line 1152: This is used to determine if we have the information needed to fetch the customer's cards from Stripe and display them in the app.

packages/front-end/components/Settings/CreditCardModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      // Trigger form validation and wallet collection
+      const { error: submitError } = await elements.submit();
+      if (submitError) {
+        throw new Error(
+          submitError.message || "Unable to validate card inputs"
+        );
+      }
+
+      const { setupIntent } = await stripe.confirmSetup({
```
mknowlton89, line 36: Should we proxy this request?

packages/front-end/components/Billing/StripeProviderWrapper.tsx
```diff
@@ -0,0 +1,32 @@
+import { createContext, useContext, useState } from "react";
+
+interface StripeContextProps {
+  clientSecret: string | null;
+  setClientSecret: (secret: string | null) => void;
+}
+
+const StripeContext = createContext<StripeContextProps | undefined>(undefined);
+
+export const StripeProviderWrapper = ({
```
jdorn, line 10: It's better to move this logic into `StripeProvider` so when using this, you only have a single wrapping component.  With how it currently is, someone might forget to add one or the other and it would cause weird errors that would be hard to debug.

packages/front-end/components/Settings/CreditCardModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+        elements,
+        clientSecret,
+        confirmParams: {
+          return_url: `${process.env.NEXT_PUBLIC_API_HOST}/settings/billing`,
```
jdorn, line 40: Were you able to test this?  When redirecting back to the billing page, I think the modal would not be open anymore.  I guess if it added the card to the customer correctly, it would start showing up in the list of payment methods and they could mark it as default there instead, so not too bad.

packages/front-end/components/Settings/CreditCardModal.tsx
```diff
@@ -0,0 +1,91 @@
...
+      // Optionally, update the user's default payment method
+      if (defaultCard) {
+        await apiCall("/subscription/payment-methods/set-default", {
+          method: "POST",
+          body: JSON.stringify({
+            paymentMethodId: setupIntent.payment_method,
+          }),
+        });
+      }
+      refetch();
```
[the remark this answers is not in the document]
jdorn: No. The elements call must be made directly to Stripe to avoid having payment data touch our servers.  So there's no way to combine them into a single request.  This is fine.

packages/front-end/components/Settings/PaymentInfo.tsx
jdorn: Alternatively, we can just call `setCardData` directly and update the `isDefault` field (remove from the old default, set it on the new one).  It seems safe to assume the rest of the data is unaffected and does not need to be refreshed.

Same when detaching a card.  We do probably need to do a full refetch when a new card is added though.

packages/back-end/src/controllers/stripe.ts
```diff
@@ -261,3 +266,99 @@ export async function postWebhook(req: Request, res: Response) {
...
+) {
+  const context = getContextFromReq(req);
+
+  const { org } = context;
+
+  try {
+    const { clientSecret } = await createSetupIntent(org.id);
+    return res.status(200).json({ clientSecret });
+  } catch (e) {
+    return res.status(500).json({ status: 500, message: e.message });
```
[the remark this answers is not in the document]
mknowlton89: Currently if the user doesn't have a license we're not rendering any of these components. I'll update the child components though, in case we use them outside of this context down the road.

I believe the best route to go is to add these checks before calling as there are some legitimate cases where the `subscription.license` might be null, but the org does actually have a subscription/stripeCustomerId, so creating the stripe customer if it doesn't exist would cause trouble.

From what I've seen we have three different subscription types:
- The original style where all of the subscription data is saved on the org
- The pre-12/2022 license where the `organizationId` isn't persisted on the license ([source](https://github.com/growthbook/central-license-server/blob/d7bb6dffcad931b85df72de50443d777f599dc59/src/models/LicenseModel.ts#L7))
- The post-12/2022 (which is the current license)

If an org has the original style or the pre-12/2022, when we hit the license server and look up the license via the orgId, the lookup will return null, and thus the error is thrown. 

I updated the `getStripeSubscriptionStatus` function to return as `hasLicense` property that is true if the license has the `organizationId` property set, so we're only rendering (and thus looking up) the payment method details if this is true to avoid looking up records we know will throw an error.

I'll update the child components though, so when we use the CreditCardModal down the road we don't run into any issues.

packages/front-end/components/Settings/PaymentInfo.tsx
mknowlton89: I've got this all built out - I am wondering we instead want to disable the "delete" option if it's a default card.

As noted in the meeting, Stripe doesn't require any of the cards on file to be "default". But looking at Orb, if no default card is set, they use the first card in the array of cards as the default. 

Currently I have the logic on our end mirroring that.

But I could disable the delete button for the default card (and add a tooltip) so the user has to explicitly set the default card to avoid potential issues.

Thoughts?

As an aside - do you want to stick with the 3 card limit?

jdorn: Yes, lets do both.  Max of 3 cards and you cannot delete the one marked as default (if you have an active subscription).  Also, when you add your first card, it is always marked as the default (disable/remove the toggle).

I don't want to rely on the "first" card logic.  We should remove all of that from our code.  So some customers today may have no cards marked as "default", which is fine.  Going forward, new customers will always have a default.

packages/front-end/components/Billing/PaymentInfo.tsx
```diff
@@ -0,0 +1,264 @@
...
+                <span className="h4 pr-2 m-0 d-inline-block align-top">
+                  <GBAddCircle />
+                </span>
+                Add Payment Method
+              </Button>
+            </Tooltip>
+          </div>
+        </Flex>
+        {error ? (
+          <Callout status="warning">{error}</Callout>
```
jdorn, line 167: Nit: should we uses `status="error"` instead here?  Or is this usually not a real error and just a warning?

packages/front-end/components/Billing/PaymentInfo.tsx
```diff
@@ -0,0 +1,264 @@
...
+    string | undefined
+  >(undefined);
+  const [error, setError] = useState<string | undefined>(undefined);
+  const [paymentMethods, setPaymentMethods] = useState<PaymentMethod[]>([]);
+  const [loading, setLoading] = useState(false);
+  const { subscription, organization } = useUser();
+  const { apiCall } = useAuth();
+  // TODO: Remove once all orgs have moved license info off of the org - only limit by isCloud()
+  // The licenseKey is required to look up payment methods
+  const canShowPaymentInfo = isCloud() && !!organization.licenseKey;
```
jdorn, line 35: Do we want to add a feature flag check here?  That way we can test with our prod stripe key first before opening it up to everyone.

packages/shared/src/enterprise/license.ts
```diff
@@ -677,6 +677,62 @@ export async function postResendEmailVerificationEmailToLicenseServer(
   );
 }
 
+export async function createSetupIntent(licenseKey: string) {
```
jdorn, line 680: Now that we have `packages/back-end/src/enterprise`, it probably makes sense to move these methods there since they can only be called from the back-end anyway (the front-end doesn't have access to `CLOUD_SECRET`).

If this requires a bigger refactor, don't worry about it.  We can do it later in a standalone PR since the same is true for most of this file (only works from back-end).

jdorn (approved):

PR #3594, opened 2025-01-30
Add GB SDK integration with back-end

ahdriel:
### Features and Changes

We now can use GrowthBook feature flags in our back-end code!

Details:
- This is available only for logged-in routes, which means it applies mostly for our Cloud App.
  - This is by design to avoid creating a circular dependency for features/SDK initialization.
- Uses the same SDK Clients as the front-end (including the ID)
- Explicitly check for IS_CLOUD for initialization 

### Testing

Manual testing
- Call `req.gb?.isOn('feature-name')` and ensure it returns the appropriate value
  - You can add a Force Rule to ensure true/false based on your User Id on app.growthbook.io
- Log `req.gb` and observe how all features and attributes are properly set


packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
...
+      const build = getBuild();
+      req.gb = new GrowthBook({
+        apiHost: "https://cdn.growthbook.io",
+        clientKey: GB_SDK_ID,
+        enableDevMode: true,
+        stickyBucketService: new BrowserCookieStickyBucketService({
+          jsCookie: req.cookies,
+        }),
+        trackingCallback: () => {
+          // TODO: How to send event to Jitsu?
```
ahdriel, line 268: @jdorn Is there a way to use Jitsu here? Or do we have an alternative for back-end logging?

packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
         dateCreated: new Date(),
       });
     };
+
+    if (IS_CLOUD) {
+      const build = getBuild();
+      req.gb = new GrowthBook({
+        apiHost: "https://cdn.growthbook.io",
+        clientKey: GB_SDK_ID,
```
ahdriel, line 262: Is this okay? To re-use the same IDs for front-end & back-end packages.

packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
...
+        enableDevMode: true,
```
ahdriel, line 263: I have enabled this, but unsure if this works/does anything for node.js.

packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
...
+          cloud: IS_CLOUD,
+          freeSeats: req.organization?.freeSeats,
+          discountCode: req.organization?.discountCode,
+          organizationId: req.organization?.id,
+          accountPlan: req.organization
+            ? getAccountPlan(req.organization)
+            : undefined,
+          superAdmin: user.superAdmin,
+          orgDateCreated: req.organization?.dateCreated,
+          anonymous_id: req.cookies["gb_device_id"],
```
ahdriel, line 282: In the FE code this is from Jitsu cookie -- found this as an alternative but also it does not feel right.

Maybe we just don't include this `anonymous_id` as all requests from the back-end will be authenticated?

packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
...
+          hasLicenseKey: req.organization?.licenseKey ? true : false,
+          configFile: usingFileConfig(),
+          usingSSO: usingOpenId(),
+          buildSHA: build.sha,
+          buildDate: build.date,
+          buildVersion: build.lastVersion,
+        },
+      });
+
+      await req.gb.init({ timeout: 500 });
```
ahdriel, line 294: If the timeout is reached, what happens if we attempt to use `req.gb?.isOn`?

packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
...
+          anonymous_id: req.cookies["gb_device_id"],
+          multiOrg: IS_MULTI_ORG,
+          role: req.organization?.members.find((m) => m.id === user.id)?.role,
+          hasLicenseKey: req.organization?.licenseKey ? true : false,
+          configFile: usingFileConfig(),
+          usingSSO: usingOpenId(),
+          buildSHA: build.sha,
+          buildDate: build.date,
+          buildVersion: build.lastVersion,
+        },
```
ahdriel, line 291: Should we add a new attribute for `requestSource` to differentiate between FE and BE?

Or maybe this is solved with different client ids?

packages/back-end/types/app-features.d.ts
```diff
@@ -0,0 +1,68 @@
+/**
```
ahdriel, line 1: Should likely merge this with `front-end/types/app-features` -- no need to have 2 of the same.

packages/back-end/src/services/auth/index.ts
```diff
@@ -248,6 +254,45 @@ export async function processJWT(
         dateCreated: new Date(),
       });
     };
+
+    if (IS_CLOUD) {
```
ahdriel, line 258: Also we likely need to have another flag if we want to force this for dev in localhost. For now I can just comment this out when testing.

Kevin-Chant:
closing as we're moving forward with #6270

PR #3595, opened 2025-01-30
Add api endpoints for serving Audit Logs

Kevin-Chant:
### Features and Changes

Adds endpoints for fetching both the per-entity and per-entity-type audit logs

### Testing

GET requests to `/api/v1/history/<entity_type>/<entity_id>` and `/api/v1/history/<entity_type>`

### Screenshots

image

image


Kevin-Chant:
Closing this after discussing the right approach here. We'll want to use Events rather than Audit Logs

PR #3598, opened 2025-01-31
Minor design fixups

Auz:


ahdriel (approved):
The toggle is looking much better!

One bug that I noticed is that we are showing the Empty Features state for a user that only has 1 Archived feature (as an example), preventing them from accessing it.

I don't think this affects many people, but wanted to register it before we forget.

PR #3599, opened 2025-01-31
Fixed issue with duplicate feature and custom fields

Auz:
Also cleaned up display of metrics listing and fixed the tabs. 

Screenshot 2025-01-31 at 12 55 45 AM

### DIFF
--- a/packages/front-end/components/Features/FeatureModal/index.tsx
+++ b/packages/front-end/components/Features/FeatureModal/index.tsx
@@ -123,22 +123,14 @@ const genFormDefaultValues = ({
     permissions: permissionsUtil,
     project,
   });
-  let customFieldValues = {};
-  if (customFields && featureToDuplicate) {
-    customFieldValues = featureToDuplicate.customFields
-      ? Object.fromEntries(
-          Object.entries(featureToDuplicate.customFields).filter(([key]) =>
-            customFields.some((a) => a.id.toString() === key)
-          )
-        )
-      : Object.fromEntries(
-          customFields.map((field) => [field.id.toString(), field.defaultValue])
-        ); // Create a fallback object
-  } else if (customFields) {
-    customFieldValues = Object.fromEntries(
-      customFields.map((field) => [field.id.toString(), field.defaultValue])
-    );
-  }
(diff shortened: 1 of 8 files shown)


packages/front-end/components/Features/FeatureModal/index.tsx
```diff
@@ -113,13 +115,31 @@ const genFormDefaultValues = ({
...
+          )
+        )
+      : Object.fromEntries(
+          customFields.map((field) => [field.id.toString(), field.defaultValue])
+        ); // Create a fallback object
+  } else if (customFields) {
+    customFieldValues = Object.fromEntries(
+      customFields.map((field) => [field.id.toString(), field.defaultValue])
+    );
+  }
```
jdorn, line 141: This has some weird edge cases. For example, if the feature you are duplicating is missing some custom fields, this will not add them with default values.

You can fix that and simplify this code a lot by getting rid of the if/else and always looping through `customFields` and setting the value to `featureToDuplicate?.customFields?.[key] ?? field.defaultValue`

Also, what data type is the id that you're having to call to string on it? Seems very unusual.

jdorn (approved):

packages/front-end/components/Features/FeatureModal/index.tsx
Auz: Ya, that would be simpler. I didn't want to persist a custom field value that was set but not shown as the filed ID does not apply to your project any more. The to string is not needed

jdorn: I don't quite follow.  `customFields` is already filtered to only include ones that are applicable to this feature.  So if you loop over that, you should be good.  This is what I was thinking:

```ts
const customFieldValues = customFields ? 
  Object.fromEntries(customFields.map((field) =>(
     [ field.id, featureToDuplicate?.customFields?.[key] ?? field.defaultValue ]
  )))
  : {};
```

Auz: ya, that's cleaner. I'll make the change

PR #3605, opened 2025-02-01
Fixed responsive display of the top of features

Auz:
Screenshot 2025-01-31 at 10 27 29 PM
Screenshot 2025-01-31 at 10 27 19 PM
Screenshot 2025-01-31 at 10 27 11 PM

### DIFF
--- a/packages/front-end/components/Features/FeaturesOverview.tsx
+++ b/packages/front-end/components/Features/FeaturesOverview.tsx
@@ -19,7 +19,7 @@ import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
 import Link from "next/link";
 import { BsClock } from "react-icons/bs";
 import { PiCheckCircleFill, PiCircleDuotone, PiFileX } from "react-icons/pi";
-import { Box, Card, Flex, Grid, Heading, Switch } from "@radix-ui/themes";
+import { Box, Card, Flex, Heading, Switch } from "@radix-ui/themes";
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Features/FeaturesOverview.tsx
```diff
@@ -960,10 +962,18 @@ export default function FeaturesOverview({
...
                   </Box>
                 </Flex>
-                <Flex align="center" justify="end">
+                <Flex
+                  align={{ initial: "center", xs: "center", sm: "start" }}
+                  justify="end"
+                  flexShrink="0"
+                  direction={{ initial: "row", xs: "column", sm: "row" }}
+                  // @ts-expect-error text wrap is not valid for some reason, but it is
+                  style={{ textWrap: "nowrap" }}
```
ahdriel, line 971: `text-wrap` is a [new CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap), so maybe we need to update the TS version so it is recognized?

But if we just want to use `nowrap` maybe we should use `whiteSpace: nowrap` as it has broader support?

ahdriel (approved):

packages/front-end/components/Features/FeaturesOverview.tsx
Auz: whiteSpace nowrap works, good call.

PR #3607, opened 2025-02-02
Start new experiment rules as drafts

jdorn:
### Features and Changes

Remove the "Start Immediately" toggle when creating a new experiment rule for a feature.  Now all experiments will always start as a draft.  With the change in #3422 , draft experiments will be published at the same time the feature draft is published.

This new flow will be much more intuitive.  It will also enforce consistency since "Start Immediately" bypassed all pre-launch checklists and other safeguards.

New flow: No "Start Immediately" toggle.  Show schedule checkbox instead.

image

When adding a schedule:

image


mknowlton89 (approved):
Tested locally and everything worked as expected.

PR #3608, opened 2025-02-02
Added custom fields to new experiment ref flow

Auz:
Screenshot 2025-02-02 at 7 51 12 AM

### DIFF
--- a/packages/front-end/components/CustomFields/CustomFieldInput.tsx
+++ b/packages/front-end/components/CustomFields/CustomFieldInput.tsx
@@ -1,20 +1,21 @@
 import { FC, useEffect, useMemo, useState } from "react";
 import { CustomField, CustomFieldSection } from "back-end/types/custom-fields";
 import { UseFormReturn } from "react-hook-form";
+import { Switch } from "@radix-ui/themes";
 import { filterCustomFieldsForSectionAndProject } from "@/hooks/useCustomFields";
 import Field from "@/components/Forms/Field";
 import SelectField from "@/components/Forms/SelectField";
 import MultiSelectField from "@/components/Forms/MultiSelectField";
-import Toggle from "@/components/Forms/Toggle";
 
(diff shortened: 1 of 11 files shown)


packages/front-end/components/CustomFields/CustomFieldInput.tsx
```diff
@@ -1,20 +1,21 @@
...
 import SelectField from "@/components/Forms/SelectField";
 import MultiSelectField from "@/components/Forms/MultiSelectField";
-import Toggle from "@/components/Forms/Toggle";
 
 const CustomFieldInput: FC<{
   customFields: CustomField[];
   // eslint-disable-next-line
   form: UseFormReturn<any>;
   section: CustomFieldSection;
+  setCustomFields?: (customFields: Record<string, string>) => void;
```
jdorn, line 15: We should not have both this and `form` as props.  I would recommend removing `form` entirely and instead having this component take `value` and `setValue` props.  Will have better type safety and make this component more flexible.

Auz: done

packages/front-end/components/CustomFields/CustomFieldInput.tsx
```diff
@@ -49,15 +55,16 @@ const CustomFieldInput: FC<{
...
         setLoadedDefaults(true);
       }
     }
-  }, [availableFields, form, loadedDefaults, currentCustomFields]);
+  }, [availableFields, loadedDefaults, currentCustomFields, setCustomFields]);
 
   const updateCustomField = (name, value) => {
     currentCustomFields[name] = value;
-    form.setValue("customFields", currentCustomFields);
+    setCustomFields(currentCustomFields);
```
jdorn, line 67: Minor: Should not mutate` currentCustomFields`.  Instead, use the spread operator.

jdorn (approved):

PR #3609, opened 2025-02-02
Fixed overflow bug on long rules and added a way to duplicate a rule

Auz:
Screenshot 2025-02-02 at 8 37 54 AM

### DIFF
--- a/packages/front-end/components/Features/RuleModal/index.tsx
+++ b/packages/front-end/components/Features/RuleModal/index.tsx
@@ -116,10 +116,6 @@ export default function RuleModal({
     attributeSchema,
   });
 
-  // unset the ID if we're duplicating the rule.
-  if (duplicate) {
-    rule.id = "";
-  }
   const defaultValues = {
     ...defaultRuleValues,
     ...rule,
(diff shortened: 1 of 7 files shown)


packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -563,7 +569,7 @@ export default function RuleModal({
       const res = await apiCall<{ version: number }>(
         `/feature/${feature.id}/${version}/rule`,
         {
-          method: i === rules.length ? "POST" : "PUT",
+          method: duplicate ? "POST" : i === rules.length ? "POST" : "PUT",
```
jdorn, line 572: When duplicating, we want to set the rule's id in the body to an empty string.  The POST endpoint will only generate an id if it's empty.

Auz: Good catch

packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -113,6 +115,11 @@ export default function RuleModal({
     ruleType: defaultType,
     attributeSchema,
   });
+
+  // unset the ID if we're duplicating the rule.
+  if (duplicate) {
+    rule.id = "";
```
jdorn, line 121: This mutates the original object and should not be done in the main render loop. Better to just modify `values` when submitting instead.

Auz: done

jdorn (approved):

PR #3610, opened 2025-02-02
Smith/cuped variance update

lukebrawleysmith:
### Features and Changes

Currently the in-production CUPED estimate of variance is wrong for relative case.  While numerically it is close to the true value in many cases, when the variance of the post-exposure metric is small, in-production variance estimates are too small, resulting in undercoverage (i.e., 95% confidence intervals do not capture the true lift 95% of the time).  

This change produces accurate variance estimates, and updates the docs with a derivation from first principles.  

### Testing

tested offline using notebook in simulation study

### Screenshots
for demo experiment, lift confidence intervals on the UI were unchanged for revenue

cuped_update

cuped_main

### DIFF
--- a/docs/docs/statistics/cuped-technical.mdx
+++ b/docs/docs/statistics/cuped-technical.mdx
@@ -41,11 +41,12 @@ $$
 Under a superpopulation framework and independence of random assignment, the adjusted means $\left(\bar{Y}_{T} - \theta\bar{X}_{T}\right)$ and $\left(\bar{Y}_{C} - \theta\bar{X}_{C}\right)$ are statistically independent.  
 Therefore, the variance of the difference in adjusted means is the sum of the variances of the adjusted means.  
 We denote these variances as $V_{adj, C}$ and $V_{adj, T}$, respectively, and they are defined as
+Define the control (treatment) population covariance between post-exposure and pre-exposure revenue as $\sigma_{XY,C}$ ($\sigma_{XY,T}$).
 
 $$
 \begin{align}
-V_{adj, C} &= \frac{\sigma^{2}_{YC} + \theta^{2}\sigma^{2}_{XC} - 2\theta\text{Cov}\left(Y_{C}, X_{C} \right)}{N_{C}}\\
-V_{adj, T} &= \frac{\sigma^{2}_{YT} + \theta^{2}\sigma^{2}_{XT} - 2\theta\text{Cov}\left(Y_{T}, X_{T} \right)}{N_{T}}.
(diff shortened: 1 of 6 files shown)


docs/docs/statistics/cuped-technical.mdx
```diff
@@ -0,0 +1,260 @@
...
+1. Define the control (treatment) population post-exposure variance as $\sigma^{2}_{YC}$ ($\sigma^{2}_{YT}$).
+2. Define the control (treatment) population pre-exposure variance as $\sigma^{2}_{XC}$ ($\sigma^{2}_{XT}$).
+3. Define the control (treatment) population covariance between post-exposure and pre-exposure revenue as $\text{Cov}\left(Y_{C}, X_{C} \right)$ ($\text{Cov}\left(Y_{T}, X_{T} \right)$).
+4. Define the covariance of the sample control means $$ \boldsymbol{\Lambda}_{C} = \text{Cov}\left[\bar{Y}_{C}, \bar{X}_{C}\right] =\begin{pmatrix}
+   \sigma^{2}_{Y,C} & \sigma*{XY,C}\\
+   \sigma*{XY,C} & \sigma^{2}_{X,C}
+   \end{pmatrix}/ N_{C}$$.
+5. Define the covariance of the sample treatment means $$ \boldsymbol{\Lambda}_{T} = \text{Cov}\left[\bar{Y}_{T}, \bar{X}_{T}\right] =\begin{pmatrix}
+   \sigma^{2}_{Y,T} & \sigma*{XY,T}\\
+   \sigma*{XY,T} & \sigma^{2}_{X,T}
```
lukesonnet, line 83:
```suggestion
   \sigma_{XY,T} & \sigma^{2}_{X,T}
```
Typo here I think for this line and the lines above.

packages/stats/gbstats/frequentist/tests.py
```diff
@@ -79,15 +83,48 @@ def __init__(
...
-            self.stat_a.n,
-            self.stat_b.variance,
-            self.stat_b.unadjusted_mean,
-            self.stat_b.n,
-            self.relative,
-        )
+        if (
+            isinstance(self.stat_a, RegressionAdjustedStatistic)
+            and isinstance(self.stat_b, RegressionAdjustedStatistic)
+            and self.relative
```
lukesonnet, line 89: Doesn't this change also need to be in bayesian/tests.py?

lukebrawleysmith: good callout, and updated

docs/docs/statistics/cuped-technical.mdx
lukebrawleysmith: Good catch, I removed the "Cov" notation and used \sigma_ notation, rather than \sigma*.  Please let me know if you disagree.

lukesonnet (approved):
lgtm! thank you!

PR #3617, opened 2025-02-03
add max width image widget with cluster experiment as example

royalfig:
Adds a new component to the docs:

`<MaxWidthImage>`

## Problem
When an image in the docs exceeds its container's size, it's max width is set to 100% of the container. While this often is desired, in some cases, it makes the image too big.

To remedy this, people use an `img` tag and set inline styles. However, this prevents the image from automatic optimizations done by Docusaurus (lazy loading and width and height attributes as the most important).

## Solution

The `MaxWidthImage` component gives you the best of both worlds, allowing for the automatic image optimizations and for setting a max width (or using the default).

## How It Works
Import it into the MDX doc:

```ts
import MaxWidthImage from "@site/src/components/MaxWidthImage"
```

Then, wrap a MD image with the component:

```mdx
<MaxWidthImage>
![A great alt description](/images/cluster-experiments/cluster-diagram.png)
</MaxWidthImage>
```

### Props
All props are optional. 

| Name | Description | Type | Default |
| --- | --- | --- | --- |
| `maxWidth` | The max width of the image in pixels | `number` | 800px |
| `border` | Add a 1px border to an image | `boolean` | `false` |
| `caption` | Add a caption to an image | `string` | `undefined` |

## Example
This PR updates the Cluster Experimentation documentation to show this component in action. The screenshot below shows an example with max width and a border.

CleanShot 2025-02-03 at 17 34 34


lukesonnet (approved):
This is awesome, thanks. Adding the border prop is also great, thank you.

PR #3629, opened 2025-02-05
add feedback widget

royalfig:
Adds a feedback widget to the docs.

https://github.com/user-attachments/assets/c293c6d0-e13e-41de-9453-1f5ab51c0cf0


lukesonnet:
I would consider just removing the tooltip. Maybe the CTA should just read "Was this helpful?" 

Can we make the buttons in the modal the same width, or side by side and smaller?

In any case, this is great.

lukesonnet (approved):

PR #3634, opened 2025-02-07
Added pagination to URL query and updated history on page change

Auz:


Auz:
Still a work in progress:

- [ ] Make sure direct navigation doesn't pop back to the previous state caused by race condition with the use effect hook.
- [ ] Test with other query parameters

PR #3636, opened 2025-02-07
Add CDN Usage to billing page on Cloud

jdorn:
### Features and Changes

Show Cloud CDN usage (requests + bandwidth) on the billing page.

image

TODO:
- [x] Allow looking back up to 12 months
- [x] Show total requests and bandwidth
- [x] Show daily graph of both requests and bandwidth for the selected month
- [x] Block months before Feb 2025
- [x] `?dummy=1` querystring to generate dummy usage data for testing


packages/front-end/components/Settings/Billing/CloudUsage.tsx
```diff
@@ -0,0 +1,287 @@
...
+      label: `${month} ${year}`,
+    });
+  }
+  const maxMonthsAgo = monthOptions.length - 1;
+
+  return (
+    <Frame style={{ position: "relative" }}>
+      {!usage.length && <LoadingOverlay />}
+      <Flex gap="2" align="center" mb="4">
+        <h3 className="mr-4 mb-0">CDN Usage</h3>
```
mknowlton89, line 118: What are your thoughts on displaying the percentage of their available bandwidth they've used here to add context about these numbers.

For now we could just define the limits in code, but eventually we can pull the limits from the orgs subscription.

packages/back-end/src/controllers/stripe.ts
```diff
@@ -323,3 +324,37 @@ export async function postWebhook(req: Request, res: Response) {
...
+  start.setHours(0, 0, 0, 0);
+
+  // End of the month
+  const end = new Date();
+  end.setMonth(end.getMonth() - monthsAgo + 1);
+  end.setDate(0);
+  end.setHours(23, 59, 59, 999);
+
+  const cdnUsage = await getDailyCDNUsageForOrg(org.id, start, end);
+
```
mknowlton89, line 358: @jdorn For the short term, if we want to build this in a manner that is agnostic to the billing platform, we could calculate usage percentage here and kick off a job to fire off a sendgrid email or fire a customerio call that can kick off an email. Thoughts?

jdorn: This request is only kicked off by a user visiting the usage page.  We could use the same underlying ClickHouse data to power alerts if we want to though.

mknowlton89 (approved):

PR #3638, opened 2025-02-07
Fix power variations; update power docs

lukesonnet:
### Features and Changes

* Updates power docs
* Sets it so "view docs" on this page sends you to the power docs
* Fixes a bug with setting nVariations on the power content page

### Testing

* New variations selector works better, and setting to 0 is fine, it just causes N/A in the data below
* New link works to docs


ahdriel (approved):

ahdriel:
Merging as it is low-risk.

PR #3650, opened 2025-02-11
First pass at upgrade options

mknowlton89:
### Features and Changes

Introduces new upgrade modal options.

I've built the feature flag and experiment in our GB account and have included screenshots there.

### DIFF
--- a/packages/back-end/.env.example
+++ b/packages/back-end/.env.example
@@ -33,3 +33,9 @@ EXPRESS_TRUST_PROXY_OPTS=
 CACHE_CONTROL_MAX_AGE=
 CACHE_CONTROL_STALE_WHILE_REVALIDATE=
 CACHE_CONTROL_STALE_IF_ERROR=
+
+# Logging
+# key/value pairs that should be included in all log events: https://getpino.io/#/docs/api?id=base-object
+# must be valid JSON e.g. {"hostname":"MY_HOST"}, or the string `null` to remove default log properties
+# if not present or JSON is invalid, Pino's defaults will be used
+LOG_BASE=
--- a/packages/back-end/src/util/logger.ts
+++ b/packages/back-end/src/util/logger.ts
@@ -4,7 +4,7 @@ import { Request } from "express";
 import { BaseLogger, Level } from "pino";
 import { ApiRequestLocals } from "back-end/types/api";
 import { AuthRequest } from "back-end/src/types/AuthRequest";
-import { ENVIRONMENT, IS_CLOUD, LOG_LEVEL } from "./secrets";
+import { ENVIRONMENT, IS_CLOUD, LOG_BASE, LOG_LEVEL } from "./secrets";
 
 const redactPaths = [
   "req.headers.authorization",
@@ -69,6 +69,14 @@ const isValidLevel = (input: unknown): input is Level => {
   ] as const).includes(input as Level);
 };
 
+// Only pass `base` if defined or null
+const logBase =
+  typeof LOG_BASE === "undefined"
+    ? {}
+    : {
+        base: LOG_BASE,
+      };
+
 export const httpLogger = pinoHttp({
   autoLogging: ENVIRONMENT === "production",
   level: isValidLevel(LOG_LEVEL) ? LOG_LEVEL : "info",
@@ -77,6 +85,7 @@ export const httpLogger = pinoHttp({
     remove: true,
   },
   customProps: getCustomLogProps,
+  ...logBase,
 });
 
 /**
--- a/packages/back-end/src/util/secrets.ts
+++ b/packages/back-end/src/util/secrets.ts
@@ -14,6 +14,24 @@ if (fs.existsSync(".env.local")) {
 
 export const LOG_LEVEL = process.env.LOG_LEVEL;
 
+let parsedLogBase:
+  | {
+      // eslint-disable-next-line
+      [key: string]: any;
+    }
+  | null
+  | undefined = undefined;
+try {
+  if (process.env.LOG_BASE === "null") {
+    parsedLogBase = null;
+  } else if (process.env.LOG_BASE) {
+    parsedLogBase = JSON.parse(process.env.LOG_BASE);
+  }
+} catch {
+  // Empty catch - don't pass a LOG_BASE
+}
+export const LOG_BASE = parsedLogBase;
+
 export const IS_CLOUD = stringToBoolean(process.env.IS_CLOUD);
 export const IS_MULTI_ORG = stringToBoolean(process.env.IS_MULTI_ORG);
 


packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -45,7 +55,22 @@ export default function UpgradeModal({ close, source }: Props) {
...
+    commercialFeatureLowestPlan,
+    seatsInUse,
+  } = useUser();
+  // These are some Upgrade CTAs throughout the app related to enterprise-only features
+  // we don't want to show a user the test treatments if that's the case
+  // since this test doesn't highlight enterprise features at all.
+  const featureFlagValue =
+    commercialFeatureLowestPlan?.[source] !== "enterprise"
+      ? growthbook.getFeatureValue("pro-upgrade-modal", "OFF")
+      : "OFF";
```
mknowlton89, line 73: This logic is a bit interesting so I wanted to call it out.

There are some Upgrade CTAs throughout the app that are referencing Enterprise features. If a user clicks that, this makes it so they see the control treatment (even if they've already be bucketed into the experiment). This is because the two treatments don't have any way of allowing the user to start an enterprise trial. Feels weird to say "upgrade to enterprise" but then only give the "pro" option.

jdorn: makes sense, good catch.  If this version wins, we'll have to make an enterprise-friendly version of it.

`source` is just a string though and is not always a commercial feature.  We should add a new required prop for this modal typed as `CommercialFeature | null` and add it everywhere.  Will be useful for other things as well.

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -253,6 +278,118 @@ export default function UpgradeModal({ close, source }: Props) {
     }
   };
 
+  const upgradeBanner = (
+    <>
+      <h3 className="mb-0">Upgrade to Pro</h3>
+      <p className="text-secondary mb-0">
+        Get instance access to advanced experimentation, permissioning and
```
jdorn, line 285: Typo: `instance` -> `instant`

packages/front-end/components/Settings/UpgradeModal/index.tsx
jdorn: We should also only evaluate the feature flag if `isCloud()` and `freeTrialAvailable` are both true.  Otherwise, users may be shown a "start trial" option when they've already used their free trial.  And self-hosted trials require more form inputs to support.

mknowlton89: @jdorn

Re: the new prop - sounds good - I'll get that added now.

Re: The targeting - I do have that logic built into the experiment rule (default value for the feature flag is "OFF" and then I've got an experiment rule where if the org is cloud, they don't have a licenseKey, and they're in the starter plan, they'll get added to the experiment - is your preference to add that here instead?

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -253,6 +287,115 @@ export default function UpgradeModal({ close, source }: Props) {
     }
   };
 
+  const upgradeHeader = (
+    <>
+      <h3 className="mb-1">Upgrade to Pro</h3>
+      <p
+        className="mb-0"
+        style={{ color: "var(--color-text-mid", fontSize: "16px" }}
```
mknowlton89, line 295: @ahdriel We don't have any pre-built components that wrap Radix's `Text` components that abstract any of this custom styling, correct?

It feels like we may want to wrap the `Text` component and then define custom styles via the `as` property for things like labels, subheads, spans, etc, and probably the same for Radix's `Headings`.

Just wanted to double check this didn't already exist and that there wasn't an existing pattern I wasn't aware of to handle this.

ahdriel: Yeah :/ I think maybe for the color and the font-size we need to have a global default via CSS using Radix variables? but it is a bigger sweeping change.

And yeah, we likely should create a `component/Radix/Text` that receives special props to define between low/mid/high text color (this is a concept we are adopting that does not exist in Radix).

mknowlton89: Thanks for following up and confirming - wanted to double check I wasn't missing any pre-existing solution for this. Thanks!

packages/front-end/components/Settings/UpgradeModal/index.tsx
jdorn: Any conditions that are strictly required for the code to work should live in the code directly.

mknowlton89: Sounds good - updated [here](https://github.com/growthbook/growthbook/pull/3650/commits/64fac5732346d8cb0875010e9e7419e080c74cb8).

packages/front-end/components/Auth/InAppHelp.tsx
```diff
@@ -42,6 +42,7 @@ export default function InAppHelp() {
           close={() => setUpgradeModal(false)}
           reason="To get access to live chat support,"
           source="in-app-help"
+          commercialFeature={null}
```
jdorn, line 45: Looks like we do have a commercial feature called `livechat`

packages/front-end/components/Settings/UpgradeModal/index.tsx
jdorn: This looks like a typo  - missing closing parentheses

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -257,6 +285,115 @@ export default function UpgradeModal({ close, source }: Props) {
...
+              },
+              {
+                value: "upgrade",
+                label: "Pro",
+                description: "Upgrade now. No credit card required.",
+              },
+            ]}
+          />
+        </div>
+        <p style={{ color: "var(--color-text-mid" }}>
```
jdorn, line 324: Same here, missing closing parentheses

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -257,6 +285,115 @@ export default function UpgradeModal({ close, source }: Props) {
...
+            rel="noopener noreferrer"
+          >
+            <strong className="a link-purple text-decoration-none">
+              Learn more <FaExternalLinkAlt className="pl-1" />
+            </strong>
+          </a>
+        </p>
+      </div>
+    );
+  }
```
jdorn, line 342: For both of these treatments, we might want to add a callout for Enterprise that links to https://www.growthbook.io/demo.  Want to keep the main focus on Pro, but still provide an escape hatch for those who are looking for enterprise.

> Interested in an Enterprise plan?  [Talk to Sales]

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -257,6 +285,115 @@ export default function UpgradeModal({ close, source }: Props) {
...
+            <FaCheckCircle className="mr-2" color="blue" />
+            <strong>Encrypted SDK endpoint response</strong>
+          </Flex>
+          <Flex align="center" className="pb-2">
+            <FaCheckCircle className="mr-2" color="blue" />
+            <strong>
+              Advanced experimentation: CUPED, Sequential testing, Bandits and
+              more
+            </strong>
+          </Flex>
```
jdorn, line 363: I'm going to play with these bullet items and see if I can come up with better ones to highlight for Pro.  These (especially the first 2) are not very compelling reasons to upgrade.

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -257,6 +285,154 @@ export default function UpgradeModal({ close, source }: Props) {
...
+      "Quickly see where feature flags are leveraged in your codebase",
+    prerequisites: "Set prerequisites to control feature rollouts",
+    redirects: "Run URL Redirect tests",
+    "multiple-sdk-webhooks": "Hook into Growthbook's SDK Webhooks",
+    "quantile-metrics": "Analyze data using quantile-based metrics",
+    "retention-metrics": "Boost retention with precise tracking",
+    "metric-populations": "Segment users using metric insights",
+    "multi-armed-bandits": "Run adaptive experiments with Bandits",
+    "historical-power": "Leverage historical data for power analysis",
+    "mid-experiment-power": "Assess mid-experiment power for quick adjustments",
```
mknowlton89, line 314: @lukesonnet @ahdriel Morning! I saw on a recent PR we switched the `CommercialFeature` from `mid-experiment-power` to`decision-framework`. I'm working on bullet points for the upgrade modal. 

Do you have any suggestions on the bullet point text for the new `decision-framework` CommercialFeature - I know there are going to be stages to what's included in the DMF so I don't want to oversell it in its current state.

Thanks!

P.S. @lukesonnet I know you're close to our customers, so if any of these other bullet points could be improved, I'm all ears!

jdorn (approved):
Left a minor nit comment.  Feel free to ignore.

PR #3651, opened 2025-02-11
[REST API] Change datasource validation for update experiment

lukesonnet:
### Features and Changes

* Adds `datasourceId` as an input when updating the experiment, and validates that it exists. Only allowed to make the change if no datasource already exists.
* Allows you to make non-datasource updates without validating the datasource exists, which was preventing making changes when the experiment model may have some invalid datasource.

### Testing

- [X] Update datasource works, only when no datasource is set
- [X] Update bad datasource fails expectedly
- [X] Update assignment query works
- [x] Update non-datasource on exp w invalid datasource works


packages/back-end/src/api/experiments/updateExperiment.ts
```diff
@@ -30,24 +30,39 @@ export const updateExperiment = createApiRequestHandler(
...
     }
+
     // check for associated assignment query id
     if (
-      req.body.assignmentQueryId != null &&
-      req.body.assignmentQueryId !== experiment.exposureQueryId &&
-      !datasource.settings.queries?.exposure?.some(
-        (q) => q.id === req.body.assignmentQueryId
-      )
+      req.body.assignmentQueryId !== undefined &&
```
jdorn, line 49: `!= null` matches both null and undefined.  With this change, if `assignmentQueryId` is set to null, the if block will run and then immediately throw an error.  Is that intentional?

lukesonnet: I think `null` being passed in should be rejected.

ahdriel, line 49: `!= null` handles both the scenario for `null` & `undefined`.

This means that now `null` will go inside this `if`.

Just checking if it is intentional

ahdriel (approved):

jdorn (approved):

PR #3653, opened 2025-02-12
Introduce Safe Rollouts 

msamper:
### Features and Changes

Introduces Safe Rollouts as a new feature rule. Allows you do a rollout while monitoring selected guardrail metrics and provides recommendations based on guardrail metric status, indicated duration, and safe rollout health. 

### Screenshots

Screenshot 2025-04-24 at 9 42 02 AM
Screenshot 2025-04-24 at 9 40 11 AM


packages/back-end/src/validators/features.ts
```diff
@@ -168,11 +168,35 @@ const experimentRefRule = baseRule
...
+] as const;
+export type SafeRolloutStatus = typeof safeRolloutStatus[number];
+
+export const safeRolloutRule = baseRule
+  .extend({
+    type: z.literal("safe-rollout"),
+    trackingKey: z.string(),
+    datasource: z.string(),
+    exposureQueryId: z.string(),
+    values: z.array(experimentValue),
```
msamper, line 184: Enforce exactly 2 variations for safe rollouts

packages/back-end/src/util/features.ts
```diff
@@ -559,6 +559,41 @@ export function getFeatureDefinition({
...
+          if (r.status === "released") {
+            // TODO: Check if this is fine for grabbing the released value or if we should add a new field like releasedVariationId
+            const variationValue = r.values[1].value;
+            if (!variationValue) return null;
+
+            // If a variation has been rolled out to 100%
+            rule.force = getJSONValue(feature.valueType, variationValue);
+          } else if (r.status === "rolled-back") {
+            // Don't return the rule if it's been rolled back
+            return null;
```
msamper, line 582: force the control value instead of returning null. default value could be different

packages/back-end/src/util/features.ts
```diff
@@ -559,6 +559,41 @@ export function getFeatureDefinition({
...
+          } else {
+            rule.variations = r.values.map((v) => {
+              return getJSONValue(feature.valueType, v.value);
+            });
+            const varWeights = r.coverage / 2; // Double check that this logic is correct
```
msamper, line 587: weights should always be 50/50

msamper: coverage is what will change

packages/back-end/src/util/features.ts
```diff
@@ -559,6 +559,41 @@ export function getFeatureDefinition({
...
           }
+        } else if (r.type === "safe-rollout") {
+          rule.coverage = r.coverage;
+
+          if (r.hashAttribute) {
+            rule.hashAttribute = r.hashAttribute;
+          }
+          if (r.seed) {
+            rule.seed = r.seed;
+          }
```
msamper, line 570: always have a seed

packages/back-end/src/util/features.ts
```diff
@@ -559,6 +559,41 @@ export function getFeatureDefinition({
...
+          } else if (r.status === "rolled-back") {
+            // Don't return the rule if it's been rolled back
+            return null;
+          } else {
+            rule.variations = r.values.map((v) => {
+              return getJSONValue(feature.valueType, v.value);
+            });
+            const varWeights = r.coverage / 2; // Double check that this logic is correct
+            rule.weights = [varWeights, varWeights];
+            rule.key = r.trackingKey;
```
msamper, line 589: make the tracking key a UUID

packages/back-end/src/util/features.ts
```diff
@@ -558,6 +558,42 @@ export function getFeatureDefinition({
...
+            // Return control value if rolled back. Feature default value might not be the same as the control value.
+            rule.force = getJSONValue(feature.valueType, controlValue);
+          } else {
+            rule.variations = [
+              getJSONValue(feature.valueType, r.controlValue),
+              getJSONValue(feature.valueType, r.variationValue),
+            ];
+            const varWeights = 0.5;
+            rule.weights = [varWeights, varWeights];
+            rule.key = r.trackingKey; // UUID
```
msamper, line 589: Check out Guy's UUID tracking key generation

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -0,0 +1,135 @@
...
+        type: metric.windowSettings.type ?? DEFAULT_METRIC_WINDOW,
+        windowUnit: metric.windowSettings.windowUnit ?? "hours",
+        windowValue:
+          metric.windowSettings.windowValue ?? DEFAULT_METRIC_WINDOW_HOURS,
+      },
+      properPrior: false,
+      properPriorMean: 0,
+      properPriorStdDev: DEFAULT_PROPER_PRIOR_STDDEV,
+      regressionAdjustmentDays: DEFAULT_REGRESSION_ADJUSTMENT_DAYS,
+      regressionAdjustmentEnabled: false,
```
lukesonnet, line 61: This should be enabled, or it should be whatever their org default is. I'd lean towards on if possible for sure. Probably would just mimic the regular flow for this, except we won't have a killswitch on safe rollout page.

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -0,0 +1,135 @@
...
+  };
+}
+
+export function getAnalysisSettingsFromSafeRolloutArgs(
+  args: SafeRolloutSnapshotAnalysisSettings
+): ExperimentSnapshotAnalysisSettings {
+  return {
+    dimensions: args.dimensions,
+    statsEngine: "frequentist",
+    regressionAdjusted: false,
```
lukesonnet, line 76: Same here, this should be be evaluated similarly to regular experiments.

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -0,0 +1,135 @@
...
+    pValueCorrection: null,
```
lukesonnet, line 77: This should also be evaluated similarly to regular experiments.

packages/back-end/src/services/safeRolloutSnapshots.ts
lukesonnet, line 76: Same here, this should be be evaluated similarly to regular experiments.

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -0,0 +1,135 @@
...
+    startDate: settings.startDate,
+    endDate: settings.endDate || new Date(),
+    experimentId: settings.experimentId,
+    exposureQueryId: settings.exposureQueryId,
+    manual: false,
+    segment: "",
+    queryFilter: settings.queryFilter || "",
+    skipPartialData: false,
+    defaultMetricPriorSettings: defaultMetricPriorSettings,
+    regressionAdjustmentEnabled: false,
```
lukesonnet, line 119: again same here.

packages/back-end/src/validators/features.ts
```diff
@@ -168,11 +168,39 @@ const experimentRefRule = baseRule
...
+    controlValue: z.string(),
+    variationValue: z.string(),
+    coverage: z.number(),
+    hashAttribute: z.string(),
+    seed: z.string(),
+    guardrailMetrics: z.array(z.string()),
+    status: z.enum(safeRolloutStatus),
+    maxDurationDays: z.number(),
+    startedAt: z.date(),
+  })
```
msamper, line 193: Add `lastSnapshotAttempt` and `nextSnapshotAttempt` which will be used by the agenda job that updates results

packages/back-end/src/routers/experiment-template/template.controller.ts
```diff
@@ -5,12 +5,61 @@ import { getContextFromReq } from "back-end/src/services/organizations";
...
+  await queryRunner.startAnalysis({
+    metricMap,
+    factTableMap,
+  });
+
+  return res.status(200).json({
+    status: 200,
+    snapshot,
+  });
+};
```
msamper, line 61: IGNORE: Just using this for testing that the query runner is working as I make small changes

packages/back-end/src/types/Audit.ts
```diff
@@ -49,6 +49,7 @@ export const entityEvents = {
   factTable: ["autocreate"],
   customField: ["create", "update", "delete"],
   experimentTemplate: ["create", "update", "delete"],
+  safeRolloutSnapshot: ["create", "update", "delete"],
```
msamper, line 52: Delete

packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -0,0 +1,228 @@
...
+    .array(
+      z.object({
+        x: z.number(),
+        y: z.number(),
+      })
+    )
+    .optional(),
+  chanceToWin: z.number().optional(),
+  errorMessage: z.string().optional().nullable(),
+  power: metricPowerResponseFromStatsEngineObject.optional().nullable(),
```
msamper, line 62: These two are causing issues since they can be null, but the corresponding type used by Experiment Snapshots doesn't allow them to be nullable. Need to figure out how to resolve that appropriately

packages/back-end/types/experiment-snapshot.d.ts
```diff
@@ -45,8 +45,8 @@ export interface SnapshotMetric {
     y: number;
   }[];
   chanceToWin?: number;
-  errorMessage?: string;
-  power?: MetricPowerResponseFromStatsEngine;
+  errorMessage?: string | null;
+  power?: MetricPowerResponseFromStatsEngine | null;
```
msamper, line 49: Added null as a possible type for now, but this may not be what we want to do since it'll probably have cascading effects wherever we try to access these

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -0,0 +1,479 @@
...
+} from "back-end/src/models/FactTableModel";
+import { getDataSourceById } from "back-end/src/models/DataSourceModel";
+import { CreateProps } from "back-end/src/models/BaseModel";
+import { determineNextDate, isJoinableMetric } from "./experiments";
+import { getSourceIntegrationObject } from "./datasource";
+
+export function getMetricForSnapshot(
+  id: string | null | undefined,
+  metricMap: Map<string, ExperimentMetricInterface>,
+  settingsForSnapshotMetrics: MetricSnapshotSettings[]
```
msamper, line 58: I think this is almost identical to the `getMetricForSnapshot` we use for experiment snapshots. If I'm not mistaken, the main difference as that we don't allow proper priors for safe rollouts so I set the proper prior field to false. Maybe worth just having an optional `type` value we pass into this function or `properPriorAllowed` so we can reuse the original function

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -0,0 +1,479 @@
...
+    variations: [
+      { id: "0", weight: 0.5 },
+      { id: "1", weight: 0.5 },
+    ],
+    coverage: experiment.coverage,
+  };
+}
+
+export async function createSnapshot({
+  experiment,
```
msamper, line 321: Also pass in feature to make sure the safe rollout rule is a part of the latest version and to include the `featureId` in the new snapshot

packages/back-end/src/queryRunners/SafeRolloutResultsQueryRunner.ts
```diff
@@ -0,0 +1,208 @@
...
+        // NB: This does not run a SQL query, but it is a health check that depends on the trafficHealth
+        result.health.power = analyzeExperimentPower({
+          trafficHealth,
+          targetDaysRemaining,
+          analysis: relativeAnalysis,
+          goalMetrics: this.model.settings.guardrailMetrics,
+          variationsSettings: this.model.settings.variations,
+        });
+      }
+    }
```
msamper, line 154: @lukesonnet Do we need this piece of code since it seems like it's dependent on having a relative analysis? My understanding is we only want to do `absolute` so we'd never have `isEligibleForMidExperimentPowerAnalysis` be true. I think you might have said we're not doing mid experiment power analyses for safe rollouts but just wanted to double check

lukesonnet: All of this can be skipped.

We're not doing this and I would expect `result.health.power` to be undefined for safe rollouts.

packages/back-end/src/routers/safe-rollout-snapshot/safe-rollout-snapshot.router.ts
```diff
@@ -0,0 +1,54 @@
...
+);
+
+// Get a snapshot for a safe rollout rule by dimension
+router.get(
+  "/:id/snapshot/:dimension",
+  validateRequestMiddleware({
+    params: snapshotDimensionParams,
+  }),
+  safeRolloutSnapshotController.getSnapshotWithDimension
+);
```
msamper, line 32: wouldn't be used right now since v0 will not have dimensions available

packages/front-end/components/Experiment/ResultsTableTooltip/ResultsTableTooltip.tsx
```diff
@@ -239,14 +239,24 @@ export default function ResultsTableTooltip({
         </div>
         <div className="text-muted font-weight-normal">
           (unadj.:&nbsp; [
-          {deltaFormatter(data.stats.ci?.[0] ?? 0, deltaFormatterOptions)},{" "}
-          {deltaFormatter(data.stats.ci?.[1] ?? 0, deltaFormatterOptions)}] )
+          {deltaFormatter(
+            data.stats.ci?.[0] ?? -Infinity,
```
msamper, line 243: Double check that this is okay to do. We may need an explicit check to see if the value is null instead of falsey

packages/back-end/src/controllers/features.ts
```diff
@@ -1172,20 +1176,110 @@ export async function postFeatureRule(
...
+      }
+    }
+
+    const safeRolloutCreateProps: CreateProps<SafeRolloutInterface> = {
+      ...interfaceFields,
+      coverage: 1, // hardcode to 100% for now
+      seed: interfaceFields.seed || uuidv4(),
+      trackingKey: interfaceFields.trackingKey || `srk_${uuidv4()}`,
+      status: "draft",
+      // TODO are these mandatory
```
msamper, line 1244: They should be mandatory

packages/back-end/src/models/SafeRolloutModel.ts
```diff
@@ -0,0 +1,84 @@
...
+  protected canUpdate() {
+    return true;
+  }
+  protected canDelete() {
+    return true;
+  }
+
+  public async findByIds(ids: string[]) {
+    return await this._find({ id: { $in: ids } });
+  }
```
msamper, line 79: we don't need this. we have `getByIds` built into the BaseModel

packages/back-end/src/models/SafeRolloutSnapshotModel.ts
```diff
@@ -0,0 +1,108 @@
...
+        limit: 1,
+      }
+    );
+
+    if (all[0]) {
+      return all[0];
+    }
+  }
+
+  public async updateById(
```
msamper, line 72: I think this should probably use the `afterUpdate` rather than overriding the `updateById` provided by the base model

packages/back-end/src/queryRunners/SafeRolloutResultsQueryRunner.ts
```diff
@@ -0,0 +1,187 @@
...
+        rows: rows,
+        error: healthQuery.error,
+        variations: this.model.settings.variations,
+      });
+
+      result.health = {
+        traffic: trafficHealth,
+      };
+    }
+    //TODO:updateFeatureRulesForSnapShotCreated();
```
msamper, line 134: Is this still needed/what is this referring to?

packages/front-end/components/Experiment/ResultsTableTooltip/ResultsTableTooltip.tsx
msamper: given that we're converting null to infinity at the stats level, this might not be necessary anymore

packages/back-end/src/queryRunners/SafeRolloutResultsQueryRunner.ts
ahdriel: I believe this is for the planned future functionality so we can dynamically update coverage instead of being hardcoded to 50/50.

We can remove this comment though IMO or at least expand with a proper `TODO: Add functionality to dynamically update coverage here`

packages/back-end/src/controllers/features.ts
```diff
@@ -1172,20 +1177,110 @@ export async function postFeatureRule(
...
-    revision,
-    environment,
-    rule,
-    res.locals.eventAudit,
-    resetReview
-  );
+  if (
+    (rule.type === "safe-rollout" && rule.safeRolloutId) || // make sure the safe rollout is created
+    rule.type !== "safe-rollout"
+  ) {
```
jdorn, line 1265: Instead of wrapping this in an `if` block, just throw above if it fails to create.

packages/back-end/src/controllers/features.ts
```diff
@@ -1656,26 +1749,61 @@ export async function putFeatureRule(
...
+          "dateCreated",
+          "dateUpdated",
+          "startedAt",
+        ]),
+      });
+    }
+    if (existingSafeRollout.startedAt) {
+      // not sure if we want to lock or just create a new revision?
+      // revsion is locked for the safe rollout once it's released
+      canUpdateRevision = false;
```
jdorn, line 1777: I don't think we should block this, but we should have checks to make sure only allowed fields are changed.

Some fields like `trackingKey` we never want to allow updates to. Others like `hashAttribute` are fine before it's started, but should be blocked after.

packages/back-end/src/controllers/features.ts
```diff
@@ -1656,26 +1749,61 @@ export async function putFeatureRule(
...
+    }
+  }
+  let revsisonVersion = parseInt(version);
```
jdorn, line 1780: Fix typo

packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const { ruleId, organization, featureId } = safeRollout;
+    if (!ruleId || !organization || !featureId) return;
+
+    const context = await getContextForAgendaJobByOrgId(organization);
+    const feature = await getFeature(context, featureId);
+    if (!feature) return;
+
+    const safeRolloutRule = getSafeRolloutRuleFromFeature(feature, ruleId);
+    if (!safeRolloutRule) return;
```
jdorn, line 72: We need a whole bunch of checks in here to skip snapshots if:
- The rule's environment toggle is off
- The feature is archived
- The rule itself is disabled

We also probably want some mechanism to clean up these orphaned rollout interfaces.  Otherwise, the list of rollouts to queue will continue growing indefinitely and this job will become very expensive and slow.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -967,6 +971,34 @@ export async function publishRevision(
...
+  safeRollouts.forEach((safeRollout) => {
+    // sync the status of the safe rollout to the status of the revision
+    const safeRolloutUpdates: Partial<SafeRolloutInterface> = {
+      status: safeRolloutStatusesMap[safeRollout.id].status,
+    };
+    if (!safeRollout.startedAt && safeRolloutUpdates.status === "running") {
+      safeRolloutUpdates["startedAt"] = new Date();
+    }
+
+    context.models.safeRollout.update(safeRollout, safeRolloutUpdates);
```
jdorn, line 999: Only do this update if something changed. Maybe this is already handled inside BaseModel?

packages/back-end/src/routers/safe-rollout/safe-rollout.helper.ts
```diff
@@ -0,0 +1,19 @@
...
+  SafeRolloutRule,
+} from "back-end/src/validators/features";
+
+export function getSafeRolloutRuleFromFeature(
+  feature: FeatureInterface,
+  ruleId: string
+): SafeRolloutRule | null {
+  Object.keys(feature.environmentSettings).forEach((env: string) =>
+    feature.environmentSettings[env].rules.forEach((rule: FeatureRule) => {
+      if (rule.id === ruleId) {
```
jdorn, line 13: If we're strictly typing this function to return a `SafeRolloutRule`, we should verify the type here instead of just relying on the id matching.

packages/back-end/src/models/FeatureModel.ts
msamper: BaseModel seems to handle this

packages/front-end/components/SafeRollout/SafeRolloutResults.tsx
```diff
@@ -0,0 +1,295 @@
...
+  //         sortOrder: 1,
+  //       };
+  //     case "draft":
+  //       return {
+  //         color: "gray",
+  //         status: "Draft",
+  //         sortOrder: 1,
+  //       };
+  //   }
+  // })();
```
msamper, line 135: We can delete this, right? I don't think we're using `ExperimentStatusIndicator` for safe rollouts at all

packages/back-end/src/api/openapi/schemas/FeatureSafeRolloutRule.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - condition
+  - id
+  - enabled
+  - type
+  - controlValue
+  - variationValue
+properties:
+  condition:
+    type: string
+  savedGroupTargeting:
```
msamper, line 12: Do we need to add prerequisites and scheduleRules here too?

jdorn: Yes for prerequisites.  I would suggest not supporting scheduling at all for the MVP since it's going to conflict with the scheduling we're adding soon as a fast follow.  Otherwise, the migration and backwards compatibility is going to be annoying to support.

packages/back-end/types/feature-rule.d.ts
```diff
@@ -0,0 +1,15 @@
+import { FeatureRule } from "back-end/src/validators/features";
+import { CreateSafeRolloutInterface } from "back-end/src/validators/safe-rollout";
+
+export type PostFeatureRuleBody = {
+  rule: FeatureRule;
+  environment: string;
+  interfaceFields?: CreateSafeRolloutInterface;
```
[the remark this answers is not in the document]
ahdriel: Updated to safeRolloutFields.

packages/front-end/components/SafeRollout/DecisionBanner.tsx
```diff
@@ -0,0 +1,180 @@
...
+  // If the safe rollout has been rolled back or released, explain that the safe rollout is
+  // acting as a temporary rollout with the control or variation value
+  if (safeRollout.status === "rolled-back" && rule.enabled !== false) {
+    return (
+      <Callout status="info">
+        This Safe Rollout is rolled back and a{" "}
+        <strong>Temporary Rollout</strong> is enabled. All users in the Safe
+        Rollout will receive the control value. If no longer needed, you can
+        disable or delete the safe rollout rule.
+      </Callout>
```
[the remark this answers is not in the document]
mknowlton89: Wrapping the children in a `<span>` tag resolved the issue. Likely need the same treatment below as well.

packages/front-end/components/SafeRollout/AnalysisSettingsSummary.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  const items: {
+    value: string | number | ReactElement;
+    tooltip?: string | ReactElement;
+    icon?: ReactElement;
+    noTransform?: boolean;
+  }[] = [];
+
+  return (
+    <div className="pr-3 py-2 analysis-settings-top border-bottom">
```
mknowlton89, line 81: The formatting is a bit wonky when the query fails

packages/front-end/components/SafeRollout/DecisionBanner.tsx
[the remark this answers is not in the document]
ahdriel: Oh that is me testing some weird stuff last night around the Callout component -- will revisit it as the fix seems to have unintentional side-effects.

packages/front-end/components/SafeRollout/AnalysisSettingsSummary.tsx
msamper: Do you happen to still have this screenshot? seems like it didn't upload

mknowlton89: Weird! 

Screenshot 2025-04-23 at 10 14 58 AM

ahdriel (approved):

PR #3655, opened 2025-02-12
[bugfix] Fix remote evaluation by adding missing await (#3654)

### LINKED ISSUE #3654: [Bug] Self-host backend remote evaluation not working (missing await)
### Summary

Remote feature evaluation is not working on self-hosted server due to a missing await keyword in evaluateFeatures call.

#### Expected Behavior

Remote feature evaluation should work correctly on self-hosted server

#### Current Behavior

Remote feature evaluation fails on self-hosted server
The evaluateFeatures call is missing the await keyword, causing the function to return a Promise instead of the actual evaluation results

[controllers/features.ts#L384](https://github.com/growthbook/growthbook/blob/cdc8c9ff95a9ad01577115d4b637f8f0f3e2832a/packages/back-end/src/controllers/features.ts#L384)

vadzimsharai:
### Features and Changes

Fixed remote feature evaluation in self-hosted backend by adding missing `await` keyword to `evaluateFeatures `call in `getEvaluatedFeaturesPublic` controller. Previously, the function was returning a raw `Promise` instead of waiting for the evaluation results, causing incorrect behavior in remote feature evaluation.

Fixes #3654

### DIFF
--- a/packages/back-end/src/controllers/features.ts
+++ b/packages/back-end/src/controllers/features.ts
@@ -381,7 +381,7 @@ export async function getEvaluatedFeaturesPublic(req: Request, res: Response) {
     res.set("Cache-control", "no-store");
 
     // todo: don't use link. investigate why clicking through returns the stub only.
-    const payload = evaluateFeatures({
+    const payload = await evaluateFeatures({
       payload: defs,
       attributes,
       forcedVariations,


bryce-fitzsimons (approved):
thanks for the fix!

PR #3665, opened 2025-02-13
Add rule ids to SDK payload

jdorn:
### Features and Changes

New option for SDK Connections to include feature rule ids in the SDK Payload.  This can be used to identify which rule was used during evaluation.  This setting will default to `true` for all newly created SDK Connections, but will remain off for existing connections for backwards compatibility.

Rule ids are supported in the following SDKs:
- Javascript: `>=0.17.0`
- C#: `>=1.1.0`
- Flutter: `>=3.4.0`
- Go: `>=0.1.4`
- Java: `>=0.9.3`
- Kotlin: `>=1.1.50`
- Python: `>=1.1.0`
- Swift: `>=1.0.49`

It is not supported at all in the following SDKs:
- PHP
- Ruby
- Elixir

In unsupported SDKs and versions, the field will be ignored, so this setting should be safe for everyone to enable.  The one exception is old Python versions, which will throw an error when an unknown field is encountered.  SDK Connections marked as Python `<=1.0.0` will strip out this new field in the payload to protect against this, ignoring the `includeRuleIds` setting entirely.

Testing:
- [x] Existing SDK connections should NOT get rule id added to the payload (backwards compatible)
- [x] Editing an existing SDK connection and checking the box SHOULD get the rule id added
- [x] Newly created SDK connections via the UI should default to the box being checked
- [x] Python < 1.1.0 should NOT get rule ids in the payload, even if the box is checked


packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -158,6 +159,7 @@ export async function validatePostPayload(
     includeDraftExperiments = false,
     includeExperimentNames = false,
     includeRedirectExperiments = false,
+    includeRuleIds = false,
```
Kevin-Chant, line 162: Should this also enforce the new default-true behavior?
```suggestion
    includeRuleIds = true,
```

jdorn: I think that also affects creating SDK Connections via the API.  I kind of wanted to leave that part alone and 100% backwards compatible.

Kevin-Chant: Ah in that case never mind. I assumed we'd want the API to have the new default behavior

Kevin-Chant (approved):
LGTM

PR #3668, opened 2025-02-13
Quick design changes and fixes

Auz:
- Fixed errors with numeric attribute assignment fields
- Fixed problem with inverted colors on tooltips
- Adjusted design of the upload screen shot section to match latest designs
- Added more clear text for GA4 users who are using GTM from within the app
- Showed activation metrics and segments to the Analysis Settings section


bryce-fitzsimons:
Maybe table the variations table in this PR. Bandits look really weird, as well as images of variable height
image

bryce-fitzsimons:
also shared experiments
image

Auz:
I updated the styles to wrap and be more responsive. Can you see how it looks now?

packages/front-end/components/Experiment/TabbedPage/AnalysisSettings.tsx
```diff
@@ -150,26 +151,46 @@ export default function AnalysisSettings({
...
+              <div className="col-4 mb-4">
+                <div className="h5">Activation Metric</div>
+                <div>
+                  {getExperimentMetricById(experiment.activationMetric)?.name}
+                </div>
+              </div>
+            )}
+
+            <div className="col-4 mb-4">
+              <div className="h5">Segments</div>
```
jdorn, line 176: This should be singular since you can only select one.  Also, we don't really want people using segments for experiments, so maybe only show this if a segment is set to avoid drawing too much attention to it.

packages/front-end/components/Experiment/VariationsTable.tsx
```diff
@@ -93,104 +99,157 @@ const VariationsTable: FC<Props> = ({
...
+            flexShrink="1"
+            flexBasis={{
+              // This might be a bit confusing, but 'gap' in flex box is not included in the flex basis width,
+              // which means percentages can't just be a simple division.
+              // This checks and does the math to make it fix perfectly
+              initial: `calc(100% / ${cols} - var(--space-${gap}) / ${cols} * (${cols} - 1))`,
+              sm: `calc(100% / ${cols - 1} - var(--space-${gap}) / ${
+                cols - 1
+              } * (${cols - 1} - 1))`,
+              md: `calc(100% / ${cols} - var(--space-${gap}) / ${cols} * (${cols} - 1))`,
```
jdorn, line 147: Can we use the `<Grid>` component instead of Flex to do this?  Seems to fit better with what we want here.  https://www.radix-ui.com/themes/docs/components/grid

jdorn (approved):

packages/front-end/components/Experiment/VariationsTable.tsx
Auz: done

packages/front-end/components/Experiment/TabbedPage/AnalysisSettings.tsx
Auz: fixed

PR #3673, opened 2025-02-18
Get readonly/full member/invite details for license

tzjames:
### Features and Changes

In addition to continuing to record the full seatsInUse that is member+invitees, we will now also record fullMember/readOnly/invite counts separately so that we can charge different rates for these values going forward.

### Dependencies

Dependent on Central License Server Pr: https://github.com/growthbook/central-license-server/pull/34

### Testing
set the LICENSE_SERVER_URL to be local dev server.
For MultiOrg:
- Add 5 noaccess/readonly members to an org and 2 invitees
- For one have one of their project roles be collaborator
- For another have them be in a team whose global role is collaborator
- For a third have one of the teams' project roles by a collaborator
- In a second org add one of the two remaining read only members from the first org but make them a collaborator
- In second org make one of the invitees from first org a member
- Refresh the license. Look at the license data in Mongo see 4 full members, 1 read only, and 1 invitee.

For Cloud
- Add 4 noaccess/readonly members to an org
- For one have one of their project roles be collaborator
- For another have them be in a team whose global role is collaborator
- For a third have one of the teams' project roles by a collaborator
- In a second org add one of the two remaining read only members from the first org but make them a collaborartor
- Refresh the license. Look at the license data in Mongo.

For offline key
- Add a LICENSE_KEY to .env.var that is an offline key
- Go to /admin if still on multiOrg can click "Download Report" 
- Open the file and see the data.


packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -303,25 +303,14 @@ export async function findOrganizationByStripeCustomerId(id: string) {
...
   if (IS_CLOUD) {
-    throw new Error("getAllInviteEmailsInDb() is not supported on cloud");
+    throw new Error("getAllOrgMemberInfoInDb() is not supported on cloud");
   }
-
-  const organizations = await OrganizationModel.find(
+  return await OrganizationModel.find(
     {},
-    { "invites.email": 1 }
+    "id invites.email members.id members.role members.projectRoles.role members.teams"
```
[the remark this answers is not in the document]
tzjames: That's what I had originally but since it is a Mongoose model the object syntax wasn't working.  I can try it again.

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -303,25 +303,21 @@ export async function findOrganizationByStripeCustomerId(id: string) {
   return doc ? toInterface(doc) : null;
 }
 
-export async function getAllInviteEmailsInDb() {
+export async function getAllOrgMemberInfoInDb() {
```
jdorn, line 306: Might want to add an explicit return type here since it's hard to infer from the Mongo query projection.

packages/back-end/src/services/licenseData.ts
```diff
@@ -66,38 +75,128 @@ export async function getLicenseMetaData() {
...
+
+  for (const userId of Object.keys(userIdsToEmailHash)) {
+    const roles = getMemberRoles(organizations, userId, teamIdToTeamMap);
+    const isReadOnly = roles.every(isReadOnlyRole);
+    const emailHash = userIdsToEmailHash[userId];
+
+    if (isReadOnly) {
+      readOnlyMembersSet.add(emailHash);
+    } else {
+      fullMembersSet.add(emailHash);
```
jdorn, line 174: We changed how we're getting the list of all user ids for self-hosted.  Before it only included users added to at least 1 org.  Now it includes everyone in Mongo (including orphaned users).

I think if `roles` is an empty array, that means they are orphaned and we should not add them to any of the sets.

packages/shared/src/enterprise/licenseUtil.ts
```diff
@@ -149,7 +149,11 @@ export interface LicenseInterface {
   };
   freeTrialDate?: Date; // Date the free trial was started
   installationUsers: {
-    [installationId: string]: { date: string; userHashes: string[] };
+    [installationId: string]: {
+      date: string;
+      userHashes: string[];
+      licenseUserCodes: LicenseUserCodes;
```
jdorn, line 155: Should this be optional since it won't exist on old cached licenses?

packages/shared/tsconfig.json
```diff
@@ -13,7 +13,11 @@
     "strictBindCallApply": true,
     "strictPropertyInitialization": true,
     "rootDir": "src",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "baseUrl": ".",
+    "paths": {
+      "enterprise/license": ["packages/shared/src/enterprise/license.ts"]
+    }
```
jdorn, line 20: Not 100% sure about the implications of these tsconfig changes.  We should definitely test this in the production build as well as yarn dev since they use different typescript compilers.

tzjames: ah whoops, I had meant to remove that since it didn't work and I renamed the file as well.

packages/back-end/src/services/licenseData.ts
tzjames: Whoops.  Good catch.  I confirmed that roles is empty for an orphaned user.

jdorn (approved):

PR #3674, opened 2025-02-18
[wip] decision criteria

lukesonnet:


packages/front-end/components/Experiment/TabbedPage/ExperimentStatusIndicator.tsx
```diff
@@ -342,17 +369,146 @@ function getPowerStatus({
...
+  for (const variationResult of metricStatus.variations) {
+    const allSuperStatSigPositive = goalMetrics.every((m) =>
+      variationResult.goalMetricsSuperStatSigPositive.includes(m)
+    );
+    const guardrailFailure = guardrailMetrics.some((m) =>
+      variationResult.guardrailMetricsFailing.includes(m)
+    );
 
-  if (powerAdditionalDaysNeeded === 0) {
+    if (decisionReady) {
```
lukebrawleysmith, line 402: One note, this approach can violate the false positive rate by:
1. declaring success for superstatsig pos (as discussed)
2. if the customer keeps letting the experiment run after `powerAdditionalDaysNeeded === 0`, and keeps waiting until results are significantly positive.  This is especially problematic if we have many variations, as we advocate stopping for harmful results only if all variations show harmful results, but we advocate stopping for benefit if only one variation shows stat sig benefit. 

We should include in our documentation a warning about declaring stat sig if they let the experiment keep running after `powerReached` and insignificant results.

lukebrawleysmith (commented):
This makes sense, left a few comments.

PR #3698, opened 2025-02-22
SQL code for CUPED adjustments for ratio metrics

lukebrawleysmith:
### Features and Changes

Adding SQL code for Fact Metrics for CUPED ratio metrics. 

TODO:
1. add in functionality for non-fact metrics  - update: not supporting this for now, due to additional computational burden.
2. derive formulas for computing theta for Bandits, and implement.  For now, CUPED is unavailable for bandit ratio metrics. 
3. update docs
4. front end work permitting CUPED to be turned on for ratio metrics


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2827,7 +2888,7 @@ export default abstract class SqlIntegration
                 : ""
             }
             ${
-              data.ratioMetric
+              data.ratioMetric && !data.regressionAdjusted
```
lukesonnet, line 2891: The capping and the denominator sum are used for both RA and non-RA ratio metrics. Can't we just rewrite this to just add the new RA fields in a condition within the broader data.ratioMetric block?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2847,7 +2908,54 @@ export default abstract class SqlIntegration
...
+              SUM(${data.capCoalesceDenominator} * ${
+                    data.capCoalesceDenominatorCovariate
+                  }) AS ${
+                    data.alias
+                  }_denominator_post_denominator_pre_sum_product
+            `
+                : ""
+            }
+            ${
+              !data.ratioMetric && data.regressionAdjusted
```
lukesonnet, line 2958: Then this can just be in the "else" for the data.ratioMetric, instead of this double logical check

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3371,6 +3481,34 @@ export default abstract class SqlIntegration
             d.dimension,
             d.${baseIdType}
         )
+        ${
+          ratioMetric
```
lukesonnet, line 3485: This needs to check if RA ratio metric, right?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1199,9 +1198,17 @@ export default abstract class SqlIntegration
       "denominator_cap_value",
       "covariate_sum",
       "covariate_sum_squares",
+      "denominator_covariate_sum",
+      "denominator_covariate_sum_squares",
       "main_covariate_sum_product",
       "quantile",
       "theta",
+      "main_post_denominator_post_sum_product",
```
lukesonnet, line 1206: It feels a little weird to create a different name for `main_denominator_sum_product` and only use it in this case, but not rename the other fields. I think we should just add the new fields and keep the old fields unnamed. I also think I set us down a bad path naming wise here, and we should rename `denominator_covariate_sum` and `denominator_covariate_sum_squares` with `denominator_pre_sum` and `denominator_pre_sum_squares` to match the other new fields being added below.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3440,7 +3578,11 @@ export default abstract class SqlIntegration
       }
       SUM(${capCoalesceDenominator}) AS denominator_sum,
       SUM(POWER(${capCoalesceDenominator}, 2)) AS denominator_sum_squares,
-      SUM(${capCoalesceDenominator} * ${capCoalesceMetric}) AS main_denominator_sum_product
+      SUM(${capCoalesceMetric} * ${capCoalesceDenominator}) AS ${
+            regressionAdjusted
+              ? `main_post_denominator_post_sum_product`
```
lukesonnet, line 3583: If you do the renaming above, you can get rid of this statement here and below.

packages/stats/gbstats/gbstats.py
```diff
@@ -873,6 +925,9 @@ def process_experiment_results(
                             metric_settings_bandit.main_metric_type = "count"
                         if metric_settings_bandit.covariate_metric_type == "binomial":
                             metric_settings_bandit.covariate_metric_type = "count"
+                        # after we have added the functionality for ratio_ra, remove this
```
lukesonnet, line 928: todo

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: yes, good callout, fixed

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: It's currently inside of `regressionAdjusted = True` case, so I think it's ok.  However, given that we are pivoting and not implementing CUPED for non-fact ratio metrics, the point is moot.

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: updated

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: done

packages/stats/gbstats/gbstats.py
lukebrawleysmith: added a comment

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3453,6 +3546,18 @@ export default abstract class SqlIntegration
       `
         : ""
     }
+    ${
```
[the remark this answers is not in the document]
lukebrawleysmith: Apologies, I thought I had removed it.  I must have miscopied somewhere.

packages/front-end/components/Experiment/MetricsOverridesSelector.tsx
```diff
@@ -135,16 +134,6 @@ export default function MetricsOverridesSelector({
           );
           let regressionAdjustmentAvailableForMetric = true;
           let regressionAdjustmentAvailableForMetricReason = <></>;
-          if (
```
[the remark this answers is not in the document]
lukebrawleysmith: Sonnet resolved, per offline convo

packages/shared/src/settings/resolvers/regressionAdjustmentEnabledResolver.ts
```diff
@@ -42,8 +42,6 @@ const regressionAdjustmentResolver = (
         ? "disabled by metric override"
         : "experiment-level metric override applied";
     }
-
-    // denominator metric checks
```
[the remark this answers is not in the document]
lukebrawleysmith: I added the comment back in

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2829,33 +2880,46 @@ export default abstract class SqlIntegration
...
+                  `
+                    : `,
+                    SUM(${data.capCoalesceDenominator} * ${data.capCoalesceMetric}) AS ${data.alias}_main_denominator_sum_product
+                  `
+                }` /*ends ifelse regressionAdjusted*/
+                : ` 
               ${
-                data.isPercentileCapped
-                  ? `MAX(COALESCE(cap.${data.alias}_denominator_cap, 0)) as ${data.alias}_denominator_cap_value,`
+                data.regressionAdjusted
```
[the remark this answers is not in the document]
lukesonnet: oh I see, this is non-ratio, thanks.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2958,9 +3019,7 @@ export default abstract class SqlIntegration
     // redundant checks to make sure configuration makes sense and we only build expensive queries for the cases
     // where RA is actually possible
     const regressionAdjusted =
-      settings.regressionAdjustmentEnabled &&
-      isRegressionAdjusted(metric, denominator);
-
+      settings.regressionAdjustmentEnabled && isRegressionAdjusted(metric);
```
lukesonnet: I think we need to add `denominator` back here so that `isRegressionAdjusted` can properly block non-fact metrics.

packages/back-end/src/services/stats.ts
```diff
@@ -378,8 +379,7 @@ export function getMetricSettingsForStatsEngine(
   const ratioMetric = isRatioMetric(metric, denominator);
   const quantileMetric = quantileMetricType(metric);
   const regressionAdjusted =
-    settings.regressionAdjustmentEnabled &&
-    isRegressionAdjusted(metric, denominator);
+    settings.regressionAdjustmentEnabled && isRegressionAdjusted(metric);
```
lukesonnet, line 382: Same here, we need to add denominator back in.

packages/shared/src/experiments.ts
```diff
@@ -197,14 +197,10 @@ export function isFunnelMetric(
   return !!denominatorMetric && isBinomialMetric(denominatorMetric);
 }
 
-export function isRegressionAdjusted(
-  m: ExperimentMetricInterface,
-  denominatorMetric?: ExperimentMetricInterface
-) {
+export function isRegressionAdjusted(m: ExperimentMetricInterface) {
```
lukesonnet, line 200: We need to add denominator back in and add `!(!isFactMetric(m) && isRatioMetric(m, denominatorMetric))`, note the ! and (), I think they're right.

packages/back-end/src/integrations/SqlIntegration.ts

packages/back-end/src/services/stats.ts
lukebrawleysmith: done

packages/shared/src/experiments.ts
lukebrawleysmith: updated

lukesonnet (approved):

packages/shared/src/settings/resolvers/regressionAdjustmentEnabledResolver.ts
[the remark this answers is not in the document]
lukesonnet: this comment is still deleted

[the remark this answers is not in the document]
lukesonnet: oh no it's just spelled differently below, nvm don't worry about it

packages/stats/gbstats/gbstats.py
```diff
@@ -873,6 +921,9 @@ def process_experiment_results(
                             metric_settings_bandit.main_metric_type = "count"
                         if metric_settings_bandit.covariate_metric_type == "binomial":
                             metric_settings_bandit.covariate_metric_type = "count"
+                        # TODO: after we have added the functionality for ratio_ra, remove this
+                        if metric_settings_bandit.main_metric_type == "ra_ratio":
```
lukesonnet, line 925: Wait, this metric type shouldn't even be possible? This code chunk doesn't make a lot of sense to me. There is no `MetricType` of `ra_ratio`? There also isn't a `main_metric_type` of `ratio`?

lukebrawleysmith: good call, changed so that we now test if the `statistic_type` is `ra_ratio`

PR #3708, opened 2025-02-24
fix(UI): Tooltip on getting started cards

ahdriel:
I noticed a UI bug when the user does not have permission to import from Launch Darkly in the Getting Started page.

### Before

Screenshot 2025-02-24 at 8 00 04 PM

### After

Screenshot 2025-02-24 at 7 58 55 PM


msamper (approved):
Tested and confirmed this fixes the bug!

PR #3709, opened 2025-02-25
Change visual treatment of statuses on experiment list

jdorn:
### Features and Changes

Experiment list page:
- Split the status column into two: `Status` and results (no column header)
- Only show the status column (running, stopped, etc.) when multiple status tabs are active.  Otherwise hide it.
- Only show the result column when running and/or stopped tabs are selected.  Otherwise hide it.
- Render the status and result as plain text instead of badges
- Add an indicator dot for tests that need your attention (amber) or stopped tests (gray, red, or green)
- Define sort order as part of `StatusIndicatorData` instead of in a separate function

Other pages:
- Switch all status badges to be the `solid` variant instead of mixing and matching soft and solid
- Switch "Ship now" and "Roll back now" to use amber color instead of green/red
- Switch "Won" and "Lost" to use green/red instead of gray
- Switch "Draft" to be pink instead of indigo (to differentiate from running)

General changes to the decision framework logic:
- Add additional `No data` states for when an experiment is missing a data source or has no metrics selected
- Change all badges to use sentence case instead of title case

image

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/ExperimentStatusIndicator.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/ExperimentStatusIndicator.tsx
@@ -11,7 +11,7 @@ export type StatusIndicatorData = {
   color: React.ComponentProps<typeof Badge>["color"];
   status: string;
   detailedStatus?: string;
-  important?: boolean;
+  needsAttention?: boolean;
   tooltip?: string;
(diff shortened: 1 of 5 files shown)


lukesonnet (approved):

ahdriel (approved):
Nice! Thanks for the updates.

Issue #3703, opened 2025-02-24
[Bug] Clicking ENTER on target by attributes/Saved Groups opens new Target by Prerequisites features box


natasha-growthbook:
### Summary

Clicking ENTER on target by attributes/saved groups opens new target Prerequisites features box

https://github.com/user-attachments/assets/c8a6c1af-cf0f-4200-be41-9dffb9594ae5

#### Expected Behavior

Clicking ENTER in the targeting by attributes/saved groups box, does not open new target by Prerequisites features box

#### Current Behavior

Clicking ENTER on target by attributes/saved groups opens new target by Prerequisites features box

mknowlton89:
@ahdriel Sorry for the collision again, but I've got a PR up for this here. https://github.com/growthbook/growthbook/pull/3712

PR #3712, opened 2025-02-25
Prevents event bubbling by using RadixButton

mknowlton89:
### Features and Changes

This PR implements a fix for https://github.com/growthbook/growthbook/issues/3703. Essentially, when the "Enter" key was being pressed, the "+ Add Prerequisite feature" buttons were being called. One option I found was to give these buttons a type - e.g. `type = "button"`. Without this, the default type is `type = "submit"`, which meant hitting "Enter" was simulating a click on this button to submit the form.

I explored that route, but ultimately used this as an opportunity to switch to our new `RadixButton` component.

### Testing

- [x] Ensure that replacing the button type prevents new prereq conditions from being added to the phases.0.prerequisites array.
- [x] Ensure that this doesn't introduce regressions elsewhere in the app in similar forms.


ahdriel (approved):
Confirmed this fixes the issue.

The new behavior now is the same as in all other pages, where pressing `Enter` advances to the next page -- just noting that this is different from the expected behavior described in the issue, but I think if we were going to change for this component we also should update the other ones.

Thanks for the quick fix!

PR #3715, opened 2025-02-25
Adjust field order in New Feature modal

jdorn:
### Features and Changes

Changes:
- Re-order fields to be more intuitive
- Only show project selector when there is at least 1 project
- Move alert into a tooltip

image


msamper (approved):
Much more intuitive and more like the experiment modal now, nice! LGTM

PR #3717, opened 2025-02-25
Improve UX for environment selector in feature rules

### LINKED ISSUE #3574: [Feature] Tabular selection of environment
### Description of Feature

Feedback: Preferred the tabular selection of environment (instead of the dropdown). It is very counterproductive. The environment is one of the MOST important elements. I need to have clear idea on which environment I am acting upon, not looking throughout the whole page looking for a subtle dropdown to figure this out.

Kevin-Chant:
### Features and Changes

Changes the UX for environment selection to always show the first 4 environments as separate tabs, and the remaining environments in a dropdown.
Style updates to make the dropdown fit in alongside the tabs

- Closes #3574 

#### <=5 environments
image
image

#### >5 environments
image

env-selector

The only remaining UI weirdness (visible in the above gif) is that typing in the dropdown will sometimes cause the caret to clip with the text. This is a general issue with our select containers rather than specific to this component so I've left it be for now.

### DIFF
--- a/packages/front-end/components/Features/FeatureRules.tsx
+++ b/packages/front-end/components/Features/FeatureRules.tsx
@@ -124,7 +124,6 @@ export default function FeatureRules({
                 ))}
                 <Flex
                   px="1"
-                  height="100%"
                   direction="column"
                   justify="center"
                   align="center"
@@ -151,6 +150,7 @@ export default function FeatureRules({
                           </Flex>
                           <Badge
                             ml="2"
+                            mr="3"
                             label={rulesByEnv[value].length.toString()}
                             radius="full"
                             variant="solid"
--- a/packages/front-end/components/Features/FeatureRules.tsx
+++ b/packages/front-end/components/Features/FeatureRules.tsx
@@ -1,5 +1,5 @@
 import { FeatureInterface } from "back-end/types/feature";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
 import {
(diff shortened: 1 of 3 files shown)


packages/front-end/components/Features/FeatureRules.tsx
```diff
@@ -81,79 +82,98 @@ export default function FeatureRules({
...
+                ))}
+                <Flex
+                  px="1"
+                  direction="column"
+                  justify="center"
+                  align="center"
+                  className={clsx("tab-trigger-container", {
+                    active: !!selectedDropdownEnv,
+                  })}
+                >
```
jdorn, line 133: Only render this if `dropdownEnvs` is not empty?

packages/front-end/components/Features/FeatureRules.tsx
```diff
@@ -81,79 +82,98 @@ export default function FeatureRules({
     })
   );
 
+  const [tabEnvs, dropdownEnvs] = useMemo(() => {
+    const tabEnvs: Environment[] = [];
+    const dropdownEnvs: Environment[] = [];
+    environments.forEach((env) => {
+      if (env.toggleOnList) {
```
jdorn, line 89: Thinking about this more, it might be better to just always render the first X environments instead of using `toggleOnList`.  There are valid reasons for having this setting off for some important environments.  For example, if you don't want people to accidentally toggle something on production.

It's also good to add some special case logic so we don't end up with a dropdown with a single item in it.  Something like

```ts
if (environments.length <= 5 || i < 5) {
  // tab
} else {
  // dropdown
}
```

jdorn (approved):

PR #3719, opened 2025-02-26
Allow empty lists for feature $in and $nin conditions

### LINKED ISSUE #3657: [Bug] (js sdk) Rule with empty list passes when attribute isn't set
### Summary

The "attribute is in the list" condition does not work as expected when the list "is empty" - it isn't actually empty.

In the UI, a feature is created with default value `true` and a force rule that serves `false` based on an "attribute is in the list" condition. The list is empty. Not all clients have the specified attribute set.

#### Expected Behavior

Since I have not specified any value in my rule, I expect that the rule should never match anything, always serving the feature with the default value `true`.

#### Current Behavior

**_The list is in fact not empty, but rather contains an empty string as its only value_**. For a client that doesn't have this specific attribute set, this will result in passing the condition and the feature being served with the forced value `false`.

It seems like the missing attribute is represented as `""` and the "empty" list is represented as `[""]`
```
[""].includes("") => true
```

Kevin-Chant:
### Features and Changes

Currently, leaving the text field empty when creating an "in the list" targeting condition leads not to an empty list but to the list `[""]`. This then results in the opposite of desired behavior when checking membership of an attribute that is only sometimes set.

- Closes #3657

### Testing

Create a feature with an empty list condition without these changes and inspect the SDK payload to see it rendered as `$in: [""]`.
Then, with the changes in place, create a second condition (or update the first), and observe `$in: []`

### Incorrect `main` behavior
image

### Correct behavior
image

### DIFF
--- a/packages/front-end/services/features.ts
+++ b/packages/front-end/services/features.ts
@@ -1055,10 +1055,15 @@ export function condToJson(
     } else if (operator === "$notEmpty") {
       obj[field]["$size"] = { $gt: 0 };
     } else if (operator === "$in" || operator === "$nin") {
-      obj[field][operator] = value
-        .split(",")
-        .map((x) => x.trim())
-        .map((x) => parseValue(x, attributes.get(field)?.datatype));
+      // Allow for the empty list
+      if (value === "") {
+        obj[field][operator] = [];
+      } else {
+        obj[field][operator] = value
+          .split(",")
+          .map((x) => x.trim())
+          .map((x) => parseValue(x, attributes.get(field)?.datatype));
+      }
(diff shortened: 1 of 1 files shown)


jdorn (approved):
This fix seems like something we should do anyway, but is this also a bug in some of our SDKs?

According to the issue, the attribute is "missing", which I take to mean undefined/null.  I double checked in javascript and it works as expected and returns false.  I wonder if some of the other languages have this bug though.

```ts
const condition = {value: {$in: [""]}};
const attributes = {};

// False
consiole.log(evalCondition(attributes, condition))
```

PR #3729, opened 2025-02-26
Add support to key-pair authentication for Snowflake data source

### LINKED ISSUE #3601: [Feature] Snowflake: key pair authentication
### Description of Feature

Support for Snowflake key pair authentication on cloud-hosted GrowthBook

ahdriel:
Resolves #3601

### Features and Changes

Add key-pair as a new authentication method to a Snowflake data source.

### Screenshots

Screenshot 2025-02-27 at 1 45 22 PM

Screenshot 2025-02-26 at 7 08 30 PM


packages/front-end/components/Settings/SnowflakeForm.tsx
```diff
@@ -35,19 +40,93 @@ const SnowflakeForm: FC<{
...
+            }
+          />
+        </div>
+      ) : null}
+
+      {authMethod === "key-pair" ? (
+        <>
+          <div className="form-group col-md-12">
+            <label>Private Key File</label>
+            <input
```
ahdriel, line 82: TODO: Create a custom component so we can show `(Existing)` as well for users who already have a private key uploaded.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -513,6 +514,35 @@ export async function putDataSource(
     if (params) {
       const integration = getSourceIntegrationObject(context, datasource);
       mergeParams(integration, params);
+
+      // TODO: Should the validation be done here or somewhere else?
```
jdorn, line 518: Not here. This method runs `testConnection` to verify the params before saving. So as long as running the query fails, you don't need any additional validation here.

ahdriel: Removed and the error message was fixed by adding a proper try/catch when decrypting the key.

jdorn (approved):

PR #3738, opened 2025-02-28
[WIP] Migrate ExperimentSnapshotModel to BaseModel

msamper:
### Features and Changes

Migrates over the `ExperimentSnapshotModel` to the new Base Model

TODO
- [ ] Rework tests 
- [x] QA
- [x] Check if anything is missing on the public API side


packages/back-end/src/models/BaseModel.ts
```diff
@@ -426,6 +426,18 @@ export abstract class BaseModel<
...
+      .aggregate(pipeline)
+      .toArray();
+
+    const migrated = rawDocs.map((d) =>
+      this.migrate(this._removeMongooseFields(d))
+    );
+
+    return migrated;
+  }
+
```
msamper, line 440: Do this within new `ExperimentSnapshotModel` inside of `getLatestSnapshotMultipleExperiments` and add
```
    const migrated = rawDocs.map((d) =>
      this.migrate(this._removeMongooseFields(d))
    );
    const filtered = bypassReadPermissionChecks
      ? migrated
      : await this.filterByReadPermissions(migrated);
```

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,374 @@
...
+      );
     }
+
+    return this.context.permissions.canCreateExperimentSnapshot(datasource);
+  }
+  protected canUpdate(existing: ExperimentSnapshotInterface): boolean {
+    return this.canCreate(existing);
+  }
+  protected canDelete(doc: ExperimentSnapshotInterface): boolean {
+    return this.canCreate(doc);
```
msamper, line 78: Unsure about these permissions since it seems like we didn't do any at the model level. I only found use of the `canCreateExperimentSnapshot` permission check within the public API

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,374 @@
...
-    {
-      organization,
-      id,
-      "analyses.settings": analysis.settings,
-    },
-    {
-      $set: { "analyses.$": analysis },
-    }
-  );
+  public async findRunningByQueryId(ids: string[]) {
```
msamper, line 119: Seems like we might not have context available when we need to call this. Does this function need to live elsewhere?

msamper: See `expireOldQueries` file for more details

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -458,7 +457,8 @@ export const notifyExperimentChange = async ({
   // do not fire significance events for old snapshots that have no type
   if (snapshot.type === undefined) return;
   // do not fire for snapshots where statistics are manually entered in the UI
-  if (snapshot.manual) return;
+  // TODO: Do we still need this check? Will we ever have legacy snapshots hitting this function?
```
msamper, line 460: Do we still need this check? Will we ever have legacy snapshots hitting this function?

packages/back-end/src/types/Audit.ts
```diff
@@ -49,6 +49,7 @@ export const entityEvents = {
   factTable: ["autocreate"],
   customField: ["create", "update", "delete"],
   experimentTemplate: ["create", "update", "delete"],
+  experimentSnapshot: ["create", "update", "delete"],
```
msamper, line 52: Do we want to audit this? Not sure if we'd need to audit scheduled snapshot creation

packages/back-end/src/jobs/expireOldQueries.ts
```diff
@@ -42,20 +38,17 @@ const expireOldQueries = trackJob(JOB_NAME, async () => {
   }
 
   // Look for matching snapshots and update the status
+  // TODO: How do we get the context here?
   const snapshots = await findRunningSnapshotsByQueryId([...queryIds]);
```
msamper, line 42: see comment on `ExperimentSnapshotModel`

msamper:
Investigating some bugs with the validators

msamper:
> Investigating some bugs with the validators

Update: It seems like the problem is that when we call the stats library to `process_multiple_experiment_results` it will return `None` for a bunch of properties that may not be available or are not relevant (i.e. `banditResult`) Then when we convert the results to JSON with `json.dumps(...)` the `None` values are converted to `Null`. The `ExperimentSnapshotInterface` I followed for creating the different validators marked most of these fields that can have `None` values as optional but not as nullable so it seems like the interface was just typed incorrectly and I'll need to fix that for the validators to avoid the `ZodError`s I'm seeing

packages/back-end/src/controllers/reports.ts
```diff
@@ -142,7 +139,7 @@ export async function postReportFromSnapshot(
 
   // Save the snapshot
   snapshot.report = doc.id;
-  await createExperimentSnapshotModel({ data: snapshot, context });
+  await context.models.experimentSnapshots.create(snapshot);
```
msamper, line 142: remove `organization`, `dateCreated`, or `dateUpdated`

packages/back-end/src/services/experiments.ts
```diff
@@ -645,7 +640,7 @@ export async function createManualSnapshot({
     triggeredBy: "manual",
   };
 
-  return await createExperimentSnapshotModel({ data, context });
+  return await context.models.experimentSnapshots.create(data);
```
msamper, line 643: remove `organization`, `dateCreated`, and `dateUpdated` from `data`

packages/back-end/src/services/reports.ts
```diff
@@ -420,10 +416,9 @@ export async function createReportSnapshot({
     snapshotData.health.traffic.dimension = {};
   }
 
-  const snapshot = await createExperimentSnapshotModel({
-    data: snapshotData,
-    context,
-  });
+  const snapshot = await context.models.experimentSnapshots.create(
```
msamper, line 419: remove `organization`, `dateCreated`, and `dateUpdated` from `data`

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,362 @@
...
-  if (shouldUpdateExperimentAnalysisSummary) {
-    const experimentModel = await getExperimentById(
-      context,
-      experimentSnapshotModel.experiment
+    return this.context.permissions.canReadSingleProjectResource(
+      experiment?.project
     );
+  }
+  protected canCreate(doc: ExperimentSnapshotInterface): boolean {
+    // const { datasource, experiment } = this.getForeignRefs(doc);
```
msamper, line 58: Grabbing the datasource didn't seem to work. Assuming it's because the datasource is on the experiment and not the snapshot. If this is the right permission to use, what would be the right way to grab the datasource here?

lukesonnet: `datasourceId` is in the `snapshot.settings` object. I think to keep the benefits of the base model we probably need to allow for foreign refs to come from nested fields with some kind of optional argument that specifies a nested field to also look in, but I obv defer to Jeremy on this.

packages/back-end/src/jobs/expireOldQueries.ts
msamper, line 42: try to make a `getContextForGlobalOrg` to get this to work

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: No, we won't, but those legacy items get JIT migrated to put this field in `snapshot.settings.manual`, so if you want to use that instead that would be great!

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,362 @@
...
-  return doc ? toInterface(doc) : null;
-}
+  public async updateSnapshotAnalysis({
+    id,
+    analysis,
+  }: {
+    id: string;
+    analysis: ExperimentSnapshotAnalysis;
+  }) {
+    const snapshotCollection = await super._dangerousGetCollection();
```
msamper, line 159: we have updateOne and findOne. Use internal methods

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,362 @@
...
-    status: "running",
-    dateCreated: { $gt: earliestDate },
-    queries: { $elemMatch: { query: { $in: ids }, status: "running" } },
-  });
+  public async addOrUpdateSnapshotAnalysis(
+    params: AddOrUpdateSnapshotAnalysisParams
+  ) {
+    const { id, analysis } = params;
+    const organization = this.context.org.id;
+    const snapshotCollection = await super._dangerousGetCollection();
```
msamper, line 192: use internal methods

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,362 @@
...
+  }
 
-export async function findSnapshotById(
-  organization: string,
-  id: string
-): Promise<ExperimentSnapshotInterface | null> {
-  const doc = await ExperimentSnapshotModel.findOne({ organization, id });
-  return doc ? toInterface(doc) : null;
-}
+  public async updateSnapshotAnalysis({
```
lukesonnet, line 152: TODO(luke): i wanna QA this.

packages/back-end/src/types/Audit.ts
lukesonnet: I think, yes?

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,362 @@
...
-    await updateSnapshotAnalysis({ organization, id, analysis });
+    const snapshots = await super._find({
+      status: "running",
+      dateCreated: { $gt: earliestDate },
+      queries: { $elemMatch: { query: { $in: ids }, status: "running" } },
+    });
+
+    return snapshots.map((doc) => this.migrate(doc));
   }
-}
```
msamper, line 124: create an intermediary method used by canUpdate and updateOnPhase delete that checks permissions based on the experiment

packages/back-end/src/types/Audit.ts
msamper: Jeremy recommended not auditing here. Will likely be very noisy

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -1,517 +1,362 @@
...
+    //   );
+    // }
 
-    if (experimentModel && isLatestPhase) {
-      await updateExperimentAnalysisSummary({
-        context,
-        experiment: experimentModel,
-        experimentSnapshot: experimentSnapshotModel,
-      });
+    // return this.context.permissions.canCreateExperimentSnapshot(datasource);
```
msamper, line 66: change `canCreateExperimentSnapshot` to take in an experiment instead of a datasource and have public API use `canRunExperimentQueries` instead

msamper:
We decided this is too risky given how many areas of the code this touches so we're closing this PR for now

Issue #3737, opened 2025-02-27
Quantile metrics inappropriately showing results when sample size is too small


lukebrawleysmith:
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/47faf35e-d46d-443f-a287-39c606aa523f" />

### Summary

Need to suppress some tooltip data when a quantile metric does not have enough data. 

#### Expected Behavior

As shown in the screenshot below, our UI correctly does not display the `p-value` when there is not enough data for a quantile metric.  `% change`, `95% confidence interval`, and `p-value`,  should all be suppressed in the TT.  

#### Current Behavior

`% change`, `95% confidence interval`, and `p-value` are shown in the TT.

PR #3744, opened 2025-02-28
Hide exp data if exp doesn't have enough data

mknowlton89:
### Features and Changes

Per the linked issue, we want to hide data (e.g. `percent changed`, `confidence interval`, and `chance to win`) in the event that we don't have enough data. This PR adds a number of `data.rowResults.enoughData` checks in the Tooltip to show it only when the experiment has enough data.

This PR also fixes an alignment issue with the `Significant`/`Not Significant` tooltip.

- Closes https://github.com/growthbook/growthbook/issues/3737

### Dependencies

None

### Testing

- [x] Ensure that if a particular metric doesn't have enough data, we're only showing the "Not enough data" warning and not anything else
- [x] Ensure that if a particular metric DOES have enough data, we're showing everything we need to

### Screenshots

Before - Not enough data -  (Light Mode)
Screenshot 2025-03-03 at 7 33 00 AM

After - Not enough data - (Light Mode)
Screenshot 2025-03-03 at 7 31 43 AM

Before - Not enough data - (Dark Mode)
Screenshot 2025-03-03 at 7 33 11 AM

After - Not enough data - (Dark Mode)
Screenshot 2025-03-03 at 7 31 52 AM

Before - enough data -  (Light Mode)
Screenshot 2025-03-03 at 7 36 02 AM

After - enough data - (Light Mode)
Screenshot 2025-03-04 at 9 57 42 AM

Before - enough data - (Dark Mode)
Screenshot 2025-03-03 at 7 35 52 AM

After - enough data - (Dark Mode)
Screenshot 2025-03-04 at 9 57 34 AM


lukebrawleysmith:
I'm not the person to review line-by-line, but I tested the changes locally, and confirmed same story as the screenshots above.  Thanks Michael!

packages/front-end/components/Experiment/ResultsTableTooltip/ResultsTableTooltip.tsx
```diff
@@ -411,103 +413,106 @@ export default function ResultsTableTooltip({
...
+                    })}
+                  >
+                    {data.statsEngine === "bayesian"
+                      ? percentFormatter.format(data.stats.chanceToWin ?? 0)
+                      : pValText}
+                  </div>
+                </div>
+              </>
+            ) : null}
+            {!addLiftWarning && data.rowResults.enoughData ? (
```
ahdriel, line 512: question: Why are we inverting the condition here? From should add lift warning to not?

ahdriel (approved):
One additional question: In the screenshots shared where it has enough data, the `after` has a `CUPED affects results` banner while the before does not.

Just want to make sure this is just a product from how we took the screenshots and not a bug.

Thanks for the improvements 🚀

packages/front-end/components/Experiment/ResultsTableTooltip/ResultsTableTooltip.tsx
mknowlton89: Thank you for catching that - this is an artifact from me forcing a variation to display for testing. I'll remove the bang operator.

mknowlton89: This is what was causing the "CUPER affects results" banner to display in the `after` screenshots.

PR #3747, opened 2025-03-01
Remove API keys from user context (not used)

jdorn:
### Features and Changes

We were sending down a sanitized list of all API keys in an account to the front-end for every request.  It only included the id and name, so no secrets were being leaked, but it still felt weird to do this.  When digging in, it turns out this data is not actually used in our front-end at all.  This PR stops sending this to the front-end.


jdorn:
@greptileai

PR #3753, opened 2025-03-03
Awaits upgrade call

mknowlton89:
### Features and Changes

Previously, we weren't awaiting the `startPro` trial, so the loading state wasn't being displayed properly.

### Testing

- [x] Ensure the loading state is displayed when `Upgrade Now` is selected
- [x] Ensure no regressions are introduced as a result of this update


packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -530,7 +529,11 @@ export default function UpgradeModal({
               <PiCaretRight />
             </>
           }
-          submit={featureFlagValue !== "OFF" ? () => onSubmit() : undefined}
+          submit={
+            featureFlagValue !== "OFF"
+              ? async () => await onSubmit()
```
jdorn, line 534: nit: just pass in `onSubmit` instead of `async () => await onSubmit()`

PR #3755, opened 2025-03-03
Fix unhealthy status badges for exps with no traffic query

lukesonnet:
This PR adds a fallback for `totalUsers` and `srm` in the `analysisSummary` to populate them even when the health query is off. 

This enables the top-level unhealthy status for users even if they don't have the health query activated.  Previously, those users were not getting their `analysisSummary` set so we would show them unhealthy banners on their results tab but no top-level status on the experiments list.

This also solves a bug where bandit users with the health query off were getting a "No data" warning even if they did have data. This was happening because `analysisSummary` was getting set for these cases because (1) `totalUsers` was 0 owing to the fact that the experiment snapshot units array wasn't missing but rather of length 0 and (2) bandit gets SRM from a different place than non-bandits. Item (2) is not true for standard experiments, so this bug was limited to bandits only.

### Testing

https://www.loom.com/share/892ea9dcc7004ac296cec8fd958f343c

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -2814,17 +2814,18 @@ export async function updateExperimentAnalysisSummary({
 
   const standardSnapshot =
     experimentSnapshot.type === "standard" &&
-    experimentSnapshot.analyses?.[0].results.length === 1;
-  const totalUsers = overallTraffic?.variationUnits.length
-    ? overallTraffic.variationUnits.reduce((acc, a) => acc + a, 0)
-    : standardSnapshot
-    ? // fall back to first result for standard snapshots if overall traffic
-      // is missing
-      experimentSnapshot?.analyses?.[0]?.results?.[0]?.variations?.reduce(
-        (acc, a) => acc + a.users,
-        0
-      )
-    : null;
+    experimentSnapshot.analyses?.[0]?.results?.length === 1;
+  const totalUsers =
+    (overallTraffic?.variationUnits.length
+      ? overallTraffic.variationUnits.reduce((acc, a) => acc + a, 0)
+      : standardSnapshot
+      ? // fall back to first result for standard snapshots if overall traffic
+        // is missing
+        experimentSnapshot?.analyses?.[0]?.results?.[0]?.variations?.reduce(
+          (acc, a) => acc + a.users,
+          0
+        )
+      : null) ?? null;
 
   const srm = getSRMValue(experiment.type ?? "standard", experimentSnapshot);
 
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -2812,14 +2812,23 @@ export async function updateExperimentAnalysisSummary({
   const overallTraffic = experimentSnapshot.health?.traffic?.overall;
   const snapshotHealthPower = experimentSnapshot.health?.power;
 
-  const totalUsers = overallTraffic?.variationUnits?.reduce(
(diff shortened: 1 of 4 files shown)


packages/shared/src/health/health.ts
```diff
@@ -102,8 +102,23 @@ export function getSRMValue(
...
+
+      if (healthQuerySRM !== undefined) {
+        return healthQuerySRM;
+      }
+      // fall back to main results SRM for no dimension split snapshots
+      // and without health query SRM
+      // if no dimension && only one overall result (e.g. no dim splits)
+      if (
+        snapshot.type === "standard" &&
+        snapshot.analyses?.[0].results.length === 1
```
jdorn, line 116: If `snapshot.analyses` is an empty array, this will throw an error.

packages/back-end/src/services/experiments.ts
```diff
@@ -2812,14 +2812,23 @@ export async function updateExperimentAnalysisSummary({
   const overallTraffic = experimentSnapshot.health?.traffic?.overall;
   const snapshotHealthPower = experimentSnapshot.health?.power;
 
-  const totalUsers = overallTraffic?.variationUnits?.reduce(
-    (acc, a) => acc + a,
-    0
-  );
+  const standardSnapshot =
+    experimentSnapshot.type === "standard" &&
+    experimentSnapshot.analyses?.[0].results.length === 1;
```
jdorn, line 2817: Same here.  Empty analyses array will throw an error.

packages/back-end/src/services/experiments.ts
```diff
@@ -2812,14 +2812,23 @@ export async function updateExperimentAnalysisSummary({
...
+  const totalUsers = overallTraffic?.variationUnits.length
+    ? overallTraffic.variationUnits.reduce((acc, a) => acc + a, 0)
+    : standardSnapshot
+    ? // fall back to first result for standard snapshots if overall traffic
+      // is missing
+      experimentSnapshot?.analyses?.[0]?.results?.[0]?.variations?.reduce(
+        (acc, a) => acc + a.users,
+        0
+      )
+    : null;
```
jdorn, line 2827: I believe the type of `totalUsers` will be `number | undefined | null` here.  We might want to coerce it to only have `null` or `undefined` and not both.

jdorn (approved):

packages/back-end/src/services/experiments.ts
```diff
@@ -2812,14 +2812,24 @@ export async function updateExperimentAnalysisSummary({
...
+        experimentSnapshot?.analyses?.[0]?.results?.[0]?.variations?.reduce(
+          (acc, a) => acc + a.users,
+          0
+        )
+      : null) ?? null;
 
   const srm = getSRMValue(experiment.type ?? "standard", experimentSnapshot);
 
-  if (overallTraffic && totalUsers !== undefined && srm !== undefined) {
+  if (overallTraffic && srm !== undefined) {
```
ahdriel, line 2832: question: Does this (overallTraffic being defined) mean that we are still only updating `analysisSummary.health` (where both SRM & Multiple Exposures live) if we have executed the health query?

Shouldn't we remove this check given that we can get SRM & Multiple Exposures even without the Health query?

Maybe I am misunderstanding something.

lukesonnet: Oh that's a good call. What's happening is that due to mongoose or something else, I'm not sure, this field does always exist, it's just empty. I'll remove that check in my notifications PR, thank you.

PR #3756, opened 2025-03-03
Enable 'Edit status' escape hatch for drafts with data

jdorn:
### Features and Changes

Allow editing an experiment draft status directly (without going through the "Start Experiment" flow), but only if it already has results.  Also, when going from `draft` -> `running` for an experiment, only reset the phase start date if there are no results yet.  

This enables a common workflow, where you need to quickly revert back to a draft to make a tweak to the implementation settings.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -2749,12 +2749,14 @@ export async function getChangesToStartExperiment(
 
   const changes: Changeset = {};
 
-  // use the current date as the phase start date
-  phases[lastIndex] = {
-    ...phases[lastIndex],
-    dateStarted: new Date(),
-  };
-  changes.phases = phases;
+  // If the experiment doesn't have any results yet, reset the phase start date to now
+  if (!experiment.analysisSummary?.snapshotId) {
+    phases[lastIndex] = {
+      ...phases[lastIndex],
+      dateStarted: new Date(),
+    };
+    changes.phases = phases;
+  }
 
   // Bandit-specific changes
   if (experiment.type === "multi-armed-bandit") {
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -2749,12 +2749,14 @@ export async function getChangesToStartExperiment(
...
   const changes: Changeset = {};
 
-  // use the current date as the phase start date
-  phases[lastIndex] = {
-    ...phases[lastIndex],
-    dateStarted: new Date(),
-  };
-  changes.phases = phases;
+  // If the experiment doesn't have any results yet, reset the phase start date to now
+  if (!experiment.analysisSummary?.snapshotId) {
```
lukesonnet, line 2753: Relying on this means that if I stopped an experiment before we landed this code, and then made these changes, I'd start a new phase using this backdoor even though results existed, and that's a pretty unintuitive behavior, IMO.

IMO we should mimic the logic that enables or disables the backdoor here, specifically:
```
const hasResults = !!analysis?.results?.[0];
```

jdorn: This doesn't start a new phase.  It only resets the latest phase start date.  And this would only happen if you first set the old stopped experiment status to `draft` and then to `running`.

I think that is sufficiently rare that it's not worth an extra db lookup to pull in snapshot data.

lukesonnet (approved):

packages/back-end/src/services/experiments.ts
lukesonnet: Resetting the latest phase date is less destructive, so I'm ok with this.

PR #3757, opened 2025-03-03
Only refresh fact table columns when needed

jdorn:
### Features and Changes

Updating anything on a fact table was triggering a full refresh, which ran a SQL query to check if any columns have changed.  This is not necessary when just changing things like `tags` or `projects`. 

This PR changes the logic to only run a full refresh if the SQL definition changes.

### DIFF
--- a/packages/back-end/src/api/fact-tables/updateFactTable.ts
+++ b/packages/back-end/src/api/fact-tables/updateFactTable.ts
@@ -58,7 +58,9 @@ export const updateFactTable = createApiRequestHandler(
     const data: UpdateFactTableProps = { ...req.body };
 
     await updateFactTableInDb(req.context, factTable, data);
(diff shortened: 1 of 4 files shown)


packages/back-end/src/api/fact-tables/updateFactTable.ts
```diff
@@ -69,3 +71,7 @@ export const updateFactTable = createApiRequestHandler(
     };
   }
 );
+
+export function needsColumnRefresh(changes: UpdateFactTableProps): boolean {
```
lukesonnet, line 75: We should add a way for this column refresh button to force a column refresh: https://github.com/growthbook/growthbook/blob/main/packages/front-end/components/FactTables/ColumnList.tsx#L97

lukesonnet (changes requested):

lukesonnet (approved):

PR #3758, opened 2025-03-03
Enabling usage of experimental data in mid-experiment power calculations

lukebrawleysmith:
### Features and Changes

Currently mid-experiment power ignores the data currently collected in the experiment.  The experiment is considered ready to make a shipping decision when the uncertainty is low enough to detect the target MDE.  

This PR permits customers to instead stop the experiment when their experimental data reach statistical significance.  Additionally, days left will be determined by statistical significance, rather than being able to detect the target MDE.  This feature is enabled only if sequential testing is turned on.  

### Testing

Unit tests are included, and I have tested the UI (will test the UI more).


packages/enterprise/src/power.ts
```diff
@@ -234,27 +244,52 @@ export function calculateMidExperimentPowerSingle(
...
-      const s2 = sigmahat2Delta * firstPeriodPairwiseSampleSize;
-      const nTotal =
-        firstPeriodPairwiseSampleSize * (scalingFactorLowPowerWarning + 1);
-      halfwidth = sequentialIntervalHalfwidth(
-        s2,
-        nTotal,
-        params.sequentialTuningParameter,
-        params.alpha / numTests
-      );
+      if (response.deltaPosterior === undefined) {
```
lukesonnet, line 248: Is it worth it to create two different interfaces upstream? One for non-sequential power and one for sequential power? Then we don't need to have all these undefined checks as we will rely on TS to ensure we pass the needed fields for the sequential version.

Or is this risky because the `response` is coming from the stats engine and we can't be sure all of the data exists?

PR #3760, opened 2025-03-04
Added graphs to power calculator

Auz:
Screenshot 2025-03-03 at 8 32 32 PM


jdorn (approved):
Not tested, but the code is isolated and relatively low risk.  If it breaks, it will just break the power calculator front-end.

PR #3762, opened 2025-03-04
Add notifications for Decision Framework

lukesonnet:
### Features and Changes

* Adds new `experiment.decision.ship`, `experiment.decision.rollback`, and `experiment.decision.review` events to notify experimenters when a decision is ready for their experiment.
* Achieves this by moving the logic determining the experiment status to the `shared` package, returning a standardized payload for the current status along with status metadata used in notifications, and then on the front-end using that to render a status, and on the back-end using that to build the event
* Used this shared logic to also update the `experiment.warning` events, because the `decision` only fires if there are no warnings and I wanted that logic to live in one place.

### Testing

Added some unit testing for the new notifications, and tested a variety of statuses changes between experiments and that the old warnings.

Tested in dev that all 3 warning types fire when expected, and only when changing:
Screenshot 2025-03-04 at 7 16 13 AM
Screenshot 2025-03-04 at 7 16 05 AM
Screenshot 2025-03-04 at 7 12 23 AM

And that the warnings still fire as expected:
Screenshot 2025-03-04 at 7 05 43 AM

There is a bit of an edge case that I'm not currently handling [because of the approach I'm using to compare the last status to the current status](https://github.com/growthbook/growthbook/pull/3762/files#diff-b0f54422db1d261a45376c43c437523d8ac0406cb743cb32b2c4cd382bc8f6b1R491-R499), which relies on the current experiment model:
https://www.loom.com/share/529b2b4fd82f4a58a164cb07d7071024


packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -502,13 +504,10 @@ export async function getLatestSnapshotMultipleExperiments(
 
 export async function createExperimentSnapshotModel({
   data,
-  context,
 }: {
   data: ExperimentSnapshotInterface;
-  context: Context;
 }): Promise<ExperimentSnapshotInterface> {
   const created = await ExperimentSnapshotModel.create(data);
-  await notifyExperimentChange({ context, snapshot: created });
```
ahdriel, line 511: Does this mean that we won't send notification for synchronous changes where the snapshot is created in a final format? (Unsure if we even have this though, maybe for manual snapshots?)

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -132,29 +127,20 @@ export const notifyAutoUpdate = ({
...
   experiment,
-  results,
-  snapshot,
+  currentStatus,
 }: {
   context: Context;
   experiment: ExperimentInterface;
-  results: ExperimentReportResultDimension;
-  snapshot: ExperimentSnapshotDocument;
+  currentStatus: RunningExperimentStatusData;
```
ahdriel, line 134: question: So we only call this function for experiments that are in the `running` status?

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -132,29 +127,20 @@ export const notifyAutoUpdate = ({
...
-
-  const triggered = multipleExposureHealth.status === "unhealthy";
+  if (currentStatus.status !== "unhealthy") {
+    return;
+  }
+  const multipleExposureData = currentStatus.unhealthyData.multipleExposures;
+  if (!multipleExposureData) {
+    return;
+  }
+  const triggered = true;
```
ahdriel, line 143: suggestion: Should we update this to check if this is a new status or if the experiment was already unhealthy?

I think that was the original intent of this variable (as in previously triggered).

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -441,37 +411,92 @@ export const notifySignificance = async ({
...
+          return "ship";
+        case "ready-for-review":
+          return "review";
+        case "rollback-now":
+          return "rollback";
+      }
+    })();
+
+    if (currentStatus.status !== lastStatus?.status) {
+      dispatchEvent({
```
ahdriel, line 442: question: why are we not using `memoizeNotification` here?

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -441,37 +411,92 @@ export const notifySignificance = async ({
...
-  if (snapshot.type === undefined) return;
-  // do not fire for snapshots where statistics are manually entered in the UI
-  if (snapshot.manual) return;
-
   await notifySignificance({ context, experiment, snapshot });
 
-  const results = getDefaultAnalysisResults(snapshot);
+  const healthSettings = getHealthSettings(
+    context.org.settings,
+    [...accountFeatures[getEffectiveAccountPlan(context.org)]].includes(
```
ahdriel, line 473: question: can we use `orgHasPremiumFeature` here?

packages/back-end/test/events/experiment.test.ts
```diff
@@ -846,4 +857,249 @@ describe("experiments events", () => {
       },
     });
   });
+
+  it("dispatches decision update when decision to ship", async () => {
```
ahdriel, line 861: Thanks for the tests 🚀

packages/back-end/types/experiment.d.ts
```diff
@@ -36,7 +36,20 @@ export type DecisionFrameworkExperimentRecommendationStatus =
...
+  srm?: boolean;
+  multipleExposures?: { rawDecimal: number; totalUsers: number };
+  lowPowered?: boolean;
+};
+
+export type RunningExperimentStatus =
+  | DecisionFrameworkExperimentRecommendationStatus
+  | { status: "no-data" }
+  | { status: "unhealthy"; unhealthyData: ExperimentUnhealthyData }
+  | { status: "before-min" };
```
ahdriel, line 50: suggestion: Can we make this more explicit? `before-min-duration` maybe?

packages/back-end/src/models/ExperimentSnapshotModel.ts
lukesonnet: I don't think we ever do this in practice. I could probably leave this here because no statuses would change so it would be unlikely to trigger any kind of notification, but I think we should only have one place where the "state of the experiment" is evaluated, including updating the experiment `analysisSummary` and firing any notifications. Having this here is a bit of an odd-man out. If we find out we're missing these notifications, maybe we should also be updating `analysisSummary`?

packages/back-end/types/experiment.d.ts
lukesonnet: Of course, done.

packages/back-end/src/services/experiments.ts
```diff
@@ -2890,7 +2888,25 @@ export async function updateExperimentAnalysisSummary({
...
+  experiment: ExperimentInterface;
+  experimentSnapshot: ExperimentSnapshotInterface;
+}): Promise<ExperimentInterface> {
+  const analysisSummary = await getExperimentAnalysisSummary({
+    context,
+    experiment,
+    experimentSnapshot,
+  });
+
+  return await updateExperiment({
```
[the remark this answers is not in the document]
lukesonnet: Ah, that makes sense, thanks.

packages/back-end/src/services/experiments.ts
```diff
@@ -2844,7 +2842,7 @@ export async function updateExperimentAnalysisSummary({
 
   const srm = getSRMValue(experiment.type ?? "standard", experimentSnapshot);
 
-  if (overallTraffic && srm !== undefined) {
+  if (srm !== undefined) {
```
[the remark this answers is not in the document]
lukesonnet: Yeah, this is kind of arbitrary now, but I didn't want to grapple with changing `.health` to accommodate undefined values...

packages/back-end/src/services/experiments.ts
```diff
@@ -2811,15 +2809,15 @@ export async function getChangesToStartExperiment(
   return changes;
 }
 
-export async function updateExperimentAnalysisSummary({
+export async function getExperimentAnalysisSummary({
```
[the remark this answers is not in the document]
lukesonnet: We are in `computeResultsStatus`

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: Oh wow yeah thank you I was only seeing the FE hooks and forgot about that method.

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -441,37 +411,92 @@ export const notifySignificance = async ({
...
 export const notifyExperimentChange = async ({
   context,
+  experiment,
   snapshot,
+  lastAnalysisSummary,
 }: {
   context: Context;
+  experiment: ExperimentInterface;
   snapshot: ExperimentSnapshotDocument;
+  lastAnalysisSummary?: ExperimentAnalysisSummary;
```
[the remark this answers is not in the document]
lukesonnet: Better, thanks.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: I kind of want this one to always fire if it differs from the last status, even if it has fired for the experiment before.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: I think that's what the memoization does, but now that you mention it, I think I realize that in order for the `memoizeNotification` to remove this unhealthy status if it flicks off, I need to not `return` early.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: Hmm, we call it for all experiments, because `getRunningExperimentStatus` can be computed for stopped experiments as if they were running. Maybe it's a misnomer?

lukesonnet (commented):
responded to all comments. if I didn't immediately do the implied action, I left the comment unresolved.

packages/back-end/src/services/experiments.ts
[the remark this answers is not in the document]
ahdriel: Oh true! Missed that

packages/back-end/src/models/ExperimentSnapshotModel.ts
ahdriel: That's an interesting point -- agreed that we likely need to refactor some stuff to ensure we have a single place where 'new snapshot was created (regardless of how), do what's needed'. Can do that on a later PR.

For `analysisSummary` at the start we required the health query, therefor it would always be async via the QueryRunner.

packages/back-end/src/services/experimentNotifications.ts
ahdriel: Yeah -- to me this seems to be a subset of data that exists only for Experiments that are running. Should we rename it to not include the name of one of our statuses?

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -255,18 +255,20 @@ export async function updateSnapshot({
...
 
     if (experimentModel && isLatestPhase) {
-      await updateExperimentAnalysisSummary({
+      const updatedExperiment = await updateExperimentAnalysisSummary({
         context,
         experiment: experimentModel,
         experimentSnapshot: experimentSnapshotModel,
       });
+
+      await notifyExperimentChange({
```
ahdriel, line 264: question: We are calling this for less snapshots/experiments now as we have moved it inside `if (shouldUpdateExperimentAnalysisSummary) && (experimentModel && isLatestPhase)`.

Just making sure this is intentional, as it can affect other notification and not just the new Decision Framework notifications.

potential issue: This in conjunction with the removal of the checks inside `notifyExperimentChange` means we will not be calling this function for bandit snapshots for example.

lukesonnet: Fewer snapshots:

Maybe? If you look at the code inside `notifyExperimentChange` it kind of does many of the exact same checks and I wanted to put more of the logic in one place instead of having summary update for some snapshots and notifications fire for others.

In truth, the current experiment status and the notifications that are behind it should probably only ever update at the same time. If there are deviations, we want to be very intentional about it.

Bandits:

Why would we no longer be calling this for bandits? Bandits have an `experimentModel` and we only want to fire for any analysis done in the `latestPhase`. They also will be *snapshot* type === "standard"?

packages/back-end/src/models/ExperimentSnapshotModel.ts
lukesonnet: I'm not sure I follow, but do you think I should just add this back in here?

ahdriel: No need to add it back.

packages/back-end/src/models/ExperimentSnapshotModel.ts
ahdriel: Thanks for the explanation!

True about bandit (I got `snapshot.type` mixed up with `experiment.type`), so please disregard.

I think these changes make sense and we should proceed!

ahdriel (approved):
:shipit:

PR #3767, opened 2025-03-05
CUPED ratio bug fixes

lukebrawleysmith:
### Features and Changes
This PR addresses two minor bugs w/r/t CUPED for ratio metrics:
1. we now currently assert that if one test statistic in `BaseABTest` is of type `RegressionAdjustedRatioStatistic`, the other also is
2. when baseline variance was 0, we were not correctly propagating through `RatioStatistics`, but were rather using the `RegressionAdjustedRatioStatistics` in the test (issue only in Bayesian engine).  

Below is example where baseline values are missing and CUPED is turned on: 

### Screenshots
Results without CUPED are showing:
bayesian_standard_after_fix

Results with CUPED were not showing due to `ZERO_NEGATIVE_VARIANCE` error
bayesian_cuped_before_fix

After the fix, results with CUPED are showing and identical to the non-CUPED case:
bayesian_cuped_after_fix

### DIFF
--- a/packages/stats/gbstats/bayesian/tests.py
+++ b/packages/stats/gbstats/bayesian/tests.py
@@ -148,8 +148,6 @@ class EffectBayesianABTest(BayesianABTest):
         config: EffectBayesianConfig = EffectBayesianConfig(),
     ):
         super().__init__(stat_a, stat_b, config)
-        self.stat_a = stat_a
-        self.stat_b = stat_b
         self.config = config
 
     def compute_result(self):
--- a/packages/stats/gbstats/models/tests.py
+++ b/packages/stats/gbstats/models/tests.py
@@ -75,6 +75,16 @@ class BaseABTest(ABC):
                 self.stat_a.theta = theta
                 self.stat_b.theta = theta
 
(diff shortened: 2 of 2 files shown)


packages/stats/gbstats/bayesian/tests.py
```diff
@@ -148,8 +148,6 @@ def __init__(
         config: EffectBayesianConfig = EffectBayesianConfig(),
     ):
         super().__init__(stat_a, stat_b, config)
-        self.stat_a = stat_a
-        self.stat_b = stat_b
```
lukesonnet, line 152: I don't understand why we're deleting this here. Is this just redundant?

lukebrawleysmith: It is redundant, and it causes errors.  We already create `self.stat_a` and `self.stat_b`  in the `super().__init__` method.  The `super().__init__` method  appropriately changes `stat_a` to a `Ratio` statistic if the pre-exposure variance is 0.  We inappropriately overwrite this change here.

lukesonnet (approved):

PR #3768, opened 2025-03-05
Fact Table JSON Columns

jdorn:
### Features and Changes

For columns in a fact table that have JSON-encoded strings, detect nested keys and data types.  Allow using these nested keys for filtering and as metric values.

image

TODO:
- [x] New `json` column data type, auto-detect from sample rows
- [x] Auto-detect nested keys and data types from sample rows
- [x] Allow editing nested keys and data types
- [x] SQL code for each data source to transform dot notation column access (e.g. `context_json.cloud`) into proper SQL
- [x] Ability to use a nested key in an inline filter (stored in dot notation)
- [x] Ability to use a nested key as a metric value column (stored in dot notation)
- [ ] Ability to use a nested key in a user filter (stored in dot notation)

Testing:
- [x] Metric analysis query
- [x] Experiment results query


packages/back-end/types/fact-table.d.ts
```diff
@@ -38,6 +41,7 @@ export interface ColumnInterface {
   alwaysInlineFilter?: boolean;
   topValues?: string[];
   topValuesDate?: Date;
+  jsonFields?: JSONColumnFields;
```
lukesonnet, line 44: nit:
Should `JSONColumnField = Record<string, {dataType: FactTableColumnType}>`? I find the naked use of returning the datatype in various functions like `getColumnExpression` to be a bit confusing. Or the fact that this is a map of datatypes could be clearer in the name.

packages/back-end/src/util/sql.ts
```diff
@@ -209,19 +209,80 @@ export function replaceCountStar(aggregation: string, col: string) {
...
+  const fields: Record<string, FactTableColumnType> = {};
+
+  testValues.forEach((str) => {
+    if (typeof str !== "string") return;
+    try {
+      const obj = JSON.parse(str);
+      Object.keys(obj).forEach((key) => {
+        if (fields[key]) return;
+        if (obj[key] === null || obj[key] === undefined) return;
+        if (Object.keys(fields).length > 50) return;
```
lukesonnet, line 236: Any particular reason for this limit of 50 fields? Just too unwieldy to use effectively? Likely to be garbage?

lukesonnet (commented):
How comfortable are you relying on the `.` notation? Is a more heavy-handed rewrite that stores column-path more explicitly in the Metric model worth it?

packages/back-end/types/fact-table.d.ts
jdorn: yeah, good call. I can see adding more info as well in the future, so an object is better

packages/back-end/src/util/sql.ts
jdorn: Since these fields are stored in the fact table itself and we pass that around everywhere, trying to add some limits to the size of that object.

lukesonnet (approved):

PR #3774, opened 2025-03-05
getting the correct attribute for SampleSize

lukebrawleysmith:
### Features and Changes

Currently inside of the `getMetricSampleSize` method we return the `value` attribute for the sample size.  We want to return the `users` attribute.  This PR addresses this issue.  

### Screenshots

Example where our app currently inappropriately returns a sample size error for having fewer than 150 users, even though the number of users per variation is 200.  
WrongSampleSizes

After the fix, the app appropriately displays results.  
RightSampleSizes

### DIFF
--- a/packages/shared/src/experiments.ts
+++ b/packages/shared/src/experiments.ts
@@ -542,7 +542,7 @@ export function getMetricSampleSize(
         baselineValue: baseline?.stats?.count,
         variationValue: stats?.stats?.count,
       }
-    : { baselineValue: baseline.value, variationValue: stats.value };
+    : { baselineValue: baseline.users, variationValue: stats.users };
 }
 
 export function hasEnoughData(


lukesonnet (changes requested):
In our UI, we actually call this "metric total" so this isn't a bug, this is just a weird formulation of when you might want to say you have "enough data". It actually works reasonably well for Proportion/Binomial metrics.

This fix is essentially introducing a bug, but it's counterintuitive we call it "sample size" under the hood. We should add a "min users" sample size, but it isn't really metric level, it's experiment level and akin to min experiment length.

lukebrawleysmith:
Thanks for explaining.  I agree this approach can be better for Binomial metrics than just checking the sample size, as this can help ensure there are sufficient conversions.  I wll close.

PR #3779, opened 2025-03-06
feature(timeseries): Add new collection + generate timeseries when a new snapshot is created

ahdriel:
### Features and Changes

- Add a new collection `Metric Time Series` which can be used to store a simplified version of `SnapshotMetric` for timeseries.
  - This collection is append-only at the moment but we can create a backfill mechanism if we want
  - In preparation for Safe Rollouts, this is keyed by `source` & `sourceId` as we will want to have TimeSeries that are not tied to a Experiment in the near future
- `ExperimentTimeSeries` handles adding a new datapoint for the time series when a new Snapshot is generated

Two additional changes were needed to support this work:

- Making `auditLog` optional on `BaseModel`
- Returning if notifications were sent by `experimentNotifications`
  - This one feels hacky and it is an area that deserves love & a bigger refactor


packages/back-end/src/validators/experiment-time-series.ts
```diff
@@ -0,0 +1,85 @@
...
+    .object({
+      users: z.number(),
+      count: z.number(),
+      stddev: z.number(),
+      mean: z.number(),
+    })
+    .optional(),
+  pValue: z.number().optional(),
+  pValueAdjusted: z.number().optional(),
+  uplift: z
```
lukesonnet, line 22: I think let's remove `uplift` for now.

lukesonnet (commented):
This seems sensible to me. I don't know how to handle the "limit" problem, but otherwise I like this approach unless there's a big DB cost we uncover.

jdorn:
> 60 snapshots, 5 metrics, and 3 variations (control, a and b) creates a document of ~620kb.

This is too big. We don't want a single document growing beyond 1mb.  It's not uncommon for some experiments to have 50+ metrics.

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -0,0 +1,129 @@
...
+      .map((a) => {
+        return [a.settings.differenceType, a];
+      })
+  );
+
+  return {
+    snapshotId: snapshot.id,
+    analysisDate: snapshot.dateCreated,
+    settingsHash: md5(
+      JSON.stringify({ ...snapshot.settings, metricSettings: [] })
```
bryce-fitzsimons, line 72: do we need this for every row, or just top level?

packages/back-end/src/types/Audit.ts
```diff
@@ -49,6 +49,7 @@ export const entityEvents = {
   factTable: ["autocreate"],
   customField: ["create", "update", "delete"],
   experimentTemplate: ["create", "update", "delete"],
+  experimentTimeSeries: ["create", "update", "delete"],
```
msamper, line 52: We probably don't need to audit these if they're automated. My snapshots base model PR makes auditing optional since we don't want to do it for snapshots either

ahdriel:
@lukesonnet @jdorn I have updated the approach to be per-metric as we discussed. There are still some items to be finished that are described in the PR description.

packages/back-end/src/validators/metric-time-series.ts
```diff
@@ -0,0 +1,72 @@
...
+    id: z.string(),
+    organization: z.string(),
+    dateCreated: z.date(),
+    dateUpdated: z.date(),
+
+    metricId: z.string(),
+    source: z.enum(["experiment"]),
+    sourceId: z.string(),
+
+    lastSettingsHash: z.string(),
```
lukesonnet, line 53: Did we decide to only hash the snapshot and not the metric settings?

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -0,0 +1,98 @@
...
+    (analysis) => analysis.settings.differenceType === "relative"
+  );
+  const absoluteAnalysis = experimentSnapshot.analyses.find(
+    (analysis) => analysis.settings.differenceType === "absolute"
+  );
+  const scaledAnalysis = experimentSnapshot.analyses.find(
+    (analysis) => analysis.settings.differenceType === "scaled"
+  );
+
+  // NB: Using relative as a base, but it should match absolute & scaled
```
lukesonnet, line 34: This is where the `value`, `denominator` and `stats` fields could also come from relative and match absolute and scaled if we want to refactor the data model to make it even leaner.

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  // NB: Explicitly naming all fields to benefit from type safety
+  // when SnapshotMetric and MetricTimeSeriesDataPoint change
+  return {
+    value: metric.value,
+    denominator: metric.denominator,
+    expected: metric.expected,
+    ci: metric.ci,
+    stats: metric.stats,
```
lukesonnet, line 93: This doesn't drop `count` from `stats`, does it?

packages/back-end/src/models/MetricTimeSeriesModel.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    if (updates.length > 0) {
+      await this._dangerousGetCollection().bulkWrite(
+        updates.map((u) => ({
+          updateOne: {
+            filter: { id: u.id },
+            update: {
+              $set: {
+                lastSettingsHash: u.lastSettingsHash,
+                dataPoints: u.dataPoints,
```
lukesonnet, line 95: There is also the `$push` approach to just add to an array rather than re-setting the entire array. You can see that in `addOrUpdateSnapshotAnalysis`. Your approach is probably safer and powerful but IDK how the read/write costs compare with `bulkWrite` like this.

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -405,6 +413,8 @@ export const notifySignificance = async ({
       })
     )
   );
+
+  return emailSent;
```
lukesonnet, line 417: We don't want to just use `emailSent`. I think it just should return `true` at this point since they could get slack notifications even if `isEmailEnabled()` is false. Or it could be `experimentChanges.length` which would be the same as `true` at this point.

However, it might be best if we just included changes where the goal or guardrail metrics change status instead of secondary metrics, but I think that should be the core logic for notifications anyways. My main concern right now is this could end up tagging too many data points if there are 100s of metrics in an experiment.

Our paths forward right now:
(1) ignore this and maybe tag a ton of snapshots if there are many metrics (not ideal)
(2) change it so notifications only fire for goal and guardrails (ehh, don't know if that should happen here, might be part of a broader strategy)
(3) Ignore this notification altogether and instead use the `analysisSummary` and just check if anything in the analysis summary changed significance. I can write this if need be, but basically you have access to the updated `experiment`, and we just need some logic to know if any of the goal or guardrail metrics in the `experimentAnalysisSummaryVariationStatus` changed their status at all.

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -0,0 +1,232 @@
...
+    const denominatorFactTableId = factMetric.denominator?.factTableId;
+    const denominatorFactTable = denominatorFactTableId
+      ? factTableMap?.get(denominatorFactTableId)
+      : undefined;
+
+    const numeratorFilters = numeratorFactTable?.filters.filter((it) =>
+      factMetric.numerator.filters.includes(it.id)
+    );
+
+    // TODO: Some of these have a `updatedate`, should we include it or filter it out?
```
lukesonnet, line 213: I'm not sure which you mean, but if they are on the fact table, I would filter it out to be a bit "riskier" but with fewer false positives.

lukesonnet (commented):
looking good so far, lmk if you want to sync on the notifications comment

packages/back-end/src/services/experimentNotifications.ts
ahdriel: Good callout! I was not sure about how often these changes/notifications would happen.

Is (3) what we are doing with `notifyDecision`? Or something similar? I think for now we can go with 3 and adopt a similar approach and pass the `previousAnalysisSummary` to `updateExperimentTimeSeries` too, if we logic is different.

Otherwise `notifyDecision` is being taken into consideration to tag the datapoint as `triggered-alert`.

lukesonnet: `notifyDecision` will be a subset of the cases when one of the goal or guardrails changes status. I'm basically looking for a middle ground that should capture the cases when a CI crosses 0, which should trigger a `notifyExperimentChanges` notification, but ignores the `notifyExperimentChanges` for secondary metrics.

So I think adapting it is good if not too big of a lift.

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -0,0 +1,294 @@
...
+  // Only update time series for dimensionless snapshots, but if we want to
+  // support dimensions for time series, we should revisit this
+  if (
+    experimentSnapshot.dimension !== null &&
+    experimentSnapshot.dimension !== ""
+  ) {
+    return;
+  }
+
+  const metricsIds = getAllMetricIdsFromExperiment(experiment);
```
lukesonnet, line 54: You can either get them from the snapshot like you mentioned offline, or you can keep using this. If you keep using this, you should turn off `includeActivationMetric`, and I think you need to include `metricGroups` as some IDs can be metric groups instead of individual metrics.

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -0,0 +1,294 @@
...
+
+function getMetricSettingsHash(
+  metricId: string,
+  metricSettings: MetricForSnapshot,
+  factMetrics?: FactMetricInterface[],
+  factTableMap?: Map<string, FactTableInterface>
+): string {
+  const factMetric = factMetrics?.find((metric) => metric.id === metricId);
+  if (!factMetric) {
+    return hashObject(metricSettings);
```
lukesonnet, line 218: We might want to have a fallback here in case the metricSettings object is undefined, as it can be if somehow the metric got deleted but is still in the experiment model.

packages/back-end/src/services/experimentNotifications.ts
ahdriel: Done -- added in `getHasSignificantDifference` function

packages/back-end/src/services/experimentTimeSeries.ts
ahdriel: Doing both -- using experimentSnapshotSettings to get the metrics, but still using the same function + passing in metricGroups + removed activation metric.

packages/back-end/src/validators/metric-time-series.ts
```diff
@@ -0,0 +1,88 @@
...
+    pValue: z.number().optional(),
+    pValueAdjusted: z.number().optional(),
+    chanceToWin: z.number().optional(),
+  })
+  .strict();
+export type MetricTimeSeriesValue = z.infer<typeof metricTimeSeriesValue>;
+
+const metricTimeSeriesVariation = z.object({
+  id: z.string(),
+  name: z.string(),
```
ahdriel, line 18: How much would this save? If the `id` is not the index then we can rely on it and remove `name`, otherwise we need to figure out something else.

packages/back-end/src/validators/metric-time-series.ts
```diff
@@ -0,0 +1,88 @@
+import { z } from "zod";
+
+const metricTimeSeriesValue = z
+  .object({
+    value: z.number(),
```
ahdriel, line 5: How to do this transparently so we store the condensed versions (v, d, e) and save space while keeping the same DX?

packages/back-end/src/models/MetricTimeSeriesModel.ts
```diff
@@ -0,0 +1,230 @@
...
+      }
+      return acc;
+    }, new Map<string, typeof dataPoints>());
+
+    // Order of array is defined by order of insertion, which is the opposite of the desired order
+    const sortedDataPoints = Array.from(lastDataPointPerDay.values())
+      .flat()
+      .reverse();
+
+    if (sortedDataPoints.length <= 300) {
```
ahdriel, line 223: Make it a year?

bryce-fitzsimons (approved):

PR #3781, opened 2025-03-06
Introduce inline cloud starter to pro upgrade

mknowlton89:
### Features and Changes

License server companion PR - https://github.com/growthbook/central-license-server/pull/38

#### General Flow
- Added a FF to the UpgradeModal to support in-line and Stripe Hosted Checkout upgrades
- If in-line upgrade, front end makes a call to the license server which does a few things:
   - Creates a license if one doesn't exist
   - Sets up a stripe Payment Intent which returns a client secret
 - Then, the front-end can use this client secret to render Stripe's `PaymentElement`
 - When the user adds a payment method, we then make a call to Stripe to confirm the payment, using the `client_secret`
 - After the payment is confirmed, we make a call to the license server to create the Orb subscription

### Dependencies

-  https://github.com/growthbook/central-license-server/pull/38

### Testing

- [x] Ensure a cloud org in the starter plan can upgrade to a pro plan successfully
- [x] Ensure that starting the upgrade flow, exiting, and coming back works correctly (ensure it doesn't create multiple stripe customers, nor multiple licenses, but simply reuses the ones created on the initial setup flow).


packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -509,6 +527,14 @@ export default function UpgradeModal({
           header={`🎉 Your 14-day Enterprise Trial starts now!`}
           isTrial={true}
         />
+      ) : showCloudProUpgrade && clientSecret ? (
```
romain-growthbook, line 530: Looks like the logic here could be combined also `clientSecret` is vague. Maybe a single `cloudProUpdateSetup` that is either `undefined`/`null` or `{ clientSecret: string }`.

romain-growthbook (commented):
Some initial comments. Looking good so far.

packages/back-end/src/controllers/stripe.ts
```diff
@@ -88,6 +90,28 @@ export const postNewProTrialSubscription = withLicenseServerErrorHandling(
   }
 );
 
+export const postNewProSubscriptionIntent = withLicenseServerErrorHandling(
```
mknowlton89, line 93: Move to orb.ts?

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -689,11 +689,22 @@ export async function getOrganization(
...
         license =
-          (await licenseInit(org, getUserCodesForOrg, getLicenseMetaData)) ||
-          null;
+          (await licenseInit(
+            org,
+            getUserCodesForOrg,
+            getLicenseMetaData,
+            // if licese is set, but no stripe subscription, force refresh
+            // this can happen when a new subscription is created
+            !license?.stripeSubscription ? true : false
```
mknowlton89, line 706: This still necessary?

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -549,43 +576,22 @@ export default function UpgradeModal({
...
                         <span>You are currently using the </span>
                         <b className="mx-1"> {licensePlanText} </b> version of
-                        Growthbook with{" "}
-                        <Link
-                          href="/settings/team"
-                          className="mx-1 font-weight-bold"
-                        >
-                          {currentUsers} team members
-                        </Link>
-                        ↗
```
mknowlton89, line 559: The way we calculate seatsInUse between Stripe and Orb has diverged, to keep things simple, I've removed this.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -689,11 +689,23 @@ export async function getOrganization(
...
-          null;
+          (await licenseInit(
+            org,
+            getUserCodesForOrg,
+            getLicenseMetaData,
+            // if license is set, but no stripe or orb subscription, force refresh
+            !license?.stripeSubscription || !license?.orbSubscription
+              ? true
+              : false
+          )) || null;
```
mknowlton89, line 708: This is all getting a bit hard to follow. In `CloudProUpgradeModal` when a user successfully subscribes, this code, paired with the `refreshOrganization` call means the org's subscription status updates automatically, which is handy if they're upgrading from an upgrade CTA to get a certain feature (ex: Bandits).

mknowlton89, line 708: This is all getting a bit hard to follow. In `CloudProUpgradeModal` when a user successfully subscribes, this code, paired with the `refreshOrganization` call means the org's subscription status updates automatically, which is handy if they're upgrading from an upgrade CTA to get a certain feature (ex: Bandits).

mknowlton89: Open to suggestions on if there is a better way to handle this - maybe the `redirectWithTimeout` function?

packages/shared/src/enterprise/models/licenseModel.ts
```diff
@@ -25,7 +25,6 @@ const licenseSchema = new mongoose.Schema({
   },
   orbSubscription: {},
   stripeSubscription: {},
-  billingPlatform: String,
```
mknowlton89, line 28: @jdorn Wanted to call this change out to you specifically to make sure this wasn't going to introduce any unwanted side effects.

I've removed this and upgrade the `getSubscriptionFromLicense` function so that we infer the `billingPlatform` based on whether the license has a `stripeSubscription` or `orbSubscription`. I've retained the `billingPlatform` property on the `SubscriptionInfo` type, though, as it is handy on the frontend.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -689,11 +689,23 @@ export async function getOrganization(
   if (licenseKey || process.env.LICENSE_KEY) {
     // automatically set the license data based on org license key
     license = getLicense(licenseKey || process.env.LICENSE_KEY);
-    if (!license || (license.organizationId && license.organizationId !== id)) {
+    if (
+      !license ||
+      (license.organizationId && license.organizationId !== id) ||
+      (license && !license.stripeSubscription) ||
+      (license && !license.orbSubscription)
```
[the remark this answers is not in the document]
tzjames: Also, the license can have neither, if it is an enterprise account.  Or a "free" one after having formerly been enterprise.  I think it needs to be:
```
if(
      !license ||
      (license.organizationId && license.organizationId !== id) ||
      ((license.plan === "pro" || license.plan === "pro_sso")  && !license.stripeSubscription && !license.orbSubscription))
)
```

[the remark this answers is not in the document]
tzjames: Though can it happen even that a license is pro, but doesn't have a subscription.  Doesn't all the license endpoints that create the license ensure that one or the other exists before returning?

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -689,11 +689,23 @@ export async function getOrganization(
...
+    ) {
       try {
         license =
-          (await licenseInit(org, getUserCodesForOrg, getLicenseMetaData)) ||
-          null;
+          (await licenseInit(
+            org,
+            getUserCodesForOrg,
+            getLicenseMetaData,
+            // if license is set, but no stripe or orb subscription, force refresh
```
tzjames, line 704: I think this could just be `!!license // if the license is set but in not a correct state, force refresh`.

packages/back-end/src/routers/organizations/organizations.controller.ts
tzjames: I believe the old stripe controllers call licenseInit with force when making the subscription.  Alternatively if we have the license returned when making a new subscription, which seems reasonable, then we can have it save directly by calling `verifyAndSetServerLicenseData(license)` from within licenseUtil.ts;

mknowlton89: @tzjames I was able to revert these changes now that `postNewInlineSubscriptionToLicenseServer` gets the updated license from the license-server and then calls `verifyAndSetServerLicenseData`. Thank you for that suggestion - that is much cleaner than this mess on the org controller.

PR #3788, opened 2025-03-11
Support multiple billing platforms

jdorn:
### Features and Changes

Update license interface to support both `stripe` and `orb` as billing platforms.


packages/shared/src/enterprise/licenseUtil.ts
```diff
@@ -147,6 +158,19 @@ export interface LicenseInterface {
     discountMessage?: string; // The message of the discount
     hasPaymentMethod?: boolean;
   };
+  orbSubscription?: {
+    id: string;
+    customerId: string;
+    qty: number;
+    trialEnd: Date | null;
+    status: Stripe.Subscription.Status;
```
mknowlton89, line 166: Orb has slightly different statuses (`Orb.Subscription.Status`), which can be "active", "ended", or "upcoming".  Whereas Stripe has `active`, `canceled`, `incomplete`, `incomplete_expired`, `past_due`, `paused`, `trialing`, `unpaid`.

I can map Orb's statuses to Stripes if we want to align on Stripe's status types.
"active" => "active"
"ended" => "cancelled"
"upcoming" => "paused" - we don't have a concept of this today, and don't anticipate we will in the future.

packages/shared/src/enterprise/licenseUtil.ts
```diff
@@ -147,6 +158,19 @@ export interface LicenseInterface {
...
+    current_period_end: number;
+    cancel_at: number | null;
+    canceled_at: number | null;
+    cancel_at_period_end: boolean;
+    planId: string;
+    hasPaymentMethod: boolean;
```
mknowlton89, line 172: We'll probably also want to add `portalUrl` here.

packages/shared/src/enterprise/licenseUtil.ts
jdorn: yeah, let's align on Stripe's statuses

packages/shared/src/enterprise/licenseUtil.ts
```diff
@@ -147,6 +158,19 @@ export interface LicenseInterface {
...
   };
+  orbSubscription?: {
+    id: string;
+    customerId: string;
+    qty: number;
+    trialEnd: Date | null;
+    status: Stripe.Subscription.Status;
+    current_period_end: number;
+    cancel_at: number | null;
+    canceled_at: number | null;
```
mknowlton89, line 169: I'm not sure this is going to be possible to extract from Orb as it isn't stored on their `subscription` object. They do have an endpoint to fetch subscription schedules, but without testing, I'm not confident that would also have the information we want.

mknowlton89: So we can store the `canceled_at` timestamp, but it might get cumbersome to manage the state.

We'll need to update `updateOrbSubscriptionInDb` to persist the state unless an org re-subscribes at a certain point. Not impossible, but we'll have to manage it manually (as opposed to stripeSubscriptions) where they manage it.

packages/shared/src/enterprise/licenseUtil.ts
jdorn: This whole license object is visible to everyone in the org, even readonly users.  Is it safe to expose the portal URL like that?  Or should we use the existing `/subscriptions/manage` endpoint that fetches the URL from the license server after doing permission checks?

mknowlton89: Ah - I see. I forgot we pass the full `license` to the front-end. My initial thought was to only share the portal URL if the user had permission, but that doesn't fix us sharing the full license.

I'm good with updating `subscription/manage`, to handle both stripe and orb subscriptions.

mknowlton89 (approved):

PR #3798, opened 2025-03-12
[api] Add user filter to fact metric API and add validation

### LINKED ISSUE #3783: Add aggregateFilter via API for FactMetric
Hi! How to add `aggregateFilter` (User Filter) via API for FactMetric?

We tried this with bulk import, but it seems it didn’t work. 

Could you please add this possibility?

```yml
- id: level-completed-1-count
  data: 
   name: User with Completed 1 Level
   metricType: proportion
   numerator: 
    factTableId: user-datasource
    aggregateFilter: ">=1"
    aggregateFilterColumn: compleated
```

lukesonnet:
### Features and Changes

* Adds `aggregateFilter` and `aggregateFilterColumn` for Fact Metric Create and Update endpoints.
* Adds additional validation to make API usage safer and adds it to the base model validation for fact metrics.

- Closes #3783 

### Testing

Tested REST API in Hoppscotch
* Fails when not retention or count metric
* Fails when using neither $$count or non-numeric column (or non-existing column)
* Fails when filter is misspecified
* Creates or Updates fact metrics correctly when it should

Screenshot 2025-03-12 at 4 43 27 PM
Screenshot 2025-03-12 at 4 44 29 PM
Screenshot 2025-03-12 at 4 43 51 PM
Screenshot 2025-03-12 at 4 45 14 PM

Screenshot 2025-03-12 at 4 45 19 PM
Screenshot 2025-03-12 at 4 47 35 PM

UI also:
* Still works as expected with additional validation
* Still fails when aggregate filter is incorrect (validation moved from modal to model)

Screenshot 2025-03-12 at 4 58 30 PM


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -26,6 +32,55 @@ const BaseClass = MakeModelClass({
...
+  metricType,
+  numerator,
+  factTable,
+}: {
+  metricType: FactMetricType;
+  numerator: ColumnRef;
+  factTable: FactTableInterface;
+}): void {
+  // error if one is specified but not the other
+  if (!!numerator.aggregateFilter !== !!numerator.aggregateFilterColumn) {
```
jdorn, line 46: Does this break the UI if you have an aggregate filter and then remove it? Does the column value stick around and cause this check to break?

jdorn (approved):

packages/back-end/src/models/FactMetricModel.ts
lukesonnet: Yes, it does break the UI and it also breaks it if you add one then switch metric types.

I'll sanitize the fields on save.

PR #3808, opened 2025-03-13
REST API: add "withRevisions" option to getFeature endpoint

bryce-fitzsimons:
### Features and Changes

add a new optional filter for the `getFeatures` REST API endpoint (api/v1/features/{id})
```
  - in: query
    name: withRevisions
    schema:
      type: string
      enum:
        - all
        - drafts
        - published
        - none
    required: false
    description: Also return feature revisions (all, draft, or published statuses)
```


docs/src/partials/event-webhook/_event-webhook-list.md
```diff
@@ -182,6 +182,78 @@ Triggered when a feature is created
                 date: string;
                 publishedBy: string;
             };
+            revisions?: {
```
bryce-fitzsimons, line 185: Not sure what to do about this. Webhooks will never send the full payload... @romain-growthbook

Kevin-Chant: Which source of truth is autogenerating this, `packages/back-end/src/api/openapi/schemas/Feature.yaml`? Can we do an equivalent to Typescript's `extends` with files like `Feature` and `FeatureWithRevisions` and then point the docs generator at the file without the revisions?

Kevin-Chant (commented):
Looks pretty good; I'd like to see the webhook docs trimmed down before we ship it though

docs/src/partials/event-webhook/_event-webhook-list.md
Kevin-Chant: @bryce-fitzsimons great minds think alike :joy: 
image

bryce-fitzsimons:
> Looks pretty good; I'd like to see the webhook docs trimmed down before we ship it though

Yeah I just tried extracting to Feature + FeatureWithRevisions in hopes that the autogenerated webhooks docs will look better...

docs/src/partials/event-webhook/_event-webhook-list.md
bryce-fitzsimons: Nevermind, ended up extending with a new `FeatureWithRevisions` type

packages/back-end/src/services/features.ts
```diff
@@ -1265,6 +1317,7 @@ export function getApiFeatureObj({
       publishedBy: publishedBy || "",
       version: feature.version,
     },
+    revisions: revisionDefs,
```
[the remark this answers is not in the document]
bryce-fitzsimons: Ah i missed the annotated return type of this method, should be `ApiFeatureWithRevisions`. I think it's ok to double dip since there is shared generation code between the two related types ApiFeature and ApiFeatureWithRevisions.

PR #3815, opened 2025-03-17
Add log_base to shared utils and use in license logger

Kevin-Chant:
### Features and Changes

We added control of the log base used by Pino in #3694, but missed that there's a standalone logger (previously in the Enterprise package, now shared) that wasn't inheriting that logbase.
A customer wrote in asking if we could also fix the logs emitted for "Verifying license data: ..." which seems to be the only usage of a separate Pino instance.

This PR moves the processing out of back-end/secrets (as it's not really a secret anyway) and into shared/utils so it can be imported from anywhere we use Pino

### Testing

Define a `LOG_BASE` in your `.env.local` or similar with example value
```
LOG_BASE={"defaultIncludedLogKey":"defaultValue"}
```
Starting the server should prompt the license server message

### Screenshots

image

### DIFF
--- a/packages/shared/src/enterprise/licenseUtil.ts
+++ b/packages/shared/src/enterprise/licenseUtil.ts
@@ -4,7 +4,7 @@ import type Stripe from "stripe";
 import pino from "pino";
 import { pick, sortBy } from "lodash";
 import AsyncLock from "async-lock";
-import { stringToBoolean } from "shared/util";
+import { parseProcessLogBase, stringToBoolean } from "shared/util";
 import { ProxyAgent } from "proxy-agent";
 import cloneDeep from "lodash/cloneDeep";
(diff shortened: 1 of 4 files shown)


packages/shared/src/enterprise/licenseUtil.ts
```diff
@@ -17,7 +17,11 @@ export const LICENSE_SERVER_URL =
 // mimic behavior in back-end/src/util/secrets.ts
 const APP_ORIGIN = process.env.APP_ORIGIN || "http://localhost:3000";
 
-const logger = pino();
+const logBase = parseProcessLogBase();
+
+const logger = pino({
+  ...logBase,
```
bryce-fitzsimons, line 23: do we need a cloned logBase here?

bryce-fitzsimons (approved):

PR #3824, opened 2025-03-19
Switch to new usage.

romain-growthbook:
This PR switches to new usage.

Questions:
* This uses a mongodb connection specifically for it, maybe we can do better?
* Code location?

### DIFF
--- a/packages/back-end/src/util/mongo.util.ts
+++ b/packages/back-end/src/util/mongo.util.ts
@@ -1,5 +1,5 @@
 import type { Document } from "mongodb";
-import { MongoClient } from "mongodb";
+import { MongoClient, Db } from "mongodb";
 import type { Document as MongooseDocument } from "mongoose";
 import mongoose from "mongoose";
 import { getEffectiveAccountPlan } from "shared/enterprise";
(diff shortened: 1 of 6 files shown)


packages/back-end/src/util/mongo.util.ts
```diff
@@ -157,3 +161,40 @@ export function removeMongooseFields<T>(
...
+
+const backendLimits = ({
+  requests: cdnRequests,
+  bandwidth: cdnBandwidth,
+}: StoredLimits) => ({ cdnRequests, cdnBandwidth });
+
+export const getOrgUsageLimits = async (
+  organization: string
+): Promise<UsageLimits> => {
+  const mongodb = new MongoClient(MONGODB_URI);
```
tzjames, line 178: We can't reuse Mongoose but we could make this a singleton: 
```
let cachedClient: MongoClient | null = null;
let cachedDb: Db | null = null;

export async function getMongoUsageDb(): Promise<Db> {
  if (!cachedClient) {
    cachedClient = new MongoClient(MONGODB_URI);
    await cachedClient.connect();
    cachedDb = cachedClient.db("usage"); // Pre-select the "usage" database
  }
  return cachedDb!;
}

export async function closeMongoUsageDb(): Promise<void> {
  if (cachedClient) {
    await cachedClient.close();
    cachedClient = null;
    cachedDb = null;
  }
}
```

Then we should call `closeMongoClient` in server.ts in SIGINT and SIGTERM for correctness sake.

romain-growthbook:
Closing, superseeded by https://github.com/growthbook/central-license-server/pull/38/files and https://github.com/growthbook/growthbook/pull/3781

PR #3827, opened 2025-03-20
Update stats dependencies

ahdriel:
### Features and Changes

Update dependencies to fix critical CVEs.

The last one (as reported by AWS Inspector) is due to `execa@1.0.0` and fixed on 2.0.0; for us this happens because of `wsrun` which depends on `jest-changed-files` but there are no `wsrun` releases that end up using execa v2.


lukesonnet (approved):
I think safe on the gbstats side. Mostly touches dev dependencies and the one non-dev dependency is a minor update for a minor part of gbstats

PR #3843, opened 2025-03-24
New usage limits

romain-growthbook:


packages/back-end/src/enterprise/billing/index.ts
```diff
@@ -45,13 +46,46 @@ export async function deletePaymentMethodById(
...
+const keyToUsageData: Record<string, OrganizationUsage> = {};
+
+export async function getUsage(organization: string) {
+  const cacheCutOff = new Date();
+  cacheCutOff.setHours(cacheCutOff.getHours() - 1);
+
+  Object.keys(keyToUsageData).forEach((organization) => {
+    if (keyToUsageData[organization]?.cdn.lastUpdated <= cacheCutOff)
+      delete keyToUsageData[organization];
+  });
```
mknowlton89, line 82: What are your thoughts on only invalidating the cache for the org we're building the context for, rather than for all orgs? It might mean the cache has some stale data, but would reduce the amount we'd have to loop through all orgs in the cache. (Genuine question - not a suggestion wrapped in a question). Is knowingly keeping stale data in a cache an anti-pattern?

Something like:
```ts
// Check only the requested organization's cache entry
  if (keyToUsageData[organization]?.cdn.lastUpdated <= cacheCutOff) {
    delete keyToUsageData[organization];
  }
```

romain-growthbook: Could be. I'm concerned with the cache growing too big over time so I think it would be smart to simply cleanup orgs that we haven't seen in a while.

tzjames (changes requested):
I think just querying the usage db directly from growthbook repo is simpler.  Let me know if you disagree.

packages/back-end/src/enterprise/billing/index.ts
jdorn: One downside of this is if the license server goes down.  Ideally in that case, we can continue to use stale values vs throwing errors.  Cleaning up in a loop like this makes that impossible.

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
```diff
@@ -123,6 +124,7 @@ export default function authenticateApiRequestMiddleware(
 
       req.context = new ReqContextClass({
         org,
+        usage: getUsage(org.id),
```
jdorn, line 127: Why not make this lazy load instead of kicking off the promise immediately for every context object?  We only need this data in a few endpoints

romain-growthbook: yeah can do good idea.

packages/back-end/src/enterprise/billing/index.ts
```diff
@@ -45,13 +46,46 @@ export async function deletePaymentMethodById(
...
+
+  if (keyToUsageData[organization]) return keyToUsageData[organization];
+
+  const usage = await getUsageDataFromServer(organization);
```
jdorn, line 91: We need to catch exceptions here and handle them.  Otherwise, this will be an uncaught promise rejection, which can crash the entire Node process.

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
romain-growthbook: I wish you could actually do that with promise.

packages/back-end/src/enterprise/billing/index.ts
```diff
@@ -45,13 +46,54 @@ export async function deletePaymentMethodById(
...
+  });
   return res;
 }
+
+export async function getUsageDataFromServer(
+  organization: string
+): Promise<OrganizationUsage> {
+  const url = `${LICENSE_SERVER_URL}cdn/${organization}/usage`;
+
+  const usage = await callLicenseServer({ url, method: "GET" });
```
jdorn, line 65: This is typed as `any` here and we're just trusting that the response has all of the required properties.  Has the potential to cause runtime errors.

packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
jdorn: I don't think we're ever going to use this data twice in a single request.  So I would just change it to:

```ts
usage: getUsage
```

The places that use it can call the function.  If a request does call this twice, it won't de-dupe, but the in-memory cache layer will take care of that.

packages/back-end/src/enterprise/billing/index.ts
romain-growthbook: Happy to add a `zod` validation or something similar. The whole `shared` stack could also use that kind of run-time check.

packages/back-end/src/enterprise/billing/index.ts
romain-growthbook: Loop removed.

mknowlton89 (approved):
LGTM - I'd like Jeremy/James to give feedback as well, though, given they had most of the feedback.

PR #3845, opened 2025-03-24
Rework bandit SRM to run in SQL

lukesonnet:
### Features and Changes

Moves SRM into the health query SQL for bandits to deal with issue of late-landing data causing some false positives.

This also means SRM no longer comes as part of the results analysis and instead lives in the `health` object of the snapshot rather than the `analysis` object of the snapshot and the `banditEvents` object that lives on the experiment model.

### Testing

- [x] Run bandit sql and compare SRMs
- [x] Ensure old bandit SRM still flows through
- [x] Ensure health query of old experiments is unchanged


packages/back-end/src/util/migrations.ts
```diff
@@ -562,6 +562,16 @@ export function upgradeExperimentDoc(
           range: [0, 1],
         };
       }
+
+      // move bandit SRM to health.srm
+      if (phase.banditEvents) {
+        phase.banditEvents = phase.banditEvents.map((event) => ({
+          ...event,
+          health: {
+            srm: event.banditResult.srm,
```
lukesonnet, line 571: only do this if bandit result has srm

bryce-fitzsimons (approved):
looks reasonable to me

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2118,27 +2173,92 @@ export default abstract class SqlIntegration
...
       )
+      ${
+        banditDates && banditWeightCaseWhen
+          ? `
+        , __unitsByVariationBanditPeriod AS (
+          SELECT
+            d.variation AS variation
+            , d.bandit_period AS bandit_period
+            , ${banditWeightCaseWhen}
+            , COUNT(*) AS units
```
lukebrawleysmith, line 2190: One limitation here is that if there are no units in variation V, the expected vs actual is not computed.  This results in a residual of 0 for the cell corresponding to V, and the resultant test statistic is too small.

lukebrawleysmith (changes requested):
Other than the 0 units in a variation case, it looks good to me.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: We need to compute it for the case when the bandit period total unit count is > 0, right?

lukebrawleysmith (approved):

PR #3847, opened 2025-03-24
Decision Criteria model and creator

lukesonnet:
### Features and Changes

The first step in allowing for customizable decision criteria and improved control and explanation of shipping recommendations at the experiment level. This PR creates an organization-level setting of a default "Decision Criteria" which can be applied to experiments and evaluate them for a shipping recommendation.

Follow-up PRs will add: The ability to select this and modify it at the experiment level. More context explaining the decision reached at the experiment level.

All FE changes should be behind a feature flag that is off by default for all orgs.

Key changes:
* New `DecisionCriteriaModel` and associated validators, controllers, and routers <- currently org level, but later could be extended to project level. Has a series of rules and conditions that can be evaluated. Doesn't do any validation to make sure the rules make sense, but the user type here should be kind of an "experimentation expert" if they end up building their own.
* New DecisionCriteriaModal and DecisionFrameworkSettings components <- the ability on the Organization settings to set your org default decision criteria and to customize your own.
* Revised EDF evaluation logic to use these new settings to render an experiment status; using the "Clear Signals" default shipping as the orgs default shipping criteria.

Design doc: https://www.figma.com/design/AZRYftQm2mn3tWhZ0fXaRd/Analysis-Plan-3.6?node-id=29-10983

### Testing

- [X] Add unit tests for decision criteria logic
- [X] Test saving and editing Decision Criteria
- [x] Check without pro/enterprise
- [x] Check without user permissions to manage
- [x] Check evaluation flows through to front-end

### Screenshots

New Decision Framework Settings
Screenshot 2025-03-27 at 10 35 09 AM
image

Decision Framework table (fixed GB defaults at the top, custom new decision criteria below)
Screenshot 2025-03-27 at 10 35 22 AM

Decision Criteria for GB default (not editable)
Screenshot 2025-03-27 at 10 36 59 AM

Blank state when creating new
Screenshot 2025-03-27 at 9 21 49 AM

<details>

<summary>Screenshots from experiment page</summary>

Only changes are some tweaks in the business logic for which status appears and the tooltips
Screenshot 2025-03-27 at 12 06 14 PM
Screenshot 2025-03-27 at 11 40 53 AM
Screenshot 2025-03-27 at 11 19 02 AM
Screenshot 2025-03-27 at 11 18 28 AM

</details>


packages/front-end/components/DecisionCriteria/DecisionCriteriaModal.tsx
```diff
@@ -0,0 +1,638 @@
...
+                      ))}
+                    </Select>
+                  </Box>
+
+                  <Box style={{ flex: 1 }}>
+                    <Select
+                      size={selectSize}
+                      value={condition.metrics}
+                      setValue={(value) => {
+                        if (value === "guardrails") {
```
ahdriel, line 454: question: Is it intentional we are calling updateCondition twice for `guardrails` or should we have an else?

packages/front-end/components/DecisionCriteria/DecisionCriteriaModal.tsx
```diff
@@ -0,0 +1,638 @@
...
+            className="appbox mb-1 mt-1"
+            direction="column"
+            gap="2"
+            p="2"
+          >
+            <Flex justify="between" align="center" mb="1">
+              <Text weight="bold" size="2">
+                Rule {ruleIndex + 1}
+              </Text>
+              {ruleIndex > 0 && !disabled && (
```
ahdriel, line 411: nit: Should we check for overall length of rules to allow users to remove rule 1 and keep rule 2 for example?

packages/front-end/components/DecisionCriteria/DecisionCriteriaTable.tsx
```diff
@@ -0,0 +1,229 @@
...
+                      }}
+                    >
+                      View
+                    </DropdownMenuItem>
+                  )}
+
+                  <DropdownMenuItem
+                    onClick={() => {
+                      setDefaultCriteriaId(criteria.id);
+                      setDropdownOpen("");
```
ahdriel, line 163: question: Do we need to manually handle the dropdown? I would expect simple open/pick flow to be supported without this kind of manual bookkepping.

packages/back-end/src/enterprise/models/DecisionCriteriaModel.ts
```diff
@@ -0,0 +1,32 @@
...
+    return this.context.permissions.canCreateDecisionCriteria();
+  }
+  protected canRead(): boolean {
+    return true;
+  }
+  protected canUpdate(): boolean {
+    return this.context.permissions.canUpdateDecisionCriteria();
+  }
+  protected canDelete(): boolean {
+    return this.context.permissions.canDeleteDecisionCriteria();
```
ahdriel, line 30: question: We have UI checks that prevent us from removing Decision Criteria that are set to default for the org.

Should we add the same checks for the API?

ahdriel (commented):
On dark mode the table colors look a bit disabled to me... kinda washed out. Maybe we can update some?

Also the table needs a border-radius to match all the other settings on the page.

Screenshot 2025-03-31 at 8 25 04 PM

lukesonnet:
>Also the table needs a border-radius to match all the other settings on the page.

Our gbtable doesn't have this, so I'm just using our regular styling. Do you have an example of a table that does this correctly?

Responded to other comments, left open the ones that were bigger or still unresolved.

packages/front-end/styles/global.scss
```diff
@@ -1009,6 +1009,20 @@ main.main {
   td {
     vertical-align: middle;
   }
+
+  &.decision-criteria-table .dark {
```
ahdriel, line 1013: issue: This style is not being applied because there's no `.dark` element inside the table.

Do we need to update it to be like the rule below (line 1027)? `.dark &.decision-criteria-table {`?

Otherwise I really think these colors work much better! Thanks for the update.

packages/front-end/components/DecisionCriteria/DecisionCriteriaModal.tsx
```diff
@@ -0,0 +1,638 @@
...
+      } catch (error) {
+        console.error("Error creating decision criteria:", error);
+      }
+    }
+    mutate();
+  };
+  // Only render the modal if it's open
+  if (!open) return null;
+
+  const selectSize = "2";
```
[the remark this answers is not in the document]
ahdriel: I would repeat it as it is not something we expect to change often/parameterize.

Similar to how we use `Text size="2"` in lots of places but don't use a const.

Very nit, no need to change if you prefer it this way.

packages/front-end/components/DecisionCriteria/DecisionCriteriaTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  return (
+    <>
+      <table className="appbox table gbtable responsive-table decision-criteria-table">
+        <thead>
+          <tr>
+            <th style={{ width: "120px" }}>Org Default</th>
+            <th className="w-100">Name</th>
+            <th style={{ width: "80px", textAlign: "center" }}>Rules</th>
+            <th style={{ width: "40px" }}></th>{" "}
```
ahdriel, line 50: nit: Is this space intentional? I think we should likely remove it.

ahdriel (approved):
Thanks for the updates!

Not sure about the table radius, from looking into it likely requires a bigger effort to change (not just adding a single css property), so we can leave it for later.

lukebrawleysmith:
This looks good, thanks for creating this!  My only comment/question is, should we preclude customers from creating non-sensical decision criteria (see attached)?  Or warn them?  
Screenshot 2025-04-03 at 9 17 47 AM

lukebrawleysmith (approved):
This looks good to me, I left 1 comment in conversation.

PR #3851, opened 2025-03-25
Embed Orb Customer Portal on /usage and /billing

mknowlton89:
### Features and Changes

This PR embeds the Orb Customer Portal for the MVP of usage based billing. It's embedded both on `/settings/usage` and `/settings/billing`.

License server companion PR - https://github.com/growthbook/central-license-server/pull/49

### Dependencies

N/A

### Testing
/billing page tests
- [x] Ensure that an org with an active orb subscription can see their customer portal
- [x] Ensure that an org with a stripe subscription doesn't see the portal or any errors
- [x] Ensure that an org without any subscription doesn't see the portal or any errors
- [x] Ensuer that an org with an enterprise subscription doesn't see the portal or any errors

/usage page tests
- [x] Ensure that an org with an active orb subscription can see their customer portal (and thus, their usage)
- [x] Ensure that an org with a stripe subscription sees our Clickhouse usage component, and no Orb usage or errors
- [x] Ensure that an org without any subscirption sees our Clickhouse usage component, and no Orb usage or errors
- [x] Ensure that an org with an enterprise subscription sees our Clickhouse usage component, and no Orb usage or errors


packages/front-end/pages/settings/billing.tsx
```diff
@@ -105,6 +103,17 @@ const BillingPage: FC = () => {
...
+        <>
+          <PaymentInfo />
+          {organization.id ? (
+            <div className="p-3 app-box border">
+              <h3>Invoices & Usage</h3>
+              <OrbPortal orgId={organization.id} />
+            </div>
+          ) : null}
+        </>
+      ) : null}
```
mknowlton89, line 116: This logic makes it so orgs with in-active orb subscriptions will see the portal - this way they can get old invoices if needed. This means it will also show usage (which will be empty). Open to feedback on this, as this could be confusing if an org saw it, but got a message that they're over their included usage in a free plan. 

The `/usage` route will display the correct Clickhouse-based usage data. Given that embedding the portal is an MVP and we'll eventually build our own direct integration, I think this is ok, but open to feedback.

packages/front-end/pages/settings/billing.tsx
```diff
@@ -105,6 +103,12 @@ const BillingPage: FC = () => {
           </p>
         )}
       </div>
+      {subscription?.status ? (
+        <>
+          <PaymentInfo />
+          <OrbPortal />
```
jdorn, line 109: We should only render this if the billing platform is Orb

packages/back-end/package.json
```diff
@@ -91,6 +91,7 @@
     "nunjucks": "^3.2.4",
     "openai": "^3.2.1",
     "openid-client": "^5.1.10",
+    "orb-billing": "^4.71.1",
```
jdorn, line 94: Do this in the license server instead

tzjames (approved):
LGTM

PR #3853, opened 2025-03-25
Enrich metrics with project names for project search

lukesonnet:
### Features and Changes

Previously the `project:` advanced filter would only look for the project id which was not useful. Now we enrich the metrics with project name and search on that (is this too slow???)

### Testing

Tested on local and it works as advertised now

### DIFF
--- a/packages/front-end/components/Metrics/MetricsList.tsx
+++ b/packages/front-end/components/Metrics/MetricsList.tsx
@@ -224,6 +224,7 @@ const MetricsList = (): React.ReactElement => {
   const {
     getDatasourceById,
     mutateDefinitions,
+    getProjectById,
     project,
     ready,
   } = useDefinitions();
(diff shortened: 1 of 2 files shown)


bryce-fitzsimons (approved):

packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -323,7 +325,7 @@ const MetricsList = (): React.ReactElement => {
         return item.type;
       },
       tag: (item) => item.tags,
-      project: (item) => item.projects,
+      project: (item) => item.projectNames,
```
jdorn, line 328: Should we concat these arrays so you can search by either id or name?  Main use case for searching by id is constructing a direct link to a filtered view like `?q=project:prj_abc123`.  Much more natural to use ids for this

jdorn (approved):

PR #3857, opened 2025-03-25
Update PHP docs

ahdriel:
### Features and Changes

As we have updated the PHP SDK and added support for new features, we need to update docs and snippets here to reflect the latest version.


docs/docs/lib/php.mdx
```diff
@@ -34,7 +34,7 @@ $growthbook = Growthbook\Growthbook::create()
 
 // Load feature flags from the GrowthBook API
 // Make sure to use caching in production! (see 'Loading Features' below)
-$growthbook->loadFeatures("sdk-abc123", "https://cdn.growthbook.io");
+$growthbook->initialize("sdk-abc123", "https://cdn.growthbook.io");
```
jdorn, line 37: Should we add a callout so existing users aren't confused?
> Note: Prior to version X, this method was called `loadFeatures` instead of `initialize`

jdorn (approved):

docs/docs/lib/php.mdx
ahdriel: Agreed! Updated

PR #3861, opened 2025-03-26
Expose more fields in the various REST API experiment endpoints

bryce-fitzsimons:
add support for:
- sequentialTestingEnabled
- sequentialTestingTuningParameter
- activationMetric
- segmentId
- queryFilter

fix partially broken fields:
- disabledStickyBucketing
- bucketVersion
- minBucketVersion

disambiguate `id` from `trackingKey` (not always the same!)


packages/back-end/src/api/openapi/payload-schemas/PostExperimentPayload.yaml
```diff
@@ -41,6 +41,15 @@ properties:
     type: array
     items:
       type: string
+  activationMetric:
+    type: string
+    description: Users must convert on this metric before being included
+  segmentId:
```
lukesonnet, line 47: We don't use `id` in the name of any other field (e.g. `activationMetricId`). Can we just use the same naming and call it `ID` in the description?

packages/back-end/src/api/openapi/payload-schemas/PostExperimentPayload.yaml
```diff
@@ -201,6 +210,12 @@ properties:
   regressionAdjustmentEnabled:
     type: boolean
     description: Controls whether regression adjustment (CUPED) is enabled for experiment analyses
+  sequentialTestingEnabled:
```
lukesonnet, line 213: I wish these were all in "settings" just like the return payload :/ 

not worth changing now...

packages/back-end/src/api/openapi/payload-schemas/PostExperimentPayload.yaml
```diff
@@ -201,6 +210,12 @@ properties:
...
+    type: boolean
+    description: Only applicable to frequentist analyses
+  sequentialTestingTuningParameter:
+    type: number
+    description: 5000 is default
```
lukesonnet, line 218: Technically the org default is the default? IDK if that's a distinction worth making on this niche setting.

packages/back-end/src/services/experiments.ts
```diff
@@ -2414,11 +2430,11 @@ export function postExperimentApiPayloadToInterface(
     hypothesis: payload.hypothesis || "",
     goalMetrics: payload.metrics || [],
     secondaryMetrics: payload.secondaryMetrics || [],
-    metricOverrides: [],
     guardrailMetrics: payload.guardrailMetrics || [],
-    activationMetric: "",
-    segment: "",
-    queryFilter: "",
+    activationMetric: payload.activationMetric || "",
```
lukesonnet, line 2434: Seems crazy we don't do any validation here (just like we don't for `goalMetrics`).

lukesonnet (approved):

packages/back-end/src/api/openapi/payload-schemas/PostExperimentPayload.yaml
bryce-fitzsimons: it's already `segmentId` in the response object (settings). I don't want to deviate from that

lukesonnet: Yep, that takes precedence. I didn't notice that, ty.

packages/back-end/src/services/experiments.ts
bryce-fitzsimons: This whole controller needs a lot of love. But for now, just trying to make it more useful.

lukesonnet: yep! i agree hence the ✅

PR #3862, opened 2025-03-26
Re-enable of Move license server code from shared to backend

Kevin-Chant:
Reverts growthbook/growthbook#3860 to reinstate #3848

The root issue with the broken import was fixed in #3863 so it should be safe to revert the revert now

### Testing
Run the following commands to check that the import now works

```
yarn workspace back-end build
yarn workspace back-end start:with-tracing
```

compare to the commit without the fix and observe the import error on start
```
git checkout 0897728714e8f00d1f19ab97121f9f6ed51abc29
yarn workspace back-end build
yarn workspace back-end start:with-tracing
```


packages/back-end/src/tracing.ts
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import "./init/aliases";
```
Kevin-Chant, line 16: This is the missing line that caused the import errors. It seems like this is the only place it's currently missing (the only entrypoint other than server.ts) so this issue shouldn't happen again unless we add new scripts

romain-growthbook (approved):
If the only change compared to the previous one is the addition of the import then this should be good to go!

Kevin-Chant:
> If the only change compared to the previous one is the addition of the import then this should be good to go!

Yep, and Jeremy suggested moving it to its own PR (#3863) so now this is a true revert with no other changes

PR #3864, opened 2025-03-27
support bandit experiments in REST API, more validation

bryce-fitzsimons:
Create or modify Bandits via REST:
- Create and Update both use the same validation as our controller / services
- Added additional standard validation for all experiements (standard and bandit)

Fix/support prerequisites on features (both top-level and rule) and experiments.

image
image
image


packages/back-end/src/api/openapi/payload-schemas/UpdateExperimentPayload.yaml
```diff
@@ -10,6 +10,11 @@ properties:
   name:
     type: string
     description: Name of the experiment
+  type:
+    type: string
+    enum:
+      - "standard"
+      - "multi-armed-bandit"
```
Kevin-Chant, line 17: I don't see any validation in the controller for this, but we only allow conversion of experiments in draft state, right?

Kevin-Chant (changes requested):
One check about `type` in the update payload, and need to update tests to add `prerequisites` to the expected arguments in a couple places

packages/back-end/src/api/openapi/payload-schemas/UpdateExperimentPayload.yaml
bryce-fitzsimons: good catch, fixed

Kevin-Chant (approved):
LGTM once test suite is passing

PR #3867, opened 2025-03-27
fix(timeseries): Add sourcePhase to MetricTimeSeries collection

ahdriel:
### Features and Changes

When refactoring I forgot to add `phase` to the time series which are tied to an Experiment, which is required.
So this PR adds the `sourcePhase` field and uses it.

Another change that is required is the migration to delete the old index if it exists, because the old index is unique, we need to drop it otherwise we cannot create new records with different sourcePhase values.


Kevin-Chant (approved):
LGTM

PR #3878, opened 2025-03-31
Fix usage graphs for previous months

jdorn:
### Features and Changes

Fix CDN usage graphs for date edge cases (e.g. looking at Feb numbers when it's current Mar >28).

### DIFF
--- a/packages/back-end/src/controllers/subscription.ts
+++ b/packages/back-end/src/controllers/subscription.ts
@@ -412,15 +412,15 @@ export async function getUsage(
 
   // Beginning of the month
   const start = new Date();
-  start.setMonth(start.getMonth() - monthsAgo);
-  start.setDate(1);
-  start.setHours(0, 0, 0, 0);
+  start.setUTCDate(1);
+  start.setUTCHours(0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


packages/back-end/src/controllers/subscription.ts
```diff
@@ -412,15 +412,15 @@ export async function getUsage(
...
+  start.setUTCHours(0, 0, 0, 0);
+  start.setUTCMonth(start.getUTCMonth() - monthsAgo);
 
   // End of the month
-  const end = new Date();
-  end.setMonth(end.getMonth() - monthsAgo + 1);
-  end.setDate(0);
-  end.setHours(23, 59, 59, 999);
+  const end = new Date(start);
+  end.setUTCMonth(end.getUTCMonth() + 1);
```
romain-growthbook, line 421: This won't work if the month has `31` days and the following only `30`.. Like today!

```javascript
> d = new Date
2025-03-31T17:40:03.439Z
> d.setUTCMonth(d.getUTCMonth() + 1);
1746121203439
> d
2025-05-01T17:40:03.439Z
```

romain-growthbook (approved):

PR #3883, opened 2025-03-31
Mk/hide limit line

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Settings/Usage/CloudUsage.tsx
+++ b/packages/front-end/components/Settings/Usage/CloudUsage.tsx
@@ -194,9 +194,9 @@ export default function CloudUsage() {
             start={startDate}
             end={endDate}
             limitLine={
-              hasSubscription && limits.cdnRequests !== "unlimited"
-                ? limits.cdnRequests
-                : null
+              hasSubscription || limits.cdnRequests === "unlimited"
+                ? null
+                : limits.cdnRequests
             }
           />
         </Box>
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Settings/Usage/CloudUsage.tsx
```diff
@@ -127,6 +129,11 @@ export default function CloudUsage() {
   }
   const maxMonthsAgo = monthOptions.length - 1;
 
+  // Orgs with non-canceled subscriptions don't have a limit
+  const hasSubscription = ["active", "trialing", "past_due"].includes(
```
tzjames, line 133: Elsewhere we do:
```
  const isAtLeastPro = ["pro", "pro_sso", "enterprise"].includes(
    effectiveAccountPlan || ""
  );
```
which also captures enterprise and is perhaps more robust to new subscription states (but I guess not new plan types).  We do that in like 5 different places in the code now, so we could make it a function that we do in one place so that the logic resides just there.

The effectiveAccountPlan will be starter if it is pro and the subscription is not one of those three above, so effectively the code above is the same thing as what you have, as enterprise should be taken care of with `limits.cdnRequests === "unlimited"`, however I think `isAtLeastPro` is the safe terminology here.

tzjames (approved):
Elsewhere we do:
```
  const isAtLeastPro = ["pro", "pro_sso", "enterprise"].includes(
    effectiveAccountPlan || ""
  );
```
which ends up including

romain-growthbook:
Fixed in https://github.com/growthbook/growthbook/pull/3884

PR #3893, opened 2025-04-02
Adds IS_CLOUD check to getUsage

mknowlton89:
### Features and Changes

Adds `IS_CLOUD` check to the `getUsage` controller method and returns an error if the visitor isn't a cloud org.

Opted to do it this way, instead of updating app.ts and including the endpoint within the `IS_CLOUD` check, as this way we can return a 400 error with a message, as opposed to a 404 error which could get confusing.

If anyone feels strongly, I can update the return type of the endpoint so as to return a 200 with a message, so we don't have an error, but I feel like this is probably fine and makes the code clean on the front end.

### Dependencies

None

### Testing

- [x] Ensure that if an org isn't a cloud org, we don't actually hit the license server from the `getUsage` endpoint, but the controller function throws an error.

### DIFF
--- a/packages/back-end/src/controllers/subscription.ts
+++ b/packages/back-end/src/controllers/subscription.ts
@@ -34,6 +34,7 @@ import {
   updateDefaultPaymentMethod,
   getPaymentMethodsByLicenseKey,
 } from "back-end/src/enterprise/billing/index";
+import { IS_CLOUD } from "back-end/src/util/secrets";
 
 function withLicenseServerErrorHandling<T>(
   fn: (req: AuthRequest<T>, res: Response) => Promise<void>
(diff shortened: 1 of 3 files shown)


packages/back-end/src/controllers/subscription.ts
```diff
@@ -397,6 +398,9 @@ export async function getUsage(
   req: AuthRequest<unknown, unknown, { monthsAgo?: number }>,
   res: Response<{ status: 200; cdnUsage: DailyUsage[]; limits: UsageLimits }>
 ) {
+  if (!IS_CLOUD) {
```
romain-growthbook, line 401: Why not disabling the endpoint altogether in `src/app.ts`:
```typescript
if (!IS_CLOUD)
  app.get("/billing/usage", subscriptionController.getUsage);
```

mknowlton89: If we go that way, which is a completely valid way to go, anytime a user hits `/settings/usage` from their browser, in their console they'll get a 404: Route not found error. Adding it to the controller method allows us to return a 400 error with a message that is a bit more accurate. 

I feel like the 404 errors could make the logs more confusing, compared to the 400 error with the message.

The frontend is uses the `useApi` hook, so we can't do an `IS_CLOUD` check and then conditionally call it due to the rules around hooks.

romain-growthbook (approved):

PR #3897, opened 2025-04-03
Fix table query header opacity

lukesonnet:
Table header on query results was showing through. Couldn't fix it with z-index so just set header BG to be opaque instead.

https://www.loom.com/share/6e3d53aa2bd9439090bf218d17da1647

After
Screenshot 2025-04-03 at 8 35 05 AM

Screenshot 2025-04-03 at 8 34 53 AM

Before
Screenshot 2025-04-03 at 8 32 19 AM

Screenshot 2025-04-03 at 8 32 29 AM


mknowlton89 (approved):
LGTM. Left 1 small nit comment.

I spent ~10 minutes trying to see if I could make it work without having to define the new global className, but couldn't come up with anything better.

PR #3898, opened 2025-04-03
Move visual editor / redirect paywall to publish step

jdorn:
### Features and Changes

Instead of blocking free accounts from creating Visual Editor or URL Redirect experiments entirely, allow creating drafts and only block when trying to start the experiment.

This will let free users try these premium features before deciding whether or not to upgrade.

TODO:
- [ ] Testing
- [ ] Add screenshots to PR
- [ ] Document new `PremiumCallout` component in the design system

### DIFF
--- a/packages/back-end/src/api/sdk-connections/validations.ts
+++ b/packages/back-end/src/api/sdk-connections/validations.ts
@@ -26,10 +26,10 @@ type CapabilitiesParams = { [k in CapabilitiesParamKey]?: boolean };
 
 const premiumFeatures = [
   ["encrypt-features-endpoint", "encryptPayload"],
-  ["visual-editor", "includeVisualExperiments"],
+  // ["visual-editor", "includeVisualExperiments"],
   ["hash-secure-attributes", "hashSecureAttributes"],
   ["remote-evaluation", "remoteEvalEnabled"],
-  ["redirects", "includeRedirectExperiments"],
+  // ["redirects", "includeRedirectExperiments"],
   ["cloud-proxy", "proxyEnabled"],
 ] as const;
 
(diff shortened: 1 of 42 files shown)


packages/back-end/src/api/sdk-connections/validations.ts
```diff
@@ -26,10 +26,10 @@ type CapabilitiesParams = { [k in CapabilitiesParamKey]?: boolean };
 
 const premiumFeatures = [
   ["encrypt-features-endpoint", "encryptPayload"],
-  ["visual-editor", "includeVisualExperiments"],
+  // ["visual-editor", "includeVisualExperiments"],
```
mknowlton89, line 29: I think these can be cleaned up?

packages/front-end/components/Experiment/TabbedPage/StartExperimentModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+        ) : null}
+
+        {needsVisualEditorUpgrade ? (
+          <PremiumCallout
+            commercialFeature="visual-editor"
+            id="start-experiment-modal"
+            mb="3"
+          >
+            This experiment contains visual editor changes, which require a paid
+            plan.
```
mknowlton89, line 97: I'm wondering if it would add value to add a doc link at the end here that links to a doc article about how to preview visual editor / url redirect experiments without a pro/enterprise plan.

The idea being the first visual editor / url redirect experiment the org creates could be a test experiment that they launch to internal users in order to test out the functionality. When instead they should be keeping it in a draft and using the Dev Tools to preview it.

mknowlton89 (approved):

packages/front-end/components/Experiment/TabbedPage/StartExperimentModal.tsx
jdorn: The visual changes section has preview links, but they are easy to miss.  I think this is a general problem, not specific to this change.  It's not super clear on a draft experiment how to preview and QA the experiment before starting.

image

PR #3899, opened 2025-04-03
Auto-compute dimension slices

lukesonnet:
### Features and Changes

If pre-defined dimension slices are available, query by grouping by all of them and make pre-computed dimension results available.

Follow-up PRs:
* Use dimension slices for post-stratification
* Better distinction between stale and up to date dimension values
* Automatic heterogeneous treatment effect detection
* Auto-refresh dimension slices
* Better dimension management via the Dimensions page

### Testing

- [X] Re-aggregates results
- [X] Computes pre-defined dimensions on the same snapshot
- [] Works with Quantile metrics
- [] Works when no pre-defined dimensions are created
- [] Works when pre-defined dimension slices are empty

### Screenshots

TODO

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -1238,9 +1238,14 @@ export default abstract class SqlIntegration
           };
         }
 
+        const dimensionData: Record<string, string> = {};
+        Object.entries(row).filter(([key, _]) => key.startsWith("dim")).map(([key, value]) => {
+          dimensionData[key] = value;
+        });
+
         return {
           variation: row.variation ?? "",
-          dimension: row.dimension || "",
+          ...dimensionData,
           users: parseInt(row.users) || 0,
           count: parseInt(row.users) || 0,
           ...metricData,
@@ -1263,9 +1268,13 @@ export default abstract class SqlIntegration
     const { rows, statistics } = await this.runQuery(query, setExternalId);
     return {
       rows: rows.map((row) => {
+        const dimensionData: Record<string, string> = {};
+        Object.entries(row).filter(([key, _]) => key.startsWith("dim")).map(([key, value]) => {
+          dimensionData[key] = value;
+        });
         return {
           variation: row.variation ?? "",
-          dimension: row.dimension || "",
+          ...dimensionData,
           users: parseInt(row.users) || 0,
           count: parseInt(row.users) || 0,
           main_sum: parseFloat(row.main_sum) || 0,
@@ -1913,6 +1922,7 @@ export default abstract class SqlIntegration
             return `, e.${d.id} AS dim_${d.id}`;
           })
           .join("\n")}
+        , 'all' as dimension
       FROM
           __rawExperiment e
       WHERE
@@ -1981,6 +1991,7 @@ export default abstract class SqlIntegration
           "'__multiple__'",
           "max(e.variation)"
         )} AS variation
+        , MAX(e.dimension) AS dimension
         , MIN(${timestampColumn}) AS first_exposure_timestamp
         ${unitDimensions
           .map(
@@ -2603,36 +2614,19 @@ export default abstract class SqlIntegration
       maxHoursToConvert
     );
 
-    if (params.dimensions.length > 1) {
-      throw new Error(
-        "Multiple dimensions not supported in metric analysis yet. Please contact GrowthBook."
-      );
-    }
-    const dimension = params.dimensions[0];
-    let dimensionCol = this.castToString("''");
-    if (dimension?.type === "experiment") {
-      dimensionCol = `dim_exp_${dimension.id}`;
-    } else if (dimension?.type === "user") {
-      dimensionCol = `dim_unit_${dimension.dimension.id}`;
-    } else if (dimension?.type === "date") {
-      dimensionCol = `${this.formatDate(
-        this.dateTrunc("first_exposure_timestamp")
-      )}`;
-    } else if (dimension?.type === "activation") {
-      dimensionCol = this.ifElse(
-        `first_activation_timestamp IS NULL`,
-        "'Not Activated'",
-        "'Activated'"
-      );
-    }
 
+    // ensure always at least length 1 with plain dim
+    const dimensionCols = params.dimensions.length > 0 ? params.dimensions.map((d) => this.getDimensionCol(d)) : ["dimension"];
+
+    // auto compute activation splits
     const timestampColumn =
-      activationMetric && dimension?.type !== "activation"
+      activationMetric && !!params.dimensions.find((d) => d.type === "activation")
         ? "first_activation_timestamp"
         : "first_exposure_timestamp";
 
     const distinctUsersWhere: string[] = [];
-    if (activationMetric && dimension?.type !== "activation") {
+    // TODO auto compute activation splits
+    if (activationMetric && params.dimensions.map((d) => this.getDimensionCol(d))) {
       distinctUsersWhere.push("first_activation_timestamp IS NOT NULL");
     }
     if (settings.skipPartialData) {
@@ -2696,11 +2690,11 @@ export default abstract class SqlIntegration
       }
       __distinctUsers AS (
         SELECT
-          ${baseIdType},
-          ${dimensionCol} AS dimension,
-          variation,
-          ${timestampColumn} AS timestamp,
-          ${this.dateTrunc("first_exposure_timestamp")} AS first_exposure_date
+          ${baseIdType}
+          , ${dimensionCols.join(", ")}
+          , variation
+          , ${timestampColumn} AS timestamp
+          , ${this.dateTrunc("first_exposure_timestamp")} AS first_exposure_date
           ${banditDates?.length ? this.getBanditCaseWhen(banditDates) : ""}
           ${
             raMetricSettings.length > 0
@@ -2760,15 +2754,15 @@ export default abstract class SqlIntegration
       }
       , __userMetricJoin as (
         SELECT
-          d.variation AS variation,
-          d.dimension AS dimension,
-          ${banditDates?.length ? `d.bandit_period AS bandit_period,` : ""}
-          ${cumulativeDate ? `dr.day AS day,` : ""}
-          d.${baseIdType} AS ${baseIdType},
+          d.variation AS variation
+          , ${dimensionCols.map((c) => `d.${c} AS ${c}`).join(", ")}
+          ${banditDates?.length ? `, d.bandit_period` : ""}
+          ${cumulativeDate ? `, dr.day` : ""}
+          , d.${baseIdType} AS ${baseIdType}
           ${metricData
             .map(
               (data) =>
-                `${this.addCaseWhenTimeFilter(
+                `, ${this.addCaseWhenTimeFilter(
                   `m.${data.alias}_value`,
                   data.metric,
                   data.overrideConversionWindows,
@@ -2806,12 +2800,13 @@ export default abstract class SqlIntegration
         }
       )
       ${
+        // TODO put quantiles in their own query
         eventQuantileData.length
           ? `
         , __eventQuantileMetric AS (
           SELECT
           m.variation
-          , m.dimension
+          , ${dimensionCols.map((c) => `m.${c} AS ${c}`).join(", ")}
           ${eventQuantileData
             .map((data) =>
               this.getQuantileGridColumns(
@@ -2824,22 +2819,22 @@ export default abstract class SqlIntegration
           __userMetricJoin m
         GROUP BY
           m.variation
-          , m.dimension
+          , ${dimensionCols.map((c) => `m.${c}`).join(", ")}
         )`
           : ""
       }
       , __userMetricAgg as (
         -- Add in the aggregate metric value for each user
         SELECT
-          umj.variation,
-          umj.dimension,
-          ${banditDates?.length ? `umj.bandit_period AS bandit_period,` : ""}
-          ${cumulativeDate ? "umj.day," : ""}
-          umj.${baseIdType},
+          umj.variation
+          , ${dimensionCols.map((c) => `umj.${c} AS ${c}`).join(", ")}
+          ${banditDates?.length ? `, umj.bandit_period` : ""}
+          ${cumulativeDate ? `, umj.day` : ""}
+          , umj.${baseIdType}
           ${metricData
             .map(
               (data) =>
-                `${this.getAggregateMetricColumn({
+                `, ${this.getAggregateMetricColumn({
                   metric: data.metric,
                   useDenominator: false,
                   valueColumn: `umj.${data.alias}_value`,
@@ -2873,11 +2868,11 @@ export default abstract class SqlIntegration
             : ""
         }
         GROUP BY
-          umj.variation,
-          umj.dimension,
-          ${cumulativeDate ? "umj.day," : ""}
-          ${banditDates?.length ? `umj.bandit_period,` : ""}
-          umj.${baseIdType}
+          umj.variation 
+          , ${dimensionCols.map((c) => `umj.${c}`).join(", ")}
+          ${cumulativeDate ? `, umj.day` : ""}
+          ${banditDates?.length ? `, umj.bandit_period` : ""}
+          , umj.${baseIdType}
       )
       ${
         percentileData.length > 0
@@ -2892,11 +2887,11 @@ export default abstract class SqlIntegration
         regressionAdjustedMetrics.length > 0
           ? `
         , __userCovariateMetric as (
-          SELECT
-            d.variation AS variation,
-            d.dimension AS dimension,
-            d.${baseIdType} AS ${baseIdType},
-            ${regressionAdjustedMetrics
+          SELECT 
+            d.variation AS variation
+            , ${dimensionCols.map((c) => `d.${c} AS ${c}`).join(", ")}
+            , d.${baseIdType} AS ${baseIdType}
+            , ${regressionAdjustedMetrics
               .map(
                 (metric) =>
                   `${this.getAggregateMetricColumn({
@@ -2932,9 +2927,9 @@ export default abstract class SqlIntegration
             m.timestamp >= d.min_preexposure_start
             AND m.timestamp < d.max_preexposure_end
           GROUP BY
-            d.variation,
-            d.dimension,
-            d.${baseIdType}
+            d.variation
+            , ${dimensionCols.map((c) => `d.${c}`).join(", ")}
+            , d.${baseIdType}
         )
         `
           : ""
@@ -2951,21 +2946,19 @@ export default abstract class SqlIntegration
           : `
       -- One row per variation/dimension with aggregations
       SELECT
-        m.variation AS variation,
-        ${
-          cumulativeDate ? `${this.formatDate("m.day")}` : "m.dimension"
-        } AS dimension,
-        COUNT(*) AS users,
+        m.variation AS variation
+        , ${dimensionCols.map((c) => `m.${c} AS ${c}`).join(", ")}
+        , COUNT(*) AS users
         ${metricData.map((data) => {
           return `
-           '${data.id}' as ${data.alias}_id,
+           , '${data.id}' as ${data.alias}_id
             ${
               data.isPercentileCapped
-                ? `MAX(COALESCE(cap.${data.alias}_value_cap, 0)) as ${data.alias}_main_cap_value,`
+                ? `, MAX(COALESCE(cap.${data.alias}_value_cap, 0)) as ${data.alias}_main_cap_value`
                 : ""
             }
-            SUM(${data.capCoalesceMetric}) AS ${data.alias}_main_sum,
-            SUM(POWER(${data.capCoalesceMetric}, 2)) AS ${
+            , SUM(${data.capCoalesceMetric}) AS ${data.alias}_main_sum
+            , SUM(POWER(${data.capCoalesceMetric}, 2)) AS ${
             data.alias
           }_main_sum_squares
             ${
@@ -3003,41 +2996,42 @@ export default abstract class SqlIntegration
             }
             ${
               data.ratioMetric
-                ? `,
+                ? `
                 ${
                   data.isPercentileCapped
-                    ? `MAX(COALESCE(cap.${data.alias}_denominator_cap, 0)) as ${data.alias}_denominator_cap_value,`
+                    ? `, MAX(COALESCE(cap.${data.alias}_denominator_cap, 0)) as ${data.alias}_denominator_cap_value`
                     : ""
                 }
-                SUM(${data.capCoalesceDenominator}) AS 
-                  ${data.alias}_denominator_sum,
-                SUM(POWER(${data.capCoalesceDenominator}, 2)) AS 
+                , SUM(${data.capCoalesceDenominator}) AS 
+                  ${data.alias}_denominator_sum
+                , SUM(POWER(${data.capCoalesceDenominator}, 2)) AS 
                   ${data.alias}_denominator_sum_squares
                 ${
                   data.regressionAdjusted
-                    ? `, 
-                  SUM(${data.capCoalesceCovariate}) AS ${data.alias}_covariate_sum,
-                  SUM(POWER(${data.capCoalesceCovariate}, 2)) AS ${data.alias}_covariate_sum_squares,
-                  SUM(${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_denominator_pre_sum,
-                  SUM(POWER(${data.capCoalesceDenominatorCovariate}, 2)) AS ${data.alias}_denominator_pre_sum_squares,              
-                  SUM(${data.capCoalesceMetric} * ${data.capCoalesceDenominator}) AS ${data.alias}_main_denominator_sum_product, 
-                  SUM(${data.capCoalesceMetric} * ${data.capCoalesceCovariate}) AS ${data.alias}_main_covariate_sum_product, 
-                  SUM(${data.capCoalesceMetric} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_main_post_denominator_pre_sum_product, 
-                  SUM(${data.capCoalesceCovariate} * ${data.capCoalesceDenominator}) AS ${data.alias}_main_pre_denominator_post_sum_product,
-                  SUM(${data.capCoalesceCovariate} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_main_pre_denominator_pre_sum_product, 
-                  SUM(${data.capCoalesceDenominator} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_denominator_post_denominator_pre_sum_product
+                    ? `
+                  , SUM(${data.capCoalesceCovariate}) AS ${data.alias}_covariate_sum
+                  , SUM(POWER(${data.capCoalesceCovariate}, 2)) AS ${data.alias}_covariate_sum_squares
+                  , SUM(${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_denominator_pre_sum
+                  , SUM(POWER(${data.capCoalesceDenominatorCovariate}, 2)) AS ${data.alias}_denominator_pre_sum_squares
+                  , SUM(POWER(${data.capCoalesceDenominatorCovariate}, 2)) AS ${data.alias}_denominator_pre_sum_squares
+                  , SUM(${data.capCoalesceMetric} * ${data.capCoalesceDenominator}) AS ${data.alias}_main_denominator_sum_product
+                  , SUM(${data.capCoalesceMetric} * ${data.capCoalesceCovariate}) AS ${data.alias}_main_covariate_sum_product
+                  , SUM(${data.capCoalesceMetric} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_main_post_denominator_pre_sum_product
+                  , SUM(${data.capCoalesceCovariate} * ${data.capCoalesceDenominator}) AS ${data.alias}_main_pre_denominator_post_sum_product
+                  , SUM(${data.capCoalesceCovariate} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_main_pre_denominator_pre_sum_product
+                  , SUM(${data.capCoalesceDenominator} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_denominator_post_denominator_pre_sum_product
                   `
-                    : `,
-                    SUM(${data.capCoalesceDenominator} * ${data.capCoalesceMetric}) AS ${data.alias}_main_denominator_sum_product
+                    : `
+                    , SUM(${data.capCoalesceDenominator} * ${data.capCoalesceMetric}) AS ${data.alias}_main_denominator_sum_product
                   `
                 }` /*ends ifelse regressionAdjusted*/
                 : ` 
               ${
                 data.regressionAdjusted
-                  ? `,
-                SUM(${data.capCoalesceCovariate}) AS ${data.alias}_covariate_sum,
-                SUM(POWER(${data.capCoalesceCovariate}, 2)) AS ${data.alias}_covariate_sum_squares,
-                SUM(${data.capCoalesceMetric} * ${data.capCoalesceCovariate}) AS ${data.alias}_main_covariate_sum_product
+                  ? `
+                , SUM(${data.capCoalesceCovariate}) AS ${data.alias}_covariate_sum
+                , SUM(POWER(${data.capCoalesceCovariate}, 2)) AS ${data.alias}_covariate_sum_squares
+                , SUM(${data.capCoalesceMetric} * ${data.capCoalesceCovariate}) AS ${data.alias}_main_covariate_sum_product
                 `
                   : ""
               }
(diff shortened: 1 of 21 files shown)


packages/stats/gbstats/gbstats.py
```diff
@@ -121,37 +123,41 @@ def detect_unknown_variations(
...
+    if dimension == "pre:date":
+        dimension_column_name = "dim_pre_date"
+    elif dimension == "pre:activation":
+        dimension_column_name = "dim_exp_activation"
+    elif dimension.startswith("exp:"):
+        dimension_column_name = "dim_exp_" + dimension.split(":")[1]
+    elif dimension.startswith("precomputed:"):
+        dimension_column_name = "dim_exp_" + dimension.split(":")[1]
+    elif dimension.startswith("dim_"):
+        dimension_column_name = "dim_unit_" + dimension
```
ahdriel, line 142: Does this mean the column will be named `dim_unit_dim_{something}`?

packages/stats/gbstats/gbstats.py
```diff
@@ -121,37 +123,41 @@ def detect_unknown_variations(
...
-    return dfc
+def get_dimension_column_name(dimension: str) -> str:
+    dimension_column_name = "dimension"
+    if dimension == "pre:date":
+        dimension_column_name = "dim_pre_date"
+    elif dimension == "pre:activation":
+        dimension_column_name = "dim_exp_activation"
+    elif dimension.startswith("exp:"):
+        dimension_column_name = "dim_exp_" + dimension.split(":")[1]
+    elif dimension.startswith("precomputed:"):
```
ahdriel, line 139: question: Do we need to support the `pre:` here too or not?

packages/stats/gbstats/gbstats.py
```diff
@@ -168,17 +174,28 @@ def get_metric_df(
...
             i = var_id_map[key]
             dimensions[dim]["total_users"] += row.users
             prefix = f"v{i}" if i > 0 else "baseline"
-            for col in ROW_COLS:
+
+            # Sum here in case multiple rows per dimension
+            for col in SUM_COLS:
+                # Special handling for count, if missing returns a method, so override with user value
+                if col == "count" and callable(getattr(row, col)):
+                    dimensions[dim][f"{prefix}_count"] += getattr(row, "users", 0)
```
ahdriel, line 194: question: Before these changes, we used `=` instead of `+=`, but now we use the latter for both. Is it intentional?

packages/front-end/components/Experiment/ResultsTable.tsx
```diff
@@ -107,6 +106,10 @@ const ROW_HEIGHT = 56;
 const METRIC_LABEL_ROW_HEIGHT = 44;
 const SPACER_ROW_HEIGHT = 6;
 
+export enum RowError {
+  QUANTILE_AGGREGATION_ERROR = "Quantile metrics cannot be re-aggregated across pre-computed dimension breakdowns.",
```
ahdriel: question: Should we just have the identifier in the enum like `QUANTILE_AGGREGATION_ERROR = 'QUANTILE_AGGREGATION_ERROR'`? Then we can define either in shared (if needed for both back-end and front-end)?

Right now the object with this error would have the full string instead of just an indentifier.

I think it is fine for now if we are only dynamically constructing this error, but if we ever save this somewhere then updating the string would be harder and we would need to keep multiple versions around.

packages/front-end/components/Report/ConfigureReport.tsx
```diff
@@ -208,7 +208,10 @@ export default function ConfigureReport({
             <DimensionChooser
               value={form.watch("experimentAnalysisSettings.dimension") || ""}
               setValue={(v) =>
-                form.setValue("experimentAnalysisSettings.dimension", v)
+                form.setValue(
+                  "experimentAnalysisSettings.dimension",
+                  v ?? undefined
```
ahdriel: On value we are using an empty string as fallback but here we are using undefined. Should we consolidate on one?

packages/front-end/components/GeneralSettings/ExperimentSettings/index.tsx
```diff
@@ -196,6 +196,68 @@ export default function ExperimentSettings({
...
+                        <Text size="3" className="font-weight-semibold">
+                          Pre-computed Dimension Breakdowns
+                        </Text>{" "}
+                        <GBInfo />
+                      </PremiumTooltip>
+                    </label>
+                  </Box>
+                  <Box>
+                    <Text>
+                      Pre-compute dimension breakdowns using dimension columns
```
ahdriel, line 248: Would be nice to link to more docs here when we have them.

packages/front-end/components/Experiment/ResultsTable.tsx
```diff
@@ -1074,7 +1106,7 @@ function getChangeTooltip(
     </>
   );
   let intervalText = <></>;
-  if (hasRisk && statsEngine === "bayesian") {
+  if (statsEngine === "bayesian") {
```
ahdriel, line 1225: question: Why have we removed this check?

packages/front-end/components/Experiment/AnalysisSettingsBar.tsx
```diff
@@ -161,8 +165,12 @@ export default function AnalysisSettingsBar({
           ) : null}
           <div className="col-auto form-inline pr-5">
             <DimensionChooser
-              value={dimension}
+              value={precomputedDimension ?? dimension ?? ""}
               setValue={setDimension}
```
ahdriel: suggestion: Would be nice to find a way to abstract this so we only deal with dimension and not 2 different separate fields for precomputed and ondemand.

Especially weird IMO where the `value` and the `setValue` do not match -- means we have some kind of implicit behavior somewhere that is likely harded to reason about.

packages/front-end/components/Experiment/Results.tsx
```diff
@@ -158,9 +162,9 @@ const Results: FC<{
   const showBreakDownResults =
     !draftMode &&
     hasData &&
-    snapshot?.dimension &&
-    snapshot.dimension.substring(0, 8) !== "pre:date" && // todo: refactor hardcoded dimension
-    analysis?.settings?.dimensions?.length; // todo: needed? separate desired vs actual
+    ((snapshot?.dimension &&
+      snapshot.dimension.substring(0, 8) !== "pre:date") ||
+      analysis?.settings?.dimensions?.length);
```
ahdriel: suggestion: If we do it like this we change the type of `showBreakDownResults` from `number | boolean | undefined` to only `boolean`.

```suggestion
    ((snapshot?.dimension &&
      snapshot.dimension.substring(0, 8) !== "pre:date") ||
      (analysis?.settings?.dimensions?.length ?? 0) > 0);
```

packages/front-end/components/Experiment/Results.tsx
```diff
@@ -176,6 +180,11 @@ const Results: FC<{
     );
   }
 
+  // cannot re-aggregate quantile metrics across pre-computed dimensions
+  const showQuantileReaggregationError = analysis?.settings?.dimensions.some(
```
ahdriel: Not sure what's a better name but this confused me initially because it felt we would show the error by default if any dimension is precomputed for the snapshot, which is partially true because this will only affect for quantile metrics.

packages/front-end/components/Experiment/Results.tsx
```diff
@@ -325,9 +336,9 @@ const Results: FC<{
           statsEngine={analysis?.settings?.statsEngine || DEFAULT_STATS_ENGINE}
           differenceType={analysis.settings?.differenceType}
         />
-      ) : showBreakDownResults ? (
+      ) : showBreakDownResults && snapshot ? (
         <BreakDownResults
-          key={snapshot.dimension}
+          key={analysis?.settings?.dimensions?.[0] ?? snapshot?.dimension}
```
ahdriel: nit: I think we can remove the `?` from the snapshot given the above check.

packages/back-end/types/query.d.ts
```diff
@@ -49,7 +49,7 @@ export interface QueryInterface {
   // eslint-disable-next-line
   result?: Record<string, any>;
   queryType?: QueryType;
-  rawResult?: Record<string, number | string | boolean | object>[];
+  rawResult?: Record<string, number | string | boolean | object | undefined>[];
```
ahdriel: question: when do we update this to `unknown` or `any`? 😂

packages/back-end/types/experiment-snapshot.d.ts
```diff
@@ -111,6 +111,8 @@ export interface DimensionForSnapshot {
   // The same format we use today that encodes both the type and id
   // For example: `exp:country` or `pre:date`
   id: string;
+  // Pre-defined dimension levels, if they exist
+  slices?: string[];
```
ahdriel, line 115: question: why are we calling it `slices` here and `pre-computed` or `pre-defined` dimensions in other places?

Should the key be `precomputed` instead?

packages/back-end/types/experiment-snapshot.d.ts
```diff
@@ -164,6 +166,7 @@ export interface SnapshotBanditSettings {
 // Also used to determine when to show "out-of-date" in the UI
 export interface ExperimentSnapshotSettings {
   manual: boolean;
+  type?: SnapshotType;
```
ahdriel: question: Do we need this here? Shouldn't the type belong only to the Snapshot itself?

What happens if the values are not the same between the snapshot and the snapshot settings?

packages/back-end/src/types/Integration.ts
```diff
@@ -399,10 +398,10 @@ export type ExperimentMetricQueryResponseRows = {
   quantile_lower?: number;
   quantile_upper?: number;
   quantile_nstar?: number;
+  [key: string]: number | string | undefined;
```
ahdriel, line 401: question: Can we add a comment on what kinds of dynamically named fields we expect here?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1252,9 +1252,16 @@ export default abstract class SqlIntegration
           };
         }
 
+        const dimensionData: Record<string, string> = {};
+        Object.entries(row)
+          .filter(([key, _]) => key.startsWith("dim"))
```
jdorn: Nit: including a separator like `dim_` would make this a little more robust and less likely to match other words.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2624,36 +2644,23 @@ export default abstract class SqlIntegration
...
-        "'Activated'"
-      );
-    }
+    const dimensionCols = params.dimensions.map((d) => this.getDimensionCol(d));
 
+    // auto compute activation splits
     const timestampColumn =
-      activationMetric && dimension?.type !== "activation"
+      activationMetric &&
+      !!params.dimensions.find((d) => d.type === "activation")
```
jdorn: Is this right?  Previously, it was checking if the dimension type is NOT activation, but now it's doing the inverse.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2781,15 +2803,19 @@ export default abstract class SqlIntegration
...
-          d.dimension AS dimension,
-          ${banditDates?.length ? `d.bandit_period AS bandit_period,` : ""}
-          ${cumulativeDate ? `dr.day AS day,` : ""}
-          d.${baseIdType} AS ${baseIdType},
+          d.variation AS variation
+          ${
+            dimensionCols.length > 0
+              ? `, ${dimensionCols.map((c) => c.alias).join(", ")}`
+              : ""
+          }
```
jdorn: I don't remember exactly when or with which data source, but I think we've run into bugs in the past because of missing aliases. Might be safer to leave these in, even if it seems duplicative (although we're very inconsistent with this already throughout the SQL)

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3024,54 +3090,63 @@ export default abstract class SqlIntegration
...
-                  SUM(${data.capCoalesceMetric} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_main_post_denominator_pre_sum_product, 
-                  SUM(${data.capCoalesceCovariate} * ${data.capCoalesceDenominator}) AS ${data.alias}_main_pre_denominator_post_sum_product,
-                  SUM(${data.capCoalesceCovariate} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_main_pre_denominator_pre_sum_product, 
-                  SUM(${data.capCoalesceDenominator} * ${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_denominator_post_denominator_pre_sum_product
+                    ? `
+                  , SUM(${data.capCoalesceCovariate}) AS ${data.alias}_covariate_sum
+                  , SUM(POWER(${data.capCoalesceCovariate}, 2)) AS ${data.alias}_covariate_sum_squares
+                  , SUM(${data.capCoalesceDenominatorCovariate}) AS ${data.alias}_denominator_pre_sum
+                  , SUM(POWER(${data.capCoalesceDenominatorCovariate}, 2)) AS ${data.alias}_denominator_pre_sum_squares
+                  , SUM(POWER(${data.capCoalesceDenominatorCovariate}, 2)) AS ${data.alias}_denominator_pre_sum_squares
```
jdorn: Accidental duplicated line?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3086,13 +3161,47 @@ export default abstract class SqlIntegration
...
+      dimensionCol = {
+        value: this.ifElse(
+          `first_activation_timestamp IS NULL`,
+          "'Not Activated'",
+          "'Activated'"
+        ),
+        alias: "dim_exp_activation",
+      };
+    }
+    return dimensionCol;
```
jdorn: This can be made more typesafe and future-proof by using a `switch` statement and early returns. Typescript will know if you forget one of the cases and fail to compile. For example, here you are missing the `datedaily` and `datecumulative` cases, but there's no type error and it just fails silently.

```ts
switch(dimension.type) {
    case "experiment":
        return ...
    case "user":
        return ...
}
```

The important part is to not have a default case or a final ending return statement.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3086,13 +3161,47 @@ export default abstract class SqlIntegration
...
+        alias: "dim_pre_date",
+      };
+    } else if (dimension?.type === "activation") {
+      dimensionCol = {
+        value: this.ifElse(
+          `first_activation_timestamp IS NULL`,
+          "'Not Activated'",
+          "'Activated'"
+        ),
+        alias: "dim_exp_activation",
```
jdorn: To avoid conflicts with experiment dimensions, should we name this `dim_activation` instead?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3260,36 +3370,22 @@ export default abstract class SqlIntegration
...
-      );
-    }
+    // ensure always at least length 1 with plain dim
+    const dimensionCols = params.dimensions.map((d) => this.getDimensionCol(d));
 
+    // auto compute activation splits
     const timestampColumn =
-      activationMetric && dimension?.type !== "activation"
+      activationMetric &&
+      !!params.dimensions.find((d) => d.type === "activation")
```
jdorn: Same here with the inverted logic. Is this a typo?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3456,15 +3581,21 @@ export default abstract class SqlIntegration
           quantileMetric === "event"
             ? `
         LEFT JOIN __quantileMetric qm
-        ON (qm.dimension = umj.dimension AND qm.variation = umj.variation)`
+        ON (qm.variation = umj.variation AND ${dimensionCols
+          .map((c) => `qm.${c.alias} = umj.${c.alias}`)
+          .join(" AND ")})`
```
jdorn: This is missing the check for `dimensionCols.length > 0` that we do elsewhere. The SQL will break here if the array is empty.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3697,7 +3860,11 @@ export default abstract class SqlIntegration
   ${"ignoreNulls" in metric && metric.ignoreNulls ? `WHERE m.value != 0` : ""}
   GROUP BY
     m.variation
-    , ${cumulativeDate ? `${this.formatDate("m.day")}` : "m.dimension"}
+    ${
+      dimensionCols.length > 0
+        ? `, ${dimensionCols.map((c) => `m.${c.alias}`).join(", ")}`
+        : ""
+    }
```
jdorn: Nit: this is a simpler way to do this without having the check the length everywhere.

```ts
${dimensionCols.map((c) => `, m.${c.alias}`).join("")}
```

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3260,36 +3370,22 @@ export default abstract class SqlIntegration
...
-      )}`;
-    } else if (dimension?.type === "activation") {
-      dimensionCol = this.ifElse(
-        `first_activation_timestamp IS NULL`,
-        "'Not Activated'",
-        "'Activated'"
-      );
-    }
+    // ensure always at least length 1 with plain dim
+    const dimensionCols = params.dimensions.map((d) => this.getDimensionCol(d));
```
jdorn, line 3321: Is the comment here true? Where is it enforcing that `params.dimensions.length > 0`?

packages/stats/gbstats/gbstats.py
lukesonnet: Yes, we are now summing across rows if the same dimension is split across other rows (other levels of other dimensions, e.g. chrome US and chrome UK need to be combined).

packages/stats/gbstats/gbstats.py
lukesonnet: that is covered above by pre:date and pre:activation

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Good idea. I think I was trying to also capture `dimension`, but that shouldn't be kept around at all.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Oof, thank you.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yep, it was Clickhouse. Thanks.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: This must have been old, thanks.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yep, typo, thanks.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: ty

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: ty, typo, mistake in my memory of the logic.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: ty

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: good suggestion, thank you.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: yep, sg, ty.

packages/stats/gbstats/gbstats.py
lukesonnet: Yes, e.g. `dim_unit_dim_2s1by13xdlq758m60`.

packages/front-end/components/Experiment/ResultsTable.tsx
lukesonnet: It was unnecessary. The stats engine here comes from the analysis and is sufficient. Maybe it was a holdover from when we used to display risk in the results table. I'm not sure, but I'm going to rip out even more references since we don't use it at all.

packages/shared/src/experiments.ts
```diff
@@ -1176,3 +1177,39 @@ export function getAdjustedCI(
...
+  dimensionMetadata: { specifiedSlices: string[] }[],
+  nVariations: number
+): number {
+  const nLevels: number[] = [];
+  dimensionMetadata.forEach((dim) => {
+    nLevels.push(dim.specifiedSlices.length);
+  });
+
+  return nLevels.reduce((acc, n) => acc * n, 1) * nVariations;
+}
```
jdorn, line 1220: We should add unit tests for these 2 functions

packages/back-end/types/experiment-snapshot.d.ts
lukesonnet: This is the levels (e.g. Chrome, Safari), not the dimension itself.

packages/back-end/src/types/Integration.ts
lukesonnet: done, I hope this works. The typing here only includes number and undefined because it has to accommodate all explicit keys.

packages/back-end/types/experiment-snapshot.d.ts
lukesonnet: Thanks, this was a bad way to tell the query runner what type a snapshot was.

packages/front-end/components/Report/ConfigureReport.tsx
lukesonnet: Fixed it with a rewrite of how setValue works for this component.

packages/back-end/types/query.d.ts
lukesonnet: It should just be any.

packages/front-end/components/Experiment/ResultsTable.tsx
lukesonnet: Yeah, no need for the text, changed it to an identifier. The string can go somewhere shared if and when we need to.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -298,10 +308,17 @@ export const startExperimentResultQueries = async (
           .filter(Boolean)
       );
     }
+    // Only run dimensional analysis for quantile metrics
+    // if snapshot type is not standard
+    const runOverallQuantileAnalysis =
+      snapshotType === "standard" &&
```
ahdriel: The comment says `is not standard` but the check is `=== standard`. Just checking if intentional.

packages/front-end/components/Experiment/AnalysisSettingsBar.tsx
lukesonnet: As per our discussion, you can see I'm using the `setAnalysisBarSettings`, but I still have to manage the state of the actual dimension and analysis settings using other methods for now. We could continue to refactor this if need be.

packages/stats/gbstats/gbstats.py
```diff
@@ -737,18 +755,35 @@ def process_single_metric(
...
-                var_id_map=get_var_id_map(a.var_ids),
-                metric=metric,
-                analysis=a,
-            ),
-            baseline_index=a.baseline_index,
+
+    results: List[List[DimensionResponse]] = []
+    for a in analyses:
+        # skip pre-computed dimension reaggregation for quantile metrics
+        # TODO also skip if dimension columns
```
ahdriel: comment so we don't forget about this todo (if it needs to be done before merge)

packages/stats/gbstats/gbstats.py
```diff
@@ -737,18 +755,35 @@ def process_single_metric(
...
+        attempted_quantile_dimension_reaggregation = a.dimension.startswith(
+            "precomputed:"
+        ) and metric.statistic_type in ["quantile_event", "quantile_unit"]
+        attempted_quantile_overall_reaggregation = (
+            a.dimension == ""
+            and metric.statistic_type in ["quantile_event", "quantile_unit"]
+            and pdrows.columns.__contains__("dim_exp")
+        )  # TODO test
```
ahdriel: another todo

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -323,9 +338,16 @@ export const startExperimentResultQueries = async (
   }
 
   for (const [i, m] of groups.entries()) {
+    // Only run dimensional analysis for quantile metrics
+    // if snapshot type is not standard
+    const runOverallQuantileAnalysis =
+      snapshotType === "standard" &&
+      dimensionObjs.length > 0 &&
+      m.some(quantileMetricType);
+
```
ahdriel: Should we update this one too?

ahdriel (commented):
Thanks for the updates!

packages/stats/gbstats/gbstats.py
lukesonnet: This one is already done!

packages/shared/src/experiments.ts
lukesonnet: done.

packages/stats/gbstats/gbstats.py
```diff
@@ -121,37 +128,41 @@ def detect_unknown_variations(
...
     # We want to end up with one row per dimension
     for row in dfc.itertuples(index=False):
-        dim = row.dimension
+        # strip dimension of prefix before `:`
+        dimension_column_name = (
+            "dimension" if not dimension else get_dimension_column_name(dimension)
+        )
+        # fall back to one unnamed dimension if no column found
+        dim = getattr(row, dimension_column_name, "")
+        print(f"dim: {dim}")
```
ahdriel: question: Do we want to keep this `print` here or do we have a logger?

packages/back-end/src/types/Integration.ts
```diff
@@ -132,22 +132,13 @@ export type ExperimentDimension = {
 export type DateDimension = {
   type: "date";
 };
-
-export type DateCumulativeDimension = {
```
ahdriel, line 136: question: Does this mean we are not supporting these fields anymore?

I am not familiar with them, so they might already be deprecated but I am curious about the user experience in case they have previously configured these.

packages/back-end/src/services/experiments.ts
```diff
@@ -478,6 +487,32 @@ export function getSnapshotSettings({
     (q) => q.id === experiment.exposureQueryId
   );
 
+  // get dimensions for standard analysis
+  // TODO(dimensions): customize which dimensions to use at experiment level
```
ahdriel, line 490: question: is this a future TODO or before merge TODO?

packages/back-end/src/services/experiments.ts
```diff
@@ -612,7 +647,10 @@ export async function createManualSnapshot({
     experiment,
     phaseIndex,
     orgPriorSettings: orgPriorSettings,
-    settings: analysisSettings,
+    orgDisabledPrecomputedDimensions: false,
```
ahdriel, line 649: question: Should we get this from org settings?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1277,9 +1285,15 @@ export default abstract class SqlIntegration
     const { rows, statistics } = await this.runQuery(query, setExternalId);
     return {
       rows: rows.map((row) => {
+        const dimensionData: Record<string, string> = {};
+        Object.entries(row)
+          .filter(([key, _]) => key.startsWith("dim"))
```
ahdriel: question: The filter above uses `dim_` but this one uses just `dim`. Should we update them to match?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2890,15 +2901,17 @@ export default abstract class SqlIntegration
           eventQuantileData.length
             ? `
         LEFT JOIN __eventQuantileMetric qm
-        ON (qm.dimension = umj.dimension AND qm.variation = umj.variation)`
+        ON (qm.variation = umj.variation ${dimensionCols
+          .map((c) => `AND qm.${c.alias} = umj.${c.alias}`)
```
ahdriel, line 2915: question: Do we need a space here somewhere?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3024,54 +3037,58 @@ export default abstract class SqlIntegration
...
+          .join("\n")}
       FROM
         __userMetricAgg m
         ${
           eventQuantileData.length
             ? `LEFT JOIN __eventQuantileMetric qm ON (
-          qm.dimension = m.dimension AND qm.variation = m.variation
+          qm.variation = m.variation 
+          ${dimensionCols
+            .map((c) => `AND qm.${c.alias} = m.${c.alias}`)
```
ahdriel, line 3100: question: Same here. Do we need a space for multiple `AND`?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3422,30 +3462,32 @@ export default abstract class SqlIntegration
         }
       )
       ${
-        quantileMetric === "event"
+        quantileMetric === "event" // TODO(sql): put quantiles in their own query
```
ahdriel, line 3475: checking if this is a before-merge or a future work TODO.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3456,15 +3498,17 @@ export default abstract class SqlIntegration
           quantileMetric === "event"
             ? `
         LEFT JOIN __quantileMetric qm
-        ON (qm.dimension = umj.dimension AND qm.variation = umj.variation)`
+        ON (qm.variation = umj.variation ${dimensionCols
+          .map((c) => `AND qm.${c.alias} = umj.${c.alias}`)
```
ahdriel, line 3512: same here for the AND space

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3668,7 +3715,9 @@ export default abstract class SqlIntegration
     ${
       quantileMetric === "event"
         ? `LEFT JOIN __quantileMetric qm ON (
-      qm.dimension = m.dimension AND qm.variation = m.variation
+      qm.variation = m.variation ${dimensionCols
+        .map((c) => `AND qm.${c.alias} = m.${c.alias}`)
```
ahdriel, line 3729: same here for spacing

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3986,20 +4039,24 @@ export default abstract class SqlIntegration
     __banditPeriodWeights bpw
     ON (
       bps.bandit_period = bpw.bandit_period 
-      AND bps.dimension = bpw.dimension
+      ${dimensionCols
```
ahdriel, line 4051: question: Won't this create a `AND AND`?

ahdriel (changes requested):

packages/stats/gbstats/gbstats.py
lukesonnet: oof, no we dont.

packages/back-end/src/types/Integration.ts
lukesonnet: They are being deprecated AND have never been used. I remove more in a follow up PR: #4368 .

packages/back-end/src/services/experiments.ts
lukesonnet: future TODO.

packages/back-end/src/services/experiments.ts
lukesonnet: No, this is manual snapshots that come from users manually enter data and are basically (if not actually) deprecated. Precomputed dimensions doesn't make sense for these in any case.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: future work.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yes. Good catch. It wasn't possible to find this in QA because we don't allow multiple dimensions right now for quantile testing. We could also make this more explicit in this function that `dimensionCols` can only be length 1.... but I'd rather just split out quantiles altogether.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yes, but now I need to figure out why this wasn't caught in my testing.

packages/back-end/src/integrations/SqlIntegration.ts
ahdriel: Should we update it to use the `.join(" AND ")` as the other code does? Or will the new line be interpreted as a space/word separation?

PR #3905, opened 2025-04-04
Update how we calculate readAccess for Metric Groups

mknowlton89:
### Features and Changes

This PR updates how we calculate whether a user as `readAccess` to a Metric Group. Previously, the `canRead` check was calling `canReadSingleProjectResource` instead of `canReadMultiProjectResource`, given that Metric Groups can have multiple projects.

- Closes **(No issue)**

### Dependencies

- None

### Testing

- [x] Ensure that global experimenters with no project roles can see all Metric Groups
- [x] Ensure that a user with a no access global role, but with `experimenter` permissions in Project A, can see a Metric Group in Project A.
- [x] Likewise, if the Metric Group is in `All Projects` the user above should be able to view it.

Please note: The ability for a user to view metrics inside of a metric group still use the `metric` leverl permissions. E.g. if an org has a Metric Group in Project A, but it only contains metrics that are in Project B, the user above would be able to see the metric group, but not any of the metrics inside of it.


ahdriel (approved):
:shipit:

PR #3910, opened 2025-04-07
new collection and updates

gazzdingo:
### Features and Changes
updated the new collection based on chats on friday


packages/back-end/src/validators/features.ts
```diff
@@ -168,31 +168,12 @@ const experimentRefRule = baseRule
...
   .extend({
     type: z.literal("safe-rollout"),
-    trackingKey: z.string(),
-    datasource: z.string(),
-    exposureQueryId: z.string(),
     controlValue: z.string(),
     value: z.string(),
     coverage: z.number(),
-    hashAttribute: z.string(),
-    seed: z.string(),
```
msamper, line 189: Do these need to be moved? I don't think they're considered to be analysis settings. At least not from the perspective of the UI

packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -58,8 +58,8 @@ const snapshotMetricObject = z.object({
     )
     .optional(),
   chanceToWin: z.number().optional(),
-  errorMessage: z.string().optional().nullable(),
-  power: metricPowerResponseFromStatsEngineObject.optional().nullable(),
+  errorMessage: z.string().optional(),
+  power: metricPowerResponseFromStatsEngineObject.optional(),
```
msamper, line 62: Changing this to not be nullable is likely going to cause zod validator issues. I had to make these nullable because its possible for a null value to come back from the stats engine

packages/back-end/src/types/Audit.ts
```diff
@@ -49,6 +49,7 @@ export const entityEvents = {
   factTable: ["autocreate"],
   customField: ["create", "update", "delete"],
   experimentTemplate: ["create", "update", "delete"],
+  safeRollout: ["create", "update", "delete"],
```
msamper, line 52: maybe change to `safeRolloutAnalysisSettings`?

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -414,14 +414,16 @@ export async function createSnapshot({
 
 export async function createSafeRolloutSnapshot({
   context,
-  safeRollout,
+  safeRolloutRule,
   feature,
   dimension,
   useCache = true,
   triggeredBy,
+  safeRollout,
```
msamper, line 422: `safeRolloutAnalysisSettings` ?

packages/back-end/src/routers/safe-rollout-snapshot/safe-rollout-snapshot.controller.ts
```diff
@@ -144,13 +145,21 @@ export const createSnapshot = async (
   // This is doing an expensive analytics SQL query, so may take a long time
   // Set timeout to 30 minutes
   req.setTimeout(SNAPSHOT_TIMEOUT);
-
+  const safeRolloutModel = new SafeRolloutModel(context);
```
msamper, line 148: I think you should be able to access the model from the context here instead of instantiating a new instance of the model

msamper: `context.models.safeRollout.findByRuleId(...)`

packages/back-end/src/models/SafeRolloutModel.ts
```diff
@@ -0,0 +1,126 @@
...
+    startedAt: z.date(),
+    lastSnapshotAttempt: z.date().optional(),
+    nextSnapshotAttempt: z.date().optional(),
+    autoSnapshots: z.boolean().default(true),
+    ruleId: z.string(),
+    featureId: z.string(),
+    coverage: z.number(),
+    controlValue: z.string(),
+    variationValue: z.string(),
+  })
```
msamper, line 31: Is this the right validator for the analysis settings model? Or are we duplicating fields here that also exist on the rule for some reason?

packages/back-end/src/models/SafeRolloutModel.ts
```diff
@@ -0,0 +1,126 @@
...
+  seed: string;
+  guardrailMetrics: string[];
+  status: typeof safeRolloutStatus[number];
+  startedAt: Date;
+  coverage: number;
+  controlValue: string;
+  variationValue: string;
+}
+
+export class SafeRolloutModel extends BaseClass {
```
msamper, line 67: We should probably call this `SafeRolloutAnalysisSettingsModel`

packages/back-end/src/models/SafeRolloutModel.ts
```diff
@@ -0,0 +1,126 @@
+import { z } from "zod";
+import { getCollection } from "back-end/src/util/mongo.util";
+import { SafeRolloutRule } from "back-end/src/validators/features";
+import { baseSchema, MakeModelClass } from "./BaseModel";
+const COLLECTION = "safeRolloutAnalysisSettings";
+
+export const safeRolloutStatus = [
+  "running",
+  "rolled-back",
+  "released",
```
msamper, line 10: should we do `reverted` and `rolled-out`? I think thats what we ended up deciding in one of our meetings. Also happy to do it after we merge this because I'll need to change some logic with generating the payload

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -303,6 +307,136 @@ export async function getFeaturesByIds(
...
+          autoSnapshots: "$rulesArray.autoSnapshots",
+          nextSnapshotAttempt: "$rulesArray.nextSnapshotAttempt",
+          status: "$rulesArray.status",
+        },
+      },
+    },
+  ]);
+  const jsonDocuments = documents.map((doc) => doc.toJSON());
+  return jsonDocuments;
+}
```
msamper, line 362: Do we still need this? I assume now we'll use the analysis settings collection to determine feature rules that are ready to be updated so we don't have to do a complex mongo aggregation

packages/back-end/src/types/Audit.ts
gazzdingo: we decided to call it saferollout in the meeting on monday that is why i refactored to to safeRollout today

packages/back-end/src/models/SafeRolloutModel.ts
gazzdingo: i duplicated it just because it it alot easier to deal with an in the future we can move away from the rule and have most of it live here. also i was thinking we can just use this collection as well if we wanted

packages/back-end/src/models/SafeRolloutModel.ts
gazzdingo: oh i cant remember that but yeah we can updated that

packages/back-end/src/models/FeatureModel.ts
gazzdingo: that was left over woops i will delete

packages/back-end/src/models/SafeRolloutModel.ts
msamper: Do you think there's a risk at all of the rule and the safe rollout document here falling out of sync? that would be my main concern with duplicating values. I do like the idea of having an easy way to migrate over to a safe rollout rule being by reference instead, though

packages/back-end/src/validators/features.ts
```diff
@@ -168,31 +168,12 @@ const experimentRefRule = baseRule
...
-    guardrailMetrics: z.array(z.string()),
-    status: z.enum(safeRolloutStatus).default("draft"),
```
msamper, line 191: I think we need status to be behind approval flows also

PR #3919, opened 2025-04-08
feature(timeseries): Add error + loading states & remove button for old experiments

ahdriel:
### Features and Changes

- If the experiment has been stopped before Apr 4, we will not show the Time Series button for the metrics.
- Added messages to loading, error and no data states.

### Screenshots

Screenshot 2025-04-08 at 4 21 12 PM
Screenshot 2025-04-08 at 4 21 01 PM
Screenshot 2025-04-08 at 4 21 23 PM

### DIFF
--- a/packages/front-end/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper.tsx
+++ b/packages/front-end/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper.tsx
@@ -1,3 +1,4 @@
+import { Flex } from "@radix-ui/themes";
 import { DifferenceType, StatsEngine } from "back-end/types/stats";
 import { ExperimentStatus } from "back-end/src/validators/experiments";
 import { MetricTimeSeries } from "back-end/src/validators/metric-time-series";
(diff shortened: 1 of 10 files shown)


packages/front-end/components/Experiment/ResultsTable.tsx
```diff
@@ -172,12 +174,17 @@ export default function ResultsTable({
...
   const gb = useGrowthBook<AppFeatures>();
-  const showTimeSeriesButton =
+  let showTimeSeriesButton =
     baselineRow === 0 &&
     tableRowAxis === "metric" &&
     !disableTimeSeriesButton &&
     gb.isOn("experiment-results-timeseries");
 
+  // Disable time series button for stopped experiments before we added this feature (& therefore data)
+  if (status === "stopped" && endDate <= "2025-04-03") {
```
lukesonnet, line 184: I assume you tested this works; something weird about endDate string comparisons just bugging me :)

packages/front-end/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper.tsx
```diff
@@ -171,3 +185,18 @@ export default function ExperimentMetricTimeSeriesGraphWrapper({
     />
   );
 }
+
+function Message({ children }: { children: React.ReactNode }) {
+  return (
+    <Flex
+      align="center"
+      height="220px"
```
lukesonnet, line 193: Should we make it smaller? Probably awkward if loading is big and then it shrinks smaller for error state?

packages/front-end/components/Experiment/ResultsTable.tsx
lukesonnet: works for me

ahdriel: Yeah -- it works because we use ISO string which is (by design) lexically comparable like this -- but agreed that it can be weird at first glance 👀

packages/front-end/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper.tsx
ahdriel: All of the states would have the same height because the graph is 220px, this way we avoid any layout change (besides the expansion of the drawer).

PR #3920, opened 2025-04-09
Experiment Template Promo

mknowlton89:
### Features and Changes

This PR introduces a promo card for `Experiment Templates` and is rendered on the `Experiments` index page. It is rendered for any organization that doesn't have any `Experiment Templates`, so long as the user viewing the page has permission to create `Experiment Templates`.

If any organization does not have an `Enterprise` plan, the CTA directs users to `Contact Sales` and links to the https://www.growthbook.io/demo. (Note: I think there are discussions about building a specific page for this, but it doesn't exist yet).

If an organization DOES have an `Enteprise` plan, the CTA opens the `Create Template Modal` directly.

### Dependencies

- None

### Testing

- [x] Ensure that if an organization already has templates, the promo isn't rendered.
- [x] Ensure that if the user viewing the /experiments page doesn't have `canCreateExperimentTemplates` permission, the promo isn't rendered.
- [x] Ensure that the promo looks good on both light and dark modes
- [x] Ensure that the promo looks good at all screen sizes (mobile and desktop)
- [x] Ensure that if the org does not have an `Enterprise` plan, the CTA reads "Contact Sales" and opens a new tab at https://www.growthbook.io/demo
- [x] Ensure that if the org does have an `Enterprise` plan, the CTA reads "Create Template` and opens the `Create Template Modal` directly.

### Screenshots
Screenshot 2025-04-09 at 3 50 55 PM
Screenshot 2025-04-09 at 3 51 06 PM


packages/front-end/pages/experiments/index.tsx
```diff
@@ -725,6 +729,14 @@ const ExperimentsPage = (): React.ReactElement => {
                         onPageChange={setCurrentPage}
                       />
                     )}
+                    {canAddTemplate && !templates.length ? (
```
jdorn, line 732: I think we also want to limit this to only show once they have at least 5 experiments.  Before that point, templates are not really that valuable and this will distract from the main thing we want them to do - create their first few experiments.

jdorn (commented):
Run the png image through `pngquant` to reduce the size.  When I tested it, it went from 142K down to 43K

packages/front-end/pages/experiments/index.tsx
mknowlton89: Sounds great - `pngquant` was able to reduce the size to 44K and I've updated the display logic.

jdorn (approved):

PR #3929, opened 2025-04-10
Add common tracing API.

romain-growthbook:
This PR abstracts away the tracing metrics API and let each respective tracing fill them up with their own implementation.

This makes it possible to switch between opentelemetry and datadog as metrics provider.


ahdriel (approved):
Changes LGTM

One question is: why are we separating the implementation between OTel and DD? Would we be able to use [OTel tracing with DataDog client](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/nodejs/otel/) instead or something along those lines?

Unsure about the tradeoffs here though.

romain-growthbook:
> Changes LGTM
> 
> One question is: why are we separating the implementation between OTel and DD? Would we be able to use [OTel tracing with DataDog client](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/nodejs/otel/) instead or something along those lines?

That's what we've been doing.

> Unsure about the tradeoffs here though.

We want to enable profiling which is not supported by OTel so we're switching to a native dd tracing to get this feature enabled.

PR #3930, opened 2025-04-11
Performance improvements for base db model

jdorn:
From profiling data, one of the most expensive code paths in our back-end is creating Zod validators.  Typically this is only done once at server start up and is no big deal, but the base db class does this unnecessarily multiple times on every incoming request.

This PR moves the Zod creation up a level so it's only done once and re-used.


ahdriel (approved):
Nice find!

PR #3937, opened 2025-04-14
Fix dd metrics

romain-growthbook:
This is a better approximation of how DD collects metrics from OTL.

### DIFF
--- a/packages/back-end/src/tracing.datadog.ts
+++ b/packages/back-end/src/tracing.datadog.ts
@@ -1,14 +1,12 @@
 import "./init/aliases";
 import tracer from "dd-trace";
-import { setMetrics } from "./util/metrics";
+import { Attributes, setMetrics } from "./util/metrics";
 
 tracer.init({
   logInjection: true,
 });
 
 const COLLECTION_INTERVAL_SECONDS = 15;
-
-type Attributes = Record<string, string | number>;
 
 class Counter {
   name: string;
--- a/packages/back-end/src/tracing.datadog.ts
+++ b/packages/back-end/src/tracing.datadog.ts
@@ -6,22 +6,22 @@ tracer.init({
   logInjection: true,
 });
 
-const COLLECTION_INTERVAL = 15;
+const COLLECTION_INTERVAL_SECONDS = 15;
+
+type Attributes = Record<string, string | number>;
 
 class Counter {
   name: string;
   value: number;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  attributes?: any;
+  attributes?: Attributes;
 
   constructor(name: string) {
     this.name = name;
     this.value = 0;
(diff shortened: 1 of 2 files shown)


packages/back-end/src/tracing.datadog.ts
```diff
@@ -6,15 +6,59 @@ tracer.init({
   logInjection: true,
 });
 
+const COLLECTION_INTERVAL = 15;
+
+class Counter {
+  name: string;
+  value: number;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  attributes?: any;
```
ahdriel, line 15: should this be `object` instead of `any`? Or would this work with a string for example?

packages/back-end/src/tracing.datadog.ts
```diff
@@ -6,15 +6,63 @@ tracer.init({
   logInjection: true,
 });
 
+const COLLECTION_INTERVAL = 15;
```
ahdriel, line 9: suggestion: Can we add the unit for this? is it seconds or milliseconds? For example `COLLECTION_INTERVAL_MS = `

packages/back-end/src/tracing.datadog.ts
```diff
@@ -6,15 +6,63 @@ tracer.init({
...
+class Counter {
+  name: string;
+  value: number;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  attributes?: any;
+
+  constructor(name: string) {
+    this.name = name;
+    this.value = 0;
+    setInterval(this.collect, COLLECTION_INTERVAL);
```
ahdriel, line 20: question: How/when are these cleaned up?

packages/back-end/src/tracing.datadog.ts
```diff
@@ -6,15 +6,63 @@ tracer.init({
...
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  add(v: number, attributes?: any) {
+    this.value += v;
+    this.attributes = attributes;
```
ahdriel, line 26: question: Should `attributes` be managed separately if they'll be overriden if we make multiple calls to `add`?

packages/back-end/src/tracing.datadog.ts
romain-growthbook: Looked back at `dd-trace` it should be `Record<string, string|number>`

packages/back-end/src/tracing.datadog.ts
romain-growthbook: It will be but I think that this is also the case with OTL. Attributes are always the same, they are used to statically tag a metric to filter it when plotting values

ahdriel (approved):

jdorn (approved):

PR #3945, opened 2025-04-16
Add decision criteria data to getExperiment api endpoint

Kevin-Chant:
### Features and Changes

Adds a new (optional) `decisionCriteriaStatus` field to the Experiment API object and pulls the data in the get endpoint (but not the list endpoint yet).
Also refactors some of the code from the frontend hook into the shared/enterprise folder so it can be more readily imported elsewhere.

This will be consumed initially by the Jira app, but could be useful elsewhere too

### Testing

Hit the `/experiments/{id}` endpoint and observe the new data being returned
Also double check that the list endpoint `/experiments` still works as expected

### Screenshots

image


packages/back-end/src/api/experiments/getExperiment.ts
```diff
@@ -11,9 +18,40 @@ export const getExperiment = createApiRequestHandler(getExperimentValidator)(
...
+    if (settings?.defaultDecisionCriteriaId) {
+      try {
+        decisionCriteria ||=
+          (await req.context.models.decisionCriteria.getById(
+            settings!.defaultDecisionCriteriaId!
+          )) ?? DEFAULT_DECISION_CRITERIA;
+      } catch {
+        // Empty catch
+      }
+    }
```
ahdriel, line 36: suggestion: We need to find better names for these.

We get the default decision criteria for an org, but it can return `undefined` if `settings.defaultDecisionCriteriaId` is not one of the hardcoded `DEFAULT_DECISION_CRITERIA(s)`; and if we get undefined back but we have a `settings.defaultDecisionCriteriaId` set then we fetch a `decisionCriteria` by `default criteria id`.

I am finding it confusing that we use `default` in 2 contexts here, one meaning the ones we (GB) provide to users out of the box, and two being the ones they set as default for their org/experiments.

Maybe we change the `DEFAULT_DECISION_CRITERIA` to be `PRESET_DECISION_CRITERIA` which would change the first function to be `getPresetDecisionCriteriaForOrg`?

cc @lukesonnet

packages/back-end/src/api/experiments/getExperiment.ts
```diff
@@ -11,9 +18,40 @@ export const getExperiment = createApiRequestHandler(getExperimentValidator)(
...
+      status,
+      detailedStatus,
+    }))(
+      getStatusIndicatorData(
+        experiment,
+        false,
+        healthSettings,
+        decisionCriteria
+      )
+    );
```
ahdriel, line 49: suggestion: That took me a bit to understand that we just want to limit the number of fields returned from `statusData`; maybe we can simplify it to use if / 2 variables instead?

packages/back-end/src/services/experiments.ts
```diff
@@ -1324,7 +1325,11 @@ function getExperimentMetric(
 
 export async function toExperimentApiInterface(
   context: ReqContext | ApiReqContext,
-  experiment: ExperimentInterface
+  experiment: ExperimentInterface,
+  decisionCriteriaStatus?: Pick<
```
ahdriel, line 1329: question(nit): I am not sure how I feel about passing extra info to `experiment.toExperimentApiInterface` and it just returns it back.

Can you help me understand why this is needed / alternatives?

packages/shared/src/enterprise/decision-criteria/statusIndicatorData.ts
```diff
@@ -0,0 +1,187 @@
+import {
```
ahdriel, line 1: While some of these statuses in this file are enterprise, I don't think all of them are.

Thoughts about moving this outside of the enterprise folder? We can keep the functions that create the `ExperimentResultStatusData` in enterprise so we can differentiate between what's paid and what's not.

packages/back-end/src/api/experiments/getExperiment.ts
Kevin-Chant: Yeah probably better not to be fancy here

packages/back-end/src/services/experiments.ts
Kevin-Chant: This is one that I wasn't sure about either. My idea was that since this information isn't part of the `ExperimentInterface`, but it should be in the response, that this would be one way to merge it in. I could also do that merging in the API code but that means that `toExperimentApiInterface` would only be producing a partial object that should be augmented elsewhere.

packages/back-end/src/api/openapi/schemas/ExperimentWithDecisionCriteriaStatus.yaml
```diff
@@ -0,0 +1,18 @@
+allOf:
+  - $ref: "./Experiment.yaml"
+  - type: object
+    properties:
+      decisionCriteriaStatus:
```
ahdriel, line 5: hmm I think this name is not entirely correct -- we might have a status that is not powered by the decision criteria at all.

For the front-end we use:
```
export type ExperimentResultStatus =
  | DecisionFrameworkExperimentRecommendationStatus
  | { status: "no-data" }
  | { status: "unhealthy"; unhealthyData: ExperimentUnhealthyData }
  | { status: "before-min-duration" };
```

Which is the underlying data that we use to figure out the label (`Running`, `Stopped`, etc) + detailed status and tooltip if needed.

Maybe should use `detailedStatus` ? (which might be powered by decision criteria but might not?)

packages/back-end/src/api/experiments/getExperiment.ts
```diff
@@ -11,12 +18,38 @@ export const getExperiment = createApiRequestHandler(getExperimentValidator)(
...
+    const healthSettings = getHealthSettings(
+      settings,
+      orgHasPremiumFeature(req.context.org, "decision-framework")
+    );
+    let decisionCriteria = getPresetDecisionCriteriaForOrg(settings);
+    if (settings?.defaultDecisionCriteriaId) {
+      try {
+        decisionCriteria ||=
+          (await req.context.models.decisionCriteria.getById(
+            settings!.defaultDecisionCriteriaId!
```
ahdriel, line 31: nit: I think we can remove both `!` here.

ahdriel (approved):
Nice! Thanks for the updates

packages/back-end/src/api/experiments/getExperiment.ts
lukesonnet: Yes, I'll refactor this in my banner PR.

PR #3946, opened 2025-04-16
fix: BigQuery client initialisation

vmallya-123:
### Features and Changes

This includes changes to make Bigquery work in auto credentials mode, where the GCP project which has Bigquery can be in a different project than where Growthbook is deployed. The Bigquery client was either not being initialised with the project or it was being initialised with Growthbook project id which was incorrect.

We encountered this issue when we were trying to deploy Growthbook on our K8s

### Testing

BQ will connect to default project even if Growthbook is deployed to different GCP project


packages/back-end/src/integrations/BigQuery.ts
```diff
@@ -36,11 +36,13 @@ export default class BigQuery extends SqlIntegration {
...
     if (!IS_CLOUD && this.params.authType === "auto") {
-      return new bq.BigQuery();
+      return new bq.BigQuery({
+        projectId: this.params.defaultProject,
+      });
     }
 
     return new bq.BigQuery({
-      projectId: this.params.projectId,
+      projectId: this.params.defaultProject,
```
jdorn, line 45: I don't think these changes are backwards compatible and could break existing installations.  Will need to think of a different solution here.

vmallya-123: for non auto the project would be inferred from the service account json file (i have tested this). Only for auto if default project is empty then maybe we initialise it without project
```
if (!IS_CLOUD && this.params.authType === "auto") {
  const options = {};
  if (this.params.defaultProject) {
    options.projectId = this.params.defaultProject;
  }
  return new bq.BigQuery(options);
}
```

jdorn:

PR #3949, opened 2025-04-16
Add saferollout feature events

lukesonnet:
Adds three new safe rollout events that fire when a safe rollout snapshot runs:
* `feature.saferollout.unhealthy`
* `feature.saferollout.ship`
* `feature.saferollout.rollback`

They fire when a safe rollout snapshot is successful and the decision framework renders a decision. There is currently no notification for a successful update with no actionable status.


packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
```diff
@@ -466,6 +468,140 @@ export function getExperimentResultStatus({
...
+        daysNeeded: Infinity, // sequential relied upon solely for safe rollouts
+      })
+    : undefined;
+
+  // If rollback now, return rollback now
+  if (decisionStatus?.status === "rollback-now") {
+    return {
+      status: "rollback-now",
+      variationIds: decisionStatus.variationIds,
+      sequentialUsed: false,
```
msamper, line 573: should this be true?

lukesonnet: yes ty

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -312,7 +322,7 @@ function getSafeRolloutSnapshotSettings({
     dimensions: settings.dimensions.map((id) => ({ id })),
     startDate: safeRollout.startedAt || new Date(), // TODO: What do we want to do if startedAt is not set?
     endDate: new Date(),
-    experimentId: safeRolloutRule.trackingKey,
+    experimentId: safeRolloutRule.trackingKey || "",
```
msamper, line 325: is this needed? the tracking key should always be defined on the rule

msamper (approved):
Looks good to me aside from the one comment I left. Let me know if you have any guidance on how to test this

PR #3950, opened 2025-04-16
Safe Rollouts snapshot API endpoints

lukesonnet:
Adds create snapshot API endpoints for safe rollouts, and adds some stubs to suggest where the main API endpoints could live, but I'm unsure of our strategy here.


lukesonnet:
https://growthbookapp.slack.com/archives/C08HK09B481/p1744844577416149?thread_ts=1744843125.013279&cid=C08HK09B481

PR #3951, opened 2025-04-17
Introduce in-app metric group promos

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/front-end/enterprise/components/MetricGroupInlineForm.tsx
+++ b/packages/front-end/enterprise/components/MetricGroupInlineForm.tsx
@@ -1,0 +1,91 @@
+import { useForm } from "react-hook-form";
+import { Flex } from "@radix-ui/themes";
+import { MetricGroupInterface } from "back-end/types/metric-groups";
+import { useState } from "react";
+import { useAuth } from "@/services/auth";
+import Field from "@/components/Forms/Field";
+import Button from "@/components/Radix/Button";
+import Callout from "@/components/Radix/Callout";
+import track from "@/services/track";
+
+export default function MetricGroupInlineForm({
+  selectedMetricIds,
+  datasource,
+  mutateDefinitions,
+  onChange,
+  cancel,
+}: {
+  datasource: string;
+  selectedMetricIds: string[];
+  mutateDefinitions: () => Promise<void>;
+  onChange: (metrics: string[]) => void;
+  cancel: () => void;
+}) {
+  const [error, setError] = useState<string | undefined>(undefined);
+  const [loading, setLoading] = useState(false);
+  const { apiCall } = useAuth();
+  const form = useForm<{ name: string; metrics: string[]; datasource: string }>(
+    {
+      defaultValues: {
+        name: "",
+        datasource,
+        metrics: selectedMetricIds,
+      },
+    }
+  );
+
+  const submit = form.handleSubmit(async () => {
+    setError(undefined);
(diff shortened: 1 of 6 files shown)


packages/front-end/enterprise/components/MetricGroupInlineForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+        body: JSON.stringify(value),
+      });
+      await mutateDefinitions();
+
+      onChange([res.metricGroup.id]);
+    } catch (e) {
+      setError(
+        `Unable to save Metric Group. ${
+          e.message ? `Reason:  ${e.message}` : ""
+        }`
```
jdorn, line 64: I saw in the screenshot in the PR description an error about invalid data source.  Is that a real error?  If so, can we detect that ahead of time and disable the metric groups prompt so they don't even see this form?

mknowlton89: Re: The "Invalid Data Source", it is a real error, but I don't think its one that'll actually be hit often. I forced it to happen locally just to get an error state.

The error can happen [here](https://github.com/growthbook/growthbook/blob/084bbba85914dfefc8bc56fb75bdd337edf7f209/packages/back-end/src/routers/metric-group/metric-group.controller.ts#L35), where we look up the Data Source by ID. As we're populating the list of data sources on the front end, this shouldn't happen, but it's possible since `getDataSourceById` can return the Data Source or null.

packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -410,6 +413,18 @@ export default function TabbedPage({
             : "d-none d-print-block"
         }
       >
+        {!metricGroups.length ? (
```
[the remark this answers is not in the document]
jdorn: add this in addition to the check that's already there, so they should not have any metric groups yet AND have 2+ metrics in the experiment

packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -197,6 +198,43 @@ export default function TabbedPage({
...
+        experiment.secondaryMetrics.some((secondaryMetric) =>
+          isMetricGroupId(secondaryMetric)
+        ) ||
+        experiment.guardrailMetrics.some((guardrailMetric) =>
+          isMetricGroupId(guardrailMetric)
+        )
+      ) {
+        return false;
+      }
+    }
```
jdorn, line 224: We can simplify both of these checks and just do `if (metricGroups.length) return false;`  We don't care about weird edge cases like someone had metric groups as enterprise, but then downgraded to free.

packages/front-end/enterprise/components/MetricGroupInlineForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+  );
+
+  const submit = form.handleSubmit(async () => {
+    setError(undefined);
+    setLoading(true);
+    const value = {
+      name: form.watch("name"),
+      metrics: form.watch("metrics"),
+      datasource: form.watch("datasource"),
+    };
```
jdorn, line 44: Don't use `form.watch` inside the submit handler.  Do this instead:

```ts
const submit = form.handleSubmit(async (value) => {
```

jdorn (approved):

PR #3954, opened 2025-04-18
Give old portals priority over radix portals

bryce-fitzsimons:
before:
image

after:
image


lukesonnet (approved):
Approved, but I'm not sure the full scope or if there's other parts of the app we should be testing for this (e.g. results tooltips or something else with Modals)

bryce-fitzsimons:
> Approved, but I'm not sure the full scope or if there's other parts of the app we should be testing for this (e.g. results tooltips or something else with Modals)

I checked as many places as I could (including results tooltips), but possible some small regressions could occur.

PR #3955, opened 2025-04-18
Introduce In-App Bandit Promo

mknowlton89:
### Features and Changes

This PR introduces a new in-app promo for `Multi-Arm Bandits`. These promos live in two places.

1. Within `/experiment/id` page in the `Implementation` section. The promo is rendered if the experiment isn't a bandit AND has more than 2 variations.

2. Within the `Edit Variations` modal if the experiment isn't already a bandit AND has more than 2 variations.

In both cases, we're leveraging the `PremiumCallout` so the CTA is different depending on whether or not the organization has the ability to create Bandits.

There may be a future where the CTA for organizations that can create bandits (aka those that have the feature and has sticky bucketing enabled), but I think that's a good future step to take, for now, like the other promos, it either prompts the org to upgrade to a plan that has the feature or if the org already has the feature, directs them in a new tab to the docs to learn more.

### Dependencies

- NA

### Testing

- [x] Ensure that the in-app promo is rendered only when an experiment has more than 2 variations and isn't already a bandit.
- [x] Ensure that the promo works as expected from all of the various entry points, not just the experiment id page (but also from places like the template page, new experiment form, etc.

### Screenshots

Screenshot 2025-04-18 at 10 33 52 AM
Screenshot 2025-04-18 at 10 34 02 AM
Screenshot 2025-04-18 at 10 34 25 AM
Screenshot 2025-04-18 at 10 34 38 AM
Screenshot 2025-04-18 at 10 55 01 AM
Screenshot 2025-04-18 at 10 55 12 AM
Screenshot 2025-04-18 at 10 55 49 AM
Screenshot 2025-04-18 at 10 56 00 AM


packages/front-end/components/Features/FeaturesOverview.tsx
```diff
@@ -531,6 +532,18 @@ export default function FeaturesOverview({
...
+          commercialFeature={null}
+          dismissable={true}
+          docSection="devTools"
+          mb="3"
+        >
+          Download our [Browser] DevTools extension to easily preview and QA
+          feature flags in your application.
+        </PremiumCallout>
+        {/* <Callout status="wizard">Hi</Callout> */}
+
```
jdorn, line 546: I think that was added by accident to this PR

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -81,6 +82,19 @@ export default function Implementation({
             </Button>
           ) : null}
         </div>
+        {experiment.variations.length > 2 &&
+        experiment.type !== "multi-armed-bandit" ? (
```
jdorn, line 86: Only show when the experiment is a draft.

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -81,6 +82,19 @@ export default function Implementation({
...
+            id="exp-implementation-bandit-promo"
+            commercialFeature="multi-armed-bandits"
+            docSection="bandits"
+            dismissable={true}
+            mx="3"
+            mb="5"
+          >
+            Bandits can help you quickly find the best performing variant.
+          </PremiumCallout>
+        ) : null}
```
jdorn, line 97: When they already have Pro, instead of linking to docs, we should have a CTA that converts the experiment to a bandit instead.  This action already exists in the More menu at the top of the experiment page, so it's just a matter of opening the same confirmation modal from here.

image

packages/front-end/components/Features/FeatureVariationsInput.tsx
```diff
@@ -277,188 +280,192 @@ export default function FeatureVariationsInput({
...
+                  id="exp-implementation-bandit-promo"
+                  commercialFeature="multi-armed-bandits"
+                  docSection="bandits"
+                  dismissable={true}
+                  mt="4"
+                  mb="1"
+                >
+                  Bandits can help you quickly find the best performing variant.
+                </PremiumCallout>
+              ) : null}
```
jdorn, line 400: Same here.  Instead of linking to docs for Pro users, we want to just let them convert this to a bandit right away.  This is much trickier to do here vs on the experiment page.  Honestly, might be better to just hide this callout completely for existing Pro users for now.  Just removing the `docSection` prop should do that for you automatically.

packages/front-end/components/Radix/PremiumCallout.tsx
```diff
@@ -12,7 +12,7 @@ import Link from "@/components/Radix/Link";
 import styles from "./RadixOverrides.module.scss";
 
 export type Props = {
-  commercialFeature: CommercialFeature;
+  commercialFeature: CommercialFeature | null;
```
jdorn, line 15: I think we should revert all the changes to this file.  I want to keep this component purely for premium features.

packages/front-end/components/Features/FeatureVariationsInput.tsx
mknowlton89: @jdorn Yea - I think it might get a bit funky to try and convert this to a bandit while the modal is open.

mknowlton89: Re: `Honestly, might be better to just hide this callout completely for existing Pro users for now. Just removing the docSection prop should do that for you automatically.` 

I'm sorta inclined to leave it in. Since it can be dismissed, I do like the idea of rendering the callout with the link to the docs to give a clear path for someone to learn what bandits are. The user can always dismiss the callout if they'd like.

Thoughts?

jdorn: I'd probably remove it from here.  Let them create the experiment draft without distractions, then they will see the callout on the experiment page where they land on after and can switch to a bandit there.  I think that will be a better UX.

packages/front-end/components/Radix/PremiumCallout.tsx
```diff
@@ -18,6 +18,7 @@ export type Props = {
   renderWhenDismissed?: (undismiss: () => void) => React.ReactElement;
   children: React.ReactNode;
   docSection?: DocSection;
+  cta?: React.ReactElement;
```
jdorn, line 21: The `docSection` prop is really just a shortcut for doing `cta={<DocLink docSection="...">View Docs</DocLink>}`.

I would get rid of the `docSection` prop and also change the logic here slightly.  I think the `Upgrade Now` and `Talk to Sales` CTAs should take precedence.  So only render this `cta` prop when the user already has the feature.

mknowlton89:
Closing this as it's stale. I can re-open if this gets re-prioritized.

PR #3958, opened 2025-04-18
Add JS SDK plugins for backend DevTools (plugins: generic, NextJS, Express)

bryce-fitzsimons:
### Supports debugging back-end SDKs with DevTools browser extension

Implements (via SDK plugin) much of the boilerplate mentioned in these docs:
https://docs.growthbook.io/tools/chrome-extension#back-end-debugging

- New devtools plugins
  - Generic base plugin for setting devtools state in the SDK
  - NextJS specific plugin with entrypoints for RSCs and middleware/apis
  - helper method for logging state back to DevTools + example React-based implementation
- Adds dev mode + dev log support to user-scoped SDKs (Node)


packages/sdk-js/src/plugins/devtools.ts
```diff
@@ -0,0 +1,229 @@
...
+        payload: gb.getDecryptedPayload?.() || {
+          features: gb.getFeatures?.(),
+          experiments: gb.getExperiments?.(),
+        },
+        attributes: gb.getAttributes(),
+      },
+    };
+  } else if (gb instanceof UserScopedGrowthBook) {
+    // UserScopedGrowthBook SDK
+    // @ts-expect-error private access
```
bryce-fitzsimons, line 207: I don't love this, but not sure if worth addressing in the SDK. Should we make public getters?

packages/sdk-js/src/GrowthBookClient.ts
```diff
@@ -369,6 +369,23 @@ export class UserScopedGrowthBook<
   ) {
     this._gb = gb;
     this._userContext = userContext;
+    this.logs = [];
+
+    if (userContext.enableDevMode) {
+      const cb = userContext.onFeatureUsage;
```
bryce-fitzsimons, line 375: not sure if I love this...

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -332,6 +334,7 @@ export type GlobalContext = {
 export type UserContext = {
   enabled?: boolean;
   qaMode?: boolean;
+  enableDevMode?: boolean;
```
bryce-fitzsimons, line 337: presuming that devMode should be user context rather than a property of the user-scoped SDK...

packages/sdk-js/src/plugins/devtools.ts
jdorn: I don't think this will work.  During bundling, all variable names that start with underscore are mangled.  A public getter for both the parent GrowthBookClient instance and the `UserContext` makes sense.

packages/sdk-js/src/plugins/devtools.ts
```diff
@@ -0,0 +1,229 @@
...
+      } catch (e) {
+        console.error("cannot parse devtools payload", e);
+      }
+    }
+
+    devtoolsPlugin(state)(gb);
+  };
+}
+
+type LogUnionWithSource = LogUnion & { source?: string };
```
jdorn, line 157: Ideally `source` would live within `SdkInfo` instead since it will be identical for all log entries.  I know the published devtools extension expects it this way though, so kind of annoying to change now.

packages/sdk-js/src/plugins/devtools.ts
```diff
@@ -0,0 +1,229 @@
...
+    return {
+      logs: gb.logs.map((log) => ({ ...log, source, clientKey })),
+      sdkInfo: {
+        apiHost,
+        clientKey,
+        version: gb.version,
+        payload: gb.getDecryptedPayload?.() || {
+          features: gb.getFeatures?.(),
+          experiments: gb.getExperiments?.(),
+        },
```
jdorn, line 201: I don't think we need this since `getDecryptedPayload` will fall back to using `getFeatures()` and `getExperiments()` internally already.  Same with the GrowthBookClient version below.

packages/sdk-js/src/types/growthbook.ts
jdorn: We should also set this in `GrowthBook._getUserContext` for consistency.

packages/sdk-js/src/plugins/devtools.ts
bryce-fitzsimons: agree, only a minor change in devtools

packages/sdk-js/src/GrowthBookClient.ts
bryce-fitzsimons: ok, I have it somewhat abstracted with deduping, similar to how we wrap `onFeatureUsage` inside `_trackFeatureUsage` for the standard SDK

jdorn (approved):

PR #3960, opened 2025-04-21
Adds archetype in-app promo to feature flag page

mknowlton89:
### Features and Changes

This PR introduces an in-app promo via the `PremiumCallout` for Archetypes within the [fid].tsx page.

In the event an organization does not have any archetypes, we'll render the promo callout. This is a dismiss-able callout. If the org has access to the feature (e.g they're pro or enterprise) it'll like out to the docs. Otherwise, it'll prompt the org to upgrade.

### Dependencies

N/A

### Testing

- [x] Ensure that the correct CTA/icon is used depending on whether or not the org has access to archetypes
- [x] Ensure that the styling looks correct for both light and dark modes
- [x] Ensure that the callout isn't rendered if the org has previously created at least 1 archetype.
- [x] Ensure that when a callout is dismissed, it remains dismissed for the duration of the browser session

### Screenshots
`Starter` Org Experience
Screenshot 2025-04-21 at 10 30 49 AM
Screenshot 2025-04-21 at 10 30 58 AM

`Pro` & `Enterprise` Org Experience
Screenshot 2025-04-21 at 10 33 30 AM
Screenshot 2025-04-21 at 10 33 38 AM

### DIFF
--- a/packages/front-end/components/Features/FeaturesOverview.tsx
+++ b/packages/front-end/components/Features/FeaturesOverview.tsx
@@ -27,6 +27,7 @@ import {
 import { FeatureUsageLookback } from "back-end/src/types/Integration";
 import { Box, Flex, Heading, Switch, Text } from "@radix-ui/themes";
 import { RxListBullet } from "react-icons/rx";
+import { ArchetypeInterface } from "back-end/types/archetype";
 import Button from "@/components/Radix/Button";
 import { GBAddCircle, GBEdit } from "@/components/Icons";
 import LoadingOverlay from "@/components/LoadingOverlay";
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Features/FeaturesOverview.tsx
```diff
@@ -1073,6 +1082,18 @@ export default function FeaturesOverview({
                         Default Value.
                       </p>
                     )}
+                    {!archetypes.length ? (
```
jdorn, line 1085: This is only relevant when you have a complex feature flag.  Only show if there are 2+ rules.  Because of that, I'm wondering if this makes sense beneath the environment tabs instead of above them.

packages/front-end/components/Features/FeaturesOverview.tsx
```diff
@@ -113,6 +116,12 @@ export default function FeaturesOverview({
   const router = useRouter();
   const { fid } = router.query;
 
+  const { data } = useApi<{
+    archetype: ArchetypeInterface[];
+  }>("/archetype");
+
+  const archetypes = data?.archetype || [];
```
jdorn, line 123: This will cause flashing on the page.  Even when you do have archetypes, this will initially be an empty array on first load, so the banner will show for a second until this api call finishes.

jdorn (commented):
On further thought, I think we should probably not do this PR at all and rethink the design approach entirely.  Basically, there's the `Simulate` tab on the top of the page and the goal of this is to let people know that exists.

This design is not accomplishing that.  It's way too far down the page and removed from the tabs.  Feels like we need an entirely different UX approach for highlighting a tab the user might not know about yet.

PR #3961, opened 2025-04-21
Add custom meta data in-app promo

mknowlton89:
### Features and Changes


packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -150,6 +153,19 @@ export default function SetupTabOverview({
                 and parameters of your experiment
               </Box>
             )}
+            {!customFields.length ? (
```
jdorn, line 156: I think we should hide this if the description is empty.  If someone isn't even putting things in the description, they probably don't care about custom fields.

jdorn (approved):

PR #3965, opened 2025-04-21
Mechanism for generating sticky bucket assignments docs in synchronous context

bryce-fitzsimons:
Needed for `initSync()` hydration. Inspired by this conversation:
https://growthbookusers.slack.com/archives/C01T6Q1SVFV/p1742319278303349?thread_ts=1742313479.892719&cid=C01T6Q1SVFV


packages/sdk-js/src/GrowthBook.ts
```diff
@@ -1151,6 +1153,20 @@ export class GrowthBook<
...
+    payload: FeatureApiResponse
+  ) {
+    if (!("getAllAssignmentsSync" in stickyBucketService)) {
+      throw new Error(
+        "generating StickyBucketAssignmentDocs docs requires StickyBucketServiceSync"
+      );
+    }
+    const ctx = this._getEvalContext();
+    const attributes = getStickyBucketAttributes(ctx, payload);
+    return stickyBucketService?.getAllAssignmentsSync(attributes);
```
jdorn, line 1168: Remove optional chaining here, not needed

jdorn (commented):
Should we update this code in `initSync`?

```ts
    if (
      this._options.stickyBucketService &&
      !this._options.stickyBucketAssignmentDocs
    ) {
      throw new Error(
        "initSync requires you to pass stickyBucketAssignmentDocs into the GrowthBook constructor"
      );
    }
```

If the docs aren't passed in, but we have a sync sticky bucket service, populate the docs automatically.

bryce-fitzsimons:
> Should we update this code in `initSync`?
> 
> ```ts
>     if (
>       this._options.stickyBucketService &&
>       !this._options.stickyBucketAssignmentDocs
>     ) {
>       throw new Error(
>         "initSync requires you to pass stickyBucketAssignmentDocs into the GrowthBook constructor"
>       );
>     }
> ```
> 
> If the docs aren't passed in, but we have a sync sticky bucket service, populate the docs automatically.

Agree, we can remove this error and go straight to generating the docs with `generateStickyBucketAssignmentDocsSync()`. That method will error if not using a Sync SB service.

packages/sdk-js/src/sticky-bucket-service.ts
```diff
@@ -170,7 +196,10 @@ export class ExpressCookieStickyBucketService extends StickyBucketService {
     }
     return doc;
   }
-  async saveAssignments(doc: StickyAssignmentsDocument) {
+  async getAssignments(attributeName: string, attributeValue: string) {
+    return await this.getAssignmentsSync(attributeName, attributeValue);
+  }
```
jdorn, line 201: These async wrappers around sync functions are going to be identical for every sync service.  Should we just add them to `StickyBucketServiceSync` class directly instead?

packages/sdk-js/src/sticky-bucket-service.ts
```diff
@@ -89,6 +89,32 @@ export abstract class StickyBucketService {
...
+    attributes: Record<string, string>
+  ): Record<StickyAttributeKey, StickyAssignmentsDocument> {
+    const docs: Record<string, StickyAssignmentsDocument> = {};
+    Object.entries(attributes)
+      .map(([attributeName, attributeValue]) =>
+        this.getAssignmentsSync(attributeName, attributeValue)
+      )
+      .forEach((doc) => {
+        if (doc) {
+          const key = `${doc.attributeName}||${doc.attributeValue}`;
```
jdorn, line 110: We have this same delimited string formatting duplicated now 7 times (4 in this file, 3 in core.ts).  We should probably abstract it to a function.  

```ts
export function getStickyBucketKey(attributeName: string, value: string) {
  return `${attributeName}||${value}`;
}
```

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -50,7 +50,9 @@ import {
   getAllStickyBucketAssignmentDocs,
   decryptPayload,
   getApiHosts,
+  getStickyBucketAttributes,
 } from "./core";
+import { StickyBucketServiceSync } from "./sticky-bucket-service";
```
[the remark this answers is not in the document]
bryce-fitzsimons: Hmm, no matter how I import it, the `dist/bundles/esm.js` bundle always includes the actual SB classes. Even if I remove this import entirely and replace with `any` for the interfaces...

jdorn (approved):

PR #3966, opened 2025-04-22
Docs: added missing info about host and allowed paths for CDN page

vladenisov:
### Features and Changes

Documentation Update. When setting up the CDN, I encountered some non-obvious errors that I had to deal with in slack or through support.

### DIFF
--- a/docs/docs/self-host/cdn.mdx
+++ b/docs/docs/self-host/cdn.mdx
@@ -23,7 +23,7 @@ GrowthBook Cloud provides its own CDN for free for low/medium traffic apps (less
 
 1. Add a CDN that sits in front of your GrowthBook API server
 **Note**: Make sure to use the domain `api.growthbook.io`, not `cdn.growthbook.io`.
-2. Configure the CDN to cache requests to `/api/features/*` (and deny all other routes  except `/sub/*`)
+2. Configure the CDN to cache requests to `/api/features/*` (and deny all other routes except `/sub/*`)
 3. Point your SDKs to your CDN's domain instead of your GrowthBook API Server
 4. (optional) Configure a Webhook to invalidate the CDN cache when features change within GrowthBook
 
--- a/docs/docs/self-host/cdn.mdx
+++ b/docs/docs/self-host/cdn.mdx
@@ -22,7 +22,8 @@ GrowthBook Cloud provides its own CDN for free for low/medium traffic apps (less
 ## How it Works
 
 1. Add a CDN that sits in front of your GrowthBook API server
-2. Configure the CDN to cache requests to `/api/features/*` (and deny all other routes)
+**Note**: Make sure to use the domain `api.growthbook.io`, not `cdn.growthbook.io`.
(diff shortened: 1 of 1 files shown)


jdorn:
Thanks for the contribution!  I believe the changes here are incorrect though.  

1. When adding your own CDN in front of GrowthBook Cloud, you should use `https://cdn.growthbook.io` as the origin, not `https://api.growthbook.io`.  Accessing the GrowthBook Cloud API directly from the SDK imposes very strict rate limiting and is not considered stable, so we may change the format at any time, which would break your integration.
2.  While you could technically support streaming updates via `/sub/*` requests on your own CDN, it's not recommended since it is very difficult (and sometimes impossible) to configure properly and can get very expensive.

The docs can definitely be more clear about this so thanks for bringing this to our attention.  We can update them.

denisov-vlad:
@jdorn That's strange, because I received such solutions from the Support team. Also, your suggestions cannot be implemented now.

> When adding your own CDN in front of GrowthBook Cloud, you should use https://cdn.growthbook.io as the origin, not https://api.growthbook.io. Accessing the GrowthBook Cloud API directly from the SDK imposes very strict rate limiting and is not considered stable, so we may change the format at any time, which would break your integration.

This isn't possible. Proxying from your own CDN to GB's CDN will cause a Fastly error on GB's side. I've tried to set up it with Cloudflare. Also, found the same error messages in your Slack.

> While you could technically support streaming updates via /sub/* requests on your own CDN, it's not recommended since it is very difficult (and sometimes impossible) to configure properly and can get very expensive.

So how to configure SDK (iOS/Android) to use different endpoints for general API and streaming updates? We don't have a goal to proxy everything through a CDN, we just need a solution to make it all work.

jdorn:
I don't think the error you were seeing was from Fastly on our side, but likely something in your Cloudflare setup.  I did a quick test with the following simple Cloudflare worker script and it runs as expected without errors:

```ts
export default {
  async fetch(request, env, ctx) {
    const url = new URL(request.url);
    const pathname = url.pathname;

    // Restrict which paths are proxied
    if (!pathname.startsWith("/api/features/")) {
      return new Response("404 Not Found", {status: 404});
    }

    // Cache responses to the CDN
    let response = await caches.default.match(request);
    if (!response) {
      response = await fetch(`https://cdn.growthbook.io${pathname}`);
      ctx.waitUntil(caches.default.put(request, response.clone()))
    }

    return response;
  }
};
```

As for streaming, the Javascript SDK has a `streamingHost` option that you could point directly to `https://cdn.growthbook.io`, but unfortunately, it looks like our Swift and Kotlin SDKs do not support this right now.  I can put in a feature request there for you.  In the meantime, if you want to use your own CDN, you will need to disable streaming in the SDK.

If streaming updates are critical, instead of your own CDN, you can deploy our [GrowthBook Proxy Server](https://docs.growthbook.io/self-host/proxy).  However, in our experience, when you add in the infrastructure and maintenance costs, it's usually cheaper to just use https://cdn.growthbook.io directly and pay overage pricing.  If you have extremely high traffic volume, we're happy to work with you on volume discounts.

jdorn:

PR #3970, opened 2025-04-22
Vercel native integration

romain-growthbook:
This is the first pass on the native vercel integration. Should implement most of it, pending for more production testing.

### DIFF
--- a/packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
+++ b/packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
@@ -13,6 +13,7 @@ import {
 import {
   addMemberToOrg,
   addMembersToTeam,
+  getOrganizationById,
 } from "back-end/src/services/organizations";
 import {
   createOrganization,
@@ -42,6 +43,11 @@ import {
   updateTeamRemoveManagedBy,
 } from "back-end/src/models/TeamModel";
 import {
+  postCancelSubscriptionToLicenseServer,
+  postNewVercelSubscriptionToLicenseServer,
+} from "back-end/src/enterprise";
+import { getLicenseByKey } from "back-end/src/enterprise/models/licenseModel";
+import {
   userAuthenticationValidator,
   systemAuthenticationValidator,
   UpsertInstallationPayload,
@@ -51,21 +57,24 @@ import {
   BillingPlan,
 } from "./vercel-native-integration.validators";
 
-const FREE_BILLING_PLAN: BillingPlan = {
-  description: "Free Billing Plan",
-  id: "free-billing-plan",
-  name: "Free",
+const STARTER_BILLING_PLAN: BillingPlan = {
+  description:
+    "Growthbook's free plan. Add up to 3 users. Enjoy unlimited feature flag evaluations, community support, up to 1M CDN requests/month, and up to 5GB of CDN Bandwidth/month.",
+  id: "starter-billing-plan",
+  name: "Starter (Free) Plan",
   type: "subscription",
 };
 
-const NON_FREE_BILLING_PLAN: BillingPlan = {
-  description: "Non Free Billing Plan",
-  id: "non-free",
-  name: "Non free",
+const PRO_BILLING_PLAN: BillingPlan = {
+  description:
+    "Enjoy all the benefits of our starter plan, plus add up to 100 members (each member incurs a cost of $20/month), get in-app chat support, advanced experimentation features, up to 2M CDN requests/month, and up to 20GB of CDN Bandwidth/month.",
+  id: "pro-billing-plan",
+  name: "Pro Plan",
   type: "subscription",
+  cost: "20.00/month per user + usage",
 };
 
-const billingPlans = [FREE_BILLING_PLAN, NON_FREE_BILLING_PLAN] as const;
+const billingPlans = [STARTER_BILLING_PLAN, PRO_BILLING_PLAN] as const;
 
 const VERCEL_JKWS_URL = "https://marketplace.vercel.com/.well-known/jwks";
 
@@ -298,7 +307,7 @@ export async function upsertInstallation(req: Request, res: Response) {
     userId: user.id,
     name:
       payload.account.name ||
-      `Vercel instalation ${authentication.payload.installation_id}`,
+      `Vercel installation ${authentication.payload.installation_id}`,
     isVercelIntegration: true,
     restrictLoginMethod: `vercel:${req.params.installation_id}`,
   });
@@ -311,16 +320,13 @@ export async function upsertInstallation(req: Request, res: Response) {
 
   const integrationModel = new VercelNativeIntegrationModel(context);
 
-  const billingPlan = FREE_BILLING_PLAN;
-
   await integrationModel.create({
     installationId: authentication.payload.installation_id,
     upsertData: { payload, authentication: authentication.payload },
     resources: [],
-    billingPlanId: billingPlan.id,
   });
 
-  return res.send({ billingPlan });
+  return res.sendStatus(204);
 }
 
 export async function getInstallation(req: Request, res: Response) {
@@ -335,7 +341,10 @@ export async function getInstallation(req: Request, res: Response) {
 }
 
 export async function updateInstallation(req: Request, res: Response) {
-  const { integration, integrationModel } = await authContext(req, res);
+  const { integration, integrationModel, org, user } = await authContext(
+    req,
+    res
+  );
 
   const { billingPlanId } = req.body as UpdateInstallation;
 
@@ -343,20 +352,74 @@ export async function updateInstallation(req: Request, res: Response) {
 
   if (!billingPlan) return res.status(400).send("Invalid billing plan!");
 
+  // Check if current billing plan is different from new plan
+  if (integration.billingPlanId !== billingPlanId) {
+    if (billingPlanId === "pro-billing-plan") {
+      // The user is upgrading from the starter plan to the pro plan
+      try {
+        const result = await postNewVercelSubscriptionToLicenseServer(
+          org,
+          req.params.installation_id,
+          user?.name || ""
+        );
+        await updateOrganization(org.id, { licenseKey: result.id });
+      } catch (e) {
+        throw new Error(
+          `Unable to create new subscription. Reason: ${e.message} || "Unknown`
+        );
+      }
+    } else {
+      // The user is downgrading from the pro plan to the starter plan
+      try {
+        const license = await getLicenseByKey(org.licenseKey || "");
+
+        if (!license) {
+          return res
+            .status(404)
+            .send(`Unable to locate license for org: ${org.id}`);
+        }
+        await postCancelSubscriptionToLicenseServer(license.id);
+      } catch (e) {
+        throw new Error(
+          `Unable to cancel subscription. Reason: ${e.message} || "Unknown`
+        );
+      }
+    }
+  }
+
   await integrationModel.update(integration, { billingPlanId });
 
   return res.send({ billingPlan });
 }
 
 export async function deleteInstallation(req: Request, res: Response) {
-  const { context, integrationModel, integration } = await authContext(
+  const { context, integrationModel, integration, org } = await authContext(
     req,
     res
   );
 
+  if (integration.billingPlanId === "pro-billing-plan") {
+    const license = await getLicenseByKey(org.licenseKey || "");
+
+    if (!license) {
+      return res.status(404).send(`Invalid license for org: ${org.id}`);
+    }
+
+    await postCancelSubscriptionToLicenseServer(license.id);
+  }
+
   await removeManagedBy(context, { type: "vercel" });
   await integrationModel.deleteById(integration.id);
 
+  await updateOrganization(
+    org.id,
+    {
+      isVercelIntegration: false,
+    },
+    { restrictLoginMethod: 1 }
+  );
+
+  //MKTODO: Remove the finalized: true before we ship this to production
   return res.status(200).send({ finalized: true });
 }
 
@@ -378,6 +441,35 @@ export async function provisionResource(req: Request, res: Response) {
   const billingPlan = billingPlans.find(({ id }) => id === billingPlanId);
 
   if (!billingPlan) return res.status(400).send("Invalid billing plan!");
+
+  if (!integration.billingPlanId) {
+    // The installation doesn't have a billing plan yet, so we need to create a new one
+    if (billingPlanId === "pro-billing-plan") {
+      try {
+        // Get fresh org with updated name
+        const updatedOrg = await getOrganizationById(org.id);
+
+        if (!updatedOrg) {
+          throw new Error("Organization not found");
+        }
+        const result = await postNewVercelSubscriptionToLicenseServer(
+          updatedOrg,
+          req.params.installation_id,
+          user?.name || ""
+        );
+        await updateOrganization(org.id, { licenseKey: result.id });
+      } catch (e) {
+        throw new Error(
+          `Unable to create new subscription. Reason: ${e.message} || "Unknown`
+        );
+      }
+    }
+
+    // Then, update the integration with the new billing plan
+    await integrationModel.update(integration, {
+      billingPlanId,
+    });
+  }
 
   const resourceId = uuidv4();
 
@@ -429,7 +521,6 @@ export async function provisionResource(req: Request, res: Response) {
   };
 
   await integrationModel.update(integration, {
-    billingPlanId,
     resources: [...integration.resources, resource],
   });
 
@@ -490,7 +581,6 @@ async function removeManagedBy(
 export async function deleteResource(req: Request, res: Response) {
   const {
     context,
-    org,
     integrationModel,
     integration,
     resource,
@@ -506,14 +596,6 @@ export async function deleteResource(req: Request, res: Response) {
 
   await deleteVercelSdkWebhook(context);
 
-  await updateOrganization(
-    org.id,
-    {
-      isVercelIntegration: false,
-    },
-    { restrictLoginMethod: 1 }
-  );
-
   return res.sendStatus(204);
 }
 
@@ -524,10 +606,22 @@ export async function getProducts(req: Request, res: Response) {
 
   if (!slug) return res.status(400).send("Invalid request!");
 
-  if (integration.billingPlanId !== FREE_BILLING_PLAN.id)
-    return res.json({ plans: [NON_FREE_BILLING_PLAN] });
+  const existingBillingPlan = billingPlans.find(
+    ({ id }) => id === integration.billingPlanId
+  );
 
-  return res.json({ plans: billingPlans });
+  if (!existingBillingPlan) {
+    return res.json({ plans: billingPlans });
+  }
+
+  const additionalProjectBillingPlan: BillingPlan = {
+    ...existingBillingPlan,
+    name: "Add GrowthBook Project to your existing plan",
+    description: "",
+    cost: "",
+  };
+
+  return res.json({ plans: [additionalProjectBillingPlan] });
 }
 
 export async function postVercelIntegrationSSO(req: Request, res: Response) {
--- a/packages/back-end/src/enterprise/licenseUtil.ts
+++ b/packages/back-end/src/enterprise/licenseUtil.ts
@@ -19,6 +19,7 @@ import {
   LicenseUserCodes,
   SubscriptionInfo,
 } from "shared/enterprise";
+import { OrganizationInterface } from "back-end/types/organization";
 import { getLicenseByKey, LicenseModel } from "./models/licenseModel";
 import { LICENSE_PUBLIC_KEY } from "./public-key";
 
@@ -412,6 +413,29 @@ export async function postNewInlineSubscriptionToLicenseServer(
       cloudSecret: process.env.CLOUD_SECRET,
       organizationId,
       nonInviteSeatQty,
+    }),
+  });
+
+  verifyAndSetServerLicenseData(license);
+  return license;
+}
+
+export async function postNewVercelSubscriptionToLicenseServer(
+  organization: OrganizationInterface,
+  installationId: string,
+  userName: string
+): Promise<LicenseInterface> {
+  const url = `${LICENSE_SERVER_URL}subscription/new-vercel-native-subscription`;
+  const license = await callLicenseServer({
+    url,
+    body: JSON.stringify({
+      cloudSecret: process.env.CLOUD_SECRET,
+      organizationId: organization.id,
+      companyName: organization.name,
+      ownerEmail: organization.ownerEmail,
+      name: userName,
+      nonInviteSeatQty: organization.members.length,
+      installationId,
     }),
   });
 
--- a/packages/front-end/pages/settings/billing.tsx
+++ b/packages/front-end/pages/settings/billing.tsx
@@ -15,7 +15,7 @@ const BillingPage: FC = () => {
 
   const permissionsUtil = usePermissionsUtil();
 
-  const { accountPlan, subscription, canSubscribe } = useUser();
+  const { accountPlan, subscription, canSubscribe, organization } = useUser();
 
   const { apiCall } = useAuth();
   const { refreshOrganization } = useUser();
@@ -69,6 +69,19 @@ const BillingPage: FC = () => {
       <div className="container pagecontents">
         <div className="alert alert-danger">
           You do not have access to view this page.
+        </div>
+      </div>
+    );
+  }
+
+  if (organization.isVercelIntegration) {
+    return (
+      <div className="container pagecontents">
+        <div className="alert alert-info">
+          This page is not available for customers whose plan is managed by
+          Vercel. Please go to your Vercel Integration Dashboard for any billing
+          information. If you&apos;d like to cancel your subscription, you can
+          do so by uninstalling the GrowthBook Integration in Vercel
         </div>
       </div>
     );
--- a/packages/shared/src/enterprise/license-consts.ts
+++ b/packages/shared/src/enterprise/license-consts.ts
@@ -93,6 +93,7 @@ export interface LicenseInterface {
     tooltipText: string; // The text to show in the tooltip
     showAllUsers: boolean; // True if all users should see the notice rather than just the admins
   };
+  vercelInstallationId?: string;
   stripeSubscription?: {
     id: string;
     qty: number;


packages/back-end/src/services/vercel-native-integration.service.ts
```diff
@@ -0,0 +1,276 @@
...
+  origin: string;
+  category: "flag" | "experiment";
+  name?: string;
+  description?: string;
+  isArchived?: boolean;
+  createdAt?: number;
+  updatedAt?: number;
+};
+
+const FEATURE_ORIGIN = "app.growthbook.io";
```
jdorn, line 23: I think we have `APP_ORIGIN` as a constant already (set to `https://app.growthbook.io` on cloud)

packages/back-end/src/models/VercelNativeIntegrationModel.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  protected canDelete(): boolean {
+    return this.context.permissions.canManageIntegrations();
+  }
+}
+
+export const findVercelInstallationByInstallationId = async (
+  installationId: string
+): Promise<VercelNativeIntegration> => {
```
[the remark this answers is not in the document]
romain-growthbook: The problem is that the `BaseModel` class is too tightly integrated with `context` and `organization`. These methods make it possible to find the data when there is no context not organization available.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.router.ts
```diff
@@ -0,0 +1,58 @@
...
+  "/v1/installations/:installation_id",
+  validateRequestMiddleware({ body: updateInstallationValidator }),
+  vercelController.upsertInstallation
+);
+
+router.patch(
+  "/v1/installations/:installation_id",
+  validateRequestMiddleware({}),
+  vercelController.upsertInstallation
+);
```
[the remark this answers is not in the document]
romain-growthbook: This is a typo, thanks for pointing out.

packages/back-end/src/controllers/auth.ts
```diff
@@ -115,7 +115,29 @@ export async function postOAuthCallback(req: Request, res: Response) {
   }
 }
 
-async function sendLocalSuccessResponse(
+export async function setResponseCookies(
```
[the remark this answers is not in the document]
romain-growthbook: Done

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -1524,6 +1531,12 @@ const onExperimentCreate = async ({
   experiment: ExperimentInterface;
 }) => {
   await logExperimentCreated(context, experiment);
+
+  if (context.org.isVercelIntegration)
+    await createVercelExperimentationItemFromExperiment({
+      experiment,
+      organization: context.org,
+    });
```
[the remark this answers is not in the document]
romain-growthbook: Put a catch + log

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,563 @@
...
+  name: "Free",
+  type: "subscription",
+};
+
+const NON_FREE_BILLING_PLAN: BillingPlan = {
+  description: "Non Free Billing Plan",
+  id: "non-free",
+  name: "Non free",
+  type: "subscription",
+};
```
romain-growthbook, line 47: There are place holders.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,563 @@
...
+  });
+
+  const context = new ReqContextClass({
+    org,
+    auditUser: null,
+    user,
+  });
+
+  // The model is attached to the top-level org so we
+  // need a different contect here!
```
romain-growthbook, line 201:
```suggestion
  // need a different context here!
```

packages/back-end/src/services/vercel-native-integration.service.ts
```diff
@@ -0,0 +1,372 @@
...
+    return;
+  }
+
+  const sdks = sdkConnections.map(({ id }) => id);
+
+  const webhookParams = {
+    name: "Sync vercel integration edge config",
+    endpoint: `${VERCEL_URL}/v1/installations/${nativeIntegration.installationId}/resources/${resource.id}/experimentation/edge-config`,
+    payloadFormat: "vercelNativeIntegration",
+    payloadKey: "gb_payload",
```
bryce-fitzsimons, line 359: Should this be the connection clientKey?

packages/back-end/src/app.ts
```diff
@@ -309,6 +310,27 @@ const origins: (string | RegExp)[] = [APP_ORIGIN];
 if (CORS_ORIGIN_REGEX) {
   origins.push(CORS_ORIGIN_REGEX);
 }
+
+if (process.env.NODE_ENV === "development" || IS_CLOUD) {
+  app.use(
+    "/vercel",
+    cors({
+      credentials: false,
+      origin: "*",
```
mknowlton89, line 319: Should we define the origin here rather than using a wildcard?

Something like:

```ts
app.use(
  "/vercel",
  cors({
    credentials: false,
    origin: process.env.NODE_ENV === "production" 
      ? ["https://vercel.com", "https://*.vercel.com"] 
      : "*",
  }),
  vercelRouter
);
```

packages/back-end/src/services/organizations.ts
```diff
@@ -103,6 +103,10 @@ export function validateLoginMethod(
   org: OrganizationInterface,
   req: AuthRequest
 ) {
+  // If organization is vercel intergration and no loginMethod is provided, add it.
+  if (org.isVercelIntegration && !req.loginMethod)
+    req.loginMethod = { id: "vercel" };
```
mknowlton89, line 108: My linter is complaining and wanting us to also add  `clientId` and `metaData` properties here.

packages/back-end/types/organization.d.ts
```diff
@@ -297,6 +296,7 @@ export interface OrganizationInterface {
   stripeCustomerId?: string;
   restrictLoginMethod?: string;
   restrictAuthSubPrefix?: string;
+  isVercelIntegration?: boolean;
```
mknowlton89, line 299: Would it be possible to remove this and instead add it as an optional property to `ReqContext` - where the value is based on whether or not the org has a record in the `vercelNativeIntegration` collection? 

The thought is it could make clean up of the org easier if they remove the integration - as if the org uninstalls the integration, we delete the record from Mongo and this value would be automatically updated.

I supposed we'd still need to handle user access and authentication as we'd need to allow them to login with email/password like a normal account - so we'd need to change the `restrictLoginMethod` when this happens, so might not yield much benefit.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
mknowlton89:
```ts
const STARTER_BILLING_PLAN: BillingPlan = {
  description:
    "Growthbook's free plan. Add up to 3 users. Enjoy unlimited feature flag evaluations, community support, up to 1M CDN requests/month, and up to 5GB of CDN Bandwidth/month.",
  id: "starter-billing-plan",
  name: "Starter (Free) Plan",
  type: "subscription",
};

const PRO_BILLING_PLAN: BillingPlan = {
  description:
    "All the benefits of our starter plan, plus add up to 100 members (each member incurs a cost of $20/month), get in-app chat support, access to advance experimentation, up to 2M CDN requests/month, and up to 20GB of CDN Bandwidth/month.",
  id: "pro-billing-plan",
  name: "Pro Plan",
  type: "subscription",
  cost: "20.00/month per user + usage",
};
```
Here is what I have in my other PR, FWIW

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+  // We don't support installation-level billing plans!
+  return res.status(400).send("Invalid request!");
+}
+
+export async function deleteInstallation(req: Request, res: Response) {
+  const { integrationModel, integration } = await authContext(req, res);
+
+  await integrationModel.deleteById(integration.id);
+
+  return res.status(200).send({ finalized: true });
```
mknowlton89, line 332: I think we may want to leave off the `finalized: true` for production so if the org does have an outstanding invoice, they can work on issuing it.

mknowlton89 (commented):
Looking good - great work on this.

Left just a few comments - I initially focused on just the integration code (around the controller, router, etc).

I'm not sure I have enough context to provide meaningful feedback on how this works with SDKs and our webhooks.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
mknowlton89: For additional context, when a user uninstalls the integration, we cancel their subscription and do so immediately. (As opposed to normal subscriptions where the org has access through the end of their term.) When this is done, we'll calculate any usage they have for the month and immediately process an invoice. If we set `{finalized: true}`, I don't think we'll have the ability to send the final invoice to Vercel to collect.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+      ...organization.members,
+      {
+        id,
+        role: "admin",
+        dateCreated: new Date(),
+        limitAccessByEnvironment: false,
+        environments: [],
+      },
+    ],
+  });
```
jdorn, line 118: This is missing a bunch of stuff.  Better to use the `addMemberToOrg` method.  This is how we do it for other SSO connections:

```ts
  await addMemberToOrg({
    organization,
    userId,
    ...getDefaultRole(organization),
  });
  try {
    await sendNewMemberEmail(
      userName || "",
      userEmail || "",
      organization.name,
      organization.ownerEmail
    );
  } catch (e) {
    req.log.error(e, "Failed to send new member email");
  }
```

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+
+  const organizationId = resource?.organizationId || integration.organization;
+  const org = await findOrganizationById(organizationId);
+
+  if (!org) return failed(400, "Invalid installation!");
+
+  const user = await findOrCreateUser({
+    email: userEmail || org.ownerEmail,
+    organization: org,
+  });
```
jdorn, line 191: One potential issue here: if the user was just added to the org, this local `org` object will not have the user in the `members` array yet.  That will likely affect the permissions available to the context object.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+  // The model is attached to the top-level org so we
+  // need a different context here!
+  const topLevelOrg = await findOrganizationById(integration.organization);
+  if (!topLevelOrg) return failed(400, "Invalid installation!");
+
+  const topLevelContext = new ReqContextClass({
+    org: topLevelOrg,
+    auditUser: null,
+    user,
+  });
```
jdorn, line 208: Another potential issue: the user might be part of a child organization, but not the parent one, so again permissions here might be messed up.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+  const user = await findOrCreateUser({
+    email: userEmail || org.ownerEmail,
+    organization: org,
+  });
+
+  const context = new ReqContextClass({
+    org,
+    auditUser: null,
+    user,
+  });
```
jdorn, line 197: For both of these, it's missing a couple fields.  It looks like you might have copied from `getContextForAgendaJobByOrgObject`, but that's when we don't know who the user is.  In this case, I think it's better to use `getContextFromReq` as a base.

Specifically, we want to specify:
- auditUser (type = "dashboard")
- teams

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+  await authContext(req, res);
+  return res.json();
+}
+
+export async function updateInstallation(req: Request, res: Response) {
+  // We don't support installation-level billing plans!
+  return res.status(400).send("Invalid request!");
+}
+
+export async function deleteInstallation(req: Request, res: Response) {
```
jdorn, line 327: Should we lock down all of these API endpoints to auth token `type = "system"`?  Otherwise, couldn't any user in the org (even a readonly one) click the "Open in GrowthBook" link in vercel, capture their token, and use that to hit these endpoints?

packages/back-end/src/app.ts
romain-growthbook: It's a good idea to tighten up the security but given that the documentation does not make any mention of the calling host, I'm a little hesitant to make assumptions that vercel may feel free to break at any time.

packages/back-end/src/services/organizations.ts
romain-growthbook: yes I left on purpose b/c it's a little wonky over there, I wanted to revisit with some input from @jdorn

packages/back-end/types/organization.d.ts
romain-growthbook: This is gonna be tricky. `ReqConext` actually requires an org.. And as you noted, we do need to clear up other field in the org when it gets out of the vercel integration context.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
romain-growthbook: @mknowlton89 Can I leave this change to you for when the billing logic is in place?

packages/back-end/src/services/organizations.ts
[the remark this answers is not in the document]
romain-growthbook: Ok.

packages/front-end/pages/vercel.tsx
```diff
@@ -0,0 +1,47 @@
...
+    if (!code || !state) return;
+
+    const fn = async () => {
+      try {
+        const ret = await fetch(
+          `${getApiHost()}/auth/sso/vercel?code=${code}&state=${state}&resourceId=${resourceId}`,
+          {
+            method: "POST",
+            credentials: "include",
+            body: JSON.stringify({ code, state, resourceId }),
```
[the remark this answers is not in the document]
romain-growthbook: Refactorization left over.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,531 @@
...
+  id: "non-free",
+  name: "Non free",
+  type: "subscription",
+};
+
+const billingPlans = [FREE_BILLING_PLAN, NON_FREE_BILLING_PLAN] as const;
+
+const VERCEL_JKWS_URL = "https://marketplace.vercel.com/.well-known/jwks";
+
+const vercelJKWSKey = createRemoteJWKSet(new URL(VERCEL_JKWS_URL));
```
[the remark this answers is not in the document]
romain-growthbook: yes. I tried a lazy load also here before and it was giving me some weird errors. I think that it's fine for now, to be maybe revisited when we need to?

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
romain-growthbook: Not sure if I know this concept but in this specific case, all organizations are created via vercel so I don't think that they can be nested. Also, this context is really just used to pass around to create models and I don't think the permissions reauirements will get really deep/intricate, it's really just updating the native integration, creating webhook, that kind of stuff.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
romain-growthbook: We don't have `getContextFromReq` because these queries come from vercel with their custom authentication SSO auth.

We can add `auditUser` but this could be misleading as this is not exactly the user making the request but in fact vercel.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
romain-growthbook: Wether `user` or `system`, the JWT token is always checked against vercel private key, it's only the payload that differs:

```typescript
const checkAuth = async <T extends string | "user">({
  token,
  type,
}: {
  token: string;
  type: T;
}): Promise<CheckedAuth<T> | { status: "error"; message: string }> => {
  try {
    const payload = await jwtVerify(token, await vercelJKWSKey);

    if (type === "user")
      return {
        status: "authenticated",
        authentication: userAuthenticationValidator.parse(payload),
      } as CheckedAuth<T>;

    if (type === "system")
      return {
        status: "authenticated",
        authentication: systemAuthenticationValidator.parse(payload),
      } as CheckedAuth<T>;

    return { status: "error", message: `Unsupported authentication: ${type}` };
  } catch (_err) {
    return { status: "error", message: "Invalid credentials" };
  }
};
```

The described attack would mean that the attacker has full control of the vercel private key in which case, `system` or `user` mode wouldn't matter.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
mknowlton89: @romain-growthbook Yup - sounds like a plan to me.

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -268,16 +270,23 @@ export async function findOrganizationById(id: string) {
   return doc ? toInterface(doc) : null;
 }
 
+type UpdatableKeys = Exclude<
+  keyof OrganizationInterface,
+  "id" | "dateCreated" | "dateUpdated"
+>;
```
jdorn, line 276: 1. Rename this type so it's clear these are the keys that can be removed, not the ones that can be updated
2. This list of keys is way too broad.  We never want to remove `settings`, or `name`, or `ownerEmail`, etc.  Instead of Exclude, I would just make this a short allow list with just the minimum set of keys required for this PR.  We can always expand it later if we need to.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -76,6 +82,20 @@ export class ProjectModel extends BaseClass {
...
+      {
+        organization: this.context.org.id,
+        managedBy,
+      },
+      {
+        $unset: {
+          managedBy: 1,
+        },
+      }
+    );
```
jdorn, line 99: I think it's ok in this case, but just want to note this does not do any permission checks and will not create an audit log entry.  If we ever expose this functionality outside of validated vercel routes, we should definitely fix this.

packages/back-end/src/services/vercel-native-integration.service.ts
```diff
@@ -0,0 +1,418 @@
...
+
+  if (!org) throw new Error("Internal error: no org found");
+
+  if (!org.isVercelIntegration) return;
+
+  const nativeIntegration = await findVercelInstallationByOrganization(org.id);
+
+  const user = await getUserByEmail(
+    nativeIntegration.upsertData.authentication.user_email
+  );
```
jdorn: Don't get the user like this. We don't need a user at all for this function.  Just create a request context with admin scope.

packages/back-end/src/services/vercel-native-integration.service.ts
```diff
@@ -0,0 +1,418 @@
...
+            endpoint: `${VERCEL_URL}/v1/installations/${nativeIntegration.installationId}/resources/${resource.id}/experimentation/edge-config`,
+            payloadFormat: "vercelNativeIntegration",
+            payloadKey: sdkConnection.key,
+            httpMethod: "PUT",
+            managedBy: {
+              type: "vercel",
+              resourceId: resource.id,
+            },
+            headers: JSON.stringify({
+              Authorization: `Bearer ${nativeIntegration.upsertData.payload.credentials.access_token}`,
```
jdorn: Do not hard-code the access token here.  Use the new webhook secrets to store this instead (can do this when the installation is first created so we don't need to do anything special here).

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -240,6 +255,7 @@ export async function createSDKConnection(params: CreateSDKConnectionParams) {
 
   if (IS_CLOUD) {
     await addCloudSDKMapping(connection);
+    await syncVercelSdkConnection(connection.organization);
```
jdorn: I think we should remove the `syncVercelSdkConnection` method from this file entirely.  We only want to do this when a resource is provisioned or upserted in the vercel controller, so it should be called there instead.  The code here fires for every SDK connection, not just the ones managed by vercel.

packages/back-end/src/services/organizations.ts
```diff
@@ -971,17 +986,37 @@ export async function addMemberFromSSOConnection(
...
+      if (installationId !== req.vercelInstallationId) return null;
+
+      const installation = await findVercelInstallationByInstallationId(
+        installationId
+      );
+      if (!installation) {
+        return null;
+      }
+      organization = await findOrganizationById(installation.organization);
+    }
```
jdorn, line 1023: The 4 lines below this need to be inside an `else` block.  Otherwise, the code here for vercel won't do anything (`organization` immediately gets overridden).

packages/front-end/pages/vercel.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+        if (!ret.ok) throw new Error(`Request failed: ${await ret.text()}`);
+
+        const { organizationId, projectId } = await ret.json();
+
+        setProject(projectId);
+        router.push(`/?org=${organizationId}`);
+      } catch (err) {
+        console.log("Ignored:", err);
+        router.push("/");
```
jdorn: Feels like we should just render this error on the page instead of redirecting to `/`.

packages/front-end/pages/vercel.tsx
```diff
@@ -0,0 +1,47 @@
+import { useEffect } from "react";
+import { useRouter } from "next/router";
+import { getApiHost } from "@/services/env";
+import { useProject } from "@/services/DefinitionsContext";
+
+const VercelPage = () => {
```
jdorn: Just a nit, but do we want to nest this page inside the `oauth` directory?  We do the same for other auth callback pages we have.  Just thinking we should leave the top-level `/vercel` route open in case we want to use it for something else in the future.

packages/back-end/src/models/SdkConnectionModel.ts
romain-growthbook: yeah this is a left over for when we wanted to support all connections in the vercel edge payload.

packages/back-end/src/services/vercel-native-integration.service.ts
romain-growthbook: I'll do that but this'll require some change in the webhook secrets. Currently we don't track to which webhook or sdk connections they're attached to and we do want to be able to clean/update them programatically.

bryce-fitzsimons:
Reminder: update webhooks docs to change config key to client key instead of gb_payload:
https://docs.growthbook.io/app/webhooks/sdk-webhooks#vercel-edge-config

packages/back-end/src/models/WebhookSecretModel.ts
```diff
@@ -47,6 +47,10 @@ export class WebhookSecretDataModel extends BaseClass {
     });
   }
 
+  public async findByKey(key: string) {
+    return this._findOne({ key });
+  }
```
jdorn, line 54: I'd prefer not exposing secret values outside of the model.  It looks like this is only used to check if a key exists so we can delete it.  Instead, let's just make a `deleteByKey` method that does this internally.

packages/back-end/src/jobs/sdkWebhooks.ts
```diff
@@ -320,20 +328,27 @@ export async function fireSdkWebhook(
 ) {
   const webhookContext = getContextForAgendaJobByOrgObject(context.org);
 
+  const sendPayload =
+    webhook.httpMethod !== "GET" &&
+    sendPayloadFormats.includes(webhook.payloadFormat ?? "standard");
+
+  if (!sendPayload) return;
```
jdorn: This will break.

On main, if `sendPayload` is false, it still sends the webhook, it just sets the payload to an empty string.  This PR skips sending the webhook entirely, which is wrong.

romain-growthbook: I can change. What's the purpose of sending an empty payload?

packages/back-end/src/models/VercelNativeIntegrationModel.ts
```diff
@@ -0,0 +1,105 @@
...
+  teamId: z.string(),
+});
+
+export type Resource = z.infer<typeof installationResourceValidator>;
+
+const vercelNativeIntegrationValidator = z
+  .object({
+    id: z.string(),
+    // All installation need at least one org.
+    // Each additional resources map to a different org.
```
jdorn, line 29: Nit: remove comment, no longer accurate

packages/front-end/pages/sdks/SdkWebhooks.tsx
```diff
@@ -68,21 +70,48 @@ export default function SdkWebhooks({
...
+          {webhook.name}
+          {webhook.managedBy?.type ? (
+            <div>
+              <Badge
+                label={`Managed by ${capitalizeFirstLetter(
+                  webhook.managedBy.type
+                )}`}
+              />
+            </div>
+          ) : null}
```
jdorn, line 83: Nit: We should abstract this badge to its own component since it is copy/pasted a bunch of times in here.

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
```diff
@@ -0,0 +1,601 @@
...
+    payload.account.contact.email,
+    payload.account.name
+  );
+
+  const org = await createOrganization({
+    email: payload.account.contact.email,
+    userId: user.id,
+    name:
+      payload.account.name ||
+      `Vercel instalation ${authentication.payload.installation_id}`,
```
mknowlton89:
```suggestion
      `Vercel installation ${authentication.payload.installation_id}`,
```

installation typo

jdorn (approved):

PR #3971, opened 2025-04-22
REST API: Enable easier result exporting

lukesonnet:
### Features and Changes

This makes it easy to get out a cleaned, almost completely denormalized, dataset like experiment result (main missing issue is that the baseline variation values are on a different row from the test variation values and effects).

Future work will use the new experiment rows type to write results back to a data warehouse.

### Testing

Tested on local and got back cleaned results when results existed, otherwise got back an empty array

### DIFF
--- a/packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
+++ b/packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
@@ -9,9 +9,14 @@ required:
   - metricName
   - metricId
   - baselineVariationId
+  - baselineVariationUnits
+  - baselineVariationNumerator
+  - baselineVariationDenominator
+  - baselineVariationMean
+  - baselineVariationStdDev
   - variationName
   - variationId
-  - variationUsers
+  - variationUnits
   - variationNumerator
   - variationDenominator
   - variationMean
@@ -21,11 +26,11 @@ required:
   - chanceToBeatControl
   - relativeRisk
   - pValue
-  - pValueAdjusted
+  - pValueCorrected
   - ciLower
   - ciUpper
-  - ciLowerAdjusted
-  - ciUpperAdjusted
+  - ciLowerCorrected
+  - ciUpperCorrected
 properties:
   experimentName:
     type: string
@@ -67,9 +72,9 @@ properties:
     type: string
     description: "The id of the variation that is the baseline for this set of results."
     nullable: true
-  baselineVariationUsers:
+  baselineVariationUnits:
     type: number
-    description: "Number of users in the baseline variation. This is the denominator for non-ratio metrics."
+    description: "Number of units (e.g. users) in the baseline variation. This is the denominator for non-ratio metrics."
     nullable: true
   baselineVariationNumerator:
     type: number
@@ -95,9 +100,9 @@ properties:
     type: string
     description: "The unique identifier of the variation being compared to the baseline."
     nullable: true
-  variationUsers:
+  variationUnits:
     type: number
-    description: "Number of users in the variation being compared. This is the denominator for non-ratio metrics."
+    description: "Number of units (e.g. users) in the variation being compared. This is the denominator for non-ratio metrics."
   variationNumerator:
     type: number
     description: "Metric numerator value for the variation being compared."
@@ -135,9 +140,9 @@ properties:
     type: number
     description: "The p-value for the statistical test comparing the variations. Only present if the Frequentist engine is being used."
     nullable: true
-  pValueAdjusted:
+  pValueCorrected:
     type: number
-    description: "The p-value adjusted for multiple comparisons. Only present if multiple comparisons corrections are turned on and the Frequentist engine is being used."
+    description: "The p-value corrected for multiple comparisons. Only present if multiple comparisons corrections are turned on and the Frequentist engine is being used."
     nullable: true
   ciLower:
     type: number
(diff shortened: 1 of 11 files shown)


packages/back-end/src/api/openapi/paths/getExperimentResults.yaml
```diff
@@ -29,4 +29,9 @@ get:
               - experiment
             properties:
               result:
-                $ref: "../schemas/ExperimentResults.yaml"
\ No newline at end of file
+                $ref: "../schemas/ExperimentResults.yaml"
+              cleanedResults:
+                type: array
+                description: "The results for the experiment de-normalized. Each element in this array represents the results for a single variation on a single metric. If a dimension was requested, then each element will be the results for a single value of the dimension. If multiple difference types (e.g. relative, absolute, scaled impact) were in the results, then each element will contain the results for a single difference type. Therefore, each element in this array will be the results for a single variation on a single metric for a single difference type for a single value of the dimension (which will be null for the overall results)."
```
jdorn, line 35: This wording is pretty hard to parse.  Here's my attempt at a simpler version:
> The results for the experiment de-normalized with no nesting.  One row per variation, metric, dimension value, and difference type.

Difference types are really weird here.  It feels wrong that the return value of the API should depend on what someone happened to click on in the UI.  Makes it really hard to plan around and use the data if the specific rows returned are so unpredictable.

To fix this, I would probably recommend defaulting to only returning `relative`, since (at least as of now) that will always be available.  Then add a querystring to opt-in to including additional difference types (e.g. `?differenceTypes=relative,absolute`.  If the analysis for a requested difference type is missing, we can calculate it and store in the snapshot first before returning.  We should benchmark this, but I think it should be pretty fast.  That way, the user will always get back exactly what they request.

packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
```diff
@@ -0,0 +1,157 @@
...
+  - variationNumerator
+  - variationDenominator
+  - variationMean
+  - variationStdDev
+  - differenceType
+  - effect
+  - chanceToBeatControl
+  - relativeRisk
+  - pValue
+  - pValueAdjusted
```
jdorn, line 24: In the product, we use `adjustment` to refer to CUPED and `correction` to refer to multiple-comparisons.  So should this be `pValueCorrected` instead?

Speaking of which, how does this work for CUPED, Sequential, and Informative Priors?  Are we just including the final p-value / CTW / CIs?  Is there any way to tell what type of results are going to be returned?

I'm thinking about someone dumping all experiment results into a single SQL table in order to do a meta-analysis.  The rows should either be standardized and comparable, or have additional columns that allow for filtering (`statsEngine`, `regressionAdjustment`, `pValueConnectionMethod`, etc.)

packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
```diff
@@ -0,0 +1,157 @@
+type: object
+required:
```
jdorn, line 2: Should we add a statsEngine field instead of relying on certain fields like pValue being null as the only way to differentiate?

packages/back-end/src/services/experimentResults.ts
```diff
@@ -0,0 +1,130 @@
...
+      dimension
+    : null;
+
+  const rows: ExperimentResultRow[] = [];
+
+  snapshot.analyses.forEach((analysis) => {
+    // skip analyses where the control variation is not the baseline variation
+    if (analysis.settings.baselineVariationIndex !== 0) {
+      return;
+    }
```
jdorn, line 69: Would CUPED cause duplicate rows?  I forget if we store CUPED on/off as 2 separate analyses or not.  If they are separate analyses, I think that would result in duplicate copies of every row.  Are there any other analysis settings like that we should watch out for?

packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
```diff
@@ -0,0 +1,157 @@
...
+  - differenceType
+  - effect
+  - chanceToBeatControl
+  - relativeRisk
+  - pValue
+  - pValueAdjusted
+  - ciLower
+  - ciUpper
+  - ciLowerAdjusted
+  - ciUpperAdjusted
```
jdorn, line 28: Probably out of scope here, but how are you thinking about health tab data?  If I'm going to be doing a meta-analysis across a bunch of experiments, I'm definitely going to want to be able to filter on ones with/without an SRM or multiple exposures error.

packages/back-end/src/api/openapi/paths/getExperimentResults.yaml
lukesonnet: > It feels wrong that the return value of the API should depend on what someone happened to click on in the UI.

We now automatically compute all 3 difference types. In the future we could auto-compute pre-specified dimensions and then choose whether or not to return them.

packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
lukesonnet: >So should this be pValueCorrected instead?

Yes, we could, although the language is a bit weird. I think it's better on average though.

>Speaking of which, how does this work for CUPED, Sequential, and Informative Priors? Are we just including the final p-value / CTW / CIs?

Just the final values. I was debating exporting the snapshot settings along with the results but thought that was too much data in one row. We could also tell users to pull this data from the snapshot collection but I think that's actually not exposed right now and we could handle that for them.

packages/back-end/src/services/experimentResults.ts
lukesonnet: >I forget if we store CUPED on/off as 2 separate analyses or not

We don't right now. I want to hook that up at some point, but it's a bit complicated as the CUPED logic is squirrely and if you toggle it off in the UI it actually turns off the setting in the analysis form.

There are no other duplicates because right now the only thing that varies across analyses are `differenceType` and `baselineVariationIndex`.

packages/back-end/src/api/openapi/schemas/ExperimentResultsRow.yaml
lukesonnet: I wasn't considering it, but yeah we could also include the SRM p-value and SRM status as two columns, as well as number of multiple exposures.

PR #3972, opened 2025-04-22
Ensure p-value adjustment applied to analysisSummary

lukesonnet:
### Features and Changes

pValue adjustment needs to be applied on the back-end when we compute analysis summaries for the decision making framework.

### Testing

Adjusted p-values were not being used before when setting analysis summary, now they are.

We should probably just save the adjusted p-values.


ahdriel (approved):
Just some CI errors but changes LGTM.

PR #3973, opened 2025-04-23
Mk/fact table promo

mknowlton89:
### Features and Changes


packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -400,6 +411,42 @@ const MetricsList = (): React.ReactElement => {
           <TagsFilter filter={tagsFilter} items={unpaginatedItems} />
         </div>
       </div>
+      {/* only show fact table promo if org doesn't have any AND they haven't previously dismissed the callout */}
+      {!factTables && !factTablePromoDismissed ? (
```
jdorn, line 415: `factTables` is an array, so I don't think this will ever be true?

packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -225,10 +230,16 @@ const MetricsList = (): React.ReactElement => {
...
+    metricGroups,
     project,
     ready,
+    factTables,
   } = useDefinitions();
   const { getUserDisplay } = useUser();
+  const [factTablePromoDismissed] = useLocalStorage(
+    "premium-callout:metrics-list-fact-table-promo",
+    false
+  );
```
jdorn, line 242: This will not work very well since values are read once from localStorage, but not kept in sync after.  So if you dismiss the fact table banner, I think this will still be false until you refresh the page.  Maybe that behavior is fine here, but just calling it out in case you weren't aware.

I also don't love relying on the special key formatting outside of the callout component.

mknowlton89: @jdorn Yea, definitely an oddball pattern.

In my testing, dismissing the banner did result it in immediately not rendering as seen [here](https://www.loom.com/share/fc7e34b7752e403cb5a0571338b17b2a).

I definitely see what you mean about not liking having to rely on the special key formatting outside of the component. Is there another pattern you would suggest to hoist the state out of the child?

PR #3978, opened 2025-04-24
Safe Rollout Status Badge, Help Text, and CTA

jdorn:
### Features and Changes

Replace safe rollout decision banner with a badge, help text, and CTA more integrated into the rule UI.

image


packages/front-end/components/SafeRollout/DecisionHelpText.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return (
+      <HelperText status="info" mb="3">
+        All users will receive the control value. If this is no longer needed,
+        you can disable or delete this rule.
+      </HelperText>
+    );
+  } else if (safeRollout.status === "released") {
+    return (
+      <HelperText status="info" mb="3">
+        All users will receive the variation value. If this is no longer needed,
```
ahdriel, line 58: nit: We refer to it as `Rollout` value in the `SERVE` section, should we also use the same here?

packages/front-end/components/SafeRollout/DecisionHelpText.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <HelperText status="warning" mb="3">
+        Your safe rollout has been running for over 24 hours and no traffic has
+        come in. Check your implementation and ensure that the tracking callback
+        is firing as expected.
+      </HelperText>
+    );
+  } else if (decisionStatus.status === "unhealthy") {
+    return (
+      <HelperText status="warning" mb="3">
+        {decisionStatus.unhealthyData.srm
```
ahdriel, line 88: nit: The test can be unhealthy for multiple reasons but we are going to render only one here.

ahdriel (approved):

packages/front-end/components/SafeRollout/DecisionCTA.tsx
```diff
@@ -0,0 +1,81 @@
...
+    variant = "solid";
+  }
+
+  return (
+    <Button
+      type="button"
+      onClick={openStatusModal}
+      variant={variant}
+      size="sm"
+      color={color}
```
msamper, line 74: Are we okay with the button becoming red? Kari has some concerns with red since it indicates a destructive action usually, but maybe `Reverting` is destructive in this context so its fine

packages/front-end/components/SafeRollout/SafeRolloutStatusBadge.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  if (safeRollout.status === "rolled-back") {
+    color = "red";
+    label = "Reverted";
+  } else if (safeRollout.status === "released") {
+    color = "green";
+    label = "Released";
+  } else if (decisionStatus?.status === "no-data") {
+    color = "amber";
+    label = "No Data";
```
msamper, line 50: `no-data` covers the state you asked me to add where there's no results in over 24 hours. I think you mentioned wanting it to be an `unhealthy` state so this would probably be the place to do it (i.e. `status === "no-data" || status === "unhealthy"` -> "Unhealthy") if you still think that makes sense. FWIW I think no data as the status is fine.

PR #3981, opened 2025-04-24
Safe Rollout to now include fact metrics

gazzdingo:
### Features and Changes
- Metrics now are showing fact metrics
- identifier type is now added to the experiment assignment table field
- if they only have 1 datasource it will pre-fill the the field with it or if there is a default
-  

### DIFF
--- a/packages/front-end/services/features.ts
+++ b/packages/front-end/services/features.ts
@@ -35,6 +35,7 @@ import isEqual from "lodash/isEqual";
 import { ExperimentLaunchChecklistInterface } from "back-end/types/experimentLaunchChecklist";
 import { SavedGroupInterface } from "shared/src/types";
 import { SafeRolloutRule } from "back-end/src/validators/features";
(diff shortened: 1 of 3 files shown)


packages/front-end/services/features.ts
```diff
@@ -676,7 +679,10 @@ export function getDefaultRuleValue({
   const hashAttribute = hashAttributes.includes("id")
     ? "id"
     : hashAttributes[0] || "id";
-
+  let defaultDataSource = settings?.defaultDataSource;
+  if (datasources && !defaultDataSource && datasources.length === 1) {
```
lukesonnet, line 683: I think @jdorn wanted to just default to picking the first one even if the length was > 1

jdorn: I'm fine with this for now. Less risky

lukesonnet (approved):

msamper (approved):

PR #3983, opened 2025-04-24
Add metrics in metric groups to override

lukesonnet:
Metric groups were not being expanded for metric overrides. This fixes that.


ahdriel (approved):
Nice catch!

PR #3987, opened 2025-04-25
don't allow cta on locked revision

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Features/Rule.tsx
+++ b/packages/front-end/components/Features/Rule.tsx
@@ -253,7 +253,7 @@ export const Rule = forwardRef<HTMLDivElement, RuleProps>(
                         <Flex gap="3" align="center">
                           <div>Safe Rollout</div>
                           <SafeRolloutStatusBadge rule={rule} />
-                          {rule.enabled !== false && (
+                          {!locked && rule.enabled !== false && (
                             <div className="ml-auto">
                               <DecisionCTA
                                 rule={rule}


ahdriel (approved):
Nice! Looks good to me -- locked is defined as below:

```
const isLocked =
    (revision.status === "published" || revision.status === "discarded") &&
    (!isLive || drafts.length > 0);
```

packages/front-end/components/Features/Rule.tsx
```diff
@@ -253,7 +253,7 @@ export const Rule = forwardRef<HTMLDivElement, RuleProps>(
                         <Flex gap="3" align="center">
                           <div>Safe Rollout</div>
                           <SafeRolloutStatusBadge rule={rule} />
-                          {rule.enabled !== false && (
+                          {!locked && rule.enabled !== false && (
```
jdorn, line 256: This will still show the CTA for drafts?  If so, I don't think it solves the bug.

The status endpoint currently will re-use an existing draft if there is one.   That is bad since when you submit the status modal it auto-publishes the draft.  That could mean unintentional changes could be released along with the status change you are doing.  The CTA should always create a brand new draft with only those changes.

jdorn: To clarify, the fix here is fine.  We just need to also change the back-end status endpoint to never re-use an existing draft.

gazzdingo: sorry i think i miss understood what the bug here, i can update the endpoint to create a new revision, i think we should only create a new revision if it is going to auto publish WDYT?

gazzdingo: or should we create a new revision no matter what

msamper (approved):
LGTM

PR #3988, opened 2025-04-25
Initial pass at Enterprise upgrade modal tweaks

mknowlton89:
### Features and Changes

This PR updates the `UpgradeModal`, specifically when an org is clicks the Upgrade CTA while they're already a `Pro` subscriber, or if the org clicks on the Upgrade Modal from a enterprise-only feature (e.g. Metric Groups).

Previously, we were still using the old style two column upgrade modal. Now, we're using the same Upgrade Modal style that we are for Pro, but the content is specific to `Enterprise` features/benefits.

This allows us to simplify the code around this modal pretty substantially - I do have a follow-up PR planned for this file that removes all references to the Free Trials, now that we're not doing those anymore. However, I'd like to do that work after this PR and in it's own PR so that if we ever want to add it back, it could make it easier to do so. As a result, you'll noticed some `es-lint ignore` comments - those will be addressed in the follow up PR.

### Dependencies

None

### Testing

- [x] Ensure that when clicking an upgrade from a `Enterprise-only` feature (e.g. Metric Groups) - We render the Enterprise modal and the CTA directs users to https://www.growthbook.io/demo
- [x] Ensure that if an org is a `Starter` org and clicks on an Pro feature (or the Upgrade) button on the /getstarted page we render the "Pro" upgrade modal
- [x] Ensure that if an org is a `Pro` org and clicks on an Enterprise feature (or the Upgrade) button on the /getstarted page we render the "Enterprise" upgrade modal
- [x] Ensure that if an org previously had an Enterprise license that has expired, we render the `notice` that calls this out.

### Screenshots
Screenshot 2025-04-28 at 1 32 52 PM
Screenshot 2025-04-28 at 1 33 13 PM


packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -80,37 +78,46 @@ export default function UpgradeModal({
...
+        effectiveAccountPlan &&
+        ["pro", "pro_sso"].includes(effectiveAccountPlan)
+      ) {
+        return "enterprise";
+      }
+    } else {
+      return commercialFeatureLowestPlan?.[commercialFeature] || "starter";
+    }
+  }
+
```
mknowlton89, line 101: Previously, if no `commercialFeature` was passed in, it would be possible for an `Enterprise` org to get the `Pro` upgrade modal.

Now, if no `commercialFeature` is passed in, the `lowestAvailablePlan` is the plan ` higher than their current plan.

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -378,6 +386,7 @@ export default function UpgradeModal({
     </p>
   );
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mknowlton89, line 389: I'll remove `trialAndUpgradeTreatment` in a follow up PR where I remove all trial flows

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -80,37 +78,46 @@ export default function UpgradeModal({
...
+      }
+
+      if (
+        effectiveAccountPlan &&
+        ["pro", "pro_sso"].includes(effectiveAccountPlan)
+      ) {
+        return "enterprise";
+      }
+    } else {
+      return commercialFeatureLowestPlan?.[commercialFeature] || "starter";
```
tzjames, line 98: A couple of questions about this line:
Is typescript making sure that `commercialFeature` is always a valid key of `commercialFeatureLowestPlan`?  
There is no "starter" plan modal, should we be defaulting to "pro"?
`starter` plan is only on cloud, while `oss` is for self-serve.  Though I don't think there is much difference in the logic for how we handle the two right now, we should probably default to the correct "free" version.

tzjames: Ok I see below that we just check to see if this equals "enterprise" below. What about renaming this function `showEnterpriseTreatment` or `showEnterpriseVersion` and have it return a boolean?

tzjames (approved):
Looks good.  Just some small naming suggestions.

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -664,239 +731,36 @@ export default function UpgradeModal({
...
-                        </div>
-                      )}
-                    </div>
-                  </div>
-                </div>
-              </div>
-            )}
+            {lowestPlan === "enterprise"
+              ? enterpriseTreatment()
+              : upgradeOnlyTreatment()}
```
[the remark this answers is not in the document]
mknowlton89: Agreed - I previously made that change, and then stashed the changes and never added it back. Doing so now.

packages/front-end/components/Settings/UpgradeModal/index.tsx
mknowlton89: I'm good with that change, I'll make it now.

PR #3990, opened 2025-04-25
fixed issue with revsions being created for features even if there are no changes

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/back-end/src/controllers/features.ts
+++ b/packages/back-end/src/controllers/features.ts
@@ -1873,15 +1873,25 @@ export async function putFeatureRule(
     }
   }
   let hasChanges = false;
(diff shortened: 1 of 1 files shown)


packages/back-end/src/controllers/features.ts
```diff
@@ -1873,7 +1872,27 @@ export async function putFeatureRule(
       });
     }
   }
-
+  let hasChanges = false;
+  const liveRule = feature.environmentSettings[environment].rules.find(
+    (r) => r.id === rule.id
+  );
```
jdorn, line 1878: I think this is wrong?  What if you are looking at a draft?  I think you'd want to compare against the existing draft to see if there are changes.

PR #4007, opened 2025-04-29
fix JWT auth race condition

itsgrimetime:
# Fix for intermittent 502 errors caused by JWT token expiration

## Problem

When JWT tokens expire in GrowthBook deployments (especially behind proxies like Cloudflare), users (or rest API clients) can experience intermittent 502 errors. This happens because:

1. When a token expires, the server returns a 401 Unauthorized response
2. The client attempts to refresh the token but abruptly closes the connection before completion
3. The proxy interprets this as a broken connection and returns a 502 error to the user (this is seen by our users as a "failed to refresh" modal at random times).

This seems to happen more frequently during times of high traffic, e.g. many requests from REST API clients paired with normal human usage of the app.

We noticed:
- "jwt expired" errors followed by 401 responses
- "Incoming request ended abruptly: context canceled" messages in proxy logs
- Successful token refresh operations that don't prevent subsequent failures

## Solution

This PR attempts to improve the token refresh mechanism in the `apiCall` function by:

1. Adding better error handling for token refresh operations with try/catch blocks
2. Preventing abrupt connection closures when authentication failures occur
3. Logging token refresh failures to aid future debugging
4. Maintaining consistent session error state for better user feedback

## Implementation Details

adds two levels of error handling:
1. An inner try/catch specifically around the token refresh operation
2. An outer try/catch for the entire API call to ensure graceful error handling

These additions aim to ensure that even when authentication fails, the connection is maintained long enough to properly handle the error, preventing the proxy from experiencing the "context canceled" condition that leads to 502 errors.

## Testing

This change has been tested in our self-hosted environment where 502 errors were previously occurring at relatively high rates (>5% at times of high traffic). The fix appears to successfully reduce (and maybe completely prevent) the abrupt connection termination and provides a better user experience when authentication issues arise.

### DIFF
--- a/packages/front-end/services/auth.tsx
+++ b/packages/front-end/services/auth.tsx
@@ -207,10 +207,8 @@ export const AuthProvider: React.FC<{
   const [token, setToken] = useState("");
   const [orgId, setOrgId] = useState<string | null>(null);
   const [organizations, setOrganizations] = useState<UserOrganizations>([]);
-  const [
-    specialOrg,
-    setSpecialOrg,
-  ] = useState<Partial<OrganizationInterface> | null>(null);
+  const [specialOrg, setSpecialOrg] =
+    useState<Partial<OrganizationInterface> | null>(null);
   const [authComponent, setAuthComponent] = useState<ReactElement | null>(null);
   const [initError, setInitError] = useState("");
   const [sessionError, setSessionError] = useState(false);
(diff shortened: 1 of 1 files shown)


jdorn:
I'm having a little trouble following this fix.  All of the `catch` blocks rethrow an error, so the main difference seems to be calling`setSessionError` in more places.  That will cause the app to render a `You've been logged out` modal instead of the page you were previously on.

The part I'm confused by is how that would fix the following:

> Preventing abrupt connection closures when authentication failures occur

The only thing I can think of is that by skipping rendering of the page, it prevents further API requests from being sent (that would also presumably fail).  So maybe Cloudflare is doing something like "if >x 401 errors in short period of time, assume the back-end is down and return 502"?  Some sort of automated spam / DDOS protection?

PR #4011, opened 2025-04-30
Initial helm chart

kosmoz:
### Features and Changes

Add initial version of a helm chart to deploy growthbook on Kubernetes

### Testing

Test with minikube/kind. The chart can be installed with helm (`helm install <releasename> .`).

### Screenshots
-

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -1,0 +1,28 @@
+on: push
+jobs:
+  build-chart:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: charts/growthbook
+    steps:
+      - uses: actions/checkout@v4
+      - uses: azure/setup-helm@v4
+      - name: Rebuild helm dependencies
+        run: helm dependency build
+      - name: Lint helm chart
+        run: helm lint --strict --with-subcharts .
+      - name: Package helm chart
+        run: helm package .
+      - name: Login to OCI registry
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+      - name: Push helm chart to OCI registry
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        run: |
+          for chart in *.tgz; do
+              helm push "$chart" oci://registry-1.docker.io/growthbook/charts
+          done
--- a/charts/growthbook/Chart.yaml
+++ b/charts/growthbook/Chart.yaml
@@ -1,0 +1,32 @@
+apiVersion: v2
+name: growthbook
+description: Open Source Feature Flagging and A/B Testing Platform
+icon: https://avatars.githubusercontent.com/u/65404594
+home: https://www.growthbook.io/
+sources:
+  - https://github.com/growthbook/growthbook/
+
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: "3.5.0"
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "3.5.0"
+
+dependencies:
+  - name: mongodb
+    repository: oci://registry-1.docker.io/bitnamicharts
+    version: "^16.5.0"
+    condition: mongodb.enabled
+  - name: server
+    version: "0.0.0"
+    alias: frontend
+  - name: server
+    version: "0.0.0"
+    alias: backend
--- a/charts/growthbook/templates/ingress.yaml
+++ b/charts/growthbook/templates/ingress.yaml
@@ -1,0 +1,43 @@
+{{- if .Values.ingress.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "growthbook.fullname" . }}
+  labels:
+    {{- include "growthbook.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with .Values.ingress.className }}
+  ingressClassName: {{ . }}
+  {{- end }}
+  {{- if .Values.ingress.tls }}
+  tls:
+    {{- range .Values.ingress.tls }}
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
+  rules:
+    {{- range .Values.ingress.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .paths }}
+          - path: {{ .path }}
+            {{- with .pathType }}
+            pathType: {{ . }}
+            {{- end }}
+            backend:
+              service:
+                name: {{ printf "%s-%s" $.Release.Name .service }}
+                port:
+                  number: {{ (index $.Values .service).service.port }}
+          {{- end }}
+    {{- end }}
+{{- end }}
--- a/charts/growthbook/values.yaml
+++ b/charts/growthbook/values.yaml
@@ -1,0 +1,172 @@
+# Default values for growthbook.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+global:
+  env:
+    - name: APP_ORIGIN
+      # full URL of your frontend, e.g. https://gb.example.com
+      value: https://growthbook.local
+
+frontend:
+  replicaCount: 1
+  podAnnotations: {}
+  podLabels: {}
+  podSecurityContext: {}
+  securityContext: {}
+  resources: {}
+  autoscaling:
+    enabled: false
+    minReplicas: 1
+    maxReplicas: 100
+    targetCPUUtilizationPercentage: 80
+    # targetMemoryUtilizationPercentage: 80
+  service:
+    port: 3000
+  command: [npx, wsrun, -p, front-end, --no-prefix, -c]
+  args:
+    - start
+  env:
+    - name: API_HOST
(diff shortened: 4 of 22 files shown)


docs/docs/self-host/index.mdx
```diff
@@ -72,3 +75,519 @@ docker-compose pull growthbook
...
+  </TabItem>
+</Tabs>
+
+### Accessing GrowthBook
+
+After deploying with Helm, access GrowthBook using the Ingress hostnames you configured in your values file:
+
+- **Minikube:**
+  - Run `minikube tunnel` in a separate terminal to expose the Ingress IP to your host.
+  - Frontend: `https://growthbook.${MINIKUBE-IP}.sslip.io`
```
tzjames, line 554: on macOS with Docker desktop and the docker VM Driver the minikube ip is only available within Docker's network namespace and requests to 127.0.0.1 got forwarded to minikube.  So I had to put a redirect in my etc/hosts:
```
echo "127.0.0.1 growthbook.192.168.49.2.sslip.io" | sudo tee -a /etc/hosts
echo "127.0.0.1 api.192.168.49.2.sslip.io" | sudo tee -a /etc/hosts
```

docs/docs/self-host/index.mdx
```diff
@@ -72,3 +75,519 @@ docker-compose pull growthbook
 docker-compose stop growthbook
 docker-compose up -d growthbook
 ```
+
+## Kubernetes (Helm) Installation
+
+:::tip
+Our official Helm chart is available on [Artifact Hub](https://artifacthub.io/packages/helm/one-acre-fund/growthbook) to deploy GrowthBook on Kubernetes:
```
tzjames: Shouldn't this link to GHCR instead as that is where we are deploying.  I think the one-acre-fund is the old one they had made.

docs/docs/self-host/index.mdx
```diff
@@ -72,3 +75,519 @@ docker-compose pull growthbook
...
+- For production and scalable cloud-native deployments, use GCP (GKE) or AWS (EKS).
+- For local testing and development, use Minikube.
+:::
+
+### Add the Helm Repository
+
+First, add the GrowthBook Helm repository to your local Helm installation:
+
+```bash
+helm repo add growthbook oci://ghcr.io/growthbook/charts/growhtbook
```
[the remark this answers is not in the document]
tzjames: Even if we fix the typo, I still don't think it will work.  I don't think `helm repo` works with oci based chart repos.  The install commands need to reference it directly as far as I can tell like 
`helm install growthbook oci://ghcr.io/growthbook/charts/growthbook -f values-minikube.yaml`

docs/docs/self-host/index.mdx
```diff
@@ -72,3 +75,519 @@ docker-compose pull growthbook
...
+      paths:
+        - path: /*
+          pathType: ImplementationSpecific
+          service: backend
+  tls: []
+```
+      </details>
+      Installation:
+      ```bash
+      helm install growthbook growthbook/growthbook -f values-aws.yaml
```
tzjames, line 225: i think this needs to be:
`helm install growthbook oci://ghcr.io/growthbook/charts/growthbook -f values-aws.yaml

docs/docs/self-host/index.mdx
```diff
@@ -72,3 +75,519 @@ docker-compose pull growthbook
...
+            service:
+              name: backend
+              port:
+                number: 3100
+  tls: []
+```
+      </details>
+      Installation:
+      ```bash
+      helm install growthbook growthbook/growthbook -f values-gcp.yaml
```
tzjames, line 329: and this should be: `helm install growthbook oci://ghcr.io/growthbook/charts/growthbook -f values-gcp.yaml`

docs/docs/self-host/index.mdx
```diff
@@ -72,3 +75,519 @@ docker-compose pull growthbook
...
+  tls:
+    - secretName: growthbook-tls
+      hosts:
+        - growthbook.${MINIKUBE-IP}.sslip.io
+        - api.${MINIKUBE-IP}.sslip.io
+```
+      </details>
+      Installation:
+      ```bash
+      helm install growthbook growthbook/growthbook -f values-minikube.yaml
```
tzjames: and this: `helm install growthbook oci://ghcr.io/growthbook/charts/growthbook -f values-minikube.yaml`

.github/workflows/chart.yaml
```diff
@@ -0,0 +1,32 @@
+on: push
```
tzjames: I think we will need to publish this sometimes as needed, like right after landing this change.  It would probably be good to also have a workflow dispatch option...
```
on: 
  push:
  workflow_dispatch:
    inputs:
      publish:
        description: 'Publish chart to OCI registry'
        required: false
        default: false
        type: boolean
```
and then changing the two ifs below to:
`if: ${{ startsWith(github.ref, 'refs/tags/') || inputs.publish }}`

charts/growthbook/Chart.yaml
```diff
@@ -0,0 +1,32 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: "3.5.0"
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "3.5.0"
```
tzjames: when we publish a new growthbook version, this seems like something we may forget to update.  What do you think about having the workflow automatically populate these values with the latest tag?

tzjames (changes requested):
Looks good.  At least I was able to git minikube to work with a few minor tweaks.

charts/growthbook/Chart.yaml
tzjames: At any rate we are at 4.1 now.

charts/growthbook/templates/ingress.yaml
```diff
@@ -0,0 +1,43 @@
...
+          {{- range .paths }}
+          - path: {{ .path }}
+            {{- with .pathType }}
+            pathType: {{ . }}
+            {{- end }}
+            backend:
+              service:
+                name: {{ printf "%s-%s" $.Release.Name .service }}
+                port:
+                  number: {{ (index $.Values .service).service.port }}
```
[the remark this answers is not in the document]
kosmoz: The ingress has separate rules for the backend and frontend services, which is what this lookup logic achieves. It works as intended.

charts/growthbook/values.yaml
```diff
@@ -0,0 +1,172 @@
...
+    - name: JWT_SECRET
+      valueFrom:
+        secretKeyRef:
+          name:
+          key:
+    - name: ENCRYPTION_KEY
+      valueFrom:
+        secretKeyRef:
+          name:
+          key:
```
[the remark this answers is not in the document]
kosmoz: Since the JWT_SECRET and ENCRYPTION_KEY variables are required for the application to function, I think it is desirable for the deployment to fail if they are not specified. I would therefore not change this.

docs/docs/self-host/index.mdx
kosmoz: Thanks, I added this to the documentation

docs/docs/self-host/index.mdx
kosmoz: Yes, updated

docs/docs/self-host/index.mdx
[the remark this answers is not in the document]
kosmoz: You're right, I removed this section and updated the instructions to be compatible with OCI charts

docs/docs/self-host/index.mdx
kosmoz: updated

docs/docs/self-host/index.mdx
kosmoz: updated

docs/docs/self-host/index.mdx
kosmoz: updated

.github/workflows/chart.yaml
kosmoz: added

charts/growthbook/Chart.yaml
kosmoz: IMO the best way is to use a release automation tool for this kind of updates. Personally, I like to use https://github.com/googleapis/release-please but there are others as well.

For now I updated the chart and app version to 4.1.0

.github/workflows/chart.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/checkout@v4
+      - uses: azure/setup-helm@v4
+      - name: Rebuild helm dependencies
+        run: helm dependency build
+      - name: Lint helm chart
+        run: helm lint --strict --with-subcharts .
+      - name: Package helm chart
+        run: helm package .
+      - name: Login to OCI registry
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
```
tzjames: needs an `|| inputs.publish` here as well.

docs/docs/self-host/index.mdx
```diff
@@ -564,6 +555,13 @@ After deploying with Helm, access GrowthBook using the Ingress hostnames you con
   - Backend/API: `https://api.${MINIKUBE-IP}.sslip.io`
   - (Replace `${MINIKUBE-IP}` with the output of `minikube ip`)
 
+  When running Minikube on MacOS with Docker Desktop, you may need to add some rules to your `/etc/hosts` file:
+
+  ```bash
+  echo "127.0.0.1 growthbook.192.168.49.2.sslip.io" | sudo tee -a /etc/hosts
+  echo "127.0.0.1 api.192.168.49.2.sslip.io" | sudo tee -a /etc/hosts
```
tzjames:
```suggestion
  echo "127.0.0.1 growthbook.${MINIKUBE-IP}.sslip.io" | sudo tee -a /etc/hosts
  echo "127.0.0.1 api.${MINIKUBE-IP}.sslip.io" | sudo tee -a /etc/hosts
```

charts/growthbook/values.yaml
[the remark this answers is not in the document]
tzjames: agreed

charts/growthbook/Chart.yaml
tzjames: thanks I'll look into that.

.github/workflows/chart.yaml
tzjames:
```suggestion
        if: ${{ startsWith(github.ref, 'refs/tags/') || inputs.publish }}
```

PR #4014, opened 2025-04-30
saferollout doesnt trigger draft changes if the rule is not updated

gazzdingo:
### Features and Changes
updates for saferollout only go though its own controller and removed out the saferollouts from the features controller

### DIFF
--- a/packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
+++ b/packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
@@ -7,6 +7,10 @@ import { getIntegrationFromDatasourceId } from "back-end/src/services/datasource
 import { SafeRolloutResultsQueryRunner } from "back-end/src/queryRunners/SafeRolloutResultsQueryRunner";
 import { getFeature } from "back-end/src/models/FeatureModel";
 import { SNAPSHOT_TIMEOUT } from "back-end/src/controllers/experiments";
+import {
+  SafeRolloutInterface,
+  validateCreateSafeRolloutFields,
+} from "back-end/src/validators/safe-rollout";
 
 // region GET /safe-rollout/:id/snapshot
 /**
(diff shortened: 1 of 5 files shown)


packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
```diff
@@ -170,3 +174,45 @@ export async function putSafeRolloutStatus(
...
+  if (safeRolloutFields) {
+    const validatedSafeRolloutFields = await validateCreateSafeRolloutFields(
+      safeRolloutFields,
+      context
+    );
+
+    await context.models.safeRollout.update(safeRollout, {
+      ...validatedSafeRolloutFields,
+    });
+  }
```
gazzdingo, line 212: not sure if we want an error here?

packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -577,6 +589,9 @@ export default function RuleModal({
         // eslint-disable-next-line
         delete (values as any).value;
       } else if (values.type === "safe-rollout") {
+        // delete for now
```
ahdriel, line 592: suggestion: Can we expand on the `why` or for how long the `for now` is? Maybe `delete until we can [blank]` or something?

packages/back-end/types/feature-rule.d.ts
```diff
@@ -10,6 +10,5 @@ export type PostFeatureRuleBody = {
 export type PutFeatureRuleBody = {
   rule: Partial<FeatureRule>;
   environment: string;
```
ahdriel, line 12: I think this means we also need to update the openapi.yaml files right? To add the new route for updating safeRollout and remove `safeRolloutFields` from the update.

ahdriel (approved):
I think this approach is okay for now but I am not the biggest fan of having one API to create a safe-rollout rule but having a different one (or 2 in this case) if I want to update it.

packages/back-end/types/feature-rule.d.ts
gazzdingo: is that part of our the openapi.yaml?

packages/front-end/components/Features/RuleModal/index.tsx
gazzdingo: left over i can remove that now

packages/back-end/src/routers/safe-rollout/safe-rollout.router.ts
```diff
@@ -12,7 +15,13 @@ const snapshotParams = z.object({ id: z.string() }).strict();
 const statusBody = z
   .object({ status: z.enum(safeRolloutStatusArray) })
   .strict();
-
+const safeRolloutBody = z
+  .object({
+    safeRolloutFields: createSafeRolloutValidator,
```
ahdriel, line 20: Are we sure we want the `create` validator here? Or is there a chance that we will receive just a single field to be updated for example?

PR #4017, opened 2025-04-30
Hide metric groups help text in Edit Metrics form

jdorn:
### Features and Changes

The Metric Selector is already cluttered, so adding more help text and another CTA was too much.  Removing for now until we get a chance to re-design this component.

Before:

image

After:

image

### DIFF
--- a/packages/front-end/components/Experiment/MetricsSelector.tsx
+++ b/packages/front-end/components/Experiment/MetricsSelector.tsx
@@ -213,11 +213,15 @@ const MetricsSelector: FC<{
     }
   });
 
-  const showMetricGroupHelper =
+  let showMetricGroupHelper =
     hasCommercialFeature("metric-groups") &&
     selected.length >= 2 &&
     !metricListContainsGroup &&
     datasource;
+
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -291,7 +295,7 @@ const MetricsSelector: FC<{
       helpText={
         <>
           {helpText}
-          {showMetricGroupHelper ? (
+          {showMetricGroupHelper && datasource ? (
```
ahdriel, line 298: question (nit): Why are we checking `datasource` explicitly here in addition of the check that happens in `showMetricGroupHelper`?

ahdriel (approved):
Code LGTM

packages/front-end/components/Experiment/MetricsSelector.tsx
jdorn: Typescript was complaining.  Before, it was smart enough to know that if `showMetricGroupHelper` was true, then datasource was definitely defined.  But my change confused Typescript.

PR #4025, opened 2025-05-02
Gm/exec dashboard

Auz:
Screenshot 2025-05-01 at 8 37 35 PM


packages/back-end/src/models/ExecReportModel.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  globallyUniqueIds: false,
+});
+
+export class ExecReportModel extends BaseClass {
+  protected canCreate(): boolean {
+    return this.context.permissions.canManageExecReports();
+  }
+  protected canRead(doc: ExecReportInterface): boolean {
+    return this.context.hasPermission("readData", doc.projects || "");
```
jdorn: Use `canReadMultiProjectResource` instead.  Has a little more logic to properly handle all the edge cases.

packages/back-end/src/routers/exec-report/exec-report.validators.ts
```diff
@@ -0,0 +1,28 @@
...
+  endDate: z.string(),
+});
+
+export const execReportValidator = z
+  .object({
+    id: z.string(),
+    organization: z.string(),
+    name: z.string().min(1),
+    description: z.string(),
+    dateRange: z.union([z.string(), DateRangeSchema]),
```
jdorn: A union between string and object is always a little weird.  Not a huge deal, but might be better to make it always an object with a type discriminator.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -620,6 +620,13 @@ export class Permissions {
     );
   };
 
+  public canManageExecReports = (): boolean => {
+    return this.checkProjectFilterPermission(
+      { projects: [] },
```
jdorn: Exec reports have `projects`, so this should not be treated like a global permission here.  Have this function accept the report object and use the projects in the permissions check.

packages/front-end/package.json
```diff
@@ -31,6 +31,7 @@
     "@visx/axis": "^2.1.0",
     "@visx/curve": "^2.1.0",
     "@visx/event": "^2.6.0",
+    "@visx/gradient": "^3.12.0",
```
jdorn, line 34: This is a different major version than all our other visx libraries. Don't know if it will cause an issue or not.  We should probably update them all to 3.x at some point.

packages/front-end/components/Tooltip/Tooltip.tsx
```diff
@@ -44,20 +45,28 @@ const Tooltip: FC<Props> = ({
...
   const [open, setOpen] = useState(state ?? false);
+  const [fadeIn, setFadeIn] = useState(false);
   const [alreadyHovered, setAlreadyHovered] = useState(false);
+  // we have to split the open state from the fade in state as if we set it all at once,
+  // the fade won't work as the opacity is not technically changing. By delaying slightly
+  // when we set the opacity to 1 (`tooltip-visible`) it lets the CSS transition work.
+  const openTimeout = useRef<NodeJS.Timeout | null>(null);
+  const fadeInTimeout = useRef<NodeJS.Timeout | null>(null);
+  const fadeOutTimeout = useRef<NodeJS.Timeout | null>(null);
   const closeTimeout = useRef<NodeJS.Timeout | null>(null);
```
jdorn: Having 4 different timers makes this very fragile and prone to race conditions.  I'm having trouble reasoning through all the user interactions and how they interact with each other and the timers.

packages/front-end/components/Experiment/TabbedPage/ExperimentStatusIndicator.tsx
```diff
@@ -24,6 +24,19 @@ export default function ExperimentStatusIndicator({
   );
 }
 
+export function ExperimentDot({ color }: { color: string }) {
```
jdorn: `color` here should not be typed as string since it only works with very specific values.  Instead, use the type from StatusIndicatorData.

```ts
color: StatusIndicatorData["color"]
```

packages/front-end/components/ExecReports/ExecReport.tsx
```diff
@@ -0,0 +1,452 @@
...
+  const projectsOptions = useProjectOptions(
+    () => {
+      return true;
+    },
+    selectedProjects || [],
+    [...projects, ...disallowedProjects]
+  );
+
+  const { experiments: allExperiments, error, loading } = useExperiments(
+    "",
```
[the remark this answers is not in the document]
Auz: I do it lower down in the filterResults. I did adjust it to apply the current project, though, by default if not otherwise specified.

jdorn (approved):
I didn't test it myself, but the changes are fairly low risk and shouldn't be a big deal if they break.

PR #4034, opened 2025-05-06
Update workflow notifications

ahdriel:
### Features and Changes

We have validated it works, so we updated the format to include the URL and removed the test branch.

### DIFF
--- a/.github/workflows/ci-nochanges.yml
+++ b/.github/workflows/ci-nochanges.yml
@@ -1,4 +1,4 @@
-name: CI No Changes
+name: CI
 on:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-nochanges.yml
```diff
@@ -1,7 +1,7 @@
-name: CI
+name: CI No Changes
```
jdorn, line 1: I don't think we can change this. We need the name to match exactly. "CI" is a required status check, so we need a job with that name to always pass, whether or not there are meaningful changes.

PR #4039, opened 2025-05-07
Added search filters for metrics and features

Auz:
Added dropdown filters to allow users to easily set syntactic search parameters easily (and of course, search more easily)

Screenshot 2025-05-07 at 11 46 19 AM


packages/front-end/components/Search/SearchFilters.tsx
```diff
@@ -0,0 +1,349 @@
...
+  items,
+  open,
+  setOpen,
+  updateQuery,
+  syntaxFilters,
+  operator = "",
+  heading,
+}) => {
+  const [filterSearch, setFilterSearch] = useState<string>("");
+  const showSearchFilter = useMemo(() => items.length > 10, [items]);
```
jdorn, line 104: Radix dropdowns have built-in type to search.  It only does a prefix search, so not quite as flexible as yours, but I think the UX is much better with the Radix one - the keyboard navigation feels much more natural.  Would simplify the code a lot here too.

packages/front-end/components/Search/SearchFilters.tsx
```diff
@@ -0,0 +1,349 @@
...
+            onClick={(e) => e.stopPropagation()}
+            onKeyDown={(e) => {
+              if (e.key !== "ArrowUp" && e.key !== "ArrowDown") {
+                e.stopPropagation();
+              }
+            }}
+          />
+        )}
+      </Box>
+      <Box overflowY="auto">
```
jdorn, line 157: I would add a maxHeight of 300px here

packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -271,32 +256,21 @@ const MetricsList = (): React.ReactElement => {
...
-      if (!showArchived) {
-        items = items.filter((m) => {
-          return !m.archived || recentlyArchived.has(m.id);
-        });
-      }
-      items = filterByTags(items, tagsFilter.tags);
-      return items;
-    },
-    [showArchived, recentlyArchived, tagsFilter.tags]
-  );
```
jdorn, line 286: There's a bug with the archived filter in the new version.  Once you show archived items, there's no way to unshow them.  Removing the filter still keeps them in the list.

Auz: Fixed

jdorn (approved):

PR #4041, opened 2025-05-08
[WIP] Experiment reporting dashboard

Kevin-Chant:


packages/back-end/src/models/BaseModel.ts
```diff
@@ -311,6 +311,16 @@ export abstract class BaseModel<
   ): Promise<z.infer<T>> {
     return this._updateOne(existing, updates, { writeOptions });
   }
+  public dangerousUpdateBypassPermission(
```
Kevin-Chant, line 320: This seemed like a utility that would be generally useful for `BaseModel`, but I wasn't 100% sure on the right approach. Open to suggestions and feedback here

packages/front-end/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper.tsx
```diff
@@ -39,7 +40,6 @@ export default function ExperimentMetricTimeSeriesGraphWrapper({
   pValueAdjustmentEnabled: boolean;
   firstDateToRender: Date;
 }) {
-  const { phase } = useSnapshot();
```
Kevin-Chant, line 63: Had to switch phase to prop-drilling so dashboards can re-use this visualization without providing a snapshot context

packages/front-end/components/Experiment/UsersTable.tsx
```diff
@@ -91,39 +93,42 @@ const UsersTable: FC<{
...
+          {(results || []).map((r, i) => {
+            if (!includeVariation(r, dimensionValuesFilter)) return null;
+            return (
+              <tr key={i}>
+                <td className="border-right">{r.name || <em>unknown</em>}</td>
+                {variations.map((v, i) => (
+                  <td key={i} className="pl-4">
+                    {numberFormatter.format(r.variations[i]?.users || 0)}
+                  </td>
+                ))}
```
Kevin-Chant, line 105: This diff is only indentation. The new `if () return null` clause is the only actual change

packages/back-end/src/enterprise/models/DashboardInstanceModel.ts
```diff
@@ -0,0 +1,229 @@
...
+    return this._createOne({
+      experimentId,
+      isDefault: true,
+      isDeleted: false,
+      userId: "",
+      editLevel: "organization",
+      enableAutoUpdates: true,
+      title: "Default Dashboard",
+      blocks,
+    });
```
Kevin-Chant: The default dashboard is currently editable (block changes) by all org members with edit perms on that experiment, but only manageable by those with `superDeleteReport`

packages/back-end/src/enterprise/models/DashboardTemplateModel.ts
Kevin-Chant, line 1: Mostly boilerplate as we aren't actually using this model yet (the default dashboard isn't persisted via a template). Should be safe to start with this and expand when we land the template builder

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -0,0 +1,271 @@
...
+    ...new Set(
+      dashboard.blocks
+        .filter((block) => block.type === "sql-explorer" && block.savedQueryId)
+        .map((block: SqlExplorerBlockInterface) => block.savedQueryId!)
+    ),
+  ]);
+
+  for (const savedQuery of savedQueries) {
+    // TODO: is this safe to run in the background or should this be awaited?
+    executeAndSaveQuery(context, savedQuery, datasource);
```
Kevin-Chant: Would like input here

packages/front-end/pages/api/init.ts
```diff
@@ -57,6 +58,7 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
     USE_FILE_PROXY: USING_FILE_PROXY,
     SUPERADMIN_DEFAULT_ROLE,
     NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY,
+    EXPERIMENT_REFRESH_FREQUENCY,
```
ahdriel, line 61: question: Right now this is specific to the dashboard refresh rate, should we qualify it with that? Otherwise it can be confused with the default experiment results refresh.

packages/front-end/pages/api/init.ts
```diff
@@ -30,6 +30,7 @@ export interface EnvironmentInitValue {
   superadminDefaultRole: string;
   ingestorOverride: string;
   stripePublishableKey: string;
+  experimentRefreshFrequency: number;
```
ahdriel, line 33: suggestion: Should we add the unit as well? `_REFRESH_FREQUENCY_HOURS`?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
```diff
@@ -0,0 +1,661 @@
...
+        requireBaselineVariation && value.length > 0
+          ? [...value, baselineVariation.id]
+          : value,
+    });
+  };
+
+  return (
+    <div
+      id="edit-drawer"
+      className={clsx("sidebarLeft", {
```
ahdriel, line 255: issue: This works for the sidebar but I am unable to scroll to the end of the page when the Edit Drawer is open. As in the drawer is on top of the content.

packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -118,14 +132,26 @@ export default function TabbedPage({
...
+      ];
+      if (experimentTabs.includes(tabName)) {
+        if (tabName === "dashboards" && !dashboardsEnabled) {
+          tabName = "overview";
+          tabPathSegments = [];
+        }
+        const tabPath = tabPathSegments.join("/");
+        setTab(tabName);
+        setTabPath(tabPath);
+        // Drop the tab path from the URL after reading it into state
```
ahdriel, line 147: question: Why do we need to do this? It makes harder to share the URL (i.e. I would not be able to copy the URL after loading it from a message)

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -0,0 +1,567 @@
...
+                    <Flex gap="4" align="center">
+                      <Select
+                        style={{
+                          marginLeft: "var(--space-3)",
+                          minWidth: "200px",
+                          borderBottom: "1px solid var(--slate-a5)",
+                        }}
+                        variant="ghost"
+                        value={dashboardId}
+                        setValue={setDashboardId}
```
ahdriel, line 320: issue: When changing this selector the URL doesn't update so I am unable to share a direct link with others.

I think this is a capability we want based on the other code we have around handling `/dashboard/{dashboardName}` paths.

packages/shared/src/util/types.ts
```diff
@@ -63,3 +63,24 @@ export function ensureAndReturn<T>(x: T): NonNullable<T> {
...
+
+export function isString(data: unknown): data is string {
+  return typeof data === "string";
+}
+
+export function isNumber(data: unknown): data is number {
+  return typeof data === "number";
+}
+
+export function partialToFull<T>(obj: Partial<T>): T {
```
ahdriel: question: Can we use `Required` instead which is part of the standard library?

We are not really checking if the object has all the defined properties, so I think a type-only solution should be enough.

packages/shared/src/enterprise/dashboards/utils.ts
```diff
@@ -0,0 +1,217 @@
...
+  data: DashboardBlockInterfaceOrData<T>
+): data is T {
+  const block = data as T;
+  return !!(block.id && block.uid && block.organization);
+}
+
+export function isDifferenceType(
+  value: string
+): value is DimensionBlockInterface["differenceType"] {
+  return ["absolute", "relative", "scaled"].includes(value);
```
ahdriel: suggestion: Ooh is there a way to link both values? Because I think TS will not complain if we update either but then the logic will be wrong.

packages/back-end/src/validators/saved-queries.ts
```diff
@@ -79,6 +79,7 @@ export const savedQueryValidator = z
     sql: z.string(),
     dataVizConfig: z.array(dataVizConfigValidator).optional(),
     results: queryExecutionResultValidator,
+    linkedDashboards: z.array(z.string()).optional(),
```
ahdriel: suggestion: Should we suffix it with `ids`?

packages/back-end/src/enterprise/models/DashboardInstanceModel.ts
```diff
@@ -0,0 +1,231 @@
...
+  ],
+});
+
+export const toInterface: ToInterface<DashboardInstanceInterface> = (doc) => {
+  const dashboard = removeMongooseFields(doc);
+  dashboard.blocks = dashboard.blocks.map(blockToInterface);
+  return dashboard;
+};
+
+export class DashboardInstanceModel extends BaseClass {
```
ahdriel: question: Should we name this just `DashboardModel`?

It would follow the pattern we have with `Experiments` and `ExperimentTemplate` already.
I guess I am just curious on why we decided that `Instance` is an important value add here.

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
```diff
@@ -0,0 +1,136 @@
...
+import {
+  CreateDashboardBlockInterface,
+  DashboardBlockInterface,
+} from "back-end/src/enterprise/validators/dashboard-block";
+import {
+  removeMongooseFields,
+  ToInterface,
+} from "back-end/src/util/mongo.util";
+
+export const dashboardBlockSchema = new mongoose.Schema(
```
ahdriel, line 13: question: Why this one does not extends from `BaseClass`?

I am curious if we could use only zod validators for the models instead of both mongoose and zod

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
```diff
@@ -0,0 +1,136 @@
...
+  {
+    organization: String,
+    id: String,
+    uid: String,
```
ahdriel, line 17: question: Why do we need both `id` and `uid`?

packages/back-end/src/enterprise/models/DashboardInstanceModel.ts
```diff
@@ -0,0 +1,231 @@
...
+    const isOwner = this.context.userId === existing.userId;
+    const isAdmin = this.context.permissions.canSuperDeleteReport();
+
+    const canManage = isOwner || isAdmin;
+    if (canManage) return true;
+    if (
+      "title" in updates ||
+      "editLevel" in updates ||
+      "enableAutoUpdates" in updates
+    ) {
```
ahdriel: question: Why is this needed? Curious if we should clarify a bit more

packages/back-end/src/enterprise/validators/dashboard-block.ts
```diff
@@ -0,0 +1,226 @@
...
+  trafficTableBlockInterface,
+  trafficGraphBlockInterface,
+  sqlExplorerBlockInterface,
+]);
+
+export type DashboardBlockInterface = z.infer<typeof dashboardBlockInterface>;
+export type DashboardBlockType = DashboardBlockInterface["type"];
+
+// Utility type for the discriminated union without the backend-generated fields
+const createOmits = {
```
ahdriel, line 179: We do have `CreateProps` but I am unsure if we can use this here.

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -290,6 +294,20 @@ export async function updateSnapshot({
         );
       }
     }
+
+    const dashboards = await context.models.dashboards.findByExperiment(
```
ahdriel, line 298: question: Do we need to do checks similar to `shouldUpdateExperimentAnalysisSummary`? Or do we want to have failed snapshots and reports/bandits to also show up in the dashboards?

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -141,38 +144,15 @@ export async function refreshSavedQuery(
...
-      error,
-      duration,
-      sql,
-    },
-    dateLastRan: new Date(),
-  });
-
   res.status(200).json({
     status: 200,
+    debugResults,
```
ahdriel, line 155: question: Do we want to keep this? Or maybe update the name of the property?

packages/back-end/src/routers/dashboards/dashboards.router.ts
```diff
@@ -0,0 +1,83 @@
...
+    enableAutoUpdates: z.boolean().optional(),
+    blocks: z
+      .array(z.union([createDashboardBlockInterface, dashboardBlockInterface]))
+      .optional(),
+  })
+  .strict();
+
+router.get(
+  "/",
+  validateRequestMiddleware({}),
```
ahdriel: nit: Do we need this or can we skip it if we don't need to validate request params?

packages/back-end/src/routers/dashboards/dashboards.router.ts
```diff
@@ -0,0 +1,83 @@
...
+  dashboardsController.getAllDashboards
+);
+
+router.get(
+  "/by-experiment/:experimentId",
```
ahdriel, line 40: Ooh this makes sense but i also want to put it under `/experiments/:experimentId/dashboards` which also makes sense.

Thoughts? This is the first `/by-{something}` route we have but it does not mean we should not adopt this pattern.

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -0,0 +1,304 @@
...
+}
+
+export async function createDashboard(
+  req: AuthRequest<z.infer<typeof createDashboardBody>, never, never>,
+  res: ResponseWithStatusAndError<SingleDashboardResponse>
+) {
+  const context = getContextFromReq(req);
+  if (!context.hasPremiumFeature("dashboards")) {
+    throw new Error("Must have a commercial License Key to create Dashboards");
+  }
```
ahdriel, line 68: question: Should this check be on the Model layer?

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -0,0 +1,304 @@
...
+  if (!dashboard) throw new Error("Cannot find dashboard");
+  const experiment = await getExperimentById(context, dashboard.experimentId);
+  if (!experiment)
+    throw new Error("Cannot update dashboard without an attached experiment");
+  const datasource = await getDataSourceById(context, experiment.datasource);
+  if (!datasource) throw new Error("Failed to find connected datasource");
+
+  // This is doing an expensive analytics SQL query, so may take a long time
+  // Set timeout to 30 minutes
+  req.setTimeout(SNAPSHOT_TIMEOUT);
```
ahdriel: I think if we are adopting the pattern of `create a query, return the id, provide and endpoint to get the latest status` then we can remove this increased timeout.

Otherwise we do need to wait on line `executeAndSaveQuery` otherwise this is also not needed and we should remove it.

packages/front-end/hooks/useDashboards.ts
```diff
@@ -0,0 +1,45 @@
+import { useMemo } from "react";
+import { DashboardInstanceInterface } from "back-end/src/enterprise/validators/dashboard-instance";
+import useApi from "./useApi";
+
+export function useAllDashboards() {
+  const { data, error, mutate } = useApi<{
+    dashboards: DashboardInstanceInterface[];
+  }>("/dashboards/");
```
ahdriel: nit: Should we remove the trailing slash?

packages/front-end/pages/api/init.ts
Kevin-Chant: No, this is making it so the frontend (currently only used for dashboards) can read the experiment refresh rate. Dashboard refreshes are handled by the existing experiment results refresh job which is why I needed to surface that env var

packages/front-end/components/Experiment/TabbedPage/index.tsx
Kevin-Chant: Two reasons
1. I was running into a bunch of issues when I initially tried connecting dashboard changes to updating the url but then that triggered other useEffects and had a ton of flicker issues
2. It's a lot cleaner of a URL in general and I wasn't sure I wanted to change the visible format of the urls for our experiment pages

So for simplicity I just went with this for now but I could be convinced we should change this

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
Kevin-Chant: Fair, but the content that it's on top of is only the comments section right? I don't think I want to artificially lengthen the main content area with blank space when the drawer is open since we're trying to direct attention to the block they're editing

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
Kevin-Chant: See comment above for the issues I had with that approach, but given the Share button I'm thinking it's okay for now. Would be happy to revisit this as we iterate though

packages/shared/src/util/types.ts
Kevin-Chant: So the usage is currently
```
{...partialToFull(objectProps)}
```
do you think a raw cast like this would be preferable? I leaned away from it because of the `as unknown` which always worries me, but like you said this helper isn't actually giving much type safety either
```
{...(objectProps as unknown as ObjectProps<T>)}
```
I can't find a way to make Required work in this case because `objectProps` is initialized as `Partial<ObjectProps<T>>`

packages/back-end/src/enterprise/models/DashboardInstanceModel.ts
Kevin-Chant: I think that's just a relic from some of the conversations we had where we were adding extra differentiation between a single dashboard _instance_ and a template. Good call :+1:

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
Kevin-Chant: Our baseclass doesn't support schemas with discriminated unions. This is something I tried really hard to make work but gave up and focused on getting the mvp out. I do want to update baseclass and transition the block model to use it as well in the future; that would make things a lot more streamlined

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
Kevin-Chant: Good question, I was just taking the schema from ReportModel since that was the base I was working off :sweat_smile:

packages/back-end/src/enterprise/models/DashboardInstanceModel.ts
Kevin-Chant: I'll add a comment, but some edits are really a "manage" action while others are a more standard "edit", so some require elevated permissions (`canManage` vs `canEdit`)

packages/back-end/src/enterprise/validators/dashboard-block.ts
Kevin-Chant: Yeah unfortunately our type helpers like CreateProps don't support discriminated unions. This is also something I want to revisit if I can get BaseClass working for the block model

packages/back-end/src/models/ExperimentSnapshotModel.ts
Kevin-Chant: Indentation is a bit hard to follow here but this block is actually still under the ` if (shouldUpdateExperimentAnalysisSummary) {` check

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
Kevin-Chant: This is for the sql explorer which does actually use this property name. Diff is a bit misleading because I'm not actually adding a new property here, just merging two possible return statements

packages/back-end/src/routers/dashboards/dashboards.router.ts
Kevin-Chant: I had it under `/experiments/:eId/dashboards` initially because dashboards were actually stored on the experiment model for a bit. I moved it to this when it got its own router & controller, but I could be persuaded to change it to match
```
app.get("/experiment/:id/reports", reportsController.getReportsOnExperiment);
```
My only complaint is that the raw routes we have in `app.ts` don't match up nicely with the newer router/controller setup & don't use `validateRequestMiddleware` so I'd prefer to keep it grouped with the other dashboard code

packages/back-end/src/routers/dashboards/dashboards.controller.ts
Kevin-Chant: It is there too, but again due to our assumptions in BaseClass I can't delegate everything to the model's helpers. In this case it's because I need to run a db create for the blocks before storing them in a dashboard, and `CreateProps` doesn't also omit `id` etc from child objects. See the comment a few lines below
```
  // Duplicate permissions checks to prevent persisting the child blocks if the user doesn't have permission
```

packages/front-end/pages/api/init.ts
ahdriel: Ohh I see -- thanks for clarifying!

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
ahdriel: That makes sense! Regarding calling attention to the drawer:
- Should we change the background color to be lighter in dark mode?
- And should we increase the max-height if we have the vertical space? It feels a bit constrained for no reason on my monitor at the moment

Screenshot 2025-07-30 at 10 32 15 AM

_disregard the error notice at the bottom left_

Not blockers though -- but would make a better UX.

packages/front-end/components/Experiment/TabbedPage/index.tsx
ahdriel: While the Share button is a good add, I do believe our current experience is less than ideal because of a technical limitation on our end.

I don't know how hard it would be to fix it, so it might not be worth, but I (personally) strongly believe that having the URL sync with the dashboard would be a way better (and more aligned with what users would expect) experience.

packages/shared/src/util/types.ts
ahdriel: Yeah, I think the `partialToFull` function is masking the casting we are doing and `as unknown` makes it more explicit that it is somewhat unsafe and to be careful around it.

As an alternative we could change `partialToFull` to use a Zod validator and then we are really ensuring that the object matches the type instead of just casting it.

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
ahdriel: It looks like that Reports use `uid` only for the public URL sharing (for logged out users). I would check with @bryce-fitzsimons why we decided to have it as a separate property, but would vote to remove it for now if we are not using it for Dashboards.

packages/back-end/src/enterprise/models/DashboardInstanceModel.ts
ahdriel: Ohh I see -- that is interesting. Yeah I think our current permissioning model does not have a very good pattern to handle `is this user allowed to change these properties for this model` while we only have broader can create/edit whole models.

packages/back-end/src/models/ExperimentSnapshotModel.ts
ahdriel: Ohh I see! Thanks for checking

packages/back-end/src/routers/dashboards/dashboards.router.ts
ahdriel: Yeah the code colocation makes a lot of sense.

I think it is fine to keep as-is given that while we do require a dashboard to be tied to a specific experiment I also see a path where we might remove that (generic dashboard showing different experiment results, maybe the blocks would be tied to a experiment in this case).

packages/front-end/components/Experiment/TabbedPage/index.tsx
bryce-fitzsimons: also agree, but I think this can be a fast follow.

We might want to think about what sharing looks like ideally. Is there a dedicated shareable link within the experiment page? Is there a stand-alone url for the dashboard by itself (similar to a report)? Will there be a notion of public/org-only shares (and SSR metadata)? I think doing it properly will take a bit more effort...

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
bryce-fitzsimons: I'd keep it. We're definitely going to fast with shareable dashboards in the future, and it would be better to use the public uid when doing so (for both security and url cleanliness reasons)

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
```diff
@@ -0,0 +1,136 @@
...
+  experimentId: String,
+});
+
+const sqlExplorerBlockSchema = new mongoose.Schema({
+  savedQueryId: String,
+  dataVizConfigIndex: Number,
+});
+
+dashboardBlockSchema.index({
+  uid: 1,
```
bryce-fitzsimons, line 98: why do blocks have a uid? isn't the id enough?

packages/back-end/src/enterprise/validators/dashboard-block.ts
```diff
@@ -0,0 +1,228 @@
+import { z } from "zod";
+import { DistributiveOmit } from "shared/util";
+
+export const differenceTypes = ["absolute", "relative", "scaled"] as const;
+
+const baseBlockInterface = z
+  .object({
+    organization: z.string(),
+    id: z.string(),
+    uid: z.string(),
```
bryce-fitzsimons: why not just id?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/TimeSeriesBlock.tsx
```diff
@@ -0,0 +1,70 @@
+import { TimeSeriesBlockInterface } from "back-end/src/enterprise/validators/dashboard-block";
+import { expandMetricGroups } from "shared/experiments";
+import ExperimentMetricTimeSeriesGraphWrapper from "@/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper";
+import useOrgSettings from "@/hooks/useOrgSettings";
+import { useDefinitions } from "@/services/DefinitionsContext";
+import { getMetricResultGroup } from "@/components/Experiment/BreakDownResults";
+import { BlockProps } from ".";
+
+export default function TimeSeriesBlock({
```
bryce-fitzsimons: might be good to rename this to ExperimentTimeSeriesBlock (and also the underlying model's "time-series" type). There are many types of time series displays that could be included in dashboards, most of which are not tied to experiments (safe rollouts, plain metric analysis, plus new stuff...)

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/DescriptionBlock.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { DescriptionBlockInterface } from "back-end/src/enterprise/validators/dashboard-block";
+import { ScrollArea } from "@radix-ui/themes";
+import Markdown from "react-markdown";
+import { Box } from "spectacle";
+import { BlockProps } from ".";
+
+export default function DescriptionBlock({
```
bryce-fitzsimons: Rename to ExperimentDescription (and underlying type field)? Other things that will soon appear on dashboards will also have descriptions

bryce-fitzsimons: Or abstract the block to take a `description?: string` so that it is reusable

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/MetricBlock.tsx
```diff
@@ -0,0 +1,150 @@
...
+  expandMetricGroups,
+  ExperimentMetricInterface,
+} from "shared/experiments";
+import useOrgSettings from "@/hooks/useOrgSettings";
+import ResultsTable from "@/components/Experiment/ResultsTable";
+import { useDefinitions } from "@/services/DefinitionsContext";
+import { getMetricResultGroup } from "@/components/Experiment/BreakDownResults";
+import { BlockProps } from ".";
+
+export default function MetricBlock({
```
bryce-fitzsimons: rename to ExperimentMetricBlock (and change type field). Standalone metrics are one of the first expansions of dashboards I'd imagine we'll build

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/TrafficGraphBlock.tsx
```diff
@@ -0,0 +1,48 @@
+import { TrafficGraphBlockInterface } from "back-end/src/enterprise/validators/dashboard-block";
+import TrafficCard from "@/components/HealthTab/TrafficCard";
+import useOrgSettings from "@/hooks/useOrgSettings";
+import Callout from "@/components/Radix/Callout";
+import { BlockProps } from ".";
+
+export default function TrafficGraphBlock({
```
bryce-fitzsimons: ExperimentTrafficGraph perhaps? I think we'll be looking at other types of traffic soon outside of experiments (feature, "pageload" metric, etc)

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardUpdateDisplay.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const textColor = refreshError || numFailed > 0 ? "red" : undefined;
+  const content = refreshError
+    ? "Update Failed"
+    : numFailed > 0
+    ? "One or more queries failed"
+    : snapshot.runStarted
+    ? `Updated ${ago(snapshot.runStarted).replace("about ", "")}`
+    : "Not started yet";
+  const tooltipBody = refreshError ? refreshError : undefined;
```
bryce-fitzsimons, line 57: Is this true for stopped experiments (do the dashboards keep auto-updating)? If no, then we should probably change/remove the lightning bolt and tooltip.

image

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
bryce-fitzsimons: also agree, even in light mode the "faint gray on faint gray-purple" is an odd look. But not blocking at all.

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
```diff
@@ -141,6 +153,112 @@ export default function SavedQueriesList({ savedQueries, mutate }: Props) {
...
                     </td>
                     <td>{query.results?.results?.length || 0}</td>
+                    <td>
+                      <div
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          e.preventDefault();
+                        }}
+                      >
+                        <Tooltip
```
ahdriel, line 163: issue(non-blocker): This Tooltip looks broken on Dark Mode

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
```diff
@@ -176,6 +294,46 @@ export default function SavedQueriesList({ savedQueries, mutate }: Props) {
...
+                                return (
+                                  <div>
+                                    <Callout
+                                      status="warning"
+                                      mb="2"
+                                    >{`This saved query is in use by ${
+                                      dashboardIds.length
+                                    } dashboard${
+                                      dashboardIds.length === 1 ? "" : "s"
+                                    }. If deleted, SQL Explorer widgets will be cleared.`}</Callout>
```
ahdriel: nit: Should we say `deleted` or `removed` which is the language we use in other scenarios?

packages/front-end/components/Tooltip/Tooltip.tsx
```diff
@@ -27,6 +27,7 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
   shouldDisplay?: boolean;
   usePortal?: boolean;
   state?: boolean;
+  ignoreMouseEvents?: boolean; // Prevent the tooltip from reacting to mouseEnter and mouseExit events
```
ahdriel, line 30: suggestion (non-blocker): I think this component is suffering from being semi-controlled when it should be either controlled or uncontrolled.

We also have the new Radix Tooltip and Radix Popover, so maybe we don't need to fix this, but I do think this new prop is because of the above.

packages/front-end/components/HealthTab/TrafficCard.tsx
```diff
@@ -97,14 +110,16 @@ export default function TrafficCard({
   }, [trafficByDate, variations, cumulative]);
 
   const sortedDimensionSlices = useMemo(() => {
-    return traffic.dimension[selectedDimension]?.sort(compareDimsByTotalUsers);
+    return traffic.dimension?.[selectedDimension]?.sort(
```
ahdriel, line 115: suggestion: can we remove hte `?.` for dimension as it should be defined always?

packages/front-end/enterprise/components/Dashboards/DashboardSnapshotProvider.tsx
```diff
@@ -0,0 +1,325 @@
...
+    defaultSnapshot,
+    loading: snapshotsLoading,
+    error: snapshotsError,
+    mutateSnapshot: mutateDefault,
+    mutateSnapshotsMap: mutateSnapshotsMap,
+    refreshStatus,
+    snapshotsMap,
+  } = useContext(DashboardSnapshotContext);
+
+  const refreshing = refreshStatus !== "running";
```
ahdriel: question: Does this mean that a `refreshStatus` of `succeeded` means this is `refreshing`?

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -0,0 +1,567 @@
...
+  const { hasCommercialFeature } = useUser();
+
+  const hasDashboardFeature = hasCommercialFeature("dashboards");
+  const canCreate = permissionsUtil.canCreateReport(experiment);
+  const canUpdateDashboard = experiment
+    ? permissionsUtil.canViewReportModal(experiment.project)
+    : true;
+  const isOwner = userId === dashboard?.userId || !dashboard?.userId;
+  const isAdmin = permissionsUtil.canSuperDeleteReport();
+  const canEdit =
```
ahdriel, line 126: question: Should this belong in our permissionsUtil or a similar layer?

It is okay for now, but would be nice to centralize everything.

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -0,0 +1,567 @@
...
+      setDashboardId(defaultDashboard?.id ?? dashboards[0].id);
+    }
+  }, [dashboardId, dashboards, defaultDashboard]);
+
+  const {
+    userId,
+    settings: { updateSchedule },
+  } = useUser();
+  const [isEditing, setIsEditing] = useState(false);
+  const [showUpgradeModal, setShowUpgradeModal] = useState(false);
```
ahdriel, line 92: question: When we show the upgrade modal how can we ensure the user understands this is only in the Enterprise plan and not included in Pro?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
```diff
@@ -0,0 +1,486 @@
...
+  },
+  dimension: {
+    name: "Dimension Results",
+  },
+  "time-series": {
+    name: "Time Series",
+  },
+  "traffic-graph": {
+    name: "Traffic Time Series",
+    hideTitle: true,
```
ahdriel, line 55: ui issue: when `hideTitle` is true we are still keeping the space there but not showing either the default or the custom block title.

Screenshot 2025-07-30 at 4 27 14 PM

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
```diff
@@ -0,0 +1,486 @@
...
+                ...otherBlocks.slice(i + direction),
+              ]);
+            }
+          }}
+          mutate={mutate}
+        />
+        {isEditing && (
+          <Flex justify="center" mb="1em" position="relative">
+            {(hoverAddBlock === i || addBlockDropdown === i) && (
+              <div
```
ahdriel, line 318: nit ui: It is not centered / it gets rendered over the + button

Screenshot 2025-07-30 at 4 22 23 PM

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/index.tsx
```diff
@@ -0,0 +1,388 @@
...
+      {!definitionsReady ||
+      experimentsLoading ||
+      dashboardSnapshotLoading ||
+      dashboardContextLoading ||
+      (blockHasSavedQuery && savedQueryLoading) ? (
+        <BlockLoadingSnapshot />
+      ) : blockNeedsConfiguration ? (
+        <BlockNeedsConfiguration block={block} />
+      ) : !snapshot || !analysis || !analysis.results[0] ? (
+        <BlockMissingData />
```
ahdriel, line 357: I am seeing this even for a SQL Explorer block, which I don't think depends on the experiment data itself.
(it would be the last block here, which has a valid visualization in the sql explorer tab)

Screenshot 2025-07-30 at 4 34 19 PM

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
Kevin-Chant: Woah, why does your dark mode look so... _dark_? @bryce-fitzsimons is that the same experience you get? Mine feels a lot more reasonable to me

image

As for the drawer, it should have a handle at the top to expand/contract it in the next version so I'm inclined to leave it for now

packages/front-end/components/Experiment/TabbedPage/index.tsx
Kevin-Chant: ^ Yeah I'm inclined to wait and make sure we get it right on the next iteration. The stopgap seems good enough to unblock MVP but I agree that it deserves more careful thought

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
Kevin-Chant: Didn't you say above that you'd suggest keeping both `uid` and `id`? Or were you thinking about the Dashboard itself rather than blocks? And if so, then when sharing a Dashboard by uid, what would we want to do with the id/organization fields of the constituent blocks?

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
Kevin-Chant: This is actually an intentional choice for our Tooltip component. Personally I disagree with it but it's not a bug; the theme is flipped (I think to create more contrast so it's visible?)
```
<RadixTheme flip={true}>
```
image

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
Kevin-Chant: This is on the delete action for the query, so I went with "deleted" to match the title and CTA. Which other scenarios do you mean?

image

packages/front-end/components/Tooltip/Tooltip.tsx
Kevin-Chant: Yeah the semi-control isn't helping things but actually a lot of the bugginess is inherent to the Tooltip component because it was happening in a non-controlled context too :grimacing:

Wdym about the Radix tooltip/popovers? Are they WIP and coming soon?

packages/front-end/components/HealthTab/TrafficCard.tsx
Kevin-Chant: @bryce-fitzsimons wasn't this one of the crashes you ran into? Maybe it was because of your docker issues so it's an unrealistic error case, but I remember adding it because of an actual error on the page

packages/front-end/enterprise/components/Dashboards/DashboardSnapshotProvider.tsx
Kevin-Chant: Thanks, this got left behind when I changed the way refreshes are handled. The whole block is unused now, but previously the variable names were different

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
Kevin-Chant: Probably, but we can't check `userId === Object.userId` in our permissions layer currently. That feels like a good change to make so we can move other cases of owner checks there

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
Kevin-Chant: That's built into the upgrade modal itself because `dashboards` is in the `enterprise` plan in `license-consts`. 
```
  function shouldShowEnterpriseTreatment(): boolean {
    // if no commercialFeature is provided, determine what plan to show based on org's current plan
    if (!commercialFeature) {
      if (
        !effectiveAccountPlan ||
        ["oss", "starter"].includes(effectiveAccountPlan)
      ) {
        return false;
      }

      return ["pro", "pro_sso"].includes(effectiveAccountPlan);
    } else {
      return commercialFeatureLowestPlan?.[commercialFeature] === "enterprise";
    }
  }
```

packages/back-end/src/enterprise/models/DashboardBlockModel.ts
bryce-fitzsimons: I think the `uid` only makes sense for things that are public, like a shareable url. Not sure this translates to blocks. I could be missing something though? Are the blocks fetchable by public URL or similar?

Kevin-Chant: I didn't know whether they'd be public or not on their own, so I just tossed the `uid` on them as well in case we wanted to be able to share individual blocks

packages/front-end/components/HealthTab/TrafficCard.tsx
bryce-fitzsimons: indeed, there are some bad assumptions further up the chain about the availability of `traffic`. When the datasource is unreachable, traffic comes in as `null` (I think...). I think leave this for now...

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
bryce-fitzsimons: Some other time, we'll have to add an optional override for `flip` and manually fix some of these ugly tooltips...

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
Kevin-Chant: image

That's because the first line is for the title + edit dropdown, and the description gets its own full-width line. That said, it looks like the TrafficCard itself also has way too much y margin so I'm stripping that out

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
bryce-fitzsimons: And/or embrace something like HoverCard

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
Kevin-Chant: Yup! Height of 9px puts it too high and 10 too low. Somehow the icon is in between a pixel boundary which makes 0 sense to me :upside_down_face: 

I can at least put it behind the button though

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/index.tsx
Kevin-Chant: So that only happens when the entire experiment snapshot is missing (should be very rare). In most cases, sql explorer blocks will be fine rendering independently when other blocks are missing data.
I just tried making these sorts of non-snapshot-dependent blocks ignore this check, but it's actually kinda complicated so I'm planning to leave it as is for now

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
bryce-fitzsimons: Mine looks like this... they're all kind of equally dark, no?
image

Kevin-Chant:
@ahdriel when you pull today's changes you'll need to recreate existing dashboards because I changed the model in a non-backwards-compatible way. Just a heads up

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
ahdriel: Yeah they all feel kinda similar to me -- making the background of the drawer lighter would call more attention to it which would be better.

Or maybe we use a purple border or something? Should check with Kari.

Kevin-Chant: Huh. On my phone those 3 screenshots are the same luminosity but on my computer mine is like 3x brighter and easier to see the details. Either way I agree with needing some tuning of styles here but I'd like to put that in fast follows so we don't block on it

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
ahdriel: Oh I should've clarified that I meant the last part, `widgets will be cleared` -- should it also say `deleted` or `removed` there?

Kevin-Chant: Whoops completely missed it when skimming the message. Will swap to deleted

packages/front-end/components/Tooltip/Tooltip.tsx
ahdriel: We are already using `import { Tooltip } from "@radix-ui/themes";` but we don't have a `components/Radix/Tooltip` because there was no need for a wrapper yet.

The same for [Popover](https://www.radix-ui.com/themes/docs/components/popover), but we are using it only in a single place ([here](https://github.com/growthbook/growthbook/blob/kc/experiment-reports/packages/front-end/components/SafeRollout/Results/ResultsTable.tsx#L510)).

packages/front-end/components/HealthTab/TrafficCard.tsx
ahdriel: Got it -- yeah, would be nice to fix the types up in the chain so we can trust it more.

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
ahdriel: Nice! I just learned something new

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
ahdriel: Screenshot 2025-07-31 at 8 47 10 AM

Oh related to Traffic Time Series (not TrafficCard tho), we do have now a box inside a box which I don't think we need.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
ahdriel: Also the + is not centered in the box 👀 
Screenshot 2025-07-31 at 8 48 53 AM

ahdriel (approved):
Thanks for your patience & for addressing the comments 🙇

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentDimensionBlock.tsx
```diff
@@ -0,0 +1,134 @@
...
+  );
+  const secondaryMetrics = expSecondaryMetrics.filter((mId) =>
+    expandedMetricIds.includes(mId)
+  );
+  const guardrailMetrics = expGuardrailMetrics.filter((mId) =>
+    expandedMetricIds.includes(mId)
+  );
+
+  return (
+    <BreakDownResults
```
bryce-fitzsimons, line 104: Can we please make the metric names consistent with the non-dimension view? Ie, no hover state, not clickable, thinner font?
image

Kevin-Chant: I removed the tooltip, but the font styling is just part of our experiment results table styling so I'm leaving that as-is

PR #4042, opened 2025-05-08
Fix CSS injection vulnerability in organization appearance settings

jjoelk:
## Security Fix: CSS Injection Vulnerability in Organization Appearance Settings

This PR addresses a critical security vulnerability in the organization appearance settings where unsanitized user input from the `primaryColor` and `secondaryColor` fields could lead to CSS injection attacks.

### CVE Information
- **CVE ID Request**: 1864028 (pending assignment)
- **Vulnerability Type**: DOM-based Style Injection via React dangerouslySetInnerHTML

### Changes Made
- Added hex color validation function to ensure primaryColor and secondaryColor are valid hex codes
- Refactored backgroundShade function to be more robust and handle invalid input
- Implemented CSS variables for improved security
- Only inject custom styles if color values pass validation

This patch prevents potential CSS injection attacks through malicious color values and ensures the application only applies styles with validated inputs.

### Security Impact

The current implementation inserts user-controlled input directly into a CSS string without proper validation, which is then inserted into the DOM using `dangerouslySetInnerHTML`. This allows an attacker with organization admin privileges to break out of the CSS context and inject arbitrary styles.

This vulnerability could be exploited to:
- Create convincing phishing interfaces that appear legitimate
- Display fake security alerts or notices to users
- Manipulate the application UI in ways that could mislead users
- Enable social engineering attacks against GrowthBook users
- Create timing-based attacks that could potentially leak sensitive information like API keys

### Testing

The fix has been tested to ensure:
- Valid hex color codes (#RGB and #RRGGBB formats) continue to work normally
- Invalid color values are properly rejected
- Attempts to inject malicious CSS are blocked
- The visual appearance remains unchanged for legitimate color values
- Timing-based attacks are no longer possible

### Dependencies

None - this is a self-contained fix that requires no additional dependencies.

### Screenshots

The screenshot below demonstrates a serious security concern - a CSS injection attack that can be used to create a timing-based attack potentially exposing API keys and secrets. This shows why proper input validation and CSS context separation is critical.

read key

For transparency: I initially reported this vulnerability privately to security@growthbook.io per the security policy. The response I received indicated this was not considered a security issue. My follow-up messages providing additional evidence and explaining the security implications in more detail did not receive any reply.

Given the clear security implications demonstrated in the screenshots and the potential for user deception, CSS-based timing attacks, and UI manipulation that could lead to credential theft, I've created this PR with a complete fix.

The changes are minimal, non-disruptive to legitimate functionality, and follow security best practices for handling user input that affects styling. I've also submitted a CVE request (1864028) for this vulnerability.

### DIFF
--- a/packages/front-end/components/Layout/Layout.tsx
+++ b/packages/front-end/components/Layout/Layout.tsx
@@ -313,20 +313,30 @@ const otherPageTitles = [
   },
 ];
 
-const backgroundShade = (color: string) => {
+// Validates if a string is a valid hex color code (#RGB or #RRGGBB format)
+const isValidHexColor = (color?: string): boolean => {
+  if (!color) return false;
+  return /^#([0-9A-F]{3}){1,2}$/i.test(color);
+};
+
+const backgroundShade = (color: string): "light" | "dark" => {
+  if (!isValidHexColor(color)) {
+    return "light"; // Default to light if invalid color
+  }
+  
   // convert to RGB
-  // @ts-expect-error TS(2769) If you come across this, please fix it!: No overload matches this call.
-  const c = +("0x" + color.slice(1).replace(color.length < 5 && /./g, "$&$&"));
-  const r = c >> 16;
-  const g = (c >> 8) & 255;
-  const b = c & 255;
+  const normalizedColor = color.slice(1);
+  const expandedColor = normalizedColor.length === 3 
+    ? normalizedColor.split('').map(c => c + c).join('') 
+    : normalizedColor;
+  
+  const r = parseInt(expandedColor.slice(0, 2), 16);
+  const g = parseInt(expandedColor.slice(2, 4), 16);
+  const b = parseInt(expandedColor.slice(4, 6), 16);
+  
   // http://alienryderflex.com/hsp.html
   const hsp = Math.sqrt(0.299 * (r * r) + 0.587 * (g * g) + 0.114 * (b * b));
-  if (hsp > 127.5) {
-    return "light";
-  } else {
-    return "dark";
-  }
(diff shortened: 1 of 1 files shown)


jdorn:
Thanks for the PR!  To clarify why this was not considered a critical security vulnerability, it requires the attacker to already be an organization admin, which means they already have full access to the account and would only be able to attack themselves.

That being said, definitely still worth fixing this.  These settings are super old and undocumented, so I'd be in favor of just removing them completely instead of trying to sanitize the values.  Are you able to update your PR with that approach?

jjoelk:
Hey @jdorn,


Thanks for the feedback! You're absolutely right about my choice of words - calling it "critical" was probably too strong given the admin privilege requirement. My bad on that one.

The main security concern I was trying to highlight is around persistence. Once this payload is injected, it stays active even after the compromised admin changes their password, gets their access revoked, or leaves the organization entirely. So if an admin account gets compromised (which happens more often than we'd like), this creates a way for attackers to maintain a foothold that persists beyond the initial compromise.

Regarding removing the feature entirely - I totally get that these settings are old and undocumented, but I'm a bit concerned that some organizations might already be using the color customization for their branding. Removing it could break their existing UI customizations and potentially cause some frustration for teams that have already styled their GrowthBook instance to match their company colors.

The sanitization approach in the PR is pretty lightweight and maintains backward compatibility while closing the security gap. It only validates that the input is actually a valid hex color, which seems like what the feature was intended for anyway.

But honestly, you know the codebase and user base way better than I do, so I think it's better for you to decide what approach makes the most sense for GrowthBook.

Thanks again for taking the time to review this!

jdorn:

PR #4046, opened 2025-05-08
Update visual-editor.mdx

august-growthbook:
Add section about toggling on visual experiments at the SDK level.


royalfig (approved):
Looks good!

PR #4048, opened 2025-05-08
auto roll-back if Guard metrics and slow rampup

gazzdingo:
### Features and Changes
- there is a new setting in the safe rollout that will allow users to automatically rollback a safe rollout if any of the guardrail metrics are failing.
- it will also ramp up the safe rollout it will start at 10% and make its way upto 100% over 25% of the total time 
 


packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -36,7 +37,15 @@ const safeRolloutNotification = [
   "rollback",
 ] as const;
 export type SafeRolloutNotification = typeof safeRolloutNotification[number];
-
+const rampUpSchedule = z.object({
+  enabled: z.boolean(),
+  step: z.number(),
+  steps: z.array(z.number()), // percentage of traffic to release each step
```
gazzdingo, line 43: not sure if we want this to start with and corner our selfs into this approach since it is hard coded. also enabled but we will need someone to track which step they are on

gazzdingo: @ahdriel

packages/back-end/src/util/features.ts
```diff
@@ -572,7 +577,19 @@ export function getFeatureDefinition({
...
           } else {
-            rule.coverage = 1; // Always 100% right now
+            if (
+              safeRollout?.rampUpSchedule.rampUpCompleted ||
+              !safeRollout?.rampUpSchedule.enabled
+            ) {
+              rule.coverage = 1; // Always 100% right now
+            } else {
+              rule.coverage =
+                (1 *
```
gazzdingo, line 587: lol i dont need 1 * i can just set the coverage from the step

packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -15,7 +15,7 @@ export type SafeRolloutStatus = typeof safeRolloutStatusArray[number];
 
 export const MaxDuration = z.object({
   amount: z.number(),
-  unit: z.enum(["days", "hours", "minutes"]),
+  unit: z.enum(["days", "hours", "weeks", "minutes"]),
```
ahdriel, line 18: nit: Can we order it from bigger to smaller?

packages/back-end/src/validators/safe-rollout.ts
ahdriel: @gazzdingo Why do we have a different schedule for the analysis & the rampup?

Should we use the same & then we can always ensure we run the analysis right before changing the schedule in case the results are not the latest?

packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -36,7 +37,15 @@ const safeRolloutNotification = [
...
 ] as const;
 export type SafeRolloutNotification = typeof safeRolloutNotification[number];
-
+const rampUpSchedule = z.object({
+  enabled: z.boolean(),
+  step: z.number(),
+  steps: z.array(z.number()), // percentage of traffic to release each step
+  nextUpdate: z.date().optional(),
+  lastUpdate: z.date().optional(),
+  rampUpCompleted: z.boolean(),
```
ahdriel: Should we keep track of when the ramp up changes happened? For example, went to step 2 at date X and then step 3 at date Y.

packages/back-end/src/controllers/features.ts
```diff
@@ -1203,6 +1203,13 @@ export async function postFeatureRule(
       featureId: feature.id,
       status: rule.status,
       autoSnapshots: true,
+      rampUpSchedule: {
+        enabled: safeRolloutFields.rampUpSchedule.enabled, // this is used so that we can disable the ramp up schedule using feature Flag
+        step: 0,
+        steps: [0.1, 0.25, 0.5],
```
ahdriel, line 1209: question: Do we need to include `1` here? Or when would we go to a 100% coverage (50/50 split) for a Safe Rollout rule?

packages/back-end/test/api/features.test.ts
```diff
@@ -36,17 +36,22 @@ describe("features API", () => {
...
-    getFeature.mockReturnValue(undefined);
-    addTags.mockReturnValue(undefined);
-    createInterfaceEnvSettingsFromApiEnvSettings.mockReturnValue(
+    (createFeature as jest.Mock).mockImplementation((v) => v);
+    (getFeature as jest.Mock).mockReturnValue(undefined);
+    (addTags as jest.Mock).mockReturnValue(undefined);
+    (createInterfaceEnvSettingsFromApiEnvSettings as jest.Mock).mockReturnValue(
       "createInterfaceEnvSettingsFromApiEnvSettings"
     );
     getSavedGroupMap.mockReturnValue("savedGroupMap");
```
ahdriel, line 57: Should we cast these 2 as `jest.Mock` too? (`getSavedGroupMap` & `getApiFeatureObj`)

packages/back-end/test/features.test.ts
```diff
@@ -1322,6 +1355,7 @@ describe("SDK Payloads", () => {
         environment: "production",
         groupMap: groupMap,
         experimentMap: experimentMap,
+        safeRolloutMap: safeRolloutMap,
```
ahdriel, line 1358: question: Do we have tests that are validating the logic of the ramp up schedule? ensuring the coverage being returned matches what we would expect from a SafeRollout.

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+      feature,
+      revision,
+      mergeResult.result,
+      "auto-publish status change"
+    );
+  }
+  return status;
+}
+
+export function determineNextSnapshotAttempt(
```
ahdriel, line 160: nit: Should we name this `determineNextSafeRolloutSnapshotAttempt`?

I know this is in the SafeRollout file but when it is being used in different places it can be easy to get it mixed up (or selecting the wrong import from VSCode by mistake given they are all named the same).

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+
+export interface UpdateRampUpScheduleParams {
+  context: ReqContext | ApiReqContext;
+  safeRollout: SafeRolloutInterface;
+  status: SafeRolloutStatus;
+}
+export async function updateRampUpSchedule({
+  context,
+  safeRollout,
+  status,
```
ahdriel, line 34: question: Why do we need the status as a separate field? Can't we use `safeRollout.status`?

Any way we can name the field differently if we need it separate? Maybe `desiredStatus` or something?

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+  );
+  const safeRolloutStatus = getSafeRolloutResultStatus({
+    safeRollout: updatedSafeRollout,
+    healthSettings,
+    daysLeft,
+  });
+  let status: SafeRolloutStatus = updatedSafeRollout.status;
+  if (
+    safeRolloutStatus?.status &&
+    ["unhealthy", "rollback-now"].includes(safeRolloutStatus.status)
```
ahdriel, line 96: are we sure we want to rollback in all `unhealthy` scenarios? I think `No data` for example is a scenario that might happen, which might be safer to roll back rather than not, but just wanted to check.

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+    case "hours":
+      maxDurationInSeconds = safeRollout.maxDuration.amount * 3600;
+      break;
+    case "minutes":
+      maxDurationInSeconds = safeRollout.maxDuration.amount * 60;
+      break;
+    default:
+      throw new Error("Invalid max duration unit");
+  }
+  const fullRampUpTimeInSeconds = maxDurationInSeconds * 0.25; // hard coded for now this is the ramp up time
```
ahdriel, line 196: question: what is the 0.25 here?

packages/back-end/src/models/SafeRolloutModel.ts
```diff
@@ -33,10 +33,34 @@ export class SafeRolloutModel extends BaseClass {
   protected canDelete() {
     return true;
   }
+  protected migrate(
```
ahdriel, line 36: suggestion: I think we also need to add `autoRollback` here for old snapshots.

packages/back-end/src/validators/safe-rollout.ts
gazzdingo: this runs at the same time or more often but always runs the analysis

packages/back-end/src/controllers/features.ts
gazzdingo: we will go 100% at the end i just thought it was redundant to have it at 1 since the next step is 100%

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
gazzdingo: yeah no worries good point

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
gazzdingo: because the status can get updated in between this step eg you check to see if we should rollback then we would use that status to determine but thinking about it we should just skip this function if the new status is rolled back (i had it like this then moved it into its own file so i just copied it across without thinking about it)

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
gazzdingo: i asked luke about this and he was like anyway. my first opinion was to not revert is unhealthy since it might be some data issues they are willing to deal with. so let me remove this

packages/back-end/src/models/SafeRolloutModel.ts
gazzdingo: good point

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
```diff
@@ -0,0 +1,215 @@
...
+        ),
+        rampUpCompleted,
+        lastUpdate: new Date(),
+        nextUpdate: nextRampUp,
+      },
+    });
+  }
+}
+
+export async function checkAndRollbackSafeRollout({
```
ahdriel, line 67: question: Do we need to add a check here to see if `safeRollout.autoRollback` is set to true? Or is the check happening somewhere else?

packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -15,15 +15,32 @@ export type SafeRolloutStatus = typeof safeRolloutStatusArray[number];
...
 });
 // NB: These are the fields that the user submit, and the rest of the interface
 // is generated by us based on this information + feature & rule definition.
+const rampUpSchedule = z.object({
+  enabled: z.boolean(),
+  step: z.number(),
+  steps: z.array(
+    z.object({
+      percent: z.number(),
+      dateRampedUp: z.date(),
```
ahdriel, line 28: question: Should this be `.optional()`? Or what value will this have for the steps that have not been ramped up yet?

packages/back-end/src/util/features.ts
```diff
@@ -572,7 +577,17 @@ export function getFeatureDefinition({
...
+            if (
+              safeRollout?.rampUpSchedule.rampUpCompleted ||
+              !safeRollout?.rampUpSchedule.enabled
+            ) {
+              rule.coverage = 1; // Always 100% right now
+            } else {
+              rule.coverage =
+                safeRollout?.rampUpSchedule?.steps[
+                  safeRollout?.rampUpSchedule.step
+                ]?.percent ?? 1;
```
ahdriel, line 589: suggestion: Because of the `if` this is always defined, so we can remove the fallback to `1`.

```
rule.coverage =
  safeRollout.rampUpSchedule.steps[
    safeRollout.rampUpSchedule.step
  ].percent;
```

packages/back-end/src/validators/safe-rollout.ts
```diff
@@ -15,15 +15,32 @@ export type SafeRolloutStatus = typeof safeRolloutStatusArray[number];
...
+});
+export type RampUpSchedule = z.infer<typeof rampUpSchedule>;
+
 export const createSafeRolloutValidator = z.object({
   datasourceId: z.string(),
   exposureQueryId: z.string(),
   guardrailMetricIds: z.array(z.string()),
   maxDuration: MaxDuration,
+  autoRollback: z.boolean(),
+  rampUpSchedule: rampUpSchedule.partial().optional(),
```
ahdriel, line 43: question: Why do we have `partial()` here? What fields from rampUpSchedule do we expect to receive and which ones will we generate?

An alternative would be to use `Pick` to be explicit about which fields we need for creation.

gazzdingo: yeah we can use pick :)

packages/back-end/src/enterprise/saferollouts/safeRolloutUtils.ts
gazzdingo: i think that might be a bug

packages/back-end/src/util/features.ts
gazzdingo: i think we should keep this incase steps does not exist or if rampup steps goes out of bounds

ahdriel (approved):

PR #4062, opened 2025-05-13
Fix suspicious lift to rely on difference type when possible

lukesonnet:
### Features and Changes

Previous "suspicious lit" code relied just on the variation means, but that could disagree with the displayed lift if CUPED was enabled. This fixes this issue at least for Absolute and Relative difference types.

However, because we can't back out the % lift from scaled impact, Suspicious flags could show up for scaled impact but not others in this case.


packages/shared/src/experiments.ts
```diff
@@ -624,14 +625,24 @@ export function isSuspiciousUplift(
...
+    return (Math.abs(stats.expected ?? 0) / baseline.cr) >= maxPercentChange;
+  } else {
+    // This means scaled impact could show up as suspicious even when
+    // it doesn't show up for other difference types if CUPED is applied
+    // and CUPED causes the value to cross the threshold
+    return Math.abs(stats.cr - baseline.cr) / baseline.cr >= maxPercentChange;
+  }
 }
 
 export function isBelowMinChange(
```
lukebrawleysmith, line 648: this method also needs to be updated

lukebrawleysmith (commented):
one comment, otherwise lgtm

packages/shared/src/experiments.ts
```diff
@@ -624,14 +625,24 @@ export function isSuspiciousUplift(
...
   if (!baseline?.cr || !stats?.cr) return false;
 
   const maxPercentChange =
     metric.maxPercentChange ?? metricDefaults?.maxPercentageChange ?? 0;
 
-  return Math.abs(baseline.cr - stats.cr) / baseline.cr >= maxPercentChange;
+  if (differenceType === "relative") {
+    return (stats.expected ?? 0) >= maxPercentChange;
+  } else if (differenceType === "absolute") {
+    return (Math.abs(stats.expected ?? 0) / baseline.cr) >= maxPercentChange;
```
lukebrawleysmith, line 639: nit - I would put Math.abs() around `baseline.cr` in the denominator.  If the control mean is negative, and the lift is large negative, we want to throw an error.

lukebrawleysmith (approved):

PR #4065, opened 2025-05-13
feat(Features): Add requireProjectForFeatures setting + add validation of Projects when creating & updating features

### LINKED ISSUE #3427: [Feature] Force all feature flags to be created within a Project
Submitting on behalf of a large enterprise customer.

Require a project to be selected when creating a new feature flag. This could be provided as a global/org-wide configuration option in Settings so Admins can opt into the behavior (or not).

Could be as simple as:
- Add a project dropdown to the "create feature" prompt
- Add a config rule to require project selection
- Disable the create button in the "create feature" prompt until a project is selected

ahdriel:
### Features and Changes

Closes #3427

We're introducing a new organization setting called requireProjectForFeatures.

- This setting is turned off by default.
- When Enabled: If you turn this setting on, every new feature must be linked to a project.
  - UI: The "None" option for selecting a project will be hidden when creating or editing features.
  - API: If you try to create a new feature without specifying a project, an error will be returned.
  - Updating a feature
    - Features without a project: If a feature doesn't currently have a project assigned, you can still update its other details without assigning a project.
    - Features with a project: Once a feature is linked to a project, it follows the same rules as creation and will throw an error if we attempt to remove the project.

### Testing

Unit tests added
Manual testing UI
- Enable setting
  - See how Project is now required when creating a Feature
  - When editing a Feature information, observe how we cannot select `None` as the project

### Screenshots

Screenshot 2025-05-13 at 2 14 07 PM

### DIFF
--- a/packages/back-end/src/controllers/features.ts
+++ b/packages/back-end/src/controllers/features.ts
@@ -459,6 +459,15 @@ export async function postFeatures(
       "Feature keys can only include letters, numbers, hyphens, and underscores."
     );
   }
+
+  if (org.settings?.requireProjectForFeatures && !otherProps.project) {
+    throw new Error("Must specify a project for new features");
+  }
+  // Validate projects - We can remove this validation when FeatureModel is migrated to BaseModel
+  if (otherProps.project) {
+    await context.models.projects.ensureProjectsExist([otherProps.project]);
+  }
+
   const existing = await getFeature(context, id);
   if (existing) {
     throw new Error(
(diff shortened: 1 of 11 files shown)


packages/front-end/components/Features/EditFeatureInfoModal.tsx
```diff
@@ -83,7 +86,7 @@ const EditFeatureInfoModal: FC<{
             value={form.watch("project")}
             onChange={(v) => {
               form.setValue("project", v);
-              setShowProjectWarningMsg(true);
+              setShowProjectWarningMsg(v !== feature.project);
```
ahdriel, line 89: To fix a small UI bug where if I selected the same project the warning would show when in reality nothing was changing.

packages/back-end/src/controllers/features.ts
```diff
@@ -2114,6 +2123,19 @@ export async function putFeature(
...
+  // For a feature created before requireProjectForFeatures was enabled, allow updates to happen until the feature is associated with a project
+  if (
+    org.settings?.requireProjectForFeatures &&
+    feature.project &&
+    updates.project === ""
+  ) {
+    throw new Error("Must specify a project");
+  }
+  // Validate projects - We can remove this validation when FeatureModel is migrated to BaseModel
+  if (updates.project) {
```
jdorn, line 2135: Only do this check if the project is changing.

jdorn (approved):

PR #4069, opened 2025-05-14
Webhook Secrets

jdorn:
### Features and Changes

Adds a new secure way to include secret tokens in your webhook requests, either in the endpoint or in a header.

Previously, these would have to be entered as plain text in the webhook definition.  Any user with read access to your organization would be able to see this.

Now, you first define Webhook Secrets (key/value pairs).  Then, you can reference these by key in your webhook definitions (e.g. `https://example.com?secret={{ MY_SECRET }}`).  The real value is swapped in right before sending the request and is never revealed in the GrowthBook UI or in logs.


packages/front-end/components/EventWebHooks/EventWebHookAddEditModal/EventWebHookAddEditModal.tsx
```diff
@@ -237,7 +237,8 @@ const EventWebHookAddEditSettings = ({
           helpText={
             isDetailedWebhook && (
               <>
-                Must accept <code>{form.watch("method")}</code> requests
+                Must accept <code>{form.watch("method")}</code> requests.
+                Supports <strong>Webhook Secrets</strong>.
```
bryce-fitzsimons, line 241: link to docs?

packages/front-end/components/EventWebHooks/EventWebHookAddEditModal/EventWebHookAddEditModal.tsx
```diff
@@ -270,7 +271,10 @@ const EventWebHookAddEditSettings = ({
                 {!validHeaders ? (
                   <div className="alert alert-danger mr-auto">Invalid JSON</div>
                 ) : (
-                  <div>JSON format for headers.</div>
+                  <div>
+                    JSON format for headers. Supports{" "}
+                    <strong>Webhook Secrets</strong>.
```
bryce-fitzsimons, line 276: link to docs?

bryce-fitzsimons:
Should we also update the various Edge SDK docs that might benefit from secrets, such as
https://docs.growthbook.io/lib/edge/cloudflare#configuring-a-sdk-webhook ?

bryce-fitzsimons (approved):
conditionally approved, see comments

packages/back-end/src/models/WebhookSecretModel.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public async getAllForFrontEnd(): Promise<WebhookSecretFrontEndInterface[]> {
+    const docs = await this.getAll();
+    return docs.map((doc) => {
+      return omit(doc, ["value"]);
+    });
+  }
+
+  public async getBackEndSecretsReplacer(): Promise<(s: string) => string> {
```
romain-growthbook, line 50: This should be updated to handle string and objects so we don't have to have to duplicate complex or limited code at the calling level:

```typescript
    public async getBackEndSecretsReplacer<T extends string | Record<string, string>>(): Promise<(s: T) => string> {
      const secrets = await this.getAll();
      const replacements: Record<string, string> = {};
      for (const secret of secrets) {
        replacements[secret.key] = secret.value;
      }

      const stringReplacer = (s: string) => {
        const template = Handlebars.compile(s, {
          noEscape: true,
          strict: true,
        });
          return template(replacements);    
      }

    return (s: T) => {
      if (typeof s === "string") return stringReplacer(s);

      return Object.keys(s).reduce<Record<string, string>>((s, key) => ({
         ...s,
          [key]: stringReplacer(s[k])
        }), {})
    }
  }
```
```

Issue #4064, opened 2025-05-13
[Feature] Add Experiment Query Retrieval to API


kabam-blambert:
### Description of Feature

Add an endpoint to the growthbook API /queries/{queryId}.

Right now when using the API you can get details for an experiment and this includes query ids for all of the metrics in an experiment as far as I can tell. 

When doing post experiment deep dives I often need to modify these queries and so far I've done that manually. But it'd be great to automate this and have a programmatic way to retrieve an experiment's queries. 

This would probably be pretty easy to implement and I'd be happy to work on it.

kabam-blambert:
I see in the code that there is already a getQueriesbyIds method in the Query model. We could just expose that?

PR #4070, opened 2025-05-14
Add in a 'Queries' endpoint to API to be able to retrieve individual queries used in experiments.

kabam-blambert:
### Why
We'd like to automate some additional analyses that run after Growthbook experiments finish. Generally to do this we have to get all the relevant queries from the experiment results page and then manually alter the query in our analysis tools. This would let us automate this step and generally streamline the whole process.

### Features and Changes

- Added in a GET endpoint /queries/:id, which allows you to pull information related to queries used to generate experiment results. 
- Fix small issues in packages/back-end/README.md. Specifically some of the SQL queries in the 'data generation' section were incorrect. Replaced experiment_viewed with viewed_experiment and changed all instances of camelCase with snake case which seems to be what is used for column names in the generated data anyways.

- Closes **([Feature] Add Experiment Query Retrieval to API [#4064](https://github.com/growthbook/growthbook/issues/4064))**

### Dependencies

None

### Testing

I did not see many specific test suites for each endpoint. I followed the data generation flow, imported one of the experiments that were generated, and was able then to query the associated experiment query. This works as expected and returns an error message when searching for a query that doesn't exist. If I missed where test suites live I'm happy to add tests.


packages/back-end/src/routers/query/query.controller.ts
Kevin-Chant, line 1: I don't think we need the non-api controller/router, right? If we just want to expose an API route then the changes in the `/api` folder are sufficient.

packages/back-end/src/routers/query/query.router.ts
Kevin-Chant, line 1: Ditto

Kevin-Chant (approved):

PR #4071, opened 2025-05-14
early stopping for safe rollouts

lukebrawleysmith:
### Features and Changes

We want to be able to early stop a Safe Rollout if there is strong evidence of improvement.  This PR computes CTW for frequentist results, and uses CTW to look for evidence for early stopping.


packages/back-end/src/services/experiments.ts
```diff
@@ -3123,17 +3124,17 @@ export async function computeResultsStatus({
...
-
         const resultsStatus = getMetricResultStatus({
           metric,
           metricDefaults,
           baseline: baselineMetric,
           stats: currentMetric,
           ciLower,
-          ciUpper,
+          ciUpper: guardrailMetric
+            ? 0.95 ** (1 / numGuardrailMetrics)
```
lukesonnet, line 3136: I don't think we should do this like so because ciUpper is used in a few ways. Can we instead pass down something like `guardrailCorrectionData: {numGuardrails: number, epsilon: X}` which is optional in `getMetricResultStatus` and then if it exists compute the needed alpha and status in `getMetricResultStatus`?

packages/shared/src/experiments.ts
```diff
@@ -766,6 +766,7 @@ export function getMetricResultStatus({
       clearSignalResultsStatus = "lost";
     }
   }
+  const guardrailSafeStatus = (stats.chanceToWin ?? -1.0) > ciUpper;
```
lukesonnet, line 769: Can we compute `chanceToWin` here if chanceToWin is missing is missing instead of in gbstats, so that this makes no gbstats changes and we don't rely on any new data coming out of the stats engine? You can add a stats helper in this file and use the `normal` imported from statslib to make the right computation?

packages/back-end/src/services/experiments.ts
lukebrawleysmith: I now pass in number of guardrail metrics, and don't change ciUpper.

packages/shared/src/experiments.ts
lukebrawleysmith: done

packages/back-end/src/services/experimentNotifications.ts
```diff
@@ -324,6 +325,7 @@ export const computeExperimentChanges = async ({
         ciUpper,
         pValueThreshold,
         statsEngine: statsEngine,
+        guardrailAlpha: DEFAULT_GUARDRAIL_ALPHA,
```
lukesonnet, line 328: Let's just use the hardcoded `DEFAULT_GUARDRAIL_ALPHA` in `getMetricResultStatus` instead of wiring it through. There's even more work to let it be customizable so I'd set that elsewhere. It might feel somewhat trivial to extract it out to a constant but it helps set expectations that it is customizable at some point.

lukesonnet (approved):

lukesonnet (changes requested):
sorry misclick, meant to just comment, only blocking change is the `ci === undefined` comment.

packages/back-end/src/services/experimentNotifications.ts
lukebrawleysmith: done

packages/shared/src/experiments.ts
```diff
@@ -766,6 +768,18 @@ export function getMetricResultStatus({
       clearSignalResultsStatus = "lost";
     }
   }
+  const ciLowerGuardrail = stats.ci ? stats.ci[0] : Number.NEGATIVE_INFINITY;
```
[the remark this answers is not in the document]
lukebrawleysmith: Confirmed that if the interval is one-sided, then ci[0] equals `Number.NEGATIVE_INFINITY`.  I checked via the console log:
console.log("equality", stats.ci?.[0] === Number.NEGATIVE_INFINITY);

packages/shared/src/enterprise/power.ts
```diff
@@ -10,9 +10,31 @@ import {
   DEFAULT_SEQUENTIAL_TESTING_TUNING_PARAMETER,
 } from "shared/constants";
 import { MetricPowerResponseFromStatsEngine } from "back-end/types/stats";
-import { sequentialDiscriminant, sequentialRho } from "shared/power";
 import { eachDayOfInterval, formatISO, subDays } from "date-fns";
 
+export function sequentialRho(
```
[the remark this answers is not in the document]
lukebrawleysmith: removed duplication

packages/shared/src/experiments.ts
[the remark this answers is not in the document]
lukesonnet: This logical statement is checking if `stats.ci` exists, not if `stats.ci[0]` exists. I feel like it should be checking the latter, otherwise, you just end up with either stats.ci as is  if stats.ci is defined or [-Inf, Inf] if stats.ci is undefined.

That does not appear to be the right logic.

packages/back-end/src/validators/experiments.ts
```diff
@@ -189,7 +189,7 @@ export type ExperimentAnalysisSummaryHealth = z.infer<
 export const goalMetricStatus = ["won", "lost", "neutral"] as const;
 export type GoalMetricStatus = typeof goalMetricStatus[number];
 
-export const guardrailMetricStatus = ["lost", "neutral"] as const;
+export const guardrailMetricStatus = ["safe", "lost", "neutral"] as const;
```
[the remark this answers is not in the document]
lukebrawleysmith: Good callout.  I added a test, and I think we are good here.

packages/back-end/src/services/experimentNotifications.ts
lukesonnet: I'm still seeing this passed through a bunch of places in this PR. Did you push these changes up?

packages/shared/src/enterprise/decision-criteria/constants.ts
```diff
@@ -76,6 +76,26 @@ export const DO_NO_HARM_DECISION_CRITERIA: DecisionCriteriaData = {
   defaultAction: "rollback",
 };
 
+export const EARLY_SAFE_ROLLOUT_DECISION_CRITERIA: DecisionCriteriaData = {
```
lukesonnet, line 79: Does this get used anywhere? Let's remove it if not.

packages/shared/src/experiments.ts
```diff
@@ -798,6 +799,19 @@ export function getMetricResultStatus({
       clearSignalResultsStatus = "lost";
     }
   }
+  const ciLowerGuardrail = stats.ci?.[0] ?? Number.NEGATIVE_INFINITY;
```
lukesonnet, line 802: Now, if `stats.ci` is missing altogether, you will get `[-Inf, Inf]` as the guardrails. This is probably a weird edgecase where some stats are missing, but will we get "safe" in this case? I think we'd want to default to it not being safe in this case. Either we can hope it falls through or we can wrap all this in something like:

```
let guardrailSafeStatus: boolean = false;
if (stats.ci) {
  const ciLowerGuardrail = stats.ci[0] ?? Number.NEGATIVE_INFINITY;
  ...
}
```

packages/shared/src/experiments.ts
```diff
@@ -806,9 +820,52 @@ export function getMetricResultStatus({
...
+  lower: number,
+  upper: number,
+  pValueThreshold: number,
+  inverse: boolean = false
+): number {
+  if (
+    lower === Number.NEGATIVE_INFINITY &&
+    upper === Number.POSITIVE_INFINITY
+  ) {
+    return 0;
```
lukesonnet, line 838: Oh, I see this covers the case I was talking about in my other comment. I guess 0 will make sure it doesn't return "safe", so this is probably fine, but I wonder if it's better to be more explicit instead of relying on just this check (which should still remain, of course).

lukesonnet (approved):
Some more comments, but they're just nits. If you're happy with the stats here and that the new "safe" status won't change our existing decision criteria logic for lost/neutral, it's good with me.

packages/shared/src/enterprise/decision-criteria/constants.ts
lukebrawleysmith: removed

packages/shared/src/experiments.ts
lukebrawleysmith: good call, and updated

packages/shared/src/experiments.ts
lukebrawleysmith: thx, I accepted your previous comment

packages/back-end/src/services/experimentNotifications.ts
lukebrawleysmith: It's currently imported only into `shared/src/experiments.ts`, so I think we are ok.

lukesonnet: Sorry this comment should have been deleted before submitting my review, I was looking at outdated code.

lukebrawleysmith:
Thanks Luke.  I double checked, and I see anywhere in our prod code that we use for `neutral` for decision making for guardrails.  Also, I reran my guardrail sim, I calculated the probability of stopping due to CTW > 0.95, and it is less than 0.05 for 2-sided, one-sided greater and one-sided lesser.

PR #4072, opened 2025-05-15
Update sdkInfo generation process and automate it

royalfig:
- Updated template for larger doc support
- Made the autogenerated file more obvious
- Added automation to the GH Action so that versions are updated on every docs build


docs/package.json
```diff
@@ -15,7 +15,7 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "typecheck": "tsc",
-    "gen-sdk-resources": "ts-node --skipProject ./scripts/gen-sdk-resources.ts"
+    "gen-sdk-resources": "npx -y tsx ./scripts/gen-event-webhook-doc.ts"
```
bryce-fitzsimons, line 18: is this meant to call `gen-sdk-resources.ts` or is there a new entrypoint?

PR #4076, opened 2025-05-15
feat(timeseries): Create TimeSeries data for Safe Rollouts

ahdriel:
### Features and Changes

Save the underlying data for TimeSeries for Safe Rollouts. It uses the same strategy as Experiments and integrates into MetricTimeSeries.

* Updating Safe Rollout Snapshots now will save a per-metric data point in the MetricTimeSeries collection
* Added new endpoint to fetch the time series

UI is being worked on a separate PR.


packages/back-end/src/validators/metric-time-series.ts
```diff
@@ -58,7 +58,7 @@ export const metricTimeSeriesSchema = z
     dateUpdated: z.date(),
 
     metricId: z.string(),
-    source: z.enum(["experiment"]),
+    source: z.enum(["experiment", "safe-rollout"]),
```
lukesonnet, line 61: Just checking: Is it a problem that sometimes in safe rollouts one side or the other of the CI is always Infinity? I see in the datatype for the CI it says the values must be a number. Will this work if we have one end of the CI is infinity? I think it should, but worth testing.

ahdriel: Good point! Zod does treat `Infinity` as a valid `z.number()`, but after parsing it actually becomes `null` instead of keeping it as Infinity 🤔 (playground).

I'll think more about the implications of this and what we might need to update to account for that.

ahdriel: Ohh my understanding was wrong. Zod does properly treat `Infinity` as a valid number and accepts it, and it is properly stored in our Database.

The playground though (and our API) show the result parsed into JSON and the JSON schema does not accept Infinity as a valid number value, therefore it gets converted to `null`.

So while this is indeed a problem, it is in a different layer (fetching the data from the client) instead of Zod validation.

ahdriel: And confirmed we are good because we updated the types to be `null | number` to account for this and handle it by falling back to `Infinity` ([example](https://github.com/growthbook/growthbook/blob/main/packages/front-end/components/Experiment/ResultsTableTooltip/ResultsTableTooltip.tsx#L232)).

packages/back-end/src/services/safeRolloutTimeSeries.ts
```diff
@@ -0,0 +1,210 @@
...
+export async function updateSafeRolloutTimeSeries({
+  context,
+  safeRollout,
+  safeRolloutSnapshot,
+}: {
+  context: ReqContext;
+  safeRollout: SafeRolloutInterface;
+  safeRolloutSnapshot: SafeRolloutSnapshotInterface;
+}) {
+  // TODO: Confirm if we want the same behavior here initially
```
lukesonnet, line 33: Yes.

packages/back-end/src/services/safeRolloutTimeSeries.ts
```diff
@@ -0,0 +1,210 @@
...
+    return;
+  }
+
+  const metricGroups = await context.models.metricGroups.getAll();
+  const metricsIds = expandMetricGroups(
+    safeRollout.guardrailMetricIds,
+    metricGroups
+  );
+
+  // TODO: Do we have more than one analysis type for safeRollout?
```
lukesonnet, line 50: No, we shouldn't. In the future maybe we will, but for now it's just absolute analysis. Seems like this code could be agnostic to this, though? Not sure, happy to just look at the first for now.

packages/back-end/src/services/safeRolloutTimeSeries.ts
```diff
@@ -0,0 +1,210 @@
...
+  if (variations.length !== 2) {
+    logger.warn(
+      `Safe Rollout ${safeRollout.id} has ${variations.length} variations, expected 2`
+    );
+  }
+
+  let factMetrics: FactMetricInterface[] | undefined = undefined;
+  let factTableMap: Map<string, FactTableInterface> | undefined = undefined;
+  const factMetricsIds: string[] = metricsIds.filter(isFactMetricId);
+  if (factMetricsIds.length > 0) {
```
lukesonnet, line 69: nice

packages/back-end/src/services/safeRolloutTimeSeries.ts
```diff
@@ -0,0 +1,210 @@
...
+    factMetrics = await context.models.factMetrics.getByIds(factMetricsIds);
+    factTableMap = await getFactTableMap(context);
+  }
+
+  const timeSeriesVariationsPerMetricId = metricsIds.reduce((acc, metricId) => {
+    acc[metricId] = variations.map((_, variationIndex) => ({
+      id: safeRolloutSnapshot.settings.variations[variationIndex].id,
+      name: variationIndex === 0 ? "control" : "variation",
```
lukesonnet, line 77: Do we have these names hard coded somewhere in the safe rollout interface? I couldn't find in a brief check.

packages/back-end/src/services/safeRolloutTimeSeries.ts
```diff
@@ -0,0 +1,210 @@
...
+      stats:
+        analysisResults?.variations[variationIndex]?.metrics[metricId]?.stats,
+      absolute: convertMetricToMetricValue(
```
lukesonnet, line 80: nit: You could check the analysis settings here to be more future proof?

packages/back-end/src/validators/metric-time-series.ts
lukesonnet: Very annoying that we can't pass Infinity back and forth

lukesonnet (approved):
Looks good to me. Left a couple nits.

packages/back-end/src/services/safeRolloutTimeSeries.ts
ahdriel: We don't store that information in a SafeRolloutSnapshotAnalysis -- we construct and have `absolute` as hardcoded when sending the data to gbstats; so I am unable to check the analysis type via the settings here at the moment.

packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
```diff
@@ -217,3 +218,41 @@ export async function putSafeRollout(
...
+  }
+
+  const { id } = req.params;
+  const safeRollout = await context.models.safeRollout.getById(id);
+  if (!safeRollout) {
+    throw new Error("Safe rollout not found");
+  }
+
+  const timeSeries = await context.models.metricTimeSeries.getBySourceAndMetricIds(
+    "safe-rollout",
```
gazzdingo, line 247: Nit: would be nice to have this as named params since the list is getting a bit long

gazzdingo (approved):
Nothing blocking on my end

PR #4077, opened 2025-05-16
saferollout timeseries UI

ahdriel:
### Features and Changes

### DIFF
--- a/packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
+++ b/packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
@@ -40,6 +40,7 @@ import { QueryStatusData } from "@/components/Queries/RunQueriesButton";
 import { useDefinitions } from "@/services/DefinitionsContext";
 import ResultsMetricFilter from "@/components/Experiment/ResultsMetricFilter";
 import { ResultsMetricFilters } from "@/components/Experiment/Results";
+import SparklineGraph from "@/components/Experiment/SparklineGraph";
 import Tooltip from "@/components/Tooltip/Tooltip";
 import { SSRPolyfills } from "@/hooks/useSSRPolyfills";
 import ChangeColumn from "./ChangeColumn";
@@ -80,7 +81,7 @@ export type ResultsTableProps = {
 };
 
 export default function ResultsTable({
-  id: _,
+  id,
   isLatestPhase,
   status,
   queryStatusData,
@@ -333,6 +334,12 @@ export default function ResultsTable({
                 {!noMetrics ? (
                   <>
                     <th
+                      style={{ width: 130 * tableCellScale }}
+                      className={clsx("axis-col label", { noStickyHeader })}
+                    >
+                      Graph
+                    </th>
+                    <th
                       style={{ width: 120 * tableCellScale }}
                       className={clsx("axis-col label", { noStickyHeader })}
                     >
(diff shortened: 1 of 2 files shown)


packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
```diff
@@ -510,6 +553,20 @@ export default function ResultsTable({
...
+                            <td style={{ padding: 0, height: 1 }}>
+                              <SafeRolloutTimeSeriesGraph
+                                data={metricTimeSeries}
+                                xExtent={metricTimeSeriesDateExtent}
+                                ssrPolyfills={ssrPolyfills}
+                              />
+                            </td>
+                          ) : j > 0 && showTimeSeries ? (
+                            <td></td>
+                          ) : null}
```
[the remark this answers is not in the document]
msamper: hmm I think `Within bounds` is still accurate because there is data, just not time series data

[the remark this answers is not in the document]
ahdriel: Screenshot 2025-05-28 at 11 08 12 AM

Done -- added a message when we don't have Time Series data for a metric.

packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
```diff
@@ -253,9 +253,22 @@ export const getSafeRolloutTimeSeries = async (
...
+  // If the rollout variation has a CI of 0,0, remove the data point as it is invalid
+  const filteredTimeSeries = timeSeries.filter((ts) => {
+    return (
+      ts.dataPoints.filter((dp) => {
+        return (
+          dp.variations[1].absolute?.ci &&
+          dp.variations[1].absolute.ci[0] !== 0 &&
+          dp.variations[1].absolute.ci[1] !== 0
+        );
+      }).length > 0
```
msamper, line 265: You could do a `.some()` here I think and it might make it easier to read since you're just checking that the 0,0 CI condition is true for some data point instead of actually needing the filtered data

packages/front-end/components/Experiment/SafeRolloutTimeSeriesGraph.tsx
```diff
@@ -0,0 +1,514 @@
...
+  getMetricFormatter,
+} from "@/services/metrics";
+import { useDefinitions } from "@/services/DefinitionsContext";
+import { useCurrency } from "@/hooks/useCurrency";
+import { SSRPolyfills } from "@/hooks/useSSRPolyfills";
+import styles from "./SafeRolloutTimeSeriesGraph.module.scss";
+
+type SafeRolloutTimeSeriesGraphProps = {
+  data: MetricTimeSeries;
+  xExtent?: [undefined, undefined] | [Date, Date];
```
msamper, line 31: nit: `xRange` or `xDateRange` might be more clear?

msamper: actually I see you're using a function called extent so maybe that name is fine

msamper (commented):
Think I found a bug. When you update the snapshot for a safe rollout the time series graph doesn't seem to update until after you click away from the browser and click back in

packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
```diff
@@ -638,3 +704,11 @@ function getChangeTooltip(changeTitle: string, differenceType: DifferenceType) {
...
   return <>{changeElem}</>;
 }
+
+function Message({ children }: { children: React.ReactNode }) {
+  return (
+    <Flex align="center" justify="center" position="relative" width="100%">
+      {children}
+    </Flex>
+  );
+}
```
mknowlton89: Small nit: This is only used once in this file and no where else - should we just do this inline?

PR #4078, opened 2025-05-16
[OpenAPI] Add custom fields to public APIs for features and experiments

msamper:
### Features and Changes


jdorn (commented):
I think a better solution here is to let users specify an `id` when creating a custom field instead of auto-generating one.  We can have strict validation on this to only allow `[a-z0-9_-]`.  Then, we can just use the custom field id in the REST API without needing to look up a mapping.  It will also be much more robust because `id` can be readonly after creation, whereas `name` can be changed at any time.

This will be a little annoying for existing custom fields with opaque, auto-generated ids, but I think that's ok.  Custom fields aren't used that much right now, so it's a good time to make this change before they get more popular.

docs/src/partials/event-webhook/_event-webhook-list.md
```diff
@@ -675,6 +681,9 @@ Triggered when a feature is updated
                 date: string;
                 publishedBy: string;
             } | undefined;
+            customFields?: ({
+                [x: string]: any;
+            } | undefined) | undefined;
```
ahdriel, line 782: question: Why is this one defined differently than the others?

I see that it is the same for other properties so there's likely a reason that I am not aware of

packages/back-end/src/models/CustomFieldModel.ts
```diff
@@ -51,12 +50,11 @@ export class CustomFieldModel extends BaseClass {
     if (!customFields) {
       return null;
     }
-    customFields.fields.forEach((field) => {
+    return customFields.fields.find((field) => {
```
ahdriel: nit: This is changing the return type to be `customField | null | undefined`

Should we add a `|| null` to not have `undefined` as a possible value here?

docs/src/partials/event-webhook/_event-webhook-list.md
msamper: That's a good question.. I'm not sure. This is auto-generated but let me see if I can make some sense of it

packages/back-end/src/models/CustomFieldModel.ts
msamper: Good catch! Yeah I think we should. Let me fix

docs/src/partials/event-webhook/_event-webhook-list.md
ahdriel: I think it is related to this list having only the changed attributes, so we might not have customFields defined in the object at all

PR #4086, opened 2025-05-16
REST API: implement SDK clientKey filter for listFeatures

bryce-fitzsimons:
Pass in an optional `?clientKey=sdk-abc123` to filter responses by what is in the SDK payload. Useful for 3rd party integrations like Vercel's flag explorer toolbar

### DIFF
--- a/packages/back-end/src/api/features/listFeatures.ts
+++ b/packages/back-end/src/api/features/listFeatures.ts
@@ -1,3 +1,4 @@
+import { getConnectionSDKCapabilities } from "shared/sdk-versioning";
 import { getFeatureRevisionsByFeaturesCurrentVersion } from "back-end/src/models/FeatureRevisionModel";
 import { ListFeaturesResponse } from "back-end/types/openapi";
 import { getAllPayloadExperiments } from "back-end/src/models/ExperimentModel";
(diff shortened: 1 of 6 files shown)


packages/back-end/src/api/features/listFeatures.ts
```diff
@@ -24,13 +27,44 @@ export const listFeatures = createApiRequestHandler(listFeaturesValidator)(
       req.query.projectId
     );
 
+    // If SDK clientKey is provided, get the SDK connection and use its projects/environment
+    let filteredFeatures = features;
+    if (req.query.clientKey) {
+      const sdkConnection = await findSDKConnectionByKey(req.query.clientKey);
+      if (!sdkConnection) {
+        throw new Error("Invalid SDK connection key");
+      }
```
jdorn, line 36: Need to also verify that the sdk connection is part of the organization.

packages/back-end/src/api/features/listFeatures.ts
```diff
@@ -24,13 +27,44 @@ export const listFeatures = createApiRequestHandler(listFeaturesValidator)(
...
+
+      const payload = await getFeatureDefinitions({
+        context: req.context,
+        capabilities: getConnectionSDKCapabilities(sdkConnection),
+        environment: sdkConnection.environment,
+        projects: sdkConnection.projects,
+        encryptionKey: sdkConnection.encryptPayload
+          ? sdkConnection.encryptionKey
+          : "",
```
jdorn, line 45: We should always pass in an empty string (or undefined) here.  Otherwise, the response might be encrypted and `payload.features` will be an empty object.

bryce-fitzsimons: good catch

jdorn (approved):

PR #4090, opened 2025-05-19
[fix] Reordering project-scoped environments is unpredictable

gazzdingo:
### Features and Changes

When re-ordering environments, the local array index within the UI was being used instead of the global index in the organization's full environment list.  Normally, these are identical so there are no issues, but if some environments are project-scoped AND the UI is filtered to a single project, the array indexes could get out of sync.

Out-of-sync indexes could cause one of three potential bugs (depending on multiple factors):
- Duplicate environment ids
- Deleted environment ids
- Incorrect ordering

Luckily, all of these outcomes are fully reversible without data loss (even a deleted environment can just be recreated with the same name and it picks up where it left off).  However, this bug can cause serious disruption to an organization's work flow.

This PR fixes the root problem of mismatched indexes and adds multiple safety nets to avoid similar issues in the future.  Even if another front-end bug with indexes is introduced in the future (hopefully not!), the back-end will now reject the request if something doesn't look right.

### DIFF
--- a/packages/back-end/src/routers/environment/environment.controller.ts
+++ b/packages/back-end/src/routers/environment/environment.controller.ts
@@ -48,7 +48,7 @@ export const putEnvironmentOrder = async (
 ) => {
   const context = getContextFromReq(req);
   const { org } = context;
-  const { envId, direction } = req.body;
+  const { envId, newIndex } = req.body;
   const existingEnvs = org.settings?.environments;
 
   if (!existingEnvs) {
@@ -77,16 +77,15 @@ export const putEnvironmentOrder = async (
   }
 
   const updatedEnvs = [...existingEnvs];
-  const newEnvIndex = envIndex + direction;
 
-  if (newEnvIndex < 0 || newEnvIndex >= existingEnvs.length) {
+  if (newIndex < 0 || newIndex >= existingEnvs.length) {
     return res.status(400).json({
       status: 400,
-      message: `Invalid direction: ${direction}`,
+      message: `Invalid new index: ${newIndex}`,
     });
   }
   updatedEnvs.splice(envIndex, 1);
-  updatedEnvs.splice(newEnvIndex, 0, updatedEnvs[envIndex]);
+  updatedEnvs.splice(newIndex, 0, updatedEnvs[envIndex]);
 
   try {
     await updateOrganization(org.id, {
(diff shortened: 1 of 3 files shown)


packages/front-end/pages/environments.tsx
```diff
@@ -207,13 +207,14 @@ const EnvironmentsPage: FC = () => {
...
-                                newEnvs.splice(i - 1, 0, e);
+                                const newIndex = environments.findIndex(
+                                  (env) => env.id === e.id
+                                );
                                 await apiCall(`/environment/order`, {
                                   method: "PUT",
                                   body: JSON.stringify({
-                                    environments: newEnvs.map((env) => env.id),
+                                    envId: e.id,
+                                    newIndex: newIndex - 1,
```
bryce-fitzsimons, line 217: should actually get the global index of `filteredEnvironments[i-1]`

packages/front-end/pages/environments.tsx
```diff
@@ -227,13 +228,14 @@ const EnvironmentsPage: FC = () => {
...
-                                newEnvs.splice(i + 1, 0, e);
+                                const index = environments.findIndex(
+                                  (env) => env.id === e.id
+                                );
                                 await apiCall(`/environment/order`, {
                                   method: "PUT",
                                   body: JSON.stringify({
-                                    environments: newEnvs.map((env) => env.id),
+                                    envId: e.id,
+                                    newIndex: index + 1,
```
bryce-fitzsimons, line 238: should actually get the global index of `filteredEnvironments[i+1]`

bryce-fitzsimons (changes requested):
small changes for UI

packages/back-end/src/routers/environment/environment.controller.ts
```diff
@@ -68,21 +67,25 @@ export const putEnvironmentOrder = async (
...
-    updatedEnvs.push(env);
-  });
+  if (newIndex < 0 || newIndex >= existingEnvs.length) {
+    return res.status(400).json({
+      status: 400,
+      message: `Invalid new index: ${newIndex}`,
+    });
+  }
+  updatedEnvs.splice(envIndex, 1);
+  updatedEnvs.splice(newIndex, 0, updatedEnvs[envIndex]);
```
jdorn, line 88: Won't this break because at this point `updatedEnvs[envIndex]` doesn't exist anymore?

packages/front-end/pages/environments.tsx
jdorn: If we change this, we also need to change the render condition for this button.  It currently does:

```ts
i < environments.length - 1 &&
```

jdorn (approved):

bryce-fitzsimons (approved):

PR #4092, opened 2025-05-20
Create AIPompts model and add AI suggestions.

tzjames:
### Features and Changes

Adds AI Prompts model and incorporates AI suggestions in the following extra places:
  `experiment-analysis`: To create a short summary of experiment results when stopped
  `metric-description`: To create a description of a metric based upon the sql,
  `experiment-hypothesis`: To reformat an experiment hypothesis to match a particular format.
 `visual-changeset-copy-transform-energetic`: Pre-existing reformatting of copy in the visual editor to be more energetic
  `visual-changeset-copy-transform-concise`: Pre-existing reformatting of copy in the visual editor to be more concise
  `visual-changeset-copy-transform-humorous`: Pre-existing reformatting of copy in the visual editor to be more humorous

Adds a front end settings page where the default prompts can be overwritten.

Adds logging of all AI usages to clickhouse when on cloud.

TODO:
- [ ] Incorporate any new design updates
- [ ] Add model and temperature overrides to each prompt.
- [ ] Show error message on the front end when they go over their limits.


packages/back-end/src/services/organizations.ts
```diff
@@ -164,6 +165,28 @@ export function getConfidenceLevelsForOrg(context: ReqContext) {
...
 
+export function getAISettingsForOrg(
+  context: ReqContext,
+  includeKey: boolean = false
+): {
+  aiEnabled: boolean;
+  openAIAPIKey: string;
+  openAIDefaultModel: TiktokenModel;
+} {
+  const aiEnabled = IS_CLOUD
```
tzjames, line 189: Do we want it to be cloud only?  If they are providing their own `openAIAPIKey` I don't see why we should limit to cloud.

packages/shared/src/ai.ts
```diff
@@ -0,0 +1,49 @@
...
+    "\nFor the Conclusions where you state the name of the winning variation and a good, but brief rationale behind that based upon the results of the metrics." +
+    "\nThere should be no other headers or sections in your response besides '### Key Findings' and '### Conclusions'" +
+    "\nYour output should be in the form of a markdown text, and should be no longer than 2000 characters." +
+    "\nIt should not be wrapped in triple backticks." +
+    "\nThere is no need to provide metric ids in your answer",
+  "metric-description":
+    "Write a concise description in markdown of the metric that will be helpful for other users who may want to use this metric in their AB tests. Paraphrase what this metric is used to show or measure.",
+  "experiment-hypothesis":
+    "A hypothesis is a statement that can be tested. It should be clear, concise, specific, and falsifiable. It should include how the user or product behaviour is expected to change, and what metrics we're trying to move with this experiment. It does not need a title.",
+  "similar-experiments": "", // STOPSHIP(Graham): This doesn't seem to be used anywhere yet in current branch?
```
tzjames: @Auz Did you check in all your similar-experiments code.  I don't see any other uses when I search "similar-experiments"

packages/back-end/src/services/organizations.ts
Auz: I dont think we are...

packages/front-end/components/License/OptInModal.tsx
```diff
@@ -0,0 +1,169 @@
...
+    noPermission: (
+      <>
+        You must be an administrator to enable this feature. Please contact your
+        administrator.
+      </>
+    ),
+    version: "2025-06-19",
+  },
+  "managed-warehouse": {
+    // STOPSHIP(graham): Update this text to match the actual agreement
```
tzjames: @graham I added this to fix `yarn type-check`.  Though maybe it doesn't belong in this PR at all.

docs/docs/self-host/env.mdx
```diff
@@ -139,6 +139,12 @@ Store uploads in a Google Cloud Storage bucket.
 - **GOOGLE_APPLICATION_CREDENTIALS** (not required when deployed to GCP with an instance service account)
 - **USE_FILE_PROXY** set this to true for access to uploads to proxy through your self hosted server allowing you to keep the bucket private.
 
+## Open AI Key
```
jdorn: OpenAI should not have a space

packages/back-end/src/models/AgreementModel.ts
```diff
@@ -0,0 +1,59 @@
...
+    return this.canCreate();
+  }
+
+  public getAgreementsForOrg = async () => {
+    const existing = await this._find({
+      organization: this.context.org.id,
+    });
+
+    return existing ? existing : [];
+  };
```
jdorn: Nit: The built-in `getAll()` method already does this exact thing.

packages/back-end/src/models/AgreementModel.ts
```diff
@@ -0,0 +1,59 @@
...
+  public getAgreementForOrg = async (agreement: AgreementType) => {
+    const existing = await this._findOne({
+      organization: this.context.org.id,
```
jdorn: You don't need to pass in `organization` here.  That is added automatically to every query.

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +205,227 @@ export async function deleteSavedQuery(
...
+    datasource.id,
+    context.org.id
+  );
+  if (!informationSchema) {
+    return res.status(404).json({
+      status: 404,
+      message: "No informationSchema found for the datasource",
+    });
+  }
+  const maxTables = 100;
```
jdorn: This could get really expensive.  Most data warehouses are large and we will often hit this 100 limit.  It's also not uncommon for tables to have hundreds of columns, which might lead to a ton of token usage and unfocused results.

I wonder if a better approach is to do this in 2 steps.
1. Give only a list of table names (not columns) and have the AI return the ones it thinks it will need to answer the question.
2. Do the full prompt, but only include the tables and full schemas from step 1

The nice thing is we should already have the list of table names for every data source already.  We then only need to fill out missing schemas for a few tables at a time instead of blocking on all 100.

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +205,227 @@ export async function deleteSavedQuery(
...
+  const instructions =
+    "You are a data analyst and SQL expert.\n" +
+    "Generate a SQL query to answer the provided question inputted. The resultant query should be ONLY based on the provided context of tables and structure, and return only valid SQL that is executable on the specified data source." +
+    "\n\nTable structure: " +
+    "\n" +
+    schemasString +
+    "\n\nInput: " +
+    input +
+    "\n\nKeep in mind that the table names may be sharded, so you should use the generic table names for tables that have wild cards, such as 'events_*'. If you are querying sharded tables, and you're asked for a date range, you should use something like: \n((_TABLE_SUFFIX BETWEEN '{{date startDateISO \"yyyyMMdd\"}}' AND '{{date endDateISO \"yyyyMMdd\"}}') OR\n" +
+    "   (_TABLE_SUFFIX BETWEEN 'intraday_{{date startDateISO \"yyyyMMdd\"}}' AND 'intraday_{{date endDateISO \"yyyyMMdd\"}}'))\n\n when constructing the query to make sure it's fast and takes use of the sharding. If you do this code, be sure to replace the dates with the correct range.\n";
```
jdorn: Only add these lines about sharding and _TABLE_SUFFIX if the datasource type is `bigquery`.  That is very unique to BigQuery and will just be confusing if used with other engines.

Also, I would add the database type (e.g. `clickhouse` or `postgres`) to the prompt.  Different databases use different syntax, so this is important.

packages/back-end/src/models/MetricModel.ts
```diff
@@ -550,3 +551,33 @@ export async function removeTagInMetrics(organization: string, tag: string) {
     }
   );
 }
+
+export async function generateMetricEmbeddings(
+  context: ReqContext | ApiReqContext,
+  metricsToGenerateEmbeddings: MetricInterface[]
```
jdorn, line 557: This is the legacy metric interface. We should really only be supporting fact metrics going forward with new features like this.

Auz: I dont think this code is used currently. I can remove it if you like

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
Auz: Done

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
Auz: Done

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -361,20 +431,32 @@ export async function postGenerateSQL(
...
+    "\n\n" +
+    "Use date ranges if possible and it makes sense with the input. If the input requests a date range, like 'in the past month', use the current date: " +
+    new Date().toISOString() +
+    " as the starting point.";
+
+  if (datasource.type === "bigquery") {
+    instructions +=
+      "\n\nBe sure to use the schema name in the query, ie: databaseName.schemaName.tableName.\n" +
+      "Keep in mind that the table names may be sharded, so you should use the generic table names for tables that have wild cards, such as 'events_*'. If you are querying sharded tables, and you're asked for a date range, you should use something like: \n((_TABLE_SUFFIX BETWEEN '{{date startDateISO \"yyyyMMdd\"}}' AND '{{date endDateISO \"yyyyMMdd\"}}') OR\n" +
+      "   (_TABLE_SUFFIX BETWEEN 'intraday_{{date startDateISO \"yyyyMMdd\"}}' AND 'intraday_{{date endDateISO \"yyyyMMdd\"}}'))\n\n when constructing the query to make sure it's fast and takes use of the sharding. If you do this code, be sure to replace the dates with the correct range.\n";
```
jdorn: The `{{...}}` macros won't work in the SQL Explorer.  The example here should just hardcode dates

```sql
_TABLE_SUFFIX BETWEEN '20250101' AND '20250131'
```

And I would remove the `intraday_` part since it's confusing.

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -361,20 +431,32 @@ export async function postGenerateSQL(
...
+      return `Database: ${value.databaseName}, Table: ${value.tableName}, Schema name: ${value.tableSchema}, Columns: [${columnsDescription}]`;
     }
   ).join("\n");
 
-  const instructions =
+  let instructions =
     "You are a data analyst and SQL expert.\n" +
     "Generate a SQL query to answer the provided question inputted. The resultant query should be ONLY based on the provided context of tables and structure, and return only valid SQL that is executable on the specified data source." +
+    "\n\nThe database is a " +
+    datasource.type +
```
jdorn: Will need to special case `growthbook_clickhouse` to just render as `clickhouse` here.

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
jdorn: I would suggest either always doing the 2-step process or setting this limit much lower, like 10.

Auz: switched to 10.

packages/back-end/src/services/organizations.ts
```diff
@@ -176,6 +177,27 @@ export function getConfidenceLevelsForOrg(context: ReqContext) {
...
+  includeKey: boolean = false
+): {
+  aiEnabled: boolean;
+  openAIAPIKey: string;
+  openAIDefaultModel: TiktokenModel;
+} {
+  const openAIKey = process.env.OPENAI_API_KEY || "";
+  const aiEnabled = IS_CLOUD
+    ? context.org.settings?.aiEnabled !== false
+    : !!(context.org.settings?.aiEnabled && openAIKey);
```
jdorn, line 191: This logic doesn't feel right.  Even on cloud, `openAIKey` should be required for ai to be enabled.  And for self-hosted, do we still have the org setting or is the existence of the env var enough?

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+      status: 404,
+      message: "AI configuration not set or enabled",
+    });
+  }
+  if (!req.organization) {
+    return res.status(404).json({
+      status: 404,
+      message: "Organization not found",
+    });
+  }
```
jdorn: Nit: this check is not necessary

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+      "The list should be in the format: 'databaseName.schemaName.tableName'. " +
+      "The tables must be one of the following tables:" +
+      "\n" +
+      tablesInfo
+        .map(
+          (table) =>
+            `${table?.databaseName}.${table?.schemaName}.${table?.tableName}`
+        )
+        .join(", ") +
+      "Return at most 20 tables. Return only the table names, separated by commas, without any additional text or explanations.";
```
jdorn: Is this conflicting since above you say to return `databaseName.schemaName.tableName` and here you say to only return table name.  Also, I believe OpenAI lets you provide a JSON schema for the result.  Is it worth doing that here to make sure an array of strings is returned?

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+
+  let instructions =
+    "You are a data analyst and SQL expert.\n" +
+    "Generate a SQL query to answer the provided question inputted. The resultant query should be ONLY based on the provided context of tables and structure, and return only valid SQL that is executable on the specified data source." +
+    "\n\nThe database is a " +
+    (datasource.type === "growthbook_clickhouse"
+      ? "clickhouse"
+      : datasource.type) +
+    " database. Be sure to make queries valid for that database type." +
+    (isGA ? " This database and tables from Google Analytics." : "") +
```
jdorn: "are from"

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+
+  if (datasource.type === "bigquery") {
+    instructions +=
+      "\n\nBe sure to use the schema name in the query, ie: databaseName.schemaName.tableName.\n" +
+      "Keep in mind that the table names may be sharded, so you should use the generic table names " +
+      "for tables that have wild cards, such as 'events_*'. If you are querying sharded tables, and " +
+      "you're asked for a date range, you should use something like: " +
+      "\n(_TABLE_SUFFIX BETWEEN 'yyyyMMdd' AND 'yyyyMMdd')" +
+      "\nWhere yyyy is the full year, MM is the month, and dd is the day.\n when constructing the query " +
+      "to make sure it's fast and takes use of the sharding. If you use this code, be sure to replace the dates with the correct range.\n";
```
jdorn: takes use of -> makes use of

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+    prompt: "Return SQL only, no markdown, no explanation, no comments.",
+    type: "generate-sql-query",
+    isDefaultPrompt: true,
+    temperature: 0.1,
+  });
+
+  // sometimes, even though we ask it not to, it returns in markdown:
+  const cleanSQL = aiResults.startsWith("```")
+    ? aiResults.replace(/```sql|```/g, "").trim()
+    : aiResults;
```
jdorn: Would it be better to use JSON Schema to force the AI to output the format we want?  For example, making it output `{ sql: string }`.  I'm ok with how it is here too, just might not be as robust.

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+
+    if (filteredTablesInfo.length === 0) {
+      return res.status(404).json({
+        status: 404,
+        message: "No relevant tables found based on AI response",
+      });
+    }
+
+    // for the filteredTables, lets loop through them and fetch their schemas, from getInformationSchemaTableById() or fetchOrCreateTableSchema()
+    for (const table of filteredTablesInfo) {
```
jdorn: This for loop is identical to the one below

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -188,3 +208,330 @@ export async function deleteSavedQuery(
...
+        `${table?.databaseName}.${table?.schemaName}.${table?.tableName}`
+      )
+    );
+
+    if (filteredTablesInfo.length === 0) {
+      return res.status(404).json({
+        status: 404,
+        message: "No relevant tables found based on AI response",
+      });
+    }
```
jdorn, line 430: Should we check here to make sure it listened to the `max 20` part of the prompt?

packages/back-end/src/enterprise/services/openai.ts
```diff
@@ -0,0 +1,207 @@
...
+ * The MODEL_TOKEN_LIMIT is the maximum number of tokens that can be sent to
+ * the OpenAI API in a single request. This limit is imposed by OpenAI.
+ *
+ */
+const MODEL_TOKEN_LIMIT = 128000;
+// Require a minimum of 30 tokens for responses.
+const MESSAGE_TOKEN_LIMIT = MODEL_TOKEN_LIMIT - 30;
+
+let _openai: OpenAI | null = null;
+let _openAIModel: TiktokenModel = "gpt-4o-mini";
```
jdorn: This is a per-org setting and cannot be defined at the top-level like this.

packages/back-end/src/enterprise/services/openai.ts
```diff
@@ -0,0 +1,207 @@
...
+      organization: context.org.id,
+      type,
+      model: _openAIModel,
+      numPromptTokensUsed: response.usage?.prompt_tokens ?? 0,
+      numCompletionTokensUsed: response.usage?.completion_tokens ?? 0,
+      temperature,
+      usedDefaultPrompt: isDefaultPrompt,
+    });
+  } catch (e) {
+    logger.error(e, "Failed to log AI usage to Clickhouse");
```
jdorn: Because you are not awaiting the above, this will never catch here.  Best bet is to make `logCloudAIUsage` never throw and always just fail silently.  Then you can remove this try/catch completely.

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+        status: 404,
+        message: "Metric not found",
+      });
+    }
+    if (!req.organization) {
+      return res.status(404).json({
+        status: 404,
+        message: "Organization not found",
+      });
+    }
```
jdorn: Nit: This check is not needed

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+    const datasource = await getDataSourceById(context, factTableDatasource);
+    const factMetricName = factMetric.name;
+    const projectNames = factMetric?.projects
+      ? await Promise.all(
+          factMetric.projects.map(async (p) => {
+            const project = await context.models.projects.getById(p);
+            return project?.name || "";
+          })
+        )
+      : [];
```
jdorn, line 777: Don't use `Promise.all` for mongo queries.  There is a `context.getProjects` method that returns all projects for the org in a single query (with a cache layer).  Do that once and then map id to project in a loop.

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+      "Timeframes are common to all metrics, and not important to describe (eg: this SQL is common for all: AND timestamp BETWEEN '{{startDate}}' AND '{{endDate}}')",
+      "Fact metrics of type 'proportion' fact metrics are used to measure the percent of experiment users who exist in a fact table",
+      "Fact metrics of type 'retention' fact metrics are used to measure the percent of experiment users who exist in a fact table a certain period after experiment exposure",
+      "Fact metrics of type 'mean' fact metrics are used to measure the average of a numeric value among all experiment users",
+      "Fact metrics of type 'ratio' fact metrics are used to measure the ratio of two numeric values among experiment users",
+      "Fact metrics of type 'quantile' fact metrics are used to measure the quantile of values after aggregating per user",
```
jdorn: For all of these types, you say `Fact metrics of type '...' fact metrics`.  Remove the second `fact metrics`.

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+      : [];
+
+    const priorKnowledge = [
+      "You are a data analyst who provides concise and accurate descriptions of metrics used in an A/B tests for less technical users to understand what this metric does",
+      "Do not start with the metric name as a title. The metric is defined with the following properties: \n" +
+        "- the metric name is: " +
+        metric.name +
+        "\n- it is based on the SQL:" +
+        metric.sql +
+        "\n" +
```
jdorn, line 906: The `metric.aggregation` setting is very important.  It should default to `SUM` if empty.  This tells us how to aggregate multiple rows to get a single value per user.

jdorn: These fields are also super important to include:
- `metric.type`
- `metric.denominator` (this is the id of another metric, so may want to do a lookup if defined)
- `metric.windowSettings`
- `metric.cappingSettings`
- `metric.tags` ?

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+        JSON.stringify(factMetricDenominator) +
+        "\n" +
+        "- the fact table has the following filters: " +
+        JSON.stringify(factTable?.filters ?? []) +
+        "\n" +
+        "- the fact table has the following columns defined: " +
+        JSON.stringify(factTableColumns) +
+        "\n" +
+        "- the fact metric is of type:" +
+        factMetric.metricType,
```
jdorn: Also include `factMetric.windowSettings`, `factMetric.cappingSettings`, and `factMetric.quantileSettings`.  Also, should we include tags?

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+  res: Response<{
+    status: number;
+    message?: string;
+    retryAfter?: number;
+    similar?: {
+      metric: MetricInterface;
+      similarity: number;
+    }[];
+  }>
+) {
```
jdorn, line 1012: This only support legacy metrics and not fact metrics.  Is that intentional?

packages/back-end/src/controllers/metrics.ts
```diff
@@ -690,3 +706,406 @@ export const getMetricNorthstarData = async (
...
+  // Sort and filter
+  const SIMILARITY_THRESHOLD = full ? 0 : 0.55;
+  const similarMetrics = similarities
+    .filter((s) => s && s.similarity && s.similarity >= SIMILARITY_THRESHOLD)
+    .sort((a, b) => (b?.similarity ?? 0) - (a?.similarity ?? 0))
+    .slice(0, full ? 100 : 5); // Get top 5 similar experiments
+
+  // loop through similarMetrics and get the full experiment object
+  const similarMetricIds = similarMetrics.map((s) => s.id);
+  const similarMetricObjects = await getMetricsByIds(context, similarMetricIds);
```
jdorn: You already have `allMetrics`.  No need to re-fetch these from Mongo.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -165,6 +173,426 @@ export async function getExperiments(
...
+      status: 404,
+      message: "AI configuration not set or enabled",
+    });
+  }
+  if (!req.organization) {
+    return res.status(404).json({
+      status: 404,
+      message: "Organization not found",
+    });
+  }
```
jdorn: Nit: not needed

packages/back-end/src/controllers/experiments.ts
```diff
@@ -165,6 +173,426 @@ export async function getExperiments(
...
+
+  const phase = experiment.phases.length - 1;
+  const snapshot =
+    (await getLatestSnapshot({
+      experiment: experiment.id,
+      phase,
+      type: "standard",
+    })) || undefined;
+
+  const winnerVariationName = experiment.variations[winner].name;
```
jdorn: This will throw if winner is an invalid index.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -165,6 +173,426 @@ export async function getExperiments(
...
+  const releasedVariationName =
+    experiment.variations.find((v) => v.id === releasedVariationId)?.name || "";
+
+  const experimentMetricIds = experiment.goalMetrics.concat(
+    experiment.guardrailMetrics,
+    experiment.secondaryMetrics
+  );
```
jdorn, line 239: This does not handle metric groups and also we should not do this concat manually here.

```ts
  const allMetricGroups = await context.models.metricGroups.getAll();
  const experimentMetricIds = expandMetricGroups(
    getAllMetricIdsFromExperiment(experiment, false),
    allMetricGroups
  )
```

packages/back-end/src/controllers/experiments.ts
```diff
@@ -165,6 +173,426 @@ export async function getExperiments(
...
+  // Sort and filter
+  const SIMILARITY_THRESHOLD = full ? 0 : 0.6;
+  const similarExperiments = similarities
+    .filter((s) => s && s.similarity && s.similarity >= SIMILARITY_THRESHOLD)
+    .sort((a, b) => (b?.similarity ?? 0) - (a?.similarity ?? 0))
+    .slice(0, full ? 100 : 5); // Get top 5 similar experiments
+
+  // loop through similarExperiments and get the full experiment object
+  const similarExperimentIds = similarExperiments.map((s) => s.id);
+  const similarExperimentObjects = await getExperimentsByIds(
```
jdorn: You already have `previousExperiments` which should contain these. No need to fetch from Mongo again.

packages/back-end/src/controllers/metrics.ts
Auz: yes, intentional at the moment - this function is not currently used in the front end.

PR #4094, opened 2025-05-20
Project out all log except where needed from featureRevisions

tzjames:
### Features and Changes
FeatureRevisions can be large - some are 10MB or more.  When fetching multiple revisions this can create a waterfall of `getMore` requests as it only fetches a certain amount at a time hence taking 5s to fetch 25 10MB revisions.  Which then adds a significant memory footprint on the server.

The logs for revisions account from 50-90% of the size of the feature revisions themselves.  It is only needed for the getRevisionLogs endpoint.  It is not part of the API schema.  This change removes it everywhere except for the `getRevisionLogs` endpoint.  Hence this should help on the speed of those requests and the memory footprint.

### Testing

On the feature overview page, fetched the logs and still saw them.
Saw on the featureAPI schema (`packages/back-end/src/api/openapi/schemas/FeatureWithRevisions.yaml`)  that logs are not returned.  
Did a search for `\Wlog[^\(\w\.:\\]` on the front-end and back-end and saw that `RevisionLog` and `FeatureRevisionModel` are the only place where it is referencing the revision logs.


jdorn (approved):
2 more places we should use projection to reduce the fields returned:
1. In `hasDraft`, we only care whether or not a doc exists, so it's definitely a waste to return all 10MB from the db.  Can probably just select a single field.
2. In `createRevision`, we do a `find` to get the latest version id.  We should only return the `version` field here since that's all that is needed.

tzjames:
@jdorn Whoops.  Thanks for finding that.  I should have searched for `.find`

PR #4097, opened 2025-05-21
Clickhouse Materialized Columns

Kevin-Chant:
### Features and Changes

Adds a form to the edit datasource page for creating and managing materialized columns in the inbuilt datasource.
This requires changing the format for our materialized views from implicit to explicit (with a `TO` statement) so the clickhouse setup helpers have been updated as well.

### Testing

Testing steps:

0. Setup

Log into Clickhouse and run the following commands in the sql terminal to set up your dev db
```
CREATE DATABASE <your_name>_dev;
CREATE TABLE <your_name>_dev.enriched_events AS default.enriched_events;
CREATE USER <your_name> IDENTIFIED WITH sha256_password BY <password> DEFAULT DATABASE <your_name>_dev;
GRANT CURRENT GRANTS ON <your_name>_dev.* TO <your_name> WITH GRANT OPTION;
```

Get the connection info from the left sidebar and set the environment variables

```
CLICKHOUSE_HOST=<full_url_with_port>
CLICKHOUSE_ADMIN_USER=<your_name>
CLICKHOUSE_ADMIN_PASSWORD=<password>
CLICKHOUSE_DATABASE=<your_name>_dev
CLICKHOUSE_MAIN_TABLE=enriched_events
```

<details>
<summary>Apply this diff</summary>

```diff
diff --git a/packages/back-end/src/app.ts b/packages/back-end/src/app.ts
index d4f29f7da..e41e76c7a 100644
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -738,12 +738,10 @@ app.post(
   datasourcesController.fetchBigQueryDatasets
 );
 
-if (IS_CLOUD) {
-  app.post(
-    "/datasource/create-inbuilt",
-    datasourcesController.postInbuiltDataSource
-  );
-}
+app.post(
+  "/datasource/create-inbuilt",
+  datasourcesController.postInbuiltDataSource
+);
 
 // Information Schemas
 app.get(
diff --git a/packages/front-end/pages/admin.tsx b/packages/front-end/pages/admin.tsx
index 457b6f9ce..0b6ffc257 100644
--- a/packages/front-end/pages/admin.tsx
+++ b/packages/front-end/pages/admin.tsx
@@ -212,7 +212,7 @@ function OrganizationRow({
             <FaPencilAlt />
           </a>
         </td>
-        {isCloud() && (
+        {
           <td className="p-0 text-center">
             <Tooltip
               body={
@@ -242,7 +242,7 @@ function OrganizationRow({
               </a>
             </Tooltip>
           </td>
-        )}
+        }
         <td style={{ width: 40 }} className="p-0 text-center">
           <a
             href="#"
```
</details>

1. Create the datasource via http://localhost:3000/admin and copy the ID

2. Insert a row into your main events table and verify it appears in your test db's `events` table.
```
INSERT INTO <your_name>_dev.enriched_events SELECT '2025-05-03 20:26:50' as timestamp, <org_id> as organization, <datasource_id> as datasource, client_key, environment, event_name, properties_json, user_id, properties_json as context_json, url, url_path, url_host, url_query, url_fragment, device_id, page_id, session_id, sdk_language, sdk_version, page_title, utm_source, utm_medium, utm_campaign, utm_term, utm_content, event_uuid, ip, geo_country, geo_city, geo_lat, geo_lon, ua, ua_browser, ua_os, ua_device_type, date_created FROM default.enriched_events WHERE event_name = 'page-load' LIMIT 1;
```

3. Auto-generate fact tables (http://localhost:3000/fact-tables) and click into the page views table. Modify the WHERE clause to use `page-load` for the event name and confirm to force it to refresh. In the columns, find `context_json` and change the type to JSON and save, then force another refresh. If you click to change `context_json` again it should have populated the subfields.

4. Go back to your datasource page, and scroll down to add materialized columns. Create multiple with different types e.g. `configFile` and `page_url`. Fully refresh the clickhouse console between each addition to see the new column being added

5. Insert an extra row into the source table to see the materialized view transform it and send it to the final events table
```
  INSERT INTO <your_name>_dev.enriched_events SELECT * FROM <your_name>_dev.enriched_events WHERE event_name = 'page-load' LIMIT 1;
```

6. Delete one or more materialized columns and check that they were dropped in clickhouse
### Screenshots

image

image

image

image


packages/back-end/src/routers/fact-table/fact-table.validators.ts
```diff
@@ -1,4 +1,16 @@
...
+export const factTableColumnTypes = [
+  "number",
+  "string",
+  "date",
+  "boolean",
+  "json",
+  "other",
+  "",
+];
+// Duplicate of the above as we can't use `as const` without breaking imports in the frontend
```
jdorn, line 13: This doesn't sound right.  Maybe something else is going on that's stopping `as const` from working?

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -336,6 +343,71 @@ export async function updateDataSource(
...
+        )
+    );
+    await updateMaterializedColumns({
+      datasource,
+      columnsToAdd,
+      columnsToDelete,
+      columnsToRename,
+      finalColumns,
+      originalColumns,
+    });
```
jdorn, line 408: It looks like this will always run on every update, even if nothing changes?

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -336,6 +343,71 @@ export async function updateDataSource(
...
   }
 
+  if (
+    isDataSourceType<GrowthbookClickhouseDataSource>(
+      datasource,
+      "growthbook_clickhouse"
+    ) &&
+    isPartialWithMaterializedColumns(updates)
+  ) {
+    updates.settings!.materializedColumns = sanitizeMaterializedColumns(
```
jdorn, line 353: Instead of `updates.settings!` everywhere, can we just put that into the `if` condition?  `if (... && updates.settings && updates.settings.materializedColumns)`.

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -336,6 +343,71 @@ export async function updateDataSource(
...
+    );
+    const originalColumnMap = Object.fromEntries(
+      originalColumns.map((col) => [col.sourceField, col])
+    );
+    const columnsToDelete: MaterializedColumn[] = [],
+      columnsToRename: { from: string; to: string }[] = [];
+
+    originalColumns.forEach((col) => {
+      if (
+        !Object.prototype.hasOwnProperty.call(newColumnMap, col.sourceField)
```
jdorn, line 372: This syntax is a little awkward and unnecessarily complex.  Either just use the `key in obj` syntax or make it an actual Map object so you get the `.has` method.

packages/back-end/src/models/DataSourceModel.ts
jdorn: Actually, looking at this more, I think we should make dedicated CRUD endpoints and methods instead of using the generic update route for everything.  
- Code will be much cleaner and easier to follow
- Avoids race conditions (2 people changing different parts of the materializedColumns array at the same time)
- Avoids unintentional changes (we had something similar recently in environments caused by passing the whole array to the back-end. A bug meant re-ordering the array would sometimes delete things by accident)
- It will map much better to the front-end actions (click the "delete" button, it calls a dedicated "delete" endpoint)

packages/back-end/src/routers/fact-table/fact-table.validators.ts
Kevin-Chant: Here's the actual error message
```
| Module parse failed: Unexpected token (12:2)
 | You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
 | |   "other",
 | |   "",
 | > ] as const;
```

packages/front-end/components/Settings/EditDataSource/DataSourceMetrics.tsx
Kevin-Chant, line 1: The diff looks large but this is just shuffling elements to fix the button text wrapping

Kevin-Chant:
@jdorn I think I've addressed all of your comments other than the `as const` issue. That one I'm not actually sure how to debug myself so I'd appreciate some insight if you've seen this before

packages/back-end/src/controllers/datasources.ts
```diff
@@ -860,3 +873,298 @@ export async function fetchBigQueryDatasets(
...
+  const columnName = sanitizeMatColumnName(userInput.columnName);
+  return {
+    datatype: userInput.datatype,
+    sourceField,
+    columnName,
+  };
+}
+
+function sanitizeMatColumnSourceField(userInput: string) {
+  if (!/^[a-zA-Z0-9 _-]*$/.test(userInput)) {
```
jdorn, line 1178: Make sure there's at least 1 alphanumeric character.  Don't want to match a string full of spaces / punctuation.  Probably don't want to allow the string to start with spaces at all since that's kind of weird and also a likely typo for someone to make in the UI.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -860,3 +873,298 @@ export async function fetchBigQueryDatasets(
...
+  }
+  return userInput;
+}
+
+function sanitizeMatColumnName(userInput: string) {
+  if (!/^[a-zA-Z_][a-zA-Z0-9_]*$/.test(userInput)) {
+    throw new Error(
+      "Invalid input. Column names must start with a letter or underscore and only use alphanumeric characters or '_'"
+    );
+  }
```
jdorn, line 1204: We need a block list of reserved column names so we don't accidentally overwrite one of the default columns.

We also might want to block known SQL keywords (`count`, `select`, `from`, `where`, etc.) since it can cause some weirdness when querying.

packages/back-end/src/services/clickhouse.ts
```diff
@@ -288,3 +349,93 @@ WITH FILL
...
+
+  logger.info(`Updating materialized columns; dropping view ${eventsViewName}`);
+  await client.command({ query: `DROP VIEW IF EXISTS ${eventsViewName}` });
+
+  let err = undefined;
+  try {
+    const addClauses = columnsToAdd
+      .map(
+        ({ columnName, datatype }) =>
+          `ADD COLUMN IF NOT EXISTS ${columnName} ${getClickhouseDatatype(
```
jdorn, line 407: Should we escape the column names in all of these queries?  Just thinking about if someone uses a SQL reserved keyword.  Because we don't allow spaces or punctuation, the damage someone could do is pretty limited, so maybe not a big deal.

PR #4100, opened 2025-05-21
Introduce auto-complete logic to EditSqlModal

mknowlton89:
### Features and Changes

This PR introduces auto-complete logic for the EditSqlModal - specifically, it will pull in databases, schemas, and tables when in the FROM clause of a query, and it will pull in column data (including the column's datatype) when typing in the SELECT, WHERE, GROUP BY, or ORDER BY clause, we'll look at the all of the tables mentioned in the query and pull columns/datatypes for each of those tables.

Additionally, this PR includes a new `sqlKeywords.ts` file that contains sql keywords that we can use to power auto completions that also rely on knowing the current context (e.g. is the user's cursor in the `SELECT` clause, the `FROM` clause, etc, and we'll filter the suggestions based on that. 

This has been tested with a BigQuery data source, as it's has some of the more advanced SQL syntax (e.g FROM clauses needing to be wrapped in backtics), so it'd be great to test this with other data source types.

At a high level, this PR is doing a few things:

1) It's leveraging React-Ace's built-in autocomplete functionality, where we pass the CodeTextArea and array of autocomplete options
2) We're fetching the information schema for a particular data source that will ultimately power the auto completions. This is done after a 300ms debounce to reduce the amount we're fetching.
3) With our informationSchemas, those only go to the table level (not the column level) eg. we store information about databases, schemas, and tables. From there, we have a single record in a separate Mongo collection for each table, that contains the columns and their associated datatype.
4) When a user is typing, we have a `getCurrentContext` function that takes in the user's cursor data (aka, where their cursor is in the query) and use that to determine what part of the query the user is editing (SELECT, FROM, WHERE, GROUP BY, ORDER BY) - we use this context to determine what type of autocomplete suggestions to pass to the CodeTextArea
5) If the user is in the SELECT part of the statement, we parse the query to identify which tables are mentioned in the query, and with those, we fetch the `informationSchemaTables` collection to get columns/datatypes for those tables. We do have a short-lived cache that stores these, so we don't have to hit mongo multiple times for the same table
6) And lastly, we have some rather complicated logic around what value we pass in for each autocomplete index - for some data sources, we need to wrap FROM clauses in backticks, but to make matters more complicated, sometimes we do and sometimes we don't need to add backticks - e.g. if the FROM clause is blank, and the user selects a table name directly from the autocomplete, we show the full path, including backticks (if the data source type needs it). Whereas is the FROM clause already has a database and schema included (e.g. FROM `database.schema) - we don't need the "value" of the autocomplete for each table to be prefixed with a backtick. Aka, we don't want the result to be `database.schema.`table`. Instead, we have to take into consideration the existing FROM clause, and apply backticks not just based on the data source type, but based on the existing clause.

### Fast Follows
- [ ] Update the `SQL_KEYWORD_DEFINITIONS` to filter based on dialect.

### Dependencies
None

### Testing

- [x] Ensure that getCurrentContext is accurate - e.g. when in the SELECT portion of the query, ensure that it returns `type: "SELECT"`, and ensure this is correct for all `Keyword` types. (SELECT, FROM, WHERE, GROUP BY, ORDER BY)
- [x] Ensure that if the getCurrentContext.type is one of SELECT, WHERE, GROUP BY, or ORDER BY, and there is a table, we fetch the columns/datatype for the table and the columns are available as autocomplete options
- [x] Ensure that when it a FROM clause, the formatting is correct based on the existing query and the data source type.


packages/front-end/test/services/sqlAutoComplete.test.ts
```diff
@@ -0,0 +1,775 @@
...
+  mockCalls = [];
+};
+
+describe("getCurrentContext", () => {
+  it("should detect SELECT context", () => {
+    const cursorData: CursorData = { input: ["SELECT "], row: 0, column: 7 };
+    const result = getCurrentContext(cursorData);
+    expect(result).toEqual({
+      type: "SELECT",
+      suggestions: [],
```
ahdriel, line 273: suggestion: Should we validate that the `suggestions` gets populated appropriately?

packages/front-end/test/services/sqlAutoComplete.test.ts
```diff
@@ -0,0 +1,775 @@
...
+  InformationSchemaTablesInterface,
+} from "back-end/src/types/Integration";
+import {
+  getCurrentContext,
+  getSelectedTables,
+  getAutoCompletions,
+} from "@/services/sqlAutoComplete";
+import { CursorData } from "@/components/Segments/SegmentForm";
+
+/** This test suite tests the sqlAutoComplete logic - throughout the file you'll see different sql formats, especially with backticks
```
ahdriel, line 44: Awesome tests! 🎉

packages/front-end/styles/global.scss
```diff
@@ -3578,6 +3578,10 @@ input:disabled {
 .ace_editor .ace_gutter {
   background: var(--color-panel-solid) !important;
 }
+.ace_editor.ace_autocomplete { 
+  width: 500px !important;
+  background-color: var(--color-panel-solid)!important;
```
ahdriel: nit: missing space

```suggestion
  background-color: var(--color-panel-solid) !important;
```

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -143,6 +150,58 @@ export default function CodeTextArea({
     );
   }, [editor, onCtrlEnter]);
 
+  useEffect(() => {
+    if (typeof window !== "undefined" && editor) {
+      import("ace-builds").then((ace) => {
```
ahdriel: question: Do we need this import? We should have it available in the `AceEditor`.

I am concerned this might increased our bundle size.

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -143,6 +150,58 @@ export default function CodeTextArea({
...
+        const langTools = ace.require("ace/ext/language_tools");
```
ahdriel: suggestion: Should we add this to the dynamic import above to ensure we have it all in a single bundle?

packages/front-end/services/sqlAutoComplete.ts
```diff
@@ -0,0 +1,647 @@
...
+ * @returns Updated array of selected table IDs
+ */
+function addEventTableName(
+  selectedTables: string[],
+  eventName: string,
+  cursorData: CursorData,
+  informationSchema: InformationSchemaInterface
+): string[] {
+  const sql = cursorData.input.join("\n");
+  if (!sql.includes("eventName")) return selectedTables;
```
ahdriel, line 123: question: Should we have this in a const somewhere? And can you clarify why eventName is important here?

packages/front-end/services/sqlKeywords.ts
```diff
@@ -0,0 +1,589 @@
...
+  },
+  {
+    value: "NULL",
+    meta: "KEYWORD",
+    score: COMPLETION_SCORES.LESS_COMMON_KEYWORD,
+    caption: "NULL",
+  },
+];
+
+export function getSqlKeywords(): AceCompletion[] {
```
ahdriel, line 582: question: Why do we needs this as a function? Can we have it as hard-coded and reference `SQL_KEYWORD_DEFINITIONS` directly on its usage instead?

packages/front-end/components/SchemaBrowser/EditSqlModal.tsx
```diff
@@ -157,6 +175,58 @@ export default function EditSqlModal({
...
+      if (!isAutocompleteEnabled) {
+        setInformationSchema(undefined);
+        return;
+      }
+      try {
+        const response = (await apiCall(
+          `/datasource/${datasourceId}/schema`
+        )) as {
+          informationSchema: InformationSchemaInterface;
+        };
```
ahdriel: nit: We can pass the return type as the generic parameter

```suggestion
        const response = await apiCall<{
          informationSchema: InformationSchemaInterface;
        }>(
          `/datasource/${datasourceId}/schema`
        );
```

mknowlton89:
> Awesome work! This is very powerful and also seems we can extend on it a lot.
> 
> Some questions and additional feedback:
> 
> * I noticed this is not available in the SQL Explorer, should we make it available there?
> * Can we remove the `def.` prefix from the autocomplete options?
>   Screenshot 2025-07-22 at 4 33 48 PM
> * And when I selected that option the table name got included without the def (nice!) but it did not add the backticks that are required in my case (`dummy-data`.`sales`). It is a MySQL data source.
>   Screenshot 2025-07-22 at 4 33 54 PM
> * The spacing between the Test Query & 3 dots looks off
>   Screenshot 2025-07-22 at 4 37 16 PM
> * Should variables always be higher than columns?
>   Screenshot 2025-07-22 at 4 34 17 PM

Thanks for the thorough review!

I'll answer a few questions here and the others inline.

1.  I noticed this is not available in the SQL Explorer, should we make it available there? - For sure, there is a bit of work we'll need to do to the Sql Explorer so it can support template variables that I plan to do after this PR is completed.
2. Can we remove the def. prefix from the autocomplete options? - I'll need to think through that - that's part of the `path` for your data source, I think it should be ok to remove for a MySQL data source, but let me test to be sure.
3. And when I selected that option the table name got included without the def (nice!) but it did not add the backticks that are required in my case (`dummy-data`.`sales`). It is a MySQL data source. - I think this might actually be resolved with my other PR, I'll pull in main once that is merged and re-test this, and if that doesn't fix it, I'll address.
4. The spacing between the Test Query & 3 dots looks off - Good catch
5. Should variables always be higher than columns? Probably not. One thing that would be cool to introduce is some logic similar to how fuzzy search works where the score can change based on the search context, but I think that might be a future improvment.

packages/front-end/services/sqlAutoComplete.ts
mknowlton89: @ahdriel To clarify, are you suggesting that we define a const like `SQL_INCLUDES_EVENTNAME` as a top-level const?

As for why `eventName` is important - when creating legacy metrics, we let the user specify an `eventName` via the create metric modal, and then when we set up the metric's initial query, we use the `{{ eventName }}` template variable.

And in order for the sql auto complete to suggest columns, we need to parse the query to identify what tables are being referenced in it, and in cases where the base sql uses the `{{ eventName }}` template variable, we wouldn't be able to do that without this logic

E.G. A user might be trying to create a metric for the number of items `added_to_cart` , so their event name could be `added_to_cart`.

We would then know in the `EditSqlModal` that we need to pull column suggestions for the `added_to_cart` table. 

Screenshot 2025-07-23 at 7 06 20 AM

packages/front-end/services/sqlKeywords.ts
mknowlton89: Good question - currently, we don't need it. But one future improvement I'd like to make is adding a `dialect` to each `SqlKeyWordDefinition` so we can filter the sql keywords based on the data source dialect.

I built this originally intending to include that here, but then axed it to reduce scope and try and get the base logic out, and this is a side effect of that.

packages/front-end/components/Forms/CodeTextArea.tsx
mknowlton89: Great suggestion - I was able to make it work without the import

packages/front-end/components/Forms/CodeTextArea.tsx
mknowlton89: Good call - Claude and I made a few big changes to how we handle imports in this [commit](https://github.com/growthbook/growthbook/pull/4100/commits/a64db23a97c68542ccf5cbbc75ddfbec58b4c66b).

Let me know what you think.

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -1,10 +1,41 @@
...
 import Field, { FieldProps } from "./Field";
 
+export type AceCompletion = {
+  caption: string;
+  value: string;
+  meta: string;
+  score: number;
+};
+
+interface AceEditorProps {
```
ahdriel: suggestion: If we are forwarding the props to Ace, should we use the exported `IAceEditorProps` instead?

And we can extend it with any custom props we might have like
```
type EditorProps = IAceEditorProps & {
    completions?: AceCompletion[]
}
```

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -76,10 +107,89 @@ const AceEditor = dynamic(
...
     ace.config.setModuleUrl("ace/mode/yaml_worker", yamlWorkerUrl.default);
 
-    return reactAce;
+    const langTools = ace.require("ace/ext/language_tools");
+
+    // Return a wrapper component that handles completions
+    const AceEditorWithCompletions = React.forwardRef<
+      HTMLElement,
+      AceEditorProps
+    >((props: AceEditorProps) => {
```
ahdriel: issue: We need to add the `ref` here and pass it along to the editor.

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -76,10 +107,89 @@ const AceEditor = dynamic(
...
+      const handleLoad = (editorInstance: Ace.Editor) => {
+        setEditor(editorInstance);
+        // Call the original onLoad if provided
+        if (onLoad) {
+          onLoad(editorInstance);
+        }
+      };
+
+      // Update completions whenever they change
+      React.useEffect(() => {
```
ahdriel: nit: We can use `useEffect` directly without changing the import to include `React`

packages/front-end/services/sqlKeywords.ts
ahdriel: I would vote to use the hard-coded array directly for now and avoid the mapping into the same structure.

packages/front-end/services/sqlAutoComplete.ts
```diff
@@ -160,21 +166,24 @@ function handleColumnCompletions(
...
-  }));
+  const databaseCompletions = informationSchema.databases
+    .map((db) => ({
+      value: db.path,
+      meta: COMPLETION_TYPES.DATABASE,
+      score: COMPLETION_SCORES.DATABASE,
+      caption: db.databaseName,
+    }))
+    // Filter out completions whose value is empty - This can happen for MySQL Data Sources that don't support all 3 levels (db, schema, table)
+    .filter((completion) => completion.value !== "");
```
ahdriel: question: Should the editor ignore (or filter out) the empty values? Because if the user selects it nothing will happen right which would always be a confusing experience.

packages/front-end/components/Forms/CodeTextArea.tsx
mknowlton89: Thanks for catching this - I don't think we actually need the forwardRef here. 

I've removed it and everything seems to be working as expected. I think the `onLoad` functionality still provides access to the editor.

It's possible I'm missing something - I've not used the `forwardRef` component a lot (and it's deprecated).

packages/front-end/services/sqlAutoComplete.ts
mknowlton89: @ahdriel Let me know if I'm misunderstanding your question, but this filter logic is removing any completions where the value is an empty string before we send the suggestions to the editor.

Forgive me if you already know this, but I'll add a comment below just to document it.

When we're building the list of database suggestions (the highest level of the informationSchema hierarchy) if the data source type doesn't support all 3 levels like MySQL, the `path` for all databases is an empty string.

So with the logic about, we're building that, and then filtering out any suggestions that don't have a path.

In looking at this further, I think we can eliminate the need to loop through the db array twice with the following logic:

```ts
const databaseCompletions: AceCompletion[] = [];
  for (const db of informationSchema.databases) {
    // Not all data sources support all 3 levels (db, schema, table) - and if they don't, the path will be empty
    // So we need to check if the path is empty before adding it to the completions
    if (db.path) {
      databaseCompletions.push({
        value: db.path,
        meta: COMPLETION_TYPES.DATABASE,
        score: COMPLETION_SCORES.DATABASE,
        caption: db.databaseName,
      });
    }
  }
```

This allows us to avoid adding it to the completions array in the first place. Most data sources don't have a ton of databases, so probably not a crazy performance improvement, but a better pattern none the less.

Let me know if that answers your question or if I misunderstood.

packages/front-end/components/Forms/CodeTextArea.tsx
ahdriel: I think it only matters if we need to get the ref for the editor. And if we don't then it does not change anything.

Agree with the decision of removing it for now to be easier to migrate to React 19.

packages/front-end/services/sqlAutoComplete.ts
ahdriel: Thanks for clarifying the logic & I agree with the implementation.

My question was for an additional check in the editor itself, for scenarios where we might have a bug and pass in completions with value being an empty string. Similar to the fix here, I don't see a valid scenario where we want a completion to have `''` for the value, so we could have the same filter in the Editor (and potentially log with a warning that a certain completion was discarded because the value was empty).

Does that make sense?

ahdriel (approved):
🎉 🎉

packages/front-end/services/sqlAutoComplete.ts
mknowlton89: 100% - I'll make that adjustment

PR #4106, opened 2025-05-22
[fix] fixed an issue with auto rollback not getting set to undefined

gazzdingo:
### Features and Changes
added a sanity check for saferollout to always set false when auto rollback is undefined
(https://growthbook.sentry.io/issues/6623345951/)


gazzdingo:
We are holding off on this to make sure that it wasnt a stale webpage

msamper:
Closed because there doesn't seem to be a real issue here. See explanation here: https://growthbookapp.slack.com/archives/C08HK09B481/p1748451932376379

PR #4107, opened 2025-05-22
Insights section

Auz:


packages/front-end/components/Layout/Layout.tsx
```diff
@@ -59,7 +55,7 @@ const navlinks: SidebarLinkProps[] = [
   {
     name: "Metrics and Data",
     href: "/metrics",
-    path: /^(metric|segment|dimension|datasources|fact-|metric-group)/,
+    path: /^(metric$|metrics|segment|dimension|datasources|fact-|metric-group)/,
```
jdorn: Metric detail pages are `/metric/{id}` and this change will stop it matching that

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -91,6 +87,50 @@ const navlinks: SidebarLinkProps[] = [
     ],
   },
   {
+    name: "Insights",
+    href: "/insights",
```
jdorn: Is there an `/insights` page? Should we change this to `/dashboard`?

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -91,6 +87,50 @@ const navlinks: SidebarLinkProps[] = [
...
+    Icon: GBLibrary,
+    path: /^(insights|library|^learning*|timeline|metric-effect|correlations)/,
```
jdorn: This regex needs some cleaning up.
- Missing `presentations` and `dashboard`
- Remove `insights` and `library` (doesn't look like they match anything)
- `^learning*` should just be `learning`

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -107,13 +147,8 @@ const navlinks: SidebarLinkProps[] = [
...
         path: /^idea/,
       },
-      {
-        name: "Presentations",
-        href: "/presentations",
-        path: /^presentation/,
-      },
     ],
-  },
+  },*/
```
jdorn: If the insights tab is behind a feature flag, we should uncomment this, add back presentations, and add:

```ts
filter: ({ gb }) => !gb?.isOn("insights")
```

packages/front-end/components/MetricExperiments/MetricExperiments.tsx
```diff
@@ -274,12 +283,25 @@ const MetricExperiments: FC<MetricAnalysisProps> = ({
...
+  numPerPage = NUM_PER_PAGE,
+  differenceType = "relative",
 }) => {
   const { data } = useApi<{
     data: ExperimentWithSnapshot[];
-  }>(`/metrics/${metric.id}/experiments`);
-  const metricExperiments = (data?.data ?? []).filter((e) =>
-    bandits ? e.type === "multi-armed-bandit" : e.type !== "multi-armed-bandit"
+  }>(`/metrics/${metric.id}/experiments`, {
+    shouldRun: dataWithSnapshot ? () => false : undefined,
```
jdorn, line 294: Nit: this is a slightly weird way to do this. Would be cleaner as:

```ts
shouldRun: () => !dataWithSnapshot
```

packages/front-end/components/Experiment/CompletedExperimentList.tsx
```diff
@@ -0,0 +1,345 @@
...
+      {openCarousel && (
+        <ExperimentCarouselModal
+          experiment={experiments[openCarousel.experimentIndex]}
+          currentVariation={openCarousel.variationId}
+          currentScreenshot={openCarousel.index}
+          imageCache={imageCache}
+          close={() => {
+            setOpenCarousel(null);
+          }}
+          restrictVariation={false}
```
ahdriel, line 59: question: Do we want to pass `true` here as it is a new behavior we are adding?

packages/front-end/enterprise/components/ExecReports/ExecExperimentImpact.tsx
```diff
@@ -40,7 +40,6 @@ function getColor(val: number | undefined): string {
 }
 
 export default function ExecExperimentImpact({
```
ahdriel, line 43: question: Do we need to update this component to handle metric groups?

packages/front-end/pages/learnings/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    searchParams.get("endDate")
+      ? new Date(searchParams.get("endDate")!)
+      : new Date(today.getTime() + 7 * 24 * 60 * 60 * 1000) // 7 days in the future
+  );
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search);
+    params.set("startDate", startDate.toISOString().slice(0, 10)); // Keep only YYYY-MM-DD
+    params.set("endDate", endDate.toISOString().slice(0, 10)); // Keep only YYYY-MM-DD
+    const newUrl = `${window.location.pathname}?${params.toString()}`;
+    window.history.pushState({}, "", newUrl);
```
ahdriel: issue (non-blocker): I believe these URL manipulations are breaking some of the back/forward buttons in the browser.

packages/front-end/components/Experiment/CompletedExperimentList.tsx
jdorn: That's what I was originally thinking, but the winning screenshot by itself isn't that useful unless you have something to compare it against.  So including all the screenshots in the popup modal makes sense.

packages/front-end/enterprise/components/ExecReports/ExecExperimentImpact.tsx
jdorn: Yes, we do.  Good catch.  For all the pages with metric dropdowns (learnings, effects, correlations), I added logic to restrict the options to only ones that are used in experiments, but I'm not accounting for metric groups.

I'm using `getAllMetricIdsFromExperiment`, which has a `metricGroups` option, but it's optional and I had no idea it was even there.

packages/front-end/pages/learnings/index.tsx
jdorn: I ran into this too on a few pages and it was pretty annoying.  I think it's an easy fix though.  It happens because `useEffect` runs on page load and we only really want this to run when someone changes one of the filters.

Instead of using `useEffect`, we can just call this from the `setDate` methods of the date pickers.

jdorn (approved):

PR #4110, opened 2025-05-23
Adds format button to SqlEditor

mknowlton89:
### Features and Changes

This PR introduces a new `Format` button to the EditSqlModal - when clicked, this will take in the underlying Datasource's type and format the query based on that type using the `sql-formatter` package that already exists in the application.

As this `sql-formatter` package already existed in our codebase, and in the backend we were formatting queries for users, I've opted to move the actual format logic to the `shared` directory so we can use the same exact formatting logic on the front-end as we do on the back-end to prevent any formatting issues.

I've based the UX based on existing patterns with Sql code editors, where once you format a query, we store the previous state and allow the user to `Clear Formatting`. The option to clear the formatting does go away if the user edits the query after formatting.

### Dependencies

None

### Testing

- [x] Ensure the formatting works as expected with a BigQuery query
- [x] Ensure the formatting works as expected with a Postgres query
- [x] Ensure if there is an error formatting the query, we render the error and icon next to the format button to give context to the user about why we couldn't format the query
- [x] Ensure that if there is an error, the error state is cleared the moment the user edits the query
- [x] Ensure the existing sql format tests continue to work without issue
- [x] Ensure that we're able to handle queries with template variables correctly
- [x] Ensure comments are handled correctly
- [ ] Ensure that the formatting works with other data source types - Clickhouse, Redshift, Databricks, etc


packages/back-end/src/util/sql.ts
```diff
@@ -12,6 +12,10 @@ Object.keys(helpers).forEach((helperName) => {
   Handlebars.registerHelper(helperName, helpers[helperName]);
 });
 
+// Re-export FormatDialect and format for backward compatibility
+export type { FormatDialect };
+export { format };
+
```
mknowlton89, line 18: This is a bit of a weird pattern - we were previously exporting these from here. This makes the diff much smaller, but happy to change if we don't want to introduce this pattern.

packages/front-end/hooks/useSqlFormatter.ts
```diff
@@ -0,0 +1,139 @@
...
+
+      const dialect = getSqlFormatterLanguage(datasourceType);
+      if (!dialect) return sql;
+
+      try {
+        // Format the SQL - using shared format function
+        const {
+          sql: sqlWithoutTemplates,
+          placeholders,
+        } = replaceTemplateVariables(sql);
```
mknowlton89, line 82: The formatter wasn't able to handle template variables, this updates the query to remove them before formatting, and then adds them back.

mknowlton89:
@lukebrawleysmith @lukesonnet - Added you as reviewers here - if you have time, I'd love to get you two using the `Format` option (video walkthrough above) with some of your data sources just to see if there are any gotchas I haven't found yet.

lukebrawleysmith:
I tested it locally and the formatting is helpful, thanks a bunch!  

Is it worth having the additional functionality of fixing incorrectly ordered statements?  e.g., if `group by` is before `where`?  
Also, perhaps the formatter can check for variable type in `where` statements?  

These are nits, as when the customer tests the query these errors will get thrown, and they can throw the query into any LLM for more help.  But if someone doesn't know much sql, it could save them time.

packages/front-end/services/sqlFormatter.ts
```diff
@@ -0,0 +1,83 @@
+import { DataSourceType } from "back-end/types/datasource";
+import { format } from "shared/sql";
+import { FormatDialect } from "shared/src/types";
+
+export function canFormatSql(datasourceType: DataSourceType): boolean {
+  return !!getSqlDialect(datasourceType);
+}
+
+function getSqlDialect(datasourceType: DataSourceType): FormatDialect | "" {
```
mknowlton89, line 9: Not all data source types have a dialect. So in those cases, we return an empty string so we don't try to format the query with the wrong language when we call `sql-formatter`'s `format` function.

packages/front-end/components/SchemaBrowser/EditSqlModal.tsx
```diff
@@ -217,24 +230,40 @@ export default function EditSqlModal({
...
+                        disabled={!canRunQueries}
+                        type="button"
+                      >
+                        <span className="pr-2">
+                          <FaPlay />
+                        </span>
+                        Test Query
+                      </Button>
+                    </Tooltip>
+                    {canFormat ? (
```
mknowlton89, line 252: Now hiding the button if we know we can't correctly format the query based on the datasource type.

ahdriel (approved):
Looking good! Thanks for the updated and the super helpful comments 🙇

PR #4114, opened 2025-05-27
Move feature revision logs to their own model

tzjames:
### Features and Changes
Feature Revision logs can be up to 10 times the size of the rest of the revision.   This creates the danger of revisions becoming larger than the maximum document size for Mongo.  So this moves future revision logs over to their own model.  Revision.log is filtered out from most queries and are only sent when specifically loading the logs.  Since this is a rare event, and it will be even rarer to do it for old versions we won't bother moving over old revision.logs to the new model.

### Testing
On main branch.
Load a feature.
Make changes.
View logs.
Change to this branch.
Make changes.
View logs.
See that the logs are the merged `revision.log` and `FeatureRevisionLogModels`.
Check Mongo to see that the `FeatureRevision` has the old logs and `FeatureRevisionLogModels` have the new logs.

### DIFF
--- a/packages/back-end/src/models/FeatureRevisionLogModel.ts
+++ b/packages/back-end/src/models/FeatureRevisionLogModel.ts
@@ -1,4 +1,7 @@
-import { featureRevisionLogValidator } from "back-end/src/validators/feature-revision-log";
+import {
+  FeatureRevisionLogInterface,
+  featureRevisionLogValidator,
+} from "back-end/src/validators/feature-revision-log";
 import { MakeModelClass } from "./BaseModel";
 
 export const COLLECTION_NAME = "featurerevisionlog";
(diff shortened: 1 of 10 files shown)


packages/back-end/src/validators/feature-revision-log.ts
```diff
@@ -0,0 +1,24 @@
...
+import { baseSchema } from "back-end/src/models/BaseModel";
+import { eventUser } from "./events";
+
+export const featureRevisionLog = z
+  .object({
+    featureId: z.string(),
+    version: z.number(),
+    user: eventUser,
+    approvedBy: eventUser.optional(),
+    timestamp: z.date(),
```
jdorn, line 11: There is already a `dateCreated` field in base model, so no need for `timestamp` here.

packages/back-end/src/validators/feature-revision-log.ts
```diff
@@ -0,0 +1,24 @@
+import { z } from "zod";
+import { baseSchema } from "back-end/src/models/BaseModel";
+import { eventUser } from "./events";
+
+export const featureRevisionLog = z
+  .object({
+    featureId: z.string(),
+    version: z.number(),
+    user: eventUser,
+    approvedBy: eventUser.optional(),
```
jdorn, line 10: It doesn't look like `approvedBy` is actually used anywhere.  I think we can remove this.

packages/back-end/src/models/FeatureRevisionLogModel.ts
```diff
@@ -0,0 +1,79 @@
...
+    const { feature } = this.getForeignRefs(doc);
+    if (!feature) {
+      throw new Error("Feature not found for FeatureRevisionLog");
+    }
+    return (
+      this.context.permissions.canCreateFeature(feature) ||
+      this.context.permissions.canManageFeatureDrafts(feature)
+    );
+  }
+  protected canUpdate(existing: FeatureRevisionLogInterface): boolean {
```
jdorn, line 48: Do we ever want to allow updating or deleting a revision log?  That's kind of the point of a log like this, it's an audit log of all changes that happened.

I could see an argument for deleting when the parent feature gets deleted, but I can't think of a reason to ever allow updating.

PR #4118, opened 2025-05-28
show when user can stop early for safe rollouts

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
+++ b/packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
@@ -496,6 +496,18 @@ export function getSafeRolloutDaysLeft({
 
   return daysLeft;
 }
+// checks to see if each guardrail metric has the status of "safe"
+function canSafeRolloutStopEarly(safeRollout: SafeRolloutInterface) {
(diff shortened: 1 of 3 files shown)


packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
```diff
@@ -602,11 +615,13 @@ export function getSafeRolloutResultStatus({
       daysLeft,
     };
   }
-
-  if (daysLeft <= 0 && resultsStatus) {
+  const showCanShip = canStopEarly && daysLeft <= 0;
```
msamper, line 618: Should this be `canStopEarly || daysLeft <= 0`?

msamper: seems like the value of `canStopEarly` shouldn't block returning the ship decision when `daysLeft` is 0

packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
```diff
@@ -602,11 +615,13 @@ export function getSafeRolloutResultStatus({
...
+  // check and see if the guardrail metrics are passing
+  if (showCanShip && resultsStatus) {
     // If no days left, return ship decision
```
msamper, line 621: update comment to reflect new conditions

PR #4124, opened 2025-05-29
Introduced managedBy field on the front end

mknowlton89:
### Features and Changes
This PR is introducing some new front-end logic when a resource has a `managedBy` property.

`Projects` & `Teams`
- We're rendering a Managed by `[VENDOR]` badge and blocking deletion if it has a managedBy.type value set.

`SDK Connections`
- We're rendering a Managed by `[VENDOR]` badge and blocking deletion if it has a managedBy.type value set.
- We are allowing a user to duplicate an SDK even if it's managed via an external source
- We do allow editing the SDK Connection, but lock disable the project field for certain vendors. As projects for an SDK connection are a subset of the SDK Connection's selected environment, we are filtering the environments to only include environments that are in `All Projects` or environments that include the SDK Connection's existing project in it's projects array.

`SDK Webhooks`
- We're rendering a Managed by `[VENDOR]` badge and blocking deletion if it has a managedBy.type value set.
- We're also blocking the editing of the webhook via the UI
- I CAN update the logic so the user can edit the name, but not sure its worth it - let me know if you disagree
- We're now also hiding the endpoint, method, format, and secret values when showing the webhook.

Subscriptions
- My billing PR will block the user from seeing the billing page and has a callout that they can go to the external vendor

Users
- Not doing anything different for users, they can be managed normally.

### Dependencies

https://github.com/growthbook/growthbook/pull/3970

### Testing

See testing plan above

### Screenshots
Screenshot 2025-05-29 at 3 08 21 PM

Screenshot 2025-05-29 at 3 08 28 PM

### DIFF
--- a/packages/front-end/components/Settings/Teams/TeamsList.tsx
+++ b/packages/front-end/components/Settings/Teams/TeamsList.tsx
@@ -13,6 +13,7 @@ import { roleHasAccessToEnv, useAuth } from "@/services/auth";
 import Tooltip from "@/components/Tooltip/Tooltip";
 import usePermissionsUtil from "@/hooks/usePermissionsUtils";
 import Badge from "@/components/Radix/Badge";
+import { capitalizeFirstLetter } from "@/services/utils";
 
 const TeamsList: FC = () => {
   const { teams, refreshOrganization, organization } = useUser();
@@ -66,10 +67,9 @@ const TeamsList: FC = () => {
                       {t.managedBy?.type ? (
                         <div>
                           <Badge
-                            label={`Managed by ${
-                              t.managedBy.type.charAt(0).toUpperCase() +
-                              t.managedBy.type.slice(1)
-                            }`}
+                            label={`Managed by ${capitalizeFirstLetter(
+                              t.managedBy.type
+                            )}`}
                           />
                         </div>
                       ) : null}
--- a/packages/front-end/pages/project/[pid].tsx
+++ b/packages/front-end/pages/project/[pid].tsx
@@ -18,6 +18,7 @@ import { useAuth } from "@/services/auth";
 import usePermissionsUtil from "@/hooks/usePermissionsUtils";
 import Frame from "@/components/Radix/Frame";
 import Badge from "@/components/Radix/Badge";
+import { capitalizeFirstLetter } from "@/services/utils";
 
 function hasChanges(value: ProjectSettings, existing: ProjectSettings) {
   if (!existing) return true;
@@ -127,10 +128,7 @@ const ProjectPage: FC = () => {
         {p.managedBy?.type ? (
           <div className="mb-2">
             <Badge
-              label={`Managed by ${
-                p.managedBy.type.charAt(0).toUpperCase() +
-                p.managedBy.type.slice(1)
-              }`}
+              label={`Managed by ${capitalizeFirstLetter(p.managedBy.type)}`}
             />
           </div>
         ) : null}
--- a/packages/front-end/pages/projects/index.tsx
+++ b/packages/front-end/pages/projects/index.tsx
@@ -14,6 +14,7 @@ import usePermissionsUtil from "@/hooks/usePermissionsUtils";
 import Tooltip from "@/components/Tooltip/Tooltip";
 import Button from "@/components/Radix/Button";
 import Badge from "@/components/Radix/Badge";
+import { capitalizeFirstLetter } from "@/services/utils";
 
 const ProjectsPage: FC = () => {
   const { projects, mutateDefinitions } = useDefinitions();
@@ -108,10 +109,9 @@ const ProjectsPage: FC = () => {
                     {p.managedBy?.type ? (
                       <div>
                         <Badge
-                          label={`Managed by ${
-                            p.managedBy.type.charAt(0).toUpperCase() +
-                            p.managedBy.type.slice(1)
-                          }`}
+                          label={`Managed by ${capitalizeFirstLetter(
+                            p.managedBy.type
+                          )}`}
                         />
                       </div>
                     ) : null}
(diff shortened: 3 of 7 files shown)


packages/front-end/components/Settings/Teams/TeamsList.tsx
```diff
@@ -60,6 +63,16 @@ const TeamsList: FC = () => {
                           {t.name}
                         </Link>
                       }
+                      {t.managedBy?.type ? (
+                        <div>
+                          <Badge
+                            label={`Managed by ${
+                              t.managedBy.type.charAt(0).toUpperCase() +
+                              t.managedBy.type.slice(1)
```
romain-growthbook, line 71: Would be nice to make this a utility so it can be re-used.

packages/front-end/pages/project/[pid].tsx
```diff
@@ -123,6 +124,16 @@ const ProjectPage: FC = () => {
             Back to all projects
           </Link>
         </div>
+        {p.managedBy?.type ? (
+          <div className="mb-2">
+            <Badge
+              label={`Managed by ${
+                p.managedBy.type.charAt(0).toUpperCase() +
+                p.managedBy.type.slice(1)
```
Comment (romain-growthbook) [inline packages/front-end/pages/project/[pid].tsx:132]: Same here.

packages/front-end/pages/projects/index.tsx
```diff
@@ -78,7 +79,9 @@ const ProjectsPage: FC = () => {
           <tbody>
             {projects.map((p) => {
               const canEdit = permissionsUtil.canUpdateProject(p.id);
-              const canDelete = permissionsUtil.canDeleteProject(p.id);
+              const canDelete =
+                // If the project has the `managedBy` property, we block deletion.
+                permissionsUtil.canDeleteProject(p.id) && !p.managedBy?.type;
```
romain-growthbook, line 84: Should we update the `canDeleteProject` itself since this logic should apply to each caller. Otherwise we'd have to duplicate this logic at every call site and possibly miss one.

packages/front-end/pages/projects/index.tsx
```diff
@@ -102,6 +105,16 @@ const ProjectsPage: FC = () => {
                     ) : (
                       <span className="font-weight-bold">{p.name}</span>
                     )}
+                    {p.managedBy?.type ? (
+                      <div>
+                        <Badge
+                          label={`Managed by ${
+                            p.managedBy.type.charAt(0).toUpperCase() +
+                            p.managedBy.type.slice(1)
```
romain-growthbook, line 113: ditto

packages/front-end/pages/sdks/SdkWebhooks.tsx
```diff
@@ -68,21 +69,49 @@ export default function SdkWebhooks({
...
       <tr key={webhook.name}>
-        <td style={{ minWidth: 150 }}>{webhook.name}</td>
+        <td style={{ minWidth: 150 }}>
+          {webhook.name}
+          {webhook.managedBy?.type ? (
+            <div>
+              <Badge
+                label={`Managed by ${
+                  webhook.managedBy.type.charAt(0).toUpperCase() +
+                  webhook.managedBy.type.slice(1)
```
romain-growthbook, line 79: ditto

packages/front-end/pages/sdks/[sdkid].tsx
```diff
@@ -50,7 +51,9 @@ export default function SDKConnectionPage() {
 
   const canDuplicate = permissionsUtil.canCreateSDKConnection(connection);
   const canUpdate = permissionsUtil.canUpdateSDKConnection(connection, {});
-  const canDelete = permissionsUtil.canDeleteSDKConnection(connection);
+  const canDelete =
+    permissionsUtil.canDeleteSDKConnection(connection) &&
+    !connection.managedBy?.type;
```
Comment (romain-growthbook) [inline packages/front-end/pages/sdks/[sdkid].tsx:56]: Same here.

packages/front-end/pages/sdks/[sdkid].tsx
```diff
@@ -70,6 +73,17 @@ export default function SDKConnectionPage() {
         ]}
       />
 
+      {connection.managedBy?.type ? (
+        <div className="mb-2">
+          <Badge
+            label={`Managed by ${
+              connection.managedBy.type.charAt(0).toUpperCase() +
+              connection.managedBy.type.slice(1)
```
Comment (romain-growthbook) [inline packages/front-end/pages/sdks/[sdkid].tsx:81]: ditto

packages/front-end/pages/settings/team/[tid].tsx
```diff
@@ -75,6 +76,16 @@ const TeamPage: FC = () => {
             via the <b>Edit Permissions</b> button below.
           </div>
         )}
+        {team.managedBy?.type ? (
+          <div>
+            <Badge
+              label={`Managed by ${
+                team.managedBy.type.charAt(0).toUpperCase() +
+                team.managedBy.type.slice(1)
```
Comment (romain-growthbook) [inline packages/front-end/pages/settings/team/[tid].tsx:84]: ditto

romain-growthbook (commented):
Looking good!

Had some remarks, nothing blocking but would be nice to address.

Typically, I see two calling sites for `canDeleteProject` and also 2 for `canDeleteSDKConnection` we should make sure that both calling sites are updated to handle `manageBy` or to update the underlying logic to work with it.

I understand that the underlying code for each do not currently have the full model but perhaps it would be worth changing that?

packages/front-end/components/Settings/Teams/TeamsList.tsx
mknowlton89: Good call

packages/front-end/pages/projects/index.tsx
romain-growthbook: Got it.

PR #4128, opened 2025-06-01
Allow restricting webhook secrets by origin

jdorn:
### Features and Changes

Webhook secrets are never revealed in the GrowthBook app, but it's still possible for any user in the organization to reveal them by adding a new webhook to a server they control (or something local like ngrok).  Then, they just need to fire a test event and inspect the request headers as it is received.

This PR lets you specify an allowlist of webhook origins for each secret.  When configured, the secret can only be used when sent to one of these origins.

image

### DIFF
--- a/packages/back-end/src/models/WebhookSecretModel.ts
+++ b/packages/back-end/src/models/WebhookSecretModel.ts
@@ -47,11 +47,17 @@ export class WebhookSecretDataModel extends BaseClass {
     });
   }
 
-  public async getBackEndSecretsReplacer() {
+  public async getBackEndSecretsReplacer(origin: string) {
     const secrets = await this.getAll();
     const replacements: Record<string, string> = {};
     for (const secret of secrets) {
-      replacements[secret.key] = secret.value;
+      if (
+        !secret.allowedOrigins ||
+        !secret.allowedOrigins.length ||
+        secret.allowedOrigins.includes(origin)
+      ) {
+        replacements[secret.key] = secret.value;
+      }
     }
 
     return secretsReplacer(replacements);
--- a/packages/front-end/components/EventWebHooks/WebhookSecretModal.tsx
+++ b/packages/front-end/components/EventWebHooks/WebhookSecretModal.tsx
@@ -40,6 +40,7 @@ export default function WebhookSecretModal({
         if (existingId) {
           const body: UpdateWebhookSecretProps = {
             description: data.description,
+            allowedOrigins: data.allowedOrigins,
           };
           // Don't update the value if it's empty
           if (data.value) {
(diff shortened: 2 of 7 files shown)


packages/back-end/src/models/WebhookSecretModel.ts
```diff
@@ -47,11 +47,17 @@ export class WebhookSecretDataModel extends BaseClass {
...
     const secrets = await this.getAll();
     const replacements: Record<string, string> = {};
     for (const secret of secrets) {
-      replacements[secret.key] = secret.value;
+      if (
+        !secret.allowedOrigins ||
+        !secret.allowedOrigins.length ||
+        secret.allowedOrigins.includes(origin)
+      ) {
+        replacements[secret.key] = secret.value;
```
romain-growthbook, line 59: Should this raise instead? If the replacement is not defined, the substitution will fail later but it'll be much harder to track.

Raising here would also give a chance to send back some meaningful error like "secret is restricted to origins: <list>, origin <caller origin> is not found in this list"

packages/front-end/components/EventWebHooks/WebhookSecretModal.tsx
```diff
@@ -79,6 +82,12 @@ export default function WebhookSecretModal({
         textarea
         placeholder="(optional)"
       />
+      <StringArrayField
+        value={form.watch("allowedOrigins")}
+        onChange={(value) => form.setValue("allowedOrigins", value)}
```
romain-growthbook, line 102: Might want to do some validation/normalization like remove trailing slashes. Otherwise `https://www.google.com/` won't work:

```typescript
new URL("https://www.google.com/").origin
'https://www.google.com'
```

packages/back-end/src/models/WebhookSecretModel.ts
jdorn: We can't throw here.  We don't know if the variable is actually being used in the webhook or not yet.  I agree the error message from handlebars is not ideal.  Open to ideas there.

packages/front-end/components/EventWebHooks/WebhookSecretModal.tsx
jdorn: Yeah, good point.  Will probably do this on save and add some more helpful validation

```ts
// Catch common error: missing http(s)
if (value.allowedOrigins.some((url) => !url.startsWith("http"))) {
  throw new Error("Origins must start with http or https");
}

// Make sure values are valid URLs and normalize to just the origin
// new URL throws if the input is invalid
const allowedOrigins = value.allowedOrigins.map((url) => (new URL(url)).origin);
```

packages/back-end/src/models/WebhookSecretModel.ts
romain-growthbook: I think that, more generally, with this change, it looks like webhook secrets are really understood as being attached to a specific webhook. It wouldn't make sense to restrict by origin at the secret level unless you're thinking of a specific webhook you want to allow to use.

In this case, perhaps the solution is to have a `webookIds` array and let the user pick the webhooks this secret is allowed to be used in.

PR #4134, opened 2025-06-03
Add fact metric audit history to fact metric page

lukesonnet:
Adds audit history as a modal you can open from the fact metric dropdown (similar to experiments and features) and uses radix + styling/ordering similar to the experiment dropdown for consistency.

Screenshot 2025-06-03 at 12 02 39 PM
Screenshot 2025-06-03 at 12 02 36 PM

### DIFF
--- a/packages/front-end/pages/fact-metrics/[fmid].tsx
+++ b/packages/front-end/pages/fact-metrics/[fmid].tsx
@@ -54,6 +54,7 @@ import DataList, { DataListItem } from "@/components/Radix/DataList";
 import useOrgSettings from "@/hooks/useOrgSettings";
 import { AppFeatures } from "@/types/app-features";
 import { useCurrency } from "@/hooks/useCurrency";
+import HistoryTable from "@/components/HistoryTable";
 
 function FactTableLink({ id }: { id?: string }) {
   const { getFactTableById } = useDefinitions();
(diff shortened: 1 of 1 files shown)


ahdriel (approved):

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -356,6 +367,36 @@ export default function FactMetricPage() {
...
+          subtitle="Are you sure you want to delete this metric? This action cannot be undone."
+          yesText="Delete"
+          noText="Cancel"
+          modalState={showDeleteModal}
+          setModalState={setShowDeleteModal}
+          onConfirm={async () => {
+            await apiCall(`/fact-metrics/${factMetric.id}`, {
+              method: "DELETE",
+            });
+            mutateDefinitions();
```
Comment (ahdriel) [inline packages/front-end/pages/fact-metrics/[fmid].tsx:393]: nit question: Do we need it here? Or would it be refreshed when navigating to `/metric`?

This showed a `Metric not found message` before redirecting me away, but not a big problem.

Comment (lukesonnet) [reply packages/front-end/pages/fact-metrics/[fmid].tsx:393]: I'm not sure, but it makes it clear it was deleted. A better flow is possible but I'm not going to touch that for now.

PR #4135, opened 2025-06-03
Ensure all string literals are casted to strings in SQL (fixes redshift bug)

### LINKED ISSUE #4132: [Bug] SQL Conversion Error in GrowthBook
### Summary

When using the Power Calculator in GrowthBook with a SQL query that includes string literals, users may encounter an error related to type conversion, specifically: "failed to find conversion function from 'unknown' to text". This issue arises when a string literal is interpreted as an "unknown" type instead of a specific text type (i'm using Redshift datasource. [Similar problem](https://repost.aws/questions/QUEpkcVGysTe29mD-sbr2XBA/redshift-regression-failed-to-find-conversion-function-from-unknown-to-numeric)).

#### Expected Behavior

The Power Calculator should properly handle string literals in SQL queries without resulting in type conversion errors

#### Current Behavior

Image

Executing a SQL query that includes an empty string literal (e.g. '') without explicit type casting results in an error message:
SQL Error [XX000]: ERROR: failed to find conversion function from "unknown" to text

Currently, when I try to run a SQL query with an empty string literal (like ''), I get an error message indicating that the system can't convert it properly. This issue disrupts the ability to use the Power Calculator for metric calculations.

lukesonnet:
### Features and Changes

Fixes a bug with power calculator and other redshift queries as string literals need to be explicitly cast as strings. This does affect other engines since we now add more casts, but this only should apply to power calculator, bandit, and standalone fact metric analysis queries.

- Closes #4132

### Testing

My redshift instance isn't working, but I tested the queries generated manually in  the redshift console.

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -718,7 +718,7 @@ export default abstract class SqlIntegration
         SELECT
           ${settings.userIdType}
           , MIN(${timestampDateTimeColumn}) AS first_exposure_timestamp
-          , '' as variation
+          , ${this.castToString("''")} as variation
         FROM
           __source
         WHERE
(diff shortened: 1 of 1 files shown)


jdorn (approved):
Looks ok to me and relatively safe. Probably worth testing in a few other db engines just to make sure there are no regressions though.

lukesonnet:
Redshift, Snowflake, Postgres, BQ, Clickhouse all work in the power calculator. Checked PG and BQ in bandits, all seemed fine. Merging now.

PR #4138, opened 2025-06-04
Introduce sql explorer

mknowlton89:
### Features and Changes


packages/front-end/components/SchemaBrowser/DatasourceTableData.tsx
```diff
@@ -80,77 +81,81 @@ export default function DatasourceSchema({
...
-        height: "50%",
-      }}
-    >
-      <div className="d-flex justify-content-between px-2 pt-2">
-        <label className="font-weight-bold">
-          <div>
-            <FaTable />{" "}
+    <AreaWithHeader
+      //MKTODO: Can't figure out this color
+      backgroundColor="#192130"
```
mknowlton89: I for the life of my can't figure out the correct approach to get this color to work as expected given our colors aren't solid, so the background their on can change them.

packages/front-end/components/SchemaBrowser/SchemaBrowserWrapper.tsx
```diff
@@ -26,63 +27,68 @@ export default function SchemaBrowserWrapper({
...
-          <span className="pl-1">{datasourceName}</span>
-        </label>
-        {informationSchema && !informationSchema.error && (
-          <label>
-            <Tooltip
-              body={
-                <div>
+    <AreaWithHeader
+      //MKTODO: Can't figure out this color
+      backgroundColor="#192130"
```
mknowlton89: Same here. I for the life of my can't figure out the correct approach to get this color to work as expected given our colors aren't solid, so the background their on can change them.

packages/back-end/src/app.ts
```diff
@@ -782,6 +786,12 @@ app.get(
   informationSchemasController.getInformationSchema
 );
 
+// Saved Queries
+app.get("/saved-queries", savedQueriesController.getSavedQueries);
+app.post("/saved-queries", savedQueriesController.postSavedQuery);
+app.put("/saved-queries/:id", savedQueriesController.putSavedQuery);
+app.delete("/saved-queries/:id", savedQueriesController.deleteSavedQuery);
```
jdorn: Not a big deal, but the newer style of doing this is to create a router file instead of adding all the routes directly to `app.ts`.  Look in `src/routers/` for examples.

packages/back-end/src/services/datasource.ts
```diff
@@ -140,6 +140,46 @@ export async function testDataSourceConnection(
...
+  datasource: DataSourceInterface,
+  query: string,
+  limit?: number
+): Promise<{
+  results?: TestQueryRow[];
+  duration?: number;
+  error?: string;
+  sql?: string;
+}> {
+  if (!context.permissions.canRunTestQueries(datasource)) {
```
jdorn: Change this to `canRunSqlExplorerQueries`

packages/back-end/src/controllers/datasources.ts
```diff
@@ -691,6 +692,42 @@ export async function testLimitedQuery(
   });
 }
 
+export async function runQuery(
+  req: AuthRequest<{
+    query: string;
+    datasourceId: string;
+    limit?: number;
```
jdorn, line 699: If we're going to accept a limit in the request, we need to validate it.  We don't want someone passing in `limit: 100000` and using up all the server memory.

packages/back-end/src/controllers/savedQueries.ts
```diff
@@ -0,0 +1,151 @@
...
+    res.status(200).json({
+      status: 200,
+      savedQueries,
+    });
+  } catch (error) {
+    res.status(500).json({
+      status: 500,
+      message: `Failed to fetch saved queries: ${error}`,
+    });
+  }
```
jdorn: For all of these controller methods, if all you are doing is returning an error response, it's better to just let the error throw and don't wrap in a try/catch.  The default error handler will take care of it (as well as log in Sentry).

packages/back-end/src/controllers/savedQueries.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (!datasource) {
+    throw new Error("Cannot find datasource");
+  }
+
+  try {
+    await context.models.savedQueries.create({
+      name,
+      sql,
+      datasourceId,
+      dateLastRan: new Date(dateLastRan),
```
jdorn: There's a `getValidDate` method in shared that does a little extra validation (and falls back to `Date.now()` if there's an issue).

packages/back-end/src/controllers/savedQueries.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export async function postSavedQuery(
+  req: AuthRequest<{
+    name: string;
+    sql: string;
+    datasourceId: string;
+    results: TestQueryRow[];
+    dateLastRan: string;
+    dataVizConfig?: DataVizConfig[];
+  }>,
```
jdorn: There are `CreateProps` and `UpdateProps` helper types for the base model you should use instead.  Lets you have a single source of truth for the object shape (the zod validator).

Usually, we export helpers from the validator file.  So something like this:

```ts
export type SavedQuery = z.infer<typeof savedQueryValidator>;
export type SavedQueryCreateProps = CreateProps<SavedQuery>;
export type SavedQueryUpdateProps = UpdateProps<SavedQuery>;
```

packages/back-end/src/controllers/savedQueries.ts
```diff
@@ -0,0 +1,151 @@
...
+      dataVizConfig: DataVizConfig[];
+    }> = {};
+
+    if (name !== undefined) updateData.name = name;
+    if (sql !== undefined) updateData.sql = sql;
+    if (datasourceId !== undefined) updateData.datasourceId = datasourceId;
+    if (results !== undefined) updateData.results = results;
+    if (dateLastRan !== undefined)
+      updateData.dateLastRan = new Date(dateLastRan);
+    if (dataVizConfig !== undefined) updateData.dataVizConfig = dataVizConfig;
```
jdorn: Cleaner if you can just use `req.body`.

```ts
const updateData = {
  ...req.body,
  dateLastRan: req.body.dateLastRan ? getValidDate(req.body.dateLastRan) : undefined
}
```

packages/back-end/src/controllers/savedQueries.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    await context.models.savedQueries.updateById(id, updateData);
```
jdorn: If you already have the `savedQuery` object, use the `update` method instead of `updateById` to avoid an extra db lookup.  Alternatively, get rid of the `getById` call above (updateById does the same validation to make sure it exists before updating)

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -65,6 +77,25 @@ export default function CodeTextArea({
...
+      styleElement.type = "text/css";
+      document.head.appendChild(styleElement);
+    }
+
+    styleElement.innerHTML = aceThemeOverrides;
+
+    return () => {
+      styleElement.remove();
+    };
+  }, []);
```
jdorn: Why do this vs just add the styles to `global.scss`?

packages/back-end/src/validators/saved-queries.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from "zod";
+
+// TODO: Add a proper type for the data viz config
+export const dataVizConfigValidator = z.any();
```
mknowlton89, line 5: Hopefully we can define this soon

packages/back-end/src/controllers/datasources.ts
mknowlton89: I don't think we still need this now that you've built your `ensureMaxLimit` function?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1435,6 +1444,7 @@ export default abstract class SqlIntegration
     sql: string,
     timestampCols?: string[]
   ): Promise<TestQueryResult> {
+    console.log("runTestQuery", sql);
```
jdorn: Remove

packages/back-end/src/services/datasource.ts
```diff
@@ -140,6 +140,45 @@ export async function testDataSourceConnection(
...
+    throw new Error("Unable to test query.");
+  }
+
+  const sql = integration.getFreeFormQuery(query);
+  try {
+    const { results, duration } = await integration.runTestQuery(sql, [
+      "timestamp",
+    ]);
+    return {
+      results,
```
jdorn, line 177: There are a few edge cases that `ensureLimit` does not handle, so there's a small chance more than 1000 rows will be returned here.  I think it's worth adding a sanity check here to truncate the array if it's bigger than 1000.

packages/back-end/src/models/SavedQueryDataModel.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+  }
+  protected canCreate(doc: SavedQuery): boolean {
+    const { datasource } = this.getForeignRefs(doc);
+    return this.context.permissions.canRunSqlExplorerQueries({
+      projects: datasource?.projects || [],
+    });
+  }
+  protected canUpdate(existing: SavedQuery): boolean {
+    const { datasource } = this.getForeignRefs(existing);
```
jdorn: A couple things:
1. For both create/update, if `datasource` is undefined in the foreign refs, that means the id is invalid, so we should throw instead of using the null coalesce operator.  Note: the same technically applies to read/delete, but I think it's ok to be more lenient there (e.g. if the data source is deleted, we still want to be able to delete a saved query).
2. For update, if the datasource id is changing, we need to validate the new one as well (both that the foreign ref exists and that the user has permission)

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -88,6 +88,11 @@ const navlinks: SidebarLinkProps[] = [
         href: "/datasources",
         path: /^datasources/,
       },
+      {
+        name: "SQL Explorer",
+        href: "/sql-explorer",
+        path: /^sql-explorer/,
+      },
```
jdorn, line 97: Should we add a feature flag for this?

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -0,0 +1,583 @@
...
+  const [selectedDatasourceId, setSelectedDatasourceId] = useState(
+    datasourceId || ""
+  );
+  const [dirty, setDirty] = useState(name ? false : true);
+  const [loading, setLoading] = useState(false);
+  const [isRunningQuery, setIsRunningQuery] = useState(false);
+  const [
+    queryExecution,
+    setQueryExecution,
+  ] = useState<QueryExecutionResult | null>(null);
```
jdorn: I think it makes better sense to store this data in the model itself and not in a temporary state here.  Will simplify quite a bit and having the execution stats in Mongo will be helpful as well.

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -0,0 +1,583 @@
...
+
+  // Pre-fill results if we're editing a saved query with existing results
+  useEffect(() => {
+    if (results && results.length > 0) {
+      setQueryExecution({
+        results,
+        sql,
+      });
+    }
+  }, [results, sql]);
```
jdorn: I haven't tested it, but this useEffect is setting off alarm bells to me that it's likely to cause bugs.  I think if we make the change I mentioned above where we store query execution stats in the model directly, we can get rid of the hook here entirely.

packages/back-end/src/controllers/savedQueries.ts
```diff
@@ -0,0 +1,151 @@
...
+    });
+  } catch (error) {
+    res.status(500).json({
+      status: 500,
+      message: `Failed to fetch saved queries: ${error}`,
+    });
+  }
+}
+
+export async function postSavedQuery(
```
jdorn: For all of these methods, should we add commercial feature checks?  Thinking maybe for the getAll endpoint, we can make it just return an empty array and maybe a flag saying some saved queries are hidden.  So if someone downgrades from Pro to Free, we don't delete any of their queries, we just don't show them in the UI anymore.

packages/front-end/components/SavedQueries/SavedQueriesList.tsx
```diff
@@ -0,0 +1,247 @@
...
+    setSelectedSavedQuery({
+      ...query,
+      id: "",
+      name: `${query.name}-copy`,
+    });
+  }, []);
+
+  const handleRowClick = useCallback(
+    (query: SavedQuery) => {
+      // Only allow row click for queries the user can edit
```
ahdriel: question: Do we still want this if the modal is the only way to see the graphs at the moment?

mknowlton89: Great callout - you're correct in that this doesn't make sense anymore.

PR #4145, opened 2025-06-04
add new hook/endpoint/ui to better surface sdk webhooks and their errors

bryce-fitzsimons:
Surface SDK webhooks and their errors on the SDK Connections listing page

- modify `useSDKConnections()` hook to optionally also fetch SDK webhooks
- new endpoint to get SDK webhooks
- new column on listing page with error tooltips

image


packages/back-end/src/routers/sdk-connection/sdk-connection.controller.ts
```diff
@@ -191,6 +191,45 @@ export const checkSDKConnectionProxyStatus = async (
...
+
+      webhooksByConnection[connection.id] = webhooks;
+    })
+  );
+
+  res.status(200).json({
+    status: 200,
+    webhooks: webhooksByConnection,
+  });
+};
```
jdorn, line 235: This could potentially be dozens or hundreds of parallel Mongo queries.  Can we use the `findAllSdkWebhooksByConnectionIds` instead which fetches all webhooks in a single query.  Then, group them in code after.

packages/front-end/pages/sdks/SdkWebhooks.tsx
```diff
@@ -24,6 +24,7 @@ import usePermissionsUtil from "@/hooks/usePermissionsUtils";
 import ClickToReveal from "@/components/Settings/ClickToReveal";
 import Badge from "@/components/Radix/Badge";
 import { capitalizeFirstLetter } from "@/services/utils";
+import Callout from "@/components/Radix/Callout";
```
jdorn, line 28: I just noticed this file should be in `components`, not `pages`.  Right now, you can go here and cause an error: https://app.growthbook.io/sdks/SdkWebhooks

packages/front-end/hooks/useSDKConnections.ts
```diff
@@ -1,8 +1,37 @@
...
+) {
+  const { includeWebhooks = false } = options;
+
+  const connectionsResponse = useApi<{
     connections: SDKConnectionInterface[];
   }>(`/sdk-connections`);
+
+  const webhooksResponse = useApi<{
+    webhooks: Record<string, WebhookInterface[]>;
+  }>(`/sdk-connections/webhooks`, { shouldRun: () => includeWebhooks });
```
jdorn: This feels like it should either be a separate hook, or we should merge the API calls and just have `GET /sdk-connections` optionally return back webhooks as well.

packages/back-end/src/routers/sdk-connection/sdk-connection.controller.ts
```diff
@@ -191,6 +191,45 @@ export const checkSDKConnectionProxyStatus = async (
...
+      const webhooks = await findAllSdkWebhooksByConnection(
+        context,
+        connection.id
+      );
+
+      // If user does not have write access, remove the shared secret
+      if (!context.permissions.canUpdateSDKWebhook(connection)) {
+        webhooks.forEach((w) => {
+          w.signingKey = "";
+        });
```
jdorn: Not a big deal, but do we ever need this field in this context?  Feels like we should just return the minimum set of fields needed in this context.  So just id, name, and error.  Can make a new `SdkWebhookSummaryInfo` type

packages/front-end/hooks/useSDKConnections.ts
bryce-fitzsimons: moved to separate hook

packages/back-end/src/routers/sdk-connection/sdk-connection.controller.ts
bryce-fitzsimons: added a summary type (with a few other summary fields, just in case)

packages/back-end/src/routers/sdk-connection/sdk-connection.controller.ts
```diff
@@ -191,6 +194,39 @@ export const checkSDKConnectionProxyStatus = async (
...
+  // Group webhooks by connection ID
+  const webhooksByConnection: Record<string, WebhookSummary[]> = {};
+  connections.forEach((connection) => {
+    const webhooks = allWebhooks
+      .filter((w) => w.sdks.includes(connection.id))
+      .map((w) =>
+        pick(w, ["id", "name", "endpoint", "lastSuccess", "error", "created"])
+      );
+    webhooksByConnection[connection.id] = webhooks;
+  });
```
jdorn, line 229: Nit: This is `O(n^2)`, if you loop over `allWebhooks` instead it will be `O(n)`.  Not a big deal in this case since the numbers are small.

jdorn (approved):

packages/back-end/src/routers/sdk-connection/sdk-connection.controller.ts
bryce-fitzsimons: good catch, made this more efficient

PR #4148, opened 2025-06-05
feat(refactor): added useProjectDefinitions hook

### LINKED ISSUE #3600: [Feature] Hooks for project-specific fact tables, metrics, segments, etc.
### Description of Feature

Often it is important to retrieve entities by specific project (fact tables, metrics) and we call `useDefinitions()` to load them all and then filter by project. It probably would be more efficient re: requests and for future devs if we had hooks that took project as an input and returned a filtered list. Either that or we modify `useDefinitions()` or something similar to take project.

Examples of what could be simplified by this:
* https://github.com/growthbook/growthbook/blob/bcc5cf7fae7af05b468c3d368aa9137ae053edf5/packages/front-end/components/Experiment/MetricSelector.tsx#L91-L108
* https://github.com/growthbook/growthbook/blob/bcc5cf7fae7af05b468c3d368aa9137ae053edf5/packages/front-end/pages/fact-tables/index.tsx#L104-L108
* https://github.com/growthbook/growthbook/blob/bcc5cf7fae7af05b468c3d368aa9137ae053edf5/packages/front-end/components/FactTables/FactTableModal.tsx#L51-L53

And the comment that inspired this:
* https://github.com/growthbook/growthbook/pull/3564#discussion_r1934705017

devcodes9:
### Features and Changes
- Introduces useProjectDefinitions hook to centralize project-scoped filtering of definitions (metrics, fact metrics, fact tables, data sources).
- Closes #3600

### Dependencies
None

### Testing
- Try using Metric Selector
- Try interacting with Fact table and Fact Metric modals
- Check the Metrics list


jdorn:

PR #4151, opened 2025-06-05
Fix login code.

romain-growthbook:


packages/back-end/src/services/organizations.ts
```diff
@@ -379,6 +382,14 @@ export async function addMemberToOrg({
...
     true
   );
+
+  await BluebirdPromise.each(teams, (teamId) =>
+    addMembersToTeam({
+      organization: updatedOrganization,
+      userIds: [userId],
+      teamId,
+    })
+  );
```
jdorn: There's a much simpler way to do this.  Just add the teams array directly when building the members object above:

```ts
  const members: Member[] = [
    ...organization.members,
    {
      id: userId,
      role,
      limitAccessByEnvironment,
      environments,
      projectRoles,
      dateCreated: new Date(),
      externalId,
      managedByIdp,
      teams
    },
  ];
```

jdorn (approved):

PR #4159, opened 2025-06-09
Enable url redirect when clicking on experimentation items.

romain-growthbook:


packages/front-end/pages/oauth/vercel-integration.tsx
```diff
@@ -8,10 +8,19 @@ const VercelPage = () => {
...
+      state,
+      resource_id: resourceId,
+      experimentation_item_id: experimentationItemId,
+    } = router.query;
 
     if (!code || !state) return;
 
+    const baseUrl =
+      "/" +
+      (experimentationItemId ? experimentationItemId.replace(":", "/") : "");
```
jdorn: Because of the `catch` block below, this is a potential security vulnerability.  An attacker can craft a URL that will then redirect anywhere they want.  Even if the code/state fails to validate (i.e. it's coming from an attacker, not actually Vercel), the redirect will still happen.

Example URL to trigger the attack:
```
/oauth/vercel-integration?code=foo&state=bar&experimentation_item_id=%2Fgoogle.com
```

I would make 2 changes to fix this:
1. Change the `catch` behavior to render the error message on the page without redirecting.  This will be better for debugging as well.
2. Validate the item id matches a format you expect.  `^(features|experiment):`.  If it doesn't match, just redirect to `/` instead.

PR #4160, opened 2025-06-09
Added validation to prevent deletion of data sources with associated fact metrics/tables

### LINKED ISSUE #3073: [Bug] Deleting a data source does not delete the fact metrics
It does delete metrics (I believe), but not the fact table or fact table metrics. They show up as coming from an unknown data source. 

Screenshot 2024-10-04 at 1 21 57 PM

Clicking on 'unknown' goes to a 404 page for the deleted data source.

devcodes9:
### Features and Changes
Prevent a datasource from being deleted if there are still Fact Metrics or Fact Tables pointing at it, so we don’t accidentally orphan downstream data. Added validations for the same.

- Fixes #3073 

### Testing
- Create a DataSource A
- Create a Fact Metric (and/or Fact Table) tied to A
- Attempt to delete A
- Delete all metrics & tables associated to A, then delete A again

### Screenshots
image


jdorn:

PR #4161, opened 2025-06-09
Add memory (btye, kB) formatting for fact metrics

lukesonnet:
### Features and Changes

Adds Memory (byte or kilobyte) formatting for fact table columns to improve formatting.

### Testing

Created fact metrics with both formatting and ran them in an analysis.

### Screenshots

Screenshot 2025-06-09 at 4 29 58 PM
Screenshot 2025-06-09 at 4 30 01 PM

Screenshot 2025-06-09 at 5 07 43 PM
Screenshot 2025-06-09 at 5 08 04 PM

### DIFF
--- a/packages/front-end/services/metrics.tsx
+++ b/packages/front-end/services/metrics.tsx
@@ -226,6 +226,50 @@ export function formatDurationSeconds(value: number) {
 
   return f;
 }
+
+export function formatMemoryBytes(value: number) {
+  if (value < 1000 * 1000) {
+    return formatNumber(value, {
+      style: "unit",
+      unit: "byte",
+    });
+  } else if (value < 1000 * 1000 * 1000) {
+    return formatNumber(value / 1000, {
+      style: "unit",
+      unit: "kilobyte",
+    });
+  } else if (value < 1000 * 1000 * 1000 * 1000) {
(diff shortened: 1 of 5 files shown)


packages/front-end/services/metrics.tsx
```diff
@@ -266,6 +266,14 @@ export function formatPercentagePoints(value: number) {
   return `${number} pp`;
 }
 
+export function formatBytes(value: number) {
+  return formatByteSizeString(value, true);
+}
+
+export function formatKilobytes(value: number) {
+  return formatByteSizeString(value / 1000, true);
```
jdorn: Should this be `*1000`? Maybe I'm misunderstanding the formats, but if you tell us a column is in KB and the metric value is 23, I'd expect to call the formatByte method with 23000

lukesonnet: ugh, it should be `/ 1024`. The unit is in kilobytes, and I'm making it bytes for the function that expects bytes.

lukesonnet: Ugh, it should be `* 1024` i got it backwards again.

jdorn (approved):

PR #4174, opened 2025-06-11
Disable/Enable User Access Action

### LINKED ISSUE #3870: [Feature] Temporarily disable user access
### Description of Feature

The ability to temporarily disable a user's access so that their seat can be reallocated during extended leave for example.

devcodes9:
### Features and Changes
- This update introduces the ability to temporarily disable organization members. Disabled users retain their membership but are prevented from signing in or accessing the app until re-enabled.
- Member statistics (numMembers) now exclude disabled users in counting of active seats
- Added PUT /member/:id/:disable which toggles a member’s disabled status via the putMemberAccess controller.

- Closes #3870 

### Testing
- Navigate to Settings > Team.
- Verify the new “Status” column correctly shows Active or Disabled for each member.
- Clicked the ellipsis menu for a team member and selected “Disable User”.
- Confirm the confirmation modal appears and requires user action.
- After confirming, the member's status updates to Disabled in the list.
- Verified the "Enable User" option appears for disabled members and successfully re-enables them.
- Check Logging in as the disabled user.
- Confirm after disabling, we can reallocate that saved seat to someone else.

### Screenshots
[Screen recording (demo video)](https://drive.google.com/file/d/1YnO3IjqD3dus4WEGho_kpA4GbUwp4iTw/view?usp=sharing)


jdorn:
I don't think we need this functionality.  The user can just be deleted from the organization.  We keep all user id references intact, so if they are added back to the org later, everything will be restored.

Having an explicit disabled state is nice, but introduces extra complexity and risk in critical surfaces such as billing and authorization, which doesn't seem worth it to me.

devcodes9:
> I don't think we need this functionality. The user can just be deleted from the organization. We keep all user id references intact, so if they are added back to the org later, everything will be restored.
> 
> Having an explicit disabled state is nice, but introduces extra complexity and risk in critical surfaces such as billing and authorization, which doesn't seem worth it to me.

That makes sense.
Thanks for reviewing @jdorn
Should we close this PR and issue?

jdorn:

PR #4175, opened 2025-06-11
add workflow to handle version event

madhuchavva:
### Features and Changes

- Workflow that handles a dispatched event for SDK release and updates its version in sdk-versioning.

### DIFF
--- a/.github/workflows/update-sdk-versions.yml
+++ b/.github/workflows/update-sdk-versions.yml
@@ -1,0 +1,114 @@
+# This workflow receives a repository_dispatch event from the SDK repositories
+# and updates the SDK version in the sdk-versions directory.
+# It then creates a pull request to the main branch.
+
+name: Update SDK Versions
+
+on:
+  repository_dispatch:
+    types: [update-sdk-version]  # Triggered by SDK repositories
+  workflow_dispatch:
+    inputs:
+      sdk:
+        description: 'SDK to update (python, js, etc.)'
+        required: true
+        default: 'python'
+      version:
+        description: 'Version to update to'
+        required: true
+
+jobs:
+  update-sdk-versions:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get Update Parameters
+        id: params
+        run: |
+          if [ "${{ github.event_name }}" == "repository_dispatch" ]; then
+            echo "sdk=${{ github.event.client_payload.sdk }}" >> $GITHUB_OUTPUT
+            echo "version=${{ github.event.client_payload.version }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/update-sdk-versions.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get Update Parameters
+        id: params
+        run: |
+          if [ "${{ github.event_name }}" == "repository_dispatch" ]; then
+            echo "sdk=${{ github.event.client_payload.sdk }}" >> $GITHUB_OUTPUT
+            echo "version=${{ github.event.client_payload.version }}" >> $GITHUB_OUTPUT
+            echo "source_repo=${{ github.event.client_payload.source_repo }}" >> $GITHUB_OUTPUT
```
jdorn, line 41: Should we have any validation to make sure it's coming from a repo within the growthbook organization?

.github/workflows/update-sdk-versions.yml
```diff
@@ -0,0 +1,114 @@
...
+          
+          # Create a new branch
+          BRANCH_NAME="update-${SDK}-sdk-${VERSION}"
+          git checkout -b "$BRANCH_NAME"
+          
+          # Update the JSON file (add new version at the beginning)
+          jq --arg v "$VERSION" \
+             '.versions = ([{"version": $v}] + .versions)' \
+             "$SDK_FILE" > "${SDK_FILE}.tmp" && mv "${SDK_FILE}.tmp" "$SDK_FILE"
+
```
jdorn, line 173: We should also update the version in the file `docs/src/data/SDKInfo.json`

.github/workflows/update-sdk-versions.yml
madhuchavva: yeah, it's possible to add that check. We can send `source_repo` in the event payload and verify it in our repos before we move on. Let me add that.

.github/workflows/update-sdk-versions.yml
madhuchavva: done!

.github/workflows/update-sdk-versions.yml
madhuchavva: added a validation for the source_repo

.github/workflows/update-sdk-versions.yml
```diff
@@ -28,87 +32,186 @@ jobs:
...
+              rm -f "${SDK_FILE}.tmp"
+              exit 1
+            fi
           fi
+
+      - name: Update SDK Documentation
+        run: |
+          SDK="${{ steps.params.outputs.sdk }}"
+          VERSION="${{ steps.params.outputs.version }}"
+          SDKINFO_FILE="docs/src/data/SDKInfo.json"
```
jdorn, line 137: Ryan just changed this recently to be a `.ts` file instead of `.json`. There's a script to re-generate it, but might be easiest to just remove this step for now to unblock you.

madhuchavva:
Switched to a different model now and this is no longer needed.

PR #4177, opened 2025-06-11
Update SDKs versions and add supported features section

royalfig:
- Updates SDK Resources generation script
- Adds Supported Features section to SDK docs

CleanShot 2025-06-12 at 11 42 26

### DIFF
--- a/docs/docs/lib/edge/lambda.mdx
+++ b/docs/docs/lib/edge/lambda.mdx
@@ -11,7 +11,7 @@ import Pill from '@site/src/components/Pill';
 import SdkResources from '@site/src/components/SdkResources';
 import SdkSupportedFeatures from '@site/src/components/SdkSupportedFeatures';
 
-# Lambda@Edge Edge App & SDK <Pill></Pill>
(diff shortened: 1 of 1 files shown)


docs/docs/lib/edge/lambda.mdx
```diff
@@ -9,8 +9,9 @@ toc_max_heading_level: 5
 import ExternalLink from '@site/src/components/ExternalLink';
 import Pill from '@site/src/components/Pill';
 import SdkResources from '@site/src/components/SdkResources';
+import SdkSupportedFeatures from '@site/src/components/SdkSupportedFeatures';
 
-# Lambda@Edge Edge App & SDK <Pill>beta</Pill>
+# Lambda@Edge Edge App & SDK <Pill></Pill>
```
Auz: Do we need the pill if there is nothing in it?

Auz (approved):

docs/docs/lib/edge/lambda.mdx
royalfig: Thanks. I think there was a type issue here, which is why it got deleted. It's now resolved.

PR #4185, opened 2025-06-13
Base Moments class for science code

lukebrawleysmith:
Currently to conduct inference using our codebase, we need to pass in 2 TestStatistics. To be able to use post-stratification in a Bayesian or sequential framework, we need to be able to pass in post-stratification results, rather than test statistics.

This PR changes our science codebase so that the first step in all tests is to calculate the first and second moments (i.e., the mean and squared standard error, respectively) of the effect estimate. These moments are then passed into either frequentist non-sequential, frequentist sequential, or the Bayesian testing engines.  
As an additional benefit, this PR simplifies our code base.  

No results should change as a result of this PR.

This PR does not contain any post-strat work, and is a pre-requisiste for post-strat.

Testing: 
All current unit tests work.  I also compared results to the main branch for frequentist (with and without sequential) and Bayesian engines and the results were unchanged.


packages/stats/gbstats/models/tests.py
```diff
@@ -97,9 +260,54 @@ def __init__(
...
+# Tests
+class BaseABTest(ABC):
+    def __init__(
+        self,
+        stats: List[Tuple[TestStatistic, TestStatistic]],
+        config: BaseConfig = BaseConfig(),
+    ):
+        self.stats = stats
+        stats_a, stats_b = zip(*self.stats)
+        self.stat_a = reduce(operator.add, stats_a)
```
lukesonnet: Shouldn't we only do this in the non post-stratify case? so doesn't this belong in the `compute_moments_result` method or somewhere else?

Should there be an error if the length is > 1 and there is a non-summable statistic (e.g. a quantile statistic)?

packages/stats/gbstats/models/tests.py
```diff
@@ -97,9 +260,54 @@ def __init__(
...
+        self.stat_b = reduce(operator.add, stats_b)
+        self.config = config
+        self.alpha = config.alpha
+        self.relative = config.difference_type == "relative"
+        self.scaled = config.difference_type == "scaled"
+        self.traffic_percentage = config.traffic_percentage
+        self.total_users = config.total_users
+        self.phase_length_days = config.phase_length_days
+        self.moments_result = self.compute_moments_result()
```
lukesonnet, line 284: Let's not compute in the initialization step.

Instead, let's change `compute_result` to be non-abstract in this case.

Let's change it to be something like:

```
def compute_result(self) -> TestResult:
    moments = self.compute_moments_result()
    result = self.compute_statistics(moments)
    return result
    
@abstractmethod
def compute_statistics():
    pass
```
Where compute_statistics is abstract and implemented by the child classes.

Not sure if this will work with type hinting, but lets try and I can maybe explore the type hinting more in depth.

lukebrawleysmith: Per offline conversation, I am leaving `compute_result` in the `__init__` method, as it would require broader re-write that is out of scope (e.g., changing p-value to not be a property and moving inside of `compute_result`).

packages/stats/gbstats/models/tests.py
lukebrawleysmith: changed logic so that we pass stats to `compute_moments_result` and added error

packages/stats/gbstats/models/tests.py
```diff
@@ -1,29 +1,54 @@
...
     compute_theta_regression_adjusted_ratio,
 )
 from gbstats.models.settings import DifferenceType
+from gbstats.utils import isinstance_union, frequentist_diff, frequentist_variance
 
 
 # Configs
+@dataclass
+class EffectMomentsConfig:
+    difference_type: DifferenceType = "relative"
```
lukesonnet: Should this be just `relative` or `absolute`? Seems scaled impact is just absolute here, right?

packages/stats/gbstats/models/tests.py
```diff
@@ -97,9 +247,64 @@ def __init__(
...
+        )
+
+
+def sum_stats(
+    stats: List[Tuple[TestStatistic, TestStatistic]]
+) -> Tuple[TestStatistic, TestStatistic]:
+    stats_a, stats_b = zip(*stats)
+    summable_check_a = all(isinstance(stat_a, SummableStatistic) for stat_a in stats_a)
+    summable_check_b = all(isinstance(stat_b, SummableStatistic) for stat_b in stats_b)
+    if len(stats_a) > 1 and (not summable_check_a or not summable_check_b):
```
lukesonnet, line 261: nit: if len is 1 and it is not summable, does reduce operator work? Can we add some unit tests for this case?

Maybe instead we should check `len(stats)` and if that is 1 we just return the 0th items directly?

packages/stats/gbstats/frequentist/tests.py
```diff
@@ -330,8 +213,10 @@ def p_value(self) -> float:
 
     @property
     def confidence_interval(self) -> List[float]:
-        halfwidth: float = t.ppf(1 - self.alpha / 2, self.dof) * np.sqrt(self.variance)
-        return two_sided_confidence_interval(self.point_estimate, halfwidth)
+        halfwidth: float = t.ppf(1 - self.alpha / 2, self.dof) * self.moments_result.standard_error  # type: ignore
```
lukesonnet: nit: what's this type ignore?

packages/stats/gbstats/models/tests.py
lukebrawleysmith: good call, updated

packages/stats/gbstats/models/tests.py
lukebrawleysmith: I added unit tests.  
I'm slightly inclined to leave as is, as the code is shorter and less verbose, and we have unit tests verifying it works.  If you feel differently I'm fine adding the len == 1 case.

packages/stats/gbstats/frequentist/tests.py
lukebrawleysmith: good call, I changed the t.ppf output from array to float, and the type error went away

PR #4190, opened 2025-06-16
Fix controlled tooltips not rendering

### LINKED ISSUE #4163: [Bug] Attributes : Unable to view references
### Summary

Unable to view references of an attribute  on the UI

#### Expected Behavior

Should look like this as per [this](https://github.com/growthbook/growthbook/pull/2930) PR

Image

#### Current Behavior

Pop up with reference information does not render.

Kevin-Chant:
### Features and Changes

Fixes an issue where controlled tooltips (via the `state` prop) weren't rendering due to the new fade in/out logic.

- Closes #4163

### Testing

Check the tooltips on the `/attributes` page

### Screenshots

image

### DIFF
--- a/packages/front-end/components/Tooltip/Tooltip.tsx
+++ b/packages/front-end/components/Tooltip/Tooltip.tsx
@@ -6,6 +6,7 @@ import {
   useEffect,
   CSSProperties,
   useRef,
+  useCallback,
 } from "react";
 import { usePopper } from "react-popper";
 import clsx from "clsx";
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Tooltip/Tooltip.tsx
```diff
@@ -54,11 +55,36 @@ const Tooltip: FC<Props> = ({
...
+
+  const handleMouseEnter = useCallback(() => {
+    clearTimeouts();
+    timeoutRef.current = setTimeout(() => {
+      setOpen(true);
+      setTimeout(() => setFadeIn(true), 50);
+    }, delay);
+  }, [clearTimeouts, timeoutRef, setOpen, setFadeIn, delay]);
+
+  const handleMouseLeave = useCallback(() => {
```
ahdriel, line 73: question: If this is truly a controlled component then we should not close it automatically based on mouseLeave, no?

We can still have a onMouseLeave listener prop here that the parent can use to update `state` if they want

Kevin-Chant: Hmm, yeah this is a weird component because it's partially/optionally controlled. I went with keeping the mouseLeave handler & just invoking it manually because the main issue was not having a trigger ref to invoke it

ahdriel: > this is a weird component because it's partially/optionally controlled

I don't think we should allow this behavior though -- we should either have a controlled input where both open and onOpenChange need to be provided or neither and then it manages itself.

PR #4191, opened 2025-06-16
fixed the side nav hover color

gazzdingo:
### Features and Changes
updated the color of the side nav hover color to use radix 
AFTER
image
BEFORE
image

### DIFF
--- a/packages/front-end/components/Layout/SidebarLink.module.scss
+++ b/packages/front-end/components/Layout/SidebarLink.module.scss
@@ -39,8 +39,12 @@ $topbarheight: 60px;
   }
 
   &:hover {
-    background-color: color-mix(in srgb, var(--violet-9) 30%, transparent);
-
+    background-color: $violet-9-30;
+    @supports (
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Layout/SidebarLink.module.scss
```diff
@@ -39,7 +39,7 @@ $topbarheight: 60px;
   }
 
   &:hover {
-    background-color: scale-color($dark-purple, $lightness: +12%);
+    background-color: color-mix(in srgb, var(--violet-9) 30%, transparent);
```
jdorn: Doesn't look like `color-mix` is super widely supported yet across browsers.  Does it fall back gracefully?

gazzdingo: what is our oldest browser support that we want? this has had support since 2023 for most browsers and has 95.66% support.
 the only other thing we can do is fall back to a hardcoded color (not from radix)

PR #4195, opened 2025-06-17
Added tabs to the results page to toggle between Metric Groups

### LINKED ISSUE #3504: [Feature] Add tabs to the results page to toggle between Metric Groups
### Description of Feature

Something like the following should provide one tab per metric group, with all metrics as the default first metric to provide quick views of metric groups together.

Screenshot 2025-01-08 at 2 57 05 PM

devcodes9:
### Features and Changes
- New state for the selected metric group tab.
- Added a tab bar to select metric group.
- Results are now conditionally rendered based on the selected group.

- Closes #3504 

### Testing
- Go to results tab in experiments page
- Verify that the metric groups tabs are visible. By default, all metrics are displayed, and switching to the metric groups tab only filters the relevant metrics.

### Screenshots

https://github.com/user-attachments/assets/2c67f200-5f92-422c-95db-61053998ed22

image


jdorn:
We're going to want a different design for this.  There is an implementation already in the branch `jd/metric-variants` (along with tons of unrelated changes). Probably best to use the version in that branch and separate it into its own PR, but will take a bit of work.

jdorn:

PR #4196, opened 2025-06-17
Vercel Native onboarding improvements

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Layout/Layout.tsx
+++ b/packages/front-end/components/Layout/Layout.tsx
@@ -376,7 +376,7 @@ const backgroundShade = (color: string) => {
 const Layout = (): React.ReactElement => {
   const [open, setOpen] = useState(false);
   const settings = useOrgSettings();
-  const { accountPlan, license, subscription, organization } = useUser();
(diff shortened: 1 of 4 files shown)


packages/front-end/components/Layout/Layout.tsx
```diff
@@ -386,7 +386,8 @@ const Layout = (): React.ReactElement => {
 
   const [upgradeModal, setUpgradeModal] = useState(false);
   const showUpgradeButton =
-    ["oss", "starter"].includes(accountPlan || "") ||
+    (["oss", "starter"].includes(accountPlan || "") &&
+      !organization.isVercelIntegration) ||
```
mknowlton89: 50/50 on this change. On one hand, think hiding this could be beneficial, since they can't actually do it (they just get instructed to go to Vercel).

jdorn: I would probably not do this change.  There are several places in the app we tell people to upgrade, but we don't link to the upgrade modal directly and rely on that Upgrade button always being in the left nav.

jdorn (approved):

PR #4202, opened 2025-06-18
Introduce post-stratification

lukebrawleysmith:
This PR introduces the science code for post stratification as well as releasing it for users. By default it is enabled, but users have the ability to opt-out on the settings page. This will affect results for any users who already have pre-computed dimensions set-up (by improving their accuracy)!

This PR also contains a substantial refactoring of how `gbstats.py` passes around data, using data classes more often and using data frames less often, in order to improve the developer experience.


packages/stats/tests/test_shared_models.py
```diff
@@ -256,23 +257,19 @@ def test_quantile_failure(self):
...
+            stats=[(stat_a, stat_b)],
+            config=FrequentistConfig(difference_type="absolute"),
+        )
         moments = EffectMoments(
-            [(stat_a, stat_b)], config=EffectMomentsConfig(difference_type="absolute")
+            [(test.stat_a, test.stat_b)],
+            config=EffectMomentsConfig(difference_type="absolute"),
         )
-        self.assertEqual(moments.variance, -1.2010673868312758)
+        self.assertEqual(moments.variance, -1.1023019547325101)
```
lukebrawleysmith, line 293: note that the previous unit test was poorly written (by me).  The theta value of 0.8333 from `test_initialize_theta` method was used in the `test_negative_variance` method, instead of being recalculated after stat_a is updated.  `theta` is now recalculated in `test_negative_variance`, which changes the variance to -1.1.

packages/stats/gbstats/models/statistics.py
```diff
@@ -57,8 +57,9 @@ def mean(self):
         return self.sum / self.n
 
     def __add__(self, other):
-        if not isinstance(other, SampleMeanStatistic):
-            raise TypeError("Can add only another SampleMeanStatistic instance")
+        assert isinstance(
+            other, Union[ProportionStatistic, SampleMeanStatistic]
+        ), "Can add only another ProportionStatistic or SampleMeanStatistic instance"
```
lukesonnet: Can we still use the isinstance and TypeError flow? Raising a TypeError instead of an AssertionError seems better, and you can disable assertions in prod if you want.

Also, why are we allowing additions of different statistic types? Unless we have a valid use case I think we should continue to disallow it.

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +326,963 @@ def scaled_impact_eligible(self) -> bool:
...
     def compute_result(self) -> TestResult:
         pass
+
+
+@dataclass_with_arbitrary_types_allowed
+class StrataResult:
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    metric_type: Literal["count", "ratio"]
+    n: int
+    mean: np.ndarray  # Expected shape: (2,) for count and (4,) for ratio
```
lukesonnet: Could we instead use a named data class for the mean and the covariance rather than an ndarray? If it makes a lot of the later math annoying then maybe we can consider an entirely different approach.

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +326,963 @@ def scaled_impact_eligible(self) -> bool:
     @abstractmethod
     def compute_result(self) -> TestResult:
         pass
+
+
+@dataclass_with_arbitrary_types_allowed
```
lukesonnet: Why is this necessary?

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +326,963 @@ def scaled_impact_eligible(self) -> bool:
...
+        return StrataResult(
+            metric_type=self.metric_type,
+            n=self.n,
+            mean=self.mean,
+            covariance=self.covariance,
+        )
+
+
+# Regression version of Algorithm 1 for count metrics
+class CreateStrataResultRegressionAdjusted:
```
lukesonnet: Why does this not inherit from the parent class `BaseCreateStrataResult`?

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +326,963 @@ def scaled_impact_eligible(self) -> bool:
...
+
+    @property
+    def unadjusted_baseline_mean(self) -> float:
+        if self.mean[2] == 0:
+            return 0
+        else:
+            return self.mean[0] / self.mean[2]
+
+
+class PostStratification:
```
lukesonnet: Should this be named PostStratificationEffectMoments?

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +326,963 @@ def scaled_impact_eligible(self) -> bool:
...
+
+    @property
+    def covariance(self) -> np.ndarray:
+        return self.contrast_matrix.dot(self.strata_covariance).dot(
+            self.contrast_matrix.T
+        )
+
+
+# Algorithm 1 for count metrics
+class CreateStrataResult(BaseCreateStrataResult):
```
lukesonnet: What should the structure of these classes be and what is their naming convention?

Also, what is the deal with proportion metrics?

It's a bit unclear. But it feels like it should be

CreateStrataResultBase
CreateStrataResultCount(CreateStrataResultBase)
CreateStrataResultCountRegressionAdjusted(CreateStrataResultCount)
CreateStrataResultRatio(CreateStrataResultBase)
CreateStrataResultRatioRegressionAdjusted(CreateStrataResultCount)

Is that right? Feels right to me, and if that leads to a bit of duplication that might be alright since it's a bit cleaner of a split.

Then you shouldn't need to do things like type hinting like this `def metric_type(self) -> Literal["count", "ratio"]:`, you could instead just do `Literal["count"]`.

lukesonnet: In general, I think it's better to have one abstract parent class and two or 3 children classes than 1 abstract grand parent, 1 parent, and 1 child, where you use both the middle and the 3rd child class, unless it's a really small final child class.

But if this will result in a ton of duplication, then let me know and then we can consider a different organization.

packages/stats/gbstats/gbstats.py
```diff
@@ -142,6 +154,7 @@ def get_dimension_column_name(dimension: str) -> str:
         dimension_column_name = "dim_unit_" + dimension
 
     return dimension_column_name
+    # Overall columns
```
lukesonnet: is this a TODO?

packages/stats/gbstats/gbstats.py
```diff
@@ -150,9 +163,18 @@ def get_metric_df(
...
+        dimension_cols = dfc.filter(like="dim_exp_")
+        num_dimensions = len(dimension_cols.columns)
+        if num_dimensions == 1:
+            dfc = dfc.rename(columns={dimension_cols.columns[0]: "dimension"})
+        else:
+            dfc["dimension"] = dimension_cols.astype(str).agg("_".join, axis=1)
+
     # Each row in the raw SQL result is a dimension/variation combo
     # We want to end up with one row per dimension
     for row in dfc.itertuples(index=False):
```
lukesonnet, line 202: So this currently won't work to post-stratify on dimensions b and c if you want to analyze by dimension `a` because in that case the `dimension_column_name` won't be "dimension".

In fact the following lines don't need to be in the for loop, it should be before it.

```
dimension_column_name = (
    "dimension" if not dimension else get_dimension_column_name(dimension)
)
```
        
Can we combine the logic where we both set the right dimension_column_name and build the post-strat single column concatenated before we go into the for loop?

packages/stats/gbstats/gbstats.py
```diff
@@ -462,6 +588,8 @@ def analyze_row(s: pd.Series) -> pd.Series:
         )
         return s
 
+    df_2 = df.apply(analyze_row, axis=1)
+    df_2.to_csv(dir_desktop + "updated_df.csv")
```
lukesonnet: errant lines

packages/stats/gbstats/gbstats.py
```diff
@@ -351,103 +359,221 @@ def analyze_metric_df(
...
+# Run A/B test analysis for each variation and dimension
+def analyze_metric_df(
+    df: pd.DataFrame,
+    metric: MetricSettingsForStatsEngine,
+    analysis: AnalysisSettingsForStatsEngine,
+) -> pd.DataFrame:
+    num_variations = df.at[0, "variations"]
+    df = initialize_df(df, analysis)
+    dir_desktop = "/Users/lukesmith/Desktop/"
+    df.to_csv(dir_desktop + "initial_df.csv")
```
lukesonnet: additional errant lines

packages/stats/gbstats/gbstats.py
```diff
@@ -351,103 +359,221 @@ def analyze_metric_df(
...
+    return s
+
+
+def run_post_stratification(
+    df: pd.DataFrame,
+    metric: MetricSettingsForStatsEngine,
+    analysis: AnalysisSettingsForStatsEngine,
+) -> pd.DataFrame:
+
+    # need to add a check for quantile metrics
```
lukesonnet: Ideally we can kind of do this and check it against the config in as high of a place as possible.

packages/stats/gbstats/gbstats.py
```diff
@@ -351,103 +359,221 @@ def analyze_metric_df(
...
+            )
+            stats_variation.append(stat)
+            total_users_variation += s["total_users"]
+        total_users = total_users_control + total_users_variation
+        stats = list(zip(stats_control, stats_variation))
+        test = get_configured_test(
+            stats, total_users, analysis=analysis, metric=metric, post_stratify=True
+        )
+        res = test.compute_result()
+        s_output = store_test_results(test, res, s_output, variation)
```
lukesonnet: I don't love the way we keep overwriting the output like this. Can we instead use a set of dataclasses, fill them with each run (e.g. test results and mid exp power) and then turn them into a dataframe row if we need to?

packages/stats/gbstats/models/tests.py
lukesonnet:
```
@dataclass
class StrataResultMeanCount:
     control_mean: number
     delta: number

     @property
     def numpyArray(self):
         return np.array([self.control_mean, self.delta])

etc...
```

lukesonnet: dataclass StrataResult = Union[StrataResultCount, StrataResultRatio]

lukebrawleysmith: done

packages/stats/gbstats/models/tests.py
lukebrawleysmith: it was necessary before, to permit numpy attribute in the dataclass.  No longer needed and removed

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: Updated so we still use the isinstance and TypeError flow. 

If we have different `ProportionStatistics` across strata cells, we sum them to get the variation statistics inside of the `sum_stats` method.  We transform the `ProportionStatistics` to `SampleMeanStatistics` to get the correct second moment.  

Alternatively, I could transform all of the `ProportionStatistics` to `SampleMeanStatistics` inside of `sum_stats`.  I feel like it's better to make this explicit in `sum_stats`, though, rather than having to remember to make the transformation.

packages/stats/gbstats/models/tests.py
lukebrawleysmith: good call, and updated

packages/stats/gbstats/gbstats.py
lukebrawleysmith: removed

packages/stats/gbstats/gbstats.py
lukebrawleysmith: removed

packages/stats/gbstats/models/tests.py
lukebrawleysmith: This is helpful thanks.  I refactored so there is one base class `CreateStrataResultBase` with 4 subclasses (CreateStrataResultCount
CreateStrataResultCountRegressionAdjusted
CreateStrataResultRatio
CreateStrataResultRatioRegressionAdjusted).  
I removed the type hint.  
Binomial metrics will be treated as Count metrics, to ensure the second moment is right.

packages/stats/gbstats/models/tests.py
lukebrawleysmith: I refactored to inherit from `BaseCreateStratatResult`

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +345,1218 @@ def scaled_impact_eligible(self) -> bool:
...
+        stat_a: RegressionAdjustedStatistic,
+        stat_b: RegressionAdjustedStatistic,
+        theta: Optional[float] = None,
+    ):
+        self.stat_a = stat_a
+        self.stat_b = stat_b
+        self.theta = theta
+
+    @property
+    def n(self) -> int:
```
lukesonnet: This is identical to the parent class. Is there a reason it is repeated like this? Are there other methods like this that are repeated?

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +345,1218 @@ def scaled_impact_eligible(self) -> bool:
...
+    def compute_result(self) -> StrataResultCount:
+        if self._has_zero_variance():
+            return self._default_output(error_message=ZERO_NEGATIVE_VARIANCE_MESSAGE)
+        regression_coefs = np.array([self.stat_b.mean, self.stat_a.mean])
+        regression_coefs_covariance = (
+            CreateStrataResult.compute_regression_coefs_covariance(
+                self.len_alpha, self.n, self.n_a, self.n_b, self.lambda_a, self.lambda_b
+            )
+        )
+        mean = CreateStrataResult.mean(self.contrast_matrix, regression_coefs)
```
lukesonnet: Tiny nit: Is there a particular reason why you use static methods for the mean but a non-static method for the covariance? I would prefer consistency here.

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +345,1218 @@ def scaled_impact_eligible(self) -> bool:
...
+class CreateStrataResultRegressionAdjusted(CreateStrataResultBase):
+    def __init__(
+        self,
+        stat_a: RegressionAdjustedStatistic,
+        stat_b: RegressionAdjustedStatistic,
+        theta: Optional[float] = None,
+    ):
+        self.stat_a = stat_a
+        self.stat_b = stat_b
+        self.theta = theta
```
lukesonnet: What is this `theta` used for? I don't see it used anywhere.

packages/stats/gbstats/gbstats.py
lukesonnet: This issue still seems unresolved.

packages/stats/gbstats/gbstats.py
lukebrawleysmith: resolved by Sonnet's gbstats refactor

packages/stats/gbstats/gbstats.py
lukebrawleysmith: resolved by Sonnet's gbstats refactor

packages/stats/gbstats/models/statistics.py
```diff
@@ -494,16 +510,20 @@ def compute_theta_regression_adjusted_ratio(
     a: RegressionAdjustedRatioStatistic, b: RegressionAdjustedRatioStatistic
 ) -> float:
     # set theta equal to 1, so the partial derivatives are unaffected by theta
-    a_one = copy.deepcopy(a)
-    b_one = copy.deepcopy(b)
-    a_one.theta = 1
-    b_one.theta = 1
+    a_one = RegressionAdjustedRatioStatistic(
+        **{k: v for k, v in asdict(a).items() if k != "theta"},
```
lukesonnet: Please use `replace(a, theta = 1)` using the `dataclasses.replace` method instead of this dict comprehension. Here and the other places we do this in `tests.py`.

packages/stats/gbstats/models/tests.py
```diff
@@ -291,15 +271,79 @@ def __init__(
...
         if self.config.post_stratify:
-            raise NotImplementedError("Post-stratification not implemented")
+            is_summable_list = all(
+                isinstance(stat_a, SummableStatistic)
+                and isinstance(stat_b, SummableStatistic)
+                for stat_a, stat_b in self.stats
+            )
+            if not is_summable_list:
+                raise ValueError("Post-stratification requires summable statistics.")
+            summable_stats = cast(
```
lukesonnet: I'm not a huge fan of using `cast` here just to make it clear to the type checker we are doing the right thing. I'd rather build the list from scratch and validate along the way:

```
            summable_stats: List[Tuple[SummableStatistic, SummableStatistic]] = []
            for stat_a, stat_b in self.stats:
                if isinstance(stat_a, SummableStatistic) and isinstance(stat_b, SummableStatistic):
                    summable_stats.append((stat_a, stat_b))
                else:
                    raise ValueError("Post-stratification requires summable statistics.")
```

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +359,1218 @@ def scaled_impact_eligible(self) -> bool:
...
+        return (
+            self.stat_a.pre_statistic.variance + self.stat_b.pre_statistic.variance <= 0
+        )
+
+    def compute_result(self) -> StrataResultCount:
+        if self._has_zero_variance():
+            return CreateStrataResult._default_output(
+                error_message=ZERO_NEGATIVE_VARIANCE_MESSAGE
+            )
+        xtx_inv_result = invert_symmetric_matrix(self.xtx)
```
lukesonnet, line 673: Inverting matrices is always risky to me in general. Are we aware of the situations when this would fail? Are there reasonable fallbacks to non-poststratified results in this case?

Should we, if post-stratify fails for any reason, try to run regular analysis, and return the regular analysis should the post strat analysis fail?

packages/stats/gbstats/utils.py
```diff
@@ -101,3 +103,72 @@ def isinstance_union(obj, union):
...
+    """
+    Represents the result of a symmetric matrix inversion operation.
+    """
+
+    success: bool
+    inverse: Optional[np.ndarray] = None
+    error: Optional[str] = None
+
+
+def invert_symmetric_matrix(v: np.ndarray) -> MatrixInversionResult:
```
lukesonnet, line 133: What is the biggest matrix we can invert? Is that size a problem ever?

packages/stats/gbstats/utils.py
```diff
@@ -101,3 +103,72 @@ def isinstance_union(obj, union):
...
+    """
+    n = v.shape[0]
+    if v.shape[1] != n:
+        return MatrixInversionResult(
+            success=False, error="Input matrix must be square."
+        )
+
+    try:
+        # Compute the Cholesky factorization of v
+        v_cholesky = la.cholesky(v, lower=True, check_finite=True)
```
lukesonnet, line 153: Why are we doing the cholesky decomposition directly instead of just using `la.inv`?

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +359,1218 @@ def scaled_impact_eligible(self) -> bool:
...
+        if self.stat_a.unadjusted_mean == 0:
+            return self._default_output(BASELINE_VARIATION_ZERO_MESSAGE)
+
+        # if any cells have 0 users in a variation, add that cell to the cell with the largest number of users
+        cells_for_analysis = self.create_cells_for_analysis(self.stats)
+        # if there is only one strata cell, run the regular effect moments test
+        if len(cells_for_analysis) == 1:
+            return EffectMoments(cells_for_analysis, EffectMomentsConfig(difference_type="relative" if self.relative else "absolute")).compute_result()  # type: ignore
+        strata_results = []
+        for _, cell in enumerate(cells_for_analysis):
```
lukesonnet: No need for enumerate?

packages/stats/gbstats/models/tests.py
```diff
@@ -315,3 +359,1218 @@ def scaled_impact_eligible(self) -> bool:
...
+            summed_cells = sum_stats(cells_for_analysis)
+            return cast(List[Tuple[SummableStatistic, SummableStatistic]], summed_cells)
+        return cells_for_analysis
+
+    def compute_result(self) -> EffectMomentsResult:
+        if self._has_zero_variance():
+            return self._default_output(ZERO_NEGATIVE_VARIANCE_MESSAGE)
+        if self.stat_a.mean == 0:
+            return self._default_output(BASELINE_VARIATION_ZERO_MESSAGE)
+        if self.stat_a.unadjusted_mean == 0:
```
lukesonnet: Aren't these checks weird since we haven't summed the stats yet? It doesn't feel like we should ever build `self.stat_a` since we only ever use it for these checks, and they don't seem to be the right checks we need to make.

packages/stats/gbstats/models/tests.py
lukebrawleysmith: Good catch - I removed this.  
I checked that `PostStratificationSummaryRatio` does not have copied methods from `PostStratificationSummary` and I checked that the methods in `CreateStrataResultBase` are not copied in any of its 4 child classes.

packages/stats/gbstats/models/tests.py
lukebrawleysmith: The `covariance_unadjusted` method is used in both `CreateStrataResult` and `CreateStrataResultRatio`, so keeping this method static reuses code.  However, I did change `CreateStrataResult.mean` to `self.mean` where appropriate.

packages/stats/gbstats/models/tests.py
lukebrawleysmith: Good catch, I think I used it before, when `theta` was calculated in the Test class.   I removed it.

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: done

packages/stats/gbstats/gbstats.py
lukebrawleysmith: @lukesonnet , just want to confirm, this is resolved, right?  run_mid_experiment_power returns a PowerResponse object.  res is either a Bayesian or Frequentist Test result.  

We still overwrite in these lines, but this is a preprocessing step. 
if metric.statistic_type in ["quantile_event", "quantile_unit"]:
    d[f"v{i}_count"] = d[f"v{i}_quantile_n"]

packages/stats/gbstats/models/tests.py
lukebrawleysmith: done

packages/stats/gbstats/models/tests.py
lukebrawleysmith: Great questions.  
>Are we aware of the situations when this would fail?  Are there reasonable fallbacks to non-poststratified results in this case?
Matrix inversion would fail if the sample size in either variation is 0 or if the pre-statistic variance across both variations is zero.  
If the sample size is too small in a cell, we combine data from that cell into the largest cell. 
The unit test `test_missing_variation_data` ensures this works. 

If the pre-statistic variance is 0 for either variant across all cells, then we fallback from CUPED to unadjusted inference.  
For ratio, I had a default in place where we fallback to simple unadjusted difference in ratios if there is no baseline variance:
Ratio case: https://github.com/growthbook/growthbook/pull/4202/files#diff-d92e03fff80b640ec096d27454b16c7abcf44f34f8d2423e278c5fdd96d15cceR1053
I mistakenly deleted the mean metric case, and it is now included. 
The unit tests `test_post_strat_count_effect_moments_fallback` and `test_post_strat_ratio_effect_moments_fallback` ensure this works. 

If the pre-statistic variance is 0 for either variant for a single cell, then we fallback from CUPED to unadjusted inference for that cell.  
The unit tests `test_post_strat_mean_reg_gbstats_single_cell_fallback` and `test_post_strat_ratio_reg_gbstats_single_cell_fallback`ensure this works. 

>Should we, if post-stratify fails for any reason, try to run regular analysis, and return the regular analysis should the post strat analysis fail?
I added logic inside of the `compute_moments_result` method of `BaseABTest` where if  `EffectMomentsPostStratification` fails for any reason, we try `EffectMoments`.

packages/stats/gbstats/utils.py
lukebrawleysmith: la.inv permits inversion of any non-singular matrix.  Cholesky exploits the symmetry of the matrix, and is faster.

packages/stats/gbstats/models/tests.py
lukebrawleysmith: embarrassing, fixed

packages/stats/gbstats/models/tests.py
lukebrawleysmith: accepted and implemented change

packages/stats/gbstats/utils.py
lukebrawleysmith: Sorry, I thought that I answered this before.  The biggest matrix this function could invert would be bigger than 1k rows.  All the matrices post-strat works with have no more than 8 rows.

PR #4205, opened 2025-06-19
fix(CodeEditor): Improve editor bundling + fix worker

ahdriel:
### Features and Changes

Some languages used by Ace uses a Web Worker and it attempts to load `worker-json.js` on the relative URL of the page by default, which does not work.

So we fix that issue by configuring Next to load worker files from ace-build as a static asset and using the absolute URL for the worker.

Additionally we are now bundling the whole editor in a single bundle and avoiding a waterfall of requests as we did before.
A potential further optimization would be to load just what's needed for JSON if that's the language we are using, but as this code does not change often this will be a one-time cost for users.

### Testing

Open the Settings > Webhooks page with Cleared Cache
Click on `New Event webhook`
Observe how the `ace-editor` is now the only bundle being loaded
Observe how there's no error around `worker-json.js` anymore
Observe how the editor still works as expected and works on both light & dark mode


jdorn (commented):
Code looks ok, but will definitely want to test this pretty extensively with a production build and within Docker.

ahdriel:
@jdorn I have tested it with production and Docker builds and it worked as expected and showed the same behavior as the dev environment.

Is there any specific test you want me to do? I opened different editors and validated that the experience is the same as the current one for users.

PR #4210, opened 2025-06-20
Experimentation left navigation re-organization

Auz:
Moved namespaces to experimentation section. Added a page for the power calculator and the experiment templates.


packages/front-end/pages/experiments/index.tsx
```diff
@@ -76,21 +58,16 @@ const ExperimentsPage = (): React.ReactElement => {
 
   const { templates } = useTemplates();
 
-  const [tabs, setTabs] = useLocalStorage<string[]>("experiment_tabs", []);
+  const [tab, setTab] = useState<string>("all");
```
jdorn, line 50: Do we want to do `useLocalStorage` here so it remembers the last tab the user had selected?

packages/front-end/pages/experiments/index.tsx
```diff
@@ -414,224 +262,71 @@ const ExperimentsPage = (): React.ReactElement => {
...
+                      </TabsContent>
+                    );
+                  })}
+                  {canAddTemplate &&
+                  !templates.length &&
+                  allExperiments.length >= 5 ? (
+                    <div className="row justify-content-center m-3">
+                      <ExperimentTemplatePromoCard
+                        hasFeature={hasTemplatesFeature}
+                        onClick={() => setOpenTemplateModal({})}
```
jdorn: Do we still want it to do this? Or should this just link to the new templates page?

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -45,17 +44,44 @@ const navlinks: SidebarLinkProps[] = [
     path: /^(features)/,
   },
   {
-    name: "Experiments",
+    name: "Experimentation",
     href: "/experiments",
-    path: /^experiment/,
+    path: /^(experiments$|experiment\/|bandit|namespaces)/,
```
jdorn: Does this match the `experiments/templates` route?  Also, power-calculator is missing.

packages/front-end/pages/experiments/index.tsx
Auz: I thought about that - what I'm concerned with is that you click on experiments expecting to see all experiments, but its on your last filter, which requires you to see that the old tab is selected. Open to be wrong, but I feel this is more expected behaviour.

packages/front-end/pages/experiments/index.tsx
Auz: Fixed.

packages/front-end/components/Layout/Layout.tsx
Auz: fixed

jdorn (approved):

packages/front-end/pages/experiments/index.tsx
tzjames: https://growthbookapp.slack.com/archives/C02BHUS8NBE/p1754389045626759 someone only cares about running experiments and is annoyed about having to constantly go back to that tab.   I'd have to agree that I think that is the more common use case.  People who can't find an experiment will hopefully see they are on "running experiments" or some other tab.

PR #4212, opened 2025-06-21
Improvements for Managed ClickHouse

jdorn:
### Features and Changes
- [x] Renamed `context_json` to `attributes` and `properties_json` to `properties`.  This is just done in the materialized view, but we can/should eventually do this in the ingestor instead.
- [x] Renamed "Materialized Columns" to "Key Attributes" in the UI to better reflect the purpose.
- [x] Added a `type` field to Materialized Columns to specify an attribute as an `Identifier` or `Dimension` (or Other).  This is used to auto-generate identifier types and experiment assignment queries.
- [x] Greatly simplified the base materialized view structure.  There's now a minimal set of top-level fields that are always present (`timestamp`, `client_key`, `event_name`, `properties`, `attributes`, `environment`, `sdk_language`, `sdk_version`, `event_uuid`, and `ip`).  Everything else is now managed through Materialized Columns.
- [x] Hid most sections on the data source detail page - identifiers, identifier joins, experiment assignment queries, jupyter notebook settings.  These are all now either not applicable or handled automatically by materialized columns.
- [x] Remove the Page Views fact table that's created by default.  Doesn't provide a lot of value and is too opinionated on event names / properties.
- [x] Add event tracking instructions to the data source detail page with info for different SDKs/languages.
- [ ] Switch materialized view to use real JSON columns instead of a String for properties and attributes
- [ ] Create a separate materialized view for Experiment Viewed events. Keep up-to-date with materialized column changes.
- [ ] When materialized column modal opens, refresh fact table columns and update select options.
- [ ] Block deleting the main events fact table or editing the SQL (unless the data source was deleted and the fact table is orphaned now)
- [ ] Hide materialized views from schema browser
- [ ] Update fact metric column selectors to remove duplicates and prefer materialized columns vs the nested JSON version


packages/front-end/components/Settings/EditDataSource/ClickhouseMaterializedColumns/AddEditMaterializedColumnsModal.tsx
```diff
@@ -182,31 +168,75 @@ export default function AddMaterializedColumnsModal({
...
+                (device_id, user_id, etc). <strong>Dimensions</strong> can be
+                used to slice and dice experiment results (account_type,
+                country, etc).
+              </>
+            }
+          />
           <Field
             label="Column Name"
-            helpText="This named column will be available in metric queries. Must start with a letter or underscore and use only alphanumeric characters and '_'"
+            helpText="This SQL column the attribute will be stored in. Must start with a letter or underscore and use only alphanumeric characters and '_'"
```
Kevin-Chant: Nit
```suggestion
            helpText="The SQL column the attribute will be stored in. Must start with a letter or underscore and use only alphanumeric characters and '_'"
```

Kevin-Chant (commented):
Only one nit from reading through the diff. I'll play with it and see if I can break anything

packages/back-end/src/controllers/datasources.ts
```diff
@@ -293,98 +290,62 @@ export async function postManagedClickHouse(
...
     "growthbook_clickhouse",
     params,
     datasourceSettings,
-    new_datasource_id
+    "managed_clickhouse"
   );
   res.status(200).json({
     status: 200,
-    id: new_datasource_id,
+    id: "managed_clickhouse",
```
tzjames: We had talked in the past about one day allowing them to have multiple managed_clickhouse datasources.  You don't think it is likely we will provide that in the future now?

packages/back-end/src/controllers/datasources.ts
```diff
@@ -944,15 +910,25 @@ export async function postMaterializedColumn(
...
-      ...datasource.settings,
-      materializedColumns: finalColumns,
-    },
+
+  // Make sure it doesn't exist already
+  if (
+    originalColumns.some(
+      (col) =>
+        col.columnName === newColumn.columnName ||
+        col.sourceField === newColumn.sourceField
```
tzjames, line 925: Are you sure you want the sourceField check?  Sometimes when renaming a column they might want two columns with the same source field, that way queries don't break.  Then they update their queries.  Then delete the old column.  That way there is no downtime.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -1034,47 +1004,58 @@ export async function updateMaterializedColumn(
...
-    settings: {
-      ...datasource.settings,
-      materializedColumns: finalColumns,
-    },
+  // Make sure there is only 1 column with this source field or name
+  if (
+    finalColumns.filter(
+      (col) =>
+        col.sourceField === newColumn.sourceField ||
+        col.columnName === newColumn.columnName
```
tzjames, line 1030: same here

packages/back-end/src/services/clickhouse.ts
```diff
@@ -38,30 +45,32 @@ const REMAINING_COLUMNS_SCHEMA: Record<string, ClickHouseDataType> = {
...
   geo_city: "String",
   geo_lat: "Float64",
   geo_lon: "Float64",
   ua: "String",
-  ua_browser: "LowCardinality(String)",
-  ua_os: "LowCardinality(String)",
-  ua_device_type: "LowCardinality(String)",
+  ua_browser: "String",
+  ua_os: "String",
+  ua_device_type: "String",
```
tzjames, line 61: These are not low cardinality enough?

packages/back-end/src/services/clickhouse.ts
```diff
@@ -274,24 +405,86 @@ AS ${featureUsageMaterializedViewSql}`,
...
-    query: `DROP DATABASE ${user}`,
-  });
+  // Backfilling data can take a while, so lock the datasource for 30 minutes
+  await lockDataSource(context, datasource, 1800);
 
-  logger.info(`Clickhouse user ${user} deleted`);
+  try {
+    // Drop the entire database and recreate it
+    logger.info(`Dropping Clickhouse database ${database}`);
+    await runCommand(client, `DROP DATABASE ${database} IF EXISTS`);
```
tzjames: `IF EXISTS` needs to come before the database name.

packages/front-end/pages/datasources/[did].tsx
```diff
@@ -318,52 +330,283 @@ mixpanel.init('YOUR PROJECT TOKEN', {
...
-              <DataSourceInlineEditIdentifierTypes
-                onSave={updateDataSourceSettings}
-                onCancel={() => undefined}
-                dataSource={d}
-                canEdit={canUpdateDataSourceSettings}
-              />
-            </Frame>
+            {isManagedClickHouse ? (
+              <>
+                {superAdmin ? (
```
Comment (tzjames) [inline packages/front-end/pages/datasources/[did].tsx]: Should this be on the /admin page instead which is where we have the create managed clickhouse button.

packages/back-end/src/controllers/datasources.ts
jdorn: If we want to do this in the future, we can just switch back to generated ids.  I think this will be nice to have for now.

packages/back-end/src/controllers/datasources.ts
jdorn: I think I'd prefer to wait until someone asks for this and then think through a proper solution for their use case.  If it's just about renaming a column, a proper solution might be to just update all the metrics that reference the old column automatically.

packages/front-end/pages/datasources/[did].tsx
Comment (jdorn) [reply packages/front-end/pages/datasources/[did].tsx]: Yeah, I'll move it

packages/back-end/src/services/clickhouse.ts
jdorn: I moved all of these fields to be managed through Materialized Columns so I needed the types here to match those types  and we don't have a low cardinality option there.

ClickHouse has this warning for LowCardinality:
> If a dictionary contains more than 100,000 distinct values, then ClickHouse can perform worse in comparison with using ordinary data types.

The performance improvements from users doing things correctly probably outweighs the risk of them picking a high cardinality field.  I just don't know for sure.

PR #4213, opened 2025-06-23
Gl/homepage revamp

gazzdingo:
### Features and Changes


packages/back-end/src/models/AuditModel.ts
```diff
@@ -88,7 +134,7 @@ export async function findAuditByEntity(
       "entity.id": id,
     },
     options
-  );
+  ).limit(100);
```
gazzdingo: remove this

packages/front-end/pages/needing-attention.tsx
```diff
@@ -0,0 +1,573 @@
...
+
+    const recentFeatures = Object.entries(recentlyUsed)
+      .filter(([_, id]) => id !== null)
+      .map(([key, id]) => {
+        let label = key
+          .replace("Id", "")
+          .replace(/([A-Z])/g, " $1")
+          .trim()
+          .replace(/^./, (str) => str.toUpperCase());
+        //find the feature or experiment by id
```
gazzdingo: remove this

packages/back-end/src/models/AuditModel.ts
```diff
@@ -46,10 +45,34 @@ const AuditModel = mongoose.model<AuditInterface>("Audit", auditSchema);
...
+      ? {
+          object: json.parent.object,
+          id: json.parent.id,
+        }
+      : undefined,
+    reason: json.reason,
+    details: json.details,
+    dateCreated: json.dateCreated,
+  };
+  return transformed as AuditInterface;
```
msamper: delete?

packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
```diff
@@ -259,3 +260,22 @@ export const getSafeRolloutTimeSeries = async (
...
 };
 // endregion GET /safe-rollout/:id/time-series
+
+// region GET /safe-rollout
+/**
+ * GET /safe-rollout
+ * Get all safe rollout rules
+ */
+export const getSafeRollouts = async (
+  req: AuthRequest<null, { environment: string }>,
```
msamper: delete environment param. seems like its not needed

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -32,9 +27,9 @@ import { usePageHead } from "./PageHead";
 
 const navlinks: SidebarLinkProps[] = [
   {
-    name: "Get Started",
+    name: "Home",
     href: "/getstarted",
```
msamper: should we change the path too if it's no longer called get started?

packages/front-end/pages/getstarted.tsx
```diff
@@ -69,6 +92,60 @@ const GetStartedPage = (): React.ReactElement => {
...
+                    style={{ color: "white" }}
+                  >
+                    {feature.title}
+                  </Text>
+                  <Text size="3" style={{ color: "white" }}>
+                    {feature.description}
+                  </Text>
+                </Flex>
+              </Card>
+            </Link>
```
msamper: Use the `<OverviewCard>` component we use in the Product Overview section below for consistency. You'll probably just have to add a prop for overlay text

packages/front-end/pages/needing-attention.tsx
```diff
@@ -0,0 +1,573 @@
...
+  ExperimentStatusDetailsWithDot,
+} from "@/components/Experiment/TabbedPage/ExperimentStatusIndicator";
+import UserAvatar from "@/components/Avatar/UserAvatar";
+
+type FeaturesAndRevisions = FeatureRevisionInterface & {
+  feature: FeatureInterface;
+  safeRollout: SafeRolloutInterface | undefined;
+};
+
+const NeedingAttentionPage = (): React.ReactElement | null => {
```
msamper: It seems like this is a component rather than a page since its used on the home page. I think we probably don't need routing to this so we should probably move this into `components/GetStarted` and rename the component

packages/front-end/pages/needing-attention.tsx
```diff
@@ -0,0 +1,573 @@
...
+          source="get-started"
+          commercialFeature={null}
+        />
+      )}
+      {showVideoId && (
+        <YouTubeLightBox
+          close={() => setShowVideoId("")}
+          videoId={showVideoId}
+        />
+      )}
```
msamper: what is this for? it seems like `showVideoId` is always an empty string

packages/front-end/pages/needing-attention.tsx
```diff
@@ -0,0 +1,573 @@
...
+  const orgIsUsingFeatureAndExperiment = hasFeatures && hasExperiments;
+  return !orgIsUsingFeatureAndExperiment ? null : (
+    <>
+      {upgradeModal && (
+        <UpgradeModal
+          close={() => setUpgradeModal(false)}
+          source="get-started"
+          commercialFeature={null}
+        />
+      )}
```
msamper: Do we need this? `upgradeModal` seems to always be false and I don't think theres a premium feature within this component

msamper (commented):
Left a few comments. There's some code within the `NeedingAttention` component that seems unnecessary from what I could tell so you can probably delete it. I'm going to run through the UI and see if I spot anything that looks off

packages/front-end/pages/needing-attention.tsx
gazzdingo: good point it started out as a page then i changed it halfway through :)

packages/front-end/pages/needing-attention.tsx
gazzdingo: delete that is leftover from when it was its own page

packages/front-end/pages/needing-attention.tsx
gazzdingo: same as above leftover

packages/front-end/pages/getstarted.tsx
gazzdingo: feels like i would be hacking that componet to get it to do what i want for it. not sure if it is worth it

msamper:
Screenshot 2025-06-27 at 10 58 07 AM
Just took a look at the UI and it looks like the advanced feature cards are too large and they also don't have a big enough margin so they take up more horizontal screen real estate than the rest of the content on the page. They should be the same size as the `OverviewCard`. The Featured docs section also seems to have been pushed down and it should still be on the top of the page like it was before. 
There's also a few things I noticed were missing from the UI that I see on the design. It looks like we want a new `Create` button to act as a shortcut for creating new features or experiments. I also see that the Get Started section should be collapsible now and the `DocumentationSidebar` should now have a Premium Features section

packages/front-end/pages/getstarted.tsx
```diff
@@ -60,6 +61,28 @@ const GetStartedPage = (): React.ReactElement => {
   // Also used for the `Launch Setup Flow` button to keep it aligned
   const DOCUMENTATION_SIDEBAR_WIDTH = "minmax(0, 245px)";
 
+  // Advanced Features Cards Section
+  const advancedFeatures = [
```
msamper: I think these should rotate based on product usage from what I see on the design, but not sure if we're just not doing this for the first version

packages/front-end/pages/index.tsx
```diff
@@ -63,5 +62,5 @@ export default function Home(): React.ReactElement {
     );
   }
 
-  return <LoadingOverlay />;
+  return <GetStartedAndHomePage />;
```
gazzdingo, line 69: @jdorn do we want this to still load until the redirects are checked?

gazzdingo: might reduce a flicker on slower connections

packages/front-end/components/GetStarted/index.tsx
```diff
@@ -0,0 +1,343 @@
...
+                  Feature Flag
+                </DropdownMenuItem>
+                <DropdownMenuItem
+                  onClick={() => {
+                    open("experiments");
+                  }}
+                >
+                  Experiment
+                </DropdownMenuItem>
+                {/* <DropdownMenuItem
```
gazzdingo, line 215: @jdorn let me know if you want all these in here they are not in the design so i comment them out but i feel like it is handy for quick links

gazzdingo:
Not sure if we want the advanced Features to auto rotate. Or if we should use that to highlight new released features in the future

packages/front-end/components/GetStarted/NeedingAttention.tsx
```diff
@@ -0,0 +1,578 @@
...
+  ExperimentDot,
+  ExperimentStatusDetailsWithDot,
+} from "@/components/Experiment/TabbedPage/ExperimentStatusIndicator";
+import UserAvatar from "@/components/Avatar/UserAvatar";
+type FeaturesAndRevisions = FeatureRevisionInterface & {
+  feature: FeatureInterface;
+  safeRollout: SafeRolloutInterface | undefined;
+};
+
+const NeedingAttentionPage = (): React.ReactElement | null => {
```
msamper: `NeedingAttentionSection`?

packages/front-end/components/GetStarted/NeedingAttention.tsx
```diff
@@ -0,0 +1,578 @@
...
+        e.statusIndicator?.detailedStatus &&
+        e.statusIndicator?.detailedStatus === "No data" &&
+        !withinFirstDay;
+
+      return (
+        e?.statusIndicator?.detailedStatus &&
+        (e.statusIndicator?.detailedStatus === "unhealthy" ||
+          e.statusIndicator?.detailedStatus === "warning" ||
+          (hasNoData && isRunning && !isArchived))
+      );
```
msamper: I'm looking at `getDetailedRunningStatusIndicatorData` which seems to generate the `detailedStatus` strings you're looking at here and we may want to include `Roll back now` and `Ship now`. I also think `unhealthy` should be `Unhealthy` and I don't see `warning` as a possible `detailedStatus`. Right now `detailedStatus` on `StatusIndicatorData` is typed as a string, but if we're going to match it in this way for this filtering logic, I think we should consider using string literal union types to make it clear what the possible detailed statuses are

packages/front-end/components/GetStarted/index.tsx
```diff
@@ -0,0 +1,414 @@
...
+        href: "https://docs.growthbook.io/app/data-pipeline",
+        commercialFeature: "pipeline-mode",
+      },
+      {
+        imgUrl: "/images/get-started/advanced/fact-tables.png",
+        title: "Fact Tables",
+        description: "Enable automatic query optimization",
+        href: "/fact-tables",
+      },
+    ];
```
msamper: should we pull this array out of the component since it's a constant?

packages/front-end/components/GetStarted/index.tsx
```diff
@@ -0,0 +1,414 @@
...
+                href="/setup"
+                className="font-weight-bold"
+                style={{ color: "inherit" }}
+              >
+                Launch the setup flow
+              </Link>{" "}
+              <PiArrowSquareOut />
+            </Callout>
+          </Grid>
+        )}
```
msamper: Looking at the design, it looks like this is in the wrong place. We want this to show up inside of the get started section when it is expanded. Right now its rendering above the `Home` header
Screenshot 2025-07-01 at 1 31 17 PM

packages/front-end/components/GetStarted/index.tsx
```diff
@@ -0,0 +1,414 @@
...
+                </Box>
+              )}
+              <Flex
+                direction="row"
+                justify="between"
+                mt={orgIsUsingFeatureOrExperiment ? "7" : undefined}
+              >
+                {orgIsUsingFeatureOrExperiment && (
+                  <Text size="4" weight="medium" mb="3" as="div">
+                    Getting Started
```
msamper: nit: I think the title is `Get Started`

packages/front-end/pages/getstarted.tsx
```diff
@@ -1,210 +1,5 @@
...
-          </Box>
-        </Grid>
-      </Container>
-    </>
-  );
-};
-
-export default GetStartedPage;
+export default function GettingStarted() {
+  return <GetStartedAndHomePage />;
```
msamper, line 4: Maybe just keep the code from `<GetStartedAndHomePage />` here? doesn't seem to be much benefit to having it in a separate component if this page file is just 5 lines

gazzdingo: i like it so that we can use it in other places like the index and this. in the future if we wanted to add extra props to show things or not it will make it way easier

msamper (approved):
Changes look good!

Issue #4207, opened 2025-06-19
[Bug] API: scheduleRules MUST always contain "enabled: false"


spanishkangaroo:
### Summary

The "scheduleRules" body part on a feature must always contain the `"enabled: false"` section or the UI will fail to render that rule feature.

#### Expected Behavior

Either enforce the body to always contain the "enabled: false" (I guess, modify the OpenAPI definition accordingly) or set it if empty.

#### Current Behavior

For example, if this body is sent in API operation "updateFeature", the API returns OK, but then the UI fails to parse and then render the rule:

```
"scheduleRules": [
    {
        "enabled": true,
        "timestamp": "2025-06-25T18:00:00Z"
    }
],
```

ahdriel:
Fixed in [#4214](https://github.com/growthbook/growthbook/pull/4214)

Issue #4206, opened 2025-06-19
[Bug] API: scheduleRules correctly parse timezone that cannot be shown afterwards


spanishkangaroo:
### Summary

The "scheduleRules" body on a feature can contain a date time expressed with timezone format, but then, the UI fails to parse it.

#### Expected Behavior

DO NOT accept datetimes with that format, just the +XX:YY format, for example. Or add logic to parse it OK in the UI.

#### Current Behavior

For example, if "2025-06-13T11:55:00.000CET" is sent inside "scheduleRules" in API operation "updateFeature", the API returns OK, but then the UI fails to parse and then render the date.

ahdriel:
Fixed in [#4214](https://github.com/growthbook/growthbook/pull/4214)

PR #4214, opened 2025-06-23
Improve validation of scheduleRules

mknowlton89:
### Features and Changes

Our validation and documentation around `scheduleRules` for creating a Feature and updating a Feature needed some improvement.

Our scheduling requires that the scheduleRules array contains exactly two rules. Likewise, each rule index has a `timestamp` column that we need to be in an ISO date format. There is additional logic that we also weren't validating. Things like: each index has to have a different value for the `enabled` property, and both indexes can't have a null timestamp.

In this PR, I built a `validateScheduleRules` function to handle all of this custom validation. I also added plan validation to the `postFeature` and `updateFeature` endpoints so that `Starter` orgs can't create feature schedules.

- Closes **(https://github.com/growthbook/growthbook/issues/4206)**
- Closes **(https://github.com/growthbook/growthbook/issues/4207)**

### Dependencies

- None

### Testing

**postFeature**
- [x] Ensure that only Pro/Ent orgs can create a feature with scheduleRules
- [x] Ensure that the schedule rules must contain exactly 2 rules
- [x] Ensure that the schedule rules have opposite `enabled` values (e.g. they can't both be `disabled` or both be `enabled`
- [x] Ensure that the schedule rules don't both have `null` timestamps
- [x] Ensure that the timestamps passed in are valid ISO timestamps - e.g. `"2025-06-13T11:55:00.000CET"` fails, but `"2025-06-27T16:09:37.769Z"` doesn't

**updateFeature**
- [x] Ensure that only Pro/Ent orgs can update a feature with scheduleRules in the payload
- [x] Ensure that the schedule rules must contain exactly 2 rules
- [x] Ensure that the schedule rules have opposite `enabled` values (e.g. they can't both be `disabled` or both be `enabled`
- [x] Ensure that the schedule rules don't both have `null` timestamps
- [x] Ensure that the timestamps passed in are valid ISO timestamps - e.g. `"2025-06-13T11:55:00.000CET"` fails, but `"2025-06-27T16:09:37.769Z"` doesn't


packages/shared/src/util/features.ts
```diff
@@ -196,6 +197,70 @@ export function validateFeatureValue(
...
+  try {
+    const date = new Date(timestamp);
+    return (
+      !isNaN(date.getTime()) &&
+      timestamp.endsWith("Z") &&
+      timestamp.length >= 20
+    );
+  } catch {
+    return false;
+  }
```
Kevin-Chant: Rather than doing regex and string matching directly, can we just check that the date object returns the same string?
```
!isNaN(date.getTime()) && date.toISOString() === timestamp
```

mknowlton89: We can - just tested that out and it works as expected. I think it's a good improvement as well, thanks for the suggestion.

PR #4223, opened 2025-06-25
SQL Explorer Detail Page

jdorn:
### Features and Changes

Add a dedicated detail page for every saved query plus a way to refresh results without opening up the modal.

image

### DIFF
--- a/packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
+++ b/packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
@@ -8,6 +8,7 @@ import {
   SavedQueryUpdateProps,
 } from "back-end/src/validators/saved-queries";
 import { orgHasPremiumFeature } from "back-end/src/enterprise";
+import { runFreeFormQuery } from "back-end/src/services/datasource";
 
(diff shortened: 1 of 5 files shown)


mknowlton89 (approved):
Code looks great. I especially like the `Refresh` pattern that doesn't return the refreshed results, just updates them in the db, and then the front-end seamlessly refetches the updates. Really cool.

Some UX feedback that we can either address in this PR or in follow ups:
- I think the id page could benefit from a `More Menu` next to the `Edit` button, so users can easily duplicate/delete a saved query. This can bring it in alignment with other detail pages we have.
- It might make sense to display the connected Data Source somewhere on the detail page. It feels a bit odd that the index page has more info that the detail page.
- User's might also benefit from having a `Visualization` empty state on the detail page to increase awareness/adoption. I think its possible some users might not see the `Visualization` tab in the SQL Explorer modal.
- In that same vein, it might make sense to have a dedicated `Create Visualization` cta somewhere on the Data Source page. It might be a bit more natural than just a single `Edit` cta - as I'd click that to edit the SQL, but maybe not to add/edit a visualization.

None of these are blockers, just quality of life improvements after playing with the page a bit.

PR #4227, opened 2025-06-26
REST api endpoint for reverting a feature to a specific revision

jdorn:
### Features and Changes

New REST endpoint for reverting a feature to a specific revision. Acts exactly the same as the "revert" action in the front-end.

### DIFF
--- a/packages/back-end/src/api/features/revertFeature.ts
+++ b/packages/back-end/src/api/features/revertFeature.ts
@@ -1,0 +1,131 @@
+import { filterEnvironmentsByFeature, MergeResultChanges } from "shared/util";
+import { isEqual } from "lodash";
+import {
+  getRevision,
+  markRevisionAsPublished,
+} from "back-end/src/models/FeatureRevisionModel";
+import { ToggleFeatureResponse } from "back-end/types/openapi";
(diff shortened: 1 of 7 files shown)


packages/back-end/src/api/features/revertFeature.ts
```diff
@@ -0,0 +1,131 @@
...
+      ) {
+        context.permissions.throwPermissionError();
+      }
+      changes.defaultValue = revision.defaultValue;
+    }
+
+    const changedEnvs: string[] = [];
+    environmentIds.forEach((env) => {
+      if (
+        revision.rules[env] &&
```
Kevin-Chant, line 76: @jdorn this means you can't revert to completely remove a broken environment, right?

Kevin-Chant (approved):
Approving so we can merge sooner, but one question about the actual revert functionality

packages/back-end/src/api/features/revertFeature.ts
jdorn: Not sure.  I exactly copied the logic from our existing internal revert endpoint.  It's very possible there was a bug there

PR #4229, opened 2025-06-26
issue-4074: implement get coderefs by feature id

paigeruppel-upstart:
### Features and Changes

Define and implement getCodeRefs - /code-references/{id} - returns code references for a given feature id

- Closes https://github.com/growthbook/growthbook/issues/4074

### Testing

start the application and seed test data
invoke the new endpoint

### Screenshots

local instance, hitting the new endpoint with a feature flag that does not have code references:
image

**TODO** example response for a feature flag with code references

PR #4230, opened 2025-06-27
Api endpoint for listing code refs

Kevin-Chant:
### Features and Changes

Adds an endpoint for listing all code refs for an organization, paginated with the standard query params

Also cleans up a few unnecessary changes I added in #4229 for simplicity

### DIFF
--- a/packages/back-end/src/api/openapi/paths/getCodeRefs.yaml
+++ b/packages/back-end/src/api/openapi/paths/getCodeRefs.yaml
@@ -14,6 +14,7 @@ responses:
     content:
       application/json:
         schema:
-          type: array
-          items:
-            $ref: "../schemas/CodeRef.yaml"
+          codeRefs:
+            type: array
+            items:
+              $ref: "../schemas/CodeRef.yaml"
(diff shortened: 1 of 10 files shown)


packages/back-end/src/models/FeatureCodeRefs.ts
```diff
@@ -129,15 +133,12 @@ export const getAllCodeRefsForFeature = async ({
   }).then((docs) => docs.map(toInterface));
 };
 
-export const getCodeRefsForFeature = async ({
```
Kevin-Chant, line 132: This diff is actually from removing the helper from #4229 since it was superfluous, and adding a new one for the full list

packages/back-end/src/api/openapi/paths/getCodeRefs.yaml
Kevin-Chant, line 1: This isn't in-use by any customers yet, so it's safe to change the return shape for now

packages/back-end/src/models/FeatureCodeRefs.ts
paigeruppel-upstart: cool thank you Kevin!

paigeruppel-upstart:
@Kevin-Chant - pestering you a bit - any chance we could get this merged today? thank you again

bryce-fitzsimons (approved):
looks good to me

PR #4237, opened 2025-06-30
feat(dimensions): Update dimension editor to add priority and custom slices

lukesonnet:
### Features and Changes

Allow customers to set custom dimension values as well as dimension priorities (priorities are behind a feature flag for now). It currently only has an effect on the dimension levels used in the health tab, but will eventually be used for auto-dimensions and post-stratification.

This required changes to:

1. The `ExposureQuery` model to save whether custom slices were used and what those values are.

```
export interface ExperimentDimensionMetadata {
  dimension: string;
  specifiedSlices: string[];
  customSlices?: boolean;
}
```
Here the custom slices are saved as the specified slices if they are set, otherwise it falls back to the traffic derived slices. We also store `customSlices` to keep the FE aware of when custom slices have been used as well as keep the specified slices from being overwritten by future "auto updating" slices that are updated via agenda job.

2. The *Update Dimension* modal now lets you specify slices + the priority (which gets saved as the order of the `dimensions` array in the `ExposureQuery`.
image

3. The `Dimension` landing page has also been updated for more visibility of experiment dimensions
image

### Testing

- [X] Specific slices and order are saved in exposure query
- [X] Custom slices appear in health tab
- [X] Health tab onboarding still works

### DIFF
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -47,7 +47,6 @@ import { getUserById } from "back-end/src/models/UserModel";
 import { AuditUserLoggedIn } from "back-end/types/audit";
 import {
   createDimensionSlices,
-  getLatestDimensionSlices,
   getDimensionSlicesById,
 } from "back-end/src/models/DimensionSlicesModel";
 import { DimensionSlicesQueryRunner } from "back-end/src/queryRunners/DimensionSlicesQueryRunner";
@@ -761,25 +760,6 @@ export async function getDimensionSlices(
   const { id } = req.params;
 
   const dimensionSlices = await getDimensionSlicesById(org.id, id);
-
-  res.status(200).json({
-    status: 200,
-    dimensionSlices,
-  });
-}
-
-export async function getLatestDimensionSlicesForDatasource(
-  req: AuthRequest<null, { datasourceId: string; exposureQueryId: string }>,
-  res: Response
-) {
-  const { org } = getContextFromReq(req);
-  const { datasourceId, exposureQueryId } = req.params;
-
-  const dimensionSlices = await getLatestDimensionSlices(
-    org.id,
-    datasourceId,
-    exposureQueryId
-  );
 
   res.status(200).json({
     status: 200,
--- a/packages/front-end/components/Settings/EditDataSource/DimensionMetadata/DimensionSlicesRunner.tsx
+++ b/packages/front-end/components/Settings/EditDataSource/DimensionMetadata/DimensionSlicesRunner.tsx
@@ -203,23 +203,23 @@ const RefreshData = ({
             </Text>
           </Flex>
         ) : null}
-          <RunQueriesButton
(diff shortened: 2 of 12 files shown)


packages/back-end/src/controllers/datasources.ts
```diff
@@ -768,25 +767,6 @@ export async function getDimensionSlices(
   });
 }
 
-export async function getLatestDimensionSlicesForDatasource(
```
jdorn, line 771: Nit: I think we can also delete the `getLatestDimensionSlices` method from the model?  I don't see anywhere using that now.

packages/front-end/pages/dimensions/index.tsx
```diff
@@ -86,15 +157,69 @@ const DimensionsPage: FC = () => {
...
+                      <a
+                        className="dropdown-item"
+                        href={`/datasources/${item.datasourceId}#${EAQ_ANCHOR_ID}`}
+                        onClick={(e) => {
+                          e.preventDefault();
+                          router.push(
+                            `/datasources/${item.datasourceId}#${EAQ_ANCHOR_ID}`
+                          );
+                        }}
+                      >
```
jdorn, line 201: Nit: Does a normal `<Link>` component not work here?

packages/front-end/components/Settings/EditDataSource/ExperimentAssignmentQueries/ExperimentAssignmentQueries.tsx
```diff
@@ -305,13 +308,68 @@ export const ExperimentAssignmentQueries: FC<ExperimentAssignmentQueriesProps> =
...
+  dataSource: DataSourceInterfaceWithParams,
+  onSave: (dataSource: DataSourceInterfaceWithParams) => void
+) => async (
+  customDimensionMetadata: CustomDimensionMetadata[],
+  dimensionSlices?: DimensionSlicesInterface
+) => {
+  const copy = cloneDeep<DataSourceInterfaceWithParams>(dataSource);
+  const exposureQuery = copy.settings?.queries?.exposure?.[editingIndex];
+
+  if (exposureQuery) {
```
jdorn: Nit, if you invert this logic, it reduces the nesting levels and makes it easier to read.

```ts
if (!exposureQuery) {
  return;
}
```

Also, should we throw an error or something in that case?  When would this even happen?

packages/front-end/components/Settings/EditDataSource/DimensionMetadata/DimensionSlicesRunner.tsx
```diff
@@ -0,0 +1,473 @@
...
+            e.preventDefault();
+            try {
+              setError("");
+              await refreshDimensionSlices();
+            } catch (e) {
+              setError(e.message);
+              console.error(e);
+            }
+          }}
+        >
```
jdorn: This will be a form nested inside another form (`<Modal>` is wrapped in a form already).  This will likely cause issues in browsers.  It looks like the `RunQueriesButton` has an `onSubmit` prop you can specify instead of wrapping in a form.

PR #4238, opened 2025-07-01
Add python process pool to reduce CPU and startup costs

jdorn:
There are two huge performance problems with our stats engine today:
1. We generate a temporary `.py` file, write it to the file system, and instantiate a new Python shell every time. This is wildly inefficient.
2. We have no concurrency limits, so too many calls to the stats engine at once can overload the server's CPU and impact other parts of the app.

This PR fixes the above problems:
1. We use a long running Python process and stdin/stdout to pass data between Node and Python.
2. We create a pool of Python processes with a max size.  Requests are queued until a process becomes available.

Quick benchmarking on my laptop:
* Single call time went from 2.8s down to 0.8s.  CPU load decreased from 0.18 to 0.10
* With a pool size of 5, CPU maxed out at ~0.45 no matter how many calls were made in parallel.  Before, CPU would grow unbounded and eventually freeze the computer.

TODO:
- [x] Test within docker build
- [x] Test error handling and edge cases
  - [x] Python service gets killed while stats are running
  - [x] Python service times out processing request
  - [x] Python analysis code throws exception
  - [x] Node passes invalid JSON to stats engine
  - [x] Node passes JSON to stats engine that's missing required fields
  - [x] Python outputs invalid JSON in response (e.g. using Infinity or NaN)
  - [x] Python outputs JSON that is missing expected fields
  - [x] Pool / locking behavior with many parallel calls
- [x] Benchmarch CPU usage when calls are happening in parallel.  Is 10 the right concurrency limit to start? (answer: no, we should match # of CPUs, so 4 in prod)

### DIFF
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -169,8 +169,8 @@ function getAvgCPU(pre: os.CpuInfo[], post: os.CpuInfo[]) {
     user += postTimes.user - preTimes.user;
     system += postTimes.sys - preTimes.sys;
     total +=
-      Object.values(postTimes).reduce((n, sum) => n + sum, 0) -
-      Object.values(preTimes).reduce((n, sum) => n + sum, 0);
+      Object.values(postTimes).reduce((sum, n) => sum + n, 0) -
+      Object.values(preTimes).reduce((sum, n) => sum + n, 0);
   });
 
   return { user: user / total, system: system / total };
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -13,8 +13,14 @@ type PythonServerResponse<T> = {
   results: T;
 };
 
+// The stats engine usually finishes within 1 second
+// We use an overly conservative timeout to account for high load
+const STATS_ENGINE_TIMEOUT_MS = 15_000;
+const MAX_POOL_SIZE = 4;
+
 class PythonStatsServer<Input, Output> {
   private python: ChildProcess;
+  private pid = -1;
   private promises: Map<
     string,
     {
@@ -31,7 +37,8 @@ class PythonStatsServer<Input, Output> {
     this.python = spawn(command, [...pythonArgs, "-u", script], {
       stdio: ["pipe", "pipe", "pipe"],
     });
-    logger.info("Started Python stats server, pid " + this.python.pid);
+    this.pid = this.python.pid || -1;
+    logger.debug(`Python stats server (pid: ${this.pid}) started`);
     this.promises = new Map();
 
     this.python.stdout?.on("data", (data) => {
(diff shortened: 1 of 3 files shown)


packages/back-end/src/services/python.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  post.forEach((cpu, i) => {
+    const preTimes = pre[i]?.times || { user: 0, sys: 0 };
+    const postTimes = cpu.times;
+
+    user += postTimes.user - preTimes.user;
+    system += postTimes.sys - preTimes.sys;
+    total +=
+      Object.values(postTimes).reduce((n, sum) => n + sum, 0) -
+      Object.values(preTimes).reduce((n, sum) => n + sum, 0);
```
tzjames: It doesn't matter in terms of correctness but the accumulator comes first for reduce, so it should be:
``` 
Object.values(postTimes).reduce((sum, n) => sum + n, 0) -
Object.values(preTimes).reduce((sum, n) => sum + n, 0);
```

packages/back-end/src/services/python.ts
```diff
@@ -0,0 +1,173 @@
...
+        ExperimentDataForStatsEngine[],
+        MultipleExperimentMetricAnalysis[]
+      >(path.join(__dirname, "..", "..", "scripts", "stats_server.py"));
+    },
+    destroy: async (server) => server.destroy(),
+    validate: async (server) => server.isRunning(),
+  },
+  {
+    min: 1,
+    max: 10,
```
tzjames: You said that we run only 5 at a time.  I assume that is referring to the `defaultLockLimit` when we initialize `Agenda`.  Why would this need to be 10?  Perhaps we can make it an env var and use it in both places?

packages/back-end/src/services/stats.ts
```diff
@@ -226,55 +205,14 @@ export function getBanditSettingsForStatsEngine(
 async function runStatsEngine(
   statsData: ExperimentDataForStatsEngine[]
 ): Promise<MultipleExperimentMetricAnalysis[]> {
-  const escapedStatsData = JSON.stringify(statsData).replace(/\\/g, "\\\\");
-  const start = Date.now();
-  const cpus = os.cpus();
-  const options = process.env.GB_ENABLE_PYTHON_DD_PROFILING
```
tzjames, line 232: Do we still want to allow Python profiling?

tzjames (approved):

packages/back-end/src/services/python.ts
jdorn: The "5 at a time" is from Agenda.  The stats engine is also called manually from the UI and with upcoming dashboard changes, a single experiment refresh might end up kicking off multiple Python calls.  I didn't want the UI to hang because of background refreshes.

I just guessed at 10.  I'm thinking we watch it in prod under load and see if that needs adjusting.

tzjames: Ah.  Perhaps we should add dd metric for monitoring of the pool to see how many are running.  Then we can see effect on CPU and can have a better idea of the maximum.  We can also monitor the number pending in case the number of jobs is greater than the max pool size for a length of time so we can scale more.

jdorn: I'm using the `generic-pool` library.  It looks like both OpenTelemetry and DataDog instrument this already - https://github.com/DataDog/dd-trace-js/blob/41ce55c79c4d08c2a3e80549d9f3547232af986c/packages/datadog-instrumentations/src/generic-pool.js

Not really sure what their instrumentation actually looks like (is it creating spans? metrics?)

tzjames: I think it creates metrics, probably prefixed with generic_pool.  I think it is fine to land and then check, if you don't have dd set up locally.

PR #4240, opened 2025-07-01
Query Param for ability to delete code refs that are not sent up in the request.

gazzdingo:
### Features and Changes
Code refs was not deleting old references it was only adding them this is a PR to remove reference that are not found

### Testing
https://www.loom.com/share/4ad9bc0c0afd42e8b35a0ef1d8cbdc35


packages/back-end/src/api/code-refs/postCodeRefs.ts
```diff
@@ -5,13 +5,55 @@ import { postCodeRefsValidator } from "back-end/src/validators/openapi";
...
+      }
+    );
+
+    const existingCodeRefsForRepoBranch = allExistingCodeRefs.filter(
+      (codeRef) => codeRef.repo === repo && codeRef.branch === branch
+    );
+
+    const existingFeatures = existingCodeRefsForRepoBranch.map(
+      (codeRef) => codeRef.feature
+    );
```
Kevin-Chant: Needs a dedupe to prevent unnecessary db calls. Maybe
```suggestion
    const existingFeatures = [...new Set(existingCodeRefsForRepoBranch.map(
      (codeRef) => codeRef.feature
    ))];
```

Kevin-Chant (approved):
One improvement for reducing db calls, but otherwise this LGTM

PR #4241, opened 2025-07-01
Adds a bit more extensibility to SqlExplorerModal

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
+++ b/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
@@ -57,6 +57,10 @@ export interface Props {
   };
   id?: string;
   mutate: () => void;
+  allowSave?: boolean;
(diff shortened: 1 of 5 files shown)


packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -57,13 +57,21 @@ export interface Props {
   };
   id?: string;
   mutate: () => void;
+  allowSave?: boolean; // Controls if user can save query AND also controls if they can create/save visualizations
```
Kevin-Chant: Personally I find true-if-undefined booleans a bit confusing. Would prefer swapping to `disableSave?: boolean` and inverting the logic

Kevin-Chant (approved):
LGTM. One preference for the optional prop but that doesn't make a big difference

PR #4247, opened 2025-07-02
Temporarily revert new homepage

jdorn:
### Features and Changes

Move new homepage to `/new-home` for now while we work on fixing some bugs.  Revert index behavior to previous


packages/front-end/pages/new-home.tsx
```diff
@@ -0,0 +1,5 @@
+import GetStartedAndHomePage from "@/components/GetStarted";
+
+export default function NewHome(): React.ReactElement {
```
gazzdingo: dont need this since we can just go to `/getstarted` to test

gazzdingo (approved):

PR #4248, opened 2025-07-02
Hot fix links

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/front-end/components/GetStarted/AdvancedFeaturesCard.tsx
+++ b/packages/front-end/components/GetStarted/AdvancedFeaturesCard.tsx
@@ -1,18 +1,17 @@
 import { AspectRatio, Box, Text } from "@radix-ui/themes";
-import Link from "next/link";
-
 import { CommercialFeature } from "shared/src/enterprise/license-consts";
 import PaidFeatureBadge from "@/components/GetStarted/PaidFeatureBadge";
+import { DocSection, DocLink } from "@/components/DocLink";
 import styles from "./AdvancedFeaturesCard.module.scss";
 
 export default function AdvancedFeaturesCard({
-  href,
+  docSection,
   imgUrl,
   title,
   description,
   commercialFeature,
 }: {
-  href: string;
+  docSection: DocSection;
   imgUrl: string;
   title: string;
   description?: string;
@@ -53,7 +52,7 @@ export default function AdvancedFeaturesCard({
   return (
     <Box width="100%" height="100%">
       <AspectRatio ratio={16 / 9}>
-        <Link href={href}>{card}</Link>
+        <DocLink docSection={docSection}>{card}</DocLink>
       </AspectRatio>
     </Box>
   );
--- a/packages/front-end/components/GetStarted/AdvancedFeaturesCard.tsx
+++ b/packages/front-end/components/GetStarted/AdvancedFeaturesCard.tsx
@@ -1,4 +1,5 @@
 import { AspectRatio, Box, Text } from "@radix-ui/themes";
+import Link from "next/link";
 
 import { CommercialFeature } from "shared/src/enterprise/license-consts";
 import PaidFeatureBadge from "@/components/GetStarted/PaidFeatureBadge";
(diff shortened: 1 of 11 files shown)


packages/front-end/components/GetStarted/NeedingAttention.tsx
```diff
@@ -405,14 +415,9 @@ const NeedingAttention = (): React.ReactElement | null => {
...
-                <Button
-                  mt="2"
-                  onClick={() => {
-                    window.location.href = "/experiments";
-                  }}
-                >
-                  View all experiments
-                </Button>
+                <Link href="/experiments" style={{ textDecoration: "none" }}>
+                  <Button mt="2">View all experiments</Button>
```
jdorn: There is a `LinkButton` component you should use instead.

packages/front-end/components/GetStarted/NeedingAttention.tsx
```diff
@@ -523,14 +534,9 @@ const NeedingAttention = (): React.ReactElement | null => {
...
-                <Button
-                  mt="2"
-                  onClick={() => {
-                    window.location.href = "/features";
-                  }}
-                >
-                  View all feature flags
-                </Button>
+                <Link href="/features" style={{ textDecoration: "none" }}>
+                  <Button mt="2">View all feature flags</Button>
```
jdorn: Same here

packages/front-end/components/GetStarted/AdvancedFeaturesCard.tsx
```diff
@@ -52,9 +53,7 @@ export default function AdvancedFeaturesCard({
   return (
     <Box width="100%" height="100%">
       <AspectRatio ratio={16 / 9}>
-        <a href={href} target="_blank" rel="noreferrer">
-          {card}
-        </a>
+        <Link href={href}>{card}</Link>
```
jdorn: We should still do `target="_blank"` for the ones that link to our docs.

gazzdingo: using docsLinks now

packages/front-end/components/GetStarted/index.tsx
```diff
@@ -52,62 +57,56 @@ const advancedFeatureList: AdvancedFeature[] = [
...
     description: "Use temp tables for intermediate steps",
-    href: "https://docs.growthbook.io/app/data-pipeline",
+    docSection: "pipelineMode",
     commercialFeature: "pipeline-mode",
   },
   {
     imgUrl: "/images/get-started/advanced/fact-tables.png",
-    title: "Fact Tables",
+    title: "Query Optimization",
     description: "Enable automatic query optimization",
```
jdorn: Change the description copy:
> Improve SQL performance and reduce costs

I also think we should probably remove the `commercialFeature` from this one.  Only some of the optimizations require a paid license.  There are basic optimizations on that docs page that anyone can do.

packages/front-end/components/GetStarted/index.tsx
```diff
@@ -161,7 +167,19 @@ const GetStartedAndHomePage = (): React.ReactElement => {
...
-    return advancedFeatureList.sort(() => Math.random() - 0.5).slice(0, 3);
+    const advancedFeatureListWithAnalytics = [...advancedFeatureList];
+    if (isCloud()) {
+      advancedFeatureList.push({
+        imgUrl: "/images/get-started/advanced/feature-flag.png",
+        title: "Feature Flag Analytics",
+        description: "View flag evaluations in real time",
+        docSection: "managedWarehouseTracking",
+        commercialFeature: "managed-warehouse",
+      });
```
jdorn, line 177: I don't think this works?  You are pushing onto the `advancedFeatureList` array and not the `advancedFeatureListWithAnalytics` array

packages/front-end/components/GetStarted/index.tsx
gazzdingo: cool that is easy

packages/front-end/components/GetStarted/index.tsx
gazzdingo: good catch

PR #4249, opened 2025-07-02
add back homepage

gazzdingo:
### Features and Changes
Once fixes for home page go out we can re add back the home page

### DIFF
--- a/packages/front-end/components/Layout/Layout.tsx
+++ b/packages/front-end/components/Layout/Layout.tsx
@@ -27,10 +27,10 @@ import { usePageHead } from "./PageHead";
 
 const navlinks: SidebarLinkProps[] = [
   {
-    name: "Get Started",
-    href: "/getstarted",
+    name: "Home",
+    href: "/",
     Icon: BsHouse,
-    path: /^getstarted/,
+    path: /^$/,
     className: styles.first,
   },
   {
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Layout/Layout.tsx
```diff
@@ -343,8 +343,8 @@ const otherPageTitles = [
     title: "Personal Access Tokens",
   },
   {
-    path: /^getstarted/,
-    title: "Get Started",
+    path: /^$/,
+    title: "Home",
```
gazzdingo: i think we should just keep this as get started but im not sure where this renders?

packages/front-end/pages/index.tsx
```diff
@@ -71,5 +67,5 @@ export default function Home(): React.ReactElement {
     );
   }
 
-  return <LoadingOverlay />;
+  return <GetStartedAndHomePage />;
```
jdorn: The flicker is pretty bad with this.  We should wait until things are loaded before rendering.

```ts
if (featuresLoading || experimentsLoading) {
  return <LoadingOverlay />;
}

return <GetStartedAndHomePage />;
```

gazzdingo: cool that was my comment in the old PR to check if we should add a loading screen

jdorn (approved):

PR #4250, opened 2025-07-02
updating covariance calculation for binomial metrics in CUPED

lukebrawleysmith:
currently the covariance calculation used in regression adjusted statistics that have binomial metric components for the pre- and post- statistics is wrong in that it does not leverage the fact that these are binomial random variables.  Because our variance calculations do make this assumption, this can lead to a covariance matrix that is negative definite, potentially leading to negative variance estimates, pre-post correlations bigger than 1, etc.  

Practically speaking, this issue surfaces only when sample sizes are small in CUPED analyses of binomial metrics, as the standard sample variance and the binomial variance estimation becomes close as n increases.  

This PR changes adds a check for binomial statistics in our covariance calculation to obviate future errors at small sample sizes.

### DIFF
--- a/packages/stats/gbstats/models/statistics.py
+++ b/packages/stats/gbstats/models/statistics.py
@@ -141,6 +141,10 @@ class RegressionAdjustedStatistic(Statistic):
     post_pre_sum_of_products: float
     theta: Optional[float]
 
+    def __post_init__(self) -> None:
+        if not isinstance(self.post_statistic, type(self.pre_statistic)):
+            raise TypeError("post_statistic and pre_statistic must be of the same type")
+
     def __add__(self, other):
         if not isinstance(other, RegressionAdjustedStatistic):
             raise TypeError("Can add only another RegressionAdjustedStatistic instance")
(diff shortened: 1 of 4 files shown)


packages/stats/gbstats/models/statistics.py
```diff
@@ -341,6 +398,15 @@ def cov_m_post_d_post(self) -> float:
     def cov_m_post_m_pre(self) -> float:
         if self.n <= 1:
             return 0
+        if isinstance(self.m_statistic_post, ProportionStatistic) and isinstance(
```
lukesonnet: If all of these covariance are the same just with two different statistics and one sum of products that gets passed in, maybe we should have one generic method for that that splits the covariance between binomial and non-binomial

packages/stats/tests/test_shared_models.py
```diff
@@ -262,19 +262,17 @@ def test_initialize_theta(self):
         moments = EffectMoments(
             [(stat_a, stat_b)], config=EffectMomentsConfig(difference_type="absolute")
         )
-        self.assertEqual(moments.stat_a.theta, 0.9722222222222223)  # type: ignore
-        self.assertEqual(moments.stat_b.theta, 0.9722222222222223)  # type: ignore
+        self.assertEqual(moments.stat_a.theta, 0.8333333333333334)  # type: ignore
+        self.assertEqual(moments.stat_b.theta, 0.8333333333333334)  # type: ignore
 
     def test_negative_variance(self):
```
lukesonnet, line 268: Is there a different test we can add for this negative variance check?

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: good call, updated

packages/stats/tests/test_shared_models.py
lukebrawleysmith: good call, updated

lukesonnet (approved):

PR #4254, opened 2025-07-03
Bump react types to v19

### LINKED ISSUE #4204: [Bug] Incompatibility with React 19 types - deprecation of global JSX namespace
### Summary

The most recent version of `@growthbook/growthbook-react` is incompatible with React 19 types. Specifically, the deprecation of the global JSX namespace as specified here: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript.

Rather than use `JSX.Element`, you will either need to use `React.JSX.Element` or import JSX from react.

https://github.com/growthbook/growthbook/blob/main/packages/sdk-react/src/GrowthBookReact.tsx#L157

#### Expected Behavior

When compiling with React 19 types, typecheck should not fail.

#### Current Behavior

When compiling with React 19 types, build fails with `Cannot find namespace 'JSX'` error.

benruehl:
### Features and Changes

Updates `@types/react` and `@types/react-dom` from v18 to v19. This should solve typescript build errors for react@19 applications.

- Closes #4204

### Testing

Create a react@19 app with typescript, add dependency to growthbook-react sdk and run tsc build.

### DIFF
--- a/packages/sdk-react/src/GrowthBookReact.tsx
+++ b/packages/sdk-react/src/GrowthBookReact.tsx
@@ -110,7 +110,7 @@ export function FeaturesReady({
   children: React.ReactNode;
   timeout?: number;
   fallback?: React.ReactNode;
-}) {
+}): React.ReactElement {
   const gb = useGrowthBook();
   const [hitTimeout, setHitTimeout] = React.useState(false);
   const ready = gb ? gb.ready : false;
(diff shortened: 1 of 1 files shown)


benruehl:
Thanks @jdorn, good suggestion. `React.ReactElement` seems to work well.
I think the explicit return types are not necessary. Would you like me to add them anyway?

jdorn:
From my testing, without the explicit return types, it looked typescript was inferring `JSX.Element` and adding that to the declaration files.

Kobe (approved):

benruehl:
Anything I can do to get this merged? @jdorn

jdorn (approved):

PR #4255, opened 2025-07-03
Show next scheduled update

### LINKED ISSUE #3304: [Feature] Show next scheduled update
We should modify the date logic in the UI to show when the next scheduled update is and to show more metadata about whether or not a update is scheduled and why:

image

This will help debug/explain when an update is not scheduled, as well as help users know they can wait until the next scheduled update to get refreshed results.

devcodes9:
### Features and Changes
- Added next scheduled update
- Closes #3304 

### Testing
- Go to any experiment
- Verify that the Next Scheduled Update time appears correctly based on the cron/frequency set.

### Screenshots
image

Note: While the issue also suggests displaying why an update is not scheduled, this has not been implemented in this PR as I wasn’t fully sure how to implement that, as I couldn’t find relevant metadata to support it.


jdorn:
Thanks for the contribution.  I think we will want a different visual treatment here to keep the UI clean.

jdorn:

PR #4271, opened 2025-07-08
computing MDE for midexperiment power

lukebrawleysmith:
Currently GrowthBook customers do not have a mechanism for easily determining whether their target MDE is too large.  This PR calculates the MDE during the middle of the experiment.  By comparing the outputted MDE to the target MDE, the customer can see if their target MDE is much too big.  

TODO: 
align on how and when to surface the target MDE
all front end work

### DIFF
--- a/packages/stats/gbstats/power/midexperimentpower.py
+++ b/packages/stats/gbstats/power/midexperimentpower.py
@@ -3,7 +3,7 @@ from dataclasses import field
 import numpy as np
 from pydantic.dataclasses import dataclass
 from scipy.stats import norm
-
+import copy
 from gbstats.models.tests import TestResult, frequentist_variance_all_cases
 from gbstats.frequentist.tests import (
     sequential_interval_halfwidth,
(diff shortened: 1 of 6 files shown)


packages/stats/gbstats/power/midexperimentpower.py
```diff
@@ -193,7 +200,7 @@ def power(self, scaling_factor) -> float:
             num_2 = (
                 adjusted_variance * self.prior_effect.mean / self.prior_effect.variance
             )
-            num_3 = self.target_mde
+            num_3 = copy.deepcopy(mde)
```
lukesonnet: Why is this copy necessary? seems like `num_3` is just used as an input below?

lukebrawleysmith: I was just being hyper careful, I don't think a copy is needed.

packages/stats/tests/test_midexperimentpower.py
```diff
@@ -89,39 +89,86 @@ def setUp(self):
...
                 self.result_bayes.scaling_factor, scaling_factor_true, places=4
             )
         else:
             raise ValueError("scaling_factor_bayes is None")
+        if self.mde_bayes:
+            self.assertAlmostEqual(self.mde_bayes, mde_true, places=4)
+        else:
+            raise ValueError("mde_bayes is None")
+
+    # Current MDE at target power of 0.8 should result in computed power of 0.8
```
lukesonnet, line 157: @lukebrawleysmith , does this test look right?

lukebrawleysmith: yes, and it's a good sanity check

PR #4277, opened 2025-07-10
Updates SQL Explorer front end

mknowlton89:
### Features and Changes

This PR introduces a few new changes:

- A user can now duplicate a visualization
- We've updated the tab to use a dropdown so users can now choose to duplicate or delete a visualization
- I've also updated the sql tab so the edit icon is only rendered when that tab is focused, to be more consistent with the other tabs
- I've also updated the max number of visualizations from 3 to 10
- I've also added a SQL icon to the sql tab to differentiate that tab vs the visualization tabs
- We've also added logic so that if there are more than 3 variations, we start to truncate the visualization tab names to save space.

### Dependencies

- None

### Testing

- [x] Ensure this works for both light and dark modes
- [x] Ensure that you can add up to 10 visualizations
- [x] Ensure that the sql tab only shows the edit icon when the tab is focused
- [x] Ensure that the saved query input field is rendered in a way that matches the mock and uses the correct buttons
- [x] Ensure that the horizontal scroll works well when a saved query has a high number of visualizations or if the visualizations have long names
- [x] Ensure the duplicate visualization button is only enabled when there are less than 10 visualizations - and if not, a tooltip is rendered
- [x] Ensure that if there are 10 visualizations, the add visualization button is still rendered, but it's disabled with a tooltip explaining why
- [x] Ensure the duplicate viz functionality works correctly

### Screenshots
Screenshot 2025-07-14 at 1 37 22 PM
Screenshot 2025-07-14 at 1 37 33 PM
Screenshot 2025-07-14 at 1 37 52 PM
Screenshot 2025-07-14 at 1 37 58 PM
Screenshot 2025-07-14 at 1 38 04 PM
Screenshot 2025-07-14 at 1 38 24 PM
Screenshot 2025-07-14 at 1 40 35 PM

### DIFF
--- a/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
+++ b/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
@@ -307,6 +307,18 @@ export default function SqlExplorerModal({
     }
   };
 
+  const getTruncatedTitle = (
+    title: string,
+    totalVisualizationCount: number
+  ): string => {
+    // Only truncate if there are 4 or more visualizations
+    if (totalVisualizationCount < 4) return title;
+
+    const maxLength = 20;
+    if (title.length <= maxLength) return title;
+    return title.substring(0, maxLength) + "...";
+  };
+
   // Filter datasources to only those that support SQL queries
   // Also only show datasources that the user has permission to query
   const validDatasources = datasources.filter(
@@ -451,7 +463,14 @@ export default function SqlExplorerModal({
                     style={{ paddingRight: "0px" }}
                   >
                     <Flex align="center" gap="2">
-                      {config.title || `Visualization ${index + 1}`}
+                      <span
+                        title={config.title || `Visualization ${index + 1}`}
+                      >
+                        {getTruncatedTitle(
+                          config.title || `Visualization ${index + 1}`,
+                          dataVizConfig.length
+                        )}
+                      </span>
                       {!readOnlyMode && tab === `visualization-${index}` ? (
                         <DropdownMenu
                           trigger={
--- a/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
+++ b/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
@@ -6,7 +6,7 @@ import {
   FaCheck,
   FaTimes,
 } from "react-icons/fa";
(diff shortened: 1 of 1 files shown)


packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -304,6 +307,18 @@ export default function SqlExplorerModal({
...
+  const getTruncatedTitle = (
+    title: string,
+    totalVisualizationCount: number
+  ): string => {
+    // Only truncate if there are 4 or more visualizations
+    if (totalVisualizationCount < 4) return title;
+
+    const maxLength = 20;
+    if (title.length <= maxLength) return title;
+    return title.substring(0, maxLength) + "...";
```
ahdriel, line 319: very nit: should it be `maxLength - 3` to account for the added `...`?

Otherwise the text will be longer in the cases of 21 or 22 characters.

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -376,7 +393,6 @@ export default function SqlExplorerModal({
                       <Flex align="center" gap="2">
                         <input
                           type="text"
-                          className="form-control"
                           value={tempName}
```
ahdriel, line 396: question (nit): Should this still be named `tempName`?

ahdriel (approved):
🙏  Thanks for the updates -- looks so much better!

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
mknowlton89: I think it makes sense to keep this. The tempName is the state for when a saved queries name is being edited, but not yet saved. If the user dismisses the changes we wipe the tempName, otherwise we use the tempName to update the form's `name` field, so I think it still makes sense.

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
mknowlton89: The 20 char limit isn't anything special, just an arbitrary limit Kari and I landed on, so it's safe, but I think this is a good callout.

PR #4282, opened 2025-07-14
Don't send orgs with trackingDisabled:true to data-enrichment endpoint

tzjames:
### Features and Changes
Pro orgs that stop paying may continue to have a growthbook clickhouse datasource even though they are free.  We will be nice and in general let them continue to track events up to a limit.  But if any org is costing us too much money we may want to hard cut them off quickly.  This PR sets a new property on OrganizationModel that will always be undefined unless set by hand to be true.  If `trackingDisabled:true` for an org, then we won't send it to the Ingestor and the Ingestor will return ClientKeyInvalid 400 error.

Notes:
- Although the query is grabbing across multiple orgs, this doesn't seem to warrant the "_dangerous" prefix like the other functions in the ingestor as all it returns is a subset of the already know orgIds passed in.  
- Filtering by orgIds is necessary as we don't want to create an index for trackingDisabled as we don't want to slow down writes.

### Testing

In Mongo set an org with an sdk key being the growthbook sdk key. 
Have .env.local:
```
IS_CLOUD=true
IS_MULTI_ORG=true
```
Make sure the env var TRACKING_DISABLED is not set. 
Start the ingestor.
Login to the org with growthbook sdk key.  
Hit localhost:3000.
In the ingestor logs see 200 returned.
In Mongo set `trackingDisabled:true`
Go to two different pages.
In the ingestor logs see 200 returned for the first one (it will still be in cache, fetch done in the backgroun)
See 400 InvalidClientKey returned for the second one.

### DIFF
--- a/packages/back-end/src/api/ingestion/ingestion.router.ts
+++ b/packages/back-end/src/api/ingestion/ingestion.router.ts
@@ -7,12 +7,14 @@ import {
 import { SDKConnectionInterface } from "back-end/types/sdk-connection";
 import { _dangerousGetSdkConnectionsAcrossMultipleOrgs } from "back-end/src/models/SdkConnectionModel";
 import { _dangerousGetAllGrowthbookClickhouseDataSources } from "back-end/src/models/DataSourceModel";
+import { getOrganizationIdsWithTrackingDisabled } from "back-end/src/models/OrganizationModel";
 
 interface SdkInfo {
   organization: string;
   client_key: string;
   datasource: string;
   environment: string;
+  trackingDisabled: boolean;
 }
 
 interface GetDataEnrichmentResponse {
@@ -21,12 +23,17 @@ interface GetDataEnrichmentResponse {
   };
 }
 
-function sdkInfo(conn: SDKConnectionInterface, datasource: string): SdkInfo {
+function sdkInfo(
+  conn: SDKConnectionInterface,
+  datasource: string,
+  trackingDisabled: boolean
+): SdkInfo {
   return {
     organization: conn.organization,
     client_key: conn.key,
     environment: conn.environment,
     datasource,
+    trackingDisabled,
   };
 }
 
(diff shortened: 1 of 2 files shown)


packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -43,9 +44,20 @@ export const getDataEnrichment = createApiRequestHandler({
     const dataSourcesByOrgId = Object.fromEntries(
       dataSources.map((ds) => [ds.organization, ds.id])
     );
+    const orgIds = Object.keys(dataSourcesByOrgId);
+
+    const orgIdsWithTrackingDisabled = await getOrganizationIdsWithTrackingDisabled(
+      orgIds
+    );
+
+    const orgIdsWiithTrackingEnabled = orgIds.filter(
```
mknowlton89: Small typo here - `orgIdsWithTrackingEnabled`

mknowlton89 (approved):
LGTM - One typo, but looks solid.

packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -43,9 +44,20 @@ export const getDataEnrichment = createApiRequestHandler({
...
+      (x) => !orgIdsWithTrackingDisabled.has(x)
```
jdorn, line 54: Instead of filtering them out, do we want to include them with a flag so we can return a more informative error in the ingestor instead of just "ClientKeyInvalid 400"?

tzjames: I did have that to begin with. I was going to return 403 forbidden. This seemed easier as I didn't have to touch the other repo. It is unclear how much we will use this and how much we will stick with putting them in the overages table so it makes sense to me to spend more time on that. Especially as we might want to just reach out to them by email if we ever actually use it

PR #4284, opened 2025-07-14
Add BACKEND_PORT env var

ari-party:
### Features and Changes

- Add support for setting the port for the back-end using env vars. If the `PORT` env var is set, it sets the port for the back-end and front-end, causing a conflict.

Made this change due to the fact [Railway](https://railway.com) [always sets the `PORT` env var](https://docs.railway.com/guides/public-networking#railway-provided-port), and are unable to deploy GrowthBook because of this.

### DIFF
--- a/docs/docs/self-host/env.mdx
+++ b/docs/docs/self-host/env.mdx
@@ -63,6 +63,9 @@ growthbook:
     - API_HOST=http://<my ip or url>:4100
 ```
 
+If you for some reason have a **PORT** environment variable set,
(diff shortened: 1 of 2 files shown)


ari-party:
@royalfig

`API_HOST=https://growthbook-api.domain.tld:3100`
`APP_ORIGIN=https://growthbook.domain.tld:3000`

The domains are reverse proxied from 443 -> 3000/3100
I have tried with and without suffixing the port to the env vars

These are my Docker logs:
1752525281_X5JBdrLVzt

PR #4287, opened 2025-07-15
feat(metrics): Improve tracing + add trackJobs to all jobs

ahdriel:
### Features and Changes

Add additional logs for when the job start and finish.
Also wrap jobs that we missed with `trackJobs` to have proper metrics.

### DIFF
--- a/packages/back-end/src/jobs/updateInformationSchema.ts
+++ b/packages/back-end/src/jobs/updateInformationSchema.ts
@@ -74,12 +74,7 @@ const updateInformationSchema = trackJob(
 let agenda: Agenda;
 export default function (ag: Agenda) {
   agenda = ag;
-
-  agenda.define(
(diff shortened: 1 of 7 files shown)


packages/back-end/src/jobs/updateInformationSchema.ts
```diff
@@ -74,12 +74,7 @@ const updateInformationSchema = trackJob(
 let agenda: Agenda;
 export default function (ag: Agenda) {
   agenda = ag;
-
-  agenda.define(
-    UPDATE_INFORMATION_SCHEMA_JOB_NAME,
-
-    updateInformationSchema
-  );
+  agenda.define(UPDATE_INFORMATION_SCHEMA_JOB_NAME, updateInformationSchema);
```
tzjames, line 77: did you mean to add trackJob here.

tzjames (approved):
This should provide more visibility!

PR #4288, opened 2025-07-15
Add several fixes for a job that runs forever

tzjames:
### Features and Changes

There was a client that started a query to a postgres client that never resolved it's promise (either successfully or rejecting).  Once the lockLifetime was reached another server picked it up.  Eventually all jobs servers were running this one job leaving no room for other jobs to run.

This change:

1. Gets rid of lockLifetimeout of 30 minutes (will default back to 10 minutes)
2. Calls touch() every 3 minutes, to make sure that another job doesn't pick it up.  If the server dies, then another server should pick it up 10 minutes from the last touch().
3. Adds a `query_timeout` to the postgres client call. 
4. Adds a timeout to `runPostgresQuery` (in case the postgres client still doesn't reject when the timeout is reached)
5. Adds a timeout to the updateSingleExperimentJob (in case some other client also doesn't timeout promptly).

### Testing
...

### DIFF
--- a/packages/back-end/src/jobs/updateExperimentResults.ts
+++ b/packages/back-end/src/jobs/updateExperimentResults.ts
@@ -17,7 +17,10 @@ import {
 import { getContextForAgendaJobByOrgId } from "back-end/src/services/organizations";
 import { getMetricMap } from "back-end/src/models/MetricModel";
 import { notifyAutoUpdate } from "back-end/src/services/experimentNotifications";
-import { EXPERIMENT_REFRESH_FREQUENCY } from "back-end/src/util/secrets";
+import {
+  EXPERIMENT_REFRESH_FREQUENCY,
+  MAX_QUERY_TIMEOUT_MS,
+} from "back-end/src/util/secrets";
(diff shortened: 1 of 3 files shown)


packages/back-end/src/jobs/updateExperimentResults.ts
```diff
@@ -106,132 +104,184 @@ async function updateSingleExperiment(job: UpdateSingleExpJob) {
...
-        experiment,
-        isScheduled: true,
-      });
-      if (tempChanges.banditStage === "exploit") {
-        reweight = true;
-      }
-    }
+  startTouch();
+
+  const TIMEOUT = MAX_QUERY_TIMEOUT_MS + 1000; // Allow some buffer for the query client to close properly after it timesout
```
jdorn: This job will do many Mongo lookups, then run multiple queries, and then call the stats engine.  Depending on concurrency settings, queries may run sequentially as well.  So the job timeout needs to be significantly longer than the individual query timeout.

Maybe it's 1 hour max per query and 2 hours max per job (or something like that).  We could make a new env var if we want for the job timeout.

PR #4293, opened 2025-07-15
OpenAPI: Add feature/{id}/revisions paginated endpoint

bryce-fitzsimons:
Add a new OpenAPI `/features/{fid}/revisions` paginated endpoint.

image

image


packages/back-end/src/api/features/getFeatureRevisions.ts
```diff
@@ -0,0 +1,22 @@
...
+import { getFeatureRevisionsValidator } from "back-end/src/validators/openapi";
+
+export const getFeatureRevisions = createApiRequestHandler(
+  getFeatureRevisionsValidator
+)(async (req) => {
+  const allRevisions = await getFeatureRevisionsByStatus({
+    context: req.context,
+    organization: req.organization.id,
+    featureId: req.params.id,
+    limit: 10000, // fetch a large number, then paginate in-memory
```
jdorn: I'm a little worried about this. We've had performance issues with revisions in the past. Fetching all in memory can be hundreds of MBs or even a GB.

packages/back-end/src/api/openapi/schemas/FeatureRevision.yaml
```diff
@@ -0,0 +1,33 @@
+type: object
+required:
+  - baseVersion
+  - version
+  - comment
+  - date
+  - status
+  - rules
+properties:
```
jdorn, line 9: I noticed in the screenshot, `organization` and possibly other fields are also being returned. We should restrict the fields to just what is listed here. Might need to specify `additionalProperties: false` here in the JSON schema to make Typescript detect this and throw type errors.

bryce-fitzsimons: `additionalProperties: false` doesn't seem to do anything here. I'll take a deeper dive though...

packages/back-end/src/api/features/getFeatureRevisions.ts
bryce-fitzsimons: fair point. shifted the pagination to the DB lookup

packages/back-end/src/api/openapi/schemas/FeatureRevision.yaml
bryce-fitzsimons: fixed: needed to scrub the field from the response payload. could have also modified the toInterface() call but that seems risky

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -50,7 +50,7 @@ featureRevisionSchema.index({ organization: 1, status: 1 });
 
 type FeatureRevisionDocument = mongoose.Document & FeatureRevisionInterface;
 
-const FeatureRevisionModel = mongoose.model<FeatureRevisionInterface>(
+export const FeatureRevisionModel = mongoose.model<FeatureRevisionInterface>(
```
jdorn: Don't export the model.  Instead, export a `countDocuments` method.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -161,6 +163,7 @@ export async function getFeatureRevisionsByStatus({
   })
     .select("-log") // Remove the log when fetching all revisions since it can be large to send over the network
     .sort({ version: -1 })
```
jdorn: Sorting by newest first makes the offset unstable.  If a new version is added while your script is running, you could end up fetching the same revision twice.  Newest first makes sense for the other places that call this method, so maybe for the new REST endpoint only, we do oldest first.

packages/back-end/src/models/FeatureRevisionModel.ts
bryce-fitzsimons: whoops, fixed

packages/back-end/src/models/FeatureRevisionModel.ts
bryce-fitzsimons: that's true, switched it to "asc" (1) for this endpoint

jdorn (approved):

PR #4295, opened 2025-07-15
Add job health check by refreshing lock if process is still running

ahdriel:
### Features and Changes

In some scenarios of long-running jobs the process would still be active but the lock would have expired because we were still waiting for a different resource and the job would be picked up by a different runner which is wrong.

So we reduced the job lock timeout to 10minutes and added a lifecycle job that will update the lock periodically if the process is still running.

If the process die for whatever reason then the lock will work as expected and the job will be picked up again by a different runner.


packages/back-end/src/services/jobLifecycle.ts
```diff
@@ -0,0 +1,51 @@
+import { Job, JobAttributesData } from "agenda";
+import { logger } from "back-end/src/util/logger";
+import { MAX_QUERY_TIMEOUT_MS } from "back-end/src/util/secrets";
+
+const TOUCH_INTERVAL_MS = 3 * 60 * 1000;
+const JOB_TIMEOUT_MS = MAX_QUERY_TIMEOUT_MS + 1000; // Allow some buffer for the query client to close properly after it times out
```
tzjames: Jeremy commented in my other PR that some jobs do multiple queries, and recommended we have a separate timeout for jobs, perhaps with the default being 2 hours.

packages/back-end/src/services/jobLifecycle.ts
```diff
@@ -0,0 +1,51 @@
+import { Job, JobAttributesData } from "agenda";
+import { logger } from "back-end/src/util/logger";
+import { MAX_QUERY_TIMEOUT_MS } from "back-end/src/util/secrets";
+
+const TOUCH_INTERVAL_MS = 3 * 60 * 1000;
```
tzjames: Actually thinking of it further we might as well set this to 9 minutes.  The default lock timeout is 10 minutes, might as well make as few db calls as necessary.

packages/back-end/src/services/jobLifecycle.ts
```diff
@@ -0,0 +1,51 @@
...
+      const errorMsg = `Agenda job timed out after ${JOB_TIMEOUT_MS}ms: ${job.attrs.name}`;
+
+      stopTouch();
+      logger.error(new Error(errorMsg));
+      reject(new Error(errorMsg));
+    }, JOB_TIMEOUT_MS);
+  });
+
+  try {
+    // NB: This does not mean the job will stop executing, if it has side effects it will still happen
```
tzjames, line 50: good point.  I'd maybe rephrase it slightly for clarity:
```
// Note: If the timeoutPromise finishes first, it does not stop the job from continuing to execute.  However it will mark the job as failed and log an error, which frees up the "slot" for another job to run, in case the defaultLockLimit is reached, and also prevents other jobs from picking it up unless they have retry logic.
```
Though.. now if we add the comment above the function then it is not needed here.

packages/back-end/src/services/jobLifecycle.ts
```diff
@@ -0,0 +1,51 @@
+import { Job, JobAttributesData } from "agenda";
+import { logger } from "back-end/src/util/logger";
+import { MAX_QUERY_TIMEOUT_MS } from "back-end/src/util/secrets";
+
+const TOUCH_INTERVAL_MS = 3 * 60 * 1000;
+const JOB_TIMEOUT_MS = MAX_QUERY_TIMEOUT_MS + 1000; // Allow some buffer for the query client to close properly after it times out
+
+// As some functions may take a while to run, this will update the job's lock if the process is still running
```
tzjames: We could add to the end of this comment some more comments:
```
.
//This prevents the lockLifetime being reached as long as the job is running, and hence stops other servers from picking up the job.
//This also adds a timeout which allows the job to keep running but marks it as failed, which frees up the "slot" for another job to run, in case the defaultLockLimit is reached, and also prevents other jobs from picking it up unless they have retry logic.

File: packages/back-end/src/services/jobLifecycle.ts [hunk 5]
```diff
@@ -0,0 +1,51 @@
...
+  fn: (job: Job<T>) => Promise<void>
+) => async (job: Job<T>) => {
+  let touchTimer: NodeJS.Timeout | null = null;
+  let timeoutTimer: NodeJS.Timeout | null = null;
+  let finished = false;
+
+  function startTouch() {
+    touchTimer = setInterval(() => {
+      if (!finished) {
+        job.touch().catch((e) => {
```
tzjames, line 19: We could add a 
```
logger.info(`Updating the lock for a long running job ${job.attrs.name} ${job.attrs._id}`);
```

packages/back-end/src/services/jobLifecycle.ts
```diff
@@ -0,0 +1,51 @@
...
+          logger.error(e, `Failed to touch Agenda job ${job.attrs.name}`);
```
tzjames: These are not the correct arguments for logger.error.  It really only outputs the first arg as a message.  So I think we intend for this to be: 
```
logger.error(`Error while trying to touch Agenda job ${job.attrs.name}: ${e}`);
```

tzjames (approved):
I added a bunch of small points, but overall looks great.
I liked how you overrode agenda.define.

packages/back-end/src/services/jobLifecycle.ts
```diff
@@ -0,0 +1,51 @@
...
+  function stopTouch() {
+    finished = true;
+    if (touchTimer) clearInterval(touchTimer);
+  }
+
+  startTouch();
+
+  const timeoutPromise = new Promise<never>((_, reject) => {
+    timeoutTimer = setTimeout(() => {
+      const errorMsg = `Agenda job timed out after ${JOB_TIMEOUT_MS}ms: ${job.attrs.name}`;
```
tzjames: Add `${job.attrs._id}` to the error message so we can easily go from an error in the logs to the db to look up more details about the job that timed out.

ahdriel: Should we render all of the jobs attributes instead? Or should we keep this message compact just with the job id and the `starting job` log has all the attributes logged?

tzjames: sure I guess that might save us a little bit of time at the cost of a few bytes of storage.

tzjames: so should be worth it.

tzjames (approved):

Issue #2096, opened 2024-02-02
Allow passing headers/user-agent for proxy requests


biku:
### Summary

Proxy to API check in Web UI fails

#### Expected Behavior

pressing re-check should change to green and show "connected"

#### Current Behavior

It does not work in our self hosted setup, which runs on AWS in kubernetes. The default WAF requires an **user-agent** be passed, otherwise we get 
```
{
    "status": 200,
    "result": {
        "status": 403,
        "body": "<html>\r\n<head><title>403 Forbidden</title></head>\r\n<body>\r\n<center><h1>403 Forbidden</h1></center>\r\n</body>\r\n</html>\r\n",
        "error": "Proxy healthcheck returned a non-successful status code",
        "version": "",
        "url": "https://our-proxy-url/healthcheck"
    }
}
```

It probably can be done globally to have a user-agent header always be included in all requests. 
For this particular problem we found that inserting
```
headers: { "user-agent": "growthbook" }
``` 
[here](https://github.com/growthbook/growthbook/blob/a1c63b864a73340d5856ec3b761cf7dde1faf6c5/packages/back-end/src/models/SdkConnectionModel.ts#L372) works

PR #4296, opened 2025-07-15
feat(http): Add User-Agent header to fetch requests

nodirnasirov:
### Features and Changes
As an enterprise user of GrowthBook, we need the User-Agent header in each request. This is a requirement that we cannot work around. The User-Agent header has been added to all outgoing requests made by the backend, ensuring compliance with enterprise standards.

- Added User-Agent header to all requests in the backend, specifically in the testProxyConnection function.
- Closes https://github.com/growthbook/growthbook/issues/2096

### Dependencies

None

### Testing

- Verify that all outgoing requests from the backend include the User-Agent header.
- Test the testProxyConnection function to ensure the header is correctly set and does not interfere with the request functionality.
- Run integration tests to confirm no regressions in proxy connection handling.

### Screenshots
None

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -205,6 +205,27 @@
         "./packages/back-end/src/scripts/**/*",
         "./packages/back-end/**/*.test.{ts,tsx,js,jsx}"
       ]
+    },
+    {
+      "rules": {
+        "no-restricted-imports": [
+          "error",
+          {
+            "paths": [
+              {
+                "name": "node-fetch",
+                "message": "Use `import { fetch } from \"back-end/src/util/http.util\";` instead.",
+                "importNames": ["default"]
+              }
+            ]
+          }
+        ]
+      },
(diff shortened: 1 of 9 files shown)


ahdriel:
@nodirnasirov Thank you for the contribution!

I have updated the PR to make your changes more broadly applicable to all requests made by our backend application.

I asked for a coworker to review our changes and as soon as it is approved I'll merge it.

mknowlton89 (approved):
Looks good - tested locally and it works as intended.

The only concern I have is that this could introduce a breaking change if other self-hosted orgs were relying on the user-agent being set to a different value as this PR will now overwrite any existing user-agent.

bryce-fitzsimons:
Just a thought, should the UA be an environment variable so at least self-hosted users can undo this change?

ahdriel:
@bryce-fitzsimons Good suggestion! Done

PR #4300, opened 2025-07-16
Add query_timeout to Postgres

ahdriel:
### Features and Changes

We are adding a 1 hour default timeout for Postgres queries -- with a fallback that will release the client in case the server does not respect the timeout.


tzjames (approved):

lukesonnet:
FYI, In some other cases, we do have a UI-exposed setting in a form for timeouts: https://github.com/growthbook/growthbook/blob/db62d7cff7d0f4cef9cf307bf16a1cd26a6e3eef/packages/front-end/components/Settings/PrestoForm.tsx#L138-L147

PR #4303, opened 2025-07-16
Create ability to run external python server

tzjames:
### Features and Changes

When python is crunching numbers it can send CPU skyrocketing which then effects the speed we can return other requests.  This PR sends python requests to an external server if `EXTERNAL_PYTHON_SERVER_URL` is set. If `PYTHON_SERVER_MODE` is set then the server runs only the `/stats` route which passes the input to the python engine and returns.  That will allow us to create a micro-service for handling python requests which we can independently provision and scale.  Our main app then will be able to have less memory and probably can get by with fewer servers.

- [ ] Add datadog monitoring for the Pool and the python jobs
- [ ] Add env var for size of pool

### Testing

1. `yarn dev`
2. Evaluate the sample experiment dataset. 
3. Copy the html output of the page.
4. kill the server
5. Run in one terminal `PORT=3999 PYTHON_SERVER_MODE=true yarn dev:back-end`
6. In another terminal run `EXTERNAL_PYTHON_SERVER_URL=http://localhost:3999 yarn dev`
7.  Evaluate the sample experiment dataset. 
8. Copy the html output of the page
9. Compare the two html outputs and see no difference.
10. Kill the python server.
11.  Evaluate the sample experiment dataset. 
12. See the error:
Screenshot 2025-07-16 at 5 41 47 PM
13. RUN `PYTHON_SERVER_INPUT_SIZE_LIMIT=1b EXTERNAL_PYTHON_SERVER_URL=http://localhost:3999 yarn dev`
14. Evaluate the sample experiment dataset. 
15. See the error:
Screenshot 2025-07-16 at 5 50 30 PM
16. Change the code to send to the server `{test: 1}` instead of `statsData`
17. Evaluate the sample experiment dataset. 
18. See the error:
Screenshot 2025-07-16 at 5 38 34 PM

### DIFF
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -7,6 +7,7 @@ import cors from "cors";
 import asyncHandler from "express-async-handler";
 import compression from "compression";
 import * as Sentry from "@sentry/node";
+import { stringToBoolean } from "shared/util";
 import { populationDataRouter } from "back-end/src/routers/population-data/population-data.router";
 import decisionCriteriaRouter from "back-end/src/enterprise/routers/decision-criteria/decision-criteria.router";
 import { usingFileConfig } from "./init/config";
@@ -91,7 +92,7 @@ const informationSchemasController = wrapController(
 
 import { isEmailEnabled } from "./services/email";
 import { init } from "./init";
-import { getCustomLogProps, httpLogger } from "./util/logger";
+import { getCustomLogProps, httpLogger, logger } from "./util/logger";
 import { usersRouter } from "./routers/users/users.router";
 import { organizationsRouter } from "./routers/organizations/organizations.router";
 import { uploadRouter } from "./routers/upload/upload.router";
@@ -122,6 +123,7 @@ import { findOrCreateGeneratedHypothesis } from "./models/GeneratedHypothesis";
 import { getContextFromReq } from "./services/organizations";
 import { templateRouter } from "./routers/experiment-template/template.router";
 import { safeRolloutRouter } from "./routers/safe-rollout/safe-rollout.router";
+import { runStatsEngine } from "./services/stats";
 
 const app = express();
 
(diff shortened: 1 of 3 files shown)


packages/back-end/src/services/python.ts
```diff
@@ -17,7 +17,9 @@ type PythonServerResponse<T> = {
 // The stats engine usually finishes within 1 second
 // We use an overly conservative timeout to account for high load
 const STATS_ENGINE_TIMEOUT_MS = 60_000;
-const MAX_POOL_SIZE = 4;
+const MAX_POOL_SIZE = process.env.GB_STATS_ENGINE_POOL_SIZE
+  ? Number(process.env.GB_STATS_ENGINE_POOL_SIZE)
+  : 4;
```
ahdriel, line 24: nit: I think the other env variables we have are prefixed with the full `GROWTHBOOK`

packages/back-end/src/app.ts
```diff
@@ -133,763 +135,819 @@ if (SENTRY_DSN) {
...
+  app.use(cookieParser());
 
-// Health check route (does not require JWT or cors)
-app.get("/healthcheck", (req, res) => {
-  // TODO: more robust health check?
-  res.status(200).json({
-    status: 200,
-    healthy: true,
+  // Health check route (does not require JWT or cors)
+  app.get("/healthcheck", (req, res) => {
```
ahdriel: question: Should we keep the healthcheck endpoint for the python server as well?

packages/back-end/src/app.ts
```diff
@@ -133,763 +135,819 @@ if (SENTRY_DSN) {
...
+  app.use(
+    bodyParser.json({
+      limit: process.env.PYTHON_SERVER_INPUT_SIZE_LIMIT || "50mb",
+    })
+  );
+  app.use(compression());
+  app.use(httpLogger);
+  app.post("/stats", async (req, res) => {
+    try {
+      const results = await runStatsEngine(req.body);
```
ahdriel: issue: I think we need to validate the payload here to ensure it matches what we expect

packages/back-end/src/services/python.ts
tzjames: we are a bit inconsistent here.  But just further down this file we already have `GB_ENABLE_PYTHON_DD_PROFILING`. If two letters are good enough for Datadog, it should be good enough for Growthbook too.

packages/back-end/src/app.ts
tzjames: Yes good catch.  I think it would be good to have it.

packages/back-end/src/app.ts
tzjames: This will not be accessible from the internet.  Also stats_server.py does validating.

packages/back-end/src/services/python.ts
ahdriel: 😂  truer words have never been typed -- hoping we can consolidate on a pattern for our env variables in the near future

jdorn: Should we change this to just use the number of CPU cores instead? Do you think there's a benefit to controlling it independently from that?

packages/back-end/src/services/stats.ts
```diff
@@ -202,14 +202,37 @@ export function getBanditSettingsForStatsEngine(
...
+        },
+        body: JSON.stringify(statsData),
+      }
+    );
+    if (!retVal.ok) {
+      let { error } = await retVal.json();
+      if (!error) {
+        error = `Stats server errored with: ${retVal.status} - ${retVal.statusText}`;
+      }
+      logger.error("Error fetching from stats engine:", error);
```
jdorn: I think we normally pass an Error object here as the 2nd argument, not a string. Probably doesn't matter too much though.

packages/back-end/src/services/python.ts
tzjames: One benefit is that it makes it slightly easier in terraform to calculate the autoscale threshold. Another might be that if we ever have it back on the api-server we might want it to be # cores - 1 or -2 to account for the app and api processes.  self-hosted people could also theoretically want to specify how many they want to use.  They might have 1 core, but still want multiple to run at the same time?  I guess that would just slow them both down to take twice as long because there is probably not much sleeping time in that script.    Anyhow I don't think it hurts to have separate control. In terraform we can always calculate the one from the other so we change it in only one place.

packages/back-end/src/app.ts
```diff
@@ -145,6 +147,27 @@ if (ENVIRONMENT !== "production") {
   app.set("json spaces", 2);
 }
 
+if (stringToBoolean(process.env.PYTHON_SERVER_MODE)) {
+  // increase max payload json size to 50mb as a single query can return up to 3000 rows
+  // and we pass the results of all queries at once into python
+  app.use(
+    bodyParser.json({
+      limit: process.env.PYTHON_SERVER_INPUT_SIZE_LIMIT || "50mb",
```
ahdriel: question: Should we increase the payload limit just to the `/stats` endpoint instead?

`app.post('/stats', bodyParser.json({...}), async (req, res) => {...})`;

packages/back-end/src/app.ts
```diff
@@ -145,6 +147,27 @@ if (ENVIRONMENT !== "production") {
...
+    })
+  );
+  app.use(compression());
+  app.use(httpLogger);
+  app.post("/stats", async (req, res) => {
+    try {
+      const results = await runStatsEngine(req.body);
+      res.status(200).json({ results });
+    } catch (error) {
+      logger.error("Error running stats engine:", error);
```
ahdriel: Pino actually expects the error (if existing) as the first argument -- I have a separate PR fixing a lot of cases where we are not doing that, but we should update it here

```suggestion
      logger.error(error, "Error running stats engine");
```

ahdriel (approved):

PR #4305, opened 2025-07-17
Add datadog and cloudwatch metrics for python server

tzjames:
### Features and Changes

As we plan to break out python to its own microservice, it would be good to have accurate sense of how the python servers are doing.  This PR adds datadog metrics for how individual python calls go, and also on the pool size.  It also, if on cloud, will send cloudwatch the pool_size metric that we can use to autoscale the python servers.

pool_size is used instead of pool_busy because we only send the metric once a minute, and there is a good chance that there are no python processes running at that instant, but were in the recent past.  generic-pool will automatically increase the pool size if needed, and only decreases it if it has been idle for a minute so pool_busy should show the maximum number of busy servers in the last minute.  We would then be able to scale if the pool_size >= MAXIMUM_POOL_SIZE *0.75.

### Testing

`yarn build`
`. $(cd packages/stats && poetry env info --path)/bin/activate && yarn start:with-datadog`
Kick off a number of experiment results processings.
See the values in [datadog](https://us5.datadoghq.com/metric/explorer?fromUser=true&start=1752750383288&end=1752751152262&paused=true#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGABuMMjyWExSAHQaCJ4aTpkYaGhtdjhoFXBQwABUPAAEAEZYY8CMOM08DRhtmfgiCAAUAJQgfKCR0bmxWABMicktaRnZyvmFxSDH5VU1dXMiTS0W7Z1oTlB4bSeSQiJyqDRLKojcZTGZzBa7Cj7KI5R6lLAAZnOKSugJuuTuECKiSeuUq1Vq9Q+zVaTgcfTSEAAXnBoZNprN5ggeIjkYc0TosAAWbGXdJ41GE4klQXPClvRo0770v4TPAaLBs2GchE8AC6VFc7jwmFC4SNqhNGBi6ISexAFqtNsFZ3tjswzriWLdbktHqO6JF+qozSwaByoHknUQCFRUBwMHapo0RMSaFEcCcckUynS6agaYzTnoTGUIl9nURSQg9kwWCzClR6ZESn1PD4IFr4gAwlJhDAUCITWgeEA).

### DIFF
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -177,7 +177,6 @@ class PythonStatsServer<Input, Output> {
           );
           clearTimeout(timer);
 
-          // Datadog metrics
           metrics.getCounter("python.stats_calls_resolved").increment();
           metrics
             .getHistogram("python.stats_call_duration_ms")
@@ -192,7 +191,6 @@ class PythonStatsServer<Input, Output> {
           );
           clearTimeout(timer);
 
-          // Datadog metrics
           metrics.getCounter("python.stats_calls_rejected").increment();
           metrics
             .getHistogram("python.stats_call_duration_ms")
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -2,11 +2,13 @@ import { ChildProcess, spawn } from "child_process";
 import os from "os";
 import path from "path";
 import { randomUUID } from "crypto";
+import { CloudWatch } from "aws-sdk";
 import { createPool } from "generic-pool";
 import { MultipleExperimentMetricAnalysis } from "back-end/types/stats";
 import { logger } from "back-end/src/util/logger";
 import { ExperimentDataForStatsEngine } from "back-end/src/services/stats";
-import { ENVIRONMENT } from "back-end/src/util/secrets";
+import { ENVIRONMENT, IS_CLOUD } from "back-end/src/util/secrets";
+import { metrics } from "back-end/src/util/metrics";
 
 type PythonServerResponse<T> = {
   id: string;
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/python.ts
```diff
@@ -176,6 +191,13 @@ class PythonStatsServer<Input, Output> {
             reason
           );
           clearTimeout(timer);
+
+          // Datadog metrics
```
ahdriel: nit: This interface also works with OTEL -- so should we remove it?

packages/back-end/src/services/python.ts
```diff
@@ -207,6 +229,67 @@ export const statsServerPool = createPool(
...
+          ],
+        },
+      ],
+    });
+  } catch (error) {
+    // When not running on AWS, no need to publish to cloudwatch or warn us every ten seconds.
+  }
+}
+
+function monitorServicePool() {
```
ahdriel, line 252: question: All of these pool metrics are per-process right?

Do they get tagged automatically with the task or machine identifier?

ahdriel (approved):

packages/back-end/src/services/python.ts
tzjames: These pool metrics are per server that is running the pool.  Unfortunately they don't get tagged with the task or machine identifier. Or perhaps fortunately, because then datadog counts it as one custom metric per machine name which has over a month a fairly high cardinality.

PR #4306, opened 2025-07-17
Updates SqlExplorer validation

mknowlton89:
### Features and Changes

While debugging a recent user issue, I noticed that we were using different validation for the `EditSqlModal` and the `SqlExplorerModal`.

In the `EditSqlModal` we were using the `validateSql` function, whereas in `SqlExplorerModal` we were using a one-off `isReadOnlyQuery` helper.

I originally noticed this as within `EditSqlModal` we have a check to throw an error if the query contains a trailing comma, but that was missing in the `SqlExplorerModal`.

As a result, I've currently opted to align both modals so they use the same `validateSql` helper. `validateSql` is much more restrictive than `isReadOnlyQuery` as `isReadOnlyQuery` accepts queries like `EXPLAIN`.  

If we think orgs will want and benefit from the less restrictive `isReadOnlyQuery` check, I can simply update it to also check for the trailing comma, otherwise we can align both modals on the `validateSql` helper.

The tertiary benefit of using `validateSql` is if we eventually want to support required columns in the `EditSqlModal`, `validateSql` has that logic built in.

### Dependencies

None

### Testing

- [x] Ensure an error is thrown if the query doesn't start with `SELECT`
- [x] Ensure an error is thrown if the query contains a trailing comma

### DIFF
--- a/packages/back-end/src/services/datasource.ts
+++ b/packages/back-end/src/services/datasource.ts
@@ -1,5 +1,4 @@
 import { AES, enc } from "crypto-js";
-import { isReadOnlySQL } from "shared/sql";
 import { ENCRYPTION_KEY } from "back-end/src/util/secrets";
 import GoogleAnalytics from "back-end/src/integrations/GoogleAnalytics";
 import Athena from "back-end/src/integrations/Athena";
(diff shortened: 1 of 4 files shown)


jdorn:
The validation requirements are different between the modals.  We want something really restrictive when they are editing a metric or experiment query.  But with SQL explorer, we need to allow a much wider variety.  An example is using CTEs (`WITH`) - will break a metric query, but is super useful in the SQL explorer.

Also, the `ensureLimit` function removes trailing semicolons already before running the query, so I don't think this is a real issue.

PR #4308, opened 2025-07-17
Some fixes for python monitoring stats

tzjames:
### Features and Changes

A few of the events were hallucinations by Copilot.  There are just two that are emitted according to https://www.npmjs.com/package/generic-pool.

The cloudwatch metric doesn't seem to be populated yet.  Also the env var for the task doesn't exist - we could get it by making a fetch, but I don't think it is necessary for our purposes.  Will output the error on prod in order to debug.

### DIFF
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -235,17 +235,19 @@ function publishPoolSizeToCloudWatch(value: number) {
       MetricData: [
         {
           MetricName: "PoolSize",
-          Timestamp: new Date(),
           Value: value,
           Unit: "Count",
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/python.ts
```diff
@@ -235,17 +235,19 @@ function publishPoolSizeToCloudWatch(value: number) {
       MetricData: [
         {
           MetricName: "PoolSize",
-          Timestamp: new Date(),
           Value: value,
```
ahdriel, line 238: As you pointed out, we might need to figure out the permissions so this change works.

PR #4309, opened 2025-07-17
Refactor InfoSchema path generation to the front end

mknowlton89:
### Features and Changes

This PR makes a few changes to our `InformationSchemas`.

Most notably, it moves the `path` generation for an information schemas parts (e.g. catalog, schemas, tables, etc) from the backend when we're building the information schema to the front end, and it's now done in a just in time manner rather than persisting the path within the db.

This simplifies the `InformationSchema` generation and allows us to make changes to the path logic without requiring the user to refresh the information schema or table object.

Additionally, we were previously generating the path for each level of an information schema object (the catalog, schema level, table level and even column level) but we've only ever used the path for tables. So we've removed the path object from all levels other than the table. This allows us to greatly simplify the logic of the path generation.

I've also added a new suite to tests for the `getTablePath` so that we can verify the logic is working as expected for different data source types.

This PR came about because of a bug where the path for tables on our Managed Warehouse Data Sources was incorrect. And in testing, I found a bug with MySql/MariaDB data sources where if any part of the path contained a special character (e.g. a dash) the path formatting was incorrect and needed to be wrapped in backticks.

### Dependencies

- None

### Testing

- [x] Ensure double clicking on a table name for a MySQL data source updates the query and when ran, works successfully, even if the db or table contains a special character like a dash.
- [x] Ensure double clicking on a table name for a Clickhouse data source updates the query and when ran, works successfully, even if the db or table contains a special character like a dash. (I haven't tested this in a browser, but via the test suite itself)


packages/back-end/src/util/informationSchemas.ts
```diff
@@ -17,17 +17,24 @@ type RowType = {
 export function getPath(dataSourceType: DataSourceType, path: RowType): string {
   const pathArray = Object.values(path);
   const returnValue = pathArray.join(".");
+
   switch (dataSourceType) {
-    // MySQL only supports path's that go two levels deep. E.G. If the full path is database.schema.table.column, it only supports table.column.
+    // MySQL and ClickHouse both support paths that go two levels deep
+    // Backticks help avoid issues with reserved words or special characters
     case "mysql":
+    case "clickhouse":
```
jdorn: The managed warehouse has a different type - `growthbook_clickhouse`.  This change is fine for mysql and regular clickhouse.  For the managed warehouse, we don't even want the db part.  Just the table name by itself.  So it should output `events`, not `org_abc123.events`.

mknowlton89: Sounds good. And we control the table names completely, correct?

So we shouldn't have to worry about the table containing special characters or reserved words and thus don't need to wrap the output in backticks

packages/back-end/src/util/informationSchemas.ts
```diff
@@ -17,17 +17,26 @@ type RowType = {
...
+          .slice(1)
+          .map((part) => "`" + part + "`") // Wrap each path part in backticks for safety
+          .join(".");
       }
 
     case "bigquery":
-      return "`" + returnValue + "`"; // BigQuery requires backticks around the path
+      return "`" + returnValue + "`"; // BigQuery requires backticks around the full path
+    case "growthbook_clickhouse":
+      return pathArray[pathArray.length - 1];
```
jdorn: This whole function is very fragile the way it is written.  It depends on the order of object keys, which is bad.  This `growthbook_clickhouse` case doesn't handle column paths, only table paths.

It needs a rewrite and a suite of unit tests.

```ts
expect(
  getPath("growthbook_clickhouse", {
    tableCatalog: "catalog";
    tableSchema: "schema";
    tableName: "table";
    columnName: "column";
  })
).toBe("table.column");
```

mknowlton89: I'm working through a refactor right now.

@jdorn Looking through it - when building a informationSchema we're calling `getPath` at the `db` level, the `schema` level, and the `table` level, but we're only using the path for tables. 

We're also building the path for each column in an `InformationSchemaTable` record.

I cant think of any uses for the path outside of the table. We currently use it to power the double click feature in the SchemaBrowser. and it'll eventually be used in the auto-complete suggestion feature.

Is there a reason im not seeing?  If not, we could deprecate the path for all but the `tables` and it could vastly simplify the logic of get path.

jdorn: Yeah, I think that's fine. I'd also be in favor of not storing the path in Mongo and generating it on demand when we need it.

mknowlton89: Yea, I think that is the ideal solution - eliminates the need for orgs to refresh their schemas/tables to get the new path. I think that + only generating the path for tables has greatly reduced the complexity.

Just pushed up changes that do both of those tasks (move the generation to the front end and eliminates the path for databases, schemas, and columns, and just builds it for tables.

ahdriel (approved):
Thanks for the improvements! I have tested it and worked as expected now 🎉

PR #4325, opened 2025-07-21
Improve error monitoring for GrowthBook Cloud

ahdriel:
### Features and Changes

For Cloud we use `Sentry` as our error monitoring tool (and self-hosted can also use our hooks by providing their own Sentry DSN if desired).

This PR improves the integration by:
* Reporting the proper environment
* Handling server-side rendering
* Creating a release and uploading source maps

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,10 +22,10 @@ jobs:
       id-token: write
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Login to Docker Hub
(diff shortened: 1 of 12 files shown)


bryce-fitzsimons:
Lot of changes in package.json. Are those all consequences of adding `@sentry/nextjs`?

ahdriel:
@bryce-fitzsimons Indeed they are!

They are mostly additive though, so the risk is low as we are not changing versions of libraries already in use.

Checking the bundle size it ends up adding 20kb to the _app chunk.

bryce-fitzsimons (approved):
approved. keep an eye on error monitoring when we go live though

PR #4326, opened 2025-07-21
Adds Big Number viz type to SqlExplorer

mknowlton89:
### Features and Changes

This PR introduces a new visualization type to the Sql Explorer that simply renders a large number. Please see the screenshots below.

This visualization does introduce a new pattern - in that its a visualization that doesn't require both an x and a y axis. 

## To Dos / Questions
- [x] It feels like we should probably give orgs the specify if the value column is a currency, and if so, which currency, and render the currency symbol in the visualization.
- [x] ~~It feels like we should also give orgs the ability to specify the number of decimal points rendered (regardless of it is a currency figure or just a number)~~
- [ ] Currently, I'm storing the `value` for the visualization in the y-axis, and it all works, but I'm curious to hear opinions from others on if we should instead update the `dataVizConfigValidator` to support visualisations with and without axis. So instead of storing the `value` in the y-axis, we'd add a new `value` property (and probably some sort of configuration property that allows orgs to specify currency, decimal places, etc)

- Closes - None

### Dependencies

- None

### Testing

- [x] Ensure all of the form validation works as expected with the new single axis.
- More test steps TBD based on To Dos and Questions above

### Screenshots

Screenshot 2025-08-04 at 7 01 06 AM

Screenshot 2025-08-04 at 7 29 17 AM

Screenshot 2025-08-04 at 10 51 38 AM

Screenshot 2025-08-04 at 10 51 17 AM

Screenshot 2025-08-04 at 10 51 24 AM

Screenshot 2025-08-04 at 10 51 31 AM


packages/back-end/src/validators/saved-queries.ts
```diff
@@ -50,13 +50,25 @@ export type dimensionAxisConfiguration = z.infer<
...
+      .nonempty()
+      .optional(),
+  }),
+  // Big value chart type that doesn't need xAxis
+  z.object({
+    title: z.string().optional(),
+    chartType: z.literal("big-value"),
+    yAxis: z.array(yAxisConfigurationValidator).nonempty(),
+  }),
+]);
```
mknowlton89: @ahdriel Working on adding a new "Big Number" data viz type (similar to Mode's Big Number) viz.

It doesn't look like EChartsReact has anything similar to this (maybe a `gauge`), but it would require a lot of weird customization.

One big difference here is that there is no `xAxis` as it's not a chart. I've updated the types here to use a discriminatedUnion, but curious if you have opinions/ideas on a different way to handle this.

It's made some checks on the front end a bit painful and I've had to introduce some logic like `requiresXAxis` and `supportsDimensions`.

Curious if you have thoughts/ideas or if you think this is the best approach.

ahdriel: I do like your approach! I am curious if the painfulness that we are feeling on the front-end is because we need to do a bigger refactor to more clearly separate our types of "charts" from echarts.

Meaning that ideally we would have a `switch(chartType)` and then TS would infer that we either have an xAxis or not depending on the type.

I'll explore it further tomorrow to see if we can make things easier on ourselves.

mknowlton89: @ahdriel Made a few changes in this [commit](https://github.com/growthbook/growthbook/pull/4326/commits/74e418ad8d7974c15b9023322c532d5a93d23fa2) to simplify some things on the front end. Let me know if this is the direction you were thinking.

packages/back-end/src/validators/saved-queries.ts
```diff
@@ -50,14 +50,59 @@ export type dimensionAxisConfiguration = z.infer<
   typeof dimensionAxisConfigurationValidator
 >;
 
-export const dataVizConfigValidator = z.object({
+const chartTypesThatRequireXAndYAxis = z.enum([
```
lukesonnet: nit: I feel like this is too restrictive of a design, since we may have multiple y-axes in the future. Is there not a way to just have the axes defined on the chart type directly instead of having a group for each type of chart depending on the number/type of axes it has?

lukesonnet:
> Currently, I'm storing the value for the visualization in the y-axis, and it all works, but I'm curious to hear opinions from others on if we should instead update the dataVizConfigValidator to support visualisations with and without axis. So instead of storing the value in the y-axis, we'd add a new value property (and probably some sort of configuration property that allows orgs to specify currency, decimal places, etc)

TBH feels fine to have it be the `yAxis`, or maybe it's something like the `primaryAxis`. But I don't think it's a big deal. But maybe the Axis has certain formatting values (e.g. `value` and `ticks` and `label` that all can be formatted and styled separately).

packages/back-end/src/validators/saved-queries.ts
mknowlton89: @lukesonnet  Are you thinking something more like [this](https://github.com/growthbook/growthbook/pull/4326/commits/0f53a4a7bedf5aef6c2a7cf3b8d5ce0098434838)?

Or, alternatively, make the `chart` an object that has a type field, along with axis' info, and additional properties (dimensions, format, etc)?

lukesonnet: I think more like the latter. I do like the linked commit a bit more, but ideally the devex could be something like when you go to create a chart it automatically can update `requiresXAxis` because `requresXAxis` is just `true` if the config contains a mandatory xaxis for any chart configs that may exist.

But I defer to you and Adriel here, tbh. It's just nice if there's a base chart that kind of can handle this logic and then the dev ex is just to create a new chart type and not worry about updating ancillary methods; they just have to focus on the config.

mknowlton89: Cool - I felt that was as well after I made the last changes but wanted to confirm.

Just pushed some more updates in this [commit](https://github.com/growthbook/growthbook/pull/4326/commits/e85562b0fff1e499a1f89904da839301a503346a) that defines a base chart that can be extended. I've also updated the helper functions to use the chart's config object directly so we don't have to manually update any of the helpers when creating a new chart type.

lukesonnet: Awesome, yeah I like the future dev ex of this approach. I'll leave it to Adriel probably for the full review?

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -128,21 +130,24 @@ export function DataVisualizationDisplay({
   dataVizConfig: Partial<DataVizConfig>;
 }) {
   const isConfigValid = useMemo(() => {
-    const parsed = dataVizConfigValidator.strip().safeParse(dataVizConfig);
+    const parsed = dataVizConfigValidator.safeParse(dataVizConfig);
```
ahdriel, line 133: question: Why are we removing the `strip()`?

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
```diff
@@ -120,368 +122,501 @@ export default function DataVizConfigPanel({
...
-              setValue={(v) => {
-                if (!v || !dataVizConfig.xAxis) return;
-                onDataVizConfigChange({
-                  ...dataVizConfig,
-                  xAxis: {
-                    ...dataVizConfig.xAxis,
-                    type: v as "string" | "number" | "date",
+      {dataVizConfig.chartType === "big-value" ? (
+        <>
+          <Select
```
ahdriel, line 141: Can we update the new Selects to also have `size="2"`? I think for secondary configuration (basically everything except Name & Graph Type) use the smaller size

ahdriel (approved):

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
mknowlton89: We're now using the `discriminatedUnion` which doesn't play well with `.strip()` and AFAIK, discriminatedUnion has it's own built-in validation.

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
mknowlton89: Yes, great call.

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
ahdriel: My concern is that we might fail validation when moving from a config that has dimension to one that doesn't and we don't completely reset the object.

When testing this worked as expected though, so I think my concern might be unfounded.

PR #4330, opened 2025-07-21
Attempt v3 ot sentry-release action

ahdriel:
### Features and Changes

After looking more into the logs and debugging the v1 was just failing silently but was not working as expected, so we are using v3 again & trying to fix the error by adding a trailing slash to indicate it is a directory.


bryce-fitzsimons (approved):
approved, please keep an eye on deployment

PR #4336, opened 2025-07-22
Set the default lock limit as an env var

tzjames:
### Features and Changes
Our jobs servers run nothing else, so they can have a higher limit.

### DIFF
--- a/packages/back-end/src/services/queueing.ts
+++ b/packages/back-end/src/services/queueing.ts
@@ -10,7 +10,7 @@ export const getAgendaInstance = (): Agenda => {
     const config: AgendaConfig = {
       // @ts-expect-error - For some reason the Mongoose MongoDB instance does not match (missing 5 properties)
       mongo: mongoose.connection.db,
-      defaultLockLimit: 5,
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/queueing.ts
```diff
@@ -10,7 +10,7 @@ export const getAgendaInstance = (): Agenda => {
     const config: AgendaConfig = {
       // @ts-expect-error - For some reason the Mongoose MongoDB instance does not match (missing 5 properties)
       mongo: mongoose.connection.db,
-      defaultLockLimit: 5,
+      defaultLockLimit: Number(process.env.GB_DEFAULT_LOCK_LIMIT) || 5,
```
ahdriel: nit: Should we add `JOBS` to the variable name to be clearer about what it affects?

ahdriel (approved):

packages/back-end/src/services/queueing.ts
tzjames: huh, I had GB_AGENDA_DEFAULT_LOCK_LIMIT added.  I guess I hadn't saved it.

PR #4337, opened 2025-07-22
Allows user to remove 5 row limit in EditSqlModal

mknowlton89:
### Features and Changes

We got some feedback from users that the 5 row limit within the `EditSqlModal` was too limiting. This PR adds a checkbox to the `EditSqlModal` that allows orgs to disable the 5 row limit. If a user does this, we will still apply the 1000 row limit max to the query to prevent things breaking and to prevent the user from running really expensive queries. In an attempt to reduce confusion, I have added a tooltip around the checkbox that outlines a 1000 row limit will be applied if the box is unchecked.

### Dependencies

- None

### Testing

- [x] Ensure if the 5 row limit box is CHECKED, when you run `Test Query` we wrap the user's query with a `LIMIT 5` clause.
- [x] Ensure if the 5 row limit box is UNCHECKED, when you run the `Test Query` we wrap the user's query with a `LIMIT 1000` clause.
- [x] Ensure the user can not pass in a limit greater than 1000 - if this were to happen, ensure an error is thrown before we attempt to run the query.

### Screenshots
Screenshot 2025-07-22 at 2 36 58 PM

Screenshot 2025-07-28 at 11 57 14 AM


Kevin-Chant (approved):
LGTM

PR #4342, opened 2025-07-24
Fixed experiment date sort order when experiment dates are missing

Auz:


packages/front-end/services/experiments.ts
```diff
@@ -39,12 +39,23 @@ import { useOrganizationMetricDefaults } from "@/hooks/useOrganizationMetricDefa
...
 
+export function experimentDate(exp: ExperimentInterfaceStringDates): string {
+  return (
+    (exp.archived
+      ? exp.dateUpdated
+      : exp.status === "running"
+      ? exp.phases?.[exp.phases?.length - 1]?.dateStarted
+      : exp.status === "stopped"
+      ? exp.phases?.[exp.phases?.length - 1]?.dateEnded
+      : exp.dateCreated) ?? new Date().toISOString() // fallback to now
```
ahdriel, line 55: question: Is this the fix? And when does it happen that we don't have even `dateCreated` to fallback to?

Auz: its not super clear, but the parenthesis there is wrapped around the status check

ilyausorov:
Woo hoo

PR #4343, opened 2025-07-25
feat: add env var for python stats server minimum pool size

ari-party:
### Features and Changes

This PR introduces `GB_STATS_ENGINE_POOL_SIZE_MIN` environment variable, this sets the `min` property in the pool options [here](https://github.com/growthbook/growthbook/blob/adf3ade4f2ab3a51a6ea87351eb5c6ffba52e3d7/packages/back-end/src/services/python.ts#L226).

Similar to how the `GB_STATS_ENGINE_POOL_SIZE` environment variable defines the `max` property. (Introduced in #4303)

Currently, a stats server process is always started, unless `PYTHON_SERVER_MODE` is set. This occurs even if no features utilizing the stats server are used (e.g. experiments), leading to unnecessary consumption of system resources.

[generic-pool/node-pool](https://github.com/coopernurse/node-pool)'s [acquire](https://github.com/coopernurse/node-pool/blob/master/README.md#poolacquire) function correctly handles the `min` property if it is set to `0`, ensuring a stats server is only spun up when needed.

### DIFF
--- a/packages/back-end/src/server.ts
+++ b/packages/back-end/src/server.ts
@@ -1,5 +1,6 @@
-import "./instrumentation";
 import "./init/aliases";
+
+import "./instrumentation";
 import app from "./app";
 import { logger } from "./util/logger";
 import { getAgendaInstance } from "./services/queueing";
--- a/packages/front-end/components/Forms/CodeTextArea.tsx
+++ b/packages/front-end/components/Forms/CodeTextArea.tsx
@@ -1,9 +1,21 @@
 import dynamic from "next/dynamic";
-import { useEffect, useState, useRef } from "react";
+import { useEffect, useState, useRef, createElement } from "react";
 import type { Ace } from "ace-builds";
+import type { IAceEditorProps } from "react-ace";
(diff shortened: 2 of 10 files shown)


packages/back-end/src/services/python.ts
```diff
@@ -19,6 +19,9 @@ type PythonServerResponse<T> = {
 // The stats engine usually finishes within 1 second
 // We use an overly conservative timeout to account for high load
 const STATS_ENGINE_TIMEOUT_MS = 60_000;
+const MIN_POOL_SIZE = process.env.GB_STATS_ENGINE_POOL_SIZE_MIN
+  ? Number(process.env.GB_STATS_ENGINE_POOL_SIZE_MIN)
+  : 1;
```
jdorn, line 24: Will need validation to make sure:
1. It is a valid number (not `NaN`)
2. It is not negative
3. It is less than or equal to the max pool size

tzjames:
Thanks for your work here.  I'm taking this over and adding some validation here: https://github.com/growthbook/growthbook/pull/4361.

PR #4351, opened 2025-07-28
ClickHouse: Refactor extractJSONField to handle JSON type checks for better extraction logic

### LINKED ISSUE #4130: [Bug] Real JSON is not supported in ClickHouse
### Summary

[This](https://github.com/growthbook/growthbook/blob/00118a47d79db4838724b1927f8a9dcc2c27ed1c/packages/back-end/src/integrations/ClickHouse.ts#L119) implementation works only if the column has a String type, not the native JSON type.

Here are the new JSON docs for ClickHouse: https://clickhouse.com/docs/sql-reference/data-types/newjson

#### Expected Behavior

Distinguish between data types and use the correct functions for working with JSON depending on the type. It is important to note that this data type is relatively new, and many people continue to use strings to store JSON, so there is no need to completely remove the old method.

#### Current Behavior

Query error for metric with native JSON type.

denisov-vlad:
### Features and Changes

Added logic to determine the type of JSON column before retrieving values by key.

If JSON is stored as a string, the functions have been changed from `simpleJSONExtract*` to `JSONExtract`, which allows you to get the correct float value if an integer is requested.

- Closes #4130 

### Testing

1. Create fact table from simple query:

```
SELECT 
    '{"float": 1.0, "string": "string", "int": 123}' as json_string,
    json_string::JSON as json
```

2. Set both column types to JSON

3. Try to extract values.

### DIFF
--- a/packages/back-end/src/integrations/ClickHouse.ts
+++ b/packages/back-end/src/integrations/ClickHouse.ts
@@ -118,9 +118,21 @@ export default class ClickHouse extends SqlIntegration {
   }
   extractJSONField(jsonCol: string, path: string, isNumeric: boolean): string {
     if (isNumeric) {
-      return `simpleJSONExtractFloat(${jsonCol}, '${path}')`;
+      return `
+if(
+  toTypeName(${jsonCol}) = 'JSON', 
+  toFloat64(${jsonCol}.${path}),
+  JSONExtractFloat(${jsonCol}, '${path}')
+)
+      `;
     } else {
-      return `simpleJSONExtractString(${jsonCol}, '${path}')`;
+      return `
+if(
+  toTypeName(${jsonCol}) = 'JSON',
+  ${jsonCol}.${path}.:String,
+  JSONExtractString(${jsonCol}, '${path}')
+)
+      `;
     }
   }
 


packages/back-end/src/integrations/ClickHouse.ts
```diff
@@ -118,9 +118,21 @@ export default class ClickHouse extends SqlIntegration {
...
+  toFloat64(${jsonCol}.${path}),
+  JSONExtractFloat(${jsonCol}, '${path}')
+)
+      `;
     } else {
-      return `simpleJSONExtractString(${jsonCol}, '${path}')`;
+      return `
+if(
+  toTypeName(${jsonCol}) = 'JSON',
+  ${jsonCol}.${path}.:String,
```
jdorn, line 132: Is the dot before the colon a typo or intentional?

vladenisov: @jdorn that's clickhouse's syntax: https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns

> It is also possible to read sub-columns of a Dynamic type using the special syntax json.some.path.:TypeName:

jdorn: Ok, great. Just double checking since the syntax seemed a little unusual.

PR #4355, opened 2025-07-28
fix(sdk-webhooks): Use webhook secrets

ahdriel:
### Features and Changes

The recommended way of authenticating webhooks is via Webhook Secrets, but our SDK Connection flow was still hard-coding keys in the headers, meaning it could be seen by users of the platform after it had been set.

This fixes the flow by requiring the user to select an Existing Webhook Secret and injecting it into the headers automatically.

#### Before

It was a string field.

Screenshot 2025-07-28 at 4 34 52 PM

#### After

Screenshot 2025-07-28 at 4 34 38 PM

### DIFF
--- a/packages/front-end/components/EventWebHooks/WebhookSecretModal.tsx
+++ b/packages/front-end/components/EventWebHooks/WebhookSecretModal.tsx
@@ -9,9 +9,13 @@ import StringArrayField from "../Forms/StringArrayField";
 export default function WebhookSecretModal({
   existingId,
   close,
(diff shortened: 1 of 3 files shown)


jdorn (approved):
It would be nice if you could choose to create a new webhook secret directly within this flow instead of linking to a separate page (and losing everything else you've typed so far in the modal).

I'll approve this, but if it's quick I say we do the above now.

ahdriel:
@jdorn Done!

PR #4356, opened 2025-07-29
load secrets first

tzjames:
### Features and Changes

if IS_CLOUD=true and SSO_CONFIG={} in .env.local `yarn dev` will end up saying "No SSO Connection Configured" which is not true.  That is because the "./instrumentation" import in server.ts kicked off a series of imports that ended up loading secrets.ts after sso.ts, but secrets.ts needs to be loaded first to get `dotenv` loaded. 

### Testing
Set IS_CLOUD=true and SSO_CONFIG={} in .env.local
`yarn dev`
localhost:3000
See no error msg, but get taken to the SSO login page.

### DIFF
--- a/packages/back-end/src/api/api.router.ts
+++ b/packages/back-end/src/api/api.router.ts
@@ -5,7 +5,7 @@ import rateLimit from "express-rate-limit";
 import bodyParser from "body-parser";
 import * as Sentry from "@sentry/node";
 import authenticateApiRequestMiddleware from "back-end/src/middleware/authenticateApiRequestMiddleware";
-import { getBuild } from "back-end/src/util/handler";
+import { getBuild } from "back-end/src/util/build";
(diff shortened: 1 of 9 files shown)


tzjames:
> `secrets.ts` imports from shared which itself kicks off tons of imports from back-end. Feels like we do want instrumentation in place before we do that. Really, the only part we need is the dotenv setup. Maybe we should move that into a separate standalone init script instead.

Well I guess the real issue was that instrumentation was importing `getBuild` from `util/handler.ts` which was importing a bunch of other back-end files, including one that read from `sso.ts`.  I've moved getBuild to its own file that imports nothing else.  So theoretically we don't need to also move dotenv from secrets.ts.  I had already done it though and it does protect us from someone else putting another import above app.ts in server.ts or another import before secrets in app.ts.

ahdriel (approved):
Thanks for the fix!

PR #4357, opened 2025-07-29
(WIP) - Sql Explorer - add template variable support

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -1414,7 +1414,17 @@ export default abstract class SqlIntegration
 
   getFreeFormQuery(sql: string, limit?: number): string {
     const limitedQuery = this.ensureMaxLimit(sql, limit ?? SQL_ROW_LIMIT);
-    return format(limitedQuery, this.getFormatDialect());
+
+    const sqlContainsTemplateVariables =
+      !!sql.match(/\{\{\s*[^}]+\s*\}\}/g)?.length || false;
+
+    const queryToRun = sqlContainsTemplateVariables
+      ? compileSqlTemplate(limitedQuery, {
+          startDate: new Date(), // MKTODO: Update this logic, just a placeholder for now
+        })
+      : limitedQuery;
+
+    return format(queryToRun, this.getFormatDialect());
   }
 
   getTestQuery(params: TestQueryParams): string {


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1414,7 +1414,17 @@ export default abstract class SqlIntegration
 
   getFreeFormQuery(sql: string, limit?: number): string {
     const limitedQuery = this.ensureMaxLimit(sql, limit ?? SQL_ROW_LIMIT);
-    return format(limitedQuery, this.getFormatDialect());
+
+    const sqlContainsTemplateVariables =
```
mknowlton89, line 1418: Not sure if we even need this - we can just call compileSqlTemplate regardless

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1414,7 +1414,17 @@ export default abstract class SqlIntegration
...
+      !!sql.match(/\{\{\s*[^}]+\s*\}\}/g)?.length || false;
+
+    const queryToRun = sqlContainsTemplateVariables
+      ? compileSqlTemplate(limitedQuery, {
+          startDate: new Date(), // MKTODO: Update this logic, just a placeholder for now
```
mknowlton89, line 1423: Need to think this through and discuss with others. In the `EditSqlModal` it makes sense to require the `startDate`, but less so via the SqlExplorer.

Namely, with the sql explorer we can introduce date filtering in two places - in the sql, or in the visualizations (or both).

The benefit of doing it in the sql is that with a 1000 row limit, orgs will be less likely to bump into that limitation if they're filtering in the query vs getting everything and then filtering in the visualizations.

The benefit of doing it in the visualizations is that you can do it on a per viz basis, though, I guess that could lead to weird experiences if you're looking at two visualizations, but they're from different date ranges.

If we want to support the date filtering at the sql level (and want to support template variables, which seems like a great idea) we need to update the UI so users can add a start/end date.

mknowlton89: @lukesonnet Do you have opinions here on the above?

As I've been working through this it felt like we'd want to add  date filtering to the visualizations to give users more flexibility, but it could introduce weird issues with looking at visualizations from two different periods and also doesn't help with the 1000 row limit.

lukesonnet: I think that we will want to support more generic template variables as I wrote in slack, so I think a generic UI to pick a date in the SQL itself is correct. Additional filtering at the graph level also makes sense, but we should be prepared to force all template variables to be specified at the SQL level.

This UI would be useful in other places as well, whenever there's a template variable but no context to set it (e.g. for `{{ experimentId }}` or other generic template variables that come from custom experiment fields, a future direction we may want to take).

Specifically for dates, we probably want to let them set "last X days" type settings though, to keep the dashboard fresh, in addition to hardcoding dates.

mknowlton89:
Closing this as it's stale. I can re-open if this gets re-prioritized.

PR #4361, opened 2025-07-30
parse and validate python vars

tzjames:
### Features and Changes
We don't want python to run at all in some circumstances, hence it would be good to be able to set the minimum to 0.
Also the stats_engine has timed out on particularly big data a few times, so we are bumping it's default value and setting it to a env var that we can more easily control without redeploying.

### Dependencies

This takes over https://github.com/growthbook/growthbook/pull/4343/files

### Testing
``` 
parseEnvInt(-1, 4, {
  min: 1,
  name: "GB_STATS_ENGINE_POOL_SIZE",
})
```
See the warning and it return 4.

``` 
parseEnvInt(undefined, 4, {
  min: 1,
  name: "GB_STATS_ENGINE_POOL_SIZE",
})
```
See it return 4.

### DIFF
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -16,12 +16,44 @@ type PythonServerResponse<T> = {
   results: T;
 };
 
+function parseEnvInt(
+  value: string | undefined,
+  defaultValue: number,
+  opts?: { min?: number; max?: number; name?: string }
+): number {
+  const num = parseInt(value || "") || defaultValue;
+  if (
+    (opts?.min !== undefined && num < opts.min) ||
+    (opts?.max !== undefined && num > opts.max)
+  ) {
+    logger.warn(
+      `Invalid value for ${opts?.name || "environment variable"}: "${
+        value ?? ""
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/python.ts
```diff
@@ -16,12 +16,44 @@ type PythonServerResponse<T> = {
   results: T;
 };
 
+function parseEnvInt(
```
ahdriel, line 19: small issue: This will still return a value bigger or smaller than min/max if that's the `defaultValue`.

I guess it can also be the engineer responsibility to make sure that defaultValue falls within min & max.

ahdriel (approved):

packages/back-end/src/services/python.ts
tzjames: Yes.  It is not exported atm, and the default value for min will always be less than the max because max's minimum value is 1.  

We could have it have it set to the max or the minimum if the default is out of range, but then the default is not the default which is not intuitive.  So I left it up to developer to ensure it.

PR #4365, opened 2025-07-31
fix(timeseries): Fix bug when first data point is invalid

ahdriel:
### Features and Changes

As fixes to some graph inconsistencies I added a filter to TimeSeries in the API, to return only valid ones.

But this created a separate issue where we would not know that the record already existed in the DB and would attempt to create a new one, which would fail. Meaning that if the first data point added to a timeseries was invalid, then it would never be able to recover.

To fix it we are now:
- Returning all records, regardless of the status
- We are checking if the datapoints are valid when saving, and dropping if they are not
- Moved the checks to shared as we will have some records that are considered invalid for graphing purposes

### DIFF
--- a/packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
+++ b/packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
@@ -298,7 +298,8 @@ export default function ResultsTable({
   });
 
   const filteredMetricTimeSeries = useMemo(() => {
-    return filterInvalidMetricTimeSeries(metricTimeSeries?.timeSeries || []);
+    if (!metricTimeSeries) return undefined;
+    return filterInvalidMetricTimeSeries(metricTimeSeries.timeSeries);
   }, [metricTimeSeries]);
 
   const metricTimeSeriesMap = useMemo(() => {


packages/back-end/src/models/MetricTimeSeriesModel.ts
```diff
@@ -300,7 +306,7 @@ export class MetricTimeSeriesModel extends BaseClass {
       return sortedDataPoints;
     }
 
-    // Drop the oldest data points (earliest in the array order)
-    return sortedDataPoints.slice(-300);
+    // Drop the oldest data points
+    return sortedDataPoints.slice(0, 300);
```
jdorn: Is this change right? It looks like you push new entries to the end of the array, so the slice you had before seems like it would be correct.

ahdriel: Thanks for the callout!

It was wrong because my mocked data was wrong. I was inserting [earlier, older], while our code correctly calls it with the order [older, earlier].

So I have fixed it and added a new `.sort` to ensure the dataPoints are processed in the intended order, regardless of how they might come in.

jdorn (approved):

PR #4370, opened 2025-08-01
fix(time-series): Avoid going over array boundary

ahdriel:
### Features and Changes

With #4365 I changed `reduceRight` to `reduce` by mistake and did notice it, so I am undoing this change to fix a bug.
Also adding additional checks to ensure we don't go over the array boundary.


msamper (approved):
Lgtm

jdorn (approved):

PR #4372, opened 2025-08-04
Adds Sql Autocomplete to Sql Explorer

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
+++ b/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { useForm } from "react-hook-form";
 import { FaPlay, FaExclamationTriangle } from "react-icons/fa";
 import {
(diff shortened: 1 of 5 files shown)


packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -329,6 +342,57 @@ export default function SqlExplorerModal({
...
+          datasource?.type,
+          apiCall,
+          "SqlExplorer"
+        );
+        setAutoCompletions(completions);
+      } catch (error) {
+        console.error("Failed to fetch autocompletions:", error);
+        setAutoCompletions([]);
+      }
+    }, 300); // 300ms debounce
```
Kevin-Chant: Unless I'm misunderstanding, this doesn't actually do any debouncing. It just delays each call by 300ms but you could stack them up within that time. Is that a concern with the `useEffect` dependencies here?

mknowlton89: Oof - what a great catch. I've just made an update to the logic [here](https://github.com/growthbook/growthbook/pull/4372/commits/3f65b2972ffc0f1bb0c40bc472fde3e60c91adab) - I don't do this often, so if you have any suggestions I'm all ears.

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -344,12 +344,13 @@ export default function SqlExplorerModal({
...
   // Update autocompletions when cursor or schema changes
   useEffect(() => {
-    const timeoutId = setTimeout(async () => {
+    const fetchCompletions = async () => {
       if (!isAutocompleteEnabled) {
         setAutoCompletions([]);
         return;
       }
       try {
+        console.log("fetching completions");
```
mknowlton89: Whoops - will remove this

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
Kevin-Chant: Actually turns out that wasn't my finest comment and we realized I was wrong in Slack :sweat_smile:

Issue #4323, opened 2025-07-21
[Feature] Add a managedBy field to dimensions


jakemainwaring22:
### Description of Feature

Some entities have a `managedBy` property, e.g., `"managedBy": "api",` as defined [here](https://github.com/growthbook/growthbook/blob/9cee2d67635d20fccc6923627c04b6aa3538d834/packages/back-end/src/api/openapi/schemas/FactMetric.yaml#L23) for fact metrics. It would be great if this field could be added to GrowthBook dimensions as well to provide more info on how it was created.

PR #4374, opened 2025-08-05
Adds managedBy field to dimensions

mknowlton89:
### Overview

This PR introduces the concept of `Official Dimensions`, based off the pattern of [Official Metrics](https://github.com/growthbook/growthbook/pull/2007). This adds a new `managedBy` property to the `DimensionInterface`.

If this property has a value of `config` or `api` - we prevent users from being able to edit or delete the dimension from the UI.

### Managed By
There are 2 different places people manage their GrowthBook dimensions:

In a config.yml file (self-hosted only)
In the GrowthBook UI
~~Via the REST API~~ this is coming in a follow-up PR

This PR adds a new managedBy property to dimensions. It has 3 possible values:

"" (empty string) - can be managed via both the UI and REST API
"api" - can only be managed via the REST API
"config" - can only be managed via a config.yml file. 

By default, we will infer this value. Anything defined in config.yml will get a value of "config". Everything else will get a value of "" (empty string). When creating/updating resources via the REST API, you can opt-in to setting this field to "api" to lock down the UI.

For now, we have simple logic to determine official status. If the managedBy property is empty, it is "ad-hoc", otherwise it is considered "official". We may make this more explicit in the future.

### Config.yml Changes
There is a new environment variable `ALLOW_CREATE_DIMENSIONS`. When set to true, new dimensions can be created via the UI even while config.yml is being used. This creates 2 classes of dimensions - Official ones that can only be changed via config.yml and "ad-hoc" dimensions that can be managed via the UI.

- Closes **(https://github.com/growthbook/growthbook/issues/4323)**

### Testing

- [x] Ensure a user isn't able to edit/delete an `Official Dimension` (i.e. `managedBy`: `config`)
- [x] Ensure we render the Official icon next to any official dimensions in the app with a proper tooltip

### Screenshots
Screenshot 2025-08-05 at 3 18 25 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Get up and running in 1 minute with:
 ```sh
 git clone https://github.com/growthbook/growthbook.git
 cd growthbook
-docker-compose up -d
+docker compose up -d
 ```
 
 Then visit http://localhost:3000
@@ -49,7 +49,7 @@ The included [docker-compose.yml](https://github.com/growthbook/growthbook/blob/
 ```sh
 git clone https://github.com/growthbook/growthbook.git
 cd growthbook
-docker-compose up -d
+docker compose up -d
 ```
 
 Then visit http://localhost:3000 to view the app.
--- a/docker-compose.proxy.yml
+++ b/docker-compose.proxy.yml
@@ -1,4 +1,3 @@
-version: "3"
 services:
   mongo:
     image: "mongo:latest"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3"
 services:
   mongo:
     image: "mongo:latest"
--- a/docs/docs/guide/nextjs-and-growthbook.mdx
+++ b/docs/docs/guide/nextjs-and-growthbook.mdx
@@ -51,7 +51,7 @@ You will need a GrowthBook account. You can either run GrowthBook locally or usi
 ```bash
 git clone https://github.com/growthbook/growthbook.git
 cd growthbook
-docker-compose up -d
+docker compose up -d
 ```
 
 After that, visit `http://localhost:3000` and create your first user account.
--- a/docs/docs/guide/nextjs-app-router.mdx
+++ b/docs/docs/guide/nextjs-app-router.mdx
@@ -38,7 +38,7 @@ You will need a GrowthBook account. You can either run GrowthBook locally or usi
 ```bash
 git clone https://github.com/growthbook/growthbook.git
 cd growthbook
-docker-compose up -d
+docker compose up -d
 ```
 
 After that, visit `http://localhost:3000` and create your first user account.
--- a/docs/docs/self-host/env.mdx
+++ b/docs/docs/self-host/env.mdx
@@ -80,7 +80,7 @@ If you change the `ENCRYPTION_KEY`, you will need to migrate any existing data s
 
 ```bash
 # If you didn't have an ENCRYPTION_KEY before, leave OLD_KEY blank below
-docker-compose run growthbook yarn migrate-encryption-key OLD_KEY
+docker compose run growthbook yarn migrate-encryption-key OLD_KEY
 ```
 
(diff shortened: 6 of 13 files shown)


packages/back-end/src/services/organizations.ts
```diff
@@ -858,15 +858,15 @@ export async function importConfig(
               ...d,
             };
             delete updates.organization;
-
-            await updateDimension(k, organization.id, updates);
+            await updateDimension(context, existing, updates);
```
mknowlton89, line 861: @bryce-fitzsimons @jdorn My understanding of this code block is a bit fuzzy.

We're doing a dimension lookup and if it already exists (which it could now exist in Mongo via the UI or from the REST API), we're then updating it.

This could throw an error if the existing dimension was from the REST Api.

But also, if the existing dimension isn't a config managed dimension, but the dimension is now in the config file, it feels like that dimension should not be managed by the config? Aka, have a `managedBy: "config"` property.

packages/back-end/src/services/organizations.ts
```diff
@@ -858,15 +858,15 @@ export async function importConfig(
...
           } else {
             await createDimension({
```
mknowlton89, line 863: I'm confused why we're ultimately calling `createDimension` for config.yml managed dimensions - if an org uses the config.yml file, when we do a lookup we're looking for it within the config.yml file directly, so why are we making a `createDimension` call at all here?

packages/back-end/src/services/organizations.ts
mknowlton89: Ah, I see now - this was a feature added here to allow an org to take a config file and then import it and thus add the resources to mongo. So this is different from an org using the config file.

packages/back-end/src/services/organizations.ts
mknowlton89: See comment above - I understand now

packages/front-end/pages/dimensions/index.tsx
```diff
@@ -296,11 +304,10 @@ const DimensionsPage: FC = () => {
                           expandable={true}
                         />
                       </td>
-                      {/* @ts-expect-error TS(2345) If you come across this, please fix it!: Argument of type 'Date | null' is not assignable t... Remove this comment to see the full error message */}
-                      {!hasFileConfig() && <td>{ago(s.dateUpdated)}</td>}
-                      {!hasFileConfig() && (
+                      {s.dateUpdated ? <td>{ago(s.dateUpdated)}</td> : null}
```
bryce-fitzsimons: should this text be hidden if `managedBy` is truthy or perhaps `orgCanCreateDimensions`?

mknowlton89: I've just made a few tweaks to this table now that we're allowing orgs to have managed dimensions and ad-hoc dimensions. So currently the table renders all columns and we simply render a `-` if the dimension doesn't have a `dateUpdated` value (which will happen for config managed dimensions).

Likewise, we're now always rendering the edit column, and just hiding the edit options for official dimensions.

packages/front-end/pages/dimensions/index.tsx
```diff
@@ -296,11 +304,21 @@ const DimensionsPage: FC = () => {
                           expandable={true}
                         />
                       </td>
-                      {/* @ts-expect-error TS(2345) If you come across this, please fix it!: Argument of type 'Date | null' is not assignable t... Remove this comment to see the full error message */}
-                      {!hasFileConfig() && <td>{ago(s.dateUpdated)}</td>}
-                      {!hasFileConfig() && (
+                      <td>
+                        {s.dateUpdated ? (
```
mknowlton89: @bryce-fitzsimons This logic assumes that dimensions created via the REST API will contain a `dateUpdated` value.

bryce-fitzsimons: cool, they will have the dates set

Issue #4328, opened 2025-07-21
[Feature] Add CRUD operations for dimensions


jakemainwaring22:
### Description of Feature
Currently, the only [endpoints](https://docs.growthbook.io/api/#tag/dimensions) available for dimensions are to get all dimensions or get a single dimension:
<img width="231" height="186" alt="Image" src="https://github.com/user-attachments/assets/a2d0ab49-61ce-4ee2-af6a-ef2a4629035c" />

Adding the full set of CRUD operations for dimensions, e.g., closer to what we have with metrics, could be very valuable:
<img width="234" height="219" alt="Image" src="https://github.com/user-attachments/assets/939f6da7-7a28-49a1-99fa-7c023a286aa5" />

Being able to create and modify dimensions in this way would give folks more options for managing them programmatically. Happy to discuss further if it would be helpful!

PR #4376, opened 2025-08-05
REST API: add additional CRUD operations for dimensions

bryce-fitzsimons:
### Features and Changes

Closes https://github.com/growthbook/growthbook/issues/4328

Depends on https://github.com/growthbook/growthbook/pull/4374

### DIFF
--- a/docs/docs/self-host/ferretdb.mdx
+++ b/docs/docs/self-host/ferretdb.mdx
@@ -1,0 +1,61 @@
+---
+title: FerretDB
+description: How to use FerretDB with GrowthBook
+---
+
+# FerretDB
+
+FerretDB is a MongoDB-compatible database that is open source and free to use. It serves as a drop-in replacement for MongoDB, converting MongoDB wire protocol queries to SQL and using PostgreSQL as its backend storage engine.
+
+:::info
+For organizations that require a 100% open-source stack, FerretDB provides a truly open-source MongoDB-compatible database alternative—free from SSPL licensing restrictions—allowing you to run GrowthBook without any proprietary database dependencies.
+:::
+
+## Install FerretDB
+
+Follow the the [FerretDB installation guide](https://docs.ferretdb.io/installation/ferretdb/docker/) to install FerretDB. Make sure your FerretDB instance is running and accessible before proceeding.
+
+## Start GrowthBook with FerretDB
+
+```bash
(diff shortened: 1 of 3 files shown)


packages/back-end/src/api/openapi/payload-schemas/PostDimensionPayload.yaml
```diff
@@ -0,0 +1,32 @@
...
+  query:
+    type: string
+    description: SQL query or equivalent for the dimension
+  managedBy:
+    type: string
+    description: Where this dimension must be managed from. If not set (empty string), it can be managed from anywhere.
+    enum:
+      - ""
+      - api
+      - config
```
mknowlton89: I think we may want to remove the "config" option from the enum as I don't think we want to encourage users to create/update dimensions via the REST api and set `managedBy: "config"` as they'd get in a really weird state where it couldn't be removed/edited.

I think removing the config option here might make some of the linters unhappy, but I think we solved this issue with the Metrics endpoints -

packages/back-end/src/api/dimensions/postDimension.ts
```diff
@@ -0,0 +1,41 @@
...
+    const organization = req.organization.id;
+
+    const datasourceDoc = await getDataSourceById(
+      req.context,
+      req.body.datasourceId
+    );
+    if (!datasourceDoc) {
+      throw new Error("Invalid data source");
+    }
+
```
mknowlton89, line 22: Updating the enum options might fix this itself, but I wonder if we want to add a safe guard to this endpoint (and the update endpoint) to return an error if the payload's `managedBy` property is set to `"config"`, given how the user would then be unable to edit/delete the Dimension.

It may be that we get this for free by updating the yaml, but not sure without testing.

bryce-fitzsimons: The zod validator should give us this for free (given the yaml fix you suggested)

packages/back-end/src/api/openapi/payload-schemas/PostDimensionPayload.yaml
bryce-fitzsimons: good call. This also has the side effect of updating the zod validator, giving us endpoint safety (you can't pass in "config" to either create or update endpoints)

PR #4385, opened 2025-08-07
Dashboard UI improvements

Kevin-Chant:
### Features and Changes

* The "+" button for adding blocks was adding a lot of clutter by repeating every block. Now it only shows when hovering the area between two blocks or after the last block
* The edit drawer has a handle for resizing vertically with a min and max height
* Various block types have had their titles moved from the internal graph to the block's outer container
* Nested appboxes have been removed

#### Blocks without +
image

#### Hovering for +
image

#### Edit drawer
image

#### Block title & appbox removal
image

### DIFF
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlockEditDrawer.tsx
@@ -293,29 +293,25 @@ export default function DashboardBlockEditDrawer({
           order={2}
           defaultSize={50}
           minSize={40}
-          style={{ background: "var(--background-color)" }}
+          style={{
+            background: "var(--background-color)",
+            position: "relative",
+          }}
         >
+          <div
+            style={{
+              position: "absolute",
+              bottom: "0px",
+              width: "100%",
+              height: "20px",
+              background:
+                "linear-gradient(180deg, rgba(0,0,0,0) 0%, rgba(0,0,0,0.1) 100%)",
+              zIndex: 10000,
+              pointerEvents: "none",
+            }}
+          />
           {block && (
-            <Flex
-              direction="column"
-              pt="5"
-              px="4"
-              gap="2"
-              height="100%"
-              position="relative"
-            >
-              <div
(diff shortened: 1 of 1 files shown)


packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
```diff
@@ -312,59 +307,77 @@ function DashboardEditor({
...
+                trigger={
+                  <IconButton
+                    onMouseEnter={() => {
+                      setHoverAddBlock(i);
+                    }}
+                    onMouseLeave={() => {
+                      setHoverAddBlock(undefined);
+                    }}
+                    className={clsx({
+                      "d-none":
```
bryce-fitzsimons, line 352: the layout shift here is a bit jarring. could we instead toggle visibility and pointerEvents?

bryce-fitzsimons:
Any idea what's going on here?
1. Why is there a weird bottom shadow covering the page (on all pages, not just dashboards)? I don't think this is related to your work, but it's kinda odd...
2. Why does the edit drawer create a weird gap in the shadow?
image

bryce-fitzsimons:
design nit: the edit drawer would look better with some sort of upward drop shadow IMO
image

Kevin-Chant:
@bryce-fitzsimons for the bottom shadow that should be part of the edit drawer (showing that you can scroll) but I can't see it on any other tabs or when the drawer is closed. Changing it to solid blue so you can't miss it, it doesn't appear anywhere else (chrome or FF).
image

But it wasn't supposed to be affected by the padding of the drawer so I've fixed that. Can you check if that shadow on your other tabs is from this component and if so why it's showing up there for you?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
Kevin-Chant: True, but I added that intentionally because I thought it'd be easy to miss the button otherwise, and I've seen that sort of shift when moving blocks before. Either way, this won't be visible to any customers yet, it's just so Kari can test the design in production and iterate.

PR #4390, opened 2025-08-08
Add Filtering to SqlExplorer Visualizations

mknowlton89:
### Features and Changes

This PR introduces the concept of filtering to Sql Explorer Visualizations. With this, organizations can now add multiple filters to the results of a SQL Query to filter the data further in a way unique to a particular visualization.

###A few things to note:
1. We're allowing organizations the ability to apply multiple filters to a visualization. Currently, for a row to be included in the visualization, it needs to pass ALL filter checks. We don't currently support "or" logic.
2. Currently, all of the filtering is happening before the rows are aggregated - we don't support the ability to filter on aggregated rows.
3. The list of filtering options supported (outlined below) is not the final list, but just the initial list for MVP.

#### Number Filters
- Number Range
- Greater Than
- Greater Than or Equal To
- Less Than
- Less Than or Equal To
- Equal To

#### Date Filters
- Today
- Last 7 Days
- Last 30 Days
- Date Range

#### String Filters
- Includes (this creates an array of options, the the column's value must be included in the `includes` array in order to pass the filter test.
- Contains (this is a string subtext search. E.X If a column has values like "pro-legacy" and "pro", a user can enter "pro" and both of the above plan types will pass the filter test.

### Fast-Follows:
- Add additional date filter options (last week, last month, last quarter, last year, last X days, last X months, last X years, month-to-date, quarter-to-date, year-to-date)
- Update UI when filtering by date with a callout that clarifies how date filtering works with timezones (E.G. `This filter will display all available timestamp data from Jul 22, 2024 12:00am UTC to Nov 18, 2024 11:59pm UTC.`)

### Future Plans
- Add additional `min`/`max` fields to the Y-Axis section that offers orgs some lightweight post-aggregation filters. This is handy for things like removing outliers post-aggregation.
- Add another tab the SQL results section where a user can override any inferred column data types. This would allow us to remove all of the `Type` fields within the DataVizConfigPanel, but the user could keep the ability to override any data types.

### Dependencies

- None

#### Happy Paths:
String Filters
- [x] Ensure the contains string type filter works as expected
- [x] Ensure the includes string type filter works as expected

Number Filters
- [x] Ensure the equals number type filter works as expected
- [x] Ensure the less than number type filter works as expected
- [x] Ensure the less than or equals to number type filter works as expected
- [x] Ensure the greater than number type filter works as expected
- [x] Ensure the greater than or equals to number type filter works as expected
- [x] Ensure the numberRange number type filter works as expected

Date Filters
- [x] Ensure the today date type filter works as expected
- [x] Ensure the last 7 days date type filter works as expected
- [x] Ensure the last 30 days date type filter works as expected
- [x] Ensure the dateRange date type filter works as expected

#### Error Paths/Edge Cases:
- [x] Ensure a user can not save a `saved-query` if any of the filters have required fields that are not finalized
- [x] Ensure that dates of various kinds work correctly when using the date filter. E.G. A date like `"2022-07-16T09:32:39.000Z"` works, and also dates formatted like: `"2025-06-15"`.
- [x] Ensure when filtering by date, the date values entered in the filter are in UTC time, not in any local timezones
- [x] Ensure that the string filtering correctly normalizes all text to lowercase before filtering


packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -134,6 +134,144 @@ export function DataVisualizationDisplay({
...
+                  filterDate.getFullYear(),
+                  filterDate.getMonth(),
+                  filterDate.getDate(),
+                );
+                const todayOnly = new Date(
+                  today.getFullYear(),
+                  today.getMonth(),
+                  today.getDate(),
+                );
+                return filterDateOnly.getTime() === todayOnly.getTime();
```
mknowlton89: I need to validate this logic.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+                        startDate: thirtyDaysAgo.toISOString().split("T")[0],
+                        endDate: today.toISOString().split("T")[0],
+                      },
+                    };
+                  } else if (type === "number") {
+                    newFilter = {
+                      column: firstColumnFilterOption.column,
+                      type: "number",
+                      filterType: "equalTo",
+                      config: { value: "0" },
```
ahdriel, line 221: suggestion: Should we start with `greaterThan` `0`? Currently when I add this filter all my data vanishes because very few of my dataPoints have value of 0.

The goal is to create a less jarring experience for the user (being broad by default) and then allow them to tighten the filters if needed.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+                onClick={() => {
+                  // Get the first column filter option
+                  const firstColumnFilterOption = columnFilterOptions[0];
+                  const type = firstColumnFilterOption.knownType;
+
+                  // Add new filter with default values based on the type
+                  let newFilter: FilterConfiguration;
+
+                  if (type === "date") {
+                    // Default to last 30 days for date filters
```
ahdriel, line 202: question: Any way we can apply the same defaults when changing between Types?

For example, if I start on String and select Date, the option selected is `Custom Date Range`.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+                      type: "number",
+                      filterType: "equalTo",
+                      config: { value: "0" },
+                    };
+                  } else {
+                    newFilter = {
+                      column: firstColumnFilterOption.column,
+                      type: "string",
+                      filterType: "contains",
+                      config: { value: "" },
```
ahdriel: suggestion: Should we default by the `Select values`? I think it can be more helpful / easier than the string field.

Not a strong opinion at all though, so feel free to ignore if you disagree.

packages/back-end/src/validators/saved-queries.ts
```diff
@@ -62,6 +134,7 @@ const formatEnum = z.enum([
 const baseChartConfig = z.object({
   title: z.string().optional(),
   yAxis: z.array(yAxisConfigurationValidator).nonempty(),
+  filter: z.array(filterConfigurationValidator).optional(),
```
ahdriel: nit: `filters`?

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -238,6 +238,118 @@ export default function SqlExplorerModal({
...
         );
       }
+
+      // Validate filters
+      if (config.filter && config.filter.length > 0) {
+        config.filter.forEach((filter, filterIndex) => {
+          const vizTitle = config.title || `${index + 1}`;
+
+          // Validate required filter fields
+          if (!filter.column) {
```
ahdriel, line 249: Would be nice to have all of these validation integrated with zod somehow -- not required for this PR, just food for thought as we keep evolving the DataViz stuff.

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -134,6 +134,154 @@ export function DataVisualizationDisplay({
...
+
+    return rows.filter((row) => {
+      return filters.every((filter) => {
+        const { column } = filter;
+        const rowValue = row[column];
+
+        // Handle null/undefined values
+        if (rowValue == null) return false;
+
+        switch (filter.filterType) {
```
ahdriel: question: Should we filter on `type` first?

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
```diff
@@ -1,61 +1,74 @@
...
+  return "string";
+}
 
 export default function DataVizConfigPanel({
   sampleRow,
+  rows,
   dataVizConfig,
   onDataVizConfigChange,
 }: {
   sampleRow: Record<string, unknown>;
```
ahdriel: question: This is currently used in a single place and we are passing `rows` to it.

Should we remove `sampleRow` and make `rows` required?

And if `rows` is empty maybe we return `null`?

packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
```diff
@@ -0,0 +1,302 @@
...
+}: {
+  dataVizConfig: Partial<DataVizConfig>;
+  onDataVizConfigChange: (dataVizConfig: Partial<DataVizConfig>) => void;
+  axisKeys: string[];
+}) {
+  if (!supportsDimension(dataVizConfig)) {
+    return null;
+  }
+
+  const dimensions = dataVizConfig.dimension || [];
```
ahdriel: suggestion: We should explicitly cast dimensions as `dimensionAxisConfiguration[]`.

Otherwise the `|| []` removes the TS ability to know what this is and on line `94` the dimension gets treated as `any`.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+export type ColumnFilterOption = {
+  column: string;
+  knownType: "date" | "number" | "string";
+};
+
+type Props = {
+  dataVizConfig: Partial<DataVizConfig>;
+  onDataVizConfigChange: (dataVizConfig: Partial<DataVizConfig>) => void;
+  sampleRow: Record<string, unknown>;
+  rows?: Record<string, unknown>[];
```
ahdriel: question: Same here -- do we need both?

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+              {filters.length > 0 &&
+                filters.map((filter, index) => {
+                  return (
+                    <DataVizFilter
+                      key={index}
+                      filterIndex={index}
+                      columnFilterOptions={columnFilterOptions}
+                      dataVizConfig={dataVizConfig}
+                      onDataVizConfigChange={onDataVizConfigChange}
+                      rows={rows}
```
ahdriel: question: Should the DataVizFilter component receive the whole dataVizConfig? Or should we pass in just the specific filter we are rendering and keep the 'add to dataVizConfig filters' part in this component?

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    />
+                  );
+                })}
+              <a
```
ahdriel: question: Can we use `Link` from `components/Radix` here instead?

packages/front-end/components/DataViz/DataVizFilter.tsx
```diff
@@ -0,0 +1,472 @@
...
+      ...dataVizConfig,
+      filter: newFilters,
+    });
+  };
+
+  const createDefaultFilterForType = (
+    column: string,
+    type: "string" | "number" | "date",
+  ): FilterConfiguration => {
+    if (type === "date") {
```
ahdriel: suggestion: Can we do a exhaustive switch here?

The goal is to make TypeScript complain if we add a new filter type but don't update the function. Currently if we add a new one it would be treated as string.

packages/front-end/components/DataViz/DataVizFilter.tsx
```diff
@@ -0,0 +1,472 @@
...
+      // Default to last 30 days for date filters
+      const today = new Date();
+      const thirtyDaysAgo = new Date(today);
+      thirtyDaysAgo.setDate(today.getDate() - 30);
+
+      return {
```
ahdriel: Should this logic of the `default` state of a filter be merged with DataVizFilterPanel logic lines 185-214?

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
mknowlton89: For sure - great suggestion.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
mknowlton89: Can you expand on this a bit - I'm not sure I follow.

You selected a field and the `Type` was `String`, and then changed the type to `Date`? In that case, the default filterType is `dateRange` and we pre-fill the endDate as the current date, and the start date as 30 days in the past.

Is your suggestion that instead of doing the date range we just selected the `filterType` of `last30Days`?

Sorry if I misunderstood.

packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
mknowlton89: Agreed

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
mknowlton89: @ahdriel to clarify - is this suggestion to reduce all of the `isNan` checks in each of the individual switch blocks? That is a smart improvement that I'll make.

If your suggestion was for a different optimization, just let me know.

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
mknowlton89: Oooh yea - great call.

We don't even actually have to handle the rows being empty currently.

The parent elements of DataVizConfigPanel require `rows` and we have an empty state if there are no rows.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
mknowlton89: Nope - great suggestion.

packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
mknowlton89: Will do.

This had a side effect of TS not being happy with setting the default value for a dimension's `fieldName` to an empty string.

As a result, I've just set the default value of the dimension to be the first index of the `axisKeys`.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
mknowlton89: For sure

packages/front-end/components/DataViz/DataVizFilter.tsx
mknowlton89: Took a pass at this in this commit.

It's definitely better and the `type` options are now based on `FilterConfiguration["type"]` so only known types can be passed in.

But I'm not sure of a way to make it more exhaustive outside of writing some custom logic to make sure each type is accounted for. If you do know of a way, I'd love to implement it.

packages/front-end/components/DataViz/DataVizFilter.tsx
mknowlton89: Yea, good call.

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
ahdriel: I think this has been fixed with the latest changes -- but the short version is that the defaults were not being applied when I changed the type of a filter. But they were if the field type was detected when I added a new filter.

Now the default for each type is properly applied regardless if it is a new filter, or if it is me changing between Types for an existing filter.

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
ahdriel: It was less about the `isNan` checks and more that at this moment the `filter.filterType` is unique between all `filter.type`, but in this function we are not checking that `filter.filterType = last7Days` belong to the `filter.type = 'date'`.

It would be one additional check to work on the way of `filter.type` -> `filter.filterType` instead of skipping the first check.

packages/front-end/components/DataViz/DataVizFilter.tsx
ahdriel: I just pushed [this small commit](https://github.com/growthbook/growthbook/pull/4390/commits/2fbae7dd6d7d754b670916992ce616692ef64e7c) that makes TypeScript complain if we ever have a `FilterConfiguration['type']` value that is not being handled in this function.

mknowlton89: TIL!

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
mknowlton89: Ah - I see now.

Just added this small commit to address it - https://github.com/growthbook/growthbook/pull/4390/commits/b986aa26ede1b04cf302580c5d08f8ab48718ba0

Its still not a great DX as the array of `filterTypes` for each `filter.type` is manually written, so if we add a new `filterType` to the `FilterConfiguration` we'll need to add it here as well. Though, I guess we'd also need to add it to the `filterOptions` defined within `DataVizFilter.tsx`.

I do see a way to use the `filterOptions` defined within `DataVizFilter` and based on the `type` finding the index of the `filterOptions` and using the `supportedTypes` array to do the validation.

But I'm wondering if this is the best place to do this - in the commit above this check is running for every row of the results.

I feel like this may be better validation to do when saving the dataVizConfig, as we could do it once per filter rather than for every row and every filter.

mknowlton89: We are currently doing this logic right before save here - https://github.com/growthbook/growthbook/pull/4390/files#diff-925591c6d977a3254beb53223acada9e0f40b6a0a65340bca13e7b70668e1294R283-R288

mknowlton89: Just pushed up this PR that updates the `type` and `filterType` compatibility check before saving to use the `filterOptions` so it's a bit more dynamic. Still need to test a bit more, ran out of time before the engineer all hands.

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
```diff
@@ -56,26 +56,26 @@ export function inferFieldType(
 }
 
 export default function DataVizConfigPanel({
-  sampleRow,
+  // sampleRow,
```
ahdriel: nit: can we remove the commented out code?

packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
```diff
@@ -21,7 +24,8 @@ export default function DataVizDimensionPanel({
     return null;
   }
 
-  const dimensions = dataVizConfig.dimension || [];
+  const dimensions =
```
ahdriel: nit: I think casting it as `const dimensions: dimensionAxisConfiguration[] = dataVizConfig.dimension || [];` will make the typechecker happy & also complain if `dataVizConfig.dimension` changes for whatever reason, while the current implementation will make TS happy even if `typeof dataVizConfig.dimension` changes and don't match `dimensionAxisConfiguration`.

packages/front-end/components/DataViz/DataVizFilter.tsx
```diff
@@ -97,222 +84,216 @@ export default function DataVizFilter({
...
-      filter: newFilters,
-    });
-  };
-
-  const createDefaultFilterForType = (
+  const createFilterConfig = (
     column: string,
-    type: "string" | "number" | "date",
+    type: FilterConfiguration["type"],
+    filterType?: string,
```
ahdriel: nit (non-blocking): `filter.type` and `filter.filterType` are very confusing names unfortunately. Maybe we can figure out a more clear naming scheme in the near future?

packages/front-end/components/DataViz/DataVizFilter.tsx
```diff
@@ -97,222 +84,216 @@ export default function DataVizFilter({
...
-                  ),
-              )
-              .map((filterOption) => (
-                <SelectItem key={filterOption.value} value={filterOption.value}>
-                  {filterOption.label}
+  return (
+    <>
+      {filter.column !== undefined && filter.type !== undefined && (
+        <Flex direction="column" gap="2">
+          {filter.column !== undefined && (
```
ahdriel: nit: We can remove this check now that the top condition was updated

packages/front-end/components/DataViz/DataVizFilterPanel.tsx
```diff
@@ -71,22 +71,19 @@ function getColumnFilterOptions(
 export default function DataVizFilterPanel({
   dataVizConfig,
   onDataVizConfigChange,
-  sampleRow,
+  // sampleRow,
```
ahdriel: nit: Same here to remove commented out code.

ahdriel (approved):
Thanks for the updates ! 🚀

packages/front-end/components/DataViz/DataVizFilter.tsx
mknowlton89: Agreed.

I do think now would be the time to change it.

We currently have

```ts
  z.object({
    column: z.string(),
    type: z.literal("date"), // I think maybe we can change it to columnType
    filterType: z.enum(["today", "last7Days", "last30Days"]), // we could change this to 
    config: z.object({}).optional(), // No config needed
  }),
  ```
  
  My thought is we do something like:
  
```ts
  z.object({
    column: z.string(),
    columnType: z.literal("date"), // I think maybe we can change it to columnType
    filterMethod: z.enum(["today", "last7Days", "last30Days"]), // we could change this to 
    config: z.object({}).optional(), // No config needed
  }),

mknowlton89: Added these updates [here](https://github.com/growthbook/growthbook/pull/4390/commits/5bc6073b582118b7d502e285ff47a5fe3c3194a0).

ahdriel: Nice! `filterType`, `columnType` + `filterMethod` are much clearer now. Thanks for the update!

Issue #4267, opened 2025-07-08
[Feature] Add support for MS Teams notifications


august-growthbook:
Similar to the notifications offered for Slack and Discord, but in MS Teams.

alyssamichelle:
Triage from the open-issues audit:

**Suggested action:** Fold into a GitHub Discussion

**Why:** Same Teams ask as #2003. This is a product-direction ask. Please convert or fold into a Discussion rather than keeping it as a standalone issue.

- [Slack (closest chat)](https://docs.growthbook.io/integrations/slack)

alyssamichelle:
Advised to close after review: this is now being discussed in https://github.com/growthbook/growthbook/discussions/6832 (Which third-party integrations are worth first-party support?).

PR #4399, opened 2025-08-11
sdk destory() - add option to destroy all sse streams

bryce-fitzsimons:
Closes https://github.com/growthbook/growthbook/issues/4267

### DIFF
--- a/packages/sdk-js/src/feature-repository.ts
+++ b/packages/sdk-js/src/feature-repository.ts
@@ -548,7 +548,7 @@ function destroyChannel(channel: ScopedChannel, key: string) {
   streams.delete(key);
 }
 
-function clearAutoRefresh() {
+export function clearAutoRefresh() {
   // Clear list of which keys are auto-updated
   supportsSSE.clear();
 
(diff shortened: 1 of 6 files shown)


packages/sdk-js/src/feature-repository.ts
```diff
@@ -548,7 +548,7 @@ function destroyChannel(channel: ScopedChannel, key: string) {
   streams.delete(key);
 }
 
-function clearAutoRefresh() {
+export function clearAutoRefresh() {
```
gazzdingo, line 554: nit: should we update this to something like removeAutoRefreshStreams seems more clear since we are exporting it now

gazzdingo (approved):
LGTM just one nit

packages/sdk-js/src/feature-repository.ts
bryce-fitzsimons: it's not a public SDK method, we're just letting the base SDK use this util function. I might leave it as-is since it also cleans up state subscriptions which aren't strictly about SSE.

PR #4400, opened 2025-08-12
Adds managedBy field to segments

mknowlton89:
### Features and Changes

This PR picks up the work left off here (https://github.com/growthbook/growthbook/pull/4374) and adds the `managedBy` field to Segments. Now, when self-hosted orgs create Segments via the `config.yml` file, the segment will have a `managedBy: "config"` property, which will surface the `OfficialBadge` in the UI.

As a fast-follow to this PR, I'll update the Segment REST API endpoints to support the `managedBy` field, so orgs can create/update segments and optionally pass in a `managedBy: "api"` property. If set, the Segment will then only be editable via the Rest API, and not the UI.

This matches the pattern of [Official Dimensions](https://github.com/growthbook/growthbook/pull/4374) and [Official Metrics](https://github.com/growthbook/growthbook/pull/2007). 

One noticable difference between this PR and the Official Dimension/Official Metrics pr is that we've now introduced a `readOnly` mode for the Segment Modals. Given that Segment's are sql or Fact table based resources and they don't have a dedicated /segment/id page, if we were to hide the `Edit` CTA like we do for Metrics/Dimensions, no user would be able to view the actual sql that powers the Segment.

So I've introduced a readOnly mode for the two Segment Modals that goes into effect IF the Segment is an Official Segment (ie. it was created via the config file or via the API). This allows users to view the actual definition of the Segment.

Likewise, this readOnly mode also impacts non-official Segments by taking into consideration the user's permissions. Even if the Segment isn't an official Segment, if the user doesn't have create/edit permissions, we're still allowing them to open the modal, but it'll open in readonly mode so they can't make changes, but they can view the details of the Segment.

### Dependencies

- None

### Testing

- [x] Ensure if a Segment is Official (e.g. managedBy equals "config" or "api") - the modal always opens in `readOnly` mode
- [x] Ensure if a Segment is Official (e.g. managedBy equals "config" or "api") - the Official Segment badge is rendered
- [x] Ensure if a Segment isn't an Official Segment, if the user has permission to create/update, readOnly mode is not set
- [x] Ensure if a Segment isn't an Official Segment, if the user doesn't have create/update permissions, readOnly mode is set
- [x] Ensure that the `GET` requests for Segments on `/segments` correctly returns both Config segments and mongo segments if `STORE_SEGMENTS_IN_MONGO` is true
- [x] Ensure that the `GET` requests for Segments on `/segments` correctly returns ONLY Config segments and no mongo segments if `STORE_SEGMENTS_IN_MONGO` is false
- [x] Ensure that the `GET` requests for Segments on `/segments` only returns Mongo segments if no config.yml is set

### Screenshots
Screenshot 2025-08-12 at 3 04 34 PM

Read-Only mode for API managed Segment
Screenshot 2025-08-12 at 3 04 45 PM

### DIFF
--- a/docs/docs/self-host/config.mdx
+++ b/docs/docs/self-host/config.mdx
@@ -518,7 +518,7 @@ sql: SELECT user_id, timestamp as date FROM pages WHERE country='US'
 ```
 
 Segment support for the config.yml was added in February 2023. If you are using the config.yml file and have previously created segments stored in MongoDB, in order to access,
-you will need to add the environment variable `STORE_SEGMENTS_IN_MONGO` or update your config.yml file to include these segments.
+you will need to add the environment variable `STORE_SEGMENTS_IN_MONGO` or update your config.yml file to include these segments. Likewise, if you'd like to create new segments via the UI in addition to those in the config.yml, you'll want to set this environment variable to `TRUE`.
 
 ### Organization Settings
 
--- a/packages/back-end/src/init/config.ts
+++ b/packages/back-end/src/init/config.ts
@@ -237,6 +237,7 @@ export function getConfigSegments(organization: string): SegmentInterface[] {
       organization,
       dateCreated: new Date(),
       dateUpdated: new Date(),
+      managedBy: "config",
     };
   });
 }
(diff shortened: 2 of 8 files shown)


bryce-fitzsimons (approved):
looks good, pretty similar to dimensions (not thoroughly tested though)

Issue #4392, opened 2025-08-08
[Feature] Add "Dependent Experiments" to the "Dependents" section on a Feature's detail page


august-growthbook:
Currently only Dependent Features are shown on a Feature's detail page in the "Dependents" section.

It would be great to also show Dependent Experiments in this same section.

<img width="977" height="309" alt="Image" src="https://github.com/user-attachments/assets/9b6d14b0-4af4-49ae-a80f-1603a39b789a" />

albertoastiazaran-qz:
Maybe a button that manually triggers the query/scan instead of automatically doing it when loading a feature page?

PR #4401, opened 2025-08-12
Feature pages: fix dependent experiment list, fix stale feature check

bryce-fitzsimons:
- Closes https://github.com/growthbook/growthbook/issues/4392
- Fixes the stale feature check (part of the check looks at dependent experiments, which was broken)

### DIFF
--- a/.github/workflows/notify-on-failure.yml
+++ b/.github/workflows/notify-on-failure.yml
@@ -29,4 +29,4 @@ jobs:
           webhook-type: webhook-trigger
           payload: |
             workflow_name: "${{ github.event.workflow_run.name }}"
-            workflow_url: "${{ github.event.workflow_run.url }}"
+            workflow_url: "${{ github.event.workflow_run.html_url }}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,12 @@ WORKDIR /usr/local/src/app
 # Set node max memory
 ENV NODE_OPTIONS="--max-old-space-size=8192"
 RUN apt-get update && \
-  apt-get install -y wget gnupg2 build-essential && \
-  echo "deb https://deb.nodesource.com/node_$NODE_MAJOR.x buster main" > /etc/apt/sources.list.d/nodesource.list && \
-  wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
-  echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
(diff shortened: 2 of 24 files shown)


packages/front-end/pages/features/[fid].tsx
```diff
@@ -69,6 +70,7 @@ export default function FeaturePage() {
   const experiments = data?.experiments;
   const safeRollouts = data?.safeRollouts;
   const [error, setError] = useState<string | null>(null);
+  const { experiments: allExperiments } = useExperiments();
```
Comment (bryce-fitzsimons) [inline packages/front-end/pages/features/[fid].tsx:73]: adds a bit of overhead, but we do use this hook extensively in our codebase

packages/front-end/pages/features/[fid].tsx
```diff
@@ -274,9 +276,9 @@ export default function FeaturePage() {
   }, [feature, features, envs]);
 
   const dependentExperiments = useMemo(() => {
```
Comment (ahdriel) [inline packages/front-end/pages/features/[fid].tsx:278]: question: Before this was being set using `experiments` from the `apiCall` to `/feature/${fid}${queryString}` (line 99).

Should we remove that part of the code if we are not using it anymore? Or should we investigate why the API response is not what we expect and why getting all experiments fixes this?

Same comment applies to `packages/front-end/pages/features/index.tsx`.

Comment (bryce-fitzsimons) [reply packages/front-end/pages/features/[fid].tsx:278]: This API call is specifically to return the linked experiments (experiment ref rules, basically). That's an entirely different calculation from prerequisite dependents, which does not have this data encoded in the model anywhere and is not easily queryable.

Comment (bryce-fitzsimons) [reply packages/front-end/pages/features/[fid].tsx:278]: I don't have a good sense about whether we should remove these linked experiments from the endpoint response, but I suspect no...

Comment (tzjames) [reply packages/front-end/pages/features/[fid].tsx:278]: From what I can tell that endpoint is only called here. Since we have all experiments with useExperiments we could generate the linkedExperiments on the front-end like we do in isFeatureStale.

tzjames (approved):
Looks good and should fix the bug.

If you want I think we could just call useExperiment from isFeatureStale and don't need to pass it through as the two places we call it already have useExperiment in it.  

I think we can do away with fetching the linkedExperiments in the backend as well, as we already calculate it on the front-end.  That is probably better done in a separate refactor PR.  

I could also see renaming "experiments" variable to be "linkedExperiments" for clarity.

packages/front-end/pages/features/[fid].tsx
Comment (bryce-fitzsimons) [reply packages/front-end/pages/features/[fid].tsx:278]: that's what I gathered as well, but would like to chat with a few folks before cleaning this up further. We'll keep it as a fast follow

PR #4404, opened 2025-08-12
Implement new experiment results popover UI

bryce-fitzsimons:
- Update results tooltip to new style guide
- Fix bugs with dashboard results tables (stats engine locked to org default, Bayesian violin plots broken)

image

image

image

### DIFF
--- a/packages/front-end/components/Experiment/BreakDownResults.tsx
+++ b/packages/front-end/components/Experiment/BreakDownResults.tsx
@@ -81,7 +81,7 @@ const BreakDownResults: FC<{
   secondaryMetrics: string[];
   guardrailMetrics: string[];
   metricOverrides: MetricOverride[];
-  id?: string;
+  idPrefix?: string;
   dimensionId: string;
   dimensionValuesFilter?: string[];
   isLatestPhase: boolean;
@@ -117,7 +117,7 @@ const BreakDownResults: FC<{
   goalMetrics,
   secondaryMetrics,
   metricOverrides,
-  id,
+  idPrefix,
   guardrailMetrics,
   isLatestPhase,
   phase,
@@ -387,7 +387,7 @@ const BreakDownResults: FC<{
               columnsFilter={columnsFilter}
               rows={table.rows}
               dimension={dimension}
-              id={(id ? `${id}_` : "") + table.metric.id}
+              id={(idPrefix ? `${idPrefix}_` : "") + table.metric.id}
               hasRisk={_hasRisk}
               tableRowAxis="dimension" // todo: dynamic grouping?
               labelHeader={
--- a/packages/front-end/components/Experiment/BreakDownResults.tsx
+++ b/packages/front-end/components/Experiment/BreakDownResults.tsx
@@ -81,6 +81,7 @@ const BreakDownResults: FC<{
   secondaryMetrics: string[];
   guardrailMetrics: string[];
   metricOverrides: MetricOverride[];
+  id?: string;
   dimensionId: string;
   dimensionValuesFilter?: string[];
   isLatestPhase: boolean;
@@ -116,6 +117,7 @@ const BreakDownResults: FC<{
   goalMetrics,
   secondaryMetrics,
   metricOverrides,
+  id,
   guardrailMetrics,
   isLatestPhase,
   phase,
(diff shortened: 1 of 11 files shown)


packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentMetricBlock.tsx
```diff
@@ -37,10 +34,7 @@ export default function ExperimentMetricBlock({
     [experiment, metricGroups]
   );
 
-  const statsEngine =
-    ssrPolyfills?.useOrgSettings()?.statsEngine ||
-    hookStatsEngine ||
-    "frequentist";
+  const statsEngine = analysis.settings.statsEngine;
```
lukesonnet, line 40: Just curious why this change was made?

packages/front-end/components/FlagCard/FlagCard.tsx
```diff
@@ -218,6 +217,26 @@ export default function FlagCard({
...
+
+        {!data?.isGuardrail && data.rowResults.suspiciousChange ? (
+          <CardItem
+            label="Suspicious"
+            tooltip={data.rowResults.suspiciousChangeReason}
+            value={<span style={{ color: "var(--pink-a11)" }}>suspicious</span>}
+          />
+        ) : null}
+
+        {data.rowResults.guardrailWarning ? (
```
lukesonnet, line 234: Where did this come from? It seems like previously this was only used for Safe Rollouts; did we want to highlight non-significant negative guardrail trends here?

It already will probably have a risk warning, and I don't know if it adds a ton. Kind of nbd either way, just wondering if it's additive.

Screenshot 2025-08-13 at 9 44 28 AM

I think something like this makes sense if we continue with the overhaul of how guardrails work in an experiment and make them one-sided or something.

packages/front-end/components/FlagCard/FlagCard.tsx
```diff
@@ -218,6 +217,26 @@ export default function FlagCard({
             }
           />
         ) : null}
+
+        {!data?.isGuardrail && data.rowResults.suspiciousChange ? (
+          <CardItem
+            label="Suspicious"
+            tooltip={data.rowResults.suspiciousChangeReason}
+            value={<span style={{ color: "var(--pink-a11)" }}>suspicious</span>}
```
lukesonnet: See convo in slack about this.

lukesonnet (approved):

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentMetricBlock.tsx
bryce-fitzsimons: The snapshot has its own stats engine that doesn't necessarily match the org settings, but rather whatever was selected in that experiment at the time of generation.

packages/front-end/components/FlagCard/FlagCard.tsx
bryce-fitzsimons: It's been there awhile, pre DMF. I do not know if we want to keep this or not.

packages/front-end/components/Experiment/BreakDownResults.tsx
```diff
@@ -385,7 +387,7 @@ const BreakDownResults: FC<{
               columnsFilter={columnsFilter}
               rows={table.rows}
               dimension={dimension}
-              id={table.metric.id}
+              id={(id ? `${id}_` : "") + table.metric.id}
```
ahdriel: suggestion: Should we name this `idPrefix`?

packages/front-end/components/FlagCard/FlagCard.tsx
lukesonnet: I'm indifferent now that I see it was indeed there before.

packages/front-end/components/Experiment/BreakDownResults.tsx
bryce-fitzsimons: we could, was trying to match the naming from CompactResults

bryce-fitzsimons: changed it

packages/front-end/components/FlagCard/FlagCard.tsx
bryce-fitzsimons: changed the value to be a bit more informative

PR #4410, opened 2025-08-13
update llms.txt generation config

royalfig:
The llms.txt file was being generated with the wrong URL. This updates the path config for llms.txt gen


ahdriel (approved):
The new llms.txt links properly to our docs! Nice fix.

I also noticed that the order of the listed links changed. Not sure of the impact and if it is intentional or not, but wanted to point it out.

royalfig:
@ahdriel that was intentional but something I'll likely rework in the future

PR #4411, opened 2025-08-13
feat(Callout): Add iconTooltip prop for ease of use

ahdriel:
### Features and Changes

In many scenarios we want to provide additional explanation about small callouts, so we are now providing `iconTooltip` that can be used for that.

Screenshot 2025-08-13 at 2 09 16 PM


ahdriel:
Will keep using the old pattern of icon with tooltip at the end of the text, so we don't need this.

PR #4412, opened 2025-08-13
Use shared component for Experiment Results popover/tooltip

ahdriel:
### Features and Changes

In #4404 we updated the main experiment results table to use the new design.
Now we are merging those changes with `AnalysisResultPopover` and using that component for both Experiment Results Table and Safe Rollout.

Here's an easier way to see a diff of what's changing in the render of ResultsTableTooltip: https://www.diffchecker.com/9WWyNVAS/

We are also adding example data for different scenarios in `/design-system`.


packages/front-end/pages/design-system/index.tsx
```diff
@@ -57,6 +66,251 @@ export default function DesignSystemPage() {
   const [selectValue, setSelectValue] = useState("carrot");
   const [activeControlledTab, setActiveControlledTab] = useState("tab1");
 
+  // Mock data for AnalysisResultPopover scenarios
```
ahdriel, line 70: Will keep this here for now but I am thinking about a different structure that might be more scalable as we mock more scenarios for different components.

packages/front-end/components/AnalysisResultPopover/AnalysisResultPopover.tsx
```diff
@@ -394,6 +426,7 @@ export default function AnalysisResultPopover({
                     getFactTableById={
                       ssrPolyfills?.getFactTableById || getFactTableById
                     }
+                    asTd={false}
                   />
                 </TableCell>
               </TableRow>
```
bryce-fitzsimons, line 415: these shouldn't be here since they are now rendered in `<FlagCard>`:
```
      <Flex direction="column" gap="2">
        {maybeRenderLiftWarning()}
        {maybeRenderSuspiciousChange()}
      </Flex>
```

packages/front-end/pages/design-system/index.tsx
bryce-fitzsimons, line 70: can we move this out of index.tsx to declutter?

packages/front-end/pages/design-system/index.tsx
```diff
@@ -925,6 +1179,237 @@ export default function DesignSystemPage() {
           <ResultsIndicator results="lost" />
         </Flex>
       </div>
+
+      <div className="appbox p-3">
+        <h3>Analysis Result</h3>
+        <Flex direction="column" gap="4">
+          <div>
+            <b>Frequentist, Relative, Significant Win (p-value adjusted)</b>
+            <Flex gap="3" mt="2">
```
bryce-fitzsimons, line 1189: can we put an outlined box around these so it replicates an actual popup?

packages/front-end/components/AnalysisResultPopover/AnalysisResultPopover.tsx
ahdriel: Good catch! It means that we were rendering the Suspicious Change twice for Safe Rollouts (both in FlagCard & AnalysisResultPopover).

I don't see the lift warning in the FlagCard though -- so I am keeping this one around.

packages/front-end/pages/design-system/index.tsx
ahdriel: See: https://github.com/growthbook/growthbook/pull/4412#discussion_r2274723537

Answer is yes, and I plan on doing it as a fast-follow because it opens a different set of questions around how we organize our code too.

packages/front-end/pages/design-system/index.tsx
ahdriel: done!

bryce-fitzsimons (approved):
LGTM

PR #4420, opened 2025-08-15
Add info about ID tokens for MS Azure AD

august-growthbook:
An Enterprise customer reported that the `User.read` permission associated with MS Graph does not work, since GrowthBook uses OIDC which is not compatible with MS Graph. Thus, ID tokens need to be enabled.

### DIFF
--- a/docs/docs/sso.mdx
+++ b/docs/docs/sso.mdx
@@ -203,7 +203,7 @@ When setting up Auth0, please ensure that you've enabled offline access, and che
 
 :::note
 
-In Azure, register an Application, instead of Enterprise, as we use OpenID Connect, not SAML.
+In Azure, register an Application, instead of Enterprise, as we use OpenID Connect, not SAML. Ensure the Application has ID tokens enabled.
 
 :::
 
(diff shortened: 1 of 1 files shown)


docs/docs/sso.mdx
```diff
@@ -212,7 +212,7 @@ In Azure, register an Application, instead of Enterprise, as we use OpenID Conne
   <div>
     <ol>
       <li>
-        Register an Application (we use OpenID Connect, so choose regular app, not Enterprise)
+        Register an Application (we use OpenID Connect, so choose regular app, not Enterprise). Because GrowthBook uses OIDC, ensure ID tokens are enabled, as the Microsoft Graph permission `User.read` does not work with OIDC.
```
tzjames, line 215:
```suggestion
        Register an Application (we use OpenID Connect, so choose regular app, not Enterprise). Ensure ID tokens are enabled, as the Microsoft Graph permission `User.read` does not work with OIDC.
```

tzjames: I feel like that first phrase is not quite needed as in the parenthesis right before it we say "we use OpenID Connect", so no need to repeat it.  

I also don't quite understand the second bit, but maybe that's just cause I'm not staring at their UI.  I don't understand how having "ID tokens enabled" is related to Microsoft Graph permission `User.read`?  Does checking that disable Microsoft Graph?

tzjames (approved):

PR #4423, opened 2025-08-18
Update Dashboards with Dimensions

Kevin-Chant:
### Features and Changes

Adds a helper for updating all dashboards associated with an experiment, and hooks into the `createSnapshot` flow to run that update whenever a `standard` snapshot is created.
Creates the minimum amount of snapshots needed to cover all dashboards with auto-update enabled, and saves those snapshots into the dashboards on query completion.
Refreshing the experiment's results page will also refresh all dashboards, but refreshing a single dashboard will not affect the others.
Also enables the use of precomputed dimensions within dashboards

### Testing

1. Create a dashboard with at least one dimension block, at least one sql explorer block, and some regular metric results. Make sure that auto-update is enabled
2. Inspect the snapshot IDs used for the dashboard via the network tab (request to `/dashboards/${dashId}/snapshots`)
3. The update icon should show the `nextSnapshotAttempt` for the experiment (or "starting soon" when it's not scheduled for the future)
4. Restart your server or use mongo to start the `updateExperimentResults` job manually
5. Observe that the snapshots have been updated and at least one was triggered by `update-dashboards`
6. Create a second dashboard, add a dimension block, and manually refresh the results
7. Check that the snapshots for the first dashboard are unchanged
8. Refresh the experiment's results tab and observe that both dashboards update their snapshots (in the background without a loading state)

### DIFF
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/index.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/index.tsx
@@ -265,16 +265,6 @@ export default function DashboardBlock<T extends DashboardBlockInterface>({
       })}
       direction="column"
     >
-      {blockHasFieldOfType(block, "snapshotId", isString) && (
-        <>
-          <Text>
-            {block.snapshotId} : {snapshot?.runStarted?.toString()}
-          </Text>
-          <Text>
-            {snapshot?.type} / {snapshot?.triggeredBy}
-          </Text>
-        </>
-      )}
       {isEditing && !editingBlock && disableBlock === "none" && (
         <div
           style={{


packages/back-end/src/jobs/updateDashboardResults.ts
```diff
@@ -0,0 +1,268 @@
...
+
+      const {
+        regressionAdjustmentEnabled,
+        settingsForSnapshotMetrics,
+      } = await getSettingsForSnapshotMetrics(context, experiment);
+
+      const metricMap = await getMetricMap(context);
+      const factTableMap = await getFactTableMap(context);
+
+      let reweight =
```
bryce-fitzsimons: Not sure what's going on here, but it doesn't seem safe to run this calc on a dashboard. The reweight decision for bandits should be part of the main experiment refresh only.

packages/back-end/src/jobs/updateDashboardResults.ts
```diff
@@ -0,0 +1,268 @@
...
+          phaseIndex: experiment.phases.length - 1,
+          defaultAnalysisSettings: analysisSettings,
+          additionalAnalysisSettings: getAdditionalExperimentAnalysisSettings(
+            analysisSettings
+          ).concat(additionalAnalysisSettings),
+          settingsForSnapshotMetrics: settingsForSnapshotMetrics || [],
+          metricMap,
+          factTableMap,
+          useCache: true,
+          type: "standard",
```
bryce-fitzsimons: is this what we actually want, a "standard" snapshot generated by the dashboard? These are the "official" snapshots for the experiment, which I think would be better managed by the standard jobs, right?

packages/shared/src/enterprise/dashboards/utils.ts
```diff
@@ -92,6 +97,30 @@ export function getBlockAnalysisSettings(
   };
 }
 
+export function blockSnapshotSettingsMatch<
+  B extends DashboardBlockInterfaceOrData<DashboardBlockInterface>
+>(snapshotSettings: ExperimentSnapshotSettings, block: B) {
+  const blockSettings = getBlockSnapshotSettings(block);
+  // TODO: check this logic w/ Luke/Bryce
+  if (!blockSettings.dimensionId)
+    return snapshotSettings.dimensions.length === 0;
```
bryce-fitzsimons: It's been a long while since I worked with this, but I think that snapshots can optionally have dimensions as well as the standard analysis. So returning false here just because `snapshotSettings.dimensions.length === 0` might not be appropriate since these snapshots could also have the mainline analysis.

packages/back-end/types/experiment-snapshot.d.ts
```diff
@@ -136,7 +136,7 @@ export interface ExperimentSnapshotAnalysisSettings {
   };
 }
 
-export type SnapshotType = "standard" | "exploratory" | "report";
+export type SnapshotType = "standard" | "exploratory" | "report" | "dashboard";
```
lukesonnet: My main concern with this flow is that sometimes we rely on "standard" as the main snapshot analysis and I'm worried about a variety of snapshots all getting out of sync with one another.

I think we might want to consider one of the two following flows (cc: @jdorn or @bryce-fitzsimons for input here)...

1. Update dashboards when a "main" snapshot gets updated, and just also update the dashboard with that experiment. That will lead us to a situation where you just have one update "schedule" per experiment and it updates the dashboard and the experiment. This makes the most sense for "Experiment" dashboards, but maybe less sense for generic dashboards. Alternatively, we could let experiment dashboards be linked to the experiment schedule (e.g. that kicks off dashboard updates), and only run the dashboard updater for non-experiment dashboards.

Or

2. Let dashboards update on their own, and if the main experiment udpated in the last X hours, use that timestamp for new snapshots and don't kick off the main snapshot, just use that one that ran X hours ago.


In either case, I think we don't want a new `SnapshotType`, I think we want `SnapshotTriggeredBy` to be `"schedule" | "manual" | "dashboard"`; or something like that. Because in either case, the dashboard should use the `"standard"` type for the standard blocks, if they exist. This is more in line with our initial thought of dashboards as a way to share the "main" results.

If we think of dashboards more as ways to do custom analyses, then this is a little murkier since you may never have the "standard" block in the dashboard, but in any case I think we should improve the linkage between the dashboards and the standard snapshot in this PR.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/index.tsx
```diff
@@ -265,6 +265,16 @@ export default function DashboardBlock<T extends DashboardBlockInterface>({
...
+      {blockHasFieldOfType(block, "snapshotId", isString) && (
+        <>
+          <Text>
+            {block.snapshotId} : {snapshot?.runStarted?.toString()}
+          </Text>
+          <Text>
+            {snapshot?.type} / {snapshot?.triggeredBy}
+          </Text>
+        </>
+      )}
```
Kevin-Chant: For ease of debugging, will be removed before merging

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -447,8 +446,8 @@ export async function postGenerateSQL(
             tableName: table.tableName,
           });
           dbSchemas.set(table.id, tableSchemaData);
-        } catch (error) {
-          errors.push(error);
```
Kevin-Chant, line 451: This was causing a typecheck error in the front-end due to a lengthy import chain. I opted to remove `errors` since it wasn't being used anywhere

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2843,6 +2847,8 @@ export async function createExperimentSnapshot({
     reweight,
     type: snapshotType,
     triggeredBy: triggeredBy ?? "manual",
+    precomputeDimensionsNonStandard,
+    startAnalysis,
```
lukesonnet: nit: Should it read "doNotStartAnalysis" or the like so that the default behavior clearer than in the coalesce we do inside the method? That name sucks though so totally up to you.

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -295,22 +296,41 @@ export async function updateSnapshot({
...
+        },
+      );
+    }
+  };
 
+  if (
+    experimentSnapshotModel.status === "success" &&
+    // Only use main snapshots or those triggered specifically for dashboards
+    (experimentSnapshotModel.triggeredBy === "update-dashboards" ||
+      experimentSnapshotModel.type === "standard")
```
lukesonnet, line 327: Won't this cause other dashboards to be updated by one dashboard that is updated manually if they happen to be on the same experiment and a snapshot satisfies the other dashboard's blocks?

packages/back-end/src/services/experiments.ts
```diff
@@ -1167,14 +1187,32 @@ export async function createSnapshot({
...
   );
-  await queryRunner.startAnalysis({
-    snapshotType: type,
-    snapshotSettings: data.settings,
-    variationNames: experiment.variations.map((v) => v.name),
-    metricMap,
-    queryParentId: snapshot.id,
-    factTableMap,
-  });
+  if (startAnalysis ?? true) {
```
lukesonnet: Line 1134 above means it still gets put into the DB in the "running" state. Doesn't that still cause the main results to think this snapshot is running?

packages/back-end/src/services/experiments.ts
```diff
@@ -3327,3 +3365,132 @@ export async function validateExperimentData(
...
+    [BlockSnapshotSettings, ExperimentSnapshotAnalysisSettings]
+  >((block) => {
+    const blockSnapshot = previousSnapshotMap.get(block.snapshotId);
+    if (!blockSnapshot || !blockSnapshot.analyses[0])
+      throw new Error(
+        "Error updating dashboard results, could not find snapshot",
+      );
+    return [
+      getBlockSnapshotSettings(block),
+      getBlockAnalysisSettings(block, blockSnapshot.analyses[0].settings),
```
lukesonnet: Is it always the 0th analysis in this case?

packages/back-end/src/services/experiments.ts
```diff
@@ -3327,3 +3365,132 @@ export async function validateExperimentData(
...
+      regressionAdjustmentEnabled,
+      snapshotSettings.dimensionId,
+    );
+
+    const queryRunner = await createSnapshot({
+      experiment,
+      context,
+      phaseIndex: experiment.phases.length - 1,
+      defaultAnalysisSettings: analysisSettings,
+      additionalAnalysisSettings: getAdditionalExperimentAnalysisSettings(
```
lukesonnet, line 3467: Is this necessary so it is easy to power the difference type dropdowns in the Dashboard for these blocks? I guess it's pretty harmless to add these on.

packages/back-end/src/controllers/experiments.ts
Kevin-Chant: Yeah I was struggling with naming but I think I'll swap it to "preventStartingAnalysis"

packages/back-end/src/models/ExperimentSnapshotModel.ts
Kevin-Chant: No because the manual update flow doesn't use `triggeredBy: "update-dashboards"`, but that's pretty weird so I cleaned it up by separating out `manual-dashboard` for contrast

packages/back-end/src/services/experiments.ts
Kevin-Chant: It does initially but in dashboards controller it checks whether the snapshot is needed and deletes it if not. Probably not as clean as the optimal solution but I can confirm it doesn't visually affect the main results
```
else {
    await deleteSnapshotById(context.org.id, mainSnapshot.id);
  }
```

packages/back-end/src/services/experiments.ts
Kevin-Chant: Nice catch, thanks

packages/back-end/src/services/experiments.ts
Kevin-Chant: Yeah it's not necessary because they can always just post a new analysis with their settings when the page is loaded, but I figured it'd be better to just include them to begin with

packages/shared/src/enterprise/dashboards/utils.ts
```diff
@@ -103,15 +108,32 @@ export function getBlockAnalysisSettings(
...
 
-export function dashboardCanAutoUpdate({
-  blocks,
-}: {
-  blocks: DashboardBlockInterfaceOrData<DashboardBlockInterface>[];
-}) {
-  // Only update dashboards where all the blocks will stay up to date with each other
-  return !blocks.find((block) =>
-    ["sql-explorer", "experiment-dimension"].includes(block.type),
+export function snapshotSatisfiesBlock(
```
bryce-fitzsimons, line 111: This method seems to be for checking whether we can (re)use a snapshot if the block and snapshot dimensions match (or exist, for precomputed). Are there other facets of a snapshot that would make them unusable if there isn't a match? For example, `phase`.

I also wonder about some of the upcoming features like metric dimensions/slices which unfurl a bunch of ephemeral metric-dimension rows into the `snapshot.analyses` and `snapshot.settings.metricSettings`. If one dashboard block has more of of these rows than another, would they be able to share a snapshot? I don't have a good sense of how things are plumbed for dashboards, so maybe this is not a real issue.

Kevin-Chant: For posterity - we talked about this offline and came to the conclusion that it's not an issue yet but will need to be kept up to date as we release changes to snapshots

PR #4425, opened 2025-08-18
CRUD Rest API endpoints for Segments

mknowlton89:
### Features and Changes

Recently, I released a PR that adds the `managedBy` field to `Segments`. I had planned a fast follow to update our REST API endpoints for Segments to utilize this new property. I didn't realize at the time that we didn't fully support CRUD operations for Segments via the REST API - only `GET` and `LIST`.

This PR adds full CRUD endpoints for Segments.

One thing to callout, with Segments, GrowthBook supports two different `TYPES` - we support `SQL` Segments and `Fact Table` Segments. The create and update endpoints support both types. I opted for pretty opinionated validation on those endpoints, open to feedback if others think I'm being too opinionated. But essentially, if it's a `SQL` Segment, I'm throwing an error if any of the `Fact Table` Segment properties are included (`factTableID` and `filters`). Likewise, if it's a `Fact Table` Segment, I'm throwing an error if any `sql` is provided.

Additionally, in the `updateSegments` endpoint, I'm blocking a user's ability to update a Segment from a `SQL` type to a `Fact Table` type and visa versa. There are just too many edge cases to handle comfortably, and I don't think it'll be a common pattern anyway.

If this does turn into an issue, we could always update, or have a separate endpoint responsible for handling this change.

## Important Callout
When the original `Segment.yaml` was created, it was done in a way where the properties are slightly mismatched with that of a `Segment`. Namely, the Segment.yaml, uses `identifierType` instead of `userIdType`, `query` instead of `sql`, and `datasourceId` instead of `datasource`.

I've done this PR in a way that honors those original properties, despite their being off from the corresponding `Segment` type our app uses in order to maintain maximum backwards compatibility. 

I also noticed `toSegmentApiInterface` was missing the `projects` property (this might've been due to the function being created before Segments supported project-scoping). I also updated this to support the new `managedBy` property, and also removed the fallback for the `userIdType` as that field is required and not optional, so we shouldn't ever have to use the fallback.

### Dependencies

- None

### Testing

- [x] Ensure the LIST endpoints works correctly
- [x] Ensure the GET /:id endpoint works correctly
- [x] Ensure you're able to delete a Segment correclty
- [x] Ensure you can create a SQL based Segment correctly
- [x] Ensure you can create a Fact Table based Segment correctly
- [x] Ensure you can update a Segment correctly (note: we don't support updating a segment in a way that changes it's TYPE)
- [x] Ensure that if you pass in a factTableId and/or filter properties for a SQL Segment, an error is thrown
- [x] Ensure that if you pass in a sql property for a Fact Table Segment, an error is thrown
- [x] Ensure that if you don't pass in a sql property for a SQL Segment, an error is thrown
- [x] Ensure that if you don't pass in a factTableId property for a Fact Table Segment, an error is thrown


packages/back-end/src/services/segments.ts
```diff
@@ -6,7 +6,7 @@ export function toSegmentApiInterface(segment: SegmentInterface): ApiSegment {
     id: segment.id,
     name: segment.name,
     owner: segment.owner || "",
-    identifierType: segment.userIdType || "user_id",
```
mknowlton89: So when `toSegmentApiInterface` was created, it was done with some deviations - the SegmentInterface uses: `userIdType`, `datasource`, and `sql` properties instead of `identifierType`, `datasourceId`, and `query` properties.

I was running into a lot of issues with the `identifierType` property, so I've adjusted it here. But I'm pretty hesitant since our API isn't versioned.

It's also really odd to me that the shape of the data the REST API returns is different from that with which the user needs to pass in. But again, hesitant to change because our API isn't versioned.

@bryce-fitzsimons Curious if you have an opinion on the best way to go?

mknowlton89: @bryce-fitzsimons I ultimately chose to go the direction of not updating this to change identifierType to userIdType. As noted in the PR description, there are a few other fields that are slightly mismatched, but I think we should probably leave them in order to maintain maximum backwards compatibility.

bryce-fitzsimons: Regarding the fallback value ("user_id"), probably best to leave it in for legacy purposes, unless we want to write a migration.

We chatted about renaming fields, and unfortunately are probably stuck with the schema that currently exists. Something to consider for API v2 perhaps...

bryce-fitzsimons (approved):
Minor change requested, otherwise looks good. Preemptively approving

PR #4438, opened 2025-08-19
Holdout Improvements

msamper:
### Features and Changes

Miscellaneous improvements to holdouts based on feedback from the initial version


packages/back-end/src/controllers/features.ts
```diff
@@ -2436,10 +2436,19 @@ export async function deleteFeatureById(
...
-        feature.id,
-      );
+      try {
+        await context.models.holdout.removeFeatureFromHoldout(
+          feature.holdout.id,
+          feature.id,
+        );
+      } catch (e) {
+        // This is not a fatal error, so don't block the request from happening
+        req.log.warn("Error removing feature from holdout", {
```
gazzdingo: should we use our own error logger and just use warning on that? i dont think that will block the request

packages/front-end/components/Experiment/TabbedPage/ExperimentActionButtons.tsx
```diff
@@ -25,7 +25,9 @@ export default function ExperimentActionButtons({
     runningStatus === "rollback-now";
   const displayCTAText = () => {
     if (holdout) {
-      return !holdout?.analysisStartDate ? "Start Analysis" : "Stop Holdout";
+      return !holdout?.analysisStartDate
+        ? "Start Analysis Period"
```
gazzdingo, line 29: did you get this copy from luke i thought it was going to be something like stop Experiment Enrollment

msamper: luke approved!

gazzdingo (approved):
LGTM

PR #4446, opened 2025-08-21
Upgrades ECharts from 5.6.0 to 6.0

mknowlton89:
### Features and Changes

One of the improvements we've been interested in with ECharts V6 is the x-axis label layout.

Currently, if there are a large number of dimensions, the dimension labels overlap the a-axis label as seen below.
V5.6.0
Screenshot 2025-08-21 at 1 04 15 PM

Echarts V6.0 does improve this experience
Screenshot 2025-08-21 at 1 08 00 PM

Unfortunately, on small screens or with a very large number of dimensions, the issue persists.

V5.6.0
Screenshot 2025-08-21 at 1 30 58 PM

V6.0
Screenshot 2025-08-21 at 1 25 33 PM

After doing some deep digging into ECharts docs, I came across some docs that said we could give the legend a `type = 'scroll'` property, which can be used to add a horizontal scroll to the list of dimensions, in this case.

I did try a vertical scroll, but it unfortunately doesn't work well with our types of graphs.

Screenshot 2025-08-28 at 11 31 48 AM

### Testing

- [x] Ensure v6 doesn't introduce any regressions
- [x] Ensure the visuals look good on all screen sizes and the horizontal scroll works both in the modal, and on the saved-query id page.

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -2427,7 +2427,7 @@ export default abstract class SqlIntegration
       metric.cappingSettings.type === "percentile" &&
       !!metric.cappingSettings.value &&
       metric.cappingSettings.value < 1 &&
-      !isCappableMetricType(metric);
+      isCappableMetricType(metric);
 
     const capCoalesceMetric = this.capCoalesceValue({
       valueCol: `m.${alias}_value`,
@@ -3192,14 +3192,14 @@ export default abstract class SqlIntegration
       metric.cappingSettings.type === "percentile" &&
       !!metric.cappingSettings.value &&
       metric.cappingSettings.value < 1 &&
-      !isCappableMetricType(metric);
+      isCappableMetricType(metric);
 
     const denominatorIsPercentileCapped =
       denominator &&
       denominator.cappingSettings.type === "percentile" &&
       !!denominator.cappingSettings.value &&
       denominator.cappingSettings.value < 1 &&
-      !isCappableMetricType(denominator);
+      isCappableMetricType(denominator);
     const capCoalesceMetric = this.capCoalesceValue({
       valueCol: "m.value",
       metric,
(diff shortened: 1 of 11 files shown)


packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -464,6 +464,7 @@ export function DataVisualizationDisplay({
                 color: textColor,
               },
               top: "bottom",
+              type: "scroll",
```
ahdriel, line 467: question: Do we want to keep this?

ahdriel (approved):
Thanks for the upgrade & validation 🎉 

One idea is that we can cap the length of the text and add a tooltip when the legend is hovered over:

```
legend: {
  textStyle: {
    color: textColor,
  },
  top: "bottom",
  formatter: (name: string) => {
    return name.length > 20 ? name.substring(0, 18) + "..." : name;
  },
  tooltip: {
    show: true,
  },
},
```

What do you think?

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
mknowlton89: @ahdriel Thanks for the review, and for the tooltip suggestion.

While v6 does reduce the overlap we experienced in v5, its still there in some cases.
Screenshot of with your code snippet above.
Screenshot 2025-08-28 at 2 59 48 PM

I did go even more aggressive with the truncation (truncated after 8 chars), but it the overlap issue can still persist at small screen sizes.

And while I don't LOVE the scroll, I do think its the better alternative as it means there shouldn't ever be an overlap issue, even on really small screen sizes.

E.G.
Screenshot 2025-08-28 at 3 06 15 PM

ahdriel: Thanks for checking! Yeah, I agree that scroll seems better in this case.

It looks like the overlap happens any time we have 2 lines, pushing the content towards the top (over the axis label) instead of creating a new line below it unfortunately.

LGTM! Merge it away

PR #4450, opened 2025-08-22
Dashboards redesign

Kevin-Chant:
### Features and Changes

Redesign of the experiment dashboards tab including moving the edit flow to a fullscreen takeover and changing the edit drawer into a sidebar


packages/back-end/src/enterprise/models/DashboardBlockModel.ts
```diff
@@ -155,3 +180,67 @@ export async function createDashboardBlock(
 
   return toInterface(block);
 }
+
+export function migrate(
```
bryce-fitzsimons, line 184: Do we need migrate? I thought dashboards weren’t public yet

Kevin-Chant: Unfortunately we do because the default dashboards have been created automatically despite the tab not being rendered. We could nuke the mongo collections but Jeremy pointed out that it'd break for self-hosted so we opted to add migrations

packages/front-end/styles/global-radix-overrides.scss
```diff
@@ -1,7 +1,7 @@
 // Button
 .rt-Button.rt-variant-ghost {
   margin: 0 !important;
-  &:not(.rt-IconButton) {
+  &:not(.rt-IconButton) &:not(.rt-SelectItem) {
```
jdorn: Is this a typo?

PR #4454, opened 2025-08-22
Ji/free managed warehouse

tzjames:
### Features and Changes

This is part of a 3 PR set, with a change in the [license server](https://github.com/growthbook/central-license-server/pull/80) and in the [ingestor](https://github.com/growthbook/growthbook-ingestor/pull/47).  The overall goal is to allow free orgs to also use the managed warehouse.  If they use more than 1M events per month those events will now go into a `overage_events` table instead of the main `enriched_events` table.  We show them a warning that they are over and prompt them to upgrade to pro.  If they do we copy over the events from the `overage_events` table and put it into the `enriched_events` table which will then load it into their materialized views.  

1. Adds `managedClickhouseEvents` as an optional limit and usage in `OrganizationUsage` type.
2. Uses the new /subscription/usages endpoint instead of the cdn/:id/usages endpoint when calling the license server
3. Modifies the data-enrichment endpoint such that all free orgs who are over the usage limit will have an `overLimit:true` property added to their `SdkInfo`.
4. Adds an AccountPlanNotice to let them know about the managedWarehouse overages.
5. Displays a Managed Warehouse Events graph on the /settings/usage page that combines the events in the `enriched_events` table and the `overage_events` table.
6. Modifies the add Growthbook Managed Warehouse modal to allow free orgs to sign up.
7. Modifies the upgrade modal to show pricing info for the managed warehouse.

### Dependencies
Both of the following need to land first:
- [ ] https://github.com/growthbook/central-license-server/pull/80 ... to get the new /subscription/usages endpoint
- [ ] https://github.com/growthbook/growthbook-ingestor/pull/47 ... so the ingestor can handle the new property.
- [ ] The overage_events clickpipe on production needs to be set up.

### Testing

1. https://www.notion.so/growthbook/Dev-Managed-Data-Warehouse-251a857e74a080eda450d70045cb3770
2. Do Test plan from [the license server PR](https://github.com/growthbook/central-license-server/pull/80) 
3. Set IS_CLOUD=true locally with a valid SSO_CONFIG.
4. Remove the license_key from the organization in Mongo.
5. Delete the Managed Warehouse from the organization.
6. See the overage AcountPlanNotice.
7. See in Clickhouse new events going into the overage_events table
8. Go to the datasource page
9. Click to create the Growthbook Managed Warehouse
10. Click on the Upgrade button in the AccountPlanNotice.
11. See the note about being able to Access all your Managed Warehouse Events
12. Upgrade to Pro.
13. See the events removed from the overage_events table.
14. See the events that had been in the overage_events table now visible in the organization's db. 

### Screenshots

Screenshot 2025-08-22 at 5 01 11 PM

Screenshot 2025-08-22 at 4 31 44 PM

### DIFF
--- a/packages/back-end/src/api/ingestion/ingestion.router.ts
+++ b/packages/back-end/src/api/ingestion/ingestion.router.ts
@@ -7,13 +7,18 @@ import {
 import { SDKConnectionInterface } from "back-end/types/sdk-connection";
 import { _dangerousGetSdkConnectionsAcrossMultipleOrgs } from "back-end/src/models/SdkConnectionModel";
 import { _dangerousGetAllGrowthbookClickhouseDataSources } from "back-end/src/models/DataSourceModel";
-import { getOrganizationIdsWithTrackingDisabled } from "back-end/src/models/OrganizationModel";
+import {
+  _dangerouslyFindAllOrganizationsByIds,
+  getOrganizationIdsWithTrackingDisabled,
+} from "back-end/src/models/OrganizationModel";
+import { getUsages } from "back-end/src/enterprise/billing";
 
 interface SdkInfo {
   organization: string;
   client_key: string;
   datasource: string;
   environment: string;
+  overLimit: boolean;
 }
 
 interface GetDataEnrichmentResponse {
@@ -22,12 +27,17 @@ interface GetDataEnrichmentResponse {
   };
 }
 
-function sdkInfo(conn: SDKConnectionInterface, datasource: string): SdkInfo {
+function sdkInfo(
+  conn: SDKConnectionInterface,
+  datasource: string,
+  overLimit: boolean,
+): SdkInfo {
   return {
     organization: conn.organization,
     client_key: conn.key,
     environment: conn.environment,
     datasource,
+    overLimit,
   };
 }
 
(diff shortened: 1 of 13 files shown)


packages/back-end/src/api/ingestion/ingestion.router.ts
```diff
@@ -7,13 +7,18 @@ import {
...
+  getOrganizationIdsWithTrackingDisabled,
+} from "back-end/src/models/OrganizationModel";
+import { getUsages } from "back-end/src/enterprise/billing";
 
 interface SdkInfo {
   organization: string;
   client_key: string;
   datasource: string;
   environment: string;
+  overLimit: boolean;
```
mknowlton89, line 21: Nit, but I wonder if changing this to something like `overManagedWarehouseLimit` would be more contextual - given there are a few different `limits` orgs may run into (cdn requests, bandwidth, etc)?

mknowlton89 (approved):

packages/back-end/src/api/ingestion/ingestion.router.ts
tzjames: Being in the ingestor router and only consumed by the ingestor I'm leaving this for now. (Too much to test again to change it.). If we ever do your excellent suggestion of moving the ingestion router to the license server to simplify this setup then we could consider renaming it then.

PR #4459, opened 2025-08-25
fix(dimensions): Dimension selector should reset other filters

lukesonnet:
### Features and Changes

The dimension selector was not properly re-setting the other filters (for non-precomputed dimensions) following #3899 . Instead of having two colliding side-effects from changing the dimension, this lets you control whether changing the dimension should directly have a side-effect on the other settings.

Pretty clunky, but until we re-work this, I think this is safer.

### Testing

Before and after in Loom:
https://www.loom.com/share/61e92bd0be2441c297933f7b9ac20edc

Also tested that switching dimensions WITH pre-computed dimensions (which does not require re-setting the other filters) works as expected.

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
@@ -230,8 +230,6 @@ export default function ExperimentHeader({
   const holdoutsEnabled =
     useFeatureIsOn("holdouts_feature") && hasHoldoutsFeature;
 
-  console.log("holdoutsEnabled", holdoutsEnabled);
-
   const hasUpdatePermissions = !holdout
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Dimensions/DimensionChooser.tsx
```diff
@@ -219,6 +218,8 @@ export default function DimensionChooser({
           initialOption="None"
           value={value}
           onChange={(v) => {
+            console.log(v);
+            console.log(value);
```
bryce-fitzsimons: 👁️

lukesonnet: ugh

ahdriel (approved):

lukesonnet:
Yes, it does.

Issue #4318, opened 2025-07-21
[Feature] Organization_id available on the UI


natasha-growthbook:
### Description of Feature

I would like to have the org_id available on the UI like it is available for airgapped licenses on [this](https://github.com/growthbook/growthbook/pull/2422) PR. 

Currently, this can be fetched using the REST API.

PR #4461, opened 2025-08-25
Renders org id on settings page for all orgs

mknowlton89:
### Features and Changes

Previously, in the `General Settings` section of an org's `/settings` page, we were only rendering the organization's id for self hosted org's that didn't have multiple organizations.

This PR removes that logic so all orgs, regardless of if they're self-hosted or using our cloud platform, and regardless of how many orgs they have, will see their org id listed below the org owner's email. 

Unlike the organization name and email, this field is read-only.

I have a self-hosted instance with multiple orgs and I can confirm that switching between them does work correctly, where the org id is updating correctly.

- Closes **(https://github.com/growthbook/growthbook/issues/4318)**

### Dependencies

- None

### Testing

- [x] Ensure that the org id renders for cloud and self-hosted orgs
- [x] Ensure that if self-hosted and the user is a member of multiple organizations, switching between the orgs updates the UI correctly.

### Screenshots

Before:
Screenshot 2025-08-25 at 2 34 23 PM

After:
Screenshot 2025-08-25 at 2 33 23 PM

### DIFF
--- a/packages/front-end/components/GeneralSettings/OrganizationAndLicenseSettings.tsx
+++ b/packages/front-end/components/GeneralSettings/OrganizationAndLicenseSettings.tsx
@@ -65,11 +65,9 @@ export default function OrganizationAndLicenseSettings({
                     </a>
                   )}
                 </Box>
-                {!isCloud() && !isMultiOrg() && (
-                  <Box>
-                    <Text weight="medium">Organization Id: </Text> {org.id}
-                  </Box>
-                )}
+                <Box>
+                  <Text weight="medium">Organization Id: </Text> {org.id}
+                </Box>
               </Flex>
               {canEdit && (
                 <Button


packages/front-end/components/GeneralSettings/OrganizationAndLicenseSettings.tsx
```diff
@@ -65,11 +65,9 @@ export default function OrganizationAndLicenseSettings({
                     </a>
                   )}
                 </Box>
-                {!isCloud() && !isMultiOrg() && (
```
mknowlton89, line 68: Reading through the GH Issue for this, I think the correct move is to just always show this, but let me know if there are any edge cases I need to be aware of that I might be missing.

bryce-fitzsimons (approved):

PR #4463, opened 2025-08-25
feat(SqlIntegration): Return columns for every query executed

ahdriel:
### Features and Changes

Certain databases allow us to execute a query with `LIMIT 0` where we do a dry-run and validate the query syntax & connection.

Another information we need to validate for the Incremental Refresh, is the returned column names.

So for BigQuery & Presto we are now explicitly returning `columns` as a `string[]`.


jdorn (approved):

packages/back-end/src/integrations/BigQuery.ts
```diff
@@ -76,8 +76,9 @@ export default class BigQuery extends SqlIntegration {
       await setExternalId(job.id);
     }
 
-    const [rows] = await job.getQueryResults();
+    const [rows, _, queryResultsResponse] = await job.getQueryResults();
```
ahdriel, line 79: This is now dependent on the upgrade of BigQuery library which should be done in a separate PR.

ahdriel: Unblocked as #4491 was merged.

PR #4464, opened 2025-08-26
Allow manual seed selection for rollouts & safe rollouts

### LINKED ISSUE #4460: [Feature] Allow manual seed selection for rollouts & safe rollouts
### Description of Feature

Currently we don't allow assigning seeds to rollout rules (nor safe rollouts) despite widespread SDK support. Currently all rollout rules on the same feature always overlap in the same hash space; rollouts in different features always use a different hash space. Sometimes you need to change this behavior (ex: (1) let rollouts stack up/ multiply, (2) keep consistent hashing between different features)

bryce-fitzsimons:
Closes #4460

Also starts unifying some UI around rules, as well as using the standard "units" label for rollouts and experiments (instead of "users").

image

image

### DIFF
--- a/packages/front-end/components/Features/RuleModal/RolloutFields.tsx
+++ b/packages/front-end/components/Features/RuleModal/RolloutFields.tsx
@@ -85,7 +85,7 @@ export default function RolloutFields({
             className="mb-3"
           />
           <SelectField
-            label="Enroll based on attribute"
+            label="Sample based on attribute"
             options={attributeSchema
               .filter((s) => !hasHashAttributes || s.hashAttribute)
               .map((s) => ({ label: s.property, value: s.property }))}
--- a/packages/front-end/components/Features/RolloutPercentInput.tsx
+++ b/packages/front-end/components/Features/RolloutPercentInput.tsx
@@ -14,7 +14,7 @@ export interface Props {
 export default function RolloutPercentInput({
   value,
   setValue,
-  label = "Percent of Users",
+  label = "Percent of Units",
   className,
 }: Props) {
   return (


packages/front-end/components/Features/RuleModal/RolloutFields.tsx
```diff
@@ -77,7 +82,7 @@ export default function RolloutFields({
             setValue={(coverage) => {
               form.setValue("coverage", coverage);
             }}
-            className="mb-1"
+            className="mb-3"
           />
           <SelectField
             label="Enroll based on attribute"
```
lukesonnet: Should this read `Sample based on attribute` to match the top-level text and be consistent also with safe rollouts?

Should we also update the RolloutPercentInput text above to read `Units` instead of users so at least this component is consistent?

packages/front-end/components/Features/RuleModal/RolloutFields.tsx
```diff
@@ -89,6 +94,34 @@ export default function RolloutFields({
...
+            >
+              Advanced Options{" "}
+              {!advancedOptionsOpen ? <PiCaretDown /> : <PiCaretUp />}
+            </span>
+            {advancedOptionsOpen && (
+              <div className="mt-3">
+                <Field
+                  label="Seed"
+                  type="input"
+                  {...form.register("seed")}
```
lukesonnet, line 117: Under the hood do we fall back to the `rule.id` or the `feature.id`?

lukesonnet (commented):
Two questions!

packages/front-end/components/Features/RuleModal/RolloutFields.tsx
bryce-fitzsimons: There are 2 hoods:
- Under the app level hood, we only put a rule seed into SDK payload if its non-empty.
- Under the SDK hood, we use `rule.seed` if truthy, otherwise fallback to the `feature.id`

lukesonnet: I see, I thought the `id` in the `rule.seed || id` line in `core.ts` was referring to the rule, but it refers to the feature. Thanks.

lukesonnet:
Have we done any testing to make sure it gets respected? If so, I'm happy to approve, unless someone else wants to take a pass.

packages/front-end/components/Features/RuleModal/RolloutFields.tsx
bryce-fitzsimons: Sure, i can make those updates. FYI, the input field will now say "Percent of Units", which is kind of odd, but 🤷

bryce-fitzsimons:
> Have we done any testing to make sure it gets respected? If so, I'm happy to approve, unless someone else wants to take a pass.

I've done some, but another set of eyes is always appreciated.

lukesonnet (approved):
Approving as it looks right but haven't been able to test.

PR #4473, opened 2025-08-27
Fix getAllResults() in GrowthBook, add to GrowthBookClient

### LINKED ISSUE #3375: [Bug] getAllResults returns an empty Map on the latest sdk-js version
### Summary

When upgrading to @growthbook/growthbook@1.3.0 the `growthBook.getAllResults()` method returns an empty Map. This caused our users not to be assigned to experiments. Downgrading did not fix the issue easily because the sdk-react [is pointing to the latest sdk-js internally](https://github.com/growthbook/growthbook/blob/main/packages/sdk-react/package.json#L41C6-L41C29), forcing us to make the downgrade the version manually.

#### Expected Behavior

When upgrading to @growthbook/growthbook@1.3.0 the `growthBook.getAllResults()` method returns the experiment results Map without requiring the subscription to the GrowthBook instance. If this was by design, it should have been announced as a breaking change. 

#### Current Behavior

When upgrading to @growthbook/growthbook@1.3.0 the `growthBook.getAllResults()` method returns an empty Map unless there's a subscription to the GrowthBook instance. 

Code Example: 

https://gist.github.com/fabianmarinog/89acfdc3fe922543544950f186617082

bryce-fitzsimons:
Closes #3375

### DIFF
--- a/packages/sdk-js/src/GrowthBook.ts
+++ b/packages/sdk-js/src/GrowthBook.ts
@@ -131,6 +131,7 @@ export class GrowthBook<
 
     this.log = this.log.bind(this);
     this._saveDeferredTrack = this._saveDeferredTrack.bind(this);
+    this._onExperimentEval = this._onExperimentEval.bind(this);
     this._fireSubscriptions = this._fireSubscriptions.bind(this);
     this._recordChangedId = this._recordChangedId.bind(this);
 
@@ -593,7 +594,7 @@ export class GrowthBook<
 
   public run<T>(experiment: Experiment<T>): Result<T> {
     const { result } = runExperiment(experiment, null, this._getEvalContext());
-    this._fireSubscriptions(experiment, result);
+    this._onExperimentEval(experiment, result);
     return result;
   }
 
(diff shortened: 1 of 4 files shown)


packages/sdk-js/src/GrowthBook.ts
```diff
@@ -830,18 +830,22 @@ export class GrowthBook<
     }
   }
 
-  private _fireSubscriptions<T>(experiment: Experiment<T>, result: Result<T>) {
-    const key = experiment.key;
+  private _onExperimentEval<T>(experiment: Experiment<T>, result: Result<T>) {
+    if (this._subscriptions.size > 0) {
+      this._fireSubscriptions(experiment, result);
+    }
+    this._assigned.set(experiment.key, { experiment, result });
```
jdorn: We should move this line above the subscriptions stuff.  Otherwise, calling `getAllResults` from inside a subscription function will not work as expected.

packages/sdk-js/src/GrowthBookClient.ts
```diff
@@ -369,6 +369,15 @@ export class UserScopedGrowthBook<
   private _gb: GrowthBookClient;
   private _userContext: UserContext;
   public logs: Array<LogUnion>;
+  private _assigned: Map<
```
jdorn: Do we need to add `getAllResults` to this class?  I'm not sure what benefit that provides over `userContext.trackedExperiments` which already exists.

I'm worried about the memory implications of this change given there will be potentially thousands of instances created per second and there's no destroy method.

bryce-fitzsimons: figured it was an oversight, but i can see not wanting this for GrowthBookClient

packages/sdk-js/src/GrowthBook.ts
bryce-fitzsimons: spoke offline, we'll need to send the new additions over to _fireSubscriptions so they aren't skipped

PR #4485, opened 2025-08-28
fix(sql): Improve handling of unit quantiles with ignore zero

### LINKED ISSUE #4477: [Bug] Route all fact metric queries through fact metric generator
### Summary

Quantile metrics fail when added to an experiment with "Exclude In-Progress Conversions" because they route through the old metric query generator. There may be other issues with fact metrics going through this old query generator.

We should resolve this by running all fact metrics through the fact metric query generator, even if they are singles.

lukesonnet:
### Features and Changes

Ignore zeros for unit quantiles was not updated to work with COUNT DISTINCT aggregations (and was a bit unsafe in implementation before anyways).

This fixes that by relying on `NULLIF({unit_aggregation}, 0)` in the user aggregation step so that the percentile methods will ignore 0 aggregate values as they will be null.

- Closes #4477 

### Testing

See Loom below.

### Screenshots

https://www.loom.com/share/2a25b676f8164f8e9b53f7ebea736b27

### DIFF
--- a/packages/front-end/components/FactTables/FactMetricModal.tsx
+++ b/packages/front-end/components/FactTables/FactMetricModal.tsx
@@ -2101,7 +2101,9 @@ export default function FactMetricModal({
                       <>
                         {form
                           .watch("numerator")
-                          ?.column?.startsWith("$$") ? undefined : (
+                          ?.column?.startsWith(
+                            "$$distinctUsers",
+                          ) ? undefined : (
                           <div className="col-auto">
                             <div className="form-group">
                               <label htmlFor="quantileIgnoreZeros">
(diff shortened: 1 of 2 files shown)


packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -2101,7 +2101,9 @@ export default function FactMetricModal({
                       <>
                         {form
                           .watch("numerator")
-                          ?.column?.startsWith("$$") ? undefined : (
+                          ?.column?.startsWith(
+                            "$$distinctUsers",
+                          ) ? undefined : (
```
lukesonnet, line 2119: We also can allow ignoring zeros after counting rows with `$$count`

PR #4488, opened 2025-08-28
[HOT-FIX] Fix bug with feature creation when no holdouts exist

msamper:
### Features and Changes

When submitting the new feature object to the api to create a feature, we were spreading the wrong variable when conditionally adding the holdout object depending on whether or not the id was `none`

### Testing

- [x] Feature creation works when no holdouts exist
- [x] Experiment creation works when no holdouts exist
- [x] Feature creation works when holdouts exist and holdout is selected
- [x] Experiment creation works when holdouts exist and holdout is selected
- [x] HoldoutSelect still defaults to the right value 
- [x] HoldoutSelect None option still works
- [x] Test Holdout deletion and ensure linkages are removed from Features & Experiments

### DIFF
--- a/packages/front-end/components/Features/FeatureModal/index.tsx
+++ b/packages/front-end/components/Features/FeatureModal/index.tsx
@@ -126,7 +126,9 @@ const genFormDefaultValues = ({
         tags: featureToDuplicate.tags,
         environmentSettings,
         customFields: customFieldValues,
-        holdout: featureToDuplicate.holdout,
+        holdout: featureToDuplicate.holdout?.id
+          ? featureToDuplicate.holdout
+          : undefined,
       }
     : {
         valueType: "" as FeatureValueType,
@@ -137,6 +139,7 @@ const genFormDefaultValues = ({
         tags: [],
         environmentSettings,
         customFields: customFieldValues,
+        holdout: undefined,
       };
 };
 
@@ -361,7 +364,7 @@ export default function FeatureModal({
         {holdoutsEnabled && (
           <HoldoutSelect
             selectedProject={selectedProject}
-            selectedHoldoutId={form.watch("holdout")?.id ?? ""}
+            selectedHoldoutId={form.watch("holdout")?.id}
             setHoldout={(holdoutId) => {
               form.setValue("holdout", { id: holdoutId, value: "" });
             }}
--- a/packages/front-end/components/Features/FeatureModal/index.tsx
+++ b/packages/front-end/components/Features/FeatureModal/index.tsx
@@ -232,14 +232,14 @@ export default function FeatureModal({
       submit={form.handleSubmit(async (values) => {
         const { defaultValue, ...feature } = values;
         const valueType = feature.valueType;
-        const { holdout, ...featureWithoutHoldout } = feature;
+        const { holdout } = feature;
 
         if (!valueType) {
           throw new Error("Please select a value type");
         }
 
         const newDefaultValue = validateFeatureValue(
-          featureWithoutHoldout,
+          feature,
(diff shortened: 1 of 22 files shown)


packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -69,6 +69,7 @@ export const HoldoutSelect = ({
   useEffect(() => {
     // If still loading, don't set anything
     if (loading) return;
+    if (!hasHoldouts || holdoutsWithExperiment.length === 0) return;
```
msamper: Doing this to not set `none` if there are no holdouts or if the org doesn't have the holdouts feature

gazzdingo (approved):
LGTM have not tested though

packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -82,36 +83,42 @@ export const HoldoutSelect = ({
...
       }
     }
-  }, [holdoutsWithExperiment, setHoldout, loading, selectedHoldoutId]);
+  }, [
+    holdoutsWithExperiment,
+    setHoldout,
+    loading,
+    selectedHoldoutId,
+    hasHoldouts,
+  ]);
```
jdorn: Code like this always makes me nervous.  We're calling setState methods inside useEffect with a ton of dependencies.  This makes it super easy for bugs to creep in and run this code more often than we want.  This would have the effect of overwriting things in the form the user has already selected.

packages/front-end/components/Features/FeatureModal/index.tsx
```diff
@@ -253,15 +256,14 @@ export default function FeatureModal({
...
+          ...featureWithoutHoldout,
           defaultValue: parseDefaultValue(defaultValue, valueType),
-          ...(holdout?.id &&
-            holdout.id !== "none" && {
-              holdout: {
-                id: holdout.id,
-                value: parseDefaultValue(defaultValue, valueType),
-              },
-            }),
+          ...(holdout?.id && {
```
jdorn: Is this still necessary?  Could we just spread `feature` and it will have `{ holdout: {id: "", ...} }` and the back-end will ignore it?

msamper: Not necessary. Happy to just remove it to simplify

packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -72,54 +72,53 @@ export const HoldoutSelect = ({
...
-      (!holdoutsWithExperiment.some((h) => h.id === selectedHoldoutId) &&
-        selectedHoldoutId !== "none")
+      !holdoutsWithExperiment.some((h) => h.id === selectedHoldoutId) &&
+      selectedHoldoutId !== ""
     ) {
       if (holdoutsWithExperiment.length === 0) {
-        setHoldout("none");
+        setHoldout("");
       } else {
         setHoldout(holdoutsWithExperiment[0].id);
```
jdorn, line 81: Does this `else` block ever actually fire now?  It looks like the intention is to default to the first holdout, but I don't think that works anymore.  I might just be reading the code wrong though.

msamper: Yeah it seems to be firing in my testing because the default form values for holdouts in both features and experiments are undefined and it's correctly setting the first holdout. I'll triple check though

msamper: confirmed it's firing

PR #4490, opened 2025-08-28
fix(phases-dates): Format phase start/end time in UTC

### LINKED ISSUE #4478: [Bug] datetime and date in phase selector don't match
### Summary

If the time in the phase editor is in the first 6 hours of a day, the date that truncates the time shows it as the previous day, likely due to some time zone error.

Image

Image

ahdriel:
### Features and Changes

Fixes #4478

When editing the phase, the datepicker correctly shows the time in UTC and it gets saved in Mongo also in UTC.
But when showing in the UI we are converting from UTC to the user's local time, leading to confusion.

So now we format the timestamp in the UTC timezone as well to ensure the values match.

https://www.loom.com/share/8e3b4dd171ad487eaf3b4e84a9e48879

### DIFF
--- a/packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
+++ b/packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
@@ -464,12 +464,6 @@ export function DataVisualizationDisplay({
                 color: textColor,
               },
               top: "bottom",
-              formatter: (name: string) => {
-                return name.length > 20 ? name.substring(0, 18) + "..." : name;
-              },
-              tooltip: {
-                show: true,
-              },
             },
           }
         : null),
--- a/packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
+++ b/packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
@@ -464,6 +464,12 @@ export function DataVisualizationDisplay({
                 color: textColor,
               },
               top: "bottom",
+              formatter: (name: string) => {
+                return name.length > 20 ? name.substring(0, 18) + "..." : name;
+              },
+              tooltip: {
+                show: true,
+              },
             },
           }
         : null),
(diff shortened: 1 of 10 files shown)


lukesonnet (changes requested):
Thanks Adriel. Can you also fix the date used in the analysis settings bar? You can see it on the loom next to the update button.

It may be worth it to check for other usages of `date` on the experiment page or even in the experiments list.

ahdriel:
@lukesonnet Good catch!

I searched for places where we use phase `dateStarted` & `dateEnded` and fixed the ones I found.

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -464,6 +464,12 @@ export function DataVisualizationDisplay({
                 color: textColor,
               },
               top: "bottom",
+              formatter: (name: string) => {
```
lukesonnet: intentional?

ahdriel: not at all -- leftover from a code review.

lukesonnet (approved):
For some of these we may want to make it clear the date is UTC (e.g. the timelines) but I think being consistent is more important so this lgtm.

packages/shared/src/dates.ts
```diff
@@ -5,14 +5,23 @@ import differenceInHours from "date-fns/differenceInHours";
 import addMonths from "date-fns/addMonths";
 import formatRelative from "date-fns/formatRelative";
 import previousMonday from "date-fns/previousMonday";
+import { formatInTimeZone } from "date-fns-tz";
 
-export function date(date: string | Date): string {
+export function date(date: string | Date, inTimezone?: string): string {
```
ahdriel, line 10: FWIW, if we upgrade to date-fns v4 a lot of this is built-in -- but that's a bigger project that should be done separately.

PR #4492, opened 2025-08-28
new end point to find by tracking key

gazzdingo:
### Features and Changes

### DIFF
--- a/packages/back-end/src/models/HoldoutModel.ts
+++ b/packages/back-end/src/models/HoldoutModel.ts
@@ -111,7 +111,7 @@ export class HoldoutModel extends BaseClass {
    * @returns The holdout
    */
   public async getByExperimentId(experimentId: string) {
-    const holdouts = await this._find({ experimentId });
-    return holdouts[0];
+    const holdout = await this._findOne({ experimentId });
+    return holdout;
   }
 }
--- a/packages/front-end/pages/holdouts/lookup.tsx
+++ b/packages/front-end/pages/holdouts/lookup.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect } from "react";
 import { useRouter } from "next/router";
 import LoadingOverlay from "@/components/LoadingOverlay";
 import Link from "@/components/Radix/Link";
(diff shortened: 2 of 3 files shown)


packages/back-end/src/models/HoldoutModel.ts
```diff
@@ -105,4 +105,13 @@ export class HoldoutModel extends BaseClass {
     const { [featureId]: _, ...linkedFeatures } = holdout.linkedFeatures;
     await this.updateById(holdoutId, { linkedFeatures });
   }
+  /*
+   * Gets the first holdout that is linked to an experiment
+   * @param experimentId - The id of the experiment
+   * @returns The holdout
+   */
+  public async getByExperimentId(experimentId: string) {
+    const holdouts = await this._find({ experimentId });
```
msamper: use `_findOne` instead

packages/front-end/pages/holdouts/lookup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const {
+    data,
+    isLoading,
+    error: apiError,
+  } = useApi<{ holdoutId: string | null }>(path || "", {
+    shouldRun: () => !!trackingKey,
+  });
+
+  useEffect(() => {
+    if (data && data.holdoutId && data.holdoutId !== "not found") {
```
msamper:
```suggestion
    if (data && data.holdoutId) {
```
I don't think the `holdoutId` would ever be `not found`

packages/front-end/pages/holdouts/lookup.tsx
```diff
@@ -0,0 +1,84 @@
...
+      router.push(`/holdout/${data.holdoutId}`);
+    }
+  }, [data, router]);
+
+  useEffect(() => {
+    if (apiError) {
+      setError(apiError.message || "An error occurred");
+    }
+  }, [apiError]);
```
msamper: Do you need a useEffect for this? I think you could use the error from `useApi` directly below

packages/front-end/pages/holdouts/lookup.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  if (error) {
+    return <div className="alert alert-danger">{error}</div>;
+  }
```
msamper:
```suggestion
  if (apiError) {
    return <div className="alert alert-danger">{apiError.message || "An error occurred"}</div>;
  }
```

packages/front-end/pages/holdouts/lookup.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  if (isLoading) {
+    return <LoadingOverlay />;
+  }
+
+  // Show different messages based on the response
+  if (data) {
+    if (data.holdoutId && data.holdoutId !== "not found") {
```
msamper:
```suggestion
    if (data.holdoutId) {
```

packages/front-end/pages/holdouts/lookup.tsx
```diff
@@ -0,0 +1,84 @@
...
+          Error looking up holdout: {apiError.message}
+        </Callout>
+        <div className="mt-3 px-3">
+          <Link href="/holdouts" size="3">
+            All Holdouts
+          </Link>
+        </div>
+      </div>
+    );
+  }
```
msamper: Seems like this would never get hit because of the conditional checking the error above. This seems more stylized and helpful though. Maybe move this up and replace with the condition on lines 36-38

packages/front-end/pages/holdouts/lookup.tsx
```diff
@@ -0,0 +1,84 @@
...
+            No holdout found for tracking key: <code>{trackingKey}</code>
+          </Callout>
+          <div className="mt-3 px-3">
+            <Link href="/holdouts" size="3">
+              All Holdouts
+            </Link>
+          </div>
+        </div>
+      );
+    }
```
msamper: The API throws an error if no holdout was found so not sure if we need this

packages/front-end/pages/holdouts/lookup.tsx
gazzdingo: sorry that was from testing lol

packages/front-end/pages/holdouts/lookup.tsx
gazzdingo: testing again

PR #4493, opened 2025-08-28
chore(sql): Add testing for experiment fact metric SQL

lukesonnet:
### Features and Changes

Adds testing for over 2,000 different possible fact metrics. If SQL changes unexpectedly for certain metric configs, this should alert us to it. If changes are intentional, then the tests will provide a way to diff the SQL to help verify.

You can then run `yarn test -u` to update the stored SQL if the changes are expected and correct.


ahdriel (approved):
🚀

PR #4494, opened 2025-08-29
Fix holdout remaining as none when switching projects

msamper:
### Features and Changes

When switching from a project with no holdouts to a project with holdouts, the selected holdout was remaining as none when it should be set to the first valid holdout in the array.


packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -325,7 +325,7 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
     },
   });
 
-  const [selectedProject, setSelectedProject] = useState(form.watch("project"));
+  const selectedProject = form.watch("project");
```
msamper, line 328: There was an unnecessary `useState` here

packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -66,22 +66,39 @@ export const HoldoutSelect = ({
     });
   }, [holdouts, experimentsMap, selectedProject, getDatasourceById]);
 
+  const [userSelectedNone, setUserSelectedNone] = useState(false);
```
jdorn: Don't use state for this. Just derive this from the form value.

packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -66,22 +66,39 @@ export const HoldoutSelect = ({
...
+  const shouldReCheckHoldout = useRef(true);
```
jdorn: A ref and useEffect is a weird way to solve this problem and super twisty. I recommend taking a step back and doing this differently

gazzdingo (approved):

PR #4498, opened 2025-08-29
initial holdout docs

royalfig:
Adds Holdout docs:

- Doc on running holdouts in GrowthBook
- Transforms existing KB article into deeper dive on holdouts theory/explanation

[] Fill out best practices section 
[] Add additional theory to KB article if needed


docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+Holdout experiments (or simply "holdouts") measure the long-term impact of features by maintaining a control group that doesn't receive new functionality. While most users experience your latest features and improvements, a small percentage remains on the original version, providing a baseline to measure cumulative effects over time.
+
+Learn more about holdouts in our [Knowledge Base](/kb/experiments/holdouts).
+
+## How Holdouts Work in GrowthBook
+
+In GrowthBook, users in the holdout treatment group receive the same experience as the general population: they participate in experiments, receive feature releases, and see all your latest changes. The holdout control group, however, stays on the baseline experience.
+
+This approach measures the cumulative impact of all changes over time by comparing the general population to the holdout control group.
```
msamper: might be worth explaining this section
Screenshot 2025-08-29 at 2 08 33 PM
We can maybe explain that the holdout size they select will also be used to retain an equivalent amount of users from the treatment group in order to measure cumulative impact

msamper: We may also want to not use the word `users` since it's not always accurate. I think @lukesonnet prefers to use the word `units` to be generic, but I don't know if that's widely understood

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+![Add Holdout](/images/holdouts-add-modal.png)
+
+1. Navigate to **Experiments** → **Holdouts** and click **Add Holdout**
+
+2. Configure your holdout settings:
+
+- **Name**: Choose a descriptive name (e.g., `q1-2025-product-holdout`)
+- **Holdout Size**: Defaults to 5% (meaning 5% of users remain in the control group)
+- **Description**: Document the purpose and scope of your holdout
+- **Key Metric**: Select a metric to track the impact of the holdout
```
msamper: Should this be `Goal Metrics`? Could also be more than one metric

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+## Creating a Holdout
+
+![Add Holdout](/images/holdouts-add-modal.png)
+
+1. Navigate to **Experiments** → **Holdouts** and click **Add Holdout**
+
+2. Configure your holdout settings:
+
+- **Name**: Choose a descriptive name (e.g., `q1-2025-product-holdout`)
+- **Holdout Size**: Defaults to 5% (meaning 5% of users remain in the control group)
```
msamper: Can explain what I mentioned above here too of an equivalent % being used for measurement on the treatment (general population) group

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+- **Description**: Document the purpose and scope of your holdout
+- **Key Metric**: Select a metric to track the impact of the holdout
+
+3. Click **Start Holdout** to begin
+
+### Adding Experiments to a Holdout
```
msamper, line 39: Maybe worth having a section for adding Features to a holdout?

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+![Add Holdout to an experiment](/images/holdouts-experiment-with-holdout.png)
+
+Once created, you can use your holdout in new experiments:
+
+1. When creating a new experiment, you'll see a **Holdout** field
+2. Your active holdout will selected in the dropdown by default
+3. Complete the rest of your experiment setup normally
```
msamper, line 47: Should we let them know that features added to the experiment will automatically inherit the holdout?

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+This approach measures the cumulative impact of all changes over time by comparing the general population to the holdout control group.
+
+## Creating a Holdout
+
+![Add Holdout](/images/holdouts-add-modal.png)
+
+1. Navigate to **Experiments** → **Holdouts** and click **Add Holdout**
+
+2. Configure your holdout settings:
+
```
msamper, line 25: I think another important holdout specific setting is `Enabled Environments`. These are the environments for which we will automatically add a Holdout Rule in any linked features while the holdout is running. Typically this should just be `production` unless they're maybe doing some sort of debugging. Maybe that's another best practice to add

docs/docs/running-experiments/running-holdouts.mdx
msamper: Same is true when creating an experiment from a feature that's linked to a holdout

docs/docs/kb/experiments/holdouts.mdx
```diff
@@ -1,106 +1,28 @@
...
 
-Why would I want to run a holdout?
+## Why run a holdout?
 
-- Holdouts are a great way to measure long-term impact. The impact of features changes over time, as does user behavior, and running an experiment for an extended period can help you understand these effects.
-- Holdouts can help you measure the impact of multiple features at once. Experiments can interact with each other in unexpected ways and these interactions can change as time goes on.
+- **Measure long-term impact**: The impact of features changes over time, as does user behavior, and running a holdout experiment for an extended period will help you understand these effects.
+- **Understand cumulative effects**: Features interact in unexpected ways, and these interactions shift over time. Holdouts capture the combined impact of multiple changes, revealing synergies and conflicts that individual experiments miss..
```
lukesonnet:
```suggestion
- **Understand cumulative effects**: Features interact in unexpected ways, and these interactions shift over time. Holdouts capture the combined impact of multiple changes, revealing synergies and conflicts that individual experiments miss.
```

lukesonnet: As a longer explanation, it might be worth going into a pitch about why alternatives are not as good, with something like so:

Some approaches to measuring cumulative effects involve summing up the effect of individual experiments. Often, analysts then shrink this total by some amount to account for statistical bias (like we do on our [impact dashboard](/insights#scaled-impact)) and to account for the general understanding that long term impacts often shrink or cannibalize one another. But even then, often these adjusted totals still overstate or mis-state the total impact (for example, see [this analysis by AirBnB](https://medium.com/airbnb-engineering/selection-bias-in-online-experimentation-c3d67795cceb)).

We could then add, if we wanted, or we could cut this altogether and save this for somewhere else:

Holdouts expose users to a combined set of features over a long period of time, naturally measuring their interactions (positive and negative) and serving as a back-test that deals with the selection bias issue in summing up shipped experiments. For this reason, they are the gold standard for measuring cumulative, long-run impact.

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+slug: /app/holdouts
+---
+
+Holdout experiments (or simply "holdouts") measure the long-term impact of features by maintaining a control group that doesn't receive new functionality. While most users experience your latest features and improvements, a small percentage remains on the original version, providing a baseline to measure cumulative effects over time.
+
+Learn more about holdouts in our [Knowledge Base](/kb/experiments/holdouts).
+
+## How Holdouts Work in GrowthBook
+
+In GrowthBook, users in the holdout treatment group receive the same experience as the general population: they participate in experiments, receive feature releases, and see all your latest changes. The holdout control group, however, stays on the baseline experience.
```
lukesonnet: I would lead with the holdout control group as this is what most people would expect and clarifying the bit about measurement later, and then matching the language in the UI.

```suggestion
In GrowthBook, units (e.g. users) in the holdout group are withheld from new features, experiments, and bandits. Units not in the holdout group will: participate in experiments, receive feature releases, and see all your latest changes. A subset of these units not in the holdout are used for measurement; these units are the same as the rest of the general population, we just select a sub-sample of units to serve as a comparison group for your holdout population.
```

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet: I've added a suggestion that addresses the first issue a bit. As for `units` vs. `users`. I wouldn't mind us using `Experimental units (e.g. users)` at first and then using `units` throughout, if you think that isn't too jarring to you, @royalfig .

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet:
```suggestion
- **Holdout Size**: Defaults to 5% (meaning 5% of traffic is placed in the holdout)
```

I mention "traffic" because it is still using our hashing on the fly and I wanted to make it clear it doesn't like place "5%" in a holdout and then just puts every one else in a new bin.

---

Both for this line and the one about metrics below I think it might make sense to have some additional color, IDK where you would want to put that. This is a bit of list of "what to do" but we also might want a section about "why" or "what happens". IDK if you want to do that inline or link it elsewhere, @royalfig . Maybe it goes in best practices, but it kind of links to each one of the settings.

 For this one:

"We also will use this % as the amount of traffic in the general population to use for measurement.

For example, if you pick 5% for your holdout group, then you will have 5% in the holdout and 95% not in the holdout, but that 95% is split into 5% that are used for measurement and 90% that are not. While using the full 95% for measurement would increase the statistical power of your holdout, it does not represent good statistical value to do so. The statistical power of a test is mostly limited by the size of the smaller group, which in this case is the holdout group, and adding a ton of additional tracking events and ballooning the holdout sample for this (often) long running experiment will cause more events to fire and queries to run longer, often without a meaningful reduction in uncertainty in your test. Please reach out if you wish to implement a different size measurement population."

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet:
```suggestion
This approach measures the cumulative impact of all changes over time by comparing the general population to the holdout group.
```

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+This approach measures the cumulative impact of all changes over time by comparing the general population to the holdout control group.
+
```
lukesonnet, line 17:
```suggestion
Under the hood, holdouts work by using prerequisites to divert a selected percentage of traffic away from features and experiments towards default values selected for your feature flags.
```

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet: Yes, and I think there should be a default as to what happens next, in line with the projects above.

Suggestion:

```suggestion
## Adding to a Holdout

Once your holdout is created, it is selected by default in new feature and experiment creation flows for the selected projects. Whenever someone creates a new feature, experiment, or bandit in a project the holdout was assigned to, a new field for "Holdout" is added to the creation flow and it defaults to the created holdout. If there are multiple holdouts assigned to an project (not recommended), then the feature or experiment creator will have to select the correct holdout to use, as only one can be used at a time per feature or experiment.


### Adding Experiments to a Holdout
```

Then add

```
### Adding Features to a Holdout

New features created for an experiment will automatically be added to that experiment's holdout. However, you may want to add other features that do not go through an experiment to a holdout to ensure that the all changes are included in the holdout, even those you did not experiment on.

... HOW TO ...
```

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet: I kind of mention that in the "features" section in my comment above. 

We could also add here:

"Experiments created from a feature that has a holdout applied will automatically inherit the holdout from that feature."

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+3. Click **Start Holdout** to begin
+
+### Adding Experiments to a Holdout
+
+![Add Holdout to an experiment](/images/holdouts-experiment-with-holdout.png)
+
+Once created, you can use your holdout in new experiments:
+
+1. When creating a new experiment, you'll see a **Holdout** field
+2. Your active holdout will selected in the dropdown by default
```
lukesonnet:
```suggestion
2. Your active holdout for the given project will selected in the dropdown by default
```

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+3. Complete the rest of your experiment setup normally
+
+The experiment will run as usual, with the holdout tracking long-term cumulative effects in the background.
+
+## Holdout Lifecycle
+
+### Active Period
+
+During the active period (typically a quarter), you'll:
```
lukesonnet:
```suggestion
During the active period (often a quarter), you'll:
```

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet: Yes, I think so.
```suggestion
- **Goal Metrics**: Select the key metrics for measuring cumulative impact
```

Best practices, but in this case it's important since we might be filtering out a lot of their key metrics:

"We prevent adding metrics with conversion windows to a holdout. This is because traffic is exposed to the holdout whenever they hit an included feature or experiment. This can often be days or weeks apart from their exposure to a later feature or experiment, so conversion windows based around their first exposure event no longer make sense. Furthermore, the goal of a holdout is to measure long-range effects; so we allow you to add metrics without conversion windows or with lookback windows."

docs/docs/running-experiments/running-holdouts.mdx
lukesonnet: We also definitely need to add `Projects`.

Projects: select the projects in which the holdout should be active. The holdout will be added by default to new features, experiments, and bandits in this project and will only be available to add to these projects.


" and will only be available to add to this project." <-- is this true @msamper ?

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+- Continue adding new experiments to the holdout
+- Launch features as normal
+- Let the holdout accumulate data on cumulative impact
+
+### Analysis Period
+
```
lukesonnet, line 64: This notion will not be totally obvious to people, so I suggest some set up:

Throughout the active holdout period, you can use the holdout to measure cumulative impact on your metrics as new features and experiments are launched. However, it is often best to measure holdout impact for some set period of time *after* all new changes are frozen in place and the general population has had a chance to experience them together. For this reason, we make it easy to measure impacts over a set analysis period.

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+After your planned duration, transition to the analysis period:
+
+1. **Stop adding new experiments**: The holdout enters analysis mode where no new experiments can be added
```
lukesonnet:
```suggestion
1. **Stop adding new experiments**: The holdout enters analysis mode where no new experiments or features can be added
```

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+2. **Extended measurement**: Continue measuring for 2-4 weeks to capture delayed effects and reduce statistical noise
+3. **Dynamic lookback window**: GrowthBook automatically applies a lookback window based on when each experiment was added, ensuring clean measurement periods
+
+During analysis, the system:
+
+Retains existing user assignments to avoid rebucketing
```
lukesonnet:
```suggestion
Continues to split traffic into your holdout and general population groups
```

Much clearer, the other looks like it uses sticky bucketing or some kind of data store.

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+Applies the appropriate lookback window to exclude results from before the analysis period
+Uses the minimum of the metric's lookback window and the analysis lookback window for each metric
```
lukesonnet:
```suggestion
Uses the minimum of the metric's lookback window (if it exists) and the analysis lookback window for each metric
```

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+### Understanding Holdout Results
+
+Holdout results show the cumulative impact of all features and experiments included during the holdout period:
+
+- **Positive lift**: Your combined changes improved metrics beyond the baseline
+- **Negative or neutral lift**: Consider whether individual wins masked broader negative effects
+- **Metric-specific impacts**: Different metrics may show varying long-term effects
+
+The analysis compares your general population (who received all features) against the holdout control group (who remained on the baseline), revealing the true long-term value of your product changes.
```
lukesonnet:
```suggestion
The analysis compares your general population (who received all features and were experimented on) against the holdout control group (who remained on the baseline), revealing the true long-term value of your product changes.
```

docs/docs/running-experiments/running-holdouts.mdx
msamper: Yup! that change was made. If a holdout is in project x, it will only show up if an experiment or features selected project is x

docs/docs/running-experiments/running-holdouts.mdx
```diff
@@ -0,0 +1,85 @@
...
+- **Goal Metrics**: Select the key metrics for measuring cumulative impact. Note: We prevent adding metrics with conversion windows to a holdout. This is because traffic is exposed to the holdout whenever they hit an included feature or experiment. This can often be days or weeks apart from their exposure to a later feature or experiment, so conversion windows based around their first exposure event no longer make sense. Furthermore, the goal of a holdout is to measure long-range effects; so we allow you to add metrics without conversion windows or with lookback windows.
+
+3. Click **Start Holdout** to begin
+
+## Adding to a Holdout
+
+Once your holdout is created, it is selected by default in new feature and experiment creation flows for the selected projects. Whenever someone creates a new feature, experiment, or bandit in a project the holdout was assigned to, a new field for "Holdout" is added to the creation flow and it defaults to the created holdout. If there are multiple holdouts assigned to an project (not recommended), then the feature or experiment creator will have to select the correct holdout to use, as only one can be used at a time per feature or experiment.
```
lukesonnet:
```suggestion
Once your holdout is created, it is selected by default in new feature and experiment creation flows for the selected projects. Whenever someone creates a new feature, experiment, or bandit in a project the holdout was assigned to, a new field for "Holdout" is added to the creation flow and it defaults to the created holdout. If there are multiple holdouts assigned to a project (not recommended), then the feature or experiment creator will have to select the correct holdout to use, as only one can be used at a time per feature or experiment.
```

msamper (approved):

PR #4503, opened 2025-09-02
Add project-scoping to prelaunch checklist tasks

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/back-end/src/controllers/experimentLaunchChecklist.ts
+++ b/packages/back-end/src/controllers/experimentLaunchChecklist.ts
@@ -63,10 +63,15 @@ export async function postExperimentLaunchChecklist(
 }
 
 export async function getExperimentCheckListByOrg(
-  req: AuthRequest,
+  req: AuthRequest<null, { num: string }, { projectId?: string }>,
   res: Response,
 ) {
   const { org } = getContextFromReq(req);
+
+  let projectId = "";
+  if (typeof req.query?.projectId === "string") {
+    projectId = req.query.projectId;
+  }
 
   if (!orgHasPremiumFeature(org, "custom-launch-checklist")) {
     return res.status(200).json({
@@ -77,13 +82,24 @@ export async function getExperimentCheckListByOrg(
 
   const checklist = await getExperimentLaunchChecklist(org.id, "");
 
+  const filteredTasks: ChecklistTask[] = [];
+
+  checklist?.tasks.forEach((task) => {
+    if (!task.projects?.length) {
+      filteredTasks.push(task);
+      return;
+    }
+    if (task.projects?.includes(projectId)) {
+      filteredTasks.push(task);
+      return;
+    }
+  });
+
   return res.status(200).json({
     status: 200,
-    checklist,
+    checklist: { ...checklist, tasks: filteredTasks },
   });
 }
-
-//TODO: Add getExperimentCheckListByProject method
 
 export async function putExperimentLaunchChecklist(
   req: AuthRequest<{ tasks: ChecklistTask[] }, { id: string }>,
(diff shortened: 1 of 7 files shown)


packages/back-end/src/app.ts
```diff
@@ -651,7 +651,11 @@ app.put(
 );
 app.get(
   "/experiments/launch-checklist",
-  experimentLaunchChecklistController.getExperimentCheckListByOrg,
+  experimentLaunchChecklistController.getExperimentCheckListForOrg,
+);
+app.get(
+  "/experiment/:id/launch-checklist",
+  experimentLaunchChecklistController.getExperimentCheckListForExperiment,
 );
```
mknowlton89, line 659: Now that a checklist's items can differ from experiment to experiment (depending on what project an experiment is in), I created a new endpoint to fetch the pre-launch checklist for a specific experiment. 

This checklist is separate from the general endpoint to fetch the checklist for the `/settings` page that doesn't take into consideration a particular experiment.

packages/back-end/src/controllers/experimentLaunchChecklist.ts
```diff
@@ -83,7 +84,52 @@ export async function getExperimentCheckListByOrg(
...
+  checklist?.tasks.forEach((task) => {
+    if (!task.projects?.length) {
+      filteredTasks.push(task);
+      return;
+    }
+    if (task.projects.includes(projectId)) {
+      filteredTasks.push(task);
+      return;
+    }
+  });
```
Kevin-Chant, line 126: Nit: this can be simplified
```suggestion
  const filteredTasks = (checklist?.tasks ?? []).filter((task) => !task.projects?.length || task.projects.includes(projectId));
```

packages/back-end/src/models/ExperimentLaunchChecklistModel.ts
```diff
@@ -12,11 +12,13 @@ const experimentLaunchChecklistSchema = new mongoose.Schema({
   dateCreated: Date,
   dateUpdated: Date,
   updatedByUserId: String,
+  // Need to deprecate projectId at this level if we add support for project-level checklist items
```
Kevin-Chant, line 15: Is it the case that this field isn't in use and has never been used? If so, I would lean towards deleting it with this change to prevent it causing conflicts in the future

packages/back-end/types/experimentLaunchChecklist.d.ts
```diff
@@ -3,6 +3,7 @@ export interface ChecklistTask {
   completionType: "manual" | "auto";
   url?: string;
   customFieldId?: string;
+  projects: string[];
```
Kevin-Chant: This field needs to be marked as optional to enforce type safety (or add a JIT migration)
```suggestion
  projects?: string[];
```

Kevin-Chant (changes requested):
Only one substantive change, code mostly looks good. I'm also not strongly opinionated on whether this is the right approach - this does seem odd given our existing patterns but on the other hand I like the simplicity. Might be worth waiting for other opinions on that front.

packages/back-end/src/models/ExperimentLaunchChecklistModel.ts
mknowlton89: @Kevin-Chant Yea... that's the case - we added it to the original model as we thought we'd eventually project scope it. Though, now if we go the route of project-scoping the tasks, rather than the checklist, we should remove this entirely.

packages/back-end/types/experimentLaunchChecklist.d.ts
mknowlton89: Ah, good catch. TY

mknowlton89:
Closing this PR - after doing some thinking, I think this could limit us in the future, as opposed to supporting full checklists at the project level. I think the biggest limitation down the road would be from a permissions standpoint - currently, in order to manage an org's checklist, a user would need the admin permission globally. And if we went the route of having a single org-level list, the permissions could get hairy, especially if a user had the permission to manage a specific project, but wasn't an admin globally, they'd need to get an admin to update their project's checklist.

Likewise, going the alternate route will allow a user to have manage project permissions for a subset of projects, and allow them to adjust the checklist for those projects, but none that they don't have permission for.

PR #4504, opened 2025-09-02
fix(power): ensure threshold passes through to power calc

lukesonnet:
### Features and Changes

We previously weren't properly passing through the bayesian and frequentist org threshold settings if they were different from the default.

Also added information to the stats engine settings section and provided you with a UI to override, thereby partly addressing #4505.

### Testing

Changed organization thresholds and observed directionally correct updates to power calculation.

Screenshot 2025-09-02 at 10 54 20 AM

Screenshot 2025-09-02 at 10 54 03 AM

Also seem loom for evidence the override is working: 
https://www.loom.com/share/d8963d9c18a54f26a166af3a90eb6031

Ignore how similar the %ages are to your threshold in that loom, this is correct with these thresholds and I tested it again with more reasonable data and all was well.

### DIFF
--- a/packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
+++ b/packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
@@ -12,6 +12,10 @@ import usePValueThreshold from "@/hooks/usePValueThreshold";
 
 type StatsEngineWithSequential = "bayesian" | "frequentist" | "sequential";
 
+export type StatsEngineSettingsWithAlpha = StatsEngineSettings & {
+  alpha: number;
+};
+
 export function alphaToChanceToWin(alpha: number): number {
   return parseFloat((100 * (1 - alpha)).toFixed(6));
 }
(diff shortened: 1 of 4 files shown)


lukebrawleysmith:
I tested frequentist (sequential and unadjusted) and Bayesian cases and see that organizational settings are now being propagated through.  LGTM

packages/front-end/components/PowerCalculation/PowerCalculationStatsEngineSettingsModal.tsx
```diff
@@ -1,16 +1,33 @@
...
+import useConfidenceLevels from "@/hooks/useConfidenceLevels";
+import usePValueThreshold from "@/hooks/usePValueThreshold";
 
 type StatsEngineWithSequential = "bayesian" | "frequentist" | "sequential";
 
+export type StatsEngineSettingsWithAlpha = StatsEngineSettings & {
+  alpha: number;
+};
+
+export function alphaToChanceToWin(alpha: number): number {
```
lukebrawleysmith, line 19: For experiments `alpha` is specified for two-sided test.  CTW is directional.  I think the solution here is `100 * (1 - 0.5 * alpha)`.

lukebrawleysmith (commented):
one comment around 1-sided vs 2-sided test, otherwise LGTM

PR #4511, opened 2025-09-03
Custom validation hooks

jdorn:
### Features and Changes

For self-hosted deployments, allow custom javascript validation snippets for features, experiments, and more.  User-supplied code is run inside a V8 isolate for security with strict CPU and memory limits.

Example: Require a non-empty description for features:

```js
if (!feature.description) throw new Error("A description is required")
```

Example: Call an external API that scans for PII in the default feature value and rejects if found

```js
// Boolean features are always fine
if (feature.valueType === "boolean") return;

const res = await fetch("https://example.com/api/v1/detect-pii", {
  method: "POST",
  body: JSON.stringify({ text: feature.defaultValue })
});
const json = await res.json();

if (json.hasPII) throw new Error("PII detected in default value");
```

Limits per invocation:
- Max memory: 32MB
- Max CPU active time: 100ms
- Max overall execution time (including fetches): 5s
- Max fetch response size: 500KB

image

image

TODO:
- [x] `runCustomValidation` method that uses V8 isolates and enforces limits
- [x] New data model to store custom validation snippets
- [x] CRUD endpoints for custom validation snippets (self-hosted only)
- [x] Front-end to manage custom validation snippets (self-hosted only)
- [x] `validateFeature` hook
   - [x] Create feature (app)
   - [x] Create feature (rest)
   - [x] Update feature (app)
   - [x] Update feature (rest)
- [x] `validateFeatureDraft` hook
   - [x] Create revision (app)
   - [x] Create revision (rest)
   - [x] Update revision (app)
   - [x] Update revision (rest)
- [ ] `validateExperiment` hook
   - [ ] Create experiment (app)
   - [ ] Create experiment (rest)
   - [ ] Update experiment (app)
   - [ ] Update experiment (rest)
   - [ ] Start experiment (app)
   - [ ] Make changes (app)

### DIFF
--- a/packages/back-end/package.json
+++ b/packages/back-end/package.json
@@ -16,7 +16,7 @@
     "start:with-tracing": "node --require ./dist/tracing.opentelemetry.js dist/server.js",
     "start:with-datadog": "node --require ./dist/tracing.datadog.js dist/server.js",
     "repl": "node --require ts-node/register/transpile-only --experimental-repl-await repl",
-    "test": "DISABLE_PYTHON_MONITOR=true GB_STATS_ENGINE_MIN_POOL_SIZE=0 jest --forceExit --verbose --detectOpenHandles",
+    "test": "DISABLE_PYTHON_MONITOR=true GB_STATS_ENGINE_MIN_POOL_SIZE=0 NODE_OPTIONS='--no-node-snapshot' jest --forceExit --verbose --detectOpenHandles",
     "type-check": "tsc --pretty --noEmit",
     "unused-export-check": "npx ts-unused-exports tsconfig.json --findCompletelyUnusedFiles --allowUnusedTypes --showLineNumber",
     "generate-dummy-data": "ts-node ./test/data-generator/new-generator.ts",


packages/back-end/src/enterprise/sandbox/sandbox-eval.ts
```diff
@@ -0,0 +1,259 @@
...
+} from "back-end/src/routers/custom-hooks/custom-hooks.validators";
+import { FeatureInterface } from "back-end/types/feature";
+import { ReqContextClass } from "back-end/src/services/context";
+import { getContextForAgendaJobByOrgObject } from "back-end/src/services/organizations";
+import { FeatureRevisionInterface } from "back-end/types/feature-revision";
+
+const MEMORY_MB = 32;
+const CPU_TIMEOUT_MS = 100; // max active CPU time (excluding fetch)
+const WALL_TIMEOUT_MS = 5000; // max total run time (including fetch)
+const MAX_FETCH_RESP_SIZE = 500 * 1024; // 500KB max response size from fetch calls
```
tzjames: Maybe set these to be env vars in case a self hosted instance wants to change it for whatever reason.

packages/back-end/package.json
```diff
@@ -74,6 +74,7 @@
     "generate-password": "^1.7.1",
     "googleapis": "^59.0.0",
     "handlebars": "^4.7.8",
+    "isolated-vm": "5.0.4",
```
tzjames, line 78: All calls to `node` in this file where Isolate might run, ie. start, start-with-tracing, watch-dev, etc... need `--no-node-snapshot` to run correctly on macOS, node>20.

packages/front-end/pages/settings/custom-hooks.tsx
```diff
@@ -0,0 +1,430 @@
...
+            />
+          ))}
+          <Button
+            onClick={runTest}
+            disabled={!form.watch("code")}
+            variant="outline"
+          >
+            <Flex align="center" gap="3">
+              <Text>Run Test</Text>
+              <Kbd size="1">Ctrl + Enter</Kbd>
```
tzjames: On Mac this is `Command + Enter`

packages/back-end/src/enterprise/sandbox/sandbox-eval.ts
```diff
@@ -0,0 +1,259 @@
+// sandbox-runner.ts
```
tzjames: Unneeded comment.  Also wrong file name,

packages/back-end/src/enterprise/sandbox/sandbox-eval.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  const hooks = await adminContext.models.customHooks.getByHook(
+    hookType,
+    project,
+  );
+  for (const hook of hooks) {
+    const res = await _runCustomHook(adminContext, hook, functionArgs);
+    if (!res.ok) {
+      const message =
+        (res.error || "Custom hook error") + (res.log ? `\n${res.log}` : "");
```
tzjames, line 98: When creating the custom hook and testing, it seems useful to have the logs.  But this also prints out when creating an actual feature, which I wasn't expecting.

packages/front-end/pages/settings/custom-hooks.tsx
```diff
@@ -0,0 +1,430 @@
...
+            method: "POST",
+            body: JSON.stringify(value),
+          });
+        }
+
+        if (onSave) onSave();
+      })}
+    >
+      <Flex align="start" gap="5">
+        <div style={{ width: "50%" }} className="border-right pr-4">
```
tzjames, line 208: When the right column is too long (feature+revision+return val+ some logs), the divider wasn't going all the way down.  With the following change it was:
```suggestion
      <Flex align="stretch" gap="5" style={{ alignItems: "stretch" }}>
        <div
          style={{ width: "50%", display: "flex", flexDirection: "column" }}
          className="border-right pr-4"
        >
```

packages/back-end/src/enterprise/sandbox/sandbox-eval.ts
```diff
@@ -0,0 +1,259 @@
...
+          const { _body, _error, ...rest } = await hostFetch.applyPromise(undefined, args);
+          if (_error) {
+            throw new Error(_error);
+          }
+          return {
+            ...rest,
+            text: async () => _body,
+            json: async () => JSON.parse(_body),
+          };
+        };
```
tzjames, line 224: This doesn't work.  It isn't the right syntax, plus functions can't be returned.  So we can't have realistic fetch behavior.  The best we can do is:
```
globalThis.fetch = async (...args) => {
          const { _body, _error, ...rest} = await hostFetch.apply(undefined, args, {
            arguments: { copy: true },
            result: { copy: true, promise: true }
          });
          if (_error) {
            throw new Error(_error);
          }
          let json;
          try {
               json = JSON.parse(_body);
          } catch {
               json = null;
          }
          return {
            ...rest,
            text: _body,
            json
          };
        };
```
but then it wouldn't be functioning like normal fetch as .text and .json wouldn't be functions and would have to be accessed directly.
Or we can do:
```
  globalThis.responseText = (response) => response._bodyText;
        globalThis.responseJson = (response) => JSON.parse(response._bodyText);
        
        globalThis.fetch = async (...args) => {
          const { _body, _error, ...rest} = await hostFetch.apply(undefined, args, {
            arguments: { copy: true },
            result: { copy: true, promise: true }
          });
          if (_error) {
            throw new Error(_error);
          }
          return {
            ...rest,
            _bodyText: _body
          };
        };
```
then they would have to do:
```
let data = await fetch("https://httpbin.org/uuid");
console.log(data);
console.log(responseText(data)); 
console.log(responseJson(data).uuid); 
```

packages/back-end/package.json
jdorn: I only changed it for `watch-dev` since I figured it's pretty rare to run `start` outside of a Docker container. I might need to also add it for tests.

packages/back-end/src/enterprise/sandbox/sandbox-eval.ts
jdorn: I was able to get it to work

packages/back-end/package.json
tzjames: Ok.  I kind of think it is such a minor performance difference and only on node start up that we might as well have it on both.  But I guess we can add it later if people complain.

PR #4512, opened 2025-09-03
Miscellaneous Holdout tweaks

msamper:
### Features and Changes

- Copy tweaks
- Shows all project holdouts for an experiment with a selected project
- Shows `lookbackStartDate` instead of phase `startDate` for analysis period
- Tweaks color of skipped holdout rule

### Screenshots

Screenshot 2025-09-03 at 10 06 09 AM


packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -24,7 +24,7 @@ export const HoldoutSelect = ({
 }) => {
   const { getDatasourceById } = useDefinitions();
   const { hasCommercialFeature } = useUser();
-  const { holdouts, experimentsMap } = useHoldouts(selectedProject);
+  const { holdouts, experimentsMap } = useHoldouts();
```
lukesonnet, line 27: When do we use this project prop for useHoldouts then? I guess we filter below so this looks fine and tracks with my testing.

msamper: it's used on the holdouts list page to only show holdouts belonging to a given project

PR #4514, opened 2025-09-03
[feat] Dismissible Callouts

msamper:
### Features and Changes

Allows Callouts to be dismissible, similar to PremiumCallouts. Makes Safe Rollout promo and Holdout promo dismissible.

### Screenshots

Screenshot 2025-09-03 at 10 47 06 AM

Screenshot 2025-09-03 at 10 53 50 AM

### DIFF
--- a/packages/front-end/components/Radix/Callout.tsx
+++ b/packages/front-end/components/Radix/Callout.tsx
@@ -45,7 +45,7 @@ export default forwardRef<
     size = "md",
     icon,
     contentsAs = "text",
-    dismissible,
+    dismissible = false,
     id,
     renderWhenDismissed,
     ...containerProps


packages/front-end/components/Radix/Callout.tsx
```diff
@@ -22,6 +24,9 @@ export default forwardRef<
     size?: "sm" | "md";
     icon?: ReactNode | null;
     contentsAs?: "text" | "div";
+    dismissible?: boolean;
+    id?: string;
+    renderWhenDismissed?: (undismiss: () => void) => React.ReactElement;
```
ahdriel: suggestion: We can type it like this to make TypeScript complain if we don't provide an `id` when `dismissible` is true, or to complain when `renderWhenDismissed` is provided but `dismissible` is not true.

```
type DismissibleProps = {
  dismissible: true;
  id: string;
  renderWhenDismissed: (undismiss: () => void) => ReactNode;
};

type UndismissableProps = {
  dismissible?: false;
  id?: string;
  renderWhenDismissed?: never;
};

type CalloutProps = {
  children: ReactNode;
  status: Status;
  size?: "sm" | "md";
  icon?: ReactNode | null;
  contentsAs?: "text" | "div";
} & (DismissibleProps | UndismissableProps) &
  MarginProps;

export default forwardRef<HTMLDivElement, CalloutProps>(function Callout(
```

msamper: TIL, great suggestion!

PR #4516, opened 2025-09-03
chore(clickhouse): Update property that was deprecated

ahdriel:
### Features and Changes

While looking over some server logs I noticed the following line:

```
[2025-09-03T22:50:57.042Z][WARN][@clickhouse/client][Config] "host" is deprecated. Use "url" instead.
```

So I am doing that.


msamper (approved):
lgtm! checked clickhouse client docs and this looks to be in line with the parameters they expect now

PR #4519, opened 2025-09-04
Investigate increased CPU usage

ahdriel:
### Features and Changes


ahdriel:
See #4541 for the fix

PR #4522, opened 2025-09-04
Add baseSetting fallback to metric value resolution

itsgrimetime:
When the settings API is called, the metric settings fields `delayHours`, `windowType`, `windowHours`, `winRisk`, `loseRisk` are being resolved by `genMetricOverrideResolver`, which:

Checks experiment-level metric overrides -> checks metric-level values -> falls back to `null` if neither exist

This should probably fall back to the organization's base settings (from `ctx.baseSettings`) like the `genDefaultResolver` does, not to `null`.

The OpenAPI spec shows these fields as required - but nullable, leading to issues when attempting to validate responses via something like pydantic

### DIFF
--- a/packages/shared/src/settings/resolvers/genMetricOverrideResolver.ts
+++ b/packages/shared/src/settings/resolvers/genMetricOverrideResolver.ts
@@ -13,6 +13,7 @@ export default function genMetricOverrideResolver(
   fieldName: keyof Omit<
     MetricOverride,
     | "id"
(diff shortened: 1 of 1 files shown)


ahdriel:
@itsgrimetime Nice find! Thanks for the fix.

Can you expand on the OpenAPI point to make sure I understand it? You are saying that the current spec is wrong even after the fix you provided in this PR? Would love to fix this for you as well if that's the case.

ahdriel (approved):

PR #4534, opened 2025-09-08
Allow sso enabled orgs to customize defaultRole to include env limits…

mknowlton89:
… and project-scoped roles

### Features and Changes

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -2053,7 +2053,7 @@ export async function putLicenseKey(
 }
 
 export async function putDefaultRole(
-  req: AuthRequest<{ defaultRole: string }>,
+  req: AuthRequest<{ defaultRole: MemberRoleWithProjects }>,
   res: Response,
 ) {
   const context = getContextFromReq(req);
@@ -2068,7 +2068,10 @@ export async function putDefaultRole(
     );
   }
 
-  if (!isRoleValid(defaultRole, org)) {
+  if (
+    !isRoleValid(defaultRole.role, org) ||
+    !areProjectRolesValid(defaultRole.projectRoles, org)
+  ) {
     throw new Error("Invalid role");
   }
 
@@ -2079,11 +2082,7 @@ export async function putDefaultRole(
   updateOrganization(org.id, {
     settings: {
       ...org.settings,
-      defaultRole: {
-        role: defaultRole,
-        limitAccessByEnvironment: false,
-        environments: [],
-      },
+      defaultRole,
     },
   });
 
--- a/packages/front-end/components/Settings/Team/UpdateDefaultRoleForm.tsx
+++ b/packages/front-end/components/Settings/Team/UpdateDefaultRoleForm.tsx
@@ -2,11 +2,13 @@ import { useState } from "react";
 import { useForm } from "react-hook-form";
 import { RESERVED_ROLE_IDS, getDefaultRole } from "shared/permissions";
 import Button from "@/components/Button";
-import SelectField, { GroupedValue } from "@/components/Forms/SelectField";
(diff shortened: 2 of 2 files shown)


packages/front-end/components/Settings/Team/ProjectRolesSelector.tsx
```diff
@@ -98,12 +97,13 @@ export default function ProjectRolesSelector({
...
-                    ...defaultRole,
-                  }) as ProjectMemberRole,
-                );
+                newProjectRoles.push({
+                  project: newProject,
+                  role: defaultRole.role,
+                  limitAccessByEnvironment:
+                    defaultRole.limitAccessByEnvironment,
+                  environments: defaultRole.environments,
+                });
```
mknowlton89: This change was necessary because now that the defaultRole can have projectRoles, if we simply spread it in, it means the `projectRole` then has a `projectRoles` array inside of it, so the nesting is too deep.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -2079,11 +2082,7 @@ export async function putDefaultRole(
   updateOrganization(org.id, {
     settings: {
       ...org.settings,
-      defaultRole: {
-        role: defaultRole,
-        limitAccessByEnvironment: false,
-        environments: [],
-      },
+      defaultRole,
```
jdorn, line 2105: We're not doing full validation on what is being passed into this endpoint.  `isRoleValid` and `areProjectRolesValid` only do surface level checks.

For example, we need `environments` to be an array of strings, but someone could pass in a different data type and we would just save it here without validation.  That has the potential to break the rest of the app.

mknowlton89: I added some zod validation middleware to this endpoint to ensure the shape of the data matches.

I've also switched from the `isRoleValid` and `isProjectRoleValid` and started using `validateRoleAndEnvs` as it's a much more exhaustive check.

And finally, I have made a tweak to the `validateRoleAndEnvs` so that we only validate the envs IF `limitAccessByEnvironments` is set to true.

On the front-end, if a user toggles the `limitAccessByEnvironment` toggle off, we don't reset the `environments` array to an empty array, so that if they re-enable it, they don't have to re-add all of the environments.

However, this could cause an issue if an environment was deleted - even if `limitAccessByEnvironments` was disabled, the endpoint would throw an error due to the deleted env. Which is a bad front end experience since when the toggle is disabled, we hide the input field for environments, so the user could get really confused as to what environment the error is referring to.

In addition to that change, within `SingleRoleSelector` we previously only rendered active environments, which again could cause confusion around deleted environments as the endpoint might throw an error saying you can't use an env, but that env wouldn't be rendered in the list of envs because it's not active.

So now, I'm pulling all of the active envs, and then adding any deleted envs (aka, an env that is in the user's environments array, but isn't in the active envs array) and indicating that it's deleted. So this way, a user can remove the deleted env. Previous to this change, an org could get into a spot where they can't make any changes to the user's role due to a deleted env.

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
```diff
@@ -67,7 +67,28 @@ export default function SingleRoleSelector({
     groupedOptions.push({ label: "Custom", options: customOptions });
   }
 
-  const availableEnvs = useEnvironments();
+  const activeEnvs = useEnvironments();
+
+  // Create a list of envs that combines current active envs, and any deleted envs that are still in the value.environments
+  const allEnvs = useMemo(() => {
+    const activeEnvIds = new Set(activeEnvs.map((env) => env.id));
+    const envOptions = [...activeEnvs];
```
jdorn: Nit: type this as `SingleValue[]`.  Then, you can just pass it directly into the select field below and keep all of the logic contained within this memo.

jdorn (approved):

PR #4536, opened 2025-09-08
Generate readable change summaries for webhooks/slack messages for features and experiments

### LINKED ISSUE #3917: [Feature] Slack Webhooks- Change log
### Description of Feature

We have a feedback/feature request for the Slack integration. The notifications themselves have been useful to know about the occurrence of an event/change, but it is difficult to understand what changed. For example, going into an event, you see the new config but you don't see the diff. 

Is there any way to at least show the diff (like how one can see it with green/red on the Audit Log within an experiment). The most user friendly option would be to have the change be messaged in a more interpretable way (like "percentage rollout changed from X to Y" or "changed variant allocation") but I recognize that's more difficult

bryce-fitzsimons:
Closes #3917

- Adds a new `changes` property to `feature.updated` and `experiment.updated` events. Captures the relevant changes in those objects.
- Adds an _even-more-human-readable_ version of changes when using Slack / Discord events formats.
- Adds general user attribution to all feature and experiment webhooks, changes the null state from "unknown user" to "updated automatically" (which is true 99% of the time: bandits, etc)

Example slack messages:
*Changing a variation*
image

*Changing a bunch of targeting stuff*
image

*Changing metrics and settings* (we can't get the actual named metrics / groups without a bunch of lookups!
image

*Adding, modifying, reordering feature rules*
image

*Generic experiment change made due to automatic update*
image

### DIFF
--- a/packages/back-end/package.json
+++ b/packages/back-end/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
   },
   "scripts": {
     "build:emails": "mkdir -p dist/templates && cp -r src/templates/* dist/templates",
@@ -33,7 +33,7 @@
     "@clickhouse/client": "^1.0.1",
     "@databricks/sql": "^1.8.1",
     "@dqbd/tiktoken": "^1.0.21",
-    "@google-cloud/bigquery": "5",
+    "@google-cloud/bigquery": "^8.1.1",
     "@google-cloud/storage": "^5.20.5",
     "@growthbook/growthbook": "^1.6.1",
     "@growthbook/proxy-eval": "^1.0.3",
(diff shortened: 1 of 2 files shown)


packages/back-end/src/events/handlers/webhooks/event-webhooks-utils.ts
```diff
@@ -36,3 +38,691 @@ export const getEventWebHookSignatureForPayload = <T>({
...
+
+  return fieldChanges;
+}
+
+export function getObjectDiff(
+  prev: Record<string, unknown>,
+  curr: Record<string, unknown>,
+  ignoredKeys: string[] = [],
+  nestedObjectConfigs: NestedObjectConfig[] = [],
+  maxDepth: number = 10,
```
jdorn: Nit: Use named parameters here

PR #4538, opened 2025-09-09
Optionally disable api root path or hide sensitive info

### LINKED ISSUE #4073: [Feature] Authenticate the root path
### Description of Feature

The [root path of the backend](https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/app.ts#L183) exposes important information that should not be available to anyone. This path should also be authenticated as other internal paths. 

Is there any reason for the root path not to be authenticated? It exposes api_host, app_origin, and build information that can be used by malicious actors. 

Image

Kevin-Chant:
### Features and Changes

Adds four optional environment variables for self-hosted instances to fully disable or partially omit the information exposed on the `/` root path of the api.

- Closes #4073 

### Testing

Add the following variables to your `packages/back-end/.env.local` and toggle them on individually to observe their effects
```
DISABLE_API_ROOT_PATH=false
API_ROOT_HIDE_API_HOST=false
API_ROOT_HIDE_APP_ORIGIN=false
API_ROOT_HIDE_BUILD_INFO=false
```

#### Without anything disabled

image

#### Fully disabled

image

#### Enabled but hiding all optional fields

image

### DIFF
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -13,8 +13,12 @@ import decisionCriteriaRouter from "back-end/src/enterprise/routers/decision-cri
 import { usingFileConfig } from "./init/config";
 import { AuthRequest } from "./types/AuthRequest";
 import {
+  API_ROOT_HIDE_API_HOST,
+  API_ROOT_HIDE_APP_ORIGIN,
+  API_ROOT_HIDE_BUILD_INFO,
   APP_ORIGIN,
   CORS_ORIGIN_REGEX,
+  DISABLE_API_ROOT_PATH,
   ENVIRONMENT,
   EXPRESS_TRUST_PROXY_OPTS,
   IS_CLOUD,
@@ -197,19 +201,33 @@ app.get("/robots.txt", (_req, res) => {
 
 app.use(compression());
 
-app.get("/", (req, res) => {
-  res.json({
-    name: "GrowthBook API",
-    production: ENVIRONMENT === "production",
-    api_host:
-      process.env.API_HOST ||
-      req.protocol + "://" + req.hostname + ":" + app.get("port"),
-    app_origin: APP_ORIGIN,
-    config_source: usingFileConfig() ? "file" : "db",
-    email_enabled: isEmailEnabled(),
-    build: getBuild(),
+if (!DISABLE_API_ROOT_PATH) {
+  app.get("/", (req, res) => {
+    res.json({
+      name: "GrowthBook API",
+      production: ENVIRONMENT === "production",
+      ...(API_ROOT_HIDE_API_HOST
+        ? {}
+        : {
+            api_host:
+              process.env.API_HOST ||
+              req.protocol + "://" + req.hostname + ":" + app.get("port"),
+          }),
+      ...(API_ROOT_HIDE_APP_ORIGIN
+        ? {}
+        : {
+            app_origin: APP_ORIGIN,
+          }),
+      config_source: usingFileConfig() ? "file" : "db",
+      email_enabled: isEmailEnabled(),
+      ...(API_ROOT_HIDE_BUILD_INFO
+        ? {}
+        : {
+            build: getBuild(),
+          }),
+    });
   });
-});
+}
 
 app.use(httpLogger);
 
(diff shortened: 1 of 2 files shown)


docs/docs/self-host/env.mdx
```diff
@@ -209,3 +209,7 @@ Logs collection is supported but disabled by default. You can set the `GROWTHBOO
...
 - **ROBOTS_TXT_PATH** - Define an absolute path to a robots.txt file that will be served on the `/robots.txt` path of your backend domain. If this env var is not defined, the app will look for a robots.txt file in the parent folder of the built `server.js` (which should be the `packages/back-end` directory in the default configuration)
+- **DISABLE_API_ROOT_PATH** - We include a default path in the API server to expose certain settings. Set to "true" to disable.
+- **API_ROOT_HIDE_API_HOST** - The value for the API_HOST variable is included in the payload from the default path. Set to "true" to omit.
+- **API_ROOT_HIDE_APP_ORIGIN** - The value for the APP_ORIGIN variable is included in the payload from the default path. Set to "true" to omit.
+- **API_ROOT_HIDE_BUILD_INFO** - Details about the build in-use are included in the payload from the default path. Set to "true" to omit.
```
jdorn: I don't know if I agree with the suggestion to add the fine-grained controls instead of just a single DISABLE flag.  I definitely don't like them all in the docs since it's so niche and could distract from more important options.  Even if we make them undocumented, there's a tiny maintenance cost to having them in the code, but not a big deal.

Kevin-Chant: Makes sense and I think I agree. I'll go ahead and remove them in favor of the single disable

packages/back-end/src/app.ts
```diff
@@ -197,19 +201,33 @@ app.get("/robots.txt", (_req, res) => {
...
-    name: "GrowthBook API",
-    production: ENVIRONMENT === "production",
-    api_host:
-      process.env.API_HOST ||
-      req.protocol + "://" + req.hostname + ":" + app.get("port"),
-    app_origin: APP_ORIGIN,
-    config_source: usingFileConfig() ? "file" : "db",
-    email_enabled: isEmailEnabled(),
-    build: getBuild(),
+if (!DISABLE_API_ROOT_PATH) {
```
jdorn: Instead of removing the route entirely, I think it's better to keep it and just return a hard-coded response.  Otherwise, the request flows through and ends up being caught by one of the middlewares.  Right now, that throws a 401 error, which is ok, but if we change that middlware in the future, it could end up throwing a 404 instead.  Just better to have that logic of what to return live within the route itself.

packages/back-end/src/util/secrets.ts
```diff
@@ -279,6 +279,11 @@ export const CLICKHOUSE_MAIN_TABLE = process.env.CLICKHOUSE_MAIN_TABLE || "";
 export const CLICKHOUSE_DEV_PREFIX =
   process.env.CLICKHOUSE_DEV_PREFIX || "test_";
 
+// Note: the Visual Editor relies on the information in this path, so disabling it will prevent some features from working correctly.
```
tzjames, line 282: This comment seems more important to include as part of the docs than as a comment.

docs/docs/self-host/env.mdx
```diff
@@ -209,3 +209,4 @@ Logs collection is supported but disabled by default. You can set the `GROWTHBOO
 - **CDN_HOST** - When set, this will update the implementation instructions within GrowthBook to override the `API_HOST` in cases where a CDN is used.
 - **EXPRESS_TRUST_PROXY_OPTS** - Express' [trust proxy](https://expressjs.com/en/5x/api.html#trust.proxy.options.table) setting value. Supports boolean (true/false), string values, and integer values for trusting the _n_<sup>th</sup> hop from the front-facing proxy server as the client. Leavy empty or specify `false` to use Express' default behavior. If you are running GrowthBook behind a proxy or load balancer, this is required to track the correct user IP for audit log events.
 - **ROBOTS_TXT_PATH** - Define an absolute path to a robots.txt file that will be served on the `/robots.txt` path of your backend domain. If this env var is not defined, the app will look for a robots.txt file in the parent folder of the built `server.js` (which should be the `packages/back-end` directory in the default configuration)
+- **DISABLE_API_ROOT_PATH** - We include a default path in the API server to expose certain settings. Set to "true" to disable.
```
tzjames:
```suggestion
- **DISABLE_API_ROOT_PATH** - We include a default path in the API server to expose certain settings. Set to "true" to withhold that information. The Visual Editor however relies on the information in this path, so disabling it will prevent some of its features from working correctly.
```

PR #4539, opened 2025-09-09
Add api key or system-type to event logs user field

### LINKED ISSUE #4421: [Feature] Audit Logs: user: null when using PAT for authentication
### Description of Feature

I know that API keys show as no user / "user": null in the audit logs but also PAT updates also seem to result in "user": null.

Is there a way for PAT authenticated updates to render in the audit logs?

Kevin-Chant:
### Features and Changes

* Adds a new type to the `EventUser` union: `{type: "system"}` for automatic/system-generated events
* Updates the `createEvent` helper to properly log the existing `api_key` when present. This was already working for AuditLogs but not for Events.
* Updates the `createEvent` helper to log `{type: "system"}` when neither `userId` nor `apiKey` are present

- Closes #4421

### Testing

Update an experiment in the following ways and then check the `/events` page to verify they are tagged appropriately:
* Manually update an experiment (metadata or refreshing results)
* Using the API authenticated with a Personal Access Token
* Using the API authenticated with an API key
* Automatically via the `updateExperimentResults` job

### Screenshots

image

### DIFF
--- a/packages/back-end/test/events/webhook.test.ts
+++ b/packages/back-end/test/events/webhook.test.ts
@@ -50,7 +50,7 @@ describe("webhook test events", () => {
           object: "webhook",
           projects: [],
           tags: [],
-          user: null,
+          user: { type: "system" },
         },
         dateCreated: expect.any(Date),
         event: "webhook.test",


packages/back-end/src/models/EventModel.ts
```diff
@@ -257,7 +257,14 @@ export const createEvent = async <
...
           }
-        : null,
+        : context.apiKey
+          ? {
+              type: "api_key",
+              apiKey: context.apiKey,
+            }
+          : {
+              type: "system",
+            },
```
Kevin-Chant, line 267: This is the riskiest part of the PR. I based the logic on what we have [here](https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/services/context.ts#L239) but am open to suggestions

gazzdingo: i think that is fine and we have used that pattern in the past so i dont think its to risky

PR #4541, opened 2025-09-09
fix: Cache job metrics + Use default export interval for OTEL

ahdriel:
### Features and Changes

While investigating increased CPU usage when running backend `with-tracing`, which enables OTEL, we have 2 changes that reduce CPU usage.

- Caching metric handler for jobs (as these are now shared by all jobs we expect memory usage to be lower in a scenario where many jobs are running in parallel)
- Use default value for `exportIntervalMillis` (60s) instead of 1s
  - The previously configured value is too aggressive and means we will be spending CPU cycles serializing the data, and sending it to the collector.
- Added `GB_DISABLE_JOB_LOGS` env, to keep OTEL metrics but disabled logs if needed.

### DIFF
--- a/packages/back-end/src/services/tracing.ts
+++ b/packages/back-end/src/services/tracing.ts
@@ -13,90 +13,92 @@ const normalizeJobName = (jobName: string) => {
     .toLowerCase();
 };
 
-export const trackJob = <T extends JobAttributesData>(
-  jobNameRaw: string,
-  fn: (job: Job<T>) => Promise<void>
-) => async (job: Job<T>) => {
-  let hasMetricsStarted = false;
+export const trackJob =
+  <T extends JobAttributesData>(
+    jobNameRaw: string,
+    fn: (job: Job<T>) => Promise<void>,
+  ) =>
+  async (job: Job<T>) => {
+    let hasMetricsStarted = false;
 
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/tracing.ts
```diff
@@ -16,41 +19,32 @@ export const trackJob =
...
+      getJobsRunningCounter().increment(attributes);
       hasMetricsStarted = true;
     } catch (e) {
       logger.error(
         { err: e, job: job.attrs },
-        `Error initializing counter for job`,
-      );
-    }
-    try {
-      histogram = metrics.getHistogram(`jobs.duration`);
```
jdorn, line 42: This was inside a try/catch block before but now it's not. Could creating the counter/histogram ever throw errors? Should we call `?.record` so if the counter is null it doesn't break?

ahdriel: It still is within a try/catch, but on a layer up, when we call `getJobsDurationHistogram`.

```
try {
        const end = performance.now();
        const elapsed = end - startTime;
        getJobsDurationHistogram().record(elapsed, attributes);
      } catch (e) {
        logger.error(
          { err: e, job: job.attrs },
          `Error recording duration metric for job`,
        );
      }
```

PR #4548, opened 2025-09-11
add defaultProject to teams, apply it on first load if set

bryce-fitzsimons:
- Add a defaultProject option for teams
- Default the selected project to whatever is allowed for the user's team(s) default project(s).
- Allow selection of projects outside of the defaultProject(s) constraint, just reset to the constraint whenever the app loads.

image


packages/front-end/components/Teams/TeamModal.tsx
```diff
@@ -69,11 +80,21 @@ export default function TeamModal({
...
+        </div>
       )}
+      <SelectField
+        label="Default Project"
+        value={form.watch("defaultProject")}
+        onChange={(p) => form.setValue("defaultProject", p)}
+        name="project"
+        initialOption="All Projects"
+        options={availableProjects}
+      />
```
jdorn: Only render if there are project options.

Also, should we call the initial value "all projects" or "none"? I can see arguments for either way

packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -141,7 +143,36 @@ export function useDefinitions() {
...
+function useTeamProjectConstraint() {
+  const { user, teams } = useUser();
+  const [project, setProject] = useLocalStorage(LOCALSTORAGE_PROJECT_KEY, "");
+
+  useEffect(() => {
+    if (!user?.teams || !teams || teamConstraintApplied) return;
+
+    const defaultProjects = new Set<string>();
+    teams.forEach((team) => {
+      if (team.defaultProject && user.teams!.includes(team.id)) {
```
jdorn: I know you are checking if user.teams exists right above this so the `!` is safe, but if that check above ever gets changed, this code could crash the app at runtime without a compile time error. Might be better to add the check here as well just in case.

packages/front-end/components/Teams/TeamModal.tsx
bryce-fitzsimons: was thinking All Projects so that it matches what we render in the left nav project selector

packages/front-end/services/DefinitionsContext.tsx
bryce-fitzsimons: wasn't exactly sure what you meant, but added a bit of redundant safety here...

packages/front-end/services/DefinitionsContext.tsx
```diff
@@ -141,7 +143,36 @@ export function useDefinitions() {
...
+  const [project, setProject] = useLocalStorage(LOCALSTORAGE_PROJECT_KEY, "");
+
+  useEffect(() => {
+    if (!user?.teams || !teams || teamConstraintApplied) return;
+
+    const defaultProjects = new Set<string>();
+    (teams || []).forEach((team) => {
+      if (team?.defaultProject && user!.teams!.includes(team.id)) {
+        defaultProjects.add(team.defaultProject);
+      }
```
mknowlton89: Could we not replace the `!` with `?` to be safer? Did some quick tests and didn't see any regressions.

```ts
if (team?.defaultProject && user.teams?.includes(team.id)) {
        defaultProjects.add(team.defaultProject);
      }
```

bryce-fitzsimons: good callout, switching to ?

Issue #4476, opened 2025-08-27
Ability to mark a resource as "Official" from in the UI (currently only possible via API)


jdorn:
Applies to Metrics, Fact Tables, Fact Metrics, Fact Table Filters, and Segments.

PR #4553, opened 2025-09-11
Allow users to mark resources as Official in the UI

mknowlton89:
### Features and Changes

With the introduction of `Official` resources (e.g. Official Metrics, Official Segments, etc) we want to give organizations the ability to mark resources as `Official` via the UI - Official resources can not be managed via the GrowthBook UI, but instead need to be managed via our API (in most cases, via Github).

Its likely that orgs have resources that were created prior to us releasing our `Official` resource support - this gives these orgs the ability to mark resources as official, preventing future changes from being made to them via the UI.

A resource is considered `Official` if it has a value for it's `managedBy` property. Currently, we support a resource being managed via the `config` - aka, the config.yml file, or via the `api` which is when a resource is managed only via our Rest API.

Given that only resources created via the `config.yml` file get `managedBy: "config"` this PR works by updating a resource's `managedBy` property and setting the value to `API`.

Meaning, when an org marks a resource as official manually via the UI, future changes to that can only be made via the Rest API. Either directly, or via a Github action (or similar).

It is possible for an org to undo this by updating the resource via the REST API and setting the `managedBy` property to an empty string. This would convert it back to a non-official resource.

For all Metrics, Fact Metrics, Fact Tables - to convert one of these to an official resource, a user will need to go to the resource id page, click the More Menu in the top-right corner, and select "Convert to Official [resourceName]".

For Segments and Fact Filters - since those don't have their own id page, a user can convert these to official by navigating to the list view, and select the "More Menu" for the row that corresponds to the resource they want to convert to official, and select the "Convert to Official [resourceName]".

Given that doing this will prevent future changes to the resource in the app, I built a "OfficialResourceModal" that displays an info callout that outlines the ramifications of the change, to prevent any accidental conversions.

- Closes https://github.com/growthbook/growthbook/issues/4476

### Dependencies

- None

### Fast Follows
- Currently, the PUT Segment Rest API docs are incorrect - This PR fixes the docs - https://github.com/growthbook/growthbook/pull/4563
- We have the concept of official dimensions, but we need to do some work to the UI - currently we're using a really old table style that just has edit and delete icons - we need to convert this to use the `DropdownMenu` component so we can add additional edit options, including `Convert to Official Dimension`.

### Testing

- [x] Ensure you can convert a Metric to an official metric via the UI
- [x] Ensure you can convert a Fact Metric to an official  fact metric via the UI
- [x] Ensure you can convert a Fact Table to an official fact table via the UI
- [x] Ensure you can convert a Fact Filter to an official fact filter via the UI
- [x] Ensure you can convert a Segment to an official segment via the UI
- [x] Ensure that once a resource is convert to "official" it can not be edited/deleted via the UI

### Screenshots

Screenshot 2025-09-12 at 10 12 34 AM

### DIFF
--- a/packages/back-end/src/routers/segment/segment.controller.ts
+++ b/packages/back-end/src/routers/segment/segment.controller.ts
@@ -238,20 +238,17 @@ export const putSegment = async (
     managedBy,
   } = req.body;
 
-  const datasource = segment.datasource;
-
   if (!context.permissions.canUpdateSegment(segment, { projects })) {
     context.permissions.throwPermissionError();
   }
 
-  const datasourceDoc = await getDataSourceById(context, datasource);
+  const datasourceDoc = await getDataSourceById(context, segment.datasource);
   if (!datasourceDoc) {
     throw new Error("Invalid data source");
   }
 
   await context.models.segments.updateById(id, {
     owner: owner || segment.owner,
-    datasource: datasource || segment.datasource,
     userIdType: userIdType || segment.userIdType,
     name: name || segment.name,
     sql: sql || segment.sql,
--- a/packages/back-end/src/routers/segment/segment.validators.ts
+++ b/packages/back-end/src/routers/segment/segment.validators.ts
@@ -35,5 +35,6 @@ export const updateSegmentValidator = segmentValidator
     organization: true,
     dateCreated: true,
     dateUpdated: true,
+    datasource: true,
   })
   .partial();


packages/back-end/src/routers/segment/segment.controller.ts
```diff
@@ -248,16 +250,17 @@ export const putSegment = async (
...
-    userIdType,
-    name,
-    sql,
-    description,
-    type,
-    factTableId,
-    filters,
-    projects,
+    owner: owner || segment.owner,
+    datasource: datasource || segment.datasource,
```
Kevin-Chant: I would remove this from the changeset if we're not allowing users to modify it anymore

packages/back-end/src/routers/segment/segment.validators.ts
```diff
@@ -29,9 +29,11 @@ export const createSegmentValidator = segmentValidator.omit({
...
-  dateUpdated: true,
-});
+export const updateSegmentValidator = segmentValidator
+  .omit({
+    id: true,
+    organization: true,
+    dateCreated: true,
+    dateUpdated: true,
+  })
+  .partial();
```
Kevin-Chant: This should also omit `datasource` to match the logic in the controller, right?

packages/front-end/components/OfficialResourceModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+}: {
+  close: () => void;
+  resourceType: ResourceType;
+  onSubmit: () => Promise<void>;
+}) {
+  return (
+    <Modal
+      open={true}
+      trackingEventModalType={`convert-to-official-${resourceType}`}
+      trackingEventModalSource="segment-list"
```
Kevin-Chant: Looks like this got copied and not changed. Might need to be a prop to `OfficialResourceModal`

mknowlton89: Whoooooops - Updating!

packages/back-end/src/routers/segment/segment.controller.ts
mknowlton89: I've got a question out to Jeremy about this right now - I had originally made this change, but wanted to verify before I did so.

mknowlton89:
Closing this in lieu of this PR - https://github.com/growthbook/growthbook/pull/4574

PR #4554, opened 2025-09-11
Allow passing includeArchived as query param to ListFeatures API

AlexHartford:
### Features and Changes

Currently, there's no way to intentionally omit archived features from the ListFeatures API.

The GrowthBook CLI depends on this API to drive code generation for types. Without a way to omit them in code, we're stuck with old references to archived features which creates noise especially for large teams doing a lot of experimentation.

This PR adds the ability to specify `includeArchived` as a query parameter to the ListFeatures API. `includeArchived` is already built functionally, this is simply allowing API consumers to specify the behavior.

- Works towards closing https://github.com/growthbook/growthbook-cli/issues/43
- If this PR is accepted, I will implement the update to the CLI to fully close this issue.

### Testing

- Hit the ListFeatures API with the `includeArchived` query parameter


packages/back-end/src/api/features/listFeatures.ts
```diff
@@ -19,7 +19,7 @@ export const listFeatures = createApiRequestHandler(listFeaturesValidator)(
   async (req): Promise<ListFeaturesResponse> => {
     const features = await getAllFeatures(req.context, {
       project: req.query.projectId,
-      includeArchived: true,
+      includeArchived: req.query.includeArchived || true,
```
jdorn: I don't think this will work since if you set it to false, it will do `false || true` and will still evaluate as true.

docs/src/data/SDKInfo.ts
```diff
@@ -3,7 +3,7 @@
 export default {
   js: {
     name: "JS SDK",
-    version: "1.6.1",
+    version: "1.6.2",
```
jdorn, line 6: It looks like you ran the `yarn bump-sdk-version` script.  Instead, you should run `yarn generate-api-types`.  

The change you are making does not require any version bumps.  Can you revert all of those changes from this PR?

jdorn (changes requested):

packages/back-end/src/api/features/listFeatures.ts
AlexHartford: Good catch, thank you!

docs/src/data/SDKInfo.ts
AlexHartford: Yes!

AlexHartford:
Thank you for the review @jdorn, and bearing with me on that first rough draft.

This PR should now be implemented properly. If I missed anything please let me know!

jrparish:
@jdorn Would you be able to take another look at this one when you get a chance?

The more we get into growthbook, the more the list continues to grow, so it would be nice to be able to clean some of this up so devs don't use keys that aren't available. 

Thanks!

PR #4561, opened 2025-09-11
feat(pipeline): Validate settings when saving

ahdriel:
### Features and Changes

- Add new `mode` property to pipeline settings in preparation of future work
- Validate pipeline settings when saving. When that happens we will create a new table, insert a row, drop the table and report back.
  - The current experience does not validate the settings, meaning users can enable this and failure would happen when refreshing results.

### Screenshots

https://www.loom.com/share/aabc6f256e1843d187ea32779fc6c4e8

### DIFF
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -585,7 +585,11 @@ export async function postValidatePipelineSettings(
     ...datasource,
     settings: {
       ...datasource.settings,
-      pipelineSettings,
+      pipelineSettings: {
+        ...pipelineSettings,
+        // Ensure minimum retention hours for testing
+        unitsTableRetentionHours: 1,
+      },
     },
   });
 


packages/back-end/src/controllers/datasources.ts
```diff
@@ -523,6 +531,155 @@ export async function putDataSource(
...
+    results.insert = {
+      result: "failed",
+      resultMessage: e instanceof Error ? e.message : String(e),
+    };
+  }
+
+  try {
+    if (results.create.result === "success") {
+      await integration.runTestQuery(
+        getPipelineValidationDropTableQuery({
```
ahdriel: @lukesonnet I should check `dropUnitsTable()` before running this right?

And for BigQuery I should update the createTable query to have an expiration of 1 minute in the `options`.

Let me know if this makes sense.

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -35,3 +53,31 @@ export function snowflakeCreateTableOptions(
       24,
   )}`;
 }
+
+export function getPipelineValidationCreateTableQuery({
```
ahdriel, line 57: TODO: This should integrate more with `createUnitsTable`, ensuring we are testing the same query we are going to use when refreshing results (with the exception that we won't have any data, just the table structure).

packages/back-end/src/types/Integration.ts
```diff
@@ -40,6 +40,8 @@ export class DataSourceNotSupportedError extends Error {
   }
 }
 
+export type DataType = "string" | "number" | "boolean" | "date" | "timestamp";
```
lukesonnet: Can we split this into "float" and "integer". I've wanted to do that in the incremental-refresh PR.

packages/back-end/src/controllers/datasources.ts
lukesonnet: I'm not sure I understand why you want to check `dropUnitsTable()`? Is it so you test that query directly? I'm not sure I understand why it's necessary for validating the connection.

> And for BigQuery I should update the createTable query to have an expiration of 1 minute in the options.

I don't think that matters if you're going to drop it, right? But I guess you might as well in case the drop command fails.

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -1,6 +1,24 @@
...
+  resultMessage?: string;
+};
+
+export type PipelineValidationResults = {
+  create: PipelineValidationResult;
+  insert: PipelineValidationResult;
+  drop: PipelineValidationResult;
+};
+
+export const DATA_SOURCE_TYPES_THAT_SUPPORT_PIPELINE_MODE: readonly DataSourceType[] =
```
lukesonnet, line 18: Eventually we might want it to differentiate based on "ephemeral" vs. "incrementalRefresh". I'm not entirely sure, but for v1 we probably will just have trino for incremental refresh to unblock the customer.

packages/shared/src/enterprise/pipeline.ts
lukesonnet: I'm not 100% sure. I guess you want to do this so we get the data types or something? But that's more about validating OUR queries rather than THEIR connection. I see how this relates to your other comment now.

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -35,3 +53,31 @@ export function snowflakeCreateTableOptions(
...
+    "string",
+  )}, created_at ${integration.getDataType("timestamp")})`;
+}
+
+export function getPipelineValidationInsertQuery({
+  tableFullName,
+}: {
+  tableFullName: string;
+}): string {
+  return `INSERT INTO ${tableFullName} (test_col, created_at) VALUES ('growthbook', CURRENT_TIMESTAMP)`;
```
lukesonnet: So for ephemeral mode we actually don't need INSERT, and ephemeral mode works in our free BQ test instance. We always are doing CREATE TABLE xxx AS (XXX), and this is free.

I don't think our customers will mostly be on free tier while be paying us for pipeline mode... but if this fails for other reasons, it shouldn't block ephemeral pipeline mode.

So really we should have the following
* Ephemeral mode
    * create table (CREATE TABLE ${xxx} AS (1 as x... perhaps more data types))
    * drop table (only used right now for databricks, and you can opt out of it). Technically we don't need this for all DBs as some people can just rely on the temp table disappearing and I would understand why an org might not want to grant us DROP. I feel like we should test the minimal necessary set, so maybe the logic of which test queries to run should depend on the settings. But this is not a big deal.
* Refresh mode
    * create, insert, and drop as you have it here would probably be good enough. Maybe there are separate permissions to test selecting partitions from a partitions table, but probably not.

packages/back-end/src/controllers/datasources.ts
ahdriel: Sorry my message was not clear at all -- I meant `should I check .dropUnitsTable()` in the SQL Integration, and just run the `drop` step if it returns true?

I think DataBricks is the only integration where `dropUnitsTable()` is set to `true`.

For BigQuery, for example, we use the `CREATE TABLE OPTIONS` with an expiration timestamp which automatically drops the table after it expires.

packages/shared/src/enterprise/pipeline.ts
ahdriel: Yep. Trying to keep the `incremental` changes to a minimum in this PR but this will evolve very soon.

packages/shared/src/enterprise/pipeline.ts
ahdriel: Good call! I'll update to skip the `insert data` step if the Mode is Ephemeral (which is the only one right now).

lukesonnet: Should we also skip drop unless databricks + it is enabled?

ahdriel: Updated!

PR #4574, opened 2025-09-15
Adds new permission, policy, and permission check helper

mknowlton89:
### Features and Changes

### DIFF
--- a/packages/shared/src/permissions/permissions.constants.ts
+++ b/packages/shared/src/permissions/permissions.constants.ts
@@ -45,6 +45,7 @@ export const POLICIES = [
   "DecisionCriteriaFullAccess",
   "SqlExplorerFullAccess",
   "HoldoutsFullAccess",
+  "ManageOfficialResources",
 ] as const;
 
 export type Policy = (typeof POLICIES)[number];
@@ -127,6 +128,16 @@ export const POLICY_PERMISSION_MAP: Record<Policy, Permission[]> = {
   TemplatesFullAccess: ["readData", "manageTemplates"],
   SqlExplorerFullAccess: ["readData", "runSqlExplorerQueries"],
   HoldoutsFullAccess: ["readData", "createAnalyses", "runQueries"],
+  ManageOfficialResources: [
+    "readData",
+    "createSegments",
+    "runQueries",
+    "manageFactTables",
+    "manageFactMetrics",
+    "manageFactFilters",
+    "createDimensions",
+    "manageOfficialResources",
+  ],
 };
 
 export const POLICY_DISPLAY_GROUPS: { name: string; policies: Policy[] }[] = [
@@ -164,6 +175,7 @@ export const POLICY_DISPLAY_GROUPS: { name: string; policies: Policy[] }[] = [
       "FactMetricsFullAccess",
       "DimensionsFullAccess",
       "SegmentsFullAccess",
+      "ManageOfficialResources",
     ],
   },
   {
@@ -384,6 +396,11 @@ export const POLICY_METADATA_MAP: Record<
     displayName: "Holdouts Full Access",
     description: "Create, edit, and delete holdouts",
   },
+  ManageOfficialResources: {
+    displayName: "Manage Official Resources",
+    description:
+      "Create, edit, and delete official resources. For example: Manage resources like Fact Tables, Metrics, Segments, etc that have been marked as 'Official'.",
+  },
 };
 
 export const DEFAULT_ROLES: Record<DefaultMemberRole, Role> = {
@@ -540,6 +557,7 @@ export const PROJECT_SCOPED_PERMISSIONS = [
   "manageCustomFields",
   "manageTemplates",
   "manageExecReports",
+  "manageOfficialResources",
 ] as const;
 
 export const GLOBAL_PERMISSIONS = [
--- a/packages/shared/src/permissions/permissionsClass.ts
+++ b/packages/shared/src/permissions/permissionsClass.ts
@@ -171,6 +171,21 @@ export class Permissions {
   };
 
   //Project Permissions
+  public canManageOfficialResources = (
+    resource: Pick<
+      | SegmentInterface
+      | FactTableInterface
+      | FactMetricInterface
+      | MetricInterface,
+      "projects"
+    >,
+  ): boolean => {
+    return this.checkProjectFilterPermission(
+      resource,
+      "manageOfficialResources",
+    );
+  };
+
   public canCreateSegment = (
     segment: Pick<SegmentInterface, "projects">,
   ): boolean => {


packages/back-end/src/controllers/metrics.ts
```diff
@@ -531,6 +556,30 @@ export async function putMetric(
...
 
+  if (metric.managedBy === "admin") {
+    if (
+      !context.permissions.canManageOfficialResources({
+        projects: metric.projects,
+      })
+    ) {
+      throw new Error(
+        "You do not have permission to update an official metric",
+      );
```
mknowlton89: Whoops - this should call context.permissions.throwPermissionError()

packages/back-end/src/controllers/metrics.ts
```diff
@@ -531,6 +556,30 @@ export async function putMetric(
...
+          projects: req.body.projects || metric.projects,
+        })
+      ) {
+        throw new Error(
+          "You do not have permission to update an official metric",
+        );
+      }
+    }
+  }
+
```
mknowlton89: I think this can be simplified to:

```ts
if (metric.managedBy === "admin" || req.body.managedBy === "admin") {
   if (
        !context.permissions.canManageOfficialResources({
          projects: req.body.projects || metric.projects,
        })
      ) {
              context.permissions.throwPermissionError()
          }
}
```

mknowlton89: Disregard - the above is more correct. We essentially need to check a few things:

1) If the existing metric has `managedBy: "admin"` then we need to check that the user has permission to update (given the existing metric's projects)
2) If the updates include `managedBy: "admin"` then we need to check that the user has permission to update and if they're trying to change the projects, use those, otherwise use the existing metric's projects
3) If neither of the above are true, just check the normal `canUpdateMetric` permission.

I suppose we could create a `canUpdateOfficialResources` helper (and also a canCreate and canDelete), where the update takes in the current and existing like `canUpdateMetric`.

docs/docs/official-resources.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+<MaxWidthImage border>
+![Example of an official metric](/images/metrics/official-metric.png)
+</MaxWidthImage>
+
+## Use Cases
+
+Official Resources are particularly useful for:
+
+- **External System Management**: Resources are managed by external systems (like data warehouses or CI/CD pipelines)
```
jdorn:
```suggestion
- **External System Management**: Resources are managed by external systems (like a version control system)
```

docs/docs/official-resources.mdx
```diff
@@ -0,0 +1,83 @@
...
+- **Cross-Project Consistency**: Resources need to be identical across multiple projects
+- **Governance**: Organizations want to prevent unauthorized changes to critical resources
```
jdorn, line 20: This is the number one use case, so I would move it first in the list

docs/docs/official-resources.mdx
```diff
@@ -0,0 +1,83 @@
...
+- Legacy Metrics
+- Segments
+- Dimensions
+
+## 3 Ways to Create Official Resources
+
+### 1. config.yml
+
+When resources are created via the `config.yml` file, they are automatically marked as "Official" and are not editable in the app. To make changes to these resources, you must update the `config.yml` file.
+
```
jdorn, line 37: We should add a note that not all resources are supported via config.yml (specifically fact tables/filters/metrics are not supported).

In general, I think we should discourage people from using `config.yml`, so might want to put this as the last option to make it less prominent.

packages/back-end/src/api/metrics/deleteMetric.ts
```diff
@@ -14,8 +14,18 @@ export const deleteMetricHandler = createApiRequestHandler(getMetricValidator)(
...
-      req.context.permissions.throwPermissionError();
+    if (metric.managedBy === "admin") {
+      if (
+        !req.context.permissions.canDeleteOfficialResources({
+          projects: metric.projects,
+        })
+      ) {
+        req.context.permissions.throwPermissionError();
+      }
+    } else {
```
jdorn: I don't think this is an else/if situation. We always want to check `canDeleteMetric`.  If it's managedBy admin, then we want to do an additional check on top of that.

Also, you can just revert all of these changes since you are already doing this check inside of `deleteMetricById`.  Same goes for all the update endpoints.

We currently aren't doing managedBy checks inside of the Model `insertMetric` and `insertMetrics`, but it makes sense to add it there too and keep it consistent.  That way, there will be no changes at all to any controllers or rest endpoints and everything will be self-contained inside of the model.

packages/back-end/src/models/MetricModel.ts
```diff
@@ -170,6 +170,14 @@ export async function deleteMetricById(
   if (metric.managedBy === "api" && context.auditUser?.type !== "api_key") {
     throw new Error("Cannot delete a metric managed by the API");
   }
+  if (
+    metric.managedBy === "admin" &&
+    !context.permissions.canDeleteOfficialResources({
+      projects: metric.projects,
+    })
+  ) {
+    throw new Error("Cannot delete an official metric");
```
jdorn: Either throw the built-in permission error or change this language to be `You do not have permission to ...`.  Otherwise, the error is not clear.  Same for updating below.

packages/front-end/components/Metrics/MetricName.tsx
```diff
@@ -77,6 +77,11 @@ export function OfficialBadge({
                 <>
                   a <code>config.yml</code> file
                 </>
+              ) : managedBy === "admin" ? (
+                <>
+                  an Admin or someone with the{" "}
+                  <code>ManageOfficialResources</code> policy
```
jdorn: I don't know if this is the right place to put the policy name.  Might be better to just have this say `This metric is being managed by admins`.  Not 100% accurate in all cases, but easier to understand.

packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -73,9 +73,18 @@ export function useCombinedMetrics({
 
   const combinedMetrics = [
     ...inlineMetrics.map((m) => {
-      const canDuplicate = permissionsUtil.canCreateMetric(m);
-      const canEdit = permissionsUtil.canUpdateMetric(m, {});
-      const canDelete = permissionsUtil.canDeleteMetric(m);
+      const canDuplicate =
+        m.managedBy === "admin"
+          ? permissionsUtil.canCreateOfficialResources(m)
+          : permissionsUtil.canCreateMetric(m);
```
jdorn: We shouldn't block duplicating.  It's going to be fairly common for someone without admin permission to duplicate an official metric and tweak it, which is totally fine.  We should just make sure in that case, that when they duplicate, the "Official" status is not copied over if they don't have permission for it.

packages/front-end/components/Metrics/MetricsList.tsx
```diff
@@ -73,9 +73,18 @@ export function useCombinedMetrics({
...
+      const canEdit =
+        m.managedBy === "admin"
+          ? permissionsUtil.canUpdateOfficialResources(m, {})
+          : permissionsUtil.canUpdateMetric(m, {});
+      const canDelete =
+        m.managedBy === "admin"
+          ? permissionsUtil.canUpdateOfficialResources(m, {})
+          : permissionsUtil.canDeleteMetric(m);
```
jdorn: This has the same else/if issue as the back-end.  Also, there are additional checks below that skip rendering edit/delete actions if `managedBy` is not empty, regardless of what the permissions say.

I think it makes sense to move all of the logic here.  So instead of having `managedBy` checks during rendering, we can just do something like this:

```ts
let canEdit = permissionsUtil.canUpdateMetric(m, {});
let canDelete = permissionsUtil.canDeleteMetric(m);

// Additional check if managed by admins
if (m.managedBy === "admin" && !permissionsUtil.canUpdateOfficialResources(m, {}) {
  canEdit = false;
  canDelete = false;
}
// Disable entirely if managed by API or config.yml
else if (m.managedBy) {
  canEdit = false;
  canDelete = false;
}
```

packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -127,6 +128,17 @@ export const POLICY_PERMISSION_MAP: Record<Policy, Permission[]> = {
...
   HoldoutsFullAccess: ["readData", "createAnalyses", "runQueries"],
+  ManageOfficialResources: [
+    "readData",
+    "createSegments",
+    "createMetrics",
+    "runQueries",
+    "manageFactTables",
+    "manageFactMetrics",
+    "manageFactFilters",
+    "createDimensions",
```
jdorn: I don't think we want/need all these dependencies.  Just have `manageOfficialResources` only (and maybe `readData`?).  This will allow more fine-grained control, since they can separately grant access to specific resource types as needed.

packages/front-end/pages/metric/[mid].tsx
```diff
@@ -133,11 +136,18 @@ const MetricPage: FC = () => {
...
   const canEditMetric =
-    permissionsUtil.canUpdateMetric(metric, {}) && !metric.managedBy;
+    metric.managedBy === "admin"
+      ? permissionsUtil.canUpdateOfficialResources(metric, {})
+      : permissionsUtil.canUpdateMetric(metric, {}) && !metric.managedBy;
   const canDeleteMetric =
-    permissionsUtil.canDeleteMetric(metric) && !metric.managedBy;
+    metric.managedBy === "admin"
+      ? permissionsUtil.canDeleteOfficialResources(metric)
+      : permissionsUtil.canDeleteMetric(metric) && !metric.managedBy;
```
Comment (jdorn) [inline packages/front-end/pages/metric/[mid].tsx]: Same here.  Fix these permission checks to always check for `canCreateMetric`, etc. in addition to `canCreateOfficialResources`.  

```ts
const canDeleteMetric =
    permissionsUtil.canDeleteMetric(metric) && (
      // Either not Official
      !metric.managedBy
      || (
        // Or it's Official and the user has permission to delete anyway
        metric.managedBy === "admin" && 
        permissionsUtil.canDeleteOfficialResources(metric)
      )
    )
```

packages/back-end/src/models/MetricModel.ts
mknowlton89: Whoops.

packages/front-end/components/Metrics/MetricsList.tsx
mknowlton89: Good call. As we've encapsulated the logic into the regular metric permission checks (`canUpdateMetric`, `canDeleteMetric`) I think we can simplify this even further to:

```ts
const canDuplicate = permissionsUtil.canCreateMetric({
        // Don't pass in managedBy as we allow non-admins to duplicate official metrics - the duplicated metric will be non-official
        projects: m.projects,
      });
      let canEdit = permissionsUtil.canUpdateMetric(m, {});
      let canDelete = permissionsUtil.canDeleteMetric(m);

      // Additional check if managed by admins
      if (m.managedBy && ["api", "config"].includes(m.managedBy)) {
        canEdit = false;
        canDelete = false;
      }
 ```
 
 We don't need the `m.managedBy === "admin"` check as that's already handled in the `canUpdateMetric` call.

File: packages/front-end/pages/metric/[mid].tsx (hunk 12 above)
Comment (mknowlton89) [reply packages/front-end/pages/metric/[mid].tsx]: Updated in [this commit](https://github.com/growthbook/growthbook/pull/4574/commits/13d1212c4ad07888e51052180320001dda7f0e90) - As we've encapsulated the logic inside of the regular metric permission checks (`canCreateMetric`, `canUpdateMetric`, and `canDeleteMetric` we can just check that.

File: packages/shared/src/permissions/permissions.constants.ts (hunk 11 above)
Comment (mknowlton89) [reply packages/shared/src/permissions/permissions.constants.ts]: @jdorn It feels like we should also give this policy the `runQueries` permission as well. Let me know if you disagree.

Comment (mknowlton89) [reply packages/shared/src/permissions/permissions.constants.ts]: Removing the permissions like `createMetrics` does lead to some weird experiences, specifically around duplication.

Ex:
If a user only has the `createMetrics` permission but not `manageOfficialResources` they can create metrics and duplicate metrics (even official metrics, but the duplicated metric won't be official).

However, if a user only has the `manageOfficialResources` policy, (which doesn't contain the `createMetrics` permission) they can then only duplicate official metrics, but they can't duplicate non-official metrics.

To get around this, the easiest way is to introduce a new helper to the `PermissionsClass` called `canDuplicateMetric`, which is true if the user has either `createMetrics` or `manageOfficialResources` permissions.

And on the front end, when we're determining whether to render the `Add Metric` button on `/metrics` we'll just need to check if the user `canCreateMetric` or `manageOfficialResources`. And if the latter, we'll probably just need to make the `managedBy` checkbox in the form read-only. Otherwise if a user with `manageOfficialResources` attempted to create a non-official resource, they'd get a permission error.

Comment (mknowlton89) [reply packages/shared/src/permissions/permissions.constants.ts]: Eh, this logic is getting gnarly again.

@jdorn Can you walk me through the scenario where an org might want someone to be able to manage official metrics, but not non-official metrics? 

Making this happen is possible, but it just requires a lot of logic that will make it more difficult to maintain and make things possibly more flaky.

Comment (jdorn) [reply packages/shared/src/permissions/permissions.constants.ts]: There is no use case for someone being able to manage official metrics, but not non-official metrics.

However, there is a use case for someone being able to only edit metrics, but not fact tables for example.  If that person gets the "ManageOfficialResources" policy, they should still only be able to edit metrics, not fact tables.

Comment (mknowlton89) [reply packages/shared/src/permissions/permissions.constants.ts]: Gotcha. Thanks for that clarity. I think I've got a working solution that will be easy to maintain. Running through my manual test plan now.

File: packages/back-end/src/controllers/metrics.ts [hunk 13]
```diff
@@ -450,9 +451,19 @@ export async function postMetrics(
...
+  if (
+    managedBy === "admin" &&
+    !context.hasPremiumFeature("manage-official-resources")
+  ) {
+    throw new Error(
+      "Your organization's plan does not support creating official metrics.",
+    );
+  }
+
+  if (!context.permissions.canCreateMetric(req.body)) {
```
jdorn: Nit: Since you added these checks in the model, you can remove all of this here.

jdorn (approved):
Because of the breaking change (setting `managedBy: api` now requires admin permissions), I'll do a quick check on GrowthBook Cloud to see if anyone will be impacted there.  Otherwise, good to go

PR #4576, opened 2025-09-15
feat: Allow unsetting of license key

tzjames:
### Features and Changes

Enterprise trial clients that want to downgrade to pro need their license key unset so that they can sign up for pro like normal. This PR allows them to do it.

### Testing

Go to /admin.  
Select an org with a license key.
Click edit pencil.
Delete the license key and save.
Click on the org again and edit.
See the license key is gone.


mknowlton89:
Tested locally and it worked as expected.

I wonder if there is a use case of updating the behavior of the `ShowLicenseInfo` component to allow an org to do this themselves?

mknowlton89 (approved):

PR #4577, opened 2025-09-15
Disallow creating legacy metrics for new orgs

msamper:
### Features and Changes

- If an organization does not already have at least 1 legacy metric, completely hide the ability to create new legacy metrics.
- If an organization has at least 1 fact table, default to the fact metric form. Doesn't matter if they have 1000 legacy metrics, if they started implementing fact tables we want that to be the default for all newly created metrics.
- Adds a new org metric setting to disable legacy metric creation

### Testing

- [x] Open the new metrics modal from an org with no legacy metrics. Should open the fact metric modal and not have the ability to create new legacy metrics.
- [x] Open the new metrics modal from an org with no fact tables and at least 1 legacy metric. Should open the legacy metric modal and include the ability to create new fact metrics.
- [x] Open the new metrics modal from an org with at least 1 fact table and some legacy metrics. Should open the fact metric modal not have the ability to create new legacy metrics.


packages/front-end/components/FactTables/NewMetricModal.tsx
```diff
@@ -62,30 +63,30 @@ export function MetricModal({
...
   // TODO: add an org setting to explicitly override this default
   let defaultType: "fact" | "legacy" = "fact";
-  if (filteredFactTables.length === 0) {
-    defaultType = "legacy";
-  } else if (filteredMetrics.length > filteredFactMetrics.length) {
+  if (filteredMetrics.length === 0) {
+    defaultType = "fact";
+  } else if (filteredFactTables.length > 0) {
+    defaultType = "fact";
+  } else if (filteredFactTables.length === 0 && filteredMetrics.length > 0) {
```
jdorn: Because the default value is fact already, this code might be easier to read if there was a single if block to set it to legacy with all the conditions.

packages/front-end/components/FactTables/NewMetricModal.tsx
```diff
@@ -62,30 +63,30 @@ export function MetricModal({
...
   const filteredFactTables = factTables
     .filter((f) => !datasource || f.datasource === datasource)
     .filter((f) => isProjectListValidForProject(f.projects, project));
 
   // Determine the most appropriate default type based on what the org has already created
-  // - If there are no fact tables yet, always default to legacy
-  // - If there are more legacy metrics than fact metrics, default to legacy
+  // - If there are no metrics, default to fact
+  // - If there are fact tables, default to fact
+  // - If there are no fact tables and there is at least one legacy metric, default to legacy
```
jdorn: We need to handle the case when there are no metrics and no fact tables yet. The default in this modal will be fact, but you can't make a fact metric unless you create a fact table first.

I think whenever possible, we should just avoid showing this modal in that case. So on the metrics page, prompt them to create a fact table first (link to the fact tables page) before prompting them to open this modal.

packages/front-end/components/FactTables/NewMetricModal.tsx
```diff
@@ -62,30 +63,24 @@ export function MetricModal({
...
   // - Otherwise, default to fact
   // TODO: add an org setting to explicitly override this default
   let defaultType: "fact" | "legacy" = "fact";
-  if (filteredFactTables.length === 0) {
-    defaultType = "legacy";
-  } else if (filteredMetrics.length > filteredFactMetrics.length) {
+  if (filteredFactTables.length === 0 && filteredMetrics.length > 0) {
     defaultType = "legacy";
   }
```
jdorn, line 77: I think it's safer to make this last if condition just `if (filteredFactTables.length === 0)`.  Otherwise, this modal can get into a broken state where defaultType is `fact`, but there are no fact tables yet.

Instead, we should focus on the entrypoints to link to `/fact-tables` when there are no metrics or fact tables yet.  You already fixed the ones on the metric list.  There is another entrypoint on `DataSourceMetrics` that we can fix as well if we want, although I don't think it's used very often.

jdorn (approved):

PR #4579, opened 2025-09-15
fix(custom-fields): Update feature custom field inputs when selected project changes

msamper:
### Features and Changes

Custom fields in the `FeatureModal` do not change when the feature's `selectedProject` changes. We're also setting the default values for the custom fields twice. Once within the `CustomFieldInput` component and another time when generating default values for the features. 

This PR fixes the bug and also removes unnecessary code around custom field default values when generating the Feature default values.


packages/front-end/components/Features/FeatureModal/index.tsx
```diff
@@ -158,33 +148,30 @@ export default function FeatureModal({
...
-      : undefined,
   });
 
   const form = useForm({ defaultValues });
 
+  const selectedProject = form.watch("project");
+  const customFields = filterCustomFieldsForSectionAndProject(
+    useCustomFields(),
+    "feature",
+    selectedProject,
```
jdorn, line 164: I understand changing `project` to `selectedProject`.  I'm unclear about all the other changes in this PR.

These are the main questions I have when looking at all of the other changes:
- Does removing the `hasCommercialFeature` check mean that custom fields are now available to all plans?
- If the feature being duplicated is missing some of the custom fields, what happens? (e.g. you add a new custom field after a feature was created and never update that feature, then you duplicate it)
- If you change projects, it doesn't look like any field values are updated. Does this actually work or does it only fix itself if you then go and edit the custom field values?

I haven't tested anything myself, so it could all be fine. Just, without context, these changes seem a little risky.

msamper: 1. No, the custom field inputs are still gated by `hasCommercialFeature` further down in the file.
2. The fields that existed on the feature being duplicated have the values copied over correctly in the form and the new field is empty
3. This actually seems to be broken today for experiments as well which is what I based these changes on. Changing the project and causing different fields to render doesn't remove any field values that are no longer valid for the project. This means that custom fields that are potentially not valid could be persisted in mongo for an experiment. We just don't render those values in the UI because we filter fields by the experiment project when presenting them along with their values on the overview page. I think it's probably worth fixing in this PR though. I can add that fix in as well.

PR #4581, opened 2025-09-15
feat(dashboards): Allow setting a dashboard as the default view for an experiment

msamper:
### Features and Changes

Allows users to set a dashboard as the default view for an experiment. Setting a dashboard as default, replaces the usual tabbed page view with the default dashboard on load. If needed, users can click on `View Regular Experiment View` to get back to the standard view.

Setting and removing a dashboard as the default view for an experiment can be done via the more menu for the dashboard the user would like to use or remove.

### Testing

- [x] Set a dashboard as the default view for an experiment and ensure the experiment page is re-rendered with the simple default dashboard view
- [x] Click on `View Regular Experiment View` and make sure it takes you back to the standard tabbed page view
- [x] Remove a dashboard as the default view and refresh to make sure the experiment loads with the standard tabbed view

### Screenshots

**Experiment with a dashboard set as default view** 
Screenshot 2025-09-17 at 4 22 24 PM

**Dashboard more menu with new option** 
Screenshot 2025-09-17 at 4 10 03 PM


packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -418,10 +436,22 @@ export default function TabbedPage({
...
           )}
+
+        {showDashboardView && (
+          <DashboardsTab
+            experiment={experiment}
+            initialDashboardId={experiment.defaultDashboardId ?? ""}
+            isTabActive
+            showDashboardView
+            switchToExperimentView={() => setShowDashboardView(false)}
+          />
```
Kevin-Chant, line 447: Probably out of scope for this PR, but do we want to lock exiting the dashboard view behind any permissions? That way non-SMEs only get the curated view

packages/front-end/enterprise/components/Dashboards/DashboardWorkspace.tsx
```diff
@@ -35,6 +35,7 @@ interface Props {
   mutate: () => void;
   submitDashboard: SubmitDashboard<UpdateDashboardArgs>;
   close: () => void;
+  switchToExperimentView?: () => void;
```
Kevin-Chant: I don't think `DashboardWorkspace` should take this prop because it's a fullscreen takeover for editing. IMO people should need to finish editing before they toggle this

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -68,19 +68,27 @@ interface Props {
...
+  showDashboardView = false,
+  switchToExperimentView,
+  mutateExperiment,
 }: Props) {
   const [dashboardId, setDashboardId] = useState(initialDashboardId);
   useEffect(() => {
-    if (initialDashboardId) {
+    if (experiment.defaultDashboardId && showDashboardView) {
+      setDashboardId(experiment.defaultDashboardId);
+    } else if (initialDashboardId) {
```
Kevin-Chant: Nit: would it be better to just use `initialDashboardId` to pass `defaultDashboardId` in (and maybe change the prop name)? Or is there a reason this isn't coming in through that prop?

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -394,6 +402,34 @@ export default function DashboardsTab({
                                   </Text>
                                 </Button>
                               )}
+                              {mutateExperiment && (
+                                <Button
```
Kevin-Chant: This should also check for update permissions on the experiment, right?

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -394,6 +402,34 @@ export default function DashboardsTab({
...
+                                    );
+                                    mutateExperiment();
+                                  }}
+                                >
+                                  <Text weight="regular">
+                                    {experiment.defaultDashboardId ===
+                                    dashboard.id
+                                      ? "Remove as default view"
+                                      : "Set as default view"}
+                                  </Text>
```
Kevin-Chant: Nit: might be worth a quick tooltip to explain what this means (if Kari agrees)

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -394,6 +402,34 @@ export default function DashboardsTab({
...
+                                      },
+                                    );
+                                    mutateExperiment();
+                                  }}
+                                >
+                                  <Text weight="regular">
+                                    {experiment.defaultDashboardId ===
+                                    dashboard.id
+                                      ? "Remove as default view"
+                                      : "Set as default view"}
```
Kevin-Chant: Nit: use title case to match other dropdown options
```suggestion
                                      ? "Remove as Default View"
                                      : "Set as Default View"}
```

Kevin-Chant (commented):
Only a few comments but looks pretty good and works as I'd expect in testing

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
msamper: Yes! Not sure why I did this. I was already passing in the defaultDashboardId into the prop from the parent component

packages/front-end/components/Experiment/TabbedPage/index.tsx
msamper: hmm I'm not sure what permission would be appropriate for locking down switching back to the standard view. The one that comes to mind is `createAnalyses`, but I imagine there might be some folks without that permission that might still be interested in digging in deeper

msamper: I think it's probably fine to not lock it down for now

Kevin-Chant: Yeah I think it'd be new, just mentioning it because it came up in some discussions before. Probably fine to leave it as-is for now but we might revisit it in the future

msamper: sounds good!

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
msamper: pinged Kari on Slack about this!

msamper: added!

PR #4582, opened 2025-09-16
(fix) Don't display Minimum numerator total as percentage when displayAsPercentage is enabled

msamper:
### Features and Changes


lukesonnet:
I haven't quickly approved because I'm concerned that we call this method elsewhere where this fix will be inappropriately used. I haven't had time to audit, but if you do I would appreciate it, or I can get to it later.

PR #4583, opened 2025-09-16
Add separate entrypoint for editing dashboard metadata

Kevin-Chant:
### Features and Changes

Removes the in-line title editing for dashboards in favor of a separate item in the moremenu. This new item pops the standard dashboard modal so users can toggle the "allow organization members to edit" setting after dashboard creation.

This also fixes an issue where title editing wasn't checking for the `canManage` permission, which would lead to permission errors being thrown

### Testing

Create two user accounts, one who owns the dashboard and a second in the same organization.
Check that the observed behavior matches expectations for the following states
| User Role / Dashboard Org Editing | Read-only | Experimenter | Admin |
|- | - | - | - |
| Org editing Disabled | No access | No access | Full access |
| Org editing Enabled | No access | Can edit blocks but not metadata | Full access |

### DIFF
--- a/packages/back-end/src/routers/dashboards/dashboards.controller.ts
+++ b/packages/back-end/src/routers/dashboards/dashboards.controller.ts
@@ -132,7 +132,7 @@ export async function updateDashboard(
         "enableAutoUpdates" in updates) &&
       !canManage
     ) {
-      context.permissions.throwPermissionError();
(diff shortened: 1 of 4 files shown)


bryce-fitzsimons (approved):
approved.

nit: perhaps should be "Edit Dashboard Settings" instead of "Details" (which could imply editing the blocks themselves)

PR #4590, opened 2025-09-17
Filter out Mixpanel from DataSourceTypeSelector options

msamper:
### Features and Changes

Removes Mixpanel as an option for new data sources and updates Mixpanel documentation to direct users to use the Data Pipelines approach 

### Screenshots

Screenshot 2025-09-17 at 4 57 28 PM


jdorn (commented):
Mixpanel supports exporting to BigQuery and Snowflake with [Data Pipelines](https://docs.mixpanel.com/docs/data-pipelines).  So someone using Mixpanel can still use GrowthBook, it just takes a few extra steps.

Instead of deleting the Mixpanel docs completely, we should change it to be about this Data Pipelines approach.  We might also have external links pointing to that existing docs URL that would be good not to break if possible.

PR #4591, opened 2025-09-18
Show chart of seat usage per installation for self-hosted enterprise …

tzjames:
### Features and Changes

Some organizations have different departments using their own infrastructure, and sharing the org wide license.  As the different departments need to communicate how many seats each are using it would be nice if they could see a chart of which department is using how many seats.  We have all that data already on the license, however only with an installationId identifying the installation.  That is not very user friendly.

This PR defaults to using the organization name as the installation name.  For IS_MULTI_ORG installations we need a new field, so we allow them to set an installationName that gets set to the installation object itself.  If it is not set we default to the installationId which at least is a unique identifier.

This PR shows "seats in use" on the license info if there is just one installation.  If there is more than one it shows a chart with the number of seats used in each installation.  

### Dependencies

The license server PR needs to land in order to start seeing the installation names in the chart instead of just the installation ids.

#### Single Installation
These should show a "Seats in use" field in the license section.
1. Go to http://localhost:3000/settings and see a "Seats in use" field.
Screenshot 2025-09-18 at 4 26 05 PM

#### Setup for multi-orgs
1. manually change the installation id on the one document in installation collection in mongo.  
2. Click refresh license on http://localhost:3000/settings

#### Airgapped
These are rare.  While we could calculate the number of seats being used, it is not on the license and no need to support this, so we just hide the field.
1. Set `IS_CLOUD=false` `LICENSE_KEY=<airgapped license key>` `MULTI_ORG=false`.  
2. Restart the server.
3. Go to http://localhost:3000/settings and see no Seats in use, chart, on installation name field. 
4. Click on Edit Organization and see no installation name setting.

#### Existing multi- installation
Existing multi installations won't have installationName set.  So if an org upgrades growthbook, but is using an existing license from the cache it may not have installationName set.  In this case we should show the installationId.
1. Set `IS_CLOUD=false` `LICENSE_KEY=<dev license key to dev server>` `MULTI_ORG=false`.  
2. Restart the server.
3. In mongo growthbook.licenses delete the installationName's that are set in the cache.
5. Go to http://localhost:3000/settings and see that the chart defaults to the installationIds.
6. Click on Edit Organization and see no installation name setting.

#### remote license multi- installation
1. click Refresh button in the license section.  This will send the org name as the installation to the license server and then return it.
2. see the chart with the current installation showing the current org name.
3. Click "Edit" in Organization settings.
4. Change the name of the current org.
5. See the chart get immediately updated with the new organization name for the current installation
6. In Mongo see that no default installation.name got save on the installation object. (this doesn't really matter now but early versions of the code defaulted to this value even if `IS_MULTI_ORG` was `false`)

#### remote license multi- installation with multi-org at an installation.
The license information is on the /admin page for multi-org installations.  It uses the installation.name for installationName and installation.id as the default.
1. Set `IS_CLOUD=false` `LICENSE_KEY=<dev license key to dev server>` `MULTI_ORG=true`.
2. Restart the server.  
3.  Go to http://localhost:3000/admin
4. Click the Refresh license button.
5. See the installation id for the name of the current installation.
7. Go to http://localhost:3000/settings
8. See installation name in the organization settings, having the default being the installationId.
9. Click "Edit" in Organization settings.
10. Change the installation name of the current org.
11.  Go to http://localhost:3000/admin
12. See the installation name in the chart updated to what was entered in the form.

### Screenshots
Single installation && cloud data:
Screenshot 2025-09-18 at 4 26 05 PM

Multi installation chart:
Screenshot 2025-09-18 at 4 24 55 PM

Multi-org multi-installation:
Organization settings with installation name
Screenshot 2025-09-18 at 5 22 47 PM
Edit org settings with installation name
Screenshot 2025-09-18 at 5 22 54 PM
Admin page showing license info
Screenshot 2025-09-18 at 4 52 44 PM

### DIFF
--- a/packages/back-end/src/enterprise/models/licenseModel.ts
+++ b/packages/back-end/src/enterprise/models/licenseModel.ts
@@ -34,6 +34,7 @@ const licenseSchema = new mongoose.Schema({
     of: {
       _id: false,
       date: Date,
+      installationName: String,
       userHashes: [String],
       licenseUserCodes: {
         invites: [String],
--- a/packages/back-end/src/models/InstallationModel.ts
+++ b/packages/back-end/src/models/InstallationModel.ts
@@ -2,30 +2,46 @@
  * This file is used to store the installation ID of the app.
  * Each installation (dev, prod, etc.) has a unique ID.
  * This is used as a singleton to identify the installation when registering licenses.
+ * The installation name is a user-settable friendly name for the installation
+ * for multi-org installations where we can't use the org name.
  */
 import { randomUUID } from "crypto";
 import mongoose from "mongoose";
+import { InstallationInterface } from "back-end/types/installation";
 
 const InstallationSchema = new mongoose.Schema({
   id: String,
+  name: String,
 });
 
-export type InstallationDocument = mongoose.Document & {
-  id: string;
-};
+export type InstallationDocument = mongoose.Document & InstallationInterface;
 
-export const InstallationModel = mongoose.model<InstallationDocument>(
+export const InstallationModel = mongoose.model<InstallationInterface>(
   "Installation",
   InstallationSchema,
 );
 
-export async function getInstallationId(): Promise<string> {
+function toInterface(doc: InstallationDocument): InstallationInterface {
+  return doc.toJSON();
+}
+
+export async function getInstallation(): Promise<InstallationInterface> {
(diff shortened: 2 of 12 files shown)


mknowlton89 (approved):
Tested locally and it looks great!

PR #4593, opened 2025-09-18
Run all fact metrics through new query

jdorn:
### Features and Changes

Since version 2.7, we've enabled Fact Table Query Optimization for Enterprise customers.  This involved combining multiple fact metrics together in a single query.  To accomplish this, we wrote a brand new (and improved) query generator.

There were still multiple instances when we would run fact metrics through the old query generator:

- If "Exclude in-progress data" was turned on for an experiment
- If the org was not on the Enterprise license
- If optimization was manually disabled in organization settings
- If percentile capping was used in either MySQL or Redshift
- If a quantile metric was used in either MySQL or Redshift
- If a ratio metric had different fact tables for the numerator and denominator

Over time, the two query generators have diverged significantly which has made maintenance difficult.  This PR now routes all fact metrics for all organizations to the new query generator.

Optimization (combining metrics in a single query) is still enterprise-only, so even though the new query generator is used, it will still run a separate query for each metric for non-enterprises.

TODO:
- [x] Merge code in branch `cross-fact-metric-query`
- [x] Test on experiments with "Exclude in-progress data"
- [x] Test on non-enterprise organizations (or when optimization is disabled in settings)
- [x] Test on Redshift with percentile capping
- [x] Test on Redshift with a quantile metric
- [x] Test with a ratio metric with different numerator/denominator fact tables
- [ ] Mark old query path as "deprecated" and only allow non-fact metrics in params

### DIFF
--- a/docs/docs/lib/csharp.mdx
+++ b/docs/docs/lib/csharp.mdx
@@ -10,121 +10,1093 @@ import SdkSupportedFeatures from '@site/src/components/SdkSupportedFeatures';
 
 # C#
 
-View the full documentation on [GitHub](https://github.com/growthbook/growthbook-c-sharp).
+The GrowthBook C# SDK supports all modern .NET platforms including .NET 6+, .NET Framework 4.6.1+, and .NET Standard 2.0+.
 
 <SdkResources sdk="csharp" />
 
 ## Installation
 
-```csharp
-    dotnet add package growthbook-c-sharp
+Install via NuGet Package Manager:
+
+```bash
+dotnet add package growthbook-c-sharp
 ```
 
-## Quick Usage
+Or via Package Manager Console:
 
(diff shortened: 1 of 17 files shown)


packages/front-end/components/Queries/QueryStatsRow.tsx
```diff
@@ -33,10 +37,12 @@ export default function QueryStatsRow({
   const usingPipelineMode = queries.some(
     (q) => q.queryType === "experimentUnits",
   );
-  const factTableOptimizedMetrics = queries
-    .filter((q) => q.queryType === "experimentMultiMetric")
-    .map((q) => getNumberOfMetricsInQuery(q))
-    .reduce((sum, n) => sum + n, 0);
+  const factTableOptimizedMetrics = hasOptimizedQueries
```
jdorn: Should this be `!hasOptimizedQueries` instead?

lukesonnet: fixed ty

PR #4594, opened 2025-09-19
Get signed urls for s3 and gcs

tzjames:
### Features and Changes
The File Proxy was inefficient as all data gets routed through the api.  
- By using signedUrls instead the api authenticates the url, returns a signed url and then the client can call directly. 
- Local uploads and downloads still work as before. 
-  As signed urls expire we add the expiry date to the image cache so that we don't keep fetching new signed images for the same src image, but only do so once the signed image expires.
- We get rid of the USE_FILE_PROXY completely

- Closes [this issue](https://github.com/orgs/growthbook/projects/11/views/1?pane=issue&itemId=128363360)

### Testing

For UPLOAD_METHOD in 's3', 'google-cloud', 'local':
Go to an experiment.
Upload a screen shot.
See the uploaded screen sheet.
Click between tabs, and in developer console see that no new images are downloaded.
After the 15 minute time out (or change that to 1 min) click around and see that the image is now redownloaded.

### DIFF
--- a/packages/back-end/src/routers/upload/upload.controller.ts
+++ b/packages/back-end/src/routers/upload/upload.controller.ts
@@ -37,10 +37,9 @@ export async function putUpload(
 ) {
   // Only handle direct uploads for local storage
   if (UPLOAD_METHOD !== "local") {
-    return res.status(400).json({
-      error:
-        "Direct uploads are only supported for local storage. Use /upload/signed-url-for-upload for cloud storage.",
-    });
+    throw new Error(
+      "Direct uploads are only supported for local storage. Use /upload/signed-url-for-upload for cloud storage.",
+    );
   }
 
   const contentType = req.headers["content-type"] as string;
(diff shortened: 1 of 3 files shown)


packages/back-end/src/routers/upload/upload.controller.ts
```diff
@@ -18,7 +31,17 @@ const extensionsToMimetype: Record<string, string> = {
...
+export async function putUpload(
+  req: AuthRequest<Buffer>,
+  res: Response<UploadResponse>,
+) {
+  // Only handle direct uploads for local storage
+  if (UPLOAD_METHOD !== "local") {
+    throw new Error(
+      "Direct uploads are only supported for local storage. Use /upload/signed-url-for-upload for cloud storage.",
+    );
+  }
```
jdorn, line 43: Not a big deal, but this change will break the UI during the deploy since the old front-end will still call this endpoint. Won't take effect until a full page load.

PR #4596, opened 2025-09-19
Mk/fact metric official UI

mknowlton89:
### Features and Changes


packages/front-end/pages/metric/[mid].tsx
```diff
@@ -481,6 +481,7 @@ const MetricPage: FC = () => {
             ) : null}
             {!metric.managedBy &&
             canEditMetric &&
+            permissionsUtil.canCreateOfficialResources(metric) &&
```
Comment (mknowlton89) [inline packages/front-end/pages/metric/[mid].tsx:484]: Small bug fix from previous PR - this ensures that only users with permission in orgs that have the feature will see this button.

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -156,17 +156,38 @@ export class FactMetricModel extends BaseClass {
...
+      !this.context.hasPremiumFeature("manage-official-resources")
+    ) {
+      throw new Error(
+        "Your organization's plan does not support creating official fact metrics.",
+      );
+    }
   }
 
   protected async beforeUpdate(existing: FactMetricInterface) {
+    // Check the admin permission here?
```
Kevin-Chant: Nit: remove comments

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -156,17 +156,38 @@ export class FactMetricModel extends BaseClass {
...
     if (existing.managedBy === "api" && !this.context.isApiRequest) {
-      throw new Error("Cannot update fact metric managed by API");
+      throw new Error(
+        "Cannot update fact metric managed by API if the request isn't from the API.",
+      );
     }
   }
 
   protected async beforeDelete(existing: FactMetricInterface) {
+    // Check the admin permission here?
```
Kevin-Chant: Same

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -2449,6 +2450,22 @@ export default function FactMetricModal({
                           </div>
                         )}
                       </div>
+                      {permissionsUtil.canUpdateOfficialResources(
+                        { projects: form.watch("projects") },
+                        {},
+                      ) && hasCommercialFeature("manage-official-resources") ? (
+                        <Checkbox
```
Kevin-Chant: Personally I don't think this should be hidden under "Show Advanced Settings". But if that's the way we want it to be, it shouldn't be nested under the `<TabsContent>` here because that lists it as an "Analysis Setting"

mknowlton89: Ah - I missed the sub-tabs there.

I've moved it out of the tabs, but kept it in the advanced settings section for now. I am going to sync with Kari as I think we can improve the flow. I'm curious to get her thoughts on the form and how we can improve the flow. I'd be interesting in making each `Create Metric` button a split button (if the org has the commercial feature) and giving the user the option to create a Metric, or an Official Metric, so call attention to the new feature.

PR #4598, opened 2025-09-21
Metric dimensions

bryce-fitzsimons:
V1 of metric dimensions on standard experiment results table

todo:
- [x] REST API support for fact metric `enableMetricDimensions`
- [x] consider supporting column definition CRUD for fact tables (including `isDimension`, `dimensionLevels`)
- Integrate with experiment dashboards (fast follow)


packages/shared/src/experiments.ts
```diff
@@ -110,13 +110,55 @@ export function getColumnRefWhereClause(
...
+    // Check if the dimension column exists in this fact table
+    const dimensionColumn = factTable.columns.find(
+      (col) => col.column === dimensionInfo.dimensionColumn,
+    );
+
+    if (!dimensionColumn || dimensionColumn.deleted) {
+      // Dimension column doesn't exist in this fact table, skip dimension filtering
+      // This can happen when the dimension column is in a different fact table
+      // (e.g., experiment events) but the metric queries another fact table (e.g., orders)
+      return [...where];
```
lukesonnet: Why are you returning here? I know this is an edge case, but don't you want to proceed as usual for this metric?

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -74,15 +75,47 @@ export async function runColumnTopValuesQuery(
     throw new Error("Top values not supported on this data source");
   }
 
-  const sql = integration.getColumnTopValuesQuery({
-    factTable,
-    column,
-    limit: 100,
-  });
+  // Try progressive lookback periods
+  const lookbackPeriods = [14, 365, 1234];
```
lukesonnet: I thought we were ditching this? I guess for most folks it won't bind since something will get found.

packages/shared/src/constants.ts
```diff
@@ -30,6 +30,7 @@ export const DEFAULT_SEQUENTIAL_TESTING_TUNING_PARAMETER = 5000;
 
 // Query settings
 export const DEFAULT_TEST_QUERY_DAYS = 30;
+export const MAX_METRIC_DIMENSION_LEVELS = 10;
```
lukesonnet: Can we make this 20? I really think 10 is too small. If people want more control, they can pick a smaller set of levels which we already help them do.

packages/back-end/src/models/MetricModel.ts
```diff
@@ -581,3 +582,52 @@ export async function generateMetricEmbeddings(
...
   return `Name: ${metric.name}\nDescription: ${metric.description}`;
 };
+
+/**
+ * Expands dimension metrics for fact metrics with enableMetricDimensions and adds them to the metricMap
+ */
+export function expandDimensionMetricsInMap(
+  metricMap: Map<string, ExperimentMetricInterface>,
+  factTableMap: FactTableMap,
+  baseMetrics: ExperimentMetricInterface[],
```
lukesonnet, line 642: nit: Should we change this to `FactMetricInterface[]` so that it is clear that it only supports fact metrics whenever a dev has to use this method?

packages/back-end/src/services/experiments.ts
```diff
@@ -190,10 +192,16 @@ export async function getExperimentMetricById(
...
-  if (isFactMetricId(metricId)) {
-    return context.models.factMetrics.getById(metricId);
+  // Handle dimension metric IDs by extracting the parent metric ID
+  const dimensionInfo = parseDimensionMetricId(metricId);
+  const actualMetricId = dimensionInfo.isDimensionMetric
+    ? dimensionInfo.parentMetricId
+    : metricId;
+
+  if (isFactMetricId(actualMetricId)) {
+    return context.models.factMetrics.getById(actualMetricId);
```
lukesonnet, line 207: is it crazy to override this method or have some other helper that can automatically get the base ID even if you pass in the full, built ID?

packages/back-end/src/services/experiments.ts
```diff
@@ -570,7 +578,13 @@ export function getSnapshotSettings({
   const currentDate = new Date();
 
   const metricSettings = expandMetricGroups(
-    getAllMetricIdsFromExperiment(experiment),
+    getAllExpandedMetricIdsFromExperiment(
```
lukesonnet: Doesn't expandMetricGroups kind of need to happen *before* we get the expanded metric IDs from the experiment? Seems like `getAllExpandedMetricIdsFromExperiment` contains `expandMetricGroups` via `getAllMetricIdsFromExperiment` so `expandMetricGroups` is redundant here.

I feel like `expandMetricGroups` should become some internal method we don't call and it only gets called from `getAllExpandedMetricIdsFromExperiment` or the like, but anyways, I guess it doesn't matter too much here.

packages/back-end/src/services/experiments.ts
```diff
@@ -2811,14 +2826,64 @@ export async function getSettingsForSnapshotMetrics(
...
+            col.isDimension &&
+            !col.deleted &&
+            (col.dimensionLevels?.length || 0) > 0,
+        );
+
+        dimensionColumns.forEach((col) => {
+          const dimensionLevels = col.dimensionLevels || [];
+
+          // Create a metric for each dimension level
+          dimensionLevels.forEach((value) => {
```
lukesonnet: This logic feels like it is repeated in a bunch of places in this PR. Is there a better way to handle this?

packages/back-end/src/jobs/refreshFactTableColumns.ts
bryce-fitzsimons: Oh yeah "forgot" to remove this (or, hoped you didn't notice!). I can remove

packages/shared/src/experiments.ts
bryce-fitzsimons: no good reason, will undo this

packages/back-end/src/services/experiments.ts
bryce-fitzsimons: Yeah I think so. IMO the model methods should be a thin layer around the data. Passing in an (invalid?) ephemeral ID here would imply that something is materialized rather than computed, I believe...

packages/back-end/src/models/MetricModel.ts
bryce-fitzsimons: maybe? I prefer not to because this is meant to be a chained step that accepts whatever is in the metrics list already.

packages/back-end/src/services/experiments.ts
bryce-fitzsimons: Indeed it was redundant. However, I did encounter a bug here: when creating the snapshot, dimension expansion was not happening when the metric was in a group. Fixed.

I'm not sure how I feel about expandMetricGroups yet, there are some use cases where you want to do it without bring up dimensions. Let's punt for now...

packages/back-end/src/services/experiments.ts
bryce-fitzsimons: Good call out. I created a shared function. It doesn't quite handle the table row generation, but at least its a bit more consolidated across a few other uses.

packages/front-end/components/Metrics/MetricName.tsx
```diff
@@ -230,18 +242,47 @@ export default function MetricName({
...
+            style={{
+              fontSize: "1em",
+              lineHeight: "1em",
+              marginTop: "-2px",
+              marginLeft: "4px",
+              color: "var(--blue-11)",
+            }}
+          />
+        ) : null}
       </span>
```
mknowlton89, line 272: @bryce-fitzsimons From what I can tell this introduced a bug. Given that `officialBadgePosition` is optional, and we don't set a default prop, if we don't pass in a value for `officialBadgePosition` the badge isn't rendered, even if the metric is official.

Looking at the code, it's not clear to me why if we want to render it to the right, we don't use the `OfficialBadge` component - is there a specific reason or should we be using the `OfficialBadge`?

If so, I think we'll likely want to specify a default badge position (I think left would make more sense given the current UI).

mknowlton89: Draft PR for reference - https://github.com/growthbook/growthbook/pull/4640

Issue #3986, opened 2025-04-25
[Bug] `scripts/migrate-encryption-key.js` is broken


BlueTogepi:
### Summary

`scripts/migrate-encryption-key.js` is broken with the imports.
To replicate this, use the [docker-compose.yaml](https://github.com/growthbook/growthbook/blob/main/docker-compose.yml) and run `docker-compose run growthbook yarn migrate-encryption-key dev` like in the docs.

#### Expected Behavior

It should not error.

#### Current Behavior

Here's the error trace

```
% docker-compose run growthbook yarn migrate-encryption-key dev    
[+] Creating 1/0
 ✔ Container growthbook-mongo-1  Created                                                                                                                                                                                                                    0.0s 
[+] Running 1/1
 ✔ Container growthbook-mongo-1  Started                                                                                                                                                                                                                    0.3s 
yarn run v1.22.22
$ yarn workspace back-end migrate-encryption-key dev
$ node dist/scripts/migrate-encryption-key.js dev
node:internal/modules/cjs/loader:1051
  throw err;
  ^

Error: Cannot find module 'back-end/src/models/DataSourceModel'
Require stack:
- /usr/local/src/app/packages/back-end/dist/scripts/migrate-encryption-key.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1048:15)
    at Module._load (node:internal/modules/cjs/loader:901:27)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at require (node:internal/modules/helpers:130:18)
    at Object.<anonymous> (/usr/local/src/app/packages/back-end/dist/scripts/migrate-encryption-key.js:4:27)
    at Module._compile (node:internal/modules/cjs/loader:1233:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1287:10)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Module._load (node:internal/modules/cjs/loader:938:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/usr/local/src/app/packages/back-end/dist/scripts/migrate-encryption-key.js'
  ]
}

Node.js v20.5.1
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed.
Exit code: 1
Command: /usr/bin/node
[...]

PR #4602, opened 2025-09-22
fix: scripts with broken imports

tzjames:
### Features and Changes

- Closes 
https://github.com/growthbook/growthbook/issues/3986

### Testing
a bit annoying to test as I'd have to rebuild docker image, but this just reverts changes in 3032 so this should be safe.

### DIFF
--- a/packages/back-end/src/scripts/generate-openapi.mjs
+++ b/packages/back-end/src/scripts/generate-openapi.mjs
@@ -36,14 +36,14 @@ async function run() {
   // Step 3: Add additional named types for easier access
   // Export each schema as a named type
   output += `import { z } from "zod";\n`;
-  output += `import * as openApiValidators from "back-end/src/validators/openapi";\n`;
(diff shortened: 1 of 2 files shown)


packages/back-end/src/scripts/generate-openapi.mjs
```diff
@@ -36,14 +36,14 @@ async function run() {
   // Step 3: Add additional named types for easier access
   // Export each schema as a named type
   output += `import { z } from "zod";\n`;
-  output += `import * as openApiValidators from "back-end/src/validators/openapi";\n`;
+  output += `import * as openApiValidators from "../src/validators/openapi";\n`;
```
jdorn: Not sure about this change. This isn't an import within the scripts directory. It's generating a file that will live in types. I checked the other files in types and they use absolute imports.

tzjames: good point

PR #4603, opened 2025-09-22
fix: prevent datasources with null UserIdType from killing the UI

tzjames:
### Features and Changes

Somehow a [user](https://growthbookapp.slack.com/archives/C05QYS2UFRD/p1758540063742019) got into the state with a userIdType of null.  It crashed the UI so they couldn't delete the datasource.  They shouldn't get into that state, but this at least shows it.  

### Testing

Create a datasource and manually set a userIdType to null. See the datasource page load.


tzjames:
We modified that by hand in the database.  Hopefully we can figure out how it got into this state and prevent it from happening again.

PR #4610, opened 2025-09-23
Improve Design System page

ahdriel:
### Features and Changes

Improve Design System page
- Add sidebar for each UI component
- Separate each component stories into its own `.stories.tsx` component
  - The design system file was getting very big with all the mocks/setup
- Rename `AnalysisSummaryPopover` to `AnalysisSummaryResults` and move it to `/ui/`
  - The previous name is misleading as the component is responsible only for the content, not how it is rendered (even though it is usually rendered inside a Popover)

#### Before
Screenshot 2025-09-23 at 2 23 38 PM

#### After
Screenshot 2025-09-23 at 2 23 50 PM


msamper (commented):
Thanks so much for doing this! So much easier to navigate through these now. I noticed a few of the components seem to not be rendering properly given the more limited real estate they have now. 
The ones I caught that looked off are:

- Premium Callout
- Stepper
- Radio Cards

ahdriel:
@msamper Thanks for the review!

>Premium Callout

I have not updated this, as I think this is something we should fix in the component as well as we cannot always guarantee a min-width where this component gets rendered.

>Stepper

Can you clarify what looks broken in the design-system page? From my testing it is working as expected.

>Radio Cards

I updated the configuration to more clearly separate the settings and the rendering of the component

msamper:
> @msamper Thanks for the review!
> 
> > Premium Callout
> 
> I have not updated this, as I think this is something we should fix in the component as well as we cannot always guarantee a min-width where this component gets rendered.
> 
> > Stepper
> 
> Can you clarify what looks broken in the design-system page? From my testing it is working as expected.
> 
> > Radio Cards
> 
> I updated the configuration to more clearly separate the settings and the rendering of the component

Makes sense, re: Premium Callout. I agree we should fix in the component. 
With the stepper, on my end it seems like its no longer taking the full width of the container its in

**Now**
Screenshot 2025-09-25 at 10 45 31 AM
**Before**
Screenshot 2025-09-25 at 10 46 13 AM

PR #4623, opened 2025-09-24
Mk/official fact table

mknowlton89:
### Features and Changes

This PR is a fast-follow of PR https://github.com/growthbook/growthbook/pull/4574 and updates Fact Tables to include the new `Official` status/logic.

### Dependencies

None

### Testing

GB App Testing
- [x] Ensure that if a fact table doesn't have a `managedBy` property, any user with the `ManageOfficialResources` can convert it to an "Official Table", so long as the org has the feature.
- [x] Once that fact table has been converted to "Official" ensure that the `managedBy` property equals `admin`
- [x] Then, ensure someone who just has permission to edit that fact table can't edit or delete it, but they can duplicate it
- [x] If a user without the `ManageOfficialResources` policy duplicates an official fact table, ensure the new fact table isn't Official
- [x] Then, ensure someone who has the `ManageOfficialResources` policy on their role CAN make edits to it or delete it
- [x] Create a new custom role that JUST has the `ManageOfficialResources` and `createFactTables` policies and ensure they can edit official fact tables that have the `managedBy = "admin"` value.
- [x] Ensure that an official fact table that is managedBy the API still can't be edited/deleted via the UI
- [x] Ensure that only enterprise orgs see the "Convert to Official Table" icon in [ftmid].tsx MoreMenu
- [x] Ensure that only enterprise orgs see the checkbox in the fact table form to mark a table as "Official".
- [x] Ensure that only enterprise orgs can set `managedBy: "admin"` for a fact table on creation. (I have left the logic as-is for updating/deleting tables - I don't want orgs to not be able to make changes to/delete official tables after they've downgraded. We'll just block them from creating new ones).

Rest API Testing
- [x] Ensure that fact tables that have the `managedBy` property set to "admin" can only be created/updated/deleted by someone with the `ManageOfficialResources` policy (aka, an admin, or a custom role that contains that policy)
- [x] Ensure that only enterprise orgs can set `managedBy: "admin"` for a fact table on creation via the REST API. (I haven't touched the logic around updating/deleting - I don't want orgs to get into a state where they can't edit/delete official tables if they've downgraded, we just block creating new ones)
- [x] Ensure that fact tables that have the `managedBy` property set to `api` can only be created/updated/deleted by someone with the `ManageOfficialResources` policy (aka, an admin, or a custom role that contains the policy)


packages/back-end/src/models/FactTableModel.ts
```diff
@@ -206,18 +225,6 @@ export async function createFactTable(
...
-  factTables: Omit<CreateFactTableProps, "datasource">[],
-  datasource: string,
-): Promise<FactTableInterface[]> {
-  const factTablesToCreate = factTables.map((factTable) =>
-    createPropsToInterface(context, { ...factTable, datasource }),
-  );
-
-  return (await FactTableModel.insertMany(factTablesToCreate)).map(toInterface);
-}
-
```
mknowlton89, line 225: As far as i can tell this isnt being used anymore.

packages/shared/src/permissions/permissionsClass.ts
```diff
@@ -600,15 +600,29 @@ export class Permissions {
...
 
   public canCreateFactTable = (
-    factTable: Pick<FactTableInterface, "projects">,
+    factTable: Pick<FactTableInterface, "projects" | "managedBy">,
   ): boolean => {
+    if (factTable.managedBy && ["admin", "api"].includes(factTable.managedBy)) {
+      if (!this.canCreateOfficialResources(factTable)) {
+        return false;
+      }
+    }
```
Kevin-Chant, line 609: Should these checks be including `api` or only checking `canCreateOfficialResources` for `admin`? It looks like the checks for the request being API are happening elsewhere, so this is imposing an extra layer of permissions that wasn't happening before, right? Is that intentional?

packages/front-end/pages/fact-tables/[ftid].tsx
```diff
@@ -181,9 +207,10 @@ export default function FactTablePage() {
             <OfficialBadge type="Fact Table" managedBy={factTable.managedBy} />
           </h1>
         </div>
-        {canEdit && (
-          <div className="ml-auto">
-            <MoreMenu>
+        {/* {canEdit && ( */}
```
Comment (Kevin-Chant) [inline packages/front-end/pages/fact-tables/[ftid].tsx]: Is there a reason this is commented instead of removed?

packages/shared/src/permissions/permissionsClass.ts
Kevin-Chant: Oh I answered my own question by checking the original PR, never mind!
> setting managedBy: api now requires admin permissions

Kevin-Chant (approved):
LGTM with one unnecessary comment to remove

PR #4625, opened 2025-09-24
feat(dimensions): Add job to refresh precomputed dimensions traffic data

ahdriel:
### Features and Changes

- Add new job to refresh precomputed dimensions data based on traffic
  - Target only exposure queries where there are dimensions, and one of them is not custom (so it is based on traffic data)
  - Skip updates if they were refreshed in the last 7 days (we can add a config for this in the near future), making it per-EAQ or DataSource if we want

### DIFF
--- a/packages/back-end/src/jobs/updateExposureQueriesDimensionSlices.ts
+++ b/packages/back-end/src/jobs/updateExposureQueriesDimensionSlices.ts
@@ -1,0 +1,266 @@
+import Agenda, { Job } from "agenda";
+import { subDays } from "date-fns";
+import { logger } from "back-end/src/util/logger";
+import { getContextForAgendaJobByOrgId } from "back-end/src/services/organizations";
+import {
+  _dangerousGetAllDatasourcesWithExposureQueriesWithAutomaticDimensionSlices,
+  getDataSourceById,
+  updateDataSource,
+} from "back-end/src/models/DataSourceModel";
+import { usingFileConfig } from "back-end/src/init/config";
+import { getIntegrationFromDatasourceId } from "back-end/src/services/datasource";
+import {
+  createDimensionSlices,
+  _dangerousGetAllDimensionSlicesByIds,
+} from "back-end/src/models/DimensionSlicesModel";
+import { DimensionSlicesQueryRunner } from "back-end/src/queryRunners/DimensionSlicesQueryRunner";
+
+const QUEUE_EXPOSURE_QUERIES_DIMENSION_SLICES_UPDATES =
+  "queueExposureQueriesDimensionSlicesUpdates";
+
+const UPDATE_SINGLE_DATASOURCE_EXPOSURE_QUERIES_DIMENSION_SLICES =
+  "updateSingleExposureQueryDimensionSlices";
+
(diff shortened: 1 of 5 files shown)


packages/back-end/src/jobs/updateExposureQueriesDimensionSlices.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  await startUpdateJob();
+
+  async function startUpdateJob() {
+    const job = agenda.create(
+      QUEUE_EXPOSURE_QUERIES_DIMENSION_SLICES_UPDATES,
+      {},
+    );
+    job.unique({});
+    job.repeatEvery("1 day");
```
lukesonnet, line 58: It should be every 7 days, and eventually maybe customizable.

packages/back-end/src/jobs/updateExposureQueriesDimensionSlices.ts
```diff
@@ -0,0 +1,266 @@
...
+> {
+  try {
+    const potentialDataSourcesToUpdate =
+      await _dangerousGetAllDatasourcesWithExposureQueriesWithAutomaticDimensionSlices();
+
+    // Include only exposure queries that have at least one non-custom slice
+    const dataSourcesAndExposureQueries = potentialDataSourcesToUpdate
+      .map((ds) => {
+        const exposureQueries = ds.settings?.queries?.exposure || [];
+        const queriesWithAutomaticDimensionSlices = exposureQueries.filter(
```
lukesonnet, line 97: If an exposure query has no `dimensionMetadata` but they do not have `disablePrecomputedDimensions` we should consider updating for a kind of "auto-onboarding" to pre-computed dimensions.

Maybe, however, there should be a separate onboarding flow that signals they want this automatic job to run?

packages/back-end/src/jobs/updateExposureQueriesDimensionSlices.ts
```diff
@@ -0,0 +1,266 @@
...
+      lookbackDays: DEFAULT_LOOKBACK_DAYS,
+    });
+
+    await queryRunner.waitForResults();
+    const outputModel = queryRunner.model;
+
+    exposureQuery.dimensionSlicesId = outputModel.id;
+    exposureQuery.dimensionMetadata = exposureQuery.dimensions.map((d) => ({
+      dimension: d,
+      specifiedSlices:
```
lukesonnet, line 246: If the output has no slices or errors, we should consider not updating the results at all. I would rather we preserve stale data than overwrite it with nothing.

packages/back-end/src/jobs/updateExposureQueriesDimensionSlices.ts
ahdriel: While we run the `queueUpdates` job every day, in the job we check for slices that have been updated more than 7 days ago ([this line / parameter](https://github.com/growthbook/growthbook/pull/4625/files#diff-ea9bb28b02ab66b3eb377aeebc006683e2f548470d98224c80aeac3574189fcfR116)).

So each individual dimensionSlice has their own 7 days clock, instead of us updating every single record every Saturday for example.

This will also allow us to have different schedules per user in the future.

PR #4626, opened 2025-09-24
feat(exposure-debug): Add a user exposure debugging tool

msamper:
### Features and Changes

Adds a new debugging tool for viewing all exposures for a given unit id within a specified lookback window.
Runs a query that does a UNION of all the exposure queries for a the specified `userIdType` and `unitId`

### Testing

- [x] Lookup exposures for a valid unit ID with exposures and verify that results are returned
- [x] Lookup exposures for a unit ID without exposures and verify it's handled gracefully
- [x] Verify all the available dimensions are rendered as columns
- [x] Lookup exposures for userIdType that has multiple exposure queries and verify the generated query is correct

### Screenshots

Screenshot 2025-09-24 at 5 25 46 PM

**With two exposure queries for the same userIdType** 
Screenshot 2025-09-25 at 9 44 15 AM


packages/front-end/pages/exposure-debugger.tsx
```diff
@@ -0,0 +1,289 @@
...
+                <div className="col-sm-2">
+                  <SelectField
+                    labelClassName="font-weight-bold"
+                    value={form.watch("lookbackDays") ?? ""}
+                    onChange={(value) => form.setValue("lookbackDays", value)}
+                    options={[
+                      { value: "7", label: "Last 7 days" },
+                      { value: "30", label: "Last 30 days" },
+                      { value: "90", label: "Last 90 days" },
+                      { value: "365", label: "Last year" },
```
msamper, line 180: Could use some feedback on some good options here. For now I went with past week, month, quarter, and year

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -228,6 +228,11 @@ const navlinks: SidebarLinkProps[] = [
         href: "/archetypes",
         path: /^archetypes/,
       },
+      {
+        name: "Debugger",
```
bryce-fitzsimons: This is not a particularly helpful label. Something more specific to User Exposures

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2411,6 +2413,75 @@ export default abstract class SqlIntegration
...
+        // Get all available dimensions for this exposure query
+        const availableDimensions = exposureQuery.dimensions || [];
+        const tableAlias = `t${i}`;
+
+        // Create dimension columns for ALL possible dimensions
+        const dimensionSelects = allDimensionNames.map((dim) => {
+          if (availableDimensions.includes(dim)) {
+            return `${tableAlias}.${dim} AS ${dim}`;
+          } else {
+            return `null AS ${dim}`;
```
lukesonnet: I remember at some point that basic "NULL" as a data type that differs across engines and may not be UNION-able to certain datatypes. We could:

(1) not worry about this until it crops up or 
(2) just render all dimensions as strings and use `this.castToString()` on both the actual dimension value and the explicit NULL here.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2411,6 +2413,75 @@ export default abstract class SqlIntegration
...
+
+        const dimensionSelectString = dimensionSelects.join(", ");
+
+        return `
+          SELECT timestamp, experiment_id, variation_id, ${dimensionSelectString} FROM (
+            ${compileSqlTemplate(exposureQuery.query, {
+              startDate: startDate,
+            })}
+          ) ${tableAlias}
+          WHERE ${exposureQuery.userIdType} = '${params.unitId}' AND timestamp >= ${this.toTimestamp(startDate)}
```
lukesonnet: We may want to cast to string on the userIdType. I don't think we enforce it is a string.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2411,6 +2413,75 @@ export default abstract class SqlIntegration
...
+        `;
+      })
+      .join("\nUNION ALL\n")}
```
lukesonnet: I think we should put this into a CTE and use `SQL_ROW_LIMIT` to limit this to 1000 rows. I would also order by timestamp in DESC order as well so the most recent exposures are on top. It always shows the most recent 1k rows. 

Nice-to-have: warning saying 1k row max is reached if rows are 1k long.

packages/front-end/pages/exposure-debugger.tsx
lukesonnet: These seem fine. If you want to add custom, you can do something like I did here: https://github.com/growthbook/growthbook/blob/main/packages/front-end/components/MetricAnalysis/MetricAnalysis.tsx#L330-L363, but it's kind gross.

packages/front-end/pages/exposure-debugger.tsx
```diff
@@ -0,0 +1,303 @@
...
+                    ]}
+                    sort={false}
+                    required
+                  />
+                </div>
+
+                <div className="col-sm-2">
+                  <Button onClick={onSubmit} disabled={loading}>
+                    Lookup
```
lukesonnet: nit: Should this just say "Query" so it matches other language and it's clearer that it's going to query? IDK I don't have strong opinions.

packages/front-end/pages/exposure-debugger.tsx
```diff
@@ -0,0 +1,303 @@
...
+                                      className="d-inline-block"
+                                      style={{
+                                        width: 150,
+                                        lineHeight: "14px",
+                                      }}
+                                    >
+                                      {variation.name}
+                                    </span>
+                                  </div>
+                                ) : (
```
lukesonnet, line 284: This stuff is nice and all, but we should consider also returning the raw variation id anyways. I really like it when I get the actual data in my debugging tool.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2411,6 +2413,75 @@ export default abstract class SqlIntegration
...
+    query: string,
+  ): Promise<UserExperimentExposuresQueryResponse> {
+    const { rows, statistics } = await this.runQuery(query);
+    return {
+      rows: rows.map((row) => {
+        return {
+          timestamp: row.timestamp,
+          experiment_id: row.experiment_id,
+          variation_id: row.variation_id,
+          ...row,
```
lukesonnet, line 2486: Does this correctly handle NULLs? I think it does, but you might want to check.

packages/back-end/src/integrations/SqlIntegration.ts
msamper: Addressed this in my latest commit. Let me know if that's what you had in mind re: CTE

packages/back-end/src/integrations/SqlIntegration.ts
msamper: Went ahead and used `this.castToString()` to be safe

packages/front-end/pages/exposure-debugger.tsx
msamper: Added the raw variation id in another column

packages/front-end/pages/exposure-debugger.tsx
```diff
@@ -0,0 +1,315 @@
...
+                      <Callout status="warning">
+                        Row limit reached. Only showing the most recent
+                        exposures.
+                      </Callout>
+                    )}
+                  </div>
+
+                  <div className="table-responsive">
+                    <table className="table gbtable">
+                      <thead>
```
lukesonnet, line 230: SortableTH? Might be nice.

bryce-fitzsimons (approved):
looks good, did not test thoroughly

packages/front-end/pages/exposure-debugger.tsx
msamper: I'll add this separately. I'm having some issues with sorting the dimension columns and I think there might be a bug with sorting `null` values or something in our `useSearch`

PR #4628, opened 2025-09-25
use ts-node for migrate-encryption-key script

tzjames:
### Features and Changes

This [PR](https://github.com/growthbook/growthbook/pull/4602) tried to fix it, but since it imports other files it fails. 
It turns out using ts-node instead will make it work.  So this PR does that and undoes the relative imports in the previous fix.

### Testing
`yarn migrate-encryption-key dev`
See no import errors thrown.


ahdriel (approved):
Nice solution! Easier than having to deal with compiled vs non-compiled code, etc.

PR #4634, opened 2025-09-25
Metric Explorer

jdorn:
### Features and Changes

Lightweight Metric Explorer.  Part of our upcoming Product Analytics suite.

This MVP is very bare bones.  Future improvements:

- Ability to group by a metric dimension
    - + new Bar Chart visualization for this
- Ability to select multiple metrics and graph them together
- Ability to add filters to individual metrics (applies to both numerator and denominator)
- Ability to add filters globally (that also applies to the population when applicable)
- More time granularity (weekly, hourly options in addition to daily)
- Smarter caching (switching from 90 days to 30 days should re-use the results when possible)
- Include all identifier types in the SQL (currently requires a new query when changing the identifier)
- Add support for Quantile metrics

### DIFF
--- a/packages/back-end/src/routers/metric-analysis/metric-analysis.controller.ts
+++ b/packages/back-end/src/routers/metric-analysis/metric-analysis.controller.ts
@@ -12,6 +12,7 @@ import { getExperimentMetricById } from "back-end/src/services/experiments";
 import { getIntegrationFromDatasourceId } from "back-end/src/services/datasource";
 import { getContextFromReq } from "back-end/src/services/organizations";
 import { AuthRequest } from "back-end/src/types/AuthRequest";
+import { metricAnalysisSettingsValidator } from "./metric-analysis.validators";
 
 export const postMetricAnalysis = async (
   req: AuthRequest<CreateMetricAnalysisProps>,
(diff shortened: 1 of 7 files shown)


packages/back-end/src/routers/metric-analysis/metric-analysis.controller.ts
```diff
@@ -129,14 +131,40 @@ export async function cancelMetricAnalysis(
...
   res: Response<{
     status: 200;
     metricAnalysis: MetricAnalysisInterface | null;
   }>,
 ) {
   const context = getContextFromReq(req);
 
+  // If we're trying to match specific analysis settings
+  if (req.query.settings) {
+    const rawSettings = JSON.parse(req.query.settings || "{}");
```
Kevin-Chant, line 148: Do we want to `try/catch` this?

packages/front-end/components/DataViz/MetricExplorer.tsx
```diff
@@ -0,0 +1,590 @@
...
+                        form.setValue("analysisSettings", newAnalysisSettings);
+                      }
+
+                      updateResults();
+                    }}
+                  />
+
+                  {metric && factTable && (
+                    <Select
+                      label="Unit"
```
Kevin-Chant: Can we mark these fields as required? The empty state on the graph doesn't indicate if you're missing required config

Kevin-Chant (approved):
LGTM. Two minor notes but they're non-blocking

Kevin-Chant:
I'm splitting out the frontend code from this PR since we plan to land that separately, and merging the backend code on its own

PR #4638, opened 2025-09-29
Statsig importer: optional mapping for attributes

bryce-fitzsimons:
Skip attribute conversion (Statsig to GB) if unchecked


tzjames (approved):
Approving to not slow you down, but there is still lint to fix.

Also perhaps it's a bit of style difference but I would have not had `skipAttributeMapping` as an argument to `mapStatsigAttributeToGB` as the name implies that function shouldn't run.  I'd have done:
```
const statsigAttribute =  experiment.idType || "user_id"
const hashAttribute = skipAttributeMapping ?  statsigAttribute : mapStatsigAttributeToGB(statsigAttribute);
```
instead of
```
const hashAttribute = mapStatsigAttributeToGB(experiment.idType || "user_id", skipAttributeMapping);
```
However your way saves a bunch of characters and I don't feel terribly strongly.

PR #4645, opened 2025-09-30
Show dashboard IDs in url to make sharing easier

Kevin-Chant:
### Features and Changes

Currently, the selected dashboard isn't persisted into the URL which makes it harder to share and prevents browser navigation between dashboards.

This fixes the somewhat hacky initial implementation of `tabPath` by not hiding it from the user and properly updating it on tab change or dashboard change.

### Testing

* Navigate the different experiment tabs and between multiple dashboards
* Use the dropdown moremenu on a dashboard and confirm that the share button leads directly to that dashboard
* The browser back/forward buttons should be able to switch between dashboards
* Copying the url directly should lead to a specific dashboard
* Deleting a dashboard (while another one exists) should load that other dashboard and update the url

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/index.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/index.tsx
@@ -111,7 +111,9 @@ export default function TabbedPage({
     `tabbedPageTab__${experiment.id}`,
     "overview",
   );
-  const [tabPath, setTabPath] = useState("");
+  const [tabPath, setTabPath] = useState(
+    window.location.hash.replace(/^#/, "").split("/").slice(1).join("/"),
+  );
 
   const router = useRouter();
 


packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
mknowlton89, line 134: I ran into one edge case that I thought I'd call out.

I created two dashboards, then deleted one. Then, if I tried to navigate directly to the deleted dashboard, when I landed there, all I saw was an empty dashboards dropdown modal and an empty screen. It feels like maybe we want to have a "Dashboard not found" section that coaches users to create a new one.

Kevin-Chant: Yeah dashboard deletion isn't ideal right now. I've also got an option to make it just pick the first dashboard from the list but I'm not sure what UX we actually want. For now I'm gonna leave it alone but I'd like to fix it in the near future

PR #4646, opened 2025-09-30
Pivot table MVP

mknowlton89:
### Features and Changes

This PR adds a new `Pivot Table` visualization option to the Sql Explorer & Product Analytics Dashboard builder. It currently supports only a single level for columns, rows, and dimensions, though it was built in a manner to where we can extend each to n-dimensions later.

This is an MVP of the feature. Future fast follows will be:
- Additional levels of depth for columns, rows, & dimensions - need to update the UI to support this.
- Configurable colors / enable/disable opacity shading
- Collapsible roll-up rows (maybe - this hasn't been confirmed)

### Testing

- [x] Build out a pivot table with some sample data and ensure the aggregation and transformation is correct
- [x] Ensure that if a column is a `date` then we format it according to the granularity

### DIFF
--- a/packages/back-end/package.json
+++ b/packages/back-end/package.json
@@ -58,6 +58,7 @@
     "async-lock": "^1.4.0",
     "aws-sdk": "^2.696.0",
     "body-parser": "^1.19.0",
+    "bs58": "^6.0.0",
     "compression": "^1.7.4",
     "cookie-parser": "^1.4.5",
     "cors": "^2.8.5",
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -3,9 +3,10 @@ import cloneDeep from "lodash/cloneDeep";
 import * as bq from "@google-cloud/bigquery";
 import { SQL_ROW_LIMIT } from "shared/sql";
 import {
-  DATA_SOURCE_TYPES_THAT_SUPPORT_PIPELINE_MODE,
+  PIPELINE_MODE_SUPPORTED_DATA_SOURCE_TYPES,
   getPipelineValidationCreateTableQuery,
   getPipelineValidationDropTableQuery,
+  getPipelineValidationInsertQuery,
   type PipelineValidationResults,
 } from "shared/enterprise";
 import { AuthRequest } from "back-end/src/types/AuthRequest";
@@ -535,6 +536,7 @@ export async function putDataSource(
 
 /**
  * Validates the pipeline settings for the given data source.
+ * Ensures we can create, insert and drop tables.
  */
 export async function postValidatePipelineSettings(
   req: AuthRequest<
@@ -566,9 +568,13 @@ export async function postValidatePipelineSettings(
     });
   }
 
-  if (!DATA_SOURCE_TYPES_THAT_SUPPORT_PIPELINE_MODE.includes(datasource.type)) {
+  if (
+    !PIPELINE_MODE_SUPPORTED_DATA_SOURCE_TYPES[pipelineSettings.mode].includes(
+      datasource.type,
+    )
+  ) {
     return res.status(400).json({
-      message: "This data source does not support pipeline mode.",
+      message: `This data source does not support pipeline mode ${pipelineSettings.mode}`,
     });
   }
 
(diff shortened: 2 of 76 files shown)


packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -687,6 +842,241 @@ export function DataVisualizationDisplay({
...
+                          fontWeight: row.isBold ? "bold" : "normal",
+                          background:
+                            typeof value === "number" && maxCellValue > 0
+                              ? `color-mix(in srgb, #5071de ${Math.round(
+                                  Math.max(
+                                    0,
+                                    Math.min(1, value / maxCellValue),
+                                  ) * 85,
+                                )}%, transparent)`
+                              : undefined,
```
mknowlton89: I need to test this with negative values.

packages/front-end/services/dataVizConfigUtilities.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // For other chart types, store as single object (first element)
+  // If empty array, set to undefined
+  return {
+    ...config,
+    xAxis: (valueArray.length > 0 ? valueArray[0] : undefined) as Partial<
+      BarChart | LineChart | AreaChart | ScatterChart
+    >["xAxis"],
+  } as Partial<DataVizConfig>;
```
Kevin-Chant: These two typecasts are unnecessary; I'd remove them both for simplicity and for extra type safety in the future

packages/front-end/services/dataVizConfigUtilities.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export function updateXAxisConfig(
+  config: Partial<DataVizConfig>,
+  updates: Partial<xAxisConfiguration>,
+): Partial<DataVizConfig> {
+  const configs = getXAxisConfig(config);
+  const current = configs[0];
```
Kevin-Chant, line 55: This is fine for now because we're only doing a single x axis for pivot tables in the mvp right? I assume we'll pass an index in instead of 0 when we expand it?

packages/front-end/components/DataViz/PivotTable.tsx
Kevin-Chant, line 1: Like we said in Slack I think we should move this file wholesale to `packages/front-end/components/SqlExplorer/PivotTable.tsx`. I don't see anything that needs to be in `Dashboards/`

packages/front-end/components/DataViz/PivotTable.tsx
```diff
@@ -0,0 +1,603 @@
...
+          return new Intl.DateTimeFormat(undefined, {
+            year: "numeric",
+          }).format(d);
+        case "none":
+        default:
+          return d.toLocaleDateString();
+      }
+    }
+    return String(xVal);
+  };
```
Kevin-Chant, line 131: Nit: This feels like it should be a helper on its own rather than defined inside another function. You could curry the values of `isDateType` and `xConfig` to preserve the current usage

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -282,67 +285,120 @@ export function DataVisualizationDisplay({
     });
   }, [dataVizConfig.filters, rows]);
 
-  // TODO: Support multiple y-axis and dimension fields
-  const xConfig = requiresXAxis(dataVizConfig)
-    ? dataVizConfig.xAxis
-    : undefined;
+  // TODO: Support multiple y-axis fields
+  const xAxisConfigs = getXAxisConfig(dataVizConfig);
+  const xConfig = xAxisConfigs[0];
```
Kevin-Chant, line 290: Double checking here as well - is the impact of this 0 index that we cant add multiple X axes yet (and will be changed in a future PR)?

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
```diff
@@ -149,18 +172,19 @@ export default function DataVizConfigPanel({
                 onDataVizConfigChange({
                   ...dataVizConfig,
                   chartType: v as DataVizConfig["chartType"],
-                });
+                } as Partial<DataVizConfig>);
```
Kevin-Chant: Unnecessary typecast
```suggestion
                });
```

packages/front-end/services/dataVizConfigUtilities.ts
mknowlton89: That's correct.

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
mknowlton89: Yup - that is correct. I've got this listed as a fast-follow in Notion for the PA Dashboards project.

PR #4652, opened 2025-10-01
Hide Segments sidenav link

mknowlton89:
### Features and Changes

This PR hides the `Segments` side-nav link for organizations that don't have any Segments, but leaves it for orgs with existing Segments. It also introduces a few lightweight notes in the docs that outlines we've begun to phase out Segments in favor of Dimensions. 

- Closes https://github.com/growthbook/growthbook/issues/4651

### Dependencies

- None

### Testing

- [x] Log in as an org that has existing Segments, and ensure that the `Segments` link under the `Metrics and Data` tab is rendered.
- [x] Log in as an org that doesn't have any existing Segments, and ensure that the `Segments` link under the `Metrics and Data` tab is NOT rendered.

### DIFF
--- a/packages/front-end/components/Layout/SidebarLink.tsx
+++ b/packages/front-end/components/Layout/SidebarLink.tsx
@@ -7,6 +7,7 @@ import { FiChevronRight } from "react-icons/fi";
 import { GrowthBook, useGrowthBook } from "@growthbook/growthbook-react";
 import { GlobalPermission } from "back-end/types/organization";
 import { Permissions } from "shared/permissions";
+import { SegmentInterface } from "back-end/types/segment";
 import { AppFeatures } from "@/types/app-features";
 import { isCloud, isMultiOrg } from "@/services/env";
 import { PermissionFunctions, useUser } from "@/services/UserContext";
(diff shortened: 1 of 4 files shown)


packages/front-end/components/Layout/SidebarLink.tsx
```diff
@@ -40,7 +42,7 @@ export type SidebarLinkProps = {
 
 const SidebarLink: FC<SidebarLinkProps> = (props) => {
   const { permissions, superAdmin } = useUser();
-  const { project } = useDefinitions();
+  const { project, segments } = useDefinitions();
```
ahdriel, line 45: question: Does this mean that we are now loading all segments on all page loads?

Or maybe we were already doing that but just not surfacing it?

The question here is around performance in case a company has lots of segments.

ahdriel (approved):

packages/front-end/components/Layout/SidebarLink.tsx
mknowlton89: Good callout.

My understanding (and I did a bit of testing), is that the `useDefinitions` hook ultimately leverages the `useApi` hook that then leverages the `useSWR` hook - so in my testing, when loading for the first time, we hit the `/organizations/definitions` endpoint, but then the response is cached, so we're not constantly hitting that same endpoint.

But also, this shouldn't have any impact as we're not making any additional calls with this, just leveraging the existing data from an existing call.

PR #4656, opened 2025-10-02
chore(SQL): Move cross-table ratio metric to fact SQL fns

lukesonnet:
### Features and Changes

Move cross-table ratio metrics to the fact metric SQL functions (e.g. `getExperimentFactMetricsQuery`). It tries to group all ratio metrics that have the same numerator and denominator fact table pairing into one group and then run that via the new query. Previously, they were passing through `getExperimentMetricQuery` which soon should not be used for any fact metrics (#4593).

This also unlocks CUPED for these cross-table ratio metrics.

TODO:
- [x] Improve comment so that it renders with a more relevant name in Async Query modal
- [x] Implement for bandits
- [x] Consider throwing clear implementation error if 3+ fact tables appear in one set of metrics (shouldn't happen if get metric groups is working) [this is kind of done in `getFactMetricCTE`]
- [x] Test metric analysis for these and regular fact metrics
- [x] Test multiple cross-table metrics with same and different denominators
- [x] Measure rows scanned/data usage vs. legacy approach

### Testing

Existing single table ratios (and all other) fact metric queries tested by `sqlintegration.test.ts` were unaffected.

Added more tests for cross-table ratio metrics and they look correct.

Manually tested a single cross-table ratio metric and results were identical before and after.

### DIFF
--- a/packages/back-end/src/models/FeatureCodeRefs.ts
+++ b/packages/back-end/src/models/FeatureCodeRefs.ts
@@ -1,6 +1,10 @@
 import mongoose from "mongoose";
-import { omit } from "lodash";
 import { FeatureCodeRefsInterface } from "back-end/types/code-refs";
+import {
+  ToInterface,
+  getCollection,
+  removeMongooseFields,
+} from "back-end/src/util/mongo.util";
 import { ApiCodeRef } from "back-end/types/openapi";
 import { OrganizationInterface, ReqContext } from "back-end/types/organization";
 import { ApiReqContext } from "back-end/types/api";
@@ -41,10 +45,10 @@ const FeatureCodeRefsModel = mongoose.model<FeatureCodeRefsInterface>(
   featureCodeRefsSchema,
 );
 
-function toInterface(doc: FeatureCodeRefsDocument): FeatureCodeRefsInterface {
-  const ret = doc.toJSON<FeatureCodeRefsDocument>();
-  return omit(ret, ["__v", "_id"]);
-}
+const COLLECTION = "featurecoderefs";
+
+const toInterface: ToInterface<FeatureCodeRefsInterface> = (doc) =>
+  removeMongooseFields(doc);
 
 export function toApiInterface(doc: FeatureCodeRefsDocument): ApiCodeRef {
   return {
@@ -93,12 +97,16 @@ export const upsertFeatureCodeRefs = async ({
     { upsert: true },
   );
 
-  return await FeatureCodeRefsModel.find({
-    feature,
-    repo,
-    branch,
-    organization: organization.id,
-  }).then((docs) => docs.map(toInterface));
+  const docs = await getCollection(COLLECTION)
+    .find({
+      feature,
+      repo,
+      branch,
+      organization: organization.id,
+    })
+    .toArray();
+
+  return docs.map((d) => toInterface(d));
 };
 
 export const getFeatureCodeRefsByFeatures = async ({
@@ -112,12 +120,16 @@ export const getFeatureCodeRefsByFeatures = async ({
   features: string[];
   organization: OrganizationInterface;
 }): Promise<FeatureCodeRefsInterface[]> => {
-  return await FeatureCodeRefsModel.find({
-    repo,
-    branch,
-    feature: { $in: features },
-    organization: organization.id,
-  }).then((docs) => docs.map(toInterface));
+  const docs = await getCollection(COLLECTION)
+    .find({
+      repo,
+      branch,
+      feature: { $in: features },
+      organization: organization.id,
+    })
+    .toArray();
+
+  return docs.map((d) => toInterface(d));
 };
 
 export const getAllCodeRefsForFeature = async ({
@@ -127,10 +139,14 @@ export const getAllCodeRefsForFeature = async ({
   feature: string;
   organization: OrganizationInterface;
 }): Promise<FeatureCodeRefsInterface[]> => {
-  return await FeatureCodeRefsModel.find({
-    feature,
-    organization: organization.id,
-  }).then((docs) => docs.map(toInterface));
+  const docs = await getCollection(COLLECTION)
+    .find({
+      feature,
+      organization: organization.id,
+    })
+    .toArray();
+
+  return docs.map((d) => toInterface(d));
 };
 
 export const getAllCodeRefsForOrg = async ({
@@ -138,7 +154,11 @@ export const getAllCodeRefsForOrg = async ({
 }: {
   context: ReqContext | ApiReqContext;
 }): Promise<FeatureCodeRefsInterface[]> => {
-  return await FeatureCodeRefsModel.find({
-    organization: context.org.id,
-  }).then((docs) => docs.map(toInterface));
+  const docs = await getCollection(COLLECTION)
+    .find({
+      organization: context.org.id,
+    })
+    .toArray();
+
+  return docs.map((d) => toInterface(d));
 };
--- a/packages/front-end/components/Experiment/BreakDownResults.tsx
+++ b/packages/front-end/components/Experiment/BreakDownResults.tsx
@@ -28,6 +28,7 @@ import { useDefinitions } from "@/services/DefinitionsContext";
 import {
   applyMetricOverrides,
   ExperimentTableRow,
+  compareRows,
 } from "@/services/experiments";
 import ResultsTable, {
   RESULTS_TABLE_COLUMNS,
@@ -43,6 +44,7 @@ import {
 import ResultsMetricFilter from "@/components/Experiment/ResultsMetricFilter";
 import { SSRPolyfills } from "@/hooks/useSSRPolyfills";
 import useOrgSettings from "@/hooks/useOrgSettings";
+import { useOrganizationMetricDefaults } from "@/hooks/useOrganizationMetricDefaults";
 import UsersTable from "./UsersTable";
 
 const numberFormatter = Intl.NumberFormat();
@@ -113,6 +115,13 @@ const BreakDownResults: FC<{
     metric: ExperimentMetricInterface,
   ) => React.ReactElement | string;
   noStickyHeader?: boolean;
+  sortBy?: "metric-tags" | "significance" | "change" | null;
+  setSortBy?: (s: "metric-tags" | "significance" | "change" | null) => void;
+  sortDirection?: "asc" | "desc" | null;
+  setSortDirection?: (d: "asc" | "desc" | null) => void;
+  analysisBarSettings?: {
+    variationFilter: number[];
+  };
 }> = ({
   experimentId,
   dimensionId,
@@ -149,11 +158,17 @@ const BreakDownResults: FC<{
   hideDetails,
   renderMetricName,
   noStickyHeader,
+  sortBy,
+  setSortBy,
+  sortDirection,
+  setSortDirection,
+  analysisBarSettings,
 }) => {
   const [showMetricFilter, setShowMetricFilter] = useState<boolean>(false);
 
   const { getDimensionById, getExperimentMetricById, metricGroups, ready } =
     useDefinitions();
+  const { metricDefaults } = useOrganizationMetricDefaults();
 
   const _pValueThreshold = usePValueThreshold();
   const pValueThreshold =
@@ -242,18 +257,23 @@ const BreakDownResults: FC<{
           getExperimentMetricById(metricId),
       )
       .filter(isDefined);
-    const sortedFilteredMetrics = sortAndFilterMetricsByTags(
-      metricDefs,
-      metricFilter,
-    );
+    // Only use tag-based sorting when sortBy is "metric-tags"
+    const sortedFilteredMetrics =
+      sortBy === "metric-tags"
+        ? sortAndFilterMetricsByTags(metricDefs, metricFilter)
+        : metricDefs.map((m) => m.id);
 
-    return Array.from(new Set(sortedFilteredMetrics))
+    const tables = Array.from(new Set(sortedFilteredMetrics))
       .map((metricId) => {
         const metric =
           ssrPolyfills?.getExperimentMetricById?.(metricId) ||
           getExperimentMetricById(metricId);
         if (!metric) return;
-        const ret = sortAndFilterMetricsByTags([metric], metricFilter);
+        // Only filter by tags when sortBy is "metric-tags"
+        const ret =
+          sortBy === "metric-tags"
+            ? sortAndFilterMetricsByTags([metric], metricFilter)
+            : [metric.id];
         if (ret.length === 0) return;
 
         const { newMetric, overrideFields } = applyMetricOverrides(
@@ -309,6 +329,27 @@ const BreakDownResults: FC<{
         };
       })
       .filter((table) => table?.metric) as TableDef[];
+
+    // Sort rows within each table by significance or change if sortBy is set
+    if (
+      (sortBy === "significance" || sortBy === "change") &&
+      metricDefaults &&
+      sortDirection
+    ) {
+      const sortOptions = {
+        sortBy,
+        variationFilter:
+          analysisBarSettings?.variationFilter ?? variationFilter ?? [],
+        metricDefaults,
+        sortDirection,
+      };
+      return tables.map((table) => ({
+        ...table,
+        rows: [...table.rows].sort((a, b) => compareRows(a, b, sortOptions)),
+      }));
+    }
+
+    return tables;
   }, [
     results,
     expandedGoals,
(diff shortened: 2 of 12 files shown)


packages/back-end/src/types/Integration.ts
```diff
@@ -53,8 +53,9 @@ export type MetricAggregationType = "pre" | "post" | "noWindow";
 export type FactMetricData = {
   alias: string;
   id: string;
-  metric: ExperimentMetricInterface;
+  metric: FactMetricInterface;
   ratioMetric: boolean;
+  crossFactTableRatio: boolean;
```
ahdriel: I think this is fine for now, but we have a lot of competing `bool`s that maybe we should use enum or a different strategy in the future.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -85,7 +85,9 @@ export function getFactMetricGroup(metric: FactMetricInterface) {
   // Ratio metrics must have the same numerator and denominator fact table to be grouped
   if (isRatioMetric(metric)) {
     if (metric.numerator.factTableId !== metric.denominator?.factTableId) {
-      return "";
+      return metric.denominator?.factTableId
+        ? `${metric.numerator.factTableId} ${metric.denominator.factTableId} (cross-table ratio metrics)`
+        : metric.id;
```
ahdriel, line 95: Question: why are we falling back to the `metric.id` now and before it was an empty string?

Does it even matter?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2502,6 +2502,11 @@ export default abstract class SqlIntegration
     alias: string,
   ): FactMetricData {
     const ratioMetric = isRatioMetric(metric);
+    const crossFactTableRatio =
+      isFactMetric(metric) &&
```
ahdriel: question: According to our types this should be a FactMetricInterface -- is there a chance where a `FactMetricInterface` returns `false` for `isFactMetric`?

I think we can keep it here for additional safety, though, but I am curious about the answer.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2536,7 +2541,7 @@ export default abstract class SqlIntegration
       columnRef: isFactMetric(metric) ? metric.numerator : null,
     });
     const capCoalesceDenominator = this.capCoalesceValue({
-      valueCol: `m.${alias}_denominator`,
+      valueCol: `${crossFactTableRatio ? "m2" : "m"}.${alias}_denominator`,
```
ahdriel: question: What does `m2` or `m` mean here?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2644,17 +2650,87 @@ export default abstract class SqlIntegration
...
+    getCrossFactTableDenominator,
+  }: {
+    dimensionCols: DimensionColumnData[];
+    baseIdType: string;
+    regressionAdjustedMetrics: FactMetricData[];
+    baseMetricTable: string;
+    getCrossFactTableDenominator: boolean;
+  }): string {
+    return `
+          SELECT
```
ahdriel: nit: Should we reduce the indentation here? Or is this because of where this function is called in the bigger SQL statement?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3879,6 +4048,23 @@ export default abstract class SqlIntegration
         ON (c.${baseIdType} = m.${baseIdType})
         `
         : ""
+    }
+              ${
+                someCrossFactTableRatio
+                  ? `
```
ahdriel: nit: another weird indentation level

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: It corresponds to the alias for the CTE that will need to be used in the final analysis step of the query.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Reduced it. It should all get reformatted anyways (except for ClickHouse, where we need a formatter that works).

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Removed it. No chance; this was forgotten after I updated the type.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
lukesonnet: As discussed offline, it creates a separate "group" just for that metric, which is probably unnecessary but this logic will change again in #4593.

packages/back-end/src/types/Integration.ts
lukesonnet: I'm leaving this for now, thanks.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -85,7 +85,9 @@ export function getFactMetricGroup(metric: FactMetricInterface) {
   // Ratio metrics must have the same numerator and denominator fact table to be grouped
   if (isRatioMetric(metric)) {
     if (metric.numerator.factTableId !== metric.denominator?.factTableId) {
-      return "";
+      return metric.denominator?.factTableId
+        ? `${metric.numerator.factTableId} ${metric.denominator.factTableId} (cross-table ratio metrics)`
```
jdorn: Question: If there are 2 metrics with the same 2 fact tables, but one has the value `A/B` and another `B/A`, they will get different group ids here.  Is that intentional?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -838,6 +839,14 @@ export default abstract class SqlIntegration
       "m0",
     );
 
+    const factTable = params.factTableMap.get(
+      metric.numerator?.factTableId || "",
+    );
+    if (!factTable) {
+      throw new Error("Unknown fact table");
```
ahdriel: question: Do we still want to Error out when the `metric.numerator` is undefined?

Currently if undefined we will fallback to `""` which I hope will not match anything 😂

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2644,17 +2658,148 @@ export default abstract class SqlIntegration
...
+                        `m.timestamp >= d.${metric.alias}_preexposure_start AND m.timestamp < d.${metric.alias}_preexposure_end`,
+                        `${metric.alias}_value`,
+                        "NULL",
+                      ),
+                    })} as ${metric.alias}_value`
+                  : ""
+              }
+                ${
+                  // only get denominator if ratio metric AND either the denominator is in the same table as the numerator OR
+                  // getDenominatorOnly is true
```
ahdriel: We should update/remove this comment now that `getDenominatorOnly` is no more. (but `getCrossFactTableDenominator` still is)

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2987,61 +3219,32 @@ export default abstract class SqlIntegration
...
+            : ""
+        }
       ${
         banditDates?.length
           ? this.getBanditStatisticsCTE({
               baseIdType,
               factMetrics: true,
               metricData,
               dimensionCols,
+              someCrossFactTableRatio:
```
ahdriel: nit: Should it be named `has` as the other boolean properties in this function?

`hasCrossFactTableRatioMetric`?

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
lukesonnet: Yes, that is intentional to keep things simple, but I think we support that use case based on the latest changes I made in that PR. I'd have to confirm that numerators from the additional table work well and might imply some slightly more generic naming, which is possible. LMK if you think it's important to explore this before landing.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yes, and that will return `undefined` factTable and throw this error. IDK this is the pattern we use a bunch so I'd rather leave it as is for now.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yes, thank you, will do.

packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
lukesonnet: I will timebox a far more generic solution.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -836,15 +844,10 @@ export default abstract class SqlIntegration
...
       "m0",
     );
 
-    const factTable = params.factTableMap.get(
-      metric.numerator?.factTableId || "",
-    );
-    if (!factTable) {
-      throw new Error("Unknown fact table");
-    }
     // TODO(sql): What happens if ratio metric is cross table?
```
ahdriel: question: Do we need to move/remove this comment?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2508,13 +2511,10 @@ export default abstract class SqlIntegration
       "attributionModel" | "regressionAdjustmentEnabled" | "startDate"
     > & { endDate?: Date },
     activationMetric: ExperimentMetricInterface | null,
+    factTablesWithIndices: { factTable: FactTableInterface; index: number }[],
```
ahdriel, line 2525: nit: For metrics you added the property `index` to the `Metric` object while here we are creating a new object and passing FactTable as a property of it.

Should we adopt this pattern for the Metric as well? Allow us more clearly to separate what is a configured Metric and what is specific for this analysis based on the index.

No strong opinions here, just trying to consolidate in a single approach (as an intermediate step).

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2541,33 +2541,43 @@ export default abstract class SqlIntegration
       metric.cappingSettings.value < 1 &&
       isCappableMetricType(metric);
 
+    const numeratorSourceIndex =
+      factTablesWithIndices.find(
+        (f) => f.factTable.id === metric.numerator?.factTableId,
+      )?.index ?? 0;
```
ahdriel, line 2558: question: Is this fallback to `0` safe? Or should we fallback to `-1` to signify not found (as JS indexOf does)?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2541,33 +2541,43 @@ export default abstract class SqlIntegration
...
+    const denominatorSourceIndex =
+      factTablesWithIndices.find(
+        (f) => f.factTable.id === metric.denominator?.factTableId,
+      )?.index ?? 0;
+    const numeratorAlias = `${numeratorSourceIndex === 0 ? "" : numeratorSourceIndex}`;
```
ahdriel, line 2563: nit: should we include the prefix here for these too? Becomes more clear how we are differentiating them

`=== 0 ? 'm' : `m${numeratorSourceIndex}`

Alternatively, would always including the index simplify things?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2727,63 +2735,50 @@ export default abstract class SqlIntegration
...
+        factTables.add(denominatorFactTable);
       }
     });
 
-    if (!mainFactTable) {
-      throw new Error("No main fact table found");
+    if (factTables.size === 0) {
+      throw new Error("No fact tables found");
+    }
+    if (factTables.size > 2) {
```
ahdriel, line 2786: suggestion: Can we add a `todo` here on what do we need to do to support more than 2?

Only if it is simple enough, otherwise a reason on why we only support 2 can also help.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2811,12 +2806,19 @@ export default abstract class SqlIntegration
...
-      factTables;
+
+    const factTable = factTablesWithIndices[0]?.factTable;
+    const additionalFactTables = factTablesWithIndices
+      .slice(1)
+      .map((f) => f.factTable);
+
+    if (!factTable) {
+      throw new Error("No main fact table found");
+    }
```
ahdriel: We can remove this as `getFactTablesForMetrics` will error out if .size is 0.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2811,12 +2806,19 @@ export default abstract class SqlIntegration
...
       metricsWithIndices,
       factTableMap,
     );
-    const { mainFactTable: factTable, additionalDenominatorFactTable } =
-      factTables;
+
+    const factTable = factTablesWithIndices[0]?.factTable;
+    const additionalFactTables = factTablesWithIndices
+      .slice(1)
+      .map((f) => f.factTable);
```
ahdriel: Suggestion: can we move this closer to the queryName as it is being used only for that now?

Or better we can do something like this and remove these 2 variables:

```
const queryName = `${
    factTablesWithIndices.length === 1
    ? `Fact Table`
    : `Cross-Fact Table`}: ${factTablesWithIndices.map((f) => f.name).join(" & ")}`;
```
(might contain errors as I wrote this in the web, but you get the idea)

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3018,124 +3036,66 @@ export default abstract class SqlIntegration
...
-              SELECT
-                d.variation AS variation
-                ${dimensionCols.map((c) => `, ${c.value} AS ${c.alias}`).join("")}
-                ${banditDates?.length ? `, d.bandit_period AS bandit_period` : ""}
-                , d.${baseIdType} AS ${baseIdType}
-                ${metricData
-                  .filter((data) => data.crossFactTableRatio)
-                  .map((data) =>
-                    data.ratioMetric
+      ${[factTable, ...additionalFactTables]
```
ahdriel: nit: Can we use `factTablesWithIndices` directly here?

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Used this for metrics. It's better, even if it's a bit awkward in a few places. Thanks.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: It's just to likely keep things working for the majority of cases where there is only one fact table and I missed some edge case. I could probably throw an error instead but I feel this will throw just as opaque of an error if it happens to be wrong.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: I don't want to include `m` because it sometimes is used with a `c` when it comes from the cuped/covariate table.

Always including an index *would* simplify things, but then our old query snapshot would break (nbd) and most users would start seeing `__factTable0` and `m0` in their queries, which is a little ugly when it's just one, but also nbd. LMK what you think about the latter and I can fix it if you think it's worth while.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: It's mostly because I haven't tested it and we won't currently use it. I added a comment.

We only support two fact tables because:
* No metric type allows more than 2, so it can only happen right now if you have a fact metric group with 2+ metrics that span multiple tables, and...
* There isn't currently a compelling reason to allow 3+ tables to be in the same set of fact metrics except in cases like the following: you have a main fact table with 4-5 metrics, and 2 cross-fact table ratio metrics where in each of them the main table is crossed with another table or two.

Furthermore, we may not want to ever support scanning 4 or 5 fact tables for the sake of keeping individual queries manageable.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: nice, ty.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Still need to test.

packages/back-end/src/integrations/SqlIntegration.ts
ahdriel: I see, I see. I agree with you, and given the bigger implications let's keep how it is.

ahdriel (approved):

PR #4657, opened 2025-10-02
support additional operators, fix some condition overwrite issues

bryce-fitzsimons:
### Features and Changes

### DIFF
--- a/packages/front-end/services/importing/statsig/transformers/ruleTransformer.ts
+++ b/packages/front-end/services/importing/statsig/transformers/ruleTransformer.ts
@@ -127,7 +127,12 @@ function transformTargetingConditions(
     any: "$in",
     none: "$nin",
     str_contains_any: "$regex",
-    str_contains_none: "$not",
+    str_contains_none: "$regex",
+    str_matches: "$regex",
+    any_case_sensitive: "$in",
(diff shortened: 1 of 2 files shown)


lukesonnet (approved):

packages/front-end/services/importing/statsig/transformers/ruleTransformer.ts
```diff
@@ -163,18 +168,55 @@ function transformTargetingConditions(
...
+          // Reverse case: existing $nin, new $nin
+          conditionObj[gbAttributeName] = {
+            $and: [
+              { [gbAttributeName]: { $nin: existingCondition.$nin } },
+              { [gbAttributeName]: { $in: values } },
+            ],
+          };
+        } else {
+          // No conflict, merge normally
+          conditionObj[gbAttributeName] = { [gbOperator]: values };
```
ahdriel: This will override the condition in place right?

Just checking if this is intentional or if we want to merge it with existing ones by doing `conditionObj[gbAttributeName][gbOperator] = values` for example.

bryce-fitzsimons: good idea

PR #4658, opened 2025-10-02
chore(search): Replace Fuse with MiniSearch

msamper:
### Features and Changes

- Replace Fuse.js with MiniSearch. Fuse.js was giving us unexpected results and it doesn't seem to be actively maintained anymore.
- Remove unnecessary `searchFields` for experiment search. Likely contributing to unexpected results in addition to any bugs that may have been present with Fuse

### Testing

**Repeat for all pages with search**
- [x] Search for a word in an item name and ensure results look as expected
- [x] Search for part of a word and ensure results look as expected

### Screenshots

**Full list of experiments in project**
Screenshot 2025-10-02 at 4 14 41 PM

**Search results**
Screenshot 2025-10-02 at 4 14 10 PM
Screenshot 2025-10-02 at 4 15 56 PM


packages/front-end/services/search.tsx
```diff
@@ -116,21 +116,33 @@ export function useSearch<T>({
...
+    );
+    const fields = Object.keys(keys);
+    const storeFields = Object.keys(items[0] || {});
+
+    const miniSearchInstance = new MiniSearch({
+      fields,
+      storeFields,
+      searchOptions: {
+        boost: keys,
+        fuzzy: 0.2,
```
msamper: Probably need to tweak this

packages/front-end/services/search.tsx
```diff
@@ -141,8 +153,9 @@ export function useSearch<T>({
 
     let filtered = items;
     if (searchTerm.length > 0) {
-      filtered = fuse.search(searchTerm).map((item) => item.item);
+      filtered = miniSearch.search(searchTerm) as T[];
```
msamper: Filter out search result specific fields? I.e. `match` and `score`
https://lucaong.github.io/minisearch/types/MiniSearch.SearchResult.html

packages/front-end/services/search.tsx
```diff
@@ -116,21 +116,33 @@ export function useSearch<T>({
...
+  const miniSearch = useMemo(() => {
+    const keys: Record<string, number> = Object.fromEntries(
+      searchFields.map((f) => {
+        const [key, weight] = (f as string).split("^");
+        const weightNum = weight ? parseFloat(weight) : 1;
+        return [key, weightNum];
+      }),
+    );
+    const fields = Object.keys(keys);
+    const storeFields = Object.keys(items[0] || {});
```
msamper: Feels unsafe and should find a better way to do this. MiniSearch requires you specify the fields you want to be returned with each search result otherwise it defaults to just the id

msamper: Could consider keeping just the id which is the default and then filtering from `items` if the `id` is in the results. Would be ideal to just get the fields we need here though

packages/front-end/services/experiments.ts
```diff
@@ -365,20 +365,7 @@ export function useExperimentSearch({
...
-      "description",
-      "tags",
-      "status",
-      "ownerName",
-      "metricNames",
-      "results",
-      "analysis",
-      "isWatched",
-    ],
+    searchFields: ["name^3", "trackingKey^2", "hypothesis^2", "description"],
```
mknowlton89, line 368: Are we scaling down the `searchFields` options given the new filtering options on the experiment page?

packages/front-end/services/search.tsx
jdorn: Yes, I'd recommend creating a Map of id to item in this memo. Then returning

```ts
ids.map(id=>itemsMap.get(id))
```

Will be safer and probably more performant.

packages/front-end/services/experiments.ts
msamper: Yes exactly and there were also some search fields like `analysis` that were causing unexpected search results

packages/front-end/components/Experiment/ImportExperimentList.tsx
```diff
@@ -56,6 +56,7 @@ const ImportExperimentList: FC<{
       exposureQueryName: item.exposureQueryId
         ? getExposureQuery(datasource?.settings, item.exposureQueryId)?.name
         : "experiments",
+      id: item.trackingKey,
```
msamper, line 59: miniSearch requires an id and these `pastExperiments` didn't have an id property so I'm adding the tracking key as an ID (assuming it'll be unique)

packages/front-end/components/GetStarted/NeedingAttention.tsx
```diff
@@ -241,7 +255,7 @@ const NeedingAttention = (): React.ReactElement | null => {
   const {
     items: featureFlagsNeedingAttention,
     SortableTH: SortableTHFeatureFlags,
-  } = useSearch({
+  } = useSearch<ComputedFeaturesAndRevisions>({
```
msamper, line 258: Was getting some type errors here before adding this type

PR #4669, opened 2025-10-06
Fix assorted dashboard errors

Kevin-Chant:
### Features and Changes

A few errors in dashboards came through Sentry recently; this PR attempts to address the most likely causes

### Testing

Manually tested that the frontend still works and nothing was made worse

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -3466,7 +3466,9 @@ export async function updateExperimentDashboards({
     .filter((block) => blockHasFieldOfType(block, "snapshotId", isString))
     .filter((block) => block.experimentId === experiment.id);
   const blocksNeedingSnapshot = blocksWithSnapshots.filter(
-    (block) => !snapshotSatisfiesBlock(mainSnapshot, block),
+    (block) =>
+      block.snapshotId.length > 0 &&
+      !snapshotSatisfiesBlock(mainSnapshot, block),
   );
   const previousSnapshotIds = [
     ...new Set(blocksNeedingSnapshot.map((block) => block.snapshotId)),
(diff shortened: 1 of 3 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -3466,7 +3466,9 @@ export async function updateExperimentDashboards({
     .filter((block) => blockHasFieldOfType(block, "snapshotId", isString))
     .filter((block) => block.experimentId === experiment.id);
   const blocksNeedingSnapshot = blocksWithSnapshots.filter(
-    (block) => !snapshotSatisfiesBlock(mainSnapshot, block),
+    (block) =>
+      block.snapshotId.length > 0 &&
+      !snapshotSatisfiesBlock(mainSnapshot, block),
```
Kevin-Chant, line 3471: The below error was triggered a couple of times, and one possible cause is updating blocks while the dashboard is being edited (so `snapshotId` is still an empty string)
> Error updating dashboard results, could not find snapshot

packages/back-end/src/services/experiments.ts
```diff
@@ -3483,10 +3485,14 @@ export async function updateExperimentDashboards({
...
     const blockSnapshot = previousSnapshotMap.get(block.snapshotId);
-    if (!blockSnapshot || !blockSnapshot.analyses[0])
+    if (!blockSnapshot)
       throw new Error(
         "Error updating dashboard results, could not find snapshot",
       );
+    if (!blockSnapshot.analyses[0])
+      throw new Error(
+        "Error updating dashboard results, referenced snapshot missing analysis",
+      );
```
Kevin-Chant, line 3495: If that wasn't the root cause, I split out the error message so we can tell which actual case is causing the issue

packages/front-end/enterprise/components/Dashboards/DashboardSnapshotProvider.tsx
```diff
@@ -108,9 +108,9 @@ export default function DashboardSnapshotProvider({
     const allQueries = allSnapshots.flatMap(
       (snapshot) => snapshot.queries || [],
     );
-    const snapshotError = snapshotsMap
-      .values()
-      .find((snapshot) => snapshot.error)?.error;
+    const snapshotError = allSnapshots.find(
+      (snapshot) => snapshot.error,
+    )?.error;
```
Kevin-Chant, line 113: I don't see why this could cause an error but seemingly this line had `find is not a function` happening many times

packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -195,7 +195,7 @@ export default function DashboardsTab({
       dashboard={dashboard}
       mutateDefinitions={mutateDashboards}
     >
-      {isEditing && dashboard ? (
+      {canEdit && isEditing && dashboard ? (
```
Kevin-Chant, line 198: There's a lot of permissions errors as well, and while I don't see how a user can get into the editing state when they dont have permission, it seems better to just prevent rendering it in the first place

PR #4671, opened 2025-10-06
Convert migrate-encryption-key to mjs

tzjames:
### Features and Changes

[One PR](https://github.com/growthbook/growthbook/pull/4602) tried to fix it, but since it imports other files it fails. 
[Then another PR](https://github.com/growthbook/growthbook/pull/4628) tried using ts-node.  That works except our docker image we build doesn't have ts-node.
So this PR changes the file to mjs and has it read from dist directly. 

### Testing
`yarn migrate-encryption-key dev`
See no import errors thrown.

### DIFF
--- a/packages/back-end/src/scripts/migrate-encryption-key.mjs
+++ b/packages/back-end/src/scripts/migrate-encryption-key.mjs
@@ -1,0 +1,81 @@
+import pkg from "crypto-js";
+const { AES, enc } = pkg;
+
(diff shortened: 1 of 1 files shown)


ahdriel (approved):
Nice! Thanks for fixing it.

I tested both locally and within a docker container and it worked.

One thing I noticed is that we are initializing the MongoDB jobs, so maybe we want to consider adding `CRON_ENABLED=false` for this script?

PR #4673, opened 2025-10-06
feat(sql): Use BigQuery query result types for Fact Tables

ahdriel:
### Features and Changes

This changes our logic for BigQuery to use the information returned by the Query to determined the types of columns on our end, instead of always testing against the returned values to infer what type it is.

### DIFF
--- a/packages/back-end/src/jobs/refreshFactTableColumns.ts
+++ b/packages/back-end/src/jobs/refreshFactTableColumns.ts
@@ -138,7 +138,48 @@ export async function runRefreshColumnsQuery(
 
   const typeMap = new Map<string, FactTableColumnType>();
   const jsonMap = new Map<string, JSONColumnFields>();
-  determineColumnTypes(result.results).forEach((col) => {
+
+  result.columns?.forEach((col) => {
+    // If the underlying SQL engine returned the datatype, use it
+    if (col.dataType !== undefined) {
+      // For JSON, only return if we have the field information, otherwise skip
+      // so we can infer from the returned data
+      if (
+        col.dataType === "json" &&
+        col.fields !== undefined &&
+        col.fields.length > 0
+      ) {
+        typeMap.set(col.name, "json");
+        jsonMap.set(
+          col.name,
+          col.fields.reduce(
+            (acc, field) => ({
+              ...acc,
+              [field.name]: {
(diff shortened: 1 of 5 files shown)


packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -138,7 +138,48 @@ export async function runRefreshColumnsQuery(
...
+  const resultWithoutKnownColumns = result.results?.[0];
+  if (resultWithoutKnownColumns) {
+    for (const col of Object.keys(resultWithoutKnownColumns)) {
+      if (typeMap.has(col)) {
+        delete resultWithoutKnownColumns[col];
+      }
+    }
+  }
+
+  determineColumnTypes([resultWithoutKnownColumns]).forEach((col) => {
```
jdorn: I think this is a regression. Before, we would pass in all result rows, but now we are only passing in the first. That means if the first row has a null value, this will fail to infer whereas it previously would get the first non-null value from all the rows.

ahdriel: Good catch! I missed that we iterate over all `rows` after getting the first one in `determineColumnTypes`.

Fixed.

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -138,7 +138,38 @@ export async function runRefreshColumnsQuery(
...
+          ),
+        );
+      } else if (col.dataType !== "json") {
+        typeMap.set(col.name, col.dataType);
+      }
+    }
+  });
+
+  const columnsToSkip = new Set(typeMap.keys());
+  determineColumnTypes(result.results, columnsToSkip).forEach((col) => {
```
jdorn: There is another small regression.  On main, selecting a JSON string would be inferred as type `json`, but here it would just be set to type `string`.

Instead of `columnsToSkip`, maybe the whole existing typeMap should be passed in.  Then, `determineColumnTypes` can decide if it needs to further refine any of the column types based on the sample row values.

ahdriel: Good catch! I updated the code to handle this scenario.

jdorn (approved):

PR #4675, opened 2025-10-07
Sort by significance results table column filter

bryce-fitzsimons:
### Features and Changes

### DIFF
--- a/packages/front-end/components/Experiment/CompactResults.tsx
+++ b/packages/front-end/components/Experiment/CompactResults.tsx
@@ -36,6 +36,7 @@ import { useDefinitions } from "@/services/DefinitionsContext";
 import {
   applyMetricOverrides,
   ExperimentTableRow,
+  compareRowsBySignificance,
 } from "@/services/experiments";
 import { QueryStatusData } from "@/components/Queries/RunQueriesButton";
 import {
@@ -48,6 +49,7 @@ import MetricTooltipBody from "@/components/Metrics/MetricTooltipBody";
 import { SSRPolyfills } from "@/hooks/useSSRPolyfills";
 import { useUser } from "@/services/UserContext";
 import { AppFeatures } from "@/types/app-features";
+import { useOrganizationMetricDefaults } from "@/hooks/useOrganizationMetricDefaults";
 import DataQualityWarning from "./DataQualityWarning";
 import ResultsTable from "./ResultsTable";
 import MultipleExposureWarning from "./MultipleExposureWarning";
@@ -105,6 +107,11 @@ const CompactResults: FC<{
       levels: string[];
     }>;
   }>;
+  sortBy?: "metric-tags" | "significance" | null;
+  setSortBy?: (s: "metric-tags" | "significance" | null) => void;
+  analysisBarSettings?: {
+    variationFilter: number[];
+  };
 }> = ({
   experimentId,
   editMetrics,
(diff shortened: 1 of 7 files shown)


packages/front-end/components/Experiment/CompactResults.tsx
```diff
@@ -444,6 +494,10 @@ const CompactResults: FC<{
     getFactMetricLevels,
     shouldShowMetricSlices,
     customMetricSlices,
+    sortBy,
+    analysisBarSettings?.variationFilter,
+    metricDefaults,
+    variationFilter,
```
jdorn, line 510: Did you make sure these are all stable and won't cause the `useMemo` to run on every render?

packages/front-end/components/Experiment/ResultsTable.tsx
```diff
@@ -173,11 +180,43 @@ export default function ResultsTable({
...
+        sortBy === "significance"
+          ? "Sorted by significance"
+          : "Sort by significance"
+      }
+    >
+      <a
+        role="button"
+        className={sortBy === "significance" ? "link-purple" : "text-muted"}
+        onClick={() =>
+          setSortBy(sortBy === "significance" ? null : "significance")
```
jdorn: Is there value in reverse sorting (i.e. click once to sort in normal order, click again to sort in reverse order, click a 3rd time to unset the sort)?

Thinking about wanting to quickly look at any big losers among a list of hundreds of metrics.

packages/front-end/services/experiments.ts
```diff
@@ -45,6 +45,60 @@ import { useExperimentStatusIndicator } from "@/hooks/useExperimentStatusIndicat
...
+    aFiltered.some(
+      (v) => v && hasEnoughData(aBaseline, v, a?.metric, metricDefaults),
+    );
+  const bHasEnoughData =
+    bBaseline &&
+    bFiltered.some(
+      (v) => v && hasEnoughData(bBaseline, v, b?.metric, metricDefaults),
+    );
+
+  if (!aHasData || !aHasEnoughData || !bHasData || !bHasEnoughData) return 0;
```
jdorn: This logic doesn't feel quite right.  The variations without enough data are still being considered below.  Should we filter those out instead?  I think that would also avoid the need for this `return 0` here since the length checks below should suffice.

packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -149,6 +149,42 @@ export default function TabbedPage({
       filterByTag: false,
     },
   );
+  const [sortBy, setSortBy] = useLocalStorage<
+    "metric-tags" | "significance" | null
```
jdorn: We can do this later, but there might also be value in sorting by the "Percent Changed" column.

packages/front-end/components/Experiment/CompactResults.tsx
bryce-fitzsimons: they are stable; no run-away memoization. They're all from state variables and not render-derived props.

packages/front-end/services/experiments.ts
bryce-fitzsimons: I think `return -1` is correct, right? We want the not-enough-data/no-data rows to push to the bottom since they aren't actionable

bryce-fitzsimons: i redid this, should be better now I think

PR #4678, opened 2025-10-07
Upgrade zod

tzjames:
### Features and Changes

When trying to use Vercel's ai, there was compatibility issues with zod.  So attempting to upgrade zod here.

This is a second attempt to upgrade zod.  It has the following differences with [the earlier attempt here](https://github.com/growthbook/growthbook/pull/4176):

1. I was able to get the gen script in docs working by using a fork of the current npm package and then patching it.   It feels a bit kludgy, but this is not very critical seeing that the other PR left it in a broken state.
2. The other PR imports from zod/v4. This was needed for transition period when zod 4.0.0 wasn't yet published. Many of the imports where already `import {z} from "zod"` so nothing needed to be done and this accounts for the large number of additional files changed in the other PR and not in this one.
3. The other PR had a change in jest that doesn't seem necessary for tests to work.
4. There are a few differences in BaseModel, but the changes in this PR are perhaps simpler.
5. There are a few differences in generate-openapi.mjs, but again this one might be simpler.  `yarn generate-api-types` works and the generated file doesn't change from before the PR.
6. Since modifying the packages, I had to make sure the `yarn setup` script also ran `yarn` in docs so that it loads.  I made it run in parallel so save some 10s or so of build time.

### Testing

`yarn type-check`
`yarn lint`
in backend: `yarn generate-api-types` -> see that generated/spec.yaml doesn't change.
in docs: `yarn gen` -> See that docs/src/partials/event-webhook/_event-webhook-list.md has only more informative changes.

### DIFF
--- a/packages/back-end/test/models/BaseModel.test.ts
+++ b/packages/back-end/test/models/BaseModel.test.ts
@@ -396,10 +396,10 @@ describe("BaseModel", () => {
     expect(model.create({ id: "aabb" })).rejects.toMatchObject({
       issues: [
         {
-          code: "invalid_type",
           expected: "string",
+          code: "invalid_type",
           path: ["name"],
-          message: "Required",
+          message: "Invalid input: expected string, received undefined",
         },
       ],
     });


docs/src/partials/event-webhook/_event-webhook-list.md
```diff
@@ -26,295 +26,7 @@ Triggered when a feature is created
   <summary>Payload</summary>
 
 ```typescript
-{
```
ahdriel, line 29: issue: We should fix this, it should not be changing to `any`.

tzjames: Good catch.  We run `yarn doc-gen` as a precommit hook.  I think the wrong version of zod was running during that gen.  I've updated zod in the docs package as well and that command seem to be working now.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -558,7 +534,7 @@ export abstract class BaseModel<
       forceCanUpdate?: boolean;
     },
   ) {
-    updates = this.updateValidator.parse(updates);
+    updates = this.updateValidator.parse(updates) as UpdateProps<z.infer<T>>;
```
Kevin-Chant: I can't tell what exactly caused this typecast to become necessary but it does worry me slightly that the result of `updateValidator.parse` is compiling as `Record<string, never>`. Did you get to the bottom of what happened here?

packages/back-end/src/util/handler.ts
```diff
@@ -79,7 +79,7 @@ export function createApiRequestHandler<
           if (!validated.success) {
             allErrors.push(`Request params: ` + validated.errors.join(", "));
           } else {
-            req.params = validated.data;
+            req.params = validated.data as z.output<ParamsSchema>;
```
Kevin-Chant: Similar question here

packages/back-end/src/models/BaseModel.ts
tzjames: I can't say I really understand UpdateZodObject as it was. Is it really needed?  We could just set it directly as 
```
export type UpdateZodObject<T extends BaseSchema> = z.ZodType<
  UpdateProps<z.infer<T>>
>;
```
And let typescript do the work.  But I'm not sure if we are losing safety by doing that.

packages/back-end/src/util/handler.ts
tzjames: this was not needed.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -54,12 +54,8 @@ export type CreateRawShape<T extends z.ZodRawShape> = {
...
 export type CreateZodObject<T> =
-  T extends z.ZodObject<
-    infer RawShape,
-    infer UnknownKeysParam,
-    infer ZodTypeAny
-  >
-    ? z.ZodObject<CreateRawShape<RawShape>, UnknownKeysParam, ZodTypeAny>
+  T extends z.ZodObject<infer RawShape, infer UnknownKeysParam>
+    ? z.ZodObject<CreateRawShape<RawShape>, UnknownKeysParam>
     : never;
```
Kevin-Chant: I'd like @jdorn's input here because he might know this better, but it seems that we can simplify `CreateZodObject` in the same way as `UpdateZodObject`. `type-check` passes with this change but I'm not sure if there's a reason not to apply it

```suggestion
export type CreateZodObject<T extends BaseSchema> = z.ZodType<
  CreateProps<z.infer<T>>
>;
```

tzjames: He seemed to be ok with it in our call yesterday.

PR #4679, opened 2025-10-07
Create feature rules in multiple environments at once

jdorn:
### Features and Changes

When adding a rule to a feature, it's common to want to add the same rule to multiple environments at once.  This PR adds a new field when creating a rule where you can select which environments to create it in.

Screenshot from 2025-10-07 18-08-41

By default, only the current environment tab is checked (just like how it works today).  However, you can change this default behavior in General Settings on the Features tab:

Screenshot from 2025-10-07 18-11-43

### DIFF
--- a/packages/back-end/src/controllers/features.ts
+++ b/packages/back-end/src/controllers/features.ts
@@ -1250,7 +1250,11 @@ export async function postFeatureRule(
   const context = getContextFromReq(req);
   const { org } = context;
   const { id, version } = req.params;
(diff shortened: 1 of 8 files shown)


bryce-fitzsimons (approved):
Seems reasonable.

I wonder if we should consider doing something for editing and removing rules as well. When created across environments, use the same rule id. Then when doing other actions (update/remove), if they have the same rule.id, have an option to perform that action across environments.

PR #4683, opened 2025-10-08
Api Rate Limit Per Org

tzjames:
### Features and Changes
Have a unique api limit that can be set on a per org basis.

### Testing

Create a personal API token <PAT>
```
for i in {1..65}; do
  echo "Request $i:"
  curl -H "Authorization: Bearer <PAT>" \
       -w "HTTP Status: %{http_code}, Time: %{time_total}s\n" \
       -s \
       http://localhost:3100/api/v1/
  echo "---"
done
```
See that the first 60 succeed, then the last 5 return 429.

Then set `apiRateLimit: 5` on the org in Mongo
Do the same command and see the first 5 succeed and then next 60 fail.

### DIFF
--- a/packages/back-end/src/api/api.router.ts
+++ b/packages/back-end/src/api/api.router.ts
@@ -7,7 +7,7 @@ import * as Sentry from "@sentry/node";
 import authenticateApiRequestMiddleware from "back-end/src/middleware/authenticateApiRequestMiddleware";
 import { getBuild } from "back-end/src/util/build";
 import { ApiRequestLocals } from "back-end/types/api";
-import { IS_CLOUD, SENTRY_DSN } from "../util/secrets";
+import { SENTRY_DSN } from "../util/secrets";
 import featuresRouter from "./features/features.router";
 import experimentsRouter from "./experiments/experiments.router";
 import snapshotsRouter from "./snapshots/snapshots.router";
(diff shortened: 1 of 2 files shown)


packages/back-end/src/api/api.router.ts
```diff
@@ -78,18 +78,28 @@ if (SENTRY_DSN) {
...
     legacyHeaders: false,
     keyGenerator: (req: Request & ApiRequestLocals) => req.apiKey,
-    message: {
-      message: `Too many requests, limit to ${overallRateLimit} per minute`,
+    message: (req: Request & ApiRequestLocals) => {
+      const rateLimitForWarning = IS_CLOUD
+        ? 60
+        : req.context.org.apiRateLimit ||
+          Number(process.env.API_RATE_LIMIT_MAX) ||
+          60;
```
jdorn: Is this logic right?  Won't this always say `60` on cloud, even if the org has a lower limit?

tzjames: yes.  But since there are multiple servers we can't get an accurate number here without polling how many servers there are and then it is all probabilistic depending upon which server the load balancer forwards them to.  So the number doesn't mean much, but the probability for throwing a 429 for the default of 16 with 8 servers is less than 1% when doing 60 req/minute.  Perhaps we should just remove the number we are warning about completely.

PR #4684, opened 2025-10-08
chore(sample-data): Use fact metrics for sample data source

### LINKED ISSUE #4578: [Chore] Change demo metrics to fact metrics
### Description of Feature

Our demo metrics in demo-datasource-project.controller.ts use legacy metrics, but instead we should use fact tables (orders and page views) and fact metrics.

As part of this we should: (1) use a proper "retention" metric type, (2) remove other conversion windows, and (3) create a proper order count metric with the same naming system as the other purchases and a proper ratio metric.

msamper:
### Features and Changes
Replaces the sample data experiment with a 2-armed experiment and uses fact metrics instead of legacy metrics.

- Closes #4578 

**To-Do**
- [x] Place new sample data route behind a feature flag tied to an experiment
- [x] Fix demo feature not appearing within the sample datasource project
- [x] Enable CUPED for sample data experiment
- [x] Use `gbdemo-add-to-cart-cta` tracking key
- [x] Insert sample data for `gbdemo-add-to-cart-cta` into `experiment_viewed`

### Testing

- [x] Make sure CUPED is disabled on the sample data experiment for free accounts


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -284,6 +285,30 @@ export class FactMetricModel extends BaseClass {
     }
   }
 
+  public async deleteAllFactMetricsForAProject(projectId: string) {
+    const factMetrics = await this._find({
+      projects: [projectId],
+    });
+    await Promise.all(
+      factMetrics.map((factMetric) => this.deleteById(factMetric.id)),
+    );
```
jdorn: Only delete if this is the only project for a fact metric.

Don't use Promise.all directly, use `promiseAllChunks` instead.  Alternatively, add a `deleteMultiple` function to base model.  Also, don't use `ById` methods when you already have the full object

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -284,6 +285,30 @@ export class FactMetricModel extends BaseClass {
...
+  public async removeProjectFromFactMetrics(projectId: string) {
+    // Find all fact metrics for the project
+    const factMetrics = await this._find({
+      projects: [projectId],
+    });
+    // Update each fact metric to remove the project
+    await Promise.all(
+      factMetrics.map((factMetric) =>
+        this.updateById(factMetric.id, {
+          projects: factMetric.projects.filter((p) => p !== projectId),
```
jdorn: This will not work since empty array = all projects

packages/front-end/hooks/useDemoDataSourceProject.ts
```diff
@@ -44,7 +48,9 @@ export const useDemoDataSourceProject = (): UseDemoDataSourceProject => {
 
   const currentProjectIsDemo = currentProjectId === demoProjectId;
 
-  const demoFeatureId = getDemoDataSourceFeatureId();
+  const demoFeatureId = useNewSampleData
+    ? "gbdemo-add-to-cart-cta"
+    : getDemoDataSourceFeatureId();
```
jdorn: Keep feature flag id the same as before

packages/front-end/pages/features/index.tsx
```diff
@@ -433,12 +430,7 @@ export default function FeaturesPage() {
   // If "All Projects" is selected and some experiments are in a project, show the project column
   const showProjectColumn = !project && allFeatures.some((f) => f.project);
 
-  // Ignore the demo datasource
-  const hasFeatures = allFeatures.some(
-    (f) =>
-      f.project !==
-      getDemoDatasourceProjectIdForOrganization(organization.id || ""),
-  );
+  const hasFeatures = allFeatures.length > 0;
```
jdorn: Revert this

packages/back-end/src/routers/project/project.controller.ts
```diff
@@ -215,6 +221,25 @@ export const deleteProject = async (
...
+    } catch (e) {
+      return res.json({
+        status: 403,
+        message: "Failed to delete fact tables",
+      });
+    }
+  } else {
+    await removeProjectFromFactTables(id, context);
+    await context.models.factMetrics.removeProjectFromFactMetrics(id);
+  }
```
jdorn, line 235: Remove this else block and delete these methods. They are not needed

packages/back-end/src/models/FactMetricModel.ts
msamper: Double checked and the `_find()` here is only matching when the projects array matches `[projectId]` exactly

PR #4686, opened 2025-10-08
Add JIT patches for missing banditEvents

bryce-fitzsimons:
Not sure why, but sometimes `phases[0].banditEvents` are missing entirely from bandit experiments. This should add some safety when running analyses and writing back to the model.

### DIFF
--- a/packages/front-end/components/Experiment/CompactResults.tsx
+++ b/packages/front-end/components/Experiment/CompactResults.tsx
@@ -14,6 +14,10 @@ import {
   PValueCorrection,
   StatsEngine,
 } from "back-end/types/stats";
+import {
+  FactTableInterface,
+  FactMetricInterface,
+} from "back-end/types/fact-table";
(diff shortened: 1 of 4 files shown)


packages/back-end/src/services/experiments.ts
```diff
@@ -641,6 +641,25 @@ export function getSnapshotSettings({
     )
     .filter(isDefined);
 
+  // JIT initialize banditEvents in memory if missing
+  if (
+    experiment.type === "multi-armed-bandit" &&
+    phase &&
+    (!phase.banditEvents || phase.banditEvents.length === 0)
+  ) {
```
ahdriel, line 649: suggestion: Can we add a `logger.warn` here (and maybe also in the `resetSettings` below)?

It is something I am interested in tracking even if the underlying fix is more complex.

bryce-fitzsimons: yep, sounds good

PR #4688, opened 2025-10-08
Add an optional "default env" org setting for feature editing

bryce-fitzsimons:
Org setting that lets you optionally specify a default environment when on a feature detail page, rather than remembering the last used environment.

image

### DIFF
--- a/packages/back-end/types/organization.d.ts
+++ b/packages/back-end/types/organization.d.ts
@@ -244,7 +244,7 @@ export interface OrganizationSettings {
   experimentPageMarkdown?: string;
   metricListMarkdown?: string;
   metricPageMarkdown?: string;
-  preferredEnvironment?: string; // empty means "remember previous environment"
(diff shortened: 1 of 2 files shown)


packages/back-end/types/organization.d.ts
```diff
@@ -244,6 +244,7 @@ export interface OrganizationSettings {
   experimentPageMarkdown?: string;
   metricListMarkdown?: string;
   metricPageMarkdown?: string;
+  preferredEnvironment?: string; // empty means "remember previous environment"
```
ahdriel: In some places we are using `null` and others we are using `undefined`. And would we allow an empty string here?

Is there a way to consolidate or do they mean different things?

bryce-fitzsimons: decided to use null as the "remember previous environment" option. undefined is for legacy support

PR #4696, opened 2025-10-09
Improve slice UI, auto-select levels, show when levels are missing, etc

bryce-fitzsimons:
### Features and Changes

- try to auto-populate slice levels when configuring column
- let user bring in latest top values when clicking refresh
- warn in the UI when slice levels are empty

image

image

image

### DIFF
--- a/packages/front-end/components/FactTables/ColumnModal.tsx
+++ b/packages/front-end/components/FactTables/ColumnModal.tsx
@@ -501,6 +501,28 @@ export default function ColumnModal({ existing, factTable, close }: Props) {
         placeholder={form.watch("column")}
       />
 
+      {showDescription ? (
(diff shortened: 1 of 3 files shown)


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5244,7 +5244,7 @@ WITH
       ${column.column} AS value,
       COUNT(*) AS count
     FROM __factTable
-    WHERE timestamp >= ${this.toTimestamp(start)}
+    -- WHERE timestamp >= ${this.toTimestamp(start)}
```
bryce-fitzsimons:
```suggestion
    WHERE timestamp >= ${this.toTimestamp(start)}
```

PR #4701, opened 2025-10-10
feat(uploads): closes #4700, AWS role assumption implementation

### LINKED ISSUE #4700: [Feature] Support AWS role assumption for S3
### Description of Feature

Many organizations enforce AWS access restrictions by requiring role assumption for enhanced security. To support this, the application should be updated to obtain credentials via an STS call when accessing AWS resources. This feature will improve compatibility with corporate security policies and ensure secure access management.

nodirnasirov:
### Features and Changes

Closes #4700 
S3 buckets are kept private in many enterprises, this PR implements STS role assumption to allow AWS S3 bucket access.

- Closes #4700

### DIFF
--- a/packages/back-end/src/services/files.ts
+++ b/packages/back-end/src/services/files.ts
@@ -13,34 +13,30 @@ import {
 } from "back-end/src/util/secrets";
 
 let s3: AWS.S3;
-function getS3(): AWS.S3 {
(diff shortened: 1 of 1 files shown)


tzjames:
Thanks for making this PR.  In my testing the sts call needed to be awaited, otherwise s3 is returned immediately and it is not ready yet.

PR #4715, opened 2025-10-15
metric slice support in dashboards

bryce-fitzsimons:
### Features and Changes

- Add metric slice support to dashboards
- Allow left-side pinning and expansion AND editor-panel pinning
- Standardize experiment row generation across a bunch of different components
- Fix bugs

image
image

### DIFF
--- a/docs/docs/self-host/env.mdx
+++ b/docs/docs/self-host/env.mdx
@@ -131,6 +131,7 @@ Store uploads in an AWS S3 bucket.
 - **S3_DOMAIN** (defaults to `https://${S3_BUCKET}.s3.amazonaws.com/`)
 - **AWS_ACCESS_KEY_ID** (not required when deployed to AWS with an instance role)
 - **AWS_SECRET_ACCESS_KEY** (not required when deployed to AWS with an instance role)
+- **AWS_ASSUME_ROLE** (optional role to assume, that has `s3:PutObject` and `s3:GetObject` to the bucket)
 
 ### google-cloud
 
--- a/docs/docs/self-host/production.mdx
+++ b/docs/docs/self-host/production.mdx
@@ -30,6 +30,12 @@ If you change the `ENCRYPTION_KEY`, you will need to migrate any existing data s
 docker compose run growthbook yarn migrate-encryption-key OLD_KEY
 ```
 
+### Data Sources
+
+In general, GrowthBook only requires read access to your data warehouse. The one exception is if Pipeline Mode is enabled, in which case GrowthBook will need write access to a specific schema that you specify in order to create temporary tables for experiment analysis.
+
+We do our best to protect against SQL injection attacks and destructive queries, but because a large part of GrowthBook's functionality is running arbitrary user-defined SQL, there is always a risk. For this reason, we recommend following the principle of least privilege and creating a dedicated database user for GrowthBook with only the necessary permissions. That means only read access to your data, plus write access to a dedicated schema created specifically for GrowthBook (if Pipeline Mode is enabled).
+
 ### Networking
 
 The most secure thing you can do is to put your entire GrowthBook deployment behind a firewall and/or corporate VPN. This way, you can restrict access to only those who need to sign into the GrowthBook platform. This can drastically reduce the attack vector and serves as an important first layer of security.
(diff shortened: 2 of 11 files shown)


packages/back-end/src/enterprise/validators/dashboard-block.ts
```diff
@@ -121,16 +127,19 @@ const experimentMetricBlockInterface = baseBlockInterface
     ),
     snapshotId: z.string(),
   })
+  .merge(metricSliceSettingsInterface)
```
Kevin-Chant, line 133: Just curious, why have this as a `merge` rather than directly putting the fields in the object? Is it just for readability that they're connected?

packages/back-end/src/enterprise/validators/dashboard-block.ts
```diff
@@ -14,6 +14,12 @@ const baseBlockInterface = z
   })
   .strict();
 
+export const pinSources = ["experiment", "custom", "none"] as const;
```
Kevin-Chant: Nit: Import this from shared rather than redeclaring

packages/back-end/src/enterprise/validators/dashboard-block.ts
bryce-fitzsimons: yeah mostly for readability so its clear why they're in there, and so it's easy to compose on future blocks.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentTimeSeriesBlock.tsx
```diff
@@ -1,71 +1,369 @@
...
+    const source = pinSource || "experiment"; // Default to "experiment" if undefined
+    if (source === "experiment") {
+      return experiment.pinnedMetricSlices;
+    } else if (source === "custom") {
+      return pinnedMetricSlices;
+    } else {
+      // source === "none"
+      return undefined;
+    }
+  }, [pinSource, experiment.pinnedMetricSlices, pinnedMetricSlices]);
```
Kevin-Chant: Is it worth pulling this and some of the other repeated logic out into helpers/hooks for dashboards? I like the move towards `useExperimentTableRows` and the more we can standardize the better it protects from code drift over time

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -188,6 +196,190 @@ export default function EditSingleBlock({
...
+        (group) =>
+          group.label === "Secondary Metrics" ||
+          group.label === "Metric Groups",
+      );
+    }
+    if (selector === "experiment-guardrail") {
+      return metricOptions.filter(
+        (group) =>
+          group.label === "Guardrail Metrics" ||
+          group.label === "Metric Groups",
```
Kevin-Chant: Seems like these shouldn't allow for metric groups directly but rather unroll any relevant metric groups that are included by the filter.

This screenshot is for an experiment with one goal metric and a metric group for the secondary metrics. The metric group is selectable for pinning despite using "All Goal Metrics", and once selected it doesn't handle the group properly


image

Kevin-Chant: @bryce-fitzsimons I just pushed a change to unroll all the groups; mind taking a look when you're online to see if you agree with the approach?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentTimeSeriesBlock.tsx
Kevin-Chant: @bryce-fitzsimons I also took a stab at this; would appreciate your eyes here too

bryce-fitzsimons: Agreed. I see your new hook and agree with the approach.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
bryce-fitzsimons: good catch, i didn't realize the pin expander widget wasn't working with groups

bryce-fitzsimons: i agree with DRY solutions like yours; custom hook is a good idea

Kevin-Chant (commented):
:white_check_mark: approved for dashboards code, but still needs review for the table row generation and experiment results components

PR #4716, opened 2025-10-16
Product Analytics Dashboards

Kevin-Chant:
### Known Bugs
- ~~When adding a Metric block, when configuring the block, we render a "Loading" state on the dashboard, and that state won't change until the user clicks "Save & Close" - it looks like we'll provide a live preview, but in my testing, it doesn't actually work that way.~~
- ~~With the `SqlExplorer` if you go to add a visualization to an existing `saved-query` an error is thrown unless you re-run the query (This should be solved in a separate PR, but just calling it out).~~

### DIFF
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
@@ -522,8 +522,6 @@ export default function EditSingleBlock({
     });
   };
 
-  console.log("block", block);
-
   return (
     <>
       {showSqlExplorerModal && (


packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -51,63 +56,143 @@ export class DashboardModel extends BaseClass {
...
+        enableAutoUpdates: true,
+        experimentId: null,
+        $or: [
+          {
+            nextUpdate: {
+              $exists: true,
+              $lte: new Date(),
+            },
+          },
+          {
```
mknowlton89, line 86: @Kevin-Chant Do we need this? Is this handling dashboards that have never been updated? Do we want to auto-update dashboards that have never been updated?

Kevin-Chant: Yeah it's necessary for the first update after enabling auto-updates (including first update after creation). The `nextUpdate` is only scheduled after an update is finished (see `services/dashboards.ts:189`) so anything with `null` here is meant to be updated with the next batch

packages/back-end/src/jobs/updateDashboards.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  const context = await getContextForAgendaJobByOrgId(orgId);
+
+  const { org: organization } = context;
+
+  const dashboard = await context.models.dashboards.getById(dashboardId);
+  if (!dashboard) return;
+
+  // Disable auto snapshots for the dashboard so it doesn't keep trying to update if schedule is off
+  if (organization?.settings?.updateSchedule?.type === "never") {
```
mknowlton89: Is this still relevant now that the update schedule is scoped to the dashboard and not based on the org level setting? Or does this org level setting make it so no resources can be auto-updated?

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -35,6 +37,16 @@ import { ReqContext } from "back-end/types/organization";
...
+/**
+ * Ensures all dataVizConfig items have IDs. Adds IDs to any items that don't have them.
+ */
+function ensureDataVizIds(dataVizConfig: DataVizConfig[]): DataVizConfig[] {
+  return dataVizConfig.map((config) => ({
+    ...config,
+    id: config.id || `data-viz_${uuidv4()}`,
+  }));
+}
+
```
mknowlton89, line 49: Previously, we were only referencing visualizations by `title` - this could lead to issues if a saved-query had two visualizations of the same name.

If this happened, when adding a visualization to a dashboard block, the user wouldn't be able to independently show/hide the visualizations with the same name, they could only show/hide both. By introducing ids, this remedies the issue. Its possible, but highly unlikely that two visualizations in the same block would have duplicate ids, so this feels safe enough.

packages/front-end/components/CustomFields/CustomFieldInput.tsx
```diff
@@ -15,7 +15,7 @@ const CustomFieldInput: FC<{
   className?: string;
 }> = ({
   customFields,
-  currentCustomFields,
+  currentCustomFields = {},
```
mknowlton89, line 18: Not sure what this change is meant to accomplish

packages/front-end/components/SqlExplorer/BigValueChart.tsx
```diff
@@ -6,7 +6,8 @@ import { formatCurrency } from "@/services/metrics";
 
 type Props = {
   value: number;
-  format: BigValueFormat;
+  format?: BigValueFormat;
+  formatter?: (value: number) => string;
```
mknowlton89, line 10: Do we need this, or can we rely on the existing `format` prop?

packages/front-end/pages/metric-explorer.tsx
```diff
@@ -0,0 +1,13 @@
+import { Box } from "@radix-ui/themes";
```
mknowlton89: Think we can safely remove this page, correct?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
```diff
@@ -0,0 +1,139 @@
...
+          }
+        >
+          <SelectItem value="avg">
+            {metric?.metricType === "proportion" ? "Proportion" : "Average"}
+          </SelectItem>
+          <SelectItem value="sum">
+            {metric?.metricType === "proportion" ? "Unit Count" : "Sum"}
+          </SelectItem>
+        </Select>
+      )}
```
mknowlton89, line 120: I think we should add a format option here?

packages/front-end/components/CustomFields/CustomFieldInput.tsx
Kevin-Chant: I'm not sure either but it was part of @jdorn's original PR for metric explorer so I assumed it was fixing some nullpointer

packages/front-end/components/SqlExplorer/BigValueChart.tsx
Kevin-Chant: Same as above, this was one of the changes Jeremy had with the metric explorer components

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
Kevin-Chant: What sort of formatting are you thinking?

packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -51,63 +56,143 @@ export class DashboardModel extends BaseClass {
...
-      "enableAutoUpdates" in updates
-    ) {
-      return false;
+    if (existing.experimentId) {
+      if (!this.context.hasPremiumFeature("dashboards")) {
+        throw new Error("Your plan does not support updating dashboards.");
+      }
+      const { experiment } = this.getForeignRefs(existing);
+      if (!experiment) throw new Error("Experiment not found.");
+      return this.context.permissions.canUpdateReport(experiment);
```
Kevin-Chant, line 185: > Editing privileged fields (metadata/settings) requires canManage

@mknowlton89 this behavior got dropped with these changes, was that intentional?

packages/front-end/components/Settings/UpgradeModal/index.tsx
```diff
@@ -282,6 +282,8 @@ export default function UpgradeModal({
   };
 
   const bullets: Partial<Record<CommercialFeature, string>> = {
+    "share-product-analytics-dashboards":
+      "Create product analytics dashboards and control who can view and edit them.",
```
Kevin-Chant, line 286: Should the branding for PA be capitalized?

packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -408,6 +410,11 @@ export const POLICY_METADATA_MAP: Record<
     description:
       "Create, edit, and delete official resources. For example: Manage resources like Fact Tables, Metrics, Segments, etc that have been marked as 'Official'.",
   },
+  GeneralDashboardsFullAccess: {
+    displayName: "General Dashboards Full Access",
+    description:
+      "Create, edit, and delete general dashboards used for building Product Analytics dashboards.",
```
Kevin-Chant: Nit: phrasing
```suggestion
      "Create, edit, and delete Product Analytics dashboards.",
```

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
mknowlton89: Sorry, you can disregard - I was thinking we'd do the same formatting available in the Sql Explorer (e.g. be able to select a number and type of currency, or accounting, etc, but I spoke with Jeremy, and this is already built into the Metric logic (I just don't have any good local metrics built out like this).

packages/front-end/components/Settings/UpgradeModal/index.tsx
mknowlton89: I went back and forth on this as well. Kari's designs had it lowercase. I can sync with her.

Kevin-Chant: Either way is fine by me just wanted to call attention to it :+1:

packages/back-end/src/enterprise/models/DashboardModel.ts
mknowlton89: I spoke to Jeremy about this and he did confirm that we don't want to give any special treatment to `superAdmins`. So now it will be based off the user's permission.

I do see now that this changed introduced a gap though.

I need to add back the logic about what a non-owner can update.

PR #4717, opened 2025-10-16
Allow refreshing/editing columns of Official fact tables via UI

jdorn:
### Features and Changes

With the addition of metric slices plus better column type detection, we should allow editing and refreshing columns via the UI, even for managed fact tables.  We still check for edit permission, just not the elevated permissions required for managed resources.


bryce-fitzsimons (approved):
looks good. I assume nothing special is needed for the REST API (doesn't look like it).

PR #4722, opened 2025-10-17
Add org setting for saved group size limit

Kevin-Chant:
### Features and Changes

* Adds an organization setting `savedGroupSizeLimit` to allow control of the maximum ID List size
* Updates the relevant frontend components for managing ID Lists to respect the size constraint
* Adds a new permission policy for bypassing the configured limit and shows a checkbox on the frontend for users with this permission
* Applies the limit to both the app controller and api controller endpoints for managing saved groups

### Testing

1. Create an ID list with a list of values
2. In org settings -> SDK Configuration set a limit below the number of values in that ID List
3. Attempt to add an item to the existing list and delete an item to see that an error is thrown
4. Attempt to create a new list with too many items and observe that the bypass and submit button are both disabled
5. Create a custom role with the new bypass permission and assign it to a user
6. Attempt to create a new list with too many items and observe that you can now bypass the size limit
7. Check that the POST and PUT endpoints for saved groups both enforce the limit for id lists
8. Delete the size limit and verify that you can add items to existing lists and create new lists above the previous limit

#### Creating a list without bypass permission
image

#### Creating a list with bypass permission
image

#### Deleting from a list that's already over the limit
image

#### Adding to a list that's already over the limit
image

#### API update existing list
image

#### API create new list
image


packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -324,6 +331,11 @@ export const POLICY_METADATA_MAP: Record<
     displayName: "Saved Groups Full Access",
     description: "Create, edit, and delete saved groups",
   },
+  SavedGroupsBypassSizeLimit: {
+    displayName: "Saved Groups Bypass Size Limit",
+    description: "Bypass org-defined size limits for saved groups",
+  },
+
```
mknowlton89, line 338: Quick ask: Can you also update the user-permissions.mdx docs article to add this new policy

packages/front-end/components/GeneralSettings/SavedGroupSettings.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <Box width="220px" flexShrink="0">
+              <Heading size="4" as="h4">
+                Saved Group Settings
+              </Heading>
+            </Box>
+
+            <Flex align="start" direction="column" flexGrow="1" pt="6">
+              <Box mb="6" width="100%" mt="2">
+                <Text as="label" className="font-weight-semibold" size="3">
+                  ID List Size Limit
```
mknowlton89, line 22: Small nit: Wonder if it would make sense to add a tooltip here to outline why someone might want to set this value? I'll defer to you as you're closer to the issue this is solving than I am.

mknowlton89 (approved):

PR #4724, opened 2025-10-17
fix experiment tab history

tzjames:
### Features and Changes
Fix the tab behavior with the back button.

### Testing
Go to an experiment. 
Click through a bunch of tabs.
Click back button. 
See the tabs in the reverse order.
Click back button from initial tab.
See the page from before you loaded the experiment (all experiments list page)


tzjames:
Abandoning this for now.  There are more issues with history than just on this page, and don't want to spend more time figuring out exactly what we want and need everywhere.

PR #4725, opened 2025-10-17
fix(navigation): Handle cmd+click properly

ahdriel:
#### Sidebar

Cmd+Click for top-level links (Home and Features) does not work. Which is less of a problem for items that have sublinks (those handled Cmd+Click properly), but impacts Home and Features more.

Now all items properly respect cmd+click.

#### Tables

Similarly, clicking on some Table rows would not respect Cmd+Click. Now they do.

### DIFF
--- a/packages/front-end/components/Settings/DataSources.tsx
+++ b/packages/front-end/components/Settings/DataSources.tsx
@@ -41,9 +41,24 @@ const DataSources: FC = () => {
       <tbody>
         {filteredDatasources.map((d, i) => (
           <tr
-            className="nav-item"
+            className="nav-item cursor-pointer"
             key={i}
             onClick={(e) => {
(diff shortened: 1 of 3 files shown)


packages/front-end/components/Settings/DataSources.tsx
```diff
@@ -41,9 +41,24 @@ const DataSources: FC = () => {
       <tbody>
         {filteredDatasources.map((d, i) => (
           <tr
-            className="nav-item"
+            className="nav-item cursor-pointer"
             key={i}
             onClick={(e) => {
+              // If clicking on a link or button, default to browser behavior
```
lukesonnet, line 47: Should probably have a `tr` that does this automatically.

lukesonnet (approved):

PR #4728, opened 2025-10-18
Prevent background flashing if user has dark mode

ahdriel:
### Features and Changes

If you are on dark mode, regardless if forced or system setting, when loading GrowthBook we would default to a white background, which creates a bad experience.

So now we inject the theme in `head` so the background is properly set.

While doing it I also simplified some of the logic in the UIThemeProvider and removed `theme--`, keeping only `-theme` classes.

https://www.loom.com/share/5e516c89f46a473f81e5a2f8a3ff2fa7


mknowlton89 (approved):
Tested locally and it works as expected.

PR #4731, opened 2025-10-21
Block multi-statement queries in SQL Explorer

jdorn:
### Features and Changes

SQL Explorer only supports a single SELECT statement at a time.  Add a sanity check to disallow semicolons in query strings during validation step.

### DIFF
--- a/packages/shared/test/sql.test.ts
+++ b/packages/shared/test/sql.test.ts
@@ -246,6 +246,10 @@ describe("isReadOnlySQL", () => {
     const sql = "/* Outer comment -- */ DROP TABLE users;\nSELECT * FROM users";
     expect(isReadOnlySQL(sql)).toBe(false);
   });
+  it("handles block comment inside line comment", () => {
+    const sql = `-- /*\nDROP TABLE users\n-- */ SELECT 1`;
+    expect(isReadOnlySQL(sql)).toBe(false);
+  });
 });
 describe("isMultiStatementSQL", () => {
   it("should return true for multiple statements", () => {


packages/shared/src/sql.ts
```diff
@@ -69,11 +69,20 @@ export function ensureLimit(sql: string, limit: number): string {
...
+  if (!isReadOnlySQL(sql)) {
+    throw new Error("SQL is not read-only");
+  }
+  return sql;
+}
+
 export function isReadOnlySQL(sql: string) {
   const normalized = sql
     .trim()
     .replace(/--.*$/gm, "") // remove line comments
```
bryce-fitzsimons: move after /* */ check

bryce-fitzsimons (approved):

docs/docs/self-host/production.mdx
```diff
@@ -97,7 +103,24 @@ Store uploads in a Google Cloud Storage bucket.
 
 ## Scaling
 
-GrowthBook instances are stateless and can be horizontally scaled behind a load balancer. We recommend each instance have at least 2GB of RAM and 1vCPU. You should run at least 3 instances at a time for maximum fault tolerance.
+### GrowthBook Container
+
+GrowthBook container instances are stateless and can be horizontally scaled behind a load balancer. We recommend each instance have at least 2GB of RAM and 1vCPU. You should run at least 3 instances at a time for maximum fault tolerance.
```
tzjames: We routinely go over 2GB. I think we should recommend a minimum of 4.

PR #4743, opened 2025-10-22
feat(sql): Add more context to runQuery for tagging

lukesonnet:
### Features and Changes

We can add much more context to the query/connection payload for certain warehouses (e.g. snowflake; bigquery) so that our customer's can do better meta-analysis of query costs on their end.

This PR passes around more user and experiment context when necessary and then adds it only for the snowflake integratino.

### Screenshots

From testing in screenshot, the query tags are getting passed through correctly.

Screenshot 2025-10-23 at 2 52 56 PM

Also, the external ID is being set correctly in mongo DB.

### DIFF
--- a/packages/back-end/src/services/snowflake.ts
+++ b/packages/back-end/src/services/snowflake.ts
@@ -7,6 +7,7 @@ import {
 } from "back-end/src/types/Integration";
 import { TEST_QUERY_SQL } from "back-end/src/integrations/SqlIntegration";
 import { QueryMetadata } from "back-end/types/query";
+import { logger } from "back-end/src/util/logger";
 
 type ProxyOptions = {
   proxyHost?: string;
@@ -52,6 +53,7 @@ function getSnowflakeQueryTagString(queryMetadata?: QueryMetadata) {
 
   // if still too long, just send the application key
   if (json.length > 2000) {
+    logger.warn("Snowflake query tag is too long, truncating", { json });
     json = JSON.stringify({
       application: "growthbook",
     });


packages/back-end/src/queryRunners/ExperimentResultsQueryRunner.ts
```diff
@@ -427,6 +434,16 @@ export const startExperimentResultQueries = async (
   return queries;
 };
 
+function getAdditionalQueryMetadataForExperiment(
+  experiment: ExperimentInterface,
+): AdditionalQueryMetadata {
+  return {
+    experimentOwner: experiment.owner,
+    experimentProject: experiment.project ?? "All",
```
ahdriel: nit: oooh, do you think `All` projects make sense in the context of a query metadata? What about `None`?

packages/back-end/src/queryRunners/ExperimentReportQueryRunner.ts
```diff
@@ -26,6 +30,7 @@ export type ReportQueryParams = {
   metricMap: Map<string, ExperimentMetricInterface>;
   factTableMap: FactTableMap;
   metricGroups: MetricGroupInterface[];
+  experimentQueryMetadata: ExperimentQueryMetadata | null;
```
ahdriel: nit: Should this be named just `metadata`? Or maybe `experimentMetadata`?

packages/back-end/src/services/snowflake.ts
```diff
@@ -25,10 +29,42 @@ function getProxySettings(): ProxyOptions {
...
+      json = JSON.stringify({
+        ...Object.fromEntries(
+          Object.entries(metadata).filter(([key]) => !tagKeys.includes(key)),
+        ),
+      });
+    }
+  }
+
+  // if still too long, just send the application key
+  if (json.length > 2000) {
```
ahdriel, line 55: nit: Do we want to add a log here?

```
logger.warning({ json }, 'Dropped Snowflake queryTag because of excessive length');
```

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -172,14 +175,28 @@ export default abstract class SqlIntegration
...
+      metadata?: QueryMetadata,
+    ) => {
       if (isMultiStatementSQL(sql)) {
         throw new Error("Multi-statement queries are not supported");
       }
-      return originalRunQuery.call(this, sql, setExternalId);
+      metadata = {
+        ...metadata,
+        userId: this.context.userId,
+        userName: this.context.userName,
```
ahdriel, line 189: nit: Should this happen when we create the Integration and set it in `additionalMetadata` property?

I am happy to skip this for now as well as it creates the possibility of overriding it, etc, and we would need to handle that -- but it would be nice to have a single place with the default context metadata (application, user).

ahdriel (approved):

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Maybe, but we create the integration in a ton of places, so I like the idea of it always being set.

PR #4753, opened 2025-10-27
fix minStickyBucketVersion comparison

bryce-fitzsimons:
Bucket versions > 0 run into a bug where minBucketVersion doesn't allow bucketVersion of the same number through. This causes rebucketing when not in the base case.

### DIFF
--- a/packages/back-end/src/controllers/reports.ts
+++ b/packages/back-end/src/controllers/reports.ts
@@ -45,6 +45,7 @@ import {
   generateExperimentReportSSRData,
 } from "back-end/src/services/reports";
 import { ExperimentResultsQueryRunner } from "back-end/src/queryRunners/ExperimentResultsQueryRunner";
+import { getAdditionalQueryMetadataForExperiment } from "back-end/src/services/experiments";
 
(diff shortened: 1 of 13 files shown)


gavrashenko:
Hello!

Are there any estimates to release?

packages/sdk-js/src/core.ts
```diff
@@ -1013,7 +1013,7 @@ function getStickyBucketVariation({
 
   // users with any blocked bucket version (0 to minExperimentBucketVersion) are excluded from the test
```
glifchits:
```suggestion
  // users with any blocked bucket version (0 to minExperimentBucketVersion, non-exclusive) are excluded from the test
```

bryce-fitzsimons: good callout, perhaps `(0 to minExperimentBucketVersion - 1)` is more clear

bryce-fitzsimons:
```suggestion
  // users with any blocked bucket version (0 to minExperimentBucketVersion - 1) are excluded from the test
```

PR #4756, opened 2025-10-28
Include Saved Queries in Dashboard View Queries Modal

Kevin-Chant:
### Features and Changes

Currently the `AsyncQueriesModal` only takes in standard queries as a prop which means Dashboards can't show the results, durations, or errors for Saved Queries. This PR adds a new component to render a single Saved Query and passes props through the modal to include them in the list of total queries for a dashboard

### Testing

- Create a dashboard without a sql explorer block and ensure the view queries button functions as normal
- Add a sql explorer block and create a saved query. That query should appear at the end of the list of queries
- Refresh the dashboard. The results should update when the query was last ran

### Screenshots

image

### DIFF
--- a/packages/front-end/components/Queries/ViewAsyncQueriesButton.tsx
+++ b/packages/front-end/components/Queries/ViewAsyncQueriesButton.tsx
@@ -42,7 +42,7 @@ const ViewAsyncQueriesButton: FC<{
     <>
       {ctaComponent ? (
         ctaComponent(() => {
-          if (!queries.length) return;
+          if (totalQueries === 0) return;
           setOpen(!open);
         })
       ) : (


packages/front-end/components/PowerCalculation/PowerCalculationSettingsModal/SetParamsStep.tsx
```diff
@@ -219,6 +219,7 @@ const PopulationDataQueryInput = ({
       {queryModalOpen ? (
         <AsyncQueriesModal
           queries={populationData?.queries?.map((q) => q.query) ?? []}
+          savedQueries={[]}
```
Kevin-Chant, line 222: This is the only other existing use of `AsyncQueriesModal` so I opted to make the prop required rather than optional

packages/front-end/components/SavedQueries/ExpandableSavedQuery.tsx
```diff
@@ -0,0 +1,107 @@
...
+    if (comments && comments[2]) {
+      title = comments[2];
+    }
+  }
+
+  return (
+    <div className="mb-4">
+      <h4 className="d-flex align-items-top">
+        {savedQuery.results.error ? (
+          <FaExclamationTriangle className="text-danger mr-2" title="Failed" />
```
Kevin-Chant, line 26: So from the logic I see [here](https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/routers/saved-queries/saved-queries.controller.ts#L201), it looks like it's not actually possible to have an error or non-existent results on a `savedQuery`; it just won't be updated in the DB.

I still kept some of this error handling from `ExpandableQuery` in case it changes, but much of this component will likely never be rendered

mknowlton89: Looking at the saved query model, it looks like we support storing and error on the results, but my understanding matches yours in that we're not actively using it today. I did a quick search in Mongo and we don't currently have any records that have a `results.error` property set.

I think the approach you have of supporting it is correct.

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
```diff
@@ -53,6 +53,33 @@ export async function getSavedQueries(req: AuthRequest, res: Response) {
...
+    });
+  }
+
+  const { ids } = req.params;
+  const savedQueryIds = ids.split(",");
+
+  const docs = await context.models.savedQueries.getByIds(savedQueryIds);
+
+  // Lookup table so we can return queries in the same order we received them
+  const map = new Map(docs.map((d) => [d.id, d]));
```
mknowlton89, line 75: Very small nit: I don't love the `map` name. I don't have any better suggestions, though.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardViewQueriesButton.tsx
```diff
@@ -21,10 +21,10 @@ export default function DashboardViewQueriesButton({
   buttonProps = {},
   hideQueryCount = false,
 }: Props) {
-  const { allQueries, snapshotError, refreshStatus } = useContext(
-    DashboardSnapshotContext,
-  );
-  const count = (allQueries ?? []).length;
+  const { allQueries, savedQueriesMap, snapshotError, refreshStatus } =
+    useContext(DashboardSnapshotContext);
```
mknowlton89, line 25: I haven't been able to track down the cause yet, but I noticed when creating a brand new experiment dashboard I was getting the `View failed queries` cta, as the `refreshStatus` was coming back as `failed`.

Still looking into why this might be.

mknowlton89: Update for future clarity - this has been resolved on main already

packages/back-end/src/routers/saved-queries/saved-queries.controller.ts
Kevin-Chant: True, I just copied this endpoint wholesale from the Datasource controller :joy:

PR #4760, opened 2025-10-28
Revised PA Dashboard creation flow

mknowlton89:
### Features and Changes

This PR does a few things:
- This PR introduces a new `Create Dashboard` flow that takes a user directly to the `DashboardWorkspace` via a new `/product-analytics/dashboards/new` route. Then, when a user adds their first block to the dashboard, only then do we save the dashboard. Upon clicking the "Done Editing" button, we direct the user to the newly created dashboard page. If the user doesn't add a block, they can click the `Exit without saving` which will return the user to the dashboard index page. (The "Done Editing" is disabled if the user hasn't added any blocks).
- This PR also adds projects filtering to the Sql Explorer Modal for data sources when opened from the dashboard editor.
- This PR also adds projects filtering to the Metric Explorer for fact metrics when opened from the dashboard editor.

### Dependencies

- None

### Testing

- [x] Ensure you can create a new dashboard from the index page
- [x] Ensure you can edit existing PA dashboards 
- [x] Ensure you can create new experiment dashboards from an experiment
- [x] Ensure you can edit an existing experiment dashboard
- [x] Ensure that the data sources available via the SQLExplorerModal are a subset of the Dashboard's projects list
- [x] Ensure that if a Dashboard has a SQL Explorer Block that contains a data source who's projects are no longer a subset of the Dashboard's projects (because either the dashboard's projects changed, or the data source's projects change) that the existing data source is still visible in the dashboard.
- [x] Ensure that the fact metrics available via the Metric Explorer are a subset of the Dashboard's projects
- [x]  Ensure that if a Dashboard has a Metric Explorer Block that contains a fact metric who's projects are no longer a subset of the Dashboard's projects (because either the dashboard's projects changed, or the fact metric's projects change) that the existing data source is still visible in the dashboard.

### DIFF
--- a/packages/sdk-js/src/core.ts
+++ b/packages/sdk-js/src/core.ts
@@ -1011,9 +1011,9 @@ function getStickyBucketVariation({
     expFallbackAttribute,
   );
 
-  // users with any blocked bucket version (0 to minExperimentBucketVersion) are excluded from the test
+  // users with any blocked bucket version (0 to minExperimentBucketVersion - 1) are excluded from the test
   if (expMinBucketVersion > 0) {
-    for (let i = 0; i <= expMinBucketVersion; i++) {
+    for (let i = 0; i < expMinBucketVersion; i++) {
       const blockedKey = getStickyBucketExperimentKey(expKey, i);
       if (assignments[blockedKey] !== undefined) {
         return {
--- a/packages/sdk-js/test/cases.json
+++ b/packages/sdk-js/test/cases.json
@@ -6411,6 +6411,249 @@
       }
     ],
     [
+      "uses a sticky bucket when sticky bucket version == experiment.minBucketVersion === experiment.bucketVersion",
+      {
+        "attributes": {
+          "deviceId": "d123",
+          "anonymousId": "ses123",
+          "foo": "bar",
+          "country": "USA"
+        },
+        "features": {
+          "exp1": {
+            "defaultValue": "control",
+            "rules": [
+              {
+                "key": "feature-exp",
+                "seed": "feature-exp",
+                "hashAttribute": "id",
+                "fallbackAttribute": "deviceId",
+                "hashVersion": 2,
+                "bucketVersion": 3,
+                "minBucketVersion": 3,
+                "condition": { "country": "USA" },
+                "variations": ["control", "red", "blue"],
+                "meta": [{ "key": "0" }, { "key": "1" }, { "key": "2" }],
+                "coverage": 1,
+                "weights": [0.3334, 0.3333, 0.3333],
+                "phase": "0"
+              }
+            ]
+          }
+        }
+      },
+      [
+        {
+          "attributeName": "deviceId",
+          "attributeValue": "d123",
+          "assignments": {
+            "feature-exp__3": "2"
+          }
+        }
+      ],
+      "exp1",
+      {
+        "bucket": 0.6468,
+        "featureId": "exp1",
+        "hashAttribute": "deviceId",
+        "hashUsed": true,
+        "hashValue": "d123",
+        "inExperiment": true,
+        "key": "2",
+        "stickyBucketUsed": true,
+        "value": "blue",
+        "variationId": 2
+      },
+      {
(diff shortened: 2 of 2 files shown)


packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
```diff
@@ -101,12 +104,41 @@ export default function SqlExplorerModal({
...
+  let filteredDatasources: DataSourceInterfaceWithParams[] = [];
+
+  // If the dashboard has a projects list, only include datasources that are contain all of the projects in the list or are in 'All Projects'
+  if (projects.length) {
+    filteredDatasources = datasources.filter((d) => {
+      if (!d.projects || !d.projects.length) {
+        return true;
+      }
+
+      // Always include the existing datasource if it exists, this will prevent issues if the datasource or the dashboard's projects have changed since the query was created.
```
mknowlton89, line 119: I think this is the correct approach, but calling it out.

If a user adds a saved-query to a dashboard, and later, either the dashboard projects change or the data source's projects change, a user could find themselves with a broken ui as the data source is now no longer available.

This code makes it so the UI won't break, but it could introduce other weirdness if the data source's projects were changed in order to hide it from others.

Note, the list of data sources we get here has already taken into account a user's `readAccess`, so it won't leak something that a user shouldn't see, but could still lead to weirdness.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -186,8 +188,28 @@ export default function EditSingleBlock({
...
+            return true;
+          }
+
+          // Always include the existing fact metric. This will prevent issues if the fact metric or the dashboard's projects have changed since the block was created.
+          if (
+            blockHasFieldOfType(block, "factMetricId", isString) &&
+            factMetric.id === block.factMetricId
+          ) {
+            return true;
+          }
```
mknowlton89, line 205: Same story here. (Referencing this [comment](https://github.com/growthbook/growthbook/pull/4760/files#r2473338154)).

packages/front-end/pages/product-analytics/dashboards/index.tsx
```diff
@@ -251,7 +266,11 @@ export default function DashboardsPage() {
                 title="Explore & Share Custom Analyses"
                 description="Create curated dashboards to visualize key metrics and track performance."
                 leftButton={
-                  <Button onClick={() => setShowCreateModal(true)}>
+                  <Button
+                    onClick={() =>
+                      router.push("/product-analytics/dashboards/new")
+                    }
+                  >
```
jdorn, line 274: Nit: we have a `LinkButton` component that takes an href instead of onClick

packages/front-end/pages/product-analytics/dashboards/new.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }, [project, userId]);
+
+  const [dashboard, setDashboard] = useState<DashboardInterface>(
+    createTemporaryDashboard,
+  );
+
+  const handleSubmitDashboard: SubmitDashboard<UpdateDashboardArgs> =
+    useCallback(
+      async (args) => {
+        console.log("args", args);
```
jdorn: Remove

packages/front-end/pages/product-analytics/dashboards/new.tsx
```diff
@@ -0,0 +1,137 @@
...
+              blocks: args.data.blocks || dashboard.blocks,
+            }),
+          });
+
+          if (res.status === 200) {
+            setDashboard(res.dashboard);
+          } else {
+            console.error(res);
+            throw new Error("Failed to create dashboard");
+          }
```
jdorn: Nit: This if/else is not needed.  `apiCall` throws if the response is non-200, so if it reaches here you can assume it succeeded.  Same with the one below.

PR #4761, opened 2025-10-28
Add Incremental Pipeline Mode (incremental refresh)

ahdriel:
### Feature

Add new option **Incremental** to Pipeline Mode.

For experiment analysis, our current pipeline always scans the source data (Exposure Queries and Metrics) for the whole duration of the Experiment. Meaning that if I have a 3 week old experiment refreshing every 6 hours, 3 weeks worth of data will be scanned every time it refreshes.

**Incremental** solves this by storing intermediate tables with pre-aggregated values and querying the source tables only for the new data (the delta between the last run and the new one).

This creates more cost-efficient updates for tables that have a lot of records/data.

#### Changes

- New `ExperimentIncrementalRefreshQueryRunner` for incremental query execution
- New `IncrementalRefreshModel` to track incremental refresh state (with intermediate table metadata)
- Initial support for BigQuery and Presto/Trino

### Testing

1. Enable incremental refresh in Data Source settings -> Pipeline Mode -> Incremental
2. Test with different configurations and metric types
3. Verify Full Refresh recreates all tables from scratch
4. Verify that Incremental refresh uses the intermediate tables and scans less rows on fact tables


packages/back-end/src/types/Integration.ts
```diff
@@ -47,14 +47,61 @@ export type DataType =
...
+
+export type FactMetricRegressionAdjustmentSettings = {
+  // used only for "firstExposure" window type
+  hours: number;
+  minDelay: number;
+  alias: string;
+  // used only for "phaseStart" window type
+  covariateStartDate: Date;
+  covariateEndDate: Date;
+};
```
ahdriel: Any way we can / should we separate these settings if they are used separately? Specially as they are not optional.

packages/back-end/src/types/Integration.ts
```diff
@@ -70,16 +117,41 @@ export type FactMetricData = {
...
-    alias: string;
-  };
+  raMetricSettings: FactMetricRegressionAdjustmentSettings;
   metricStart: Date;
   metricEnd: Date | null;
   maxHoursToConvert: number;
 };
 
+export type FactMetricSourceData = {
+  factTable: FactTableInterface;
```
ahdriel, line 135: question: Should this be an array? As a metric can refer to 2 tables?

Not for now, but relates to the cross-fact table metrics.

packages/back-end/src/models/QueryModel.ts
```diff
@@ -22,6 +22,8 @@ const querySchema = new mongoose.Schema({
     type: String,
     unique: true,
   },
+  name: String,
+  title: String,
```
ahdriel: question: What is the difference between both?

Or should we remove `name`?

packages/back-end/src/models/IncrementalRefreshModel.ts
```diff
@@ -0,0 +1,67 @@
...
+} from "back-end/src/validators/incremental-refresh";
+import { MakeModelClass, UpdateProps } from "./BaseModel";
+
+export const COLLECTION_NAME = "incrementalrefresh";
+
+const BaseClass = MakeModelClass({
+  schema: incrementalRefreshValidator,
+  collectionName: COLLECTION_NAME,
+  idPrefix: "ir_",
+  auditLog: {
```
ahdriel: question: Do we need auditLog for this collection?

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -287,6 +287,9 @@ export async function refreshDashboardData(
         factTableMap: await getFactTableMap(context),
         experimentQueryMetadata:
           getAdditionalQueryMetadataForExperiment(experiment),
+        // TODO(adriel): Dashboard is using IncrementalRefresh -- should we allow it or force it to be non-Incremental?
```
ahdriel: Need to check this one

packages/front-end/components/Queries/ExpandableQuery.tsx
```diff
@@ -141,11 +141,11 @@ const ExpandableQuery: FC<{
                 </tbody>
               </table>
             </div>
-          ) : (
+          ) : query.query.startsWith("SELECT") ? (
```
ahdriel, line 144: question: Do we want to keep this?

packages/back-end/src/models/IncrementalRefreshModel.ts
lukesonnet: We don't expose it anywhere. I want these kind of events to be logged with the experiment and its associated pipeline, but I don't think we should rely on this internal model for that.

PR #4777, opened 2025-10-30
metric groups bugfix

lukebrawleysmith:
### Features and Changes

Currently when all of the metrics in an experiment come from a metric group, CUPED is disabled because we do not pass the metric ids associated with the metric group when getting snapshot settings for the results tab.  This PR fixes this.  

### Screenshots

Before and after attached
cupid_disabled
cuped_properly_enabled

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
@@ -15,7 +15,6 @@ import { DEFAULT_STATS_ENGINE } from "shared/constants";
 import {
   getAllMetricIdsFromExperiment,
   getAllMetricSettingsForSnapshot,
-  expandMetricGroups,
 } from "shared/experiments";
 import { isDefined } from "shared/util";
 import { useDefinitions } from "@/services/DefinitionsContext";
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
```diff
@@ -111,11 +112,26 @@ export default function ResultsTab({
   const hasRegressionAdjustmentFeature = hasCommercialFeature(
     "regression-adjustment",
   );
-
-  const allExperimentMetricIds = getAllMetricIdsFromExperiment(
+  const { getMetricGroupById } = useDefinitions();
+  const metricAndMetricGroupIds = getAllMetricIdsFromExperiment(
```
bryce-fitzsimons: shouldn't we just pass in the `metricGroups` object into `getAllMetricIdsFromExperiment()` instead of the manual expansion below?

bryce-fitzsimons: you can pull the groups out of the definitions like so:

```
  const {
    getDatasourceById,
    getExperimentMetricById,
    getMetricById,
    getProjectById,
    metrics,
    metricGroups, // here
    datasources,
  } = useDefinitions();
```

bryce-fitzsimons (changes requested):
left some feedback. the fix is much simpler I believe...

packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
lukebrawleysmith: accepted, thanks

PR #4778, opened 2025-10-30
chore(precommit): Don't run script if we don't need to

ahdriel:
### Features and Changes

The pre-commit hook to run `gen-event-webhook-doc.ts` was too broad, running if any file in the back-end package changed.

Now we are more conservative and run it only if the script or if any file in `back-end/events` is updated.


lukesonnet (approved):
As long as the related check still runs in the GH workflow this seems harmless.

PR #4780, opened 2025-10-31
Have experiments API return signed URLs + add experiment tests

tzjames:
### Features and Changes
Makes sure that URLs are signed in API responses.  If people have a script to get all experiments they will need to download it within the 15 minutes before it expires.

We were missing tests in general for experiments so I added a bunch for all the routes.

### Testing
`yarn test`
Add a screenshot to an experiment
`curl -X GET "http://localhost:3100/api/v1/experiments" \
  -H "Authorization: Bearer <BEARER TOKEN>"`
Click on the screenshot link in the response and see that it loads.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1593,10 +1593,8 @@ export async function toExperimentApiInterface(
         description: v.description || "",
         screenshots: await Promise.all(
           v.screenshots.map(async (s) => {
-            if (!s.path) return s.path;
             try {
-              const url = await getSignedImageUrl(s.path);
-              return url || s.path;
+              return await getSignedImageUrl(s.path);
             } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/back-end/test/api/api.setup.ts
```diff
@@ -9,7 +9,7 @@ import { getAgendaInstance } from "back-end/src/services/queueing";
 
 jest.mock("back-end/src/util/secrets", () => ({
   ...jest.requireActual("back-end/src/util/secrets"),
-  CRON_ENABLED: 1,
+  CRON_ENABLED: 0,
```
tzjames, line 12: I'm not sure why we would have wanted CRON_ENABLED while running api tests.  Some of the jobs were throwing errors during the test runs.

bryce-fitzsimons (approved):

PR #4781, opened 2025-10-31
improve json editing / value modal, add copy to clipboard

bryce-fitzsimons:
- Use the Code Editor widget for editing JSON values (features)
- Fix the fullscreen display for JSON values
- Add a CTA to "copy to clipboard" in fullscreen mode
- Add copy button on rule display for string & json, clean up button UI

image

image

image

image

image

### DIFF
--- a/packages/front-end/components/Forms/CodeTextArea.tsx
+++ b/packages/front-end/components/Forms/CodeTextArea.tsx
@@ -180,7 +180,8 @@ const AceEditor = dynamic(
 
 export type Language = "sql" | "json" | "javascript" | "python" | "yml";
 
-export const FIVE_LINES_HEIGHT = 98;
+export const FIVE_LINES_HEIGHT = 97;
+export const TEN_LINES_HEIGHT = 194;
 
 type CodeTextAreaFieldProps = Omit<
   FieldProps,
@@ -227,7 +228,7 @@ export default function CodeTextArea({
   wrapperClassName,
   completions,
   resizable = false,
-  defaultHeight = 200, // for resizable
+  defaultHeight = TEN_LINES_HEIGHT, // for resizable
   showCopyButton = false,
   showFullscreenButton = false,
   ...otherProps


packages/front-end/components/Features/FeatureValueField.tsx
```diff
@@ -120,6 +134,59 @@ export default function FeatureValueField({
   }
 
   if (valueType === "json") {
+    if (useCodeInput) {
```
ahdriel, line 137: question: When do we want NOT to use the code editor for a json field?

packages/front-end/components/Features/ValueDisplay.tsx
```diff
@@ -85,37 +89,89 @@ export default function ValueDisplay({
...
-                onClick={() => setModalOpen(true)}
-              >
-                <PiCornersOutBold size={15} />
-              </IconButton>
-            </Tooltip>
-          </Flex>
+        </div>
+        {!isFullscreen && (
+          <div
+            className="d-flex"
```
ahdriel: suggestion: Should we use `Flex` here?

packages/front-end/components/Features/ValueDisplay.tsx
```diff
@@ -85,37 +89,89 @@ export default function ValueDisplay({
...
+            copySuccess ? (
+              <Button style={{ width: 100 }} icon={<PiCheck />} color="gray">
+                Copied
+              </Button>
+            ) : (
+              <Button
+                style={{ width: 100 }}
+                icon={<PiCopy />}
+                onClick={() => {
+                  if (!copySuccess) performCopy(value);
```
ahdriel, line 162: nit: Should we have `handleCopyButtonClick` and use for both fullscreen and not-fullscreen values?

packages/front-end/components/Features/RuleModal/SafeRolloutFields.tsx
```diff
@@ -394,6 +394,9 @@ export default function SafeRolloutFields({
               renderJSONInline={true}
               disabled={disableFields || controlValueDisabled}
               useDropdown={true}
+              useCodeInput={true}
+              showFullscreenButton={true}
+              codeInputDefaultHeight={98}
```
ahdriel: This feels like a very magical number, can we make it a const and add more context to it?

And also why only for SafeRollout?

packages/front-end/components/Features/RuleModal/ExperimentRefFields.tsx
```diff
@@ -163,6 +163,9 @@ export default function ExperimentRefFields({
                   valueType={feature.valueType}
                   feature={feature}
                   renderJSONInline={false}
+                  useCodeInput={true}
+                  showFullscreenButton={true}
+                  codeInputDefaultHeight={80}
```
ahdriel: nit: Same here regarding why we need `80`?

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -205,15 +216,26 @@ export default function CodeTextArea({
   onCtrlEnter,
   wrapperClassName,
   completions,
+  resizable = false,
+  defaultHeight = 200, // for resizable
```
ahdriel: question: Does this apply even if resizable is `false`? I would expect so, so maybe we need to remove the comment?

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -205,15 +216,26 @@ export default function CodeTextArea({
...
+  showFullscreenButton = false,
   ...otherProps
 }: Props) {
   // eslint-disable-next-line
   const fieldProps = otherProps as any;
   const { theme } = useAppearanceUITheme();
   const [editor, setEditor] = useState<null | Ace.Editor>(null);
+  const [isFullscreen, setIsFullscreen] = useState(false);
   const containerRef = useRef<HTMLDivElement>(null);
+  const editorUid = useMemo(() => uuid4(), []);
```
ahdriel: nit: React does have `useId` but I am not familiar with tradeoffs between uuidv4 and their version of an id.

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -254,42 +299,180 @@ export default function CodeTextArea({
...
           <>
+            <style jsx>{`
+              .code-editor-fullscreen {
+                position: fixed;
+                top: 0;
+                left: 0;
+                right: 0;
+                bottom: 0;
+                z-index: 1050;
+                background-color: ${theme === "light" ? "#fff" : "#1e1e1e"};
```
ahdriel: suggestion: We should use `var(--color-surface-solid)` instead

packages/front-end/components/Forms/CodeTextArea.tsx
```diff
@@ -254,42 +299,180 @@ export default function CodeTextArea({
...
-                placeholder={placeholder}
-                fontSize="1em"
-                completions={completions}
-                {...heightProps}
-                setOptions={
-                  language === "sql"
+              {isFullscreen ? (
+                <Flex align="center" gap="3" mb="2" justify="between">
+                  <label className="mb-0 d-block font-weight-bold">
+                    {fieldProps.label}
```
ahdriel, line 341: question: Should we improve the typing for `fieldProps` if we expect `label` to exist?

ahdriel (approved):

packages/front-end/components/Features/FeatureValueField.tsx
bryce-fitzsimons: no sure, i left it there as a "just in case"

ahdriel: I vote for us to check the usage or at least make `true` the default behavior and update the old usages if we believe this is a better experience.

packages/front-end/components/Features/ValueDisplay.tsx
bryce-fitzsimons: will leave it alone since there is slightly different behavior with event propagation

packages/front-end/components/Forms/CodeTextArea.tsx
bryce-fitzsimons: unfortunately no, when not resizable and when applying the defaultHeight, the button layout gets messed up.

Can look more into some of these weird layout issues later, or clean up the combinations that will never be used. But it doesn't break for any of our current usages.

packages/front-end/components/Forms/CodeTextArea.tsx
bryce-fitzsimons: sure, we can use the native hook

PR #4782, opened 2025-10-31
Mk/metric exp extensions

mknowlton89:
### Features and Changes

This PR introduces functionality to our Product Analytics Dashboards (and by extension our Experiment Dashboards) where a user can add `row filters` to a Fact Metric's analysis via the Metric Explorer block on a dashboard.

This initial PR only supports leveraging previously defined `row` filters from a Fact Metric's parent Fact Table. Eventually, we would like to support the ability for a user to write adhoc filters (or build them via a nice UI) inside the Metric Explorer to support users building adhoc reports. E.G. Let's say a user wants to filter by order value, and there is only a `value > 100` filter defined on the Fact Table, but the user wants to use a `value > 150` for a particular Dashboard. The user should be able to do that without having to go create it on the Fact Table first.

Again, that functionality doesn't exist in this PR, just calling it out as a future improvement.

To do this, we had to update the shape of the `metricAnalysisSettingsValidator`. I opted to make them required properties that can be nullable, to match the pattern of the `populationIds`, but we can swap the definition to optional properties if there are downstream effects I'm not aware of.

To accomplish the actual filtering, I've updated the `getMetricAnalysisQuery` function - now, this function takes in a metric and the metricAnalysisSettings, and if the metricAnalysisSettings contains filters (numeratorFilters or denominatorFilters) we add those filters to a metric's existing numeratorFilters/denominatorFilters right before we pass the metric to the query builder. This allows us to re-use the existing logic to generate sql with filters, without having to update the actual metric.

### Dependencies

- None

### Testing

- [x] Ensure that the filters work as expected and actually filter the data
- [x] Ensure that if a user switches from one metric to another inside of the Metric Explorer, we reset the `numeratorFilters` and `denominatorFilters` arrays
- [x] Ensure that we only show `denominatorFilters` in the UI if the metric type supports it
- [x] Ensure that the graph on the [fmid].tsx page still works as expected
- [x] Ensure this doesn't introduce any regressions with North Star Metrics
- [ ] Ensure this doesn't introduce any regressions with experiment analysis **(WOULD LOVE HELP WITH TESTING THIS WITH ADDITIONAL EXPERIMENTS OUTSIDE OF MY DEV ENV)**


packages/back-end/src/routers/metric-analysis/metric-analysis.controller.ts
```diff
@@ -51,14 +56,33 @@ export const postMetricAnalysis = async (
...
+
+  console.log("metricBefore", metricObj);
+
+  // This is just a test to see if we can add filters to the metric obj in an adhoc manner
+  if (data.filters) {
+    metricObj.numerator.filters = [
+      ...metricObj.numerator.filters,
+      ...data.filters,
+    ];
+  }
```
mknowlton89: @bryce-fitzsimons As an exercise to familiarize myself with metric analysis, this is one of the paths I explored this morning.

This resolves the super simple use case of allowing a user to add an additional row filter (that is pre-defined on the fact table) to a `mean` fact metric via the Metric Explorer block.

It essentially just updates the `postMetricAnalysis` endpoint to take in a `filters` array, and before we pass the metric to `createMetricAnalysis`, we're spreading in any of those `filters` to the metric's filters array. (In this case, since it's a mean, we're only spreading it into the numerator's filters array. If we truly want to support it this way, we'll need to support numerator and denominator metrics, but just working through a POC.

CC: @lukesonnet @lukebrawleysmith - I'm very unfamiliar with metric analysis so unsure if this will give us any unintended consequences.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/MetricExplorerBlock.tsx
```diff
@@ -172,6 +174,14 @@ export default function MetricExplorerBlock({
...
       ) : ["running", "queued"].includes(metricAnalysis.status || "") ? (
         <LoadingOverlay />
+      ) : "dataset" in chartData &&
+        Array.isArray(chartData.dataset) &&
+        !chartData.dataset[0]?.source?.length ? (
+        <Box p="4" style={{ textAlign: "center" }}>
+          <Text style={{ color: "var(--color-text-mid)", fontWeight: 500 }}>
+            The query ran successfully, but no data was returned.
+          </Text>
+        </Box>
```
mknowlton89, line 184: New pattern when a query runs successfully, but doesn't return any data.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/MetricExplorerBlock.tsx
```diff
@@ -103,6 +103,8 @@ export default function MetricExplorerBlock({
         },
         axisLabel: {
           color: textColor,
+          rotate: -45,
+          hideOverlap: true,
```
mknowlton89, line 107: Rotates the axisLabel for better readability with a large number of data points

packages/back-end/src/api/openapi/payload-schemas/PostFactMetricAnalysisPayload.yaml
```diff
@@ -18,6 +18,16 @@ properties:
...
+  numeratorFilters:
+    type: array
+    items:
+      type: string
+    description: Array of Fact Table Filter Ids
+  denominatorFilters:
+    type: array
+    items:
+      type: string
+    description: Array of Fact Table Filter Ids
```
mknowlton89: So, this came about because I updated the `metricAnalysisSettingsValidator` so include these new properties, and similar to the `populationId`, I made it required, but nullable.

After going down this path, I'm not sure this is what we want - this would enable users via the REST API to build a metric analysis in a similar way to the Dashboard Builder, which in itself doesn't seem like a bad idea, but not sure if we want to support it.

The alternative is I can make the properties optional, and then we don't have to support them via the REST API.

@lukesonnet @lukebrawleysmith Do you know if any orgs using this endpoint would like this functionality?

lukebrawleysmith: I don't know of any, but that doesn't mean it wouldn't be helpful.

packages/back-end/src/services/metric-analysis.ts
```diff
@@ -7,8 +8,37 @@ import { MetricAnalysisQueryRunner } from "back-end/src/queryRunners/MetricAnaly
 import { getFactTableMap } from "back-end/src/models/FactTableModel";
 import { Context } from "back-end/src/models/BaseModel";
 import { SegmentInterface } from "back-end/types/segment";
+import { MetricAnalysisParams } from "../types/Integration";
 import { getIntegrationFromDatasourceId } from "./datasource";
 
+// When creating an analysis for metrics via a Dashboard, we sometimes apply adhoc filters to the analysis, that aren't a part of the metric itself (e.g. adding additional row filters)
+// This function takes the metric and applies these adhoc settings before running the analysis
+export function updateMetricByAnalysisSettings(
```
Kevin-Chant: Nit naming suggestion: `getMetricWithFiltersApplied` or similar. Starting with `update` made it seem a lot more worrisome in the `SqlIntegration` usage

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/MetricExplorerBlock.tsx
Kevin-Chant: Should/could this be applied at the `DashboardBlock/index.tsx` level or is the pattern specific to metric analyses?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -729,12 +729,21 @@ export default function EditSingleBlock({
...
                         ?.label || "Metric",
                     factMetricId: value,
+                    ...(isMetricExplorer && {
+                      metricAnalysisId: "",
+                      analysisSettings: {
+                        ...block.analysisSettings,
+                        numeratorFilters: null,
+                        denominatorFilters: null,
+                      },
+                    }),
```
Kevin-Chant, line 756: Good addition, I think I planned to do some of this resetting and then forgot about it. Were there other metric analysis fields that are worth resetting here as well?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
```diff
@@ -55,175 +73,371 @@ export default function MetricExplorerSettings({ block, setBlock }: Props) {
   // TODO: reset invalid values when metric changes
```
Kevin-Chant: Oh, yeah this is that TODO I mentioned. Can we get rid of it by expanding your changes in the other file?

packages/front-end/enterprise/components/Dashboards/DashboardSnapshotProvider.tsx
```diff
@@ -34,6 +34,7 @@ import { getQueryStatus } from "@/components/Queries/RunQueriesButton";
 
 export const DashboardSnapshotContext = React.createContext<{
   experiment?: ExperimentInterfaceStringDates;
+  dashboard?: DashboardInterface;
```
Kevin-Chant: This seems unused (and is a bad pattern IMO). I removed it locally and type-check still passed - what's the impetus for adding it here?

packages/back-end/src/services/metric-analysis.ts
mknowlton89: Oooof yes - that's so much better

packages/front-end/enterprise/components/Dashboards/DashboardSnapshotProvider.tsx
mknowlton89: Whoops - I think this was an artifact from a change I was exploring that I didn't remove all the way. Removed now.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/MetricExplorerBlock.tsx
mknowlton89: It's unique to the metric anaylses - the Sql Explorer data is in a different shape. I think it makes sense to keep them separate.

With that said, I did look at a sql explorer block that had a filter applied which resulted in no results, and we just show an empty graph. So I'll put up a PR shortly to handle that.

Also, documenting it here so I don't forget, sql explorer big-int visualizations don't seem to respect any `filters`, so we should address that.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
mknowlton89: Probably makes sense to reset the `userIdType`, `populationType`, and the `populationId` as well after looking at it further. It's likely those could change when switching metrics.

packages/back-end/src/routers/metric-analysis/metric-analysis.validators.ts
```diff
@@ -21,6 +21,8 @@ export const metricAnalysisSettingsValidator = z
 
     populationType: metricAnalysisPopulationTypeValidator,
     populationId: z.string().nullable(),
+    numeratorFilters: z.array(z.string()).nullable(),
```
lukesonnet: Sorry to be really annoying, but given how we're spreading these, should they be called "additionalNumeratorFilters"? Maybe that's overkill. Maybe a comment here is enough. I also think doing nothing is reasonable since `getMetricWithFiltersApplied` is clear enough.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -889,7 +890,8 @@ export default abstract class SqlIntegration
   }
 
   getMetricAnalysisQuery(params: MetricAnalysisParams): string {
-    const { metric, settings } = params;
+    const metric = getMetricWithFiltersApplied(params);
+    const { settings } = params;
```
lukesonnet, line 895: tiny tiny nit: it might be nice if it was impossible to get metric from `params` directly and instead have to pass it through the validator above....

Related, it might be better to do this directly in the MetricAnalysisQueryRunner. It makes stuff like this feel a bit better, and it's moving a way from a pattern we use too much in SqlIntegration. I feel like these methods have too much override type logic that should be handled by the runner.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
```diff
@@ -52,178 +70,373 @@ export default function MetricExplorerSettings({ block, setBlock }: Props) {
...
+                  </Flex>
+                </Flex>
+              </Text>
+            </div>
+          }
+          transitionTime={100}
+        >
+          <Box p="4" height="fit-content">
+            <Flex direction="column" gap="4">
+              <MultiSelectField
```
lukesonnet, line 370: I know we discussed this offline, and I don't think there's a good way to do it, but is there a way to indicate these are "additional" metric filters? A tooltip is the worst way to do this: "Row Filters specified here are combined with any existing filters on the metric.".

Fine if you want to override to keep it cleaner, I don't think it's a big deal.

packages/back-end/src/routers/metric-analysis/metric-analysis.validators.ts
mknowlton89: Not annoying at all. I think it's a valid suggestion.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
mknowlton89: I made two changes.

1) I did add a tooltip - while I wish I could come up with a different pattern, I didn't find any that were better.
2) I did update the empty state from `Add filters` to `Add additional filters` to make the empty state just a tad more contextual.

packages/back-end/src/integrations/SqlIntegration.ts
mknowlton89: @lukesonnet Just to confirm, you're thinking we update the signature of `getMetricAnalysisQuery` to the following, where we split up the metric and the params?

```ts
getMetricAnalysisQuery(
    metric: FactMetricInterface,
    params: Omit<MetricAnalysisParams, "metric">,
  ): string;
```

And the `metric` that would be passed in there would already contain the `numerator` and `denominator` filter arrays that have already had the `additionalNumeratorFilters` and `additionalDenominatorFilters` spread in?

I did all of this work in this [commit](https://github.com/growthbook/growthbook/pull/4782/commits/d0993c9e69dd8ea086c04324149d78e50b5bff7a). I think this is what you meant, but would love confirmation.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerSettings.tsx
```diff
@@ -52,178 +71,385 @@ export default function MetricExplorerSettings({ block, setBlock }: Props) {
...
+                      <Text as="label" size="3" weight="medium">
+                        Row Filters
+                      </Text>
+
+                      <Text as="span" size="2" mb="2">
+                        {" "}
+                        (Denominator)
+                      </Text>
+                      <Tooltip
+                        body="Row Filters specified here are combined with any existing denominatorfilters defined on the metric."
```
Kevin-Chant:
```suggestion
                        body="Row Filters specified here are combined with any existing denominator filters defined on the metric."
```

PR #4792, opened 2025-11-04
improve feature rule UI

bryce-fitzsimons:
- Standardize feature rule display
- Add useful links when editing saved groups

image

image

### DIFF
--- a/packages/front-end/components/Features/ExperimentRefSummary.tsx
+++ b/packages/front-end/components/Features/ExperimentRefSummary.tsx
@@ -284,26 +284,26 @@ export default function ExperimentRefSummary({
             </Table>
           </Box>
           <Box mt="3">
-          {!isBandit && (
-            <ExperimentSplitVisual
-              values={experiment.variations.map((variation, j) => {
-                return {
-                  name: variation.name,
-                  value:
-                    variations.find((v) => v.variationId === variation.id)
-                      ?.value ?? "null",
-                  weight: phase.variationWeights?.[j] || 0,
-                };
-              })}
-              coverage={effectiveCoverage}
-              label="Traffic split"
-              unallocated="Not included (skips this rule)"
-              type={type}
-              showValues={false}
-              stackLeft={true}
-              showPercentages={true}
-            />
-          )}
+            {!isBandit && (
+              <ExperimentSplitVisual
+                values={experiment.variations.map((variation, j) => {
+                  return {
+                    name: variation.name,
+                    value:
+                      variations.find((v) => v.variationId === variation.id)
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -441,6 +443,24 @@ export default function ConditionInput(props: Props) {
...
+                          group?.type === "list"
+                            ? `/saved-groups/${group.id}`
+                            : "/saved-groups#conditionGroups";
+                        return (
+                          <Link
+                            href={link}
+                            target="_blank"
+                            style={{ position: "relative", zIndex: 1000 }}
+                          >
+                            {o.label} <PiArrowSquareOut />
```
msamper, line 460: This feels a bit cluttered on the UI because the `PiArrowSquareOut` icon appears next to the icon to delete the group, but I also understand wanting to make it clear it's a link. I think its fine for now, but I'm sure Kari would have some ideas

packages/front-end/components/Features/ExperimentRefSummary.tsx
```diff
@@ -139,158 +139,183 @@ export default function ExperimentRefSummary({
...
+                label="Traffic split"
+                unallocated="Not included (skips this rule)"
+                type={type}
+                showValues={false}
+                stackLeft={true}
+                showPercentages={true}
+              />
             </Box>
+          )}
+          <Flex direction="row" gap="2" mb="3">
```
msamper, line 309: I think this may need some top margin for non bandit experiments

packages/front-end/components/Features/SavedGroupTargetingDisplay.tsx
```diff
@@ -30,31 +31,38 @@ export default function SavedGroupTargetingDisplay({
...
+              {s.ids.length > 1 && "("}
               {s.ids.map((id) => {
                 const group = getSavedGroupById(id);
                 const link =
                   group?.type === "list"
                     ? `/saved-groups/${group.id}`
                     : "/saved-groups#conditionGroups";
                 return (
-                  <Link
+                  <LinkButton
```
msamper: I don't know that this should be a `LinkButton` but it's already styled to look like a button today so this is fine for now. I feel like it goes against the patterns we've been following with links as of late in the app. Maybe it's more jarring now because it's purple

packages/front-end/components/Features/ConditionInput.tsx
bryce-fitzsimons: agree, we can follow up during a proper redesign

PR #4794, opened 2025-11-05
Separate query results collection

jdorn:
### Background

Our current `queries` Mongo collection is a bottleneck for how much data we can return from a single SQL query.  Because of inefficient data structures and Mongo document size limitations, we only support query results that are less than ~1MB.

That means for experiments with many metrics, we currently have to split across multiple smaller queries, each of which scan the same source data.  This can lead to 5x the query cost or more in some situations.

### Changes

This PR changes the `queries` collection to just hold metadata about the query and introduces a new optimized collection to hold the actual results.
- Results are encoded in a more space efficient data structure. Some quick testing shows 70% reduction in raw data size (33% after compression).  
- Results are automatically split across multiple Mongo documents if needed so we can scale way beyond single-document limits.

In addition, this PR adds an in-memory cache layer in the QueryRunner class because once a query succeeds, the results are indempotent and do not need to be fetched again.  This should drastically reduce the amount of data fetched from Mongo during experiment refreshes.

### Future Work

This PR alone will not reduce warehouse query costs.  After this lands, we can improve our fact table optimization logic to include more metrics per query.

### DIFF
--- a/packages/back-end/test/queryRunners/QueryRunner.test.ts
+++ b/packages/back-end/test/queryRunners/QueryRunner.test.ts
@@ -6,7 +6,7 @@ import {
 import { QueryInterface, QueryStatus } from "back-end/types/query";
 import { SourceIntegrationInterface } from "back-end/src/types/Integration";
 import { updateQuery } from "back-end/src/models/QueryModel";
-import { ReqContext } from "back-end/types/organization";
+import { ReqContext } from "back-end/types/request";
 
 jest.mock("back-end/src/models/QueryModel");
 
@@ -44,11 +44,19 @@ class TestQueryRunner extends QueryRunner<
 
   async executeQuery(
     query: QueryInterface,
-    run: unknown,
-    process: unknown,
-    onFailure: unknown,
+    {
+      run,
+      process,
+      onFailure,
+      onSuccess,
+    }: {
+      run: unknown;
+      process?: unknown;
+      onFailure: unknown;
+      onSuccess?: unknown;
+    },
   ) {
-    this.executeQuerySpy(query, run, process, onFailure);
+    this.executeQuerySpy(query, { run, process, onFailure, onSuccess });
     // Don't actually execute for tests
     return Promise.resolve();
   }
@@ -164,25 +172,31 @@ describe("QueryRunner", () => {
       // Query A should NOT execute (has timer)
       expect(runner.executeQuerySpy).not.toHaveBeenCalledWith(
         expect.objectContaining({ id: "qry_A" }),
-        expect.any(Function),
-        expect.any(Function),
-        expect.any(Function),
+        expect.objectContaining({
+          run: expect.any(Function),
+          process: expect.any(Function),
+          onFailure: expect.any(Function),
+        }),
       );
 
       // Query B SHOULD execute (no timer, no dependencies)
       expect(runner.executeQuerySpy).toHaveBeenCalledWith(
         expect.objectContaining({ id: "qry_B" }),
-        expect.any(Function),
-        expect.any(Function),
-        expect.any(Function),
+        expect.objectContaining({
+          run: expect.any(Function),
+          process: expect.any(Function),
+          onFailure: expect.any(Function),
+        }),
       );
 
       // Query C SHOULD execute (no timer, no dependencies)
       expect(runner.executeQuerySpy).toHaveBeenCalledWith(
         expect.objectContaining({ id: "qry_C" }),
-        expect.any(Function),
-        expect.any(Function),
-        expect.any(Function),
+        expect.objectContaining({
+          run: expect.any(Function),
+          process: expect.any(Function),
+          onFailure: expect.any(Function),
+        }),
       );
 
       // Clean up timer
@@ -221,9 +235,11 @@ describe("QueryRunner", () => {
       // Query A should NOT execute (dependency is still running)
       expect(runner.executeQuerySpy).not.toHaveBeenCalledWith(
         expect.objectContaining({ id: "qry_A" }),
-        expect.any(Function),
-        expect.any(Function),
-        expect.any(Function),
+        expect.objectContaining({
+          run: expect.any(Function),
+          process: expect.any(Function),
+          onFailure: expect.any(Function),
+        }),
       );
     });
 
@@ -258,6 +274,7 @@ describe("QueryRunner", () => {
 
       // Query A should be marked as failed
       expect(updateQuery).toHaveBeenCalledWith(
+        mockContext,
         expect.objectContaining({ id: "qry_A" }),
         expect.objectContaining({
           status: "failed",
@@ -268,9 +285,12 @@ describe("QueryRunner", () => {
       // Query A should NOT execute
       expect(runner.executeQuerySpy).not.toHaveBeenCalledWith(
         expect.objectContaining({ id: "qry_A" }),
-        expect.any(Function),
-        expect.any(Function),
-        expect.any(Function),
+        expect.objectContaining({
+          run: expect.any(Function),
+          process: expect.any(Function),
+          onFailure: expect.any(Function),
+          onSuccess: expect.any(Function),
+        }),
       );
     });
 
@@ -306,9 +326,11 @@ describe("QueryRunner", () => {
       // Query A SHOULD execute (dependency succeeded)
       expect(runner.executeQuerySpy).toHaveBeenCalledWith(
         expect.objectContaining({ id: "qry_A" }),
-        expect.any(Function),
-        expect.any(Function),
-        expect.any(Function),
+        expect.objectContaining({
+          run: expect.any(Function),
+          process: expect.any(Function),
+          onFailure: expect.any(Function),
+        }),
       );
     });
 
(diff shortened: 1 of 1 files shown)


packages/back-end/src/models/QueryModel.ts
```diff
@@ -102,9 +124,29 @@ export async function countRunningQueries(
...
+  context: ReqContext,
   query: QueryInterface,
   changes: Partial<QueryInterface>,
 ): Promise<QueryInterface> {
+  if (query.organization !== context.org.id) {
+    throw new Error("Cannot update query from different organization");
+  }
+
+  // If we're setting results, store them in a separate collection
+  // Some legacy queries have processed results that differ from raw results, so skip those
```
lukesonnet, line 128: Is this true only for legacy queries? We also have a meaningful `process` statement in the past experiments query runner: https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/queryRunners/PastExperimentsQueryRunner.ts#L44

So this becomes "if you use `process` when it changes `rows`, then you can't use this other collection". I feel like this should be more explicit somehow.

packages/shared/src/sql.ts
```diff
@@ -182,3 +183,90 @@ function stripCommentsAndStrings(sql: string): {
...
+type SqlResultChunkData = Pick<
+  SqlResultChunkInterface,
+  "columns" | "numRows" | "data"
+>;
+
+export function encodeSQLResults(
+  // Raw SQL results
+  results: Record<string, unknown>[],
+  // 4MB default chunk size (document max is 16MB, but leave plenty of room for overhead)
+  chunkSize: number = 4000000,
```
lukesonnet: So this isn't a limit, since each chunk can have up to 4000000-1 and then whatever the size of the next row is, right? I understand it isn't a big issue given the buffer, but want to make sure I understand.

packages/back-end/src/models/QueryModel.ts
jdorn: yes, I agree.  I want to get rid of the `process` option entirely.  That logic really shouldn't live in the query itself.  Can maybe do that in a followup PR

packages/shared/src/sql.ts
jdorn: yes, that is correct.  A single row should be pretty small, even with thousands of columns so I'm not worried about it.  Especially given the very conservative limits in the code

packages/back-end/src/models/QueryModel.ts
lukesonnet: Ok, we use it extensively in the incremental refresh PR for side-effects, but we want to move that to an onSuccess method instead, like we added an onFailure method.

jdorn: Created an `onSuccess` method and deprecated `process`.  There are still 3 places that use `process`, but we can deal with them later.

packages/shared/src/sql.ts
```diff
@@ -191,3 +192,90 @@ function stripCommentsAndStrings(sql: string): {
...
+): SqlResultChunkData[] {
```
ahdriel, line 204: nit: Can we add the unit to the prop name?


```suggestion
  // 4MB default chunk size (document max is 16MB, but leave plenty of room for overhead)
  chunkSizeBytes: number = 4_000_000,
): SqlResultChunkData[] {
```

packages/shared/src/sql.ts
```diff
@@ -191,3 +192,90 @@ function stripCommentsAndStrings(sql: string): {
...
+  }
+
+  function getSize(value: unknown): number {
+    if (value === null || value === undefined) return 1;
+    if (typeof value === "boolean") return 1;
+    if (typeof value === "number") return 8;
+    if (typeof value === "string") return value.length + 5;
+
+    // Each nested field has overhead, so just multiply by 2 to be extra conservative
+    return 1 + JSON.stringify(value).length * 2;
```
ahdriel, line 230: nit/issue: I don't think we have objects/arrays on many of the queries, but calling JSON.stringify for each row can potentially be super expensive.

I don't know a better solution at the moment though 🥲

packages/shared/src/sql.ts
```diff
@@ -191,3 +192,90 @@ function stripCommentsAndStrings(sql: string): {
...
+    encodedResults.push(currentChunk);
+  }
+
+  return encodedResults;
+}
+
+export function decodeSQLResults(
+  chunks: SqlResultChunkData[],
+): Record<string, unknown>[] {
+  const results: Record<string, unknown>[] = [];
```
ahdriel, line 262: nit(performance): If we know the size of the array we could pre-allocate all of it here instead of relying on the runtime to resize it as we keep adding more data.

I am not sure how the allocation happens though given that an array of 400 null objects certainly takes less memory than a populated array of the same size.

packages/back-end/src/models/QueryModel.ts
```diff
@@ -60,10 +61,31 @@ function toInterface(doc: QueryDocument): QueryInterface {
...
+  const docs = await QueryModel.find({
+    organization: context.org.id,
+    id: { $in: ids },
+  });
+  const queries = docs.map((doc) => toInterface(doc));
+
+  if (includeResults) {
+    for (const query of queries) {
+      if (query.hasChunkedResults) {
+        const result = await context.models.sqlResultChunks.getResultsByQueryId(
```
ahdriel: question: Should we fetch all of the chunks at once from Mongo instead of doing 1 query per queryId?

I don't know if we have additional memory considerations (if Mongo has a max size for a response) or not, but we could avoid additional trips to the db.

packages/back-end/src/models/QueryModel.ts
```diff
@@ -60,10 +61,31 @@ function toInterface(doc: QueryDocument): QueryInterface {
   return omit(ret, ["__v", "_id"]);
 }
 
-export async function getQueriesByIds(organization: string, ids: string[]) {
+export async function getQueriesByIds(
+  context: ReqContext,
+  ids: string[],
+  includeResults: boolean = true,
```
ahdriel: nit: This is semi-true, as we will still return the results if the query has `hasChunkedResults: false`

packages/back-end/src/models/QueryModel.ts
```diff
@@ -102,9 +124,29 @@ export async function countRunningQueries(
...
+  }
+
+  // If we're setting results, store them in a separate collection
+  // Some legacy queries have processed results that differ from raw results, so skip those
+  if (
+    changes.result &&
+    changes.result === changes.rawResult &&
+    changes.rawResult.length > 0
+  ) {
+    await context.models.sqlResultChunks.createFromResults(
```
ahdriel, line 134: question: Does this mean that we potentially can have multiple chunks of the same data associated with the Query?

- Create query with results -- those get added to sqlResultChunk
- Update query with a new result array -- it would create new sqlResultChunks instead of updating the already existing ones

I think this might be less of a problem in our code given how queries are executed and updated, but it is a possibility.

Or it'll fail if this happens for whatever reason given the unique index requirement.

packages/back-end/src/validators/queries.ts
```diff
@@ -15,3 +15,17 @@ export const queryPointerValidator = z
...
+export const sqlResultChunkValidator = z
+  .object({
+    organization: z.string(),
+    dateCreated: z.date(),
+    dateUpdated: z.date(),
+    id: z.string(),
+    queryId: z.string(),
+    chunkNumber: z.number(),
+    numRows: z.number(),
+    columns: z.array(z.string()),
```
ahdriel: nit/question: Why are we saving the column names too?

We could either use Object.keys on the `data` object, or if we are keeping the columns array, not use a record for `data` and instead rely on the index to reconstruct the results.

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -59,7 +59,9 @@ export type StartQueryParams<Rows, ProcessedRows> = {
     query: string,
     setExternalId: ExternalIdCallback,
   ) => Promise<QueryResponse<Rows>>;
-  process: (rows: Rows) => ProcessedRows;
+  /** @deprecated */
```
ahdriel, line 59: Love to see it!

packages/shared/src/sql.ts
jdorn: JSON.stringify is pretty fast (parsing is the slow part)

packages/shared/src/sql.ts
jdorn: The max number of rows is a few thousand, so growing the array organically should be fine. V8 is very good at that.  I think this would start to matter a lot more once you start getting to hundreds of thousands of elements.

packages/back-end/src/models/QueryModel.ts
jdorn: Yeah, it will fail with a unique key constraint error.  Queries are immutable in our code base today (and I imagine they will stay that way), so this should never happen right now, but I agree it is a little bit fragile

packages/back-end/src/models/QueryModel.ts
```diff
@@ -60,10 +61,23 @@ function toInterface(doc: QueryDocument): QueryInterface {
   return omit(ret, ["__v", "_id"]);
 }
 
-export async function getQueriesByIds(organization: string, ids: string[]) {
+export async function getQueriesByIds(
+  context: ReqContext,
+  ids: string[],
+  includeChunkedResults: boolean = true,
```
lukesonnet, line 67: Is there an argument for us to return different types based on whether it is a chunked query without its chunked results or whether it is a chunked query WITH results so that consumers of this method and its types aren't confused about whether the data is attached?

PR #4796, opened 2025-11-05
Streamline Experiment Dashboard creation flow

Kevin-Chant:
### Features and Changes

Product Analytics dashboards are launching with a simplified creation flow which puts the user directly into the edit mode and asks for the dashboard metadata on save. This PR applies those same changes to Experiment Dashboards as well to reduce friction

This also fixes a minor bug where general dashboards on creation weren't showing the `shareLevel` or `editLevel` selectors

### Testing

- [x] Editing an existing experiment dashboard shouldn't prompt a modal on hitting "Done editing"
- [x] Creating a dashboard then clicking "Exit without saving" before making changes shouldn't POST the new dashboard
- [x] Creating a dashboard and adding a block should POST the dashboard and enable the snapshot provider
- [x] Creating a dashboard and adding a block should prompt the DashboardModal on "Done editing"
- [x] The metadata for each field of that modal should be saved to the new dashboard
- [x] Editing the dashboard again after saving should not prompt the modal a second time
- [x] The above should still apply to general dashboards

### Screenshots

image

image

image

image

### DIFF
--- a/packages/front-end/pages/product-analytics/dashboards/new.tsx
+++ b/packages/front-end/pages/product-analytics/dashboards/new.tsx
@@ -79,6 +79,7 @@ export default function NewDashboardPage() {
             body: JSON.stringify({
               blocks: args.data.blocks,
               title: args.data.title ?? dashboard.title,
+              shareLevel: args.data.shareLevel ?? dashboard.shareLevel,
               editLevel: args.data.editLevel ?? dashboard.editLevel,
               enableAutoUpdates:
                 args.data.enableAutoUpdates ?? dashboard.enableAutoUpdates,


packages/front-end/enterprise/components/Dashboards/DashboardsTab.tsx
```diff
@@ -238,8 +262,15 @@ function DashboardsTab({
           dashboard={dashboard}
           submitDashboard={submitDashboard}
           mutate={mutateDashboards}
-          close={() => setIsEditing(false)}
+          close={() => {
+            setIsEditing(false);
+            setDashboardFirstSave(false);
+            if (dashboardId === "new") {
+              setDashboardId(dashboard.id === "new" ? "" : dashboard.id);
```
Kevin-Chant, line 269: This check is because the `dashboardId` in state is no longer exactly `dashboard.id`. Instead it's being used similar to the `/new` url path to show that we're in the initial creation of a new dashboard, but the underlying `dashboard.id` is updated as soon as that dashboard is persisted to the db (so if it's still `new`, it was never persisted)

PR #4798, opened 2025-11-05
Fixes dimension group bug when switching from bar or area to other ch…

mknowlton89:
### Features and Changes

While dogfooding the SqlExplorerModal via the Product Analytics Dashboards, we came across a bug when switching from 1 graph type to another, with regards to how we're handling dimensions.

The `dimension` object has a `display` property that can be either `grouped` or `stacked` for `bar` and `area` charts, but for all other chart types, we only support `grouped`.

We found a bug where if a user had a bar or area chart and had a dimension that had the display set to `stacked`, if the user switched to a chart type that didn't support `stacked` we weren't properly handling that.

This PR adds a new helper to normalize a `dimension.display` property to account for this.

Now, on change, it will reset to `grouped` if the chart type doesn't support `stacked`.

I did also update the validators so only `bar` and `area` charts can have stacked dimensions.

### Testing
- [x] Ensure you can still save a bar or area chart with a dimension that has the display set to stacked
- [x] Ensure that if you have an existing bar or area chart with a dimension that has the display set to `stacked`, switching to a different chart resets the display to `grouped`.
- [x] Ensure this doesn't introduce any regressions with creating charts without dimensions or anything else.
- [x] Ensure you can still save a saved-query after these changes, and the blocks are also valid for a dashboard after these changes and can be saved.


packages/front-end/services/dataVizConfigUtilities.ts
```diff
@@ -75,3 +75,41 @@ export function updateXAxisConfig(
 
   return setXAxisConfig(config, updated);
 }
+
+// Not all chart types support stacked dimensions, so this function ensures only bar and area can have a "stacked" display
+export function normalizeDimensionsForChartType(
```
lukesonnet, line 80: I wonder if this can be generalized to normalize all values with some general configuration validation service, but just an idea for later, perhaps.

mknowlton89: This is a great idea.

PR #4801, opened 2025-11-05
feat(incremental): Add experiment-level enable/disable

ahdriel:
### Features and Changes

To facilitate onboarding with IncrementalRefresh, we want to ensure older experiments keep working as expected. So instead of throwing if they use any settings that are not supported by Incremental Refresh, we will automatically fallback to the regular ExperimentResultsQueryRunner.

Additionally, if an error is thrown for an Experiment in the Incremental Refresh pipeline, we provide an easy way to exclude the experiment (and re-enable it if wanted) from incremental, so we don't block users.

### Screenshots

Screenshot 2025-11-05 at 1 43 51 PM

Screenshot 2025-11-05 at 1 52 38 PM

### DIFF
--- a/packages/back-end/src/services/dataPipeline.ts
+++ b/packages/back-end/src/services/dataPipeline.ts
@@ -1,0 +1,163 @@
+import {
+  ExperimentMetricInterface,
+  isFactMetric,
+  isRatioMetric,
+  quantileMetricType,
+} from "shared/experiments";
+import { ExperimentSnapshotSettings } from "back-end/types/experiment-snapshot";
+import { FactTableMap } from "back-end/src/models/FactTableModel";
+import { IncrementalRefreshInterface } from "back-end/src/validators/incremental-refresh";
+import { OrganizationInterface } from "back-end/types/organization";
+import { orgHasPremiumFeature } from "back-end/src/enterprise";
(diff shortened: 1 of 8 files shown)


packages/back-end/src/services/dataPipeline.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (
+    (settings.pipelineSettings?.includedExperimentIds !== undefined &&
+      !settings.pipelineSettings?.includedExperimentIds.includes(
+        experiment.id,
+      )) ||
+    (settings.pipelineSettings?.excludedExperimentIds !== undefined &&
+      settings.pipelineSettings?.excludedExperimentIds.includes(experiment.id))
+  ) {
+    throw new Error(
+      "Experiment is not included in the Pipeline Incremental scope",
```
lukesonnet, line 84: This is a bit cryptic, but I guess this error shouldn't really ever be shown unless we made a mistake.

lukesonnet (commented):
Right now won't this throw errors for many existing experiments but keep falling back to the old runner?

I thought wanted to not even throw errors in the case of old experiments and instead grandfather them in with no errors unless they explicitly start adding them to the included experiment ids?

Then for new experiments I thought wanted to only fall back if they explicitly opted in to the old runner?

PR #4804, opened 2025-11-05
feat(incremental): Add support for efficient dimension analysis

lukesonnet:
### Features and Changes

Adds a separate `ExperimentIncrementalRefreshExploratoryQueryRunner` to be used for ad-hoc queries on top of existing incremental refresh intermediate tables.

Uses that class to allow unit, experiment, and date cohort dimension queries to leverage the existing intermediate data.

Eventually we can use that query runner to also do days since exposed time series on demand.

### Testing

* [X] unit query joins correctly
* [X] existing experiment query joins correctly
* [X] new exposure query dimensions don't show up if they aren't in the exposure query model
* [X] adding a new metric just skips that metric
* [X] removing a metric ignores that removed metric
* [X] ~dashboard dimensions pick the right query runner~ THEY DONT, there is a dependency we need to introduce, and instead I'm blocking for V0
* [ ] changes to core experiment settings throw an error

### DIFF
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/index.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/index.tsx
@@ -21,7 +21,7 @@ import {
   DASHBOARD_WORKSPACE_NAV_HEIGHT,
 } from "@/enterprise/components/Dashboards/DashboardWorkspace";
 import Button from "@/ui/Button";
-import useExperimentalRefreshMode from "@/hooks/useExperimentalRefreshMode";
+import useExperimentPipelineMode from "@/hooks/useExperimentPipelineMode";
 import { BLOCK_SUBGROUPS, BLOCK_TYPE_INFO, isBlockTypeAllowed } from "..";
 import EditSingleBlock from "./EditSingleBlock";
 
@@ -96,7 +96,7 @@ export default function DashboardEditorSidebar({
   >(undefined);
 
   // TODO(incremental-refresh): remove when dimensions supported in dashboard
-  const experimentalRefreshMode = useExperimentalRefreshMode(
+  const experimentalRefreshMode = useExperimentPipelineMode(
     experiment ?? undefined,
   );
 
--- a/packages/front-end/enterprise/components/Dashboards/DashboardWorkspace.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardWorkspace.tsx
@@ -18,7 +18,7 @@ import { cloneDeep, pick } from "lodash";
 import { CREATE_BLOCK_TYPE, getBlockData } from "shared/enterprise";
 import { isDefined } from "shared/util";
 
-import useExperimentalRefreshMode from "@/hooks/useExperimentalRefreshMode";
+import useExperimentPipelineMode from "@/hooks/useExperimentPipelineMode";
 import Button from "@/ui/Button";
 import Link from "@/ui/Link";
 import Tooltip from "@/components/Tooltip/Tooltip";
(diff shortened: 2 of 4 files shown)


packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -479,6 +489,7 @@ export const startExperimentIncrementalRefreshQueries = async (
             unitsMaxTimestamp: maxTimestamp,
             experimentSettingsHash:
               getExperimentSettingsHashForIncrementalRefresh(snapshotSettings),
+            unitsDimensions: experimentDimensions.map((d) => d.id),
```
ahdriel, line 423: question: Should we store it with the `type` or is the `string` dimension.id enough here?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -7479,15 +7466,94 @@ ${this.selectStarLimit("__topValues ORDER BY count DESC", limit)}
...
 
+    const { unitDimensions } = this.processDimensions(
+      params.dimensions,
+      params.settings,
+      params.activationMetric,
+    );
+
+    const idTypeObjects = [
+      [exposureQuery.userIdType],
+      ...unitDimensions.map((d) => [d.dimension.userIdType || "user_id"]),
```
ahdriel: question: Why do we fallback to `user_id` by default here?

And is there a problem if we have multiple of the same value like `['user_id', 'user_id']`

lukesonnet: Because we do it that way elsewhere. Probably a throwback to a old version where it was an empty string or something? We could remove it almost certainly.

packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
lukesonnet: We're always going to cast to string (since we do that at the end anyways).

packages/front-end/hooks/useExperimentalRefreshMode.ts
```diff
@@ -0,0 +1,20 @@
+import { ExperimentInterfaceStringDates } from "back-end/types/experiment";
+
+import { getIsExperimentIncludedInIncrementalRefresh } from "@/services/experiments";
+import { useDefinitions } from "@/services/DefinitionsContext";
+
+// This hook returns list of projects user has permission for along with any projects already associated with a resource (e.g. metric.projects)
```
ahdriel: nit: This comment is not correct, right?

packages/front-end/hooks/useExperimentalRefreshMode.ts
```diff
@@ -0,0 +1,20 @@
...
+export default function useExperimentalRefreshMode(
```
ahdriel: question: Do we have `Experimental` because we expect the RefreshMode to go away / remove this hook?

Otherwise I would vote to call it `useExperimentRefreshMode` or even `useExperimentPipelineMode` (regular vs incremental)?

packages/front-end/pages/product-analytics/dashboards/[did].tsx
```diff
@@ -177,6 +177,7 @@ function SingleDashboardPage() {
             initialShareLevel={dashboard.shareLevel}
             dashboardOwnerId={dashboard.userId}
             isGeneralDashboard={true}
+            isIncrementalRefreshExperiment={false}
```
Comment (ahdriel) [inline packages/front-end/pages/product-analytics/dashboards/[did].tsx:180]: question: Do we expect this to change? And is tied to the `generalDashboard` prop above it?

Comment (lukesonnet) [reply packages/front-end/pages/product-analytics/dashboards/[did].tsx:180]: Yes, I considered joining them, but I hope this is temporary.

packages/front-end/hooks/useExperimentalRefreshMode.ts
lukesonnet: oops

PR #4819, opened 2025-11-07
Adds survey link to cancel flow

mknowlton89:
### Features and Changes

This PR adds a new modal to the `/billing` page for organizations after they cancel their Orb subscription. This modal takes users to a Typeform survey to ask them why they cancelled. We're appending the user's email address to the url so we can track a user's response.

### Dependencies

- None

### Testing

- [x] Start a new Orb subscription, and then after you click the "Cancel" CTA, ensure this modal fires.
- [x] Ensure that the org is being updated like it is today, when a user cancels, so the user doesn't get into a weird state where the subscription is canceled, but the application doesn't reflect that

### Screenshots
Screenshot 2025-11-10 at 7 09 44 AM

### DIFF
--- a/charts/growthbook/Chart.lock
+++ b/charts/growthbook/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: mongodb
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 16.5.1
+  version: 18.1.9
 - name: server
   repository: ""
   version: 0.0.0
 - name: server
   repository: ""
   version: 0.0.0
-digest: sha256:4d1eabf4f8723300ab9bc325ae11fe395cc1de97e3cf32c6c7540a3af3a70ace
-generated: "2025-04-17T11:49:04.396553326+02:00"
+digest: sha256:81fec7a9e7863e315f4347e877ef61d7a8af8a94d8d2f3549dce0810d79ac4ce
+generated: "2025-11-14T10:32:04.916227+01:00"
--- a/charts/growthbook/Chart.yaml
+++ b/charts/growthbook/Chart.yaml
@@ -22,7 +22,7 @@ appVersion: "4.1.0"
 dependencies:
   - name: mongodb
     repository: oci://registry-1.docker.io/bitnamicharts
-    version: "^16.5.0"
+    version: "^18.0.0"
     condition: mongodb.enabled
   - name: server
     version: "0.0.0"
--- a/packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
+++ b/packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
@@ -64,28 +64,42 @@ const STARTER_BILLING_PLAN: BillingPlan = {
   ],
 };
 
-const PRO_BILLING_PLAN: BillingPlan = {
-  description: "Full featured experimentation and growth platform",
-  id: "pro-billing-plan",
-  name: "Pro Plan",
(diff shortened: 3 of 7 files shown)


packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -46,6 +50,34 @@ export default function SubscriptionInfo() {
...
         />
       )}
+      {showCancellationSurveyModal && (
+        <Modal
+          open={true}
+          header={null}
+          trackingEventModalType="cancellation-survey"
+          close={() => setShowCancellationSurveyModal(false)}
+          submit={async () => {
+            const surveyUrl = new URL("https://form.typeform.com/to/kL75SA6F");
```
jdorn: I don't love hard-coding this.

packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -46,6 +50,34 @@ export default function SubscriptionInfo() {
...
+            if (email) {
+              surveyUrl.searchParams.set("email", email);
+            }
```
jdorn, line 66: It feels like we might need to disclose that the user's email will be included in the feedback?  Especially for self-hosted users.

packages/front-end/components/Settings/SubscriptionInfo.tsx
mknowlton89: I'll create a const for this.

packages/front-end/components/Settings/SubscriptionInfo.tsx
mknowlton89: Probably a good idea.

One thing to call out, this will only be visible to Cloud orgs canceling their Orb subscription. Stripe/self-hosted orgs manage their subscription via the Stripe portal.

I can add the caveat to the modal (example below), but I'm wondering if we just want to make it an optional field on the survey instead. I would expect we'd receive more responses if we did that (as opposed to auto-collecting it, and including the disclosure statement).

Screenshot 2025-11-12 at 11 59 27 AM

Kevin-Chant: I'd definitely lean towards it being an optional field that gets pre-populated cause I agree it would get more responses while also feeling (and being) more privacy-conscious.
Is it intentional that right now it's not including the email at all? (Did you decide that orgId is sufficient?)

packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -46,6 +52,36 @@ export default function SubscriptionInfo() {
...
+            if (organization.id) {
+              surveyUrl.searchParams.set("org_id", organization.id);
+            }
+
+            window.open(surveyUrl.toString(), "_blank");
+            setShowCancellationSurveyModal(false);
+          }}
+          cta="Share Feedback"
+          closeCta="No thanks"
+          showHeaderCloseButton={false}
```
Kevin-Chant, line 73: I'm not sure how much this would drop the response rate but personally I get way more annoyed by modals without the X in the top right. Just feels more pushy, but take it or leave it whether you change this

packages/front-end/components/Settings/SubscriptionInfo.tsx
mknowlton89: @Kevin-Chant Yea - we decided to not collect email address, and instead just pass the orgId instead to avoid any potential issues.

Kevin-Chant (approved):

PR #4822, opened 2025-11-10
Add helpers for throwing controller errors

Kevin-Chant:
### Features and Changes

Currently we have a broadly used `context.permissions.throwPermissionError` helper which automatically sets the response status code and filters the errors from being reported to Sentry. It seemed useful to add similar helpers for the other common error cases, i.e. Bad Request, Unauthorized, Not Found, and Internal Server Error

### Testing

Verify that each type of error is still treated correctly via the network inspector and frontend behavior. This required some temporary frontend changes to allow the breaking behavior
- [x] Bad request
- [x] Unauthorized
- [x] Not found
- [x] Internal server error

#### Bad request

(Behavior is identical before/after)

image

#### Unauthorized error

Before
image

After
image

#### Not found

Before
image

After

image

#### Internal server error

Before
image

After

image

### DIFF
--- a/packages/back-end/src/routers/metric-group/metric-group.controller.ts
+++ b/packages/back-end/src/routers/metric-group/metric-group.controller.ts
@@ -72,7 +72,7 @@ export const putMetricGroup = async (
     );
   }
   if (org.id !== metricGroup.organization) {
-    context.throwUnauthorizedError(
+    context.permissions.throwPermissionError(
       "You don't have access to that metric group",
     );
   }
@@ -139,7 +139,7 @@ export const putMetricGroupReorder = async (
     context.permissions.throwPermissionError();
   }
   if (metricGroup.organization !== context.org.id) {
-    context.throwUnauthorizedError(
+    context.permissions.throwPermissionError(
       "You don't have access to that metric group",
     );
   }
@@ -177,7 +177,7 @@ export const removeMetricFromGroup = async (
     context.permissions.throwPermissionError();
   }
   if (metricGroup.organization !== context.org.id) {
-    context.throwUnauthorizedError(
+    context.permissions.throwPermissionError(
       "You don't have access to that metric group",
     );
   }
--- a/packages/shared/src/util/errors.ts
+++ b/packages/shared/src/util/errors.ts
@@ -10,84 +10,10 @@ export class ErrorWrapper extends Error {
   }
 }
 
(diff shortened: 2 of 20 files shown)


packages/shared/src/util/errors.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export class UnauthorizedError extends Error {
+  status = 403;
+  constructor(message: string) {
+    super(message);
+    this.name = "UnauthorizedError";
+  }
+}
+
+export class ErrorNotFound extends Error {
```
tzjames: For consistent naming and to math `this.name` this could be `NotFoundError`

packages/shared/src/util/errors.ts
```diff
@@ -0,0 +1,93 @@
...
+  status: number;
+
+  constructor(message: string, status: number) {
+    super(message);
+    this.status = status;
+    this.name = "LicenseServerError";
+  }
+}
+
+export class VercelIntallationNotFound extends Error {}
```
tzjames: typo.  It should be `VercelInstallationNotFound`.

packages/shared/src/util/errors.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export class BadRequestError extends Error {
+  status = 400;
+  constructor(message: string) {
+    super(message);
+    this.name = "BadRequestError";
+  }
+}
+
+export class UnauthorizedError extends Error {
```
tzjames: This could be renamed `ForbiddenError` as 403 is Forbidden whereas 401 is Unauthorized.

packages/back-end/src/routers/metric-group/metric-group.controller.ts
```diff
@@ -67,10 +67,14 @@ export const putMetricGroup = async (
...
   const metricGroup = await context.models.metricGroups.getById(req.params.id);
   if (!metricGroup) {
-    throw new Error("Could not find metric group with that id");
+    return context.throwNotFoundError(
+      "Could not find metric group with that id",
+    );
   }
   if (org.id !== metricGroup.organization) {
-    throw new Error("You don't have access to that metric group");
+    context.throwUnauthorizedError(
```
tzjames: This should maybe a permissionError?  I guess they are both 403 right now.

PR #4823, opened 2025-11-10
Remove the mongo model for dashboard blocks

Kevin-Chant:
### Features and Changes

We realized that the `DashboardModel` isn't actually linked to the `DashboardBlockModel` in Mongo so while blocks are persisted to the db in the creation flow, during updates the changes only make it to the dashboard.

This PR removes the model which is mostly vestigial at this point, and moves the relevant helpers to the dashboard level instead.

### Testing

Go through the lifecycle of a dashboard to confirm that the behavior still matches the expectation
- [x] Create a new dashboard with a block
- [x] Modify the details of that block
- [x] Add a second block
- [x] Delete the first block
- [x] Rename the dashboard & change some metadata
- [x] Add a new block again
- [x] Delete the dashboard

### DIFF
--- a/packages/back-end/src/routers/dashboards/dashboards.controller.ts
+++ b/packages/back-end/src/routers/dashboards/dashboards.controller.ts
@@ -162,12 +162,10 @@ export async function updateDashboard(
 
   if (updates.blocks) {
     const migratedBlocks = updates.blocks.map(migrateBlock);
-    const createdBlocks = await Promise.all(
-      migratedBlocks.map((blockData) =>
-        dashboardBlockHasIds(blockData)
-          ? blockData
-          : generateDashboardBlockIds(context.org.id, blockData),
-      ),
+    const createdBlocks = migratedBlocks.map((blockData) =>
+      dashboardBlockHasIds(blockData)
+        ? blockData
+        : generateDashboardBlockIds(context.org.id, blockData),
     );
     updates.blocks = createdBlocks;
   }


packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -288,3 +290,92 @@ function getSavedQueryIds(doc: DashboardDocument): Set<string> {
...
+
+export function createDashboardBlock(
+  organization: string,
+  initialValue: CreateDashboardBlockInterface,
+): DashboardBlockInterface {
+  const block = {
+    ...initialValue,
+    organization,
+    id: uniqid("dshblk_"),
+    uid: uuidv4().replace(/-/g, ""),
```
Kevin-Chant, line 306: The ID is still necessary for referencing blocks, and the UID might be used in the future for permalinking to a single block (which is possible via an [embedded index](https://www.mongodb.com/docs/manual/core/indexes/index-types/index-multikey/create-multikey-index-embedded/))

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -107,44 +107,18 @@ export async function createDashboard(
   } = req.body;
 
   if (experimentId) {
-    // Quick permission check before we write to the dashboard block collection
```
Kevin-Chant, line 110: These checks were only necessary due to the blocks being written to the collection before the dashboard's internal `canCreate`/`canUpdate` helpers had a chance to run. The logic is still present [here](https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/enterprise/models/DashboardModel.ts#L106)

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -187,32 +161,12 @@ export async function updateDashboard(
...
-        if (!context.hasPremiumFeature("product-analytics-dashboards")) {
-          throw new Error(
-            "Your plan does not support updating private dashboards.",
-          );
-        }
-      }
-    }
-    const migratedBlocks = updates.blocks.map((block) => migrate(block));
+    const migratedBlocks = updates.blocks.map(migrateBlock);
     const createdBlocks = await Promise.all(
```
mknowlton89: Do we still need this `await Promise.all()` call here since `gnerateDashboardBlockIds` is sychronous.

Kevin-Chant: No, I missed this instance of it thanks

mknowlton89 (approved):
Tested with existing Dashboard's that had existing blocks and also created new dashboards/blocks without issue.

LGTM

PR #4829, opened 2025-11-11
Backend proxy option for Statsig / LaunchDarkly importer (self-hosted only)

bryce-fitzsimons:
Adds a backend proxy option to route requests to Statsig / LaunchDarkly REST APIs. Needed when GB is installed without HTTPS to avoid CORS / HTTPS downgrade security issues.

- Self-hosted only (blocked on FE and BE for cloud users)
- general pass-through proxy endpoints

### DIFF
--- a/packages/front-end/components/importing/ImportFromLaunchDarkly/ImportFromLaunchDarkly.tsx
+++ b/packages/front-end/components/importing/ImportFromLaunchDarkly/ImportFromLaunchDarkly.tsx
@@ -144,6 +144,9 @@ async function buildImportedData(
   existingEnvs: Set<string>,
   features: FeatureInterface[],
   callback: (data: ImportData) => void,
+  useBackendProxy: boolean = false,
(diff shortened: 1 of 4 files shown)


august-growthbook (approved):

packages/back-end/src/routers/importing/importing.controller.ts
```diff
@@ -0,0 +1,134 @@
...
+      status: 400,
+      message: "Missing required fields: endpoint and apiKey",
+    });
+  }
+
+  // Validate HTTP method
+  const validatedMethod = validateHttpMethod(method);
+
+  try {
+    const url = `https://statsigapi.net/console/v1/${endpoint}`;
```
Kevin-Chant, line 58: Do we need any validation of `endpoint` or permission checks for the user before fetching the urls?

bryce-fitzsimons: I don’t think so. Self hosted only, and we don’t check permissions for the non-proxied version of this call either

PR #4845, opened 2025-11-13
Add automatic CRUD API endpoint generation for BaseModel

Kevin-Chant:
### Features and Changes

Creates a new, lighter weight method of creating CRUD API endpoints for objects that use `BaseModel`. The new `ApiModel` is initially implemented only by the `DashboardModel` but can be extended to the other standard models in future PRs.

Docs are automatically generated for both the CRUD actions as well as any `customHandlers` defined for the model

### Testing

Use your favorite tool to hit each of the new endpoints for dashboards and verify the behavior.

image

Check the generated docs to make sure the model and all endpoints are included properly
image

### DIFF
--- a/packages/back-end/src/models/BaseModel.ts
+++ b/packages/back-end/src/models/BaseModel.ts
@@ -267,7 +267,7 @@ export abstract class BaseModel<
       z.ZodTypeAny,
       z.ZodTypeAny
     >,
-  ): Promise<z.infer<ApiT> | null> {
+  ): Promise<z.infer<ApiT>> {
     const id = req.params.id;
     const doc = await this.getById(id);
     if (!doc) throw new Error("Not found");
@@ -275,7 +275,7 @@ export abstract class BaseModel<
   }
   protected async handleApiCreate(
     req: ApiRequest<unknown, z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>,
-  ): Promise<z.infer<ApiT> | null> {
+  ): Promise<z.infer<ApiT>> {
     const rawBody = req.body;
     const toCreate = await this.processApiCreateBody(rawBody);
     return this.toApiInterface(await this.create(toCreate));
@@ -309,7 +309,7 @@ export abstract class BaseModel<
       z.ZodType<UpdateProps<z.infer<T>>>,
       z.ZodTypeAny
     >,
-  ): Promise<z.infer<ApiT> | null> {
+  ): Promise<z.infer<ApiT>> {
     const id = req.params.id;
     const rawBody = req.body;
     const toUpdate = await this.processApiUpdateBody(rawBody);


packages/shared/types/openapi.d.ts
```diff
@@ -371,12 +371,191 @@ export interface paths {
     /** Get list of custom fields */
     get: operations["getCustomFields"];
   };
+  "/dashboards/:id": {
+    get: operations["getDashboard"];
+    post: operations["updateDashboard"];
+    delete: operations["deleteDashboard"];
+  };
+  "/dashboards/": {
```
ahdriel: nit/question: Should this be `/dashboards` only? Does it make any difference?

packages/shared/types/openapi.d.ts
```diff
@@ -371,12 +371,191 @@ export interface paths {
     /** Get list of custom fields */
     get: operations["getCustomFields"];
   };
+  "/dashboards/:id": {
```
ahdriel: question/issue: All other urls have the parameters defined with curly braces (like `/code-refs/{id}`).

1) Is this an issue for the OpenAPI spec?
2) Should we keep consistency and use `{id}` here too?

packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -315,6 +375,71 @@ export class DashboardModel extends BaseClass {
...
   }
+
+  public toApiInterface(dashboard: DashboardInterface): ApiDashboardInterface {
+    return {
+      ...removeMongooseFields(dashboard),
+      blocks: dashboard.blocks.map(toBlockApiInterface),
+      dateCreated: dashboard.dateCreated.toISOString(),
+      dateUpdated: dashboard.dateUpdated.toISOString(),
+      nextUpdate: dashboard.nextUpdate?.toISOString(),
+      lastUpdated: dashboard.lastUpdated?.toISOString(),
```
ahdriel, line 385: question: Should this be the default for our API? Is it not already?

'This' being sending all dates as ISOStrings.

packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -315,6 +375,71 @@ export class DashboardModel extends BaseClass {
...
+    } = apiCreateDashboardBody.parse(rawBody);
+    const createdBlocks = await Promise.all(
+      blocks.map((blockData) =>
+        generateDashboardBlockIds(
+          this.context.org.id,
+          fromBlockApiInterface(blockData),
+        ),
+      ),
+    );
+    return {
```
ahdriel, line 408: issue: There are some differences in the `POST Create` and `GET Dashboard`, mainly on how we are handling `null` values.

In the dashboard case, it applies to `experimentId` and `projects` for example.

Sending a POST to create a new dashboard with this payload
```
{
  "editLevel": "private",
  "shareLevel": "published",
  "enableAutoUpdates": false,
  "title": "Test Dashboard",
  "blocks": []
}
```

Returns this

```
{
  "dashboard": {
    "id": "dash_fzty71x4dmk4ywr8x",
    "uid": "b67aa9740b214f5489073650f1124c87",
    "isDefault": false,
    "isDeleted": false,
    "userId": "",
    "editLevel": "private",
    "shareLevel": "published",
    "enableAutoUpdates": false,
    "title": "Untitled Dashboard",
    "blocks": [],
    "organization": "org_fzty723m0m8qbqj6y",
    "dateCreated": "2026-01-08T04:49:16.785Z",
    "dateUpdated": "2026-01-08T04:49:16.785Z"
  }
}
```

This response does not have `null` fields, `projects` and `experimentId`.

But GET /dashboards/dash_fzty71x4dmk4ywr8x returns in the API response.

Should we have both of them be treated the same way and return the same values?

ahdriel: One additional thing is that because `GET dashboards/:id` returns `projects: null`, I am unable to get a dashboard, modify the object and then POST it to be updated.

This gives me an error saying that `projects` cannot be `null`.

I do think we should accept `null` for projects to remove it. Or should it be an empty array instead?

ahdriel: More broadly, the above does not work because we complain about many keys that we don't allow modification (id, uid, isDefault, etc...).

Maybe we should also accept those in the request if the values match of the current object?

I am still thinking of the flow Fetch object -> modify -> post to be updated.

packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -50,6 +62,38 @@ const BaseClass = MakeModelClass({
...
+    modelPlural: "dashboards",
+    apiInterface: apiDashboardInterface,
+    schemas: {
+      createBody: apiCreateDashboardBody,
+      updateBody: apiUpdateDashboardBody,
+    },
+    includeDefaultCrud: true,
+    customHandlers: [
+      {
+        pathFragment: "/by-experiment/:experimentId",
```
ahdriel, line 78: nit: This is not throwing a `404` if the experiment id is invalid. Should it?

packages/back-end/src/enterprise/models/DashboardModel.ts
```diff
@@ -50,6 +62,38 @@ const BaseClass = MakeModelClass({
...
+        verb: "get",
+        operationId: "getDashboardsForExperiment",
+        validator: apiGetDashboardsForExperimentValidator,
+        zodReturnObject: apiGetDashboardsForExperimentReturn,
+        reqHandler: async (
+          req: ApiRequestForValidator<
+            typeof apiGetDashboardsForExperimentValidator,
+            ApiGetDashboardsForExperimentReturn
+          >,
+        ): Promise<ApiGetDashboardsForExperimentReturn> => ({
```
ahdriel, line 86: nit: Would be nice to figure out a way to reduce the typing boilerplate and still have the type-safety.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -239,12 +254,76 @@ export abstract class BaseModel<
...
+    req: ApiRequest<
+      unknown,
+      z.ZodType<{ id: string }>,
+      z.ZodTypeAny,
+      z.ZodTypeAny
+    >,
+  ): Promise<z.infer<ApiT> | null> {
+    const id = req.params.id;
+    const doc = await this.getById(id);
+    if (!doc) throw new Error("Not found");
```
ahdriel, line 273: issue: This is returning a response code 400 when it should be 404. Can we fix it?

packages/back-end/src/models/BaseModel.ts
```diff
@@ -239,12 +254,76 @@ export abstract class BaseModel<
...
+  }
+
   /***************
    * These methods are implemented by the MakeModelClass helper function
    ***************/
-  protected abstract getConfig(): ModelConfig<T, E>;
+  protected abstract getConfig(): ModelConfig<T, E, ApiT>;
   protected abstract getCreateValidator(): CreateZodObject<T>;
   protected abstract getUpdateValidator(): UpdateZodObject<T>;
+  public static getModelConfig() {
```
ahdriel, line 330: question: Why not make it `abstract` if we need it to be implemented by a derived class?

packages/back-end/src/models/BaseModel.ts
```diff
@@ -853,20 +932,28 @@ export abstract class BaseModel<
...
     E,
+    ApiT,
     WriteOptions
   > {
     getConfig() {
       return config;
     }
+    static getModelConfig(): ModelConfig<T, E, ApiT> {
+      return config;
+    }
```
ahdriel, line 984: question: Do we need both `getConfig` and `getModelConfig`?

packages/back-end/src/api/ApiModel.ts
```diff
@@ -0,0 +1,294 @@
...
+  includeDefaultCrud?: boolean;
+  crudActions?: CrudAction[];
+  crudValidatorOverrides?: Partial<CrudValidatorShapes<T>>;
+  customHandlers?: CustomHandler[];
+};
+type ApiModel = {
+  pathBase: string;
+  modelClass: ModelClass;
+};
+export const apiModels = [
```
ahdriel: This is a very important configuration now -- any way we can make it more explicit? Right now it feels a bit lost in the middle of this file.

packages/shared/types/openapi.d.ts
ahdriel: Same applies to the spec.yaml file

packages/back-end/generated/spec.yaml
```diff
@@ -5951,6 +5954,4106 @@ paths:
...
+      $skipValidatorGeneration: true
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      tags:
+        - Dashboards
+      summary: ''
```
ahdriel: question: Is there a place that we can define the summary?

Not sure if it is needed right now, but might be valuable in the future.

packages/back-end/generated/spec.yaml
```diff
@@ -5951,6 +5954,4106 @@ paths:
...
+      operationId: getDashboard
+      x-codeSamples: []
```
ahdriel: question: Should we add `curl` samples as all other endpoints have?

packages/back-end/src/services/context.ts
```diff
@@ -65,35 +65,68 @@ export type ForeignRefTypes = {
...
+  | "metricTimeSeries"
+  | "webhookSecrets"
+  | "holdout"
+  | "dashboards"
+  | "customHooks"
+  | "incrementalRefresh"
+  | "sqlResultChunks"
+  | "sdkConnectionCache";
+
+export const modelClasses = {
```
ahdriel, line 95: nit: Should this be typed with `Record<ModelName, BaseClass>` or something similar?

packages/shared/types/openapi.d.ts
ahdriel: I am talking about the trailing `/`

packages/back-end/src/scripts/generate-doc-models-for-openapi-schemas.ts
```diff
@@ -0,0 +1,218 @@
...
+    Object.values(loadedApiDoc.paths).some((path) => !isValidPathRecord(path))
+  )
+    return false;
+  return true;
+}
+
+function getOrCreatePathRecord(api: ApiShape, fullPath: string, verb: string) {
+  if (api.paths[fullPath] && "$ref" in api.paths[fullPath])
+    throw new Error(
+      "Unable to add API route at '${verb}' ${fullPath}; this path has a $ref defined",
```
ahdriel: issue: This should use backticks to interpolate `verb` and `fullPath`.

packages/back-end/src/api/ApiModel.ts
```diff
@@ -0,0 +1,294 @@
...
+    verb: "get",
+    pathFragment: "/",
+    plural: true,
+  },
+  delete: {
+    verb: "delete",
+    pathFragment: "/:id",
+  },
+  update: {
+    verb: "post",
```
ahdriel: The old discussion... should this be `PUT`, `PATCH` or `POST`?

packages/shared/types/openapi.d.ts
Kevin-Chant: Shouldn't make a difference but I'm changing it cause I agree with having it match precedent

packages/shared/types/openapi.d.ts
Kevin-Chant: Not sure if it actually causes issues, but I'm changing it to format `:id` to `{id}` during `generate-api-types`. (It has to be `:id` for the definition to work for express)

packages/back-end/src/enterprise/models/DashboardModel.ts
Kevin-Chant: It's the expected behavior but requires manually adding the logic in a `toApiInterface`. In this PR I added that helper to `BaseModel`, but for models like this one which need to override it (due to other fields) then you end up with this extra boilerplate. Definitely open to suggestions if you have a way to simplify things

packages/back-end/src/enterprise/models/DashboardModel.ts
Kevin-Chant: It could, but I'm leaning towards leaving it as-is because that way we don't need to actually pull the experiment from the DB. Lmk if you feel strongly that it should

packages/back-end/src/models/BaseModel.ts
Kevin-Chant: Thanks for flagging. I have some error helpers for 403, 404, 500, etc in #4822. I can swap this to use those once that's merged

packages/back-end/src/models/BaseModel.ts
Kevin-Chant: I actually tried that first but it turns out that Javascript doesn't allow `abstract static` methods. It doesn't need to be implemented by devs though, it's added automatically by the call to `MakeModelClass()` so it's only undefined if calling `BaseClass` directly (which I don't think we would ever do)

packages/back-end/src/models/BaseModel.ts
Kevin-Chant: Note that `getModelConfig` is `static` - I added it because in the router setup context we don't have instances of the model classes yet but still need to access their config

packages/back-end/src/api/ApiModel.ts
Kevin-Chant: I'm moving it to the top with the `CONST_NAME` format. I tried a couple ways to make it automatic when `BaseModel` is extended but couldn't quite get that working. Hopefully this is simple enough for devs to find

packages/back-end/src/services/context.ts
Kevin-Chant: I think it should but I'm not sure how to make that work. `BaseModel` is generic and the classes actually extend an intermediary abstract `Model` rather than `BaseModel` so I got lost in the typescript sauce trying to type this object. (Also, there's a large risk of a circular dependency with how this is defined)

If you know how to get that typing working do let me know

packages/back-end/src/scripts/generate-doc-models-for-openapi-schemas.ts
Kevin-Chant: Whoops, good catch :sweat_smile:

packages/back-end/src/api/ApiModel.ts
Kevin-Chant: I saw we had a number of update endpoints using `POST` so I thought that was the precedent but now I see it's mixed. In that case I'll swap it to `PUT`

packages/back-end/src/enterprise/models/DashboardModel.ts
Kevin-Chant: The treatment of `null` by mongo is a bit annoying, and worries me that we'll have to remember to add some boilerplate to every model.

What's happening is that on the `POST`, we just return the doc back to the user rather than doing another roundtrip to fetch from mongo. Then on the `GET` it finally pulls those null fields for the first time, but it doesn't know to change them to `undefined` (which is valid) hence the "`projects` cannot be `null`". In `DashboardModel`'s `migrate` we have it do exactly this for update schedule: `updateSchedule: orig.updateSchedule || undefined,`

So I'm not sure how to avoid the current state of needing to add each of these optional-but-not-nullable keys to each model's `migrate` function.

Kevin-Chant: As to your second point, I think I agree in theory with allowing those extra keys (I've seen at least one use-case involving fetch -> modify -> post), but since the PUT is just a partial of the editable fields, they should only be including the fields they actually want to change rather than sending the whole object

packages/back-end/generated/spec.yaml
Kevin-Chant: I had trouble getting them to be detailed at all, but having something is probably better than nothing. Good shout

packages/back-end/src/enterprise/models/DashboardModel.ts
Kevin-Chant: Yeah I went back and forth on a bunch of directions here. I just changed it to use some inference based on the validators provided (but it still needs a return type annotation because of some issues with our base `ApiRequest` interface)

packages/back-end/src/models/BaseModel.ts
```diff
@@ -245,12 +260,76 @@ export abstract class BaseModel<
...
   }
 
+  protected async handleApiGet(
+    req: ApiRequest<
+      unknown,
+      z.ZodType<{ id: string }>,
+      z.ZodTypeAny,
+      z.ZodTypeAny
+    >,
+  ): Promise<z.infer<ApiT> | null> {
```
ahdriel: issue: This never returns `null`. It either throws or returns the model. Can we remove `| null`?

I think it also applies to `create` and `update`.

packages/back-end/src/enterprise/models/DashboardModel.ts
ahdriel: No strong feeling here. As it is custom we can leave it as-is for the dashboards by-experiment route.

packages/back-end/src/enterprise/models/DashboardModel.ts
ahdriel: For the latter point, should the method be PATCH then? 😂

For the first point, I am also not sure. Maybe we can rely on Zod? To add field back somehow? Not sure though.

ahdriel (approved):

PR #4848, opened 2025-11-13
Experiment results filters

bryce-fitzsimons:
1. Introduces filters to experiment results
- filter by tag (refactor to make less crappy)
- filter by metric group or individual metric
- filter by slice

2. Refreshes the general UI around experiment results (and a few other general improvements)

Does not remove slice pinning. Does not affect experiment dashboards. Improvements there will be part of a follow-up effort.


packages/front-end/components/Experiment/CompactResults.tsx
```diff
@@ -497,56 +613,62 @@ export function getRenderLabelColumn({
...
-                  size={14}
-                  className={isPinned ? "link-purple" : "text-muted opacity50"}
-                  onClick={() => {
-                    if (togglePinnedMetricSlice && row?.sliceLevels) {
-                      togglePinnedMetricSlice(
-                        metric.id,
-                        row.sliceLevels,
-                        location || "goal",
-                      );
+          {(!sliceTagsFilter || sliceTagsFilter.length === 0) && (
```
bryce-fitzsimons, line 637: Holding off on removing pinning until we have a decent mechanism to persist filters to a dashboard...

packages/front-end/components/Experiment/ResultMoreMenu.tsx
```diff
@@ -1,28 +1,84 @@
...
+  DropdownMenuItem,
+  DropdownMenuSeparator,
+} from "@/ui/DropdownMenu";
+import AsyncQueriesModal from "@/components/Queries/AsyncQueriesModal";
 import usePermissionsUtil from "@/hooks/usePermissionsUtils";
-import ConfirmDialog from "@/ui/ConfirmDialog";
 import { getIsExperimentIncludedInIncrementalRefresh } from "@/services/experiments";
 import { useDefinitions } from "@/services/DefinitionsContext";
+import Badge from "@/ui/Badge";
+
```
bryce-fitzsimons, line 24: This file should probably have all of these utilities (CSV generator, notebook link generator, etc) broken out into a service rather than inline in the component...

Since this PR is already pretty large, maybe something to fast follow on?

packages/front-end/hooks/useExperimentTableRows.ts
```diff
@@ -390,33 +454,57 @@ export function generateRowsForMetric({
 
   numSlices = sliceData.length;
 
+  // If slice filter is active and metric has no slices, don't show parent row
+  if (sliceTagsFilter && sliceTagsFilter.length > 0 && numSlices === 0) {
+    return [];
+  }
+
+  // When slice filters are active and metric has slices, make parent row label-only
```
bryce-fitzsimons: Are there any circumstances wherein we want to show metric-level row stats when slice filters are set?
@lukebrawleysmith @lukesonnet

packages/front-end/components/Experiment/ResultMoreMenu.tsx
lukesonnet: Yes, for sure. Fast follow seems reasonable.

packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -603,7 +603,7 @@ const AnalysisForm: FC<{
           allowUndefined={!isBandit}
           disabled={isBandit}
         />
-        {isBandit && (
+        {!isHoldout && (
```
lukesonnet, line 606: Why is this off for holdouts?

bryce-fitzsimons: originally we only set CUPED in this form for bandits, but now we're expanding it to experiments (formerly these had their own toggle switch on results table). Only holdouts should be excluded.

lukesonnet: Holdouts can have cuped. Is there somewhere else CUPED is being set for holdouts?

bryce-fitzsimons: we will address cuped + holdouts separately (no regression in this PR, but would be good to support it)

PR #4850, opened 2025-11-14
feat(auth): Add Kerberos support for Trino/Presto

### LINKED ISSUE #4773: [Feature] Kerberos authentication support for Presto/Trino
### Description of Feature

Growthbook currently supports Basic and Custom Auth to connect to Presto/Trino (cf https://docs.growthbook.io/warehouses/prestodb-or-trino). However, some organizations rely on Kerberos to authenticate access. 

Supporting Kerberos would entail something akin to defining a principal, as well as a keytab location or a kerberos cache file location ([usually](https://web.mit.edu/kerberos/krb5-1.12/doc/admin/env_variables.html) exposed via the `KRB5CCNAME` environment variable).

Libraries such as https://www.npmjs.com/package/krb5 can be used to perform the heavy lifting.

I'm opening this issue to probe whether they'd be traction to add such support.  

Thank you!

ahdriel:
Closes #4773

### Features and Changes

Add support for Kerberos authentication for Trino/Presto data sources.

Authentication can be configured by setting the common env variables like `KRB5_CONFIG`, `KRB5_CLIENT_KTNAME` or `KRB5CCNAME`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ WORKDIR /usr/local/src/app
 # Set node max memory
 ENV NODE_OPTIONS="--max-old-space-size=8192"
 RUN apt-get update && \
-  apt-get install -y wget gnupg2 build-essential ca-certificates libkrb5-dev && \
+  apt-get install -y wget gnupg2 build-essential ca-certificates && \
   mkdir -p /etc/apt/keyrings && \
   wget -qO- https://deb.nodesource.com/gpgkey/nodesource.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && \
   echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x buster main" > /etc/apt/sources.list.d/nodesource.list && \
--- a/packages/back-end/src/util/kerberos.util.ts
+++ b/packages/back-end/src/util/kerberos.util.ts
@@ -5,11 +5,19 @@ import {
   GSS_MECH_OID_KRB5,
 } from "kerberos";
 import { logger } from "./logger";
+import { IS_CLOUD } from "./secrets";
 
 export async function getKerberosHeader(
   servicePrincipal: string,
   clientPrincipal?: string,
 ): Promise<string> {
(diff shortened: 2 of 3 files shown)


packages/back-end/src/integrations/Presto.ts
```diff
@@ -55,6 +56,18 @@ export default class Presto extends SqlIntegration {
     if (this.params?.authType === "customAuth") {
       configOptions.custom_auth = this.params.customAuth || "";
     }
+    if (this.params?.authType === "kerberos") {
+      const servicePrincipal = this.params.kerberosServicePrincipal;
+      const clientPrincipal = this.params.kerberosClientPrincipal;
+      if (!servicePrincipal || !clientPrincipal) {
+        throw new Error(
+          "Kerberos service and client principals are required for Kerberos authentication",
```
brouberol: I think we can relax this, and allow for no kerberos client principal to be passed. For example, when a single principal exists in the keytab, the SPNEGO negotiation will use it by default.

packages/back-end/src/util/kerberos.util.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export async function getKerberosHeader(
+  servicePrincipal: string,
+  clientPrincipal: string,
+): Promise<string> {
+  const startTime = performance.now();
+  const formattedServicePrincipal = formatServicePrincipal(servicePrincipal);
+  const client = await initializeClient(formattedServicePrincipal, {
+    mechOID: GSS_MECH_OID_KRB5,
+    principal: clientPrincipal,
```
brouberol: If we make `clientPrincipal` optional, possibly only pass it if non-nil.

Dockerfile
```diff
@@ -18,7 +18,7 @@ WORKDIR /usr/local/src/app
 # Set node max memory
 ENV NODE_OPTIONS="--max-old-space-size=8192"
 RUN apt-get update && \
-  apt-get install -y wget gnupg2 build-essential ca-certificates && \
+  apt-get install -y wget gnupg2 build-essential ca-certificates libkrb5-dev && \
```
jdorn: Is this needed here?  I would think `libkrb5-dev` would only be a runtime dependency and not build-time

PR #4851, opened 2025-11-14
Add confirmation to deleting PA dashboard

Kevin-Chant:
### Features and Changes

Adds a confirmation modal before allowing deletion of Product Analytics Dashboards.

The implementation is a general confirmation step for `DropdownMenuItem` since the existing `DeleteButton` doesn't play nicely with Radix dropdowns.

### Testing

Create and then delete a dashboard

### Screenshots

image

### DIFF
--- a/packages/front-end/pages/product-analytics/dashboards/index.tsx
+++ b/packages/front-end/pages/product-analytics/dashboards/index.tsx
@@ -468,7 +468,6 @@ export default function DashboardsPage() {
                                             </span>
                                           ),
                                           cta: "Delete",
-                                          submitColor: "danger",
                                           submit: async () => {
                                             await apiCall(
                                               `/dashboards/${d.id}`,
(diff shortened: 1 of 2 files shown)


packages/front-end/pages/product-analytics/dashboards/index.tsx
```diff
@@ -461,11 +461,23 @@ export default function DashboardsPage() {
...
-                                          });
-                                          mutateDashboards();
+                                        confirmation={{
+                                          confirmationTitle: (
+                                            <span>
+                                              Delete Dashboard <i>{d.title}</i>?
+                                            </span>
+                                          ),
+                                          cta: "Delete",
+                                          submitColor: "danger",
```
mknowlton89: Tiny nit: I'm trying to think of use cases where we wouldn't want to use the `"danger"` color and I'm not coming up with any.

I wonder if we should make `submitColor` optional, and set the default to `"danger"`.

I might also suggest a change from `submitColor` to `submitType` that takes an enum of `"danger"`, `"warning"`, etc, since I might try to pass in "red" here, for example.

Note: I do see the `Modal` component uses `submitColor`, so I do see the argument for keeping it as is. In either case, I think making it optional and setting a default would probably be a good dev ex.

PR #4858, opened 2025-11-17
Deploy to ECR as well

tzjames:
### Features and Changes
To lower the time of Fargate to download and start up images we will instruct fargate to load from ecr instead of DockerHub. This also avoids any issue of DockerHub rate limiting us. This second step sends the image to ECR as well as DockerHub.  We also update the step to push the built images to s3 to pull from ECR instead of DockerHub.

The last step will actually update the container image to read from ECR instead of dockerhub.

### Dependencies
https://github.com/growthbook/terraform/pull/116

### Testing
See deploy succeeds.
Login to AWS and see image updated.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,7 +31,7 @@ jobs:
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
 
       - name: Configure AWS credentials for ECR
-        uses: aws-actions/configure-aws-credentials@v1
+        uses: aws-actions/configure-aws-credentials@v5
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -30,6 +30,17 @@ jobs:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
 
+      - name: Configure AWS credentials for ECR
+        uses: aws-actions/configure-aws-credentials@v1
```
ahdriel: nit: The latest release of `configure-aws-credentials` is v5.1.0

Is there a reason why we are pinning this to v1? Or should we attempt to use a more recent version?

PR #4864, opened 2025-11-18
Move UID generation to BaseModel

Kevin-Chant:
### Features and Changes

To make it easier to migrate existing models to BaseModel we should abstract away UID generation. This PR adds it to the object conditionally if `uid` is present in the schema, meaning `uid` will become another reserved property. I don't see any existing usage outside of this pattern so that should be a safe change to make.

### Testing

- [x] Create a BaseModel object with a `uid` field (currently only Dashboards) twice to check that it generates unique `uid`s
- [x] Create a BaseModel object without a `uid` field (e.g. Fact Metric) and check that no `uid` is stored on the document

### Screenshots

image

image


ahdriel (approved):
Nice!

PR #4867, opened 2025-11-19
Custom Hooks - Incremental Changes Option

jdorn:
### Features and Changes

Add a new option to custom hooks - "Incremental Changes Only".  This affects the behavior of the hook during update operations. If the same error was already present before the update, the hook will be skipped.

This is especially useful for enforcing rules that may be difficult to fix retroactively. For example, if you require all features to have at least one tag, enabling this option will prevent users from being blocked by existing features that violate this rule when they attempt to make unrelated changes.

There are also cases where you would NOT want to enable this option. For example, if you have a hook that prevents publishing changes to "locked" features, you would want this to run every time to ensure the lock is never bypassed.

### DIFF
--- a/docs/docs/experimentation-analysis/sql-templates.mdx
+++ b/docs/docs/experimentation-analysis/sql-templates.mdx
@@ -114,4 +114,4 @@ The `1=1` part is a common SQL trick to make appending additional `AND` conditio
 
 Similar to dates, you must add surrounding quotes yourself when using string Custom Fields (e.g. use `'{{customFields.region}}'` instead of `{{customFields.region}}`).
 
-:::note
+:::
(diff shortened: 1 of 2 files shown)


jdorn:
> Conditional approval. I'm sure you tested this, but could we get into a situation where 2 legitimate failures in a row in incremental mode lead to incorrectly bypassing the hook?

This is definitely possible.  Here's a simple example:

```js
// Throw if there are no tags
if (!feature.tags.length) throw new Error("tags required");

// Throw if there is no project
if (!feature.project) throw new Error("project required');
```

If a feature currently has no tags, you can bypass the 2nd part of the check and remove a project.  Since it will throw the same error (the tags check) before and after your change, it will skip the hook.

I'm not sure how big of a deal this is or how to fix it.  We support multiple hooks, so an easy workaround for the user is to split this into 2 single-purpose hooks (probably a best practice anyway).  Maybe I can just add a "Best Practices" section in the docs  for now and talk about this.

PR #4868, opened 2025-11-19
Take 2: Simplify UIThemeProvider + prevent background flash

ahdriel:
This reverts commit 26be1f4 (from PR #4738).

### Features and Changes

The original PR had to be reverted because our public pages are SSR, and window is not available, so it broke.

This is now fixed and it works for both public and non-public pages.

### DIFF
--- a/packages/front-end/services/AppearanceUIThemeProvider.tsx
+++ b/packages/front-end/services/AppearanceUIThemeProvider.tsx
@@ -118,13 +118,6 @@ export const AppearanceUIThemeProvider: FC<PropsWithChildren> = ({
     }
   }, [preferredTheme, systemTheme]);
 
-  useEffect(() => {
-    document.documentElement.classList.remove("light-theme", "dark-theme");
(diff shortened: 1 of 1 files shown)


packages/front-end/services/RadixTheme.tsx
```diff
@@ -14,7 +14,7 @@ export const RadixTheme: FC<RadixThemeProps> = ({ children, flip }) => {
     <Theme
       accentColor="violet"
       panelBackground="solid"
-      appearance={computedTheme}
+      {...(flip ? { appearance: computedTheme } : {})}
```
ahdriel, line 17: This is an important change, where before we were explicitly passing `appearance` which would default to `light` if localStorage was unavailable, now we rely on the html class `light-theme` and `dark-theme` that the Snippet manages.

bryce-fitzsimons (changes requested):
Private/logged in pages are looking pretty broken on my end
Screenshot 2025-11-19 at 2 41 02 PM

PR #4869, opened 2025-11-19
add break-all to diff viewer

bryce-fitzsimons:
before (was causing confusion with nested structures / JSON):
image

after:
image


packages/front-end/components/HistoryTable.tsx
ahdriel, line 58: question: Should we create a wrapper that has these props by default?

ahdriel (approved):

packages/front-end/components/HistoryTable.tsx
bryce-fitzsimons: could, but didn't think worth abstracting and imposing a new default style for something so small. Can revisit in the future..

PR #4871, opened 2025-11-20
fix(dark mode): Update css overrides to use dark-theme

ahdriel:
### Features and Changes

On #4868 I removed the `--theme` classes but I forgot we also use overrides with just `.dark`.

So now this fixes it, consolidating on `light-theme` and `dark-theme`.


gazzdingo (approved):
LGTM

PR #4875, opened 2025-11-20
Fix metric analysis dates not updating

Kevin-Chant:
### Features and Changes

Currently refreshing a dashboard with a metric analysis block simply re-runs the query on the original date range. This fix makes it recompute the date range based on the lookback days specified in the block config.

### Testing

- Create a dashboard with a metric analysis block
- Inspect the block's start and end date fields, e.g. in mongo `db.dashboards.find({"blocks.type": "metric-explorer"})`
- [x] Refreshing the dashboard directly should update the dates
- [x] Auto-update should update the dates

### DIFF
--- a/packages/back-end/src/queryRunners/MetricAnalysisQueryRunner.ts
+++ b/packages/back-end/src/queryRunners/MetricAnalysisQueryRunner.ts
@@ -89,7 +89,6 @@ export class MetricAnalysisQueryRunner extends QueryRunner<
   }): Promise<MetricAnalysisInterface> {
     const updates: Partial<MetricAnalysisInterface> = {
       queries,
-      runStarted,
       error,
       result,
       status:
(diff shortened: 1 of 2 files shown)


packages/back-end/src/queryRunners/MetricAnalysisQueryRunner.ts
```diff
@@ -99,6 +98,9 @@ export class MetricAnalysisQueryRunner extends QueryRunner<
             ? "error"
             : "success",
     };
+    if (runStarted) {
+      updates.runStarted = runStarted;
+    }
```
Kevin-Chant, line 103: `runStarted` was being set to `null` on metric analyses unintentionally by another call to `updateModel` which led to an incorrect datetime in the "Updated at" display for dashboards

PR #4876, opened 2025-11-20
Include Draft Experiments for all SDKs

madhuchavva:
### Features and Changes

Decouple `Include Draft Experiments` toggle from UI/Auto Experiments and show it for backend SDKs too.

### Screenshots

image


lucas-hild-knowunity:
Hey @madhuchavva, we are a paying Growthbook customer and would quite urgently need this feature for being able to test experiments before launching them. Do you have an ETA when this will be available?

packages/back-end/src/util/features.ts
```diff
@@ -419,10 +421,8 @@ export function getFeatureDefinition({
           const exp = experimentMap.get(r.experimentId);
           if (!exp) return null;
 
-          if (!includeExperimentInPayload(exp)) return null;
-
-          // Never include experiment drafts
-          if (exp.status === "draft") return null;
+          if (!includeExperimentInPayload(exp, [], includeDraftExperiments))
+            return null;
```
jdorn, line 425: For auto-experiments (visual editor, url redirect), when we include drafts in the response, they are marked with `status: "draft"`.  This stops them from running by default, but they can be enabled manually with either dev tools or via a URL querystring.

This PR adds draft experiments to features, but it behaves differently - the draft experiment rules will be active and run automatically.  They won't be disabled by default like they are for visual/redirect experiments.

This is technically a breaking change for anyone who currently has that `enable draft experiments` toggle on.  Unfortunately, we don't have any property in the feature definition that's the equivalent of `experiment.status`, so there's no easy way for us to match the behavior of visual editor experiments.

It's probably ok, but just wanted to point this out.

PR #4877, opened 2025-11-20
Options to include metadata (custom fields, tags, projects) in SDK Payload

bryce-fitzsimons:
image

image


packages/front-end/services/importing/statsig/util.ts
```diff
@@ -0,0 +1,338 @@
...
+          condition: "not_in",
+          column: "category",
+          values: ["a", "b", "c"],
+        },
+      ],
+      metricSourceName: "DummyMetricSource",
+      valueColumn: "event_value",
+    },
+  },
+];
```
jdorn, line 388: File added by mistake

packages/back-end/src/api/openapi/payload-schemas/PostSdkConnectionPayload.yaml
```diff
@@ -28,6 +28,8 @@ properties:
     type: boolean
   includeRuleIds:
     type: boolean
+  includeProjectName:
```
jdorn: Rename this to `includeProjectId` throughout since projects also have names, but that's not what is being included.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -827,13 +828,14 @@ export abstract class BaseModel<
       this._dangerousGetCollection()
         .createIndex(index.fields as { [key: string]: number }, {
           unique: !!index.unique,
+          sparse: !!index.sparse,
```
bryce-fitzsimons: hopefully this is compatible with other mongo-like flavors

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -66,12 +80,106 @@ export class ProjectModel extends BaseClass {
     return this.context.permissions.canDeleteProject(doc.id);
   }
 
-  protected migrate(doc: MigratedProject) {
-    const settings = {
-      ...(doc.settings || {}),
-    };
+  public async getById(id: string) {
```
bryce-fitzsimons: Wasn't sure if we should be generating this on request or not

bryce-fitzsimons: the thought was it won't affect payload delivery since sdk connections will not have "include project uid" flag set. however, it's possible but extremely unlikely that users could turn this on for connections without having fetched and upgraded the project model yet

packages/back-end/src/services/SDK_PAYLOAD_FLOW.md
```diff
@@ -0,0 +1,271 @@
+# SDK Payload Generation & Serving Flow
```
bryce-fitzsimons: Included this file because it's often a point of confusion. We can remove it if we don't want it in the codebase.

packages/back-end/src/api/openapi/payload-schemas/PostSdkConnectionPayload.yaml
```diff
@@ -28,6 +28,16 @@ properties:
     type: boolean
   includeRuleIds:
     type: boolean
+  includeProjectPublicId:
+    type: boolean
+  includeCustomFields:
+    type: array
+    items:
+      type: string
+  includeTags:
```
jdorn, line 37: People are going to think this is for filtering, not metadata.  Like "if I select tag A here, the payload will only includes features with tag A".

If someone does make this mistake, they will end up exposing internal tags in their SDK payload unintentionally (or exposing back-end features to the front-end), which could be pretty bad.

The safest option is to rename this and turn it into a boolean.  `includeTagsInPayload: true`.  When true, all tags are included.  I think this will cover 95% of the use cases for tags already.  If we need to add a way to limit which tags are included, we can do it later.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -122,4 +118,69 @@ export class ProjectModel extends BaseClass {
...
+    refreshSDKPayloadCache(
+      this.context as ReqContext | ApiReqContext,
+      payloadKeys,
+    ).catch((e) => {
+      logger.error(
+        e,
+        "Error refreshing SDK payload cache after project create",
+      );
+    });
+  }
```
jdorn: Do we need this `afterCreate` hook?  When a project is first created, it has no features or experiments.  It's not possible for it to affect any SDK connections yet.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -122,4 +118,69 @@ export class ProjectModel extends BaseClass {
...
+
+  protected async afterUpdate(
+    existing: ProjectInterface,
+    updates: Partial<ProjectInterface>,
+    newDoc: ProjectInterface,
+    writeOptions?: never,
+  ): Promise<void> {
+    await super.afterUpdate(existing, updates, newDoc, writeOptions);
+    // Refresh SDK payload cache for all environments
```
jdorn: Should we only do this if the `publicId` field changes?

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -122,4 +118,69 @@ export class ProjectModel extends BaseClass {
...
+        e,
+        "Error refreshing SDK payload cache after project update",
+      );
+    });
+  }
+
+  protected async afterDelete(
+    doc: ProjectInterface,
+    writeOptions?: never,
+  ): Promise<void> {
```
jdorn: I'm also not sure if we really need this afterDelete hook, but I'm fine leaving it in.  Typically, when a project is deleted, features inside the project would also be deleted, which would already trigger any relevant payload refreshes.

packages/front-end/pages/project/[pid].tsx
```diff
@@ -155,10 +156,20 @@ const ProjectPage: FC = () => {
...
-          <Flex align="center">
+          <Flex direction="column" align="start">
             <Heading size="7" as="h1">
               {p.name}
             </Heading>
+            <Flex gap="6" mb="4">
+              <Metadata
+                label="Public ID"
+                value={<code>{p.publicId || p.id}</code>}
+              />
```
Comment (jdorn) [inline packages/front-end/pages/project/[pid].tsx:167]: Not for this PR, but we should eventually update all of the REST endpoints that accept a projectId in a querystring to work with either internal or public ids.

packages/back-end/src/routers/demo-datasource-project/new-demo-datasource-project.controller.ts
```diff
@@ -222,9 +223,13 @@ export const postDemoDatasourceProject = async (
   }
 
   try {
+    const projectName = "Sample Data";
+    const basePublicId =
+      generateProjectPublicIdFromName(projectName) || "sample-data";
     const project = await context.models.projects.create({
```
jdorn: We do this so many different places.  Should we just add this logic to the `beforeCreate` hook in the project model?  If `publicId` is empty, generate one from the name.

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -370,6 +385,46 @@ export async function editSDKConnection(
...
+        ? connection.projects
+        : (await context.models.projects.getAll()).map((p) => p.id);
+    const payloadKeys = getPayloadKeysForAllEnvs(context, projectsToUse);
+    refreshSDKPayloadCache(context, payloadKeys).catch((e) => {
+      logger.error(
+        e,
+        "Error refreshing SDK payload cache after SDK connection tags update",
+      );
+    });
+  }
```
jdorn, line 409: These two blocks are identical. We should probably combine them.

bryce-fitzsimons:
@cursoragent this PR has fallen behind main significantly. Now payloads are built in a single additive pass and cached per SDK Connection (or API key), rather than doing a 2 step shared cache with "everything" that needs to be JIT scrubbed per SDK Connection. Want to take a stab at getting this PR into the right shape again?

PR #4878, opened 2025-11-21
feat(metrics): Add distinct dates column type

lukesonnet:
### Features and Changes

Adds a fact metric column type of "Distinct Dates". This column type counts the distinct number of dates a user did an action as designated by the existence of a row in the fact table that adds passes a set of (optional) conditional filters.

This can be used to, for example, count the number of distinct days a user was active during an experiment, or the number of distinct days a user did a given action during an experiment.

To achieve this, we add a new column type to `ColumnRef`.  To handle this in `SqlIntegration.ts` we consolidate some key methods in our main fact metric queries. Specifically, we update: `getAggregationMetadata` (used by incremental refresh experiment queries and by metric analysis queries), `getAggregateMetricColumn` (used by non-incremental refresh experiment queries), and `getFactMetricColumn`, used by all three queries.

```
    if (column === "$$distinctDates") {
      // Column here should be a date (a timestamp truncated to date)
      const reAggregationFunction = nullIfZero
        ? (column: string) => `NULLIF(COUNT(DISTINCT ${column}), 0)`
        : (column: string) => `COUNT(DISTINCT ${column})`;
      const fullAggregationFunction = nullIfZero
        ? (column: string) => `NULLIF(COUNT(DISTINCT ${column}), 0)`
        : (column: string) => `COUNT(DISTINCT ${column})`;
      return {
        intermediateDataType: "date",
        partialAggregationFunction: (column: string) => `MAX(${column})`,
        finalDataType: "integer",
        reAggregationFunction,
        fullAggregationFunction,
      };
    }
``` 

#### Open questions

* Should we consider an option where you can define when a "date" starts as an offset from the timestamp? That way instead of just taking whatever `date_trunc(timestamp)` uses to create a date, you could say that your "date" starts at XX:XX:XX.XXX time and we can use that as the date?
* Should we add the ability to create daily aggregate filters (maybe as a fast follow)? Allowing someone to count as DAU only if they fulfill a certain criteria on a given day?

### Testing

* [x] Test against existing snapshot for no changes, then add new column type to update snapshot
* [x] Test mean days active in PG in dev instance
* [x] Test ratio distinct dates in PG in dev instance
* [X] Test mean days active in BQ in dev instance
* [X] Test ratio distinct dates in BQ in dev instance
* [x] Test in incremental refresh mode in BQ
* [x] Test new metric example in docs

### Screenshots

Creation (looks similar for ratio or for unit quantile)
Screenshot 2025-11-21 at 5 41 45 PM

Metric page
Screenshot 2025-11-21 at 5 41 40 PM

On metric page metric analysis
Screenshot 2025-11-21 at 5 41 36 PM

results
Screenshot 2025-11-21 at 5 41 17 PM

In product analytics
Screenshot 2025-11-21 at 5 08 09 PM

### DIFF
--- a/docs/docs/metrics/metrics.mdx
+++ b/docs/docs/metrics/metrics.mdx
@@ -86,7 +86,7 @@ Mean Metrics are useful when there are more than 2 possible states a user can be
 For the metric value, there are 4 types of aggregations you can do:
 
 - `COUNT(*)` - a count of all rows per user in your Fact Table
-- `COUNT(DISTINCT DATE(timestamp))` - the distinct number of days user completed the qualifying actions in your fact table (e.g. for a count of active days per user)
+- `COUNT(DISTINCT DATE(timestamp))` - the distinct number of days user completed the qualifying actions in your fact table (e.g. for a count of active days per user, akin to daily active users [DAU] metrics)
 - `SUM(column)` - the sum per user of any numeric column in your Fact Table
 - `MAX(column)` - the max of a numeric column per user in your Fact Table
 - `COUNT(DISTINCT column)` - the distinct count of a string column per user in your Fact Table.
--- a/packages/back-end/src/models/FactMetricModel.ts
+++ b/packages/back-end/src/models/FactMetricModel.ts
@@ -230,13 +230,6 @@ export class FactMetricModel extends BaseClass {
         );
       }
     }
-    if (data.denominator?.column === "$$distinctDates") {
-      if (!metricSupportsDistinctDates) {
-        throw new Error(
-          "$$distinctDates is only supported for mean, ratio, and quantile metrics",
-        );
-      }
-    }
 
     // validate user filter
     if (


docs/docs/metrics/metrics.mdx
```diff
@@ -63,26 +63,35 @@ The default display currency is USD, but you can change it under **Settings** 
...
+You then set the retention window, which is the amount of time after the first experiment exposure that a user must do the qualifying actions. You can also use a conversion window to only look for a certain time period after the start of the retention window. For example, if you want to measure the retention of users who came back to your site in the second full week after the experiment exposure, you would set the retention window to 7 days and the conversion window to 7 days.
+
 ### Mean Metrics
 
 Mean Metrics are useful when there are more than 2 possible states a user can be in. For example, instead of just purchase/not purchase, you might care about the average number of orders a user makes, the average revenue you earned per user in an experiment, or a count of the distinct types of items a user purchased.
 
 For the metric value, there are 4 types of aggregations you can do:
 
 - `COUNT(*)` - a count of all rows per user in your Fact Table
+- `COUNT(DISTINCT DATE(timestamp))` - the distinct number of days user completed the qualifying actions in your fact table (e.g. for a count of active days per user)
```
jdorn: We should probably put the following exact text somewhere in the docs at least so it is searchable.

> Daily Active Users (DAU)

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -2134,6 +2153,7 @@ export default function FactMetricModal({
                     }
                     includeColumn={true}
                     setDatasource={setDatasource}
+                    includeDistinctDates={true}
```
jdorn, line 2156: We don't need this now, but this could also be pretty useful for proportion metrics as an aggregate filter.  Percent of users who were active for >3 days.

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -217,6 +217,27 @@ export class FactMetricModel extends BaseClass {
...
+        throw new Error(
+          "$$distinctDates is only supported for mean, ratio, and quantile metrics",
+        );
+      }
+    }
+    if (data.denominator?.column === "$$distinctDates") {
+      if (!metricSupportsDistinctDates) {
+        throw new Error(
+          "$$distinctDates is only supported for mean, ratio, and quantile metrics",
+        );
```
jdorn: Is this check needed?  If there's a denominator, that means it is a ratio metric and supported already.  I guess this is to future-proof things if we add more restrictive conditions to the supports logic above?

lukesonnet: I was thinking of future proofing and catching if there are some issues with removing the denominator, but I think i'll just remove it for now.

PR #4879, opened 2025-11-21
Ji/paginate audits

tzjames:
### Features and Changes

Audits weren't paginate.  [Some requests](https://us5.datadoghq.com/apm/traces?query=env%3Aprod%20service%3Aprod-api%20-resource_name%3A%28GET%20OR%20POST%20OR%20%22bound%20dispatch%22%20OR%20asyncUtilWrap%20OR%20router%29&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=true&messageDisplay=inline&my_code=enabled&profiling-timeline__selected_lines=876095071&profiling-timeline__stack_order=top-down&profiling-timeline__summary_tab=flame-graph&profiling-timeline__tb=1763636361082001000&profiling-timeline__tf_f_0=0&profiling-timeline__tf_f_1=40222659180&profiling-timeline__tf_us_0=0&profiling-timeline__tf_us_1=40222659180&query_translation_version=v0&shouldShowLegend=true&sort=desc&sort_by=%40duration&sort_order=desc&spanID=5385664019675719780&spanType=all&spanViewType=errors&storage=hot&timeHint=1763636401304&trace=AwAAAZqg69SY_KtROAAAABhBWnFnNi1YSUFBQWl1LUZ0c2xEQWlWQWUAAABdMTE5YWExMjItYmFmZi00YjcxLWI0YmMtZWM5Yjg2OTk1YjcyLXN5bnRoZXRpYy10aW1lLTE3NjM2MzI4MDAwMDAwMDAwMDBjLTE3NjM2Mzk5OTkyMzEwMDAwMDFvAAFUWA&trace-to-profile__filter_by=spanChild&traceID=691ef489000000001e50c8702b90da56&traceQuery=&view=spans&start=1763636117333&end=1763636641333&paused=true) for audits are so large that the request is taking 40s and then dying with "invalid string length".

This PR paginates the audits.  It is a bit tricky since there are two separate queries that it merges together.  We solve this by keeping a cursor of the date of the last audit.  The queries then only fetch starting from before that audit when fetching additional pages.

It also uses `removeMongooseFields` as that is way more performant.

Mongoose seems to do some magic where sometimes the second are it treats as options and sometimes as projections.  I added null for the projection because sometimes it was returning with an empty array.

### Testing

Go to a an experiment that has a lot of audits.  Go to Page 2, Go to Page 3.  See no more pages.  Go back to Page 2 and Page 1 and make sure the same set of audits are still showing.

Go to /saved-groups and click audit logs (this was not paginated for me).  Saved Groups is the only place which uses `getAllHistory`.  See the total and no paginations buttons.

Go to /activity see all watched activity. 

### Screenshots
Screenshot 2025-11-21 at 12 01 21 PM
Screenshot 2025-11-21 at 12 01 12 PM
Screenshot 2025-11-21 at 1 24 25 PM

### DIFF
--- a/packages/back-end/src/models/AuditModel.ts
+++ b/packages/back-end/src/models/AuditModel.ts
@@ -1,8 +1,11 @@
 import mongoose, { FilterQuery, QueryOptions } from "mongoose";
 import uniqid from "uniqid";
-import { omit } from "lodash";
 import { AuditInterface } from "back-end/types/audit";
 import { EntityType } from "back-end/src/types/Audit";
+import {
+  removeMongooseFields,
+  ToInterface,
+} from "back-end/src/util/mongo.util";
 
 const auditSchema = new mongoose.Schema({
   id: {
@@ -41,15 +44,8 @@ type AuditDocument = mongoose.Document & AuditInterface;
 
 const AuditModel = mongoose.model<AuditInterface>("Audit", auditSchema);
 
-/**
- * Convert the Mongo document to an AuditInterface, omitting Mongo default fields __v, _id
- * @param doc
- */
-const toInterface = (doc: AuditDocument): AuditInterface => {
-  return omit(doc.toJSON<AuditDocument>(), [
-    "__v",
-    "_id",
-  ]) as unknown as AuditInterface;
+const toInterface: ToInterface<AuditInterface> = (doc: AuditDocument) => {
+  return removeMongooseFields(doc);
 };
 
 export async function insertAudit(
@@ -111,6 +107,7 @@ export async function findAuditByEntity(
       "entity.id": id,
       ...customFilter,
     },
+    null,
     options,
   );
   return auditDocs.map((doc) => toInterface(doc));
@@ -132,6 +129,7 @@ export async function findAuditByEntityList(
       },
       ...customFilter,
     },
+    null,
     options,
   );
   return auditDocs.map((doc) => toInterface(doc));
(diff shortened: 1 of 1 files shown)


packages/front-end/components/HistoryTable.tsx
```diff
@@ -185,6 +196,10 @@ const HistoryTable: FC<{
     });
   }, [data?.events, showName, getSavedGroupById]);
 
+  const currentPage = cursors.length;
+  const totalPages = data ? Math.ceil(data.total / limit) : 0;
+  const hasMore = data ? currentPage < totalPages : false;
+
```
msamper, line 202: Could we do this logic within the endpoint so the front-end only needs to be concerned with requesting the next page? Or is the reason for doing this because of the complexity with the two query results being merged?

msamper (approved):
Lgtm! Didn't have an experiment with 50+ events so I wasn't able to test locally, but the code looks good to me. Just had one comment, but more of a question than anything

packages/front-end/components/HistoryTable.tsx
tzjames: Yes the backend only knows the date to continue from, not which page it is on.  It knows the total, but not how many it has seen, so can't easily figure out without another db query if it does have more.  However the front-end knows how many it has seen, and gets the total, so it is simple for it to calculate.

PR #4882, opened 2025-11-24
PA Dashboards - Slices V2

mknowlton89:
### Features and Changes

This PR updates the Metric Explorer to support Metric Slices (auto and custom).

On the front-end this required a few changes:

1. We had to introduce a new `MetricSlicesSection` of the Metric Explorer Settings panel, this allows orgs with the `metric-slices` commercialFeature (currently Enterprise) to add auto and custom slices to a Metric's visualization.

2. To support this, we had to update the `QueryMetricAnalysisRunner` so it would return the slice data required for powering the front-end, which then ultimately required updates to the `getMetricAnalysisQuery` function. Now, in addition to returning the historgram data, in the `dates` array of the results, we're including slice data per date.

3. And finally, we needed to introduce some additional controls to the front end that would allow an organization the ability to show/hide certain slice levels from the chart (e.g. hide browser:chrome) and make it persist. To do this, we introduced a new `Display Settings` section to the Metric Explorer Settings block and added a `displaySettingsValidator` property to the MetricExplorerBlock that allows orgs to hide certain series. This `Display Settings` section will be expanded in the future to allow orgs to customize a series name and color. And we'll also likely add more controls here to allow the user to customize the chart's name. We'll also likely want to carry these changes over to the `Sql Explorer` as well for the same controls.

### Dependencies

- None

### Callouts
- One thing I'm not super happy with is how we're building the list of slice columns/levels inside of the `Display Settings` section. We're taking the `metricAutoSlices` and `metricCustomSlices` and building the list from that. At the same time, when we get the rows back from the query analysis runner, we're also building the name for each column/level combo. These both use the `formatSliceLabel` function, but it feels a bit fragile. I'd much prefer the data we're using in the `Display Settings` to be from the rows, but I haven't found a clean pattern to do that yet. I've explored building a hook, or adding it to the context, but neither option felt great. Just calling this out.

### Testing

Basic Tests:
- [x] Ensure that a pro org is not able to add slices to a metric explorer block and that the we correctly show a premium badge that explains why
- [x] Ensure that an enterprise org is able to add slices to a Metric Explorer Block - if the fact table associated with a metric has any auto slices defined, ensure the Auto Slices section is enabled and the auto slice options are present
- [x] If the same org above doesn't have any auto slices, ensure we disable the dropdown and display an appropriate message.
- [x] Ensure that an enterprise org is able to add custom slices so long as they have auto slices OR any columns on the table have a `topValue` property set. If not, ensure that we have an appropriate message coaching them how to take advantage of this feature.

Metric Analysis Tests:
- [x] Ensure if an enterprise org selects an auto slice, the graph renders each slice variant, plus a roll up `other` variant.
- [x] Ensure if an enterprise org select multiple auto slices, the graph renders each slice variant, plus a roll up `other` variant.
- [x] Ensure if an enterprise org selects an auto slice, plus adds custom slices, the graph renders each slice variant, plus a roll up `other` variant, and each custom slice variant (but there is no concept of an `other` for the custom slice variant.
- [x] Ensure that if an enterprise org adds auto slices where the number of individual slice levels exceeds 100, we render a graphic in the UI letting them know we can only process 100 slice levels and that the graph may not be 100% accurate.

Display Settings Tests:
- [x] If the block does not have a metric selected, ensure we render a message in the Display Settings section that coaches the user to select a metric.
- [x] If a block does not contain any slices, ensure we render the Display Settings section and show the non-sliced data, but we display a user's ability to show/hide it, and the name of the row is a combination of the metric name + valueType. E.G. `Orders (sum)`
- [x] If a block contains auto slices, ensure we render a collapsible row for each auto slice column, and inside of the collapsible section, we render every slice level available, with a checkbox.
- [x] If a block section contains custom slices, ensure we render a collapsible row for the `Custom Slices` and inside of the collapsible section, we render every custom slice available, with a checkbox.
- [x] Ensure this section works when a block has just auto slices, or just custom slices, as well as auto and custom slices.
- [x] Ensure that by default, each row is checked
- [x] Ensure that when a slice level is unchecked, it is removed from the graph immediately
- [x] Ensure that if a slice level is unchecked, and the block settings are saved, these settings persist from visit to visit and don't reset.
- [x] Ensure that if you use the select all / deselect all, it works as expected
- [x] Ensure that if you use the search feature, the search is correctly filtering slice options, and the select all / deselect all only affects the items visible from the search

Additional Tests:
- [x] Ensure that refreshing the block retains the slice levels and any display settings
- [x] Ensure that auto refreshing of the block retains the slice levels and any display settings
- [x] Ensure that existing dashboards continue to work without issue
- [x] Ensure that changing the metric of a block removes any slice levels and resets the displaySettings
- [x] Ensure that we reset any slices if the viz type changes to `Big Number` and we render a message in the `Metric Slices` section that tells a user slices aren't available for Big Number visualizations.

### DIFF
--- a/packages/front-end/components/Experiment/MetricSliceEditingInterface.tsx
+++ b/packages/front-end/components/Experiment/MetricSliceEditingInterface.tsx
@@ -3,7 +3,7 @@ import { FaTimes } from "react-icons/fa";
 import { PiX, PiStackBold } from "react-icons/pi";
 import { Text, Flex, IconButton } from "@radix-ui/themes";
 import { SliceLevelsData } from "shared/experiments";
-import { FactMetricInterface } from "back-end/types/fact-table";
+import { FactMetricInterface } from "shared/types/fact-table";
 import SelectField from "@/components/Forms/SelectField";
 import Field from "@/components/Forms/Field";
 import Button from "@/ui/Button";
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerCustomSliceSelector.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerCustomSliceSelector.tsx
@@ -3,7 +3,7 @@ import { FaPlusCircle } from "react-icons/fa";
 import { PiPencilSimpleFill, PiX } from "react-icons/pi";
 import { Text, Flex, IconButton } from "@radix-ui/themes";
 import { isFactMetric, SliceLevelsData } from "shared/experiments";
-import { FactMetricInterface } from "back-end/types/fact-table";
+import { FactMetricInterface } from "shared/types/fact-table";
 import { CustomMetricSlice } from "shared/validators";
 import Badge from "@/ui/Badge";
 import { useDefinitions } from "@/services/DefinitionsContext";
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerDisplaySettings.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/MetricExplorerDisplaySettings.tsx
@@ -3,8 +3,8 @@ import { Flex, Text, Box } from "@radix-ui/themes";
 import {
   DashboardBlockInterfaceOrData,
   MetricExplorerBlockInterface,
-} from "back-end/src/enterprise/validators/dashboard-block";
-import { FactTableInterface } from "back-end/types/fact-table";
+} from "shared/src/enterprise/validators/dashboard-block";
+import { FactTableInterface } from "shared/types/fact-table";
 import Collapsible from "react-collapsible";
 import { FaAngleRight } from "react-icons/fa";
 import { PiPalette } from "react-icons/pi";
--- a/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/SeriesList.tsx
+++ b/packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/SeriesList.tsx
@@ -1,8 +1,8 @@
 import {
   DashboardBlockInterfaceOrData,
   MetricExplorerBlockInterface,
-} from "back-end/src/enterprise/validators/dashboard-block";
-import { FactTableInterface } from "back-end/types/fact-table";
+} from "shared/src/enterprise/validators/dashboard-block";
+import { FactTableInterface } from "shared/types/fact-table";
 import { Box, Flex, Text, TextField } from "@radix-ui/themes";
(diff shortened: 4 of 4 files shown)


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -884,31 +885,45 @@ export default abstract class SqlIntegration
...
+    // Get fact table once for reuse - use the base metric (first one)
+    const baseMetric = metrics[0];
+    const factTable = params.factTableMap.get(
+      baseMetric.numerator?.factTableId || "",
+    );
+    if (!factTable) {
+      throw new Error("Unknown fact table");
+    }
+
+    const metricsWithIndices = metrics.map((metric, index) => ({
```
mknowlton89, line 920: We could update the metrics we pass in here to be this shape to avoid having to loop through all of the metrics again.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -163,6 +163,7 @@ import {
 
 export const MAX_ROWS_UNIT_AGGREGATE_QUERY = 3000;
 export const MAX_ROWS_PAST_EXPERIMENTS_QUERY = 3000;
+export const MAX_METRICS_IN_FACT_METRIC_QUERY = 100;
```
mknowlton89: @lukesonnet I pulled in the last changes to this branch to dust it off after the holidays. There were some conflicts with your [Daily Participation branch](https://github.com/growthbook/growthbook/pull/4948). I merged those in and resolved the conflicts. I *think* everything works as expected, but would love a second look.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -888,31 +887,45 @@ export default abstract class SqlIntegration
     finalValueColumn: string,
     finalDenominatorColumn: string,
     ratioMetric: boolean,
+    alias: string,
   ): string {
-    return `, COUNT(*) as units
```
Kevin-Chant, line 892: Why'd the `COUNT(*)` get dropped?

packages/back-end/src/integrations/SqlIntegration.ts
Kevin-Chant: I think it's fine as-is. Shouldn't impact performance and it's not enough lines of code to worry about imo

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -888,31 +887,45 @@ export default abstract class SqlIntegration
...
   }
 
   getMetricAnalysisQuery(
-    metric: FactMetricInterface,
+    metrics: FactMetricInterface[],
     params: Omit<MetricAnalysisParams, "metric">,
   ): string {
     const { settings } = params;
 
+    // Get fact table once for reuse - use the base metric (first one)
```
Kevin-Chant, line 911: Do we need any validation that the metrics are coming from the same fact table?

PR #4884, opened 2025-11-24
fix(ConditionInput): Update operator when attribute format changes + add warning for version

ahdriel:
### Features and Changes

When using a `string` attribute with format `version`, it is very important to use `$v*` operators like `$vlt`, as they handle versions as we would expect, and not like strings.

* Fixed ConditionInput operator not updating if format changes
  * If we selected `equal to` for a `string` attribute, and then changed the attribute to one with format `version`, we would not update the operator, leading to unexpected behavior.
  * Now this is fixed and we will change from $eq to $veq and vice-versa
* If a rule is uses the non-version operators on a version attribute, we will show a warning as it is likely unintentional

Screenshot 2025-11-24 at 2 43 18 PM

### DIFF
--- a/packages/front-end/components/Features/Rule.tsx
+++ b/packages/front-end/components/Features/Rule.tsx
@@ -414,16 +414,15 @@ export const Rule = forwardRef<HTMLDivElement, RuleProps>(
                     <Callout status="warning" mt="3">
                       <Flex direction="column" gap="2">
                         <Text>
-                          This rule is using string comparison operators on
-                          version string attributes which can lead to unexpected
-                          behavior. Update the following attributes to use
-                          version-specific operators (e.g. use $vgt instead of
-                          $gt):{" "}
+                          This rule uses string operators on version attributes,
+                          which can have unintended effects. Edit this rule and
+                          change{" "}
                           <strong>
                             {attributesWithVersionStringOperatorMismatches.join(
                               ", ",
                             )}
-                          </strong>
+                          </strong>{" "}
+                          to use version operators ($vgt, $vlt, etc.) instead.
                         </Text>
                       </Flex>
                     </Callout>
--- a/packages/front-end/services/features.ts
+++ b/packages/front-end/services/features.ts
@@ -72,7 +72,7 @@ export interface AttributeData {
   disableEqualityConditions?: boolean;
 }
 
-export const STRING_FORMAT_OPERATORS_MAP = {
+export const STRING_VERSION_FORMAT_OPERATORS_MAP = {
   $eq: "$veq",
   $ne: "$vne",
   $gt: "$vgt",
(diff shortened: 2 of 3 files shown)


packages/front-end/test/components/Features/ConditionInput.test.tsx
```diff
@@ -0,0 +1,272 @@
...
+  //       expect(screen.getByText("IF")).toBeInTheDocument();
+  //     });
+
+  //     // Check that onChange maintains the version operator
+  //     expect(mockOnChange).toHaveBeenCalled();
+  //     const lastCall =
+  //       mockOnChange.mock.calls[mockOnChange.mock.calls.length - 1];
+  //     const outputCondition = JSON.parse(lastCall[0]);
+  //     expect(outputCondition).toEqual({ version: { $vgt: "2.0.0" } });
+  //   });
```
msamper: Do we want to delete or uncomment these tests?

packages/front-end/services/features.ts
```diff
@@ -72,6 +72,19 @@ export interface AttributeData {
   disableEqualityConditions?: boolean;
 }
 
+export const STRING_FORMAT_OPERATORS_MAP = {
```
msamper:
```suggestion
export const STRING_VERSION_FORMAT_OPERATORS_MAP = {
```
Maybe add version to the const name? There's also a Date format so this might be ambiguous

packages/front-end/components/Features/Rule.tsx
```diff
@@ -395,6 +409,25 @@ export const Rule = forwardRef<HTMLDivElement, RuleProps>(
...
+                {attributesWithVersionStringOperatorMismatches &&
+                  attributesWithVersionStringOperatorMismatches.length > 0 && (
+                    <Callout status="warning" mt="3">
+                      <Flex direction="column" gap="2">
+                        <Text>
+                          This rule is using string comparison operators on
+                          version string attributes which can lead to unexpected
+                          behavior. Update the following attributes to use
+                          version-specific operators (e.g. use $vgt instead of
+                          $gt):{" "}
```
msamper: Will people know what the right action is here? I'm assuming they can manually edit the JSON or switch to a regular string attribute and back to the version attribute to correct it

msamper (commented):
Tested and seems to be fixed! Left some comments

packages/front-end/components/Features/Rule.tsx
msamper: This callout kind of sounds like I need to go the attributes page and make a change there

msamper (approved):
Approving to unblock. Comments aren't blocking

packages/front-end/test/components/Features/ConditionInput.test.tsx
ahdriel: Nice catch! We can remove as the last 2 tests cover these scenarios.

packages/front-end/components/Features/Rule.tsx
ahdriel: Updated the text to be clearer about having to edit this rule and not the attribute.

And both would fix -- because clicking Edit when it first loads the operator will be unselected, so they can just save it and it would fix it (pending the release of the new feature revision).

PR #4885, opened 2025-11-25
feat(new-onboarding): New onboarding experience for experimenters

msamper:
### Features and Changes

Starts new orgs on an experimentation focused Get Started page when there is an intent to experiment specified on sign up in Cloud. We assume that no specified usage intent means they are interested in experimentation.

Also streamlines the `AnalysisSettingsSummary` and removes unnecessary metadata. Removes Units table from the results page and instead includes a Units count on the `AnalysisSettingsSummary`. Moves metadata on settings used to compute results above the `AnalysisSettingsSummary`.

This new experience will be behind a feature flag.

### Testing

- [x] New org with intent to experiment gets redirected to get started and can see new view
- [x] New org only interested in feature flagging starts with setup flow and sees current get started experience

### Screenshots

**New Get Started view for experimenters**
Screenshot 2025-11-25 at 11 54 25 AM
**New Data Source/Metrics Guide**
Screenshot 2025-11-25 at 10 35 46 AM
**New AnalysisSettingsSummary**
Screenshot 2025-11-26 at 1 00 07 PM
Screenshot 2025-11-26 at 1 00 37 PM


packages/front-end/types/app-features.ts
```diff
@@ -105,4 +104,9 @@ export type AppFeatures = {
   "october16-ff": boolean;
   "general-dashboards": boolean;
   "incremental-refresh": boolean;
+  "top-banner": Record<string, unknown>;
+  "gbdemo-checkout-layout": boolean;
+  "post-stratification": boolean;
+  liebenowflag9: boolean;
```
jdorn: Remove this `liebenowflag9`.  The feature is archived (and I think it was added by accident in the first place).

packages/front-end/public/images/get-started/advanced/data-pipeline.png
jdorn, line 1: Run this through `pngquant` (or similar) to reduce the file size

packages/front-end/pages/index.tsx
```diff
@@ -40,11 +42,18 @@ export default function Home(): React.ReactElement {
...
+      ) ||
+      organization?.demographicData?.ownerUsageIntents?.length === 0 ||
+      !organization?.demographicData?.ownerUsageIntents; // If no intents, assume interest in experimentation
     if (!hasFeatureOrExperiment) {
-      if (
-        gb.isOn("use-new-setup-flow-2") &&
-        !organization.isVercelIntegration
-      ) {
+      const useNewOnboarding =
+        gb.isOn("experimentation-focused-onboarding") && isCloud();
```
jdorn: We only want to call `isOn` for people that are actually going to see the new flow.  Otherwise, it will just add noise to the experiment results.

That means don't call `isOn` if we're not on cloud or if the usage intent means they are going to get the old setup flow.  It looks like this is done correctly on the GetStarted page already, so we can just copy that logic here.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -447,6 +303,12 @@ export default function AnalysisSettingsSummary({
                   />
                 ))}
             </div>
+            <div className="col-auto" style={{ fontSize: "12px" }}>
+              <Metadata
+                label="Units"
```
jdorn: Is "Units" clear enough here about what it means?

A few options:
* Use the identifier type instead (so `user_id`, `orgId`, etc.).  Can probably do some lightweight formatting of these to make them more readable (like convert from snake/camel case to title case and pluralize it: `userId` -> `User Ids`)
* Use the term `Experiment Units` instead of just `Units`.  Still might not be 100% clear to people not familiar with the terminology and it takes up a lot of space.
* Add a user/person icon back to make it more clear what "units" refers to. However, this might be misleading for identifiers like "orgId", where it's not really a count of individual users.

packages/front-end/pages/getstarted/data-source-guide.tsx
```diff
@@ -0,0 +1,271 @@
...
+        (m) =>
+          m.projects.includes(project) && !m.projects.includes(demoProjectId),
+      )
+    : factMetrics.some((m) => !m.projects.includes(demoProjectId));
+  const hasDatasource = project
+    ? datasources.some(
+        (d) =>
+          d.projects?.includes(project) && !d.projects?.includes(demoProjectId),
+      )
+    : datasources.some((d) => !d.projects?.includes(demoProjectId));
```
jdorn, line 73: Fact tables, metrics, and data sources can also be global (empty projects array). Those are being skipped here when they shouldn't be

packages/shared/src/constants.ts
```diff
@@ -51,6 +51,7 @@ export const OWNER_JOB_TITLES = {
   engineer: "Engineer",
   dataScientist: "Data Scientist",
   projectManager: "Project Manager",
+  productManager: "Product Manager",
```
jdorn: Instead of adding a new option here, I think we should just change the display name to `PM` so it works for both.  Will keep the list of options small and I don't think we really care about the distinction between these two.

Similarly, we may want to make the "Data Scientist" label a little more broad so it encompasses data analysts and data engineers as well.  Maybe `Data & Analytics`?

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
msamper: We could also consider doing something like `Exposures` which sounds more clear than `Units` if we want to keep it generic. It feels like we should have display names for identifier types if we want the unit to be dynamic based on the actual identifier type but it seems like we only allow snake case with lowercase characters for identifier type names in the datasource page so pretty easy to just convert that for now

msamper: Screenshot 2025-11-26 at 11 55 19 AM

with `user_id`

packages/front-end/pages/_app.tsx
```diff
@@ -167,7 +167,6 @@ function App({
         }
         .radix-themes {
           --default-font-family: ${inter.style.fontFamily};
-          --font-weight-medium: 600;
```
msamper, line 170: Removed this because it's making the weight for `medium` the same as `bold` and we use `medium` in several designs. Not sure why this was added

jdorn (approved):

PR #4893, opened 2025-11-25
pin form-data to avoid security vulnerabilities

tzjames:
### Features and Changes

Form-data has some [security vulnerabilities](https://us-east-1.console.aws.amazon.com/inspector/v2/home?region=us-east-1#/findings/container-image/arn%3Aaws%3Aecr%3Aus-east-1%3A476680172222%3Arepository%2Fgrowthbook%2Fsha256%3A0f76c288cbd3ae08c97783bf756706a267c5a9e0be929760e7230790910120e7).  We pin it to the most recent major version for versions 2,3,and 4.  This is deemed safe because the [change log](https://github.com/form-data/form-data/blob/master/CHANGELOG.md) doesn't show any breaking changes within major versions.  The modules relying on older major versions are either from testing or from @types.

This is the output of `pnpm why -r --prod form-data` after pinning in `overrides`:
```
pnpm why -r --prod form-data
Legend: production dependency, optional only, dev only

back-end@0.0.1 /Users/james/growthbook/growthbook/packages/back-end (PRIVATE)

dependencies:
snowflake-sdk 2.3.3
├─┬ @google-cloud/storage 7.10.0
│ └─┬ retry-request 7.0.2
│   └─┬ @types/request 2.48.12
│     └── form-data 2.5.5
└─┬ axios 1.12.2
  └── form-data 4.0.5
```
Note ECR was complaining about version 3.0.1 and 4.0.0 being mentioned in /usr/local/src/app/node_modules/.pnpm/tedious@16.7.1/node_modules/tedious/benchmarks/package-lock.json however `pnpm why -r --prod form-data` doesn't show that, I think because they are not actually installed.   Hopefully ECR won't still mention it after this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,10 @@
     "overrides": {
       "@growthbook/growthbook": "1.6.2",
       "vite": "^6.0.0",
-      "execa": "^5.0.0"
+      "execa": "^5.0.0",
+      "form-data@2": "2.5.5",
+      "form-data@3": "3.0.4",
+      "form-data@4": "4.0.5"
     },
     "patchedDependencies": {
       "@types/presto-client@1.0.2": "patches/@types__presto-client@1.0.2.patch",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,9 @@ overrides:
   '@growthbook/growthbook': 1.6.2
   vite: ^6.0.0
   execa: ^5.0.0
+  form-data@2: 2.5.5
+  form-data@3: 3.0.4
+  form-data@4: 4.0.5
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -71,7 +71,10 @@
     "overrides": {
       "@growthbook/growthbook": "1.6.2",
       "vite": "^6.0.0",
-      "execa": "^5.0.0"
+      "execa": "^5.0.0",
+      "form-data@2": "2.5.5",
```
ahdriel, line 75: Love that we can do this now! Thanks for unlocking this 🚀

ahdriel (approved):

PR #4897, opened 2025-11-26
When deleting a project, delete all associated resources by default

jdorn:
### Features and Changes

Currently, deleting a project leaves a bunch of orphaned resources - features, experiments, metrics, data sources, etc.

This PR adds a checkbox (default checked) when deleting a project to also clean up all of these resources.  For single-project resources like features and experiments, it will simply delete them.  For multi-project resources like metrics, it will only delete if the resource is scoped to only this project and no other ones.

This "delete resources" behavior already existed when deleting the sample project.  This PR just adds it to the normal "delete project" flow as well and lets users opt-out if they want to keep the current behavior.

image

### DIFF
--- a/packages/front-end/pages/projects/index.tsx
+++ b/packages/front-end/pages/projects/index.tsx
@@ -143,7 +143,7 @@ const ProjectsPage: FC = () => {
                       {canDelete ? (
                         <DeleteButton
                           className="dropdown-item text-danger"
-                          displayName="project"
+                          displayName={p.name}
                           text="Delete"
                           useIcon={false}
                           onClick={async () => {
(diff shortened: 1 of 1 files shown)


packages/back-end/src/routers/project/project.controller.ts
```diff
@@ -286,27 +262,50 @@ export const deleteProject = async (
...
+  // custom hooks?
+  // custom fields?
+  // pre-launch checklists?
+
+  if (deleteResources && failedToDeleteResources.length > 0) {
+    const message =
+      `Project deleted, but failed to delete the following resources: ` +
+      failedToDeleteResources.join(", ");
+    res.status(400).json({
+      status: 400,
```
Kevin-Chant, line 304: Nit: status `500` instead since it's not a user error?
```suggestion
    res.status(500).json({
      status: 500,
```

Kevin-Chant (approved):

PR #4898, opened 2025-11-26
Env var to disable SDK payload caching

jdorn:
### Background

We currently cache SDK payload data in a MongoDB document.  This document includes all features across all projects.  Because of Mongo's 16MB max document size, this imposes an upper limit on the number and size of features a single organization can have.  This problem is even worse when using Mongo-compatible databases like Firestore which impose a much smaller 1MB limit.

### Changes

This PR adds a new environment variable (`SDK_PAYLOAD_CACHE=none`) to disable the SDK payload cache entirely.  When disabled, we also optimize the payload generation logic.  Since we're not caching the result for future use, we can run more tailored database queries to get only what we need for the current response.

Disabling the cache makes payload generation more computationally expensive, so this is only safe to do when there is a robust caching layer in front of GrowthBook.

### Future

In the future, we plan to add additional options to cache in S3 or GCS instead of Mongo to enable scaling beyond Mongo's size limitations while keeping payload generation fast and efficient.

### DIFF
--- a/packages/back-end/test/api/features-public.test.ts
+++ b/packages/back-end/test/api/features-public.test.ts
@@ -75,6 +75,7 @@ jest.mock("back-end/src/services/organizations", () => ({
 jest.mock("back-end/src/models/SdkPayloadModel", () => ({
   getSDKPayload: jest.fn(),
   updateSDKPayload: jest.fn(),
+  getSDKPayloadCacheLocation: jest.fn().mockReturnValue("mongo"),
 }));
 
 jest.mock("back-end/src/models/SdkConnectionModel", () => ({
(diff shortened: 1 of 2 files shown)


packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -1380,11 +1380,18 @@ export async function getExperimentMapForFeature(
 
 export async function getAllPayloadExperiments(
   context: ReqContext | ApiReqContext,
-  project?: string,
+  projects?: string[],
 ): Promise<Map<string, ExperimentInterface>> {
+  const projectFilter =
```
bryce-fitzsimons, line 1385: nit: unify the way we build the query filter between ExperimentModel and FeatureModel

PR #4899, opened 2025-11-26
Improve saved query management for dashboards

Kevin-Chant:
### Features and Changes

Currently, duplicating a dashboard sql explorer block (either by duplicating the dashboard as a whole or just the block directly) just creates a new reference to the existing saved query. This results in any changes made to the underlying query after duplication also being reflected in the original block. To address this, there are two main changes in this PR:

- When creating a dashboard, all saved queries that are referenced by other dashboards are cloned so they don't affect each other
- Within the block editor for sql explorer blocks, there are two new actions to duplicate or delete the chosen saved query

Note that when duplicating a single sql explorer block it still uses a reference to the original saved query rather than automatically duplicating the query, but it's much easier to duplicate that query as needed.

### Testing

For setup, create a dashboard and add three sql explorer blocks. Two blocks should use the same saved query and one should have a different query.
- [x] Duplicating the dashboard creates a new dash with three blocks using two queries, both of which have new IDs
- [x] Updating any of the new blocks should affect neither the existing dashboard nor the other new blocks
- [x] Within the new dashboard, duplicate one of the saved queries and add it to a new block. Changing that new query shouldn't affect the other blocks
- [x] Delete the newly duplicated saved query, it should reset the selected block's state and no longer show in the dropdown

### Screenshots

image

image

### DIFF
--- a/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
+++ b/packages/front-end/components/SchemaBrowser/SqlExplorerModal.tsx
@@ -57,17 +57,19 @@ import { filterOptions } from "../DataViz/DataVizFilter";
 import SchemaBrowser from "./SchemaBrowser";
 import styles from "./EditSqlModal.module.scss";
 
+export interface SqlExplorerModalInitial {
+  sql?: string;
+  name?: string;
+  datasourceId?: string;
+  results?: QueryExecutionResult;
+  dateLastRan?: Date | string;
+  dataVizConfig?: DataVizConfig[];
+}
+
 export interface Props {
   dashboardId?: string;
   close: () => void;
-  initial?: {
-    sql?: string;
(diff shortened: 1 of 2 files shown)


packages/shared/src/validators/saved-queries.ts
```diff
@@ -230,7 +230,7 @@ export const testQueryRowSchema = z.record(z.any());
 
 export const queryExecutionResultValidator = z.object({
   results: z.array(testQueryRowSchema),
-  error: z.string().optional(),
+  error: z.string().nullable().optional(),
```
Kevin-Chant, line 233: This is because the `error` field is `null` in mongo and isn't being coerced to either `undefined` or `""` when read from the DB automatically. We could add a `migrate` helper but this was lighter weight

PR #4921, opened 2025-12-02
feat(auth): Add Kerberos support for Trino/Presto (take 2)

### LINKED ISSUE #4773: [Feature] Kerberos authentication support for Presto/Trino

ahdriel:
Closes #4773 again

### Features and Changes

#4850 got reverted in #4911 because some users were facing `User X cannot impersonate user growthbook`.

The root cause is that we were not passing the `engine` field to `presto-client`, and it was defaulting to `Presto`, meaning that we were sending `X-Presto-User: growthbook` header which was being ignored by Trino.

By passing the proper `engine`, the proper header was being set which Trino respected and then threw an error in case the impersonation rules were not configured.

To avoid this, for basic_auth scenarios, we are going to set the user to the same user from basic_auth, bypassing impersonation. If wanted, we can explore a more comprehensive impersonation solution in the future (passing in user name or user emails to restrict data access on a per-user basis for example, instead of having a single connection).

### DIFF
--- a/packages/back-end/src/integrations/Presto.ts
+++ b/packages/back-end/src/integrations/Presto.ts
@@ -37,7 +37,6 @@ export default class Presto extends SqlIntegration {
   }
   runQuery(sql: string): Promise<QueryResponse> {
     const configOptions: ClientOptions = {
-      engine: this.params.engine,
       host: this.params.host,
       port: this.params.port,
       user: "growthbook",
@@ -52,7 +51,6 @@ export default class Presto extends SqlIntegration {
         user: this.params.username || "",
         password: this.params.password || "",
       };
-      configOptions.user = this.params.username || "";
     }
     if (this.params?.authType === "customAuth") {
       configOptions.custom_auth = this.params.customAuth || "";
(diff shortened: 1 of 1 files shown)


packages/back-end/src/integrations/Presto.ts
```diff
@@ -51,10 +52,27 @@ export default class Presto extends SqlIntegration {
         user: this.params.username || "",
         password: this.params.password || "",
       };
+      configOptions.user = this.params.username || "";
     }
     if (this.params?.authType === "customAuth") {
       configOptions.custom_auth = this.params.customAuth || "";
```
jdorn, line 56: It sounds like on Trino, we were previously not sending a `user` field at all and now we're setting it to `growthbook`.  We're patching that behavior for `basicAuth`.  Do we also need to do something similar for `customAuth`?

jdorn (approved):
Approved with a question

packages/back-end/src/integrations/Presto.ts
ahdriel: This is a good callout and something I was having a hard time figuring out a proper solution as all changes would change behavior for either trino or presto dbs.

For now I am taking a more conservative approach and setting `engine` only for Kerberos auth, keeping the same behavior for already existing Data sources. Meaning they'll keep receiving `X-Presto-User` headers that are ignored by Trino.

In the future we can work a more comprehensive solution for impersonation (like sending user email addresses and letting Trino/Presto map that to an internal user) if we want.

PR #4922, opened 2025-12-02
fix(sql): fix broken count distinct re-aggregation in metric analysis

lukesonnet:
### Features and Changes

We previously had to do some special re-aggregation for count distinct metrics in the metric analysis query. Now that we are instead relying on the shared partial, re, and full aggregation methods, this isn't necessary.

That old special re-aggregation was not compatibility with the refactor to use the new methods and was breaking metric analysis queries for count distinct metrics.

This PR removes that old special re-aggregation.

### Testing

Tested a count distinct metric analysis query in BQ.

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -941,32 +941,14 @@ export default abstract class SqlIntegration
 
     const createHistogram = metric.metricType === "mean";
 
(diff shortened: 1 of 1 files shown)


ahdriel (approved):
Thanks for the explanation!

Code LGTM and it is simplified -- I do see we are doing the re/aggregation in a previous query now (see `value_for_reaggregation` in code).

PR #4923, opened 2025-12-02
Improved Row Filters for Metrics

jdorn:
Currently, we have two fields for filtering rows in metric definitions.  `filters` lets you pick saved filters.  `inlineFilters` lets you define ad-hoc `IN` clauses for filtering.  Anything that is not a simple `IN` clause requires you to create a saved filter first, which is not ideal.

This PR replaces both of these fields with a brand new `rowFilters` array.  This supports both use cases above, as well as many additional inline filtering operators beyond `IN` clauses.

You can now add multiple row filters.  Each one has an "operator", plus additional config depending on the operator.  Here is the full list of supported operators now:
- `=`
- `!=`
- `>`
- `<`
- `>=`
- `<=`
- `in` - same as previous `inlineFilters` field
- `not_in`
- `is_null`
- `not_null`
- `is_true`
- `is_false`
- `contains`
- `not_contains`
- `starts_with`
- `ends_with`
- `sql_expr`
- `saved_filter` - same as previous `filters` field

image

TODO:
- [x] ColumnRef model changes
- [x] JIT migration to new format
- [x] Update SQL generation code (with many new test cases)
- [x] Update REST endpoints to support both old and new formats
- [x] Update UI in FactMetricModal
- [x] Update metric docs
- [x] UX / design review
- [x] Test JIT migration for existing inline filters and saved filters
- [x] Test metric create/edit flows
- [x] Test REST api endpoints for fact metrics
- [ ] Test product analytics
- [ ] Test experiment SQL
- [x] Test metric SQL


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -142,9 +146,52 @@ export class FactMetricModel extends BaseClass {
...
+        newColumnRef.rowFilters.push({
+          operator: "saved_filter",
+          values: [f],
+        });
+      }
+    }
+
+    if (inlineFilters) {
+      for (const [column, values] of Object.entries(inlineFilters)) {
+        const filteredValues = values.filter((v) => !!v);
```
lukesonnet, line 200: What if a value is `""` and that value is meaningful? Shouldn't we include that value?

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -142,9 +146,52 @@ export class FactMetricModel extends BaseClass {
...
+    }
+
     return newDoc as FactMetricInterface;
   }
 
+  public static migrateColumnRef(columnRef: LegacyColumnRef): ColumnRef {
+    const { filters, inlineFilters, ...newColumnRef } = columnRef;
+
+    // If row filters are already defined, do nothing
+    if (newColumnRef.rowFilters !== undefined) {
```
lukesonnet, line 183: question: Should this be possible? No, right? But obviously fine to have as a fallback, just want to make sure I understand that we are cleaning up the column ref on save in other endpoints.

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -313,6 +370,32 @@ export class FactMetricModel extends BaseClass {
...
+  ): LegacyColumnRef {
+    const newColumnRef: LegacyColumnRef = {
+      ...columnRef,
+      filters: [],
+      inlineFilters: {},
+    };
+
+    newColumnRef.rowFilters?.forEach((rf) => {
+      if (rf.operator === "saved_filter") {
+        newColumnRef.filters?.push(rf.values?.[0] || "");
```
lukesonnet, line 403: How I long for us only to return the new ColumnRef in a versioned API endpoint.

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -263,15 +281,12 @@ function getMetricSettingsHash(
...
-          id: it.id,
-          name: it.name,
-          value: it.value,
-        })),
+        filters: getFiltersForHash(numeratorFactTable, factMetric.numerator),
       },
       denominatorFactTable: {
         sql: denominatorFactTable?.sql,
         eventName: denominatorFactTable?.eventName,
+        // TODO: also include denominator filters?
```
lukesonnet, line 289: Oof. Yes. We must have missed this the last time around. If we add them now, then we will incorrectly "break" a bunch of filtered ratio metric time series and incorrectly state the settings changed.

I kind of think we should just add it now and not look back.

packages/back-end/src/services/safeRolloutTimeSeries.ts
```diff
@@ -199,15 +196,12 @@ function getSafeRolloutMetricSettingsHash(
...
-          id: it.id,
-          name: it.name,
-          value: it.value,
-        })),
+        filters: getFiltersForHash(numeratorFactTable, factMetric.numerator),
       },
       denominatorFactTable: {
         sql: denominatorFactTable?.sql,
         eventName: denominatorFactTable?.eventName,
+        // TODO: include denominator filters?
```
lukesonnet, line 204: ditto

packages/shared/src/validators/fact-table.ts
```diff
@@ -107,13 +107,39 @@ export const columnAggregationValidator = z.enum([
...
+  "is_true",
+  "is_false",
+  "sql_expr",
+  "saved_filter",
+] as const;
+
+export const rowFilterValidator = z.object({
+  operator: z.enum(rowFilterOperators),
+  column: z.string().optional(),
+  values: z.array(z.string()).optional(),
```
lukesonnet, line 134: Classic engineer-learns-new-trick type comment, but should this be a discriminated union on the `operator` column?

packages/shared/src/experiments.ts
```diff
@@ -189,56 +187,156 @@ export function getColumnRefWhereClause({
...
+    factTable,
+    jsonExtract,
+  );
+  const columnType = getSelectedColumnDatatype({
+    factTable,
+    column: rowFilter.column,
   });
 
-  return [...where];
+  // If a boolean column is using equals operator, convert to is_true/is_false
```
lukesonnet, line 253: Is there a chance that the column datatype is wrong? Do we do validation that means if they were able to save the filter, then actually the datatype is wrong and we shouldn't do this conversion?

packages/back-end/src/models/FactMetricModel.ts
jdorn: I agree. This is only in the migration code to replicate the "ignore empty string" logic we have today for inline filters.  Once it's using the new `rowFilters` object, empty strings are meaningful and no longer ignored.

packages/back-end/src/models/FactMetricModel.ts
jdorn: It does return the new `rowFilters` field.  It just ALSO has to return the old fields.  So this backfills the old fields when possible to avoid any backwards compatibility breaks.

lukesonnet: Understood.

packages/back-end/src/models/FactMetricModel.ts
jdorn: This is part of the migration.  Old documents in Mongo do not have a `rowFilters` field.  So if the document DOES have it, then we can skip the rest of the migration and return.

packages/shared/src/validators/fact-table.ts
jdorn: It's a trade-off.  Pros and Cons to both approaches.  I found having one consistent type easier to work with, especially on the front-end.  The downside is that you may have some orphaned values around if they slip through validation - e.g. an `is_null` operator with an array of string values that are just ignored.

packages/back-end/src/services/experimentTimeSeries.ts
jdorn: We could only add the filed if it's not empty.  I would guess 95% of denominators do not use any saved filters, so we would only invalidate a small number of time series.

lukesonnet (approved):

PR #4929, opened 2025-12-03
add top values -> auto slice levels job, add pinning for important levels

bryce-fitzsimons:
Adds a new fact table level setting that allows automatically refreshing auto slice levels based on top values. Runs as a scheduled job (but also triggered on table changes or top-level manual refreshes).

You can hold out specific levels as locked - protected from being erased by the next run top values query.

Query uses a 14 day lookback for top values.

Runs every 24 hours by default, settable via `AUTO_SLICE_UPDATE_FREQUENCY_HOURS`. Note: We could consider letting cloud users set this in org settings if preferred.

image

image

image

image

Nothing returned from top values query state:
image

### DIFF
--- a/packages/back-end/src/jobs/updateAutoSlices.ts
+++ b/packages/back-end/src/jobs/updateAutoSlices.ts
@@ -149,11 +149,15 @@ async function updateAutoSlicesForColumns(
 
         // Process results for each column
         for (const col of columnChunk) {
-          const topValues = topValuesByColumn[col.column] || [];
+          const topValues = topValuesByColumn[col.column];
 
           // Persist topValues and topValuesDate
           col.topValues = topValues;
           col.topValuesDate = new Date();
+
+          // Skip autoSlices update if no top values.
+          // (empty response may be transient, not worth losing the slices)
+          if (!topValues?.length) continue;
 
           // Update autoSlices with locked levels + new top values
           const lockedLevels = col.lockedAutoSlices || [];


packages/back-end/src/jobs/updateAutoSlices.ts
```diff
@@ -0,0 +1,165 @@
...
+    UPDATE_SINGLE_FACT_TABLE_AUTO_SLICES,
+    updateSingleFactTableAutoSlices,
+  );
+
+  await startUpdateJob();
+
+  async function startUpdateJob() {
+    const updateJob = agenda.create(QUEUE_AUTO_SLICE_UPDATES, {});
+    updateJob.unique({});
+    updateJob.repeatEvery(AUTO_SLICE_UPDATE_FREQUENCY_HOURS + " hours");
```
bryce-fitzsimons, line 46: if we want, we could convert this to an organization setting (for cloud users). Not sure if warranted though...

packages/back-end/src/util/secrets.ts
```diff
@@ -129,6 +129,9 @@ export const DEFAULT_CONVERSION_WINDOW_HOURS =
 export const METRIC_REFRESH_FREQUENCY =
   parseInt(process.env.METRIC_REFRESH_FREQUENCY || "") || 24;
 
+export const AUTO_SLICE_UPDATE_FREQUENCY_HOURS =
+  parseInt(process.env.AUTO_SLICE_UPDATE_FREQUENCY_HOURS || "") || 24;
```
lukesonnet: This should default to 7 days, not 24 hours.

packages/back-end/src/jobs/updateAutoSlices.ts
lukesonnet: Yes, we can do it later if requested.

packages/front-end/components/FactTables/FactTableModal.tsx
```diff
@@ -290,6 +292,19 @@ export default function FactTableModal({
             />
           </div>
         ) : null}
+
+        {hasCommercialFeature("metric-slices") && (
+          <div className="mt-4">
+            <Checkbox
+              label="Auto-update slice levels"
+              description="Automatically update Auto Slice levels based on top column values (14 day lookback). Locked slice levels will always be preserved."
```
lukesonnet:
```suggestion
              description="Automatically update Auto Slice levels every 7 days based on the top column values over a 14 day lookback. Locked slice levels will always be preserved."
```

Could pipe in the actual value for `7`.

packages/front-end/components/FactTables/ColumnModal.tsx
```diff
@@ -580,44 +596,315 @@ export default function ColumnModal({ existing, factTable, close }: Props) {
...
+                        <Text
+                          weight="regular"
+                          size="1"
+                          style={{ color: "var(--color-text-mid)" }}
+                        >
+                          Last run: Never
+                        </Text>
+                      )}
+                    </Flex>
+                    {!factTable.autoSliceUpdatesEnabled && (
```
lukesonnet: We should allow ad-hoc updates so that you can schedule and then still update when you know a new value has landed.

packages/front-end/pages/api/init.ts
```diff
@@ -150,6 +152,9 @@ export default function handler(req: NextApiRequest, res: NextApiResponse) {
     experimentRefreshFrequency: EXPERIMENT_REFRESH_FREQUENCY
       ? parseInt(EXPERIMENT_REFRESH_FREQUENCY)
       : 6,
+    autoSliceUpdateFrequencyHours: AUTO_SLICE_UPDATE_FREQUENCY_HOURS
+      ? parseInt(AUTO_SLICE_UPDATE_FREQUENCY_HOURS)
+      : 24,
```
lukesonnet: Can we change this default as well.

packages/front-end/services/env.ts
```diff
@@ -24,6 +24,7 @@ const env: EnvironmentInitValue = {
   ingestorOverride: "",
   stripePublishableKey: "",
   experimentRefreshFrequency: 6,
+  autoSliceUpdateFrequencyHours: 24,
```
lukesonnet: and here

packages/back-end/src/jobs/updateAutoSlices.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    // Only update auto-slice columns that are string type
+    if (
+      col.isAutoSliceColumn &&
+      col.datatype === "string" &&
+      !col.deleted &&
+      canInlineFilterColumn(factTable, col.column)
+    ) {
+      try {
+        const topValues = await runColumnTopValuesQuery(
```
lukesonnet: We should probably do this in one query instead of `columns.length` queries...

packages/back-end/src/jobs/updateAutoSlices.ts
```diff
@@ -0,0 +1,165 @@
...
+
+const updateSingleFactTableAutoSlices = async (
+  job: UpdateSingleFactTableAutoSlicesJob,
+) => {
+  const { organization, factTableId } = job.attrs.data;
+
+  if (!factTableId || !organization) return;
+
+  const context = await getContextForAgendaJobByOrgId(organization);
+
```
lukesonnet, line 72: Should we also skip if `organization` doesnt have the auto-slice premium feature if they downgraded from enterprise?

packages/back-end/src/jobs/updateAutoSlices.ts
lukesonnet: done in 49503dee1ac2ad1de71c648de4a1f9657c274f1d

packages/back-end/src/jobs/updateAutoSlices.ts
```diff
@@ -0,0 +1,185 @@
...
+          const lockedLevels = col.lockedAutoSlices || [];
+          const autoSlices: string[] = [...lockedLevels];
+          for (const value of topValues) {
+            if (autoSlices.length >= maxSliceLevels) break;
+            if (!autoSlices.includes(value)) {
+              autoSlices.push(value);
+            }
+          }
+
+          col.autoSlices = autoSlices;
```
lukesonnet, line 172: We might not want to go from "some auto slices" to "none". Even if no top-levels are found, it might be transient, and I'm concerned about us removing all of them in that case.

I think not deleting these slices might cause some experiment queries to fail, but that might be a better failure mode than silently removing all slices due to an errant update query...

bryce-fitzsimons: sure makes sense, fixed

PR #4933, opened 2025-12-03
Use pnpm instead of yarn

tzjames:
### Features and Changes
In order to be able to override transitive dependencies with security vulnerabilities we need to be on a package manager that  allows that.  Unfortunately yarn 1 (classic) only allows resolutions [by dependency graph glob](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md).  That means if the package ever gets updated and its transitive dependencies are on a higher major version than was put in resolutions, then the package would actually be downgraded and possibly broken.

For pnpm overrides can use semver.  So it is easy to make sure they are all using the same package: 
```
 "pnpm": {
    "overrides": {
      "fast-xml-parser@4": "4.5.3"
    }
  }
```

Other advantages of pnpm is that will have faster installs and eventually will use less disk space than yarn (though since we did `pnpm import` here we end up using the same packages and the same size for now).  The stricter dependency checking within pnpm should lead to fewer issues with transitive dependencies in general.

Things done to make this work:

1. Ran `pnpm import`.  This ends up using all the same packages that yarn did.  Copying info in yarn.lock over to pnpm-lock.yaml
2. Made two different pnpm-workspaces.yaml for the main repo and for docs.
3. Updated all commands using yarn in package.json files to use the respective commands with pnpm. Most are a simple swap.  Exceptions are `yarn setup` needs to be `pnpm run setup` as `pnpm setup` is a reserved command.  **We might want to rename our command to not conflict.**.  Also `wsrun` is not needed as pnpm does things recursively with `-r` and can use `--filter` to filter workspace packages.
4. Added dependencies in package.json files that yarn hoisted from other packages and made them explicit (and moved some from devDependencies that are actually needed by the app at run time).  The main tricky one is the `@opentelemetry` ones as they all had to match the main `@opentelemetry/api` one.  **I had to pin some versions to not get it to complain.**
5. Update all the commands in our github action workflows to use pnpm from yarn.
6. Handled patches (patch files have a slightly different format, must be declared in packages.json, and doesn't need the `patch-package` package to work.
7. Updated Dockerfile to use pnpm and prune unneeded files. I don't think I can use `pnpm deploy` because that I think is geared to a single project and we need both front-end and back-end. 
8. Updated all contributing docs to use pnpm commands instead of yarn.
9. Updated commands in helm charts.
10. Moved the scripts within docs to shared package as it was hard to refer to shared package since it is in a different workspace.
11. **Running `pnpm gen-sdk-resources` in docs package remove some capabilities from the Rust package.  But that probably happens on main too.**  
12. I had to increase the limit for api setup in tests as it was timing out.  I'm not sure exactly why it takes slightly longer in this branch, but it was probably the MongoMemoryServer taking a bit longer for some reason to set up.  Only one of the test files was timing out, the others it worked for since it was already loaded.
13. eslint and type-check complained about a few more things than before. Some were real, but Sentry was annoying as it complains about it not exporting values that are available and I had to move to importing by name.

### Dependencies

- [x] Make an alias for yarn to pnpm to ensure that commands people might have in their terraform or infrastructure continue to work.
- [x] yarn start:with-tracing, yarn start:with-datadog
- [x] yarn cron
- [x] yarn workspace back-end start
- [x] Pre-port changes to package.json and other linting issues to main repo 

### Testing

- [x] Make sure that the packages in the build docker image have the patches applied.
```
docker run -it --entrypoint bash growthbook-test:local -c "cd /usr/local/src/app && find node_modules/.pnpm -name '*patch_hash*' -type d | grep -E '(agenda|presto-cl
ient|@types__presto-client)'"
node_modules/.pnpm/presto-client@1.1.0_patch_hash=mbeixnouradmtp2ryk6uz642k4
node_modules/.pnpm/@types+presto-client@1.0.2_patch_hash=bv7qcx3y4ykvpvp7odxdubazpe
node_modules/.pnpm/agenda@5.0.0_patch_hash=22tgtweyd5fuhnx2hdyly6czu4
```
- [ ] Make sure helm charts work
- [x] For backwards compatibility make sure the following still work:
- [x] yarn start:with-tracing, yarn start:with-datadog
```
docker run -it growthbook-test:local yarn start:with-tracing
```
See it start the app
- [x] yarn workspace back-end start
```
docker run -it growthbook-test:local yarn workspace back-end start
```
See the back-end start
- [x] pnpm dev
- [x] pnpm build
- [x] pnpm start
- [x] Click around preview env: https://pr-4933-growthbook.fly.dev/.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN apt-get update && \
   echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" > /etc/apt/sources.list.d/nodesource.list && \
   apt-get update && \
   apt-get install -yqq nodejs && \
-  npm install -g pnpm && \
+  npm install -g pnpm@9.15.0 && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 # Copy over minimum files to install dependencies
@@ -60,8 +60,7 @@ RUN \
   && find node_modules -name "*.map" -delete \
   && find node_modules -name "CHANGELOG*" -delete \
   && find node_modules -name "LICENSE*" -delete \
-  && find node_modules -name "README*" -delete \
-  && find node_modules -type d -name "test" -o -name "tests" -o -name "__tests__" | xargs rm -rf 2>/dev/null || true
+  && find node_modules -name "README*" -delete
 RUN pnpm postinstall
 
 
@@ -76,7 +75,7 @@ RUN apt-get update && \
   echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" > /etc/apt/sources.list.d/nodesource.list && \
   apt-get update && \
   apt-get install -yqq nodejs && \
-  npm install -g pnpm && \
+  npm install -g pnpm@9.15.0 && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 COPY --from=pybuild /usr/local/src/app/requirements.txt /usr/local/src/requirements.txt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,9 +55,6 @@ jobs:
           # Stats engine
           pip install pipx
           pipx install poetry
-          cd packages/stats
-          poetry install
-          cd ../..
         env:
           CI: true
 


Kevin-Chant (commented):
Reading the diff things look pretty good, but for snyk it seems to be stuck trying to test it based on `yarn`. Their docs say `pnpm` is supported but I can't find anything in their UI for how to configure package managers.

If we can get that sorted & rebase latest `main` to fix conflicts I think it's a :white_check_mark:  from me

> pnpm setup is a reserved command

I agree with renaming ours to not conflict

Dockerfile
```diff
@@ -53,8 +53,16 @@ RUN \
...
-RUN yarn postinstall
+  && pnpm install --frozen-lockfile --prod \
+  && pnpm store prune \
+  && find node_modules -name "*.md" -delete \
+  && find node_modules -name "*.ts" ! -name "*.d.ts" -delete \
+  && find node_modules -name "*.map" -delete \
+  && find node_modules -name "CHANGELOG*" -delete \
+  && find node_modules -name "LICENSE*" -delete \
+  && find node_modules -name "README*" -delete \
+  && find node_modules -type d -name "test" -o -name "tests" -o -name "__tests__" | xargs rm -rf 2>/dev/null || true
```
jdorn: Remove this line for tests.  Too risky

Dockerfile
```diff
@@ -22,29 +22,29 @@ RUN apt-get update && \
   mkdir -p /etc/apt/keyrings && \
   wget -qO- https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && \
   echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" > /etc/apt/sources.list.d/nodesource.list && \
-  wget -qO- https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --dearmor -o /etc/apt/keyrings/yarn.gpg && \
-  echo "deb [signed-by=/etc/apt/keyrings/yarn.gpg] https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
   apt-get update && \
-  apt-get install -yqq nodejs yarn && \
+  apt-get install -yqq nodejs && \
+  npm install -g pnpm && \
```
jdorn: For this (and the one below), specify at least the major pnpm version to install

PR #4934, opened 2025-12-03
Deterministic colors for Sql Explorer dimensions

mknowlton89:
### Features and Changes

Currently, in the SQL Explorer, we aren't defining any colors used for dimensions within a graph, which can lead to a case where two graphs have similar dimensions (e.g. the dimension is `country` and there are dimension slices for `US`, `CA`, and `UK`.

Currently, since we're not defining any colors, the colors are determined by the order of the dimensions, which can lead to a dimension's color being different between two graphs, which is really confusing.

This is compounded by the fact that we don't allow users to customize the colors of the charts yet.

This PR adds some logic so that we pass a color into the chart explicitly, and the color is set deterministically, so no matter what the dimension combinations, are, a single dimension (e.g. browser:"CA") will always have the same color across charts.

This currently only affects graphs in the SQL explorer with dimensions, but will soon impact metric explorer graphs with metric slices.

Currently, this PR solves the problem for SQL Explorer by defining 11 different colors in our `PALETTE` and then using those for the dimensions. We defined 11 colors because we currently limit the number of dimensions to 10 for the SQL Explorer (but also added 1 extra for the `other` category that we automatically add if there are more than 10 unique dimensions).

When we add metric slice support, we'll need to increase the number of colors to 100 as we allow up to 100 metric slices in a metric explorer graph. This will prevent some challenges as coming up with this many unique colors is going to be really difficult.

Additionally, it will mean that any existing charts will have their colors shifted (they'll still be the same across graphs, but they will change) since the palette length will change.

We can go ahead and add these colors now to prevent the future shift, the downside is, with 100 different colors, it's more likely that two very similar colors could get added to a graph with just a few dimensions, which isn't a great experience.

One separate option is that we can have 2 palettes - 1 for Sql Explorer dimensions that has 11 colors, and one for Metric explorer slices, with 100 colors. The downside is, the same dimension  (e.g. browser:"CA") would have different colors for a Sql explorer visualization compared to a metric explorer visualization.

The alternative solution here is that we continue to let Echarts define the colors, but we build the UI that allows users to customize colors in case there are conflicts.

We should add this regardless, but will likely require a bigger lift as we don't have any concept of how to customize this now in the UI.

### DIFF
--- a/packages/front-end/services/dataVizConfigUtilities.ts
+++ b/packages/front-end/services/dataVizConfigUtilities.ts
@@ -130,28 +130,19 @@ export function normalizeDimensionsForChartType(
   }
 }
 
-// Most distinct colors first - these are prioritized when we have fewer dimensions
-// to avoid similar colors like blue/cyan or indigo/purple appearing together
-const DISTINCT_PALETTE = [
+const PALETTE = [
   blue.blue8,
   teal.teal10,
   orange.orange10,
   pink.pink10,
   amber.amber10,
   red.red10,
-];
-
-// Similar/fallback colors - only used when we need more than 6 dimensions
-const FALLBACK_PALETTE = [
   indigo.indigo10,
   purple.purple10,
   cyan.cyan10,
   mint.mint10,
   lime.lime11,
 ];
-
-// Full palette for when we have more than 6 dimensions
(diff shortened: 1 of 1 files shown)


mknowlton89:
Gonna go a different route - rather than use deterministic colors we'll generate a palette, and allow a user to override a color, and then we'll re-use that across other charts on the same dashboard.

PR #4940, opened 2025-12-03
Analysis Settings improvements & Unit Breakdown

### LINKED ISSUE #4930: [Bug] Dimensions: Unit breakdown and SRM checks missing
### Summary

Previously, we could look at any dimension and see whether there was any SRM. 
#### Expected Behavior
Now, only the first device's total anonymous IDs are displayed, and there's no way to check SRM. We should be able to see if there is SRM for a particular dimension.

Image

### LINKED ISSUE #4931: [Bug] Custom reports: Missing unit break down
### Summary

Custom reports are missing a unit. And it can only be seen by looking at the metric. No SRM checks, as this makes making custom analyses more time consuming. 
#### Expected Behavior

It should be as previously, with a breakdown and the ability to check for SRM.

msamper:
### Features and Changes

Adds back dimension unit breakdowns for `BreakDownResults`, adds unit count to custom reports `ReportAnalysisSettingsBar`, replaces `GBEdit` in results with `PiPencilSimpleFill`, adds Segment and Activation Metric to Analysis metadata if in use, fixes bug with dimension unit counts not being computed correctly

- Closes #4930 
- Closes #4931 

### Testing

- [x] Add an activation metric and a segment to an experiment and make sure it's displayed
- [x] Select a dimension in the dimension chooser and make sure the unit count looks correct
- [x] Create a custom report with and without a dimension and make sure the unit count looks correct
- [x] When a dimension is selected, make sure dimension breakdown is shown

### Screenshots

**Experiment with Segment and Activation Metric**
Screenshot 2025-12-03 at 3 42 04 PM
**Dimension breakdown added back**
Screenshot 2025-12-03 at 3 42 32 PM
**Custom Report Analysis Settings with units added**
Screenshot 2025-12-03 at 3 43 24 PM


lukesonnet (commented):
Didn't test, but I think this looks good to me!

PR #4942, opened 2025-12-04
Move sdk-connection and managed-by validator to shared

tzjames:
### Features and Changes

This is a continuation of https://github.com/growthbook/growthbook/pull/4935 and https://github.com/growthbook/growthbook/pull/4938.

It:

-  moves sdk-connection type to shared 
- sets up a new Entry Point for validators.
- moves packages/back-end/src/util/constants.ts into the shared/constants entry point

### Testing
`yarn build`


packages/shared/src/validators/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./managed-by";
```
jdorn, line 1: It's a little annoying that we have to update this index file every time we add a new validator.

PR #4944, opened 2025-12-04
perf(gbstats): Optimize post strat performance

ahdriel:
### Features and Changes

After #4202 was merged we observed a clear increase in CPU usage and Memory increase for our Python code.

This PR optimizes our performance while keeping the same behavior (except for a bug fix which affected the uncertainties with more than 3 decimal places).


packages/stats/tests/frequentist/test_post_strat.py
```diff
@@ -254,14 +254,14 @@ def setUp(self):
...
 
-        self.standard_error_count_rel = 0.01224418556792039
-        self.standard_error_count_abs = 0.3769153078989124
-        self.standard_error_count_reg_rel = 0.0022103200530259534
-        self.standard_error_count_reg_abs = 0.07401168371016856
-        self.standard_error_ratio_rel = 0.007141416587299529
-        self.standard_error_ratio_abs = 0.0050779658636993154
-        self.standard_error_ratio_reg_rel = 0.0012337665987066867
-        self.standard_error_ratio_reg_abs = 0.0012278145208316127
+        self.standard_error_count_rel = 0.01216
```
lukesonnet: Why are we truncating these values?

lukesonnet: We should have absolute ability to replicate, no?

lukebrawleysmith: Agreed, I updated so that the full values are in there now.

PR #4948, opened 2025-12-04
feat(metric): Introduce Daily Participation fact metric type

lukesonnet:
### Features and Changes

An analogue of DAU in an experiment context is the fraction of days a user is active over the number of days a user is in an experiment. Or, more generally, one may be interested in a metric type where you measure, per user, the fraction of eligible days that they did some action (were active, or met some other criterion).

This PR introduces a "daily participation" style metric, where we count the number of matching days per user, then compute the eligible days since exposure for that user, divide the two, and then take the average of that over all users in an experiment. This is a "mean" metric for statistical purposes, but we present it as a %age, as it should range from 0 to 1.

For example, if your control variation has the following two users:
* User 1: meets active criteria (literally has a row matching metric filters) on 2 days and has been in the experiment for 5 days, then they are 2/5 = 0.4
* User 2: meets active criteria on 3 out of 3 days, so they are 3/3 = 1.

The variation average with just these two users would be (total user fraction days active / number of users) = (0.4 + 1) / 2 = 0.7, or 70%. This means the average daily participation rate is 70% for users in this variation.

This metric type works with all metric delays and date windows as well.
* With no window, N eligible days = `DATE_DIFF(now(), exposure_timestamp, 'day') + 1`; in other words, the days between the exposure timestamp and now (or the end of the experiment).
* With a conversion window, N eligible days = `DATE_DIFF(LEAST(now(), exposure_timestamp + window_hours), exposure_timestamp, 'day') + 1`; in other words, the days between the exposure timestamp and the end of the conversion window (or now if that is earlier than the end of the conversion window).
* With a lookback window, N eligible days = `DATE_DIFF(now(), GREATEST(exposure_timestamp, now() - window_hours), 'day') + 1`; in other words, the days between the later of the exposure timestamp or the beginning of the lookback window and now (or the end of the experiment, if it is ended).

Other features of note:
- Does not work with metric analysis on FMID for now
- Does not work with product analytics and needs to be blocked there @mknowlton89 

TODO:
- [x] Check all TODOs in PR and self-review
- [ ] Update docs
- [x] Add "format as percent" option to modal

### Testing

- [x] Working with no, a conversion, and lookback window in BQ with and without incremental refresh
- [x] Changes to unit test sql snapshot are as expected (we have to pass timestamp around more and we removed a redundant parentheses)

### Screenshots
Screenshot 2025-12-09 at 2 50 23 PM

Screenshot 2025-12-09 at 2 50 30 PM

Screenshot 2025-12-09 at 2 50 15 PM
Screenshot 2025-12-09 at 2 50 08 PM
Screenshot 2025-12-09 at 2 49 53 PM

Generated value query in experiment context with no, lookback, and conversion windows:
```
--- no conversion window      
COUNT(DISTINCT umj.m1_value) / GREATEST(
        date_diff(
          CAST('2025-12-09 22:32:53' AS TIMESTAMP),
          CAST(MIN(umj.timestamp) AS TIMESTAMP),
          DAY
        ) + 1,
        1
      ) AS m1_value,
--- 3 day lookback window
      COUNT(DISTINCT umj.m2_value) / GREATEST(
        date_diff(
          CAST('2025-12-09 22:32:53' AS TIMESTAMP),
          --- later of exposure or 3 days ago
          GREATEST(
            CAST(MIN(umj.timestamp) AS TIMESTAMP),
            CAST('2025-12-06 22:32:53' AS TIMESTAMP)
          ),
          DAY
        ) + 1,
        1
      ) AS m2_value,
--- 4 day conversion window
      COUNT(DISTINCT umj.m3_value) / GREATEST(
        date_diff(
          -- earlier of now or 4 days after exposure
          LEAST(
            CURRENT_TIMESTAMP(),
            DATETIME_ADD(
              CAST(MIN(umj.timestamp) AS TIMESTAMP),
              INTERVAL 96 HOUR
            )
          ),
          CAST(MIN(umj.timestamp) AS TIMESTAMP),
          DAY
        ) + 1,
        1
      ) AS m3_value
```

### DIFF
--- a/packages/front-end/components/FactTables/FactMetricModal.tsx
+++ b/packages/front-end/components/FactTables/FactMetricModal.tsx
@@ -1718,7 +1718,6 @@ export default function FactMetricModal({
             stddev: DEFAULT_PROPER_PRIOR_STDDEV,
           };
         }
-        // TODO validation for new type
 
         if (values.metricType === "ratio" && !values.denominator)
           throw new Error("Must select a denominator for ratio metrics");
@@ -2026,8 +2025,6 @@ export default function FactMetricModal({
                   ) {
                     return;
                   }
-
-                  // TODO Validation for metric type
 
                   // always reset delay value when switching away from retention
                   if (
@@ -2749,12 +2746,12 @@ export default function FactMetricModal({
                         {type === "ratio" || type === "dailyParticipation" ? (
                           <Box mb="1">
                             <Checkbox
-                              label="Format average as a percentage"
+                              label="Format variation value as a percentage"
                               value={form.watch("displayAsPercentage") ?? false}
                               setValue={(v) =>
                                 form.setValue("displayAsPercentage", v === true)
                               }
-                              description="Will render variation means as a percentage rather than a proportion (e.g. 34% instead of 0.34)."
+                              description="Will render variation values as a percentage rather than a proportion (e.g. 34% instead of 0.34)."
                             />
                           </Box>
                         ) : null}


packages/back-end/src/types/Integration.ts
```diff
@@ -103,6 +103,18 @@ export type CovariateFirstExposureSettings = {
...
+/**
+ * Transformation function applied to aggregated metric values at the user level.
+ * For most metrics, this is an identity function that returns the column unchanged.
+ * For dailyParticipation metrics, this divides by the participation window to
+ * produce a participation rate.
+ */
+export type AggregatedValueTransformation = (
+  column: string,
+  initialTimestampColumn: string,
+  analysisEndDate: Date,
```
jdorn: Nit: use named parameters here

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -1966,6 +1991,12 @@ export default function FactMetricModal({
                             calculate a complex value by dividing two different
                             numeric columns in your fact tables.
                           </div>
+                          <div className="mb-2">
+                            <strong>Daily Participation</strong> metrics
+                            calculate the average number of days since exposure
```
jdorn: Should it be the "average percent of days" instead of "average number of days"

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -1144,6 +1147,8 @@ FROM
 GROUP BY variation`.trim();
 
   switch (type) {
+    // TODO
+    case "dailyParticipation":
```
jdorn: I think we just need to move this case to be part of the `mean` block instead of retention/proportion.

Then, we just have to change the experimentSQL:

```sql
SUM(m.value / u.days_since_exposure) as numerator,
```

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -107,51 +107,59 @@ function MetricType({
...
   quantileType,
 }: {
-  type: "proportion" | "retention" | "mean" | "ratio" | "quantile";
+  type:
+    | "proportion"
+    | "retention"
+    | "mean"
+    | "ratio"
+    | "quantile"
+    | "dailyParticipation";
```
Comment (jdorn) [inline packages/front-end/pages/fact-metrics/[fmid].tsx]: Should we derive this from `FactMetricInterface` instead?

```ts
type: FactMetricInterface["metricType"]
```

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -6258,12 +6242,87 @@ ${this.selectStarLimit("__topValues ORDER BY count DESC", limit)}
...
+    } else if (
+      metric.windowSettings.type === "conversion" &&
+      !overrideConversionWindows
+    ) {
+      endDateString = `LEAST(${this.getCurrentTimestamp()}, ${this.addHours(startDateString, windowHours)})`;
+    }
+
+    return this.ensureFloat(
+      `GREATEST(${this.dateDiff(startDateString, endDateString)} + 1, 1)`,
+    );
```
jdorn, line 6294: Throughout this function, I think we want to use `toTimestamp` instead of `castToTimestamp` when converting a javascript Date object to a timestamp.

This is moot if we do the above, but wrapping strings in double quotes is a syntax error in a few SQL engines IIRC.

lukesonnet: This won't work since DATE_DIFF requires more explicit types to be passed. So I'm using `toTimestamp` and then also casting it to timestamp.

packages/front-end/components/FactTables/FactMetricModal.tsx
lukesonnet: We should change it here so that it is more accurate, IMO. I also think most people don't look at the experimentSQL.

packages/front-end/services/metrics.tsx
```diff
@@ -336,6 +336,13 @@ export function getExperimentMetricFormatter(
 
   // Fact metric
   switch (metric.metricType) {
+    case "dailyParticipation":
+      return (() => {
+        if (metric.displayAsPercentage) {
+          return formatPercent;
+        }
+        return formatNumber;
+      })();
```
ahdriel: question: why is this an `iife` and not just `if/return`?

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -107,51 +107,53 @@ function MetricType({
...
+          {quantileType === "unit" ? "after aggregating per user" : ""}
+        </div>
+      );
+    case "dailyParticipation":
+      return (
+        <div>
+          <strong>Daily Participation Metric</strong> - The average of the
+          percentage of days after exposure that a user is in the Fact Table
+        </div>
+      );
```
Comment (ahdriel) [inline packages/front-end/pages/fact-metrics/[fmid].tsx:155]: Can we add the following to make it exhaustive?

```
default: {
  const exhaustiveCheck: never = type;
  throw new Error(`Unhandled MetricType type: ${exhaustiveCheck}`);
}
```

packages/front-end/components/Metrics/FactMetricTypeName.tsx
```diff
@@ -0,0 +1,16 @@
+import { FactMetricType } from "back-end/types/fact-table";
+import { capitalizeFirstLetter } from "@/services/utils";
+
+export default function FactMetricTypeName({ type }: { type: FactMetricType }) {
```
ahdriel: question: Should we make this handle both types of Metrics by default?

Also wdyt about `MetricTypeDisplayName` to emphasize that this is the formatting that we want to show users vs internal names.

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -1712,6 +1718,7 @@ export default function FactMetricModal({
             stddev: DEFAULT_PROPER_PRIOR_STDDEV,
           };
         }
+        // TODO validation for new type
```
ahdriel: Before we merge?

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -1995,6 +2027,8 @@ export default function FactMetricModal({
                     return;
                   }
 
+                  // TODO Validation for metric type
```
ahdriel: Before we merge (2)?

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -2687,19 +2746,16 @@ export default function FactMetricModal({
                           loseRiskRegisterField={form.register("loseRisk")}
                           riskError={riskError}
                         />
-                        {type === "ratio" ? (
+                        {type === "ratio" || type === "dailyParticipation" ? (
                           <Box mb="1">
                             <Checkbox
-                              label="Format ratio as a percentage"
+                              label="Format average as a percentage"
```
ahdriel: nit: Should this be `value`? or does `average` work for ratio too?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3224,6 +3266,7 @@ export default abstract class SqlIntegration
         , __userMetricJoin${f.index === 0 ? "" : f.index} as (
           SELECT
             d.variation AS variation
+            , d.timestamp AS timestamp
```
ahdriel, line 3269: question: Why do we need this now while we didn't need it before?

lukesonnet: Because it is used in the __userMetricAgg step to compute the days since exposure.

packages/front-end/components/FactTables/FactMetricModal.tsx
lukesonnet: Ah, it's done, thanks. I set the fields I needed to set.

packages/front-end/components/FactTables/FactMetricModal.tsx
lukesonnet: Same as above.

packages/front-end/components/FactTables/FactMetricModal.tsx
lukesonnet: It's the final average, not the "value" which we often mean to use the user level value in places like psuedo-code. It's pretty loose though. I originally called this "value", went to "Format variation mean as a percentage" and then settled here, but you're right, both are imprecise. I'll write "variation value", but nothing is really great.

packages/front-end/services/metrics.tsx
lukesonnet: Typescript was complaining the first way I did it, I must have had a typo or something. I have no idea why. I'll simplify, thanks.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Previously we only needed it in the CASE WHEN statement (`addCaseWhenTimeFilter`) to filter the applicable rows. We could also compute some days since exposure here but it makes more sense metric by metric in the next step.

packages/front-end/components/Metrics/FactMetricTypeName.tsx
lukesonnet: Renamed it. TBH I don't really want to spend time thinking about legacy metrics at all, so I didn't want to try and cover all those use cases.

ahdriel (approved):
:shipit:

PR #4951, opened 2025-12-05
Ji/move types and validators to shared 5

tzjames:
### Features and Changes

Move a bunch more validators and types to shared.  
The import style for validators is now to reference the file directly, as it save us having to list it in the index.ts file. 

Everything was a straight up move here except that within BaseModel:
- moved CreateProps and UpdateProps to shared/types/base-model.ts
- moved baseSchema to shared/validators/base-model.ts

Also I made changes to: 
- tsconfig - vscode kept adding files one by one automatically.  So I added the whole dir.

### Testing

`yarn build`


packages/back-end/tsconfig.json
```diff
@@ -24,5 +24,11 @@
     // This improves issue grouping in Sentry.
     "sourceRoot": "/"
   },
-  "include": ["src/**/*", "types/**/*.d.ts", "typings/**/*.d.ts"]
+  "include": [
+    "src/**/*",
+    "types/**/*.d.ts",
+    "typings/**/*.d.ts",
+    "../shared/src/validators/**/*",
+    "../shared/types/**/*.d.ts"
```
jdorn: Does this mess up the directory structure of `dist`?

jdorn (approved):

PR #4957, opened 2025-12-06
Allow images in public report markdown to be shown

tzjames:
### Features and Changes
Similar to https://github.com/growthbook/growthbook/pull/4746 and https://github.com/growthbook/growthbook/pull/4802 public reports also have metadata that could have images in them.  If the report is public we should allow anyone to see it. 

### Testing

Create a sharable public report with a description with an image in it.
Open it in a private browser.
See the image.
In dev console, network tab find the "upload/public-signed-url" endpoint.  
Copy as cURL.
In a terminal, run it and see it return a signed url.
Make the public report private again.
Rerun the cURL.
See it return 
```
{
  "status": 404,
  "message": "Report is not publicly shared"
}
```

Create a sharable public experiment with a image variations in it.
Open it in a private browser.
See the image.
In dev console, network tab find the "upload/public-signed-url" endpoint.  
Copy as cURL.
In a terminal, run it and see it return a signed url.
Make the public experiment private again.
Rerun the cURL.
See it return 
```
{
  "status": 404,
  "message": "Experiment is not publicly shared"
}
```

### DIFF
--- a/packages/back-end/src/routers/upload/upload.controller.ts
+++ b/packages/back-end/src/routers/upload/upload.controller.ts
@@ -18,7 +18,10 @@ import {
 } from "back-end/types/upload";
 import { UPLOAD_METHOD } from "back-end/src/util/secrets";
 import { getExperimentByUid } from "back-end/src/models/ExperimentModel";
-import { getReportsByExperimentId } from "back-end/src/models/ReportModel";
+import {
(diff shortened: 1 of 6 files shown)


bryce-fitzsimons (approved):
not tested, appears safe though

PR #4960, opened 2025-12-07
science code for metric drilldown results

lukebrawleysmith:
### Features and Changes

This code creates results for Metric Drilldowns table.  For a capped, CUPED-adjusted metric with an informative prior run under stratified analysis, Metric Drilldowns provides the following results:

1. CUPED-unadjusted
2. uncapped
3. flat prior
4. post-stratification turned off

if any of these features are irrelevant (e.g., CUPED is not turned on), then the table will not show the comparison row (e.g.,there will be no row for CUPED-unadjusted).  
### Testing

I created metrics that have CUPED turned off, are uncapped, etc. and matched numbers against the supplemental results in the PR.


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3417,7 +3453,10 @@ export default abstract class SqlIntegration
            , ${this.castToString(`'${data.id}'`)} as ${data.alias}_id
             ${
               data.isPercentileCapped
-                ? `, MAX(COALESCE(cap${numeratorSuffix}.${data.alias}_value_cap, 0)) as ${data.alias}_main_cap_value`
+                ? `
+                , SUM(COALESCE(m.${data.alias}_value, 0)) AS ${data.alias}_main_sum_uncapped
```
lukesonnet: Instead of typing in `COALESCE(m.${data.alias}_value, 0)` here, I think you should make a version of uncapped verisons of the `capCoalesceMetric` and `capCoalesceDenominator` which is stored in the `MetricData` by the `getMetricData` method. Create alternative strings called something like `uncappedCoalesceMetric` and `uncappedCoalesceDenominator` (and any other similar values) and use those here.

packages/back-end/src/services/stats.ts
```diff
@@ -266,10 +344,13 @@ export function getMetricSettingsForStatsEngine(
   const metric = cloneDeep<ExperimentMetricInterface>(metricDoc);
   applyMetricOverrides(metric, settings);
 
-  const denominatorDoc =
-    metric.denominator && !isFactMetric(metric)
-      ? metricMap.get(metric.denominator)
-      : undefined;
+  let denominatorDoc: ExperimentMetricInterface | undefined = undefined;
```
lukesonnet: I'm not sure I understand why this change is necessary.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3472,6 +3515,21 @@ export default abstract class SqlIntegration
                 ${
                   data.regressionAdjusted
                     ? `
+                  ${
+                    data.isPercentileCapped
+                     ? `
+                      , SUM(COALESCE(c.${data.alias}_value, 0)) AS ${data.alias}_covariate_sum_uncapped
```
lukesonnet: Don't we need these even if the value is capped with an absolute value? I think your changes here only cover the case where the data is percentile capped. Please be sure to test with both absolute and percentile capping.

Moving the logic into `getMetricData` will help this be more obvious, but I also think that this part of the code just won't work if it is nested within the `isPercentileCapped` call.

packages/back-end/src/services/stats.ts
```diff
@@ -501,6 +583,57 @@ function parseStatsEngineResult({
...
+          const ciUncapped: [number, number] | undefined = v
+            .supplementalResultsUncapped?.ci
+            ? [
+                v.supplementalResultsUncapped.ci[0] ?? -Infinity,
+                v.supplementalResultsUncapped.ci[1] ?? Infinity,
+              ]
+            : undefined;
+
+          const ciFlatPrior: [number, number] | undefined =
+            "supplementalResultsFlatPrior" in v &&
```
lukesonnet: This shouldn't be necessary, same below in L629.

packages/back-end/src/services/stats.ts
```diff
@@ -501,6 +583,57 @@ function parseStatsEngineResult({
           const ci: [number, number] | undefined = v.ci
             ? [v.ci[0] ?? -Infinity, v.ci[1] ?? Infinity]
             : undefined;
+
+          const ciCupedUnadjusted: [number, number] | undefined = v
```
lukesonnet: I would consider using a function instead of repeating this logic a bunch of times:

```
parseStatsEngineCI(ci: [number | undefined, number | undefined]): [number, number] | undefined { ... }
```

packages/back-end/src/services/stats.ts
```diff
@@ -501,6 +583,57 @@ function parseStatsEngineResult({
...
+
+          // Update CI values in supplemental results
+          if (v.supplementalResultsCupedUnadjusted) {
+            v.supplementalResultsCupedUnadjusted.ci = ciCupedUnadjusted;
+          }
+          if (v.supplementalResultsUncapped) {
+            v.supplementalResultsUncapped.ci = ciUncapped;
+          }
+          if (
+            "supplementalResultsFlatPrior" in v &&
```
lukesonnet, line 530: this shouldn't be needed either.

packages/stats/gbstats/models/statistics.py
```diff
@@ -654,3 +654,46 @@ class BanditPeriodDataRatio:
 class BanditPeriodDataCuped:
     stats: List[RegressionAdjustedStatistic]
     weights: List[float]
+
+
+def create_theta_adjusted_statistics(
```
lukesonnet, line 659: I don't love that we keep moving this around. Why are we moving it again?

packages/stats/gbstats/gbstats.py
lukesonnet: I think we've deleted this method. Try merging main again and getting rid of all these changes.

packages/stats/gbstats/models/results.py
```diff
@@ -64,27 +64,79 @@ class PowerResponse:
...
+    ci: List[float]
+    uplift: Uplift
+    errorMessage: Optional[str]
+
+
+@dataclass
+class BayesianTestResultNoDefaults(TestResultNoDefaults):
     chanceToWin: float
-    risk: Tuple[float, float]
+    risk: List[float]
```
lukesonnet: Why did we make this change?

packages/stats/gbstats/models/results.py
```diff
@@ -64,27 +64,79 @@ class PowerResponse:
...
     errorMessage: Optional[str]
-    power: Optional[PowerResponse]
 
 
+# TODO: talk to Sonnet about better way to handle defaults
 @dataclass
-class BayesianVariationResponse(BaseVariationResponse):
+class TestResultNoDefaults:
+    expected: float
+    ci: List[float]
```
lukesonnet: Why is this not a `ResponseCI`?

packages/stats/gbstats/models/results.py
```diff
@@ -64,27 +64,79 @@ class PowerResponse:
 
 
 @dataclass
-class BaseVariationResponse(BaselineResponse):
+class BaseExperimentResults:
```
lukesonnet: This doesn't seem to be used at all now?

packages/stats/gbstats/gbstats.py
```diff
@@ -476,6 +518,48 @@ def run_mid_experiment_power(
...
+            n=stat.n,
+            m_statistic=m_statistic,
+            d_statistic=d_statistic,
+            m_d_sum_of_products=stat.m_post_d_post_sum_of_products,
+        )
+    return stat
+
+
 # Run A/B test analysis for each variation and dimension
 def analyze_metric_df(
```
lukesonnet, line 511: Is there a way for us to move all of this bespoke logic outside of `analyze_metric_df` so that it continues to return a standard `DimensionResponse` and instead we just customize the `metric_data` and some of the `MetricSettingsForStatsEngine` and `AnalysisSettingsForStatsEngine` to make the stats engine do the right think for each of these events?

Then we can have one helper which does the following

`create_supplemental_analysis_settings(metric_data, metric, analysis): SupplementalAnalyses` -> Which creates something like a new dataclass of:
```
@dataclass
class SupplementalAnalysisSettings:
   metric_data: List[DimensionMetricData] <- this would have capped replaced with uncapped columns
   metric: MetricSettingsForStatsEngine <- fine tune these for the new analysis
   analysis: AnalysisSettingsForStatsEngine <- fine tune these for the new analysis


@dataclass
class SupplementalAnalyses:
    supplementalAnalysisUncapped: Optional[SupplementalAnalysisSettings]
    supplementalAnalysisFlatPrior: Optional[SupplementalAnalysisSettings]
    .... and so on
```

Then in `process_analysis` you do the main `analyze_metric_df` and then you also build the supplemental analyses, then you go into that object and run `analyze_metric_df` for each of the supplemental analysis, then add them to the result, pruning down the result to just the fields you need for the supplemental analysis result, That way you don't have to fork any of the core code at all.

IDK if this is possible, but I'd rather just try and leave the core code as it is rather than weaving in all these supplemental analyses throughout gbstats.py.

packages/stats/gbstats/models/statistics.py
lukesonnet: Maybe this is just a merge with main issue?

packages/stats/gbstats/gbstats.py
lukebrawleysmith: done

packages/back-end/src/integrations/SqlIntegration.ts

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: yes, updated

packages/back-end/src/services/stats.ts
lukebrawleysmith: What shouldn't be necessary?  We have to reformat the python output in TS to correctly handle infinity.  I did put this in a method, though, per your comment below.

packages/back-end/src/services/stats.ts
lukebrawleysmith: done

packages/back-end/src/services/stats.ts
lukebrawleysmith: Can you please lmk what is unneeded?

packages/stats/gbstats/models/statistics.py
lukebrawleysmith: I don't know why this is showing as new.  The same code in the same location is currently in main:
https://github.com/growthbook/growthbook/blob/main/packages/stats/gbstats/models/statistics.py

Perhaps I deleted and then undeleted at some point.  I'm sorry for the churn.

packages/stats/gbstats/gbstats.py
lukebrawleysmith: done

packages/stats/gbstats/models/results.py
lukebrawleysmith: changed it to ResponseCI

packages/stats/gbstats/models/results.py
lukebrawleysmith: removed

packages/stats/gbstats/models/results.py
lukebrawleysmith: I removed this object, pls ignore

lukebrawleysmith:
Removing the `FrequentistTestResultNoDefaults` object and using a ResponseCI for intervals in both `FrequentistTestResult` and `BayesianTestResult` resulted in extra lines due to type checking.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3747,11 +3866,18 @@ export default abstract class SqlIntegration
 
     // Get capping settings and final coalesce statement
     const isPercentileCapped = isPercentileCappedMetric(metric);
+    const isAbsoluteCapped = isAbsoluteCappedMetric(metric);
 
     const denominatorIsPercentileCapped = denominator
       ? isPercentileCappedMetric(denominator)
       : false;
 
+    const denominatorIsAbsoluteCapped =
```
lukesonnet: nit: Why not use `isAbsoluteCappedMetric(denominator)`?

packages/back-end/src/services/experimentQueries/constants.ts
```diff
@@ -31,6 +31,20 @@ export const RATIO_METRIC_CUPED_FLOAT_COLS = [
...
+export const RATIO_METRIC_FLOAT_COLS_UNCAPPED = RATIO_METRIC_FLOAT_COLS.map(
+  (col) => col + "_uncapped",
+);
+
+export const BASE_METRIC_CUPED_FLOAT_COLS_UNCAPPED =
+  BASE_METRIC_CUPED_FLOAT_COLS.map((col) => col + "_uncapped");
+
+export const RATIO_METRIC_CUPED_FLOAT_COLS_UNCAPPED =
+  RATIO_METRIC_CUPED_FLOAT_COLS.map((col) => col + "_uncapped");
+
```
lukesonnet, line 47: mustfix: When adding new top level constants here, you also need to update `experimentQueries.ts` directly to add these columns if capping is set so that we stay under the 1000 columns per query limit. So you'll need to modify `getNonQuantileFloatColumns` to check for capping; if there is capping, then you need to add these columns to the column response as they will be added when the metric has capping.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2755,6 +2873,18 @@ export default abstract class SqlIntegration
       capValueCol: `${alias}_denominator_cap`,
       columnRef: metric.denominator,
     });
+    const uncappedCoalesceMetric = `
```
lukesonnet: shouldfix: Shouldn't these also use `this.capCoalesceValue` like you did for the legacy metric? Otherwise I don't think it properly applies for threshold metrics.

packages/back-end/src/services/stats.ts
lukesonnet: The entire `  "supplementalResultsFlatPrior" in v &&` check. Notice how you don't have to do that kind of `in` check for your other if statements here.

Smith response: 
Removing the FrequentistTestResultNoDefaults object and using a ResponseCI for intervals in both FrequentistTestResult and BayesianTestResult resulted in extra lines due to type checking.

packages/shared/types/stats.d.ts
```diff
@@ -22,12 +22,15 @@ export type PValueErrorMessage =
...
-interface BaseVariationResponse {
+interface BaselineResponse {
   cr: number;
   value: number;
   users: number;
   denominator?: number;
   stats: MetricStats;
+}
+
+interface BaseExperimentResults {
```
lukesonnet: nit: Is `Experiment` the right name? Feels "bigger" than variation.

Let's just call them `BaseVariationResults` and those end up in `BayesianVariationResults` and those are used to construct the full `BayesianVariationResponse`, etc...

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1379,74 +1414,156 @@ export default abstract class SqlIntegration
...
+        );
+        Object.assign(
+          result,
+          parseFloatField(
+            row,
+            "denominator_post_denominator_pre_sum_product_uncapped",
+          ),
+        );
+
+        return result as ExperimentMetricQueryResponseRows[number];
```
lukesonnet: I don't think we should make this change to the way we parse the results as it requires this `as` statement. I know there's a lot of boilerplate here, but I'd rather refactor it as part of a later PR or consider another approach that doesn't require this `as` statement.

packages/stats/gbstats/bayesian/bandits.py
```diff
@@ -184,7 +185,7 @@ def compute_result(self) -> BanditResponse:
         return BanditResponse(
             users=self.variation_counts.tolist(),
             cr=(self.variation_means).tolist(),
-            ci=credible_intervals,
+            ci=cast(List[ResponseCI], credible_intervals),
```
lukesonnet: Why cast? Can't you fix the method that returns `credible_intervals` in the first place to return the right type instead of casting?

packages/stats/gbstats/models/results.py
```diff
@@ -60,29 +60,57 @@ class PowerResponse:
     scalingFactor: Optional[float]
 
 
-ResponseCI = Tuple[Optional[float], Optional[float]]
+@dataclass
+class FrequentistTestResult(TestResult):
```
lukesonnet, line 115: Please keep the names in this class in line with the names in the stats package so theya re easy to compare.

packages/stats/gbstats/gbstats.py
```diff
@@ -746,21 +849,234 @@ def process_analysis(
...
         metric_data=reduced_metric_data,
-        num_variations=len(var_names),
+        num_variations=num_variations,
         metric=metric,
         analysis=analysis,
     )
 
+    cuped_adjusted = metric.statistic_type in ["ratio_ra", "mean_ra"]
+    metric_capped = metric.capped
+    analysis_bayesian = analysis.stats_engine == "bayesian"
```
lukesonnet: Shouldn't this instead check if bayesian *and* non-flat priors are used?

packages/stats/gbstats/gbstats.py
```diff
@@ -746,21 +849,234 @@ def process_analysis(
...
+            metric_data=reduced_metric_data,
+            num_variations=num_variations,
+            metric=metric_cuped_adjusted,
+            analysis=analysis,
+        )
+    else:
+        result_cuped_adjusted = None
+    if metric_capped:
+        analysis_uncapped = dataclasses.replace(analysis, use_uncapped_metric=True)
+        result_uncapped = analyze_metric_df(
```
lukesonnet, line 897: What I'd prefer over any of the `use_uncapped` flags in the rest of the code is simply replacing the capped data with the uncapped values in `reduced_metric_data` and then the rest of the logic is completely unchanged. So you'd need a new method here, e.g. `replace_capped_data_with_uncapped_data(metric_data)` that takes the old data and returns a version with the capped columns replaced with the uncapped columns (without editing in place, so you might need to use `reduced_metric_data.copy()`).

Then you can replace `use_uncapped` everywhere else in the code.

packages/shared/types/stats.d.ts
```diff
@@ -124,6 +145,7 @@ export interface AnalysisSettingsForStatsEngine {
   num_goal_metrics: number;
   one_sided_intervals?: boolean;
   post_stratification_enabled?: boolean;
+  use_uncapped_metric?: boolean;
```
lukesonnet: What is this argument used for? Is it still needed if we do my other comment where we just replace the data when doing the uncapped analysis?

packages/back-end/src/services/stats.ts
lukebrawleysmith: supplementalResultsFlatPrior only exists on BayesianVariationResponse, not FrequentistVariationResponse, so this properly distinguishes the types

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -4129,8 +4305,16 @@ export default abstract class SqlIntegration
...
-          ? ", MAX(COALESCE(capd.value_cap, 0)) as denominator_cap_value"
+        denominatorIsPercentileCapped || denominatorIsAbsoluteCapped
+          ? `, SUM(${uncappedCoalesceDenominator}) AS denominator_sum_uncapped
+             , SUM(POWER(${uncappedCoalesceDenominator}, 2)) AS denominator_sum_squares_uncapped
+             , SUM(${uncappedCoalesceMetric} * ${uncappedCoalesceDenominator}) AS main_denominator_sum_product_uncapped
+             ${
+               isPercentileCapped
+                 ? `
+             , MAX(COALESCE(capd.value_cap, 0)) as denominator_cap_value`
+                 : ""
```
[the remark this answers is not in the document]
lukebrawleysmith: great catch!  I addressed this, thank you.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1442,9 +1439,68 @@ export default abstract class SqlIntegration
...
+            covariate_sum_squares_uncapped:
+              parseFloat(row.covariate_sum_squares_uncapped) || 0,
+          }),
+          ...(row.main_denominator_sum_product_uncapped !== undefined && {
+            main_denominator_sum_product_uncapped:
+              parseFloat(row.main_denominator_sum_product_uncapped) || 0,
+          }),
+          ...(row.main_covariate_sum_product_uncapped !== undefined && {
+            main_covariate_sum_product_uncapped:
+              parseFloat(row.main_covariate_sum_product_uncapped) || 0,
```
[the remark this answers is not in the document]
lukebrawleysmith: great catch!  I addressed this, thank you.

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: fixed

packages/back-end/src/services/experimentQueries/constants.ts
lukebrawleysmith: done

packages/back-end/src/services/stats.ts
lukesonnet: Oh I see. Thanks.

packages/back-end/src/integrations/SqlIntegration.ts

packages/shared/types/stats.d.ts
lukebrawleysmith: I decided to call it `TestResult` to be consistent with gbstats.  Happy to change.

packages/stats/gbstats/bayesian/bandits.py
lukebrawleysmith: `gaussian_credible_interval` method is in utils, which is math focused and is ignorant of gbstats data structures like `ResponseCI`.  I wanted to avoid using `ResponseCI` inside of `gaussian_credible_intervals.`.  Happy to modify if desired.

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: updated the code, obviating the `as` statement

packages/stats/gbstats/bayesian/bandits.py
lukesonnet: Instead of `cast` you can just type-hint above:

```
 credible_intervals: List[ResponseCI] = [
 ```

File: packages/stats/gbstats/gbstats.py (hunk 19 above)
Comment (lukebrawleysmith) [reply packages/stats/gbstats/gbstats.py]: done

File: packages/stats/gbstats/models/results.py (hunk 18 above)
Comment (lukebrawleysmith) [reply packages/stats/gbstats/models/results.py:115]: these now match the names in shared/types/stats.d.ts

File: packages/stats/gbstats/gbstats.py (hunk 20 above)
Comment (lukebrawleysmith) [reply packages/stats/gbstats/gbstats.py:897]: done

File: packages/shared/types/stats.d.ts (hunk 21 above)
Comment (lukebrawleysmith) [reply packages/shared/types/stats.d.ts]: deleted now

File: packages/stats/gbstats/bayesian/bandits.py (hunk 17 above)
Comment (lukebrawleysmith) [reply packages/stats/gbstats/bayesian/bandits.py]: good call, and updated

File: packages/shared/types/stats.d.ts (hunk 15 above)
Comment (lukesonnet) [reply packages/shared/types/stats.d.ts]: I think that looks good and consistency is the most important, IMO, to make it easy to see the linkage between the two.

File: packages/stats/gbstats/gbstats.py [hunk 24]
```diff
@@ -746,21 +809,266 @@ def process_analysis(
...
+    result_capped: Optional[List[DimensionResponseIndividual]],
+    result_flat_prior: Optional[List[DimensionResponseIndividual]],
+    result_unstratified: Optional[List[DimensionResponseIndividual]],
+) -> List[DimensionResponse]:
+    result = []
+
+    for d in range(num_dimensions):
+        this_dimension_result = core_result[d]
+        this_baseline_result = this_dimension_result.variations[0]
+        variations: list[VariationResponse] = []
```
lukesonnet: Unless we did some version upgrade, I think this will fail on 3.9 as `list` wasn't added until later. So please use `List` for type-hinting (same goes for `dict` vs. `Dict`, etc...)

Not sure why this isn't caught during CI, though...

packages/stats/gbstats/gbstats.py
```diff
@@ -746,21 +809,266 @@ def process_analysis(
...
+        df: Input pandas DataFrame.
+
+    Returns:
+        A DataFrame with updated values and '_uncapped' columns removed.
+    """
+    # Create a copy to avoid SettingWithCopy warnings or mutating the original
+    df = df.copy()
+
+    # Identify all columns that end with the suffix
+    capped_cols = [col for col in df.columns if col.endswith("_uncapped")]
```
lukesonnet: Shouldn't this be called `uncapped_cols`?

lukebrawleysmith: yes, updated

packages/stats/gbstats/gbstats.py
lukebrawleysmith: idk why either, I'm surprised.  Updated.

PR #4963, opened 2025-12-07
Move some integration types to shared

tzjames:
### Features and Changes

Within back-end/src/types/Integrations.ts there are 2 error classes, and the SourceIntegrationInterface has a reference to ReqContext which is only available in the back-end.  Most of the rest have references from the front-end and so I've moved them to /shared/types/integration.d.ts.

### Testing

`yarn build`
Click around in the preview env.


tzjames:
Hrmm the preview had this in its logs:
```
--> Build Summary:  (​)
2025/12/07 22:10:17 error releasing builder: deadline_exceeded: context deadline exceeded
```
and unfortunately the back-end isn't loading.  But despite that message preview didn't fail.    I'm not sure if it is related.  Running this locally after a build with `yarn start` does work.  I'm going to rerun the preview job as it would be good to have that verification before I land this.

PR #4968, opened 2025-12-08
chore(sql): refactor aggregation metadata to be more centralized

lukesonnet:
### Features and Changes

Removes old method (`getAggregateMetricColumn`) that was only essential for event quantiles and uses the new `getAggregationMetadata` method, displacing when/where the error gets thrown.

Also, refactor so that `getAggregationMetadata` happens within the `getMetricData` call so that it doesn't need to happen each time a metric gets used in a different CTE.

### Testing

* [x] Event quantiles with and without incremental refresh (expecting an error with it)


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3286,22 +3299,18 @@ export default abstract class SqlIntegration
               (data) =>
                 `${
                   data.numeratorSourceIndex === f.index
-                    ? `, ${this.getAggregateMetricColumn({
-                        metric: data.metric,
-                        useDenominator: false,
-                        valueColumn: `umj.${data.alias}_value`,
-                        quantileColumn: `qm.${data.alias}_quantile`,
-                      })} AS ${data.alias}_value`
+                    ? `, ${data.numeratorAggFns.fullAggregationFunction(
```
lukesonnet, line 3302: Note, we only ever use full aggregation function for event quantiles here, so the errors in `getAggregationMetadata` for partial and reaggregation shouldn't ever occur. But this isn't obvious from the developer point of view.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -6358,9 +6288,11 @@ ${this.selectStarLimit("__topValues ORDER BY count DESC", limit)}
       const fullAggregationFunction = nullIfZero
         ? (column: string) => `NULLIF(COUNT(DISTINCT ${column}), 0)`
         : (column: string) => `COUNT(DISTINCT ${column})`;
+
       return {
         intermediateDataType: "date",
-        partialAggregationFunction: (column: string) => `MAX(${column})`,
+        partialAggregationFunction: (column: string) =>
+          this.castToDate(`MAX(${column})`),
```
lukesonnet, line 6295: Ensures we match `intermediateDataType`, just another fix.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -6358,9 +6288,11 @@ ${this.selectStarLimit("__topValues ORDER BY count DESC", limit)}
...
         finalDataType: "integer",
```
ahdriel, line 6296: question: Is this `finalDataType` correct because it gets stored as epoch?

packages/back-end/test/__snapshots__/sqlintegration.test.ts.snap
```diff
@@ -1728,99 +1728,99 @@ exports[`full fact metric experiment query - bigquery generated fact metric SQL
...
...SE NULL END ) ), 0 ) ) as m0_value FROM __distinctUsers d JOIN __factTable m ON (m.user_id = d.user_id) WHERE m.timestamp >= d.min_preexposure_start AND m.timestamp < d.max_preexposure_end GROUP BY d.variation, d.user_id ) -- One row per variation/dimension with aggregations SELECT m.variation AS variation, COUNT(*) AS users, cast( 'type_proportion__cap_none__agg_none__col_special_distinctUsers__inline_none__filter_mobile__window_conversion__regadj_4d__aggfilter____3__aggfiltercol_qty' as string ) as m0_id, SUM( COALESCE( ( CASE WHEN m.m0_value >= 3 THEN 1 ELSE NULL END ), 0 ) ) AS m0_main_sum, SUM( POWER( COALESCE( ( CASE WHEN m.m0_value >= 3 THEN 1 ELSE NULL END ), 0 ), 2 ) ) AS m0_main_sum_squares, SUM( COALESCE( ( CASE WHEN c.m0_value >= 3 THEN 1 ELSE NULL END ), 0 ) ) AS m0_covariate_sum, SUM( POWER( COALESCE( ( CASE WHEN c.m0_value >= 3 THEN 1 ELSE NULL END ), 0 ), 2 ) ) AS m0_covariate_sum_squares, SUM( COALESCE( ( CASE WHEN m.m0_value >= 3 THEN 1 ELSE NULL END ), 0 ) * COALESCE( ( CASE WHEN c.m0_value >= 3 THEN 1 ELSE NULL END ), 0 ) ) AS m0_main_covariate_sum_product FROM __userMetricAgg m LEFT JOIN __userCovariateMetric c ON (c.user_id = m.user_id) GROUP BY m.variation"`;
```
ahdriel, line 1731: just checking that the `parentheses` here are intentional (or at least not harmful) around `SUM(COALESCE((umj.m0_value), 0))` & `COALESCE( ( ( CASE WHEN m.timestamp >= d.m0_preexposure_start AND m.timestamp < d.m0_preexposure_end THEN m0_value ELSE NULL END ) ), 0 )` (note the 3 parenthesis before case when).

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: finalDataType is the result of re and full aggregation, not of partial. Those are count distinct.

packages/back-end/test/__snapshots__/sqlintegration.test.ts.snap
lukesonnet: They should be non-harmful!

PR #4969, opened 2025-12-08
Don't include quantile metrics in Metric Explorer dropdown

mknowlton89:
### Features and Changes

Quantile metrics are not supported in the Metric Explorer block - this PR prevents them from being included in the list of `Metrics` available to build a visualization from in the Metric Explorer.

### Dependencies

- None

### Testing

- [x] Ensure that no metrics with a `metricType` that equals `quantile` is included in the `factMetricOptions` list.

### DIFF
--- a/packages/back-end/src/integrations/GoogleAnalytics.ts
+++ b/packages/back-end/src/integrations/GoogleAnalytics.ts
@@ -32,6 +32,7 @@ import {
   MetricAnalysisParams,
   ExperimentFactMetricsQueryResponse,
   UserExperimentExposuresQueryResponse,
+  DropMetricSourceCovariateTableQueryParams,
   InsertMetricSourceCovariateDataQueryParams,
   CreateMetricSourceCovariateTableQueryParams,
 } from "shared/types/integrations";
@@ -216,6 +217,11 @@ export default class GoogleAnalytics implements SourceIntegrationInterface {
     _query: string,
     _setExternalId: ExternalIdCallback,
   ): Promise<IncrementalWithNoOutputQueryResponse> {
+    throw new Error("Method not implemented.");
+  }
+  getDropMetricSourceCovariateTableQuery(
+    _params: DropMetricSourceCovariateTableQueryParams,
+  ): string {
     throw new Error("Method not implemented.");
   }
   getCreateMetricSourceCovariateTableQuery(
--- a/packages/back-end/src/integrations/Mixpanel.ts
+++ b/packages/back-end/src/integrations/Mixpanel.ts
@@ -35,6 +35,7 @@ import {
   MetricAnalysisParams,
   ExperimentFactMetricsQueryResponse,
   UserExperimentExposuresQueryResponse,
+  DropMetricSourceCovariateTableQueryParams,
   CreateMetricSourceCovariateTableQueryParams,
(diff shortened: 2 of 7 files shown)


packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -189,15 +190,13 @@ export default function EditSingleBlock({
   const blockContext = useBlockContext(blockId);
 
   // TODO: does this need to handle metric groups
-  const factMetricOptions = useMemo(
-    () =>
+  const factMetricOptions = useMemo(() => {
+    const unSupportedMetricTypes: FactMetricType[] = ["quantile"];
```
ahdriel: suggestion: Should we move this to a higher-level const somewhere else?

Right now I think it is a bit hard to find it if I were to add a new metric type to the unsupported list.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -189,15 +190,13 @@ export default function EditSingleBlock({
...
+
+    return (
       factMetrics
         // Filter fact metrics to only include those that are in 'All Projects' or have all of the projects in the projects list
```
ahdriel, line 208: We should move this comment with the `if(!projects.length...)` code right?

mknowlton89: Oh shoot - yea, it should've followed the changes. I'll move it.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
mknowlton89: For sure - I added it here initially for simplicity as if we define it outside of the useMemo hook it then needs to get added as a dependency, and wrapped in it's own useMemo.

The other option which I currently implemented is moving it to the `shared/constants.ts` file - I did rename it as I felt it needed additional context with it being defined outside of the `EditSingleBlock` file. Let me know what you think.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -189,30 +190,33 @@ export default function EditSingleBlock({
...
+      .filter((factMetric) => {
+        // Always include the existing fact metric. This will prevent issues if the fact metric or the dashboard's projects have changed since the block was created.
+        if (
+          blockHasFieldOfType(block, "factMetricId", isString) &&
+          factMetric.id === block.factMetricId
+        ) {
+          return true;
+        }
+
+        // Quantile metrics are not supported in the Dashboard Editor
```
ahdriel: Should we remove this comment now?

lukesonnet (approved):
ty michael

PR #4974, opened 2025-12-08
Allow cross-project prerequisite targeting

bryce-fitzsimons:
Allow users to select prerequisite flags that are outside of the current feature/experiment's project.
- Show a warning if projects do not match.
- No project permission checks (since we're not altering the parent project, just creating a dependency from this project)

image

image

### DIFF
--- a/packages/back-end/package.json
+++ b/packages/back-end/package.json
@@ -93,7 +93,7 @@
     "nodemailer": "^6.9.12",
     "number-format.js": "^2.0.9",
     "nunjucks": "^3.2.4",
-    "openai": "5.11.0",
+    "openai": "6.10.0",
     "openid-client": "^5.1.10",
     "pg": "^8.9.0",
     "pino": "^9.7.0",
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -974,13 +974,8 @@ export default abstract class SqlIntegration
       segment: params.segment,
     });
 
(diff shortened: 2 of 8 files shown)


packages/front-end/components/Features/PrerequisiteModal.tsx
```diff
@@ -260,6 +331,21 @@ export default function PrerequisiteModal({
               >
                 {label}
               </span>
+              {projectName ? (
+                <span
+                  className="text-muted small float-right position-relative"
+                  style={{ top: 3 }}
+                >
+                  project: <strong>{projectName}</strong>
```
jdorn, line 340: Does this look ok for really long project names? Should we truncate here?

packages/front-end/components/Features/PrerequisiteTargetingField.tsx
```diff
@@ -299,6 +358,21 @@ export default function PrerequisiteTargetingField({
                             >
                               {label}
                             </span>
+                            {projectName ? (
+                              <span
+                                className="text-muted small float-right position-relative"
+                                style={{ top: 3 }}
+                              >
+                                project: <strong>{projectName}</strong>
```
jdorn, line 367: Same here about truncating

packages/front-end/components/Features/PrerequisiteModal.tsx
bryce-fitzsimons: Added truncation. I didn't observe this with my fairly long project names, but it certainly could happen.

PR #4987, opened 2025-12-09
handle dimensions the same

ahdriel:
### Features and Changes

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -2053,9 +2053,9 @@ export default abstract class SqlIntegration
               return `, ${this.getDimensionInStatement(
                 d.id,
                 d.specifiedSlices,
-              )} AS dim_${d.id}`;
+              )} AS dim_exp_${d.id}`;
             }
-            return `, e.${d.id} AS dim_${d.id}`;
+            return `, e.${d.id} AS dim_exp_${d.id}`;
           })
           .join("\n")}
       FROM


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2053,9 +2053,9 @@ export default abstract class SqlIntegration
               return `, ${this.getDimensionInStatement(
                 d.id,
                 d.specifiedSlices,
-              )} AS dim_${d.id}`;
```
lukesonnet, line 2056: I don't thnk we want to make changes here. We want the full data to pass through in this section, and do the case when later on when we actually consume it.

packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -770,7 +780,7 @@ const startExperimentIncrementalRefreshQueries = async (
       name: TRAFFIC_QUERY_NAME,
       query: integration.getExperimentAggregateUnitsQuery({
         ...unitQueryParams,
-        dimensions: dimensionsForTraffic,
+        dimensions: dimensionObjs.length ? dimensionObjs : dimensionsForTraffic,
```
lukesonnet, line 783: I think it has to be a intersection of the `dimensionsForTraffic` and `experimentDimensions`

ahdriel:
Closing as we are going with a different solution.

PR #4990, opened 2025-12-09
Composable (recursive) condition groups

bryce-fitzsimons:
Add support for nested / recursive saved condition groups.

image

Built-in protection for circular dependencies:

TODO: 
- [x] Fix `savedGroupUpdated` logic.  Needs to take into account nested saved groups when figuring out which payloads should be refreshed.
- [x] Update statsig importer to take advantage of this new capability

### DIFF
--- a/packages/back-end/src/routers/saved-group/saved-group.controller.ts
+++ b/packages/back-end/src/routers/saved-group/saved-group.controller.ts
@@ -75,18 +75,13 @@ export const postSavedGroup = async (
   }
 
   let uniqValues: string[] | undefined = undefined;
-  // If this is a condition group, make sure the condition is valid and not empty.
-  // When skipCycleCheck=1 (used by importers), still validate general JSON/syntax
-  // but skip saved-group cyclic/invalid reference checks so users can fix them later.
+  // If this is a condition group, make sure the condition is valid and not empty
   if (type === "condition") {
     const allSavedGroups = await getAllSavedGroups(org.id);
     const groupMap = new Map(allSavedGroups.map((sg) => [sg.id, sg]));
     const conditionRes = validateCondition(
       condition,
       groupMap,
-      // When skipCycleCheck=1, skip only saved-group *cycle* checks while still
-      // enforcing JSON validity and other saved-group errors (unknown group,
-      // invalid nested condition, max depth).
       skipCycleCheck === "1",
     );
     if (!conditionRes.success) {
--- a/packages/front-end/services/importing/statsig/transformers/savedGroupTransformer.ts
+++ b/packages/front-end/services/importing/statsig/transformers/savedGroupTransformer.ts
@@ -17,7 +17,7 @@ export async function transformStatsigSegmentToSavedGroup(
   ) => Promise<unknown>,
   project?: string,
   skipAttributeMapping: boolean = false,
-  _savedGroupIdMap?: Map<string, string>,
+  savedGroupIdMap?: Map<string, string>,
 ): Promise<
   Omit<
     SavedGroupInterface,
@@ -44,7 +44,7 @@ export async function transformStatsigSegmentToSavedGroup(
         const transformed = transformStatsigConditionsToGB(
           rule.conditions,
           skipAttributeMapping,
-          _savedGroupIdMap,
+          savedGroupIdMap,
         );
 
         // Start from the base targeting condition (if any)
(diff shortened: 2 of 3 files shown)


packages/back-end/src/routers/saved-group/saved-group.controller.ts
```diff
@@ -70,8 +76,11 @@ export const postSavedGroup = async (
 
   let uniqValues: string[] | undefined = undefined;
   // If this is a condition group, make sure the condition is valid and not empty
-  if (type === "condition") {
-    const conditionRes = validateCondition(condition);
+  // unless explicitly bypassed (used by importers that handle cycle safety).
+  if (type === "condition" && skipCycleCheck !== "1") {
```
jdorn: This will also skip non-cycle validation, which I don't think we want.

Same below

packages/front-end/components/Features/ConditionDisplay/index.tsx
```diff
@@ -216,7 +249,7 @@ function getConditionParts({
           {operatorToText(operator, renderPrerequisite)}
         </span>
         {hasMultiValues(operator) ? (
-          <MultiValueDisplay value={value} />
+          <MultiValueDisplay value={value} displayMap={displayMap} />
```
jdorn: I think this might break for actual `$inGroup` and `$notInGroup` operators (list saved groups)

packages/front-end/services/importing/statsig/transformers/savedGroupTransformer.ts
```diff
@@ -37,13 +37,85 @@ export async function transformStatsigSegmentToSavedGroup(
...
         const transformed = transformStatsigConditionsToGB(
           rule.conditions,
           skipAttributeMapping,
-          undefined, // Don't resolve saved groups during saved group creation
+          _savedGroupIdMap,
         );
+
+        // Start from the base targeting condition (if any)
+        let baseConditionObj: Record<string, unknown> = {};
         if (transformed.condition && transformed.condition !== "{}") {
```
jdorn, line 52: Will this break if a saved group ONLY has referenced nested saved groups without any other conditions?

bryce-fitzsimons: shouldn't. seems to work fine on import as well

packages/front-end/services/importing/statsig/transformers/ruleTransformer.ts
```diff
@@ -28,6 +28,7 @@ export function transformStatsigConditionsToGB(
   skipAttributeMapping: boolean = false,
   savedGroupIdMap?: Map<string, string>,
   featuresMap?: Map<string, FeatureInterface>,
+  addSavedGroupsToConditions: boolean = false,
```
jdorn: Do we still need this argument? Doesn't look like it's being used anywhere

packages/front-end/services/importing/statsig/transformers/ruleTransformer.ts
```diff
@@ -197,6 +217,20 @@ function transformTargetingConditions(
       attributeName = field || "custom_field";
     } else if (type === "unit_id" && customID) {
       attributeName = customID;
+    } else if (type === "passes_gate" || type === "fails_gate") {
```
jdorn: Is this right?  I thought `passes_gate` / `failed_gate` had to do with prerequisites not saved groups?  Is this supposed to be `passes_segment` / `failed_segment` instead?  Or is this block not needed at all and we can remove it?

packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -218,10 +222,217 @@ export const scrubSavedGroups = (
...
+  if (!condition) return false;
+
+  const src =
+    typeof condition === "object"
+      ? JSON.stringify(condition)
+      : String(condition);
+
+  const errorMarkers = [
+    SAVED_GROUP_ERROR_INVALID,
+    SAVED_GROUP_ERROR_UNKNOWN,
```
jdorn: Does the importer work with this?  Wouldn't the dummy `__unknown__` saved group ids cause an `ERROR_UNKNOWN` error?

packages/front-end/services/importing/statsig/transformers/ruleTransformer.ts
bryce-fitzsimons: this was hallucinated 😞, manually removed and tied up loose ends

packages/shared/src/sdk-versioning/sdk-payload.ts
bryce-fitzsimons: it worked, but not for the correct reason. the 2-pass system masked the failed state. will fix the validator bypass logic

PR #4991, opened 2025-12-09
fix(notebook): fix columns in notebook generation

lukesonnet:
### Features and Changes

We change the structure of our return object in gbstats as part of post-strat, and this broke notebooks.

Now we return a reasonable result dataset, although it is a different structure from before.

### Testing

Tested and NB now downloads and runs as expected.

Screenshot 2025-12-09 at 3 36 30 PM

### DIFF
--- a/packages/front-end/pages/datasources/[did].tsx
+++ b/packages/front-end/pages/datasources/[did].tsx
@@ -417,16 +417,16 @@ mixpanel.init('YOUR PROJECT TOKEN', {
                   </Frame>
                 ) : null}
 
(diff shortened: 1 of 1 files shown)


packages/stats/gbstats/gbstats.py
```diff
@@ -370,42 +370,6 @@ def decision_making_conditions(metric, analysis):
     )
 
 
-def initialize_df(
```
lukesonnet, line 373: Unused method left around from old refactor

PR #5001, opened 2025-12-11
feat(feature-eval-diagnostics): Add Feature Evaluation Diagnostics 

msamper:
### Features and Changes

Adds a new tab to the features page called **Diagnostics** that includes a new tool for viewing recent feature evaluations along with any custom metadata orgs wish to include from their data source. 
Adds a new type of query that can be added to datasources to pull feature evaluation data to power the new feature evaluation diagnostics tool.

### To-Do

- [x] Include datasource selector in Diagnostics tab instead of a modal
- [x] Only allow one feature query for now per datasource
- [x] Remove `UNION` from query generator

### Testing

[Script to generate fake feature event data](https://gist.github.com/msamper/daa7156ade08d5b159e4f6288153d269)

1. Set up a data source (Postgres is easiest) with a table with dummy feature events that include `timestamp` and `feature_key` at a minimum. (Some additional columns that users might include in a real setup are: `value`, `reason`, `user_attributes`, `unit_id`) 
3. Create a new feature or select an existing one and go to the diagnostics tab
4. Go through the setup process and configure your feature usage query using the table created in step (1)
5. Go back to the feature and run the evaluations query (make sure your dummy data includes data for this feature)
6. Play around with results and make sure sorting is working correctly

### Screenshots

Screenshot 2025-12-11 at 4 44 07 PM
Screenshot 2025-12-18 at 5 19 45 PM
Screenshot 2025-12-11 at 4 44 53 PM
Screenshot 2025-12-11 at 4 45 20 PM

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -2618,10 +2618,10 @@ export default abstract class SqlIntegration
 
     return format(
       `-- Feature Evaluation Diagnostics Query
-      WITH _featureEvalQuery AS (
+      WITH __featureEvalQuery AS (
         ${featureEvalQuery}
       )
-      SELECT * FROM _featureEvalQuery
+      SELECT * FROM __featureEvalQuery
       WHERE feature_key = '${featureKey}'
       ORDER BY timestamp DESC
       LIMIT ${SQL_ROW_LIMIT}


packages/back-end/types/datasource.d.ts
```diff
@@ -171,6 +171,14 @@ export interface ExposureQuery {
...
 }
 
+export interface FeatureUsageQuery {
+  id: string;
+  name: string;
+  description?: string;
+  query: string;
+  error?: string;
+}
+
```
msamper, line 178: I think maybe this should be called something else. We use the `Feature Usage` for usage analytics so it might be confusing. Open to suggestions

msamper: maybe `FeatureEvaluationsQuery` or `FeatureExposuresQuery`?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  return (
+    <Box className="contents container-fluid pagecontents">
+      <h2>Feature Evaluation Diagnostics</h2>
+      <p>
+        View recent feature evaluations along with any custom metadata
+        you&apos;ve added.
+      </p>
+
+      {datasource && !datasourceHasFeatureUsageQuery && (
```
ahdriel: issue: This ordering is weird when I have one datasource with Feature Evaluation configured and others without.

It creates an experience where selecting a different value from the dropdown adds this message above the dropdown, which breaks the hierarchy I have in mind.

What do you think about replacing the empty section that has `View recent feature evaluations` button with this message when there's no Feature Usage Query configured?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+      )}
+      <Box width="400px">
+        <SelectField
+          label="Select a Data Source"
+          labelClassName="font-weight-bold"
+          value={form.watch("datasourceId") ?? ""}
+          onChange={(newDatasource) => {
+            form.setValue("datasourceId", newDatasource);
+          }}
+          options={validDatasources.map((d) => {
```
ahdriel, line 230: question: Why are we allowing users to select DataSources instead of selecting `Feature Usage Queries` directly?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+        />
+      </Box>
+
+      <Frame mt="4">
+        <Flex direction="row" justify="end" mb="4">
+          <Tooltip
+            content="Setup a feature usage query in your data source to view feature evaluations."
+            enabled={!datasourceHasFeatureUsageQuery && !loading}
+          >
+            <Button
```
ahdriel: suggestion: We should add a `View queries` button here. Currently mine is returning 0 results and I don't know why and have no way to debug as these queries also are not saved in the Query collection.

packages/back-end/src/app.ts
```diff
@@ -510,6 +510,10 @@ app.post(
   "/query/user-exposures",
   datasourcesController.runUserExperimentExposuresQuery,
 );
+app.post(
+  "/query/feature-eval-diagnostic",
+  datasourcesController.getFeatureEvalDiagnostics,
+);
```
ahdriel, line 516: nit: should this be named `postFeatureEvalDiagnostics`?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+            >
+              {loading
+                ? "Running..."
+                : !results
+                  ? "View recent feature evaluations"
+                  : "Refresh feature evaluations"}
+            </Button>
+          </Tooltip>
+        </Flex>
+        {error && (
```
ahdriel: issue: I was getting an error but it was not surfaced here. Need to fix it.

Here's the full response from `POST /query/feature-eval-diagnostic`:
```
{
  "status": 200,
  "error": "Every derived table must have its own alias",
  "sql": "-- Feature Evaluation Diagnostics Query\nSELECT\n  *\nFROM\n  (\n    SELECT\n      *\n    from\n      `dummy-data`.`feature_events`\n    WHERE\n      feature_key = 'feature-diagnostics-test'\n  )\nORDER BY\n  timestamp DESC\nLIMIT\n  1000"
}
```

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+          <Callout status="error">
+            <strong>Error:</strong> {error}
+          </Callout>
+        )}
+        {results && results.length === 0 && (
+          <Callout status="info">No feature evaluations found.</Callout>
+        )}
+        {/* If there's feature usage data, show a table with the data */}
+        {items.length > 0 && (
```
ahdriel: minor: The initial empty state looks weird IMO with the button on the top right.

additionally, if keeping this layout, the padding/margin is larger at the bottom which should be fixed

Image

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+import Link from "@/ui/Link";
+import Tooltip from "@/ui/Tooltip";
+import EmptyState from "@/components/EmptyState";
+
+type FeatureEvaluationDiagnosticsQueryResults = {
+  rows?: FeatureEvalDiagnosticsQueryResponseRows;
+  statistics?: QueryStatistics;
+  error?: string;
+  sql?: string;
+  truncated?: boolean;
```
ahdriel: question: Is `truncated` in use?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+};
+
+// Helper function to format a value for display
+const formatDisplayValue = (value: unknown): string => {
+  if (value === null) {
+    return "";
+  } else if (value === undefined) {
+    return "";
```
ahdriel: question/suggestion: Should we differentiate between `(empty string)`, `null` and `undefined` in the UI?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+  } else if (typeof value === "boolean") {
+    return String(value);
+  } else if (typeof value === "object") {
+    return JSON.stringify(value);
+  } else {
+    return String(value);
+  }
+};
+
+export function getFeatureDatasourceDefaults(
```
ahdriel: question: Is this function exported? If yes, should we move it to a different helper file? Otherwise we should remove the `export`.

ahdriel: nit: Also should we rename it to `getDataSourceInitialFormValue`? or something to add `Form`.

I thought this was related to datasource configuration defaults at first.

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+      FeatureEvalDiagnosticsQueryResponseRows[number] & { id: string }
+    >,
+  ) => void;
+}) {
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const { datasources, getDatasourceById } = useDefinitions();
+  const settings = useOrgSettings();
+  const { apiCall } = useAuth();
```
ahdriel, line 98: question: Would we be able to use `useApi` (which uses `swr` under the hood)?

This means we would get back `data, error, isLoading` by default without us having to manually manage these states throughout the file.

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  const datasourceId = form.watch("datasourceId");
+  const datasource = datasourceId ? getDatasourceById(datasourceId) : null;
+
+  const datasourceHasFeatureUsageQuery =
+    datasource &&
+    datasource.settings.queries?.featureUsage &&
+    datasource.settings.queries.featureUsage.length > 0;
+
+  // Extract all unique keys from results (excluding 'id' which is added internally)
```
ahdriel: We should update this comment or remove the specification of the id column given that we are actually removing 3 columns.

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+    [results, columns],
+  );
+
+  const { items, pagination, SortableTH } = useSearch({
+    items: evalItems,
+    defaultSortField: "timestamp",
+    defaultSortDir: -1,
+    localStorageKey: "feature-diagnostics",
+    searchFields: ["timestamp", ...columns],
+    pageSize: 100,
```
ahdriel: question: By default our other usages are either 10 or 20. Do we want to go with 100 or should we go lower like 50?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+      return {
+        timestamp: format(getValidDate(row.timestamp), "PPpp"),
+        ...displayValues,
+      };
+    },
+    [results, columns],
+  );
+
+  const { items, pagination, SortableTH } = useSearch({
```
ahdriel: question: I see you are using `useSearch` here but I don't see a search input. Is there another reason why?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+            feature: feature.id,
+            datasourceId: form.watch("datasourceId"),
+          }),
+        },
+      );
+      if (results.rows) {
+        // add an id to each row based on the timestamp
+        const rowsWithId = results.rows.map((row) => ({
+          ...row,
+          id: row.timestamp ?? "",
```
ahdriel: question: Why do we need the timestamp as an id? There's potential for conflict here.

Would using the array index work?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+            <strong>Error:</strong> {error}
+          </Callout>
+        )}
+        {results && results.length === 0 && (
+          <Callout status="info">No feature evaluations found.</Callout>
+        )}
+        {/* If there's feature usage data, show a table with the data */}
+        {items.length > 0 && (
+          <>
+            <table className="table experiment-table gbtable">
```
ahdriel: question: Any reason why we are using the old tables instead of ui/Table?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+                        )
+                        .join(" ")}
+                    </SortableTH>
+                  ))}
+                </tr>
+              </thead>
+              <tbody>
+                {items.map((row) => (
+                  <tr key={row.id}>
+                    <td>{format(getValidDate(row.timestamp), "PPpp")}</td>
```
ahdriel: question: We are doing the same formatting in `evalFeatures`. Do we need it here?

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -0,0 +1,314 @@
...
+                    {columns.map((key) => (
+                      <td key={key}>
+                        {typeof row[key] === "string" ||
+                        typeof row[key] === "number" ||
+                        typeof row[key] === "boolean"
+                          ? String(row[key])
+                          : ""}
```
ahdriel: question: Same here. Can we use the formatted value from useAddComputedField?

packages/front-end/pages/features/[fid].tsx
```diff
@@ -43,6 +45,9 @@ export default function FeaturePage() {
   const [lastDisplayedVersion, setLastDisplayedVersion] = useState<
     number | null
   >(null);
+  const [diagnosticsResults, setDiagnosticsResults] = useState<Array<
+    FeatureEvalDiagnosticsQueryResponseRows[number] & { id: string }
+  > | null>(null);
```
Comment (ahdriel) [inline packages/front-end/pages/features/[fid].tsx:50]: question: Why is this state here? Can we move it inside `FeatureDiagnostics` instead?

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueryModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+        ? cloneDeep<FeatureUsageQuery>(featureUsageQuery)
+        : {
+            description: "",
+            id: uniqId("tbl_"),
+            name: "",
+            query: defaultQuery,
+          },
+  });
+
+  // User-entered values
```
ahdriel: can we remove this comment?

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueryModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const userEnteredQuery = form.watch("query");
+
+  const handleSubmit = form.handleSubmit(async (value) => {
+    await onSave(value);
```
ahdriel, line 49: issue: onSave is not async, so we cannot await it.

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueryModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+        close={onCancel}
+        size="lg"
+        header={modalTitle}
+        cta="Save"
+        ctaEnabled={saveEnabled}
+      >
+        <div className="my-2 ml-3 mr-3">
+          <div className="row">
+            <div className="col-12">
+              <Field label="Display Name" required {...form.register("name")} />
```
ahdriel: question: Do we need a display name for this query?

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueries.tsx
```diff
@@ -0,0 +1,273 @@
...
+    setEditingIndex(-1);
+    onCancel();
+  }, [onCancel]);
+
+  const featureUsageQueries = useMemo(
+    () => dataSource.settings?.queries?.featureUsage || [],
+    [dataSource.settings?.queries?.featureUsage],
+  );
+
+  console.log("featureUsageQueries", featureUsageQueries);
```
ahdriel: to remove

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueries.tsx
```diff
@@ -0,0 +1,273 @@
...
+      setUiMode(uiMode);
+    },
+    [],
+  );
+
+  const handleActionDeleteClicked = useCallback(
+    (idx: number) => async () => {
+      const copy = cloneDeep<DataSourceInterfaceWithParams>(dataSource);
+
+      // @ts-expect-error TS(2532) If you come across this, please fix it!: Object is possibly 'undefined'.
```
ahdriel: issue: Can we fix this error & the other `ts-expect-errors` on this file while we are here?

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueries.tsx
```diff
@@ -0,0 +1,273 @@
...
+                <Flex>
+                  <Heading as="h4" size="3" mb="1">
+                    {query.name}
+                  </Heading>
+                  {query.description && (
+                    <p className="ml-3 text-muted">{query.description}</p>
+                  )}
+                </Flex>
+
+                <Flex gap="4"></Flex>
```
ahdriel: question: Do we need this?

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueries.tsx
```diff
@@ -0,0 +1,273 @@
...
+                {query.error && (
```
ahdriel: question: I was not able to show this. Where is this being set / when would this appear?

packages/back-end/types/datasource.d.ts
ahdriel: I do think `FeatureUsage` makes sense here because it matches the SDK callback `onFeatureUsage`.

But agreed it can be potentially confusing with usage analytics

packages/back-end/types/datasource.d.ts
```diff
@@ -171,6 +171,14 @@ export interface ExposureQuery {
   error?: string;
 }
 
+export interface FeatureUsageQuery {
+  id: string;
+  name: string;
+  description?: string;
+  query: string;
+  error?: string;
```
ahdriel: suggestion: I don't think `error` belongs here. I guess technically a query can have a single error if malformed, but the same query can have transient errors depending on when/how they are executed.

I don't have a better solution and this follows the pattern for other queries we already have in place, so I consider this non-blocking.

packages/back-end/src/services/datasource.ts
```diff
@@ -236,6 +237,50 @@ export async function runUserExposureQuery(
...
+  try {
+    const { rows, statistics } =
+      await integration.runFeatureEvalDiagnosticsQuery(sql);
+    return {
+      rows,
+      statistics,
+      sql,
+    };
+  } catch (e) {
+    return {
```
ahdriel: suggestion: Should we also change the response code to be 400?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2600,6 +2602,30 @@ export default abstract class SqlIntegration
...
+    const featureEvalQuery = this.datasource.settings?.queries?.featureUsage
+      ? this.datasource.settings.queries.featureUsage[0].query
+      : "";
+
+    return format(
+      `-- Feature Evaluation Diagnostics Query
+      SELECT * FROM (
+        ${featureEvalQuery}
+        WHERE feature_key = '${featureKey}'
+      )
```
ahdriel, line 2623: issue: `MySQL` does not support unnamed derived tables.

So we need to either add `FROM ( ) AS __featureUsage` or my recommendation would be to use `WITH AS` which follows the pattern of our other queries.

```
WITH __featureUsage AS (
    ${featureEvalQuery}
    WHERE feature_key = '${featureKey}
)
SELECT * FROM __featureUsage
ORDER BY ...
```

ahdriel: Actually the above will also break if my Feature Usage query already has a WHERE clause. For example if I have a table that has multiple kinds of events and I need to add `event_type = 'feature_evaluation'` this would not work.

So the user provided SQL query should be on its own subquery and then we add the feature_key filter on the outer query.

packages/back-end/src/services/datasource.ts
ahdriel: Or better -- let the error bubble up to the controller and then handle it there.

You can wrap the error with the sql query if wanted.

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: I could go either way with this. I was trying to follow the model we have with experiments right now where you select a datasource first and then an EAQ (in this case we only allow one feature usage query per datasource which is why it's only one thing you need to select) We could just have users select feature usage queries and that would presumably shorten the list of options in the select if the org has several datasources for some reason

msamper: I think we'd probably still want to reference the datasource name in the options though. Not sure if folks will use names for the feature usage queries that will easily help them identify them from other data source usage queries

packages/front-end/pages/features/[fid].tsx
Comment (msamper) [reply packages/front-end/pages/features/[fid].tsx:50]: I originally had it within that component, but I pulled it out so that the results from the query wouldn't be lost when navigating between tabs since we're not storing query results in mongo

packages/front-end/components/Settings/EditDataSource/FeatureEvaluationQueries/FeatureEvaluationQueryModal.tsx
msamper: Since there's only one we can probably eliminate

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: I think right now we only use `useAPI` for `GET` requests and we use `useAuth` for `POST` requests. Not sure if there's a reason why. Seems like the recommended pattern is to have `useSWRMutation` be used for remote mutations so maybe we want a wrapper hook for that?

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: I think something like 50 is probably more reasonable

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: index should work! I think originally I thought it wouldn't because of sorting within `useSearch` but we set the id before passing results into that hook

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: This is reasonable! Agree that it's a bit odd since it pushes down the select

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: It seems like I was only setting the error in the `catch` block, but we actually set errors as `error` in the response when we catch them in the controller. Should be fixed now

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: No reason! I think i just followed a file using the old table. Updated to use ui/Table

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: Probably useful for debugging so I think so

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: I'm planning on doing a separate PR that adds a lightweight query runner for this to leverage the query collection so I'll add that there to avoid creating a special new queries modal for this.

packages/back-end/types/datasource.d.ts
msamper: Removed because it's not being used in this case

packages/front-end/components/Features/FeatureDiagnostics.tsx
msamper: Fixed padding and centered the button in the initial empty state and made it larger. Let me know if you think that looks better

ahdriel:
2 design comments:

- Can we make sure the sort indicator never shows up on a new line (like in the Value column here)?
- Can should we align the header and the columns content? All content is indented but the headers are not.

Screenshot 2025-12-18 at 1 03 16 PM

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2615,10 +2618,11 @@ export default abstract class SqlIntegration
 
     return format(
       `-- Feature Evaluation Diagnostics Query
-      SELECT * FROM (
+      WITH _featureEvalQuery AS (
```
ahdriel: nit: I don't know why, we do use 2 underscores as the prefix. Can we update it here too?

`__featureEvalQuery`

packages/front-end/components/Features/FeatureDiagnostics.tsx
ahdriel: Yea -- I think we do want that, so we can consolidate our `get` and `post` data loading/handling in the FE.

PR #5008, opened 2025-12-15
Use correct injection flow for Vue Options API

siarheipashkevich:
Right now the documentation has an invalid flow for injecting `gbKey` for vue options API, I tried to adjust this by official documentation: https://vuejs.org/guide/components/provide-inject#working-with-symbol-keys

### Features and Changes


royalfig:
Good catch! Thanks for the contribution.

PR #5011, opened 2025-12-15
Fix

ahdriel:
## Features and Changes

This PR fixes a critical variance calculation bug in the incremental refresh pipeline that was causing artificially low standard errors (~54% reduction compared to the regular pipeline), making experiment results appear much more confident than they actually are. Root Cause: The metric source table was storing pre-aggregated daily values (sums) but not storing sum of squares. When reaggregating, the code was computing `SUM(POWER(daily_sum, 2))` instead of `SUM(daily_sum_squares)`, which incorrectly calculates variance and loses information about within-period variance.

Solution:
1. Schema Change: Added `_sum_squares` columns to metric source tables for both values and denominators
2. INSERT Query: Compute `SUM(POWER(value, 2))` when aggregating daily data
3. Statistics Query: Reaggregate using `SUM(sum_squares)` instead of recomputing from sums
4. Statistics Calculation: Updated final variance calculation to use pre-computed sum_squares when available (incremental refresh) or compute on-the-fly (regular pipeline)

Additional Notes:

* The fix ensures the incremental refresh pipeline produces statistically identical results to the regular pipeline
* Variance is now correctly preserved through the aggregation hierarchy (event → daily → user → variation)
* Future consideration: May need similar fixes for covariate tables used in regression-adjusted metrics

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -2725,6 +2725,12 @@ export default abstract class SqlIntegration
       capValueCol: `${alias}_denominator_cap`,
       columnRef: metric.denominator,
     });
+
+    // For incremental refresh: reference pre-computed sum_squares columns
+    const capCoalesceSumSquares = `m${numeratorAlias}.${alias}_sum_squares`;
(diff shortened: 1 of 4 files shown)


lukesonnet:
This was an AI generated PR to check some functionality, and its fix is incorrect.

PR #5014, opened 2025-12-16
fix(EDF): do not use fallback from EDF in early stopping

lukesonnet:
### Features and Changes

We need to improve how our EDF deals with early stopping (with and without sequential testing).

This PR does this by making 2 key changes: 
* only using the "Strict Criteria" decision criteria to allow for early stopping without sequential testing. This is a rigorous criteria, and with a high degree of certainty needed to pass the early stopping without sequential testing criteria, we feel it is safe to apply universally.
* ensuring EDF users with sequential testing do not get their "fallback" status of "review" prematurely just because none of the rules match. For the fallback value to ever be used, you must have reached your desired power, regardless of whether you use sequential testing.

I also fixed a bug with clear losers and text.

Follow-ups: make it work better with sequential testing #5035 

### Testing

- [X] Test that we are not prematurely rendering Review statuses with sequential testing
- [X] Test that we are not prematurely rendering Review statuses with do no harm experiments with early stopping without sequential testing

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -3504,7 +3504,7 @@ export async function computeResultsStatus({
 
           if (resultsStatus.clearSignalResultsStatus === "won") {
             metricStatus.superStatSigStatus = "won";
-          } else if (resultsStatus.resultsStatus === "lost") {
+          } else if (resultsStatus.clearSignalResultsStatus === "lost") {
             metricStatus.superStatSigStatus = "lost";
           }
           if (!variationStatus.goalMetrics) {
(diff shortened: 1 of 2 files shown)


packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
```diff
@@ -288,10 +347,27 @@ export function getDecisionFrameworkStatus({
       };
     }
 
+    // Early stopping should only say ship if one variation is a clear winner
```
lukebrawleysmith, line 366: what if you have two variations that are way better than control?  Shouldn't we ship the better of those two?

lukesonnet: Yes, probably. But I want the recommendation for now to produce a single variation to ship. I'd like to support this case more clearly in the future.

PR #5015, opened 2025-12-16
SDK Payload Cache by clientKey instead of environment

jdorn:
### Features and Changes

Change how SDK payloads are cached.

Currently, we cache an intermediate state per-environment.  Then, when a request with a clientKey comes in, we use the settings for that SDK connection to filter/scrub the intermediate state and return the final payload.

This new method instead just caches the final payload per clientKey (i.e. SDK Connection) without an intermediate step.  This has a number of benefits:
- It removes the intermediate state, which is a big source of bugs and tech debt
- It makes payload responses faster since it gets rid of the runtime post-processing step
- It makes it much easier to introduce new SDK connection settings
- It allows us to have a more accurate `dateUpdated` field in the payload response

This is a big risky change, so it will be released in stages:
1. Populate the new cache, but do not use it when serving requests
2. Compare payloads between the caches across a wide variety of SDK connection settings to look for discrepancies
3. Opt a few SDK connections into using the new cache when serving requests in order to measure performance at scale
5. Enable the new cache for everyone and monitor performance
6. Remove the old cache and clean up tech debt


bryce-fitzsimons (approved):
Reviewed line by line and tested locally. Seems to work properly.

PR #5016, opened 2025-12-16
fix(incremental-refresh): Fix pre-computed dimensions in incremental refresh

lukesonnet:
### Features and Changes

Create a separate path for the precomputed dimensions in incremental refresh to ensure that you can only pass in dimensions with specified splices to the query runner.

Ensures that when computing statistics for incremental refresh we use on-demand dimensions correctly (unit, experiment that is not pre-computed, and date) and that we only try to pre-compute dimensions with pre-specified slices.
 
### Testing

- [x] Pre-computed dimensions are calculated and analyze compute when selected
- [x] On-demand unit dimensions compute on top of incremental refresh
- [x] On-demand experiment dimensions compute on top of incremental refresh
- [x] On-demand date cohort dimensions compute on top of incremental refresh
- [x] Changes in what dimensions are in the exposure query are reflected in the incremental refresh
  - [x] Incremental refresh properly ignores new experiment dimensions added to EAQ (since we don't have full history)
  - [x] Incremental refresh properly ignores non pre-computed experiment dimension deleted from EAQ (this experiment dimension is irrelevant)
  - [X] Incremental refresh properly drops a previously pre-computed experiment dimension when it is dropped from the EAQ (since we don't have it for new users)

### DIFF
--- a/packages/back-end/test/services/dimensions.test.ts
+++ b/packages/back-end/test/services/dimensions.test.ts
@@ -4,7 +4,7 @@ import {
   trimPrecomputedDimensionsToMaxCells,
 } from "back-end/src/services/dimensions";
 
-describe("getPredefinedDimensionSlicesByExperiment", () => {
+describe("trimPrecomputedDimensionsToMaxCells", () => {
   it("returns correct number of slices", () => {
     const dimensions: ExperimentDimensionWithSpecifiedSlices[] = [
       {


packages/back-end/src/services/dimensions.ts
```diff
@@ -0,0 +1,116 @@
...
+
+type EligibleDimensions = {
+  eligibleDimensions: ExperimentDimension[];
+  // Use for traffic analysis
+  eligibleDimensionsWithSlices: ExperimentDimensionWithSpecifiedSlices[];
+  // Use for pre-computing/post-stratification
+  eligibleDimensionsWithSlicesUnderMaxCells: ExperimentDimensionWithSpecifiedSlices[];
+};
+
+export function getExposureQueryEligibleDimensions({
```
lukesonnet, line 87: This is the new method that helps separate the exposure query dimensions into those that can be stored in the units table on refresh, those that can be used for pre-computing dimensions, and those that can be used in the traffic query.

packages/back-end/src/services/dimensions.ts
```diff
@@ -0,0 +1,116 @@
...
+  exposureQuery,
+  incrementalRefreshModel,
+  nVariations,
+}: {
+  exposureQuery: ExposureQuery;
+  incrementalRefreshModel: IncrementalRefreshInterface | null;
+  nVariations: number;
+}): EligibleDimensions {
+  const allDimensions = getExposureQueryDimensions({ exposureQuery });
+  const eligibleDimensions = incrementalRefreshModel
```
ahdriel, line 98: question: How does this interact with a Full Refresh for incremental? It should be `null` so we can use `allDimensions` right?

packages/back-end/src/services/experiments.ts
```diff
@@ -725,6 +728,9 @@ export function getSnapshotSettings({
   };
 }
 
+/**
+ * @deprecated Manual (e.g. manually inputting data) snapshots are no longer supported
```
ahdriel, line 733: question: Is this a new thing or just making it more official?

And should we add some metrics to see if we can remove this code maybe later on a numbered release?

packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -416,7 +410,7 @@ const startExperimentIncrementalRefreshQueries = async (
             unitsMaxTimestamp: maxTimestamp,
             experimentSettingsHash:
               getExperimentSettingsHashForIncrementalRefresh(snapshotSettings),
-            unitsDimensions: experimentDimensions.map((d) => d.id),
+            unitsDimensions: eligibleDimensions.map((d) => d.id),
```
ahdriel, line 412: question: `UnitDimension` is a type we have with `type: user` but here we are using `ExperimentDimension`.

Are they interchangeable to some extent or is this some debt we should clean up?

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -86,7 +86,6 @@ function validateUserFilter({
   }
 }
 
-// Ensure that that
```
ahdriel, line 93: nice 😂

packages/back-end/src/services/dimensions.ts
```diff
@@ -0,0 +1,116 @@
...
+      return [];
+    }) ?? []
+  );
+}
+
+export function trimPrecomputedDimensionsToMaxCells(
+  dimensions: ExperimentDimensionWithSpecifiedSlices[],
+  nVariations: number,
+): ExperimentDimensionWithSpecifiedSlices[] {
+  let totalLevels = countDimensionLevels(dimensions, nVariations);
```
ahdriel, line 53: nit: The previous function had a check to remove dimensions with no specified slices, but this one doesnt. Which means we will sum 1 for those dimensions I believe. Just checking if it is intentional.

```
// remove dimensions that have no slices
dimensions = dimensions.filter((d) => d.specifiedSlices.length > 0);
```

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -7547,14 +7552,39 @@ ${this.selectStarLimit("__topValues ORDER BY count DESC", limit)}
...
+      ? this.getDimensionCol(dateDimension)
+      : undefined;
 
-    const allDimensionCols = [...unitDimensionCols, ...nonUnitDimensionCols];
+    const nonUnitDimensionCols = [
+      ...experimentDimensionCols,
+      ...(dateDimensionCol ? [dateDimensionCol] : []),
+      ...precomputedDimensionCols,
+    ];
+    const allDimensionCols = [...nonUnitDimensionCols, ...unitDimensionCols];
```
ahdriel, line 7587: nit: This was `...unitDimensionCols, ...nonUnitDimensionCols` before. Just checking to ensure that the order change does not matter here.

packages/back-end/test/services/dimensions.test.ts
```diff
@@ -0,0 +1,52 @@
+import { ExperimentDimensionWithSpecifiedSlices } from "shared/types/integrations";
+import {
+  countDimensionLevels,
+  trimPrecomputedDimensionsToMaxCells,
+} from "back-end/src/services/dimensions";
+
+describe("getPredefinedDimensionSlicesByExperiment", () => {
```
ahdriel: nit: Should we update this name? `services/dimension` maybe?

packages/back-end/src/services/experiments.ts
lukesonnet: Making it more official. I think we can just remove it in this release, but I'll dig up the usage. I think it is 0.

packages/back-end/test/services/dimensions.test.ts
lukesonnet: Sorry, I'm not sure I understand this question.

packages/back-end/src/services/dimensions.ts
lukesonnet: It should be yes, but now I can see it isn't always being handled correctly. Thanks for this call out.

packages/back-end/src/services/dimensions.ts
lukesonnet: Yes, that is an unnecessary check. Technically no slices is a single "other" slice so I want to leave it in as the other logic doesn't do this filtering.

packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
lukesonnet: This was a bad choice in Incremental refresh to start calling ALL experiment AND unit dimensions "Unit dimension" because they refer to the breakdown of units in an experiment. And they could either then be "ExposureQueryDimension" (what we now call ExperimentDimension) and "StandaloneDimension" (what we now call a Unit Dimension). Then we can just be clearer in the docs and code to keep it distinct from "Metric Slice/Dimension".

lukesonnet: So incremental refresh has the "new" name for both experiment and unit dimensions, "unitsDimensions".

This code is correct, the timing and naming is bad.

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: It shouldn't.

packages/back-end/test/services/dimensions.test.ts
ahdriel: I meant that this function name (getPredefinedDimensionSlicesByExperiment) does not exist anymore, so just thinking on updating the describe message to more accurately describe this file which is now services/dimensions.

PR #5028, opened 2025-12-20
Post stratification uncertainty quantilfication bugfix

lukebrawleysmith:
### Features and Changes

Post stratification currently has a bug that causes overestimation of uncertainty.  This PR:

1. fixes the bug
2. removes some unnecessary code
3. restructures post-strat code in places to match the tech-spec

### Testing
1. all unit tests pass with minor changes to the outputs after the uncertainty quantification bug fix.  
2.  compared results offline and found that post-strat now is on par or slightly reduces uncertainty compared to no post-start
3. ran a simple simulation that illustrated that post-strat has similar performance to regression with HC1 covariance, as expected

### DIFF
--- a/packages/stats/gbstats/utils.py
+++ b/packages/stats/gbstats/utils.py
@@ -119,16 +119,6 @@ def multinomial_covariance(nu: np.ndarray) -> np.ndarray:
     return np.diag(nu) - np.outer(nu, nu)
 
 
-def multinomial_third_sample_moment(n, nu):
-    """
-    Given n multinomial samples with proportion nu,
-    calculates E(X**3 / n**3).
-    """
-    num = (n - 1) * (n - 2) * nu**3 + 3 * (n - 1) * nu**2 + nu
-    den = n**2
-    return num / den
-
-
 @dataclass
 class MatrixInversionResult:
     """
--- a/packages/stats/tests/frequentist/test_post_strat.py
+++ b/packages/stats/tests/frequentist/test_post_strat.py
@@ -126,7 +126,7 @@ def compute_dof(stats: List[Tuple[TestStatistic, TestStatistic]]) -> float:
     )
 
 
-@skip("Skipping test because of compute time")
+@skip("Skipping covariance of multinomial weighted means test because it is slow")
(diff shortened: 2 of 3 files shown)


packages/stats/tests/frequentist/test_post_strat.py
```diff
@@ -125,6 +126,64 @@ def compute_dof(stats: List[Tuple[TestStatistic, TestStatistic]]) -> float:
...
+                alpha_array[sim, :, cell] = np.mean(this_alpha, axis=0)
+            combined_mean[sim, :] = alpha_array[sim, :, :] @ this_nu_hat
+        self.covariance_theoretical = (
+            PostStratificationSummary.covariance_of_multinomial_weighted_means(
+                n_total, alpha_mean, alpha_cov, nu
+            )
+        )
+        self.covariance_empirical = np.cov(combined_mean.T, ddof=1)
+
+    def test_covariance_of_multinomial_weighted_means(self):
```
lukesonnet, line 175: I'm not sure I understand this test.

packages/stats/gbstats/utils.py
```diff
@@ -119,73 +119,14 @@ def multinomial_covariance(nu: np.ndarray) -> np.ndarray:
     return np.diag(nu) - np.outer(nu, nu)
 
 
-def third_moments_matrix_vectorized(n_total: int, nu: np.ndarray) -> np.ndarray:
+def multinomial_third_sample_moment(n, nu):
```
lukesonnet: This doesn't seem to be used anywhere.

lukebrawleysmith: removed

packages/stats/tests/frequentist/test_post_strat.py
lukebrawleysmith: Suppose we have 5 strata cells.  We have an estimate of the effect in each cell.  The point estimate of the absolute effect is a weighted estimate of the cell effects, where the weights are multinomial distributed. 

This function:
1. simulates cell effects
2. generates multinomial weights
3. constructs the multinomial weighted means

It then compares the empirical covariance of 3. to the theoretical covariance, which is used in post-strat uncertainty estimation.

PR #5029, opened 2025-12-20
feature(sql): Chunk more metrics per table to improve query efficiency

lukesonnet:
### Features and Changes

Now that we are chunking sql results, it is less problematic to return more columns per query, and therefore have more metrics per query, thereby improving the efficiencies created by the fact metric query optimization (where if you had 21+ metrics coming from one table, we still had to scan the table more than once).

This PR increases the max allowable number of fact metrics per query to a dynamic value depending on the types of metrics requested. For non-ratio and ratio metrics, that's about 50-100 metrics per chunk (with ratio metrics being much larger).

We still run quantile metrics always in their own query due to the additional query demands, and we can only fit around 17 of those in a single query.

These numbers will go down as we add the uncapped versions of metrics for metric drilldown, but it should only reduce us to around 25-50 metrics per chunk (still potentially halving query costs in some cases), and we can also be smart then to only count the additional columns if a metric is capped.

### Testing

- [x] Added sensible unit tests
- [x] Test with 20 quantile metrics
- [x] Test with 120 ratio metrics
- [x] Test with legacy metrics
- [x] Test with bandits
- [x] test with capping
- [x] test with capping ratio metrics
- [x] test with and without cuped
- [x] test with overrides for CUPED and for capping
- [x] Go over all docs to ensure 1,000 limit is reasonable, consider increasing for most common/confident engines (BQ, Snowflake, Databricks, Clickhouse)

### DIFF
--- a/packages/back-end/src/services/experimentQueries/experimentQueries.ts
+++ b/packages/back-end/src/services/experimentQueries/experimentQueries.ts
@@ -31,11 +31,11 @@ import {
 export function getNonQuantileFloatColumns({
   metric,
   regressionAdjusted,
-  bandit,
+  isBandit,
 }: {
   metric: FactMetricInterface;
   regressionAdjusted: boolean;
-  bandit: boolean;
+  isBandit: boolean;
 }): string[] {
   const baseCols = (() => {
     switch (metric.metricType) {
@@ -96,7 +96,7 @@ export function getNonQuantileFloatColumns({
 
   const cols = [...baseCols, ...cupedCols, ...percentileCappingCols];
 
-  if (bandit) {
+  if (isBandit) {
     cols.push(...BANDIT_CUPED_FLOAT_COLS);
   }
 
@@ -106,11 +106,11 @@ export function getNonQuantileFloatColumns({
 export function maxColumnsNeededForMetric({
   metric,
   regressionAdjusted,
-  bandit,
+  isBandit,
 }: {
   metric: FactMetricInterface;
   regressionAdjusted: boolean;
-  bandit: boolean;
+  isBandit: boolean;
 }) {
   // id column
   const boilerplateCols = 1;
@@ -118,7 +118,7 @@ export function maxColumnsNeededForMetric({
   const floatCols = getNonQuantileFloatColumns({
     metric,
     regressionAdjusted,
-    bandit,
+    isBandit,
   });
   switch (metric.metricType) {
     case "mean":
--- a/packages/back-end/test/services/experimentQueries.test.ts
+++ b/packages/back-end/test/services/experimentQueries.test.ts
@@ -21,7 +21,7 @@ describe("experimentQueries", () => {
           maxColumnsNeededForMetric({
             metric,
             regressionAdjusted: false,
-            bandit: false,
+            isBandit: false,
           }),
         ).toBe(baseExpectedCols);
 
@@ -29,7 +29,7 @@ describe("experimentQueries", () => {
           maxColumnsNeededForMetric({
             metric,
             regressionAdjusted: true,
-            bandit: false,
+            isBandit: false,
           }),
         ).toBe(baseExpectedCols + 3);
 
@@ -37,7 +37,7 @@ describe("experimentQueries", () => {
           maxColumnsNeededForMetric({
             metric,
             regressionAdjusted: true,
-            bandit: true,
+            isBandit: true,
           }),
         ).toBe(baseExpectedCols + 4);
       });
@@ -53,7 +53,7 @@ describe("experimentQueries", () => {
           maxColumnsNeededForMetric({
             metric,
             regressionAdjusted: false,
-            bandit: false,
+            isBandit: false,
           }),
         ).toBe(4);
       });
@@ -79,7 +79,7 @@ describe("experimentQueries", () => {
             numerator: { factTableId: "ft_1" },
           }),
           regressionAdjusted: false,
-          bandit: false,
+          isBandit: false,
         });
 
         // Create 100 quantile metrics
@@ -171,7 +171,7 @@ describe("experimentQueries", () => {
                 maxColumnsNeededForMetric({
                   metric: m,
                   regressionAdjusted: false,
-                  bandit: false,
+                  isBandit: false,
                 }),
               0,
             );
@@ -273,7 +273,7 @@ describe("experimentQueries", () => {
             numerator: { factTableId: "ft_1" },
           }),
           regressionAdjusted: false,
-          bandit: false,
+          isBandit: false,
         });
         // Set max columns so only 1 quantile metric fits per chunk
         const maxColumnsPerQuery = baseColumnsNeeded + quantileColsNeeded + 1;


packages/back-end/src/services/experimentQueries/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+  "main_denominator_sum_product",
+  "denominator_pre_sum",
+  "denominator_pre_sum_squares",
+  "main_post_denominator_pre_sum_product",
+  "main_pre_denominator_post_sum_product",
+  "main_pre_denominator_pre_sum_product",
+  "denominator_post_denominator_pre_sum_product",
+];
+
+export const ALL_NON_QUANTILE_METRIC_FLOAT_COLS = RATIO_METRIC_FLOAT_COLS;
```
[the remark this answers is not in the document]
lukesonnet: Nice catch!

I added "quantile" to the `getQuantileBoundsFromQueryResponse` method and just build it there so that all of the quantile logic stays separated as much as possible.

I'll also be sure to test quantile metrics, as outlined in my PR summary.

packages/back-end/src/services/experimentQueries/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+export const N_STAR_VALUES = [
+  100, 200, 400, 800, 1600, 3200, 6400, 12800, 25600, 51200, 102400, 204800,
+  409600, 819200, 1638400, 3276800, 6553600, 13107200, 26214400, 52428800,
+];
+
+export const BASE_METRIC_FLOAT_COLS = [
+  "main_sum",
+  "main_sum_squares",
+  "main_cap_value",
+  "denominator_cap_value",
```
jdorn: Should this be in the ratio constants instead?

packages/back-end/src/services/experimentQueries/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+  "denominator_sum",
+  "denominator_sum_squares",
+  "main_denominator_sum_product",
+  "denominator_pre_sum",
+  "denominator_pre_sum_squares",
+  "main_post_denominator_pre_sum_product",
+  "main_pre_denominator_post_sum_product",
+  "main_pre_denominator_pre_sum_product",
+  "denominator_post_denominator_pre_sum_product",
+];
```
jdorn, line 32: Wondering if we should break these up more to separate out CUPED.  Would let us double the number of metrics per query when CUPED is disabled.

Might also make the code simpler to have these as exclusive arrays.  So for example, the ratio columns would not also contain the base columns.  Will make the groups more composable and let us separate out more things like capping if we want to.

```ts
const cols = [...BASE_COLS];
if (cuped) cols.push(...CUPED_COLS);
if (ratio) cols.push(...DENOMINATOR_COLS);
if (cuped && ratio) cols.push(...DENOMINATOR_CUPED_COLS);

return cols;
```

packages/shared/src/sql.ts
```diff
@@ -5,6 +5,8 @@ import { FormatError } from "../types/error";
 
 export const SQL_ROW_LIMIT = 1000;
 
+export const SQL_COL_LIMIT = 1000;
```
jdorn: Is this being used?

lukesonnet: oops, no

packages/back-end/src/services/experimentQueries/constants.ts
lukesonnet: Yes, I'll move these around as per your other suggestion.

packages/back-end/src/services/experimentQueries/constants.ts
lukesonnet: Thanks for the suggestion, I added `getNonQuantileFloatColumns`. I separated everything out to maximize the usage we will get per table and used switch statements to try and limit future changes from missing a trick here.

I'm not stoked about how it looks... but I think it's alright and allows us to be more efficient. I think there's a potential for bugs if we add more columns for an existing metric type, but I'm not sure the best way to handle that without much more metadata/structure per metric type.

Still want to test more next week.

packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -149,7 +155,21 @@ export function getIncrementalRefreshMetricSources(
...
+        applyMetricOverrides(metric, snapshotSettings);
+        return {
+          metric,
+          regressionAdjusted:
+            isRegressionAdjusted(metric) &&
+            snapshotSettings.regressionAdjustmentEnabled,
+        };
+      }),
+      maxColumnsPerQuery: integration.getSourceProperties().maxColumns,
+      bandit: !!snapshotSettings.banditSettings,
```
ahdriel: nit: Should we prefix this with `is`?

packages/back-end/src/services/experimentQueries/experimentQueries.ts
```diff
@@ -0,0 +1,305 @@
...
+    });
+    groupArrays.push(...chunks);
+  });
+
+  // Add unused fact metrics as singles to the group array
+  factMetrics.forEach((m) => {
+    if (!groupArrays.some((group) => group.some((g) => g.id === m.id))) {
+      groupArrays.push([m]);
+    }
+  });
```
ahdriel, line 302: This is a nested iteration (looking over all items in all the groups for each factMetric).

I think it would be fine given the low number of metrics/groups, but we can also do:
```
const groupedMetricIds = new Set(
	groupArrays.flatMap((group) => group.map((g) => g.id))
);
factMetrics.forEach((m) => {
	if (!groupedMetricIds.has(m.id)) { 
		groupArrays.push([m]);
	}
});
```

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1263,6 +1265,8 @@ export default abstract class SqlIntegration
       [key: string]: number;
     } = {};
     if (row[`${prefix}quantile`] !== undefined) {
+      quantileData[`${prefix}quantile`] =
```
ahdriel, line 1268: question: where is this being used? / why we need this?

ahdriel (approved):
Changes LGTM! And thanks for the detailed tests 🚀

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: I don't know that we actually need this column in general in the returned values, neither do we need the cap value in capped metrics, but they can sometimes help with debugging in screenshots and stuff like that. Probably could remove it.

I added it here in this PR because it isn't in FLOAT_COLS that gets processed as part of every query.

packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
lukesonnet: you mean `isBandit`? Sure. I think that's clearer.

packages/back-end/src/services/experimentQueries/experimentQueries.ts
lukesonnet: Yeah I was just copying the original logic, I'll use your logic, it's cleaner.

PR #5030, opened 2025-12-22
Tooltip Syncing on Dashboards

mknowlton89:
### Features and Changes

This PR updates the Dashboards to introduce a new DashboardChartsContext wrapper. This new wrapper allows us to add charts to a group. This group can then be used by ECharts so that if two charts have the same x-Axis, and a user hovers over a data point on one graph which causes the tooltip to render, we'll also render the tooltip on the other chart.

E.G
Screenshot 2025-12-22 at 9 40 23 AM

Screenshot 2025-12-22 at 9 41 01 AM

To power this, we've updated the props for `EChartsReact` to use the `onChartReady` prop that registers the chart in the context when mounted.

There are ultimately two approaches we can take with this. I've gone the simple route for now, which simply adds all charts on a dashboard to the `group`, and offloads the matching logic to Echarts.

By default, ECharts will first match by x-value, which works really well if the x-axis are the same (and the date ranges are the same).

If the x-values don't align perfectly, ECharts falls back to matching by value. This is especially handy if two charts have different ranges, but they both contain the same `x-axis`. E.G. They have different date ranges, but they both include `2022-02-02`. 

The other approach is to create multiple chart `groups` where each group has identical x-axis (both values, ranges, etc). This is much more complex, and does mean that the fuzzy matching above won't work, but it will also prevent any issues where tooltips showing in instances where the user doesn't think it should. E.g. Showing the `2022-02-02` date on two graphs that aren't actually related.

### Testing

- [x] Ensure it works as expected with multiple graphs all built off the same sql that have the same x-axis
- [x] Ensure it works as expected for multiple metric blocks that are all built off the same metric
- [x] Ensure that it works as expected for sql blocks and metric blocks that both have similar x-axis (dates)
- [x] Ensure the context is wiped when the user navigates to another page. 

### Screenshots

Screenshot from the 3rd test case (Note, there is an unrelated issue with timezones that we'll likely want to address)
Screenshot 2025-12-22 at 9 40 17 AM

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -3504,7 +3504,7 @@ export async function computeResultsStatus({
 
           if (resultsStatus.clearSignalResultsStatus === "won") {
             metricStatus.superStatSigStatus = "won";
-          } else if (resultsStatus.resultsStatus === "lost") {
+          } else if (resultsStatus.clearSignalResultsStatus === "lost") {
             metricStatus.superStatSigStatus = "lost";
           }
           if (!variationStatus.goalMetrics) {
--- a/packages/front-end/components/Experiment/TabbedPage/ExperimentDecisionExplanation.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/ExperimentDecisionExplanation.tsx
@@ -53,10 +53,20 @@ export default function ExperimentDecisionExplanation({
     return null;
   }
 
-  // fix to get deciding rule
   const getConditionText = (
     condition: DecisionCriteriaInterface["rules"][0]["conditions"][0],
   ) => {
+    const { quantity, connection } = (() => {
+      switch (condition.match) {
+        case "all":
+          return { quantity: "All", connection: "metrics are" };
+        case "any":
+          return { quantity: "At least one", connection: "metric is" };
+        case "none":
+          return { quantity: "No", connection: "metrics are" };
+      }
+    })();
+
     const metricType = (() => {
       switch (condition.metrics) {
         case "goals":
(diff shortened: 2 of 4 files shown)


packages/front-end/enterprise/components/Dashboards/DashboardChartsContext.tsx
```diff
@@ -0,0 +1,117 @@
...
+  useContext,
+  useEffect,
+  useRef,
+  useCallback,
+} from "react";
+import * as echarts from "echarts/core";
+
+interface DashboardChartsContextValue {
+  registerChart: (id: string, instance: echarts.ECharts) => void;
+  unregisterChart: (id: string) => void;
```
Kevin-Chant: Should there be some on-mount useEffect for the chart components that calls `unregisterChart`? Currently this isn't in use anywhere

mknowlton89: Ah, good catch.

I did initially have a useEffect that was returning a cleanup function that called unregister, but ultimately removed it as in my testing, if a user navigates away, the DashboardChartsProvider unmounts and React garbage collects everything automatically.

mknowlton89: For now I've removed the `unregister` function and references in the context file.

For historical context, removal of the useEffect cleanup functions that previously called `unregister` was done [in this commit](https://github.com/growthbook/growthbook/pull/5030/changes/d73e15e7c12cec7caf2a6f1fba8958c2c1e15cbe#diff-10c7b036afe6955bdabf79f9ba5f284b0564a64ac04977ff959fed1135f25752).

PR #5031, opened 2025-12-22
SavedGroupModel to use BaseClass

oelshaikh:
### Features and Changes
# Saved Groups Model Refactor

### Summary

This PR migrates the `SavedGroupModel` from a direct Mongoose-based model to one that extends the standard `BaseModel` class. This aligns the model with the codebase's modern patterns, consolidating type definitions and leveraging inherited CRUD, permission, and audit logging capabilities.

### Changes

- **Refactored `SavedGroupModel`**: The model now extends `BaseClass`, allowing it to inherit standard methods for get, update, and delete operations.
- **Context-Based Access**:
  - Replaced standalone function calls (e.g., `getSavedGroupById`) with model access via `req.context.models.savedGroups`.
  - Updated `getSavedGroupMap` to accept a `context` object instead of an `organization`, allowing it to access both the model and the organization from the context.
- **Controller Cleanup**: Removed manual auditing and redundant logic from controllers, as these are now handled automatically by the `BaseClass`.
- **Validation & Types**:
  - Updated validators to include the full `savedGroup` schema required by the base class.
  - Consolidated type definitions by moving `SavedGroupInterface` from `shared/types/groups.d.ts` to `shared/types/saved-group.d.ts`.
- **Test Updates**: Updated backend unit tests to reflect the new model structure and context usage.
- **Get All without values**: Integrated the changes from [this PR](https://github.com/growthbook/growthbook/pull/5064) into the new model.
  - Added ability to use projections in `BaseModel`'s `_find` function. 

NOTE: The majority of the file changes in this PR are just import updates.

### Testing

- [x] Ensure you are able to create, update, and delete saved groups in the UI 
- [x] Ensure you are able to create, update, and delete saved groups via REST API
- [x] Ensure that existing saved groups are available in the dropdown when creating an experiment 
- [x] Visit the saved groups page, look at the network tab in dev tools to ensure values are being omitted in API response.


packages/back-end/src/api/saved-groups/listSavedGroups.ts
```diff
@@ -1,9 +1,6 @@
 import { ListSavedGroupsResponse } from "shared/types/openapi";
 import { listSavedGroupsValidator } from "shared/validators";
-import {
-  getAllSavedGroups,
-  toSavedGroupApiInterface,
-} from "back-end/src/models/SavedGroupModel";
+import { toSavedGroupApiInterface } from "back-end/src/models/SavedGroupModel";
```
gazzdingo: i think that we can remove this now

packages/back-end/src/api/saved-groups/listSavedGroups.ts
gazzdingo: this to

packages/back-end/src/api/saved-groups/postSavedGroup.ts
gazzdingo, line 84: should we check with jeremy if we want to update this to use context user

packages/back-end/test/api/features-public.test.ts
```diff
@@ -87,10 +87,6 @@ jest.mock("back-end/src/models/ApiKeyModel", () => ({
   lookupOrganizationByApiKey: jest.fn(),
 }));
 
-jest.mock("back-end/src/services/savedGroups", () => ({
-  getSavedGroupMap: jest.fn(),
-}));
```
gazzdingo, line 92: might need to revert this

gazzdingo (changes requested):
can you check permissions

packages/back-end/src/api/saved-groups/listSavedGroups.ts
oelshaikh: Moved this to a member function to follow the pattern of other models like `FactMetricModel` and `ProjectModel`. 

It seems like the interface here is slightly different than the schema defined in the validator so it makes sense to keep this to keep the API response consistent.

packages/back-end/test/api/features-public.test.ts
oelshaikh: Seems like this was incorrectly mocked, `getSavedGroupMap` doesn't exist in `services/savedGroups`, it comes from `services/features`. Removing this mock did not cause any tests to fail.

packages/back-end/src/api/saved-groups/postSavedGroup.ts
oelshaikh: Seems like the BaseModel [already does this](https://github.com/growthbook/growthbook/blob/609716cf80eb70dd51aeb44ada7eb8b386c0e028/packages/back-end/src/models/BaseModel.ts#L454-L457)

packages/back-end/src/api/saved-groups/updateSavedGroup.ts
```diff
@@ -94,13 +88,12 @@ export const updateSavedGroup = createApiRequestHandler(
   // If there are no changes, return early
   if (Object.keys(fieldsToUpdate).length === 0) {
     return {
-      savedGroup: toSavedGroupApiInterface(savedGroup),
+      savedGroup: req.context.models.savedGroups.toApiInterface(savedGroup),
     };
   }
 
-  const updatedSavedGroup = await updateSavedGroupById(
+  const updatedSavedGroup = await req.context.models.savedGroups.updateById(
```
jdorn: If you already have the saved group object, you can use `update` instead of `updateById`. This will avoid an extra Mongo lookup.

packages/back-end/src/routers/project/project.controller.ts
```diff
@@ -281,7 +280,7 @@ export const deleteProject = async (
 
   // Clean up saved groups
   try {
-    await removeProjectFromSavedGroups(id, org.id);
+    await context.models.savedGroups.removeProject(id);
```
jdorn: Nit: the name of this method makes the id argument a little ambiguous. Looking at this code, I might think you are supposed to pass in the saved group Id and not the project id. Maybe something like `removeProjectIdFromAllGroups`would be clearer?

packages/back-end/src/routers/saved-group/saved-group.controller.ts
```diff
@@ -261,23 +238,11 @@ export const postSavedGroupAddItems = async (
...
-    entity: {
-      object: "savedGroup",
-      id: updatedSavedGroup.id,
-      name: savedGroup.groupName,
-    },
-    details: auditDetailsUpdate(savedGroup, updatedSavedGroup),
-  });
-
-  savedGroupUpdated(context, savedGroup);
+  savedGroupUpdated(context, updatedSavedGroup);
```
jdorn: Should we move this to the afterUpdate hook in the model so we don't have to remember to call it everywhere we update a saved group?

packages/back-end/src/api/saved-groups/updateSavedGroup.ts
oelshaikh: done, also done for delete

packages/back-end/src/routers/saved-group/saved-group.controller.ts
oelshaikh: done

packages/back-end/src/routers/project/project.controller.ts
oelshaikh: done

packages/back-end/src/models/SavedGroupModel.ts
```diff
@@ -1,172 +1,120 @@
...
+          },
+        },
+      });
+    }
+
+    return group;
+  }
+
+  protected migrate(legacyDoc: LegacySavedGroupInterface): SavedGroupInterface {
+    return SavedGroupModel.migrateSavedGroup(legacyDoc);
```
Kevin-Chant, line 76: Out of scope for this PR, but this makes me think we should move `migrate` for `BaseModel` to be static and just alias it in an instance method via `this.constructor.migrate()`

packages/back-end/src/models/SavedGroupModel.ts
```diff
@@ -1,172 +1,120 @@
...
+      });
+    }
+  }
+
+  public async removeProjectIdFromAllGroups(projectId: string) {
+    await this._dangerousGetCollection().updateMany(
+      { organization: this.context.org.id, projects: projectId },
+      // @ts-expect-error - Mongodb driver types are strict about $pull
+      { $pull: { projects: projectId } },
+    );
```
Kevin-Chant, line 105: Looks like the reason you were getting a `TSError` was the generic typing couldn't inference `SavedGroupInterface` so it couldn't match `{projects: string}` to `Partial<any>`. Giving it an explicit type removes the need for a `@ts-expect-error`.

Also (out of scope), I think this means we need an `updateMany` helper in `BaseModel`. We shouldn't have to use `_dangerousGetCollection` in instance methods on the subclasses.
```suggestion
    const pullOperation: UpdateFilter<SavedGroupInterface> = {
      projects: projectId,
    };
    await this._dangerousGetCollection().updateMany(
      { organization: this.context.org.id, projects: projectId },
      { $pull: pullOperation },
    );
```

packages/shared/src/validators/saved-group.ts
```diff
@@ -1,9 +1,30 @@
...
+export const savedGroupTypeValidator = z.enum(["condition", "list"]);
+
+export const savedGroupValidator = z
+  .object({
+    id: z.string(),
+    organization: z.string(),
+    groupName: z.string(),
+    owner: z.string(),
+    type: savedGroupTypeValidator,
+    source: z.string().optional(),
```
Kevin-Chant: Where did `source` come from? It's not part of the current `SavedGroupInterface`

packages/back-end/src/models/SavedGroupModel.ts
oelshaikh: Done, this is much cleaner thanks!

packages/shared/src/validators/saved-group.ts
oelshaikh: It comes from the current model definition: https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/models/SavedGroupModel.ts#L31

it seems like this might be a legacy field that isn't always set but I wanted to make sure the DB model was consistent.

Kevin-Chant: Hmm so it is. I checked every instance of `.source` and nothing's related to saved groups. I also have some very old saved groups in my local db and they don't have `source` set either. I'd actually lean towards removing this field (despite that being a change to the model) so as not to pollute the interface. @jdorn would you mind weighing in here? Maybe you know what `source` used to be for saved groups

jdorn: `source` is an old field that we use inside of `migrateSavedGroup`.  So it should be part of the `LegacySavedGroupInterface`, but does not need to be part of the saved group validator here.

PR #5034, opened 2025-12-23
Allow user to configure chart scaling for Sql Explorer line and scatter charts

mknowlton89:
### Features and Changes

This PR introduces a new `Display Settings` panel in the Data Viz Configurator for the Sql Explorer and allows users to specify on a per-chart basis (limited to only line and scatter charts). 

This PR also introduces a migration so that all charts by default are anchored to 0, including Metric Explorer visualizations. Given that we're going to deprecate the current MetricExplorer block visualizations soon, I'm not planning to add the display settings controls to the Metric Explorer block, but instead focusing on getting the new block builder built that will allow a user to create visualizations from metrics, but do so in a way similar to how we support sql blocks today.

### Testing
New dashboard
- [x] Create a new dashboard with a bar chart and a line chart and ensure the dashboard saves correctly
- [x] Then, for the line chart, disable the anchor to 0 and ensure the graph updates and the chart saves accordingly
- [x] Then, change the line chart to a bar chart and ensure that the `displaySettings` object is then removed completely upon being saved to the db.
- [x] Then, switch the bar chart back to a line chart and ensure the `anchorToZero` is enabled by default, and is saved to the db correctly.

Existing dashboard
- [x] Ensure you can open an existing dashboard that was created before the migration and make changes to it successfully
- [x] When opening for the first time, ensure that line and scatter charts ran the migration and have the `anchorToZero` enabled by default
- [x] Ensure you can change this setting and persist it on the db correctly
- [x] Ensure that you can then change the graph type to one that doesn't support the anchorToZero property and ensure that you're able to save it successfully without any errors and when looking at the config in the db, ensure there is no displaySettings property set

General
- [x] Ensure the `Display Settings` panel is only rendered for line and scatter charts

### Screenshots
Screenshot 2025-12-29 at 10 51 08 AM


packages/back-end/src/models/SavedQueryDataModel.ts
```diff
@@ -24,6 +24,50 @@ const BaseClass = MakeModelClass({
...
+        const needsMigration =
+          !displaySettings ||
+          typeof displaySettings.anchorYAxisToZero !== "boolean";
+
+        if (needsMigration) {
+          return {
+            ...config,
+            displaySettings: {
+              ...(displaySettings || {}),
+              anchorYAxisToZero: true,
```
lukesonnet, line 56: I thought we changed the default behavior? I think it's hard to know ahead of time what the default should be, but I think for bar charts it should definitely be to include 0 by default, but for line charts it could go either way, but I think you could easily argue for zooming in by default.

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
```diff
@@ -163,18 +177,47 @@ export default function DataVizConfigPanel({
...
+                  );
                   return;
                 }
                 // Update chart type and normalize dimensions if needed
                 const updatedConfig = normalizeDimensionsForChartType({
                   ...dataVizConfig,
                   chartType: v as DataVizConfig["chartType"],
                 } as Partial<DataVizConfig>);
+                // If the chart type changes to line/scatter & we don't have displaySettings, set the default
+                if (["line", "scatter"].includes(v)) {
```
lukesonnet: Should this be a const somewhere for a mapping of types to settings, then it could also be used by the migration and in DisplaySettingsPanel?

lukesonnet: I kind of like the pattern of using exhaustive switches for this kind of stuff for fact metric types, but I'm just curious if you think it's a good idea:

```
function supportsAnchorYAxisToZero(chartType) {
  // exhaustive switch with no default
}
```

So now any new chart type must confront this question.

Or it's some kind of mandatory property that has to be defined to extend chart type? But IDK how to do that in typescript.

Or maybe it can be inferred from your validators where this field exists?

packages/front-end/components/DataViz/DisplaySettingsPanel.tsx
```diff
@@ -0,0 +1,90 @@
...
+                </Flex>
+              </Text>
+            </div>
+          }
+          transitionTime={100}
+        >
+          <Box p="4" height="fit-content">
+            <Flex direction="column" gap="4">
+              <Checkbox
+                label="Anchor y-axis to zero"
```
lukesonnet: What is the behavior if the values can be negative? If the values span below and above 0, does this setting do nothing (as I would expect)?

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
```diff
@@ -771,7 +775,7 @@ export function DataVisualizationDisplay({
               : "category",
       },
       yAxis: {
-        scale: true,
+        scale: !anchorYAxisToZero,
```
lukesonnet, line 778: Eventually we may want to have even more control over the range and use `anchorYAxisToZero` as an input to a range calculator or something.

lukesonnet (approved):

packages/front-end/components/DataViz/DisplaySettingsPanel.tsx
mknowlton89: It ultimately depends on the data - the scale of the min/max values might change slightly, but 0 would still be included in that range.

packages/front-end/components/DataViz/SqlExplorerDataVisualization.tsx
mknowlton89: For sure. If `min` and `max` are provided in the chart config, ECharts will ignore the `scale` property and use the `min` and `max` values instead.

I imagine the UI would allow the user to simply check a box for anchoring the y-axis to 0 OR they could specify the min/max range.

packages/back-end/src/models/SavedQueryDataModel.ts
mknowlton89: We did. When we initially released PA Dashboards, we always anchored to 0, then we got feedback that we shouldn't so we automatically disabled the scale. 

Jeremy had strong feelings that in most cases we should anchor to 0, so this PR sets that back to the default, but now gives users the ability to disable that if they want.

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
mknowlton89: Agreed on these points - I'll explore and let you know what I end up with. Good suggestion.

packages/back-end/src/models/SavedQueryDataModel.ts
lukesonnet: Yeah, I tend to agree that reasons to NOT anchor to zero are like "nice to haves to see the result more clearly" but there are often good principled reasons to force anchoring to zero.

packages/front-end/components/DataViz/DataVizConfigPanel.tsx
mknowlton89: Added a shared `chartTypeSupportsAnchorToYAxis` so we're using the same logic everywhere.

I also added a slightly broader `chartTypeSupportsDisplaySettings` helper. This helper will help us ultimately determine if we need to render the `DisplaySettingsPanel` at all.

Currently, the logic between these two helpers is identical, but as we add more support for the DisplaySettings we can expand the `chartTypeSupportsDisplaySettings` helper.

Thanks for the suggestion for this - I had thought about it when writing the initial code but didn't take the time to implement and this has really improved things.

PR #5038, opened 2025-12-23
Improve Saved Group linking, add [sgid] page for Condition Groups

bryce-fitzsimons:
- Add new [sgid] landing page for Condition Groups
- Make all condition displays and condition editors link to this new page
- Fix references widget to include nested saved groups
- Improve overall UI

image

image

image

image

### DIFF
--- a/packages/front-end/components/SavedGroups/ConditionGroups.tsx
+++ b/packages/front-end/components/SavedGroups/ConditionGroups.tsx
@@ -240,7 +240,9 @@ export default function ConditionGroups({ groups, mutate }: Props) {
                                 )}
                                 canDelete={
                                   isEmpty(referencingFeaturesByGroup[s.id]) &&
-                                  isEmpty(referencingExperimentsByGroup[s.id]) &&
+                                  isEmpty(
+                                    referencingExperimentsByGroup[s.id],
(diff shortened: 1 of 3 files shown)


packages/front-end/pages/saved-groups/[sgid].tsx
```diff
@@ -105,26 +110,110 @@ export default function EditSavedGroupPage() {
...
+      try {
+        const parsed = JSON.parse(condition);
+        const savedGroupIds = extractSavedGroupIds(parsed);
+        if (savedGroupIds.has(targetId)) {
+          referencingGroups.push(sg);
+        }
+      } catch (e) {
+        console.error("Error parsing condition", e);
+      }
+    });
```
Comment (jdorn) [inline packages/front-end/pages/saved-groups/[sgid].tsx:114]: I'm ok with this logic, but we could simplify and just look to see if the condition string contains this quoted saved group id. Tiny chance of false positives, but much simpler logic to read and maintain.

jdorn (approved):

packages/front-end/pages/saved-groups/[sgid].tsx
Comment (bryce-fitzsimons) [reply packages/front-end/pages/saved-groups/[sgid].tsx:114]: good point, simplified the check

PR #5039, opened 2025-12-24
Migrate SdkWebhookModel to BaseModel

Kevin-Chant:
### Features and Changes

Changes the WebhookModel to inherit from BaseModel and updates usage in controllers and jobs

### Testing

1. On `main`, navigate to an SDK connection and click "Add Webhook" to create one with the old model
2. Switch to this branch and create another webhook with the same config. Use mongo to inspect that the docs are identical (other than generated data)
3. Edit an existing webhook and then delete it to check the other interactions also work

### Screenshots

image

### DIFF
--- a/packages/back-end/src/models/WebhookModel.ts
+++ b/packages/back-end/src/models/WebhookModel.ts
@@ -57,6 +57,8 @@ export class SdkWebhookModel extends BaseClass {
         newDoc.payloadFormat = "standard-no-payload";
       }
     }
+    if (!castDoc.dateCreated && castDoc.created)
+      newDoc.dateCreated = castDoc.created;
     return newDoc;
   }
 
@@ -110,7 +112,7 @@ export class SdkWebhookModel extends BaseClass {
   }
 
   public static async dangerousFindSdkWebhookByIdAcrossOrgs(id: string) {
-    const doc = getCollection(COLLECTION_NAME).findOne({
+    const doc = await getCollection(COLLECTION_NAME).findOne({
       id,
     });
     return doc ? this.migrate(removeMongooseFields(doc)) : null;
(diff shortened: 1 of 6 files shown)


packages/back-end/src/models/WebhookModel.ts
```diff
@@ -1,233 +1,139 @@
...
+    error: string,
+  ) {
+    await this.update(webhook, {
+      error,
+      lastSuccess: error ? undefined : new Date(),
+    });
+  }
+
+  public static async dangerousFindSdkWebhookByIdAcrossOrgs(id: string) {
+    const doc = getCollection(COLLECTION_NAME).findOne({
```
tzjames: this should have an `await`

packages/shared/src/validators/webhooks.ts
```diff
@@ -11,23 +11,41 @@ export const payloadFormatValidator = z.enum([
...
+  project: z.string().optional(),
+  environment: z.string().optional(),
+  featuresOnly: z.boolean().optional(),
+  signingKey: z.string(),
+  lastSuccess: z.date().nullable(),
+  error: z.string(),
+  useSdkMode: z.boolean(),
+  sdks: z.array(z.string()),
+  /** @deprecated */
+  created: z.date(),
```
tzjames: since new ones shouldn't have this shouldn't this be `.optional()`?

packages/back-end/src/models/WebhookModel.ts
```diff
@@ -1,233 +1,139 @@
...
-
-export async function countSdkWebhooksByOrg(organization: string) {
-  return await WebhookModel.countDocuments({
-    organization,
-    useSdkMode: true,
-  }).exec();
+    });
+  }
+
+  public getCreateProps(sdkConnectionId: string) {
```
tzjames: could rename this to `getDefaultCreateProps`?

packages/back-end/src/models/WebhookModel.ts
```diff
@@ -1,233 +1,139 @@
...
+    return {
+      environment: "",
+      project: "",
+      error: "",
+      created: new Date(),
```
tzjames: if we make created optional, then maybe get rid of this and make sure the front-end uses dateCreated.

PR #5047, opened 2025-12-31
Adds hasPremiumFeature check to a few enterprise feature models

mknowlton89:
### Features and Changes

There were a few enterprise features where we weren't verifying the organization's plan on the backend (only protecting this via the UI). This PR adds `hasPremiumFeature` checks to the `canCreate` method on the Holdout, Custom Field, and ExperimentTemplate models.

In the process, I did notice that these models were using a slightly different pattern for the `canCreate`, `canUpdate`, and `canDelete` methods. Now, we're throwing actual permission errors rather than generic error messages if the user doesn't have permission.

From a tech-debt perspective, I think we should go through and update all models that extend the `BaseModel` so the `canCreate` returns actual permission errors if the user doesn't have permission, rather than a generic error, but we can do this in a separate PR.

### Dependencies

None

### Testing

- [x] Remove UI restrictions for custom fields and try to create a new custom field as a free org and ensure the backend returns a 400 error with a descriptive message.
- [x] Remove UI restrictions for holdouts  and try to create a new holdout as a free org and ensure the backend returns a 400 error with a descriptive message.
- [x] Remove UI restrictions for experiment templates and try to create a new template as a free org and ensure the backend returns a 400 error with a descriptive message.

### Screenshots
Screenshot 2025-12-31 at 7 23 06 AM
Screenshot 2025-12-31 at 7 23 15 AM


packages/back-end/src/models/CustomFieldModel.ts
```diff
@@ -26,15 +26,33 @@ export class CustomFieldModel extends BaseClass {
...
-    return this.context.permissions.canManageCustomFields();
+    if (!this.context.hasPremiumFeature("custom-metadata")) {
+      throw new Error(
+        "Your organization's plan does not include the custom fields feature.",
+      );
+    }
+
+    if (!this.context.permissions.canManageCustomFields()) {
+      throw this.context.permissions.throwPermissionError();
+    }
```
jdorn: I don't think this is right. The can* methods should return a Boolean and the base class should do the throwing. If the status code being returned is wrong, we should fix that in the base class.

jdorn: Also, we should probably add a new method to the base class `hasPremiumFeature` instead that returns true by default (but can be overriden). We can check that automatically during create and update operations and throw if needed.

mknowlton89: I see.

Is the idea of the `can*` methods that those only check permissions? If so, that makes the update to the BaseModel class pretty simple - we can update that to return a permission error, instead of the generic error.

If not, then it'll take a bit more work as the `can*` methods could return false for a number of reasons. 

This is tangental, but if the `can*` methods are only permission checks, I'm not sure the naming has enough context.

I can create the new `hasPremiuimFeature` method to the base class and check that before doing the create & update operations.

Kevin-Chant (approved):
~Might as well make the new errors use `PermissionError` while we're here, otherwise~ LGTM
Edit: Nvm I just double checked the error text and it's less applicable to premium feature checks. We can add a helper for that separately
```
You do not have permission to perform this action
```

PR #5050, opened 2026-01-02
Add configurable user parameter for Presto connection

vovapyc:
### Features and Changes

Added configurable user parameter for Presto connection. Previously hardcoded as "growthbook", now users can set their own value in the UI. Defaults to "growthbook" if not specified.

- Closes **https://github.com/growthbook/growthbook/issues/5051**

### Dependencies

None

### Testing

1. Go to Data Sources → Add/Edit Presto connection
2. Verify new "User" field appears after Port field
3. Verify default value is "growthbook"
4. Save with custom value and verify it persists

### Screenshots

image

### DIFF
--- a/packages/front-end/components/Settings/PrestoForm.tsx
+++ b/packages/front-end/components/Settings/PrestoForm.tsx
@@ -85,7 +85,7 @@ const PrestoForm: FC<{
         />
       </div>
       <div className="form-group col-md-12">
-        <label>User</label>
+        <label>Query User</label>
         <input
           type="text"
           className="form-control"


packages/front-end/components/Settings/PrestoForm.tsx
```diff
@@ -84,6 +84,19 @@ const PrestoForm: FC<{
           onChange={onParamChange}
         />
       </div>
+      <div className="form-group col-md-12">
+        <label>User</label>
```
jdorn: We also have a username param when using basic auth. To make the distinction clear, I think we should set the label for this field to `Query User`

Auz (approved):

PR #5057, opened 2026-01-05
Re-enable post-strat

lukesonnet:
Re-enable post-strat once improvements to uncertainty quantification is landed #5028


lukebrawleysmith (approved):
lgtm.  I pushed one change so the Results tab now says "CUPEDps" instead of CUPED. and made the same change in a few other places.

PR #5060, opened 2026-01-06
[performance] FeatureWithoutValues interface to avoid large JSON in front-end

jdorn:
### Features and Changes

In many parts of the front-end, we fetch all feature flags across all projects into memory.  This is used to power the feature search, populate select boxes, see where saved groups are being used, etc..  In a large org with thousands of features, this can become problematic, especially if there are JSON flags with large values.  In the worst case, this can cause page loads in the app to take 10+ seconds.

This PR makes several changes related to this to improve front-end performance.

1. Do not fetch the full feature list unless it's really needed.  For example, the homepage was getting the full list just to check if the length was > 0.  This is now using a dedicated API endpoint that just returns a single boolean.
2. Do not return JSON flag values unless really needed.  For example, to check which features reference a saved group, we just need to inspect targeting conditions, not flag values.
3. Remove some unnecessary data being returned from API calls in the back-end. For example, along with the features we were also looking up all linked experiments and returning them, even though the front-end was not using that data.

TODO:
* [ ] A version of `evaluatePrerequisiteState` that does not require JSON values
* [ ] Switch GetStartedPage's `NeedingAttention` component to only fetch the specific feature/experiment ids required
* [ ] On feature page, return dependent features, prerequisite features, and staleness from back-end instead of calculating on front-end from full feature list
* [ ] Lots of testing to make sure there are no regressions
* [ ] Benchmark performance improvements


packages/front-end/pages/features/index.tsx
```diff
@@ -352,8 +358,14 @@ export default function FeaturesPage() {
                         }) ? (
                           <button
                             className="dropdown-item"
-                            onClick={() => {
-                              setFeatureToDuplicate(feature);
+                            onClick={async () => {
+                              // We need the full feature (with values) to duplicate
```
tzjames: I think this comment is taken care of?

packages/back-end/src/app.ts
```diff
@@ -761,7 +761,16 @@ app.use("/demo-datasource-project", demoDatasourceProjectRouter);
...
 // Features
 app.get("/feature", featuresController.getFeatures);
-app.get("/feature/:id", featuresController.getFeatureById);
+app.get(
+  "/features-without-values",
+  featuresController.getFeaturesWithoutValues,
+);
+// These routes are poorly named but kept for backwards compatibility
+// We should swap them later:
+//   - feature/:id -> feature/:id/details
```
tzjames, line 770: You can start the swap here, right?  Make `app.get("/feature/:id/details", featuresController.getFeatureAndDetailsById);` and update all callers on the front-end. But also keeping `app.get("/feature/:id", featuresController.getFeatureAndDetailsById);` for backwards compatibility.  Then after this lands we can get rid of the data route and have `app.get("/feature/:id", featuresController.getFeatureById);`

PR #5069, opened 2026-01-08
fix(eslint): Prevent .eslintrc unintentional rule override

ahdriel:
### Features and Changes

When adding new eslint rules my assumption was that the rules were additive, and that is true to an extent. If we use the same rule twice, then it gets overriden.

This happened for `no-restricted-imports`.

- Fix .eslintrc to ensure all rules apply to the desired files
- Fix all broken usages (mostly relative paths in the import path)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -81,6 +81,7 @@
     "no-restricted-imports": "off",
     "react/react-in-jsx-scope": "off",
     "react/prop-types": "off",
+    "react/no-unknown-property": ["error", { "ignore": ["jsx", "global"] }],
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
     "import/order": [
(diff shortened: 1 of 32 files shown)


packages/front-end/components/DataViz/DataVizFilter.tsx
```diff
@@ -3,7 +3,7 @@ import { Box, Flex, Text, TextField } from "@radix-ui/themes";
 import { PiTrash } from "react-icons/pi";
 import { Select, SelectItem } from "@/ui/Select";
 import Button from "@/ui/Button";
-import MultiSelectField from "../Forms/MultiSelectField";
+import MultiSelectField from "@/components/Forms/MultiSelectField";
 import { ColumnFilterOption } from "./DataVizFilterPanel";
```
ahdriel, line 7: question: Should we allow local imports like this or not?

Kevin-Chant: I think `./` imports are fine because they don't have the risk of `../` and can save a lot of space when you're 5 folders deep

PR #5071, opened 2026-01-08
better handling for query errors / failed status

bryce-fitzsimons:
We weren't showing the view queries button anywhere when the very first snapshot failed. We also were showing the wrong error message in place of the table.

image

### DIFF
--- a/packages/front-end/components/Report/LegacyReportPage.tsx
+++ b/packages/front-end/components/Report/LegacyReportPage.tsx
@@ -413,8 +413,8 @@ export default function LegacyReportPage({
                       }
                       notebookUrl={`/report/${report.id}/notebook`}
                       notebookFilename={report.title}
-                      queries={report.queries}
-                      queryError={report.error}
+                      legacyQueries={report.queries}
+                      legacyQueryError={report.error}
                       results={report.results?.dimensions}
                       variations={variations}
                       metrics={getAllMetricIdsFromExperiment(


packages/front-end/components/Experiment/Results.tsx
```diff
@@ -261,7 +267,18 @@ const Results: FC<{
         </div>
       )}
 
+      {status === "failed" ? (
```
lukesonnet: Doesn't this mean it always shows up even if it has data? I'm not sure we need to show this banner if we have the "error" in the topright in the case where there is data but a future update fails.

packages/front-end/styles/global.scss
```diff
@@ -2823,13 +2823,9 @@ html {
...
   .gb-multi-select__menu-portal {
     position: absolute !important;
-    top: -8px;
-    left: 290px;
-    @media (max-width: 520px) {
-      top: auto;
-      bottom: 0;
-      left: 0;
-    }
+    top: auto;
```
bryce-fitzsimons, line 2826: unrelated fix for select fields within portals/popovers (e.g. Experiment Result Filter component)

packages/front-end/components/Experiment/Results.tsx
bryce-fitzsimons: fixed, thanks

packages/front-end/components/Experiment/ResultMoreMenu.tsx
```diff
@@ -357,7 +359,9 @@ export default function ResultMoreMenu({
         variant="soft"
       >
         <DropdownMenuGroup>
-          {queryStrings.length > 0 && (
+          {queryStrings.length > 0 ||
+          status === "failed" ||
```
lukesonnet: I don't think this is needed any more. I think the original check that there are some queries is the appropriate one, and the logic of which queries to show is governed by `queryStrings`.

packages/front-end/components/Experiment/ResultMoreMenu.tsx
```diff
@@ -452,7 +456,7 @@ export default function ResultMoreMenu({
           close={() => setQueriesModalOpen(false)}
           queries={queryStrings}
           savedQueries={[]}
-          error={queryError}
+          error={snapshot ? snapshot?.error : latest?.error}
```
lukesonnet: nit: Ideally this ternary and the one that builds queryStrings could be done in one place since they go hand-in-hand.

PR #5075, opened 2026-01-09
Stop disabling dashboard update button when only sql explorer blocks exist

Kevin-Chant:
### Features and Changes

The "Update" button on dashboards was being incorrectly disabled when the only source of queries was sql explorer blocks. Sql explorer queries are already respected for the "View Queries" count and async queries modal, so we just needed to change the logic that disables the update button

### Testing

Create and save a new dashboard with only a markdown block, the update button should be disabled.
After adding a saved query, it should be enabled, and clicking it does refresh the query (though there's no visual indication if the underlying data is still the same)

### Screenshots

image

(After refreshing, the snapshot provider refreshes the saved query map and `dateLastRan` is updated)
image


mknowlton89 (approved):
Tested and works as expected!

PR #5078, opened 2026-01-09
BigQuery SQL formatting bug

mknowlton89:
### Features and Changes
Users were experiencing data corruption when saving and reopening SQL Explorer queries that contained BigQuery `NUMERIC` or `BIGNUMERIC` columns. These values would display correctly on first run (e.g., `"3066553.75"`), but after saving and reopening the query, they would appear as JSON objects:{  "s": "1",  "e": "6",   "c": [5, 4, 9, 6, 3, 0, 9, 5, 3]}
This occurred with aggregate functions like SUM() and division operations on numeric columns.

#### Root Cause
The @google-cloud/bigquery library returns NUMERIC and BIGNUMERIC values as Big.js objects with an internal structure of {s: sign, e: exponent, c: coefficient array}. When these objects were serialized to JSON (for API responses) and saved to MongoDB, they would be stored in their raw internal format rather than as human-readable strings.
While BigQuery integration already handled special date types (BigQueryDate, BigQueryTimestamp), it wasn't converting numeric Big.js objects to strings.

#### Solution
Fixed the issue at the source by extending the existing type conversion logic in BigQuery.ts:
Added Big.js detection in the runQuery method (right after the existing date handling)

#### Why This Approach
✅ Prevents the issue rather than cleaning it up after the fact
✅ Consistent with how dates are already handled in the same code
✅ Simpler - one line (String(value)) vs. 60+ lines of manual decimal reconstruction
✅ More performant - converts once at query time, not on every save
✅ Single source of truth - fix happens in one place

### Migration Path
Users with existing affected saved queries can simply re-run the query to fix the data. The conversion now happens automatically at the source, so all new query results will be stored correctly.

### Testing
Tested with BigQuery queries containing:
CAST(x AS NUMERIC)
CAST(x AS BIGNUMERIC)
SUM() aggregations on numeric columns
Division operations returning numeric results
All values now correctly save and display as strings (e.g., "3066553.75") instead of Big.js internal objects.


packages/back-end/src/integrations/BigQuery.ts
```diff
@@ -125,6 +125,18 @@ export default class BigQuery extends SqlIntegration {
...
+          typeof value === "object" &&
+          "s" in value &&
+          "e" in value &&
+          "c" in value &&
+          Array.isArray((value as Record<string, unknown>).c)
+        ) {
+          // BigQuery NUMERIC/BIGNUMERIC values are returned as Big.js objects
+          // with internal structure {s: sign, e: exponent, c: coefficient array}
+          // Convert to string to preserve precision and avoid serialization issues
+          row[key] = String(value);
```
ahdriel, line 142: Oooh -- I am curious if we will have unexpected behavior here where we expect the type to be number but we get the returned value as a string (which is intended now).

mknowlton89: That's a good callout. 

From what I can tell we shouldn't - in SqlIntegration.ts we're calling `parseInt` and `parseFloat` already, so we are already expecting strings here.

mknowlton89: E.G. https://github.com/growthbook/growthbook/blob/9c440e88a16ff9a137284d599a9d03d0d209b24e/packages/back-end/src/integrations/SqlIntegration.ts#L1236-L1242

PR #5080, opened 2026-01-11
Front-end performance improvements v1

jdorn:
### Changes

* Don't include linkedExperiments in `useFeatures()` hook data (was not being used)
* On homepage, only fetch 10 recent changes (instead of 100) and omit details (huge JSON blobs)
* On homepage, use new endpoint to check if features/experiments exist.  Was previously downloading entire list and doing `length > 0` on the front-end.
* Don't fetch saved queries in left nav (not being used)
* Resize and compress images on homepage
* Don't search default values on feature list (JSON flags could have huge values)

### Future Changes

* When fetching all features/experiments on the front-end, only return minimal fields from the back-end (id, name, etc.) unless the full objects are absolutely necessary

### DIFF
--- a/packages/back-end/src/models/FeatureModel.ts
+++ b/packages/back-end/src/models/FeatureModel.ts
@@ -210,9 +210,6 @@ export async function getAllFeatures(
   );
 }
 
-const _undefinedTypeGuard = (x: string[] | undefined): x is string[] =>
-  typeof x !== "undefined";
-
 export async function hasArchivedFeatures(
   context: ReqContext | ApiReqContext,
   project?: string,


packages/back-end/src/models/AuditModel.ts
```diff
@@ -61,19 +61,17 @@ export async function insertAudit(
...
  * @param organization
- * @param options
  */
-export async function findAuditByUserIdAndOrganization(
+export async function findRecentAuditByUserIdAndOrganization(
   userId: string,
   organization: string,
-  options?: QueryOptions,
-): Promise<AuditInterface[]> {
+): Promise<Omit<AuditInterface[], "details">> {
```
tzjames: The type here is incorrect.
```suggestion
): Promise<Omit<AuditInterface, "details">[]> {
```

PR #5084, opened 2026-01-12
Re-enable post-stratification

lukesonnet:
Re-enables post-stratification.

Post-stratification can be turned on for users if:
* In Organization Settings > Experiment Settings; they ensure "Pre-compute Dimension Breakdowns" is selected.
* They ensure post-strat is on for their experiment by either setting the experiment to accept the org default and turning post-strat on by default in Organization Settings > Stats Engine Settings; OR by overriding the setting to On at the experiment-level in the Analysis Settings.

Achieves this by:
* Adds a resolver that takes into account the org setting that enables post-strat and compares the experiment setting vs. the org default to determine whether post-strat is on or off.
* Adds the experiment on/off override to the analysis form 
* Adds the status to the experiment analysis header
* Adds additional tracking within the stats engine response object within the snapshot for whether or not post-strat was actually applied (this isn't used in the UI now, but can help in debugging in the future). This means that if post-strat is on and pre-computed dimensions are on, but there are no dimensions available to pre-compute, we still say "post-strat: Enabled" even if there was no dimensions that were used and thus post-strat did nothing.

We don't always show the post-stratification override and status to the analysis header. If pre-computed dimensions are off:
* Exp override not shown in Analysis Form as post-strat is hidden, and nothing shown in header
* Under the hood post-strat is set to off

Docs update to follow.

## Testing
* No setting appears and flag is false when pre-computed dimensions is off
* Is Outdated flag appears when expected, and resolver computes the right results given the combination of settings

## Screenshots

new field in analysis form
Screenshot 2026-01-12 at 11 24 20 AM
Screenshot 2026-01-12 at 12 18 00 PM

is outdated shows when the setting would change across runs, also you can see the new Post-Stratification: Disabled setting in the experiment settings bar.
Screenshot 2026-01-12 at 11 24 39 AM

When org level pre-computed dimensions are off, we hide these new UI elements and turn post-strat off under the hood
Screenshot 2026-01-12 at 11 31 54 AM


packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -730,12 +730,17 @@ const startExperimentIncrementalRefreshQueries = async (
     });
     queries.push(maxTimestampMetricsSourceQuery);
 
+    const dimensionsForPrecomputation = org.settings
```
lukesonnet, line 733: Unrelated bugfix; incremental refresh was doing pre-computed dimensions even if the setting was off.

packages/back-end/src/services/reports.ts
```diff
@@ -461,6 +461,9 @@ export async function createReportSnapshot({
   });
   const statsEngine =
     report.experimentAnalysisSettings.statsEngine || settings.statsEngine.value;
+  const postStratificationEnabled =
+    report.experimentAnalysisSettings.postStratificationEnabled ||
+    settings.postStratificationEnabled.value;
```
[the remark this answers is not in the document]
lukesonnet: Fixed as suggested. There isn't a report setting to turn it off now, for what it's worth, since we're deprecating support of these reports eventually. We can add it as a setting if it is requested. I think metric drilldowns will mean this isn't high priority.

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -250,9 +250,9 @@ export function getDefaultExperimentAnalysisSettingsForSafeRollout(
       (regressionAdjustmentEnabled !== undefined
         ? regressionAdjustmentEnabled
         : (organization.settings?.regressionAdjustmentEnabled ?? false)),
-    postStratificationEnabled: false,
-    //hasPostStratificationFeature &&
-    //!(organization.settings?.postStratificationDisabled ?? false),
+    postStratificationEnabled:
+      hasPostStratificationFeature &&
+      !organization.settings?.postStratificationDisabled,
```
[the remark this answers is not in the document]
lukesonnet: Fixed. Post-strat is always on for safe rollouts if we can pre-compute and if it is on at the org level.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2955,14 +2957,16 @@ export async function postSnapshot(
...
+    const postStratificationEnabled = settings.postStratificationEnabled.value;
 
-    const analysisSettings = getDefaultExperimentAnalysisSettings(
+    const analysisSettings = getDefaultExperimentAnalysisSettings({
       statsEngine,
       experiment,
-      org,
-      false,
+      organization: org,
+      regressionAdjustmentEnabled: false,
```
ahdriel: question (not related to this pr): Why do we disable regressionAdjustment for manual snapshots but we respect postStratification?

packages/back-end/src/services/experiments.ts
```diff
@@ -1404,6 +1412,8 @@ export async function createSnapshot({
       statsEngine: defaultAnalysisSettings.statsEngine,
       regressionAdjustmentEnabled:
         defaultAnalysisSettings.regressionAdjusted ?? false,
+      postStratificationEnabled:
+        defaultAnalysisSettings.postStratificationEnabled ?? true,
```
ahdriel: question: Should we use `DEFAULT_POST_STRATIFICATION_ENABLED` here instead of hardcoding `true`?

packages/shared/src/settings/resolvers/postStratificationEnabledResolver.ts
```diff
@@ -0,0 +1,58 @@
+import { SettingsResolver } from "../types";
+
+/**
+ * Resolver for post-stratification enabled setting.
+ *
+ * The org setting is `postStratificationDisabled` (inverted - when true, post-strat is OFF),
+ * but we expose it as `postStratificationEnabled` (when true, post-strat is ON).
```
ahdriel: I think this can be a source of confusion in the future as the 2 settings are named very similarly.

What do you think about having the org setting be `postStratificationEnabled` as well? And if it is `undefined` or `on`, we treat it as active.
In the UI we can still show it as `Disable Post Stratification`, but the underlying fields would match.

packages/shared/types/stats.d.ts
```diff
@@ -37,6 +37,8 @@ interface BaseVariationResponse {
   ci?: [number | null, number | null];
   errorMessage?: string;
   power?: MetricPowerResponseFromStatsEngine;
+  // Added later to gbstats model, leave as undefined
+  realizedSettings?: RealizedSettings;
```
ahdriel, line 41: nice! This is a good improvement.

Should it belong in the BaseVariationResponse or should it live a level higher? I don't know if we define what we support in gbstats on a per analysis basis or a per variation basis.

packages/back-end/src/controllers/experiments.ts
lukesonnet: If you mean Manual as in the one created by `createManualSnapshot` then I'm just going to delete that whole thing today lol.

If you mean something else, then please let me know.

packages/back-end/src/services/experiments.ts
lukesonnet: Yes, thanks.

packages/shared/types/stats.d.ts
lukesonnet: Some things may only happen per variation. I had it much higher earlier, but IMO having richer data we can use to get higher level summaries later is better.

In this case, in one variation you could have post-strat apply and in another you might not and it might fall back to the standard estimator, if there is not enough variance across cells in the case of the other variation.

packages/shared/src/settings/resolvers/postStratificationEnabledResolver.ts
lukesonnet: >What do you think about having the org setting be postStratificationEnabled as well? And if it is undefined or on, we treat it as active.

Yeah I just thought that was weird to treat undefined as on. WDYT?

packages/back-end/src/services/safeRolloutSnapshots.ts
```diff
@@ -250,9 +250,10 @@ export function getDefaultExperimentAnalysisSettingsForSafeRollout(
       (regressionAdjustmentEnabled !== undefined
         ? regressionAdjustmentEnabled
         : (organization.settings?.regressionAdjustmentEnabled ?? false)),
-    postStratificationEnabled: false,
-    //hasPostStratificationFeature &&
-    //!(organization.settings?.postStratificationDisabled ?? false),
+    postStratificationEnabled:
+      hasPostStratificationFeature &&
+      !organization.settings?.disablePrecomputedDimensions &&
+      !organization.settings?.postStratificationDisabled,
```
lukebrawleysmith: this logic looks right to me.  nit: I feel like it would be easier to read if the booleans were instead `enablePreComputedDimensions` and `postStratificationEnabled`.

lukesonnet (commented):
> suggestion: In the Settings UI, it looks like Variance Reduction section is under the engine settings (either Frequentist or Bayesian) but in reality it is shared between both.

Anything we can do to make this distinction clearer?

Let me consider as a follow-up please.

packages/back-end/src/controllers/experiments.ts
ahdriel: Yeah -- this code is inside

```
// Manual snapshot
if (!experiment.datasource) {
```

packages/front-end/components/Experiment/AnalysisForm.tsx
lukesonnet, line 597: @ahdriel , stats engine just submits "" for accept default, instead of undefined or null, which gets saved in the DB but it isn't expected since the type specifies it must be one of the Stats Engine selects.

packages/back-end/src/controllers/experiments.ts
lukesonnet: Oh sorry, I missed that. Yeah this code doesn't matter. These decisions don't matter at all, neither will ever apply. Just passing forward defaults I would think.

packages/shared/src/settings/resolvers/postStratificationEnabledResolver.ts
ahdriel: I think it is fine as it falls back to `DEFAULT_POST...` which is set to `true`. It does mean that we need to be aware of the default value/behavior though.

lukesonnet: Yeah I should make this change.  I already changed the UI to read "use by default" so it isn't even "disable" in the UI anymore.

lukesonnet: Ok, added a migration and used the constant fallback when needed.

packages/back-end/src/services/safeRolloutSnapshots.ts
lukesonnet: Yeah, it's still a little messy. I don't want to touch also the disable pre-computed setting for now, but I can at least rectify the post-strat one which makes logic in a bunch of places simpler.

lukebrawleysmith (approved):
My suggestions have been addressed.

packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -266,6 +277,7 @@ const AnalysisForm: FC<{
       submit={form.handleSubmit(async (value) => {
         const { dateStarted, dateEnded, skipPartialData, ...values } = value;
 
+        console.log(values);
```
ahdriel, line 280: Need to remove

lukesonnet: That's old, I removed it.

lukesonnet: github honestly being trash

PR #5088, opened 2026-01-13
chore(experiments): remove manual snapshot data form

lukesonnet:
### Features and Changes

Largely AI generated PR to remove the ability to create or update "manual" snapshots, which are ones where you manually type in all metric data and we do some basic statistics for you and render the result. 5 orgs have used this in the last 6 months, and those usages all seem exploratory and potentially unexpected.

Keeping this around is a bunch of dead code that references shared methods that we have to update all the time, so we should just remove it.

This experience was pretty broken anyways. I think it was totally broken with fact metrics, we don't let you pick "manual" anymore, so it really shouldnt' be something people do, and we should be tracking errors when there is no datasource instead of exposing manual data entry.

### Testing

- [x] test old manual snapshots at least load correctly

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
@@ -629,45 +629,45 @@ export default function AnalysisSettingsSummary({
             </Flex>
 
             {ds &&
-              permissionsUtil.canRunExperimentQueries(ds) &&
-              allMetrics.length > 0 && (
-                <RefreshResultsButton
-                  entityType={
-                    experiment.type === "holdout" ? "holdout" : "experiment"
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -628,7 +628,8 @@ export default function AnalysisSettingsSummary({
               ) : null}
             </Flex>
 
-            {(!ds || permissionsUtil.canRunExperimentQueries(ds)) &&
+            {ds &&
+              permissionsUtil.canRunExperimentQueries(ds) &&
```
bryce-fitzsimons: technically fine as-is, but I prefer using a ternary `?` whenever some of the inputs are not boolean.

bryce-fitzsimons (approved):
Looks good, have not tested the migrations. Small nit about rendering logic.

PR #5092, opened 2026-01-14
Ls/only use original queries for cache

lukesonnet:
### Features and Changes

When creating a new query document that used the cache for results, we effectively we re-setting the TTL on the cache. If you clicked the update button within the TTL and the query was unchanged, you never would get new data.

This changes that by only pulling in "original" (e.g. queries that didnt use the cache) documents when trying to use the cache, effectively preserving the TTL.

I think this is fine for still retrieving the old results because even though we now skip the intermediate documents that hit the cache, we previously had to ensure that `cachedQueryUsed` passed forward from the original query. This should, I think, negate the need for that feature of `cachedQueryUsed`, but it's fine to keep it around to mark whether or not a query doc was original or pulled from the cache.

### Testing

TODO

### DIFF
--- a/packages/back-end/src/models/QueryModel.ts
+++ b/packages/back-end/src/models/QueryModel.ts
@@ -153,6 +153,7 @@ export async function getRecentQuery(
   organization: string,
   datasource: string,
   query: string,
+  onlyOriginalQueries: boolean = true,
 ) {
   // Only re-use queries that were run recently
   const earliestDate = new Date();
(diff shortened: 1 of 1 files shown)


packages/back-end/src/models/QueryModel.ts
```diff
@@ -153,6 +153,7 @@ export async function getRecentQuery(
   organization: string,
   datasource: string,
   query: string,
+  onlyOriginalQueries: boolean = true,
```
ahdriel, line 156: nit: Do we even want to have this as an option?

ahdriel (approved):
Good catch!

PR #5093, opened 2026-01-14
 SDK Connections modal refresh 

oelshaikh:
### Features and Changes

- Updated the Create and Edit SDK Connections modals to be cleaner / easier to use. Figma file: https://www.figma.com/design/OunBM2DKWs0PmYsoisuP36/New-SDK-Connection-Modal-v2.0?node-id=1-2&p=f&m=dev
  - Added descriptive subtitles to some of the labels
  - Changed flex box of sdk languages to a grid (only enabled via a `variant` prop)
  - Added a search bar for SDK languages
  - Payload Security button styling and behavior updated
  - Hid some options on creation to make creation flow easier. 
    - These options are now only present when editing.
  - Tweaks to spacing and language throughout 

### Testing

- [x] Ensure ability to create a new SDK connection 
  - Cover every flow (sdk languages that support remote evaluated vs not, nextjs, etc.)
  - Ensure search functionality works
  - Ensure extra payload security options are not visible but defaults are set correctly in form
  - Ensure advanced settings are not visible but defaults are set correctly in form
- [x] Ensure ability to edit an SDK connection
  - Ensure extra payload security options are visible only when using "ciphered"
  - Ensure advanced settings collapsible is visible and works as expected
- [x] Ensure other flows that involve the  `SDKLanguageSelector` componenet (such as `/setup`) remain unchanged.  

### Screenshots

**New SDK Connection**
image

**Edit SDK Connection**
image


packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -431,12 +456,37 @@ export default function SDKConnectionForm({
...
       cta={cta}
     >
-      <div className="px-2 pb-2">
-        <Field label="Name" {...form.register("name")} required />
+      <div className="px-2">
+        <Field
+          label={
+            <div>
+              Connection Name
+              <Text as="p" size="1" weight="regular" color="gray" mb="0" mt="1">
```
bryce-fitzsimons, line 464: opinion: this text seems unnecessary. Field's `placeholder` is a better pattern IMO

packages/front-end/components/Features/SDKConnections/SDKLanguageSelector.tsx
```diff
@@ -20,27 +23,36 @@ export function SDKLanguageOption({
...
   language: SDKLanguage;
   selected: boolean;
   onClick: () => void;
+  variant?: "default" | "grid";
 }) {
+  const isGrid = variant === "grid";
+
   return (
     <div
       className={`hover-highlight d-inline-flex align-items-center cursor-pointer border rounded ${
```
bryce-fitzsimons, line 37: generally we should prefer Radix primitives (either our own in @ui/ or the standard Radix imports). So this would likely be a `<Flex>`

packages/front-end/components/Features/SDKConnections/SDKLanguageSelector.tsx
```diff
@@ -157,23 +180,72 @@ export default function SDKLanguageSelector({
...
+          return (
+            <TabsContent key={tab} value={tab}>
+              {variant === "grid" ? (
+                <Grid
+                  columns={{ initial: "1", sm: "2", md: "4" }}
+                  gapX="2"
+                  gapY="1"
+                  p="2"
+                  mb="5"
+                  height="126px"
```
bryce-fitzsimons, line 227: default height should be enough to avoid a scrollbar with 3 rows. And when hardcode a value like this, a comment would be helpful (eg: `// 3 SDKLanguageOption rows`)

packages/front-end/components/Features/SDKConnections/SDKLanguageSelector.tsx
```diff
@@ -157,23 +180,72 @@ export default function SDKLanguageSelector({
...
+                  overflowY="scroll"
```
bryce-fitzsimons, line 228: "auto" is better, don't need to show the scrollbar if no scrolling is necessary

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -646,385 +697,204 @@ export default function SDKConnectionForm({
...
-                buttonsWrapperClassName="sdk-security-button-wrapper mb-3"
-                buttonsClassName={(tab) =>
-                  clsx("sdk-security-button text-center border rounded", {
-                    selected: tab === getSecurityTabState(form.getValues()),
-                  })
-                }
-                tabContentsClassName={(tab) =>
-                  tab === "none" ? "d-none" : "noborder"
+            <label>Payload Security</label>
+            <ControlledTabs
```
bryce-fitzsimons, line 701: These buttons/tabs don't seem to be wired correctly to properly reflect the state of payload security options. For instance, if I click "Ciphered", it checks "encrypt" / "hash" / "hide names" for me. But if I manually uncheck those 3 items, the correct result is a security level of "Plain Text"; however, we remain with "Ciphered" checked which is no longer accurate.

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -646,385 +697,204 @@ export default function SDKConnectionForm({
...
+                    )}
+                    Ciphered
+                    <div className="subtitle">
+                      Obfuscated while still remaining cacheable.
+                    </div>
+                  </>
+                }
+              >
+                <>
+                  <div className="p-3">
```
bryce-fitzsimons, line 763: Looks like too much y-padding here

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -1039,199 +909,235 @@ export default function SDKConnectionForm({
...
-                  <div className="mb-2 d-flex align-items-center">
+                      {(form.watch("includeVisualExperiments") ||
+                        form.watch("includeRedirectExperiments")) && (
+                        <Checkbox
+                          value={form.watch("includeDraftExperiments")}
+                          setValue={(val) =>
+                            form.setValue("includeDraftExperiments", val)
+                          }
+                          label={
+                            <Tooltip
```
bryce-fitzsimons, line 1007: Somehow this Tooltip prevents me from clicking on the checkbox label text to toggle the checkbox. We don't have this problem with things like "Pass Saved Groups by reference"

packages/front-end/components/Features/SDKConnections/SDKLanguageSelector.tsx
```diff
@@ -20,27 +23,36 @@ export function SDKLanguageOption({
...
         boxShadow: selected ? "0 0 0 1px var(--text-color-primary)" : "",
+        ...(isGrid && { backgroundColor: "var(--color-panel-solid)" }),
       }}
       onClick={(e) => {
         e.preventDefault();
         onClick();
       }}
     >
-      <SDKLanguageLogo language={language} showLabel={true} size={30} />
+      <SDKLanguageLogo
```
bryce-fitzsimons, line 51: the "BETA" tags on some of the logos (Edge) seem to be misaligned, should be in the top-right corner

packages/front-end/components/Features/SDKConnections/SDKLanguageSelector.tsx
```diff
@@ -157,23 +180,72 @@ export default function SDKLanguageSelector({
...
+                position: "absolute",
+                top: "50%",
+                left: "12px",
+                transform: "translateY(-50%)",
+                color: "#aaa",
+                pointerEvents: "none",
+                zIndex: 10,
+              }}
+            />
+            <Field
```
bryce-fitzsimons, line 197: I was almost expecting the search field to apply to the unfiltered list, not search within a tab. Seems less useful as-is because each tab is already such a small subset of the total languages.

packages/front-end/styles/global.scss
```diff
@@ -3343,22 +3348,45 @@ input:disabled {
     box-sizing: border-box;
     flex: 1;
     padding: 24px 20px 18px;
-    max-width: 300px;
+    background-color: var(--color-surface);
```
bryce-fitzsimons, line 3351: I wonder if we could have used pure Radix with a handful of targeted overrides here instead of a ton of custom styles.

ahdriel: I think we should yes. And ideally we can avoid a lot (all?) of these `!important` too.

It might need a bit more refactor but I think it would be a good thing to do.

PR #5097, opened 2026-01-15
feat(holdouts): Holdout Scheduling

msamper:
### Features and Changes

Adds the ability to schedule the different phases of a holdout in advance. 
A new agenda job was added for managing the holdout status updates for scheduled events.

Some changes were needed to existing holdout functionality to make this work as expected. Starting a holdout was still being done via the experiment status endpoint so that was changed to happen within the holdout edit status endpoint instead. I also created a new `HoldoutInterfaceStringDates` type since dates are string when they reach the front-end and we were incorrectly typing them as `Date`

### Testing
[Design Reference](https://www.figma.com/design/fX5eE0xUWIWgtX65SLnOip/Holdout-Scheduling-4.3?node-id=63-19135&m=dev)

**Happy Path**
- [ ] Create a holdout schedule with all the phases scheduled
- [ ] Make sure schedule updates holdout status and ProgressBar as expected (see designs)
- [ ] Delete schedule and make sure it's cleared
- [ ] Try editing the schedule at during different phases and make sure you can only edit phases that have not passed yet

**Manual Intervention**
- [ ] Create a holdout schedule with all the phases scheduled
- [ ] Using the more menu, manually advance to the next phase and make sure status is updated as well as the corresponding date on the schedule. Information about the next update in the header should also be updated

**Edge Cases**
- [ ] Try to create a schedule with non-consecutive dates (ERROR)
- [ ] Try to create a schedule where you schedule a future phase without scheduling the next immediate phases. i.e. trying to schedule the holdout stop without scheduling the holdout start and analysis start (ERROR)
- [ ] Use Force Status Change to change the holdout status to something else. Schedule should be deleted.

### Screenshots

Screenshot 2026-02-11 at 5 30 53 PM
Screenshot 2026-02-12 at 10 19 19 AM
Screenshot 2026-02-03 at 9 58 25 AM
Screenshot 2026-02-03 at 10 07 49 AM

### DIFF
--- a/packages/back-end/src/jobs/updateHoldoutStatus.ts
+++ b/packages/back-end/src/jobs/updateHoldoutStatus.ts
@@ -80,19 +80,29 @@ const updateSingleHoldout = async (job: UpdateSingleHoldoutJob) => {
     throw new Error("Holdout experiment not found: " + holdout.id);
   }
   if (holdoutExperiment.archived) {
-    logger.info(
-      "Skipping status update: Holdout experiment is archived: " + holdout.id,
-    );
+    logger.info(`Skipping status update: Holdout ${holdout.id} is archived`);
     return;
   }
   if (holdoutExperiment.status === "stopped") {
-    logger.info(
-      "Skipping status update: Holdout experiment is stopped: " + holdout.id,
-    );
+    logger.info(`Skipping status update: Holdout ${holdout.id} is stopped`);
     return;
   }
 
   const now = new Date();
+  const scheduled = holdout.nextScheduledStatusUpdate;
+  if (!scheduled?.date) {
+    logger.info(
+      `Skipping status update: Holdout ${holdout.id} has no scheduled update`,
+    );
+    return;
+  }
+  if (scheduled.date > now) {
+    logger.info(
+      `Skipping status update: Holdout ${holdout.id} scheduled update is in the future (possibly rescheduled).`,
+    );
+    return;
+  }
+
   const phases = [...holdoutExperiment.phases] as ExperimentPhase[];
 
   let newNextScheduledStatusUpdate: {
@@ -103,7 +113,7 @@ const updateSingleHoldout = async (job: UpdateSingleHoldoutJob) => {
   try {
     logger.info("Start Updating Status for holdout " + holdout.id);
 
-    switch (holdout.nextScheduledStatusUpdate?.type) {
+    switch (scheduled.type) {
       case "start": {
         const changes: Changeset = await getChangesToStartExperiment(
           context,


packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    switch (holdout.nextScheduledUpdateType) {
+      case "start":
+        phases[0] = {
+          ...phases[0],
+          dateEnded: undefined,
+        };
+        if (phases[1]) {
+          phases = [phases[0]];
+        }
```
msamper: This logic may not be right. Should probably use the same logic I added to the holdout edit status endpoint here

packages/back-end/src/models/HoldoutModel.ts
```diff
@@ -1,11 +1,14 @@
 import { HoldoutInterface, holdoutValidator } from "shared/validators";
 import { ExperimentInterface } from "shared/types/experiment";
+import { getCollection } from "back-end/src/util/mongo.util";
 import { MakeModelClass } from "./BaseModel";
 import { getExperimentById } from "./ExperimentModel";
 
+const COLLECTION_NAME = "holdouts";
```
msamper, line 7: undo this change

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -701,7 +719,21 @@ export default function ExperimentHeader({
           </Box>
 
           <Flex direction="row" align="center" gap="2" flexShrink="0">
-            {canRunExperiment ? (
+            {isHoldout &&
+            holdout?.nextScheduledUpdate &&
+            holdout.nextScheduledUpdateType ? (
+              <Linkbutton variant="ghost" href={`/holdouts/${holdout?.id}`}>
```
msamper: Should be an anchor link to holdout schedule section

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,199 @@
...
+        );
+
+        if (holdout.scheduledStatusUpdates?.startAnalysisPeriodAt) {
+          newNextScheduledUpdateType = "startAnalysisPeriod";
+          newNextScheduledUpdate =
+            holdout.scheduledStatusUpdates.startAnalysisPeriodAt;
+        } else if (holdout.scheduledStatusUpdates?.stopAt) {
+          newNextScheduledUpdateType = "stop";
+          newNextScheduledUpdate = holdout.scheduledStatusUpdates.stopAt;
+        }
```
msamper: Should probably be more strict here and only allow changing the `nextScheduledUpdate` to the next consecutive event just in case they somehow schedule a start and stop without scheduling the analysis phase, for example

packages/front-end/components/Holdout/EditScheduleModal.tsx
```diff
@@ -0,0 +1,153 @@
+import { HoldoutInterfaceStringDates } from "shared/validators";
+import { useForm } from "react-hook-form";
+import { Box, Text } from "@radix-ui/themes";
+import { ExperimentInterfaceStringDates } from "shared/types/experiment";
+import { useState } from "react";
+import { useAuth } from "@/services/auth";
+import Modal from "@/components/Modal";
+import ScheduleStatusChangeInputs from "./ScheduleStatusChangeInputs";
+
+const validateSchedule = (
```
msamper: Consolidate into one validation function?

packages/front-end/components/Holdout/HoldoutSchedule.tsx
```diff
@@ -0,0 +1,151 @@
...
+      weight: secondSegmentWeight,
+      completion: getCompletion(startAnalysisPeriodDate, stopDate),
+      color: "amber",
+    },
+    {
+      id: "3",
+      weight: thirdSegmentWeight,
+      completion: 0,
+      color: "slate",
+    },
```
msamper: Maybe this should be handled by the `ProgressBar` component. i.e. if segment weights don't add up to 100 we add a `slated` striped progress bar to complete the bar

packages/front-end/components/Holdout/HoldoutSchedule.module.scss
```diff
@@ -0,0 +1,35 @@
+@keyframes textA {
```
msamper: Should probably rename this to something more meaningful. Maybe `status` for `textA` and `details` for `textB`

packages/front-end/ui/Text.tsx
```diff
@@ -82,6 +82,8 @@ export default function Text({
     style.color = "var(--color-text-mid)";
   } else if (color === "text-low") {
     style.color = "var(--color-text-low)";
+  } else if (color === "text-disabled") {
```
msamper, line 85: Update Text.stories to include this new color

packages/front-end/components/DatePicker/index.tsx
```diff
@@ -72,6 +72,16 @@ export default function DatePicker({
     ),
   );
   const [open, setOpen] = useState(false);
+
+  // TODO: Check why date is buffered
```
msamper: Let me ask Bryce about this. Realizing the useEffects I added here don't really make sense since they defeat the purpose of these buffered date states

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,196 @@
...
+        );
+
+        if (holdout.scheduledStatusUpdates?.startAnalysisPeriodAt) {
+          newNextScheduledUpdateType = "startAnalysisPeriod";
+          newNextScheduledUpdate =
+            holdout.scheduledStatusUpdates.startAnalysisPeriodAt;
+        }
+
+        await context.models.holdout.update(holdout, {
+          analysisStartDate: undefined,
```
lukesonnet: Is this just clearing it in case it was set previously for some reason?

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,196 @@
...
+          nextScheduledUpdateType: newNextScheduledUpdateType,
+          nextScheduledUpdate: newNextScheduledUpdate,
+        });
+        await updateExperiment({
+          context,
+          experiment: holdoutExperiment,
+          changes,
+        });
+        queueSDKPayloadRefresh({
```
lukesonnet, line 138: Should the model instead monitor for meaningful updates and then always trigger the payload refresh so it can happen no matter what the trigger is?

Probably wouldn't make this change now, but something to consider.

packages/back-end/src/routers/holdout/holdout.controller.ts
```diff
@@ -372,7 +376,92 @@ export const updateHoldout = async (
...
+          : undefined,
+      }),
+    };
+
+    // Compute next scheduled update and next scheduled update type
+    // Next scheduled update should be the earliest date among startAt, startAnalysisPeriodAt, and stopAt that is in the future
+    const now = new Date();
+    const potentialUpdates: Array<{
+      date: Date;
+      type: "start" | "startAnalysisPeriod" | "stop";
```
lukesonnet: nit: should this type be set somewhere else and be used across this PR

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,196 @@
...
+        // set the status to stopped for the experiment
+        await updateExperiment({
+          context,
+          experiment: holdoutExperiment,
+          changes: {
+            phases,
+            status: "stopped",
+          },
+        });
+        queueSDKPayloadRefresh({
```
lukesonnet, line 201: I wonder if this method could be moved to afterUpdate, but I think it would require too much logic about the change in states, so probably best to just keep it here.

packages/front-end/components/Experiment/EditStatusModal.tsx
```diff
@@ -4,19 +4,21 @@ import {
 } from "shared/types/experiment";
 import { useForm } from "react-hook-form";
 import { datetime } from "shared/dates";
-import { HoldoutInterface } from "shared/validators";
+import { HoldoutInterfaceStringDates } from "shared/validators";
+import { Box, Text } from "@radix-ui/themes";
```
ahdriel: question: Why are we using Text from radix-ui? Can we import it from ui/Text instead?

packages/front-end/components/Experiment/TabbedPage/startHoldoutAnalysisModal.tsx
```diff
@@ -1,3 +1,4 @@
+import { Box, Text } from "@radix-ui/themes";
```
ahdriel: question: Same here.. can we import from `ui/Text` instead?

packages/shared/src/validators/holdout.ts
```diff
@@ -19,7 +27,53 @@ export const holdoutValidator = z
     linkedFeatures: z.record(z.string(), holdoutLinkedItemValidator),
     environmentSettings: z.record(z.string(), featureEnvironment),
     analysisStartDate: z.date().optional(),
+    scheduledStatusUpdates: scheduledUpdatesValidator.optional(),
```
ahdriel: nit: Should this be named `scheduledUpdateConfiguration` or something?

To me the current naming creates a distinction between `nextScheduledUpdate` and `scheduledStatusUpdates` when in reality they are very related

packages/shared/src/validators/holdout.ts
```diff
@@ -19,7 +27,53 @@ export const holdoutValidator = z
...
+    nextScheduledUpdate: z.date().optional().nullable(),
+    nextScheduledUpdateType: z
+      .enum(["start", "startAnalysisPeriod", "stop"])
+      .optional()
+      .nullable(),
```
ahdriel: question: Why are we supporting both `optional` and `nullable`?

Do they have different meanings? If yes I think we should call it out.
If not, can we support only 1? Either `null` or `undefined`.

packages/shared/src/validators/holdout.ts
```diff
@@ -19,7 +27,53 @@ export const holdoutValidator = z
...
+  })
+  .strict();
+
+const _holdoutStringDatesValidator = holdoutValidator
```
ahdriel, line 47: oof -- this makes sense but I wish we did not have to do this.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1888,6 +1887,13 @@ export async function postExperimentStatus(
   if (experiment.organization !== org.id) {
     throw new Error("You do not have access to this experiment");
   }
+  if (experiment.type === "holdout") {
+    res.status(400).json({
+      status: 400,
+      message: "Cannot edit the status of a holdout through this endpoint",
```
ahdriel: nit: should we tell which endpoint to use?

packages/back-end/src/models/HoldoutModel.ts
```diff
@@ -38,6 +41,122 @@ export class HoldoutModel extends BaseClass {
...
+  }
+
+  public static async getAllHoldoutsToUpdate(): Promise<
+    { id: string; organization: string }[]
+  > {
+    const now = new Date();
+
+    const holdouts = await getCollection<HoldoutInterface>(COLLECTION_NAME)
+      .find({
+        nextScheduledUpdate: {
```
ahdriel: suggestion: We should create a new index for `nextScheduledUpdate` as this is now the basis for this query. See `additionalIndexes`

packages/back-end/src/models/HoldoutModel.ts
```diff
@@ -38,6 +41,122 @@ export class HoldoutModel extends BaseClass {
...
+      updates.scheduledStatusUpdates?.stopAt &&
+      !updates.scheduledStatusUpdates?.startAnalysisPeriodAt
+    ) {
+      throw new Error(
+        "To set a stop date, you must first set an analysis start date",
+      );
+    }
+
+    // Check if the dates are consecutive
+    const dateError =
```
ahdriel, line 122: nit: should we extract those variables before comparing them just to make the code smaller?

packages/back-end/src/models/HoldoutModel.ts
```diff
@@ -38,6 +41,122 @@ export class HoldoutModel extends BaseClass {
...
+    if (!holdoutExperiment) {
+      throw new Error("Holdout experiment not found");
+    }
+
+    const now = new Date();
+
+    // Check if one of the scheduled dates is in the past
+    if (
+      updates.scheduledStatusUpdates?.startAt &&
+      holdoutExperiment.status === "draft" &&
```
ahdriel, line 69: question: If we pass a string here that is not a valid date, what happens?

Or maybe nothing happens because Zod validates that this is a proper date string beforehand?

packages/back-end/src/jobs/updateHoldoutStatus.ts
msamper: I think I just copied this over from the edit status endpoint which needs to be able to move backwards in state, but shouldn't be needed here

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,196 @@
...
+  if (!holdout) return;
+
+  const holdoutExperiment = await getExperimentById(
+    context,
+    holdout.experimentId,
+  );
+
+  if (!holdoutExperiment) return;
+  if (holdoutExperiment.archived) return;
+  if (holdoutExperiment.status === "stopped") return;
```
ahdriel: suggestion: Should we log why we are skipping the holdout?

packages/front-end/components/Experiment/EditStatusModal.tsx
msamper: Ah my bad I must have missed these. These were added before your typography components were merged. Will update!

packages/shared/src/validators/holdout.ts
msamper: Yeah, I don't love this either. Was trying to think of a better solution

packages/back-end/src/models/HoldoutModel.ts
msamper: Seems like the first step in `updateOne` which is the caller of `beforeUpdate` is to validate so Zod should catch this

packages/front-end/ui/ProgressBar.tsx
```diff
@@ -0,0 +1,144 @@
...
+  color: "slate" | RadixColor | "disabled";
+  endBorder?: boolean;
+  tooltip?: string;
+};
+
+type ProgressBarProps = {
+  segments: Segment[];
+};
+
+export function ProgressBar({ segments }: ProgressBarProps) {
```
ahdriel, line 31: nit: Can we add `ProgressBar.stories` to show examples and import it in `pages/design-system`?

packages/front-end/ui/ProgressBar.tsx
```diff
@@ -0,0 +1,144 @@
...
+            weight: remainingWeight,
+            completion: 0,
+            color: "slate" as const,
+          },
+        ]
+      : segments;
+  return (
+    <Flex
+      wrap="nowrap"
+      className="h-2 w-full rounded-full bg-gray-200"
```
ahdriel: issue: Can we do this without bootstrap classes?

My understanding is that we are moving away from Bootstrap and I would love to be able to remove all of its CSS eventually.

packages/front-end/ui/ProgressBar.tsx
```diff
@@ -0,0 +1,144 @@
...
+                        ? isFirst && isLast
+                          ? "4px"
+                          : isFirst
+                            ? "4px 0 0 4px"
+                            : isLast
+                              ? "0 4px 4px 0"
+                              : "0"
+                        : isFirst
+                          ? "4px 0 0 4px"
+                          : "0",
```
ahdriel: question: Can we achieve the same with `:first-child` and `:last-child` in CSS? I feel it can be more readable than the current solution

packages/front-end/ui/ProgressBar.tsx
```diff
@@ -0,0 +1,144 @@
...
+      className="h-2 w-full rounded-full bg-gray-200"
+      style={{ height: "24px", borderRadius: "4px" }}
+      overflow="hidden"
+      my="4"
+    >
+      {segmentsWithFiller.map((segment, i) => {
+        const isFirst = i === 0;
+        const isLast = i === segmentsWithFiller.length - 1;
+        const completionPct = segment.completion / 100;
+        const completedWidth = completionPct * 100;
```
ahdriel: issue: We can just use `completionPct`. We are `/ 100` and then `* 100` so we can just skip that.

packages/front-end/ui/ProgressBar.tsx
```diff
@@ -0,0 +1,144 @@
...
+        };
+
+        return (
+          <Tooltip
+            content={segment.tooltip}
+            enabled={!!segment.tooltip}
+            key={i}
+          >
+            <Flex
+              key={i}
```
ahdriel: suggestion: The segments have ids, can we use that instead of the `index` here?

packages/front-end/ui/DropdownMenu.tsx
```diff
@@ -269,7 +271,13 @@ export function DropdownMenuItem({
       >
         <Flex as="div" justify="between" align="center" style={style}>
           <Box as="span" className={`mr-2 ${loading ? "font-italic" : ""}`}>
-            {children}
+            {tooltip ? (
+              <Tooltip body={tooltip} tipPosition="left">
+                {children}
+              </Tooltip>
+            ) : (
+              children
```
ahdriel: question: If we already have children, what do you think about letting the children be responsible for wrapping it in the tooltip if they want?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2074,6 +2061,14 @@ export async function postExperimentStop(
     return;
   }
 
+  if (experiment.type === "holdout") {
+    res.status(400).json({
+      status: 400,
+      message: "Cannot stop a holdout through this endpoint",
```
ahdriel: suggestion: Same here, can we direct users to the new endpoint?

I know right now this is internal, but we (hopefully) can merge them all in the future.

packages/back-end/src/routers/holdout/holdout.router.ts
```diff
@@ -15,5 +15,6 @@ router.delete(
   "/:id/feature/:featureId",
   holdoutController.deleteHoldoutFeature,
 );
+router.delete("/:id/schedule", holdoutController.deleteHoldoutSchedule);
```
ahdriel: question: This pattern is something we don't have in other places. Should we rely on `editHoldout` to clear out the schedule?

I mean that we don't have a dedicated endpoint to create the schedule, but we do to remove it.

packages/back-end/src/util/holdouts.ts
```diff
@@ -26,7 +26,9 @@ export function getAffectedSDKPayloadKeys(
 
   const environments = getEnabledEnvironments(holdout, allowedEnvs);
 
-  const projects = new Set(holdout.projects);
+  const projects = new Set(
+    holdout.projects.length > 0 ? holdout.projects : [""],
+  );
```
ahdriel, line 31: question: Why do we need this change?

For empty `projects` we are going from an empty set to a Set with an empty string.

packages/front-end/components/Holdout/ScheduleStatusChangeInputs.tsx
```diff
@@ -0,0 +1,170 @@
...
+    <Box my="4">
+      <Box mb="2">
+        <Text weight="medium">Start Holdout</Text>
+      </Box>
+
+      {!isRunning && !isStopped && !isArchived ? (
+        <Flex direction="row" gap="2" align="baseline">
+          <Box flexGrow="1">
+            <DatePicker
+              date={form.watch("scheduledStatusUpdates.startAt")}
```
ahdriel: nit: should we use `startDate` here?

Also applies to all other props that we are passing `form.watch` to in this file.

packages/front-end/components/Holdout/LinkedFeaturesTable.tsx
```diff
@@ -8,7 +8,7 @@ import Link from "@/ui/Link";
 import ValueDisplay from "@/components/Features/ValueDisplay";
 
 interface Props {
-  holdout: HoldoutInterface;
+  holdout: HoldoutInterfaceStringDates;
```
ahdriel, line 11: question: Are these changes because the model comes from mongoose with Dates in a string format and not a Date object?

I don't think we need to handle it in this PR, but I want to understand why we are updating the types so we can find a better solution for this in the future.

packages/front-end/components/Holdout/HoldoutSchedule.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+// These values are hardcoded to line up with the Analysis label
+const NO_STOP_DATE_WEIGHT = 90.5;
+const NO_START_ANALYSIS_DATE_WEIGHT = 87;
+
+function pickEarlierDate(
+  date1: string | undefined,
+  date2: string | undefined,
+): Date | null {
+  if (!date1 && date2) return new Date(date2);
```
ahdriel: issue: can we create the 2 Date objects first and then compare them?

Right now we are creating them many times when we don't need to

packages/back-end/src/models/HoldoutModel.ts
msamper: Good callout! will do

packages/front-end/components/Holdout/HoldoutSchedule.tsx
```diff
@@ -0,0 +1,243 @@
...
+  if (date1 && !date2) return new Date(date1);
+
+  if (!date1 || !date2) return null;
+
+  return new Date(date1) < new Date(date2) ? new Date(date1) : new Date(date2);
+}
+
+function getCompletion(startDate: Date | null, endDate: Date | null): number {
```
ahdriel: nit: should we clarify it is a percentage? `getCompletionPercentage`

packages/front-end/components/Holdout/HoldoutSchedule.tsx
```diff
@@ -0,0 +1,243 @@
...
+  );
+
+  const segments: Segment[] = [
+    {
+      id: "holdout",
+      weight: showUnscheduledSegment
+        ? !startAnalysisPeriodDate
+          ? NO_START_ANALYSIS_DATE_WEIGHT
+          : NO_STOP_DATE_WEIGHT
+        : 68,
```
ahdriel: issue: Can we move this hardcoded weight to a variable as well?

packages/front-end/components/Holdout/HoldoutSchedule.tsx
```diff
@@ -0,0 +1,243 @@
...
+      color: isDraft ? "slate" : "indigo",
+      endBorder: isDraft ? false : true,
+      tooltip:
+        holdoutSegmentCompletion === 100
+          ? "Holdout has stopped—no new Experiments or Features can be added"
+          : undefined,
+    },
+    {
+      id: "analysis",
+      weight: showUnscheduledSegment ? 0 : 32,
```
ahdriel: same for these weights

packages/front-end/components/Holdout/HoldoutSchedule.tsx
```diff
@@ -0,0 +1,243 @@
...
+      <Flex justify="between">
+        <Box>
+          {experiment.status === "draft" ? (
+            <>
+              <Text weight="medium" color="text-high">
+                Start:{" "}
+              </Text>
+              <Text
+                color={startDate ? "text-high" : "text-disabled"}
+                weight="regular"
```
ahdriel: nit: this is the default weight, should we remove this line?

packages/front-end/components/Holdout/ScheduleStatusChangeInputs.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return (
+    <Box my="4">
+      <Box mb="2">
+        <Text weight="medium">Start Holdout</Text>
+      </Box>
+
+      {!isRunning && !isStopped && !isArchived ? (
+        <Flex direction="row" gap="2" align="baseline">
+          <Box flexGrow="1">
+            <DatePicker
```
ahdriel, line 57: question: Should we clarify what timezone is being used for the scheduling? Right now we don't say anything

packages/front-end/components/Holdout/EditScheduleModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+          }),
+          ...(experiment.status !== "stopped" && {
+            stopAt: rawValue.scheduledStatusUpdates.stopAt
+              ? new Date(rawValue.scheduledStatusUpdates.stopAt).toISOString()
+              : "",
+          }),
+        }
+      : undefined;
+
+    await apiCall<{
```
ahdriel, line 61: question: how are we handling errors in this endpoint? is this being raised somewhere in the UI?

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -123,7 +134,19 @@ export default function SetupTabOverview({
         />
       ) : null}
       <div>
-        <h2>Overview</h2>
+        <Flex justify="between" align="baseline" mb="3">
+          <h2>Overview</h2>
+          {canEditExperiment &&
```
ahdriel: nit: all of these permission checks feel very important -- maybe we can find a way to centralize it

packages/front-end/components/Holdout/EditScheduleModal.tsx
msamper: Yes, the modal component handles errors with its submit callback and displays them in a callout up top

packages/front-end/components/Holdout/LinkedFeaturesTable.tsx
msamper: It's because when the Holdout object is sent to the UI from the API the dates are serialized to strings. This pattern with a `StringDates` version of a type exists today with experiments. Was thinking about this and maybe the right solution is to leverage zod to coerce the fields we know are supposed to be dates back into dates when receiving them via the API in the UI and vice versa. https://zod.dev/api?id=coercion

msamper: Adding an agenda item for this at the eng meeting!

packages/front-end/ui/ProgressBar.tsx
msamper: Yup good idea

packages/back-end/src/util/holdouts.ts
msamper: There is a bug that exists today in main where if a holdout doesn't have a project, the payload cache isn't being updated when the holdout is stopped. Luke and I investigated and it seems like the issue is coming from here because when there are no projects we return 0 sdk payload keys which results in no refresh. We do something similar already with features (https://github.com/growthbook/growthbook/blob/msamper/holdout-scheduling/packages/back-end/src/util/features.ts#L258-L262)

packages/shared/src/validators/holdout.ts
msamper: Added a comment since they have different meanings

packages/front-end/components/Holdout/ScheduleStatusChangeInputs.tsx
msamper: I think that would definitely be more clear. I tried appending it to the end of the field component but ran into some issues there. There's some custom styling we're doing with the date picker that seems to be making the appended string not visible. If we want to go that route I think I would prefer to do add it in a separate PR to get this one merged. Also open to suggestions on where else to display the timezone. To the right of the field won't work since we have a clear button

ahdriel (approved):
Nice work!

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  let newNextScheduledStatusUpdate: {
+    type: "start" | "startAnalysisPeriod" | "stop";
+    date: Date;
+  } | null = null;
+
+  try {
+    logger.info("Start Updating Status for holdout " + holdout.id);
+
+    switch (holdout.nextScheduledStatusUpdate?.type) {
```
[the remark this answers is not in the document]
msamper: The chances of this situation happening are very very small, but adding a check just in case

packages/back-end/src/routers/holdout/holdout.controller.ts
```diff
@@ -458,10 +587,13 @@ export const editStatus = async (
...
         lookbackStartDate: new Date(),
         dateEnded: undefined,
-        name: "Analysis Period",
+        name: "Analysis",
       };
       await context.models.holdout.update(holdout, {
         analysisStartDate: new Date(),
+        nextScheduledStatusUpdate: holdout.statusUpdateSchedule?.stopAt
+          ? { type: "stop", date: holdout.statusUpdateSchedule.stopAt }
+          : null,
```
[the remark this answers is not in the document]
msamper: May be nice to revisit this later because the logic in `editStatus` is pretty twisty, but not addressing this for now. The only case this could possibly happen is when a user is forcing a status change backwards and that's already a backdoor path we don't fully support/expect users to use often.

PR #5100, opened 2026-01-15
Don't create empty Experiment Updated events.

tzjames:
### Features and Changes

A customer complained about `experiment.update` events getting created when `changes` is empty, and previous_attributes only has `dateUpdated` as the only difference. It turns out this happens because for the diff we use `ApiExperiment` which doesn't have all the same properties as `ExperimentInterface`/`ExperimentModel`.  So any time an Experiment is updated where it updates internal only fields an event still gets generated showing nothing changes.  

A very common place where this happens is when a new automatic snapshot happens. The `lastSnapshotAttempt` and `nextSnapshotAttempt` get updates on the model, but the `ApiExperiment` doesn't have these fields.

This PR fixes the issue by not creating an event when the change log is empty.  However since the `dateUpdated` would still be changed in the background, it creates a new `dateApiUpdated`. For the API, the diff for events, and any other external uses of experiment we now send down that the `dateUpdated` = `dateApiUpdated`. We then only update `dateApiUpdated` if changes get made to one of the external properties.  Hence from an api perspective they never know that these background updates are happening to the experiment. 

In fixing this, a bug was discovered with the `getObjectDiff` function where top-level arrays were not properly processed with the `ignoredKeys` for `variations.screenshots` doing nothing.  Every time `toExperimentApiInterface` gets called new screenshot links get generated with a new signed image url that works for 15 minutes, and we don't want that to count as real changes.  If it was within a second the same signed image url gets generated, but in the snapshot where one is calculated beforehand and one after it wasn't properly ignoring that.  

So the main changes are:
1. Don't create an event when the change log is empty.
2. Keep a separate `dateApiUpdated` property on Experiments that tracks only when an external date is changed
3. Fix a bug in `getObjectDiff`
4. Add a unit test file to thoroughly test that `getObjectDiff` is working correctly.

This PR only adds logic to getObjectDiff to make it clear there are no regressions for things that previously worked.  A future PR will refactor it with helper functions for repeated logic.

### Testing

`yarn test`

In mongo set `nextSnapshotAttempt` on the sample experiment to some time in the past..
Restart the dev server.
See in the logs that the snapshot is being calculated:
```
 | [1] [16:41:16.479] INFO (74638): Start Refreshing Results for experiment exp_4posg253hmkcqpa9e
...
 | [1] [16:41:26.966] INFO (74638): Successfully Refreshed Results for experiment exp_4posg253hmkcqpa9e
```
See on the /events page that no new event got generated

Update the description on an experiment and note the time you update it.
See on the /events page that a new event got generated.

Again In mongo set `nextSnapshotAttempt` on the sample experiment to some time in the past..
Restart the dev server.
See in the logs that the snapshot is being calculated. Again note the time.
Wait a minute and then again Update the description on an experiment and note the time you update it.
On the /events page look at the new event's previous_attributes.dateUpdated - it should be the time from the first time you updated the description and not from when the subsequent snapshot happened when you restarted the dev server.

### DIFF
--- a/packages/back-end/src/events/handlers/webhooks/event-webhooks-utils.ts
+++ b/packages/back-end/src/events/handlers/webhooks/event-webhooks-utils.ts
@@ -555,8 +555,8 @@ export function getObjectDiff(
             };
             result.modified.push(hierarchicalMod);
           }
-        } else if (Array.isArray(prev[key]) && Array.isArray(curr[key])) {
-          // Handle top-level arrays with nested config
+        } else {
+          // Array handling for other nested configs
           const prevArray = prev[key] as unknown[];
           const currArray = curr[key] as unknown[];
 
@@ -666,23 +666,7 @@ export function getObjectDiff(
               };
               result.modified.push(hierarchicalMod);
             }
-          } else {
-            // No idField, treat as simple modification
-            const simpleMod: SimpleModification = {
-              key,
-              oldValue: prevArray,
-              newValue: currArray,
-            };
-            result.modified.push(simpleMod);
           }
-        } else {
-          // Non-array object without specific handling
-          const simpleMod: SimpleModification = {
-            key,
-            oldValue: prev[key],
-            newValue: curr[key],
-          };
-          result.modified.push(simpleMod);
         }
       } else if (Array.isArray(prev[key]) && Array.isArray(curr[key])) {
         // Handle arrays without nested config (fallback for arrays without idField)


packages/back-end/src/events/handlers/webhooks/event-webhooks-utils.ts
```diff
@@ -561,36 +685,16 @@ export function getObjectDiff(
...
-        } else {
-          // For arrays of objects, we need nested config to know the ID field
-          // If no config, fall through to simple field change
-          const simpleMod: SimpleModification = {
-            key,
-            oldValue: prevArray,
-            newValue: currArray,
-          };
-          result.modified.push(simpleMod);
-        }
```
tzjames, line 593: This was doing the same code in both cases.

PR #5103, opened 2026-01-16
fix(api): add slice results to results payload

lukesonnet:
### Features and Changes

We needed to expand the metric map and get all expanded metrics from an experiment to ensure that metric slice results were included in the results API payload.

Alternatively, we could just use the metrics in the snapshot results and return them all even if they are stripped from the experiment object or the metric is deleted from GrowthBook.

### Testing

Did some `curl`ing. Before, no slices. After, slices!

```
curl -X GET "${API_HOST}/api/v1/experiments/${EXPERIMENT_ID}/results" \
  -u "${API_SECRET}:" \
  -H "Content-Type: application/json" | jq .
```

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -111,13 +111,13 @@ const config = {
   themeConfig:
     /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
     {
-      announcementBar: {
-        id: "announcement-bar",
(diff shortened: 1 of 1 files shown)


bryce-fitzsimons (approved):
Tested, works fine with slice results as well as top level.

Wondering if perhaps we should also return the metric name and whether it has a parent metric or not (is a slice, basically). But not something we should necessary do now, maybe not important to end users.

lukesonnet:
Yes there's more work to make results endpoints more powerful and accessible that is on the exp team roadmap.

PR #5107, opened 2026-01-16
wGl/approval flows structure

gazzdingo:
### Features and Changes


packages/shared/src/validators/approval-flows.ts
```diff
@@ -0,0 +1,136 @@
...
+  details: z.string().optional(),
+  createdAt: z.date(),
+});
+
+
+
+export const approvalFlowCreateValidator = z.object({
+  entityType: z.enum(approvalEntityTypeArray),
+  entityId: z.string(),
+  title: z.string(),
```
gazzdingo: this can be removed

PR #5108, opened 2026-01-16
Restrict Zod default usage and add new default values option to BaseModel config

msamper:
### Features and Changes

We discovered that when updating resources with a model that uses the `BaseModel` if the underlying validator had a default and you passed in a partial object without those properties with defaults, the properties with default values would get overwritten because Zod's `parse` would place the in the update object. Zod doesn't allow you to stop default values from being applied when parsing. See [this](https://growthbookapp.slack.com/archives/C07NK4F016Z/p1768504794372159?thread_ts=1768497338.549619&cid=C07NK4F016Z) thread.

This PR restricts usage of Zod's `default()` and creates a new config property for the `BaseModel` called `defaultValues` to handle this issue. Now default values are applied within the `BaseModel` and only when creating a new document. We apply the default values before parsing with Zod to make sure we have a chance to apply defaults for any required fields that might otherwise throw during parsing if they're not defined.

### Testing

**Make sure the fact metric owner/tags bug is resolved**
- [x] Update a fact metrics owner and make sure tags aren't wipe and vice versa

**General Testing**
- [x] Test creation with models that have been migrated to BaseModel defaultValues and make sure documents are created correctly
- [x] Test updating with models that have been migrated and make sure defaults are not applied


packages/back-end/src/models/BaseModel.ts
```diff
@@ -557,6 +558,15 @@ export abstract class BaseModel<
       props.owner = this.context.userName || "";
     }
 
+    // Add default values if values are not provided
+    if (this.config.defaultValues) {
+      for (const [key, value] of Object.entries(this.config.defaultValues)) {
+        if (!props[key as keyof CreateProps<z.infer<T>>]) {
```
jdorn: This won't work for falsy values.  We only want to use the default if it's undefined

packages/shared/src/validators/fact-table.ts
```diff
@@ -72,6 +72,9 @@ export const createFactTablePropsValidator = z
     name: z.string(),
     description: z.string(),
     id: z.string().optional(),
+    // Only being used in middleware for fact-table POST request so this is safe
+    // Remove when we migrate FactTableModel to use the BaseModel and use defaultValues instead
+    // eslint-disable-next-line no-restricted-syntax
     owner: z.string().default(""),
```
msamper, line 78: Happy to also just remove this if we think it's fine. Fact tables aren't using the BaseModel so I can't set defaults that way and this validator is just used for the route middleware for creation which is safe

.eslintrc
```diff
@@ -278,12 +278,24 @@
...
-  ]
+                "message": "SDK packages cannot import from internal packages.",
+              },
+            ],
+          },
+        ],
+      },
+    },
+    {
+      "files": ["./packages/shared/src/validators/*","./packages/back-end/src/routers/**/*"],
```
msamper: Can also just target all of `shared` and `back-end` if we want. Wasn't sure if we want some flexibility for models that aren't on the base model and might use validators in route middleware

packages/back-end/src/models/BaseModel.ts
```diff
@@ -447,6 +459,38 @@ export abstract class BaseModel<
...
+        currentValue !== null &&
+        typeof currentValue === "object" &&
+        !Array.isArray(currentValue)
+      ) {
+        // If both are objects (not arrays), recursively merge nested defaults
+        this._applyDefaultValues(
+          currentValue as Record<string, unknown>,
+          defaultValue as Record<string, unknown>,
+        );
+      }
```
jdorn, line 490: I think you are missing an `else` here to handle default values of primitive values?

msamper: Lines 474-476 should cover primitives right? Unless I'm missing something

packages/shared/src/validators/fact-table.ts
ahdriel: Let's keep it for now but migrate FactTables to BaseModel soon, hopefully.

.eslintrc
```diff
@@ -284,6 +284,18 @@
           }
         ]
       }
+    },
+    {
+      "files": ["./packages/shared/src/validators/*"],
```
ahdriel, line 289: issue (non-blocking): eslint overrides this instead of merging

This means that any other `no-restricted-syntax` rules will not be applied to all files inside shared/src/validators.

We might need to add this configuration of selector and message into multiple places to ensure we have proper coverage without breaking other eslint rules.

packages/front-end/components/Experiment/Templates/TemplateForm.tsx
```diff
@@ -190,7 +190,7 @@ const TemplateForm: FC<Props> = ({
     if (onCreate) {
       onCreate(res.template.id);
     } else if (isEmpty(initialValue) || isNewTemplate) {
-      router.push(`/experiments#templates`);
+      router.push(`/experiments/templates`);
```
ahdriel, line 193: Nice catch!

PR #5110, opened 2026-01-19
Add API endpoints for custom fields

Kevin-Chant:
### Features and Changes

Adds the standard CRUD endpoints for custom fields to our API
Replaces the existing list endpoint with the new style of API handlers

### Testing

First, create some [custom fields](http://localhost:3000/settings/customfields) on `main` and take a snapshot of the response for the GET endpoint (`/api/v1/custom-fields`) to compare the behavior

Using Postman or your preferred tool, check the behavior of each endpoint and verify it displays appropriately on [the custom fields page](http://localhost:3000/settings/customfields)

### Screenshots

Example output of the POST endpoint

image

Docs

image

### DIFF
--- a/packages/back-end/src/models/BaseModel.ts
+++ b/packages/back-end/src/models/BaseModel.ts
@@ -321,7 +321,7 @@ export abstract class BaseModel<
       z.ZodType<UpdateProps<z.infer<T>>>,
       z.ZodTypeAny
     >,
-  ): Promise<z.infer<ApiT> | void> {
+  ): Promise<z.infer<ApiT>> {
     const id = req.params.id;
     const rawBody = req.body;
     const toUpdate = await this.processApiUpdateBody(rawBody);
--- a/packages/back-end/src/models/CustomFieldModel.ts
+++ b/packages/back-end/src/models/CustomFieldModel.ts
@@ -36,7 +36,7 @@ const BaseClass = MakeModelClass({
     crudActions: ["create", "delete", "get", "update"],
     customHandlers: [
       defineCustomApiHandler({
-        pathFragment: "/",
+        pathFragment: "",
         verb: "get",
         operationId: "listCustomFields",
         validator: {
@@ -292,7 +292,7 @@ export class CustomFieldModel extends BaseClass {
       z.ZodTypeAny,
       z.ZodTypeAny
     >,
-  ): Promise<ApiCustomField | void> {
+  ): Promise<ApiCustomField> {
     const id = req.params.id;
     const parsedBody = apiUpdateCustomFieldBody.parse(req.body);
     const containerObject = await this.updateCustomField(id, parsedBody);
--- a/packages/shared/types/openapi.d.ts
+++ b/packages/shared/types/openapi.d.ts
@@ -386,6 +386,8 @@ export interface paths {
     get: operations["getDashboardsForExperiment"];
   };
   "/custom-fields": {
+    /** Get all custom fields */
+    get: operations["listCustomFields"];
     /** Create a single customField */
     post: operations["createCustomField"];
   };
@@ -396,10 +398,6 @@ export interface paths {
     put: operations["updateCustomField"];
     /** Delete a single customField */
     delete: operations["deleteCustomField"];
-  };
-  "/custom-fields/": {
-    /** Get all custom fields */
-    get: operations["listCustomFields"];
   };
 }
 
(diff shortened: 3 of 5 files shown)


packages/back-end/src/models/BaseModel.ts
```diff
@@ -272,20 +272,20 @@ export abstract class BaseModel<
...
-  protected async handleApiGet(
+  public async handleApiGet(
     req: ApiRequest<
       unknown,
       z.ZodType<{ id: string }>,
       z.ZodTypeAny,
       z.ZodTypeAny
     >,
-  ): Promise<z.infer<ApiT>> {
+  ): Promise<z.infer<ApiT> | void> {
```
jdorn: This function does not actually return void, right? It throws before returning.

i think you can give the throw helper function a return type of never and typescript will know that it stops execution.

packages/back-end/src/models/CustomFieldModel.ts
```diff
@@ -193,4 +218,87 @@ export class CustomFieldModel extends BaseClass {
...
+  }
+  public async handleApiCreate(
+    req: ApiRequest<unknown, z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>,
+  ): Promise<ApiCustomField> {
+    const parsedBody = apiCreateCustomFieldBody.parse(req.body);
+    const containerObject = await this.addCustomField(parsedBody);
+    const created = containerObject.fields.find(
+      ({ id }) => id === parsedBody.id,
+    );
+    return this.singleFieldToApiInterface(created!);
```
jdorn: I prefer avoiding type assertions with exclamation points. Better to explicitly handle the case where it might be undefined even if it won't happen in practice. Makes the code more future proof.

packages/back-end/src/models/BaseModel.ts
Kevin-Chant: Thanks! I was trying to figure out how to make that type inference work, didn't realize it was just `never`

packages/shared/types/openapi.d.ts
```diff
@@ -389,6 +385,22 @@ export interface paths {
...
+  };
+  "/custom-fields/{id}": {
+    /** Get a single customField */
+    get: operations["getCustomField"];
+    /** Update a single customField */
+    put: operations["updateCustomField"];
+    /** Delete a single customField */
+    delete: operations["deleteCustomField"];
+  };
+  "/custom-fields/": {
```
tzjames: This changed to have a trailing slash.

packages/back-end/src/models/CustomFieldModel.ts
```diff
@@ -16,6 +22,44 @@ const BaseClass = MakeModelClass({
...
+    schemas: {
+      createBody: apiCreateCustomFieldBody,
+      updateBody: apiUpdateCustomFieldBody,
+    },
+    pathBase: "/custom-fields",
+    includeDefaultCrud: false,
+    crudActions: ["create", "delete", "get", "update"],
+    customHandlers: [
+      defineCustomApiHandler({
+        pathFragment: "/",
```
tzjames: In order to fix the extra "/" at the end of the getAll endpoint, this should just be ""

packages/back-end/src/models/BaseModel.ts
```diff
@@ -314,14 +314,14 @@ export abstract class BaseModel<
...
-  protected async handleApiUpdate(
+  public async handleApiUpdate(
     req: ApiRequest<
       unknown,
       z.ZodType<{ id: string }>,
       z.ZodType<UpdateProps<z.infer<T>>>,
       z.ZodTypeAny
     >,
-  ): Promise<z.infer<ApiT>> {
+  ): Promise<z.infer<ApiT> | void> {
```
tzjames: This void I think is now unnecessary?

packages/back-end/src/models/CustomFieldModel.ts
```diff
@@ -124,6 +168,14 @@ export class CustomFieldModel extends BaseClass {
     };
     const existing = await this.getCustomFields();
     if (existing) {
```
tzjames, line 170: Do we also need to validate that the newCustomField.id is a valid one, like not empty or strange characters?

packages/back-end/src/models/BaseModel.ts
Kevin-Chant: Whoops, must've missed this one ty

packages/back-end/src/models/CustomFieldModel.ts
Kevin-Chant: I'll add a min length in zod. I think strange characters shouldn't matter much as it should be url escaped

PR #5125, opened 2026-01-20
Publish JS/React SDKs 1.6.3

bryce-fitzsimons:
- publish new SDKs which now support $regexi operators (case-insensitive)
- In-app warnings when SDK Connections don't support it
- Bump versions, etc
- Fix script

image

todo:
- [x] once published to NPM, run `pnpm install --no-frozen-lockfile` to resolve CI issue

### DIFF
--- a/packages/front-end/components/Metrics/MetricGroupModal.tsx
+++ b/packages/front-end/components/Metrics/MetricGroupModal.tsx
@@ -77,6 +77,7 @@ const MetricGroupModal: FC<{
                 datasource: value.datasource,
                 projects: value.projects,
                 metrics: value.metrics,
+                owner: "",
               }),
             },
           );


packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -1023,3 +1038,51 @@ function ConditionAndGroupInput({
...
+  }
+
+  return (
+    <Callout
+      status={hasSDKWithCaseInsensitiveRegex ? "warning" : "error"}
+      mt="2"
+    >
+      {hasSDKWithCaseInsensitiveRegex
+        ? "Some of your SDK Connections in this project may not support case-insensitive regex."
+        : "None of your SDK Connections in this project support case-insensitive regex. Either upgrade your SDKs to version 1.6.3+ or use case-sensitive regex operators instead."}
```
jdorn: Do not hard-code a version number here.

packages/shared/src/sdk-versioning/sdk-report.ts
```diff
@@ -24,6 +24,7 @@ const allCapabilities: Record<SDKCapability, boolean> = {
   visualEditor: true,
   visualEditorDragDrop: true,
   visualEditorJS: true,
+  caseInsensitiveRegex: true,
```
jdorn, line 27: Forgot to also add this new capability to the `nocode.json` file

packages/front-end/components/Features/ConditionInput.tsx
bryce-fitzsimons: come to think of it, we shouldn't have a version number at all.

docs/src/data/SDKInfo.ts
```diff
@@ -328,7 +337,7 @@ export default {
   },
   python: {
     name: "Python SDK",
-    version: "1.4.5",
+    version: "2.0.1",
```
madhuchavva: it's `2.1.0` for python

PR #5127, opened 2026-01-21
fix(bug): Make sure we clean up listeners for QueryRunner and PythonStatsServer

### LINKED ISSUE #5086: [Bug] Potential memory leak
### Summary

I am hosting Growthbook as a self-service.

After upgrade from `v3.60` to `v4.20` and change of healhcheck endpoint,
I see degraded performance for month long running services.

<details><summary>How health check was used differently</summary>
<p>

Assumed `/api/init` would be less heavy than `/` for multi port checks.

</p>
</details> 

#### Expected Behavior

Memory, CPU, and latency should remain stable, when there are no big traffic changes.

#### Current Behavior

* Memory gradually increasing
* CPU load gradually increasing
* Timeouts from SDK clients getting more frequent
* Errors from UI getting more frequent

Image
Image

#### Workaround

* Increasing timeout for clients - makes it slightly better, but issue remains
* Restart of a GrowthBook service - solves (for short period of time)

Image

#### Details

<details><summary>Environment </summary>
<p>

* GrowthBook running on AWS Fargate (Linux/ARM64)
* 0.5 vCPU / 2 GiB Memory
* Node.js v20.5.1
* Container image: growthbook (revision 66)

</p>
</details> 

<details><summary>Observed Behavior </summary>
<p>

* Memory utilization steadily increases from ~44% to ~88% over weeks/months
* Sawtooth pattern indicates GC runs but memory ceiling keeps rising
* No file descriptor leak (stable at ~26 FDs in growthbook container)
* JS Heap appears small (~22 MB) but total RSS grows over time

</p>
</details> 

<details><summary>Debugging Performed </summary>
<p>

* Connected Chrome DevTools via Node.js inspector (port 9229)
* Heap snapshot shows:
     * 42,116 strings (11,960 kB / 49% retained)
     * 65,317 compiled code entries (7,679 kB / 31% retained)
     * 13,924 Functions (3,138 kB / 13% retained)
* Notable: `FileSystemBlobStore`, `ArrayBuffer`, `system/JSArrayBufferData` present
* Netdata monitoring shows memory usage pattern consistent with slow leak

</p>
</details> 

<details><summary>Suspected Areas: from runtime</summary>
<p>

* Possible unbounded caching or event listener accumulation
* Native buffer/ArrayBuffer not being released
* Connection pool or subscription state growing

</p>
</details>

<details><summary>Node.js snapshot stats</summary>
<p>

Image
Image
Image

</p>
</details> 

<details><summary>Suspected Areas: from static code analysis</summary>
<p>

In [packages/back-end/src/queryRunners/QueryRunner.ts:270](https://github.com/growthbook/growthbook/blob/main/packages/back-end/src/queryRunners/QueryRunner.ts#L270):

```
await new Promise<void>((resolve, reject) => {
  this.emitter.on(FINISH_EVENT, () => {  // BUG: .on() never removes listener
    if (this.error) {
      reject(this.error);
    } else {
      resolve();
    }
  });
});
```

Each call to waitForResults() adds a new listener via .on(). When the promise resolves/rejects, the listener remains attached to the EventEmitter indefinitely.

### Additional Issues Found

* Recursive timeout without limit (lines 642-661): When reusing cached queries, `setTimeout(check, 3000)` can loop indefinitely
* Unbounded cache (line 129): `finishedQueryMapCache` has no size limit

</p>
</details>

ahdriel:
Improves or hopefully fixes #5086 

### Features and Changes

- Use `.once` instead of `.on` for the QueryRunner eventEmmiter
  - This only happens once, but we were not cleaning it up properly. So we can use `once` that manages that automatically for us.
- Clear up timers for QueryRunner & PythonStatsServer properly

### DIFF
--- a/packages/back-end/test/queryRunners/QueryRunner.test.ts
+++ b/packages/back-end/test/queryRunners/QueryRunner.test.ts
@@ -37,7 +37,7 @@ class TestQueryRunner extends QueryRunner<
 
   public setQueuedQueryTimer(queryId: string, timer: NodeJS.Timeout) {
     // @ts-expect-error Setting private prop for testing
-    this.queuedQueryTimers[queryId] = timer;
+    this.pendingTimers[queryId] = timer;
   }
 
   public executeQuerySpy = jest.fn();


ahdriel:
@Kevin-Chant Thanks for the review!

I am leaning towards closing the issue as the 3 potential points mentioned there are being addressed by this PR.

I agree that it might not solve all memory leaks, but we can always re-open the issue or create a new one if we observe them happening again.

PR #5131, opened 2026-01-21
Mk/project admins

mknowlton89:
### Features and Changes

We've received requests to introduce the concept of `Project Admins`. Currently, only admins, or those with the `TeamManagementFullAccess` policy were able to control the roles of members in their organization. However, its common for organizations to utilize `projects` within GrowthBook, and some project owners don't have admin permissions as that grants permissions to other things like billing, nor does it make sense to give the project owners the `TeamManagementFullAccess` as that gives these project owners permission to edit member's roles for their project, but others as well.

With this new PR, it's now possible for an admin (or another project-admin) to give a user a `project-admin` role for a particular project.

Project admins are then able to customize other member's project-level roles for the project(s) they're admins of.

Meaning, if User 1 was a project admin of `Project A` - they could change `User 2`'s project-level role for `Project A` from `noaccess` to `engineer`. This project admin can also adjust the project name, description, stats engine, and pre-launch checklist.

But, project-admins can not add or delete projects - that's reserved for users with the `ProjectsFullAccess` policy - which for our default roles is only available for `admins`.

### A few things to call out:
- With GrowthBook, we don't really have a concept of project-members. All members are included in all projects - but the user's roles determine their access level to a project. For example, if a user's global role is `noaccess`, and for `Project A` they have an `engineer` role, they'll only be able to see `Project A` in their projects list, and only see resources that are included in `Project A`.
- This new role is only available for orgs with the `advanced-permissions` commercial feature, as those are the only orgs that can define project-level roles.
- With this new policy, it is now possible for orgs with the `custom-roles` commercial feature (enterprise-only at the time of this PR) to create custom roles with the new policy.
- This PR also introduces a new optional `displayName` property to the `Role` interface. This does a few things. 1) It improves the UI so we can now render role names as `No Access` vs `noaccess` or `Visual Editor` instead of `visualEditor`. 2) It also allows us to introduce a new pattern where new default roles we create are prefixed with the `gbDefault_` prefix. This will reduce potential collisions with organization's custom role names. Note: I have also added some validation so moving forward, no new custom roles can be created with the `gbDefault_` prefix. It would be much more likely that a collision would happen with an id like `projectAdmin`, instead of `gbDefault_projectAdmin`. Still a bit of a security through obscurity thing. I checked Mongo and we shouldn't have any custom role id collisions with `gbDefault_projectAdmin`, and the new validation should prevent any future collisions.
> [!WARNING]
> It is still plausible that a self-hosted organization could have a collision, though, very unlikely. Calling it out regardless.

### What work went into making this possible
- Previously, we had a singular `manageProjects` permission that powered the `ProjectsFullAccess` policy. With this PR, I've introduced more granular permissions around projects - `createProjects` and `deleteProjects`.
- With these new permissions, I've updated the `ProjectsFullAccess` to include all of the project-level permissions: `createProjects`, `manageProjects`, and `deleteProjects`.
- I've also created a new policy `ProjectsAdminAccess` that only includes the `manageProjects` permission. Note: It also includes the `readData` policy, as that is a requisite for the user to view any data.
- In the permission class, we had existing permission check helpers of `canCreateProject` and `canDeleteProject` - before these changes, these helpers were checking the `manageProjects` permission - now they're checking the more granular permissions `createProjects` and `deleteProjects`, respectively.
- And in the UI, I've updated the `SingleRoleSelector` can include the `projectAdmin` role for project-level roles only. To accomplish this, I added a new optional prop `includeProjectAdminRole` that is false by default - which matches the pattern of the `includeAdminRole`. This allows us to ensure that the `projectAdmin` role is only included for project-level role selection, and not global role selection.

### Dependencies

- None

#### User with projectAdmin role for a particular project
- [x] Ensure they can't create new projects
- [x] Ensure they can't delete any projects 
- [x] Ensure they can edit the project they're an admin of (name, description, stats engine, pre-launch checklist - if the org is enterprise)
- [x] Ensure they can change a member's project role for the particular project they're viewing
- [x] Ensure they CAN'T change a member's global role from either the `/settings/members` page or from the `[pid].tsx` route.
- [x] Ensure they CAN'T access the `/settings/members` page

#### User that is a projectAdmin on multiple projects
- [x] Ensure they can't create new projects
- [x] Ensure they can't delete any projects 
- [x] Ensure they can edit any project they're an admin of (name, description, stats engine, pre-launch checklist - if the org is enterprise)
- [x] Ensure they can change a member's project role for the particular project they're viewing, but they can't for any project they're not an admin of
- [x] Ensure they CAN'T change a member's global role from either the `/settings/members` page or from the `[pid].tsx` route.
- [x] Ensure they CAN'T access the `/settings/members` page

#### User with a custom role that leverages our existing `ProjectsFullAccess` policy as their global role
- [x] Ensure they can create projects
- [x] Ensure they can delete projects
- [x] Ensure they can edit projects (name, description, stats engine, pre-launch checklist - if the org is enterprise)
- [x] Ensure they can change a member's project role for the particular project they're viewing
- [x] Ensure they CAN'T change a member's global role from either the `/settings/members` page or from the `[pid].tsx` route.
- [x] Ensure they CAN'T access the `/settings/members` page

#### User with a custom role that leverages the new `ProjectAdminAccess` policy at a project level
- [x] Ensure they can't create new projects
- [x] Ensure they can't delete any projects 
- [x] Ensure they can edit the project they're an admin of (name, description, stats engine, pre-launch checklist - if the org is enterprise)
- [x] Ensure they can change a member's project role for the particular project they're viewing
- [x] Ensure they CAN'T change a member's global role from either the `/settings/members` page or from the `[pid].tsx` route.
- [x] Ensure they CAN'T access the `/settings/members` page

#### User with a custom role that leverages the new `ProjectAdminAccess` policy at the global level
- [x] Ensure they can't create new projects
- [x] Ensure they can't delete any projects 
- [x] Ensure they can edit any project in the org (name, description, stats engine, pre-launch checklist - if the org is enterprise)
- [x] Ensure they can change a member's project role for the particular project they're viewing
- [x] Ensure they CAN'T change a member's global role from either the `/settings/members` page or from the `[pid].tsx` route.
- [x] Ensure they CAN'T access the `/settings/members` page

#### General
- [x] Ensure the `projectAdmin` role option is only available for orgs with the `advanced-permissions` commercial feature (so, it's not available for free orgs at the time of this PR)
- [x] Ensure that if an admin user is viewing a project id page, they still see the standard `Edit Role` dropdown on the `MembersList` and not the `Edit Project Role` option. They have permission to edit the user's full role, not just their project role.
- [x] Ensure the `projectAdmin` is not an option when determining a user's global role
> [!NOTE]
> The `projectAdmin` role will work at the global level, but to me, it feels more appropriate as a project-scoped role. Softly head opinion.
- [x] Ensure the new displayName property works as expected for default roles
- [x] Ensure if a custom role doesn't have a displayName property, it falls back to the id
- [x] Ensure if a custom role DOES have a displayName property, it uses that
- [x] Ensure a user is able to create a new custom role with and without a displayName
- [x] Ensure a user is not able to create a new custom role with the `gbDefault_` prefix

### DIFF
--- a/packages/front-end/components/Settings/RestoreConfigYamlButton.tsx
+++ b/packages/front-end/components/Settings/RestoreConfigYamlButton.tsx
@@ -272,6 +272,7 @@ export default function RestoreConfigYamlButton({
               if (!json || typeof json !== "object") {
                 throw new Error("Could not parsed yaml file into JSON object");
               }
+              // @ts-expect-error TS(2354) If you come across this, please fix it!: Argument of type 'object' is not assignable to parameter of type 'SetStateAction<null>'.
               setParsed(json);
               parseConfig(json);
             }}


packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -487,6 +489,98 @@ export async function putMemberRole(
...
+  if (!found) {
+    return res.status(404).json({
+      status: 404,
+      message: "Cannot find member",
+    });
+  }
+
+  try {
+    await updateOrganization(org.id, {
+      members: org.members,
```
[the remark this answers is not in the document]
mknowlton89: This is by design - on the Project Id page, when we render the MemberList, we only render active members, and no pending members..

packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -523,6 +540,40 @@ export const DEFAULT_ROLES: Record<DefaultMemberRole, Role> = {
...
+      "SavedGroupsFullAccess",
+      "TagsFullAccess",
+      "DataSourceConfiguration",
+      "TemplatesFullAccess",
+      "DecisionCriteriaFullAccess",
+      "HoldoutsFullAccess",
+      "GeneralDashboardsFullAccess",
+      "ProjectAdminAccess",
+    ],
+  },
```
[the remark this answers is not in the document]
mknowlton89: Addressed in this commit - https://github.com/growthbook/growthbook/pull/5131/changes/2058934c25a92815e538c580764f6ed8d281cf85

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
```diff
@@ -32,17 +38,23 @@ export default function SingleRoleSelector({
   const deactivatedRoles = organization.deactivatedRoles || [];
 
   const isNoAccessRoleEnabled = hasCommercialFeature("no-access-role");
+  const isProjectAdminRoleEnabled = hasCommercialFeature("project-admin-role");
```
jdorn, line 41: Should we also check this permission on the back-end? Currently it's possible for someone to bypass this check in the API call.

packages/front-end/components/Teams/Roles/RoleForm.tsx
```diff
@@ -44,17 +45,31 @@ export default function RoleForm({
...
+      return false;
+    }
+
     if (!/^[a-zA-Z0-9_]+$/.test(input.id)) {
       setError("Name can only contain letters, numbers, and underscores.");
       return false;
     }
+
+    if (input.displayName && input.displayName.length > 64) {
+      setError("Display name must be 100 characters or less.");
```
jdorn: Error message says 100, but actual limit is 64.  Also, should we add this to the back-end validation as well?

jdorn (approved):
Left some minor comments, but overall looks good

packages/front-end/components/Settings/Team/SingleRoleSelector.tsx
mknowlton89: Yea, good catch.

Added this to the `validateRoleAndEnvs` helper like we do for the `noaccess` role.

Also took the time to improve the error messages there a bit.

PR #5132, opened 2026-01-21
Fix baseline stddev using theta-adjusted statistic

itsgrimetime:
I think #4202 introduced a bug that changed the baseline response to use fresh statistics instead of the theta-adjusted `stat_a` from the test, causing inconsistent stddev values between baseline and treatment variations. I noticed this when comparing results from our very-out-of-date fork (july 2025) to recent upstream.

If this is expected, feel free to close. Comments on original PR make me think it was a known bug/workaround (and maybe intended to be fixed later?)

This fix captures stat_a from within the test loop (where it has already been processed through theta adjustment like stat_b), ensuring consistent statistics computation for both baseline and treatment.

### DIFF
--- a/packages/stats/gbstats/gbstats.py
+++ b/packages/stats/gbstats/gbstats.py
@@ -438,7 +438,7 @@ def run_mid_experiment_power(
         upperBoundAchieved=mid_experiment_power_result.upper_bound_achieved,
         scalingFactor=mid_experiment_power_result.scaling_factor,
     )
-    
+
 
 # Run A/B test analysis for each variation and dimension
 def analyze_metric_df(


packages/stats/gbstats/gbstats.py
```diff
@@ -548,17 +550,18 @@ def analyze_dimension(dimensionData: DimensionMetricData) -> DimensionResponse:
...
-        for _, row in d.iterrows():
-            control_stats.append(
-                variation_statistic_from_metric_row(row, "baseline", metric)
-            )
-        stats = list(zip(control_stats, control_stats))
-        stat_a_summed, _ = sum_stats(stats)
-        baseline_data = get_metric_response(d, stat_a_summed, 0)
+        # but should be the same for the baseline (stat_a is the control/baseline statistic)
+        if baseline_stat is None:
+            # Edge case: no treatment variations, compute baseline stat directly
```
lukebrawleysmith, line 555: Why would there be no treatment variations?  `analyze_metric_df` is called only in the context of an A/B test, right?

packages/stats/gbstats/gbstats.py
```diff
@@ -548,17 +550,18 @@ def analyze_dimension(dimensionData: DimensionMetricData) -> DimensionResponse:
...
+            control_stats = []
+            for _, row in d.iterrows():
+                control_stats.append(
+                    variation_statistic_from_metric_row(row, "baseline", metric)
+                )
+            stats = list(zip(control_stats, control_stats))
+            stat_a_summed, _ = sum_stats(stats)
+            baseline_stat = stat_a_summed
```
lukebrawleysmith, line 563: In this case, won't `theta` be `None`, while in the case above, `theta` is defined?

packages/stats/tests/test_gbstats.py
```diff
@@ -971,5 +973,260 @@ def test_process_analysis_denominator(self):
                 self.assertEqual(v.denominator, 510 if i == 0 else 500)
 
 
+# Test data for 3-armed test with CUPED
+THREE_ARMED_CUPED_DF = pd.DataFrame(
```
lukebrawleysmith, line 977: Do we need a 3-armed test?  I feel like a two-armed test and the existing `RA_STATISTICS_DF` object is enough for testing.  I understand we want to show that the baseline test from 0 vs 2 is used for calculating `theta`, but a single line comment in gbstats.py might be cleaner than adding a lot of new code to our unit tests.  

Feel free to ignore, this is not blocking, I just want to prevent adding unnecessary length to our unit test files.

packages/stats/gbstats/gbstats.py
lukesonnet: If the dataframe was missing rows; i'm not sure if we'd stop the analysis somewhere upstream but I don't think we do for sure.

packages/stats/gbstats/gbstats.py
lukesonnet: Yes, that's right. I'm not too concerned about this since this is covering an edge case and think it's reasonable to fall back to an unadjusted mean when not in the context of an a/b test.

packages/stats/tests/test_gbstats.py
lukesonnet: > unnecessary length to our unit test files.
I don't think this is a problem, and this forces future changes about what we return here to be explicit, so I do not want to remove it.

lukebrawleysmith (approved):
Thanks for addressing my comments

PR #5138, opened 2026-01-22
Fix unsafe scrubbing during payload generation 

bryce-fitzsimons:
Because SdkConnection payloads are generated in bulk now, some of the previously-safe scrubbing logic is now unsafe because it alters the base definitions for other SDK connections.

- Clone features and experiments before applying any scrubbing
- Clone holdouts and saved groups which may have been improperly mutated as well


packages/shared/src/sdk-versioning/sdk-payload.ts
```diff
@@ -97,40 +96,38 @@ export const scrubFeatures = (
     });
   }
 
-  const newFeatures = cloneDeep(features);
```
bryce-fitzsimons, line 100: cloning not needed here

PR #5142, opened 2026-01-23
JS SDKs: support case insensitive versions of array operators

bryce-fitzsimons:
For `$in`, `$nin`, `$all` operators, which apply to arrays of values, introduce case-insensitive equivalents:
`$ini`, `$nini`, `$alli`.

Note: Assuming we don't need a case-insensitive version of `$eq` or `$ne`.

(Inspired by the recent addition of `$regexi`)

### Features and Changes


packages/sdk-js/src/mongrule.ts
```diff
@@ -145,12 +149,43 @@ function elemMatch(actual: any, expected: any, savedGroups: SavedGroupsValues) {
...
 }
 
-function isIn(actual: any, expected: Array<any>): boolean {
+function isIn(
+  actual: any,
+  expected: Array<any>,
+  insensitive: boolean = false,
+): boolean {
+  const caseFold = (val: any) =>
+    insensitive && typeof val === "string" ? val.toLowerCase() : val;
```
jdorn: I think array.includes is highly optimized in JavaScript. Switching to array.some might be a performance hit. Should we only do that when insensitive is true?

bryce-fitzsimons: this is a good point, forked the logic for `insensitive` here.

PR #5147, opened 2026-01-23
fix(metric-group): Clean up metric groups when metric is deleted

ahdriel:
### Features and Changes

When a metric is deleted, we were not cleaning up references to it from Metric Groups. Now we are.

### DIFF
--- a/packages/back-end/src/models/MetricGroupModel.ts
+++ b/packages/back-end/src/models/MetricGroupModel.ts
@@ -47,13 +47,13 @@ export class MetricGroupModel extends BaseClass {
   }
 
   async removeMetricFromAllGroups(metricId: string): Promise<void> {
-    const metricGroups = await this.findByMetric(metricId);
-    await Promise.all(
-      metricGroups.map((group) =>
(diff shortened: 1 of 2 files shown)


packages/back-end/src/models/MetricGroupModel.ts
```diff
@@ -45,4 +45,15 @@ export class MetricGroupModel extends BaseClass {
       metrics: metricId,
     });
   }
+
+  async removeMetricFromAllGroups(metricId: string): Promise<void> {
+    await this._dangerousGetCollection().updateMany(
+      { organization: this.context.org.id, metrics: metricId },
+      {
+        // @ts-expect-error - not sure why $pull is complaining, but it works
+        $pull: { metrics: metricId },
```
lukesonnet, line 54: We use it other places... what error is it throwing?

ahdriel: This is the only one using `$pull` via BaseModel. The others are using mongoose.

PR #5148, opened 2026-01-23
fix: ensure reset of UI analysis settings on update

lukesonnet:
### Features and Changes

Only update the analysis settings and also update the analysis settings as displayed in the table dropdowns when a snapshot succeeds running in a browser.

https://www.loom.com/share/e388f2f503244c14bd6c71c96b09f400

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
@@ -1,6 +1,6 @@
 import { ExperimentInterfaceStringDates } from "shared/types/experiment";
 import { FactTableColumnType } from "shared/types/fact-table";
-import React, { useMemo, useState, useEffect } from "react";
+import React, { useMemo, useState, useEffect, useRef } from "react";
 import { OrganizationSettings } from "shared/types/organization";
 import { ExperimentSnapshotInterface } from "shared/types/experiment-snapshot";
 import { DifferenceType, StatsEngine } from "shared/types/stats";
@@ -74,7 +74,7 @@ export interface Props {
   setSliceTagsFilter?: (tags: string[]) => void;
   sortBy?: "significance" | "change" | "custom" | null;
   sortDirection?: "asc" | "desc" | null;
-  resetAnalysisSettingsOnUpdate?: () => void;
+  onSnapshotSuccessfulUpdate?: () => void;
 }
 
 const numberFormatter = Intl.NumberFormat();
@@ -100,7 +100,7 @@ export default function AnalysisSettingsSummary({
   setSliceTagsFilter,
   sortBy,
   sortDirection,
-  resetAnalysisSettingsOnUpdate,
+  onSnapshotSuccessfulUpdate,
 }: Props) {
   const {
     getDatasourceById,
@@ -138,7 +138,7 @@ export default function AnalysisSettingsSummary({
     analysis,
     dimension: _snapshotDimension,
     precomputedDimensions,
-    mutateSnapshot: originalMutateSnapshot,
+    mutateSnapshot,
     setAnalysisSettings,
     setSnapshotType,
     setDimension: setSnapshotDimension,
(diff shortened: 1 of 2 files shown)


packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -142,13 +138,29 @@ export default function AnalysisSettingsSummary({
...
-    mutateSnapshot,
+    mutateSnapshot: originalMutateSnapshot,
     setAnalysisSettings,
     setSnapshotType,
     setDimension: setSnapshotDimension,
     phase,
   } = useSnapshot();
 
+  // Track previous latest status to detect transition from "running" to "success"
+  const [previousLatestStatus, setPreviousLatestStatus] = useState<
```
ahdriel: nit: Given this is used just for a side-effect I think we should use `useRef`, otherwise we are re-rendering this when we don't need to.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -105,6 +100,7 @@ export default function AnalysisSettingsSummary({
   setSliceTagsFilter,
   sortBy,
   sortDirection,
+  resetAnalysisSettingsOnUpdate,
```
ahdriel: issue: This is named like a `boolean` prop but it is a function.

What do you think about `onSnapshotUpdate`?

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -142,13 +138,29 @@ export default function AnalysisSettingsSummary({
...
+  // Call reset when latest status transitions from "running" to "success"
+  useEffect(() => {
+    if (previousLatestStatus === "running" && latest?.status === "success") {
+      resetAnalysisSettingsOnUpdate?.();
+    }
+    setPreviousLatestStatus(latest?.status);
+  }, [latest?.status, previousLatestStatus, resetAnalysisSettingsOnUpdate]);
+
+  // Use the original mutateSnapshot - the useEffect above will handle the reset
+  const mutateSnapshot = originalMutateSnapshot;
```
ahdriel: question: I don't understand why we need this. Can't we just use `mutateSnapshot` from the props?

packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
```diff
@@ -151,6 +151,19 @@ export default function ResultsTab({
     dimension: analysisBarSettings.dimension,
   };
 
+  const resetAnalysisSettingsOnUpdate = useCallback(() => {
+    setAnalysisSettings(null);
+    setAnalysisBarSettings({
+      ...analysisBarSettings,
+      dimension: analysisBarSettings.dimension.startsWith("precomputed:")
+        ? ""
+        : (analysisBarSettings.dimension ?? ""),
```
ahdriel: `.dimension` is already a string here, so we don't need `?? ""`

packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
```diff
@@ -151,6 +151,19 @@ export default function ResultsTab({
     dimension: analysisBarSettings.dimension,
   };
 
+  const resetAnalysisSettingsOnUpdate = useCallback(() => {
+    setAnalysisSettings(null);
+    setAnalysisBarSettings({
```
ahdriel: from Claude:

Given we are overriding values, we can use 
```
setAnalysisBarSettings((prev) => ({
    ...
}))
```
and remove `analysisBarSettings` from the dependency array, making this more stable.

lukesonnet: 🫨

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
lukesonnet: Old pattern where i wrapped the callback, ty.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
lukesonnet: done, didn't know about useRef, thanks.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
lukesonnet: It makes it seem like it's going to do too much....

Changed it to `onSnapshotSuccessfulUpdate` anyways

PR #5150, opened 2026-01-25
Performance improvements v2: tighten project scoping for useFeaturesList()

bryce-fitzsimons:
Avoid fetching all features across all projects unless absolutely necessary

- scope `useFeaturesList()` calls to use either the project from the project selector or a specific entitity (feature/experiment)'s project whenever possible. Async/skip fetch support for when the entity project needs a bit of time to load.
- Offload prerequisite state and cycle checks to the backend, lazy-load features encountered during DAG evaluation
- create new `useFeaturesNames()` hook that just gets feature names, projects, and sometimes the defaultValue (for prereq calcs)
- remove `useFeaturesList()` when not needed (prerequisite modals, status displays, and input fields)
- for things like saved goup reference lists or feature flag dependent lists, scope to the current project but with an optional toggle to search for references globally.
- delete/archive modals that should be blocked by dependency checks must use global feature list. load those features only when trying to delete/archive instead of on all entity pages (block delete while loading data)
- lots and lots of UI cleanup

### DIFF
--- a/packages/back-end/src/controllers/features.ts
+++ b/packages/back-end/src/controllers/features.ts
@@ -3353,12 +3353,15 @@ export async function getPrerequisiteStates(
   const skipRootConditions = req.query.skipRootConditions === "true";
 
   const states: Record<string, PrerequisiteStateResult> = {};
+  const featuresMap = new Map<string, FeatureInterface>();
+  featuresMap.set(feature.id, feature);
+
   for (const env of envIds) {
     states[env] = await evaluatePrerequisiteStateAsync(
       context,
       feature,
       env,
-      undefined,
+      featuresMap,
       skipRootConditions,
     );
   }
@@ -3496,7 +3499,11 @@ export async function postBatchPrerequisiteStates(
       const visited = new Set<string>();
       const visiting = new Set<string>();
 
-      const visit = async (feature: FeatureInterface): Promise<boolean> => {
+      const visit = async (
+        feature: FeatureInterface,
+        depth: number = 0,
+      ): Promise<boolean> => {
+        if (depth >= PREREQUISITE_MAX_DEPTH) return true;
         if (visiting.has(feature.id)) return true;
         if (visited.has(feature.id)) return false;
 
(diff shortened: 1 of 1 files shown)


packages/front-end/hooks/usePrerequisiteStates.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  const key =
+    enabled && targetFeatureId && featureIds.length && environments.length
+      ? `${orgId}::/feature/${targetFeatureId}/batch-prerequisite-states|${featureIds
+          .slice()
+          .sort()
+          .join(
+            ",",
+          )}|${environments.slice().sort().join(",")}|${checkPrerequisite ? `checkPrereq:${checkPrerequisite.id}:${checkPrerequisite.prerequisiteIndex ?? -1}` : ""}|${checkRulePrerequisites ? `checkRule:${checkRulePrerequisites.environment}:${checkRulePrerequisites.ruleIndex}` : ""}`
+      : null;
```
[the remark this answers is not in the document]
bryce-fitzsimons: Kind of an edge case, but patched nonetheless.

packages/back-end/src/controllers/features.ts
```diff
@@ -3309,3 +3313,608 @@ export async function postCopyEnvironmentRules(
...
+
+  const skipRootConditions = req.query.skipRootConditions === "true";
+
+  const states: Record<string, PrerequisiteStateResult> = {};
+  for (const env of envIds) {
+    states[env] = await evaluatePrerequisiteStateAsync(
+      context,
+      feature,
+      env,
+      undefined,
```
jdorn: Minor, but should we pass in a shared features map here to avoid refetching the same features across environments?

packages/back-end/src/controllers/features.ts
```diff
@@ -3309,3 +3313,608 @@ export async function postCopyEnvironmentRules(
...
+    testRevision?: FeatureRevisionInterface | null,
+  ): Promise<[boolean, string | null]> => {
+    const cyclicCheckMap = new Map<string, FeatureInterface>();
+    cyclicCheckMap.set(targetFeature.id, targetFeature);
+    optionFeatures.forEach((f) => cyclicCheckMap.set(f.id, f));
+
+    const visited = new Set<string>();
+    const stack = new Set<string>();
+
+    const visit = async (
```
jdorn, line 3568: For both of these recursive calls, is it worth having a max depth sanity check?

PR #5155, opened 2026-01-26
Use pm2 on prod

tzjames:
### Features and Changes
Since pnpm switch it seems like we have a rather serious memory leak.  One theory is that the pnpm script runner , since it buffers the output from the servers, might be the cause of the leak. Previously we used wsrun.  Here we use pm2 to manage it.  It has paramters for `autorestart` and `max_memory_restart` that for now, to keep things consistent we have off.

### Testing

See preview env work.


Dockerfile
ahdriel: question: Is this what we should replace to avoid `pnpm` in the runtime path?

pm2 does have [pm2-runtime](https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/) that can be used here and we can point to the `ecosystem.config.js`.

ecosystem.config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  apps: [
+    {
+      name: "back-end",
+      script: "dist/server.js",
+      cwd: "./packages/back-end",
+      instances: 1,
+      autorestart: process.env.PM2_AUTORESTART === "true",
```
ahdriel, line 8: question: This makes the default to be `false` for autorestart. Do we want instead to make autorestart the default behavior?

start-front-end.js
```diff
@@ -0,0 +1,20 @@
...
+  "packages/front-end/node_modules/next/dist/bin/next",
+);
+const cwd = path.join(__dirname, "packages/front-end");
+
+const child = spawn("node", [nextBin, "start"], {
+  cwd: cwd,
+  stdio: "inherit",
+  env: process.env,
+});
+
```
ahdriel: question: Should we add a `child.on('error', ...)` handler as well? Even if just for logs.

ecosystem.config.js
tzjames: I think the better thing is for the health check to start failing once we go over 7GB (we have 8GB on prod).  Fargate will automatically stop sending requests there and kick off a new one.  Ideally there would be no lost requests.  For memory leaks that is almost certainly a better.  PM2 will kill the process I think and some requests would get lost in the few seconds before the new process is full spun up and ready. 

However the more common case is that we have a customer with a bunch of data, hitting an unoptimized endpoint and we go from 1.5GB to 8GB and then it OOMS and then it takes 5-10s before 2 consecutive health checks fails, and then the LB stops sending requests. In that case PM2 auto restarting is probably better.  The difference is probably just a few seconds though either way.

For now I left it with the current scenario. But I'm open to changing it, or at least changing it for us once we are certain that we have solved the more major memory leak.

PR #5159, opened 2026-01-26
feat(metric-drilldown): Enable for Unit Dimensions breakdown

ahdriel:
### Features and Changes

The initial implementation of Metric Drilldown in #5090, included it only for the main results, but not for `Unit dimensions` breakdowns.

Now we are also using it there.

### Testing instructions

- Go to the `Results` page/table
- Select a `Unit dimension` on top left of the table
- Update results if needed
- Observe how rows are clickable and open the `MetricDrilldownModal`
- Observe how data is correct and makes sense

### DIFF
--- a/packages/front-end/components/Experiment/Results.tsx
+++ b/packages/front-end/components/Experiment/Results.tsx
@@ -91,8 +91,7 @@ const Results: FC<{
       metricRow: ExperimentTableRow;
       initialTab?: MetricDrilldownTab;
       initialSliceSearchTerm?: string;
-      dimensionInfo?: { name: string; value: string };
-      dimensionIndex?: number;
+      dimensionInfo?: { name: string; value: string; index: number };
     } | null>(null);
 
   // todo: move to snapshot property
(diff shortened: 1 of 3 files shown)


packages/front-end/components/Experiment/BreakDownResults.tsx
```diff
@@ -288,8 +308,8 @@ const BreakDownResults: FC<{
                   }}
                 >
                   {label ? (
-                    label === "__NULL_DIMENSION" ? (
-                      <em>NULL (unset)</em>
+                    label === NULL_DIMENSION_VALUE ? (
```
msamper, line 311: From what I can tell this seems to be working correctly, but I'm a bit confused about the logic here. Why is it that if the label is `NULL_DIMENSION_VALUE` we `formatDimensionValueForDisplay`? I would think if there was a null dimension value there would be not dimension value to format

msamper (approved):
Tested and looks good to me. Just a question around some of the logic that I was a bit confused about. 

Unrelated, but I noticed when the drilldown modal opens the CI chart seems to slide into place. Doesn't seem to happen with the results tab from what I can tell so I wonder why that's happening.

packages/front-end/components/Experiment/BreakDownResults.tsx
ahdriel: This is just because we are adding a `em` around the NULL_DIMENSION_VALUE here, if that's the case.

`formatDimensionValueForDisplay` handles all values, but it just returns a string (`NULL (unset)` in this case), so here we do a weird dance to keep the same behavior of wrapping it in a `<em>`.

This is a good callout that this logic is not easy to read and we should refactor.

PR #5171, opened 2026-01-27
Better UI around custom fields

bryce-fitzsimons:
shared table, better formatting, etc

### DIFF
--- a/packages/front-end/styles/global.scss
+++ b/packages/front-end/styles/global.scss
@@ -1009,7 +1009,6 @@ main.main {
   td {
     vertical-align: middle;
   }
-
   &.table-valign-top {
     td,
     th {
@@ -1039,6 +1038,12 @@ main.main {
   }
   td {
     vertical-align: middle;
+  }
+  &.table-valign-top {
+    td,
+    th {
+      vertical-align: top;
+    }
   }
 }
 


packages/front-end/components/CustomFields/CustomFieldModal.tsx
```diff
@@ -101,6 +101,7 @@ export default function CustomFieldModal({
           // unset any placeholder value, as this is not applicable to boolean fields
           value.placeholder = "";
         }
+        const sectionValue = (value.section ?? section) as CustomFieldSection;
```
ahdriel: question: Do we need this cast? I think we can remove it

packages/front-end/components/CustomFields/SortableCustomField.tsx
```diff
@@ -21,141 +76,178 @@ export function SortableCustomFieldRow(props: SortableProps) {
...
       )}
+      <td
+        style={{
+          width: W.menu,
+          minWidth: W.menu,
+          padding: "0.75rem 0.5rem 0 0",
+          textAlign: "center",
+        }}
+      >
+        <CustomFieldRowMenu
```
ahdriel, line 156: issue: This is not centered vertically with the row

packages/front-end/components/CustomFields/SortableCustomField.tsx
```diff
@@ -3,14 +3,69 @@ import { CSS } from "@dnd-kit/utilities";
...
+  );
+}
+
+export const CUSTOM_FIELD_TABLE_WIDTHS = {
+  dragHandle: 30,
+  menu: 30,
+  name: "10%",
+  key: "14%",
+  description: "20%",
+  appliesTo: "8%",
```
ahdriel: suggestion: Should we increase this? Or maybe add a min-width for this column? In my default setup `Experiment` is being rendered as

```
Experim
ent
```

packages/front-end/components/CustomFields/CustomFields.tsx
```diff
@@ -15,99 +15,224 @@ import {
...
+  showAppliesTo: boolean;
+  showRequired: boolean;
+  colSpan: number;
+  deleteCustomField: (cf: CustomField) => Promise<void>;
+  setModalOpen: (v: Partial<CustomField> | null) => void;
+  handleMoveUp: (moveId: string, aboveId: string) => void;
+  handleMoveDown: (moveId: string, belowId: string) => void;
+  canManage: boolean;
+}) {
+  const W = CUSTOM_FIELD_TABLE_WIDTHS;
```
ahdriel, line 56: nit: should we do at least `WIDTH`?

packages/front-end/components/CustomFields/CustomFields.tsx
```diff
@@ -15,99 +15,224 @@ import {
...
+
+  const filteredItems = useMemo(() => {
+    if (filter === "all") return items;
+    return items.filter((cf) => cf.section === filter);
+  }, [items, filter]);
+
+  const openAddModal = () =>
+    setModalOpen(
+      filter === "all" ? {} : { section: filter as CustomFieldSection },
```
ahdriel, line 66: I am not sure the intent of this code, but a weird behavior is that I can select `Experiment` tab and when I click on `Add Custom Field` the default `Applies To` is still `Feature`.

Would be nice if they are synced.

packages/front-end/components/CustomFields/CustomFields.tsx
```diff
@@ -15,99 +15,224 @@ import {
...
   const sensors = useSensors(
     useSensor(MouseSensor, {}),
     useSensor(TouchSensor, {}),
     useSensor(KeyboardSensor, {}),
   );
 
-  function handleDragStart(event) {
-    setActiveId(event.active.id);
+  function handleDragStart(event: { active: { id: string } }) {
+    setActiveId(String(event.active.id));
```
ahdriel: question: Isn't event.active.id already a String? Do we need to wrap it in `String`?

packages/front-end/components/CustomFields/CustomFields.tsx
```diff
@@ -15,99 +15,224 @@ import {
...
-  title: string;
-}> = ({ section, title }) => {
+export type AppliesToFilter = "all" | CustomFieldSection;
+
+function CustomFieldsTable({
+  filter,
+  items,
+  showAppliesTo,
+  showRequired,
+  colSpan,
```
ahdriel: issue: If this table is responsible for rendering the columns, it should also be able to compute colSpan.

Any specific reason why we need to have this as a prop?

packages/front-end/components/CustomFields/CustomFields.tsx
```diff
@@ -15,99 +15,224 @@ import {
...
 import Button from "@/ui/Button";
+import { Tabs, TabsList, TabsTrigger, TabsContent } from "@/ui/Tabs";
 
-const CustomFields: FC<{
-  section: CustomFieldSection;
-  title: string;
-}> = ({ section, title }) => {
+export type AppliesToFilter = "all" | CustomFieldSection;
+
+function CustomFieldsTable({
```
ahdriel, line 35: suggestion: Should we create a new file for this table?

packages/front-end/components/CustomFields/SortableCustomField.tsx
```diff
@@ -21,141 +76,178 @@ export function SortableCustomFieldRow(props: SortableProps) {
...
-  const customField = props.customField;
-
-  const draggedRowStyle = { background: "rgba(127, 207, 250, 0.3)" };
   const handleStyle = {
     fontSize: 20,
     color: "var(--slate-a6)",
     cursor: `${isDragging ? "grabbing" : "grab"}`,
   };
 
+  const sectionLabel =
```
ahdriel: question: should this be inline if it is not being reused?

packages/front-end/components/CustomFields/SortableCustomField.tsx
ahdriel: Screenshot 2026-01-27 at 3 32 19 PM

bryce-fitzsimons: intentionally top-aligned because row height is variable. We do similarly in saved group listing tables

packages/front-end/components/CustomFields/CustomFields.tsx
bryce-fitzsimons: good callout

packages/front-end/components/CustomFields/CustomFields.tsx
bryce-fitzsimons: my personal pref is to keep the table as part of the page unless we need to reuse it somewhere else

packages/front-end/components/CustomFields/SortableCustomField.tsx
ahdriel: It looks bottom aligned?

PR #5173, opened 2026-01-28
refactor(results): Remove ResultsTableTooltip in favor of CI tooltip

ahdriel:
### Features and Changes

The current ResultsTableTooltip has a lot of duplicated information that we are already surfacing in the table and it takes a lot of space.

So we are removing it, in favor of smaller more targeted tooltips, including one to teach about Metric Drilldown if the mouse is not moved for 2 seconds.

### Specifics

- Added a new useCursorTooltip hook to allow us to easily create tooltips that are tied to the mouse positioning

### Testing

- Verify that the CI tooltip shows immediately when hovering over metric change values
- Verify that the row detail tooltip appears after 2 seconds of hovering over a row
- Verify that hovering over CI and waiting for 2 seconds won't show the metric detail tooltip

### DIFF
--- a/packages/front-end/components/Experiment/ExperimentResultTooltipContent/ExperimentResultTooltipContent.tsx
+++ b/packages/front-end/components/Experiment/ExperimentResultTooltipContent/ExperimentResultTooltipContent.tsx
@@ -66,7 +66,9 @@ export default function ExperimentResultTooltipContent({
 
   const getBadgeText = () => {
     if (significant) {
-      return resultsStatus === "won" ? "WON" : "LOST";
+      if (resultsStatus === "won") return "WON";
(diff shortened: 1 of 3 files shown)


packages/front-end/hooks/useCursorTooltip.tsx
```diff
@@ -0,0 +1,223 @@
...
+              width: 1,
+              height: 1,
+              pointerEvents: "none",
+            }}
+          />
+        </Tooltip>,
+        document.body,
+      );
+    },
+    [isVisible, cursorPos],
```
msamper: Curious.. does this need to be portaled? I though the tooltip was already portaled but maybe this serves a different purpose.

ahdriel:
cursor review

PR #5180, opened 2026-01-28
Add dateCreated index to features model

bryce-fitzsimons:
REST API uses date sorting/pagination for listFeatures. Adding index for faster querying


bryce-fitzsimons:
cursor review

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -159,6 +159,8 @@ const featureSchema = new mongoose.Schema({
 
 featureSchema.index({ id: 1, organization: 1 }, { unique: true });
 featureSchema.index({ organization: 1, project: 1 });
```
jdorn: Can probably remove this index since it will be fully covered by the one below

PR #5185, opened 2026-01-29
Support for Multi range namespaces.

madhuchavva:
### Namespaces to support multiple ranges

image

image

### Only new namespaces will take up "filters" format with `hashattribute`

Old Experiments:
```json
 {
          "id": "fr_41yrisirmgibsuw0",
          "coverage": 1,
          "hashAttribute": "id",
          "namespace": [
            "ns-41yrisirmgibmj14",
            0,
            0.35
          ],
          "seed": "40a0b987-133c-4209-a71d-eed2b23dd835",
          "hashVersion": 2,
          "variations": [
            false,
            true
          ],
          "weights": [
            0.5,
            0.5
          ],
          "key": "uber-drivers-campaign",
          "meta": [
            {
              "key": "0",
              "name": "Control"
            },
            {
              "key": "1",
              "name": "Variation 1"
            }
          ],
          "phase": "0",
          "name": "uber-drivers-campaign"
        },
```

New Experiments with multiple ranges:
```json

{
          "id": "fr_41yri4k3mgilc1ha",
          "coverage": 1,
          "hashAttribute": "id",
          "filters": [
            {
              "attribute": "id",
              "seed": "ns-41yriuppmgiisat0",
              "hashVersion": 2,
              "ranges": [
                [
                  0.2,
                  0.5
                ],
                [
                  0.8,
                  1
                ]
              ]
            }
          ],
          "seed": "4b7dd032-4408-4d6d-9b90-c5f1fc9ec352",
          "hashVersion": 2,
          "variations": [
            false,
            true
          ],
          "weights": [
            0.5,
            0.5
          ],
          "key": "uber-drivers-campaign-ns-split",
          "meta": [
            {
              "key": "0",
              "name": "Control"
            },
            {
              "key": "1",
              "name": "Variation 1"
            }
          ],
          "phase": "0",
          "name": "uber-drivers-campaign-ns-split"
        }

```

### New namespaces to be created with `hashattribute`:

image

### Experiments with "Fallback Attribute" can see Legacy Namespaces only.

image

### Group Namespaces matching Experiment's hash attribute.
image

### Messaging for choosing a namespace with different hash attribute
image


packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -1068,10 +1123,21 @@ export async function putNamespaces(
     throw new Error("Namespace not found.");
   }
 
+  if (!hashAttribute) {
+    throw new Error("Hash attribute is required");
+  }
```
jdorn: What happens if someone tries to update the description of a legacy namespace in the UI?  Does that switch it to be in the new namespace format?  If so, won't that mess up any running experiments that are using the namespace?

packages/back-end/src/services/experiments.ts
```diff
@@ -1563,7 +1563,12 @@ export async function toExperimentApiInterface(
       namespace: p.namespace?.enabled
         ? {
             namespaceId: p.namespace.name,
-            range: p.namespace.range,
+            range:
+              "range" in p.namespace
+                ? p.namespace.range
+                : "ranges" in p.namespace && p.namespace.ranges?.length
+                  ? p.namespace.ranges[0]
+                  : [0, 1],
```
jdorn: Is this the right default value here? If there is no range, should we do `[0, 0]` instead?

packages/back-end/src/services/features.ts
```diff
@@ -335,26 +346,58 @@ export function generateAutoExperimentsPayload({
...
-                attribute: e.hashAttribute,
-                seed: phase.namespace.name,
-                hashVersion: 2,
-                ranges: [phase.namespace.range],
-              },
-            ]
-          : [],
-        seed: phase.seed,
-        name: e.name,
-        phase: `${e.phases.length - 1}`,
```
jdorn, line 434: I think these 3 fields were accidentally deleted?

packages/back-end/src/services/features.ts
```diff
@@ -335,26 +346,58 @@ export function generateAutoExperimentsPayload({
...
+        if (nsDefinition?.format === "multiRange") {
+          const ns = phase.namespace;
+          let ranges: [number, number][];
+
+          if ("ranges" in ns && ns.ranges) {
+            ranges = ns.ranges;
+          } else if ("range" in ns) {
+            ranges = [ns.range];
+          } else {
+            ranges = [[0, 1]];
```
jdorn: Same here. For default value, should we do `[0, 0]` to be safe?

packages/back-end/src/services/features.ts
```diff
@@ -335,26 +346,58 @@ export function generateAutoExperimentsPayload({
...
+              attribute:
+                "hashAttribute" in ns && ns.hashAttribute
+                  ? ns.hashAttribute
+                  : nsDefinition.hashAttribute!,
+              seed: nsDefinition.seed || ns.name,
+              hashVersion: 2,
+              ranges,
+            },
+          ];
+        } else {
```
jdorn: The old code here was using `filters`, so we should continue to use that for legacy namespaces.  We need the payload to be 100% identical after this change for existing namespacecs.

packages/back-end/src/services/features.ts
```diff
@@ -335,26 +346,58 @@ export function generateAutoExperimentsPayload({
...
+          } else {
+            ranges = [[0, 1]];
+          }
+
+          exp.filters = [
+            {
+              attribute:
+                "hashAttribute" in ns && ns.hashAttribute
+                  ? ns.hashAttribute
+                  : nsDefinition.hashAttribute!,
```
jdorn: Don't use `!` like this.  If typescript is saying this might be undefined, we need to handle that instead of ignoring the warning. Or, if the type is incorrect and it really should not be undefined, we should fix the type.

packages/back-end/src/services/features.ts
```diff
@@ -1249,6 +1319,8 @@ export function evaluateFeature({
         revision,
         date,
         safeRolloutMap,
+        holdoutsMap: new Map(),
+        namespaces: new Map(),
```
jdorn: For this and the places below where an empty map is passed in, the payload will be incorrect, won't it?  The code will fall back to assuming everything is a legacy namespace range, so the seed, hash attribute, and multiple ranges will all be missing.

This particular usage here is not critical since it's only for the "simulate" functionality in the app as of now, but it's still a little risky to leave it off.

packages/front-end/components/Experiment/TabbedPage/TargetingInfo.tsx
```diff
@@ -50,12 +50,38 @@ export default function TargetingInfo({
...
+        const ns = phase.namespace!;
+        if ("ranges" in ns && ns.ranges && ns.ranges.length > 0) {
+          return ns.ranges.reduce(
+            (sum, [start, end]) => sum + (end - start),
+            0,
+          );
+        } else if ("range" in ns && ns.range) {
+          return ns.range[1] - ns.range[0];
+        }
+        return 1;
```
jdorn: Should default be 0 for all of these?

packages/front-end/components/Experiment/ExpandablePhaseSummary.tsx
```diff
@@ -21,8 +21,20 @@ export default function ExpandablePhaseSummary({ i, phase, editPhase }: Props) {
...
+        if ("ranges" in ns && ns.ranges && ns.ranges.length > 0) {
+          return ns.ranges.reduce(
+            (sum, [start, end]) => sum + (end - start),
+            0,
+          );
+        } else if ("range" in ns && ns.range) {
+          return ns.range[1] - ns.range[0];
+        }
+        return 1;
+      })()
```
jdorn: This exact logic is used multiple times in multiple files.  Should we extract this to a utility function?

packages/front-end/components/Experiment/NamespaceModal.tsx
```diff
@@ -17,23 +26,45 @@ export default function NamespaceModal({
...
+
+  const form = useForm<NamespaceFormValue>({
     defaultValues: {
       label: existingNamespace?.label || existingNamespace?.name || "",
       description: existingNamespace?.description || "",
       status: existingNamespace?.status || "active",
+      hashAttribute:
+        existingNamespace?.hashAttribute ||
+        attributes.find((a) => a.hashAttribute)?.property ||
+        "id",
```
jdorn, line 50: For backwards compatibility, we may want to have 2 separate modals - one for legacy and one for new.  On the front-end, decide which modal to show based on the namespace format.

packages/shared/types/organization.d.ts
```diff
@@ -160,6 +160,12 @@ export interface Namespaces {
   label: string;
   description: string;
   status: "active" | "inactive";
+  // Hash attribute for namespace allocation (uses v2 hashing algorithm)
+  hashAttribute: string;
+  // Seed for hashing (UUIDv4)
+  seed: string;
+  // Flag to differentiate between namespace formats: legacy (single range) or multiRange (multiple ranges with own hashAttribute)
+  format?: "legacy" | "multiRange";
 }
```
jdorn, line 172: It would simplify things if we use a discriminated union here.

```ts
interface LegacyNamespace {
  format: "legacy";
  name: string;
  label: string;
  description: string;
  status: "active" | "inactive";
}

interface MultiRangeNamespace {
  format: "multiRange";
  name: string;
  label: string;
  hashAttribute: string;
  seed: string;
  description: string;
  status: "active" | "inactive";
}

export type Namespaces = LegacyNamespace | MultiRangeNamespace;
```

Then, throughout the code, we can just check the format and we'll know the exact shape of the namespace and what fields are present or not.  With the way the code is currently, we need fallbacks all throughout the code for whether or not `seed` or `hashAttribute` are empty.

packages/shared/src/validators/shared.ts
```diff
@@ -1,12 +1,27 @@
...
+  range: z.tuple([z.number(), z.number()]),
+});
+
+// MultiRange format (multiple ranges, own hashAttribute, and hashVersion)
+// Note: hashVersion defaults to 2 if not provided in application logic
+const multiRangeNamespaceValue = z.object({
+  enabled: z.boolean(),
+  name: z.string(),
+  ranges: z.array(z.tuple([z.number(), z.number()])),
+  hashAttribute: z.string().optional(), // If not provided, inherits from experiment
```
jdorn: I don't think we should have a `hashAttribute` field at all here.  It should always use the attribute from the namespace definition no matter what.  Same with `hashVersion`

packages/back-end/src/services/features.ts
```diff
@@ -131,6 +132,10 @@ export function generateFeaturesPayload({
     string,
     { holdout: HoldoutInterface; experiment: ExperimentInterface }
   >;
+  namespaces?: Map<
+    string,
+    { hashAttribute?: string; seed?: string; format?: "legacy" | "multiRange" }
```
bryce-fitzsimons: better to provide a type for `"legacy" | "multiRange"`

packages/front-end/components/Experiment/ReleaseChangesForm.tsx
```diff
@@ -596,12 +596,26 @@ function getRecommendedRolloutData({
...
-    if (
-      namespaceRange[0] > lastNamespaceRange[0] ||
-      namespaceRange[1] < lastNamespaceRange[1]
-    ) {
+    // Get total range coverage
+    const getTotalCoverage = (ns: {
+      ranges?: [number, number][];
+      range?: [number, number];
+    }) => {
+      if ("ranges" in ns && ns.ranges && ns.ranges.length > 0) {
```
bryce-fitzsimons: duplicate code

packages/front-end/components/Features/NamespaceSelector.tsx
```diff
@@ -80,51 +136,100 @@ export default function NamespaceSelector({
...
+                        onChange={(e) => {
+                          updateRange(
+                            index,
+                            0,
+                            parseFloat(e.target.value) || 0,
+                          );
+                        }}
+                      />
+                    </div>
+                    <div className="col-auto">to</div>
```
bryce-fitzsimons: avoid bootstrap styling, favor @ui/ or @radix-ui

packages/front-end/components/Features/NamespaceSelector.tsx
```diff
@@ -55,76 +174,153 @@ export default function NamespaceSelector({
...
+              {selectedNamespace &&
+                "hashAttribute" in selectedNamespace &&
+                selectedNamespace.hashAttribute && (
+                  <div className="alert alert-info mb-3">
+                    <strong>Hash Attribute:</strong>{" "}
+                    {`${selectedNamespace.hashAttribute}`}
                   </div>
+                )}
+              {selectedIsDifferentHash && (
+                <div className="alert alert-info mb-3">
```
bryce-fitzsimons: I think this should perhaps be a `<Callout status="warning" ...>`. We generally don't want people choosing the wrong hash, right?

packages/front-end/components/Features/NamespaceSelector.tsx
```diff
@@ -55,76 +174,153 @@ export default function NamespaceSelector({
...
-                      {...form.register(`${formPrefix}namespace.range.1`, {
-                        valueAsNumber: true,
-                      })}
-                    />
+          {namespace && selectedNamespace && (
+            <div className="mt-3">
+              {selectedNamespace &&
+                "hashAttribute" in selectedNamespace &&
+                selectedNamespace.hashAttribute && (
+                  <div className="alert alert-info mb-3">
```
bryce-fitzsimons: prefer radix components like `<Callout>`

packages/front-end/components/Features/NamespaceSelector.tsx
```diff
@@ -30,12 +40,121 @@ export default function NamespaceSelector({
...
+    ? filteredNamespaces.map((n) => ({ value: n.name, label: n.label }))
+    : [
+        ...matchingNamespaces.map((n) => ({
+          value: n.name,
+          label: n.label,
+        })),
+        ...(differentHashNamespaces.length
+          ? [
+              {
+                value: separatorOptionValue,
```
bryce-fitzsimons: Better to use GroupedValue for the options rather than custom UI for separators. It's basically this:
```
[
  {
    label: "Group label",
    options: [ {value, label}, {value, label} ]
  },
  {
    label: "Other options",
    options: [ {value, label}, {value, label} ]
  }
]
```

packages/front-end/components/Features/NamespaceSelector.tsx
```diff
@@ -55,76 +174,153 @@ export default function NamespaceSelector({
...
             }}
             placeholder="Choose a namespace..."
-            options={namespaces
-              .filter((n) => {
-                return n?.status !== "inactive";
-              })
-              .map((n) => ({ value: n.name, label: n.label }))}
+            options={namespaceOptions}
+            sort={false}
+            isOptionDisabled={(option) => {
```
bryce-fitzsimons: let's not do these custom formatters unless we have to

packages/back-end/src/util/features.ts
```diff
@@ -315,6 +319,51 @@ export function getHoldoutFeatureDefId(holdoutId: string) {
   return `$holdout:${holdoutId}`;
 }
 
+/**
+ * Helper function to apply namespace to a rule
+ * Handles both multiRange format (with hashAttribute and multiple ranges) and legacy format
+ */
+export function applyNamespaceToRule(
```
bryce-fitzsimons: does this also apply to AutoExperiments (the top-level `experiment: []` field on the sdk payload)? If so, maybe rename. If not, then is this a regression?

packages/front-end/components/Features/NamespaceSelector.tsx
madhuchavva: No. Jeremy mentioned that it was okay to still choose a namesapce with different hash attr and not turn this into a warn or err. We group the namespaces by the matching hash attr and push the ones matching to the top just to set a preference over choosing the others.

bryce-fitzsimons (approved):

PR #5192, opened 2026-01-30
fix behavior of dimension selector

lukesonnet:
### Features and Changes

Fix pre-defined dimensions to await the successful computation of a new analysis. Previously it ran the analysis but needed a re-render to get the completed analysis.

Fix updater to run and reset to base result when pre-defined dimension is selected (ideally it would just also compute the pre-defined dimension update, can fast follow with that). Previously the update button did nothing.

Reduced "flicker"/load time on other selectors.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -645,19 +645,19 @@ export async function parseDimension(
     if (dimension.match(/^exp:/)) {
       return {
         type: "experiment",
-        id: dimension.substr(4),
+        id: dimension.substring(4),
         specifiedSlices: slices,
       };
     } else if (isPrecomputedDimension(dimension)) {
       return {
         type: "experiment",
-        id: dimension.substr(PRECOMPUTED_DIMENSION_PREFIX.length),
+        id: dimension.substring(PRECOMPUTED_DIMENSION_PREFIX.length),
         specifiedSlices: slices,
       };
-    } else if (dimension.substr(0, 4) === "pre:") {
+    } else if (dimension.substring(0, 4) === "pre:") {
       return {
         // eslint-disable-next-line
-        type: dimension.substr(4) as any,
+        type: dimension.substring(4) as any,
       };
     } else {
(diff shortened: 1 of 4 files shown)


packages/front-end/components/Dimensions/DimensionChooser.tsx
```diff
@@ -247,35 +247,35 @@ export default function DimensionChooser({
           dimensions: [v],
         };
 
-        triggerAnalysisUpdate(
+        // check if the analysis exists in the current snapshot
+        const analysisExistsInMainSnapshot = snapshot
+          ? getSnapshotAnalysis(snapshot, newSettings)
```
lukesonnet: We have to check current snapshot, not standard snapshot, to see if the analysis exists and then we can skip the mutation below, which causes an unnecessary slowdown in loading the results.

The other choosers can just check the main snapshot, but because this selector can actually switch snapshots and adds an analysis to the `standardSnapshot`, we have to check the loaded `snapshot`.

packages/back-end/src/services/experiments.ts
```diff
@@ -646,10 +648,10 @@ export async function parseDimension(
         id: dimension.substr(4),
         specifiedSlices: slices,
       };
-    } else if (dimension.match(/^precomputed:/)) {
+    } else if (isPrecomputedDimension(dimension)) {
       return {
         type: "experiment",
-        id: dimension.substr(12),
+        id: dimension.substr(PRECOMPUTED_DIMENSION_PREFIX.length),
```
ahdriel: nit: Can we update this to `substring`?

`substr` is deprecated.

packages/front-end/components/Dimensions/DimensionChooser.tsx
ahdriel: Should this be `getSnapshotAnalysis() != null`?

This way we can force `analysisExistsInMainSnapshot` to be just a boolean instead of `false | ExperimentSnapshotAnalysis | null` as it currently is.

PR #5198, opened 2026-01-30
Remove risk from GrowthBook FE

lukesonnet:
### Features and Changes

Risk, while a valuable concept, was adding visual clutter and complexity to a variety of surfaces in the app. This removes risk from the front end and (mostly) from the docs, deprecates it in the REST API, but leaves the computation in place in case we choose to restore it or use it in future. Alternative solutions: add it to the new Metric Drilldown, add a popover and keep the tooltip, were not favored as it simply continued to bury an unused feature.

As an experimentation platform, we will plan to build future tools to answer the question of "is this risky" using our experiment decision framework and non-inferiority tests, which more directly tackle the question.

We also remove the "guardrail warning" which was a flag for whenever guardrails were negative and not statsig. This was not a helpful message and was far too noisy.

### Testing

Check rendering + errors on:

- [X] Results table
- [x] Custom reports results table
- [x] Report configuration
- [X] Metric overrides
- [X] fmid
- [x] Fact metric form
- [x] mid
- [x] Metric form
- [x] API docs

### Screenshots

https://www.loom.com/share/b486b1ab20eb4b3da53ccbd2b9d2c464


packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -272,7 +270,7 @@ const AnalysisForm: FC<{
       open={true}
       close={cancel}
       size="lg"
-      ctaEnabled={!editMetrics || !hasMetricOverrideRiskError}
+      ctaEnabled={!editMetrics}
```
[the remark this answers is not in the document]
lukesonnet: Oof I caught this and then forgot about it. Fixed.

packages/front-end/components/Experiment/MetricsOverridesSelector.tsx
```diff
@@ -288,10 +229,15 @@ export default function MetricsOverridesSelector({
...
-                            },
+                            ...(metricDefinition &&
+                            !isRetentionMetric(metricDefinition)
+                              ? [
+                                  {
+                                    label: "Lookback",
+                                    value: "lookback",
+                                  },
+                                ]
+                              : []),
```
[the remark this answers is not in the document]
lukesonnet: Not fixing this to keep it in line with the other code. Tested and works well, and I don't care about lookback windows when metric definition is missing.

docs/docs/experimentation-analysis/experiment-results.mdx
```diff
@@ -29,12 +29,6 @@ For both columns, the raw data in grey underneath shows the numerator and denomi
 
 **Chance to Win** tells you the probability that the variation is better. Anything above 95% (a customizable threshold set at the organization level) is highlighted green indicating a very clear winner. Anything below 5% is highlighted red, indicating a very clear loser. Anything in between is grayed out indicating it's inconclusive. If that's the case, there's either no measurable difference or you haven't gathered enough data yet.
 
-Furthermore, we include tooltips for **Risk**, which captures the average loss in a metric if the variation were actually to be worse than control. We use this tooltip to let you know about "risky" cases. For example, in cases like in the screenshot above, the chance to win is above 50% for a few metrics, but the average loss if the baseline was actually better is above the risk threshold, indicating that shipping the variation still has substantial risk. More information about risk is available in a tooltip when you mouseover the result:
-
-<MaxWidthImage maxWidth={500} border>
-  ![Results Tooltip](/images/results-tooltip.png)
```
ahdriel, line 35: nit: Can we remove the images/results-tooltip image as well?

packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -272,7 +270,7 @@ const AnalysisForm: FC<{
       open={true}
       close={cancel}
       size="lg"
-      ctaEnabled={!editMetrics || !hasMetricOverrideRiskError}
+      ctaEnabled={true}
```
ahdriel: nit: This is already the default, so we could remove it.

packages/front-end/components/Experiment/EditMetricsForm.tsx
```diff
@@ -196,7 +194,7 @@ const EditMetricsForm: FC<{
       size="lg"
       open={true}
       close={cancel}
-      ctaEnabled={!hasMetricOverrideRiskError}
+      ctaEnabled={true}
```
ahdriel: nit: Same here as `ctaEnabled` `true` is the default.

packages/front-end/components/Experiment/MetricsOverridesSelector.tsx
```diff
@@ -288,10 +229,15 @@ export default function MetricsOverridesSelector({
                               label: "Conversion",
                               value: "conversion",
                             },
-                            {
-                              label: "Lookback",
-                              value: "lookback",
-                            },
+                            ...(metricDefinition &&
```
ahdriel, line 232: question: Why is this needed?

packages/front-end/components/Experiment/MetricsOverridesSelector.tsx
```diff
@@ -304,223 +250,163 @@ export default function MetricsOverridesSelector({
...
+                          "conversion" ||
+                        (metricDefinition &&
+                          isRetentionMetric(metricDefinition)) ? (
+                          <div className="row mt-2">
+                            <div className="col">
+                              <Field
+                                label={
+                                  metricDefinition &&
+                                  isRetentionMetric(metricDefinition)
+                                    ? "Retention Starts After (hours)"
```
ahdriel, line 265: This is much clearer! nice

packages/front-end/components/Report/ConfigureReport.tsx
```diff
@@ -173,7 +171,7 @@ export default function ConfigureReport({
       header={`Edit Analysis`}
       useRadixButton={true}
       cta="Save and refresh"
-      ctaEnabled={!hasMetricOverrideRiskError}
+      ctaEnabled={true}
```
ahdriel: nit: same

ahdriel (approved):

packages/front-end/components/Experiment/MetricsOverridesSelector.tsx
lukesonnet: This line? Because the metric could be missing if you have an existing override that maps to a metric that no longer exists. It's kind of harmless and does nothing unless you add the metric back to the experiment.

The next line because retention metrics can't have lookback windows.

docs/docs/experimentation-analysis/experiment-results.mdx
lukesonnet: oh yeah thanks

PR #5200, opened 2026-01-30
[Draft] Product Analytics Explorer UI

oelshaikh:
### Features and Changes


packages/shared/src/validators/product-analytics.ts
```diff
@@ -0,0 +1,184 @@
+import { z } from "zod";
+import { rowFilterValidator } from "./fact-table";
+
+const baseValueValidator = z.object({
+  name: z.string(),
+  color: z.string().nullable(),
```
mknowlton89: I'm not sure the color belongs at this level.

We may eventually want to introduce this, but I don't think we should add it initially. While we haven't pushed the dashboard color PR, we want the colors to be scoped at a dashboard level, not a chart level, so if a user changes the color for Chrome on one chart, it updates for other charts that have a `Chrome` series.

packages/shared/src/validators/product-analytics.ts
```diff
@@ -0,0 +1,184 @@
...
+  tag: z.string().nullable(),
```
mknowlton89: Curious to hear about the idea for the `tag` on the value?

packages/front-end/enterprise/components/ProductAnalytics/Explorer.tsx
```diff
@@ -0,0 +1,52 @@
...
+          padding: "var(--space-3)",
+        }}
+      >
+        <MetricExplorerSettings />
+      </Box>
+    </Flex>
+  );
+}
+
+export default function MetricExplorer() {
```
mknowlton89, line 35: I'd prefer this to take the same name as the file. Maybe `ProductAnalyticsExplorer`?

packages/shared/src/validators/product-analytics.ts
oelshaikh: Agreed, initially I had color and tag tied to each series/value but realized that it breaks down once we start introducting group bys. I removed the concept of persistent colors and tags in my local changes and now they're just determined at charting time (will push soon!)

packages/front-end/enterprise/components/ProductAnalytics/SideBar/ExplorerSideBar.tsx
```diff
@@ -0,0 +1,169 @@
...
+            )}
+          </Flex>
+        </Flex>
+
+        <Box px="0" mt="3" style={{
+          border: "1px solid var(--gray-a3)",
+          borderRadius: "var(--radius-4)",
+          backgroundColor: "var(--color-panel-translucent)",
+        }}>
+          <TabsContent value="metric">
```
mknowlton89, line 107: I like this approach, but I feel like it might result in a lot of duplicated (or nearly duplicated) code.

E.G. All tabs will support the `Values` section and the `Group By` section.

We may want the tab content here to only include the content that is unique for a particular type.

E.G. The SqlTabContent will have the Data Source selector, and the section where the user can view the sql, click the cta to open the modal, and include the timestamp selector.

While the fact table section will have the fact table selector, and so on.

packages/shared/src/validators/product-analytics.ts
oelshaikh: Tag was just a short auto-generated identifier like "A", "B" so that the user could quickly find which value/series relates to which line or bar on the chart, but it also breaks down when we start adding dimensions so I removed tags in my latest changes (yet to be pushed).

packages/front-end/enterprise/components/ProductAnalytics/ExplorerContext.tsx
```diff
@@ -0,0 +1,205 @@
...
+export interface ExplorerContextValue {
+  // ─── State ─────────────────────────────────────────────────────────────
+  draftExploreState: ProductAnalyticsConfig;
+  submittedExploreState: ProductAnalyticsConfig | null;
+  hasPendingChanges: boolean;
+  exploreData: ProductAnalyticsResult | null;
+  loading: boolean;
+
+  // ─── Modifiers ─────────────────────────────────────────────────────────
+  setDraftExploreState: React.Dispatch<
```
mknowlton89, line 49: I really like this pattern - while its a bit more verbose than just having a state variable and exposing the `setState` call to all of the children, I think we'll run into way fewer issues as we expand this since we won't have to write a bunch of complicated one-off `setState()` logic

PR #5206, opened 2026-01-31
feat(results): Reinstate suspicious, not enough data, draw tooltip explainers

ahdriel:
### Features and Changes

With the removal of AnalysisSummaryTooltip in #5173 some of the information available in the tooltip was removed. We are reinstating the following ones:
- Suspicious change
- Draw (significant but below minChange)
- Not enough data

#### Screenshots

Screenshot 2026-02-02 at 1 51 34 PM
Screenshot 2026-02-02 at 1 51 21 PM
Screenshot 2026-02-02 at 1 51 14 PM
Screenshot 2026-02-02 at 7 08 15 PM

### DIFF
--- a/packages/front-end/components/Experiment/DrilldownTooltip.tsx
+++ b/packages/front-end/components/Experiment/DrilldownTooltip.tsx
@@ -54,7 +54,7 @@ export function DrilldownTooltip({ enabled, children }: DrilldownTooltipProps) {
     <>
       {children({
         onMouseMove,
-        onMouseLeave: () => triggerProps.onMouseLeave({} as React.MouseEvent),
+        onMouseLeave: close,
         onClick: close,
       })}
       {renderTooltip(
--- a/packages/front-end/hooks/useHoverTooltip.tsx
+++ b/packages/front-end/hooks/useHoverTooltip.tsx
@@ -478,10 +478,12 @@ export function HoverTooltip({
   content,
   delayMs,
   positioning,
+  enabled,
 }: HoverTooltipProps) {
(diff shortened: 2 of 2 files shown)


packages/front-end/components/Experiment/ChangeColumn.tsx
```diff
@@ -89,6 +94,12 @@ export default function ChangeColumn({
         differenceType,
         statsEngine,
         ssrPolyfills,
+        suspiciousChange: rowResults.suspiciousChange,
+        suspiciousThreshold: rowResults.suspiciousThreshold,
+        notEnoughData: !rowResults.enoughData,
+        minSampleSize,
+        minPercentChange: rowResults.minPercentChange,
+        currentMetricTotal: stats?.value ?? 0,
```
lukesonnet: Should we return this from `rowResults` as well since we compute all the other logic in there from this value?

packages/front-end/components/Experiment/useColumnStatusPopovers.tsx
```diff
@@ -0,0 +1,102 @@
...
+  differenceType,
+  statsEngine,
+  ssrPolyfills,
+  minSampleSize = 0,
+  showSuspicious = true,
+}: UseColumnStatusPopoversOptions) {
+  const popoverEnabled = !!(metric && differenceType && statsEngine);
+
+  // Get the max numerator value across baseline and variation
+  const currentMetricTotal = Math.max(baseline?.value ?? 0, stats?.value ?? 0);
```
lukesonnet: yeah, more evidence we should just pass forward the existing computation from rowResults?

PR #5211, opened 2026-02-02
Upgrade pip in dockerfile

Kevin-Chant:
### Features and Changes

Pip 24 has a security vulnerability, so we should prompt it to upgrade when building our docker image

### Testing

`docker build -t growthbook . --no-cache`

Note: My local env fails the build on both `main` and this branch, but I think the error is due to my docker version given the comment in the dockerfile about the wildcard. The actual build steps before this including the pip upgrade work fine
```
Step 41/52 : COPY buildinfo* ./buildinfo
COPY failed: no source files were specified
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,6 +84,7 @@ RUN apt-get update && \
   npm install -g pnpm@9.15.0 && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
+RUN pip install --upgrade pip
 COPY --from=pybuild /usr/local/src/app/requirements.txt /usr/local/src/requirements.txt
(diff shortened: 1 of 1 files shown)


tzjames:
This is in the wrong place.  Or at least it needs to be in the final image as well.  I've made the change and tested and I do now see: 
```
james@Jamess-MBP growthbook %  docker run --rm growthbook pip --version        
pip 26.0 from /usr/local/lib/python3.11/site-packages/pip (python 3.11)
```
So I'll approve, but I'll let you land.
I'm not sure what is going on that you can't build locally.  FWIW my docker version is:  `Docker version 24.0.2, build cb74dfc`

PR #5214, opened 2026-02-02
Recalculate nextUpdate after schedule change

mknowlton89:
### Features and Changes

Previously, if a user updated the `updateSchedule` for a dashboard, we'd update the dashboard's `updateSchedule` object correctly, but we weren't recalculating the `nextUpdate` field. Which means that the `updateSchedule` changes wouldn't actually change the update schedule until the next auto-update ran.

Now, whenever a user hits the `updateDashboard` endpoint and something about the auto-updates has changed, we're recalculating the `nextUpdate` date, so the changes are respected immediately.

One challenge that this introduced was with how we're calling the `updateDashboard` endpoint from the front-end - in those cases, the `updates` object includes existing, non-dirty fields. Which means if a user opened the `Edit Dashboard Modal` and just changed the dashboard name, the `updates` object still contained properties like `updateSchedule` and `enableAutoUpdates`.

To ensure that we're not re-calculating the nextUpdate based on non-schedule changes like dashboard name changes, we've introduced a new `shouldRecalculateNextUpdate` helper function that checks if the updates are actually changing any of the auto updating fields.

### Dependencies

- None

#### Basic Functionality Tests
- [x] Change auto-update schedule for dashboard with existing `nextUpdate` and verify `nextUpdate` is recalculated correctly
- [x] Change schedule type from "stale" to "cron" with auto-updates enabled and verify `nextUpdate` is recalculated based on cron expression
- [x] Disable auto-updates for an enabled dashboard and verify `nextUpdate` becomes null
- [x] Disable auto-updates and change schedule at the same time and verify `nextUpdate` becomes null (disabled takes precedence)
- [x] Update other dashboard fields (like title) without changing schedule and verify `nextUpdate` remains unchanged

#### Auto-Updates Disabled Tests
- [x] Enable auto-updates and change schedule at the same time and verify `nextUpdate` is calculated

#### Missing nextUpdate Tests
- [x] Change schedule for dashboard with auto-updates enabled but no existing `nextUpdate` and verify `nextUpdate` is calculated
- [x] Disable auto-updates for dashboard with no existing `nextUpdate` and verify `nextUpdate` remains unset

#### Edge Cases
- [x] Enable auto-updates and set schedule for dashboard that had neither and verify `nextUpdate` is calculated

### DIFF
--- a/docs/src/styles/custom.scss
+++ b/docs/src/styles/custom.scss
@@ -275,6 +275,28 @@
   display: none;
 }
 
+// Custom widths — shared max-width and centering so navbar and main content align
+$content-max-width: 92rem;
+
+.navbar__inner,
+.main-wrapper {
+  margin-inline: auto;
+  max-width: $content-max-width;
+  width: 100%;
+}
+
+@media (min-width: $content-max-width) {
+  .navbar {
+    padding-inline: max(0px, (100vw - $content-max-width) / 2);
+  }
+
+  .navbar__inner {
+    padding-inline-start: calc(
+      var(--ifm-menu-link-padding-horizontal) + 0.5rem
+    );
+  }
+}
+
 [class^="docItemContainer"] {
   max-width: 688px;
 }
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -694,11 +694,6 @@
       "count": 1
     }
   },
-  "packages/front-end/components/SavedGroups/SavedGroupReferencesList.tsx": {
-    "no-restricted-imports": {
-      "count": 1
-    }
-  },
   "packages/front-end/components/SchemaBrowser/EditSqlModal.tsx": {
     "no-restricted-imports": {
       "count": 1
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -1009,7 +1009,7 @@ export default abstract class SqlIntegration
           -- Get aggregated metric per user by day
           SELECT
           ${populationSQL ? "p" : "f"}.${baseIdType} AS ${baseIdType}
-            , ${this.dateTrunc("timestamp")} AS date
+            , ${this.dateTrunc("f.timestamp")} AS date
             , ${metricData.numeratorAggFns.fullAggregationFunction(`f.${metricData.alias}_value`)} AS value
             , ${metricData.numeratorAggFns.partialAggregationFunction(`f.${metricData.alias}_value`)} AS value_for_reaggregation
                   ${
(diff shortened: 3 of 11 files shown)


mknowlton89:
> Great!
> 
> This should really only change the next update if it isn't a cron schedule (right?). But I think `determineNextDate()` handles that and we should not add any special logic here.

Correct, or if the cron schedule has changed - that would also cause the `nextUpdate` date to change. E.G. If you went from doing the update every morning at 6am to every day at 7am, that would also cause the `nextUpdate` to change.

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -157,9 +161,25 @@ export async function updateDashboard(
...
+  if (updates.enableAutoUpdates === false) {
+    // Auto-updates being disabled - clear the nextUpdate
+    dashboardUpdates.nextUpdate = undefined;
+  } else if (shouldRecalculateNextUpdate(updates, dashboard)) {
+    // Recalculate nextUpdate based on the schedule
+    const schedule = updates.updateSchedule ?? dashboard.updateSchedule;
+    dashboardUpdates.nextUpdate = schedule
+      ? (determineNextDate(schedule) ?? undefined)
+      : undefined;
+  }
```
Kevin-Chant: This logic should live in `beforeUpdate` on `DashboardModel` so it can apply to both this controller and the API endpoint (and anywhere else that Dashboards get updated in the future)

mknowlton89: Ah, yea - great call.

mknowlton89: So I don't think the `beforeUpdate` is the right place for this given we need to add the `nextUpdate` field to the `updates` object that we're eventually passing into the `updateById`.

Currently, all of the `beforeUpdate` calls do some validation or impact adjacent objects. But we need to update the actual `updates` object.

The current pattern that I've built out updates the `updateById` on the DashboardModel itself, which does this logic and then ultimately calls `super.updateById`.

From what I can tell this is the pattern that will allow us to manipulate the `updates` object before passing it to Mongo.

Open to feedback on the pattern if you have other ideas.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -768,7 +768,7 @@ export abstract class BaseModel<
       );
     }
 
-    await this.afterUpdate(doc, updates, newDoc, options?.writeOptions);
+    await this.afterUpdate(doc, allUpdates, newDoc, options?.writeOptions);
```
mknowlton89, line 771: @Kevin-Chant I started getting some test failures for the `BaseModel.test.ts` file and it was because the `beforeUpdate` call is now receiving the `allUpdates` which contains the `dateUpdated` property.

And as I was fixing it, I noticed that we were still passing the `updates` object, and not the `allUpdates` to the `afterUpdate`.

It feels like we should use the `allUpdates` here as well - that sound right to you?

Kevin-Chant: Yeah for parity that makes sense. Not as important as beforeUpdate but should be better if they match

PR #5218, opened 2026-02-02
REST API: add metric usage endpoint

lukesonnet:
### Features and Changes

Add dedicated endpoint to tracking metric usage.

### Testing

Request (met_1234 doesn't exist):
```
 curl -X GET "http://localhost:3100/api/v1/usage/metrics?ids=fact__gh5wjyb5mfzwkqiz,met_3cb763qgj7mgh3psua,met_1234" \
  -u secret_user_XXXX
```

Response:
```
{
  "metricUsage": [
    {
      "metricId": "fact__gh5wjyb5mfzwkqiz",
      "experiments": [
        {
          "experimentId": "exp_3cb763q17mzmhassqau",
          "experimentStatus": "running",
          "lastSnapshotAttempt": "2025-11-19T00:07:13.113Z"
        },
        {
          "experimentId": "exp_3cb763q1n5qmevljg4y",
          "experimentStatus": "running",
          "lastSnapshotAttempt": "2026-02-10T19:18:43.221Z"
        }
      ],
      "lastSnapshotAttempt": "2026-02-10T19:18:43.221Z"
    },
    {
      "metricId": "met_3cb763qgj7mgh3psua",
      "experiments": [
        {
          "experimentId": "exp_3cb763q59hmg9mp9lr",
          "experimentStatus": "running",
          "lastSnapshotAttempt": "2026-02-10T19:18:43.211Z"
        }
      ],
      "lastSnapshotAttempt": "2026-02-10T19:18:43.211Z"
    },
    {
      "metricId": "met_1234",
      "experiments": [],
      "lastSnapshotAttempt": null
    }
  ]
}
```

[x] works if metric is missing
[x] works if metric is only added via metric group


lukesonnet:
cursor review

packages/back-end/generated/spec.yaml
```diff
@@ -3334,6 +3337,51 @@ paths:
                 properties:
                   deletedId:
                     type: string
+  /metric-usage:
```
ahdriel: Oooh -- I see the need to have a new endpoint to accept batches of metrics instead of getting single metrics under `/metrics/{id}/usage`.

question: Are there plans to add more usage-related data in the future? Should we namespace it under `/usage/metrics` maybe? What do you think?

packages/back-end/generated/spec.yaml
```diff
@@ -3334,6 +3337,51 @@ paths:
...
+    post:
```
ahdriel: Per-REST conventions we should use POST only if we are modifying a resource, which is not the case here (though this is not a fully compliant REST api either 🤷)

What do you think about changing it to a GET with query string params for the metricIds? Would be easier to cache for example as getting the same URL should return the same results (if nothing changed in the experiment obviously)

packages/back-end/src/api/openapi/payload-schemas/PostMetricUsagePayload.yaml
```diff
@@ -0,0 +1,9 @@
+type: object
+required:
+  - metricIds
+properties:
+  metricIds:
+    type: array
+    items:
+      type: string
+    description: Array of metric IDs (both fact and legacy) to get usage for
```
ahdriel: question: if we are calling it out, should we also include metric groups in the description?

packages/back-end/src/api/metric-usage/postMetricUsage.ts
```diff
@@ -0,0 +1,102 @@
...
+      lastSnapshotAttempt: exp.lastSnapshotAttempt
+        ? exp.lastSnapshotAttempt.toISOString()
+        : null,
+    }));
+
+    // Calculate summary statistics
+    const nRunningExperiments = experimentList.filter(
+      (e) => e.experimentStatus === "running",
+    ).length;
+    const nTotalExperiments = experimentList.length;
```
ahdriel: question: If we are going over all experiments regardless, wdyt about returning the number for each experiment status?

maybe in an object keyed by the experiment status?

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -791,6 +792,91 @@ export async function getExperimentsUsingMetric({
...
+  context,
+  metricIds,
+  // Optional, skips second DB call if already fetched
+  allMetricGroups,
+  excludeMetricGroupIds,
+  limit,
+}: {
+  context: ReqContext | ApiReqContext;
+  metricIds: string[];
+  allMetricGroups?: MetricGroupInterface[];
```
ahdriel: suggestion: Should we make this non-optional?

So we either provide it or pass an empty object if we don't case about it (or even undefined if we want), but we need to call it out explicitly.

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -791,6 +792,91 @@ export async function getExperimentsUsingMetric({
...
+  excludeMetricGroupIds?: boolean;
+  limit?: number;
+}): Promise<ExperimentInterface[]> {
+  if (metricIds.length === 0) {
+    return [];
+  }
+
+  // Fetch all metric groups once and build a map from metric ID to group IDs
+  const metricToGroupIds = new Map<string, string[]>();
```
ahdriel: This logic is the same between postMetricUsage and here -- or at least very similar.

Should we use a helper or something? Maybe `buildMetricIdToGroupIdsMap`?

packages/back-end/generated/spec.yaml
lukesonnet: I changed it away from that because the idea of query string params separated by commas makes my skin crawl.

But sure, I will switch it back, it makes sense.

packages/back-end/generated/spec.yaml
lukesonnet: I'll make this change, even though I'm not sure what other usage we'd do in the short term. Perhaps we could have like EAQ usage, Fact Table usage.

packages/back-end/src/api/metric-usage/postMetricUsage.ts
lukesonnet: Actually, I'm going to just remove this. It's redundant, and easy enough to find from the experiments list. I'm going to keep the `lastSnapshotAttempt` computed top-level field since it's so important. The rest they can easily count themselves and the reduced return object for the experiment makes it easy to see what they need to see.

packages/back-end/src/api/openapi/payload-schemas/PostMetricUsagePayload.yaml
lukesonnet: It does not include metric groups, it just searches metric usage in an experiment including if it is part of a metric group

packages/back-end/src/models/ExperimentModel.ts
lukesonnet: Sure, and clarified the comments here.

packages/back-end/src/models/ExperimentModel.ts
lukesonnet: The fact is we can just pass in the map for usage directly to this method as we don't need to compute it in both places. I made that change and now the logic lives in one place and the args on this method are clearer.

I'm going to let the next person that re-uses this method deal with that if they need to.

packages/back-end/generated/spec.yaml
ahdriel: I haven't seen the updates yet, but we can also take an array and the URL format would be `/usage?metricIds[]=123&metricIds[]=321`

But as you can see it would require more characters

packages/back-end/src/models/ExperimentModel.ts
ahdriel: the next person = us in 2 months lol

packages/back-end/src/api/usage/getMetricUsage.ts
```diff
@@ -0,0 +1,90 @@
...
+          if (!exp.lastSnapshotAttempt) return latest;
+          if (!latest) return exp.lastSnapshotAttempt;
+          return exp.lastSnapshotAttempt > latest
+            ? exp.lastSnapshotAttempt
+            : latest;
+        },
+        null,
+      );
+
+      return {
```
ahdriel, line 99: question: When a metric does not exist, this is the returned value:

```
{
  "metricUsage": [
    {
      "metricId": "fact__fzty7m44mkrcl0u6n",
      "experiments": [],
      "lastSnapshotAttempt": null
    }
  ]
}
```

Is this the same if I don't have permission to see any of the experiments from the metric?
Should we differentiate them?

packages/back-end/src/models/ExperimentModel.ts
lukesonnet: Yes, future me.

packages/back-end/src/api/usage/getMetricUsage.ts
lukesonnet: Added 

```
{
      "metricId": "met_1234",
      "error": "Metric not found or no permission to read the metric."
}
```

PR #5220, opened 2026-02-02
feat(results): Add Metric Drilldown to reports, dashboards & public pages

ahdriel:
### Features and Changes

In #5159 we ended up removing Drilldown from other parts of the application that use `ResultsTable` but don't use `CompactResults`.

This adds it back to:
- Public Experiment pages
- Reports (public pages included)
- Dashboards

And includes misc fixes around TimeSeries data and unauthenticated users.

### DIFF
--- a/packages/front-end/components/Report/ReportResults.tsx
+++ b/packages/front-end/components/Report/ReportResults.tsx
@@ -179,20 +179,15 @@ export default function ReportResults({
               report.experimentAnalysisSettings.customMetricSlices
             }
             statsEngine={
-              report.experimentAnalysisSettings?.statsEngine ||
-              DEFAULT_STATS_ENGINE
+              analysis?.settings?.statsEngine || DEFAULT_STATS_ENGINE
             }
             pValueCorrection={pValueCorrection}
             startDate={getValidDate(phaseObj.dateStarted).toISOString()}
             endDate={getValidDate(phaseObj.dateEnded).toISOString()}
             reportDate={snapshot.dateCreated}
             isLatestPhase={phase === phases.length - 1}
-            sequentialTestingEnabled={
-              report.experimentAnalysisSettings?.sequentialTestingEnabled
-            }
-            differenceType={
-              report.experimentAnalysisSettings.differenceType || "relative"
-            }
+            sequentialTestingEnabled={analysis?.settings?.sequentialTesting}
+            differenceType={analysis?.settings.differenceType}
             ssrPolyfills={ssrPolyfills}
             isReportContext
           >
@@ -211,10 +206,7 @@ export default function ReportResults({
                 statsEngine={
                   analysis?.settings?.statsEngine || DEFAULT_STATS_ENGINE
                 }
-                differenceType={
-                  report?.experimentAnalysisSettings?.differenceType ||
-                  "relative"
-                }
+                differenceType={analysis.settings.differenceType}
                 ssrPolyfills={ssrPolyfills}
               />
             ) : showBreakDownResults ? (
@@ -250,10 +242,7 @@ export default function ReportResults({
                 pValueCorrection={pValueCorrection}
                 settingsForSnapshotMetrics={settingsForSnapshotMetrics}
                 sequentialTestingEnabled={analysis?.settings?.sequentialTesting}
-                differenceType={
-                  report?.experimentAnalysisSettings?.differenceType ||
-                  "relative"
-                }
+                differenceType={analysis.settings.differenceType}
                 // metricFilter={metricFilter}
                 // setMetricFilter={setMetricFilter}
                 ssrPolyfills={ssrPolyfills}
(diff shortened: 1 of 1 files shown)


packages/front-end/components/MetricDrilldown/MetricDrilldownOverview.tsx
```diff
@@ -164,12 +171,14 @@ function MetricDrilldownOverview({
           Metric definition
         </Text>
         <MetricDrilldownMetadata statsEngine={statsEngine} row={row} />
-        <Flex direction="row" gap="5">
-          <MetricDrilldownMetricCard metric={metric} type="numerator" />
-          {isRatioMetric(metric) && (
-            <MetricDrilldownMetricCard metric={metric} type="denominator" />
-          )}
-        </Flex>
+        {isAuthenticated && (
```
lukesonnet, line 174: This just hides the drilldown card so we don't show "unknown" stuff that requires an API call on public pages?

packages/front-end/components/Report/ReportResults.tsx
```diff
@@ -158,7 +159,41 @@ export default function ReportResults({
             <LoadingSpinner />
           </div>
         ) : (
-          <>
+          <MetricDrilldownProvider
+            experimentId={snapshot.experiment}
+            phase={phase}
+            experimentStatus="stopped"
```
lukesonnet: Why is this set to "stopped"?

packages/front-end/components/Report/ReportResults.tsx
```diff
@@ -158,7 +159,41 @@ export default function ReportResults({
...
+              analysis?.settings?.statsEngine || DEFAULT_STATS_ENGINE
+            }
+            pValueCorrection={pValueCorrection}
+            startDate={getValidDate(phaseObj.dateStarted).toISOString()}
+            endDate={getValidDate(phaseObj.dateEnded).toISOString()}
+            reportDate={snapshot.dateCreated}
+            isLatestPhase={phase === phases.length - 1}
+            sequentialTestingEnabled={analysis?.settings?.sequentialTesting}
+            differenceType={
+              report.experimentAnalysisSettings.differenceType || "relative"
```
lukesonnet: Is there a reason this comes from the report but the sequential testing comes from the analysis? We seem to use mostly the analysis settings in the Results.tsx call.

packages/front-end/components/Experiment/ResultsTable.tsx
```diff
@@ -199,11 +203,28 @@ export default function ResultsTable({
...
+  const drilldownContext = useMetricDrilldownContext();
+  const effectiveOnRowClick = onRowClick ?? drilldownContext?.openDrilldown;
+
+  // Dev warning for missing drilldown handler
+  if (
+    process.env.NODE_ENV === "development" &&
+    warnIfNoDrilldown &&
+    !effectiveOnRowClick
+  ) {
+    console.warn(
```
lukesonnet: Do we want to leave this in here?

packages/front-end/components/Report/ReportResults.tsx
lukesonnet: Looks like this is just used by

1. `ExperimentMetricTimeSeriesGraphWrapper` to add space to the right of the graph.

TBH, I think we should just always show one day to the right to give it a buffer visually....

So we could remove it from that code entirely.

2. `showTimeRemaining` in `getRowResults`

Which is kind of useful, and I guess we don't want to show it in custom reports since it doesn't really make sense, but then it's really about it being in the report context, not that the status is "stopped."

packages/front-end/components/MetricDrilldown/MetricDrilldownOverview.tsx
ahdriel: Yeah -- and I am not sure how much we want to show on a "public" page. The user would still need to get the link from somewhere, so I guess this is less of a concern?

lukesonnet: Happy to just hide in general as a rule to play it safe, just wanted to confirm the choice here.

packages/front-end/components/Experiment/ResultsTable.tsx
ahdriel: Removed

packages/front-end/components/Report/ReportResults.tsx
ahdriel: Good catch! for the report we should use from `report.experimentAnalysisSettings`.

Fixed

lukesonnet: You mean `analysis.settings`? This already uses `report.experimentAnalysisSettings.`

ahdriel: I meant that I am updating `sequentialTestingEnabled` to also come from `report`. I think this line is correct as-is.

Fixed locally, fixing some TS errors and I'll push.

packages/front-end/components/Report/ReportResults.tsx
ahdriel: Removed from here.

lukesonnet (approved):

PR #5223, opened 2026-02-03
feat(design-system): Add Typography components Text & Heading

ahdriel:
### Features and Changes

- Add standardized `Text` & `Heading` components under `/ui/`
- Forbids additional usage of `Text` & `Heading` directly from `radix-themes`


packages/front-end/ui/Heading.tsx
```diff
@@ -0,0 +1,96 @@
...
+import type { HeadingProps as RadixHeadingProps } from "@radix-ui/themes";
+
+type HeadingSizes = "small" | "medium" | "large" | "x-large" | "2x-large";
+type HeadingWeights = "medium" | "semibold";
+type HeadingAlign = "left" | "center" | "right";
+type HeadingColors =
+  | "text-high"
+  | "text-mid"
+  | "text-low"
+  | RadixHeadingProps["color"];
```
msamper: Do we need to limit the colors or do we want to allow folks to use any Radix color? I know Kari had wanted to limit colors for another component so not sure if the same applies here

packages/front-end/ui/Heading.tsx
```diff
@@ -0,0 +1,96 @@
...
+  small: "3",
+  medium: "4",
+  large: "5",
+  "x-large": "6",
+  "2x-large": "7",
+};
+
+const radixWeightMap: Record<HeadingWeights, RadixHeadingProps["weight"]> = {
+  medium: "medium",
+  semibold: "bold",
```
msamper, line 20: Is there a reason we map `semibold` to `bold`? Is it that the heading will actually render as semibold rather than bold due to some radix theme settings?

packages/front-end/ui/Heading.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+export interface HeadingProps {
+  children: React.ReactNode;
+  as: "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
msamper: Could we do `RadixHeadingProps["as"]`?

packages/front-end/ui/Text.tsx
```diff
@@ -0,0 +1,113 @@
...
+  | "normal"
+  | "nowrap"
+  | "pre-wrap"
+  | "pre-line"
+  | "break-spaces";
+type TextColors =
+  | "text-high"
+  | "text-mid"
+  | "text-low"
+  | RadixTextProps["color"];
```
msamper: Same question here re: colors

packages/front-end/ui/Text.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [textColor, setTextColor] = useState<TextProps["color"]>("text-mid");
+  return (
+    <Flex direction="row" gap="9">
+      <Flex direction="column" gap="1">
+        <span>Unwrapped text</span>
+        <Grid columns="repeat(3, 1fr)" gap="0.5rem">
+          <Text size="medium" weight="regular" color={textColor}>
+            Unwrapped text
+          </Text>
+          <Text size="medium" weight="medium" color={textColor}>
```
msamper, line 25: do we want to include `medium/regular` and then all the variations for `small`?

packages/front-end/ui/Heading.tsx
ahdriel: This is a good callout. I know that the current mapping works as expected for the styles provided by the design team, but your raise an excellent point that this might not be needed.

I believe it is related to these settings we have in `radix-config.css`:
```
  --font-weight-light: 200;
  --font-weight-regular: 400;
  --font-weight-medium: 500;
  --font-weight-bold: 600;
```

Would you be open to merging as-is and I'll do the font-weight investigation as a follow-up either today or tomorrow?

packages/front-end/ui/Heading.tsx
ahdriel: I am going to limit it for now to just text high, mid & low.

But I added a comment that we might need to expand it to support more colors.

packages/front-end/ui/Text.stories.tsx
ahdriel: Nice catch! It was me testing some other stuff. Fixed.

PR #5227, opened 2026-02-03
Bypass pnpm in yarn shim for start commands

jdorn:
### Features and Changes

This fixes a memory leak with pnpm and keeps the Docker image fully backwards compatible for the vast majority of users.

### DIFF
--- a/docs/docs/self-host/env.mdx
+++ b/docs/docs/self-host/env.mdx
@@ -198,7 +198,7 @@ If your company has two or more largely independent products (e.g. Google has Se
 
 The GrowthBook API is instrumented with OpenTelemetry to publish observability metrics, traces, and logs.
 
-To enable, you must change the Docker CMD from the default `yarn start` to `yarn start:with-tracing`.
+To enable, you must set the environment variable `TRACING_PROVIDER=opentelemetry`.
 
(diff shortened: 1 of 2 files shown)


docs/docs/self-host/env.mdx
```diff
@@ -198,7 +198,7 @@ If your company has two or more largely independent products (e.g. Google has Se
 
 The GrowthBook API is instrumented with OpenTelemetry to publish observability metrics, traces, and logs.
 
-To enable, you must change the Docker CMD from the default `pnpm start` to `pnpm start:with-tracing`.
+To enable, you must change the Docker CMD from the default `yarn start` to `yarn start:with-tracing`.
```
ahdriel: question: We are directing users to use the `yarn` shim to bypass the pnpm runner?

I think this is confusing though, even though the results are likely better than using pnpm.

Ideally we would be able to remove this yarn shim layer eventually, and this makes harder to get there.

I don't have a better suggestion at the moment though, so just commenting to note.

jdorn: Yeah.  Maybe we should make our own `start` script and push people to start using that instead.  Hide the underlying implementation so we can iterate and improve it in the future

PR #5230, opened 2026-02-04
Add savedGroupReferences capability for Swift SDK

vladenisov:
### Features and Changes

There is no `Pass ID Lists by reference` option in Swift connections, but SDK supports it starting from the 1.0.58.

Here is the confirmation in Slack: https://growthbookusers.slack.com/archives/C05T9EY0UKC/p1770118129248409


jdorn:

PR #5231, opened 2026-02-04
Install Poetry from a mirror for Preview environment network issues.

tzjames:
### Features and Changes
The preview environment in ci has been failing with connectivity issues to pypi.  This is an attempt around that.

It:
- Get poetry from a mirror (because Fly.io's depot container can't connect to pypi.)
- Telling poetry lock not to update (because doing `poetry source add` modifies pyproject.toml and poetry complains it is then out of date)
- Download kerberos in the earlier image as well (not sure why that is not needed normally)
- Made a sentry-cli a non-optional dependency (as that was also having networking issues fetching it later, but pnpm seems to install it fine)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,7 +70,7 @@ RUN \
   && rm -rf packages/shared/node_modules \
   && rm -rf packages/sdk-js/node_modules \
   && rm -rf packages/sdk-react/node_modules \
-  && SENTRYCLI_SKIP_DOWNLOAD=1 pnpm install --frozen-lockfile --prod --no-optional \
+  && pnpm install --frozen-lockfile --prod --no-optional \
   && pnpm store prune \
   && find node_modules -type f -name "*.md" -delete \
   && find node_modules -type f -name "*.ts" ! -name "*.d.ts" -delete \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,16 +1,31 @@
 ARG PYTHON_MAJOR=3.11
 ARG NODE_MAJOR=20
+ARG PYPI_MIRROR_URL=""
 
 # Build the python gbstats package
 FROM python:${PYTHON_MAJOR}-slim AS pybuild
+ARG PYPI_MIRROR_URL
 WORKDIR /usr/local/src/app
-RUN pip install --upgrade pip
 COPY ./packages/stats .
```
ahdriel, line 9: question: Is this safe to do before `pip install`?

I think we should keep the same order -- specially as stats is the only package with python dependencies.

tzjames: Is that referring to Copying ./packages/stats?  The line before it was just upgrading pip.  Now it will use whatever pip came with the python:3.11-slim base image. I think that should be safe.  We got rid of the same thing on line 87 and then run the pip3 install  that actually builds the packages that stats needs.  So I think all is good?

PR #5238, opened 2026-02-05
Initial node 24 upgrade

tzjames:
### Features and Changes
Node 20 is soon no longer supported.  We upgrade here to Node 24.

- upgrades `isolated_vm` to get a version that is supported by node 24
- overrided `nan` to get a version that is supported by node 24 (it is used by lz4 which is used by @databricks/sql. Neither  lz4 nor @databricks/sql have a more recent version so had to do the override on nan)
- updated Docker
- updated ci
- updated recommended node version in package.json

I had to:
```
cd /Users/james/growthbook/growthbook/node_modules/.pnpm/lz4@0.6.5/node_modules/lz4
npx node-gyp rebuild
```
at one point locally.  As `lz4` used `nan` which I had to override to get a node24 compatible version.  I'm not sure all devs will need to do that.  Perhaps `pnpm clean` and then `pnpm i` would be enough. 

### Testing
ci passes


Kevin-Chant (approved):
Need to fix the merge conflicts, but LGTM.
In my case, I only needed to `nvm use 24`, `pnpm i`, and `pnpm test`. No need for any clean/reinstall

Edit: I did need to `pnpm install -g node-gyp` when I went back to main though, which was weird

PR #5241, opened 2026-02-05
Use LIMIT 0 for BQ and Snowflake column checks

lukesonnet:
### Features and Changes

The validity query just checks for certain columns and we can remove the need to scan data for BQ and Snowflake by using LIMIT 0 instead of LIMIT 1.

This will make validation queries run more quickly, be less prone to network timeouts (esp. for self-hosted users), but still provide the same functionality.

This currently only applies to EAQ validations.

### Testing

- [X] BQ validation query scans no rows and returns the right columns
- [X] Snowflake validation query scans no rows and returns the right columns

### DIFF
--- a/packages/back-end/src/services/datasource.ts
+++ b/packages/back-end/src/services/datasource.ts
@@ -352,9 +352,9 @@ export async function testQueryValidity(
 
     let columns: Set<string>;
 
-    // For datasources supporting LIMIT 0, use column metadata
-    if (integration.supportsLimitZeroColumnValidation?.()) {
-      const columnNames = results.columns?.map((c) => c.name) || [];
+    // For datasources where the result includes columns, use column metadata
+    if (results.columns) {
+      const columnNames = results.columns.map((c) => c.name);
       if (columnNames.length === 0) {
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/datasource.ts
```diff
@@ -349,10 +349,23 @@ export async function testQueryValidity(
   const sql = integration.getTestValidityQuery(query.query, testDays);
   try {
     const results = await integration.runTestQuery(sql);
-    if (results.results.length === 0) {
-      return "No rows returned";
+
+    let columns: Set<string>;
+
+    // For datasources supporting LIMIT 0, use column metadata
+    if (integration.supportsLimitZeroColumnValidation?.()) {
```
jdorn: I feel like we should just check if results.columns exists and use it if so. Doesn't matter what the supportsZeroLimit setting returns.

lukesonnet: Ok, made this change. Didn't check for length, just if `columns` exists.

lukesonnet:
cursor review

PR #5243, opened 2026-02-06
Statsig importer: Dedupe metrics while importing, fix bug that imported all metrics

bryce-fitzsimons:


packages/front-end/services/importing/statsig/statsig-importing.ts
```diff
@@ -1526,7 +1542,7 @@ export async function runImport(options: RunImportOptions) {
       case "experiments":
         return `exp-${item.experiment?.name || item.experiment?.id || index}`;
       case "metrics":
-        return `metric-${item.metric?.id || index}`;
+        return `metric-${item.metric?.name || index}`;
```
bryce-fitzsimons, line 1545: we don't track statsig's internal id in our DB, so deduping on the name is sufficient for infrequently-updated metrics.

PR #5244, opened 2026-02-06
WatchModel to BaseModel

Kevin-Chant:
### Features and Changes

Migrates `WatchModel` to use `BaseModel`

Also includes new helpers within `BaseModel` for `countDocuments` and `bulkWrite` as well as a new migration option for legacy models that were incompatible with `BaseModel`

### Dependencies

We'll want to migrate the `WatchModel` in cloud before merging this PR so mongo doesn't lock up with the huge bulk update.

### Testing

Ensure that you have existing watches in mongo (or start watching some features/experiments if not)
```
db.watches.find({})
```
The documents should lack the `id` field until you start the server, at which point it should be generated
Verify that starting the server for the first time doesn't emit any errors or warnings (e.g. `ERROR (38245): Error creating org/id unique index for watches`)

Add a new experiment to see that you're automatically enabled as a watcher and check that you can remove and re-add watching for it
Same for a new feature within the organization

### DIFF
--- a/packages/back-end/src/models/BaseModel.ts
+++ b/packages/back-end/src/models/BaseModel.ts
@@ -31,7 +31,7 @@ import {
 } from "back-end/src/services/context";
 import { ApiRequest } from "back-end/src/util/handler";
 import { ApiBaseSchema, ApiModelConfig } from "back-end/src/api/ApiModel";
-import { safeBulkWrite } from "back-end/src/util/mongo.util";
+import { dbSafeBulkWrite } from "back-end/src/util/mongo.util";
 
 export type Context = ApiReqContext | ReqContext;
 
@@ -833,11 +833,11 @@ export abstract class BaseModel<
   protected async _dangerousBulkWriteCrossOrganization(
     operations: AnyBulkWriteOperation[],
   ) {
-    return safeBulkWrite(this._dangerousGetCollection(), operations);
+    return dbSafeBulkWrite(this._dangerousGetCollection(), operations);
   }
 
   protected async bulkWrite(operations: AnyBulkWriteOperation[]) {
-    return safeBulkWrite(
+    return dbSafeBulkWrite(
       this._dangerousGetCollection(),
       operations.map((op) => {
         if ("insertOne" in op) {
--- a/packages/back-end/src/util/mongo.util.ts
+++ b/packages/back-end/src/util/mongo.util.ts
@@ -165,10 +165,11 @@ export function getCollection<T extends Document>(name: string) {
 }
 
 /**
- * Attempts to perform a bulkWrite operation if supported by the collection. If not, falls back to chunked individual operations.
+ * Attempts to perform a bulkWrite operation if supported by the database driver.
+ * If not, falls back to chunked individual operations.
  * Supports updateOne and insertOne operations. Extend as needed for other op types.
  */
-export async function safeBulkWrite(
+export async function dbSafeBulkWrite(
   collection: Collection<Document>,
   ops: AnyBulkWriteOperation<Document>[],
   options?: BulkWriteOptions,
@@ -216,7 +217,7 @@ export async function safeBulkWrite(
         }
       } else {
         throw new Error(
-          "Unsupported bulkWrite operation type in safeBulkWrite",
+          "Unsupported bulkWrite operation type in dbSafeBulkWrite",
         );
       }
     }),
--- a/packages/back-end/test/util/mongo.util.test.ts
+++ b/packages/back-end/test/util/mongo.util.test.ts
@@ -2,7 +2,7 @@ import mongoose from "mongoose";
 import { setupApp } from "back-end/test/api/api.setup";
 import {
   getConnectionStringWithDeprecatedKeysMigratedForV3to4,
-  safeBulkWrite,
+  dbSafeBulkWrite,
 } from "back-end/src/util/mongo.util";
 
 describe("mongo utils", () => {
@@ -127,7 +127,7 @@ describe("mongo utils", () => {
     });
   });
 
-  describe("safeBulkWrite", () => {
+  describe("dbSafeBulkWrite", () => {
     setupApp();
     const collectionName = "bulkWriteTests";
     const _model = mongoose.model(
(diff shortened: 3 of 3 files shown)


packages/back-end/src/models/BaseModel.ts
```diff
@@ -926,45 +951,53 @@ export abstract class BaseModel<
...
-            err,
-            `Error creating org/id unique index for ${this.config.collectionName}`,
-          );
-        }),
-    );
-
-    // If id is globally unique, create an index for that
-    if (this.config.globallyUniqueIds) {
-      promises.push(
+      migrationPromise.then(() => {
```
Kevin-Chant, line 990: @tzjames I'm open to suggestions for a better syntax to queue all of the index creation behind this promise that doesn't affect `pendingIndexOperations`

packages/back-end/src/models/BaseModel.ts
```diff
@@ -810,6 +815,26 @@ export abstract class BaseModel<
...
+    };
+    return this._dangerousGetCollection().countDocuments(query);
+  }
+
+  protected async _dangerousBulkWrite(
+    operations: AnyBulkWriteOperation[],
+    ordered: boolean = false,
+  ) {
+    this._dangerousGetCollection().bulkWrite(operations, { ordered });
+  }
```
Kevin-Chant: I also don't see an easy way to scope down the `bulkWrite` operations per org so this might have to always be a `_dangerous` helper. Open to suggestions though

tzjames: Here is Claude's suggestions...
```
 const orgId = this.context.org.id;
    const baseQuery = this.getBaseQuery();

    // Scope all operations to the current organization
    const scopedOperations = operations.map((op) => {
      if ("insertOne" in op) {
        return {
          insertOne: {
            ...op.insertOne,
            document: {
              ...op.insertOne.document,
              organization: orgId,
            },
          },
        };
      } else if ("updateOne" in op) {
        return {
          updateOne: {
            ...op.updateOne,
            filter: {
              ...baseQuery,
              ...op.updateOne.filter,
              organization: orgId,
            },
          },
        };
      } else if ("updateMany" in op) {
        return {
          updateMany: {
            ...op.updateMany,
            filter: {
              ...baseQuery,
              ...op.updateMany.filter,
              organization: orgId,
            },
          },
        };
      } else if ("deleteOne" in op) {
        return {
          deleteOne: {
            ...op.deleteOne,
            filter: {
              ...baseQuery,
              ...op.deleteOne.filter,
              organization: orgId,
            },
          },
        };
      } else if ("deleteMany" in op) {
        return {
          deleteMany: {
            ...op.deleteMany,
            filter: {
              ...baseQuery,
              ...op.deleteMany.filter,
              organization: orgId,
            },
          },
        };
      } else if ("replaceOne" in op) {
        return {
          replaceOne: {
            ...op.replaceOne,
            filter: {
              ...baseQuery,
              ...op.replaceOne.filter,
              organization: orgId,
            },
            replacement: {
              ...op.replaceOne.replacement,
              organization: orgId,
            },
          },
        };
      }
      return op;
    });

    return this._dangerousGetCollection().bulkWrite(scopedOperations, {
      ordered,
    });
```
In general I think we will only be doing it for updateOne.  So maybe just handle that case, and we can add more when needed?  Also BulkWrite is the thing that Cosmos doesn't handle, so we probably should fallback to using `promiseAllChunks`.  See this [thread](https://growthbookapp.slack.com/archives/C055GKYGT26/p1768152262104359?thread_ts=1767909142.039069&cid=C055GKYGT26)

packages/back-end/src/models/BaseModel.ts
tzjames: Seems reasonable to me.

packages/back-end/src/models/WatchModel.ts
```diff
@@ -1,97 +1,93 @@
...
-      organization,
-    })
-    .project({ userId: 1, _id: 0 })
-    .toArray();
-  return watchers.map((watcher) => watcher.userId);
-}
+  public async upsertWatch({ userId, item, type }: UpdateWatchOptions) {
+    const existing = await this.getWatchedByUser(userId);
+    existing
+      ? await this._updateOne(existing, { [type]: [...existing[type], item] })
```
tzjames: Could this end up getting duplicate items if it is called twice?

packages/back-end/src/models/WatchModel.ts
```diff
@@ -1,97 +1,93 @@
...
-  userId: String,
-  organization: String,
-  experiments: [String],
-  features: [String],
+const BaseClass = MakeModelClass({
+  schema: watchSchema,
+  collectionName: "watches",
+  idPrefix: "watch_",
+  readonlyFields: [],
+  additionalIndexes: [{ unique: true, fields: { userId: 1, organization: 1 } }],
```
tzjames: we do do a call `getExperimentWatchers` shall we have an additional index for that?

packages/back-end/src/models/WatchModel.ts
```diff
@@ -1,97 +1,93 @@
...
+    // If multiple documents match, updateOne will update the first matching document only.
+    await this._dangerousBulkWrite(
+      Array.from({ length: numNullIds }, () => ({
+        updateOne: {
+          filter: { id: null },
+          update: { $set: { id: this._generateId() } },
+        },
+      })),
+      true, // Ordered to prevent the sequential `updateOne`s from running in parallel
+    );
```
tzjames: Do we have any locking on this to make sure it is executed only once?  If we have 8 servers all starting at the same time, for Cloud we have 25k of these watches.  We might be getting these huge queries multiple times hitting Mongo.  Mongo might lock the collection until the migration is complete, so any requests adding or updating a watch might have to wait until it is done.  Maybe we might want to update cloud manually in batches?  You could consider adding some batches in here too so it doesn't do all 25k at once, but maybe like 1k at a time.

Kevin-Chant: As we talked about offline we'll go with the manual update option for cloud, and the existing migration should be sufficient for self-hosted orgs

packages/back-end/src/models/BaseModel.ts
```diff
@@ -810,6 +810,52 @@ export abstract class BaseModel<
...
+
+  protected async _dangerousBulkWriteCrossOrganization(
+    operations: AnyBulkWriteOperation[],
+  ) {
+    return safeBulkWrite(this._dangerousGetCollection(), operations);
+  }
+
+  protected async bulkWrite(operations: AnyBulkWriteOperation[]) {
+    return safeBulkWrite(
+      this._dangerousGetCollection(),
```
tzjames: I find the juxtaposition of safe and dangerous here a bit confusing.  Perhaps we could call it `crossDbCompatibleBulkWrite` or something similar.

Kevin-Chant: Agreed, though I opted to rename `safeBulkWrite` because the `_dangerous` prefix is how we've been trying to steer devs away from certain `BaseModel` helpers (or at least encourage extra caution)

PR #5251, opened 2026-02-07
Ty/attr improvements

teresayung:
### Features and Changes

This PR improves attribute discovery, metadata, admin, and documentation in four areas (per [ticket](https://www.notion.so/growthbook/Attribute-discovery-metadata-2efa857e74a0802c9583ff62a5c90fd1?source=copy_link)):

**Part 1 of the ticket is moved to another PR.**

**2. Filtering, search, and sort on the Attributes Table page**

- **Search:** Full-text search over property, description, datatype, project names, and tags.
- **Filters:** Filter by project, data type, tag, identifier, and archived (via `AttributeSearchFilters`), aligned with the features page layout.
- **Sort:** Sortable columns (e.g. property, description, datatype, tags).
- **Sticky table header** so filters and column headers stay visible when scrolling.

image

---

**3. Tags on attributes**

- **Data model:** `tags?: string[]` on `SDKAttribute` (shared types).
- **Backend:** Attributes API and router support `tags` on create/update; tag controller removes a tag from attributes when a tag is deleted.
- **Attributes UI:**
  - Create/Edit attribute modal: **Tags** field (same tag picker as elsewhere).
  - Attributes table: **Tags** column with `SortedTags`; tags included in search and in the new filters.

Attributes can be grouped and discovered by tag both on the attributes page and in the feature rule flow.

image

image

---

**4. Links and richer text in attribute descriptions**

- **Attribute description** is now **Markdown** (e.g. bold, lists, **links**).
  - Create/Edit attribute modal: description uses **MarkdownInput** instead of a plain textarea.
  - Attributes page: description column renders with the **Markdown** component so links and other Markdown render correctly.
- This supports embedded documentation links (and other formatting) in the description. A separate first-class "documentation link" property was not added; links are done via Markdown in the description.

image
Shortened URL on paste
image

---

### Dependencies

None.

### Testing

1. **Attributes page:** Use search, filters (project, datatype, tag, identifier, archived), and column sort. Confirm sticky header and that tags appear in the Tags column and in search/filters.
2. **Attribute tags:** Create/edit an attribute and add tags; confirm they appear on the attributes table and in the feature rule dropdown and tooltip.
3. **Attribute description:** Set an attribute description with Markdown (including a link). Confirm it renders as Markdown on the attributes page and in the rule tooltip.

### DIFF
--- a/packages/front-end/pages/attributes.tsx
+++ b/packages/front-end/pages/attributes.tsx
@@ -454,7 +454,7 @@ const FeatureAttributesPage = (): React.ReactElement => {
             <tbody>
               {attributeSchema?.length > 0 ? (
                 <>
-                  {filteredAttributes.map((v, i) => drawRow(v, i))}
+                  {filteredAttributes.map((v) => drawRow(v))}
                   {!filteredAttributes.length && isFiltered && (
                     <tr>
                       <td colSpan={8} className="text-center text-gray">


teresayung:
cursor review

teresayung:
cursor review

packages/back-end/src/api/attributes/postAttribute.ts
```diff
@@ -15,6 +16,11 @@ export const postAttribute = createApiRequestHandler(postAttributeValidator)(
 
     const org = req.context.org;
 
+    const tags = (req.body as { tags?: string[] }).tags ?? [];
```
bryce-fitzsimons: We shouldn't need to do this if our generated API types are correct. Currently the REST API validators are not referencing the new "tags" field. You'd need to add this to both `src/api/openapi/paths/postAttribute.yaml` and `putAttribute.yaml`:

```
          tags:
            type: array
            items:
              type: string
```

... and then run `pnpm generate-api-types`

packages/front-end/components/Features/AttributeModal.tsx
bryce-fitzsimons: Let's take the opportunity to refresh the UI here. Any time we see an "alert" classname, it should be converted to <Callout> (`@/ui/Callout`). While at it, we could ask the robot to convert this modal into Radix layouts and avoid bootstrap classnames.

packages/front-end/components/Markdown/MarkdownInput.tsx
```diff
@@ -83,6 +92,32 @@ const MarkdownInput: FC<{
     }
   }, [autofocus, textareaRef.current]);
 
+  const handlePaste = (e: React.ClipboardEvent<HTMLTextAreaElement>) => {
```
bryce-fitzsimons, line 95: nice

packages/front-end/components/Features/ConditionDisplay.tsx
```diff
@@ -571,15 +571,17 @@ export default function ConditionDisplay({
...
 }: {
   condition?: string;
   savedGroups?: SavedGroupTargeting[];
   prerequisites?: FeaturePrerequisite[];
+  project?: string;
   prefix?: ReactNode;
 }) {
   const { savedGroups } = useDefinitions();
-  const attributes = useAttributeMap();
+  const attributes = useAttributeMap(project);
```
bryce-fitzsimons: is this needed?

packages/front-end/components/SavedGroups/TruncatedConditionDisplay.tsx
```diff
@@ -9,6 +9,7 @@ interface TruncatedConditionDisplayProps {
   savedGroups?: SavedGroupTargeting[];
   prerequisites?: FeaturePrerequisite[];
   maxLength?: number;
+  project?: string;
```
bryce-fitzsimons: needed?

packages/front-end/pages/attributes.tsx
```diff
@@ -395,13 +469,37 @@ const FeatureAttributesPage = (): React.ReactElement => {
...
-            <thead>
+            <thead
+              className="sticky-top shadow-sm"
+              style={{ top: HEADER_HEIGHT_PX + "px", zIndex: 900 }}
+            >
               <tr>
-                <th>Attribute</th>
-                <th>Description</th>
-                <th>Data Type</th>
+                <SortableTH field="property">Attribute</SortableTH>
```
bryce-fitzsimons: I'd probably provide specific % widths for all columns (except description, let it fill the rest). The tag column looks really squished when there is a long description

packages/back-end/src/api/attributes/postAttribute.ts
teresayung: ahh good to know

packages/front-end/components/Features/ConditionDisplay.tsx
teresayung: `project` is passed so the attribute map that we show is scoped to that project.  When creating attributes, we have the option to assign no project (which I'm assuming means the attr is globally available) or to a specific project.  I'm passing `project` here so we get project specific attrs (in addition to the already included global attrs).  Wdyt?

packages/front-end/components/Features/AttributeModal.tsx
teresayung: changed! 
image

packages/back-end/src/api/attributes/postAttribute.ts
teresayung: https://github.com/growthbook/growthbook/pull/5251/commits/759570243d2f16381fd5f27c96adb90ac9cadb31 should address this change

packages/front-end/components/Features/ConditionDisplay.tsx
bryce-fitzsimons: We don't want to filter the attributeMap to the feature's project. The way you have it, It really only affects whether we pretty-format or fallback to JSON-displayed formatting for attributes that are outside of the feature's project scoping. But the way that we render ConditionDisplay should not depend on project at all.

packages/front-end/components/Features/AttributeModal.tsx
```diff
@@ -286,7 +292,7 @@ export default function AttributeModal({ close, attribute }: Props) {
               </Tooltip>
               . Do not use this format if you are using an incompatible SDK as
               it will break any filtering based on the attribute.
-            </div>
+            </Callout>
```
bryce-fitzsimons, line 295:
```suggestion
            <Callout status="warning">
Version string attributes are only
supported in{" "}
<Tooltip
@@ -286,7 +292,7 @@ export default function AttributeModal({ close, attribute }: Props) {
</Tooltip>
. Do not use this format if you are using an incompatible SDK as
it will break any filtering based on the attribute.
            </Callout>
```

packages/front-end/pages/attributes.tsx
```diff
@@ -125,27 +194,35 @@ const FeatureAttributesPage = (): React.ReactElement => {
     null,
   );
 
-  const drawRow = (v: SDKAttribute, i: number) => {
+  const drawRow = (v: SDKAttribute, _: number) => {
```
bryce-fitzsimons:
```suggestion
  const drawRow = (v: SDKAttribute) => {
```

packages/front-end/pages/attributes.tsx
```diff
@@ -155,15 +232,26 @@ const FeatureAttributesPage = (): React.ReactElement => {
...
-        <td className="text-gray">
+        <td style={{ minWidth: 100 }}>
+          <SortedTags tags={v.tags || []} useFlex={true} />
+        </td>
+        <td className="text-gray" style={{ minWidth: 85 }}>
           {numReferences > 0 ? (
-            <Link onClick={() => setShowReferencesModal(i)} className="nowrap">
+            <Link
+              onClick={() => {
+                const schemaIndex = attributeSchema.findIndex(
```
bryce-fitzsimons, line 257: I think it would be better to bake the original (unsorted, unfiltered) index into each attribute. Something like `const attributesWithIndex = useMemo(attributes.map((a, i) => { ...a, originalIndex: i })` or whatnot

PR #5252, opened 2026-02-08
Use the new SdkConnectionCacheModel for live traffic, deprecate SdkPayloadModel.

bryce-fitzsimons:
- Direct all live traffic to the new cache system (SdkConnectionCacheModel)
- On cache miss, regenerate (do not use old cache)
- Webhooks (global, sdk, and legacy) use the new cache
- Continue populating the old cache until we're ready to sunset it
- Introduce LATEST_SDK_PAYLOAD_SCHEMA_VERSION for cache busting
- Clean up some of the existing generation code. Lot of duplicate logic.

### DIFF
--- a/packages/back-end/src/models/SdkConnectionCacheModel.ts
+++ b/packages/back-end/src/models/SdkConnectionCacheModel.ts
@@ -4,6 +4,9 @@ import {
 } from "shared/validators";
 import { MakeModelClass } from "./BaseModel";
 
+// Increment this if we change the payload contents in a backwards-incompatible way
+export const LATEST_SDK_PAYLOAD_SCHEMA_VERSION = 1;
+
 const BaseClass = MakeModelClass({
   schema: sdkConnectionCacheValidator,
   collectionName: "sdkcache",
@@ -29,7 +32,10 @@ export class SdkConnectionCacheModel extends BaseClass {
   }
 
   public async getById(id: string) {
-    return await this._findOne({ id });
+    return await this._findOne({
+      id,
+      schemaVersion: LATEST_SDK_PAYLOAD_SCHEMA_VERSION,
+    });
   }
 
   public async upsert(
@@ -37,12 +43,15 @@ export class SdkConnectionCacheModel extends BaseClass {
     contents: string,
     auditContext?: SdkConnectionCacheAuditContext,
   ) {
-    const existing = await this.getById(id);
+    // Find existing doc by id only (ignore version) to support version upgrades
+    const existing = await this._findOne({ id });
     const updateData: {
       contents: string;
+      schemaVersion: number;
       audit?: SdkConnectionCacheAuditContext;
     } = {
       contents,
+      schemaVersion: LATEST_SDK_PAYLOAD_SCHEMA_VERSION,
     };
     if (auditContext) {
       updateData.audit = auditContext;
--- a/packages/shared/src/validators/sdk-connection-cache.ts
+++ b/packages/shared/src/validators/sdk-connection-cache.ts
@@ -12,6 +12,7 @@ export const sdkConnectionCacheAuditContextValidator = z.object({
 
 export const sdkConnectionCacheValidator = baseSchema.extend({
   contents: z.string(),
+  schemaVersion: z.number(),
   audit: sdkConnectionCacheAuditContextValidator.optional(),
 });
 


packages/back-end/src/services/features.ts
```diff
@@ -738,11 +734,15 @@ async function refreshSDKPayloadCache({
               }
             : undefined;
 
-        await context.models.sdkConnectionCache.upsert(
-          connection.key,
-          JSON.stringify(contents),
-          auditContext,
-        );
+        // Only write to cache if SDK_PAYLOAD_CACHE is not set to "none"
```
bryce-fitzsimons, line 746: are we still doing this?

bryce-fitzsimons:
cursor review

packages/back-end/src/controllers/features.ts
```diff
@@ -224,70 +238,102 @@ export async function getPayloadParamsFromApiKey(
...
+  if (storageLocation !== "none") {
+    const cached = await context.models.sdkConnectionCache.getById(params.key);
+    if (cached) {
+      try {
+        defs = JSON.parse(cached.contents);
+      } catch (e) {
+        logger.warn(e, "Failed to parse cached SDK payload, regenerating");
+      }
+    }
+  }
```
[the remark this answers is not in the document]
bryce-fitzsimons: this is a good catch. I knew there were issues with caching legacy keys, but forgot to patch. Creating a special-case cache key to resolve.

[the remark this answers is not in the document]
bryce-fitzsimons: I don't believe this is correct. We already proactively regenerate the cache whenever environment settings change. See `environment.controller.ts > putEnvironment`:

```
      if (!isEqual(existingProjects, newProjects)) {
        const connections = await findSDKConnectionsByOrganization(context);
        const affectedConnections = connections.filter(
          (c) => c.environment === id,
        );

        queueSDKPayloadRefresh({
          context,
          payloadKeys: [],
          sdkConnections: affectedConnections,
        });
      }
```

packages/back-end/src/controllers/features.ts
```diff
@@ -224,70 +238,106 @@ export async function getPayloadParamsFromApiKey(
...
+          dateUpdated: new Date(),
+          event: "cache-miss",
+          model: "sdk-connection",
+          stack: stackTrace,
+          connection: params as unknown as Record<string, unknown>,
+        })
+        .catch((e) => {
+          logger.warn(e, "Failed to write JIT generated payload to cache");
+        });
+    }
```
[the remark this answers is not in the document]
bryce-fitzsimons: you're right. pushing a fix to nuke all legacy keys when things change

packages/back-end/src/models/SdkConnectionCacheModel.ts
```diff
@@ -28,6 +32,13 @@ export class SdkConnectionCacheModel extends BaseClass {
...
   }
 
+  public async getById(id: string) {
+    return await this._findOne({
+      id,
+      schemaVersion: LATEST_SDK_PAYLOAD_SCHEMA_VERSION,
+    });
+  }
+
   public async upsert(
```
bryce-fitzsimons, line 41: @jdorn why do we `getById` and then fork `update()` and `create()`? Versus `update({}, {}, { upsert: true })`

packages/back-end/src/models/SdkConnectionCacheModel.ts
```diff
@@ -36,9 +47,11 @@ export class SdkConnectionCacheModel extends BaseClass {
     const existing = await this.getById(id);
     const updateData: {
       contents: string;
+      schemaVersion: number;
```
bryce-fitzsimons: wondering if we should skip `schemaVersion` for this lookup so we don't have orphaned docs from old versions

packages/back-end/src/services/features.ts
```diff
@@ -1018,60 +1026,7 @@ export async function getFeatureDefinitions({
   hashSecureAttributes,
   savedGroupReferencesEnabled,
 }: FeatureDefinitionArgs): Promise<FeatureDefinitionSDKPayload> {
```
jdorn, line 1033: We are still calling `getFeatureDefinitions` in a few places that we should switch to use `getFeatureDefinitionsWithCache`:

- `api/features/listFeatures.ts` - we have an SDK connection so should be straight forward
- `jobs/webhooks.ts` - these are legacy SDK webhooks, but they can still benefit from the cache

Also, to prevent other places from using this, we may want to move `getFeatureDefinitionsWithCache` to this file and then not export the underlying `getFeatureDefinitions` method.

bryce-fitzsimons: I missed these. Fixed. It ended up creating some additional complexity around legacy webhook cache key, but should be fine now.

I also stopped exporting `getFeatureDefinitions`, which had some cascading effects across existing test runners. Lot more mocks and scaffolding now, but I believe we're testing the correct layer finally.

packages/back-end/src/api/features/listFeatures.ts
```diff
@@ -90,16 +89,9 @@ export const listFeatures = createApiRequestHandler(listFeaturesValidator)(
       ) {
         throw new Error("Invalid SDK connection key");
       }
-      const payload = await getFeatureDefinitions({
+      const payload = await getFeatureDefinitionsWithCache({
         context: req.context,
-        capabilities: getConnectionSDKCapabilities(sdkConnection),
```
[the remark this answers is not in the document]
bryce-fitzsimons: good catch!

packages/back-end/src/models/SdkConnectionCacheModel.ts
```diff
@@ -4,11 +4,15 @@ import {
...
 
+// Increment this if we change the payload contents in a backwards-incompatible way
+export const LATEST_SDK_PAYLOAD_SCHEMA_VERSION = 1;
+
 const BaseClass = MakeModelClass({
   schema: sdkConnectionCacheValidator,
   collectionName: "sdkcache",
   idPrefix: "sdk-",
   globallyUniqueIds: true,
+  additionalIndexes: [{ fields: { id: 1, schemaVersion: 1 }, unique: true }],
```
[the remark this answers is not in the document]
bryce-fitzsimons: yeah. I guess we cannot use `globallyUniqueIds: true` anymore

packages/back-end/src/models/SdkConnectionCacheModel.ts
```diff
@@ -28,6 +28,10 @@ export class SdkConnectionCacheModel extends BaseClass {
     return true;
   }
 
+  public async getById(id: string) {
+    return await this._findOne({ id });
+  }
```
jdorn, line 39: Remove this.  Already provided by the base class

PR #5265, opened 2026-02-10
Add docs for Roku SDK

madhuchavva:
### Features 

Adding docs for Roku under "front-end" category - since Roku is primarily a streaming/TV platform and relies on Javascript based tooling.

image


docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+- ✅ Roku Streaming Stick (all versions)
+- ✅ Roku Express
+- ✅ Roku Premiere
+- ✅ Roku TV
+- ✅ Legacy devices (Roku 2, Roku 3)
+
+</TabItem>
+<TabItem value="js" label="js">
+
+The SDK is written in pure js and requires:
```
royalfig: I'm pretty sure this is referring to pure js as JavaScript, right? If so, let's uppercase it: JS. I think we usually say, too, vanilla JS if you mean not using any framework/lib.

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+- Features are loaded **once** when `init()` is called
+- **No automatic background refresh** is supported for now.
+- **Roku Limitation:** Unlike web SDKs, Roku does not support Server-Sent Events (SSE) for real-time streaming updates
+- To get updated features, you must reinitialize the GrowthBook instance (typically on app restart)
+- For real-time updates, implement a manual refresh mechanism:
+
+```js
+' Manual refresh example (requires creating new instance)
+sub RefreshFeatures()
+    oldAttributes = m.global.gb.attributes
```
royalfig, line 195: do we need to preserve other elements like trackingCallback?

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+})
+
+' Option 2: Offline mode
+gb = GrowthBook({
+    features: myFeatures,
+    attributes: { id: GetDeviceId() }
+})
+```
+
+### All Configuration Options
```
royalfig, line 285: Is this specifically calling out config options for offline mode? I'd maybe make that explicit if so.

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+```
+
+
+
+## Experimentation (A/B Testing)
+
+There is nothing special you have to do for feature flag experiments. Just evaluate the feature flag like you would normally do. If the user is put into an experiment as part of the feature flag, it will call the `trackingCallback` automatically in the background.
+
+```js
+result = m.global.gb.evalFeature("pricing-experiment")
```
royalfig: I think this is a little confusing, because you didn't mention `evalFeature` above as one of the main methods. 

In this example, I'd probably use one of those methods for the main example. 

Then, if you want to show how you can log out more info, I'd likely make a note about evalFeature being an advanced method for accessing additional info...

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+|---|---|---|
+| **Purpose** | Track experiment exposures | Track all feature evaluations |
+| **Frequency** | Once per experiment + user | Every feature check |
+| **Use Case** | Send to analytics for A/B testing | Debugging, usage monitoring |
+| **Fires For** | Experiments only | All features |
+| **De-duplication** | Yes (automatic) | No |
+
+
+
+**What Happens:**
```
royalfig: This is a little unclear:What's this referring to? Should it be under the next heading?

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+})
+
+' Console output:
+' [GrowthBook] Evaluating experiment: button-color-test
+' [GrowthBook] Hash input: button-color-test|user_12345
+' [GrowthBook] Hash value: 0.742
+' [GrowthBook] Assigned to variation 1 (red)
+```
+
+#### Hash Versions
```
royalfig: I don't think we should include this, right. Version 1 is deprecated and should only use v2, unless I'm missing something.

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+
+Version 2 provides more granular traffic allocation (e.g., 12.34% instead of 12.3%).
+
+```js
+' Hash version is set in the experiment rule (dashboard)
+' SDK automatically uses the correct version
+' No code changes needed
+```
+
+## Targeting & Advanced Features
```
royalfig: I think we cover targeting/attributes in other sections. Unless there's something specific to Roku in this section, I think you can omit the section altogether or just provide a link to the attributes/targeting section.

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+end for
+
+' ✅ GOOD: Check once, use many times
+maxQuality = gb.getFeatureValue("max-quality", "HD")
+for each video in videos
+    video.quality = maxQuality
+end for
+```
+
+### Development Mode
```
royalfig: You mentioned this earlier, too.

Also, does dev mode enable this verbose logging? With the other SDKs, it doesn't (if I remember correctly). I think there's a different `debug` param that enables it.

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+
+' Console output:
+' [GrowthBook] Loading features from: https://cdn.growthbook.io/...
+' [GrowthBook] Features loaded successfully: 42 features
+' [GrowthBook] Evaluating feature: new-ui
+' [GrowthBook] Condition matched: premium = true
+' [GrowthBook] Result: on = true, value = true, source = force
+```
+
+## Examples
```
royalfig: I don't think these examples need to be included here. You've already included some above.

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+else if layout = "B" then
+    ShowLayoutB()  ' 30% of users
+else if layout = "C" then
+    ShowLayoutC()  ' 20% of users
+end if
+```
+
+## Troubleshooting
+
+### Issue: SDK Not Loading
```
royalfig: I think the use of "Issue" and "Try the below" is a bit unnecessarily repetitive. Remove them and see if the section still makes sense

docs/docs/lib/roku.mdx
```diff
@@ -0,0 +1,1020 @@
...
+**Try the below:**
+1. Verify weights in GrowthBook dashboard match expectations
+2. Ensure weights array length matches variations count
+3. Test with multiple user IDs to verify distribution
+4. Check experiment is published and active
+
+## Limitations
+
+- ❌ No Server-Sent Events (SSE) streaming support (Roku limitation)
+- ❌ No Visual Editor experiments (SceneGraph only)
```
royalfig, line 758: What's scenegraph?

docs/docs/lib/roku.mdx
madhuchavva: This is a specific example talking about how users can instantiate the sdk instance over and over for refreshing features. trackingCallbacks are introduced later in the sections.

docs/docs/lib/roku.mdx
madhuchavva: We have a section for Offline Mode under Loading Features

docs/docs/lib/roku.mdx
madhuchavva: both ways are supported. I could remove this and add `getFeatureValue`

docs/docs/lib/roku.mdx
madhuchavva: good catch! section was misplaced.

docs/docs/lib/roku.mdx
madhuchavva: yeah. but i think it makes sense to do the verbose logging in the dev mode to help with debugging instead of expecting users to set two flags for it. Moreover, not all SDKs have `debug` setting. For eg: in Python, users can extend the logger instance and change it's level to see more logs.

docs/docs/lib/roku.mdx
madhuchavva: Scenegraph is XML based framework to design UI for roku apps.

packages/front-end/components/Features/SDKConnections/SDKLanguageLogo.tsx
```diff
@@ -83,6 +84,16 @@ export const languageMapping: Record<SDKLanguage, LanguageLogo> = {
...
+  roku: {
+    Icon: SiRoku,
+    color: "#4F01A3",
+    label: "Roku",
+    docs: "roku",
+    type: "frontend",
+    filters: ["browser"],
+    packageName: "growthbook-roku",
+    packageUrl: "https://www.npmjs.com/package/growthbook-roku",
+  },
```
madhuchavva, line 96: @royalfig Thanks for including this change. I'm about to create another PR for it :)

PR #5271, opened 2026-02-10
chore(build): Reduce runtime image size + misc improvements (take 2)

ahdriel:
Reverts #5270 which reverted #5250 (the original changes)

The root cause of the issue is that if `GB_ENABLE_PYTHON_DD_PROFILING` is enabled, we need to have `ddtrace-run` in the PATH, and because we were copying dependencies around, this was not the case anymore.

Using python venv and adding it to the PATH fixes this problem.

### Additional features and changes

* Use python virtualenv to copy bin folder instead of `--target`
* Use pnpm fetch as [recommended](https://pnpm.io/docker)
* Pinned python version to install `python3.11` instead of `python3`
* Changed some order of COPY and ENV to make layers more stable

### Original changes

* Image size is reduced by ~20% (3.1gb -> 2.5gb) 
* Number of packages installed from 202 to 112, which makes our images more secure and easier to maintain
* Simplify setup by using node:20-slim instead of python:slim. Installing python is easier than installing node.
* Do everything python related in the python stage, including pre-building deps and local wheel (before it was done during last stage)
* Add --no-cache-dir for all pip3 commands. Previously we were deleting the cache folder, now we skip it altogether.
* Add --no-install-recommends to all apt install. Skips nice to have packages, still install hard dependencies though.
* Add SHELL ["/bin/bash", "-o", "pipefail", "-c"] to ensure that we error out if any steps of pipe commands fail
* Pin ddtrace version
* Updated runtime from libkrb5-dev to libkrb5-3. Smaller and just what's needed for runtime.

### Testing

- [x] Docker image builds correctly
- [x] I can analyze results with a single container
- [x] Analysis run when GB_ENABLE_PYTHON_DD_PROFILING is set to true
- [x] `docker run --rm growthbook-local which ddtrace-run` returns the proper path


Dockerfile
```diff
@@ -121,29 +127,15 @@ RUN rm -f packages/front-end/tsconfig.json && \
...
-# Copy PM2 config file
-COPY ecosystem.config.js ./ecosystem.config.js
-
-# Copy yarn compatibility shim for users with custom entry points
-COPY bin/yarn ./bin/yarn
-RUN chmod +x ./bin/yarn
-ENV PATH="/usr/local/src/app/bin:${PATH}"
-
-# wildcard used to act as 'copy if exists'
+# Build metadata
```
tzjames, line 141:
```suggestion
# Verify required runtime entrypoints are installed and executable
RUN command -v python3 >/dev/null 2>&1 || (echo "ERROR: python3 is not installed or not in PATH!" && exit 1)
RUN command -v ddtrace-run >/dev/null 2>&1 || (echo "ERROR: ddtrace-run is not installed or not in PATH!" && exit 1)
RUN command -v pm2-runtime >/dev/null 2>&1 || (echo "ERROR: pm2-runtime is not installed or not in PATH!" && exit 1)

# Build metadata
```

PR #5276, opened 2026-02-11
Create API endpoints for Metric Groups

Kevin-Chant:
### Features and Changes

Adds the standard CRUD API endpoints for Metric Groups

### Testing

* Using your favorite tool, hit each of the new endpoints to check the functionality
* Also check that the docs generated properly

### Screenshots

image

image

image

### DIFF
--- a/packages/shared/src/validators/metric-group.ts
+++ b/packages/shared/src/validators/metric-group.ts
@@ -30,7 +30,7 @@ export const apiCreateMetricGroupBody = z.strictObject({
   projects: z.array(z.string()),
   metrics: z.array(z.string()),
   datasource: z.string(),
+  owner: z.string().optional().describe("Will default to the current user"),
+  archived: z.boolean().optional(),
 });
-export const apiUpdateMetricGroupBody = apiCreateMetricGroupBody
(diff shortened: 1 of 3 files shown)


packages/shared/src/validators/metric-group.ts
```diff
@@ -1,18 +1,36 @@
...
+  archived: z.boolean(),
+});
+
+export const apiCreateMetricGroupBody = z.strictObject({
+  name: z.string(),
+  description: z.string(),
+  tags: z.array(z.string()).optional(),
+  projects: z.array(z.string()),
+  metrics: z.array(z.string()),
+  datasource: z.string(),
```
lukesonnet, line 32: I would add `description`, `owner`, and `archived` as optional just to give full control unless there's reasons we are not exposing these in APIs.

In general I'd prefer the API to be as open and powerful as possible, but open to feedback here.

Kevin-Chant: Description is there, and I could add the other two. It's a bit odd to create a metric group that's already archived but if it's just an optional field then that's fine

PR #5281, opened 2026-02-11
Python: migrate from poetry to uv

vladenisov:
### Features and Changes

Migration to UV. Here is why UV better:
https://medium.com/@hitorunajp/poetry-vs-uv-which-python-package-manager-should-you-use-in-2025-4212cb5e0a14
https://www.loopwerk.io/articles/2024/python-poetry-vs-uv/

Although `pyproject.toml` looks different, everything has actually remained the same.

### Testing

Manually


jdorn:

Issue #5135, opened 2026-01-22
[Bug] Project Scoped permissions : Editing Fact Tables


natasha-growthbook:
### Summary

When leveraging project scoped permissions to edit the fact table, users with the Analyst and Experimenter role do not have access to edit the fact table as their global role is evaluated instead of their project scoped role.

### Replication Steps

Create a user with global role set to Collaborator and project role set to Analyst or Experimenter == Cannot edit fact tables in the project where they have Analyst/Experimenter role

<img width="774" height="132" alt="Image" src="https://github.com/user-attachments/assets/7d8a5f48-3c69-49ab-8fb0-3a8f5d5bab91" />

<img width="1452" height="807" alt="Image" src="https://github.com/user-attachments/assets/ab887091-3aa8-4e65-9f11-295325a5beb4" />

#### Expected Behavior

Project-scoped permissions takes precedent allowing users to edit fact tables
 
#### Current Behavior

Users will project scoped Analyst and Experimenter role are rendered permission error when editing fact tables

PR #5284, opened 2026-02-11
Bug fix for 5135

mknowlton89:
### Features and Changes

This PR fixes - https://github.com/growthbook/growthbook/issues/5135

#### Current Experience

> When leveraging project scoped permissions to edit the fact table, users with the Analyst and Experimenter role do not have access to edit the fact table as their global role is evaluated instead of their project scoped role.

#### Outline of the fix

When we're making the post request to create the fact table, the fact table is inheriting the data source's projects.

So if a data source is in All Projects, we'll throw an error because the projects array is empty (so we check the user's global role).

In most cases, it seems like when creating a resource from the index page, we default the projects list to the current project the user is viewing.

So, there are two potential fixes.

1. We adopt that same logic here - and if the user is viewing the Fact Table page from a project, we default the form to that project.
2. We will need to filter the list of data sources in the modal to only those where the user has permission for every project that the data source is a part of (or must have it globally, if the data source is in All Projects.

I believe option 1 is the correct approach - if the user has permission to create a fact table in a particular project, and the data source they're using is in All Projects that indicates to me that the data source is available for all projects.

- Closes https://github.com/growthbook/growthbook/issues/5135

### Dependencies

- None

### Testing

- [x] Ensure a user with a global `collaborator` role and an `Analyst` or `Experimenter` role for `Project B` can create a Fact Table in `Project 123` even if the Fact Table leverages a Data Source that is in `All Projects`
- [x] Ensure that the same user above CAN'T create a Fact Table on top of a Data Source that is only in `Project B`

### DIFF
--- a/packages/front-end/components/FactTables/FactTableModal.tsx
+++ b/packages/front-end/components/FactTables/FactTableModal.tsx
@@ -68,11 +68,7 @@ export default function FactTableModal({
       tags: existing?.tags || [],
       eventName: existing?.eventName || "",
       managedBy: existing?.managedBy || "",
-      projects: existing?.projects
-        ? existing.projects
-        : project
-          ? [project]
-          : [],
+      projects: existing?.projects || [],
       autoSliceUpdatesEnabled: existing?.autoSliceUpdatesEnabled ?? false,
     },
   });
@@ -169,7 +165,26 @@ export default function FactTableModal({
             const ds = getDatasourceById(value.datasource);
             if (!ds) throw new Error("Must select a valid data source");
 
+            let projects = ds.projects || [];
+
+            if (projects.length) {
+              // If the data source has projects, filter out any the user doesn't have permission to create fact tables in
+              projects = projects.filter((project) => {
+                return permissionsUtil.canCreateFactTable({
+                  projects: [project],
+                });
+              });
+            } else {
+              // If the data source is in all projects, check if the user has permission to create a fact table globally
+              if (permissionsUtil.canCreateFactTable({ projects: [] })) {
+                projects = []; // If the user does have global permissions, allow the fact table to be created in all projects
+              } else {
+                // If the user doesn't have global permission to create fact tables, use the project the user is in
+                projects = [project];
+              }
+            }
             value.columns = [];
+            value.projects = projects;
 
             const { factTable, error } = await apiCall<{
               factTable: FactTableInterface;


packages/front-end/components/FactTables/FactTableModal.tsx
```diff
@@ -68,7 +68,11 @@ export default function FactTableModal({
       tags: existing?.tags || [],
       eventName: existing?.eventName || "",
       managedBy: existing?.managedBy || "",
-      projects: existing?.projects || [],
+      projects: existing?.projects
+        ? existing.projects
+        : project
+          ? [project]
+          : [],
```
lukesonnet: Projects is required, so `existing?.projects` should only be missing for new fact tables. So we are essentially picking a default here. 

Since we used to fall back to the datasource projects if they exist, should we fall back to that here as well if it exists and if a user is in All Projects? (e.g. replace the last `: []` with `: ds.projects || []`?

Also, this is different from how fact metrics work, where we actually just pick the fact table's projects when creating a fact metric, regardless of a user's active projects. That flow doesn't have the bug this PR fixes because we only let you pick a fact table for a fact metric that you have access to in the first place.

I think that's ok, but it's something to consider. It almost feels like we should do something like:

```
existing?.projects 
  ?  existing.projects 
  : hasPermissionToCreateOnAll(ds.projects) 
    ? ds.projects 
    : project 
      ? [project]
      : []
```

Kevin-Chant: Personally I think it'd be better not to branch behavior on permissions; that could just end up with confusion if different users are seeing different results from the same action.

Inheriting from the ds projects when in All Projects seems reasonable, but would require a little more logic to support because the ds isn't fixed yet at this point. I'll leave that up to @mknowlton89

mknowlton89: Not to pour on here, but I do find it a bit odd that we're assigning a fact table projects based on the current project and/or the underlying data source's project(s), but we're not displaying that anywhere in the modal.

When creating other resources, like a feature, we show the `Projects` multi-select field. I sorta feel like if we do some of this custom logic, we should include that field here to be more transparent about what project(s) this fact table will be a part of.

And if the user changes the Data Source we can update the projects multi-select field with the new project(s).

lukesonnet: I'm fine with a selector.

I still think we should inherit from the ds projects as a reasonable fallback, which we used to do.

I don't think we should prevent user with permission on project X from creating a fact table just because the DS is in project X and Y, or is all projects.

I don't have strong feelings here beyond: I think we should unblock the above user creation flow (which is the origin of the original issue).

mknowlton89: Cool - I've updated the logic so that we do try to use the data source's projects, if possible.

Logic is:
- If the data source has projects, we start with that, and remove any projects from the list where the user doesn't have permission to create fact tables
- If the data source doesn't have projects, we check if the user has global permission to create fact tables, and if so, allow the fact table to be created in all proejcts
- If the data source doesn't have projects and the user doesn't have global permission, we fall back to the current project the user is in.

There is some baked in logic that should prevent bugs here. Like if a data source's project list includes only projects where the user doesn't have permission to create fact tables, that data source won't even show up in the data source options list.

And the `Create Fact Table` cta does check permission before we enable it.

And finally, if the user does get themselves into a situation where they're trying to create a fact table in a project/projects that they don't have permissions for, the API will still return the error, but that shouldn't be possible.

PR #5290, opened 2026-02-12
WIP: First attempt at moving variations into phases

lukesonnet:
Add `variations: z.array(variation).optional()` to the `phases` object in an experiment. The last phases variations serves as the new source of truth for the experiments variations, falling back to the experiment variations if they are missing.

Adds `disabled: z.boolean().optional()` to `variation` so that we can keep a variation metadata around but mark it as disabled so that it is ignored in payload construction and can be strategically shown or hidden when the phase is selected in the UI.

This means on experiment updates:
* We update the overall variations *and* the latest phase variations, if needed
* Whenever we get variation data, we check the latest phase and fallback to the experiment variation if need be.

I've also added the ability to remove a variation in this PR via Make Changes as a way to test this flow out. There UI there is a toy UI.

Other alternatives:
* We could instead just keep the experiment.variations list as a master list of all variations in their latest state and delete variations in each phase. This doesn't work well because we rely on variation indices within a phase, and we might want to keep the 0 vs 2 visual for later phases.
* We could instead just remove `experiment.variations` altogether in JIT migration and going forward. I think this could be fine, but might be a little riskier to merge? It also loses the notion of the "experiment level variation metadata" which can spawn phases and make it easy to compare a phase's variations to the global variations. Of course, for now, the global variations === latest phase variations, so maybe this will never diverge and this isn't a meaningful piece of data to keep around.


lukesonnet:
Closing this in favor of https://github.com/growthbook/growthbook/pull/5355 as this got a bit hairy. #5355 also uses the JIT migration and moves variations into phases entirely (option 2 listed in the summary above).

PR #5291, opened 2026-02-12
refactor(audit): Add audit logging to FactTables + refactor 

ahdriel:
### Features and Changes

- Centralize audit logging with `createModelAuditLogger`
  - Update BaseModel to use it
- Add logging for FactTables
- Migrate existing logs to use the new pattern
- Remove redundant calls from controllers / api endpoints

### DIFF
--- a/packages/back-end/src/models/SdkConnectionModel.ts
+++ b/packages/back-end/src/models/SdkConnectionModel.ts
@@ -425,7 +425,7 @@ export async function deleteSDKConnectionModel(
   sdkConnection: SDKConnectionInterface,
 ) {
   await SDKConnectionModel.deleteOne({
-    organization: sdkConnection.organization,
+    organization: context.org.id,
     id: sdkConnection.id,
   });
 


packages/back-end/src/services/audit.ts
```diff
@@ -98,3 +105,114 @@ export function auditDetailsDelete<T>(
...
   });
 }
+
+export type AuditLogConfig<Entity extends EntityType> = {
+  entity: Entity;
+  createEvent: EventTypes<Entity>;
+  updateEvent: EventTypes<Entity>;
+  deleteEvent: EventTypes<Entity>;
+  autocreateEvent?: EventTypes<Entity>;
+  omitDetails?: boolean;
```
lukesonnet, line 115: Previously it was more case-by-case whether to omit details or not due to the fact you called the full method each time. Is there any reason to allow details to be omitted differently for each event type?

I'm thinking that for updates it could be important, but not for create or deletes, in some cases?

But eh... probably not!

packages/back-end/src/models/FactTableModel.ts
```diff
@@ -289,6 +301,8 @@ export async function updateFactTable(
       },
     },
   );
+
+  await audit.logUpdate(context, factTable, { ...factTable, ...changes });
```
lukesonnet, line 305: Now that updates for columns can be async, this could fire twice per update, right?

packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -408,14 +420,16 @@ export const updateSdkConnectionsRemoveManagedBy = async (
...
-  id: string,
+export async function deleteSDKConnectionModel(
+  context: ReqContext,
+  sdkConnection: SDKConnectionInterface,
 ) {
   await SDKConnectionModel.deleteOne({
-    organization,
-    id,
+    organization: sdkConnection.organization,
+    id: sdkConnection.id,
```
Kevin-Chant:
```suggestion
    organization: context.org.id,
    id: sdkConnection.id,
```

packages/back-end/src/api/projects/deleteProject.ts
```diff
@@ -1,23 +1,13 @@
...
 import { createApiRequestHandler } from "back-end/src/util/handler";
-import { auditDetailsDelete } from "back-end/src/services/audit";
 
 export const deleteProject = createApiRequestHandler(deleteProjectValidator)(
   async (req): Promise<DeleteProjectResponse> => {
     const project = await req.context.models.projects.deleteById(req.params.id);
 
     if (!project) throw new Error("Could not find project!");
 
-    await req.audit({
```
lukesonnet, line 12: Is this the only redundant one? IDK how to easily check them all.

packages/back-end/src/models/FactTableModel.ts
ahdriel: I think so, yes. It would reflect that 2 separate updates were done to the model.

packages/back-end/src/services/audit.ts
ahdriel: I think we can add this in the future if we need it -- should not be hard.

Right now we are using it only for `dataSources` as that contains sensitive information.

ahdriel:
> Basically:
> 
> If not on the BaseModel, this PR shows how to use the same audit logging infrastructure to manually call audit events during a Model's created, update, and delete events with less overhead. Yes?

@lukesonnet You are correct. And we ensure that both BaseModel and mongoose models log it in the same way by using the same function.

packages/back-end/src/services/audit.ts
```diff
@@ -98,3 +105,114 @@ export function auditDetailsDelete<T>(
     context,
   });
 }
+
+export type AuditLogConfig<Entity extends EntityType> = {
+  entity: Entity;
+  createEvent: EventTypes<Entity>;
+  updateEvent: EventTypes<Entity>;
+  deleteEvent: EventTypes<Entity>;
```
Kevin-Chant, line 113: This might be worth changing to `events: Record<string, EventTypes<Entity>>` or similar so we can support entity types with more than just CUD

packages/back-end/src/jobs/createAutoGeneratedMetrics.ts
```diff
@@ -27,8 +25,7 @@ type CreateAutoGeneratedMetricsJob = Job<{
 const createAutoGeneratedMetrics = async (
   job: CreateAutoGeneratedMetricsJob,
 ) => {
-  const { datasourceId, organizationId, metricsToCreate, user } =
-    job.attrs.data;
+  const { datasourceId, organizationId, metricsToCreate } = job.attrs.data;
```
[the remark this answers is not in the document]
ahdriel: Ack'd -- we won't fix this right now as we have a TODO to ensure we add a more generic log of `user who triggered this async job` that should work for more scenarios than just this one.

PR #5292, opened 2026-02-13
WatchModel to BaseModel using generic primary keys

Kevin-Chant:
### Features and Changes

Fork of #5244 using the new generic primary key options

### Dependencies

This PR is based off #5293 not `main` so we should wait until that's merged and then review this against `main`

### Testing

If you've previously run #5244 locally, you'll need to delete any indices referencing ID
```
db.watches.getIndexes()
// db.watches.dropIndex("id_1_organization_1")
```
and then remove the IDs from the documents
```
db.watches.updateMany({}, {$unset: {id: true}})
```

After starting the server, check that watches work in the frontend:
- Add a new experiment to see that you're automatically enabled as a watcher
- Check that you can remove and re-add watching for it
- Same for a new feature within the organization
- Create a second user within the org and have them watch an existing feature/experiment

And then verify that the watch for the new user was created correctly + doesn't have an ID
```
db.watches.find({id: {$ne: null}})
```

The watch model already had an index on `userId/organization`, but we can verify that it would've been created automatically by dropping it and then restarting the server
```
db.watches.dropIndex("id_1_organization_1")
db.watches.getIndexes()
```

### DIFF
--- a/packages/back-end/src/api/experiments/postExperiment.ts
+++ b/packages/back-end/src/api/experiments/postExperiment.ts
@@ -16,7 +16,6 @@ import {
 } from "back-end/src/services/experiments";
 import { createApiRequestHandler } from "back-end/src/util/handler";
 import { getUserByEmail } from "back-end/src/models/UserModel";
-import { upsertWatch } from "back-end/src/models/WatchModel";
 import { getMetricMap } from "back-end/src/models/MetricModel";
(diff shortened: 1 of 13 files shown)


packages/back-end/src/models/BaseModel.ts
tzjames:
```suggestion
        "Cannot update id, uid, organization, dateCreated, or dateUpdated",
```

PR #5294, opened 2026-02-13
Add bulkWrite and countDocuments to BaseModel

Kevin-Chant:
### Features and Changes

Plucks the `bulkWrite` and `countDocuments` helpers from other in-flight branches to reduce the amount of unrelated code in each diff.

### Testing

Added unit tests for `dbSafeBulkWrite` to validate both codepaths, but these new helpers aren't being used yet (testing can mostly wait until the `WatchModel` migration)

### DIFF
--- a/packages/back-end/src/models/WebhookModel.ts
+++ b/packages/back-end/src/models/WebhookModel.ts
@@ -118,11 +118,8 @@ export class SdkWebhookModel extends BaseClass {
     return doc ? this.migrate(removeMongooseFields(doc)) : null;
   }
 
-  public async countSdkWebhooksByOrg(organization: string) {
-    return await this._dangerousGetCollection().countDocuments({
-      organization,
-      useSdkMode: true,
-    });
+  public async countSdkWebhooksByOrg() {
+    return await this._countDocuments({});
   }
 
   public getDefaultCreateProps(sdkConnectionId: string) {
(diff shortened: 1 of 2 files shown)


packages/back-end/src/models/BaseModel.ts
```diff
@@ -819,6 +814,52 @@ export abstract class BaseModel<
...
+  protected async _dangerousCountDocumentsCrossOrganization(
+    filter: ScopedFilterQuery<T>,
+  ) {
+    return this._dangerousGetCollection().countDocuments(filter);
+  }
+
+  protected async _countDocuments(filter: ScopedFilterQuery<T>) {
+    const query = this.applyBaseQuery(filter);
+    return this._dangerousGetCollection().countDocuments(query);
+  }
```
tzjames, line 826: I think Webhookmodel could end up using this?  Were you going to do that in this PR or another?

packages/back-end/src/models/BaseModel.ts
```diff
@@ -819,6 +814,52 @@ export abstract class BaseModel<
...
+
+  protected async _dangerousBulkWriteCrossOrganization(
+    operations: AnyBulkWriteOperation[],
+  ) {
+    return dbSafeBulkWrite(this._dangerousGetCollection(), operations);
+  }
+
+  protected async bulkWrite(operations: AnyBulkWriteOperation[]) {
```
tzjames, line 834: Similarly shall we test this out on MetricTimeSeries model?

Kevin-Chant: This one's a bit more hairy than `WebhookModel` was because it's passing an option I'm not sure about. I'm gonna leave it alone for now

PR #5296, opened 2026-02-13
Add Lookback Override to Analysis Settings

lukesonnet:
### Features and Changes

Adds Lookback Overrides to advanced analysis settings that behaves like a lookback override for all metric windows to a specific date or a specific number of hours in the past. This is really powerful if you want to create a custom report where you only look at the last 14 days of data across all metrics, or you want to apply this filter to your entire experiment perforce, kind of like we do with the Holdout's Analysis Period.

Does this by essentially overriding the Attribution Window selector so you can now pick: Respect Metric Windows, Ignore Conversion Windows, or Use Custom Lookback Window. It then pipes this through to a bunch of relevant places from the snapshot (e.g. if used in computing results) to ensure the context is preserved for this setting that changes important metric parameters.

Key changes in the BE are:
* Adding `"lookbackOverride"` to the `"attributionModel"`
* Adding `lookbackOverride?: LookbackOverride` to both the experiment analysis settings AND the snapshot settings (for ease of looking up the settings used in a given computation)
* Changing the phaseLookbackWindow for `getMetricForSnapshot` which actually applies this lookback window as a key override and takes precedence over any potential individual metric overrides.

Key changes in the FE:
* See screenshots below, mostly the analysis settings form and a few places where I call out this setting has been applied

### Testing

- [x] Works with no window metrics, conversion window metrics
- [x] works in reports
- [x] Correctly blends with retention (where lookback should not make retention window start sooner)
- [X] Preserves shorter lookback windows
- [X] Test future date
- [x] Rest API

### Screenshots

Setting it up
image

Screenshot 2026-02-13 at 12 58 58 AM

Screenshot 2026-02-13 at 12 58 51 AM
Screenshot 2026-02-13 at 12 59 10 AM
Screenshot 2026-02-13 at 12 58 54 AM

When it is on in an experiment
Screenshot 2026-02-13 at 12 59 24 AM

See "metric definition" (will hide the regular window if it exists)
Screenshot 2026-02-13 at 12 59 33 AM

Is Outdated works, comparing desired experiment settings to actual snapshot settings
Screenshot 2026-02-13 at 1 03 19 AM

Report setting
Screenshot 2026-02-13 at 1 03 57 AM

Report header
Screenshot 2026-02-13 at 1 03 51 AM

### DIFF
--- a/docs/src/partials/event-webhook/_event-webhook-list.md
+++ b/docs/src/partials/event-webhook/_event-webhook-list.md
@@ -1371,7 +1371,7 @@ Triggered when an experiment is created
                 inProgressConversions: "include" | "exclude";
                 /** Setting attribution model to `"experimentDuration"` is the same as selecting "Ignore Conversion Windows" for the Conversion Window Override. Setting it to `"lookbackOverride"` requires a `lookbackOverride` object to be provided. */
                 attributionModel: "firstExposure" | "experimentDuration" | "lookbackOverride";
-                /** Controls the lookback override for the experiment. If set, takes precedence over attributionModel setting. For type "window", value must be a non-negative number and valueUnit is required. */
+                /** Controls the lookback override for the experiment. For type "window", value must be a non-negative number and valueUnit is required. */
                 lookbackOverride?: {
                     type: "date" | "window";
                     /** For "window" type - non-negative numeric value (e.g. 7 for 7 days). For "date" type a date string. */
@@ -1545,7 +1545,7 @@ Triggered when an experiment is updated
                 inProgressConversions: "include" | "exclude";
                 /** Setting attribution model to `"experimentDuration"` is the same as selecting "Ignore Conversion Windows" for the Conversion Window Override. Setting it to `"lookbackOverride"` requires a `lookbackOverride` object to be provided. */
                 attributionModel: "firstExposure" | "experimentDuration" | "lookbackOverride";
-                /** Controls the lookback override for the experiment. If set, takes precedence over attributionModel setting. For type "window", value must be a non-negative number and valueUnit is required. */
+                /** Controls the lookback override for the experiment. For type "window", value must be a non-negative number and valueUnit is required. */
                 lookbackOverride?: {
                     type: "date" | "window";
                     /** For "window" type - non-negative numeric value (e.g. 7 for 7 days). For "date" type a date string. */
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
```diff
@@ -219,16 +224,6 @@ export default function ResultsTab({
...
-              {analysis?.settings?.statsEngine === "frequentist" ? (
-                <Metadata
-                  label="Sequential"
-                  value={
-                    analysis?.settings?.sequentialTesting
-                      ? "Enabled"
-                      : "Disabled"
-                  }
-                />
-              ) : null}
```
jdorn, line 231: I think we should revert this change. Showing sequential separately increases discoverability. Without that, there's no way to know that option even exists.

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -140,6 +140,15 @@ const experimentSchema = new mongoose.Schema({
       conversionDelayHours: Number,
     },
   ],
+  lookbackOverride: {
+    type: { type: String, enum: ["date", "window"] },
+    value: mongoose.Schema.Types.Mixed, // Date for "date" type, Number for "window" type
+    valueUnit: {
+      type: String,
+      enum: ["minutes", "hours", "days"],
+      default: "hours",
```
jdorn: Is this the right granularity? I would think days and weeks would be the only options people would actually want to use.

packages/front-end/components/Experiment/TabbedPage/ResultsTab.tsx
lukesonnet: I don't think this bar for a (not super commonly used) setting is the right place for discoverability, but I can see it being useful for new users who happen to switch to Frequentist. Ill revert it.

packages/back-end/src/models/ExperimentModel.ts
lukesonnet: Yeah it should be weeks instead of minutes.

packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -289,6 +296,12 @@ const AnalysisForm: FC<{
         if (experiment.status === "stopped") {
           body.phaseEndDate = dateEnded;
         }
+        // Include lookbackOverride; use undefined to clear when user selects "None"
+        if (value.lookbackOverride !== undefined) {
+          body.lookbackOverride = value.lookbackOverride;
+        } else if (experiment.lookbackOverride !== undefined) {
+          body.lookbackOverride = undefined;
+        }
```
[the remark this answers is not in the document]
lukesonnet: Fixed in new design: we don't care about clearing it, as you need both attributionWindow === "lookbackOverride" and `lookbackOverride` set for this to work.

packages/back-end/src/services/experiments.ts
```diff
@@ -1577,6 +1601,18 @@ export async function toExperimentApiInterface(
       queryFilter: experiment.queryFilter || "",
       inProgressConversions: experiment.skipPartialData ? "exclude" : "include",
       attributionModel: experiment.attributionModel || "firstExposure",
+      lookbackOverride: experiment.lookbackOverride
+        ? experiment.lookbackOverride.type === "date"
+          ? {
+              type: "date" as const,
+              value: experiment.lookbackOverride?.value.toISOString(),
+            }
```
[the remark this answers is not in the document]
lukesonnet: I am not too concerned about this? Maybe I should be, but we do the validation on entry to the DB.

PR #5299, opened 2026-02-13
fix(jobs): Add OTEL span for each job

ahdriel:
### Features and Changes

For exceptions that happen in our Jobs (managed by Agenda), our error tracking is treating all of them as a single trace because this is a long-running process.

This fixes it by manually creating spans for the job execution context.

- [x] Create span for each job execution
- [x] Add organizationId for all jobs that are associated with an organization
- [x] Add additional information to the span based on jobs attributes

#### Before

Note how the time is up to 6 minutes, showing how multiple jobs executions (this one is scheduled for every 5 minutes) are being aggregated on the same trace, making it hard to understand a per-job issues/errors.

Screenshot 2026-02-13 at 10 30 20 AM

#### After

See how we have 2 separate Trace IDs and they end in milliseconds.

Screenshot 2026-02-13 at 10 31 16 AM
Screenshot 2026-02-13 at 10 31 03 AM

### DIFF
--- a/packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
+++ b/packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
@@ -12,6 +12,7 @@ const UPDATE_SINGLE_SAFE_ROLLOUT_SNAPSHOT = "updateSingleSafeRolloutSnapshot";
 const QUEUE_SAFE_ROLLOUT_SNAPSHOT_UPDATES = "queueSafeRolloutSnapshotUpdates";
 
 type UpdateSingleSafeRolloutSnapshotJob = Job<{
+  organizationId: string;
   safeRollout: SafeRolloutInterface;
 }>;
 
@@ -45,6 +46,7 @@ export default async function (agenda: Agenda) {
     safeRollout: SafeRolloutInterface,
   ) {
     const job = agenda.create(UPDATE_SINGLE_SAFE_ROLLOUT_SNAPSHOT, {
+      organizationId: safeRollout.organization,
       safeRollout,
     });
     job.unique({ id: safeRollout.id });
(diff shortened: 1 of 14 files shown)


packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
```diff
@@ -56,12 +58,18 @@ export default async function (agenda: Agenda) {
...
   job: UpdateSingleSafeRolloutSnapshotJob,
 ) => {
-  const { safeRollout } = job.attrs.data;
+  const { safeRollout, organizationId } = job.attrs.data;
 
   const { id, organization, featureId } = safeRollout;
   if (!id || !organization || !featureId) return;
 
-  const context = await getContextForAgendaJobByOrgId(organization);
+  if (organizationId !== organization) {
```
jdorn, line 66: Any current scheduled jobs in Mongo are going to be missing `organizationId`, so we should only throw here if `organizationId && organizationId !== organization`

packages/back-end/src/jobs/createInformationSchema.ts
```diff
@@ -15,15 +15,15 @@ import { getContextForAgendaJobByOrgId } from "back-end/src/services/organizatio
 const CREATE_INFORMATION_SCHEMA_JOB_NAME = "createInformationSchema";
 type CreateInformationSchemaJob = Job<{
   datasourceId: string;
-  organization: string;
+  organizationId: string;
```
jdorn, line 18: I like the idea of making these keys consistent, but this is going to cause big migration issues.  Anything that is currently queued will fail / be corrupted, it will mess with Agenda's unique indexes, probably more things. Not worth it in my opinion.

packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
ahdriel: They will because we now are calling `migrateJobData` in the processor that happens before we call this function.

PR #5302, opened 2026-02-13
Smith/bandits no sticky bucketing

lukebrawleysmith:
### Features and Changes

Bandits can now be created and run without sticky bucketing. 

### Testing

- [x] Create a bandit with sticky bucketing disabled for the org
- [x] Create a bandit with sticky bucketing enabled for the org but disable sticky bucketing for the experiment
- [x] Edit Analysis Settings for a draft bandit
- [x] Edit Analysis Settings for a non-draft bandit (Should not be able to edit decision metric settings)
- [x] Create a bandit via a feature rule when sticky bucketing is disabled for the org
- [x] Create a bandit via a feature rule when sticky bucketing is enabled for the org
- [x] Link a bandit via a feature rule
- [x] Add Bandit from bandits page when sticky bucketing is disabled for org

### Screenshots
**Bandit Creation Flow Changes**
Screenshot 2026-02-27 at 3 03 41 PM

**Warning States**
Screenshot 2026-02-27 at 3 03 53 PM
Screenshot 2026-02-27 at 3 04 05 PM

**Analysis Settings changes**
Screenshot 2026-02-27 at 3 04 28 PM
Screenshot 2026-02-27 at 3 04 41 PM

To do:
- [x] More complete docs update; 
- [x] Include screenshots after FE changes are finalized

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -2191,9 +2191,7 @@ export default abstract class SqlIntegration
       SELECT
         e.${baseIdType} AS ${baseIdType}
         , ${
-          !!settings.banditSettings?.useFirstExposure &&
-          settings.banditSettings &&
-          settings.banditSettings.windowSettings
+          !!settings.banditSettings?.useFirstExposure && settings.banditSettings
             ? this.getFirstVariationValuePerUnit()
             : this.ifElse(
                 "count(distinct e.variation) > 1",


packages/back-end/src/services/experiments.ts
```diff
@@ -1057,6 +1060,7 @@ export async function createSnapshot({
     reweight,
     datasource,
     incrementalRefreshModel,
+    useStickyBucketing: organization.settings?.useStickyBucketing,
```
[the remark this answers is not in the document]
lukebrawleysmith: Addressed

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -326,12 +336,58 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
       banditScheduleUnit: scopedSettings.banditScheduleUnit.value,
       banditBurnInValue: scopedSettings.banditBurnInValue.value,
       banditBurnInUnit: scopedSettings.banditScheduleUnit.value,
+      banditConversionWindowValue: undefined,
+      banditConversionWindowUnit: "hours" as "hours" | "days",
```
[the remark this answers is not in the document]
lukebrawleysmith: Addressed

packages/front-end/components/Experiment/MultipleExposureWarning.tsx
```diff
@@ -15,12 +17,31 @@ const numberFormatter = new Intl.NumberFormat();
...
+    experimentType === "multi-armed-bandit";
+  if (isBandit) {
+    if (experiment) {
+      const orgStickyBucketing = !!settings?.useStickyBucketing;
+      const usingStickyBucketing =
+        orgStickyBucketing && !experiment.disableStickyBucketing;
+      if (!usingStickyBucketing) {
+        return null;
+      }
+    }
```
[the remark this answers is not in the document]
lukebrawleysmith: Addressed

packages/shared/types/experiment-snapshot.d.ts
```diff
@@ -199,6 +199,7 @@ export interface ExperimentSnapshotSettings {
   variations: SnapshotSettingsVariation[];
   coverage?: number;
   banditSettings?: SnapshotBanditSettings;
+  useStickyBucketing?: boolean;
```
lukesonnet: I don't think `useStickyBucketing` is the right name here. Since really it's a more generic "useFirstExposureOnly" or "ignoreMultipleExposures" kind of flag. 

And maybe it belongs in `banditSettings` since we don't plan to support this exact change for non-bandits in the near term?

packages/shared/src/validators/safe-rollout-snapshot.ts
```diff
@@ -170,6 +170,7 @@ const safeRolloutSnapshotSettings = z.object({
   coverage: z.number().optional(),
   phase: z.object({ index: z.string() }).optional(),
   customFields: z.record(z.string(), z.unknown()).optional(),
+  useStickyBucketing: z.boolean().optional(),
```
lukesonnet: If we move the setting into bandit settings (https://github.com/growthbook/growthbook/pull/5302/changes#r2828980670), then this won't be needed, right?

packages/front-end/pages/bandits/index.tsx
```diff
@@ -167,21 +167,14 @@ const ExperimentsPage = (): React.ReactElement => {
...
                 >
                   <Button
                     onClick={() => {
                       setOpenNewExperimentModal(true);
                     }}
                     disabled={
-                      !hasMultiArmedBanditFeature ||
-                      !hasStickyBucketFeature ||
-                      !orgStickyBucketing
+                      !hasMultiArmedBanditFeature || !hasStickyBucketFeature
```
lukesonnet: Shouldn't we remove `!hasStickyBucketFeature`?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1745,6 +1746,20 @@ export default abstract class SqlIntegration
     throw new Error("Unknown dimension type: " + (dimension as Dimension).type);
   }
 
+  private getFirstVariationValuePerUnit() {
+    return `SUBSTRING(
+        MIN(
+          CONCAT(SUBSTRING(${this.formatDateTimeString("e.timestamp")}, 1, 19), 
+            coalesce(${this.castToString(
+              `e.variation`,
+            )}, ${this.castToString(`'${NULL_VARIATION_VALUE}'`)})
```
lukesonnet, line 1771: I see why you did this with `NULL_VARIATION_VALUE`. I'm not sure what happens now if a variation is all `null` in an EAQ. Do we show it as "unknown variation found: `null`" in the UI or something? I'd want to just preserve that logic, but this is probably also fine.

packages/back-end/src/services/experiments.ts
```diff
@@ -2692,6 +2700,22 @@ export function postExperimentApiPayloadToInterface(
         settings,
       }),
     );
+    // Preserve conversion window fields from payload if provided
+    if (
+      "banditConversionWindowValue" in payload &&
+      payload.banditConversionWindowValue !== undefined
+    ) {
+      obj.banditConversionWindowValue =
+        payload.banditConversionWindowValue as number;
```
lukesonnet: Are all of these `as` statements needed? I would think something like

```
obj.banditConversionWindowValue =
        payload.banditConversionWindowValue
```
Would be sufficient given the types generated by the yaml files, but maybe I'm mistaken.

packages/back-end/src/services/experiments.ts
```diff
@@ -2847,6 +2875,18 @@ export function updateExperimentApiPayloadToInterface(
     ...(customMetricSlices !== undefined ? { customMetricSlices } : {}),
     ...(customFields !== undefined ? { customFields } : {}),
     ...(autoRefresh !== undefined ? { autoSnapshots: !!autoRefresh } : {}),
+    ...(banditScheduleValue !== undefined ? { banditScheduleValue } : {}),
+    ...(banditScheduleUnit !== undefined ? { banditScheduleUnit } : {}),
+    ...(banditBurnInValue !== undefined ? { banditBurnInValue } : {}),
+    ...(banditBurnInUnit !== undefined ? { banditBurnInUnit } : {}),
+    ...("banditConversionWindowValue" in payload &&
```
lukesonnet: Again is the "in payload" part necessary? I don't see a difference to these other settings in the yaml file.

packages/back-end/src/services/experiments.ts
lukesonnet, line 1090: I would expect getSnapshotSettings to take the conversion window override. I don't see where `banditConversionWindowValue` actually ever gets used.

packages/shared/types/experiment-snapshot.d.ts
lukebrawleysmith: done

packages/shared/src/validators/safe-rollout-snapshot.ts
lukebrawleysmith: correct

packages/front-end/pages/bandits/index.tsx
lukebrawleysmith: yes, good catch, done

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: In a bandit it shows as there being no data, under both current code and in this branch. 

Screenshot 2026-02-19 at 1 33 45 PM

packages/back-end/src/services/experiments.ts
lukebrawleysmith: removed

packages/back-end/src/services/experiments.ts
lukebrawleysmith: removed the redundant check

packages/back-end/src/services/experiments.ts
lukebrawleysmith: I missed this, and fixed it.  I confirmed in a bandit that the conversion window specified in bandit creation is being used, not the metric-specific conversion window.

packages/back-end/src/services/experiments.ts
```diff
@@ -72,7 +72,11 @@ import {
   ApiExperimentResults,
   ApiMetric,
 } from "shared/types/openapi";
-import { MetricPriorSettings } from "shared/types/fact-table";
+import {
+  MetricPriorSettings,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
lukesonnet: remove

packages/back-end/src/services/experiments.ts
```diff
@@ -2692,6 +2715,13 @@ export function postExperimentApiPayloadToInterface(
         settings,
       }),
     );
+    // Preserve conversion window fields from payload if provided
+    if (payload.banditConversionWindowValue !== undefined) {
+      obj.banditConversionWindowValue = payload.banditConversionWindowValue;
+    }
+    if (payload.banditConversionWindowUnit !== undefined) {
+      obj.banditConversionWindowUnit = payload.banditConversionWindowUnit;
+    }
```
lukesonnet, line 2818: I would add a check that if one exists the other should also exist.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1745,22 +1746,45 @@ export default abstract class SqlIntegration
...
+          )
+        ),
+        20, 
+        99999
+      )`;
+  }
+
   private getConversionWindowClause(
```
lukesonnet, line 1779: This adds the right clause to the CASE WHEN, but we use the conversion windows elsewhere in SQL Integration. Should we be overriding the setting when we pick the MetricSettingsForSnapshot so that it flows through to the entire SQL Integration file?

For example `getMetricEnd` and `getMaxHoursToConvert` also use this setting.

Maybe what we should do is instead of doing overrides here, we should go to the construction of the metric settings for snapshot in the getSnapshotSettings function and override the default metrics windows with the ones from the bandit if they exist (and leave the other windows untouched for now).

I would still leave it in `banditSettings` for posterity, but I think it should also be applied there.

packages/back-end/src/services/experiments.ts
lukebrawleysmith: done

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: Updated.  I imagine that we similar to how we want the back end to be smart and the stats engine to be dumb, we also want the SQL engine to be dumb?

packages/back-end/src/services/experiments.ts
lukebrawleysmith: done

packages/back-end/src/integrations/SqlIntegration.ts
lukesonnet: Yes, we don't want to put as much logic about why the window is the way it is into sqlintegration, just what window to use for what metric.

packages/back-end/src/services/reports.ts
lukesonnet, line 355: This method is also used in custom reports, do we allow those for bandits? If so, do we need to pass in these settings for custom reports as well?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2907,6 +2912,8 @@ export async function createExperimentSnapshot({
       experimentMetricOverrides: experiment.metricOverrides,
       datasourceType: datasource?.type,
       hasRegressionAdjustmentFeature: true,
+      banditConversionWindowValue: experiment.banditConversionWindowValue,
```
lukesonnet: To be safe, should we include a check that experiment is a bandit, and if so pass it, otherwise pass nothing?

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -341,12 +345,58 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+        );
+        form.setValue("banditConversionWindowUnit", "hours");
+      }
+    } else if (!goalMetricId) {
+      // If no goal metric, set to 1 hour
+      form.setValue("banditConversionWindowValue", 1);
+      form.setValue("banditConversionWindowUnit", "hours");
+    }
+  }, [goalMetricId, defaultConversionWindowHours, form]);
+
```
msamper: move back to `BanditRefNewFields` where all the bandit form logic is contained

packages/front-end/components/GeneralSettings/BanditSettings.tsx
```diff
@@ -34,19 +35,21 @@ export default function BanditSettings({
...
+            <Heading size="medium" as="h4">
               Bandit Settings
             </Heading>
           </Box>
         )}
         <Box
           className={clsx({
-            "w-100": page === "org-settings",
-            "col mb-2": page === "experiment-settings",
+            col: page === "experiment-settings",
```
msamper: update to use radix grid instead of bootstrap

packages/front-end/components/GeneralSettings/BanditSettings.tsx
```diff
@@ -110,6 +113,7 @@ export default function BanditSettings({
                       },
                     ]}
                     disabled={!hasBandits || lockExploratoryStage}
+                    style={{ width: 90, minWidth: 90 }}
```
msamper, line 109: apply to all conversion window inputs or consider making a conversion window input component

packages/front-end/ui/Checkbox.tsx
msamper, line 76: @ahdriel do we need to support refs for our `Text` component? I was going to migrate `Text` here but our `Text` component doesn't support refs and I'm honestly not sure what the use case is for them here since we don't have examples of a checkbox where we forward a ref

ahdriel: I think this is a good callout and we should add `ref` support to the `ui/Text` component! Gap in the design.

packages/front-end/components/Experiment/MultipleExposureWarning.tsx
```diff
@@ -15,12 +17,39 @@ const numberFormatter = new Intl.NumberFormat();
...
+      }
+    } else {
+      // If experiment object is not provided, we can't determine if sticky
+      // bucketing is enabled on the experiment. To avoid showing warnings for
+      // non-sticky-bucketing bandits, we hide the warning when we can't confirm.
+      // If org sticky bucketing is not enabled, the bandit definitely doesn't use it.
+      // If org sticky bucketing is enabled but we don't have the experiment object,
+      // we can't know if disableStickyBucketing is true, so we hide to be safe.
+      return null;
+    }
```
msamper: @lukebrawleysmith Seems like this logic will always not show the multiple exposures warning when we don't have the `experiment` object for a bandit and theres one case where don't have the experiment object `CompactResults` and one case where we do `BanditSummaryResultsTab`. Might be cleaner to just have this logic in those two components respectively. If we want to keep the logic here, then `experimentType` can be removed as a prop since it doesn't really have a purpose if `experiment` isn't passed in

msamper: Side note: do we need to update the multiple exposures block in the health tab with this logic as well?

packages/front-end/components/Features/RuleModal/BanditRefNewFields.tsx
```diff
@@ -81,6 +86,9 @@ export default function BanditRefNewFields({
   setWeight: (i: number, w: number) => void;
   variations: SortableVariation[];
   setVariations: (v: SortableVariation[]) => void;
+  /** When true, use metric default conversion window; payload should omit banditConversionWindowValue/Unit */
```
msamper: delete

packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -130,6 +130,8 @@ export default function RuleModal({
 
   const [allowDuplicateTrackingKey, setAllowDuplicateTrackingKey] =
     useState(false);
+  const [disableBanditConversionWindow, setDisableBanditConversionWindow] =
```
msamper, line 133: This might need to be initialized in the same way we initialize the version of this in `NewExperimentForm`. I'll double check

packages/shared/src/validators/features.ts
```diff
@@ -121,6 +121,8 @@ const experimentRule = baseRule
     banditScheduleUnit: z.enum(["hours", "days"]).optional(),
     banditBurnInValue: z.number().optional(),
     banditBurnInUnit: z.enum(["hours", "days"]).optional(),
+    banditConversionWindowValue: z.number().optional(),
+    banditConversionWindowUnit: z.enum(["hours", "days"]).optional(),
```
msamper: Make nullable

packages/front-end/components/Features/RuleModal/index.tsx
msamper: Can always be initialized to false. Bandits force you to go edit analysis settings on the bandit page rather than the rule itself

packages/front-end/components/GeneralSettings/BanditSettings.tsx
lukebrawleysmith: accepted thank you

packages/front-end/components/Experiment/MultipleExposureWarning.tsx
lukesonnet: >Side note: do we need to update the multiple exposures block in the health tab with this logic as well?

Yes, or we want to state that it is disabled for Bandits without Sticky Bucketing. For now I would just hide it.

packages/front-end/components/Experiment/BanditDecisionMetricSettings.tsx
```diff
@@ -0,0 +1,219 @@
...
+            !goalMetricWindow?.windowValue && (
+              <Callout status="warning" my="4">
+                Disabling the conversion window may bias results if units switch
+                variations during the experiment.
+              </Callout>
+            )}
+          {showConversionWindowWarning && !disabled && (
+            <Callout status="warning" my="4">
+              <Text>
+                To prevent counting conversions after a unit may have switched
```
lukesonnet: Limit how many conversions are counted after a unit may have switched variations by decreasing the metric conversion window to be <= 10% of the Update Cadence.

packages/shared/src/experiments.ts
```diff
@@ -859,6 +861,8 @@ export function getAllMetricSettingsForSnapshot({
   experimentMetricOverrides?: MetricOverride[];
   datasourceType?: DataSourceInterfaceWithParams["type"];
   hasRegressionAdjustmentFeature: boolean;
+  banditConversionWindowValue?: number;
```
lukesonnet: These aren't used in this method at all.

packages/shared/src/experiments.ts
```diff
@@ -692,6 +692,8 @@ export function getMetricSnapshotSettings<T extends ExperimentMetricInterface>({
   experimentRegressionAdjustmentEnabled: boolean;
   organizationSettings?: Partial<OrganizationSettings>; // can be RA and prior settings from a snapshot of org settings
   metricOverrides?: MetricOverride[];
+  banditConversionWindowValue?: number;
```
lukesonnet: Again these don't seem to be used at all unless I'm missing something.

packages/back-end/src/services/experiments.ts
```diff
@@ -631,6 +637,18 @@ export function getSnapshotSettings({
                     0,
                   ) ?? 0,
               })) ?? [],
+          useFirstExposure: useStickyBucketing,
+          windowSettings:
+            experiment.banditConversionWindowValue != null &&
```
lukesonnet: Shouldn't these checks be `!!experiment.banditConversionWindowValue` since the value will be undefined in some cases and then when it is used in `generateWindowSettings` we will override the conversion window with `undefined` values?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2175,11 +2190,15 @@ export default abstract class SqlIntegration
       -- One row per user
       SELECT
         e.${baseIdType} AS ${baseIdType}
-        , ${this.ifElse(
-          "count(distinct e.variation) > 1",
-          "'__multiple__'",
-          "max(e.variation)",
-        )} AS variation
+        , ${
+          !settings.banditSettings?.useFirstExposure && settings.banditSettings
```
lukesonnet: Isn't this logic backwards?

`!settings.banditSettings?.useFirstExposure` should read: `settings.banditSettings?.useFirstExposure`, no?

lukesonnet: I think I'm wrong and the logic is right, but the language is confusing.

I thought `useFirstExposure` referred to the case where we were not using sticky bucketing so we were always just relying on the first exposure, but it is true when we `useStickyBucketing`.

packages/back-end/src/services/experiments.ts
msamper: `!=` would capture both `undefined` and `null` since it's loose inequality but happy to change it to `!!experiment.banditConversionWindowValue` to make it more readable

lukesonnet: Whatever you think best.

packages/shared/src/experiments.ts
lukebrawleysmith: removed

packages/shared/src/experiments.ts
lukebrawleysmith: removed

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: I agree that the language is confusing.  I changed two booleans, which should address.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2175,11 +2190,15 @@ export default abstract class SqlIntegration
       -- One row per user
       SELECT
         e.${baseIdType} AS ${baseIdType}
-        , ${this.ifElse(
-          "count(distinct e.variation) > 1",
-          "'__multiple__'",
-          "max(e.variation)",
-        )} AS variation
+        , ${
+          !!settings.banditSettings?.useFirstExposure && settings.banditSettings
```
msamper: Why do we have the `&& settings.banditSettings` here? Should we be checking for `settings.banditSettings.windowSettings`?

packages/back-end/src/services/experiments.ts
lukebrawleysmith: leaving it as `!!experiment.banditConversionWindowValue` for readability

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: good call, added check for `settings.banditSettings.windowSettings`

lukebrawleysmith: yes, updated

lukesonnet: I think this is incorrect. If for some reason `useFirstExposure` is true, we just want to fork if it's a bandit, not if it's a bandit with a custom conversion window.

`!!settings.banditSettings?.useFirstExposure`

msamper (approved):
Lgtm! 🚀

PR #5303, opened 2026-02-14
feat(datasource): Attribute ↔ Identifier Type Mapping

msamper:
### Features and Changes

Allow users to specify attributes that map to a given identifier type. When creating an experiment, we now use this mapping to pick the first matching exposure query (if available).

### Testing

- [ ] Add attributes to an identifier type for a datasource and make sure selecting one of the attributes correctly maps to the exposure query with that identifier type
- [ ] Make sure the initial hash attribute and initial exposure query match (assuming the default datasource is the datasource you defined the mappings in)
- [ ] Switch between an attribute linked to an identifier type and an attribute not linked to an identifier type. The exposure query should not change

### Screenshots

Screenshot 2026-02-13 at 5 27 45 PM
Screenshot 2026-02-13 at 5 28 13 PM
Screenshot 2026-02-13 at 5 28 23 PM

### DIFF
--- a/packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
+++ b/packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
@@ -187,10 +187,10 @@ export default function ExperimentRefNewFields({
     );
     if (hashAttributeToIdentifierTypeMap.size > 0) {
       const matches =
-        (matchHashAttribute?.length ?? 0) > 0
+        matchHashAttribute && matchHashAttribute.length > 0
           ? {
               label: "Matches Hash Attribute",
-              options: matchHashAttribute!.map((q) => {
+              options: matchHashAttribute.map((q) => {
                 return {
                   label: q.name,
                   value: q.id,
@@ -200,10 +200,10 @@ export default function ExperimentRefNewFields({
           : null;
 
       const doesNotMatch =
-        (remainingExposureQueries?.length ?? 0) > 0
+        remainingExposureQueries && remainingExposureQueries.length > 0
           ? {
               label: "Does Not Match Hash Attribute",
-              options: remainingExposureQueries!.map((q) => {
+              options: remainingExposureQueries.map((q) => {
                 return {
                   label: q.name,
                   value: q.id,


packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
```diff
@@ -290,9 +350,23 @@ export default function ExperimentRefNewFields({
               options={attributeSchema
                 .filter((s) => !hasHashAttributes || s.hashAttribute)
                 .map((s) => ({ label: s.property, value: s.property }))}
-              value={form.watch("hashAttribute")}
+              value={hashAttribute}
               onChange={(v) => {
                 form.setValue("hashAttribute", v);
+                // Try and find a matching exposure query for the new hash attribute
+                const userIdType = datasource?.settings?.userIdTypes?.find(
```
lukesonnet: nit: Should we pull this out to a standalone function just to make the purpose clearer?

packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
```diff
@@ -155,6 +160,61 @@ export default function ExperimentRefNewFields({
...
+  const groupedExposureQueries: (GroupedValue | SingleValue)[] = useMemo(() => {
+    const matchHashAttribute = exposureQueries?.filter((q) => {
+      return (
+        q.userIdType === hashAttributeToIdentifierTypeMap.get(hashAttribute)
+      );
+    });
+    const remainingExposureQueries = exposureQueries?.filter(
+      (q) => !matchHashAttribute?.includes(q),
+    );
+    if (matchHashAttribute?.length) {
```
lukesonnet: I also wonder if we should show these groups if *any* mapping exists in the datasource? e.g. if you have set it up for one hash attribute but not this one, maybe we should show the grouping?

Maybe this should then just be whether  `hashAttributeToIdentifierTypeMap` has any size?

IDK, don't feel strongly about this and probably better ways to encourage adoption.

packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
```diff
@@ -465,12 +539,7 @@ export default function ExperimentRefNewFields({
                 value={form.watch("exposureQueryId") ?? ""}
                 onChange={(v) => form.setValue("exposureQueryId", v)}
                 required
-                options={exposureQueries?.map((q) => {
-                  return {
-                    label: q.name,
-                    value: q.id,
-                  };
-                })}
+                options={groupedExposureQueries}
```
lukesonnet, line 565: Should `sort` be false in this selector so we can control the order and put the matching group first?

packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
```diff
@@ -290,9 +350,23 @@ export default function ExperimentRefNewFields({
...
+                  (t) => t.attributes?.includes(v),
+                )?.userIdType;
+                if (userIdType) {
+                  const exposureQueryId = getExposureQuery(
+                    datasource?.settings,
+                    "",
+                    userIdType,
+                  )?.id;
+                  if (exposureQueryId) {
+                    form.setValue("exposureQueryId", exposureQueryId);
```
lukesonnet: I wonder if there's a way to alert people we are doing this in the background, in case they've already set the EAQ, but I think we shouldn't worry about this for now because we're going to update the flow here eventually, and if you go Next through the modal, you'll still technically need to review the EAQ.

packages/front-end/components/Features/RuleModal/ExperimentRefNewFields.tsx
```diff
@@ -155,6 +160,61 @@ export default function ExperimentRefNewFields({
   const hasHashAttributes =
     attributeSchema.filter((x) => x.hashAttribute).length > 0;
 
+  const hashAttribute = form.watch("hashAttribute");
+
+  const hashAttributeToIdentifierTypeMap = useMemo(() => {
+    const attributeToIdentifierType = new Map<string, string>();
```
msamper: change to `<string, string[]>` since attributes can map to more than one

msamper: If this Map has entries, we want to show the groupings in the `SelectField`

PR #5311, opened 2026-02-17
Upgrade next to 16 and turbopack

tzjames:
### Features and Changes
Upgrades next to version 16.
- Update a few Sass files that complained
- Had to use react-jsx in tsconfig.json with `styled-jsx`.
- Added `baseline-browser-mapping` to get rid of new warnings.
- Updated `next/dynamic` to React's `lazy` to pass type-check.
- Made an import in shared/src/enterprise/sso to relative otherwise when `IS_CLOUD=true`, next 16 wasn't properly parsing the import.
- Removed next-env.d.ts from repo because it is generated and `pnpm dev` and `pnpm build` generate different versions of the file.
- Added special handling for the AceEditor (Ace had to be imported first before the others, then raw-loader was needed for the workers.)

### Testing
In preview and dev environment
- Go to a json feature and edit the schema, make a syntax error and see that it catches is. 
- Go to Sql Explorer and see that you can enter sql and run it.
- See that you can create a visualization from the run data.
- Go to Experiment Results and see results
- Create a presentation see it works
- Click around all the other pages and see no runtime errors.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -19,6 +19,9 @@ const compat = new FlatCompat({
   recommendedConfig: js.configs.recommended,
   allConfig: js.configs.all,
 });
+// Strip invalid "name" property from Next.js flat config (ESLint 9 rejects it)
+const { name: _nextName, ...nextRecommendedConfig } =
+  nextEslintPluginNext.configs.recommended;
 
 export default defineConfig([
   globalIgnores([
@@ -31,6 +34,7 @@ export default defineConfig([
     "packages/sdk-js/scripts",
     "**/*.tsbuildinfo",
   ]),
+  nextRecommendedConfig,
   {
     extends: fixupConfigRules(
       compat.extends(
@@ -41,7 +45,6 @@ export default defineConfig([
         "plugin:@typescript-eslint/eslint-recommended",
         "plugin:@typescript-eslint/recommended",
         "plugin:prettier/recommended",
-        "plugin:@next/eslint-plugin-next/recommended",
         "plugin:react-hooks/recommended",
       ),
     ),
@@ -50,7 +53,6 @@ export default defineConfig([
       react: fixupPluginRules(react),
       "@typescript-eslint": fixupPluginRules(typescriptEslint),
       prettier: fixupPluginRules(prettier),
-      "@next/next": fixupPluginRules(nextEslintPluginNext),
       "no-async-foreach": noAsyncForeach,
     },
 


packages/front-end/next.config.js
```diff
@@ -10,12 +10,34 @@ const cspHeader = `
 const nextConfig = {
   // We already run eslint and typescript in CI/CD
   // Disable here to speed up production builds
```
Kevin-Chant: Nit: remove unused comment
```suggestion
```

packages/front-end/package.json
```diff
@@ -76,7 +76,7 @@
     "lodash": "^4.17.23",
     "md5": "^2.3.0",
     "minisearch": "^7.2.0",
-    "next": "14.2.35",
+    "next": "^16.1.6",
```
ahdriel, line 79: question: Should we upgrade the `eslint next` package s to be on v16 as well?

packages/front-end/next.config.js
```diff
@@ -8,13 +8,30 @@ const cspHeader = `
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  // We already run eslint and typescript in CI/CD
-  // Disable here to speed up production builds
-  eslint: {
-    ignoreDuringBuilds: true,
+  turbopack: {
```
ahdriel, line 13: issue: On my local environment (not sure if it happens for you as well), I am seeing the following message during dev:

```
packages/front-end dev: ⚠ Warning: Next.js inferred your workspace root, but it may not be correct.
packages/front-end dev:  We detected multiple lockfiles and selected the directory of /Users/adrielvieira/package-lock.json as the root directory.
packages/front-end dev:  To silence this warning, set `turbopack.root` in your Next.js config, or consider removing one of the lockfiles if it's not needed.
packages/front-end dev:    See https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack#root-directory for more information.
```

Following the documentation and adding the following property inside `turbopack` object

```
    root: path.join(__dirname, "../.."),
```

ensure it resolves to the root of the monorepo.

Can we add it so we don't scan extra files?

.gitignore
```diff
@@ -48,3 +48,6 @@ charts/**/*.tgz
 
 # Preview environment generated files
 preview/Dockerfile.generated
+
+# Next.js generated (path varies by dev vs build)
+packages/front-end/next-env.d.ts
```
ahdriel: nit: should we update this to `next-env.d.ts` (or `**/next-end.d.ts` if we want to be explicit?).

Then we can also update the comment above.


```suggestion
# Next.js generated
next-env.d.ts
```

packages/front-end/package.json
tzjames: Good call.  Doing.

packages/front-end/next.config.js
tzjames: I don't have that warning for whatever reason.  I had it in a different attempt at some point, and it neither helped nor harmed the issue I was trying to solve.  I'll add it and see if the build and everything passes.

PR #5313, opened 2026-02-17
[HOTFIX] Filter out invalid metrics from experiment create/update

msamper:
### Features and Changes

#5316 

Hotfix to temporarily not throw an error when an unknown metric is passed in to experiment update/create while we figure out the underlying issue.
See https://growthbookapp.slack.com/archives/C05QYS2UFRD/p1770905452163769

### Testing

- [x] Create/Update an experiment with valid metrics. Should persist all metrics
- [x] Make a create/update request with a mix of valid/invalid metrics. Invalid metrics should be removed

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -1083,7 +1083,7 @@ export async function postExperiments(
   const experimentType = data.type ?? "standard";
   const holdoutId = data.holdoutId;
 
-  // TODO: Added as a hotfix. Remove when issue is resolved.
+  // TODO: Added as a hotfix. Remove when issue #5316 is fixed.
   // Filter out invalid metric ids from the data
   if (invalidMetricIds.length) {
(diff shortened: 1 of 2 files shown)


packages/back-end/src/controllers/experiments.ts
```diff
@@ -1393,14 +1421,34 @@ export async function postExperiment(
...
+          //   status: 403,
+          //   message: "Unknown metric: " + newMetricIds[i],
+          // });
+          // return;
         }
       }
     }
+    // Filter out invalid metric ids from the data
+    if (invalidMetricIds.length) {
+      data.goalMetrics = data.goalMetrics?.filter(
```
ahdriel, line 1436: question: why are we handling these differently? One we are filtering and then using that array while here we are modifying `data` directly?

msamper: No real reason. Updated them to match

PR #5321, opened 2026-02-18
delete npm from docker image and uninstall python build tools

tzjames:
### Features and Changes
There are often security vulnerabilities in our build tools. As the build tools are not actually used in the image, they are not true security vulnerabilities.  By deleting our build tools when we are done with them, we avoid having to track down these issues in the future.

### Testing
In preview env refresh the example experiment stats.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN \
     && poetry export -f requirements.txt --output requirements.txt \
     && pip install --no-cache-dir -r requirements.txt \
     && pip install --no-cache-dir dist/*.whl ddtrace==4.3.2 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -101,6 +103,7 @@ WORKDIR /usr/local/src/app
 RUN apt-get update && \
   apt-get install -y --no-install-recommends python${PYTHON_MAJOR} ca-certificates libkrb5-3 && \
   npm install -g pnpm@10.28.2 && \
+  rm -rf /usr/local/lib/node_modules/npm && \
```
Kevin-Chant, line 106: Nit: slightly cleaner to use `RUN npm uninstall -g npm` than directly deleting the files

Kevin-Chant (approved):
LGTM. I'll leave it for you to merge so we can roll back if something goes wrong. (I did test the preview env as well though)

PR #5325, opened 2026-02-18
fix(feature-diagnostics): Ensure sorting by timestamp works as expected

ahdriel:
### Features and Changes

Fix Feature Evaluation Diagnostics sorting to use numeric timestamps (`timestampSort`) instead of formatted timestamp strings.


msamper (approved):
Thanks for fixing!

PR #5329, opened 2026-02-19
Migrate TeamModel to BaseModel and add API endpoints

Kevin-Chant:
### Features and Changes

Refactors `TeamModel` to inherit from `BaseModel` and adds the standard CRUD + some custom API handlers

### Testing

On `main`, create a team in the settings -> members page
- [x] The existing team should still be fully functional
  - [x] Details render correctly
  - [x] Can edit permissions
  - [x] Can add members
  - [x] Can remove members
- [x] New teams should work in the UI
  - [x] Can create a new team
  - [x] Details render correctly
  - [x] Can edit permissions
  - [x] Can add members
  - [x] Can remove members
- [x] In mongo, the team documents should both look correct (`db.teams.find({})`)
- [x] The new API endpoints should behave as expected
  - [x] List teams
  - [x] Get team
  - [x] Create team
  - [x] Update team
  - [x] Delete team
  - [x] Add members to team
  - [x] Remove members from team

### DIFF
--- a/packages/back-end/src/models/BaseModel.ts
+++ b/packages/back-end/src/models/BaseModel.ts
@@ -314,7 +314,7 @@ export abstract class BaseModel<
   public async handleApiList(
     _req: ApiRequest<unknown, z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>,
   ): Promise<z.infer<ApiT>[]> {
-    return (await this.getAll()).map(this.toApiInterface);
+    return (await this.getAll()).map(this.toApiInterface.bind(this));
   }
   public async handleApiDelete(
     req: ApiRequest<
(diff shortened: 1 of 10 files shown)


packages/back-end/src/models/BaseModel.ts
```diff
@@ -619,10 +619,13 @@ export abstract class BaseModel<
       throw new Error("Cannot set dateUpdated field");
     }
 
-    // Add default owner if empty
+    // Add default owner and createdBy if empty
     if ("owner" in props && !props.owner) {
       props.owner = this.context.userName || "";
     }
+    if ("createdBy" in props && !props.createdBy) {
```
msamper, line 626: Is there a difference between `createdBy` and `owner`? Is the idea to have better auditing if someone changes the `owner`?

Kevin-Chant: I don't think there's a relevant difference, some models just aren't standardized on `owner` so I added this for compatibility

PR #5330, opened 2026-02-19
Vercel Pre-Install Auth Fix

madhuchavva:
## Problem

- GET /v1/products/:slug/plans (the "List Billing Plans" endpoint) calls authContext(), which:
     - Calls checkAuth() — uses systemAuthenticationValidator which requires installation_id: z.string() (non-optional). Zod parse fails → caught as "Invalid credentials" → 401.

Vercel calls this endpoint before an installation exists (pre-install phase), so the OIDC token has a null/missing installation_id.

## Fixes:

- Make `installation_id` and `account_id` optional in `systemAuthenticationValidator` to reflect that Vercel's spec allows these to be absent pre-install and add additional check in `authContext` to ensure every endpoint that routes through authContext (installations, resources, etc.) still requires a valid installation ID.

- Rewrite `getProducts`: 
     - Verifies the bearer token directly via jwtVerify against the Vercel JWKS (signature is still checked — no security bypass)
    - Returns availableBillingPlans immediately if installation_id is absent (pre-install phase)
    - If an installation_id is present, attempts the installation lookup and returns the appropriate "add project" plan, falling back to availableBillingPlans on any error

### DIFF
--- a/packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
+++ b/packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
@@ -137,7 +137,7 @@ const checkAuth = async <T extends string | "user">({
   type: T;
 }): Promise<CheckedAuth<T> | { status: "error"; message: string }> => {
   try {
-    const payload = await jwtVerify(token, await vercelJKWSKey);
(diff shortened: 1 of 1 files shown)


madhuchavva:
cursor review

PR #5343, opened 2026-02-20
fix(nav): Ensure we refetch data if `fid` changes

ahdriel:
### Features and Changes

For performance reasons we have special logic in the Feature page on how we handle revisions and when we fetch them.

This works as expected, but we had a bug where if the url param `fid` updated, we would not fetch new data. This worked fine on our last navigation patterns as you would never be able to navigate from feature to feature without going through a different page.

But with Cmd + K this is not true anymore, so if we used Cmd+K to navigate between Features, nothing would update in the UI even though the URL was updated.

This fixes it + uses `useApi` for fetching data.

### DIFF
--- a/packages/front-end/pages/features/[fid].tsx
+++ b/packages/front-end/pages/features/[fid].tsx
@@ -1,20 +1,6 @@
 import { useRouter } from "next/router";
 import { useEffect, useState, useMemo } from "react";
-import { FeatureInterface, FeatureRule } from "shared/types/feature";
-import { FeatureCodeRefsInterface } from "shared/types/code-refs";
-import { FeatureRevisionInterface } from "shared/types/feature-revision";
-import { ExperimentInterfaceStringDates } from "shared/types/experiment";
-import {
-  filterEnvironmentsByFeature,
-  getDependentExperiments,
-  getDependentFeatures,
-  mergeRevision,
-} from "shared/util";
-import {
-  SafeRolloutInterface,
-  HoldoutInterface,
-  MinimalFeatureRevisionInterface,
-} from "shared/validators";
+import { getDependentExperiments, getDependentFeatures } from "shared/util";
 import { FeatureEvalDiagnosticsQueryResponseRows } from "shared/types/integrations";
 import LoadingOverlay from "@/components/LoadingOverlay";
 import PageHead from "@/components/Layout/PageHead";
@@ -23,7 +9,7 @@ import { useLocalStorage } from "@/hooks/useLocalStorage";
 import FeaturesOverview from "@/components/Features/FeaturesOverview";
 import FeaturesStats from "@/components/Features/FeaturesStats";
 import useOrgSettings from "@/hooks/useOrgSettings";
-import { useEnvironments, useFeaturesList } from "@/services/features";
+import { useFeaturesList } from "@/services/features";
 import { FeatureUsageProvider } from "@/components/Features/FeatureUsageGraph";
 import FeatureTest from "@/components/Features/FeatureTest";
 import { useAuth } from "@/services/auth";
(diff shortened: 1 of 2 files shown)


packages/front-end/pages/features/[fid].tsx
```diff
@@ -301,7 +354,11 @@ export default function FeaturePage() {
   const dependents = dependentFeatures.length + dependentExperiments.length;
 
   if (error) {
-    return <div className="alert alert-danger">An error occurred: {error}</div>;
+    return (
+      <div className="alert alert-danger">
+        An error occurred: {error.message}
+      </div>
+    );
```
Comment (msamper) [inline packages/front-end/pages/features/[fid].tsx]:
```suggestion
      <Callout status="error">
        An error occurred: {error.message}
      </Callout>
    );
```
should we update this to use the Callout component?

packages/front-end/pages/features/[fid].tsx
```diff
@@ -43,42 +61,164 @@ export default function FeaturePage() {
...
+      return next;
+    });
+
+    setCachedSafeRolloutsById((prev) => {
+      const next = { ...prev };
+      baseData.safeRollouts.forEach((sr) => {
+        next[sr.id] = sr;
+      });
+      return next;
+    });
```
Comment (msamper) [inline packages/front-end/pages/features/[fid].tsx]: At this point aren't we starting from a blank slate because of the reset `useEffect`? What I'm wondering is why we're using the `prev` state to build the new one for each of these state variables

packages/front-end/pages/features/[fid].tsx
```diff
@@ -43,42 +61,164 @@ export default function FeaturePage() {
...
+      return;
+    }
+
+    setCachedRevisionsByVersion((prev) => {
+      const next = { ...prev };
+      selectedVersionData.revisions.forEach((r) => {
+        next[r.version] = r;
+      });
+      return next;
+    });
```
Comment (msamper) [inline packages/front-end/pages/features/[fid].tsx]: Might be helpful to explain why we're doing this since I was a bit confused reading this. Looking at the controller, it seems like we only fetch a minimal amount of revisions so when you select a version you might need to get that revision directly.

packages/front-end/pages/features/[fid].tsx
```diff
@@ -43,42 +61,164 @@ export default function FeaturePage() {
...
+    setCachedSafeRolloutsById((prev) => {
+      const next = { ...prev };
+      selectedVersionData.safeRollouts.forEach((sr) => {
+        next[sr.id] = sr;
+      });
+      return next;
+    });
```
Comment (msamper) [inline packages/front-end/pages/features/[fid].tsx]: I might be mistaken, but I don't think these change even if you request a specific version. It seems like we return all of the possible values for experiments and safe rollouts no matter what version you're requesting. So I'm thinking we can probably just memoize these and make the dependency array include `fid`

msamper (changes requested):
Left some comments on some potential improvements. I had some trouble reading parts of this, but I think if my comments make sense there's an opportunity to simplify the states here to make it more readable. Also tested and it seems to work now!

packages/front-end/pages/features/[fid].tsx
Comment (ahdriel) [reply packages/front-end/pages/features/[fid].tsx]: Good catch!

packages/front-end/pages/features/[fid].tsx
Comment (ahdriel) [reply packages/front-end/pages/features/[fid].tsx]: That's a good point! I mistakenly though this was running on the new revisions fetch, but actually the dependencies were just `fid` and `baseData` which should be in sync (mostly, pending data fetching latency).

There is a different part of the code that has this same logic and it uses prev to keep building the cache for each new fetched revision, but it does not apply here.

I have refactored the code and simplified this part.

packages/front-end/pages/features/[fid].tsx
Comment (ahdriel) [reply packages/front-end/pages/features/[fid].tsx]: You are absolutely right! I was following the previous pattern but we don't need to have this.

Simplified to just use the ones coming from the `baseFeature`! Thanks for the callout

msamper (approved):
Looks great!! Thanks for the refactor here, the hook makes the code a lot easier to understand 🙌🏼

PR #5346, opened 2026-02-20
fix(custom-fields): Improve validation + UI

ahdriel:
### Features and Changes

This work makes custom fields be enforced more consistently through the product. Custom Fields are user-defined fields that can be required, and be attached to either a Feature or a Experiment, and additionally, they can be project-scoped.

- Creating a feature/experiment now properly validates required custom fields and prevents the creation if they are missing
- Updating a feature/experiment is a bit more relaxed at the moment to allow users time to migrate, so in this case we will only validate if we are modifying any custom fields, but if the user is modifying different fields, we won't prevent them from doing so.
- Fixed UI to properly show/hide the appropriate custom fields based on the Project selected (or only global-scoped Custom Fields if applicable)
- `MultiSelect` now has a `required` prop
- Refactored the code to ensure validations are applied in a single way
- Updated boolean custom fields to render the same order as all the other custom fields (label then field)
  - Also fixed an issue about conflicting ids for the labels
- Updated handling of default values for CustomFields to support falsy values

### Testing

- Manual testing
- Added new unit tests


packages/front-end/hooks/useCustomFields.ts
```diff
@@ -11,26 +17,26 @@ export function filterCustomFieldsForSectionAndProject(
...
+    ...v,
+    projects: (v.projects ?? []).map((p) => p.trim()).filter(Boolean),
+  }));
+
+  // if no selected project, show only globally available custom fields
+  if (!normalizedProject) {
+    return normalizedCustomFields.filter((v) => v.projects.length === 0);
+  }
+
+  // if selected project: show global fields + project scoped fields.
```
bryce-fitzsimons, line 38: Check instances of `<CustomFieldInput>`. It seems that "create new feature" does not pass in a project here.

packages/back-end/src/util/custom-fields.ts
```diff
@@ -1,70 +1,210 @@
...
+      throw new Error(
+        `Invalid number value for custom field ${customField.id} (${normalizedValue}).`,
+      );
+    }
+  }
+
+  if (customField.type === "url") {
+    const normalizedValue = toStringValue(value).trim();
+    try {
+      new URL(normalizedValue);
```
bryce-fitzsimons, line 128: Consider not doing this validation. Folks might use url snippets / relative paths that may not conform to URL() validation

packages/front-end/hooks/useCustomFields.ts
ahdriel: Can you pinpoint me to where we are missing it? I have added to 'create new feature' in this PR ([link](https://github.com/growthbook/growthbook/pull/5346/changes#diff-95090b3a6f150c4c3e28d9b64e015f0be08800d3dedb6b4bd043c4d04aecf90bR390)).

packages/back-end/src/util/custom-fields.ts
ahdriel: We use `type: url` in the front-end UI though, meaning that our UI is enforcing a proper URL format.

Would you recommend that we also relaxed the UI validation or keep the mismatched behavior between API & UI?

ahdriel:
> > Creating a feature/experiment now properly validates required custom fields and prevents the creation if they are missing
> I thought you guys wanted to prevent starting the experiment (draft -> live), not block creation.

@bryce-fitzsimons that is still the plan, but this is not part of that work.

packages/front-end/components/CustomFields/CustomFieldInput.tsx
```diff
@@ -27,42 +27,84 @@ const CustomFieldInput: FC<{
...
     if (!loadedDefaults) {
       // here we are setting the default values in the form, otherwise
       // boolean/toggles or inputs with default values will not be saved.
       if (availableFields) {
+        const nextCustomFields = { ...normalizedCustomFields };
         availableFields.forEach((v) => {
-          if (!currentCustomFields?.[v.id] && v.defaultValue) {
+          const currentValue = nextCustomFields?.[v.id];
+          const missingCurrentValue =
+            currentValue === undefined ||
```
bryce-fitzsimons, line 52: for my own info: why do we need this manual sanitization?

packages/front-end/hooks/useCustomFields.ts
bryce-fitzsimons: whoops, i missed that. disregard

packages/back-end/src/util/custom-fields.ts
bryce-fitzsimons: if it's already validating on the FE, then I suppose we can validate it on the BE as well.

One complication that we haven't addressed (although my other PR attempts to resolve) is that if you change the type of a customfield, how does it affect existing models? It could cause all model updates to suddenly stop working if a string changed to url is no longer validating.

ahdriel: For updates, either via UI or via API, we are only validating (or at least that's my goal, but more testing is certainly welcomed!) custom fields if they are being modified. Meaning that even if they become technically invalid under the new configuration, we won't error out.

But if you change a custom field and then attempt to modify a separate custom field, we will error out.

This also manifests when the user modifies the Project a resource is associated with, and we might have required custom fields for the new Project that would need to be specified. Given that our UI currently does not support this pattern, and this would be a bigger breaking change -- I adopted the more relaxed 'validate only if changing' stance for now.

Agreed that we need to find a better solution for these 2 similar problems.

packages/front-end/components/CustomFields/CustomFieldInput.tsx
ahdriel: To be able to handle falsy values like 0.

I guess I could simplify and use `!= undefined` but decided to be explicit about what is considered empty.

PR #5351, opened 2026-02-23
Migrate AuditModel to BaseModel

Kevin-Chant:
### Features and Changes

Moves AuditModel to inherit from BaseModel and updates its usage across the app

The Zod schema was rather tricky due to Audits being a generic type, so it relies on `superRefine` for runtime validation and typecasting for simplicity at compile-time

### Testing

WIP

### DIFF
--- a/packages/back-end/src/models/AuditModel.ts
+++ b/packages/back-end/src/models/AuditModel.ts
@@ -1,233 +1,176 @@
-import mongoose, { FilterQuery, QueryOptions } from "mongoose";
-import uniqid from "uniqid";
 import { AuditInterface, EntityType } from "shared/types/audit";
-import {
-  removeMongooseFields,
-  ToInterface,
-} from "back-end/src/util/mongo.util";
+import { auditSchema } from "shared/validators";
+import { CreateProps } from "shared/types/base-model";
+import { getCollection } from "back-end/src/util/mongo.util";
+import { FindOptions, MakeModelClass, ScopedFilterQuery } from "./BaseModel";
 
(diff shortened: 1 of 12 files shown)


packages/back-end/src/models/AuditModel.ts
```diff
@@ -1,233 +1,176 @@
...
+  }
+  protected canRead(): boolean {
+    return true;
+  }
+  protected canUpdate(): boolean {
+    return true;
+  }
+  protected canDelete(): boolean {
+    return true;
+  }
```
jdorn, line 34: Should we always return `false` for canUpdate and canDelete? I think we want this collection to be append only so it could add an extra layer of protection against accidentally modifying or deleting old logs.

packages/back-end/src/models/AuditModel.ts
```diff
@@ -1,233 +1,176 @@
...
+        sort: { dateCreated: -1 },
+      },
+    );
+  }
+
+  public async findAuditByEntity(
+    type: EntityType,
+    id: string,
+    options?: FindOptions<AuditInterface>,
+    customFilter?: ScopedFilterQuery<typeof auditSchema>,
```
jdorn: I generally don't like accepting arbitrary Mongo queries in models. Makes it easier if all of the query logic lives within model classes instead of spread throughout the code base.

In this example, looks like the only use case is providing an optional maxDate, so this method should have a `maxDate?: Date` argument instead of a generic query.

Kevin-Chant:
Closing to start fresh as the branch is quite stale

PR #5352, opened 2026-02-23
Ty/attr features

teresayung:
### Features and Changes

- **Attribute tooltip on feature targeting**  
  When configuring “Target by Attributes” on a feature rule, the attribute dropdown now shows a **hover tooltip** for each attribute that has a description or tags.
  - **Content:** Tooltip shows the attribute’s description (rendered as Markdown, including clickable links) and tags as pill-style badges.
  - **Placement:** Tooltip appears to the **right** of the option (`side="right"`).
  - **Hover area:** The whole option row is hoverable to show the tooltip, not just the label text.
  - **Links:** `disableHoverableContent={false}` so the tooltip stays open when hovering over it, allowing users to click links in the description. Link colors in tooltips are styled for contrast in both light and dark themes (see Dependencies).

- **Files changed**
  - `packages/front-end/components/Features/ConditionInput.tsx` – Attribute options include `description` and `tags`; custom `formatOptionLabel` with Tooltip + Markdown; full-row trigger span.
  - `packages/front-end/styles/global-radix-overrides.scss` – Link styles inside `.rt-TooltipContent` for light theme (blue-4/3) and dark theme (blue-11/12) so links are readable on the tooltip background.

- Closes https://www.notion.so/growthbook/Attribute-discovery-metadata-2efa857e74a0802c9583ff62a5c90fd1?source=copy_link

### Dependencies

None.

### Testing

1. Open a feature and add or edit a rule that uses “Target by Attributes” (e.g. a rollout or experiment rule).
2. Open the attribute dropdown (the “field” selector in the condition row).
3. **Hover** an attribute that has a description and/or tags:
   - Tooltip should appear to the **right** with description and tags.
   - Hovering anywhere on that option row (not only the text) should show the tooltip.
4. If the description contains Markdown links:
   - Move the cursor onto the tooltip; it should stay open.
   - Click a link; it should open in a new tab and links should be clearly visible (blue on dark tooltip in light theme, dark blue on light tooltip in dark theme).
5. Attributes with no description or tags (and Saved Group options) should show no tooltip.
6. Toggle dark theme (if available) and confirm tooltip link colors still look correct.

### Screenshots

image
image

Detail Attribute Page
image

### DIFF
--- a/packages/front-end/components/Forms/SelectField.tsx
+++ b/packages/front-end/components/Forms/SelectField.tsx
@@ -1,10 +1,16 @@
-import { FC, useMemo, useRef, ReactNode, useState } from "react";
+import {
+  FC,
+  useMemo,
+  useRef,
+  ReactNode,
+  useState,
+  ComponentProps,
+} from "react";
 import ReactSelect, {
   components,
   InputProps,
   FormatOptionLabelMeta,
   StylesConfig,
-  MenuPortalProps,
 } from "react-select";
 import cloneDeep from "lodash/cloneDeep";
 import clsx from "clsx";
@@ -13,7 +19,7 @@ import { RadixTheme } from "@/services/RadixTheme";
 import Field, { FieldProps } from "./Field";
 
 export const RadixThemeMenuPortal = (
-  props: MenuPortalProps<unknown, boolean, never>,
+  props: ComponentProps<typeof components.MenuPortal>,
 ) => (
   <RadixTheme>
     <components.MenuPortal {...props} />
--- a/packages/front-end/test/components/Features/ConditionInput.test.tsx
+++ b/packages/front-end/test/components/Features/ConditionInput.test.tsx
@@ -281,7 +281,7 @@ describe("ConditionInput", () => {
     });
   });
 
-  it("shows tooltip with Data type and Identifier when attribute dropdown is opened and option is hovered", async () => {
+  it("shows tooltip with 'Type' and 'Identifier' when attribute dropdown is opened and option is hovered", async () => {
     render(
       <RadixTheme>
         <TooltipProvider>
@@ -320,7 +320,7 @@ describe("ConditionInput", () => {
     fireEvent.mouseEnter(userIdOptions[0]);
 
     await waitFor(() => {
-      expect(screen.getByText("Data type:")).toBeInTheDocument();
+      expect(screen.getByText("Type:")).toBeInTheDocument();
       expect(screen.getByText("Identifier")).toBeInTheDocument();
     });
   });
--- a/packages/front-end/test/setup.ts
+++ b/packages/front-end/test/setup.ts
@@ -1,7 +1,16 @@
 import * as matchers from "@testing-library/jest-dom/matchers";
-import { expect } from "vitest";
+import { expect, beforeEach } from "vitest";
 
 expect.extend(matchers);
+
+// Portal component requires a #portal-root element in the DOM
+beforeEach(() => {
+  if (!document.getElementById("portal-root")) {
+    const portalRoot = document.createElement("div");
+    portalRoot.id = "portal-root";
+    document.body.appendChild(portalRoot);
+  }
+});
 
 // Mock ResizeObserver for tests (not available in jsdom)
 global.ResizeObserver = class ResizeObserver {


teresayung:
cursor review

packages/front-end/components/Features/ConditionDisplay.tsx
```diff
@@ -346,8 +346,44 @@ function getConditionParts({
...
+    f !== "$savedGroups" &&
+    f !== "$notSavedGroups";
+
   return conditions.map(({ field, operator, value, parentId }, i) => {
-    let fieldEl: ReactNode = (
+    let fieldEl: ReactNode = isAttributeField(field) ? (
+      <Badge
+        color="gray"
+        label={
+          <Link
```
bryce-fitzsimons: Not loving what the purple link + PiArrowSquareOut is doing to the readability of conditions. I recall we were talking about making the links less prominent, no icon.

packages/front-end/styles/global-radix-overrides.scss
```diff
@@ -364,6 +364,23 @@ p.rt-TooltipText {
   margin-bottom: 0;
 }
 
+// Links inside tooltips: ensure contrast on tooltip background (gray-12 is dark in light theme, light in dark theme)
```
bryce-fitzsimons: we're not showing links currently in the tooltip right? if not, we should revert this

packages/front-end/pages/attributes/[property].tsx
```diff
@@ -0,0 +1,390 @@
...
+                label="description"
+                header="Description"
+                headerClassName="h4"
+                containerClassName="mb-1"
+              />
+            </div>
+          </Frame>
+        </Box>
+
+        <hr />
```
Comment (bryce-fitzsimons) [inline packages/front-end/pages/attributes/[property].tsx]: remove `<hr />`

packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -548,10 +550,70 @@ function ConditionAndGroupInput({
...
+                  disableHoverableContent={false}
+                  content={
+                    <Box style={{ maxWidth: 280 }}>
+                      {opt.description && (
+                        <Box mb={opt.tags?.length ? "2" : undefined}>
+                          <Markdown style={{ fontSize: 12 }}>
+                            {opt.description}
+                          </Markdown>
+                        </Box>
+                      )}
```
bryce-fitzsimons: Probably good to add datatype as well (boolean, string, number, etc...), and for the true case an `isIdentifier` label. This probably means the tooltip needs to always show.

packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -548,10 +550,70 @@ function ConditionAndGroupInput({
...
+                      {opt.tags && opt.tags.length > 0 && (
+                        <Flex gap="1" wrap="wrap">
+                          {opt.tags.map((tag) => (
+                            <Box
```
bryce-fitzsimons: better to use the `<SortedTags>` component

packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -548,10 +550,70 @@ function ConditionAndGroupInput({
...
+                            >
+                              {tag}
+                            </Box>
+                          ))}
+                        </Flex>
+                      )}
+                    </Box>
+                  }
+                >
+                  <span
```
bryce-fitzsimons: use Radix components

packages/front-end/components/Features/ConditionInput.tsx
```diff
@@ -548,10 +550,70 @@ function ConditionAndGroupInput({
...
+              if (!hasTooltip) return label;
+              return (
+                <Tooltip
+                  side="right"
+                  disableHoverableContent={false}
+                  content={
+                    <Box style={{ maxWidth: 280 }}>
+                      {opt.description && (
+                        <Box mb={opt.tags?.length ? "2" : undefined}>
+                          <Markdown style={{ fontSize: 12 }}>
```
bryce-fitzsimons: should we truncate in case of long descriptions?

also not sure if worth addressing, but I uploaded an image to the tag description/markdown, and in tooltip format it can get a bit bulky

packages/front-end/components/Features/ConditionDisplay.tsx
```diff
@@ -346,8 +346,44 @@ function getConditionParts({
...
+            href={`/attributes/${encodeURIComponent(field)}`}
+            target="_blank"
+            color="violet"
```
bryce-fitzsimons: All links are violet by default. But we might want to ping Kari about the styling here.

packages/front-end/styles/global-radix-overrides.scss
teresayung: We are showing links inside the tooltip if they're in the description

packages/front-end/components/Features/AttributeBadge.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface AttributeBadgeProps {
+  /** Attribute id/name to display and link to the attribute page. */
+  attributeId: string;
+}
+
+/**
+ * Gray badge with a link to the attribute page. Uses theme-aware styling:
+ * day theme = gray badge, violet link; dark mode uses matching tokens.
+ */
```
bryce-fitzsimons:
```suggestion
```

packages/front-end/components/Features/AttributeOptionTooltip.tsx
```diff
@@ -0,0 +1,120 @@
...
+export function getAttributeOptionHasTooltip(
+  opt: AttributeOptionForTooltip,
+): boolean {
+  return (
+    !!opt.description ||
+    (opt.tags?.length ?? 0) > 0 ||
+    opt.datatype != null ||
+    opt.hashAttribute !== undefined
+  );
+}
```
bryce-fitzsimons: We don't need this method. datatype is never null so there's always something to show. I think we should also put the full name of the attribute on top of the tooltip so that we don't need to rely on `title`

packages/front-end/components/Features/AttributeOptionTooltip.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <Box>
+              <Text size="small" as="div">
+                <Markdown style={{ fontSize: 12 }}>
+                  {option.description}
+                </Markdown>
+              </Text>
+            </Box>
+          </Flex>
+        )}
+        {(option.datatype != null || option.hashAttribute === true) && (
```
bryce-fitzsimons: too many null checks in this area. It's never null. If there are some legacy edge cases, you could just do `Data type: {option.datatype ?? "unknown"}`

packages/front-end/components/Features/AttributeOptionTooltip.tsx
```diff
@@ -0,0 +1,120 @@
...
+              <Text size="small" as="div">
+                <strong>Identifier</strong>
+              </Text>
+            )}
+          </Flex>
+        )}
+        {option.tags && option.tags.length > 0 && (
+          <Flex direction="column" gap="1">
+            <Text size="small" as="div">
+              <strong>Tags:</strong>
```
bryce-fitzsimons: use text weight instead of `<strong>`. Also true in other areas in this file.

packages/front-end/components/Features/AttributeOptionTooltip.tsx
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export function AttributeOptionWithTooltip({
+  option,
+  children,
+}: {
+  option: AttributeOptionForTooltip;
+  children: React.ReactNode;
+}) {
+  const { theme } = useAppearanceUITheme();
```
bryce-fitzsimons: this looks unnecessary. Is the theme otherwise not being shown correctly?

packages/front-end/components/Features/ConditionDisplay.tsx
```diff
@@ -346,8 +346,16 @@ function getConditionParts({
   keyPrefix?: string;
   prefix?: ReactNode;
 }) {
+  const isAttributeField = (f: string) =>
```
bryce-fitzsimons: this is risky. We should pass in ground truth from the outside instead of guessing.

packages/front-end/styles/global-radix-overrides.scss
bryce-fitzsimons: do we still need this if we're flipping the tooltip color theme?

packages/front-end/components/Features/AttributeOptionTooltip.tsx
teresayung: It's not showing up correctly when i toggle flip theme.  The background changes but the text doesn't

bryce-fitzsimons (approved):

PR #5356, opened 2026-02-24
Use Polyglot for formatting

tzjames:
### Features and Changes
`sql-formatter` is slow and has brought down the website before when we allowed it to format large queries.  Also the version we had didn't support Clickhouse, but we have upgraded in this PR to a version that does. 
`polyglot` is a [new package ](https://github.com/tobilg/polyglot) that does the formatting much more quickly. However it can't parse everything yet so we fallback to `sql-formatter`.  

Unfortunately polyglot ESM version has a top-level await. Since shared package is CJS, the shared sql.ts can't load polyglot EMS version from the front end.  For now we check whether the function exists, and so for the front-end we just fall back to the sql-formatter.  As most front-end queries are small, the sql-formatter probably works fine.  If we ever change our mind we have two options:
- Make polyglot a dependency of front-end and just import it and use it there instead of calling the shared version.
- Compile a ESM version of sql.ts which the front-end uses, in addition to the cjs for the back-end.  Then import it dynamically if and only if window is set (being called from the front-end).  See https://github.com/growthbook/growthbook/pull/5406 for a possible implementation)

In full this PR:
- Adds polyglot to the sharedPackage with a fallback if polyglot doesn't load ok (front-end) or if it fails to format to sql-formatter.
- Adds metrics for formatting so we can keep track of how often formatting succeeds
- Adds simple tests to confirm new functionality is working.
- Updates sql-formatter

### Testing
`pnpm build`
`pnpm start:with-datadog`
start datadog agent locally
refresh an experiment and see new formatting metrics in datadog.
Go create a new dashboard and make a new sql query, type in a query and click format and see it format it in the 
For sql-formatter format:
```
SELECT
  *
FROM
  growthbook.public.pages
LIMIT
  1;
```

If polyglot formatted it, it would have looked like:
```
SELECT
  *
FROM growthbook.public.pages
LIMIT 1;
```
However that doesn't work on the front-end.

### DIFF
--- a/packages/shared/src/sql.ts
+++ b/packages/shared/src/sql.ts
@@ -46,7 +46,7 @@ const MAX_SQL_LENGTH_FOR_POLYGLOT = parseInt(
 );
 
 /** Map Polyglot Dialect to sql-formatter SqlLanguage for fallback when polyglot fails */
-function polyglotToSqlFormatter(dialect: Dialect): SqlLanguage {
+function polyglotToSqlFormatter(dialect: FormatDialect): SqlLanguage {
   switch (dialect) {
     case Dialect.PostgreSQL:
     case Dialect.MySQL:


packages/shared/src/sql.ts
```diff
@@ -1,32 +1,135 @@
...
 import { format as sqlFormat } from "sql-formatter";
 import { SqlResultChunkInterface } from "../types/query";
 import { FormatDialect } from "../types/sql";
 import { FormatError } from "../types/error";
 
+export type FormatMetricsEvent =
+  | { engine: "polyglot"; success: true; timeMs: number }
+  | { engine: "polyglot"; success: false; timeMs: number }
+  | { engine: "sqlformat"; success: true; timeMs: number }
+  | { engine: "sqlformat"; success: false; timeMs: number };
```
Kevin-Chant:
```suggestion
export type FormatMetricsEvent = {
  engine: "polyglot" | "sqlformat";
  success: boolean;
  timeMs: number;
};
```

packages/shared/src/sql.ts
```diff
@@ -1,32 +1,135 @@
...
+): void {
+  formatMetricsReporter = reporter;
+}
+
+let polyglotInitPromise: Promise<void> | null = null;
+
+export async function initPolyglotFormat(): Promise<void> {
+  if (typeof polyglotInit !== "function") return;
+  if (!polyglotInitPromise) {
+    polyglotInitPromise = polyglotInit().catch(() => {});
```
Kevin-Chant, line 33: Should we add any logging to the catch?

tzjames: We would be able to detect it in that the `polyglot.failure / (polyglot.failure + polyglot.success)` metric would be 1.  aka the polyglot error rate. If the init fails that all future formats fail.

PR #5357, opened 2026-02-24
Keep track of which queries got formatted successfully

tzjames:
### Features and Changes
As we might want to look into which queries are failing to be formatted we mark the queries that failed to be formatted as `sqlFormatted: false`

### Testing
Refresh an experiment
In mongo sort the queries collection by {createdAt: -1} see that the most recent query has `sqlFormatted: true`

### DIFF
--- a/packages/back-end/src/types/Integration.ts
+++ b/packages/back-end/src/types/Integration.ts
@@ -1,19 +1,14 @@
 import { ExperimentMetricInterface } from "shared/experiments";
 import { FormatDialect, TemplateVariables } from "shared/types/sql";
 
-/** Query string or object with format status. Used by integrations that format SQL. */
-export type PossiblyFormattedSql =
-  | string
-  | { sql: string; isFormatted?: boolean };
+/** Query with format status. Used by integrations that format SQL. */
+export type PossiblyFormattedSql = { sql: string; isFormatted?: boolean };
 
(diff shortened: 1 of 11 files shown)


packages/back-end/src/types/Integration.ts
```diff
@@ -1,5 +1,19 @@
...
+  | { sql: string; isFormatted?: boolean };
+
+export function resolveSqlQuery(q: PossiblyFormattedSql): {
+  sql: string;
+  isFormatted: boolean;
+} {
+  return typeof q === "string"
+    ? { sql: q, isFormatted: false }
+    : { sql: q.sql, isFormatted: q.isFormatted ?? false };
+}
```
jdorn, line 10: Why keep the string type in a union? It looks like we are already updating everything to use the object version anyway and it would simplify it if we just knew it was always an object.  Could get rid of the `resolveSqlQuery` method entirely.

tzjames: This is due to GoogleAnalytics and Mixpanel where there were just strings. Converting them to have a status (false) of whether they are formatted is a possibility. There is also a getDropUnitsTableQuery that is returning a string right now that we don't format.  We could call formatWithStatus on that but it is rather small. Or just hard code it that it is formatted.

PR #5358, opened 2026-02-24
fix(managed-warehouse): Ensure table identifiers are updated + misc

ahdriel:
### Features and Changes
For the managed warehouse product (backed by clickhouse) we create and manage:

- Datasource type growthbook_clickhouse
  - With columns configuration that users can override for their use case
- Table for events named ch_events

This fixes 3 bugs for the resources we manage:

- Updating a FactTable identifiers / columns, always updates `userIdTypes`
- Fix restoring previously deleted columns for `ch_events` table
- Improve loading indicator to refresh columns that are processed in the background
- Fix permission checks to allow columns to be refreshed via the UI even if they are managedBy API

### Testing

New unit tests added

### DIFF
--- a/packages/back-end/src/models/FactTableModel.ts
+++ b/packages/back-end/src/models/FactTableModel.ts
@@ -258,7 +258,7 @@ export async function updateFactTable(
   factTable: FactTableInterface,
   changes: UpdateFactTableProps,
 ) {
-  // for tables managedBy API, only allow UI updates to columns
+  // Allow changing columns even for API-managed fact tables
   if (
     factTable.managedBy === "api" &&
     context.auditUser?.type !== "api_key" &&
--- a/packages/back-end/src/api/fact-tables/postFactTable.ts
+++ b/packages/back-end/src/api/fact-tables/postFactTable.ts
@@ -43,11 +43,12 @@ export const postFactTable = createApiRequestHandler(postFactTableValidator)(
 
     // Validate userIdTypes
     if (req.body.userIdTypes) {
-      const validIdentifierTypes = new Set(
-        (datasource.settings?.userIdTypes || []).map((u) => u.userIdType),
-      );
       for (const userIdType of req.body.userIdTypes) {
-        if (!validIdentifierTypes.has(userIdType)) {
+        if (
+          !datasource.settings?.userIdTypes?.some(
+            (t) => t.userIdType === userIdType,
+          )
+        ) {
           throw new Error(`Invalid userIdType: ${userIdType}`);
         }
       }
--- a/packages/back-end/src/api/fact-tables/updateFactTable.ts
+++ b/packages/back-end/src/api/fact-tables/updateFactTable.ts
@@ -46,11 +46,12 @@ export const updateFactTable = createApiRequestHandler(
     if (!datasource) {
       throw new Error("Could not find datasource for this fact table");
     }
-    const validIdentifierTypes = new Set(
-      (datasource.settings?.userIdTypes || []).map((u) => u.userIdType),
-    );
     for (const userIdType of req.body.userIdTypes) {
-      if (!validIdentifierTypes.has(userIdType)) {
+      if (
+        !datasource.settings?.userIdTypes?.some(
+          (t) => t.userIdType === userIdType,
+        )
+      ) {
         throw new Error(`Invalid userIdType: ${userIdType}`);
       }
     }


packages/back-end/src/models/FactTableModel.ts
```diff
@@ -92,6 +92,26 @@ function toInterface(doc: FactTableDocument): FactTableInterface {
...
+    factTable.datasource === "managed_warehouse";
+
+  const allowedFields = isManagedWarehouseChEvents
+    ? new Set([
+        "columns",
+        "userIdTypes",
+        "columnsError",
+        "columnRefreshPending",
+      ])
+    : new Set(["columns", "userIdTypes"]);
```
lukesonnet: Nit: not sure why we now allow userIdTypes to be edited here in this use case, but probably doesn't matter.

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -300,15 +298,16 @@ export const putFactTable = async (
...
-    const removedColumns = detectRemovedColumns(originalColumns, data.columns);
-
-    if (removedColumns.length > 0) {
-      await cleanupMetricAutoSlices({
-        context,
-        factTableId: factTable.id,
-        removedColumns,
-      });
+    if (
+      datasource.type === "growthbook_clickhouse" &&
```
lukesonnet: nit: feels like these should be consts somewhere

packages/back-end/src/services/clickhouse.ts
```diff
@@ -797,14 +798,26 @@ export async function updateMaterializedColumns({
...
+          );
           // Destination already exists
-          if (newColumns.find((c) => c.column === to)) {
-            // Just mark the old column as deleted
+          if (existingDestinationCol) {
+            // Restore destination if it had been previously removed.
+            existingDestinationCol.deleted = false;
+            existingDestinationCol.dateUpdated = new Date();
+            // Mark the old column as deleted.
             col.deleted = true;
```
lukesonnet, line 822: Since we are simply marking columns as deleted here, but they still exist (but are no longer populated) in the materialized view, when we Refresh the columns on the fact table won't it just undelete these columns?

More generally, do we need to maintain this list and cross-reference it when doing the column refresh, rather than just posting it to the fact table?

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -54,6 +55,18 @@ const refreshFactTableColumns = async (job: RefreshFactTableColumnsJob) => {
     );
     updates.columns = columns;
     updates.columnsError = null;
+
+    if (
+      datasource.type === "growthbook_clickhouse" &&
+      factTable.id === "ch_events"
```
msamper: We should probably make this a const value somewhere

packages/back-end/src/models/FactTableModel.ts
```diff
@@ -259,8 +279,9 @@ export async function updateFactTable(
   changes: UpdateFactTableProps,
 ) {
   // Allow changing columns even for API-managed fact tables
+  // and the side-effect of updating userIdTypes.
   if (
-    Object.keys(changes).some((k) => k !== "columns") &&
+    !isAllowedApiManagedFactTableUpdate(factTable, changes) &&
     factTable.managedBy === "api" &&
     context.auditUser?.type !== "api_key"
```
msamper: nit: move `isAllowedApiManagedFactTableUpdate` down so we don't perform the check if the fact table isn't managed by the api

packages/back-end/src/services/clickhouse.ts
```diff
@@ -823,9 +836,37 @@ export async function updateMaterializedColumns({
...
 }
+
+export function getManagedWarehouseUserIdTypes(
+  datasource: GrowthbookClickhouseDataSource,
+  factTableId: string,
+  columns: ColumnInterface[],
+): string[] {
+  if (factTableId !== "ch_events") {
+    throw new Error(
+      "This function can only be called for managed warehouse datasource and table.",
```
msamper:
```suggestion
      "This function can only be called for managed warehouse datasource and fact table.",
```

packages/back-end/src/models/FactTableModel.ts
msamper: Wait shouldn't we allow `columnsError` and `columnRefreshPending` to be updated in both cases if we're allowing `column` updates? Seems like if someone tried to refresh the columns on a different fact table it would fail because in `putFactTable` we mutate `columnsError` and `columnRefreshPending` when forcing a column refresh. Maybe we don't need this ternary

ahdriel: That was me getting things mixed up -- fixed.

packages/back-end/src/services/clickhouse.ts
ahdriel: This is an excellent point! We need to test this more thoroughly before merging. cc @jdorn

packages/back-end/src/models/FactTableModel.ts
ahdriel: My understanding of the code is that `columns` would be user-provided and validated against the 'can be modified via API request', while `columnsError` and `columnRefreshPending` is a more internal update that should not go through this code here.

I'll clean this up and test more.

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -296,29 +302,36 @@ export const putFactTable = async (
...
 
-    if (removedColumns.length > 0) {
-      await cleanupMetricAutoSlices({
-        context,
-        factTableId: factTable.id,
-        removedColumns,
-      });
+    if (
+      datasource.type === "growthbook_clickhouse" &&
+      factTable.id === MANAGED_WAREHOUSE_EVENTS_FACT_TABLE_ID
```
jdorn: We should have a general solution, not just for managed warehouse

packages/back-end/src/services/clickhouse.ts
```diff
@@ -797,14 +801,26 @@ export async function updateMaterializedColumns({
             description: "",
             numberFormat: "",
           });
+        } else {
+          // If the column already exists but was previously removed, restore it.
+          existingCol.deleted = false;
+          existingCol.name = col.columnName;
+          existingCol.datatype = col.datatype;
```
jdorn: Remove name and datatype here

packages/back-end/src/services/clickhouse.ts
ahdriel: We actually do delete them in the materialized view, so I think we are good here.

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -296,29 +301,29 @@ export const putFactTable = async (
...
+      columnRefreshPending: needsBackgroundRefresh,
+    };
 
-    if (removedColumns.length > 0) {
-      await cleanupMetricAutoSlices({
-        context,
-        factTableId: factTable.id,
-        removedColumns,
-      });
-    }
```
jdorn, line 312: Was removing this intentional?

ahdriel: Yes -- this is already happening as a side-effect of `updateFactTable` and `updateFactTableColumns` -- no need to do this for a third time here.

PR #5359, opened 2026-02-24
[fix] Not able to delete teams via SCIM

jdorn:
### Features and Changes

Fix bug that was blocking teams from being deleted via SCIM (and other) API calls.

### DIFF
--- a/packages/back-end/src/models/TeamModel.ts
+++ b/packages/back-end/src/models/TeamModel.ts
@@ -132,7 +132,7 @@ export class TeamModel extends BaseClass {
       );
     }
 
-    if (team?.managedByIdp) {
+    if (team?.managedByIdp && !this.context.isApiRequest) {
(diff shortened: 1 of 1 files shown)


packages/back-end/src/models/TeamModel.ts
```diff
@@ -132,7 +132,7 @@ export class TeamModel extends BaseClass {
       );
     }
 
-    if (team?.managedByIdp) {
+    if (team?.managedByIdp && !this.context.isApiRequest) {
```
msamper, line 135: Do we want to allow them to delete teams `managedByIdp` via the API as well as the idP or should deletion be restricted to being done via the idP?

jdorn: I think it's a fine escape hatch to let people fix issues. The restriction was mostly to stop people from accidentally deleting things via the UI.

PR #5361, opened 2026-02-24
Fix SCIM team deletion

Kevin-Chant:
### Features and Changes

Currently, SCIMs are unable to delete teams (groups) in a single request due to the `beforeDelete` check in `TeamModel` still thinking that there are members present in the group.
Along with the fix of updating the in-memory org when members are deleted, this PR also adds a query param (`deleteMembers`) to the API delete endpoint to accomplish the same behavior of removing all members and then deleting the group.

### Testing

Create a team and add members to it. Then, hit the DELETE `/api/v1/teams/:teamId/removeMembersAndDelete` endpoint without the new parameter to see the expected error:
```
{
  "message": "Cannot delete a team that has members. Please delete members before retrying."
}
```

Try again with `deleteMembers` set to `"true"`, and it should allow for team deletion

### Screenshots

image

### DIFF
--- a/packages/back-end/src/models/TeamModel.ts
+++ b/packages/back-end/src/models/TeamModel.ts
@@ -1,14 +1,17 @@
 import {
   apiAddTeamMembersValidator,
   apiCreateTeamBody,
-  apiRemoveMembersAndDeleteValidator,
+  apiDeleteTeamValidator,
+  apiDeleteTeamReturn,
   apiRemoveTeamMemberValidator,
   apiTeamValidator,
   apiUpdateTeamBody,
   teamSchema,
+  ApiDeleteTeamReturn,
 } from "shared/validators";
 import { ApiTeamInterface, TeamInterface } from "shared/types/team";
 import { areProjectRolesValid, isRoleValid } from "shared/permissions";
+import { stringToBoolean } from "shared/util";
 import { IS_CLOUD } from "back-end/src/util/secrets";
 import {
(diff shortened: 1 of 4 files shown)


packages/back-end/src/api/apiModelHandlers.ts
```diff
@@ -31,7 +31,7 @@ export type CustomApiHandler<
   operationId: string;
   validator: ApiRequestValidator<ParamsSchema, BodySchema, QuerySchema>;
   zodReturnObject: ReturnShape;
-  summary?: string;
+  summary: string; // For generating docs, e.g. "Get all dashboards for an experiment"
```
Kevin-Chant, line 34: I marked this as required since without a summary the docs sidebar looks strange for custom handlers. TeamModel was the only one without them so far

packages/back-end/src/models/TeamModel.ts
```diff
@@ -94,6 +97,31 @@ const BaseClass = MakeModelClass({
...
+            organization: req.context.org,
+            userIds: getMembersOfTeam(req.context.org, team.id),
+            teamId: team.id,
+          });
+          await req.context.models.teams.delete(team);
+          return {
+            status: 200,
+          };
+        },
+      }),
```
jdorn, line 129: Instead of a new endpoint, might make sense to add a query param to the deleteTeam endpoint to opt into this behavior.

PR #5362, opened 2026-02-25
Improve feature revision selector

bryce-fitzsimons:
- Don't truncate the minimal revisions, just use in-mem windowing on the component
- Fetch full revision on demand + cache
- Ability to link (?v=versionNumber) to a revision outside the default-loaded set and lazy load its data.
- Ability to select discarded revisions via toggle switch (off by default)
- Improve UI + Radix

image

image

### DIFF
--- a/packages/back-end/src/models/FeatureRevisionModel.ts
+++ b/packages/back-end/src/models/FeatureRevisionModel.ts
@@ -110,7 +110,8 @@ export async function getMinimalRevisions(
     featureId,
   })
     .select("version datePublished dateUpdated createdBy status")
-    .sort({ version: -1 });
+    .sort({ version: -1 })
+    .limit(200);
 
   return docs.map((m) => ({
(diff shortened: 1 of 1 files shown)


packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -110,8 +110,7 @@ export async function getMinimalRevisions(
     featureId,
   })
     .select("version datePublished dateUpdated createdBy status")
-    .sort({ version: -1 })
-    .limit(25);
+    .sort({ version: -1 });
```
jdorn: I think we still need some limit. There are features with thousands of revisions.

bryce-fitzsimons: Ok, but it should be a big number (or introduce pagination which seems like overkill for metadata). Setting to 200

PR #5374, opened 2026-02-26
Fix features /sync endpoint when approvals are required

jdorn:
### Features and Changes

* When approval flows are required and existing features are synced from either the Statsig or LaunchDarkly importer, leave the revision as a draft and do not make the changes live.  Currently it was partially published, getting into a broken state.
* Minor bug - if there are more than 5 revisions AFTER the live one, the "Revert to Previous" link does not work.  It is still possible to revert by selecting the revision manually in the dropdown and then clicking revert from there.  This PR removes the link.  If so much has changed since the live revision, it's reasonable to add some friction to reverting and make you pick the revision manually.

### DIFF
--- a/packages/back-end/src/controllers/features.ts
+++ b/packages/back-end/src/controllers/features.ts
@@ -1500,9 +1500,9 @@ export async function postFeatureSync(
 
   if (
     data.defaultValue != null &&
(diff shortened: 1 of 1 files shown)


jdorn:
cursor review

PR #5376, opened 2026-02-27
Fix a memory leak in OpenIdAuthConnection

tzjames:
### Features and Changes

Fixes a small memory leak in OpenIdAuthConnection where the clientMap would get larger and larger over time. Vercel creates new connections on every request. Also ssoConnectionCache returns new objects on expiry.  Hence the clientMap gets monotonically larger over time.

### Testing
Login with open auth.

### DIFF
--- a/packages/back-end/src/services/auth/OpenIdAuthConnection.ts
+++ b/packages/back-end/src/services/auth/OpenIdAuthConnection.ts
@@ -1,3 +1,4 @@
+import crypto from "crypto";
 import { NextFunction, Request, RequestHandler, Response } from "express";
 import {
   Issuer,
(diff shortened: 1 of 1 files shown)


packages/back-end/src/services/auth/OpenIdAuthConnection.ts
```diff
@@ -56,7 +56,13 @@ const ssoConnectionCache = new MemoryCache(async (ssoConnectionId: string) => {
   throw new Error("Could not find SSO connection - " + ssoConnectionId);
 }, 30);
 
-const clientMap: Map<SSOConnectionInterface, Client> = new Map();
+// Stable key for clientMap
+function getConnectionCacheKey(connection: SSOConnectionInterface): string {
+  if (connection.id) return connection.id;
+  return `${connection.clientId}:${connection.metadata?.issuer ?? ""}`;
```
jdorn: If the other fields in a connection get updated, the cache will not be invalidated. This could lock someone out of their account if we rotate the `clientSecret` for example and it wouldn't recover until the servers restart.

I think this approach could work, we just might want to have a (relatively short) TTL on the cache entries so that stale ones get ejected without a server restart.

tzjames: What about just including all the properties that could affect login in a hash?

packages/back-end/src/services/auth/OpenIdAuthConnection.ts
```diff
@@ -56,7 +57,29 @@ const ssoConnectionCache = new MemoryCache(async (ssoConnectionId: string) => {
...
+    connection.id ??
+    `${connection.clientId}:${connection.metadata?.issuer ?? ""}`;
+  const configHash = crypto
+    .createHash("sha256")
+    .update(
+      JSON.stringify({
+        clientId: connection.clientId,
+        clientSecret: connection.clientSecret ?? "",
+        metadata: connection.metadata,
+      }),
```
Kevin-Chant, line 75: Just to confirm, we're not concerned about map size growth from this hash changing? E.g. if vercel creates new sso connections per-request, those will actually have the same `clientId`, `clientSecret`, and `metadata`?

tzjames: Yes those stay the same.  See line 333-344.

Kevin-Chant: Ah sorry. That's what I get for looking at just the diff on my phone 😅

Kevin-Chant (approved):
I'll defer to Jeremy on which parts of the connection should be hashed, but this should be strictly better than before as-is so I'm approving for now.

tzjames:
Went over it with Jeremy and the props specified are the only one used in creating the client.

PR #5377, opened 2026-02-27
Feature page: fix default revision selected on load

bryce-fitzsimons:
Should be active draft, currently selects the live revision due to async issues.


bryce-fitzsimons:
@cursor do your thing

PR #5378, opened 2026-03-02
Migrate ApiKeyModel to BaseModel

Kevin-Chant:
### Features and Changes

Migrates the ApiKeyModel to inherit from BaseModel and updates usages across the codebase

:warning: There's one notable behavioral change: previously, admins could read other users' PATs, and those keys were returned as part of the GET `/organization` endpoint. Now due to `canRead`, that endpoint will only populate the current user's keys. (The frontend behavior is unchanged as this endpoint wasn't the source of truth for the `/personal-access-tokens` or `/settings/keys` pages)

Other changes
- Uses `key` as the pkey for the schema with an optional `id` field (this should match existing mongo behavior, but differs from the basemodel standard)
- Added a `sanitize` helper in BaseModel for automatically redacting the sensitive fields unless bypassed

### Testing

- [x] PAT behavior is unchanged
  - [x] Can create new PATs
  - [x] Newly created PAT returns a non-empty key value
  - [x] Can view own PATs in /personal-access-tokens
  - [x] Cannot see other users' PATs in the listing (even as admin)
  - [x] Can delete own PAT
  - [x] Admin cannot delete another user's PAT
  - [x] Can reveal own PAT key via the reveal endpoint
- [x] API Key behavior is unchanged
  - [x] Can create new org API keys (as admin)
  - [x] Creating an org API key without permission is rejected
  - [x] Org API keys appear in /settings/keys with key value redacted
  - [x] Can delete an org API key (as admin)
  - [x] Can reveal an org API key (as admin)
  - [x] Non-admin cannot reveal an org API key
- [ ] API auth still functions
  - [x] Request using a PAT
  - [x] Request using an API key
  - [ ] Request using `SECRET_API_KEY` env var

### DIFF
--- a/packages/back-end/src/util/api-key.util.ts
+++ b/packages/back-end/src/util/api-key.util.ts
@@ -76,7 +76,8 @@ export function migrateApiKey(legacyDoc: unknown) {
   };
 }
 
-export async function lookupOrganizationByApiKey(
+// Cross-organization DB operation, lives outside of ApiKeyModel due to a circular dependency with auth middleware
+export async function dangerousLookupOrganizationByApiKey(
   key: string,
 ): Promise<Partial<ApiKeyInterface>> {
   // If self-hosting on a single org and using a hardcoded secret key
--- a/packages/back-end/src/controllers/config.ts
+++ b/packages/back-end/src/controllers/config.ts
@@ -6,7 +6,7 @@ import {
   LegacyExperimentPhase,
   LegacyVariation,
 } from "shared/types/experiment";
-import { lookupOrganizationByApiKey } from "back-end/src/util/api-key.util";
+import { dangerousLookupOrganizationByApiKey } from "back-end/src/util/api-key.util";
 import { APP_ORIGIN } from "back-end/src/util/secrets";
 import { ErrorResponse, ExperimentOverridesResponse } from "back-end/types/api";
 import {
@@ -35,7 +35,8 @@ export async function getExperimentConfig(
   const { key } = req.params;
 
   try {
-    const { organization, secret } = await lookupOrganizationByApiKey(key);
+    const { organization, secret } =
+      await dangerousLookupOrganizationByApiKey(key);
     if (!organization) {
       return res.status(400).json({
         status: 400,
@@ -103,7 +104,8 @@ export async function getExperimentsScript(
   const { key } = req.params;
 
   try {
-    const { organization, secret } = await lookupOrganizationByApiKey(key);
+    const { organization, secret } =
+      await dangerousLookupOrganizationByApiKey(key);
     if (!organization) {
       return res
         .status(400)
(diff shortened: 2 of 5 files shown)


packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -902,7 +893,7 @@ export async function getOrganization(
     : invites.map((i) => ({ email: i.email }));
 
   // Some other global org data needed by the front-end
-  const apiKeys = await getAllApiKeysByOrganization(context);
+  const apiKeys = await context.models.apiKeys.getAll();
```
Kevin-Chant, line 896: This is the helper that used to return all users' keys rather than just the current user. From what I (and Claude) could tell, this was unused and is safe to remove

packages/back-end/src/util/api-key.util.ts
```diff
@@ -33,3 +46,58 @@ export const roleForApiKey = (
...
+    if (orgs.length === 1) {
+      return {
+        id: "SECRET_API_KEY",
+        key: SECRET_API_KEY,
+        secret: true,
+        organization: orgs[0].id,
+        role: SECRET_API_KEY_ROLE,
+      };
+    }
+  }
```
Kevin-Chant, line 95: I haven't been able to test this myself yet because I have multiple orgs in the db. @tzjames if you have a single-org dev env it'd be great if you could try this out in particular

tzjames: It is pretty easy.  You can just do `MONGODB_URI=mongodb://root:password@localhost:27017/growthbook_single_org?authSource=admin` in your env vars and then you create a new database, and then you can have a single org database too for testing.  That said I tested this and it seems to work as expected.

packages/back-end/src/routers/organizations/organizations.controller.ts
tzjames: Agreed it doesn't seem to be used on the front-end.  FWIW, it was added in pull request [#8](https://github.com/growthbook/growthbook/pull/8/changes)! I don't see useUser called there, so I think it was never used from the start. I think it is safe to get rid of that.

packages/back-end/src/util/api-key.util.ts
```diff
@@ -33,3 +46,58 @@ export const roleForApiKey = (
...
+export function migrateApiKey(legacyDoc: unknown) {
+  const obj = legacyDoc as ApiKeyInterface;
+  return {
+    ...obj,
+    role: roleForApiKey(obj) || undefined,
+    dateUpdated: obj.dateUpdated ?? obj.dateCreated,
+  };
+}
+
+export async function lookupOrganizationByApiKey(
```
tzjames: Should this perhaps be a `dangerousLookupOrganizationByApiKey` static method on the `ApiKeyModel`?
It would solve a circular reference with API_KEY_COLLECTION being imported from `ApiKeyModel` which imports from this file.  I guess things are working even with it there, so maybe not a big deal, but it would go with your rules you wrote at the top.

tzjames (approved):

packages/back-end/src/util/api-key.util.ts
Kevin-Chant: I actually had it there to start but it caused a circular import issue between `api-key.util.ts` and the auth middleware IIRC. I'll still rename it to dangerous though since that's an important part of these sorts of methods

PR #5379, opened 2026-03-02
fix: show anchor y-axis to zero checkbox for bar charts

### LINKED ISSUE #4950: [Bug] Bar chart y-axis starts at non-zero value and makes differences look misleadingly large
### Summary

When rendering a bar chart, the y-axis scale automatically starts at a high non-zero value (e.g. around 2200). Because of this, bars whose values differ only slightly (e.g. 2310 vs 2360) appear to have a very large visual difference in height. This is visually misleading, especially for bar charts where users typically expect the baseline to start at zero.

In the attached screenshot, the bars are all around ~2100–2400, but the y-axis starts near 2200. As a result, one bar dominates the chart even though the actual numeric difference between the bars is relatively small.

#### Expected Behavior

For bar charts, the y-axis should:

Default to starting at 0, or

Provide a clear option to force the axis baseline to 0, and preferably make that the default for bar/column charts.

#### Current Behavior

The y-axis minimum is automatically set to a value close to the smallest data point (e.g. ~2200).

This compresses the scale and visually exaggerates differences between bars.

#### Why this matters

Bar charts are often used to compare magnitudes. When the baseline is not zero, the visual impression can be misleading even if the underlying numbers are close. This can lead users to over-interpret small differences as large effects.

#### Suggested improvements

Add a setting like yAxis.startAtZero: true / “Always start y-axis at zero” for bar/column charts.

Consider making startAtZero = true the default for bar charts, while still allowing advanced users to override it if needed.

willdumes:
## Summary

Fixes #4950

The \"Anchor y-axis to zero\" checkbox was not appearing for bar charts because `chartTypeSupportsAnchorYAxisToZero` only included `line` and `scatter`. The rendering layer already defaults `anchorYAxisToZero` to `true` when `displaySettings` is absent, so bar charts were already anchoring to zero — users just had no way to toggle it off.

Changes:
- Add `"bar"` to `chartTypeSupportsAnchorYAxisToZero` so the checkbox appears for bar charts
- Add optional `displaySettings.anchorYAxisToZero` to `barChartValidator` so the setting can be persisted
- Update `AnchorYAxisToZeroCheckbox` type cast to include `BarChart`

## Test plan
- [ ] Create a bar chart in SQL Explorer
- [ ] Confirm the "Anchor y-axis to zero" checkbox now appears in Display Settings
- [ ] Toggle it off — confirm y-axis zooms in to data range
- [ ] Toggle it on — confirm y-axis starts at zero
- [ ] Confirm line and scatter charts are unaffected

### DIFF
--- a/packages/front-end/components/DataViz/DisplaySettingsPanel/AnchorYAxisToZeroCheckbox.tsx
+++ b/packages/front-end/components/DataViz/DisplaySettingsPanel/AnchorYAxisToZeroCheckbox.tsx
@@ -1,4 +1,4 @@
-import { DataVizConfig, LineChart, ScatterChart } from "shared/validators";
+import { BarChart, DataVizConfig, LineChart, ScatterChart } from "shared/validators";
 import { chartTypeSupportsAnchorYAxisToZero } from "shared/enterprise";
 import Checkbox from "@/ui/Checkbox";
 
@@ -18,7 +18,7 @@ export default function AnchorYAxisToZeroCheckbox({
 
   // We know the chart type supports anchorYAxisToZero, so we can cast the dataVizConfig to the appropriate type.
   const configWithDisplaySettings = dataVizConfig as Partial<
-    LineChart | ScatterChart
+    BarChart | LineChart | ScatterChart
   >;
   return (
     <Checkbox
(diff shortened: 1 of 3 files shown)


jdorn:

PR #5382, opened 2026-03-02
fix(trino/presto): Always specify engine headers

ahdriel:
### Features and Changes

We were not passing `engine` to presto-client, so it was always using `presto` as the default value, even when talking to a `trino` server.

Now this is fixed.

To make this not a breaking change we are also adding a `trinoUser` field.

### DIFF
--- a/packages/back-end/src/integrations/Presto.ts
+++ b/packages/back-end/src/integrations/Presto.ts
@@ -59,8 +59,10 @@ export default class Presto extends SqlIntegration {
       timeout: this.params.requestTimeout ?? 0,
       checkInterval: 500,
     };
-    if (this.params.engine === "trino" && this.params.trinoUser) {
-      configOptions.user = this.params.trinoUser;
+    if (this.params.engine === "trino") {
+      if (this.params.trinoUser) {
+        configOptions.user = this.params.trinoUser;
+      }
     } else {
       configOptions.user = this.params.user || "growthbook";
     }


packages/front-end/components/Settings/PrestoForm.tsx
```diff
@@ -84,19 +84,37 @@ const PrestoForm: FC<{
...
+          <input
+            type="text"
+            className="form-control"
+            name="trinoUser"
+            value={params.trinoUser || ""}
+            onChange={onParamChange}
+            placeholder="growthbook"
+          />
+          <small className="form-text text-muted">
+            The user for X-Trino-User header. It is not sent if empty.
```
lukesonnet, line 99: This isn't true. We send `"growthBook"` if it is empty, no?

ahdriel: Not for `trinoUser` -- intentionally, as the behavior before was to send `X-Presto-User` which would be ignored, so now we are just not sending it.

lukesonnet:
```
    if (this.params.engine === "trino" && this.params.trinoUser) {
      configOptions.user = this.params.trinoUser;
    } else {
      configOptions.user = this.params.user || "growthbook";
    }
```

To me, this implies that we will send `this.params.user || "growthbook"` if this field is unset. I defer to you of course!

ahdriel: Nice catch! The logic was wrong, fixed it now:
```
if (this.params.engine === "trino") {
  if (this.params.trinoUser) {
    configOptions.user = this.params.trinoUser;
  }
} else {
  configOptions.user = this.params.user || "growthbook";
}
```

PR #5392, opened 2026-03-03
Fact metric SQL validation

jdorn:
### Features and Changes

Adds API-side validation for fact metric row filters, specifically for the `sql_expr` operator, before saving. This prevents runtime SQL errors by executing a `LIMIT 0` query against the datasource to test SQL validity.

- Introduces `packages/back-end/src/services/factMetricRowFilterValidation.ts` to encapsulate validation logic.
- Integrates validation into `FactMetricModel.customValidation` for both create and update operations.
- Tightens saved-filter ID validation for denominator filters when numerator and denominator share the same fact table.

### Dependencies

None

### Testing

- Added focused unit tests in `packages/back-end/test/services/factMetricRowFilterValidation.test.ts`.
- Ran targeted backend tests and a backend type-check to verify correctness.

### Screenshots

None

---
[Slack Thread](https://growthbookapp.slack.com/archives/C046L8DQD6F/p1772531700696069?thread_ts=1772531700.696069&cid=C046L8DQD6F)

<p><a href="https://cursor.com/agents/bc-dcdffd09-316c-5d5b-b8bd-a66260d4c7f5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-dcdffd09-316c-5d5b-b8bd-a66260d4c7f5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</p>

### DIFF
--- a/packages/back-end/src/models/FactMetricModel.ts
+++ b/packages/back-end/src/models/FactMetricModel.ts
@@ -401,7 +401,6 @@ export class FactMetricModel extends BaseClass {
       numeratorSqlExprFiltersToValidate.length ||
       denominatorSqlExprFiltersToValidate.length
     ) {
-      const rowFilterValidationTestDays = 1;
       const datasource = await getDataSourceById(this.context, data.datasource);
       if (!datasource) {
         throw new Error("Could not find datasource");
@@ -416,7 +415,6 @@ export class FactMetricModel extends BaseClass {
         integration,
         factTable: numeratorFactTable,
         rowFilters: numeratorSqlExprFiltersToValidate,
-        testQueryDays: rowFilterValidationTestDays,
         errorPrefix: "Invalid numerator row filter SQL: ",
       });
 
@@ -425,7 +423,6 @@ export class FactMetricModel extends BaseClass {
           integration,
           factTable: denominatorFactTable,
           rowFilters: denominatorSqlExprFiltersToValidate,
-          testQueryDays: rowFilterValidationTestDays,
           errorPrefix: "Invalid denominator row filter SQL: ",
         });
       }
--- a/packages/back-end/src/services/factMetricRowFilterValidation.ts
+++ b/packages/back-end/src/services/factMetricRowFilterValidation.ts
@@ -68,13 +68,11 @@ export async function validateFactMetricRowFilterSql({
   integration,
   factTable,
   rowFilters,
-  testQueryDays,
   errorPrefix,
 }: {
   integration: SourceIntegrationInterface;
   factTable: FactTableForRowFilterValidation;
   rowFilters: RowFilter[] | undefined;
-  testQueryDays?: number;
   errorPrefix: string;
 }): Promise<void> {
   if (!integration.getTestValidityQuery || !integration.runTestQuery) {
@@ -86,12 +84,12 @@ export async function validateFactMetricRowFilterSql({
     return;
   }
 
-  const query = `SELECT * FROM (
+  const query = `SELECT timestamp FROM (
   ${factTable.sql}
 ) f
 WHERE ${riskyFilterExpressions.join(" AND ")}`;
 
-  const sql = integration.getTestValidityQuery(query, testQueryDays, {
+  const sql = integration.getTestValidityQuery(query, 1, {
     eventName: factTable.eventName,
   });
 
--- a/packages/back-end/test/services/factMetricRowFilterValidation.test.ts
+++ b/packages/back-end/test/services/factMetricRowFilterValidation.test.ts
@@ -118,7 +118,6 @@ describe("factMetricRowFilterValidation", () => {
         integration,
         factTable,
         rowFilters: [{ operator: "in", column: "country", values: ["US"] }],
-        testQueryDays: 14,
         errorPrefix: "Invalid row filter SQL: ",
       });
 
@@ -132,7 +131,6 @@ describe("factMetricRowFilterValidation", () => {
           integration: {} as SourceIntegrationInterface,
           factTable,
           rowFilters: [{ operator: "sql_expr", values: ["amount > 100"] }],
-          testQueryDays: 14,
           errorPrefix: "Invalid row filter SQL: ",
         }),
       ).resolves.toBeUndefined();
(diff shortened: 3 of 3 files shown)


packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -322,39 +341,67 @@ export class FactMetricModel extends BaseClass {
...
+        this.context,
+        datasource,
+        true,
+      );
+
+      await validateFactMetricRowFilterSql({
+        integration,
+        factTable: numeratorFactTable,
+        rowFilters: data.numerator.rowFilters,
+        testQueryDays: this.context.org.settings?.testQueryDays,
```
jdorn: Since we don't care about return data at all, just that the query doesn't throw an error, we should hard-code 1 here to keep the scanned data as small as possible.

packages/back-end/src/services/factMetricRowFilterValidation.ts
```diff
@@ -0,0 +1,81 @@
...
+  rowFilters.forEach((rowFilter) => {
+    if (rowFilter.operator === "sql_expr") {
+      const sqlExpression = rowFilter.values?.[0]?.trim();
+      if (sqlExpression) {
+        expressions.push(`(${sqlExpression})`);
+      }
+      return;
+    }
+
+    if (rowFilter.operator === "saved_filter") {
```
jdorn: Saved filters are validated separately when saved, so we don't need to check them again here. We can assume the SQL in a saved filter is valid already.

packages/back-end/src/services/factMetricRowFilterValidation.ts
```diff
@@ -0,0 +1,81 @@
...
+): string[] {
+  if (!rowFilters?.length) return [];
+
+  const expressions: string[] = [];
+
+  rowFilters.forEach((rowFilter) => {
+    if (rowFilter.operator === "sql_expr") {
+      const sqlExpression = rowFilter.values?.[0]?.trim();
+      if (sqlExpression) {
+        expressions.push(`(${sqlExpression})`);
```
jdorn: Custom SQL may have line comments at the end, so must insert a new line before the closing parentheses.

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -250,6 +281,7 @@ export class FactMetricModel extends BaseClass {
         "Cannot update fact metric managed by API if the request isn't from the API.",
       );
     }
+    this._existingMetricFromUpdate = existing;
```
jdorn: I don't like this pattern.  Why don't we update the BaseModel instead to pass an optional argument into `customValidation` with the previous object.  When called during create, it will be undefined, but when called during update it will be populated.

There's already a `writeOptions` argument in customValidation.  Let's move that to the 3rd argument and add previousData as the 2nd argument.  There's only 1 model that uses writeOptions right now (`UrlRedirectModel`), so that's the only other one we'll need to update.

packages/back-end/src/services/factMetricRowFilterValidation.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (!integration.getTestValidityQuery || !integration.runTestQuery) {
+    return;
+  }
+
+  const riskyFilterExpressions = getRiskyRowFilterSqlExpressions(rowFilters);
+  if (!riskyFilterExpressions.length) {
+    return;
+  }
+
+  const query = `SELECT * FROM (
```
ahdriel: question: Can we use `SELECT 1` here given we don't care about the results, just about the execution?

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -322,39 +349,88 @@ export class FactMetricModel extends BaseClass {
...
+              existingMetric.denominator?.factTableId !==
+                data.denominator.factTableId,
+          })
+        : [];
+
+    if (
+      numeratorSqlExprFiltersToValidate.length ||
+      denominatorSqlExprFiltersToValidate.length
+    ) {
+      const rowFilterValidationTestDays = 1;
```
ahdriel: suggestion: Should we just hardcode 1 in `validateFactMetricRowFilterSql`?

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -322,39 +349,88 @@ export class FactMetricModel extends BaseClass {
...
+      if (!datasource) {
+        throw new Error("Could not find datasource");
+      }
+      const integration = getSourceIntegrationObject(
+        this.context,
+        datasource,
+        true,
+      );
+
+      await validateFactMetricRowFilterSql({
```
ahdriel, line 414: nit: we are doing 2 validations sequentially here, but we could do Promise.all and run concurrently.

packages/back-end/src/services/factMetricRowFilterValidation.ts
lukesonnet: Even SELECT 0 in the case of the engines that support it?

jdorn: `getTestValidityQuery` already does `LIMIT 0` (or `LIMIT 1` if the data source doesn't support it).  I'm slightly worried about query optimization engines seeing `SELECT 1` and skipping the rest of the query.  I can do `SELECT timestamp` instead, which we know 100% of fact table sql contains.

packages/back-end/src/models/FactMetricModel.ts
jdorn: Yes, I tried telling Cursor to do this, but this was the best it could do.  Fixing manually now

packages/back-end/src/models/FactMetricModel.ts
jdorn: I think this would only come up if a ratio metric adds a new, different `sql_expr` filter to both the numerator and denominator at the same time, which seems like it would be extremely rare.

PR #5398, opened 2026-03-03
feat(approval-flows): Add support to Saved Groups

ahdriel:
### Features and Changes


packages/shared/src/enterprise/approval-flows/helpers.ts
gazzdingo, line 1: isnt this duplicated on the the backend we should be able to remove the backend one

PR #5405, opened 2026-03-04
Disable webhooks automatically on multiple failures

jdorn:
### Features and Changes

There are many invalid webhooks configured on GrowthBook Cloud and they just continuously attempt to fire, adding noise to our logs.

When a webhook (legacy, sdk, or proxy) fails 10 times consecutively, disable it automatically.  It will be flagged as disabled in the UI and can be re-enabled in a few ways:
1. Clicking the "Test" button with a successful response from the endpoint
2. Updating the webhook settings (e.g. changing the URL)
3. Manually re-enabling it

### Testing
Can trigger all webhook types at once by creating an SDK connection with a proxy enabled and adding an SDK webhook to the connection.  Then, just toggle a feature on and off a bunch of times.

* SDK Webhooks
  * [x] Fire webhook with 4XX response 10 times.  Watch the `consecutiveFailures` increase each time.  After 10 times, should be marked as disabled
  * [x] Attempt to fire webhook again after disabled and it should not fire
  * [x] See webhook marked as disabled in the UI
  * [x] Re-enable webhook in UI by clicking the `Test` button and getting back a 2XX resonse
  * [ ] Fire 2 webhook calls with failures, then 1 with success and see `consecutiveFailures` get reset to 0
* Proxy Webhooks
  * [x] Fire webhook with 4XX response 10 times.  Watch the `consecutiveFailures` increase each time.  After 10 times, should be marked as disabled
  * [x] Attempt to fire webhook again after disabled and it should not fire
  * [x] See webhook marked as disabled in the UI
  * [x] Re-enable webhook in UI by clicking the `Test` button and getting back a 2XX resonse
  * [ ] Fire 2 webhook calls with failures, then 1 with success and see `consecutiveFailures` get reset to 0

### DIFF
--- a/packages/back-end/src/models/SdkConnectionModel.ts
+++ b/packages/back-end/src/models/SdkConnectionModel.ts
@@ -12,6 +12,7 @@ import {
   SDKLanguage,
 } from "shared/types/sdk-connection";
 import { ApiSdkConnection } from "shared/types/openapi";
+import { WEBHOOK_CONSECUTIVE_FAILURES_THRESHOLD } from "shared/constants";
 import { cancellableFetch } from "back-end/src/util/http.util";
 import {
   IS_CLOUD,
@@ -465,7 +466,9 @@ export async function setProxyError(
         "proxy.connected": false,
         "proxy.lastError": new Date(),
         "proxy.consecutiveFailures": consecutiveFailures,
-        ...(consecutiveFailures >= 5 ? { "proxy.enabled": false } : {}),
+        ...(consecutiveFailures >= WEBHOOK_CONSECUTIVE_FAILURES_THRESHOLD
+          ? { "proxy.enabled": false }
+          : {}),
       },
     },
   );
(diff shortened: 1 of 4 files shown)


packages/back-end/src/models/SdkConnectionModel.ts
```diff
@@ -447,19 +453,40 @@ export async function setProxyError(
...
+        organization: connection.organization,
+        id: connection.id,
+      },
+      {
+        $set: {
+          "proxy.enabled": false,
+        },
+      },
+    );
+  }
```
jdorn: A few changes:
1. Don't use `$inc`, just read the current count and add one. There's technically a chance for race conditions, but that's ok since we don't need the count to be 100% accurate.
2. Don't type the updates as `Record<string, unknown>`. If the updates are passed directly into the `updateOne` call, it will be properly type checked.
3. Don't do 2 separate updates. Combine them into 1

packages/front-end/components/Features/SDKConnections/SdkWebhooks.tsx
```diff
@@ -121,7 +121,31 @@ export default function SdkWebhooks({
...
+                        Last error: {webhook.error}
+                      </>
+                    ) : null}
+                  </div>
+                </Callout>
+              }
+            >
+              <span className="text-warning-muted">
+                <FaExclamationTriangle /> disabled
+              </span>
```
jdorn, line 146: This should be treated as an error, not a warning in the UI

PR #5412, opened 2026-03-04
fix(owner): Normalize to userId for all internal APIs

ahdriel:
## Summary

The `owner` field has been misconfigured and instead of holding just a `user id`, it sometimes holds the `username` or `email` of the owner.

To close this gap, we are updating all internal usages (made via the UI) to store the `user id`.

And we are also updating all places where they are rendered to handle all the different formats that these values might be.

Additionally, we still render the legacy string formats when editing / selecting an Owner, but if another options is picked then we remove that, forcing the update to a proper `user id` reference.

We are **not** making any changes to the public API at this moment because it would be a breaking change, but that's the next step.


packages/back-end/src/models/BaseModel.ts
```diff
@@ -621,7 +621,9 @@ export abstract class BaseModel<
 
     // Add default owner and createdBy if empty
     if ("owner" in props && !props.owner) {
-      props.owner = this.context.userName || "";
+      props.owner = this.context.isApiRequest
+        ? this.context.userName || ""
+        : this.context.userId || this.context.userName || "";
     }
     if ("createdBy" in props && !props.createdBy) {
       props.createdBy = this.context.userName || "";
```
ahdriel, line 627: Should we apply the same fix to `createdBy` in a separate PR? I have not looked into how this is being handled yet, but leaving a note so we don't forget.

Kevin-Chant: Yes I believe we should, though it's probably also worth trying to standardize those models on `owner` instead of `createdBy`. At a glance I see three usages - `GithubIntegrationInterface`, `TeamModel`, and `minimalFeatureRevisionInterface` (which could stay as-is)

packages/front-end/components/Owner/SelectOwner.tsx
```diff
@@ -27,51 +15,42 @@ export default function SelectOwner({
...
 
-  // if the resource stores owner by id, we need the id to be the value, rather than the name
-  const memberOptions =
-    ownerIdentifierType === "id"
-      ? memberUserNameAndIdOptions
-      : memberUsernameOptions;
+  const options = useMemo(() => {
+    if (!value || memberOptions.some((member) => member.value === value)) {
+      return memberOptions;
+    }
```
Kevin-Chant: Probably doesn't matter much, but how about swapping the `memberOptions.some` check for a `Set.has` for performance in large orgs?

packages/back-end/src/models/BaseModel.ts
jdorn: createdBy is different from owner. It's more of an audit log trail so should never be updated. At least that's how it's used for feature revisions.

packages/front-end/components/Owner/SelectOwner.tsx
ahdriel: Certainly does not hurt to be more performant. Done!

Kevin-Chant (approved):
LGTM!

PR #5417, opened 2026-03-04
variations in phases 1: use getLatestPhaseVariations

lukesonnet:
### Features and Changes

Adds two new methods:
* getLatestPhaseVariations <- returns the latest phase variations complete with metadata and original index from the `experiment.variations` object so that we can know if a varaition was the 1st or 2nd, etc... for front-end purposes. Eventually, this will pull from the phase `variations` object in a later PR, which would allow us to persist some defunct variations but still store their metadata at the experiment level.
* getAllVariations <- returns simply all experiment.variations. This will not change going forward.

Furthermore, I updated a bunch of front-end code to use `v.index` instead of `i`, so that when the phases variations array deviates from the experiment variations array, the front-end updates as well.

In another follow-up PR I will migrate date graph and other visual components to use variation index instead of the array index of the variations array passed in

### DIFF
--- a/packages/shared/src/experiments/variations.ts
+++ b/packages/shared/src/experiments/variations.ts
@@ -21,15 +21,10 @@ export function getLatestPhaseVariations(
   // TODO change to come from phase
   const phaseVariations = experiment.variations;
 
-  const indexById = new Map<string, number>();
-  allVariations.forEach((v, i) => {
-    indexById.set(v.id, i);
-  });
-
   let hasMissing = false;
   const result: VariationWithIndex[] = phaseVariations.map((v, i) => {
-    const foundIndex = indexById.get(v.id);
-    if (foundIndex === undefined) {
+    const foundVariation = allVariations.find((allV) => allV.id === v.id);
+    if (foundVariation === undefined) {
       hasMissing = true;
       return {
         ...v,
@@ -37,10 +32,9 @@ export function getLatestPhaseVariations(
       };
     }
     return {
-      ...allVariations[foundIndex],
+      ...foundVariation,
       // override experiment variation metadata with phase variation metadata, if present
       ...v,
-      index: foundIndex,
     };
   });
   if (!hasMissing) {


packages/back-end/src/controllers/experiments.ts
```diff
@@ -250,9 +251,11 @@ export async function postAIExperimentAnalysis(
     })) || undefined;
 
   const winnerVariationName =
-    experiment.variations[winner]?.name || "none chosen";
+    getLatestPhaseVariations(experiment)[winner]?.name || "none chosen";
   const releasedVariationName =
-    experiment.variations.find((v) => v.id === releasedVariationId)?.name || "";
+    getLatestPhaseVariations(experiment).find(
+      (v) => v.id === releasedVariationId,
+    )?.name || "";
```
jdorn: I don't think these changes here are safe.  In any case where we are pointing to a specific variation, either by id or index, we kind of need to use the full variations array and not just the one from the latest phase.

So maybe we have two helper methods and we choose which one to use based on context - `getLatestPhaseVariations(experiment)` and `getAllVariations(experiment)`.

packages/shared/src/experiments.ts
```diff
@@ -1929,3 +1930,45 @@ export function getEffectiveLookbackOverride(
...
+ */
+export function getLatestPhaseVariations(
+  experiment: ExperimentWithVariations,
+): VariationWithIndex[] {
+  const allVariations = getAllVariations(experiment);
+  // TODO change to come from phase
+  const phaseVariations = experiment.variations;
+  return phaseVariations.map((v, i) => ({
+    ...v,
+    index: allVariations.findIndex((allV) => allV.id === v.id) ?? i,
```
ahdriel: potential issue? if `findIndex` fails it'll return `-1` so the `?? i` will never apply.
so: what is the safe fallback that we can have if an index is not found via id?

I think it should be either all are in the `allVariations` and we use `id` to get the order, or we should default to `i` for the array order.

what do you think?

packages/shared/src/experiments.ts
```diff
@@ -1929,3 +1930,45 @@ export function getEffectiveLookbackOverride(
...
+type VariationWithIndex = Variation & {
+  index: number;
+};
+
+/**
+ * Returns the variations for the current/latest phase of an experiment.
+ * Today this just returns experiment.variations directly. In the future,
+ * this will merge phase-level variation status with top-level metadata.
+ */
+export function getLatestPhaseVariations(
```
ahdriel: suggestion: should we move these helpers to `src/experiments/variations.ts`?

packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
```diff
@@ -464,7 +465,7 @@ export function getExperimentResultStatus({
       srm: healthSummary.srm,
       srmThreshold: healthSettings.srmThreshold,
       totalUsersCount: healthSummary.totalUsers,
-      numOfVariations: experimentData.variations.length,
```
ahdriel, line 467: question: should we add `/** @deprecated */` to `experimentInterface`? or is it too early?

If yes we should then let's talk about eslint rules as well on Slack

packages/front-end/components/HealthTab/TrafficCard.tsx
```diff
@@ -164,10 +164,10 @@ export default function TrafficCard({
               <th className="border-top-0" style={{ paddingLeft: "32px" }}>
                 {dimensionWithIssues?.label}
               </th>
-              {variations.map((v, i) => (
+              {variations.map((v) => (
                 <th
-                  className={`border-top-0 variation with-variation-label variation${i}`}
-                  key={i}
+                  className={`border-top-0 variation with-variation-label variation${v.index}`}
+                  key={v.id}
```
ahdriel, line 170: Love to see a proper differentiation between `id` and `index` 🎉

docs/src/data/commercialFeatures.ts
```diff
@@ -1,6 +1,10 @@
 // THIS FILE IS AUTOGENERATED BY `pnpm gen-commercial-features`. DO NOT EDIT DIRECTLY.
 
 export default {
+  "adv-presentations": {
```
ahdriel, line 4: nit: is this the source of conflict? or should we do this in a separate PR?

packages/shared/src/experiments.ts
```diff
@@ -1929,3 +1930,45 @@ export function getEffectiveLookbackOverride(
...
+};
+
+type VariationWithIndex = Variation & {
+  index: number;
+};
+
+/**
+ * Returns the variations for the current/latest phase of an experiment.
+ * Today this just returns experiment.variations directly. In the future,
+ * this will merge phase-level variation status with top-level metadata.
```
ahdriel: question: will `phase-level variation` have the same shape as `Variation` currently does? or will they differ?

packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
lukesonnet: to `variations`? I think we should so that it isn't used directly, I would love to discuss further. However, setting to it is appropriate, so would we want to create setter functions?

packages/shared/src/experiments.ts
lukesonnet: No, phase level variation will be what is here (https://github.com/growthbook/growthbook/pull/5430/changes#diff-ba62e0b66745290769628b046e4ec1ac53d18920badba9fe0a6d8482f42cd775R69-R78). The PR is out of date now, but that model is the right idea.  


```
type PhaseVariation = {
  id: string,
  status: "active" | "disabled"
}
```

Then this method would maybe take a status filter you have to specify "active" or "all", and if active, would look like:

```
const allVariations = getAllVariations(experiment);
const latestPhase = experiment.phases[experiment.phases.length - 1];
const phaseVariations: PhaseVariation[] = latestPhase
  ? latestPhase.variations.filter((v) => v.status === "active")
  : experiment.variations.map((v) => ({ id: v.id, status: "active" }));
return phaseVariations.map((v, i) => {
  const fullVariation = allVariations.find((allV) => allV.id === v.id);
  const foundIndex = allVariations.findIndex((allV) => allV.id === v.id);
  return {
    ...v,
    ...(fullVariation ?? {}), // might need to fill in some missing fields
    index: foundIndex >= 0 ? foundIndex : i
  };
});
```

packages/shared/src/experiments.ts
lukesonnet: Done.

packages/shared/src/experiments.ts
lukesonnet: Yeah I think you're probably right. But TBH I'm not sure. I'll implement that change.

packages/shared/src/experiments.ts
ahdriel: I see the import is happening from `shared/variations` -- is this how we want it? IMO it should still be under `experiments` as variations always belong to an experiment, right?

ahdriel: I think this line is the reason why: https://github.com/growthbook/growthbook/pull/5417/changes/78096bfadb183b2662dc16b0b835145deb99953a..e5d9ad158fdb1433e8f429c80a3c66f4771b23ef#diff-b3fd18ab84f90701b9508e0d75f1a63bb7ba28916863eb2ebe598fd77cb4f144R1933

packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
ahdriel: Yes -- sorry my comment was not super clear.

Can you clarify when using `experiment.variations = ` would be appropriate with these changes?
Ideally that would also happen with a function (or setter if this is a class) because it has other side effects now, no?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -247,9 +248,10 @@ export async function postAIExperimentAnalysis(
     })) || undefined;
 
   const winnerVariationName =
-    experiment.variations[winner]?.name || "none chosen";
+    getAllVariations(experiment)[winner]?.name || "none chosen";
```
ahdriel: nit: should we reuse the variations instead of calling this twice?

packages/front-end/components/HealthTab/BanditSRMCard.tsx
```diff
@@ -52,11 +53,11 @@ export default function BanditSRMCard({
       getSRMHealthData({
         srm: srm ?? Infinity,
         srmThreshold,
-        numOfVariations: experiment.variations.length,
+        numOfVariations: getLatestPhaseVariations(experiment).length,
         totalUsersCount: totalUsers,
         minUsersPerVariation: DEFAULT_SRM_BANDIT_MINIMINUM_COUNT_PER_VARIATION,
       }),
-    [srm, srmThreshold, experiment.variations.length, totalUsers],
+    [srm, srmThreshold, experiment, totalUsers],
```
ahdriel: Oh this is a big change actually, can we move the `getLatestPhaseVariations().length` to a const and use it here?

packages/shared/src/enterprise/decision-criteria/decisionCriteria.ts
lukesonnet: Discussed it wouldn't be except via a getter or a setter, but will handle this in a follow-up PR.

PR #5426, opened 2026-03-05
Fix(results): minor CI popover fix

lukesonnet:
### Features and Changes

Adds `(adj.)` when a CI is adjusted and fixes the values used, before it was incorrectly showing just the upper CI threshold (e.g. 97.5%) instead of the width of the CI (95%) when alpha = 0.05.

Also removes the plus minus from the results row which was really noisy.


packages/front-end/components/Experiment/ExperimentResultTooltipContent/ExperimentResultTooltipContent.tsx
```diff
@@ -90,8 +92,17 @@ export default function ExperimentResultTooltipContent({
     ? getColumnRefFormatter(metric.numerator, getFactTableById)
     : getMetricFormatter(metric.type === "binomial" ? "count" : metric.type);
 
+  const numberFormatter = Intl.NumberFormat(undefined, {
+    maximumFractionDigits: 1,
+  });
+  const ciWidthDisplay =
+    numberFormatter.format(100 * (1 - pValueThreshold)) + "%";
+
+  console.log(pValueAdjustmentEnabled);
```
jdorn: Remove console log

packages/front-end/components/Experiment/ExperimentResultTooltipContent/ExperimentResultTooltipContent.tsx
```diff
@@ -90,8 +92,17 @@ export default function ExperimentResultTooltipContent({
     ? getColumnRefFormatter(metric.numerator, getFactTableById)
     : getMetricFormatter(metric.type === "binomial" ? "count" : metric.type);
 
+  const numberFormatter = Intl.NumberFormat(undefined, {
+    maximumFractionDigits: 1,
+  });
```
jdorn: Ideally, create this outside of the component so it doesn't create a new instance on every render

PR #5431, opened 2026-03-05
Use Polyglot for formatting

tzjames:
### Features and Changes
In https://github.com/growthbook/growthbook/pull/5356 we used polyglot for formatting.  In a last minute change though I had changed how the Dialect mapping to use Polygot's version first.  Since polyglot doesn't load correctly on the front-end, this caused a runtime error on the front-end which was not noticed until after I deployed, and it had to be reverted.  This PR gets it working again by hardcoding the dialects everywhere we need them.   We might be able to rework this if we do separate formatting on the front-end and back-end, but first we need to see how often Polyglot is successful before using it widely on the front-end.

### Testing
`pnpm build`
`pnpm start:with-datadog`
start datadog agent locally
refresh an experiment and see new formatting metrics in datadog.
Go create a new dashboard and make a new sql query, type in a query and click format and see it format it in the 
For sql-formatter format:
```
SELECT
  *
FROM
  growthbook.public.pages
LIMIT
  1;
```

If polyglot formatted it, it would have looked like:
```
SELECT
  *
FROM growthbook.public.pages
LIMIT 1;
```
However that doesn't work on the front-end.

### DIFF
--- a/packages/front-end/services/sqlFormatter.ts
+++ b/packages/front-end/services/sqlFormatter.ts
@@ -1,28 +1,28 @@
 import { DataSourceType } from "shared/types/datasource";
-import { Dialect, format } from "shared/sql";
+import { format } from "shared/sql";
 import { FormatDialect } from "shared/types/sql";
 
 export function canFormatSql(datasourceType: DataSourceType): boolean {
   return !!getSqlDialect(datasourceType);
 }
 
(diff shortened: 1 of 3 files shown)


Kevin-Chant (approved):
LGTM, second try's the charm

PR #5433, opened 2026-03-05
Fix snapshot stuck in running status and improve error tooltip

lukesonnet:
## Summary

- **Snapshot stuck in "running" when analysis setup fails:** `createSnapshot` persists the snapshot with `status: "running"` before calling `startAnalysis`. If `startAnalysis` throws (e.g., no metrics selected, integration error), nothing updated the snapshot status, leaving it permanently stuck. Now the `startAnalysis` call is wrapped in a try/catch that sets the snapshot to `status: "error"` before re-throwing.
- **Defensive empty-queries guard in `QueryRunner.startAnalysis`:** If `startQueries` returns zero queries, the runner now explicitly fails with a descriptive error ("No queries were generated for this analysis") instead of relying on the `getOverallQueryStatus()` edge case where `0 >= 0` evaluates to `"failed"`.
- **Improved error tooltip when snapshot fails with no queries:** When `latest` has a top-level error but no queries (failed before queries were created), the warning triangle tooltip now shows "Snapshot update failed: \<error\>" instead of the misleading "Click to view queries" text.

## Test plan

- [x] Verify that a snapshot which fails during `startAnalysis` (e.g., experiment with no metrics) transitions to `status: "error"` instead of remaining stuck in `"running"`
- [x] Verify that the warning triangle tooltip displays the snapshot error message when there are no queries
- [x] Verify normal snapshot creation and query execution still works end-to-end

### DIFF
--- a/packages/front-end/components/Experiment/Results.tsx
+++ b/packages/front-end/components/Experiment/Results.tsx
@@ -197,7 +197,7 @@ const Results: FC<{
     experiment.guardrailMetrics.length > 0;
 
   const isBandit = experiment.type === "multi-armed-bandit";
-
+  const hasQueries = queryStrings.length > 0;
   return (
     <>
       {!draftMode ? null : (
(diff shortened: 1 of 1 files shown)


packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -222,6 +222,20 @@ export abstract class QueryRunner<
     const queries = await this.startQueries(params);
     this.model.queries = queries;
 
+    if (queries.length === 0) {
+      const noQueriesError = "No queries were generated for this analysis";
+      logger.debug(this.model.id + " runner: " + noQueriesError);
+      const newModel = await this.updateModel({
+        status: "failed",
```
lukesonnet, line 229: IDK if we want "failed" or "succeeded", to be honest.

PR #5438, opened 2026-03-06
chore(preview): Add healthcheck + stop machine earlier + increase CPU

ahdriel:
### Features and Changes

- Increase CPU for better performance
- Add healthcheck check -- ensuring if the docker image builds but the server has problem starting this will fail
- Stop machine earlier -- save some $ if no one is accessing it; no need to keep it online for 30 minutes every deployment


.github/workflows/preview.yml
```diff
@@ -42,18 +45,54 @@ jobs:
...
           org: growthbook
           region: sjc
-          secrets:
-            APP_ORIGIN=https://pr-${{ github.event.number }}-growthbook.fly.dev
-            API_HOST=https://pr-${{ github.event.number }}-growthbook.fly.dev:4433
+          secrets: APP_ORIGIN=${{ env.APP_URL }}
+            API_HOST=${{ env.APP_URL }}:4433
             MONGODB_URI=${{ secrets.MONGODB_URI_STAGING }}
-            JWT_SECRET=${{ secrets.MONGODB_URI_STAGING }}
+            JWT_SECRET=${{ secrets.JWT_SECRET_STAGING }}
```
jdorn, line 54: Do we need to add this to our GitHub secrets or does it already exist?

ahdriel: I have added it! Thanks for checking.

And to clarify the impact, this will just invalidate all sessions as the JWT_SECRET changes, so preview users will have to log in again. So it is a safe change with not big impact.

PR #5440, opened 2026-03-06
fix(exp-overview): make hypothesis markdown; other cleanup

lukesonnet:
### Features and Changes

Changes hypothesis field to markdown.

Makes hypothesis callout dismissable.

Hides segment from overview analysis settings if not eligible.

### Screenshots

https://www.loom.com/share/97e64ec7f4e34efc9e4b854cec797dcf

### DIFF
--- a/packages/front-end/components/Experiment/EditHypothesisModal.tsx
+++ b/packages/front-end/components/Experiment/EditHypothesisModal.tsx
@@ -141,7 +141,7 @@ export default function EditHypothesisModal({
           placeholder="e.g Making the signup button bigger will increase clicks and ultimately improve revenue"
           showButtons={false}
         />
-        <Box>
+        <Box my="4">
           {!aiResponse && (
             <Flex align="start" justify="start">
               {hasAISuggestions ? (


packages/front-end/components/Experiment/EditHypothesisModal.tsx
```diff
@@ -135,16 +135,12 @@ export default function EditHypothesisModal({
...
-            name="hypothesis"
-          />
-        </div>
+        <MarkdownInput
+          value={form.watch("hypothesis")}
+          setValue={(value) => form.setValue("hypothesis", value)}
+          placeholder="e.g Making the signup button bigger will increase clicks and ultimately improve revenue"
+          showButtons={false}
+        />
         <Box>
```
msamper: nit: This needs some top padding for the preview tab. Maybe even a `<Separator>`

msamper: Scratch that.. I think that button probably shouldn't be shown in the preview tab since it lets you modify the hypothesis

lukesonnet: I'm not sure I follow. The preview tab looks fine?

msamper: Screenshot 2026-03-06 at 10 39 29 AM
This is what I'm seeing. I was suggesting adding some top padding to that check hypothesis button, but I think it actually may not belong on the preview tab if it has the ability to mutate the hypothesis you entered

lukesonnet: I think it's ok for the button since it suggests values below and then prompts you to replace yours with it if you want, It can be nice to see on the preview button. So i'll leave it but add padding, thanks!

PR #5441, opened 2026-03-06
Generate minimal SQL filters for fact tables

jdorn:
When multiple metrics from the same fact table are included in a single experiment or product analytics query, we add a rough filter to the fact table in a CTE. This looks something like this:

```sql
WITH _factTable AS (
  SELECT * FROM (
    /* user supplied fact table SQL */
  ) t
  WHERE
    -- Date filter, always applied
    timestamp >= '2026-01-01 00:00:00' AND timestamp < '2026-01-31 23:59:59'
    -- Rough metric filters, ORed together
    AND (
      (event_name='signup') -- metric 1
      OR (event_name='signup' AND device_type='mobile') -- metric 2
    )
)
```

These generated rough metric filters can be very verbose and inefficient. For example, in the above example, the filter for metric 2 is not needed at all because it's just a further refinement on top of metric 1 and thus has no effect.  This pattern is actually extremely common due to metric slices.

In some cases, the resulting filter might be so complex that it prevents the data warehouse's query optimizer from working properly and performance may be hurt as a result.

This PR adds logic to produce a minimal rough filter for a given set of metrics. So for the above, we would throw out the filter for metric 2 entirely.

### DIFF
--- a/packages/shared/test/sql.test.ts
+++ b/packages/shared/test/sql.test.ts
@@ -491,23 +491,34 @@ describe("buildMinimalOrCondition", () => {
 
   it("keeps overlapping but non-subset groups", () => {
     // {A, B} and {B, C} share condition B but neither is a subset of the other
-    expect(buildMinimalOrCondition([["A", "B"], ["B", "C"]])).toBe(
-      "((A AND B)\nOR\n(B AND C))",
-    );
+    expect(
+      buildMinimalOrCondition([
+        ["A", "B"],
+        ["B", "C"],
+      ]),
+    ).toBe("((A AND B)\nOR\n(B AND C))");
   });
 
   it("dominates a group that is a superset of multiple independent minimals", () => {
     // {A, B} and {C, D} are independent minimals
     // {A, B, C, D} is a superset of both — should be removed
     expect(
-      buildMinimalOrCondition([["A", "B"], ["C", "D"], ["A", "B", "C", "D"]]),
+      buildMinimalOrCondition([
+        ["A", "B"],
+        ["C", "D"],
+        ["A", "B", "C", "D"],
+      ]),
     ).toBe("((A AND B)\nOR\n(C AND D))");
   });
 
   it("keeps a partial-overlap chain where no group dominates another", () => {
     // {A, B}, {B, C}, {C, D} — each pair overlaps but none is a subset
     expect(
-      buildMinimalOrCondition([["A", "B"], ["B", "C"], ["C", "D"]]),
+      buildMinimalOrCondition([
+        ["A", "B"],
+        ["B", "C"],
+        ["C", "D"],
+      ]),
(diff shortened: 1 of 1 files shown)


packages/shared/src/sql.ts
```diff
@@ -256,6 +357,88 @@ export function encodeSQLResults(
...
+ */
+export function buildMinimalOrCondition(
+  filterGroups: (string | null)[][],
+): string {
+  // Remove null filters and deduplicate within each group
+  const cleanGroups = filterGroups.map((group) => [
+    ...new Set(group.filter((f): f is string => f !== null)),
+  ]);
+
+  // An empty group (no conditions) matches everything — no WHERE clause needed
```
lukesonnet: In fact, no WHERE clause is mandatory.

packages/shared/src/sql.ts
```diff
@@ -256,6 +357,88 @@ export function encodeSQLResults(
...
+  if (cleanGroups.length === 0 || cleanGroups.some((g) => g.length === 0)) {
+    return "";
+  }
+
+  // Remove groups dominated by less restrictive groups.
+  // Group i is dominated if another group j's conditions are a subset of i's
+  // (j is less restrictive, so i is redundant in the OR).
```
lukesonnet, line 296: Is this logic necessary where we scan over potentially thousands of pairs?

packages/shared/src/sql.ts
```diff
@@ -256,6 +357,88 @@ export function encodeSQLResults(
...
+      if (dominated.has(j)) continue;
+
+      const iSubJ = isSubsetOf(groupSets[i], groupSets[j]);
+      const jSubI = isSubsetOf(groupSets[j], groupSets[i]);
+
+      if (iSubJ && jSubI) {
+        // Equal sets — deduplicate, keep i
+        dominated.add(j);
+      } else if (iSubJ) {
+        // i ⊆ j — i is less restrictive, j is dominated
```
lukesonnet, line 312: special characters in code scare me

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -791,10 +785,9 @@ function generateFactTableCTE(
     `${timestampColumn} >= ${helpers.toTimestamp(dateRange.startDate)} AND ${timestampColumn} <= ${helpers.toTimestamp(dateRange.endDate)}`,
   );
 
-  // If all of the metrics are filtered, apply the filters to the raw fact table
-  // This is better for performance and makes dynamic dimensions more accurate
-  if (filters.size && !hasUnfilteredMetric) {
-    whereClauses.push(`(${Array.from(filters).join(" OR ")})`);
+  const metricsFilter = buildMinimalOrCondition(allMetricFilters);
```
lukesonnet, line 788: It might be nice to make it clear to me and future people that it's important that this is empty even if filters exist if any one metric has no filter

PR #5444, opened 2026-03-06
fix(experiments): Fix separators in MetricsSelector

msamper:
### Features and Changes

The dividers between managed and unmanaged metrics were displaying incorrectly. This PR fixes the dividers and only shows them between the two groups. If all managed metrics are selected, for example, we don't show a divider above the unmanaged metrics anymore.

### Testing

**No Official Metrics available for MetricsSelector**
- [x] No dividers should be displayed

**Official and Unofficial Metrics available for MetricsSelector**
- [x] When nothing is selected, there should only be a divider between the official and non-official metrics
- [x] When all available official metrics are selected, no divider should be rendered
- [x] When all available unofficial metrics are selected, no divider should be rendered
- [x] When there are some metrics available in official and unofficial metrics, the divider should be rendered

### Screenshots

**BEFORE**
Screenshot 2026-03-06 at 2 02 29 PM
Screenshot 2026-03-06 at 2 02 35 PM

**AFTER**
Screenshot 2026-03-06 at 2 01 53 PM
Screenshot 2026-03-06 at 2 02 02 PM

### DIFF
--- a/packages/front-end/components/Forms/MultiSelectField.tsx
+++ b/packages/front-end/components/Forms/MultiSelectField.tsx
@@ -41,6 +41,7 @@ const SortableMultiValue = SortableElement(
       e.stopPropagation();
     };
     const innerProps = { ...props.innerProps, onMouseDown };
+    // @ts-expect-error TS(2322) If you come across this, please fix it!: Type '{ innerProps: { onMouseDown: MouseEventHandl... Remove this comment to see the full error message
     return <components.MultiValue {...props} innerProps={innerProps} />;
   },
 );
@@ -55,6 +56,7 @@ const SortableMultiValueLabel = SortableHandle<any>(
 );
 
 const OptionWithTitle = (props: OptionProps<SingleValue>) => {
+  // @ts-expect-error TS(2322) If you come across this, please fix it!: Type '{ children: ReactNode; innerRef: (instance: ... Remove this comment to see the full error message
   const option = <components.Option {...props} />;
   return <div title={props.data?.tooltip}>{option}</div>;
 };


packages/front-end/components/Forms/MultiSelectField.tsx
```diff
@@ -75,6 +75,39 @@ const Input = (props: InputProps) => {
   return <components.Input onPaste={onPaste} {...props} />;
 };
 
+// Hide heading only for the first *visible* group when its label is empty.
+// If the first group has all options selected (group hidden), the second group
+// becomes the first visible one and should not show an empty heading either.
+function GroupHeading(
+  props: React.ComponentProps<typeof components.GroupHeading>,
+) {
+  const group = props.data as GroupedValue;
```
ahdriel, line 84: nit: do we need this cast here?

packages/front-end/components/Forms/MultiSelectField.tsx
```diff
@@ -75,6 +75,39 @@ const Input = (props: InputProps) => {
...
+  const label = group?.label;
+  const selectProps = props.selectProps as {
```
ahdriel, line 86: nit: same here -- shouldn't the `props` already be properly typed?

packages/front-end/components/Forms/MultiSelectField.tsx
```diff
@@ -75,6 +75,39 @@ const Input = (props: InputProps) => {
   return <components.Input onPaste={onPaste} {...props} />;
 };
 
+// Hide heading only for the first *visible* group when its label is empty.
+// If the first group has all options selected (group hidden), the second group
+// becomes the first visible one and should not show an empty heading either.
+function GroupHeading(
```
ahdriel, line 81: nit: This function logic feels more complex than it should for some reason -- but no ideas at the moment on how to simplify

packages/front-end/components/Forms/MultiSelectField.tsx
msamper: For some reason the `props` here are being shown as `any` even though I've typed them. Tried to resolve with cursor but haven't been able to successfully resolve or figure out why that is

PR #5453, opened 2026-03-09
Improves cache functionality for dashboard pa blocks

mknowlton89:
### Features and Changes

Fix dashboard explorer sidebar caching and chart type update behavior

#### Problem
The ExplorerSideBar rendered inside the dashboard block editor was not following the same caching contract as the standalone explore page. Three related issues were identified:

1. Update button always forced a fresh query (cache=never)

The Update button in the dashboard sidebar was calling handleSubmit({ force: true }), which bypasses the cache entirely on every click. The correct behavior is: auto-fetch config changes with cache=required (non-managed warehouse) or cache=preferred (managed warehouse), and only use cache=never when the user manually triggers a refresh after a cache miss (stale state).

2. Front-end-only chart type changes incorrectly invalidated the cached analysis

Switching between chart types within the same category (e.g. line → area) doesn't require new data — compareConfig/toFetchKey already classifies these as needsFetch=false. However, the ProductAnalyticsExplorerSideBarWrapper effect was unconditionally clearing explorerAnalysisId on any needsUpdate, including these purely cosmetic changes. This caused the dashboard block to drop its existing result and prompt the user to click Update unnecessarily.

3. Dashboard block rendered with the server-stored chart type instead of the locally-updated config

ProductAnalyticsExplorerBlock was passing data?.exploration.config (the config baked into the server-stored exploration result) as submittedExploreState to ExplorerChart. This meant front-end-only chart type changes were never visually reflected until a full re-run, even if the underlying data was still valid.

##### ExplorerContext.tsx
Exposed needsFetch on ExplorerContextValue so consumers can distinguish between data-requiring changes and front-end-only rendering changes.

##### ProductAnalyticsExplorerSideBarWrapper.tsx
Updated the block-sync effect to only clear explorerAnalysisId when needsFetch=true. Front-end-only changes (e.g. line → area) now preserve the existing analysis ID, so the dashboard block continues to display results without prompting the user to re-run.

##### ProductAnalyticsExplorerBlock.tsx
Changed submittedExploreState to use block.config (falling back to data?.exploration.config), so the chart renders with the current local config rather than the config that was in effect when the analysis was last run. This allows front-end-only changes to be reflected immediately.

##### ExplorerSideBar.tsx
- Changed the Update button's onClick from handleSubmit() to handleSubmit({ force: isStale }). This restores the correct two-phase flow: auto-fetches use the normal cache strategy (cache=required/cache=preferred), and the Update button only forces cache=never when there is an actual cache miss (stale state).
- Also updated the sidebar when rendering in the dashboard to hide the `Date Granularity` selector for cumulative chart options. 

### Dependencies

- None

### Testing

- [x] Ensure that a chart type change doesn't result in a new data fetch call, and the chart on the dashboard updates automatically.
- [x] Ensure that clicking the "Update" button follows the same pattern used on the exploration pages (if there was no existing fetch ref or the data source was a managed warehouse, it uses the `cache=preferred` option, otherwise it uses the `cache=required` option, and only passed the `cache=never` option when the data was stale.

### DIFF
--- a/.cursor/rules/backend/model-patterns.md
+++ b/.cursor/rules/backend/model-patterns.md
@@ -32,7 +32,7 @@ const BaseClass = MakeModelClass({
     updateEvent: "myResource.update",
     deleteEvent: "myResource.delete",
   },
-  globallyUniqueIds: true, // IDs unique across all orgs
+  globallyUniquePrimaryKeys: true, // IDs unique across all orgs
   defaultValues: {
     // Default field values
     description: "",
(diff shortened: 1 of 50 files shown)


mknowlton89:
@cursor review

Issue #5458, opened 2026-03-10
[Bug] Uploading screenshots to S3 errors when AWS_ASSUME_ROLE is set


nodirnasirov:
### Summary

on latest there is error when trying to upload screenshots with provided AWS_ASSUME _ROLE:
```
 Cannot read properties of null (reading 'masterCredentials') error
```
#### Expected Behavior

screenshots should be uploaded with no error

#### Current Behavior

fails to upload screenshots.

PR #5459, opened 2026-03-10
fixes s3 credentials refresh

nodirnasirov:
## Features and Changes
Fix Cannot read properties of null (reading 'masterCredentials') error when uploading files via S3 pre-signed URLs with an assumed IAM role (`AWS_ASSUME_ROLE`).
The AWS SDK v2 TemporaryCredentials constructor does not eagerly resolve `masterCredentials` — it only does so on the first call to `refresh()`. Synchronous SDK operations such as getSignedUrl and `createPresignedPost` attempt to access masterCredentials before it is populated, causing a null reference error.
The fix explicitly calls `refresh()` immediately after constructing `TemporaryCredentials`, ensuring credentials are fully initialized before the S3 client is used for any operation.
Closes [#5458](https://github.com/growthbook/growthbook/issues/5458)
## Dependencies
None
## Testing
Configure the app with `AWS_ASSUME_ROLE` set to a valid IAM role ARN and `UPLOAD_METHOD=s3`.
Attempt to upload a file through the UI.
Verify the upload completes successfully without the `masterCredentials` error.
## Screenshots
N/A — backend-only change, no UI impact.

### DIFF
--- a/packages/back-end/src/services/files.ts
+++ b/packages/back-end/src/services/files.ts
@@ -15,6 +15,30 @@ import {
 let s3: AWS.S3;
 let awsTempCredentials: AWS.TemporaryCredentials | null = null;
 
+function shouldRefreshCredentials(): boolean {
+  if (!awsTempCredentials) return false;
+
+  // Check if credentials need refresh using the SDK's built-in method
+  if (awsTempCredentials.needsRefresh()) return true;
+
+  // Additional safety check: if expireTime is not set, credentials may not have been loaded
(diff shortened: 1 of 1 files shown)


Kevin-Chant (approved):
This is a bit hard to verify locally (the root problem only occurs when local credentials aren't present, but that means the assume role fails because it's not running in an EC2 context).
I'll approve and merge it for now, @nodirnasirov could you ping me on slack once you've been able to verify it fixes the issue for you?

PR #5463, opened 2026-03-10
refactor(snapshot): Split snapshot into plan + creation

ahdriel:
### Features and Changes

For dashboards we had a weird condition where we created and deleted a snapshot because a lot of the values were calculated/generated inside the same function, but we avoided running the analysis which is expensive.

Now we are splitting into into a plan phase and a creation + analysis phase.

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -2960,6 +2960,15 @@ export async function createExperimentSnapshotFromPlan({
 }> {
   const metricMap = await getMetricMap(context);
   const factTableMap = await getFactTableMap(context);
+  const metricGroups = await context.models.metricGroups.getAll();
+
+  expandAllSliceMetricsInMap({
+    metricMap,
+    factTableMap,
+    experiment,
+    metricGroups,
+  });
+
   const queryRunner = await createSnapshotFromPlan({
     plan,
     context,


packages/back-end/src/services/experiments.ts
```diff
@@ -1037,7 +1038,132 @@ export function getAdditionalQueryMetadataForExperiment(
   };
 }
 
-export async function createSnapshot({
+export type ExperimentSnapshotQueryRunner =
+  | ExperimentResultsQueryRunner
+  | ExperimentIncrementalRefreshQueryRunner
+  | ExperimentIncrementalRefreshExploratoryQueryRunner;
```
lukesonnet, line 1044: The `ExperimentReportQueryRunner` is also a snapshot query runner, but it's only used for legacy reports. IDK if we care about that, we can probably ignore it.

packages/back-end/src/services/experiments.ts
```diff
@@ -1037,7 +1038,132 @@ export function getAdditionalQueryMetadataForExperiment(
...
+  experiment,
+  snapshotType,
+}: {
+  allowIncrementalRefresh: boolean;
+  isExperimentCompatibleWithIncrementalRefresh: boolean;
+  datasource: DataSourceInterface;
+  experiment: ExperimentInterface;
+  snapshotType: SnapshotType;
+}): SnapshotQueryRunnerKind {
+  if (
```
lukesonnet, line 1084: Seems like inverting this logic could make it more legible.

```
allowIncrementalRefresh &&
isIncrementalRefreshEnabledForSnapshot({ datasource, experiment }) &&
    (experiment.type === undefined || experiment.type === "standard") &&
    isExperimentCompatibleWithIncrementalRefresh
```

Then one of 2 incrementals

packages/back-end/src/services/experiments.ts
```diff
@@ -1144,17 +1266,64 @@ export async function createSnapshot({
...
+    fullRefresh,
+  });
+
+  return {
+    snapshot: data,
+    runnerKind,
+    useCache,
+    fullRefresh,
+    settingsForSnapshotMetrics,
+    defaultAnalysisSettings,
```
lukesonnet: Why are these fields separate from the copies of them that exist in the `snapshot: data`?

packages/back-end/src/services/experiments.ts
```diff
@@ -1168,106 +1337,76 @@ export async function createSnapshot({
...
+  };
+
+  if (plan.runnerKind === "incremental") {
+    await (
+      queryRunner as ExperimentIncrementalRefreshQueryRunner
+    ).startAnalysis({
+      ...analysisProps,
+      experimentId: experiment.id,
+      incrementalRefreshStartTime: new Date(),
+      fullRefresh: plan.fullRefresh && plan.snapshot.type === "standard",
```
lukesonnet, line 1394: Should we store analysisType in the plan to be used here?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2920,14 +2921,78 @@ export async function createExperimentSnapshot({
...
+  });
+
+  return createExperimentSnapshotFromPlan({
+    plan,
+    context,
+    experiment,
+  });
+}
+
+export async function createExperimentSnapshotFromPlan({
```
lukesonnet, line 2949: nit: the name of this is weird since it returns the snapshot and the query runner

`createExperimentSnapshotQueryRunner`? Eh, maybe that's more confusing

packages/back-end/src/services/experiments.ts
ahdriel: Good callout! Let's ignore it for now as it goes through a different code path (otherwise TS would be complaining here).

But I'll be looking into further refactor/reorg of this code tomorrow, so I'll keep this in mind.

packages/back-end/src/routers/dashboards/dashboards.controller.ts
```diff
@@ -196,19 +193,19 @@ export async function refreshDashboardData(
...
+    const plannedExperimentMainSnapshot = await planExperimentSnapshot({
+      context,
+      experiment,
+      dimension: undefined,
+      datasource,
+      phase: experiment.phases.length - 1,
+      useCache: false,
+      triggeredBy: "manual-dashboard",
+      type: "standard",
+      allowIncrementalRefresh: false,
```
lukesonnet: IMO I would remove this

PR #5464, opened 2026-03-11
Add API endpoint for listing experiment templates

Kevin-Chant:
### Features and Changes

Adds a rest endpoint for listing experiment templates (filtered by `projectId`)
Started in #5462 by @dannylin-ant

Also has a minor improvement to `apiConfig` to make overriding query params for list endpoints easier in the future

### Testing

- [x] Hit the new endpoint with and without `projectId=` to check that the filtering works appropriately
- [x] Check that the docs generated properly with the new param format

### Screenshots

image

image

### DIFF
--- a/packages/back-end/src/api/ApiModel.ts
+++ b/packages/back-end/src/api/ApiModel.ts
@@ -11,6 +11,7 @@ import {
 } from "back-end/src/util/handler";
 import { MetricGroupModel } from "back-end/src/models/MetricGroupModel";
 import { TeamModel } from "back-end/src/models/TeamModel";
+import { ExperimentTemplatesModel } from "back-end/src/models/ExperimentTemplateModel";
 import {
   CustomApiHandler,
   CrudAction,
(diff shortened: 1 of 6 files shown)


packages/back-end/src/api/ApiModel.ts
```diff
@@ -41,7 +43,7 @@ type CrudValidatorShapes<T extends ApiBaseSchema> = {
     z.ZodNever
   >;
   get: ApiRequestValidator<z.ZodType<{ id: string }>, z.ZodNever, z.ZodNever>;
-  list: ApiRequestValidator<z.ZodNever, z.ZodNever, z.ZodNever>;
+  list: ApiRequestValidator<z.ZodNever, z.ZodNever, z.ZodTypeAny>;
```
Kevin-Chant, line 46: We'll probably want a similar change for the other validators as well, but that's out of scope for this PR

packages/shared/src/validators/experiment-template.ts
```diff
@@ -53,6 +50,57 @@ export type ExperimentTemplateInterface = z.infer<
   typeof experimentTemplateInterface
 >;
 
+export const apiExperimentTemplateValidator = apiBaseSchema.safeExtend({
```
Kevin-Chant, line 53: The fields are duplicated intentionally so that changes to the model don't affect the API without dev interaction

PR #5466, opened 2026-03-11
feat(incremental): Ensure we don't run 2 IncrementalRefreshQueryRunner at the same time

ahdriel:
### Features and Changes

Previously our QueryRunners ran only `read-only` queries so the concurrency was a nice-to-have but not a hard requirement.

With IncrementalRefresh that is not true anymore, so concurrent runs will actually impact one another and drop tables, potentially causing the queries to fail or inserting corrupted data, depending on the order that they would finish.

This PR fixes it with the following approach:
- Add a lock for `IncrementalRefresh` when creating the snapshot
	- If unable to acquire the lock, throw an error
- Background jobs would also see the error and not do any updates
	- `stale` updateSchedules means that the already running snapshot will modify `nextScheduleUpdate` and this won't be reattempted. So the in-progress becomes the last/most up to date one.
	- `cron` updateSchedule will be retried after 10 minutes

Also ensures we refresh the UI with the latest snapshot in case this error happens to try to recover the inconsistent state.

### DIFF
--- a/packages/back-end/src/models/IncrementalRefreshModel.ts
+++ b/packages/back-end/src/models/IncrementalRefreshModel.ts
@@ -9,7 +9,7 @@ import { MakeModelClass } from "./BaseModel";
 export const COLLECTION_NAME = "incrementalrefresh";
 
 // If a lock hasn't been updated in this long, consider it stale
-const STALE_LOCK_TIMEOUT_MS = 30 * 60 * 1000; // 30 minutes
+const STALE_LOCK_TIMEOUT_MS = 60 * 60 * 1000;
 
 const BaseClass = MakeModelClass({
   schema: incrementalRefreshValidator,
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -3134,33 +3134,42 @@ export async function postSnapshot(
 
   const useCache = !req.query["force"];
 
-  const { snapshot } = await createExperimentSnapshot({
-    context,
-    experiment,
-    datasource,
-    dimension,
-    phase: phase,
-    useCache,
-    type: experiment.type === "multi-armed-bandit" ? "exploratory" : undefined,
-  });
-
-  await req.audit({
-    event: "experiment.refresh",
-    entity: {
-      object: "experiment",
-      id: experiment.id,
-    },
-    details: auditDetailsCreate({
-      phase,
+  try {
+    const { snapshot } = await createExperimentSnapshot({
+      context,
+      experiment,
+      datasource,
       dimension,
+      phase,
       useCache,
-      manual: false,
-    }),
-  });
-  res.status(200).json({
-    status: 200,
-    snapshot,
-  });
+      type:
+        experiment.type === "multi-armed-bandit" ? "exploratory" : undefined,
+    });
+
+    await req.audit({
+      event: "experiment.refresh",
+      entity: {
+        object: "experiment",
+        id: experiment.id,
+      },
+      details: auditDetailsCreate({
+        phase,
+        dimension,
+        useCache,
+        manual: false,
+      }),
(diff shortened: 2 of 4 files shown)


packages/back-end/src/models/IncrementalRefreshModel.ts
```diff
@@ -42,9 +43,94 @@ export class IncrementalRefreshModel extends BaseClass {
...
+        },
+        {
+          $set: {
+            currentExecutionSnapshotId: snapshotId,
+            dateUpdated: new Date(),
+          },
+          $setOnInsert: {
+            organization: this.context.org.id,
+            experimentId,
+            unitsTableFullName: null,
```
lukesonnet, line 81: Should we refactor "default" new settings to one place: https://app.graphite.com/github/pr/growthbook/growthbook/5466/Av%2Fincremental-refresh-split#lines-packages/back-end/src/models/IncrementalRefreshModel.ts-R38-R47

packages/back-end/src/services/experiments.ts
```diff
@@ -1038,6 +1041,428 @@ export function getAdditionalQueryMetadataForExperiment(
...
+  incomingIntent: ExperimentSnapshotRefreshIntent;
+  context: ReqContext | ApiReqContext;
+}): Promise<ExperimentSnapshotInterface> {
+  const existing = snapshot.refreshIntent ?? {};
+  const mergedIntent: ExperimentSnapshotRefreshIntent = {
+    banditReweightRequested:
+      existing.banditReweightRequested ||
+      incomingIntent.banditReweightRequested ||
+      false,
+    triggeredBySchedule:
```
lukesonnet: This feels kind of like a lie. I can't quite think through scenarios where this triggeredBy is relied on but then might be wrong since it was  in fact the user triggered snapshot, but it feels a bit worse than the banditRewieght one. Fine to leave it for now, though.

packages/back-end/src/services/experiments.ts
```diff
@@ -1038,6 +1041,428 @@ export function getAdditionalQueryMetadataForExperiment(
...
+          throw new Error(SNAPSHOT_UPDATE_IN_PROGRESS_ERROR);
+        }
+        if (incomingIntent) {
+          // Merge intent into the running snapshot and return it
+          const merged = await mergeRefreshIntent({
+            snapshot: activeSnapshot,
+            incomingIntent,
+            context,
+          });
+          return { snapshot: merged };
```
lukesonnet: I don't understand why if and only if the incoming intent is set do we just merge and return the new snapshot.

packages/back-end/src/services/experiments.ts
```diff
@@ -1384,114 +1795,228 @@ export async function createSnapshotFromPlan({
...
+        autoSnapshots: false,
+      },
+    });
+
+    await notifyAutoUpdate({ context, experiment, success: true });
+  } catch (error) {
+    // TODO: Is this new or already existing behavior?
+    logger.error(
+      error,
+      "Failed to turn off autoSnapshots after snapshot failure: " +
```
lukesonnet:
```suggestion
      "Failed to turn off autoSnapshots after scheduled snapshot failure: " +
```

packages/back-end/src/services/experiments.ts
```diff
@@ -1038,6 +1041,428 @@ export function getAdditionalQueryMetadataForExperiment(
...
+      queryRunner: queryRunner as ExperimentIncrementalRefreshQueryRunner,
+      metricMap,
+      factTableMap,
+      isIncremental: true,
+    });
+
+    return { snapshot: queryRunner.model, queryRunner };
+  }
+
+  // --- Incremental-exploratory: block if incremental writer is active ---
```
lukesonnet: This will break the current dashboard updates.

PR #5467, opened 2026-03-11
Ji/docs api self hosted fix

tzjames:
### Features and Changes

Docusaurus was interpreting 
  `- url: {domain}/api/v1"` as a relative url and inserting `docs.growthbook.io/redocusaurus/` before it.  By making it an absolute path, it displays as intended.

It also removes the variable: domain, as it wasn't be properly replaced, and there was no place to set it even for docs.growthbook.io that I am aware of.

### Screenshots

Before:
Screenshot 2026-03-11 at 5 52 34 PM

After:
Screenshot 2026-03-11 at 6 01 24 PM

### DIFF
--- a/packages/back-end/src/api/openapi/openapi.yaml
+++ b/packages/back-end/src/api/openapi/openapi.yaml
@@ -51,7 +51,7 @@ info:
 servers:
   - url: "https://api.growthbook.io/api/v1"
     description: GrowthBook Cloud
-  - url: "http://{domain}/api/v1"
+  - url: "https://{domain}/api/v1"
     description: Self-hosted GrowthBook
 tags:
   - name: projects
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -51,7 +51,7 @@ info:
 servers:
   - url: 'https://api.growthbook.io/api/v1'
     description: GrowthBook Cloud
-  - url: 'http://{domain}/api/v1'
+  - url: 'https://{domain}/api/v1'
     description: Self-hosted GrowthBook
 tags:
   - name: projects


packages/back-end/src/api/openapi/openapi.yaml
```diff
@@ -51,11 +51,8 @@ info:
 servers:
   - url: "https://api.growthbook.io/api/v1"
     description: GrowthBook Cloud
-  - url: "{domain}/api/v1"
+  - url: "http://{domain}/api/v1"
     description: Self-hosted GrowthBook
-    variables:
-      domain:
-        default: http://localhost:3100
```
jdorn, line 58: Is removing this variable intentional?

tzjames: Yes.  That variable does nothing. There was no way to set it anywhere, and it wasn't actually being replaced in the actual url.  It is an open question though of whether we want it to show {domain} or whether we want it to show localhost:3100, and then in the description say "Self-hosted Growthbook (replace localhost:3100 with your api domain)" or anything else.

tzjames: or maybe it should show {API_HOST} was that is what the env var is for running the docker container.

jdorn: The variable is not just for our docs.  It's part of the OpenAPI spec.  So anyone using a tool like Postman that can read OpenAPI specs can show an input field for that variable.  https://learn.openapis.org/specification/servers.html#server-variables

tzjames: Well when it comes to Postman, it uses the first server url and makes a `baseUrl` variable out of it automatically.  Subsequent urls are ignored.  There is a [long standing issue](https://github.com/postmanlabs/openapi-to-postman/issues/8) out for changes here, but has never been done. 

So in terms of Postman there is no need for the second url at all.  However other tools might behave differently.  For docs it just needs to start with `http(s)://` for it to be treated as absolute and it not insert the docs url into it.  If we still want to have it like it currently is for other tools, we could create a separate api-spec-for-docs.yaml and have the second url be hardcoded as whatever we want as long as it starts with `http(s)://` ... presumably `https://{domain}/api/v1`?

PR #5474, opened 2026-03-12
Allow API Keys to take arbitrary roles

Kevin-Chant:
### Features and Changes

Expands the API Key `role` field to allow any role in the organization (including custom roles when applicable)

### Testing

- [x] Existing key role (`admin`)
  - [x] Can create key and see in UI
  - [x] Using key allows read operations
  - [x] Using key allows insert/update
- [x]  Default role secret key (`collaborator`)
  - [x] Can create key and see in UI
  - [x] Using key allows read operations
  - [x] Using key fails permission check on insert/update
- [x] Custom role secret key (custom role with Saved Groups Full Access)
  - [x] Can create key and see in UI
  - [x] Using key allows read operations
  - [x] Using key fails permission check on insert/update (e.g. feature)
  - [x] Using key allows insert/update operations for Saved Groups
- [x] Creating key with deactivated role fails validation
- [x] Deleting a custom role used by an API key fails validation

### Screenshots

image

### DIFF
--- a/packages/front-end/components/Settings/ApiKeysModal.tsx
+++ b/packages/front-end/components/Settings/ApiKeysModal.tsx
@@ -28,7 +28,9 @@ const ApiKeysModal: FC<{
       options: roleList
         .filter(
           (role) =>
-            !deactivatedRoles.includes(role.id) && role.id in DEFAULT_ROLES,
+            !deactivatedRoles.includes(role.id) &&
+            role.id in DEFAULT_ROLES &&
+            role.id !== "noaccess",
         )
         .map((role) => ({
           label: role.displayName || role.id,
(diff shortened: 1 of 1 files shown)


packages/front-end/components/PersonalAccessTokens/PersonalAccessTokens.tsx
tzjames: This could be updated to reflect that they can create keys with any specific role.

packages/front-end/components/Settings/ApiKeysModal.tsx
```diff
@@ -9,18 +11,59 @@ import SelectField from "@/components/Forms/SelectField";
...
+      options: roleList
+        .filter(
+          (role) =>
+            !deactivatedRoles.includes(role.id) && role.id in DEFAULT_ROLES,
+        )
+        .map((role) => ({
+          label: role.displayName || role.id,
+          value: role.id,
+        })),
+    };
```
tzjames, line 39: We should probably filter out "No access role" here.  It just doesn't make sense to make an api key with no access.

PR #5475, opened 2026-03-12
Add date range to test SQL queries to reduce amount of data scanned

jdorn:
### Features and Changes

Throughout the product, any time we run a test SQL query (e.g. to validate the entered SQL is correct), we should include a date range filter whenever possible to limit the amount of data scanned (and the associated warehouse costs).

We were already passing in `startDate` and `endDate` SQL template variables, but that requires the user to remember to reference those from their SQL.  This PR also adds a `timestamp > startDate` WHERE clause to every test query that we know has a `timestamp` column.  It's also flexible enough to work in the future if we allow custom timestamp column names.

In addition, this PR improves client-side SQL validation and error messages, generates better default fact table SQL, and improves the fact table creation flow UX.

### Testing
* [x] Refreshing fact table columns
* [x] Update fact table SQL and click the "Test Query" button (also verify column list is updated correctly)
* [x] Update saved filter SQL
* [x] Add a `sql_expr` filter to a fact metric
* [x] Edit legacy metric SQL and click the "Test Query" button
* [x] Edit exposure SQL and click the "Test Query" button
* [x] Run Feature Diagnostics query

Some queries will NOT include a date range filter since they don't have a `timestamp` column. We should still test these for regressions:
* [ ] Edit identity join SQL and click the "Test Query" button
* [ ] Edit dimension SQL and click the "Test Query" button
* [x] Edit segment SQL and click the "Test Query" button (this does have a `date` column, but we aren't supporting that in this PR right now)


jdorn:
@greptileai

packages/front-end/components/FactTables/FactTableModal.tsx
```diff
@@ -320,7 +288,7 @@ export default function FactTableModal({
           </div>
         ) : null}
 
-        {hasCommercialFeature("metric-slices") && (
+        {hasCommercialFeature("metric-slices") && !!existing && (
```
lukesonnet, line 291: I think the goal is to streamline creation but this seems like something an experienced user would know to turn on when creating a new fact table.

packages/front-end/components/FactTables/FactTableModal.tsx
```diff
@@ -265,47 +264,16 @@ export default function FactTableModal({
                 }}
               >
                 {form.watch("sql") ? "Edit" : "Add"} SQL <FaExternalLinkAlt />
-              </button>
+              </Button>
             </div>
           </div>
         )}
 
-        {selectedDataSource && (!existing?.id || duplicate) && (
```
lukesonnet, line 273: I would actually love it if we could make this easier to edit instead of harder, but I understand why it might not belong here.

For advanced users, they don't necessarily want identifier type to be a coincidence, they want clear control over what identifier types are linked to a table.

packages/front-end/services/datasources.ts
```diff
@@ -161,6 +199,14 @@ FROM
     return [];
   },
   userIdTypes: ["user_id"],
+  getFactTableSQL: (tablePrefix, userIdTypes) => {
+    return `SELECT
+  ${userIdTypes.map((ut) => `${ut},`).join("\n  ")}
```
lukesonnet, line 204: I don't know that this encourages good practice here and in the other locations where userIdTypes is available.

If we have a succinct list we know exist for a tracker/DWH combo and I feel we should use those instead of allowing them to get passed in from the user's configured DataSource.

lukesonnet (approved):
I have some minor reservations as noted in individual comments, but overall this strikes me as an improvement, and cursory QA behaves as expected.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1656,16 +1658,24 @@ export default abstract class SqlIntegration
...
+    const tableWhereClause = timestampColumn
+      ? `WHERE ${timestampColumn} >= ${this.toTimestamp(startDate)}`
+      : "";
+
     const limitedQuery = compileSqlTemplate(
       `WITH __table as (
         ${query}
       )
-      ${this.selectStarLimit("__table", limit)}`,
+      ${this.selectStarLimit(`__table ${tableWhereClause}`, limit)}`,
```
ahdriel: I guess this is fine but it does not feel right 😂 

Should we have `where` as a parameter too?

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -75,6 +75,8 @@ async function testFilterQuery(
     throw new Error("Testing not supported on this data source");
   }
 
+  const timestampColumn = "timestamp";
```
ahdriel, line 78: nit: should we move this to a shared `DEFAULT_TIMESTAMP_COLUMN`?

packages/front-end/services/datasources.ts
```diff
@@ -664,7 +843,8 @@ export function validateSQL(sql: string, requiredColumns: string[]): void {
     (col) => !sql.toLowerCase().includes(col.toLowerCase()),
   );
 
-  if (missingCols.length > 0) {
+  // Allow `SELECT *` queries, but otherwise look for all required columns in the query
+  if (missingCols.length > 0 && !sql.match(/SELECT\s+\*/i)) {
```
ahdriel, line 847: nit: Can we add a test for this new behavior?

packages/front-end/services/datasources.ts
```diff
@@ -647,6 +807,25 @@ export function getInitialMetricQuery(
   ];
 }
 
+export function getInitialFactTableQuery(
+  datasource: DataSourceInterfaceWithParams,
+): [string[], string] {
```
ahdriel: nit: Should we return an object to reduce the chance of confusion between the strings and what they mean?

packages/back-end/src/routers/fact-table/fact-table.controller.ts
jdorn: I want to fully refactor how we deal with timestamps at some point, but didn't want to open that can of worms in this PR

packages/front-end/services/datasources.ts
jdorn: This is only for people who selected "Custom" as the event tracker.  For all the other ones, we hard-code the list of identifiers based on what we know about the event tracker.

jdorn:
cursor review

PR #5480, opened 2026-03-12
perf: Virtualize MetricsSelector dropdown with react-window

itsgrimetime:
## Summary

- Add `react-window` to virtualize the MetricsSelector dropdown menu list
- Only ~8-10 visible options are rendered instead of all matching options
- Implemented as an opt-in `virtualized` prop on `MultiSelectField`, enabled in `MetricsSelector`

## Context

Builds on #5469. After memoizing data computation, the remaining bottleneck is that react-select renders a `MetricName` component for **every matching option** in the DOM, even though only ~8-10 are visible in the scroll viewport. With hundreds of metrics, each keystroke triggers hundreds of component renders (each with `useDefinitions()` context lookups, `Tooltip`, `OfficialBadge` sub-components, etc.).

Virtualizing with `react-window` cuts rendered options from hundreds down to ~10, eliminating the remaining lag.

### Changes

| File | Change |
|------|--------|
| `VirtualizedMenuList.tsx` | New component: lightweight `react-window` `FixedSizeList` wrapper for react-select's `MenuList` |
| `MultiSelectField.tsx` | Add opt-in `virtualized` prop that swaps in `VirtualizedMenuList` |
| `MetricsSelector.tsx` | Enable `virtualized` on the multi-select |
| `package.json` | Add `react-window` (~6kb gzipped) + `@types/react-window` |

## Test plan

- [ ] Open an experiment's settings and edit metrics (Goal, Secondary, Guardrail)
- [ ] Type in the metric search box — filtering should be snappy with no lag
- [ ] Scroll through the dropdown — options should render smoothly as they enter the viewport
- [ ] Keyboard navigation (arrow keys) should still work correctly
- [ ] Verify selecting/deselecting metrics works as before
- [ ] Verify metric groups display correctly in the dropdown
- [ ] Verify other MultiSelectField usages (without `virtualized`) are unaffected


jdorn:

PR #5493, opened 2026-03-16
Get rid of stale usage flags

tzjames:
### Features and Changes
Self-hosted organizations that have blocked growthbook won't be using all of the latest features.  Here we remove our stale features.  I've replace all of them with the true variant, except `sdk-connections-new-empty-state` which didn't win it's experiment.

I've gotten rid of the experiments according to this [list](https://docs.google.com/spreadsheets/d/1NPQysc8rS8_5HSupN6zJ4gshvx-eYMXsik2Rqq6nvKw/edit?gid=1900767756#gid=1900767756), with the exception of `feature-usage-auto-refresh-interval` which I kept around for now since it seems like we may want to tweak it at some point, however I set it's default value in code to be what is on prod now.

Besides that the logic should be identical to what was serving with the exception of `factMetricList` where we hadn't been checking whether they had `metricSlices` commercial feature but now we do.

I also got rid of an argument to `useRealtimeData` because we never called it with `update=false`.

### Testing

- [x] Tested the SDKConnectionsList in the empty state: On a new db, I went to http://localhost:3000/sdks and clicked add New connections.
- [x] Tested NewExperimentForm and saw that similar experiments still work.

### DIFF
--- a/packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
+++ b/packages/front-end/components/Features/SDKConnections/SDKConnectionsList.tsx
@@ -10,8 +10,6 @@ import {
   getDisallowedProjects,
 } from "shared/util";
 import clsx from "clsx";
-import type { SDKLanguage } from "shared/types/sdk-connection";
-import { getLatestSDKVersion } from "shared/sdk-versioning";
 import { useDefinitions } from "@/services/DefinitionsContext";
 import LoadingOverlay from "@/components/LoadingOverlay";
 import { GBAddCircle, GBHashLock, GBRemoteEvalIcon } from "@/components/Icons";
@@ -45,8 +43,6 @@ export default function SDKConnectionsList() {
   const canCreateSDKConnections =
     permissionsUtil.canViewCreateSDKConnectionModal(project);
 
-  const [initialModalSelectedLanguage] = useState<SDKLanguage | null>(null);
-
   if (error) {
     return <div className="alert alert-danger">{error.message}</div>;
   }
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -202,8 +193,6 @@ export default function RuleModal({
     "multi-armed-bandits",
   );
```
Kevin-Chant, line 204: Why was the commercialFeature check removed?

packages/front-end/components/Layout/Layout.tsx
```diff
@@ -215,7 +210,6 @@ const navlinks: SidebarLinkProps[] = [
         path: /^presentation/,
       },
     ],
-    filter: ({ gb }) => !gb?.isOn("insights"),
```
Kevin-Chant, line 218: Actually it looks like since this is an inverted check the whole section should be removed

packages/front-end/components/Settings/EditDataSource/DataSourcePipeline/PipelineModeSelector.tsx
```diff
@@ -46,7 +40,7 @@ const PipelineModeSelector = ({
     },
   ];
 
-  if (isIncrementalPipelineAvailable || value === "incremental") {
+  if (value === "incremental") {
```
Kevin-Chant: If collapsing to `true || value === "incremental` this option should just be part of the original list

PR #5494, opened 2026-03-16
Use tsgo for type-checking, skipLibCheck:true, fix lint

tzjames:
### Features and Changes
This is a redo of `https://github.com/growthbook/growthbook/pull/4127` now that we are on pnpm, and also tsgo nightly has been updated a bunch.  `ci` runs with most things cached is down to 3 min 30s from 7-8min beforehand.

#### ts-go for type-checking.  A lot of other things needed to change for this to work.
- had to set `"useUnknownInCatchVariables": false,` to avoid a bunch of lint
- skiplibcheck:true needed to be used for all tsconfigs
- a `date-fns-tz` and `dotenv` needed to be upgraded
- a bunch of lint needed to be fixed.

A bunch of improvents to ci:
#### GitHub Actions (Node.js 24 support)
- **actions/checkout**: v2/v3/v4 → v5
- **actions/cache**: v4 → v5
- **actions/setup-node**: v4 → v5
These versions run on Node.js 24 and address the deprecation warning.

#### CI workflow
- Reordered steps: pnpm install before Python setup
- **ESLint cache**: key had been based on source files, so basically the cache was never used. When files change ESLint still re-lints changed files via its own cache
- **tsbuildinfo cache**: added for faster type-checking with `incremental:true` property in tsconfig
- **Poetry cache**: added for `packages/stats/.venv`
- **setup-python**: v5 → v6, with `cache-dependency-path` for pip in order for it not to hit a circular dependency and take forever
- **Poetry**: `virtualenvs.in-project true` for in-repo venv in order for the poetry cache to work.

#### package.json
- `type-check` no longer uses `--workspace-concurrency=1` - this make it run a few seconds faster .. at least locally
- `jest` removed `--forceExit`, `--verbose`, `--detectOpenHandles`- these are mostly for debugging and made the tests run slower
- `lint` will now first run without `--fix` (which takes ~2s) then only if there are issues will run it with `--fix` which takes ~22s.  This means it will run slightly slower when there are issues but if there are no issues (which is probably the majority of the time it is run) it will run much faster.

### Testing
`pnpm type-check`
See ci passes.


packages/back-end/src/enterprise/licenseUtil.ts
```diff
@@ -69,7 +72,7 @@ export function getSubscriptionFromLicense(
   };
 }
 
-type MinimalOrganization = {
+export type MinimalOrganization = {
```
Kevin-Chant: Nit: unnecessary export
```suggestion
type MinimalOrganization = {
```

packages/back-end/src/events/handlers/slack/slack-event-handler-utils.ts
```diff
@@ -1588,7 +1588,7 @@ export function formatDiffForSlack(
           }
 
           if (value.modified && value.modified.length > 0) {
-            value.modified.forEach((change: ModificationItem) => {
+            (value.modified as ModificationItem[]).forEach((change) => {
```
Kevin-Chant: The amount of casts strikes me as a code smell, but I don't have a great solution since the root issue for each is likely pretty complex.

Kevin-Chant: E.g. I looked into this one and it seems to be due to having two separate definitions of `HierarchicalValue` with different schemas

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5260,8 +5260,8 @@ export default abstract class SqlIntegration
             return `event_time BETWEEN '${start}' AND '${end}'`;
           },
           getAdditionalEvents: () => [],
-          getEventFilterWhereClause: (eventName: string) =>
-            `event_name = '${eventName}'`,
+          getEventFilterWhereClause: (eventName?: string) =>
+            eventName !== undefined ? `event_name = '${eventName}'` : "",
```
Kevin-Chant: Nit: prefer updating the signature in `shared/types/datasource.d.ts` to be non-optional

packages/back-end/src/models/BaseModel.ts
```diff
@@ -1016,7 +1018,7 @@ export abstract class BaseModel<
     if (this.config.globallyUniquePrimaryKeys) {
       promises.push(
         this._dangerousGetCollection()
-          .createIndex(pKeyIndex, { unique: true })
+          .createIndex(pKeyIndex as Record<string, 1 | -1>, { unique: true })
```
Kevin-Chant: Nit: unnecessary cast
```suggestion
          .createIndex(pKeyIndex, { unique: true })
```

packages/back-end/src/util/http.util.ts
```diff
@@ -88,7 +88,8 @@ export const cancellableFetch = async (
 
   try {
     response = await fetch(url, {
-      signal: abortController.signal,
+      signal:
+        abortController.signal as import("node-fetch").RequestInit["signal"],
```
Kevin-Chant: Can this just be an `import type` at the top of the file? Not a huge fan of inline imports

packages/back-end/src/util/logger.ts
```diff
@@ -89,32 +89,32 @@ export const httpLogger = pinoHttp({
...
+  trace: (...args: unknown[]) => {
+    httpLogger.logger.trace(...(args as Parameters<BaseLogger["trace"]>));
   },
-  warn(...args: Parameters<BaseLogger["warn"]>) {
-    httpLogger.logger.warn(...args);
+  warn: (...args: unknown[]) => {
+    httpLogger.logger.warn(...(args as Parameters<BaseLogger["warn"]>));
   },
-};
+} as BaseLogger;
```
Kevin-Chant: Nit: can we move this back to a `: BaseLogger` annotation? Looks like the definition satisfies that interface

PR #5498, opened 2026-03-16
feat(experiments): Add new controller to update feature values on draft experiments

msamper:
### Features and Changes

Adds a new controller `postExperimentFeatureValues` that is meant to be used for updating feature values and adding variations on draft experiments. For now this endpoint isn't used anywhere in the UI so it's fairly low risk, but the next step will be to implement designs for a new experience to allow editing feature values from a draft experiment overview page. This PR also does some refactoring and moves some service functions from a controller file to a services file.

### Testing

**Happy Paths**

- [x] Update existing variation metadata with no feature value changes. No feature revision created.
- [x] Update existing variation weights with no feature value changes. No feature revision created.

Experiment with one feature
- [x] Create a new variation with valid variation weights and add a value for the variation to the feature. Revision created.
- [x] Update feature values. Revision created.
- [x] Request auto-publish when making value changes. Revision is created and auto-published.
- [x] Request a target revision version when making value changes. Revision is updated.

Experiment with two+ features
- [x] Create a new variation with valid variation weights and add a value for the variation to the features. Revision created
- [x] Update feature values on one feature. Revision created for that feature.
- [x] Update feature values on all features. Revision created for all features.

**Error States**

- [x] Create a new variation, but weights do not add up to 1. Error.
- [x] variations and variationWeights don't have the same length. Error.
- [x] Attempt to make a request through this endpoint when an experiment is not in draft. Error.
- [x] Make a request but pass in less features than are actually linked to the experiment. Error.

Experiment with one feature
- [x] Create a new variation, but don't add a new ExperimentRefVariation in your request. Error.
- [x] Create a new variation, add a new ExperimentRefVariation to the feature, but use the wrong id. Error.

Experiment with two+ features
- [x] Create a new variation, but don't add a new ExperimentRefVariation to every feature in your request. Error.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1006,7 +1006,7 @@ export async function updateExperimentAndSync({
     bypassWebhooks,
   });
 
-  const shouldSyncLinked = changes.variations !== undefined;
+  const shouldSyncLinked = changes.variations && updated;
 
   if (shouldSyncLinked) {
     await syncVisualChangesetsAndUrlRedirectsForExperiment({


packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3962,314 @@ export async function getExperimentTimeSeries(
...
+      if (!mergeResult.success) {
+        throw new Error(
+          "Unable to auto-publish feature values. Please resolve conflicts before publishing.",
+        );
+      }
+
+      const changedEnvs = Object.keys(mergeResult.result.rules || {});
+      if (changedEnvs.length > 0) {
+        if (!context.permissions.canPublishFeature(feature, changedEnvs)) {
+          context.permissions.throwPermissionError();
```
msamper: It might make more sense for this to just skip auto-publishing if the user doesn't have permission to publish the feature instead of throwing a permission error that will make it seem like something went wrong

packages/back-end/src/app.ts
```diff
@@ -651,6 +651,10 @@ app.post(
   "/experiment/:id/targeting",
   experimentsController.postExperimentTargeting,
 );
+app.post(
+  "/experiment/:id/features",
+  experimentsController.postExperimentFeatureValues,
```
ahdriel, line 665: nit: should it just be `postExperimentFeatures`?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
     timeSeries,
   });
 }
+
+export async function postExperimentFeatureValues(
+  req: AuthRequest<
+    {
+      variations: Variation[];
+      variationWeights: number[];
+      features: Record<string, ExperimentRefVariation[]>;
```
ahdriel: question: Do we receive just the latest revisions from the feature that still has connection to the experiment?

Or would we also receive a feature in case any revision in the past had the Experiment-Ref rule?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+      message: "Experiment not found",
+    });
+    return;
+  }
+
+  if (experiment.status !== "draft") {
+    res.status(400).json({
+      status: 400,
+      message:
+        "Editing feature values from an experiment is only allowed for experiments in draft. To edit feature values for a running experiment, edit the values from the feature flags directly.",
```
ahdriel, line 3722: question: Is this a temporary restriction or how we want to move forward?

Or maybe we are still not sure?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+  if (variations.length !== variationWeights.length) {
+    res.status(400).json({
+      status: 400,
+      message: "variations and variationWeights must be the same length.",
+    });
+    return;
+  }
+
+  validateVariationIds(variations);
+  const variationWeightSum = variationWeights.reduce(
```
ahdriel: question: are we concerned about potential floating point inconsistencies here?

In the front-end we have this code, but ideally we can use a different technique to ensure these calculations are properly done.

```
// Make sure variation weights add up to 1 (allow for a little bit of rounding error)
const totalWeights = variationWeights.reduce(
(total: number, weight: number) => total + weight,
0,
);
const isValid = totalWeights > 0.99 && totalWeights < 1.01;
  ```

File: packages/back-end/src/controllers/experiments.ts [hunk 6]
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+  );
+  if (variationWeightSum !== 1) {
+    res.status(400).json({
+      status: 400,
+      message: "variationWeights must add up to 1.",
+    });
+    return;
+  }
+
+  if (Object.values(features).some((v) => v.length !== variations.length)) {
```
ahdriel: question: Should we also validate that `variationId` matches and not just the length?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+    res.status(400).json({
+      status: 400,
+      message: "All features must specify values for all variations.",
+    });
+    return;
+  }
+
+  if (!experiment.phases?.length) {
```
ahdriel, line 3727: nit: Move up as this is a cheap check and also lets all `experiment` validations be close to each other.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+  const variationsChanged =
+    JSON.stringify(variations) !== JSON.stringify(experiment.variations);
+
+  const changes: Changeset = {};
+
+  if (variationsChanged) {
+    changes.variations = variations;
+  }
+
+  // Mirror `postExperiment`: variationWeights are updated on the last phase only.
```
ahdriel: question: should we extract this into a helper or something?

Not specifically this logic, but the `apply variation weights changes to an experiment` logic.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+
+  const linkedFeatures = await getFeaturesByIds(context, linkedFeatureIds);
+
+  const envs = getAffectedEnvsForExperiment({
+    experiment,
+    orgEnvironments: context.org.settings?.environments || [],
+    linkedFeatures,
+  });
+
+  if (!context.permissions.canRunExperiment(experiment, envs)) {
```
ahdriel: question: why are we checking this? Should we allow modifications via `canUpdateExperiment` that would update the Feature draft as well -- even if they cannot click the `Start` button?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+    context.permissions.throwPermissionError();
+  }
+
+  if (linkedFeatures.length !== Object.keys(features).length) {
```
ahdriel: nit: Move under where we fetch `linkedFeatures`?

packages/back-end/src/controllers/experiments.ts
ahdriel: Can we do this validation early? So that this becomes hard to happen?
Because otherwise we will have some records updated and some not -- leaving us in an inconsistent state.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+
+  if (!context.permissions.canUpdateExperiment(experiment, changes)) {
+    context.permissions.throwPermissionError();
+  }
+
+  const linkedFeatureIds = experiment.linkedFeatures || [];
+
+  const linkedFeatures = await getFeaturesByIds(context, linkedFeatureIds);
+
+  const envs = getAffectedEnvsForExperiment({
```
ahdriel, line 3764: question: Are we allowing users to specify the values per-env for the Feature? if not, how are we handling updates for features that have multiple environments (applies to all?) ?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+
+    const visualChangesets = await findVisualChangesetsByExperiment(
+      experiment.id,
+      org.id,
+    );
+
+    if (visualChangesets.length) {
+      await Promise.all(
+        visualChangesets.map((vc) =>
+          syncVisualChangesWithVariations({
```
ahdriel: suggestion: should we have a higher-level `updateExperiment` that handles this + url redirects?

Could be a `updateExperiment` and `updateExperimentModel` or something

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+    });
+
+    if (!featureNeedsUpdate) {
+      continue;
+    }
+
+    const changedEnvironments = matchingRules.map(
+      (m: MatchingRule) => m.environmentId,
+    );
+    const resetReview = resetReviewOnChange({
```
ahdriel: question: Should all of this logic should be somewhere else? maybe we need a new `experiment-feature` layer that would be aware of `review requirements` etc, so we don't need to always be accounting for that.

So the controller becomes thin and responsible for gathering user input, formatting it, and passing it to the underlying function.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+
+    const requiresReview = checkIfRevisionNeedsReview({
+      feature,
+      baseRevision: base,
+      revision: updatedRevision,
+      allEnvironments: orgEnvIds,
+      settings: org.settings,
+    });
+
+    if (!requiresReview) {
```
ahdriel: nit: if it does require review, should we return something to the `UI` saying that it is not published yet?

packages/back-end/src/controllers/features.ts
```diff
@@ -641,6 +641,48 @@ export async function postFeatures(
   });
 }
 
+export async function getLiveAndBaseRevisionsForFeature({
```
ahdriel: nit: Can we move this function somewhere else as this is not a controller endpoint?

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -959,6 +959,48 @@ export async function editFeatureRule(
   );
 }
 
+export async function editFeatureRules(
```
ahdriel, line 954: suggestion: should we make `editFeatureRule` call `editFeatureRules` with an array?

So the single method becomes just a helper to wrap it in an array -- or we can also do the same in a single function and accept either array or single value and handle it inside.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+    if (
+      !context.permissions.canUpdateFeature(feature, {}) ||
+      !context.permissions.canManageFeatureDrafts(feature)
+    ) {
+      context.permissions.throwPermissionError();
+    }
+  }
+
+  // If variations have changed, update the experiment and sync visual changesets and url redirects
+  if (changes.variations) {
```
ahdriel: issue: If only `variationWeights` change, are we calling `updateExperiment` or where is getting updated?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+        ),
+      );
+    }
+  }
+
+  // Go through features and create revisions
+  for (const feature of linkedFeatures) {
+    const orgEnvIds = context.environments;
+    // Get existing experiment-ref rules for this experiment across all environments
+    const matchingRules = getMatchingRules(
```
ahdriel: question: `feature` is the live version here? or is it the draft?

Just to make sure that we are looking/comparing the right versions and not getting rules from live but the draft configuration is different, creating a mismatch.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3950,3 +3964,305 @@ export async function getExperimentTimeSeries(
...
+          revision: revision.version,
+          comment: "auto-publish experiment variation values change",
+        }),
+      });
+    }
+  }
+
+  res.status(200).json({
+    status: 200,
+    experiment,
```
ahdriel: issue: This is the stale experiment, not the modified one.

packages/back-end/src/controllers/experiments.ts
msamper: I think we're still figuring it out. We want to potentially enable editing feature values for running experiments in the future but it would most likely be through the make changes flow

packages/back-end/src/controllers/experiments.ts
msamper: Yeah I think that's a great idea. I noticed we're lacking validation with the sum of the variation weights in the back-end so maybe this would be a good helper to do that. Right now we're just trusting that the `variationWeights` are valid when we write them to the latest phase in `postExperiment` so it'd be good to fix that

packages/back-end/src/controllers/experiments.ts
msamper: I added this check because `variations` is listed as one of the keys that requires run experiment permission in the `postExperiment` controller and it's modifiable via this API
https://github.com/growthbook/growthbook/blob/msamper/exp-feature-value-edit/packages/back-end/src/controllers/experiments.ts#L1744-L1763

I think what is wrong here is that we should only be checking this permission if `variationsChanged`

packages/back-end/src/controllers/experiments.ts
msamper: No, I don't think we are. If I'm understanding your question correctly, we're determining the environments for the feature rule variation value updates based on the `matchingRules` we find in this [code block](https://github.com/growthbook/growthbook/blob/msamper/exp-feature-value-edit/packages/back-end/src/controllers/experiments.ts#L4138-L4143). This still isn't handling for the situation I brought up on [slack](https://growthbookapp.slack.com/archives/C07NK4F016Z/p1773948417757109) where there's and existing draft(s) that are modifying experiment-ref rule values since I still don't have a good solution for that case.

packages/back-end/src/controllers/experiments.ts
msamper: Yeah good point! That'd be a good thing to let them know so they're not confused

packages/back-end/src/controllers/experiments.ts
msamper: Good catch! I think this might be a bug right now because `updateExperiment` is only triggered by changes to `variations`. Will fix

packages/back-end/src/controllers/experiments.ts
msamper: `feature` here is the live feature. `getMatchingRules` can take in a revision and that will supersede the live feature for finding the matching rules, but I still need to figure out what the right way is to identify an existing draft revision that is also modifying the `experiment-ref` values. Right now this logic is always creating a new draft revision when there are changes to the live features `experiment-ref` variation values in the request

packages/back-end/src/controllers/experiments.ts
msamper: Will fix!

packages/back-end/src/controllers/experiments.ts
msamper: We would just receive features that are currently connected to the experiment. i.e. if a feature previously had an `experiment-ref` rule for this experiment and no longer does, we shouldn't allow that feature to be in `features`. I can add some validation for this

packages/back-end/src/controllers/experiments.ts
msamper: I think we should be. This is an existing problem with updating experiments because we currently do no variation weight validation on the back-end but we could solve with a variation update helper

packages/back-end/src/app.ts
msamper: It could. Only reason I did `postExperimentFeatureValues` is because `postExperimentFeatures` makes it sounds like it's a controller where you can add new features

packages/back-end/src/controllers/experiments.ts
msamper: not needed anymore since the client specifies whether or not we're auto publishing. If the permissions don't check out for some reason during validation we'll throw

packages/shared/src/experiments/experiments.ts
```diff
@@ -1483,6 +1483,24 @@ export function generateSliceStringFromLevels(
   return generateSliceString(slices);
 }
 
+/**
+ * Allowed drift when checking that variation weights sum to 1.
+ * Matches sdk-js and NewPhaseForm (0.99 … 1.01 inclusive).
+ */
+const VARIATION_WEIGHTS_SUM_TOLERANCE = 0.01;
```
lukesonnet, line 1490: I don't know why this is.... feels like it should be 0.0001 to confirm it doesn't matter for delivery (because 0.3333 and 0.3334 are different), but I don't know what kinds of issues that might raise for summing like 20 variations.

But maybe nothing to do here.

packages/back-end/src/app.ts
lukesonnet: double nit: Should the path then also be `:id/featureValues`

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3709,3 +3688,207 @@ export async function getExperimentTimeSeries(
...
   });
 }
+
+export async function postExperimentFeatureValues(
+  req: AuthRequest<
+    {
+      variations: Variation[];
+      variationWeights: number[];
+      features: Record<string, ExperimentRefVariation[]>;
+      featureRevisionOptions: Record<
```
lukesonnet: question: is there a reason these two records are separate rather than bundled together?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3709,3 +3688,207 @@ export async function getExperimentTimeSeries(
     timeSeries,
   });
 }
+
+export async function postExperimentFeatureValues(
+  req: AuthRequest<
+    {
+      variations: Variation[];
+      variationWeights: number[];
```
lukesonnet, line 3697: question: should we bundle these together so that this controller can never take mismatched variations and variation weights?

I guess maybe it's better to do the check in `validateExperimentFeatureVariations`, but feels like the controller can just be designed to not accept invalid data. Maybe it's bad though if it requires a ton of work by whoever is packaging data for the controller.

packages/back-end/src/controllers/experiments.ts
lukesonnet: If checking `linkedFeatures` is meant to accomplish this, I'm not sure it does, because I don't think we remove a feature from `linkedFeatures` if you just delete the rule on the feature side, or if the rule is only in an old revision. We may want to check that in the future.

For now, I'm somewhat less concerned, because if someone wants to re-link an old feature flag via this endpoint, it's not the desired action, but it seems... ok.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3709,3 +3688,207 @@ export async function getExperimentTimeSeries(
...
+      variationWeights,
+    );
+  }
+
+  if (!context.permissions.canUpdateExperiment(experiment, changes)) {
+    context.permissions.throwPermissionError();
+  }
+
+  // Check for permission to update each feature
+  for (const feature of linkedFeatures) {
```
lukesonnet: Why are we checking for all linked features and not just the subset in this controller?

packages/back-end/src/services/experiment-feature.ts
```diff
@@ -0,0 +1,217 @@
...
+    const matchingRules = getMatchingRules(
+      feature,
+      (r: FeatureRule) =>
+        r.type === "experiment-ref" && r.experimentId === experiment.id,
+      context.environments,
+      revision,
+    );
+
+    // TODO: Should we allow this to happen and continue?
+    // A linked feature may no longer have any experiment-ref rules, which would cause the update to fail.
```
lukesonnet: Yeah, again I'm not sure why we go with `linkedFeatures` instead of the features in the actual payload being changed as the core set we loop over.

packages/back-end/src/services/experiment-feature.ts
```diff
@@ -0,0 +1,217 @@
...
+    const { targetVersion, autoPublish, forceNewDraft } =
+      featureRevisionOptions[feature.id];
+
+    const useExistingRevision = targetVersion && !forceNewDraft && !autoPublish;
+
+    const revision = useExistingRevision
+      ? await getDraftRevision(context, feature, targetVersion)
+      : await getLiveRevisionForFeature(context, feature);
+
+    const matchingRules = getMatchingRules(
```
lukesonnet, line 198: question: where do we ever do the validation to block edits if the feature has multiple ref rules with different values? The robot seems to think this PR doesn't do this at all, and the only related code is for the front-end now where we show that banner if there are `inconsistentValues`. Do we just want to rely on the front-end for this in the future, or should the `inconsistentValues` check also occur here on validation?

packages/back-end/src/services/experiments.ts
```diff
@@ -937,6 +941,84 @@ export function resetExperimentBanditSettings({
...
+}): Promise<ExperimentInterface> {
+  const updated = await updateExperiment({
+    context,
+    experiment,
+    changes,
+    bypassWebhooks,
+  });
+
+  const shouldSyncLinked =
+    changes.variations !== undefined || changes.phases !== undefined;
```
lukesonnet: Why are phases now leading to syncing? Previously, AFAICT, it was only checking if variations changed in `postExperiment`?

lukesonnet: This might change in my PR, but is it right here? I guess it's harmless.

packages/back-end/src/models/FeatureRevisionModel.ts
```diff
@@ -646,6 +646,7 @@ export async function updateRevision(
     {
       $set: { ...changes, status, dateUpdated: new Date() },
     },
+    { new: true },
```
lukesonnet, line 652: Is this a good idea given the name of this method? Probably fine, just wanted to ask.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -935,33 +936,55 @@ export async function editFeatureRule(
...
+
+export async function editFeatureRules(
+  context: ReqContext | ApiReqContext,
+  feature: FeatureInterface,
+  revision: FeatureRevisionInterface,
+  matches: { environmentId: string; i: number }[],
+  updates: Partial<FeatureRule>,
+  user: EventUser,
+  resetReview: boolean,
+) {
```
lukesonnet, line 962: Would it be a good idea to add unit tests to this method?

packages/back-end/src/models/FeatureRevisionModel.ts
msamper: Returning the new document after the update? or are you talking about replacing `updateOne` with `findOneAndUpdate`?

packages/back-end/src/controllers/experiments.ts
msamper: Yeah I was trying to make it so that the client could just pass in the data as it's represented on the experiment rather than transforming it into a shape just used for the API. I think the only thing we'd gain on the validation side is not having to check that the length of both arrays match.

packages/shared/src/experiments/experiments.ts
msamper: I believe I was just matching the tolerance we have in the SDK here, but I can double check

packages/back-end/src/controllers/experiments.ts
msamper: Yeah you're right. I think when I originally wrote this portion of the code I thought `linkedFeatures` was updated when experiment rules were removed, but that's not the case. I think your suggestion in another comment to just use the features in the payload is correct.

packages/back-end/src/controllers/experiments.ts
msamper: I think I separated them because I wanted to keep the value updates separate from the revision selection, but I think bundling together makes more sense. Don't really have a strong opinion about it so I'm happy to change

packages/back-end/src/controllers/experiments.ts
msamper: See my response to your comment above about `linkedFeatures`

packages/back-end/src/models/FeatureModel.ts
msamper: I added a unit test for the core logic that was moved to `applyPartialFeatureRuleUpdatesToRevision`. I only have two test cases though so I may add some more

packages/back-end/src/services/experiment-feature.ts
msamper: We don't have that validation here right now, but I think we probably should to ensure we block in the case. I can add.

packages/back-end/src/services/experiments.ts
msamper: Good catch! I think this was unintentional. The only phase changes we'd be making from this new endpoint are weight updates and even in that case I don't see why that should require syncing. I'll remove

packages/back-end/src/controllers/experiments.ts
lukesonnet: Sure, that doesn't seem worth it.

packages/back-end/src/models/FeatureRevisionModel.ts
lukesonnet: I misunderstood this method, never mind!

packages/shared/src/experiments/experiments.ts
lukesonnet: yeah that's fine. I just don't really understand the motivation for that specific tolerance. don't worry about it for now unless something jumps out at you.

packages/back-end/src/app.ts
msamper: punting on this for now. we can change later if we need to. There's just not a great name I can think of for this since it also allows adding variations

PR #5499, opened 2026-03-16
Variations in Phases 1b: use variation index in graphs

lukesonnet:
### Features and Changes

Recently we ensured all variations returned from an experiment have the index from the main `experiment.variations` object using helpers like `getLatestPhaseVariations`. This means we can sometimes delete the 1st variation and only have the 0th and 2nd. To compensate for this, we need to build the index and pass it to our graph components instead of relying on an array of variation names alone.

This PR achieves that with the main components.

This shouldn't affect any users, as none of them have "disabled" variations yet.


msamper (approved):
Code lgtm and tested locally to make sure the change was no-op

PR #5504, opened 2026-03-17
feat(api): Add templateId to POST /experiments endpoint

dannylin-ant:
## Summary

Adds support for the `templateId` field when creating experiments via the REST API (`POST /api/v1/experiments`). Previously this was only possible through the UI, meaning experiments created programmatically would never appear in the template "Usage" count on the templates page. This enables support for applying template values as defaults to the experiment if a certain property is not passed in the payload but is available on the template.

## Changes

- Add optional `templateId` string field to `PostExperimentPayload.yaml`
- Validate the template exists before creating the experiment (returns `Invalid template: <id>` if not found or inaccessible)
- Wire `templateId` through `postExperimentApiPayloadToInterface()` so it's persisted on the experiment document
- Regenerate OpenAPI types and Zod validators
- Adds `templateToPostExperimentDefaults` to convert the experiment template to the shape of the `PostExperiment` request body to apply defaults when a property is not included in the payload

## Behavior

- If a `templateId` is provided, the API will use template defaults when there is a value on the template that was not passed in the `PostExperiment` request body
- Empty string normalizes to `undefined` (not stored).
- Templates the caller can't read (cross-project) are treated as not found.
- If a `templateId` is provided along with `assignmentQueryId` we throw an error. Users should not specify an `assignmentQueryId` on the experiment when using a template

## Test Plan

- `pnpm --filter back-end type-check` ✅
- `pnpm --filter shared type-check` ✅
- ESLint on changed files ✅
- Manual: Create experiment via API with `templateId`, verify it appears in template usage count on the Templates page

### DIFF
--- a/docs/src/partials/event-webhook/_event-webhook-list.md
+++ b/docs/src/partials/event-webhook/_event-webhook-list.md
@@ -1454,6 +1454,7 @@ Triggered when an experiment is created
                     levels: string[];
                 }[];
             }[] | undefined;
+            templateId?: string | undefined;
         };
     };
     user: {
@@ -1630,6 +1631,7 @@ Triggered when an experiment is updated
                     levels: string[];
                 }[];
             }[] | undefined;
+            templateId?: string | undefined;
         };
         previous_attributes: {
             id?: string | undefined;
@@ -1772,6 +1774,7 @@ Triggered when an experiment is updated
                     levels: string[];
                 }[];
             }[] | undefined;
+            templateId?: string | undefined;
         };
         changes?: {
             added: Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


packages/back-end/src/api/openapi/payload-schemas/PostExperimentPayload.yaml
```diff
@@ -3,14 +3,13 @@ required:
...
-  - assignmentQueryId
 properties:
   datasourceId:
     type: string
-    description: ID for the [DataSource](#tag/DataSource_model)
+    description: ID for the [DataSource](#tag/DataSource_model). Can only be set if a templateId is not provided.
   assignmentQueryId:
     type: string
-    description: The ID property of one of the assignment query objects associated with the datasource
+    description: The ID property of one of the assignment query objects associated with the datasource. Can only be set if a templateId is not provided.
```
Kevin-Chant: Also update the comment to match

PR #5506, opened 2026-03-18
BaseModel type system improvements

Kevin-Chant:
### Features and Changes

We noticed a bug from migrating TeamModel that BaseModel wasn't strict enough about UpdateProps so we had a runtime error without a ts warning. Separately, one of the API models had a new query param added incorrectly, and this also didn't prompt any errors.

The fix for the first issue was to make the UpdateProps type reject the default fields (org, dates, uid) and any pKey fields (id or custom per model).

The stricter type annotation needed a few classes of changes:
- Change usage of Partial<T> to UpdateProps<T>
- Stop typing canUpdate's param as the full interface
- Change typing of backend endpoint `req.body`s

For the API issue, we can tighten the type inference by tying the handler method signatures to the zod validators so they directly control the type of `req` in the handler.

### Testing

- [x] `pnpm type-check` (and ci) pass
- [x] Trying to set id or organization in an update has a ts warning (e.g. the `...team` spread from #5457)
- [x] Adding a new param to `apiDeleteCustomFieldValidator` causes it to be available in `CustomFieldModel#handleApiDelete` immediately


Kevin-Chant:
Okay @tzjames this should be ready for real this time. The merge issues were due to a real bug that our previous types didn't catch, so I also made the API handler typing more concrete to prevent that in the future

tzjames (approved):
LGTM

PR #5509, opened 2026-03-18
Update to aws-sdk v3

### LINKED ISSUE #2655: [Feature] Support custom S3 endpoint for upload storage
### Description of Feature

We use a 3rd party service which offers an S3-compatible API for object storage. We'd like to use this service for storing Growthbook uploads, but unfortunately GB currently only appears to support using official AWS endpoints without code changes. This should be configurable via environment variable.

Possible options would be to either support a custom S3 endpoint directly via a Growthbook-specific config (and then pass it as the `endpoint` option when constructing the S3 client) or to update to v3 of the AWS SDK for JS, which natively supports [service-specific endpoints](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html), including from environment variables.

It might be wise to update to v3 of the SDK anyway as [v2 is reaching end-of-support in 2025](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/).

I would be willing to contribute the required changes if a course of action is decided on.

tzjames:
### Features and Changes
Upgrades to version 3 of the sdk.
It installs the individual modules, so that the overall package size is 5% less or so.

Closes https://github.com/growthbook/growthbook/issues/2655

### Testing
With s3 credentials upload an image, and see it succeed.
Testing cloudwatch metrics is a bit harder, and will just checkout the datadog dashboard afterwards to make sure those metrics are still being posted.


packages/back-end/generated/.api-spec-hash
```diff
@@ -1 +1 @@
-4d7ceb494091756bb276ab1a7365713769aebfba
+7a6e5d4bba653ec6eb6a3129ca4434992179bf0c
```
tzjames: somehow this is wrong on main

tzjames: ah because we committed a change from cursor to the precommit hook after it was calculated.

packages/back-end/src/services/files.ts
```diff
@@ -12,63 +19,26 @@ import {
...
-
-      // Eagerly load credentials for the first time so that masterCredentials
-      // is populated before any synchronous SDK operations (e.g. getSignedUrl,
-      // createPresignedPost) attempt to access it.
-      await refreshCredentials();
-
-      s3 = new AWS.S3({
-        signatureVersion: "v4",
-        credentials: awsTempCredentials,
+      clientConfig.credentials = fromTemporaryCredentials({
```
Kevin-Chant, line 31: I would assume that leaning on the AWS libraries more heavily might handle it, but we need to make sure this doesn't re-introduce the refresh issues from #5312 and #5459.

I wasn't able to repro the issues locally because they only happen in a context where there's no `.credentials` file or env vars to seed credentials from (but a role can still be inferred, e.g. EC2/ECS)

Kevin-Chant (approved):
Approving now to unblock, but would be good to hear back about whether they can test the credentials refresh issue

PR #5510, opened 2026-03-18
Variations in Phases 2: Migrating variation active level to phases

lukesonnet:
### Features and Changes

Begin relying on the phase as the source of truth for which variations are active, and use the experiment variations solely to store metadata.

This PR adds `variations` to the `phases` object, adds a JIT migration, and ensures endpoints updating variaitons also update this object when it is correct to do so.

Importantly, we have to make some decisions about how to update `phases` when `variations` change, either from internal or external endpoints. I would say the core logic is:
* The new data model for Phases
* The new "phase accessor" method in `getLatestPhaseVariations` (https://github.com/growthbook/growthbook/pull/5510/changes#diff-c91c750717bf7b5dd875ac0714f1b86492d7e0c38975a38a489be3fb593ead06R28)
* The new migration in `upgradeExperimentDoc` (https://github.com/growthbook/growthbook/pull/5510/changes#diff-f223efbaf711ac7164ccdcb6ad916bc6370e6df49c69447d110bc4ee55207b42R606-R617)
* The new logic in the endpoints `postExperiment`, `postExperiments`, and the REST API endpoints `postExperimentApiPayloadToInterface` and `updateExperimentApiPayloadToInterface`.

In a follow-up PR, I will move `variationWeights` into the `phase.variations` object as well.

In a second follow-up PR, we will be able to remove a variation and mark it as "pass-through" or "fully delete" which will force new phase + re-randomize.

### Testing

- [x] Create new experiment (from scratch)
- [x] Edit experiment targeting on existing running experiment (manual + make changes; variation weights, order on manual mode)
- [x] Edit experiment targeting on existing draft experiment
- [x] Start a new phase from an existing experiment (manual + make changes)
- [x] Change experiment status (running/stopped)
- [X] Duplicate experiment
- [X] Create experiment from Feature Rule modal
- [X] Add existing experiment from feature rule modal
- [X] Import experiments
- [X] Compare Experiment Events modal
- [X] REST API creation
- [X] REST API updating
- [X] Configure Report using experiment metadata
- [x] Generated demo experiment
- [ ] Test payload properly excludes deleted variations (manually delete variation in the phase in the DB).
- [X] Open old experiments created pre-branch and make changes there
- [X] Bandit creation
- [X] Update bandit variations (create, re-order, etc.)
- [X] Bandit weight update


packages/shared/src/experiments/variations.ts
```diff
@@ -1,51 +1,64 @@
...
 
-  // TODO change to come from phase
-  const phaseVariations = experiment.variations;
+  const latestPhase = experiment.phases[experiment.phases.length - 1];
+  const phaseVariations = latestPhase.variations;
 
   let hasMissing = false;
-  const result: VariationWithIndex[] = phaseVariations.map((v, i) => {
+  const foundVariations: VariationWithIndexAndStatus[] = [];
+  phaseVariations.forEach((v) => {
```
lukesonnet: I want the order of `phaseVariations` to be the canonical order of the variations, but that whole thing is a bit messy right now as I have to re-order the phase variations when the payload comes in with re-ordered main variations.

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -94,6 +94,12 @@ const banditResultObject = {
   weightsWereUpdated: Boolean,
 };
 
+const phaseVariation = {
+  _id: false,
+  id: String,
+  status: String,
+};
```
msamper, line 101: Not to complicate things more but if we're going to have variation objects on the phase I kind of feel like we should consider moving weights into those objects as well to decouple them from the index of the variations array and have them map to an id instead

packages/back-end/src/services/experiments.ts
```diff
@@ -3170,18 +3178,17 @@ export function updateExperimentApiPayloadToInterface(
     ...(sequentialTestingTuningParameter !== undefined
       ? { sequentialTestingTuningParameter }
       : {}),
-    ...(variations
-      ? {
-          variations: variations?.map((v) => ({
-            id: generateVariationId(),
-            screenshots: [],
+    ...(() => {
```
msamper: I find this kind of hard to read inline like this. I would suggest pulling this out into a helper

packages/front-end/services/importing/statsig/transformers/experimentTransformer.ts
```diff
@@ -137,6 +137,10 @@ export function transformStatsigExperimentToGB(
       name: "Main",
       reason: results ? `Experiment ${results}` : "",
       variationWeights,
+      variations: variations.map((v) => ({
+        id: v.id,
+        status: "active" as const,
+      })),
```
msamper, line 143: Should we make this a shared helper? Not sure if we expect the logic for the mapping to get more complex

packages/shared/src/experiments/variations.ts
```diff
@@ -1,51 +1,64 @@
...
 
 type ExperimentWithVariations = {
   variations: Variation[];
 };
 
+type ExperimentWithVariationsAndPhases = {
+  variations: Variation[];
+  phases: Pick<ExperimentPhase, "variations">[];
+};
+
```
msamper: If these types are a subset of an experiment, might be better to use `Pick<>` instead so we can reference the types from the `ExperimentInterface` directly

packages/shared/src/experiments/variations.ts
```diff
@@ -1,51 +1,64 @@
-import { Variation } from "shared/types/experiment";
+import {
+  ExperimentPhase,
+  Variation,
+  VariationStatus,
+} from "shared/types/experiment";
 
 type ExperimentWithVariations = {
```
msamper: delete?

packages/shared/src/experiments/variations.ts
```diff
@@ -1,51 +1,64 @@
...
       ...foundVariation,
-      // override experiment variation metadata with phase variation metadata, if present
-      ...v,
-    };
+      // Add status from phase variation, if present
+      status: v.status,
+    });
   });
+  // If any missing, fall back to all variations with status "active"
   if (!hasMissing) {
```
msamper:
```suggestion
  if (hasMissing) {
```
Should the check be this instead based on the comment above or is the comment wrong? I'm not sure I understand in what cases a variation could be missing from the main experiment variations array so that might help me understand what the correct behavior should be here

msamper (commented):
Did a first passthrough mainly focused on the back-end. I'm curious about how difficult it would be to package `variationWeights` into the new phase variations. I'd love to see those be coupled with variation objects and not be dependent on just indices that could change and cause the weights to be out of sync

packages/back-end/src/models/ExperimentModel.ts
lukesonnet: > Did a first passthrough mainly focused on the back-end. I'm curious about how difficult it would be to package variationWeights into the new phase variations. I'd love to see those be coupled with variation objects and not be dependent on just indices that could change and cause the weights to be out of sync

Yes, I agree. It is another bigger migration and risky. I did that in the first pass, and would like to do it in yet another PR.

msamper: Gotcha. Another PR for that change makes sense

packages/shared/src/validators/experiments.ts
```diff
@@ -66,6 +66,19 @@ export const banditEvent = z
...
 
+// TODO(phase-update): allow "passThrough" e.g. forcibly skip a range
+// and send users to the next feature rule
+export const variationStatus = ["active"] as const;
+export type VariationStatus = (typeof variationStatus)[number];
+
+export const phaseVariation = z
+  .object({
+    id: z.string(),
+    status: z.enum(variationStatus),
```
ahdriel, line 77: question: the only scenarios are either all `phases` are `disabled` (or some other language) OR the last one is `active` while the rest is `disabled`.

Is my understanding correct? Or which scenarios do we want to support with `status` here?

packages/shared/src/validators/experiments.ts
```diff
@@ -66,6 +66,19 @@ export const banditEvent = z
 export type BanditResult = z.infer<typeof banditResult>;
 export type BanditEvent = z.infer<typeof banditEvent>;
 
+// TODO(phase-update): allow "passThrough" e.g. forcibly skip a range
+// and send users to the next feature rule
+export const variationStatus = ["active"] as const;
+export type VariationStatus = (typeof variationStatus)[number];
+
+export const phaseVariation = z
```
ahdriel, line 74: question: What does `phaseVariation` mean? and how does it differ from a `variation`?

packages/shared/src/experiments/variations.ts
ahdriel: I agree with @msamper that this should be `if (hasMissing)`.

I think this is the opposite of what we want and currently we have broken behavior for phases.

We should add tests for this function too.

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -1740,6 +1747,7 @@ const getExperimentChanges = (
     "phases",
   ];
 
+  // TODO: check variations logic here
```
ahdriel: question: TODO before merging or not?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1727,6 +1726,20 @@ export async function postExperiment(
     changes.phases = phases;
   }
 
+  // Re-order variations to match the order of the experiment.variations
```
ahdriel: question: is this code doing what the comment is saying?

It seems to me that we are respecting the order that comes in the POST body of the request.

packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
```diff
@@ -417,6 +417,10 @@ Treatment shows a larger 'Add to Cart' CTA, but with the same functionality.`,
           condition: "",
           namespace: { enabled: false, name: "", range: [0, 1] },
           variationWeights: [0.5, 0.5],
+          variations: [
+            { id: "v0", status: "active" as const },
```
ahdriel: nit: should we update these ids to be prefixed with `var_` as well?

packages/back-end/src/services/experiments.ts
```diff
@@ -2905,6 +2905,11 @@ export function postExperimentApiPayloadToInterface(
   organization: OrganizationInterface,
   datasource: DataSourceInterface | null,
 ): Omit<ExperimentInterface, "dateCreated" | "dateUpdated" | "id"> {
+  const variationIds = payload.variations.map(() => generateVariationId());
```
ahdriel, line 3289: question: If the payload come with `id` in the variations object, we are overriding it with a new one.

I know this is preexisting behavior and we should keep it, but it seems unexpected to me.

packages/back-end/src/services/experiments.ts
```diff
@@ -3220,10 +3227,23 @@ export function updateExperimentApiPayloadToInterface(
...
                 )?.map((_v, _i, arr) => 1 / arr.length),
+              variations: resolvedVariations
+                ? resolvedVariations.map((v) => ({
+                    id: v.id,
+                    status: "active" as const,
+                  }))
+                : experiment.variations.map((v) => ({
+                    id: v.id,
+                    status: "active" as const,
+                  })),
```
ahdriel: nit: should we simplify and do a `(resolvedVariations ?? experiment.variations).map()`?

packages/back-end/src/services/experiments.ts
```diff
@@ -3245,6 +3265,8 @@ export function updateExperimentApiPayloadToInterface(
     organization,
   });
 
+  // TODO validate variations?
```
ahdriel: before merging or in the future?

packages/shared/src/experiments/variations.ts
```diff
@@ -1,51 +1,64 @@
...
 export function getLatestPhaseVariations(
-  experiment: ExperimentWithVariations,
-): VariationWithIndex[] {
+  experiment: ExperimentWithVariationsAndPhases,
+): VariationWithIndexAndStatus[] {
   const allVariations = getAllVariations(experiment);
 
-  // TODO change to come from phase
-  const phaseVariations = experiment.variations;
+  const latestPhase = experiment.phases[experiment.phases.length - 1];
```
ahdriel: issue: I think this will crash during runtime if `experiment.phases` is empty.

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -375,6 +387,75 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+        })),
+      );
+      form.setValue(
+        "phases.0.variationWeights",
+        v.map((data) => data.weight),
+      );
+      form.setValue(
+        "phases.0.variations",
+        v.map((data) => ({
+          id: data.id || generateVariationId(),
```
ahdriel: question: do we need to have a shared `id` if `data.id` is missing?

Right now we will have mismatched ids between phaseVariation and experimentVariations when that happens.

packages/back-end/src/services/experiments.ts
```diff
@@ -3170,18 +3178,17 @@ export function updateExperimentApiPayloadToInterface(
...
-          })),
-        }
-      : {}),
-    ...(phases
-      ? {
-          phases: phases.map((p) => {
+          }))
+        : undefined;
+
+      const resolvedPhases = phases
```
ahdriel: question: Does this mean they can get out of sync? Like if we have `variations` but no `phases` they'll be updated separately.

Agreed with @msamper that IIFE is a bit hard to read here.

packages/shared/src/experiments/variations.ts
lukesonnet: Yeah it should be, sorry. Just a typo but I'm scared it didn't come up in my testing.

It shouldn't ever happen, but this is just a safety net to fall back to the old behavior. But since the variations in the phase refer to variations by ID, it's possible you mangle something, potentially via the REST API.

I'll fix the typo and re-do some of my checks.

packages/shared/src/experiments/variations.ts
lukesonnet: It gets used below?

packages/front-end/services/importing/statsig/transformers/experimentTransformer.ts
lukesonnet: I'm gonna punt on this for now, if you don't mind. Still sorting out the statuses in later PRs.

packages/shared/src/experiments/variations.ts
lukesonnet: It's tough because of the date issues in `ExperimentPhase`, so I'm going to leave this as is unless you have a suggestion for how to solve the issue with the dates that is more elegant than this (I tried a bit :/).

packages/shared/src/experiments/variations.ts
msamper: Ignore I completely missed the helper at the bottom of the file

packages/shared/src/experiments/variations.ts
lukesonnet: Yes, fixed in local, thanks.

packages/back-end/src/controllers/experiments.ts
lukesonnet: It is re-ordering the phase variations to match the order it comes in via the `postExperiment` endpoint, not experiment.variations, thanks. I'll correct the comment.

packages/back-end/src/routers/demo-datasource-project/demo-datasource-project.controller.ts
lukesonnet: Updated to `var_0` and `var_1`

packages/back-end/src/models/ExperimentModel.ts
lukesonnet: I did this now by checking that we now are just using variations from the latest phase, not the experiment and tested it in my dev instance. So this was before merging, and is done.

packages/shared/src/validators/experiments.ts
lukesonnet: It's the minimal set of information to know which variations are in what state for a given phase. Eventually I will also move weights in here, so that `experiment.variations` is mostly variation metadata and `phase.variations` is the source of truth for what variations are active and being used in each phase.

packages/shared/src/validators/experiments.ts
lukesonnet: I don't think I follow.

In the next PR that adds functionality, status will be one of `["active", "passthrough"]`, where `passthrough` (or something similar) means it is still getting people assigned to it and takes up variation weight to preserve the user splits across variations, but users in `passthrough` variations skip the experiment rule.

You can also "remove" a variation where it will be absent entirely from `phases.variation`, and therefore the SDK payload, but still has some metadata at the experiment level.

packages/front-end/components/Experiment/NewExperimentForm.tsx
lukesonnet: Yes, I guess it wasn't missing in my test paths, but updated. Thanks!

packages/back-end/src/services/experiments.ts
lukesonnet: Yes, ty.

packages/back-end/src/services/experiments.ts
lukesonnet: Yeah, no idea why, just preserving existing behavior and don't feel like I want to tackle this here in case I unexpectedly break an existing flow.

packages/back-end/src/services/experiments.ts
lukesonnet: Yes, this is already possible where you can update `variationWeights` with no concern for the actual length of the `experiment.variations` object, and vice-versa.

I can try and partially resolve this before merging if we think it's too problematic. See this comment thread: https://github.com/growthbook/growthbook/pull/5510#discussion_r3011560919

There is still the chance for some things to get out of sync, but it's less likely now. In general, updating variations or variation weights via this endpoint is pretty broken.

packages/back-end/src/services/experiments.ts
lukesonnet: See thread in Slack discussing the new logic, but basically now we have the following behavior in this update endpoint:

I added logic to that basically doesn't expose `phase.variations` to users at all in this endpoint, but if they set experiment.variations, I take that as gospel and override the latest phase variations with those settings. Also, to be mostly backwards compatible, if you pass new `phases` objects, I also take the `variations` from the experiment level unless the the phase in that index position already had some variation metadata.

You can see unit tests in the api/experiments.test.ts file to see some more example behavior.

packages/back-end/src/services/experiments.ts
lukesonnet: Pulled it out in-line, and the logic here changed a ton anyways!

packages/back-end/src/services/experiments.ts
msamper: I wonder why we even allow them to pass in a variation ID in the first place

packages/back-end/src/services/experiments.ts
```diff
@@ -3245,6 +3326,8 @@ export function updateExperimentApiPayloadToInterface(
     organization,
   });
 
+  // Validate that phase variations are the same as the variations coming in via the request body
```
msamper: Does this validation need to be added?

packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -353,6 +353,10 @@ const AnalysisForm: FC<{
             body.phases[phaseId].variationWeights = body.variations.map(
               () => 1 / (body?.variations?.length || 2),
             );
+            body.phases[phaseId].variations = body.variations.map((v) => ({
+              id: v.id || "",
+              status: "active" as const,
+            }));
```
msamper: Should we do this for all experiments instead of just bandits?

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -375,6 +387,81 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+    (v, i) => ({
+      id: v.id || "",
+      name: v.name,
+      key: v.key || `${i}`,
+      description: v.description,
+      screenshots: v.screenshots,
+      weight: watchedWeights[i] ?? 1 / (watchedExpVariations.length || 2),
+      status: "active" as const,
+    }),
+  );
```
msamper, line 402: Why do we need this in this in this shape? (`FormVariation`) From what I can  tell `weight` and `status` are stripped out for the child components so I'm not sure I understand what they're being added for.

packages/front-end/components/Experiment/NewExperimentForm.tsx
```diff
@@ -375,6 +387,81 @@ const NewExperimentForm: FC<NewExperimentFormProps> = ({
...
+
+  const setVariationWeight = useCallback(
+    (i: number, weight: number) => {
+      form.setValue(`phases.0.variationWeights.${i}`, weight);
+    },
+    [form],
+  );
+
+  // Props for FeatureVariationsInput derived from the combined variations
```
msamper, line 453: nit: also used for `ExperimentRefNewFields` and `BanditRefNewFields`

packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -503,29 +503,39 @@ export default function RuleModal({
...
-            },
-          ],
+          ...(() => {
+            const variations = values.values.map((v, i) => ({
+              id: uniqId("var_"),
+              key: i + "",
+              name: v.name || (i ? `Variation ${i}` : "Control"),
+              screenshots: [],
+            }));
+            const variationWeights = values.values.map((v) => v.weight);
```
msamper: nit: could we generate the `variations` and the `variationWeights` above this block where we create the experiment object? an inline function like this being spread is a bit hard to read

packages/front-end/components/Features/FeatureVariationsInput.tsx
```diff
@@ -95,16 +95,20 @@ export default function FeatureVariationsInput({
   const [editingIds, setEditingIds] = useState(
     startEditingIndexes || !idsMatchIndexes,
   );
+  console.log("editingIds", editingIds);
   const [numberOfVariations, setNumberOfVariations] = useState(
     Math.max(variations?.length ?? 2, 2) + "",
   );
+  const renormalizeVariationKeysOnSort =
```
msamper: I'm a bit confused, but not sure I fully understand what this is for. If `!valueAsId` and two other conditions are true then we set `valuesAsIds` for `SortableVariationsList` to `true`? Is that expected?

packages/shared/src/experiments/variations.ts
```diff
@@ -1,51 +1,67 @@
...
+  const foundVariations: VariationWithIndexAndStatus[] = [];
+  latestPhase.variations.forEach((v) => {
     const foundVariation = allVariations.find((allV) => allV.id === v.id);
     if (foundVariation === undefined) {
       hasMissing = true;
-      return {
-        ...v,
-        index: i,
-      };
+      return;
```
msamper: This will just move on to the next iteration of the `forEach` loop. I think you want to exit here since you return `defaultResponse` if `hasMissing`. Would recommend using a for of loop or a traditional for loop

packages/front-end/components/Features/FeatureVariationsInput.tsx
lukesonnet: Yeah, this file is crazy. That `valuesAsIds` on `SortableVariationsList` governs some other behavior, but I should create a separate prop, I'll do that.

lukesonnet: Done, I tried to make it a bit clearer, but basically, if `valuesAsId` is true, we only set the values for a feature flag (used on experiment creation from a feature flag) and the IDs are in the background, so we want to re-normalize IDs in that case in case you drag stuff around.

But there are other cases, e.g. when editing variations directly in an experiment, where if they don't have custom IDs (and you're not actively editing them), we need to renormalize those keys after sorting and deleting, because otherwise we have errors where we try to set a new key for a new variation that clashes with an existing key.

A different solution in this case could be to just pick the next available key for that variation, but I'd rather avoid that as that would be, I think, a new behavior.

packages/back-end/src/services/experiments.ts
lukesonnet: We actually don't do this anymore, we're doing that new logic in `resolveExperimentUpdateVariationsAndPhases`. I'll add a comment there about the intended behavior.

packages/shared/src/experiments/variations.ts
lukesonnet: Done with a for loop and `break`, thanks.

packages/front-end/services/features.ts
```diff
@@ -1375,30 +1375,36 @@ export function getExperimentDefinitionFromFeature(
...
+    let name = i ? `Variation ${i}` : "Control";
+    if (v?.name) {
+      name = v.name;
+    } else if (feature.valueType === "boolean") {
+      name = v.value === "true" ? "On" : "Off";
+    }
+    return {
+      id: generateVariationId(),
+      name,
+      key: v.value,
```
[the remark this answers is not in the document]
lukesonnet: weird copy past error I guess! fixed.

packages/front-end/components/Experiment/NewExperimentForm.tsx
lukesonnet: And it's passed through from there to `FeatureVariationsInput` as well.

packages/front-end/components/Experiment/NewExperimentForm.tsx
lukesonnet: Status is stripped, but not weight? We need both. I'll remove status for now, but we'll likely add it back in in a later PR if we update status in the same modal.

packages/front-end/components/Experiment/AnalysisForm.tsx
lukesonnet: It's actually redundant here since we do it in the back end. So I'll remove it from here.

packages/back-end/src/services/experiments.ts
```diff
@@ -3064,6 +3071,128 @@ export function postExperimentApiPayloadToInterface(
...
+  const hasPhasePayload = phases !== undefined;
+  const hasVariationPayload = variations !== undefined;
+
+  const resolvedVariations: ExperimentInterface["variations"] | undefined =
+    hasVariationPayload
+      ? variations.map((v) => ({
+          id: v.id || generateVariationId(),
+          screenshots: v.screenshots || [],
+          ...v,
+        }))
```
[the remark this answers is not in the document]
lukesonnet: Fixed.

packages/front-end/components/Features/RuleModal/index.tsx
lukesonnet: Done, ty.

lukesonnet:
@greptileai

packages/shared/src/validators/experiments.ts
```diff
@@ -79,6 +92,7 @@ export const experimentPhase = z
     namespace: namespaceValue.optional(),
     seed: z.string().optional(),
     variationWeights: z.array(z.number()),
+    variations: z.array(phaseVariation),
```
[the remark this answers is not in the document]
lukesonnet: We prefer to keep these kinds of fields mandatory to prevent additional boilerplate throughout the code to handle such a critical field.

PR #1334, opened 2023-06-06
[stats] Add support for daily and cumulative time series to integrations and gbstats

lukesonnet:
## Features and Changes

* Adds support for cumulative and daily time series, but does not expose them to GrowthBook users yet.
* Also refactors our general query approach to rely on `__userMetric` as our main dataset with all user and metric data (LEFT JOIN metrics on to users rather than an inner join)

## Overview

Adds ability to create cumulative and daily time series. 

Our current time series just are dimension results by user's first date bucketed. So that means on day X, we see experimental effects for all users bucketed on day X, regardless of whether those conversions happened on day X or day X+1 and so on. Furthermore, on day X, we don't have any users that were bucketed on days X-1 and backwards influencing the metric, so it doesn't tell you what happened *on day X* but rather what happened to users *bucketed on day X*.

There's ample demand for Cumulative Time Series, which show you *the total effects for all users and all metrics through day X*. In other words, what is the experiment result as if I had run my results at the end of day X?

We also might care about Daily Time Series, which show you *for all users bucketed through day X, what is the effect on metrics that were "accrued" on day X*. This probably only matters for those using the `Experiment Duration` attribution model or long conversion windows, but can let you know what is happening on given days in your experimental population as the experiment goes on. This is effectively the derivative of the Cumulative Time Series, but with appropriate statistics rather than simply diff'ing results in the front end.

This PR adds the necessary components to `gbstats` and our SQL integrations to compute these last two series: Cumulative and Daily Time Series.

## How?

* Uses`getDateTable` to build a time series of dates and then cross joining user metric results to that before aggregating. This method is a generic method that should work across all SQL engines and is very cheap. Other methods, such as UNNESTing, may be more compact in their SQL but are slightly less performant and vary across engines.
* Reworking the entire SQL query to focus on LEFT JOINing `__metric` to `__distinctUsers` to rely on this as our main count of users and to left join other auxiliary metrics on to (e.g. denominator and/or CUPED metrics). This is also where we cross join with the above date range to build our time series.
* Splits the `__userMetric` CTE into `__userMetricJoin` and `__userMetricAgg` so that custom aggregations can operate over a column of metric values that are `COALESCE(value, 0)` for users who have 1+ conversion and are NULL for users with 0 matching conversions.
[...]

PR #5512, opened 2026-03-19
perf: eliminate double fact-table scan in CUPED queries

nadapzy:
## Problem

When regression adjustment (CUPED) is enabled, `getExperimentFactMetricsQuery()` emits `__factTable` as a CTE and references it **twice**:
- `__userMetricJoin` — post-exposure metrics (LEFT JOIN)
- `__userCovariateMetric` — pre-exposure CUPED covariates (INNER JOIN)

This can cause unnecessary slot usage as query planners can more efficiently compute covariate data in a single pass through the fact table in a single join. 

This split was originally done as a part of #1334 when we needed to split out the computation for time series, but we ended up going a different direction for time series so that split in CTEs and JOINing twice no longer has any value.

## Fix

Since `metricStart` already subtracts `regressionAdjustmentHours` (see `getMetricStart` call), `__factTable` already contains all pre-exposure rows. The second scan re-reads bytes the first scan already filtered.

This PR folds the pre-exposure `CASE WHEN m.timestamp >= d.${alias}_preexposure_start ...` columns directly into `__userMetricJoin` alongside the post-exposure columns. `__userCovariateMetric` then becomes a cheap `GROUP BY` over `__userMetricJoin` — no second `__factTable` reference.

## Semantics

The original `__userCovariateMetric` uses INNER JOIN (users with zero pre-exposure events are absent). The new version uses LEFT+CASE (those users get NULL covariate values, which aggregate to 0).

Downstream consumption is unchanged:
- `LEFT JOIN __userCovariateMetric c` at lines ~4393/4503/4809 — already LEFT
- `capCoalesceValue()` wraps all columns with `COALESCE(${valueCol}, 0)` — already handles NULL

| Path | Original | New |
|---|---|---|
| User with 0 pre-exposure events | absent → `c.X=NULL` → `COALESCE→0` | present with `SUM(NULL)=0` → `c.X=0` → `COALESCE→0` |
| Final aggregation | 0 | 0 |

## Testing

- [x] Snapshot tests updated (SQL structure changes, final aggregations identical)
- [X] Manual verification against a live experiment (cross-fact table ratio metric; legacy metrics; incremental refresh all looking good)

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -967,7 +967,7 @@ export default abstract class SqlIntegration
       },
       null,
       [{ factTable, index: 0 }],
-      false,
+      "m",
       "m0",
     );
 
@@ -7725,7 +7725,7 @@ ORDER BY column_name, count DESC
 
     const { factTablesWithMetricData } = this.parseExperimentFactMetricsParams({
       ...params,
-      // Covariate data joined to single table with `m` prefix before columns are extracted
+      // Covariate data joined to single table with `m` alias before columns are extracted
       covariateTableAlias: "m",
     });
 


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -967,6 +967,7 @@ export default abstract class SqlIntegration
       },
       null,
       [{ factTable, index: 0 }],
+      false,
```
ahdriel: issue: This is breaking the CI -- should be a string.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3449,6 +3378,36 @@ export default abstract class SqlIntegration
                   `,
               )
               .join("\n")}
+            ${
+              // CUPED pre-exposure covariate columns: emitted here so that
+              // __userCovariateMetric can aggregate them from __userMetricJoin
+              // instead of re-scanning __factTable. See getCovariateMetricCTE.
```
ahdriel, line 3384: question: do we need to update this comment as we have removed `getCovariateMetricCTE`?

PR #5519, opened 2026-03-20
Implement Presto timeout and cancel

tzjames:
### Features and Changes
Presto seems to somehow die sometimes in a way that leaves queries hanging. The heartbeat keeps ticking and we never cancel the query.  

We had been overriding the default timeout of the presto client to 0, meaning it never timesout.  The presto-client normally has a timeout of 60s. The presto-database normally has a timeout of 5min.  Here we use a default timeout of 60min, which is more in line with some of our other databases, and is less likely to catch users unaware of this somewhat breaking change.  The likelihood of it breaking anyone is very low, and the fix if anyone were to hit it would be to simply set a higher default timeout of have them explicitly set it to 0.

We also implement CancelQuery.

Also there was a bug when entering in "0" as string.  The presto-client actually set it to timeout immediately instead of forever.  Casting it to a number does make it so it never timesout.

Furthermore there was a bug on the cancel query button which kicked off another query  - that is also fixed.

### Testing

- [x] With a [trino datasource](https://github.com/growthbook/trino-hive-hdfs) see that the externalId gets set. 
- [x] Do a complicated query and hit cancel immediately and see that it cancels it in the database: http://localhost:8080/ui/#
- [x] Enter in 0 in the datasource timeout settings, run the query and see it not time out.
- [x] Enter in 3s in the datasource timeout settings, run the query and see that it times out.

### DIFF
--- a/packages/front-end/components/Queries/RunQueriesButton.tsx
+++ b/packages/front-end/components/Queries/RunQueriesButton.tsx
@@ -180,12 +180,15 @@ const RunQueriesButton = forwardRef<HTMLButtonElement, Props>(
                 flipTheme={false}
               >
                 <IconButton
+                  type="button"
                   variant="solid"
                   color="tomato"
                   size="2"
                   style={{ width: 20, height: 20, padding: 2 }}
                   radius="full"
-                  onClick={async () => {
+                  onClick={async (e) => {
+                    e.preventDefault();
+                    e.stopPropagation();
                     resetFilters?.();
                     try {
                       await apiCall(cancelEndpoint, { method: "POST" });


packages/back-end/src/integrations/Presto.ts
```diff
@@ -41,24 +42,23 @@ export default class Presto extends SqlIntegration {
...
       schema: this.params.schema,
       catalog: this.params.catalog,
-      timeout: this.params.requestTimeout ?? 0,
       checkInterval: 500,
     };
+    if (
+      this.params.requestTimeout !== undefined &&
+      this.params.requestTimeout !== null
+    ) {
+      configOptions.timeout = this.params.requestTimeout;
```
ahdriel: The official docs state this option as

> The seconds that a query is allowed to run before it starts returning results, defaults to 60 seconds. Set to null or 0 to disable.

So I am curious if the `returning results` means some kind of streaming or if this will require all Trino queries to be finished in 60 seconds or less now.

If it is the latter I am concerned about the potential impact to users.

tzjames: I think it does mean that it will cancel queries after 60s.  People can just set the requestTimeout then in their connection string to fix it. I'm not sure what the transition plan should be here. Having the default be infinite is also problematic.

ahdriel: I am pro having a higher default value -- maybe 30m or 60m? Regardless, it'll be a breaking change so we should note that in the release notes.

Given the goal we want to solve right now is not resource exhaustion on our servers, I would lean towards being generous with the timeout, while still making sure that these queries are eventually resolved.

bearloga: Hi! Mikhail from WMF here, where we're using the Presto connector both in GrowthBook and Superset. Hope it's okay to chime in with an unsolicited comment!

[Presto docs](https://prestodb.io/docs/current/admin/properties.html#query-client-timeout) say that the default value for `query.client.timeout`
> how long a query runs without contact from the client application, such as the CLI, before it’s abandoned

is 5 minutes

So one option is to go with that. From personal experience, Adriel's 30m/60m suggestion is very realistic, while at the same time very reasonable.

tzjames: Ok going with 60 minutes.  It is a breaking change, but few if anyone should have queries lasting that long, and there is a simple fix to set the timeout to 0 manually or something larger.

packages/back-end/src/integrations/Presto.ts
```diff
@@ -22,6 +23,24 @@ type Row = any;
...
+    return DEFAULT_PRESTO_REQUEST_TIMEOUT_SEC;
+  }
+  if (typeof requestTimeout === "string") {
+    const t = requestTimeout.trim();
+    if (t === "") return DEFAULT_PRESTO_REQUEST_TIMEOUT_SEC;
+    const n = Number.parseInt(t);
+    return Number.isNaN(n) ? DEFAULT_PRESTO_REQUEST_TIMEOUT_SEC : n;
+  }
+  const n = Number(requestTimeout);
+  return Number.isNaN(n) ? DEFAULT_PRESTO_REQUEST_TIMEOUT_SEC : n;
```
ahdriel, line 41: nit: This is nice defensive programming, but would be good if we have a helper that parses it all as either an `int` or fallbacks to the provided value.

packages/front-end/components/Queries/RunQueriesButton.tsx
```diff
@@ -180,12 +180,15 @@ const RunQueriesButton = forwardRef<HTMLButtonElement, Props>(
                 flipTheme={false}
               >
                 <IconButton
+                  type="button"
```
ahdriel, line 183: nit: should this be default of `IconButton`?

because otherwise the default is `submit` which is usually not what we intend, right?

tzjames: Perhaps one day we can make our own IconButton, but we are using RadixUi's which keeps the default html of submit.  Most of the other places in our code base either are setting type=button, or are not in forms so it doesn't matter.

packages/back-end/src/integrations/Presto.ts
tzjames: Good idea.  I'll make another PR to handle that.

PR #5520, opened 2026-03-20
Add a cancel button for queries

tzjames:
### Features and Changes

On the Recent Queries page we didn't have a way to cancel running queries.  When queries are mistakenly in a running state and are failed to be killed for some reason or another it would be good if people had a manual was of stopping them.  This PR takes care of that. 

### Testing
In Mongo Compass sort queries by `createdAt: -1`
In the first query delete its `finishedAt` property and set it's `status` to `running`.
Go to that datasource page, click the three dots menu, click "View Queries"
Sort the queries be `startedAt` desc.
See the cancel button.
Click it.
See that it shows the failed icon.
See in Mongo compass that the query shows `failed` status.

### Screenshots
Screenshot 2026-03-20 at 4 34 13 PM
Screenshot 2026-03-20 at 4 35 32 PM
Screenshot 2026-03-20 at 4 35 25 PM

### DIFF
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -48,7 +48,7 @@ import {
   getQueriesByDatasource,
   getQueriesByIds,
   getQueryById,
-  updateQuery,
+  updateQueryIfRunning,
 } from "back-end/src/models/QueryModel";
 import { findDimensionsByDataSource } from "back-end/src/models/DimensionModel";
 import {
@@ -1076,7 +1076,9 @@ export async function cancelDataSourceQuery(
     error: `Query cancelled by user (${cancelledBy})`,
   });
   if (!updated) {
-    throw new Error("Query is no longer in running state and cannot be cancelled");
+    throw new Error(
+      "Query is no longer in running state and cannot be cancelled",
+    );
   }
 
   res.status(200).json({ status: 200 });


packages/back-end/src/controllers/datasources.ts
```diff
@@ -1020,6 +1023,58 @@ export async function getDataSourceQueries(
...
+    } catch (e) {
+      // Log but continue - we'll still mark the query as failed
+      logger.debug(e, `Failed to cancel query on warehouse: ${e.message}`);
+    }
+  }
+
+  await updateQuery(context, query, {
+    status: "failed",
+    finishedAt: new Date(),
+    error: "Query cancelled by user",
```
ahdriel: nit: Do we want to add the user who took the action here? either email or something?

Or is there another place where we can see this information?

packages/front-end/pages/datasources/queries/[did].tsx
```diff
@@ -180,40 +189,72 @@ const DataSourceQueries = (): React.ReactElement => {
...
+                          style={{ width: 20, height: 20, padding: 2 }}
+                          radius="full"
+                          onClick={async (e) => {
+                            e.stopPropagation();
+                            try {
+                              await apiCall(
+                                `/datasource/${did}/query/${query.id}/cancel`,
+                                { method: "POST" },
+                              );
+                              await mutate();
```
Comment (ahdriel) [inline packages/front-end/pages/datasources/queries/[did].tsx:262]: nit: Move `mutate` to `finally` block to ensure we have latest status in the UI even if there's an error (maybe it was already finished / cancelled by then?)

packages/front-end/pages/datasources/queries/[did].tsx
```diff
@@ -180,40 +189,72 @@ const DataSourceQueries = (): React.ReactElement => {
...
+                            } catch (err) {
+                              console.error(err);
```
Comment (ahdriel) [inline packages/front-end/pages/datasources/queries/[did].tsx]: question: Should we surface this in the UI? The user should know what happened if they clicked it.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -1020,6 +1023,58 @@ export async function getDataSourceQueries(
...
+    throw new Error("Only running queries can be cancelled");
+  }
+
+  const integration = await getIntegrationFromDatasourceId(
+    context,
+    datasourceId,
+    true,
+  );
+
+  if (integration.cancelQuery && query.externalId) {
```
ahdriel, line 1060: question: For datasources where we do not have the ability to cancel the query -- should we error out instead?

I understand the desire to mark it as `failed` on our end to allow other queries to run, but then this can break the `concurrency limit` settings we have on our end and allow more than X queries to be running at the same time, for example.

I am not sure what approach I prefer though.

packages/back-end/src/controllers/datasources.ts
```diff
@@ -1020,6 +1023,58 @@ export async function getDataSourceQueries(
...
+    try {
+      await integration.cancelQuery(query.externalId);
+    } catch (e) {
+      // Log but continue - we'll still mark the query as failed
+      logger.debug(e, `Failed to cancel query on warehouse: ${e.message}`);
+    }
+  }
+
+  await updateQuery(context, query, {
```
ahdriel: question: How does this work with our QueryRunner, which is in-memory?

If `cancelQuery` is
- supported: We need to ensure that the QueryRunner also gives up;
- not supported: The query will potentially finish sometime down the line and overwrite `failed` to `succeeded` or something.

packages/back-end/src/controllers/datasources.ts
tzjames: I think this will be a very rarely used feature, especially seeing that cancel button was broken in that a new query would immediately run and no one reported on it. I also think that the concurrency limit settings are also rarely used. Running an extra query on the very rare case someone clicks cancel is not ideal, but this scenario is pretty unlikely to happen. That said not exposing that we are not actually cancelling the query on the server when someone clicks the cancel button is not cool. I think the solution there is to actually implement cancel for the DB integrations that don't have it. Most clients actually support it. It's not been implemented because we have never done it.

packages/back-end/src/controllers/datasources.ts
tzjames: If `cancelQuery` is

- supported: then the query in the warehouse will be killed. The promise in the QueryRunner will then error and the queryModel document will be updated that it failed, losing whatever user info we might have originally had on the model. I'll put in a check to keep it.
- not supported: If the query kept on running and actually finished, then that is more of the truth than the fact that it was cancelled and failed.

PR #5534, opened 2026-03-24
Fix #4789: Apply activation date filter for activation status.

### LINKED ISSUE #4789: [Bug] Activation Status dimension grouping does not apply activation date filter
### Summary

When grouping by the Activation Status dimension, GrowthBook does not apply the activation date filter. The generated SQL classifies users as “Activated/Not Activated” based only on NULL/NOT NULL of first_activation_timestamp, while the non-dimensioned query correctly applies a date boundary on first_activation_timestamp. This leads to inconsistent activation counts and experiment results depending on whether the Activation Status dimension is used.

Example SQL snippets:

With Activation Status dimension (date filter missing; uses only NULL/NOT NULL):

```
__distinctUsers AS (
  SELECT
    user_id,

    if(first_activation_timestamp IS NULL, 'Not Activated', 'Activated') AS dimension,

    variation,
    first_exposure_timestamp AS timestamp,
    dateTrunc('day', first_exposure_timestamp) AS first_exposure_date,
    first_exposure_timestamp AS preexposure_end,
    dateSub(hour, 672, first_exposure_timestamp) AS preexposure_start
  FROM __experimentUnits

  WHERE first_exposure_timestamp <= toDateTime('some date', 'UTC')
)
```
Without Activation Status dimension (correctly applies activation date filter):

```
__distinctUsers AS (
  SELECT
    user_id,

    toString('') AS dimension,

    variation,
    first_activation_timestamp AS timestamp,
    dateTrunc('day', first_exposure_timestamp) AS first_exposure_date,
    first_exposure_timestamp AS preexposure_end,
    dateSub(hour, 672, first_exposure_timestamp) AS preexposure_start
  FROM __experimentUnits

  WHERE first_activation_timestamp IS NOT NULL
    AND first_activation_timestamp <= toDateTime('some date', 'UTC')
)
```
#### Expected Behavior
Activation status should be determined consistently with the base activation logic, including the activation date constraint.
When the Activation Status dimension is used, “Activated” users should satisfy:
first_activation_timestamp IS NOT NULL
first_activation_timestamp <= <as_of/evaluation timestamp>
“Not Activated” should include all users who do not meet the above, ensuring parity with the non-dimension query.

#### Current Behavior
With the Activation Status dimension, users are labeled “Activated” solely based on first_activation_timestamp IS NOT NULL, regardless of whether the activation happened after the evaluation timestamp. This inflates activation counts and diverges from results when the dimension is not used.

joaomsgomes:
When grouping experiment results by the Activation Status dimension, users were incorrectly classified as Activated based solely on the presence of a non-null first_activation_timestamp, without checking whether the activation occurred within the experiment's evaluation window. This led to inflated activation counts and inconsistent results compared to queries without the dimension.

### Implementation

- Modified SqlIntegration.ts to include a date boundary condition in the activation status classification logic. Users are now classified as Activated only if first_activation_timestamp is non-null and falls within the evaluation window.

### Tests

- Added a unit test verifying that the generated SQL includes the correct date boundary condition when an activation metric is configured.
- Added a unit test ensuring no activation logic is generated when no activation metric is provided.

### Features and Changes

This fix corrects a SQL generation bug in the Activation Status dimension. No new features or APIs are introduced.

### Dependencies

 None.

### Testing

Unit tests can be run with:

cd packages/back-end/test
pnpm test activation-dimension-test.test

### Screenshots
N/A — this is a backend SQL generation fix with no UI changes.

- Closes **[#4789](https://github.com/growthbook/growthbook/issues/4789)**


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -3826,7 +3825,7 @@ export default abstract class SqlIntegration
       case "activation":
         return {
           value: this.ifElse(
-            `first_activation_timestamp IS NULL`,
+            `first_activation_timestamp IS NULL OR first_activation_timestamp > first_exposure_timestamp`,
```
ahdriel: Unfortunately we cannot use `first_exposure_timestamp` here for the comparison due to how it is calculated.

The conversion window clause (`a.timestamp >= e.timestamp + delay`) already guarantees that activation timestamps are greater than exposure timestamps. So here the current code would treat all users as `Not Activated` which is wrong.

I believe we could either check `> experiment.endDate` to exclude activations that happened after the experiment has concluded, or we could simplify and remove this new condition, keeping only `first_activation_timestamp IS NULL` -- which will get filtered out in the aggregate query if needed.

tl;dr: We should undo this change.

ahdriel (commented):
Thanks for the contribution!

packages/back-end/src/integrations/SqlIntegration.ts
joaomsgomes: You're right! Thanks for the clarification.

Since the conversion window already guarantees the ordering between exposure and activation, this condition is indeed redundant and incorrect.

I'll revert this change and keep only the `first_activation_timestamp IS NULL` check.

lukesonnet:
@greptileai

joaomsgomes:
Regarding the P1 Findings: First I want to check with you @lukesonnet, would you like me to address this as well, or would you prefer to handle it yourself given your familiarity with that part of the code?
Regarding the P2 findings: both are valid and are in the test file I wrote. I'll fix the loose assertion to verify the actual timestamp value, and replace the as unknown as cast with a fully typed params object. I'll commit those changes shortly.

joaomsgomes:
Hi @lukesonnet, just following up on my previous message in case it got lost. Wanted to check if you had any thoughts on the P1 finding, or whether you'd prefer to handle that part yourself. Happy to address it on my end if that's helpful. Thanks again for the review!

joaomsgomes:
Hi @lukesonnet, just wanted to follow up once more on the P1 finding. Completely understand things can get busy, so no rush at all. I'm happy to take care of it on my end if that's easier, or leave it to you if you'd prefer to handle that part of the code yourself. Either way works for me. Thanks again for your time on this review!

joaomsgomes:
@lukesonnet ping

PR #5536, opened 2026-03-24
fix(feature-usage): Use compileSqlTemplate helper for feature usage queries

msamper:
### Features and Changes

We previously were not using our `compileSqlTemplate` helper function for the feature usage query so template variable replacement wasn't working. This PR adds usage of the helper for the feature usage query.

### Testing

- [x] Add a feature usage query that uses template variables
- [x] Make sure the query is compiled correctly when testing the query
- [x] Make sure the query is compiled correctly when running the query from a feature diagnostics tab

### DIFF
--- a/packages/back-end/test/sqlintegration.test.ts
+++ b/packages/back-end/test/sqlintegration.test.ts
@@ -1252,3 +1252,43 @@ describe("full fact metric experiment query - bigquery", () => {
     },
   );
 });
+
+describe("getFeatureEvalDiagnosticsQuery", () => {
+  beforeEach(() => {
+    jest.useFakeTimers("modern");
+    jest.setSystemTime(new Date("2025-03-24T12:00:00.000Z"));
+  });
+
+  afterEach(() => {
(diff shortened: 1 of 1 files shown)


packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2669,10 +2669,14 @@ export default abstract class SqlIntegration
       ? this.datasource.settings.queries.featureUsage[0].query
       : "";
 
+    const compiledFeatureEvalQuery = compileSqlTemplate(featureEvalQuery, {
+      startDate: oneWeekAgo,
```
ahdriel, line 2673: question: Do we surface the available fields in the UI anywhere?

If not, we likely should, but maybe as part of a separate PR.

msamper: From what I can tell I don't think we do. I think they're only listed in our docs. Agree that we should add separately
https://docs.growthbook.io/app/sql-templates

PR #5542, opened 2026-03-25
Allow scoping API keys by environment and project

Kevin-Chant:
### Features and Changes

Adds the `memberRoleWithProjects` fields to API keys so they can benefit from the same level of granularity as user permissions.
Updates the ReqContextClass to respect these new limitations on API keys when present
Changes the frontend selector for API key roles to use the existing RoleSelector
Adds new columns for the fields to the API keys table

Notably does not include the `teams` field of `memberRoleInfo` to simplify things and because we'll likely want to create Service Accounts in the near future which are a better abstraction here than applying teams to keys directly

### Testing

- [x] Creating API keys allows for specifying environment access and project roles (and environment access for project roles)
- [x] The new data is rendered correctly in the Secret API Keys table
- [x] Creating PATs doesn't have the new inputs

Create an API key with global role `No Access` and project role `Read Only`. Create at least one experiment inside the project and one outside
- [x] The global noaccess key should only yield experiments within its project from GET `/api/v1/experiments`
 
Create an API key with global role `Experimenter` but limited to one environment. Create an SDK connection for that environment and for a different environment
- [x] The environment-scoped key should be able to PUT `/api/v1/sdk-connections/<sdk-conn-in-environment>`
- [x] The environment-scoped key should not be able to PUT `/api/v1/sdk-connections/<sdk-conn-in-other-env>`

Create an API key with global role `No Access` and project role `Engineer` limited to some environments. Check the SDK PUT permissions again
- [x] The environment-scoped project role should be able to PUT `/api/v1/sdk-connections/<sdk-conn-in-environment-and-project>`
- [x] The environment-scoped project role should not be able to PUT `/api/v1/sdk-connections/<sdk-conn-in-other-env>`
- [x] The environment-scoped project role should not be able to PUT `/api/v1/sdk-connections/<sdk-conn-in-env-without-project>`

### Screenshots

image

### DIFF
--- a/packages/back-end/src/models/ApiKeyModel.ts
+++ b/packages/back-end/src/models/ApiKeyModel.ts
@@ -105,6 +105,7 @@ export class ApiKeyModel extends BaseClass {
         for (const pr of doc.projectRoles) {
           this.validateRole(pr.role);
           await this.validateProject(pr.project);
+          this.validateEnvironments(pr.environments);
         }
       }
     }


packages/back-end/src/middleware/authenticateApiRequestMiddleware.ts
```diff
@@ -57,11 +60,8 @@ export default function authenticateApiRequestMiddleware(
 
   // Lookup organization by secret key and store in req
   dangerousLookupOrganizationByApiKey(secretKey)
-    .then(async (apiKeyPartial) => {
-      const { organization, secret, id, userId, role } = apiKeyPartial;
-      if (!organization) {
-        throw new Error("Invalid API key");
-      }
```
Kevin-Chant, line 64: This check and error moved inside `dangerousLookupOrganizationByApiKey`

packages/back-end/src/util/api-key.util.ts
```diff
@@ -73,32 +73,39 @@ export function migrateApiKey(legacyDoc: unknown) {
...
     if (orgs.length === 1) {
-      return {
+      return migrateApiKey({
         id: "SECRET_API_KEY",
         key: SECRET_API_KEY,
         secret: true,
         organization: orgs[0].id,
         role: SECRET_API_KEY_ROLE,
-      };
+        dateCreated: new Date(),
```
Kevin-Chant, line 95: This is necessary for satisfying the interface as we're creating a synthetic record

packages/back-end/src/models/ApiKeyModel.ts
```diff
@@ -66,27 +71,98 @@ export class ApiKeyModel extends BaseClass {
...
+          "Your plan does not support restricting API key permissions by environment.",
+        );
+      }
+      this.validateEnvironments(doc.environments);
+      if (doc.projectRoles) {
+        if (!this.context.hasPremiumFeature("advanced-permissions")) {
+          this.context.throwPlanDoesNotAllowError(
+            "Your plan does not support project-level permissions on API keys.",
+          );
+        }
```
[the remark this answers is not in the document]
Kevin-Chant: Since it's an optional field, enforcing that no value was passed seems fine to me.

PR #5546, opened 2026-03-25
Initial setup of new exploration endpoints

mknowlton89:
### Features and Changes
This PR introduces 3 new REST API endpoints for the product analytics explorer tool.
- POST /api/v1/product-analytics/metric-exploration
- POST /api/v1/product-analytics/fact-table-exploration
- POST /api/v1/product-analytics/data-source-exploration

Three separate endpoints are intentional — each accepts a type-specific request body, which keeps the API straightforward to consume and document. The underlying data model uses a discriminated union on config.type, but exposing a single endpoint for all three types would make request validation and documentation significantly harder to follow.

#### Schema design:
- ProductAnalyticsExplorationBase defines all shared response properties (id, organization, status, dates, result, etc.)
- Each endpoint has its own response schema (ProductAnalyticsMetricExploration, - ProductAnalyticsFactTableExploration, ProductAnalyticsDataSourceExploration) that extends the base and adds a typed config field
- Request body validation is handled per-endpoint via type-specific Zod validators (metricExplorationConfigValidator, factTableExplorationConfigValidator, dataSourceExplorationConfigValidator) that each extend a shared baseExplorationConfigValidator
- Response shape: Each endpoint returns:
   - exploration — the exploration result (or null if no result is available)
   - query — the underlying Query object from the database, useful for inspecting the generated SQL
   - message — present when exploration is null, explaining why
- Cache control: All three endpoints accept a cache query parameter:
   - preferred (default): return a cached result if available, otherwise run a new query
   - never: always run a new query, ignoring cache
   - required: only return a cached result; if none exists, returns exploration: null with a message

Other changes:
- Added toExplorationApiInterface() to AnalyticsExplorationModel to convert internal model documents to the API response shape
- Extracted ExplorationCacheQuery as a shared Zod schema and type in shared/validators, used by both the REST API layer and runProductAnalyticsExploration internally

### Testing
- [x] Ensure you can pass in a valid metric exploration config and the endpoint generates a new valid exploration and returns results
- [x] Ensure you can pass in a valid fact table exploration config and the endpoint generates a new valid exploration and returns results
- [x] Ensure you can pass in a valid data source exploration config and the endpoint generates a new valid exploration and returns results
- [x] Ensure the cache query param is respected and works as expected for all three cache modes (preferred, never, required)

### DIFF
--- a/packages/back-end/src/api/api.router.ts
+++ b/packages/back-end/src/api/api.router.ts
@@ -4,6 +4,7 @@ import { Router, Request, RequestHandler } from "express";
 import rateLimit from "express-rate-limit";
 import bodyParser from "body-parser";
 import * as Sentry from "@sentry/node";
+import { parseEnvInt } from "shared/util";
 import authenticateApiRequestMiddleware from "back-end/src/middleware/authenticateApiRequestMiddleware";
 import { DashboardModel } from "back-end/src/enterprise/models/DashboardModel";
 import { CustomFieldModel } from "back-end/src/models/CustomFieldModel";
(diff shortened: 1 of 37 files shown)


mknowlton89:
@cursor review

PR #5548, opened 2026-03-26
fix: preserve cipher options when switching to Remote Evaluated mode

### LINKED ISSUE #5423: [Bug] Remote Eval: ciphered options not saving
### Summary

When enabled cipher option when using remove eval it's not saving. You can save and then go back and the saving did not persist.

https://github.com/user-attachments/assets/da77a6c5-22c4-4cd2-8928-7586c4ea1232

#### Expected Behavior

Ciphered options saved when using remote eval

#### Current Behavior

Ciphered options not saved when using remote eval

thecurioussailor:
### Features and Changes

When editing an SDK Connection with Remote Evaluated mode, the cipher options
(encryptPayload, hashSecureAttributes, includeExperimentNames) were being
unconditionally reset to their defaults every time the form loaded. This caused
previously saved cipher options to appear lost when reopening the edit modal.

The fix mirrors the existing "Ciphered" tab pattern — cipher option defaults are
only applied when no cipher options are currently set, preserving existing values
on load.

- Closes https://github.com/growthbook/growthbook/issues/5423

### Dependencies

None

### Testing

1. Open an SDK Connection in edit mode
2. Select "Remote Evaluated" mode
3. Check "Hide experiment and variation names"
4. Save
5. Reopen edit modal — cipher option should still be checked

### Screenshots
image


bryce-fitzsimons (approved):

bryce-fitzsimons:
This issue should be resolved now. If any issues remain, feel free to re-open

PR #5553, opened 2026-03-26
Add more metadata to BQ labels

lukesonnet:
### Features and Changes

BQ labels are more restrictive than the generic metadata trino and snowflake let you pass in, but we should try to pass over as much as possible.

This PR does this while fitting within BQs requirements: https://docs.cloud.google.com/bigquery/docs/labels-intro

Note: THIS WILL BREAK QUERIES NOISILY IF THE LABEL IS NOT ACCEPTED. I have confirmed in testing they don't just silently accept malformed labels.

Screenshot 2026-03-26 at 1 07 24 PM


packages/back-end/src/services/bigquery.ts
```diff
@@ -64,3 +65,48 @@ export function getFactTableTypeFromBigQueryType(
     }
   }
 }
+
+/**
```
ahdriel, line 71: Given these break all queries if we get it wrong we should add tests to ensure they are formatted properly.

ahdriel (approved):

lukesonnet:
@greptileai

PR #5555, opened 2026-03-27
Move explore endpoints to apiConfig

Kevin-Chant:
Swaps the implementation of explore endpoints from the legacy yaml format to apiConfig

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -6278,6 +6278,7 @@ paths:
           in: query
           required: false
           schema:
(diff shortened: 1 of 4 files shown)


packages/back-end/src/models/AnalyticsExplorationModel.ts
```diff
@@ -167,19 +242,7 @@ export class AnalyticsExplorationModel extends BaseClass {
...
-      datasource: exploration.datasource,
-      status: exploration.status,
-      dateStart: exploration.dateStart,
-      dateEnd: exploration.dateEnd,
-      error: exploration.error ?? null,
-      result: exploration.result,
-      config: exploration.config,
-    };
+  public toApiInterface(exploration: ProductAnalyticsExploration) {
+    return toApiInterface(exploration);
```
Kevin-Chant, line 246: Moved this outside the class to avoid a circular dependency with the apiConfig definition

PR #5558, opened 2026-03-27
Refactor: Add Utilities for parsing ints and use them everywhere.

tzjames:
### Features and Changes
- **Shared number parsing utilities** Moved parseEnvInt from `python.ts` to `shared/src/util/numbers.ts`. Added additional helpers: `parseOptionalInt`, `parseIntWithDefault`, `parseIntWithDefaultCapped`.
- **AI rate limiting (429)** (`shared/ai.ts`): adds `formatAIRateLimitRetryMessage` for human-readable copy from optional `Retry-After` seconds.

### Dependencies
None.

### Testing
- `pnpm test`
- Start back-end with a deliberately invalid integer env (e.g. `EMAIL_PORT`) and confirm a warning line and fallback default; set valid values and confirm normal behavior.

### DIFF
--- a/packages/shared/src/util/numbers.ts
+++ b/packages/shared/src/util/numbers.ts
@@ -1,9 +1,17 @@
 /**
- * Parses `value` as a base-10 integer. Use when reading datasource / form fields
+ * Parses `value` as an integer. Use when reading datasource / form fields
  * that may be string, number, or missing.
  *
- * Returns `fallback` for `undefined`, `null`, blank strings, non-finite numbers,
- * or strings that do not start with a valid integer (e.g. `"abc"`).
+ * Uses `Number(value)` (after trimming strings), then
+ * `Number.isInteger(n) ? n : Math.trunc(n)` so decimals truncate toward zero
+ * and `NaN`/`Infinity` are rejected. Unlike `parseInt`, there is no partial
+ * parse of strings (`"123abc"` → `NaN`). Blank strings are treated as missing.
+ *
+ * Results must be **safe integers**; values outside that range return
+ * `undefined`.
+ *
+ * Returns `fallback` for `undefined`, `null`, blank strings, non-finite
+ * numbers, or any other invalid input.
  */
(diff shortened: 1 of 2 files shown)


packages/back-end/src/routers/events/events.controller.ts
```diff
@@ -31,9 +32,9 @@ export const getEvents = async (
 
   const eventTypes = JSON.parse(type || "[]");
 
-  const cappedPerPage = Math.min(parseInt(perPage), 100);
+  const cappedPerPage = parseIntWithDefaultCapped(perPage, 30, 100);
```
jdorn, line 35: Where did 30 come from?

packages/shared/src/util/numbers.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (n === undefined) return fallback;
+  return n;
+}
+
+/** Parsed integer or `undefined` when missing, blank, or not a valid integer. */
+export function parseOptionalInt(value: unknown): number | undefined {
+  if (value === undefined || value === null) return undefined;
+  const t = typeof value === "string" ? value.trim() : String(value);
+  if (t === "") return undefined;
+  const n = parseInt(t);
```
jdorn: `parseInt` silently ignores non-numeric characters, truncates decimals, and has other weird quirks.  Do we want that behavior?

```ts
parseInt("123abc") -> 123
parseInt(0.0000005) -> 5
parseInt("1e3") -> 1
parseInt("1.9") -> 1
```

If we want something more strict, the `Number` constructor can work instead

```ts
Number("123abc") -> NaN
Number("1e3") -> 1000

const n = Number("1.9") // 1.9
return Number.isInteger(n) ? n : NaN // Or Math.trunc(n) if we want
```

Either way, we should at least add these as test cases so we're explicit about the expected behavior

packages/back-end/src/routers/events/events.controller.ts
tzjames: the EventsPage has const `[perPage, setPerPage] = useState(30);`.

PR #5564, opened 2026-03-30
feat(api): add POST /api/v1/experiments/{id}/visual-changeset

saurabhkashyap-ui:
### Features and Changes

**Why this matters:** Before this change, there was **no public REST way** to create or apply a **new visual changeset on an experiment** in line with the rest of the v1 API. Integrations and automation had to rely on other flows (e.g. the UI or internal paths), which blocked headless or MCP-style use cases.

**What you get now:** With **`POST /api/v1/experiments/{id}/visual-changeset`**, callers can **create/update the visual changeset for an experiment in one step** by sending the documented JSON body—the same experiment can carry a proper visual changeset payload **without** doing workarounds. This matches how other experiment resources are exposed and keeps behavior discoverable via OpenAPI.

Adds a new authenticated API endpoint: **`POST /api/v1/experiments/{id}/visual-changeset`**, which applies a visual changeset payload to an experiment.

- Implements the route handler and wires it in the experiments router.
- Documents the endpoint in OpenAPI (`openapi.yaml`, path + payload schema) and regenerates shared API types / validators (`pnpm generate-api-types`).

**Usage (high level):** send a `POST` to `/api/v1/experiments/{experimentId}/visual-changeset` with the documented JSON body (see OpenAPI / `PostExperimentVisualChangeset` types). Requires the same auth as other experiment APIs.

### Dependencies

None beyond the normal GrowthBook app stack (existing API auth, MongoDB, etc.).

### Testing

- `pnpm run setup` (if needed after pulling main)
- `pnpm generate-api-types` (clean working tree afterward)
- `pnpm lint:ci`
- `pnpm type-check`
- `pnpm test`
- Manual: call the new endpoint against a dev instance with a valid API key / session and confirm the experiment updates as expected (see request/response in OpenAPI).

### Screenshots

N/A — API-only change (no UI).

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "build:clean": "rimraf dist",
     "build:typescript": "tsc",
-    "dev": "swc src -w --out-dir dist --copy-files",
+    "dev": "swc src -w --out-dir dist",
     "build": "pnpm build:clean && pnpm build:typescript",
     "test": "jest",
     "clean": "rimraf node_modules",


packages/back-end/src/api/openapi/paths/listVisualChangesets.yaml
```diff
@@ -1,32 +1,31 @@
-get:
```
Kevin-Chant, line 1: Needed to remove `get` here to have two operations defined on the same path

packages/shared/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build:clean": "rimraf dist",
     "build:typescript": "tsc",
-    "dev": "swc src -w --out-dir dist",
+    "dev": "swc src -w --out-dir dist --copy-files",
```
Kevin-Chant: I'm dropping this change as it looks unrelated, but let me know if there was a particular reason for it @saurabhkashyap-ui

Kevin-Chant (approved):

packages/shared/package.json
saurabhkashyap-ui: @Kevin-Chant, fine by me.
swc only emits compiled JS/TS by default. '--copy-files' makes the watcher also copy non-code files under 'src' (JSON, etc.) into 'dist', which avoids broken or half-empty 'dist' in dev when something imports those assets. It’s a normal fix and matches how other packages often run SWC in watch mode.

PR #5574, opened 2026-03-31
Feat/api delete specific feature flag rule

HadarRosenthal:
### Features and Changes

Adds a dedicated REST API endpoint to delete a specific rule from a feature flag:

**`POST /api/v1/features/{id}/rule/delete`**

Previously, removing a single rule required fetching the full feature, manually removing the rule from the rules array, and sending the entire updated environment settings back via the update endpoint. This new endpoint simplifies that workflow.

**Usage:**
```bash
curl -X POST https://api.growthbook.io/api/v1/features/my-feature/rule/delete \
  -H "Authorization: Bearer secret_abc123" \
  -d '{ "environment": "production", "ruleId": "fr_abc123" }'
```

Design decisions:
- Uses POST (not DELETE) for consistency with existing action endpoints (toggle, revert)
- Identifies rules by ruleId rather than array index to avoid race conditions for external API consumers
- Creates a published revision with full audit trail
- Returns the updated feature object, consistent with other mutation endpoints
- Respects review requirements and permission checks (canUpdateFeature, canPublishFeature)

### Dependencies

none

### Testing

- Type-checks pass for shared and back-end packages
- 4 new test cases added covering:
  - Successful rule deletion with correct revision creation
  - 400 when feature not found
  - 400 when environment is invalid
  - 400 when ruleId is not found in the environment
  - 400 when required body fields are missing
- Manual testing: create a feature with rules, call the endpoint, verify the rule is removed and the response contains the updated feature

### Screenshots

N/A — backend API only, no UI changes. The endpoint auto-generates in the API docs via the OpenAPI spec:
/api/v1/features/{id}/rule/delete appears under the features tag in the ReDoc documentation.


HadarRosenthal:
cursor review

jdorn:
Thanks for the contribution.  We were already in the middle of a big refactor for feature API endpoints.  There is now a whole set of dedicated endpoints for dealing with feature drafts, including one to delete a rule - https://docs.growthbook.io/api#tag/feature-revisions/operation/deleteFeatureRevisionRule so this PR is no longer needed.

PR #5580, opened 2026-04-01
Bump packages for security vulns

Kevin-Chant:
### Features and Changes

Bumps minor versions for various packages to address known vulnerabilities

### Testing

`pnpm i && pnpm build`
`docker build -t growthbook .`
`docker run -p 3000:3000 -p 3100:3100 growthbook`

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -8,7 +8,7 @@
     "node": ">=16"
   },
   "scripts": {
-    "build:clean": "rimraf dist",
+    "build:clean": "rimraf dist tsconfig.tsbuildinfo",
     "build:typescript": "tsc",
     "dev": "swc src -w --out-dir dist",
     "build": "pnpm build:clean && pnpm build:typescript",


packages/back-end/package.json
```diff
@@ -105,7 +105,7 @@
     "mssql": "^12.2.0",
     "mysql2": "^3.11.3",
     "node-fetch": "^2.6.1",
-    "nodemailer": "^7.0.11",
+    "nodemailer": "^8.0.4",
```
Kevin-Chant, line 108: The changelog only has one breaking change here and it's irrelevant for our usage, should be safe to bump major version

packages/shared/package.json
```diff
@@ -8,7 +8,7 @@
     "node": ">=16"
   },
   "scripts": {
-    "build:clean": "rimraf dist",
+    "build:clean": "rimraf dist tsconfig.tsbuildinfo",
```
Kevin-Chant, line 11: Unrelated to vulns, but I got my local `shared` into a bad state where `pnpm build` was failing. This now cleans the build cache file as well so it doesn't get out of sync

tzjames: That happened to me once before too.

tzjames (approved):

PR #5589, opened 2026-04-02
Resolve owner email input to user id in API

Kevin-Chant:
### Features and Changes

A followup to #5412, adds a helper to resolve user input `owner` fields from email to the user ID
API endpoints will attempt to resolve the user input to a user ID if possible, otherwise storing the input as-is
The resolution is primarily handled in BaseModel, while legacy endpoints call it manually.
For API documentation, adds a reusable OwnerField and OwnerInputField so the field always has an accurate description

### Dependencies

~Should wait until #5613 lands to prevent possible conflicts~

### Testing

- [x] Create a feature in the API with your user ID directly -> should use your user id
- [x] Create a feature in the API with `owner` = your email -> should infer your user id
- [x] Create a feature in the API with an invalid user ID -> should return status 400
- [x] Create a feature in the API with `owner` = another user's email -> should infer that user id
- [x] Create a feature in the API with `owner` = a string name e.g. "Ben" -> should allow that raw string


Kevin-Chant:
cursor review

eslint-suppressions.json
```diff
@@ -913,5 +913,40 @@
     "prettier/prettier": {
       "count": 2
     }
+  },
```
tzjames: I think we don't need to have these suppressions.  We can set the owner to ownerField from z.string() in a all these places.  I'll send you a diff that you can `git apply`

tzjames (approved):
Besides working around needing the suppressions this looks good to go, so approving now, not to block you.

PR #5592, opened 2026-04-02
Remove ingestion endpoint as it has been move to the license server

tzjames:
### Features and Changes
1. Remove ingestion endpoint as it has been move to the license server
2. Also remove any function that was used only by the ingestion endpoint

### Dependencies
Can't land until 
https://github.com/growthbook/terraform/pull/131
https://github.com/growthbook/growthbook-ingestor/pull/56

### Testing
cross-fingers


Kevin-Chant (approved):
Approving pending https://github.com/growthbook/growthbook-ingestor/pull/56

PR #5601, opened 2026-04-03
feat(metrics): lower tail capping

### LINKED ISSUE #4818: [Feature] Allow capping of negative/low values
Requested by a large Enterprise customer.

Similarly to how Winsorization caps high values, we could allow for the capping of negative values. Example use case would include metrics related to refunds (negative currency amounts) or very low profit-per-user.

There is not currently a way to do this, as Winsorization (which we do offer) only caps positive values.

a potential workaround would be to create a second metric based on a different column where the value doesn't go negative at all.

@lukebrawleysmith and @lukesonnet

lukebrawleysmith:
_Requested by **Luke Sonnet** · [Slack thread](https://growthbookapp.slack.com/archives/C046L8DQD6F/p1783514636565819?thread_ts=1783514636.565819&cid=C046L8DQD6F)_

## What & why

Fact metrics can now cap (winsorize) the **lower** tail independently of the upper tail. Previously capping only trimmed the top of the distribution, so large negative values (refunds, credits, corrections) skewed metric aggregates with no way to bound them.

**Before:** a fact metric had a single capping setting (`absolute` or `percentile`) that only limited the upper tail.

**After:** a fact metric can set an upper cap and a lower cap **independently**, including a different type per tail. For example, absolute `0` on the lower bound (clamp negatives to zero) combined with a 99th-percentile cap on the upper bound. Legacy (non-fact) metrics are unchanged and remain upper-only.

## How

- **Data model:** `cappingSettingsValidator` keeps its original single-tail shape and continues to represent the *upper* tail (backward compatible). A new optional `lowerCappingSettings` field on `factMetricValidator` holds the lower tail with its own `type` / `value` / `ignoreZeros`. `getCappingTailState` and `validateCappingSettingsOrdering` evaluate the two objects independently — ordering is enforced only when both tails share a type; mixed types are allowed.
- **SQL:** the cap primitive builds the upper bound (`LEAST`) and lower bound (`GREATEST`) from separate settings objects, so mixed types work. Lower percentile caps get their own subquery and threshold column.
- **Frontend:** the fact metric modal's capping form now has two independent tail editors, each with its own type selector. The legacy metric form stays upper-only.
- Scoped to fact metrics only. The legacy metric form, `metric/[mid]` display, Mixpanel, and product-analytics keep their upper-only behavior (reverted to `main`).

Merged the latest `main` (the branch was ~293 commits behind) and resolved conflicts preferring `main`.

Closes #4818.

Updated Metric Detail page below.  

Screenshot 2026-07-22 at 4 42 47 PM


packages/back-end/src/api/fact-metrics/postFactMetric.ts
```diff
@@ -169,10 +169,29 @@ export async function getCreateMetricPropsFromBody(
...
+      usesUpperPercentile || usesLowerPercentile
+        ? cs.ignoreZeros || false
+        : false;
+    data.cappingSettings = {
+      type: upperActive ? capType : "",
+      value: upperActive ? cs.value || 0 : 0,
+      ignoreZeros: ignoreZerosForPercentiles,
+      lowerType: lowerActive ? lowType : "",
+      lowerValue: lowerActive ? (cs.lowerValue ?? 0) : 0,
+    };
```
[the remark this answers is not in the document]
lukebrawleysmith: These should be resolved with latest 2 commits.

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -1544,7 +1579,10 @@ export default function FactMetricModal({
           if (values.numerator.column !== "$$distinctUsers") {
             values.numerator.aggregateFilterColumn = "";
           } else {
-            if (values.cappingSettings?.type) {
+            if (
+              values.cappingSettings?.type === "percentile" ||
+              values.cappingSettings?.lowerType === "percentile"
+            ) {
```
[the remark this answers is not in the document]
lukebrawleysmith: should be fixed with latest commit

packages/back-end/src/api/fact-metrics/postFactMetric.ts
```diff
@@ -169,10 +169,29 @@ export async function getCreateMetricPropsFromBody(
...
   }
```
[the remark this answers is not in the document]
lukebrawleysmith: removed lowerIgnoreZeros

packages/shared/src/validators/fact-table.ts
```diff
@@ -168,8 +168,52 @@ export const cappingSettingsValidator = z
...
+      data.lowerValue != null &&
+      data.lowerValue >= data.value
+    ) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "Lower percentile (lowerValue) must be less than upper percentile (value).",
+        path: ["lowerValue"],
+      });
+    }
```
msamper: I think we should move this validation logic to the `customValidation` function on the `FactMetricModel` to co-locate it with other validation that we do

packages/back-end/src/util/migrations.ts
```diff
@@ -150,6 +150,12 @@ export function upgradeMetricDoc(doc: LegacyMetricInterface): MetricInterface {
   delete newDoc.conversionDelayHours;
   delete newDoc.conversionWindowHours;
 
+  // Lower-tail capping is for Fact Metrics (SQL) only, not legacy metrics.
+  if (newDoc.cappingSettings) {
+    delete (newDoc.cappingSettings as { lowerType?: unknown }).lowerType;
+    delete (newDoc.cappingSettings as { lowerValue?: unknown }).lowerValue;
+  }
+
```
msamper: Do we need this? Shouldn't we be blocking adding `lowerType` and `lowerValue` to legacy metrics?

packages/back-end/src/services/experimentQueries/experimentQueries.ts
```diff
@@ -80,23 +83,44 @@ export function getNonQuantileFloatColumns({
...
+          break;
+        case "ratio":
+          cols.push(
+            ...BASE_METRIC_LOWER_PERCENTILE_CAPPING_FLOAT_COLS,
+            ...RATIO_METRIC_LOWER_PERCENTILE_CAPPING_FLOAT_COLS,
+          );
+          break;
+        case "quantile":
+          break;
+      }
```
msamper, line 152: Rather than creating a second `switch`, you should be able to add these additional cols in line with the cases above when the metric `isLowerPercentileCappedMetric` Unless its possible that a metric can be lower percentile capped but not upper percentile capped

msamper (commented):
Still need to test the UI code, but these are my comments from a first pass. Will test the UI and leave additional feedback if needed.

packages/shared/src/validators/fact-table.ts
lukebrawleysmith: Happy to talk this out, but cappingSettingsValidator is used in SafeRollouts as well (https://github.com/growthbook/growthbook/blob/3c2eeb1e141ebc8807174762ac02eb94f63046fa/packages/shared/src/validators/safe-rollout-snapshot.ts#L127).  

Is it worth moving it?

packages/back-end/src/services/experimentQueries/experimentQueries.ts
lukebrawleysmith: A metric can be lower percentile capped but not upper percentile capped

packages/back-end/src/util/migrations.ts
lukebrawleysmith: We have blocked creation of lower tail capped legacy metrics.  It's a safety check.  If someone were to modify a legacy metric stored in Mongo or a config file or something, then this would ensure that the metric object is not lower capped.  

It's an edge case.  If you think we should remove it, I'm open.

packages/shared/src/validators/fact-table.ts
msamper: I think so. There's no reason to do this validation when creating a safe rollout snapshot since the capping settings are coming from the metrics themselves. I think it's also nice to be consistent and use `customValidation` on the `BaseModel` rather than using `superRefine` . We only really need to be validating when writing the Fact Metric

lukebrawleysmith: Thanks, addressed.

packages/back-end/src/util/migrations.ts
lukesonnet: Let's just remove it please!

packages/back-end/src/services/experiments.ts
```diff
@@ -3114,8 +3016,9 @@ export function toMetricApiInterface(
       goal: metric.inverse ? "decrease" : "increase",
       cappingSettings: metric.cappingSettings
         ? {
-            ...metric.cappingSettings,
             type: metric.cappingSettings.type || "none",
+            value: metric.cappingSettings.value,
+            ignoreZeros: metric.cappingSettings.ignoreZeros,
```
lukesonnet: let's also just leave this alone

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -104,6 +105,41 @@ function validateUserFilter({
   }
 }
 
+/** Upper/lower capping ordering (logic formerly on cappingSettingsValidator in Zod). */
```
lukesonnet: >logic formerly on cappingSettingsValidator in Zod

Was it? I think this comment might be wrong.

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -104,6 +105,41 @@ function validateUserFilter({
...
+function validateCappingSettingsOrdering(cs: MetricCappingSettings): void {
```
lukesonnet: Should this method be in `shared/src/experiments/experiments.ts` and also used in `FactMetricModal` instead of the duplicated logic we have there?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5057,27 +5069,124 @@ export default abstract class SqlIntegration
...
+    const cs = m.metric.cappingSettings;
+    const upper =
+      isPercentileCappedMetric(m.metric) &&
+      cs.value != null &&
+      cs.value > 0 &&
+      cs.value < 1
+        ? { pct: cs.value }
+        : undefined;
+    const lower = isLowerPercentileCappedMetric(m.metric)
+      ? { pct: cs.lowerValue as number }
```
lukesonnet: Instead of `as` can we do `isLowerPercentileCappedMetric(m.metric) && cs.lowerValue !== undefined` or something?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5057,27 +5069,124 @@ export default abstract class SqlIntegration
...
+          `${m.alias}_denominator_cap`,
+          m.denominatorSourceIndex,
+        ),
+      );
+    }
+    return out;
+  }
+
+  /** Legacy experiment metrics: one value column and `value_cap` / `value_cap_lower` outputs. */
+  private getLegacyMetricPercentileCapRows(
```
lukesonnet: Legacy metrics should not have lower percentile data, so we can just ignore those fields here and we can use the right legacy metric type for `metric` in the method signature.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5088,48 +5197,61 @@ export default abstract class SqlIntegration
...
-      columnRef: columnRef || null,
-      column: valueCol,
-      ignoreInvalid: true,
-    });
-    if (filters.length) {
-      valueCol = `(CASE WHEN ${filters.join(" AND ")} THEN 1 ELSE NULL END)`;
+    if (hasUpperAbs || hasUpperPct || hasLowerAbs || hasLowerPct) {
+      let expr = this.ensureFloat(`COALESCE(${innerCol}, 0)`);
+      if (hasUpperAbs) {
+        expr = `LEAST(${expr}, ${cs!.value})`;
```
lukesonnet: We try not to use the ! notation like this. In general, try to avoid using as or xx!.yy if there's another reasonable option, even if it requires copying some checks. You could also just consider materializing cs.value and cs.lowervalue as part of the upper checks so that the value itself can be used rather than a boolean.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5057,27 +5069,124 @@ export default abstract class SqlIntegration
...
+    if (!needsPercentileCapSubquery(m.metric)) {
+      return [];
+    }
+    const cs = m.metric.cappingSettings;
+    const upper =
+      isPercentileCappedMetric(m.metric) &&
+      cs.value != null &&
+      cs.value > 0 &&
+      cs.value < 1
+        ? { pct: cs.value }
```
lukesonnet: Why does this need to be an object. Can't it just be a number that gets passed to `row`?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -5057,27 +5069,124 @@ export default abstract class SqlIntegration
...
+  private getFactMetricPercentileCapRows(
+    m: FactMetricData,
+  ): FactMetricPercentileData[] {
+    if (!needsPercentileCapSubquery(m.metric)) {
+      return [];
+    }
+    const cs = m.metric.cappingSettings;
+    const upper =
+      isPercentileCappedMetric(m.metric) &&
+      cs.value != null &&
```
lukesonnet: Aren't these chceks redundant?

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1135,34 +1089,22 @@ export default abstract class SqlIntegration
...
-                        sourceIndex: metricData.denominatorSourceIndex,
-                      },
-                    ]
-                  : []),
-              ],
+              this.getFactMetricPercentileCapRows(metricData).map((row) => {
+                const p = `${metricData.alias}_`;
+                return {
+                  ...row,
+                  valueCol: row.valueCol.startsWith(p)
```
lukesonnet: Can't this logic just live inside the `getFactMetricPercentileCapRows` method?

lukesonnet (commented):
Left a few comments throughout. My main question:

Should we just have a second field in `FactMetricValidator` that is `lowerCappingSettings` that also accepts a full `cappingSettingsValidator`?

Feels cleaner, allows us to have separate ignore zeros (probably not important), and just block this capability from legacy metrics overall.

Not sure if you explored that option.

packages/back-end/src/util/migrations.ts
lukebrawleysmith: removed

lukebrawleysmith:
Regarding the comment about deleting check for capping settings in Migrations.ts, I removed the check. 
https://github.com/growthbook/growthbook/pull/5601#discussion_r3080391296

packages/back-end/src/services/experiments.ts
lukebrawleysmith: If we leave this alone, would it prevent customers from creating a lower-tail capped metric via API?

packages/back-end/src/models/FactMetricModel.ts
lukebrawleysmith: removed comment

packages/back-end/src/models/FactMetricModel.ts
lukebrawleysmith: I moved it to shared/src/validators, as this applies outside of experiments.  Happy to chat more if this is wrong.

packages/back-end/src/integrations/SqlIntegration.ts

packages/back-end/src/integrations/SqlIntegration.ts

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: thanks for the tip, updated

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: simplified it, thanks

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: removed

packages/back-end/src/integrations/SqlIntegration.ts
lukebrawleysmith: This logic is specific to the capValue CTE, and getFactMetricPercentileCapRows is called elsewhere.   Open to discussing, but I think it's ok in current location.

packages/shared/src/experiments/experiments.ts
```diff
@@ -494,30 +518,110 @@ export function isRegressionAdjusted(
...
+/** Any upper or lower tail capping is active (SQL / experiment analysis). */
+export function hasActiveCappingTails(metric: ExperimentMetricDefinition) {
+  return (
+    getCappingTailState(metric.cappingSettings, getLowerCappingSettings(metric))
+      .anyCap && isCappableMetricType(metric)
+  );
+}
+
+/** Short label for tooltip / metadata when capping is enabled. */
+export function formatMetricCappingSummary(metric: ExperimentMetricDefinition) {
```
lukesonnet: nit: this doesn't belong in `shared` but can live in the same file as the only component that uses it.

packages/shared/src/experiments/experiments.ts
```diff
@@ -412,6 +413,29 @@ export function getAggregateFilters({
   return filters;
 }
 
+/**
+ * For `$$distinctUsers` with aggregate user filters, each user's contribution is 1 when their
+ * aggregated value matches the filter(s) and NULL otherwise. Used in experiment SQL so capping
+ * applies to the same per-user expression as the uncapped path.
+ */
+export function wrapValueColumnWithAggregateUserFilter(
```
lukesonnet: Most metrics you can't cap + have a user filter, but I guess we do sometimes cap user filter metrics if you pick a ratio metric and use the Unique Users column type.

That said, this method is never used anywhere except in a unit test. What is it's purpose?

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -240,6 +245,16 @@ export class FactMetricModel extends BaseClass {
       newDoc.denominator = null;
     }
 
+    // Ratio metrics support only percentile capping.
```
lukesonnet, line 251: This could be breaking if this check was UI only first, but that's ok, we just should be on the lookout for any issues.

lukesonnet (commented):
Shared FE feedback offline. One additional concern re: functionality... with ignore zeros in lower tail and upper tail, can't they end up crossing in code if they differ? Maybe we should just force ignore zeros to either be on or off for both upper and lower capping. We don't have to change forms for now, but just validate that if ignore zeros is on for one percentile cap, also ensure it's also on for the other percentile cap.

packages/back-end/src/models/FactMetricModel.ts
lukebrawleysmith: Okay, I will leave it for now, thanks for the info

packages/shared/src/experiments/experiments.ts
lukebrawleysmith: moved it to same file as the only component that uses it

packages/shared/src/experiments/experiments.ts
lukebrawleysmith: I deleted the method.  It was previous code that was used to determine a user's contribution in sql in the presence of multiple filters.

lukebrawleysmith:
Re the comment below, 
I added a validator that ensures that a metric ignores zeros in both tails or neither. 

https://github.com/growthbook/growthbook/pull/5601#pullrequestreview-4774388949

lukesonnet:
@cursoragent review

lukesonnet (approved):

PR #5607, opened 2026-04-04
Feature Revision REST endpoints (beta)

bryce-fitzsimons:
Full suite of endpoints for managing a feature through drafts/revisions.

Guiding principles:
- Favor actions over huge patches
- Strip out complexity where possible, let users opt in
- Inputs don't necessarily match model format when complexity is high
- More examples and metadata in the generated docs
- Shortcuts to generate ramp-up and simple schedules on rules/features, using things like `rampActions` and `schedule`


packages/back-end/src/api/features/postFeatureRevisionPublish.ts
```diff
@@ -0,0 +1,116 @@
...
+    comment: z.string().optional().default(""),
+    adminOverride: z.boolean().optional().default(false),
+  }),
+})(async (req) => {
+  const feature = await getFeature(req.context, req.params.id);
+  if (!feature) throw new Error("Could not find feature");
+
+  if (!req.context.permissions.canUpdateFeature(feature, {})) {
+    req.context.permissions.throwPermissionError();
+  }
```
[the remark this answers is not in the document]
bryce-fitzsimons: fixing...

packages/back-end/generated/spec.yaml
```diff
@@ -26260,6 +28214,7 @@ components:
               type: number
               minimum: 0
               maximum: 1
+              description: Rollout Percent (0–1)
```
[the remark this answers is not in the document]
bryce-fitzsimons: incorrect

packages/back-end/src/services/features.ts
```diff
@@ -2007,25 +2007,27 @@ const fromApiEnvSettingsRulesToFeatureEnvSettingsRules = (
         ...(r.scheduleRules && { scheduleRules: r.scheduleRules }),
       };
       return experimentRefRule;
-    } else if (r.type === "experiment") {
-      const values = r.values || r.value;
-      if (!values) {
-        throw new Error("Missing values");
-      }
-      const experimentRule: ExperimentRule = {
-        // missing id will be filled in by addIdsToRules
```
jdorn, line 2133: Instead of completely removing this (breaking change), should we mark it as deprecated?

jdorn (commented):
Did not review all new endpoints, but changes to existing endpoints looks good.  A few breaking changes, but should have a relatively small blast radius.

jdorn:
# Code Review Feedback (Claude)

## Critical Issues

### 1. `putFeatureRevisionRule.ts` — Missing safe-rollout immutability check

Controller `putFeatureRule` (`features.ts:2518-2565`) blocks updates to `controlValue`, `variationValue`, `safeRolloutId`, `hashAttribute`, `seed`, and `trackingKey` after a safe rollout has started:

```ts
if (hasSafeRolloutStarted) {
  // ... throws "Cannot update the following fields after a Safe Rollout has started: ..."
}
```

REST `putFeatureRevisionRule.ts:121-146` has no equivalent check — the patch handler happily applies any of those fields to a started safe rollout, silently breaking the experiment.

### 2. `postFeatureRevisionRuleAdd.ts` — Missing holdout/experiment compatibility guard

Controller `postFeatureRule` (`features.ts:1745-1781`) for experiment-ref rules:

- Throws if `experiment.status !== "draft"`, the experiment has linked changes (features / redirects / visual), or the experiment is bound to a different holdout than the feature
- Otherwise, auto-links `experiment.holdoutId = feature.holdout.id` and adds the experiment to `holdout.linkedExperiments`

REST has none of this — you can attach an experiment-ref rule to a holdout-bound feature even if the experiment is running, has linked implementations, or has the wrong holdout. State will be inconsistent.

### 3. `postFeatureRevisionRevert.ts` — `archived` field omitted from revert

Controller `postFeatureRevert` (`features.ts:1334-1343`) and `postFeatureRevertDraft` (`features.ts:1520-1522`) include `archived` in the changes object:

```ts
if (revision.archived !== undefined && revision.archived !== (feature.archived ?? false)) {
  // ...permission check...
  mergeChanges.archived = revision.archived;
}
```

REST `postFeatureRevisionRevert.ts:71-181` builds `changes` with `defaultValue`, `rules`, `environmentsEnabled`, `prerequisites`, and `metadata` only — never `archived`.

Reverting to a revision with a different archive state silently fails to restore it (in both the draft and publish strategies).

### 4. `putFeatureRevisionMetadata.ts` — Project change has no permission check

Controller `putFeature` (`features.ts:3138-3151`) requires `canPublishFeature` on both the old and new project when changing project, plus:

- `ensureProjectsExist`
- `requireProjectForFeatures` org-setting check

REST `putFeatureRevisionMetadata.ts:56-73` blindly merges `project` into `changes.metadata`. With only `canUpdateFeature(feature, {})` checked at the top, a user can:

- Move a feature into a project where they have no permission
- Move a feature out of a project to escape project-scoped restrictions
- Set a project that doesn't exist
- Drop the project on an org with `requireProjectForFeatures`

### 5. `postFeatureRevisionRevert.ts` — Draft-strategy `changes` is sparse, not full state

Controller `postFeatureRevertDraft` (`features.ts:1502-1525`) populates the new draft with the full state of the target revision (all fields unconditionally if defined).

REST reuses the same sparse `changes` object built for the publish path (`postFeatureRevisionRevert.ts:186-205`). It only contains fields that differ from current state, and is missing `archived` entirely.

The resulting draft will not faithfully represent the target revision when later published.

---

## Smaller Deviations Worth a Second Look

### `putFeatureRevisionMetadata.ts` and `postFeatureRevisionRevert.ts` — `valueType` is editable

Both REST endpoints include `valueType` in the metadata patch. Controller `putFeature` does not allow `valueType` in its `metadataKeys`.

Letting a feature's value type change post-creation can corrupt rule values stored against it.

### Missing environment validation

Several REST endpoints accept `environment` from the body without checking it exists in the org's environments list. Controllers consistently validate:

```ts
environmentIds.includes(environment)
```

Affected endpoints:

- `postFeatureRevisionRuleAdd.ts`
- `postFeatureRevisionRulesReorder.ts`
- `deleteFeatureRevisionRule.ts`
- `putFeatureRevisionRule.ts`
- `putFeatureRevisionRuleRampSchedule.ts`
- `deleteFeatureRevisionRuleRampSchedule.ts`

`postFeatureRevisionToggle.ts:26-29` is the lone REST handler that does validate.

**Result:** callers can write rules into bogus or stale environment keys.

### `postFeatureRevision.ts` — `baseVersion` not validated

`postFeatureRevision.ts:33` lets the caller pass any `baseVersion`.

Controller `postFeatureCreateDraft` (`features.ts:2752-2760`):

- Always uses `feature.version`
- Asserts the live revision exists

REST will create a draft against an arbitrary or non-existent base version.

### `putFeatureRevisionHoldout.ts` — Holdout not validated

REST accepts:

```ts
{ holdout: { id, value } }
```

without checking that the holdout exists in the org.

The controller path for setting holdout (`putFeature` → `createOrUpdateDraftWithChanges` with `envelopeChanges.holdout`) goes through additional logic.

Worth confirming whether holdout-side bookkeeping (linking experiments, etc.) is being skipped.

### `postFeatureRevisionRuleAdd.ts` — Safe rollout `rampUpSchedule.enabled` hardcoded to `false`

Controller honors:

```ts
safeRolloutFields?.rampUpSchedule?.enabled
```

(`features.ts:1722`)

REST forces:

```ts
enabled: false
```

Callers cannot enable ramp-up via the REST API.

### Hardcoded `resetReview = true`

The following REST endpoints pass a literal `true` for the reset-review parameter instead of calling `resetReviewOnChange(...)`:

- `postFeatureRevisionRulesReorder.ts:91`
- `putFeatureRevisionArchive.ts:53`
- `putFeatureRevisionHoldout.ts:53`

This is stricter than the controller (which respects org settings), but inconsistent and may unexpectedly invalidate review approvals.

### `getFeatureRevisions.ts` — No `getFeature` permission check

The list endpoint queries revisions by `featureId` without first loading the feature through `getFeature(...)`.

Org-scoping prevents cross-tenant leaks, but within an org a user with no read access to the feature can still enumerate its revisions.

*(This pre-dates the PR — same in `main` — but worth flagging.)*

### `postFeatureRevisionPublish.ts` — No `publishExperimentIds` support

Controller `postFeaturePublish` (`features.ts:1149-1195`) can optionally publish related draft experiments atomically with the feature publish.

REST has no equivalent. Likely intentional, but a functional gap.

### `putFeatureRevisionPrerequisites.ts` — Stricter than controller

REST requires:

- `canUpdateFeature`
- `canManageFeatureDrafts`

Controller endpoints (`postPrerequisite` / `putPrerequisite` / `deletePrerequisite`) require only:

- `canUpdateFeature`

Stricter, not buggy — but inconsistent if the goal is to mirror controller behavior.

jdorn:
@greptileai

packages/back-end/src/api/features/postFeatureRevisionRuleAdd.ts
```diff
@@ -0,0 +1,327 @@
...
+          experiment.hasURLRedirects ||
+          experiment.hasVisualChangesets;
+        if (
+          experiment.status !== "draft" ||
+          (experiment.holdoutId &&
+            experiment.holdoutId !== feature.holdout.id) ||
+          expHasLinkedChanges
+        ) {
+          throw new BadRequestError(
+            "Failed to create experiment rule. Experiment has linked changes, is not in draft status, or is not linked to the same holdout as the feature.",
```
lukesonnet: Do we want to allow creating experiment rules for running experiments? This is currently possible in the UI and certainly reasonable. If we do allow this, then we for running experiments, we should not allow you to `updateExperiment` with a new holdout, even if the experiment didn't have a holdout.

packages/back-end/src/api/features/postFeatureRevisionRuleAdd.ts
```diff
@@ -0,0 +1,327 @@
...
+      );
+      if (!experiment) {
+        throw new NotFoundError(
+          `Could not find experiment "${ruleInput.experimentId}"`,
+        );
+      }
+
+      if (anyMissing) {
+        ruleInput.variations = ruleInput.variations.map((v, i) => ({
+          variationId: v.variationId ?? experiment.variations[i]?.id ?? "",
```
lukesonnet: This feels pretty hacky. IMO it should be all or none on the variation Id front. If any, then validate all; if none, then just use indexes and check length.

Also, can you use `getLatestPhaseVariations` instead of experiment.variations?

packages/back-end/src/api/features/postFeatureRevisionRuleAdd.ts
bryce-fitzsimons: Not sure I understand the suggestion; this error blocks for unclean experieriments on holdout-bound features. Do we want to relax this constraint?

bryce-fitzsimons: chatted briefly offline; I believe the current behavior in the PR is already correct. let me know if not though

bryce-fitzsimons:
@greptileai

packages/back-end/src/services/features.ts
```diff
@@ -2161,25 +2161,31 @@ const fromApiEnvSettingsRulesToFeatureEnvSettingsRules = (
...
-      })),
-      enabled: r.enabled != null ? r.enabled : true,
-      ...(r.prerequisites && { prerequisites: r.prerequisites }),
-      ...(r.scheduleRules && { scheduleRules: r.scheduleRules }),
-    };
-    return rolloutRule;
+    logger.error(
+      { ruleType: (r as { type?: string }).type },
+      "Unrecognized feature rule type; skipping rule",
+    );
```
jdorn: Should we make this a compile-time error if we are missing a type?  For example, looks like we don't support safe rollouts today.  If we add that later, we need to remember to update this code as well.

jdorn (approved):
Added one minor comment.  Did not test, but changes are mostly incremental and look relatively safe

packages/back-end/src/services/features.ts
bryce-fitzsimons: certainly, can add compile time type guarding here

bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@greptileai please review again

packages/back-end/src/api/features/getFeatureRevisions.ts
```diff
@@ -1,39 +1,74 @@
...
+  // unreadable projects.
+  const feature = await getFeature(req.context, req.params.id);
+  if (!feature) throw new NotFoundError("Could not find feature");
+
+  const { status, author } = req.query;
+
+  const skipPagination = stringToBoolean(req.query.skipPagination?.toString());
+  if (skipPagination && !API_ALLOW_SKIP_PAGINATION) {
+    throw new Error(
+      "skipPagination is not allowed. Set API_ALLOW_SKIP_PAGINATION=true in API environment variables. Self-hosted only.",
```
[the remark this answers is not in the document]
bryce-fitzsimons: won't fix, uncommonly used endpoint, asc is not the right sort direction in nearly every case

PR #5640, opened 2026-04-09
fix(settings): restore CDN usage graph on Usage page for Orb billing

### LINKED ISSUE #5634: [Bug] Usage graph has been replaced by duplicated invoice tab
### Summary

As of Build: 4.3.0+badc206 (2026-03-25) of the web app ( also coincides with our subscribing as pro users ), the API usage graph that used to be shown was replaced with a duplicated invoice tab. This means that the invoice can now be seen under the Settings->Billing, as intended, as well as Settings->Usage. Settings->Usage used to be a CDN usage graph over time so that users could monitor their usage on a day to day basis.

#### Expected Behavior

The usage graph is important to be able to monitor CDN usage, and we were in the middle of testing out various refresh strategies to try and minimise usage. Now we have no visibility on our CDN request quotas

#### Current Behavior

What happens instead of showing the usage graph, is that another invoice is shown

NOTE: We are on a Pro subscription

aiSynergy37:
## Summary
- fix `/settings/usage` to always render the CDN usage graph component
- avoid rendering Orb customer portal on Usage, since Orb portal is already available on Billing
- add regression tests to ensure orb-billed organizations still see Cloud Usage on the Usage page

## Testing
- `corepack pnpm --filter front-end exec vitest run test/components/settings/UsagePage.test.tsx`

Closes #5634

### DIFF
--- a/packages/front-end/pages/settings/usage.tsx
+++ b/packages/front-end/pages/settings/usage.tsx
@@ -1,7 +1,6 @@
 import usePermissionsUtil from "@/hooks/usePermissionsUtils";
 import CloudUsage from "@/components/Settings/Usage/CloudUsage";
 import { useUser } from "@/services/UserContext";
-import OrbPortal from "@/enterprise/components/Billing/OrbPortal";
 
 export default function UsagePage() {
   const permissionsUtil = usePermissionsUtil();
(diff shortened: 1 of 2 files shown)


jdorn:
When an account is on a paid plan, Orb is the source of truth for usage data. Our internal graphs are not as accurate as Orb and we don't want to mislead users about how much they will be billed, so we opt to only show the Orb portal for now.

PR #5646, opened 2026-04-10
Split experiment snapshot results into per-metric documents

lukesonnet:
### Problem

Experiment snapshots hit the 16MB MongoDB BSON document limit when users have many metric slices, because all metric results are stored inline in `analyses[].results` on the snapshot document.

### Solution

Store each metric's results as a separate document in a new `experimentsnapshotmetricresults` collection, keeping the snapshot document itself slim. Results are transparently hydrated back into the original shape on read, so all existing consumers (APIs, CSV export, notifications, front-end) continue to work without changes.

### Key changes

- **New model:** `ExperimentSnapshotMetricResultModel` (uses `BaseModel` / `MakeModelClass`) stores one document per metric × dimension, with indexes on `(snapshotId, analysisIndex, metricId, dimensionName)` and `(parentMetricId, snapshotId, analysisIndex)` for efficient per-metric and per-slice queries.
- **Split/merge logic:** `splitAnalysisResultsToMetricResultRows` and `mergeMetricResultRowsToAnalysisResults` symmetrically convert between inline results and per-metric rows. Round-trip correctness is unit tested.
- **Write path:** `updateSnapshotAnalysis` splits successful results into per-metric rows and sets `resultsStoredPerMetric: true` on the analysis, storing `results: []` on the parent document.
- **Read path:** `hydrateExperimentSnapshotMetricResults` transparently reassembles full results from the side collection when `resultsStoredPerMetric` is set. Legacy snapshots with inline results continue to work as-is.
- **Snapshot cloning (reports):** Cloned snapshots set `sourceSnapshotId` to reference the source's metric-result rows (analogous to `cachedQueryUsed` on `QueryInterface`), avoiding row duplication. Legacy sources with inline results are copied forward directly.
- **New endpoint:** `GET /snapshot/:id/metric-results` supports granular queries by `analysisIndex`, `metricIds`, `dimensionNames`, and `parentMetricId` for future lazy-loading UI work.
- **Type safety:** `CreateExperimentSnapshotInput` omits `results` from analyses (defaults to `[]`), preventing accidental >16MB inserts at creation time.

### QA checklist

- [ ] Run a standard experiment snapshot refresh — verify results display correctly and snapshot saves without error
- [ ] Run a snapshot for an experiment with many metric slices (>50 metrics) — verify it succeeds where it may have previously hit BSON limits
- [ ] Check that switching between analysis settings (stats engine, difference type) shows correct results
- [ ] Create a report from an existing experiment snapshot — verify the report loads with correct results
- [ ] Verify the external API `GET /api/v1/experiments/:id/results` returns the same shape and data as before
- [ ] Delete an experiment phase — verify the snapshot and its metric-result rows are cleaned up
- [ ] Verify experiment notifications (significance changes) still fire correctly after a snapshot refresh
- [ ] Check that dashboard snapshot blocks render correctly for experiments with per-metric stored results

### DIFF
--- a/packages/shared/src/validators/queries.ts
+++ b/packages/shared/src/validators/queries.ts
@@ -28,3 +28,27 @@ export const sqlResultChunkValidator = z
     data: z.record(z.string(), z.array(z.unknown())),
   })
   .strict();
+
+export const experimentSnapshotMetricResultValidator = z
+  .object({
+    organization: z.string(),
+    dateCreated: z.date(),
+    dateUpdated: z.date(),
+    id: z.string(),
+    snapshotId: z.string(),
+    analysisIndex: z.number(),
+    metricId: z.string(),
+    parentMetricId: z.string(),
+    dimensionName: z.string(),
+    dimensionValue: z.string(),
+    srm: z.number(),
+    variations: z.array(
+      z
+        .object({
+          users: z.number(),
+          metric: z.unknown(),
+        })
+        .strict(),
+    ),
+  })
+  .strict();
--- a/packages/shared/types/experiment-snapshot.d.ts
+++ b/packages/shared/types/experiment-snapshot.d.ts
@@ -156,6 +156,8 @@ export type SnapshotTriggeredBy =
 export interface ExperimentSnapshotMetricResultInterface {
   id: string;
   organization: string;
+  dateCreated: Date;
+  dateUpdated: Date;
   snapshotId: string;
   /** Index into `snapshot.analyses` */
   analysisIndex: number;
--- a/packages/shared/types/experiment-snapshot.d.ts
+++ b/packages/shared/types/experiment-snapshot.d.ts
@@ -148,6 +148,36 @@ export type SnapshotTriggeredBy =
   | "manual-dashboard"
   | "update-dashboards";
 
+/**
+ * One metric’s stats for a single results row (dimension / slice) on a snapshot analysis.
+ * Stored in `experimentsnapshotmetricresults` so large snapshots stay under BSON limits
+ * and the API can load metrics (and slices) incrementally.
+ */
+export interface ExperimentSnapshotMetricResultInterface {
+  id: string;
+  organization: string;
+  snapshotId: string;
+  /** Index into `snapshot.analyses` */
+  analysisIndex: number;
+  /** Metric id as keyed in `SnapshotVariation.metrics` (includes slice query string when applicable). */
+  metricId: string;
+  /**
+   * Base metric id for indexing slice rows (`parseSliceMetricId(metricId).baseMetricId`).
+   * Use this to fetch all slice documents for one logical metric.
+   */
+  parentMetricId: string;
+  /** Same as `ExperimentReportResultDimension.name` (e.g. `"All"`, slice labels). */
+  dimensionName: string;
+  /** Encodes the dimension row identity/value (e.g. `All`, slice value). */
+  dimensionValue: string;
+  srm: number;
+  /** Parallel to experiment phase variations; one entry per variation. */
+  variations: {
+    users: number;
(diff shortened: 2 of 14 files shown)


packages/shared/src/validators/queries.ts
```diff
@@ -28,3 +28,27 @@ export const sqlResultChunkValidator = z
...
   })
   .strict();
+
+export const experimentSnapshotMetricResultValidator = z
+  .object({
+    organization: z.string(),
+    dateCreated: z.date(),
+    dateUpdated: z.date(),
+    id: z.string(),
+    snapshotId: z.string(),
```
jdorn, line 38: Also add `experimentId`.  Makes things like cleanup easier when deleting an experiment. Also, allows us to do proper permission checks without needing to join through the snapshots collection first.

packages/shared/types/experiment-snapshot.d.ts
```diff
@@ -148,13 +148,50 @@ export type SnapshotTriggeredBy =
   | "manual-dashboard"
   | "update-dashboards";
 
+/**
+ * One metric’s stats for a single results row (dimension / slice) on a snapshot analysis.
+ * Stored in `experimentsnapshotmetricresults` so large snapshots stay under BSON limits
+ * and the API can load metrics (and slices) incrementally.
+ */
+export interface ExperimentSnapshotMetricResultInterface {
```
jdorn, line 156: Don't define an interface here.  Just infer it from the zod schema.

```ts
export type ExperimentSnapshotMetricResultInterface = z.infer<
  typeof experimentSnapshotMetricResultValidator
>;
```

packages/shared/src/validators/queries.ts
```diff
@@ -28,3 +28,27 @@ export const sqlResultChunkValidator = z
...
+    analysisIndex: z.number(),
+    metricId: z.string(),
+    parentMetricId: z.string(),
+    dimensionName: z.string(),
+    dimensionValue: z.string(),
+    srm: z.number(),
+    variations: z.array(
+      z
+        .object({
```
jdorn, line 47: Should we have the variationId in here as well?  Thinking forward to when people can pause variations. It wasn't as big of a deal when everything was in the same document, but feels kind of wrong now.

packages/shared/src/validators/queries.ts
```diff
@@ -28,3 +28,27 @@ export const sqlResultChunkValidator = z
...
+    organization: z.string(),
+    dateCreated: z.date(),
+    dateUpdated: z.date(),
+    id: z.string(),
+    snapshotId: z.string(),
+    analysisIndex: z.number(),
+    metricId: z.string(),
+    parentMetricId: z.string(),
+    dimensionName: z.string(),
+    dimensionValue: z.string(),
```
jdorn, line 43: Is storing a separate document per `dimensionValue` too fine-grained?  Should we just do it to the `dimensionName` level (or even just `metricId` level) instead?

150 metrics * 2 dimensions * 10 values per dimension = 3,000 documents per analysis.  That feels like way too much.

packages/shared/src/validators/queries.ts
```diff
@@ -28,3 +28,27 @@ export const sqlResultChunkValidator = z
...
   .strict();
+
+export const experimentSnapshotMetricResultValidator = z
+  .object({
+    organization: z.string(),
+    dateCreated: z.date(),
+    dateUpdated: z.date(),
+    id: z.string(),
+    snapshotId: z.string(),
+    analysisIndex: z.number(),
```
jdorn, line 39: I don't love having `analysisIndex` as part of this collection, but I don't think it's avoidable right now. This basically means if they use the baseline switcher in the UI, we are going to write potentially hundreds of documents.

lukesonnet:
Closing in favor of #5651

PR #5649, opened 2026-04-13
feat(snapshots): conditionally overflow analyses to chunk collection when >12MB

nadapzy:
## Problem

On experiments with many metrics + autoSlices + dimension cuts, the snapshot result document exceeds MongoDB's 16MB BSON limit. BigQuery completes but the Mongo write fails — analysis never renders. Root cause: `metricAutoSlices` expands to ~11.7K stat cells per dimensioned snapshot, stored entirely inside `analyses[].results[].variations[].metrics{}` on a single document.

This is a per-document limit; sharding doesn't help. #5610/#5611 (snapshot reaper) surface the failure but don't prevent it.

## Approach

Conditionally spill `analyses` to a separate `snapshotanalysisoverflow` collection **only when** the serialized array exceeds 12MB (4MB headroom under the 16MB cap for other snapshot fields + BSON overhead). Mirrors the existing `SqlResultChunkModel` pattern (#4794) used for query results.

- **Under threshold (>99% of snapshots):** behavior unchanged — `analyses` stored inline, `hasOverflowAnalyses: false`.
- **Over threshold:** `analyses` JSON-serialized, split into ~4M-char chunks, stored in `snapshotanalysisoverflow` keyed by `{organization, snapshot, chunkIndex}`. Parent doc gets `analyses: []` + `hasOverflowAnalyses: true`. Read accessors transparently rehydrate.

## Changes

- New `snapshotAnalysisOverflowValidator` (shared) + `hasOverflowAnalyses?: boolean` on `ExperimentSnapshotInterface`
- New `SnapshotAnalysisOverflowModel` (BaseModel) with `replaceForSnapshotSerialized` / `getAnalysesForSnapshot(s)` / `deleteForSnapshot(s)` + Date-field revival on parse
- `ExperimentSnapshotModel`: `prepareAnalysesForWrite()` size gate used by `updateSnapshot`, `createExperimentSnapshotModel`, and the single-analysis mutate path; chunks written before flag flip; overflow cleanup on `deleteSnapshotById` + `updateSnapshotsOnPhaseDelete`
- `addOrUpdateSnapshotAnalysis` / `updateSnapshotAnalysis` converted to read-modify-write through the size gate (handles already-overflowed snapshots)
- Read accessors (`findSnapshotById`, `getLatestSnapshot`, `findSnapshotsByIds`, `getLatestSnapshotMultipleExperiments`, `findLatestRunningSnapshotByReportId`) hydrate transparently — `context` threaded through ~13 call sites
- `back-end/src/util/overflow.ts` utils + tests

## Tradeoffs / known limitations

- **Eager hydration:** oversized snapshots still return the full payload to the client. This fixes the *write* failure; payload-size UX is a separate concern.
- **RMW for single-analysis updates:** loses the atomic `$push`; acceptable since concurrent writes to the same snapshot are rare and the prior two-step push-then-set was already non-atomic.
- **Unconditional chunk cleanup in `updateSnapshot`:** cheap indexed `deleteMany` no-op for never-overflowed snapshots (no prior-read available there); the mutate path gates on prior flag.
- Scoped to `ExperimentSnapshot` only — `SafeRolloutSnapshot` shares the structure but lacks slice expansion, lower risk.

## Testing

- `pnpm type-check` (shared/back-end/front-end) ✓
- `pnpm --filter back-end test` ✓
- `pnpm --filter shared test` ✓
- New util tests: `back-end/test/util/overflow.test.ts`

### DIFF
--- a/packages/back-end/src/api/experiments/getExperimentResults.ts
+++ b/packages/back-end/src/api/experiments/getExperimentResults.ts
@@ -16,6 +16,7 @@ export const getExperimentResults = createApiRequestHandler(
   const phase = parseInt(req.query.phase ?? experiment.phases.length - 1 + "");
 
   const snapshot = await getLatestSnapshot({
+    context: req.context,
     experiment: experiment.id,
     phase,
     dimension: req.query.dimension,
--- a/packages/back-end/src/api/snapshots/getExperimentSnapshot.ts
+++ b/packages/back-end/src/api/snapshots/getExperimentSnapshot.ts
@@ -7,7 +7,7 @@ import { createApiRequestHandler } from "back-end/src/util/handler";
 export const getExperimentSnapshot = createApiRequestHandler(
   getExperimentSnapshotValidator,
 )(async (req): Promise<GetExperimentSnapshotResponse> => {
-  const snapshot = await findSnapshotById(req.context.org.id, req.params.id);
+  const snapshot = await findSnapshotById(req.context, req.params.id);
   if (!snapshot) {
     throw new Error("Snapshot not found or no permission to access");
   }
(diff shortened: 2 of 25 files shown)


jdorn:
This was fixed in #5651

PR #5651, opened 2026-04-13
Experiment Snapshot Chunking

jdorn:
Experiment snapshots can get very large, causing MongoDB issues (when size >16MB) as well as front-end performance degradation.

This PR splits the snapshot into one chunk per metric, optimizes the storage data structure, and sets us up for huge front-end performance gains (only fetching results for metrics that are visible on screen).


packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -226,16 +255,42 @@ export async function updateSnapshotsOnPhaseDelete(
...
-  organization: string;
+  context: Context;
   id: string;
   updates: Partial<ExperimentSnapshotInterface>;
-  context: Context;
 }) {
+  const organization = context.org.id;
+
+  // If analyses have results, chunk them into separate documents
+  if (updates.analyses?.some((a) => a.results?.length > 0)) {
```
lukesonnet: Doesn't this mean an update to one analysis will delete all results from a different analysis? It feels too broad to delete all chunks associated with a snapshot just if the update payload has some results.

I don't think this ever matters in practice, so maybe fine to leave it is with a comment calling out this behavior so future changes don't mistake this important logical assumption.

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -341,15 +396,89 @@ export async function updateSnapshot({
...
+
+    // Re-fetch to get all analyses, then rebuild all chunks
+    const updated = await ExperimentSnapshotModel.findOne({
+      organization,
+      id,
+    });
+    if (updated?.settings) {
+      // Populate existing chunk data for other analyses
+      await context.models.snapshotResultChunks.populateSnapshots([
+        updated as unknown as ExperimentSnapshotInterface,
```
lukesonnet: Should we use `ExperimentSnapshotDoc` as the type here?

packages/back-end/src/models/SnapshotResultChunkModel.ts
```diff
@@ -0,0 +1,183 @@
...
+    analyses: ExperimentSnapshotAnalysis[],
+    settings: ExperimentSnapshotSettings,
+  ): Promise<AnalysisMetaEntry[]> {
+    const metricOrdering = getMetricOrdering(settings);
+    const { metricChunks, analysisMeta } = encodeSnapshotResults(
+      analyses,
+      metricOrdering,
+    );
+
+    const experimentId = settings.experimentId;
```
lukesonnet: This is actually the experiment tracking key.

ahdriel:
cursor review

packages/back-end/src/models/SnapshotResultChunkModel.ts
ahdriel: Fixed! Using the proper experiment id now

ahdriel:
cursor review

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -226,58 +326,115 @@ export async function updateSnapshotsOnPhaseDelete(
...
+      context,
+      snapshotId: id,
+      experimentId: experimentSnapshot.experiment,
+      analyses: analysisUpdates,
+      settings: experimentSnapshot.settings,
+    });
+  }
+
+  if (chunkResult) {
+    deleteExistingChunksAfterUpdate = chunkResult.metricIds.length === 0;
```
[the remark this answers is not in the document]
ahdriel: Not really a bug. `hasChunkedAnalyses` just means that those are external, not that they really exist.

So it is okay to be empty, it just means we need to check the other collection to be sure.

Added a test for this case.

ahdriel (approved):

lukesonnet (approved):

PR #5652, opened 2026-04-14
Support additional projects for features and experiments

jdorn:
### Features and Changes

Currently, features and experiments have a singular `project` field. This is great for ownership and permissions - no ambiguity about who can make changes, etc.  However, there are many cases when someone wants a feature flag to be shared between multiple projects. For example, a flag used on both front-end and back-end. Our current solution is to make a 3rd project called `common` and move the flag there. Then create an SDK connection that includes both front-end and common and one that has both back-end and common. This works, but is very un-intuitive and hacky.

This PR adds a new `additionalProjects` field to both features and experiments. This does not control permissions or ownership - that remains solely with the `project` field. This new field instead controls "visibility" - what SDK payloads the feature can be in, which list pages it shows up in, which experiments it can be tied to.

The changes should be 100% backwards compatible.


vladenisov:
Hi @jdorn,

That's great, thanks! Could you please add `project` option in targeting? It will be helpful if we have different rules for each project and don't have an attribute to split them.

PR #5655, opened 2026-04-14
prompt + tool tweaks + rawFetch jwt refresh

oelshaikh:
### Summary

- **Restructured system prompt** — Replaced the monolithic concatenated string with a structured XML-tagged template for clarity and maintainability. Added explicit workflow steps, chart/dimension/unit/filter rules, and refined response style guidance.
- **Enhanced search tool** — Search now supports empty queries (browse all items), pagination via `skip`/`limit`, and returns `totalMetrics`, `totalFactTables`, and `totalMatches` counts. Blank queries sort alphabetically; keyword queries sort by relevance.
- **Removed `getCurrentConfig` and `getConfigSchema` tools** — These tools added unnecessary round-trips. The agent now receives `config` directly in `runExploration` results and the system prompt encodes config rules inline.
- **`runExploration` now returns `config` in its result** — The normalized config is included alongside `resultCsv`, enabling the agent to make follow-up modifications without extra tool calls.
- **Batch fact table lookups in `getAvailableColumns`** — Replaced sequential `getFactTable` calls with a single `Promise.all` batch, reducing latency for multi-metric queries.
- **Tuned agent parameters** — Lowered temperature from 0.3 to 0.1 for more deterministic outputs, increased `maxSteps` from 15 to 20, and added `maxConsecutiveToolErrors: 5`.
- **Automatic JWT refresh on 401 in `fetchRaw`** — When the back-end returns a 401 with `"jwt expired"`, the front-end now transparently refreshes the token and retries the request, falling back to redirect or session error if refresh fails.

### DIFF
--- a/packages/back-end/src/enterprise/services/product-analytics-agent.ts
+++ b/packages/back-end/src/enterprise/services/product-analytics-agent.ts
@@ -34,44 +34,87 @@ import { runColumnsTopValuesQuery } from "back-end/src/jobs/refreshFactTableColu
 // Constants & system prompt
 // =============================================================================
 
-const METRICS_PREVIEW_COUNT = 5;
 const MAX_RESULT_ROWS = 200;
 
-const PA_SYSTEM_INSTRUCTIONS =
-  "Answer questions about metrics, the current configuration, and result data clearly and concisely.\n" +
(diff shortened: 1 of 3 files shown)


mknowlton89 (approved):
Code looks good - pulled it down and ran it without any issue.

PR #5656, opened 2026-04-15
Commercial features list refactor

jdorn:
### Features and Changes

Build commercial feature arrays incrementally (e.g. `pro_sso = [ ...proFeatures, "sso" ]`).  Makes it easier to see which features are unique to each tier and reduces the chance for mistakes in the future (e.g. adding to Pro, but forgetting to add to Enterprise).

I manually verified the feature lists match exactly before and after this refactor.


Kevin-Chant (approved):
Nice, I've definitely had to go back and add a feature to `pro_sso` before so this should be much safer now

PR #5663, opened 2026-04-16
Add status filter to LIST experiments endpoint

### LINKED ISSUE #5662: [Feature] Add status query parameter to GET /api/v1/experiments
### Description of Feature

We’d like to be able to filter experiments by status directly via the REST API instead of fetching everything and filtering client‑side.

There is no status query parameter to filter by experiment status (draft, running, stopped), so API consumers must:
Fetch experiments in pages (e.g. 100 at a time), and
Filter by status in their own code.
This is slow and somewhat unreliable for larger orgs with many experiments.

Requested behavior

Add a status query parameter to GET /api/v1/experiments, for example:

GET /api/v1/experiments?status=running
GET /api/v1/experiments?status=stopped
GET /api/v1/experiments?status=draft

Ideally this would:

Accept one or more statuses (e.g. status=running,stopped), and
Apply the filter at the database layer (similar to existing filters) to keep responses fast for large collections.

Why this is useful

Many integrations and internal tools only care about running experiments (or only stopped ones for reporting).
Today they must fetch all experiments and filter client‑side, which:
Increases latency and memory usage,
Complicates pagination logic, and
Is fragile for large orgs.
A server‑side status filter would align with existing search capabilities in the UI (where status is already a first‑class search field).

### Related Issues

https://github.com/growthbook/growthbook/issues/5422
https://github.com/growthbook/growthbook/issues/5421

Kevin-Chant:
### Features and Changes

Adds an optional `status` query param to the list experiments endpoint to filter by experiment status

- Closes #5662

### Testing

Hit the new endpoint with and without the param to confirm the desired behavior

### Screenshots

image

image

image

image

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -2360,6 +2360,7 @@ paths:
         - $ref: '#/components/parameters/projectId'
         - $ref: '#/components/parameters/datasourceId'
         - $ref: '#/components/parameters/experimentId'
+        - $ref: '#/components/parameters/status'
       responses:
         '200':
           content:
@@ -16710,7 +16711,20 @@ paths:
         - $ref: '#/components/parameters/limit'
         - $ref: '#/components/parameters/offset'
         - $ref: '#/components/parameters/featureId'
-        - $ref: '#/components/parameters/status'
+        - name: status
+          in: query
+          description: Filter by schedule status
+          schema:
+            description: Filter by schedule status
+            type: string
+            enum:
+              - pending
+              - ready
+              - running
+              - paused
+              - pending-approval
+              - completed
+              - rolled-back
       responses:
         '200':
           content:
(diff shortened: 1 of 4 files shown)


jdorn (approved):

packages/back-end/generated/spec.yaml
```diff
@@ -2360,6 +2360,7 @@ paths:
         - $ref: '#/components/parameters/projectId'
         - $ref: '#/components/parameters/datasourceId'
         - $ref: '#/components/parameters/experimentId'
+        - $ref: '#/components/parameters/status'
```
jdorn, line 2363: We could make this logic a little smarter.  Right now, the first `status` param the generator sees gets pulled into `components/parameters` and all subsequent ones that have different implementations remain inline.  We might want to do a first pass through all endpoints and count instances of different params.  Then, ignore ones that only appear once, and for params with multiple possible implementations, pick the one with the most references.  We could even keep multiples references around and just name it `status` and `status2` or something.

PR #5667, opened 2026-04-16
Fix experiment timeline sort order regression

jdorn:
### Features and Changes

- Updated `useExperimentSearch` in `packages/front-end/services/experiments.ts`:
  - `localStorageKey` is now **required** (no default shared key)
- Updated every `useExperimentSearch` call site to pass a **unique per-page/list key**:
  - `experiments-page`
  - `timeline-page`
  - `learnings-page`
  - `bandits-page`
  - `exec-report-completed-experiments`
  - `exec-report-all-experiments`
  - `experimentsNeedingAttention` (already unique, unchanged)
  - `ExperimentList` now requires `localStorageKey` prop; caller updated with `exec-report-running-experiments`

This prevents sort preference leakage between pages while preserving each page’s own persisted sort behavior.

### Testing

- `pnpm --filter front-end type-check`
- Manual validation in running app:
  - set `/experiments` sort to alphabetical by name
  - navigate to `/timeline` (without hard refresh)
  - confirm timeline order remains independent and not forced to match experiments sorting


jdorn:
@cursoragent revert all changes to `timeline.tsx`. A better fix is to just add the sort logic to the existing useMemo hook in `ExperimentTimeline.tsx`:

```ts
const filteredExperiments = useMemo(() => {
  // Sort in here before returning
```

jdorn:
@cursoragent looking at this deeper, the root cause is that we share the same sort field/dir across all of the places in the app where we show a list of experiments and this is persisted in localStorage.

I think a better fix is to revert all of the changes in this PR and instead in the `useExperimentSearch` hook, make the `localStorageKey` prop required instead of optional.  Every place we use the hook, pass in a unique string for that particular page.

ahdriel (approved):

PR #5670, opened 2026-04-16
Adds tracking for experiment starts

mknowlton89:
### Features and Changes

When an experiment transitions from any non-running status to `running`, the back-end now fires a fire-and-forget notification to the license server (`POST /events/experiment-started`) with the org's license, experiment ID, and timestamp. This allows the license server to track experiment starts for usage/billing purposes.

The notification is skipped for air-gapped license keys (which have no outbound license server connectivity by design). Errors are logged but do not block the experiment update.

### Dependencies

None

### Testing

- [x] Start an experiment (transition status to `running`) and verify a `POST /events/experiment-started` request is made to the license server with the correct `organizationId`, `experimentId`, and `timestamp`
- [x] Do the above for an org that has the `licenseKey` stored on the organization's object in Mongo, and as an env variable
- [x] Confirm an experiment that was already `running` and gets updated does **not** fire a duplicate notification
- [x] Confirm that an org with an air-gapped license key does **not** trigger a notification when an experiment starts
- [x] Confirm that a license server error does not break the experiment update (error should be logged, not thrown)

### Screenshots

N/A — no UI changes


packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -1876,6 +1880,20 @@ const onExperimentUpdate = async ({
       experiment: newExperiment,
       organization: context.org,
     });
+
+  if (
+    oldExperiment.status !== "running" &&
+    newExperiment.status === "running" &&
+    !isAirGappedLicenseKey(context.org.licenseKey)
```
jdorn: This line makes it seem like this PR is intended to work for self-hosted connected licenses, but I don't think it actually does.  The function uses `CLOUD_SECRET` and self-hosted stores license keys usually in an env var.

I believe this is the proper way to get the license for an org:

```ts
const license = getLicense(org.licenseKey)
```

And instead of CLOUD_SECRET and org id, we should just pass the license key to the license server.  That's enough to identify the proper org and is secure for both self-hosted and cloud.

mknowlton89:
@greptile review

mknowlton89:
@greptile review

tzjames (approved):

PR #5673, opened 2026-04-17
Add feature revision webhooks

bryce-fitzsimons:
- Family of feature.revision related webhooks
- Ability to add wildcards (up to 2 levels deep) to event webhooks rather than specifying one by one
- UI cleanups
- import cleanups

image


bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@greptileai

bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@cursor review

packages/back-end/src/services/featureRevisionEvents.ts
```diff
@@ -0,0 +1,202 @@
...
+    const tags = feature.tags ?? [];
+    // Environment filter precedence:
+    //   1. Caller-provided (e.g. specific env(s) touched for revision.updated)
+    //   2. Envs declared on the revision
+    //   3. All envs configured on the feature
+    const environments =
+      opts.environments ??
+      (revision.rules && Object.keys(revision.rules).length > 0
+        ? Object.keys(revision.rules)
+        : Object.keys(feature.environmentSettings ?? {}));
```
jdorn, line 74: Should apply project scoping to environments

packages/back-end/src/services/featureRevisionEvents.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+
+// Dispatches audit + webhook for review actions (approve, request-changes, comment).
+// Shared between the legacy controller and the REST API handler so both paths
+// stay in sync when new review types are added.
+export async function dispatchRevisionReviewEvent(
+  ctx: ReqContext | ApiReqContext,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  req: { audit: (entry: any) => Promise<void> },
```
jdorn: I think we can get rid of `req` here and just call `ctx.auditLog` which is the same thing

packages/back-end/src/services/featureRevisionEvents.ts
bryce-fitzsimons: oops, will fix

gazzdingo:
# Claud review 
## Code Review: `bryce/feature-revision-webhooks`

Compared against `origin/main`, 12 commits, ~2.5k LOC across 44 files. This adds `feature.revision.*` webhook/audit events, wildcard event subscriptions (e.g. `feature.*`, `feature.revision.*`), and a rewrite of the EventWebHook UI onto the `@/ui` design system.

## Findings (ordered by severity)

### High

**1) The UI now exposes the internal `webhook.test` event to end users.**

In `packages/front-end/components/EventWebHooks/utils.tsx` the old static list is replaced by `notificationEventNames as allNotificationEventNames` imported from `shared/validators`. That shared list is derived from `notificationEvents.webhook.test`, which is explicitly marked `noDoc: true` because it's internal (it's what the "Send Test" button fires). In the new UI it appears alongside real events and can be "subscribed to," which means a user can create a webhook that only receives its own test pings and never real events. The dynamic list should filter out entries where `notificationEvents[r][e].noDoc === true`.

```ts
// packages/front-end/components/EventWebHooks/utils.tsx
import {
  NotificationEventNameOrWildcard,
  notificationEventNames as allNotificationEventNames,
} from "shared/validators";
```

**2) New post-mutation `req.audit` calls can turn a successful write into a 5xx.**

`postFeatureRequestReview`, `postFeatureDiscard`, `postFeatureRebase`, `postFeatureCreateDraft` (and their REST counterparts) previously did not call `req.audit` after the mutation — they do now, and `audit(...)` is awaited unguarded. If audit persistence fails the client sees an error even though the revision has already been mutated, and the mutation isn't rolled back. `dispatchFeatureRevisionEvent` wraps its work in `try/catch` + `logger.error`; the audit calls that directly precede it should do the same (or `void audit(...).catch(...)`).

```ts
// packages/back-end/src/controllers/features.ts (postFeatureRequestReview)
await markRevisionAsReviewRequested(
  context,
  revision,
  res.locals.eventAudit,
  comment,
);
...
await req.audit({
  event: "feature.revision.requestReview",
  ...
});

await dispatchFeatureRevisionEvent(
  context,
  feature,
  finalRevision,
  "revision.reviewRequested",
  { reviewComment: comment ?? null },
);
```

**3) `autoPublish` archive path silently drops the new `revision.published` webhook.**

In `postFeatureArchive`, when `autoPublish === true` the new draft is both created and published in one call, but the controller only calls `recordRevisionUpdate` inside the `!autoPublish` branch and never calls `dispatchFeatureRevisionEvent(..., "revision.published", ...)`. So a subscriber watching for publish events sees archive→unarchive publishes on every other endpoint except this one. Either dispatch `revision.published` in the `autoPublish` branch or document the omission.

```ts
// packages/back-end/src/controllers/features.ts (postFeatureArchive)
if (!autoPublish) {
  await recordRevisionUpdate(context, feature, draft, "archive");
}
```

### Medium

**4) `expandEventWildcards` is dead code.**

Declared in `packages/shared/src/validators/events.ts` and exported, but never imported anywhere. Either wire it up (the obvious place is webhook dispatch: right now `getAllEventWebHooksForEvent` does its own `$in` expansion via `getWildcardPatternsForEvent`, and the Slack integration filter does its own loop — these could share `expandEventWildcards`) or delete it to avoid confusion.

**5) Import ordering violation in `event-webhooks.router.ts`.**

A `const eventNameOrWildcard = …` is declared between two blocks of `import` statements. This will fail `import/first` (ESLint) in most Next/Node configurations; it's also a readability issue. Move the `const` below the import block.

```ts
// packages/back-end/src/routers/event-webhooks/event-webhooks.router.ts
import express from "express";
import { z } from "zod";
import {
  zodNotificationEventNamesEnum,
  eventWebHookMethods,
  eventWebHookPayloadTypes,
  isEventWebhookWildcard,
} from "shared/validators";

const eventNameOrWildcard = z
  .string()
  .refine(
    (val) =>
      zodNotificationEventNamesEnum.includes(val as never) ||
      isEventWebhookWildcard(val),
    { message: "Must be a valid event name or wildcard pattern" },
  );
import { wrapController } from "back-end/src/routers/wrapController";
```

**6) Slack integration modal renders wildcards without the new label formatter.**

`SlackIntegrationAddEditModal.tsx` flattens `eventWebHookEventOptions` with `.flatMap(({ options }) => options.map((opt) => ({ label: opt.label, value: opt.value })))` and passes them straight to `MultiSelectField` without a `formatOptionLabel`. That means users see a list mixing raw event names (`feature.revision.approved`) and ambiguous wildcard labels built by `buildGroupedEventOptions`: `"All feature events"`, `"revision"` (for `feature.revision.*`), `"rampSchedule"` (for `feature.rampSchedule.*`), etc. The word "revision" on its own is particularly misleading. Pass `formatWebhookEventOptionLabel` here too, or generate non-grouped label strings like `"All feature.revision.* events"`.

**7) `buildGroupedEventOptions` produces ambiguous labels for level-2 wildcards.**

For `feature.revision.*` the option is pushed with `label: level2` → `"revision"`. In the EventWebHook modal the `formatWebhookEventOptionLabel` compensates by re-deriving the label from `value`, but the raw `label` field still leaks into every other consumer (Slack modal above, the selected-value chip, test IDs, accessibility trees, and the audit `events` display). Make the raw `label` self-describing (e.g. `"All feature.revision events"` or `"feature.revision.*"`) and leave the pretty rendering to `formatOptionLabel`.

```ts
// packages/front-end/components/EventWebHooks/utils.tsx
if (events.length > 1) {
  options.push({
    id: `${level1}.${level2}.*`,
    name: `${level1}.${level2}.*`,
    label: `${level2}`,
    value: `${level1}.${level2}.*`,
    isWildcard: true,
  });
}
```

**8) Wildcard regex accepts unknown namespaces.**

`EVENT_WEBHOOK_WILDCARD_PATTERN = /^[a-z]+(\.[a-zA-Z]+)*\.\*$/` matches `xyz.*` or `foo.bar.*` which will silently never fire. Since the list of valid resources is known at compile time, consider rejecting prefixes that don't correspond to a real resource / subgroup in `notificationEvents`. Otherwise typos in integrations or the REST API are accepted and never error.

**9) `createOrUpdateDraftWithChanges` silently falls back on merge spread.**

The signature of `updateRevision` was changed to return the updated draft, and the helper now does `return updatedDraft ?? { ...existingDraft, ...merged }`. If `updateRevision` resolves to `undefined` you return a locally-merged object that may diverge from what's in Mongo (e.g. side-effect fields like `dateUpdated`, `rules[env]`). Either make `updateRevision` guaranteed to return the fresh doc (and drop the fallback), or re-fetch with `getRevision` on the `undefined` branch instead of hand-merging. The current pattern bakes in a subtle "works most of the time" behavior.

```ts
// packages/back-end/src/controllers/features.ts
const updatedDraft = await updateRevision(
  context,
  feature,
  existingDraft,
  changes,
  logEntry,
  false,
);
return updatedDraft ?? { ...existingDraft, ...merged };
```

**10) Inconsistent semantics for `endActions: null` between controller and model.**

The controller change coerces `null ?? undefined` so the model falls through to the *template* branch, but the defensive model change treats `action.endActions = null` as an empty array (no actions). Two call sites, two different outcomes for the same input. This ships in the same PR but is unrelated to webhooks; pick one semantic for `null` and document it.

```ts
// packages/back-end/src/models/FeatureModel.ts
const endActions: RampStepAction[] =
  action.endActions !== undefined
    ? Array.isArray(action.endActions)
      ? action.endActions.map(normalizeAction)
      : []
    : template?.endPatch && Object.keys(template.endPatch).length > 0
```

### Low / informational

**11) Type widening loses compile-time safety.**

`NotificationEventNameOrWildcard = NotificationEventName | string` collapses to `string`, so TypeScript no longer catches bad event names at call sites (e.g. `events: ["feautre.*"]` would type-check). Runtime Zod validation is the only backstop. Acceptable trade-off, but call it out in the CHANGELOG / docs.

**12) Inconsistent payload shapes across resources.**

Revision webhook payloads include `orgId`; `feature.created/updated/deleted` and `experiment.*` payloads do not. Webhook consumers now get an inconsistent shape depending on which event they subscribe to. Either add `orgId` across the board or drop it (it's also available on the outer event envelope).

```ts
// packages/shared/src/validators/feature-revision-notifications.ts
// Fields present on every revision event payload.
const baseRevisionEventFields = {
  orgId: z.string(),
};
```

**13) Missing unit tests per `CLAUDE.md` ("write tests for utility/helper functions"):**

- `isEventWebhookWildcard`, `EVENT_WEBHOOK_WILDCARD_PATTERN`, `getWildcardPatternsForEvent`, `expandEventWildcards` in `shared/src/validators/event-webhook.ts` / `events.ts`
- `buildGroupedEventOptions`, `formatWebhookEventOptionLabel` in `front-end/components/EventWebHooks/utils.tsx`
- Happy/error paths for `dispatchFeatureRevisionEvent`, `recordRevisionUpdate`, `dispatchRevisionReviewEvent` in `back-end/src/services/featureRevisionEvents.ts` (project/env scoping and error-swallowing are easy to mis-refactor).

The existing `EventWebhookModel.test.ts` update only covers the `$in` query shape for `feature.created → feature.*` (a single-depth case); add a multi-depth assertion (e.g. `feature.revision.discarded → ["feature.*", "feature.revision.*"]`) to lock the contract.

**14) `dispatchFeatureRevisionEvent` swallows errors with only a `logger.error`.**

If the Mongo insert into the events collection fails (bad payload shape, disk pressure, etc.) there's no surfaced failure and no metric. Consider adding a counter or at minimum tagging the log with `webhookEvent: true` so ops can alert on it.

**15) Intentional gap in `entityEvents.feature`.**

New `revision.*` audit events are added to `entityEvents.feature` but `revision.publish` and `revision.revert` deliberately are not (they reuse `feature.publish` / `feature.revert`). Consumers filtering `entityEvents.feature` for "all revision lifecycle events" will miss those. Worth a comment in `shared/src/constants.ts` explaining the intentional gap.

**16) Silent no-op for empty comments on review submission.**

`postFeatureReviewOrComment` and `postFeatureRevisionSubmitReview` silently no-op for empty `comment` on a `Comment` action (inside `dispatchRevisionReviewEvent`). That matches the intent ("don't emit an empty event"), but `submitReviewAndComments` still ran and wrote a revision log. Consider returning a 400 for empty comments instead of quietly half-executing, or at least document it in the OpenAPI description.

**17) `EventWebHookDetail` edit/create errors are no longer shown.**

The container sets `editError` / `createError` to `null` and the modal's `error` prop is effectively always `null`. Errors now surface only via the `submit()` throwing up through `Modal`, which renders a generic failure string. The old red inline banner ("Failed to create webhook: …") was strictly more informative; verify this is intentional.

```tsx
// packages/front-end/components/EventWebHooks/EventWebHookDetail/EventWebHookDetail.tsx
isModalOpen={isEditModalOpen}
onEdit={handleEdit}
onDelete={handleDelete}
onEditModalOpen={() => setIsEditModalOpen(true)}
onModalClose={() => setIsEditModalOpen(false)}
eventWebHook={eventWebHook}
editError={null}
```

## Things that look right / nice

- Routing/filter expansion (`getAllEventWebHooksForEvent`, `getSlackIntegrationsForFilters`) correctly uses `$in` with `getWildcardPatternsForEvent` and preserves the tag/project filters.
- `dispatchFeatureRevisionEvent` correctly scopes `environments` to the feature's project and uses a sensible precedence (caller → revision rules → feature settings).
- Shared `dispatchRevisionReviewEvent` keeps the internal controller and REST handler in sync for approve / request-changes / comment paths.
- The strict `featureRevisionWebhookPayload` plus `.extend({...}).strict()` per event gives a clear contract for subscribers; the payload keys the dispatcher sets (`featureId`, `version`, `status`, `orgId`) all align with the schemas.
- `EventWebhookModel.test.ts` was updated alongside the query change rather than being left stale.

gazzdingo (approved):
LGTM 🚢

PR #5677, opened 2026-04-17
Add project-level significance

### LINKED ISSUE #5575: [Feature] Allow significance level to be scoped to Projects (currently just org-wide)
Allow the Bayesian significance level to be scoped to Projects (currently only scoped at the organization level)

lukesonnet:
## Summary

Closes #5575.

Adds project-level overrides for Bayesian (`confidenceLevel`) and Frequentist (`pValueThreshold`) significance thresholds, mirroring the existing org-level defaults.`pValueCorrection` resolution is also made project-aware on the backend (no UI)

https://www.loom.com/share/7d5fcca867104911a2c4183b9190d268

## Changes

- **Shared validators**: Extended `projectSettingsValidator`, `apiProjectValidator`, and `post/putProjectBody` with `confidenceLevel` and `pValueThreshold`.
- **Backend**: Made `getConfidenceLevelsForOrg`, `getPValueThresholdForOrg`, and `getPValueCorrectionForOrg` async and project-aware via `getScopedSettings`. Updated all call sites (experiments, experiment notifications, dashboards, reports, safe rollouts, scheduled jobs) to pass the correct `projectId`. `getDefaultExperimentAnalysisSettings` now accepts an optional `pValueThreshold` override. `ProjectModel.toApiInterface` exposes the new fields.
- **Frontend hooks**: `useConfidenceLevels` / `usePValueThreshold` accept an optional `projectId` and resolve via `getScopedSettings`. `projectId` is threaded through the experiment result rendering pipeline (`Results`, `Compact/BreakDown/DateResults`, `ResultsTable`, `PercentGraph`, time series graph, tooltips, safe rollout results, dashboard blocks).
- **UI**: New reusable `ChanceToWinThresholdField` and `PValueThresholdField` components (with shared highlight/warning helpers) used in both org-level `BayesianTab`/`FrequentistTab` and new Bayesian/Frequentist tabs on `pages/project/[pid].tsx`. Percent↔decimal conversion handled on load/save; org defaults shown as helper text.
- **OpenAPI**: Regenerated `spec.yaml`.
- **Tests**: New `organizationsProjectAware.test.ts` covering project-vs-org resolution (9 tests). Updated existing significance test mocks to `mockResolvedValue`.

## Test plan

- [x] Set a custom `confidenceLevel` on a project and verify Bayesian experiments under that project use it in the UI and in computed results statuses.
- [x] Set a custom `pValueThreshold` on a project and verify Frequentist experiments.
- [x] Also test safe rollouts and custom reports.
- [x] Presentations
- [x] Metric correlations
- [x] Verify experiments/features without a project fall back to the org default.
- [x] Verify project settings round-trip through the REST API (`GET`/`POST`/`PUT /projects`).
- [x] `pnpm type-check`, `pnpm lint`, and back-end tests pass.


lukesonnet:
@greptileai

lukesonnet:
@cursor review

packages/back-end/src/services/organizations.ts
```diff
@@ -182,8 +183,24 @@ export function getContextFromReq(req: AuthRequest): ReqContext {
...
+    organization: context.org,
+    project,
+  });
+}
+
+function confidenceLevelsFromScopedSettings(
+  settings: ReturnType<typeof getScopedSettings>["settings"],
+) {
+  const ciUpper = settings.confidenceLevel.value || 0.95;
   return {
```
ahdriel, line 205: suggestion: Can we update this + useConfidenceLevels hook + useSSRPolyfills to fallback to `DEFAULT_CONFIDENCE_LEVEL` instead of `|| 0.95`?

packages/shared/src/validators/projects.ts
```diff
@@ -10,6 +10,8 @@ export const statsEnginesValidator = z.enum(statsEngines);
 
 export const projectSettingsValidator = z.object({
```
ahdriel, line 11: question: do we see a world in the future where we allow all experiment settings to be overriden at the project-level as well?

Just curious about your thoughts.

packages/front-end/ui/AnalysisResultSummary.tsx
```diff
@@ -52,21 +52,23 @@ export interface AnalysisResultSummaryProps {
   };
   differenceType: DifferenceType;
   isBandit?: boolean;
+  projectId?: string;
```
ahdriel: nit: Should we type this as `projectId: string | undefined`?

Meaning that we always need to pass in the projectId but it might be undefined when it does not belong to a project.

This way TS will complain when we forget about this prop which is very important.

packages/front-end/hooks/usePValueThreshold.ts
```diff
@@ -1,7 +1,15 @@
 import { DEFAULT_P_VALUE_THRESHOLD } from "shared/constants";
-import useOrgSettings from "./useOrgSettings";
+import { getScopedSettings } from "shared/settings";
+import { useUser } from "@/services/UserContext";
+import { useDefinitions } from "@/services/DefinitionsContext";
 
-export default function usePValueThreshold() {
-  const settings = useOrgSettings();
-  return settings.pValueThreshold || DEFAULT_P_VALUE_THRESHOLD;
+export default function usePValueThreshold(projectId?: string) {
```
ahdriel: For all of the hooks we should check the remaining usages where projectId is not being passed in.

issue: I believe for example ExperimentResultTooltipContent.tsx is not passing projectId to usePValueThreshold when it should.

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -211,6 +211,8 @@ export class ProjectModel extends BaseClass {
       dateUpdated: project.dateUpdated.toISOString(),
```
ahdriel, line 194: issue (line 168): When creating a project via API the settings field is being ignored.

We should merge the object instead of overriding (or just pass the incoming `project` object in this case).

packages/front-end/pages/project/[pid].tsx
```diff
@@ -77,18 +88,26 @@ const ProjectPage: FC = () => {
       Object.keys(settings).forEach((k) => {
         newVal[k] = settings?.[k] || newVal[k];
       });
+      if (typeof newVal.confidenceLevel === "number") {
```
Comment (ahdriel) [inline packages/front-end/pages/project/[pid].tsx:91]: question/issue: This will run any time the form or settings are updated -- so if something changes on the settings we will keep multiplying confidenceLevel by 100 every time it happens, no?

packages/back-end/src/services/organizations.ts
```diff
@@ -192,6 +209,46 @@ export function getConfidenceLevelsForOrg(context: ReqContext) {
...
+    projectId,
+  );
+  return confidenceLevelsFromScopedSettings(settings);
+}
+
+/**
+ * Resolves all significance-related settings (confidence levels, p-value
+ * threshold, p-value correction) with a single project lookup. Use this when
+ * a caller needs more than one of these values for the same project to avoid
+ * the N round trips that calling each individual helper would incur.
```
ahdriel: question: how can we make the individual helper not have to fetch the project multiple times? would be nice to not have to have this in mind when using the helpers.

packages/shared/src/validators/projects.ts
lukesonnet: All? Hmm... no I don't think so. I think for a lot of these things, experiment templates are the answer, or programmatic fetching with a certain set of settings. This one makes more sense because I don't think we want experiment level control of significance, it gets too hard to track down why something is stat sig.

packages/front-end/ui/AnalysisResultSummary.tsx
lukesonnet: Yes, sounds good, thanks.

packages/front-end/hooks/usePValueThreshold.ts
lukesonnet: I am also going to force `string | undefined` here as the type as I think this isn't really "optional", more "in project or not in project".

packages/back-end/src/models/ProjectModel.ts
lukesonnet: Nice catch... not sure why we had that pattern. That `create` method is unnecessary and we should just use the base model method. I'm just going to remove it and add unit tests for the REST API path.

packages/back-end/src/services/organizations.ts
lukesonnet: done ty

packages/back-end/src/services/organizations.ts
lukesonnet: I now use `context.getProjects()` which already cached the projects in memory instead of calling the DB, instead of using `models.projects.getById`.

packages/front-end/pages/project/[pid].tsx
Comment (lukesonnet) [reply packages/front-end/pages/project/[pid].tsx:91]: It doesn't seem to be in my testing, no. The robot says:

- The effect only depends on `settings` (and `form` reference), not form value changes. Typing in the form does **not** re-run it.
- On each run, `newVal.confidenceLevel` gets overwritten from `settings.confidenceLevel` **before** multiplication.
- `settings.confidenceLevel` is stored as a decimal (`0.95`), because save divides by 100 before sending.
- So each effect run does: `0.95 -> 95`, not `95 -> 9500`.

Comment (lukesonnet) [reply packages/front-end/pages/project/[pid].tsx:91]: Seems brittle, though.

packages/front-end/ui/AnalysisResultSummary.tsx
lukesonnet: I made this change in a bunch of places to require projectId as explicitly undefined. It caught a bunch of bugs, but raised the broader of issue of now having to wire projectId through a bunch of niche components. I need to audit all of these in a last step, but wanted to have you take a look and see if you thought it was worth addressing now or not (e.g. look at all of the front-end changes in the latest PR that are just wiring through projectID, such as in SafeRolloutResults).

ahdriel: I think we should take some time this week to look into it -- but for now we can go with the deep prop drilling even though it is not the best solution, but I think it would be better/easier to explore a solution for that separately than this work.

ahdriel: Meaning I am okay with proceeding as-is, and I agree with the concern you brought up and we should look into it later.

packages/back-end/src/services/organizations.ts
ahdriel: Nice! So we can now:
- remove this comment that is outdated
- do we still need this function? that's more of a personal question that I'll defer to you

ahdriel (approved):

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1497,17 +1497,17 @@ export async function postExperiment(
...
-    experiment.hasVisualChangesets ||
-    (experiment.linkedFeatures && experiment.linkedFeatures.length > 0);
   if (
-    // Holdout change
     data.holdoutId &&
     data.holdoutId !== experiment.holdoutId &&
     experiment.holdoutId
   ) {
-    if (experiment.status !== "draft" || experimentHasLinkedChanges) {
+    if (
```
ahdriel: question: is this change intentional for this PR or should it be a separate one?

packages/front-end/components/Share/Presentation.tsx
```diff
@@ -68,6 +71,16 @@ const Presentation = ({
   const { getExperimentMetricById, metricGroups } = useDefinitions();
   const orgSettings = useOrgSettings();
 
+  // Note: presentations render slides for many experiments (potentially across
```
ahdriel, line 74: question: but won't this break things?

ahdriel (approved):

packages/back-end/src/services/organizations.ts
lukesonnet: Remove comment, leaving it for now.

packages/back-end/src/controllers/experiments.ts
lukesonnet: Oof. That's a reversion of another PR I landed Tuesday. Will undo this.

packages/front-end/components/Share/Presentation.tsx
lukesonnet: moved to new pattern where there's a hook that builds the map of project level settings that can be quickly accessed in these components. Will test.

lukesonnet:
@greptileai Review again please

lukesonnet:
@cursor review

packages/back-end/generated/spec.yaml
```diff
@@ -7720,12 +7720,22 @@ paths:
...
                       description: Stats engine.
                       type: string
+                    confidenceLevel:
+                      description: >-
+                        Bayesian chance-to-win threshold (stored as decimal,
+                        e.g. 0.95).
+                      type: number
+                    pValueThreshold:
+                      description: Frequentist p-value threshold (e.g. 0.05).
+                      type: number
```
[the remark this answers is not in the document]
lukesonnet: Total false positive. These are also auto-generated and the zod is correct.

PR #5683, opened 2026-04-17
feat(design-system): Add first iteration of the Dialog component

msamper:
### Features and Changes

Adds a new `Dialog` component that we will eventually use to fully replace Modals. This new component is composable to address the issue we have the current `Modal` where we have a large, ever growing list of props to try and achieve different layouts and variations of the component. I've also added a `DialogLayout` pattern component to try and capture the most common modal layout so that basic modal implementations don't need to compose the right layout every time. 

I went ahead and migrated some of our simpler modals to `DialogLayout`, but there are still ~168 files using the old `Modal` that we'll want to migrate over. 

**Next Steps**

1. Add a full width (1280px) size
2. Add a Dialog pattern equivalent of `PagedModal` 
3. Migrate over remaining `Modal` usages

### Testing

- [x] Verify the medium dialog component renders properly
- [x] Verify the large dialog component renders properly
- [x] Verify errors render properly
- [x] Verify the modals migrated over to Dialog work and look as expected

### Screenshots

**Medium Dialog**
Screenshot 2026-04-23 at 11 44 11 AM
**Large Dialog**
Screenshot 2026-04-23 at 11 44 24 AM
**Error State**
Screenshot 2026-04-23 at 11 44 38 AM
**Scrollable**
Screenshot 2026-04-23 at 11 46 05 AM


packages/front-end/ui/Dialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+import { v4 as uuidv4 } from "uuid";
+import { truncateString } from "shared/util";
+import track, { TrackEventProps } from "@/services/track";
+import Button, { Color } from "./Button";
+import ErrorDisplay from "./ErrorDisplay";
+import Text from "./Text";
+
+export type Props = {
+  open: boolean;
+  header: string;
```
ahdriel: I am slightly concerned that we will go down the same path as the Modal component and have 50+ props as we already have `cta`, `ctaColor` and `ctaEnabled` -- and if we need 2 buttons/etc then it'll keep growing.

I think we should have composable Dialog API (example here: https://base-ui.com/react/components/dialog) and then we can provide wrappers for the most common cases.

But my goal is that Dialog itself would not be aware of CTAs or CTA colors for example -- it only provides a place where those can be rendered

What do you think?

packages/front-end/ui/Dialog/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+      <RadixDialog.Content
+        size={getRadixSize(size)}
+        maxWidth={getMaxWidth(size)}
+        maxHeight="85vh"
+        style={
+          {
+            display: "flex",
+            flexDirection: "column",
+            overflow: "hidden",
+            paddingTop: "32px",
```
ahdriel, line 175: content on the right side is not being aligned properly.

The body takes up more space than the header/footer making the CTA and the content not align on the right.

Image

ahdriel (changes requested):

packages/front-end/ui/Dialog/index.tsx
msamper: I was trying to handle the content shifting when a scrollbar is present, but it's not trivial since they can be either always present or an overlay with Mac dependent on OS settings. Used the correct padding for now regardless of scrollbar presence so the content will shift slightly to the left when a scrollbar is present. This slightly misaligns the content with the footer CTAs, but I couldn't figure out a good solution without over-engineering this component. Seems like the issue is present today as well with our existing modal

packages/front-end/ui/Dialog.tsx
msamper: Implemented a composable API with a separate form wrapper that can optionally be used by consumers. Also created a component for the most common type of modal so that we don't have to compose the Dialog every time. The naming and placement of the wrapper and component might not be the best so I'm open to feedback!

msamper:
@greptileai

packages/front-end/ui/Dialog/index.tsx
ahdriel: I am curious if the [Radix Scroll Area](https://www.radix-ui.com/themes/docs/components/scroll-area) is something that can help here / was built for that.

I am in agreement with you though that we should rely on the default behavior and shift content when the toolbar is present rather than the opposite.

msamper:
@greptileai

packages/front-end/ui/Dialog/index.tsx
msamper: Taking a look!

msamper: Seems to work perfectly! added

packages/front-end/ui/Dialog/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+// Header layout primitive.
+//
+// Renders a fixed-height row at the top of the dialog. Children are laid out
+// in a space-between flex row so the common pattern of
+// <Title /> <SomeAction /> just works — no HeaderAction prop required.
+// ---------------------------------------------------------------------------
+
+function Header({ children }: { children: ReactNode }) {
+  return (
+    <Flex
```
msamper, line 199: @ahdriel Curious to see if you think this should be less opinionated about the `Header` layout. I'm referring to it being a flex box with children justified `between` specifically. I think this is a reasonable restriction, but maybe there are cases where we want to allow more customizability that I just can't think of and it would make sense for them to pass in a flexbox if they want

packages/front-end/components/Dialog/FormDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+  trackOnSubmit?: boolean;
+  close: () => void;
+  children: ReactNode;
+};
+
+// FormDialog is the opinionated wrapper around the composable Dialog
+// primitives: header + scrollable body + Cancel / Save footer, optionally
+// wired to a form submit. New dialogs with one-off layouts should compose
+// <Dialog.Root> primitives directly instead of reaching for more props here.
+export default function FormDialog({
```
ahdriel: oooh I find these names quite confusing -- `DialogForm` vs `FormDialog`.

Maybe `FormDialog` becomes `DialogLayout` or something along these lines?

Oooh -- after reading Kari's comment I also like the idea of `patterns`. So it would be something like `ui/Dialog/Patterns` which would contain the simple layout with 2 actions.

packages/front-end/ui/Dialog/DialogForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return useContext(DialogFormContext);
+}
+
+type DialogFormProps = {
+  onSubmit: () => void | Promise<void>;
+  trackOnSubmit?: boolean;
+  children: ReactNode;
+};
+
+export default function DialogForm({
```
ahdriel, line 37: question (one that I also don't have an answer for at the moment): Should these 2 new files be under `components` or also under `ui/`?

They are very related to our design system, so I lean towards `ui/`, but I understand they are not the primitives. Or is the concern that adding a component that technically has no `ui` to the `ui` folder is confusing?

packages/front-end/ui/Dialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+  useState,
+} from "react";
+import { v4 as uuidv4 } from "uuid";
+import track, { TrackEventProps } from "@/services/track";
+import ErrorDisplay from "./ErrorDisplay";
+import Text from "./Text";
+
+export type Size = "md" | "lg";
+
+export function getRadixSize(size: Size): Responsive<"3" | "4"> {
```
ahdriel: nit: why is this being exported?

packages/front-end/ui/Dialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+// Context shared between primitives.
+//
+// Dialog.Root owns error + tracking state and exposes it here so that
+// Dialog.Body can render an error automatically, and the DialogForm wrapper
+// (in components/Dialog) can report submit outcomes without the consumer
+// wiring anything up.
+// ---------------------------------------------------------------------------
+
+type DialogContextValue = {
+  size: Size;
```
ahdriel: nit: Does the `size` need to be in the `Context` of the Dialog?

packages/front-end/ui/Dialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+    ],
+  );
+
+  useEffect(() => {
+    if (open) {
+      sendTrackingEvent("modal-open");
+    } else {
+      setError(null);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ahdriel: issue: If the trackingEvent props change this will still have the old `sendTrackingEvent`.

maybe we need to also consider the previous state and trigger these events only when it changes from false to true and vice-versa?

packages/front-end/ui/Dialog/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+  );
+}
+
+// ---------------------------------------------------------------------------
+// Header layout primitive.
+//
+// Renders a fixed-height row at the top of the dialog. Children are laid out
+// in a space-between flex row so the common pattern of
+// <Title /> <SomeAction /> just works — no HeaderAction prop required.
+// ---------------------------------------------------------------------------
```
ahdriel, line 195: nit: We should remove or update this outdated comment -- HeaderAction does not exist anymore

packages/front-end/ui/Dialog/index.tsx
ahdriel: Worst case the user can render a Flex width 100% and then do whatever other layout they want inside of it.

I think this is a good default.

packages/front-end/ui/Dialog/index.tsx
```diff
@@ -0,0 +1,300 @@
...
+      {children}
+    </RadixDialog.Title>
+  );
+}
+
+function Description({ children }: { children: ReactNode | string }) {
+  return (
+    <Box flexShrink="0" pr="7">
+      <RadixDialog.Description size="2" mb="0">
+        <Text color="text-mid" size="large">
```
ahdriel: issue: I believe the browser will complain if we add a div inside of `Text` because it is rendered as a `span` by default.

Wdyt about making this a required `string` and not accept `ReactNode`?

packages/front-end/ui/Dialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+// so DialogForm consumers get error handling for free.
+// ---------------------------------------------------------------------------
+
+function Body({ children }: { children: ReactNode }) {
+  const { bodyRef, error } = useDialogContext();
+  return (
+    <ScrollArea type="auto" mt="4" mb="3" ref={bodyRef}>
+      <Box overflowX="hidden" pr="7">
+        {error && <ErrorDisplay error={error} mb="5" />}
+        {/* Left padding is used to prevent the focused outline on fields from being cut off */}
```
ahdriel: is this true? when would they get cut off? is it related to a `overflow: hidden` somewhere?

packages/front-end/ui/Dialog/DialogForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+      if (trackOnSubmit) {
+        sendTrackingEvent("modal-submit-success");
+      }
+    } catch (err) {
+      setError(err.message);
+      scrollBodyToTop();
+      setLoading(false);
+      if (trackOnSubmit) {
+        sendTrackingEvent("modal-submit-error", {
+          error: truncateString(err.message, 32),
```
ahdriel, line 63: nit/question: why are we truncating this at 32 characters?

packages/front-end/ui/Dialog.tsx
msamper: I would love your help in figuring this out haha. It is true and I'm happy to send you a loom to show you how it looks without the padding. 
I think it's because the focus state outline takes up an additional 1 or 2 px. If I remove the overflow hidden on the x axis it adds a horizontal scroll

packages/front-end/ui/Dialog/index.tsx
msamper: Makes sense to me! I don't see when the description would be anything aside from text anyway

packages/front-end/components/Dialog/FormDialog.tsx
msamper: Haha yes I agree. I like the patterns concept! I think it will make it more clear

packages/front-end/ui/Dialog/DialogForm.tsx
msamper: My concern was mainly that its not a primitive and its just composing using the Dialog API and adding form logic with the form wrapper. Our components folder is a mess, but it felt to me like that was the logical place for it to live. I like the patterns idea you proposed in the other comment though

packages/front-end/ui/Dialog.tsx
msamper: Good call! It doesn't. We just need it on the `Dialog.Root`

packages/front-end/ui/Dialog/DialogForm.tsx
msamper: This is copied over from the `Modal` component so I'm not sure. Maybe it's a limitation with the event tracker?

packages/front-end/ui/Dialog/index.tsx
msamper: Actually... we may want to be able to render doc links like we do with `EditDescriptionModal`. I'm letting the description accept `ReactNode` but I'm rendering the `Text` as a div to avoid any browser issues

packages/front-end/ui/Dialog.tsx
msamper: Managed to fix without shifting the content

ahdriel (approved):

PR #5684, opened 2026-04-17
fix(shared): Ensure import from the package and not src files

ahdriel:
### Features and Changes

When importing from `shared` we want to use `shared` package that is compiled and not the `src` files.

This was also creating a bug where some files were being imported twice and could create issues in case of a Zod validator singleton.

#### Changes

- Add `eslint` rule to prevent import from `shared/src`
- Updated all usages from `shared/src` to `shared/{file}`

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -225,6 +225,11 @@ export default defineConfig([
               group: ["*back-end*", "**/sdk-{js,react}*"],
               message: "front-end can only import from shared or itself.",
             },
+            {
+              group: ["shared/src", "shared/src/*", "shared/src/**"],
+              message:
+                "Import from the package (e.g., 'shared/validators') instead of 'shared/src/...'",
+            },
           ],
         },
       ],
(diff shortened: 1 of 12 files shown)


eslint.config.mjs
```diff
@@ -407,6 +417,11 @@ export default defineConfig([
               group: ["*back-end*", "*front-end*"],
               message: "shared cannot import from back-end or front-end.",
             },
+            {
+              group: ["shared/src", "shared/src/*", "shared/src/**"],
+              message:
+                "Within shared, use relative paths or import from shared without /src/",
```
tzjames, line 423: Yes I think when it is within it's own directory it needs to be relative.  imports within shared/src/validators to other files within shared/src/validators need to be relative otherwise it will be a circular import.  Though making a eslint rule for that might be annoying.

tzjames (approved):
Looks good.  We might need to make a cursor rule or something to prevent people's IDE's from making new ones, but now with the linter cursor should at least see that is an error and be able to fix it.  Thanks so much for doing this.

PR #5691, opened 2026-04-18
Move clickhouse functions over to the license server

tzjames:
### Features and Changes
This moves the Clickhouse Managed Warehouse functions over to the license server as they are cloud only functionality that self hosted don't need to use.

### Dependencies
Land the central-license-server change first.

### Testing
- [x] addCloudSDKMappingViaLicenseServer - Create a new sdk connection and see [usage.sdk_connections](https://console.clickhouse.cloud/services/fdb63ac1-65ec-48ef-b74f-df1854c7c3b6/console/database/usage/table/sdk_key_mapping) on clickhouse get a new row.
- [x] createClickhouseUserViaLicenseServer - Create the clickhouse datasource in the UI and see db get made and old events copied over.
- [X] dangerousRecreateClickhouseTablesViaLicenseServer - Manually `DROP TABLE feature_usage` in clickhouse db for the org.  Then in /admin choose the org and click drop and recreate table.  See the table appear again all filled out.
- [x] deleteClickhouseUserViaLicenseServer - Go to the datasource in the UI and click delete.  See that the user and database no longer exist in Clickhouse
- [x] migrateOverageEventsForOrgIdViaLicenseServer - manually in Mongo usage.organization_usages table lookup the organization and set the `usage.managedClickous.status` to `over`.  Send some more events.  See those events added to the [overage_events table](https://console.clickhouse.cloud/services/fdb63ac1-65ec-48ef-b74f-df1854c7c3b6/console/database/default/table/overage_events).  Sign up to pro. See the events removed from the clickhouse.overage_events table and added to the feature_usage table in the org's db.
- [x] updateMaterializedColumnsInClickhouseViaLicenseServer - on the datasource page add a keyAttribute.  See that a new column appears on the experiment_views table.

### DIFF
--- a/packages/back-end/src/services/licenseServerManagedClickhouse.ts
+++ b/packages/back-end/src/services/licenseServerManagedClickhouse.ts
@@ -9,13 +9,13 @@ import type {
 import type { Response } from "node-fetch";
 import { LICENSE_SERVER_URL } from "back-end/src/enterprise/licenseUtil";
 import { fetch } from "back-end/src/util/http.util";
+import { CLOUD_SECRET } from "back-end/src/util/secrets";
 
 async function postManagedClickhouse(
   path: string,
   body: unknown,
 ): Promise<Response> {
-  const cloudSecret = process.env.CLOUD_SECRET;
-  if (!cloudSecret) {
+  if (!CLOUD_SECRET) {
     throw new Error(
       "CLOUD_SECRET must be set to use license server managed ClickHouse",
     );
@@ -28,7 +28,7 @@ async function postManagedClickhouse(
     method: "POST",
     headers: {
       "Content-Type": "application/json",
-      Authorization: `Bearer ${cloudSecret}`,
+      Authorization: `Bearer ${CLOUD_SECRET}`,
     },
     body: JSON.stringify(body),
   });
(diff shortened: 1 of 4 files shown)


packages/back-end/src/services/licenseServerManagedClickhouse.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export async function createClickhouseUserViaLicenseServer(
+  orgId: string,
+  materializedColumns: MaterializedColumn[] = [],
+): Promise<DataSourceParams> {
+  const res = await postManagedClickhouse("provision", {
+    orgId,
+    materializedColumns,
+  });
+  return (await res.json()) as DataSourceParams;
```
Kevin-Chant, line 118: Should this have a zod validator parse the result to make sure both repos stay in sync?

packages/back-end/src/services/licenseServerManagedClickhouse.ts
```diff
@@ -0,0 +1,176 @@
...
+  let res: Response;
+  try {
+    res = await fetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Authorization: `Bearer ${CLOUD_SECRET}`,
+      },
+      body: JSON.stringify(body),
+    });
```
Kevin-Chant, line 61: Should we add a timeout here (per-path since some would be much longer than others?)

Kevin-Chant (approved):
Two non-blocking suggestions. Up to you if they seem worth it

packages/back-end/src/services/licenseServerManagedClickhouse.ts
tzjames: I will end up removing the code in this repo.

PR #5695, opened 2026-04-20
Return owner email from REST API

Kevin-Chant:
### Features and Changes

Adds `ownerEmail` as an optional field to models with `owner` to remove the need for an extra API call to look up the user ID. 
Creates a cached helper for mapping user IDs -> emails to enrich the data before returning

### Testing

Spot check some models to see `ownerEmail` being returned when `owner` is an ID/email and omitted when it's not

### Screenshots

image

image

image


Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

tzjames (commented):
Looks good overall.  I did notice that both `buildOwnerEmailMap` and `withOwnerEmail` are then called everywhere.  I tried out an alternative approach here: https://github.com/growthbook/growthbook/commit/53261239b829f85bf026fb182e22d8e020dfa98a
However then it is more easy to forget on new routes.  However with what you have there is nothing reminding people to pass in the map unless they look at the function signature.  I think though that you might be able to make it non-optional, and then typescript would complain if they didn't pass the map in.  So I think I would vote for that unless there are cases where we don't want to add it.  I think we pretty much always do though?

tzjames (approved):

PR #5699, opened 2026-04-20
Allow holdout to be marked as "opt-in"

lukesonnet:
## Summary

Adds optional holdout behavior via a new holdout flag, named `doNotSetAsDefaultHoldout`.

- Holdouts with `doNotSetAsDefaultHoldout: true` are no longer auto-selected by default for new experiments/features.
- Selecting `None` only shows a warning when at least one required/default holdout is available.
- `doNotSetAsDefaultHoldout` is added to the shared holdout schema, defaults to `false` on create, is persisted via holdout create/update APIs, and is editable from the holdout Overview Tab -> Implementation Section.

Toggle location
Screenshot 2026-04-20 at 10 47 04 AM

Opposite language + view with a linked experiment (just to show the UI)
Screenshot 2026-04-20 at 10 49 37 AM

When at least one is not opt-in, we get warnings + default behavior
Screenshot 2026-04-20 at 10 47 12 AM

If all holdouts are opt-in, then None is default, with no warnings
Screenshot 2026-04-20 at 10 46 47 AM

### DIFF
--- a/packages/back-end/src/models/HoldoutModel.ts
+++ b/packages/back-end/src/models/HoldoutModel.ts
@@ -19,7 +19,7 @@ const BaseClass = MakeModelClass({
   },
   globallyUniquePrimaryKeys: false,
   defaultValues: {
-    doNotSetAsDefaultHoldout: false,
+    skipAsDefaultHoldout: false,
   } as Partial<HoldoutInterface>,
   additionalIndexes: [
     {
--- a/packages/back-end/src/models/HoldoutModel.ts
+++ b/packages/back-end/src/models/HoldoutModel.ts
@@ -18,6 +18,9 @@ const BaseClass = MakeModelClass({
     deleteEvent: "holdout.delete",
   },
   globallyUniquePrimaryKeys: false,
+  defaultValues: {
+    doNotSetAsDefaultHoldout: false,
+  } as Partial<HoldoutInterface>,
   additionalIndexes: [
     {
       fields: { "nextScheduledStatusUpdate.date": 1 },
--- a/packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
@@ -5,12 +5,14 @@ import {
 } from "shared/types/experiment";
 import { VisualChangesetInterface } from "shared/types/visual-changeset";
 import { URLRedirectInterface } from "shared/types/url-redirect";
-import { useState } from "react";
+import { useMemo, useState } from "react";
 import { HoldoutInterfaceStringDates } from "shared/validators";
 import { FeatureInterface } from "shared/types/feature";
+import { Flex } from "@radix-ui/themes";
 import AddLinkedChanges from "@/components/Experiment/LinkedChanges/AddLinkedChanges";
 import LinkedChanges from "@/components/Experiment/LinkedChanges/LinkedChanges";
 import usePermissionsUtil from "@/hooks/usePermissionsUtils";
+import { useAuth } from "@/services/auth";
 import VariationsTable from "@/components/Experiment/VariationsTable";
 import TrafficAndTargeting from "@/components/Experiment/TabbedPage/TrafficAndTargeting";
 import AnalysisSettings from "@/components/Experiment/TabbedPage/AnalysisSettings";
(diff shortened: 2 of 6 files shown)


packages/back-end/src/models/HoldoutModel.ts
```diff
@@ -18,6 +18,9 @@ const BaseClass = MakeModelClass({
     deleteEvent: "holdout.delete",
   },
   globallyUniquePrimaryKeys: false,
+  defaultValues: {
+    doNotSetAsDefaultHoldout: false,
```
ahdriel: suggestion(nit): I find the double negatives of `do not: false` a bit hard to understand.

What do you think about `skipAsDefaultHoldout`?

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -99,6 +102,29 @@ export default function Implementation({
...
+      method: "PUT",
+      body: JSON.stringify({
+        doNotSetAsDefaultHoldout: !holdout.doNotSetAsDefaultHoldout,
+      }),
+    });
+    await mutate();
+  }
+
+  const holdoutDefaultStateText = useMemo(() => {
+    return holdout?.doNotSetAsDefaultHoldout
```
ahdriel: I find the UI a bit hard to understand with the clickable text. Should we use a proper toggle?

But I also understand how this is a not a main path, so happy to go as-is for now too.

ahdriel (approved):

packages/back-end/src/models/HoldoutModel.ts
lukesonnet: I agree that helps a lot, just need the default to be `false` for ease of use. Will change.

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -99,6 +102,29 @@ export default function Implementation({
...
+      ? `This holdout ${experiment.status === "draft" ? "will not be" : experiment.status === "running" ? "is not" : "was not"} selected by default for new experiments or features.`
```
lukebrawleysmith: I think the existing language should be changed, and that will help the new language.  

This holdout is a default for new experiments or features. -> New experiments and features will (not) be added to this holdout by default.  Click to change.

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -233,6 +259,25 @@ export default function Implementation({
...
+                      onClick={toggleDefaultHoldoutState}
+                    >
+                      {holdout.doNotSetAsDefaultHoldout
+                        ? "Click to make default."
+                        : "Click to make opt-in."}
+                    </button>
+                  </>
+                ) : null}
+              </Text>
+            </Flex>
```
lukebrawleysmith, line 267: Is the Overview page the best place for this decision for the user, or should it also be in Holdout creation flow?  I feel like the user should specify the Holdout they want in the creation flow, and then if they want to read a summary of their choices, they can then read the Overview page.

lukesonnet: I originally had it there, but decided against it as part of a general push to put less in the creation flow.

I think a broader orchestration/default is the right way to handle anyways.

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
lukesonnet: I think that's definitely better, I'll reconcile this with Adriel's feedback that this should be a checkbox.

lukesonnet:
@cursor review

packages/front-end/components/Holdout/HoldoutSelect.tsx
```diff
@@ -66,13 +66,28 @@ export const HoldoutSelect = ({
...
+    }
+
     // check to see if the holdout still exists and if not, set the holdout to the first valid holdout
     if (!holdoutsWithExperiment.some((h) => h.id === selectedHoldoutId)) {
-      setHoldout(holdoutsWithExperiment[0]?.id ?? "");
+      setHoldout(requiredSelectableHoldouts[0]?.id ?? "");
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [holdoutsWithExperiment]);
+  }, [holdoutsWithExperiment, requiredSelectableHoldouts]);
```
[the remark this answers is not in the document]
lukesonnet: Resolved by just removing that code. The default is `undefined` from the parent forms (unless duplicated) and so we don't need to clear it here, and clearing it was breaking duplication flows.

PR #5700, opened 2026-04-20
fix(pipeline): bootstrap incremental units table when prior state has no unitsTableFullName

nadapzy:
## Problem

In incremental Pipeline Mode, the per-experiment units table CREATE is gated on the absence of an `IncrementalRefresh` state document. Since #5466, `acquireLock()` upserts that document (with `unitsTableFullName: null`) *before* any warehouse query runs. This breaks the implicit "doc exists ⇒ table exists" assumption: if the first snapshot fails or is aborted after the lock is acquired, every subsequent "Update" sees the doc, sets `fullRefresh = false`, skips the CREATE, and the incremental UPDATE query fails reading a units table that was never created.

A related edge case: dimension-less exploratory snapshots are routed to the incremental runner in read-only mode (`fullRefresh` clamped to `false`). If no prior standard snapshot has materialized the units table, this path also reads a non-existent table.

## Fix

- Treat `incrementalRefreshModel.unitsTableFullName === null` as "needs bootstrap" when computing `fullRefresh` in `planSnapshot()`. `unitsTableFullName` is only populated after a successful CREATE, so it is the correct proxy for table existence.
- Add `hasMaterializedUnitsTable` to `getSnapshotQueryRunnerKind()` so dimension-less exploratory snapshots fall back to the `"results"` runner when the units table has not yet been created. The downstream `fullRefresh` clamp for exploratory snapshots is intentionally left in place — exploratory runs must not rebuild the units table.

## Testing

- `pnpm --filter back-end test test/services/snapshot-query-runner-kind.test.ts test/services/snapshot-planning.test.ts` (18/18 pass, including new case)

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1078,6 +1078,7 @@ export function getSnapshotQueryRunnerKind({
   experiment,
   snapshotType,
   hasSnapshotDimensions,
+  hasMaterializedUnitsTable = true,
 }: {
   allowIncrementalRefresh: boolean;
   isExperimentCompatibleWithIncrementalRefresh: boolean;
@@ -1085,6 +1086,7 @@ export function getSnapshotQueryRunnerKind({
   experiment: ExperimentInterface;
   snapshotType: SnapshotType;
   hasSnapshotDimensions: boolean;
+  hasMaterializedUnitsTable?: boolean;
 }): SnapshotQueryRunnerKind {
   if (
     allowIncrementalRefresh &&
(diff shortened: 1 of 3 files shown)


lukesonnet (approved):
Thanks for this fix!

Seems reasonable to us and the fix for the aborted main update is great! 

We're less certain that this is the right behavior for the exploratory dimension-less snapshot; we might prefer to (a) error and force you to do a full refresh and then do the exploratory analysis or (b) do the incremental update automatically and then use that as the basis for the exploratory analysis.

However, we're happy to merge as is for now, but we may revert that behavior in the future.

PR #5718, opened 2026-04-22
Fixes bug with ft explorer if ft had template variables

mknowlton89:
### Features and Changes

Fixes a bug in the fact table explorer where using a fact table whose SQL contains template variables (e.g. `{{eventName}}`) would silently fail — returning no rendered SQL and no results in the UI.

**Root cause:** `getProductAnalyticsQuery` assembled the full exploration SQL via `generateProductAnalyticsSQL` and then passed it to `compileSqlTemplate` with only `startDate`/`endDate`. Any fact-table-specific variables like `{{eventName}}` were never provided, causing `compileSqlTemplate` to throw. That error was swallowed upstream, leaving the exploration in a `running` state with no query document attached — which the UI surfaces as blank SQL and no data.

**Fix:** Before calling `generateProductAnalyticsSQL`, the fact table map is pre-processed to resolve each fact table's SQL against its own template variables (via the existing `getFactTableTemplateVariables` helper). This mirrors the pattern already used in experiment analysis queries, and works for all dataset types (`fact_table` and `metric`) since all fact tables pass through this map.

### Dependencies

None

### Testing

- [x] Create or find a fact table whose SQL uses `{{eventName}}` (e.g. a Segment-style events table filtered by event name)
- [x] Open the fact table explorer and select that fact table
- [x] Verify the query runs successfully, rendered SQL is visible, and results are returned
- [x] Verify that fact tables without template variables continue to work as before
- [x] Verify metric-type explorations that reference fact tables with `{{eventName}}` also work correctly

### Screenshots


mknowlton89:
@greptileai

jdorn (approved):

PR #5720, opened 2026-04-22
Fetch top values for all string fact-table columns

jdorn:
## Summary

Changes our fact table "top values" query to use unnest / lateral joins in all data sources instead of `UNION ALL`.  This is much more efficient when fetching top vales for multiple columns at a time and only scans the source data once.

With this change, we are now populating top values for all string columns in a fact table automatically.  This greatly improves the UI by rendering auto-complete dropdowns vs generic text input fields when applying filters.

To avoid going too crazy, we have the following limits:
- Only fetches top values for the first 50 string columns in a fact table
- Limits a single query to 25 columns
- Only fetches the top 100 values per column
- Excludes values that are longer than 100 characters

In the worst case scenario, this will run 2 queries that each return 2,500 rows and result in 500KB of data being stored in the fact table model.

## Test plan

Test refreshing a fact table column list on the major data sources. The impact radius of a bad query is fairly constrained.  It will just mean column `topValues` are not updated and we will see it in the error logs.  It will have no impact on any other queries (experiment results, etc.).  If we aren't able to test a couple of these ahead of time, it should be fine to just push live and look out for errors in production.

- [x] Postgres
- [x] BigQuery
- [x] Snowflake
- [x] ClickHouse
- [x] Databricks (had multiple AI models validate the query, but did not actually test)
- [x] Presto/Trino (had multiple AI models validate the query, but did not actually test)


jdorn:
cursor review

jdorn:
cursor review

lukesonnet (approved):

PR #5721, opened 2026-04-22
Fix schema re-registration dev issue

lillialjackson:
### Features and Changes
This change sets adds a new env var to allow for bypassing zod schema registration when not triggered by the generate-openapi script.

### Dependencies
None

### Testing

Manual testing 
### Screenshots

No UI changes


ahdriel (approved):
Thanks for the fix!

PR #5722, opened 2026-04-22
SqlIntegration refactor - part 1

jdorn:
## Background

SqlIntegration.ts is over 8,000 lines and really hard to work in, for both humans and agents.  Plus, everything is object oriented with inheritance, which makes it difficult to unit test and trace the impact of changes.

The vast majority of code is generating SQL, which is split across over 100 different methods with varying levels of inter-dependency.

## Changes

There are 2 big changes in this PR
1. Use standalone "dialect" files for each data source type instead of overriding class methods
2. Move most SQL generating methods into standalone files and make them pure/functional - no dependency on `this`

Plus, some small cleanup like removing a duplicate `formatSQL` call for the sample units CTE.

The end result is that SqlIntegration is now under 3,000 lines of code.

## Testing
- [x] Manual e2e test of experiment results
- [x] Manual e2e test of bandit results
- [x] Manual test of past experiments query (import flow)
- [ ] Manual e2e test of product analytics
- [ ] Full integration test suite across all data source types

## Future Work

This is just step 1 of the refactor.  Future work:
- Move more methods to standalone files, especially incremental refresh ones
- Remove wrapper methods in SqlIntegration and just call the SQL functions directly
- Move the dialect files to shared and use on the front-end when showing preview SQL (e.g. when creating fact metrics)
- Better organization of the SQL files to make it easier to find things
- Unit test suite on the dialect files
- Unit tests for the standalone SQL methods


jdorn:
cursor review

jdorn:
@greptileai

packages/back-end/src/integrations/dialects/base.ts
```diff
@@ -0,0 +1,136 @@
...
+  ) => `${col} ${sign} INTERVAL '${amount} ${unit}s'`,
+
+  formatDate: (col: string) => col,
+
+  formatDateTimeString: (col: string) => baseDialect.castToString(col),
+
+  selectStarLimit: (table: string, limit: number) =>
+    `SELECT * FROM ${table} LIMIT ${limit}`,
+
+  defaultSchema: "",
```
ahdriel, line 82: question: This is overriden just for Postgres, Redshift and Vertica. Is this okay?

Or is this a bug / something we should have for all dialects (or maybe we should remove from those 3 in the long run)?

On a broader aspect -- this should not live in the SqlDialect layer given this is a user configuration IMO.

packages/back-end/src/integrations/dialects/base.ts
```diff
@@ -0,0 +1,136 @@
...
+import type { DateTruncGranularity, SqlDialect } from "shared/types/sql";
+import { defaultPercentileCapSelectClause } from "back-end/src/integrations/sql/percentile-cap-select-clause";
+
+export const baseDialect: SqlDialect = {
+  escapeStringLiteral: (value: string) => value.replace(/'/g, `''`),
+
+  jsonExtract: (jsonCol: string, path: string, isNumeric: boolean) => {
+    const raw = `json_extract_scalar(${jsonCol}, '$.${path}')`;
+    return isNumeric ? baseDialect.castToFloat(raw) : raw;
+  },
```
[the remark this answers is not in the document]
ahdriel: @jdorn This is still present in this code, no?

If I override `castToFloat` in my dialect this function will ignore it.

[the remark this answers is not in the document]
jdorn: I believe I "fixed" this by just ensuring that all dialects that override `castToFloat` also now override `jsonExtract`.  Not the best long-term fix, but we rarely add new dialects so probably ok.

packages/back-end/src/integrations/dialects/base.ts
jdorn: I tried to match the behavior on `main`, but it's possible I introduced a bug here.  The scope is pretty narrow - it's only for legacy metrics that are using the query builder UI.  No fact metrics are affected and no legacy metrics that use SQL are affected.

packages/back-end/src/integrations/sql/columns/first-variation-value-per-unit.ts
```diff
@@ -0,0 +1,16 @@
+import { NULL_VARIATION_VALUE } from "shared/constants";
+import type { SqlDialect } from "shared/types/sql";
+
+export function getFirstVariationValuePerUnit(dialect: SqlDialect): string {
+  return `SUBSTRING(
+        MIN(
+          CONCAT(SUBSTRING(${dialect.castToString("e.timestamp")}, 1, 19),
```
ahdriel: On `main` this is `this.formatDateTimeString("e.timestamp")` and not `castToString`.

Should we update it to use `formatDateTimeString` as well?

packages/back-end/src/integrations/sql/fact-metrics/dimension-value-per-unit.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+  if (dimension.type === "user") {
+    return `COALESCE(MAX(${dialect.castToString(
+      `__dim_unit_${dimension.dimension.id}.value`,
+    )}),'${NULL_DIMENSION_VALUE}')`;
+  }
+  if (dimension.type === "experiment") {
+    return `SUBSTRING(
+        MIN(
+          CONCAT(SUBSTRING(${dialect.castToString("e.timestamp")}, 1, 19),
```
ahdriel: Same here -- it uses `formatDateTimeString` on main.

packages/back-end/src/integrations/sql/columns/first-variation-value-per-unit.ts
jdorn: good catch, I'm blaming that one on the robot.  I'll do a deeper audit of all of these

packages/back-end/src/integrations/dialects/base.ts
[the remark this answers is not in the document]
lukesonnet: Can we add a comment about that here so that future extensions will be more likely to see this?

lukesonnet (commented):
QA'd experiment results on Snowflake, BigQuery, and Trino. Will try and squeeze some more in today.

ahdriel (approved):

packages/back-end/src/integrations/dialects/base.ts
ahdriel: We can keep as-is, but I think this is something we should clean up eventually to make the Dialects truly independent and not tied to a datasource configuration.

lukesonnet (approved):

PR #5725, opened 2026-04-23
make tags clickable in some contexts; show tag-filtered list view

bryce-fitzsimons:
- When clicking a feature, experiment/bandit, or metric tag from a detail page, load up the filtered list view for that tab.
- When clicking a tag on a list page, just modify the search filter to search for that tag (only).

image

image


jdorn (approved):
Did not test, but code looks pretty low risk

PR #5730, opened 2026-04-23
Product Analytics - "Show As" toggle 

oelshaikh:
# Product Analytics: "Show As" toggle + unit handling fixes

## Summary

Adds a chart-level **Show As** control (`total` vs `per_unit`) to Product Analytics explorations so the same underlying metrics can be rendered as raw totals (e.g. "total events") or per-unit averages (e.g. "events per user") without editing the metric. This also fixes a bug where mean metrics without a `unit` were silently losing their denominator in SQL, which broke any per-unit rendering downstream (chart, table, and agent CSV).

## What changed

**Schema / validators** (`packages/shared`)

- New optional `showAs: "total" | "per_unit"` on `baseExplorationConfigValidator` (back-compat; read sites default to the inferred value).
- SQL layer (`sql.ts`) now always emits a denominator/unit-count column for any unit-aggregated non-ratio, non-quantile metric, so the frontend can switch between totals and per-unit on the fly.
- Regenerated `spec.yaml` to expose `showAs` on all PA endpoints.

**Shared helpers** (`packages/shared/src/enterprise/product-analytics/utils`)

- New `showAs` helpers: `inferShowAs`, `getEffectiveShowAs`, `clearInapplicableShowAs`, `showAsAppliesTo`, `getSharedUnit`.
- `getEffectiveMetricValue` now takes `{ showAs, isRatio }` — ratios always render as N/D; standard metrics respect `showAs`.
- New `buildExplorationColumns` + `getExplorationCellValue`: single source of truth for the exploration column schema (dimension columns, ratio expansion into Numerator/Denominator/Value, per-unit header labels). Used by both the Explorer result table and the agent's CSV.
- New `getMetricMixClass` + `getLockedMixClass` for mixing rules (ratio / quantile / standard).

**Explorer UI** (`packages/front-end/enterprise/components/ProductAnalytics`)

- New `ShowAsSection` sidebar control, conditionally rendered via `showAsAppliesTo` (hidden for fact_table/data_source datasets and for metric datasets where per-unit is degenerate).
- `ExplorerChart` respects the effective `showAs`, adds a Y-axis name ("Total" / "Per \<unit\>") and routes all cell/total computations through the shared helpers.
- `useExplorationTableData` reuses the shared column builder + cell extractor; display-only concerns (date formatting, rounding, "Total" fallback) stay in the hook.
- `MetricTabContent`: metric options now always show the full list and disable incompatible metrics with a tooltip (e.g. "Ratio metrics can't be combined with other metric types.") instead of silently hiding them.
- `ValueCard`: mean metrics now support unit selection.
- `ExplorerContext` normalizes configs on load and on every draft update via `fillMissingUnits` (backfills `userIdTypes[0]`) and `clearInapplicableShowAs` (strips `showAs` when it doesn't apply), so URL-loaded, AI-generated, and saved configs never disagree with what the chart renders.
- `toFetchKey` excludes `showAs` so toggling it doesn't trigger a refetch.

**AI agent** (`packages/back-end/src/enterprise/services/product-analytics-agent.ts`)

- Prompt updated with `<show_as_rules>` and tightened `<unit_rules>` (always set `unit` for standard metric types).
- `buildResultCsv` switched to the shared column schema + cell extractor, so the CSV the agent reads matches the column set, ordering, and effective `showAs` of the on-screen table/chart. Column headers now encode the mode (`"<name>"` vs `"<name> per <unit>"`).
- Snapshot summary now surfaces `showAs` transitions for follow-up turns.
- `processSSEEvent.ts`: forwards the top-level normalized config during active turns so the streaming chart matches the final persisted one.

**Misc**

- `DisplayTestQueryResults` grows an optional `columnLabels` prop so callers can pass stable keys + separate display labels (used by the exploration table to render per-unit headers without losing column identity).
- `ChatMessageList`: fix a bug where the agent's collapsed-steps expansion state could snap shut mid-turn; expansion now resets only when a new user turn begins.
- Tests updated for the default `showAs: "total"` in `DEFAULT_EXPLORE_STATE`.

## Behavior by metric type

| Metric type(s) | Unit selector | Show As toggle | Renders | Inferred default |
|---|---|---|---|---|
| `mean` | shown | **shown** | `total` → raw numerator; `per_unit` → numerator / unit count | `per_unit` when numerator aggregation is `max` or `count distinct` (totals are incoherent); otherwise `total` |
| `proportion`, `retention`, `dailyParticipation` | shown | hidden | raw numerator (per-unit would collapse to ~1 since denominator == numerator) | — |
| `ratio` | shown only when numerator is `$$distinctUsers` | hidden | always N/D (self-contained) | — |
| `quantile` | hidden | hidden | percentile value (no denominator emitted) | — |

Rules of thumb:

- **Show As toggle visibility**: requires at least one `mean` metric in the dataset. Datasets made entirely of the other types hide the toggle because per-unit is either degenerate or self-contained.
- **Mixing**: `ratio` and `quantile` can't be combined with other classes; `mean` / `proportion` / `retention` / `dailyParticipation` can mix freely with each other. Incompatible options appear disabled in the dropdown with a tooltip.
- **fact_table / data_source datasets**: Show As never applies — values render as the raw emitted number.

## Backwards compatibility

- `showAs` is optional everywhere; read sites fall back to `inferShowAs`, which only picks `per_unit` when totals would be mathematically incoherent (mean + `max` / `count distinct` aggregations). All existing saved configs render as `"total"` by default, unchanged.
- Existing configs without a `unit` on mean metrics are auto-filled at load time, restoring correct denominators without requiring a re-save.

## Testing

- Pick a `mean` metric (e.g. revenue). Show As toggle appears in the sidebar. Flip between **Event Totals** and **Per \<unit\>** — chart values, Y-axis label, and result table header (`"<name>"` vs `"<name> per <unit>"`) all update together. Toggling doesn't trigger a refetch.
- Switch to a `proportion` / `retention` / `dailyParticipation` / `ratio` / `quantile` metric — toggle is hidden.
- Add a `ratio` metric, then try to add a `mean` alongside it — incompatible options are disabled in the dropdown with a tooltip.
- Open an older mean-metric exploration that was saved without a `unit` — chart renders correctly (denominator is backfilled from `userIdTypes[0]`).
- Ask the PA AI agent "revenue per user last 7 days" then "total revenue last 7 days" — the chart flips modes and the agent's response describes the numbers matching the Y-axis label.

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -13707,6 +13707,11 @@ paths:
                                     required:
                                       - predefined
                                     additionalProperties: false
+                                  showAs:
+                                    type: string
+                                    enum:
+                                      - total
+                                      - per_unit
                                   type:
                                     type: string
                                     const: metric
@@ -13988,6 +13993,11 @@ paths:
                                     required:
                                       - predefined
                                     additionalProperties: false
+                                  showAs:
+                                    type: string
+                                    enum:
+                                      - total
+                                      - per_unit
                                   type:
                                     type: string
                                     const: fact_table
@@ -14278,6 +14288,11 @@ paths:
                                     required:
                                       - predefined
                                     additionalProperties: false
+                                  showAs:
+                                    type: string
+                                    enum:
+                                      - total
+                                      - per_unit
                                   type:
                                     type: string
                                     const: data_source
@@ -16647,6 +16662,11 @@ paths:
                   required:
                     - predefined
                   additionalProperties: false
+                showAs:
+                  type: string
+                  enum:
+                    - total
+                    - per_unit
                 type:
                   type: string
                   const: metric
@@ -16979,6 +16999,11 @@ paths:
                                 required:
                                   - predefined
                                 additionalProperties: false
+                              showAs:
+                                type: string
+                                enum:
+                                  - total
+                                  - per_unit
                               type:
                                 type: string
                                 const: metric
@@ -17280,6 +17305,11 @@ paths:
                   required:
                     - predefined
                   additionalProperties: false
+                showAs:
+                  type: string
+                  enum:
+                    - total
+                    - per_unit
                 type:
                   type: string
                   const: fact_table
@@ -17621,6 +17651,11 @@ paths:
                                 required:
                                   - predefined
                                 additionalProperties: false
+                              showAs:
+                                type: string
+                                enum:
+                                  - total
+                                  - per_unit
                               type:
                                 type: string
                                 const: fact_table
@@ -17931,6 +17966,11 @@ paths:
                   required:
                     - predefined
                   additionalProperties: false
+                showAs:
+                  type: string
+                  enum:
+                    - total
+                    - per_unit
                 type:
                   type: string
                   const: data_source
@@ -18289,6 +18329,11 @@ paths:
                                 required:
                                   - predefined
                                 additionalProperties: false
+                              showAs:
+                                type: string
+                                enum:
+                                  - total
+                                  - per_unit
                               type:
                                 type: string
                                 const: data_source
@@ -24640,6 +24685,11 @@ components:
                         required:
                           - predefined
                         additionalProperties: false
+                      showAs:
+                        type: string
+                        enum:
+                          - total
+                          - per_unit
                       type:
                         type: string
                         const: metric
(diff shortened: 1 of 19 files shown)


mknowlton89 (approved):
Looks good! Tested locally and didn’t notice any issues. 

Approving now to not block you when the opacity issue is resolved.

PR #5733, opened 2026-04-23
feat(experiments): Support feature value editing from the experiment page + misc improvements

msamper:
### Features and Changes
- Introduces `DraftSelectorDropdown`, a shared component that lets users choose where to save feature changes — publish directly, save to an existing open draft, or open a new draft. Currently wired into the **Edit Feature Flag Values** modal (accessed from the experiment page). Intended to replace `DraftSelectorForChanges` once we migrate feature modals to our new `Modal` component and do some more iterations on the design.
- Adds a dedicated `EditFeatureFlagValuesModal` for editing variation values and weights directly from the experiment page, with support for adding new variations inline
- Improves how unconfigured variation values are displayed: variations with no value defined show a "Define missing values" warning (on both the experiment page and feature page)
- Migrates the Add Feature to Experiment modal to `ModalStandard`
- Adds a pre-launch checklist item for filling in missing feature values after adding a variation

https://www.loom.com/share/7928097633e04f828e4f31ea97a5a9be

### Testing
- [x] Link a feature flag to a draft experiment and open the Edit Feature Flag Values modal — verify the draft selector dropdown shows the correct options (existing draft / new draft) and defaults to an existing draft if it contains experiment-ref rule changes
- [x] Save changes to an existing draft and verify the feature revision is updated correctly
- [x] Add a new variation from the modal and verify weights auto-rebalance and the new variation row appears
- [x] With multiple linked feature flags, add a new variation to the experiment without updating one of the feature flags — verify that feature shows "Define missing values" on both the experiment page and the feature page
- [x] Approve a scheduled experiment start and verify the Edit button on linked feature flags is hidden
- [x] Start an experiment that has a linked feature with a no-op draft (live feature > edit values > edit back to live values) no-op draft should be discarded automatically

### Screenshots
Screenshot 2026-05-29 at 10 21 29 AM
Screenshot 2026-05-29 at 10 24 36 AM
Screenshot 2026-05-20 at 2 21 58 PM
Screenshot 2026-05-20 at 2 22 09 PM
Screenshot 2026-05-20 at 2 35 45 PM
Screenshot 2026-05-20 at 2 45 17 PM


packages/back-end/src/controllers/experiments.ts
```diff
@@ -3922,13 +3922,29 @@ export async function postExperimentFeatureValues(
 
   if (variationsChanged) {
     changes.variations = variations;
+    // Also update the phase's variations array to include new/removed variations
```
lukesonnet, line 4038: Is it added to the main experiment variations array somewhere?

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
```diff
@@ -0,0 +1,601 @@
...
+
+    append({
+      id: generateVariationId(),
+      name: `Variation ${fields.length}`,
+      description: "",
+      key: String(fields.length),
+      screenshots: [],
+      weight: 0,
+      value: getDefaultVariationValue(feature.defaultValue ?? ""),
+    });
```
lukesonnet, line 306: Walk-through that produces a collision:

Start with 2 existing variations (keys "0", "1")
Add → key "2"
Add → key "3"
Remove the new variation at index 2 (key "2") → fields now ["0","1","3"]
Add → fields.length = 3 → key "3" (duplicate)

Then it errors that keys must be unique.

That's why we have to do a bunch of stuff like: https://github.com/growthbook/growthbook/blob/msamper/draft-selector-redesign/packages/front-end/components/Features/FeatureVariationsInput.tsx#L102-L105 where we do forceRenormalizeVariationKeysOnSort

Probably what's best is something like keeping track of which variations are new, and then just ensure we normalize those keys before we submit, probably to be 1, 2, 3 etc. greater than the length of `experiment.variations` and have no conflict with the keys in `experiment.variations`, or maybe it should be handled server side with "newVariation": true since we don't expose key management in this UI.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -3922,13 +3922,29 @@ export async function postExperimentFeatureValues(
 
   if (variationsChanged) {
```
lukesonnet, line 4036: This part makes me the most nervous, I'm a bit tied up the rest of the day, but wondering how thoroughly we've QAd all the variation addition/removal edge cases.

packages/front-end/components/Features/ExperimentRefSummary.tsx
```diff
@@ -266,16 +269,34 @@ export default function ExperimentRefSummary({
...
+                          <HelperText status="warning">
+                            Define missing values
+                          </HelperText>
+                        ) : type === "string" && value === "" ? (
+                          <Text
+                            color="text-low"
+                            fontStyle="italic"
+                            size="small"
+                          >
+                            Intentionally left empty
```
lukesonnet: Is this better/clearer than `SERVE ""`? That seems clearer to me, but don't have strong feelings.

packages/front-end/components/Experiment/PreLaunchChecklist.tsx
```diff
@@ -285,6 +285,35 @@ export function getChecklistItems({
             ),
           });
         });
+
+      const latestVariations = getLatestPhaseVariations(experiment);
```
lukesonnet, line 289: Does it work with blocking the Public API start endpoint? I feel like it should, but the broader issue here is with the FE doing so much of the validation for the pre-launch checklist instead of a shared service. IMO, if trivial, please resolve, if more complicated, don't worry about it for now.

packages/front-end/components/Features/DraftSelectorDropdown.tsx
```diff
@@ -0,0 +1,352 @@
...
+  DropdownMenuSeparator,
+  DropdownSubMenu,
+} from "@/ui/DropdownMenu";
+import { useFeatureRevisionsContext } from "@/contexts/FeatureRevisionsContext";
+
+export type DraftMode = "existing" | "new" | "publish";
+
+type CurrentOption = "new" | "existing" | "this" | "publish";
+
+export default function DraftSelectorDropdown({
```
lukesonnet, line 28: Has @bryce-fitzsimons seen the new selector in this PR? Should he weigh in if he hasn't?

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
```diff
@@ -0,0 +1,601 @@
...
+          gatedEnvSet={gatedEnvSet}
+          locked={ruleOnlyOnDraft}
+          lockedTooltip={
+            ruleOnlyOnDraft
+              ? "This experiment rule is added in this draft revision. Changes will be saved to it."
+              : undefined
+          }
+        />
+      }
+      submit={form.handleSubmit(async (values) => {
```
lukesonnet, line 344: It seems I can submit even with no changes. Is that our usual pattern?

lukesonnet (commented):
QAd a bit and it mostly worked well. Some nits and broader concerns left inline.

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
ahdriel: It is not -- but I would vote to make it.

I've had the experience where the submit button is disabled but I don't see any errors in the form or I need to dig to find where it is. With the submit it would immediately error out and show where the failure is (empty field or something).

packages/front-end/components/Features/DraftSelectorDropdown.tsx
msamper: Good idea! Shared with Bryce

packages/back-end/src/controllers/experiments.ts
msamper: Yup right above we add the `variations` to `changes.variations` and we persist further down in the code with a call to `updateExperimentAndSync`

packages/front-end/components/Features/ExperimentRefSummary.tsx
msamper: mmm maybe not. Maybe now that we have `Define missing values` this state isn't particularly useful

msamper: Removed

packages/front-end/components/Experiment/PreLaunchChecklist.tsx
msamper: It didn't, but I added a checklist item to `getExperimentStartChecklistStatus` to mirror the front-end checklist items. Agree that we should work on having a shared service for pre-launch checklist because right now there's a disconnect between the checklist we use for REST api vs the one we have in app on the UI

packages/front-end/components/Features/DraftSelectorDropdown.tsx
msamper: Removed the ability to publish from the edit flag values modal for now per Bryce's recommendation to let starting the experiment handle the auto-publishing once a revision is approved. Bryce also shared some feedback that the design for this dropdown may not work well for larger orgs that have 10+ revisions so we'll probably want to revisit the design with Kari. Going to only use the dropdown on the experiments page for now

packages/back-end/src/controllers/experiments.ts
msamper: Have played around some more with adding/removing variations when there are custom keys, when we have more than one implementation, when we're in a different phase, and haven't run into any issues. Of course, would appreciate your help doing some more QA if possible

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
msamper: Should be fixed. Sending the new variations to the back-end without a key and letting the back-end be responsible for assigning the keys via a new helper before we validate the variations

packages/front-end/components/Features/ExperimentRefSummary.tsx
lukesonnet: >Define missing values this state isn't particularly useful

Yeah that's what I was thinking

packages/shared/types/experiment.d.ts
```diff
@@ -270,6 +270,14 @@ export interface LinkedFeatureInfo {
   inconsistentValues: boolean;
   rulesAbove: boolean;
   environmentStates: Record<string, LinkedFeatureEnvState>;
+  /**
+   * True when the live revision has at least one experiment-ref rule for this
+   * experiment. Used by the edit-values modal to gate "Apply now": when this
```
lukesonnet: I think this comment is incorrect, it's just used as flavor text now.

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
```diff
@@ -0,0 +1,589 @@
...
+    if (!reviewSetting?.requireReviewOn) return "none";
+    const envList = reviewSetting.environments ?? [];
+    return envList.length === 0 ? "all" : new Set(envList);
+  }, [settings?.requireReviews, feature]);
+
+  const defaultDraft = useDefaultDraft(revisionList);
+
+  // The linking flow always creates a pending draft that adds the
+  // experiment-ref rule, so live doesn't have the rule yet. In that state,
+  // "Apply now" (publish to live) would fail on the back-end because there's
```
lukesonnet: same here, right? We set `canAutoPublish` to false below, so none of this stuff really applies anymore.

packages/front-end/components/Experiment/LinkedChanges/LinkedFeatureFlag.tsx
```diff
@@ -10,43 +10,53 @@ import LinkedChange from "@/components/Experiment/LinkedChanges/LinkedChange";
...
 type Props = {
   info: LinkedFeatureInfo;
   experiment: ExperimentInterfaceStringDates;
+  numLinkedChanges: number;
   onReAdd?: () => void;
   mutate?: () => void;
 };
 
 export default function LinkedFeatureFlag({
   info,
```
lukesonnet, line 34: `info` confused me throughout this process, since really its the linked feature flag with revision and is super critical to the draft we select for revision and all kind of stuff. Just a slightly better name `linkedFlagRevision` or something, whatever the robot suggests.

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
```diff
@@ -0,0 +1,589 @@
...
+    const raw = settings?.requireReviews;
+    if (raw === true) return "all";
+    if (!Array.isArray(raw)) return "none";
+    const reviewSetting = getReviewSetting(raw, feature);
+    if (!reviewSetting?.requireReviewOn) return "none";
+    const envList = reviewSetting.environments ?? [];
+    return envList.length === 0 ? "all" : new Set(envList);
+  }, [settings?.requireReviews, feature]);
+
+  const defaultDraft = useDefaultDraft(revisionList);
```
lukesonnet:
```suggestion
  // used as fallback in case somehow existing draft isn't passed in
  const defaultDraft = useDefaultDraft(revisionList);
```

IDK something like that, the naming threw me off

lukesonnet (approved):
Approving, but as discussed please shift the behavior to a "new" draft if the revision is live and there's no more recent revisions that just touch the experiment.

packages/shared/types/experiment.d.ts
msamper: updated

packages/front-end/components/Experiment/LinkedChanges/EditFeatureFlagValuesModal.tsx
msamper: Updated

packages/front-end/components/Experiment/LinkedChanges/LinkedFeatureFlag.tsx
msamper: Renamed to linkedFeatureInfo per claude's suggestion

msamper:
@greptileai

PR #5738, opened 2026-04-24
Adds copy link to explorations

lillialjackson:
### Features and Changes

This change set adds functionality to allow users to copy the link of the current exploration view to share with others in the same org.

### Dependencies

None

### Testing

Manual 

### Screenshots

Screenshot 2026-04-24 at 11 29 33 AM

Screenshot 2026-04-24 at 11 29 40 AM

### DIFF
--- a/packages/front-end/enterprise/components/ProductAnalytics/SideBar/ExplorerSideBar.tsx
+++ b/packages/front-end/enterprise/components/ProductAnalytics/SideBar/ExplorerSideBar.tsx
@@ -5,7 +5,8 @@ import {
   FactTableValue,
   ExplorationConfig,
 } from "shared/validators";
-import { PiArrowsClockwise } from "react-icons/pi";
+import { PiArrowsClockwise, PiLink } from "react-icons/pi";
+import ShareUrlPopover from "@/ui/ShareUrlPopover";
 import PaidFeatureBadge from "@/components/GetStarted/PaidFeatureBadge";
 import Text from "@/ui/Text";
 import SelectField from "@/components/Forms/SelectField";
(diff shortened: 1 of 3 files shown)


packages/front-end/ui/ShareUrlPopover.tsx
```diff
@@ -0,0 +1,73 @@
...
+      side={side}
+      align={align}
+      showCloseButton
+      showArrow={false}
+      contentStyle={{ padding: 0 }}
+      contentClassName={styles.popoverContent}
+      content={
+        <Box className={styles.inner}>
+          <Flex direction="column" gap="1" mb="5">
+            <Text
```
mknowlton89: We have a `Heading` component that we've built that is a small wrapper around Radix's `Heading` - I think it would make sense to use that here - it has some helpers built in around colors and sizing to make future changes easier.

packages/front-end/ui/ShareUrlPopover.tsx
```diff
@@ -0,0 +1,73 @@
...
+              size="5"
+              weight="bold"
+              style={{ color: "var(--color-text-high)" }}
+            >
+              {title}
+            </Text>
+            <Text size="3" style={{ color: "var(--color-text-mid)" }}>
+              {description}
+            </Text>
```
mknowlton89: And much like our custom `Heading` component, we have a custom `Text` component that wraps Radix's `Text` component that we should use here - it's got helpers similar to the `Heading` component.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/ExplorerSideBar.tsx
```diff
@@ -103,33 +104,48 @@ export default function ExplorerSideBar({
...
+            </Tooltip>
+            <ShareUrlPopover
+              trigger={
+                <Button
+                  size="sm"
+                  variant="solid"
+                  color="violet"
+                  aria-label="Share exploration link"
+                >
+                  <PiLink />
```
mknowlton89: I wonder if we want to use the `<PiLinkBold />` here - I see in the app whenever we're using the regular `PiBold` its within a button that has an icon + text. The non-bold option looks a little light compared to the mock.

packages/front-end/ui/ShareUrlPopover.tsx
```diff
@@ -0,0 +1,73 @@
...
+  description?: string;
+  side?: "top" | "right" | "bottom" | "left";
+  align?: "start" | "center" | "end";
+}
+
+export default function ShareUrlPopover({
+  trigger,
+  url,
+  title = "Share this exploration",
+  description = "Anyone in your organization with this link can open this exploration with the exact configuration applied.",
```
mknowlton89: I'm wondering if we want to make title required (and not have a default), and then keep description optional, but get rid of the default.

I feel like this component can be used elsewhere, and I'd hate for someone to accidentally not include a description and then the popover includes references to an exploration. 

That way this can be used for all sorts of resources, not just the exploration.

mknowlton89: Also, the description is slightly misleading as it stands.

I think we should update to: `Anyone in your organization with read access to the Data Source this exploration uses, can open this exploration.`

When it comes to permissions, users have global permissions, and then project-specific permissions. And we have a concept of a `No Access` role for enterprise organizations (more docs on that [here](https://docs.growthbook.io/account/user-permissions#how-does-the-no-access-role-work)).

We do some permission checking so users can only see some data sources in the data source list in the top-right corner of the exploration tools. So if a user doesn't have read access to a data source and they get a link to an exploration built on a metric that uses the data source, they'll get a `Data Source not found` error. (We should probably try to improve this experience, but that is outside of the scope of this project).

mknowlton89 (commented):
Looks awesome! Left a few pointed comments. 

The experience matches the mocks really well - that being said, the font sizes did feel pretty large compared to the rest of the page. It seems like everything was 2px bigger than all the other similar text on the page. I'd suggest running it by Johnny.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/ExplorerSideBar.tsx
```diff
@@ -103,33 +104,51 @@ export default function ExplorerSideBar({
...
+              trigger={
+                <IconButton
+                  size="2"
+                  variant="solid"
+                  color="violet"
+                  aria-label="Share exploration link"
+                  style={{ height: 32, width: 32 }}
+                >
+                  <PiLink size={20} />
+                </IconButton>
```
mknowlton89, line 146: Small nit: I think this button should be disabled until the user has made a change. I think we have this logic in place already via `isSubmittable`.

mknowlton89 (approved):
Looks good! Left one small comment about a disabled state.

I still think the sizing is too large, but I'll defer to Johnny for the final call, so going ahead and approving.

PR #5739, opened 2026-04-24
chore(design-system): Migrate simple modals to ModalStandard (part 1)

msamper:
### Features and Changes

Migrates over several usages of the `Modal` component to our new `ModalStandard` pattern


msamper:
@cursor review

msamper:
@cursor review

msamper:
@cursor review

msamper:
@cursor review

ahdriel (approved):
Looking good! Some points for quick follow-ups that I think we should handle in a separate PR:

1. The spacing between the body and the footer looks a bit too big when we have a single line.

Image

2. Similar to the 1px cutoff on the sides, we also have 1 at the bottom / top of the content -- we should investigate and see how we can address this

Image

3. The CTA in the footer is still not right aligned with the content. They are very close but there's some px of difference. Would be nice to get that fixed.

msamper:
> Looking good! Some points for quick follow-ups that I think we should handle in a separate PR:
> 
> 1. The spacing between the body and the footer looks a bit too big when we have a single line.
> 
> Image
> 2. Similar to the 1px cutoff on the sides, we also have 1 at the bottom / top of the content -- we should investigate and see how we can address this
> 
> Image
> 3. The CTA in the footer is still not right aligned with the content. They are very close but there's some px of difference. Would be nice to get that fixed.

Good catch! For (1) it's just because the text in the `DeleteButton` modal is being rendered with a margin-bottom since it's rendered as a `<p>...</p>` I'll fix that. Here's how it looks without that margin:
Screenshot 2026-05-04 at 2 07 02 PM

Happy to investigate (2) and (3) in a separate PR

PR #5744, opened 2026-04-24
added validation for attributes to make sure they are valid before sa…

gazzdingo:
### Features and Changes

Adds an **opt-in org setting** (`requireRegisteredAttributes`) that makes feature-rule and experiment saves reject typo'd attributes — mirroring the existing saved-group "Unknown attributeKey" behavior. 

When the setting is on, saves throw a `BadRequestError` listing the bad keys if any of these aren't declared (and unarchived) in `org.settings.attributeSchema`:

- `hashAttribute` (rollout, experiment, experiment-ref, safe-rollout rules; experiments)
- `fallbackAttribute` (experiment, experiment-ref rules; experiments)
- Any non-operator field name in rule/phase `condition` JSON (recurses into `$and`/`$or`/`$nor`/`$not`, skips `$`-prefixed operators)

Error wording: `Unknown attribute key(s) on rule: "account_uuid", "userID". Declare them under Settings > Targeting Attributes or fix the typo.`

**Coverage** — single shared helper called from every surface that writes a rule or experiment, so there's no drift between front-end and API or between create and update:

- Feature rules: REST `POST/PUT /api/v1/features/.../rules/...`, bulk `POST /features` + `PUT /features/:id`, internal `postFeatureRule` + `putFeatureRule`.
- Experiments: REST `POST /experiments` + `POST /experiments/:id`, internal `postExperiments` / `postExperiment` / `postExperimentTargeting` / `postExperimentPhase` / `putExperimentPhase`.
- Client-side pre-flight in `RuleModal`, `EditTargetingModal`, `NewExperimentForm` — same shared utilities as the back-end, so the pre-network error message is byte-identical.

**New UI**: single checkbox on `Settings > Targeting Attributes` — "Reject feature rules and experiments that reference unknown attributes" — gated behind `canManageOrgSettings`. Defaults off for backward compatibility.

**Notes**
- Archived attributes are treated as **invalid** (same as unknown) — intentional, so deprecating an attribute actually prevents new references.
- Prerequisite conditions are explicitly excluded (they're evaluated against `{ value: <flag_value> }`, not org attributes — already validated separately).
- Existing rules are not backfilled; validation only fires on save.

### Dependencies

None.

### Testing

Unit tests added:
- `packages/shared/test/util/features.test.ts` — `extractConditionAttributeKeys` (operators, `$and`/`$or`/`$not` recursion, `$inGroup`, dot-notation, dedupe) + `findUnregisteredAttributes` (missing, archived, dot-notation via root segment).
- `packages/back-end/test/services/attributes.test.ts` — `assertRegisteredAttributes` (setting-off no-op, unknown `hashAttribute` / `fallbackAttribute` / condition key, archived-as-unknown, aggregated error, unparseable JSON pass-through, dot-notation).

Manual testing:
1. `Settings > Targeting Attributes` → enable the new toggle.
2. Create a feature rule with `hashAttribute: "userID"` (schema has `userId`). Save → should reject client-side with the typo'd key in the error. Retry via REST API → same rejection server-side.
3. Repeat for a rule `condition` with an unknown field (`{"account_uuid": "x"}`) and for experiment targeting `hashAttribute` / `fallbackAttribute` / phase `condition`.
4. Disable the toggle → all of the above saves pass again (no-op).
5. Archive a declared attribute → saves referencing it are now rejected.


.gitignore
bryce-fitzsimons, line 1: remove

bryce-fitzsimons:
We might want this for saved groups (condition groups) as well

bryce-fitzsimons:
Might want to dedupe this with `getAttributeReferences` into a common condition walk method.

bryce-fitzsimons:
We could consider adding validation in the models (Feature, FeatureRevision, SavedGroup??) instead of individual endpoints

gazzdingo:
@greptileai

gazzdingo:
@greptileai

gazzdingo:
@greptileai

bryce-fitzsimons (approved):

PR #5749, opened 2026-04-27
REST: clarify trackingKey param for listExperiments

lukesonnet:
Deprecates `experimentId` and replaces it with the better named `trackingKey` param for `listExperiments` REST API endpoint.

### DIFF
--- a/packages/shared/src/validators/experiments.ts
+++ b/packages/shared/src/validators/experiments.ts
@@ -1282,12 +1282,18 @@ export const listExperimentsValidator = {
       ...paginationQueryFields,
       projectId: z.string().describe("Filter by project id").optional(),
       datasourceId: z.string().describe("Filter by Data Source").optional(),
+      trackingKey: z
+        .string()
+        .describe("Filter by experiment tracking key")
+        .optional(),
       experimentId: z
         .string()
         .describe(
-          "Filter the returned list by the experiment tracking key (id)",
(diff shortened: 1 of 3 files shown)


lukesonnet:
Screenshot 2026-04-27 at 12 07 37 PM
All looks good, I don't particularly care about greptiles complaints.

packages/shared/src/validators/experiments.ts
```diff
@@ -1282,12 +1282,18 @@ export const listExperimentsValidator = {
...
+        .optional(),
       experimentId: z
         .string()
         .describe(
-          "Filter the returned list by the experiment tracking key (id)",
+          "Filter the returned list by the experiment tracking key (not the internal experiment ID). Note, this was deprecated to help reduce confusion, consider using `trackingKey` instead, which is functionally identical. You cannot use both params at the same time.",
         )
-        .optional(),
+        .optional()
+        .meta({ deprecated: true }),
```
ahdriel, line 1295: ooh TIL -- what does this do? just marks it as `deprecated` for OpenAPI generated docs?

ahdriel (approved):
Thanks for the clean up! It always makes me confused.

packages/shared/src/validators/experiments.ts
jdorn: yes, anything in `.meta` gets passed on as-is when calling `z.toJSONSchema` and ends up in our OpenAPI spec, which is then used to generate the docs.

PR #5753, opened 2026-04-28
Updates canSubscribe logic

mknowlton89:
### Features and Changes

This PR simplifies the logic for the `canSubscribe` variable to reflect the changes we've made to how licenses/subscriptions are managed.

Now, we get the org's effective account plan, and we block an org's ability to upgrade if their plan is `enterprise`, or if they have a `pro` or `pro_sso` plan and their status isn't `cancelled`.

Previously, we were using some older objects to identify this - now, we're getting it entirely from the license.


mknowlton89:
@greptile review

mknowlton89:
@greptile review

mknowlton89:
@greptile review

PR #5754, opened 2026-04-28
REST: Add experiment start, stop, and modify temp rollout endpoints

lukesonnet:
## Summary

- Adds dedicated experiment lifecycle REST endpoints and shared service logic so internal and external flows use the same business rules in `experimentChanges`.
- Introduces `POST /api/v1/experiments/:id/start` and updates `POST /api/v1/experiments/:id/stop` contract/behavior, including checklist handling, winner/release validation, and stronger status guards.
- Adds `POST /api/v1/experiments/:id/modify-temporary-rollout` to toggle temporary rollout on stopped experiments, separating rollout updates from stop semantics.
- Hardens edge cases: starting experiments with no phases, running-only stop enforcement (with internal compatibility path), and explicit temporary rollout validation.
- Regenerates OpenAPI spec and validator docs/examples to match the new API surface.


lukesonnet:
@greptileai

lukesonnet:
@cursor review

lukesonnet:
@cursor review

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2110,24 +2111,26 @@ export async function postExperimentStatus(
...
+        object: "experiment",
+        id: experiment.id,
+      },
+      details: auditDetailsUpdate(experiment, updated),
+    });
+
+    res.status(200).json({
+      status: 200,
+    });
+    return;
```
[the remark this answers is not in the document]
lukesonnet: @jdorn, curious about your thoughts here about custom errors and REST vs. internal API endpoints.

packages/shared/src/validators/experiments.ts
```diff
@@ -1266,6 +1266,78 @@ const updateExperimentBody = z
...
+    enableTemporaryRollout: z
+      .boolean()
+      .describe(
+        "If true, include this stopped experiment in SDK payload and force the release variation (`releasedVariationId`) to all traffic.",
+      )
+      .optional(),
+    releasedVariationId: z
+      .string()
+      .describe(
+        "Required if enableTemporaryRollout is true. Variation ID (e.g. var_abc123) to release to 100% of traffic eligible for this experiment.",
```
jdorn, line 1295: Should we have this act as an optional override?  If omitted, the winnerVariationId will be used instead.  I'm ok with it being explicit here as well, no strong feelings either way.

packages/back-end/src/api/experiments/postExperimentStart.ts
```diff
@@ -0,0 +1,36 @@
...
+import { startExperiment } from "back-end/src/services/experimentChanges/changeExperimentStatus";
+import { toEnhancedExperimentApiResponse } from "./enhancedExperimentResponse";
+
+export const postExperimentStart = createApiRequestHandler(
+  postExperimentStartValidator,
+)(async (req) => {
+  const { experiment, updated } = await startExperiment({
+    context: req.context as ReqContext,
+    experimentId: req.params.id,
+    skipChecklist: req.body?.skipChecklist,
```
jdorn, line 17: We should also add an easy way to get current pre-launch checklist status / items for a draft.  Without that, it's kind of hard to build a workflow around this.

Totally fine if we do this in a follow up PR (along with Make Changes flow)

packages/back-end/src/controllers/experiments.ts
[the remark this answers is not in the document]
jdorn: No strong feelings.  It's fine to return extra debug fields in an error response.  Callers will just ignore fields they don't know about.

jdorn (approved):
I did not test or do a super deep line-by-line review, but general approach looks good

packages/shared/src/validators/experiments.ts
lukesonnet: I like the separation and requirement even if it's a tiny bit more unwieldy. We can relax it further later; harder to do the other direction.

lukesonnet (commented):
Will fast-follow with pre-launch checklist endpoint, and will work with the team on resolving the metadata issue for errors.

PR #5757, opened 2026-04-28
Changing org default model to haiku from gpt 5.4 mini

oelshaikh:
### Features and Changes

- Changing default model to haiku now that anthropic API is set in cloud environment. This is primarily to test product analytics AI agent performance.


config/_config_.yml
```diff
@@ -0,0 +1,233 @@
+organization:
```
jdorn: Was this committed by accident?

oelshaikh: yes, good catch, removed now.

jdorn (approved):

PR #5758, opened 2026-04-28
Updates to the explorer view data formatting and pro badge 

lillialjackson:
### Features and Changes

This change set updates the format for the results view, tooltip, and crosshair tool tip data display for the explorer pages. It also includes a small change to set the pro badge color to the current color in the "save to dashboard" button

### Dependencies

None

### Testing

Manual

### Screenshots

Screenshot 2026-04-28 at 12 39 55 PM

Screenshot 2026-04-28 at 12 40 05 PM

Screenshot 2026-04-28 at 12 40 17 PM

Screenshot 2026-04-28 at 12 40 45 PM

Screenshot 2026-04-28 at 12 40 58 PM


packages/front-end/components/GetStarted/PaidFeatureBadge.tsx
```diff
@@ -12,6 +12,7 @@ export type Props = {
   premiumText?: string | JSX.Element;
   useTip?: boolean;
   variant?: "outline" | "solid";
+  style?: CSSProperties;
```
jdorn: I don't love opening up `style` props if we can avoid it since it makes it too easy to bypass our style guides.  Maybe a more tailored prop like `inheritColor: true` or something would be better here.

jdorn (commented):
Overall looks good.  I also agree with all of the suggested P2 fixes from the Greptile review.

packages/front-end/components/GetStarted/PaidFeatureBadge.tsx
lillialjackson: That's a great point, will change

mknowlton89 (approved):
LGTM!

As I was testing, I feel like we should explore providing users with a timezone selector in case the user is in a timezone different from what the data is. E.G. The data will look odd if we're tracking the data in ET, but the user is viewing the report somewhere in Europe.

Not something we should handle in this PR, just calling it out. I'll track it.

PR #5759, opened 2026-04-28
fix(ids): use uuidv7 for BaseModel and QueryModel id generation

ahdriel:
### Features and Changes

In certain collections we are using `uniqid` to generate pseudo-random ids but we are hitting conflicts more often than what we would like.

So we are switching to use `uuidv7` (with a prefix of `2` to differentiate from `uniqid`), to reduce the chance of those collisions.

- Bump uuid package version to 14
- Create new util for generating ids (drop-in replacement for `uniqid`)
- Update BaseModel to make use of it

### Testing

New unit tests added + suite passes + manual testing

### DIFF
--- a/packages/back-end/jest.config.js
+++ b/packages/back-end/jest.config.js
@@ -1,8 +1,10 @@
 module.exports = {
   moduleFileExtensions: ["ts", "js", "node", "json"],
   transform: {
-    "^.+\\.(ts|tsx)$": "@swc/jest",
+    "^.+\\.(ts|tsx|js|mjs)$": "@swc/jest",
   },
+  // uuid@14+ ships ESM-only. Let swc transpile it for the Jest CJS runtime.
+  transformIgnorePatterns: ["node_modules/(?!\\.pnpm/uuid@|uuid/)"],
   testMatch: ["**/test/**/*.test.(ts|js)"],
   moduleNameMapper: {
     "^axios$": "axios/dist/axios.js",
--- a/packages/back-end/src/models/BaseModel.ts
+++ b/packages/back-end/src/models/BaseModel.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 
 import { v4 as uuidv4 } from "uuid";
-import uniqid from "uniqid";
 import mongoose, { FilterQuery } from "mongoose";
 import { AnyBulkWriteOperation, Collection } from "mongodb";
 import omit from "lodash/omit";
@@ -32,6 +31,7 @@ import {
 } from "back-end/src/api/ApiModel";
 import { CrudAction } from "back-end/src/api/apiModelHandlers";
 import { dbSafeBulkWrite } from "back-end/src/util/mongo.util";
+import { generateId } from "back-end/src/util/uuid";
 import {
   resolveOwnerEmail,
   resolveOwnerEmails,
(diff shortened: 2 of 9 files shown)


packages/back-end/src/models/BaseModel.ts
```diff
@@ -616,7 +615,9 @@ export abstract class BaseModel<
    * Internal methods that can be used by subclasses
    ***************/
   protected _generateId() {
-    return uniqid(this.config.idPrefix);
+    return (
+      (this.config.idPrefix ?? "") + uuidv7().replace(/-/g, "").slice(0, 22)
```
jdorn: Full uuids are 32 hex characters and the last half is the high entropy part, so we're losing quite a bit of that here.  If we want to keep ids short, we could use base58 or similar encoding instead of hex. That can fit the whole uuidv7 into 22 characters without any loss of entropy.

Also, all existing uniqid values start with `1`. We might want to prefix these uuidv7 with `2` so sorting documents from before and after this change will continue working as expected.

Since the logic to generate these is become a little complicated, probably worth adding a util function for it - basically a drop-in replacement for `uniqid`

packages/back-end/jest.config.js
```diff
@@ -1,8 +1,10 @@
 module.exports = {
   moduleFileExtensions: ["ts", "js", "node", "json"],
   transform: {
-    "^.+\\.(ts|tsx)$": "@swc/jest",
+    "^.+\\.(ts|tsx|js|mjs)$": "@swc/jest",
   },
+  // uuid@14+ ships ESM-only. Let swc transpile it for the Jest CJS runtime.
+  transformIgnorePatterns: ["node_modules/(?!\\.pnpm/uuid@|uuid/)"],
```
jdorn, line 7: We should do a quick sanity check in the PR preview env to make sure this ESM-only build works in Docker.  We use a different build process than we do on local dev.

packages/back-end/src/models/BaseModel.ts
ahdriel: Done! Updated to use bs58 like we are already using for other places and created the `util/uuid`. Also added the prefix 2.

packages/back-end/jest.config.js
ahdriel: This is a jest only change for the local test -- but I checked the docker image and ssh'd into it and was able to require uuid and also test that the function is working as expected.

The final docker image does not have jest and other testing structure though, so I am not able to test this specifically.

packages/back-end/src/util/uuid.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Replaces uniqid(prefix), with lower chances of collision
+ *
+ * uuid is prefixed by '2' to differentiate from uniqid(prefix)
+ * and is base58 to reduce the length of the id
+ *
+ * Format: `prefix + "2" + base58(uuidv7)` — e.g. `qry_21CaYDxMweieUAr1YaiBC2d`.
+ */
+export function generateId(prefix = ""): string {
+  return (
+    prefix + "2" + bs58.encode(uuidParse(uuidv7())).padStart(BASE58_WIDTH, "1")
```
jdorn: Is padStart right? Won't that break sorting?  I think it's probably fine to just have inconsistent lengths

ahdriel: From my research it wouldn't given it would be 21 for most of the users until year ~5000 or something like that. But I agree with you that this is additional complexity that we do not need. Going to remove it.

jdorn (approved):

packages/back-end/jest.config.js
jdorn: I just meant create something in the PR preview env and make sure it generates an id properly.  Probably fine though

PR #5761, opened 2026-04-29
docs(self-host): Fix SQL typo in query table name from 'identifies' to 'identifiers'

ynfle:
### Features and Changes

This fixes a type in the `config.yml` section of the self-hosting docs

### Dependencies

None

### Screenshots

Screenshot 2026-04-28 at 20-23-15 Config yml GrowthBook Docs


jdorn:
This is not a typo. Segment is one of the more popular event tracking systems and uses this as their table name when writing to a data warehouse.

PR #5771, opened 2026-04-30
feat: support archived flag in bulk-import facts endpoint

### LINKED ISSUE #5763: [Feature] Support archiving fact metrics via bulk-import/facts (Git-managed metrics)
### Description of Feature

We manage all of our fact metrics as “Official” via the GitHub + bulk-import/facts workflow, using a single metrics.json file in our repo and a CI job that calls POST /api/v1/bulk-import/facts on every push

We’d like to be able to archive fact metrics as part of this same Git-based workflow so that:

- Archiving is code‑reviewed and tracked in git alongside the metric definition.
- If we ever rebuild GrowthBook from metrics.json, the archived state is preserved.
- Today, archiving fact metrics is only documented as supported via the update REST endpoint (/api/v1/fact-metrics/:id with { "archived": true }), not via bulk-import/facts

Adding "archived": true inside the data block of a fact metric in metrics.json is accepted by the bulk import (HTTP 200) but has no effect on the metric’s archived state.

### Requested enhancement

Extend the bulk-import/facts format so that an archived flag on fact metrics (and optionally fact tables) is honored during import, e.g.:

```
{
  "factMetrics": [
    {
      "id": "cm2_var_1_2_plus_adrev_cpo_search_per_user_trino",
      "data": {
        "name": "...(DEPRECATED)",
        "metricType": "mean",
        "archived": true,
        "numerator": { ... }
      }
    }
  ]
}

```
On import, GrowthBook should:

- Create/update the metric as today, and
- Set its archived state to match the JSON (true/false), even for Official metrics.
- Workaround today

We can work around this by keeping metrics.json as‑is and adding a second CI step that calls /api/v1/fact-metrics/:id with { "archived": true } for a git‑tracked list of deprecated metric IDs, but native support in bulk-import/facts would make the Git‑driven workflow much cleaner.

Kevin-Chant:
## Summary

- Adds an optional `archived` field to the fact metric payload accepted by `POST /api/v1/bulk-import/facts`.
- Customers syncing fact metrics from version control can now archive metrics through the same bulk endpoint they use for create/update, instead of having to call the standalone update endpoint separately.

The runtime handlers (`getCreateMetricPropsFromBody` and `getUpdateFactMetricPropsFromBody`) already pass `archived` through to the model layer via spread, so this change only required exposing the field in the bulk-import Zod schema and regenerating the OpenAPI spec.

- closes #5763

## Test plan

- [x] Send a bulk-import payload with `factMetrics[].data.archived: true` for a new metric and confirm it is created in the archived state
- [x] Send a bulk-import payload with `archived: true` / `archived: false` for an existing metric and confirm the archived state is updated
- [x] Confirm `pnpm --filter back-end type-check` passes (verified locally)
- [x] Confirm `pnpm --filter back-end generate-openapi` produces the committed spec (verified locally)

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -11304,6 +11304,12 @@ paths:
                               - ''
                               - api
                               - admin
+                          archived:
+                            description: >-
+                              Set to true to archive the metric. Archived
+                              metrics are hidden by default in the UI and
+                              excluded from new experiments.
+                            type: boolean
                           metricAutoSlices:
                             description: >-
                               Array of slice column names that will be
--- a/packages/shared/src/validators/bulk-import.ts
+++ b/packages/shared/src/validators/bulk-import.ts
@@ -419,6 +419,12 @@ const postBulkImportFactsBody = z
                 'Set this to "api" to disable editing in the GrowthBook UI',
               )
               .optional(),
+            archived: z
+              .boolean()
+              .describe(
+                "Set to true to archive the metric. Archived metrics are hidden by default in the UI and excluded from new experiments.",
+              )
+              .optional(),
             metricAutoSlices: z
               .array(z.string())
               .describe(


Kevin-Chant:
cursor review

tzjames (approved):

PR #5776, opened 2026-04-30
feat(dimensions): Add timeseries for all precomputed dimensions

ahdriel:
### Features and Changes

Metric Timeseries only worked for top-level metrics, but now we are expanding it to support precomputed dimensions as we already have the data available.

Note: Future work will add support for Unit Dimensions as well.

- After a main standard snapshot is updated now we:
  - Run 3 analyses (relative, scaled, absolute) for each precomputed dimension
  - Use the analyses to create the metric time series for the dimension breakdown
- Updated MetricTimeSeriesModel to also take into consideration dimension and dimension value

### Testing

- Unit tests
- Manual testing
  - Update a Experiment results
  - Change dimension breakdown to a precomputed one
  - Click on a Metric
  - Observe that the timeseries graph is now populated


ahdriel:
cursor review

packages/back-end/src/services/experiments.ts
```diff
@@ -1911,6 +1912,92 @@ export type SnapshotAnalysisParams = {
   snapshot: ExperimentSnapshotInterface;
 };
 
+export function getPrecomputedDimensionAnalysisSettings({
+  baseSettings,
+  dimensionId,
+  differenceTypes = ["relative", "absolute", "scaled"],
+  baselineVariationIndex = baseSettings.baselineVariationIndex ?? 0,
```
lukesonnet: If this is separated out from `baseSettings` because it might not be 0, but it needs to be for the dimension time series, is there another way for us to ensure we're getting the right analysis set than relying on `baseSettings` except for this one field?

e.g. what if there's some other field that also should be fixed? Why is this one special?

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -54,6 +57,49 @@ export async function updateExperimentTimeSeries({
...
+    await getExperimentTimeSeriesContext({
+      context,
+      experiment,
+      experimentSnapshot,
+    });
+  const analyses = experimentSnapshot.analyses.filter(
+    (analysis) =>
+      analysis.settings.dimensions.length === 0 &&
+      (analysis.settings.baselineVariationIndex === undefined ||
+        analysis.settings.baselineVariationIndex === 0),
```
lukesonnet: Should there be a separate service for what is a valid time series analysis? Or should we be smarter about letting what can go in a time series doc be more flexible, and be clearer about how we know which is the "right" one for a given analysis settings requested by the user?

lukesonnet (approved):

packages/back-end/src/services/experiments.ts
ahdriel: Good point! I have added a check to ensure that we only kick this off when the update is against the control as the baseline.

We have the same limitation at the moment for non-dimension timeseries and we show `Time series is only available when comparing against Control as a baseline.`

PR #5778, opened 2026-04-30
Feat/list experiment results bulk

adittya-upstart:
### Features and Changes

  Adds `GET /api/v1/experiments/results`, a paginated, filtered bulk endpoint that returns the latest snapshot for many experiments in a single round trip.

  The existing `GET /experiments/{id}/results` requires one HTTP call per experiment. For portfolio-wide reporting (e.g. nightly warehouse pulls), that fans out to N calls and trips the per-IP rate limit. The new endpoint collapses it
  to a small number of paginated reads.

  **Implementation reuses what's already there:**

  - `getAllExperiments` for the org + per-project authz filter
  - `getLatestSnapshotMultipleExperiments` so the snapshot lookup is one Mongo aggregation, not 1+N
  - `toSnapshotApiInterface` so each item has the same `ExperimentResults` shape the per-experiment endpoint already returns
  - `paginationQueryFields` for shared `limit` / `offset` semantics (default 10, max 100)

  **Filters mirror `/experiments`:** `projectId`, `datasourceId`, `status`. `phase` and `dimension` are deliberately omitted: they don't generalize across a portfolio (different experiments have different phase counts; not every
  experiment supports every dimension). Callers that need either should keep using the per-experiment endpoint.

  **Pagination semantics:** experiments without a completed snapshot are silently omitted from `experimentResults` rather than aborting the batch, but they still occupy a slot in pagination, so `experimentResults.length` may be less
  than the page `count`. This is documented on the validator.

  **Routing:** registered before `/experiments/{id}/results` so Express literal-matches `/experiments/results` instead of treating "results" as an `:id` parameter.

  **Usage:**

  ```bash
  curl -u secret_xxx: 'https://api.growthbook.io/api/v1/experiments/results?status=running&limit=50'
  ```

  ### Dependencies

  None.

  ### Testing

  Type-check, lint, and `pnpm --filter back-end test --testPathPattern="api/experiments"` all pass locally (62 tests, 5 new). The new tests live in `packages/back-end/test/api/experiments.test.ts` and cover:

  - happy path: latest snapshot returned per experiment
  - silent skip when an experiment has no snapshot
  - empty page when no experiments match
  - `projectId` / `datasourceId` / `status` filter passthrough
  - experiments without phases are excluded from the snapshot lookup

  While adding the tests I hit a latent rate-limit collision in the test harness: `/api/v1` rate-limits per `req.apiKey` at 60 req/min, but the auth mock in `packages/back-end/test/api/api.setup.ts` left `req.apiKey` undefined so every
  test in the file shared one bucket. The existing 57 tests sat just under the cap; the new 5 tipped it. Setting `req.apiKey = randomUUID()` per request fixes it once for everyone.

  I have not exercised this against a populated dev DB yet. Happy to pair on a smoke test if a maintainer can point me at a fixture or shared environment.

  ### Screenshots

  N/A (API change, no UI).


adittya-upstart:
A few design questions worth a maintainer take before merge — none blocking, just want to flag them so the discussion happens here rather than in a follow-up:

1. **`count` vs `total` asymmetry.** I went with `count = experimentResults.length` and left `total` as "experiments matching the filter". Documented in the validator, but it does mean a paginating client can't predict total returned results without iterating. Cleaner alternatives I considered: (a) add a separate `totalWithSnapshots` field, or (b) pre-filter to experiments-with-snapshots before paginating (one extra Mongo query on `experimentSnapshots` projecting just `experiment` ids). Happy to do either if you prefer; current shape was the smallest deviation from the existing list-endpoint pattern.

2. **`includeArchived: true` is hardcoded.** Matches `listExperiments`, so it's consistent, but neither endpoint exposes a toggle. Worth adding a query param to both as a follow-up?

3. **`api.setup.ts` rate-limit fix scope.** I set `req.apiKey = randomUUID()` per request so the per-key 60/min limiter doesn't share buckets across tests. This means a future test that intentionally exercises the 429 path would need a way to reuse a key. Not a problem today (no such test exists), but worth flagging in case there are plans there.

Will keep an eye out for review and address anything else that comes up.

Kevin-Chant (approved):
LGTM, thanks for the submission!
For your design questions, I agree with the choices you made for this branch, and if we want to change `archived: true` or the random api keys in tests we can handle those in future PRs

PR #5780, opened 2026-05-01
feat(experiments): Enable scheduling the start time for an experiment

msamper:
### Features and Changes

Allows users to schedule a start time for an experiment through a two step process. Users can set a scheduled start date, make any necessary changes to the experiment configuration and then can approve an experiment for the scheduled start. Approving an experiment for the scheduled start goes through the same modal as starting the experiment and is subject to the same requirements. 

Adds a new pre-launch checklist item that can be added to an orgs pre-launch checklist to require adding a scheduled experiment start time.

Adds a new agenda job that is responsible for starting an experiment when there is a valid, approved scheduled start time. This job can easily be extended in the future to also stop an experiment when we support stop scheduling.

### Adding a Schedule
- [x] "Add Schedule" button does **not** appear for bandit experiments
- [x] Saving a future date sets `statusUpdateSchedule.startAt` and the UI now shows "Target Start: [date]" with a pencil icon
- [x] Saving a future date **does not** set `nextScheduledStatusUpdate` (requires explicit approval)
- [x] Submitting the modal with no date selected is disabled (CTA is disabled when no date and no existing schedule)
---
### Editing / Removing a Schedule (before approval)
- [x] Clicking the "Target Start: [date]" button opens the modal in "Edit Schedule" mode
- [x] Clearing the date picker and saving removes the schedule (sets `statusUpdateSchedule` to `null`)
- [x] **No "Unschedule Experiment"** red button is shown in the modal when the schedule has not been approved yet
- [x] Editing the schedule to a new date clears any existing `nextScheduledStatusUpdate` approval
---
### Schedule Past-Due Warning
- [x] If `statusUpdateSchedule.startAt` is in the past, the "Target Start" button shows a warning icon (⚠️)
---
### Start Experiment Modal — Scheduled Flow
- [x] When a future schedule exists, "Start Experiment" button label changes to **"Approve for Scheduled Start"**
- [x] After approval, the header shows **"Starts [date] at [time]"** instead of the "Start" button
- [x] When checklist is **incomplete** + schedule is future: primary CTA is disabled; secondary button says **"Schedule Anyway"**
- [x] When schedule date is **in the past**: primary CTA is disabled; secondary button says **"Start Now"** (immediate start)
- [x] When there is **no schedule**: checklist-incomplete behavior is unchanged — secondary CTA says **"Start Anyway"**
---
### Editing Schedule After Approval (Unschedule)
- [x] Clicking the "Starts [date]" button in the header opens `EditScheduleModal`
- [x] The modal shows a red **"Unschedule Experiment"** secondary button when `nextScheduledStatusUpdate` is set
- [x] Clicking "Unschedule Experiment" clears `nextScheduledStatusUpdate` and returns the header to the "Approve for Scheduled Start" button
- [x] Editing the date on an already-approved schedule clears the approval and the user must re-approve
---
### Auto-Start Agenda Job
- [x] An experiment with `nextScheduledStatusUpdate.date` in the past is auto-started by the background job (within ~1 minute)
- [x] After auto-start: experiment moves to `running`, `nextScheduledStatusUpdate` is cleared
- [x] If the experiment is already `running` or `stopped` when the job fires, it skips and clears `nextScheduledStatusUpdate`
- [x] If the experiment is archived when the job fires, it skips and clears `nextScheduledStatusUpdate`
- [x] Manually starting an experiment (bypassing the schedule) also clears `nextScheduledStatusUpdate`
---
### Edge Cases
- [x] Approving a schedule with a date that is already in the past returns a `400` error
- [x] Pre-launch checklist "schedule" item (when configured by org) shows "Add scheduled start date" link and is marked complete once `statusUpdateSchedule.startAt` is set

### Screenshots

Screenshot 2026-05-07 at 3 10 27 PM
Screenshot 2026-05-07 at 3 10 35 PM
Screenshot 2026-05-07 at 3 10 50 PM
Screenshot 2026-05-07 at 3 10 58 PM
Screenshot 2026-05-07 at 3 11 06 PM
Screenshot 2026-05-07 at 9 03 58 AM


packages/front-end/components/Experiment/TabbedPage/StartExperimentModal.tsx
```diff
@@ -58,84 +106,147 @@ export default function StartExperimentModal({
...
+    hasSchedule && !isHoldout
+      ? "Schedule Experiment to Start"
+      : isHoldout
+        ? "Start Holdout"
+        : "Start Experiment";
+
+  const primaryCta =
+    hasSchedule && parsedScheduledDate
+      ? `Start ${format(parsedScheduledDate, "MMM d, yyyy 'at' h:mm a")}`
+      : "Start Now";
```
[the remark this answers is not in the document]
msamper: This is a part of the design for this feature. Will review with Kari

packages/back-end/src/services/experimentChanges/changeExperimentStatus.ts
```diff
@@ -297,31 +294,53 @@ export async function startExperiment({
...
     changes.phases = startExperimentTarget.phases;
   }
 
-  // Publish linked feature drafts atomically with the status transition.
-  // If any draft cannot be published, abort the start.
-  const publishResult = await publishPendingFeatureDraftsForExperiment(
+  const updated = await updateExperiment({
     context,
     experiment,
+    changes: { nextScheduledStatusUpdate: null, ...changes },
```
lukesonnet, line 376: When we add `stop` in a follow-up PR, let's probably refactor this logic setting this to be `null` to a separate service that handles the state transition for both status and the next scheduled status, maybe.

lukesonnet, line 376: Image

For some reason it doesn't seem like this update is triggering an audit log event? You can see me setting the schedule here, but even though this experiment is running, there's no "start" audit log event.

packages/front-end/components/Experiment/LinkedChanges/AddLinkedChanges.tsx
```diff
@@ -160,6 +160,7 @@ export default function AddLinkedChanges({
   setUrlRedirectModal: (state: boolean) => unknown;
 }) {
   if (experiment.status !== "draft") return null;
+  if (experiment.nextScheduledStatusUpdate) return null;
```
lukesonnet, line 163: nit: Maybe there should be a service for "isEditableExperiment" or something? Perhaps we can do this as a follow-up

packages/shared/src/enterprise/decision-criteria/statusIndicatorData.ts
```diff
@@ -99,9 +106,6 @@ export function getStatusIndicatorData(
     }
   }
 
-  // TODO: Future statuses
-  // return ["indigo", "soft", "Scheduled"];
```
lukesonnet, line 103: nice!

packages/front-end/components/Experiment/TabbedPage/StartExperimentModal.tsx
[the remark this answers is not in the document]
lukesonnet: Nah, it's fine. Go with the existing design.

packages/back-end/src/jobs/updateExperimentStatus.ts
```diff
@@ -0,0 +1,141 @@
...
+          experiment,
+          changes: { nextScheduledStatusUpdate: null },
+        });
+        return;
+    }
+
+    logger.info("Successfully updated status for experiment " + experiment.id);
+  } catch (e) {
+    logger.error(e, "Failed to update experiment status " + experiment.id);
+  }
```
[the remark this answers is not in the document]
lukesonnet: This seems like something to fix. Throw an error and unset the nextScheduledStatusUpdate, but we probably want this to be pretty noisy. How do we handle a failure to schedule start a feature flag? Maybe that should be our north star.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1678,6 +1687,8 @@ export async function postExperiment(
       key === "metricOverrides" ||
       key === "lookbackOverride" ||
       key === "variations" ||
+      key === "statusUpdateSchedule" ||
+      key === "nextScheduledStatusUpdate" ||
       key === "customFields" ||
```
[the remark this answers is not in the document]
lukesonnet: I was basically going to leave a nit about this. I think this endpoint should just not accept this internally managed field at all.

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -97,20 +103,38 @@ export default function SetupTabOverview({
     new Date(experiment.phases[0].dateStarted) &&
     holdoutExperiments.length > 0 &&
     holdoutExperiments.some((e) => e.status !== "draft");
-  const canEditHoldoutSchedule =
-    isHoldout && holdout && canEditExperiment && editHoldoutSchedule;
+  const canEditSchedule = !isBandit && canEditExperiment && editSchedule;
```
lukesonnet, line 106: Should we block editing `schedule` in the BE at the `postExperiment` level for bandits, then?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1691,6 +1702,48 @@ export async function postExperiment(
...
+    changes.status &&
+    changes.status !== "draft" &&
+    experiment.nextScheduledStatusUpdate
+  ) {
+    // If the experiment moves out of a schedulable state, clear any pending
+    // scheduled status update so the agenda job doesn't try to start it.
+    changes.nextScheduledStatusUpdate = null;
+  }
+
+  if (approveScheduledStart) {
```
lukesonnet: Should this just be a separate internal endpoint? I think so....

If we leave it here, do we need to update the logic below to use `experiment.statusUpdateSchedule?.startAt` in case it exists in the same payload as `approveScheduledStart?

packages/back-end/src/services/experimentChanges/changeExperimentStatus.ts
lukesonnet: Here's what roboto says:

> postExperimentStatus and startExperiment both fire a req.audit({ event: "experiment.status", ... }) audit log entry. The agenda-driven auto-start does not — only the generic experiment.updated event from logExperimentUpdated is created, and getContextForAgendaJobByOrgObject sets auditUser: null, so we lose the "who approved this" attribution.
> 
> Two pieces:
> 
> Store approvedBy (and ideally approvedAt) inside nextScheduledStatusUpdate so the audit trail can answer "who started this experiment?"
> Emit an explicit experiment.status audit event in the auto-start path with that attribution.

lukesonnet (changes requested):
This looks great Michael. Lots of edge cases and nits in comments, but a few blockers. 3 more below:

1. When publishing an approved feature flag, you can Publish and force start easily even if a pre-launch checklist item is OFF *AND* it ignores the planned schedule. See loom below.
https://www.loom.com/share/1528198a2caa4e9bb715740c410461ab

Proposed solution:
* Change "Publish" to "Schedule" in this case, and go through the "approve to schedule" flow we have on the experiment (or change to "Approve to Scheduled Start"
* Consider a "Schedule anyway" button if pre-launch checklist isn't complete, like we have on experiment, but fine if we don't do this at all if it looks bad. 

Please make an executive decision here!

2. Should we modify the "startExperiment" public API endpoint to by default approve to start an experiment instead of start it if it has a scheduled date? I think we should. If people want to start immediately programmatically, they can remove the schedule and then use the "startExperiment" endpoint. Let's add a note to the description of this endpoint to make the fact that "startExperiment" will just stage an experiment with a future scheduled start date to start right away (and its ok if it silently just starts an experiment with a scheduled start date in the past).

3. Should we make sure we can add and modify `schedule` via the updateExperiment Public API endpoint? Just the main POST patch endpoint should be able to modify the schedule.

packages/back-end/src/controllers/experiments.ts
lukesonnet: Moved to standalone internal endpoint and service: [7f08e92](https://github.com/growthbook/growthbook/pull/5780/commits/7f08e9202e0a94c21dcc7d179405c83d0a41a49b)

packages/back-end/src/services/experimentChanges/changeExperimentStatus.ts
lukesonnet: Didn't add all that additional context, but we needed to emit a status audit event, so I added that in [7f08e92](https://github.com/growthbook/growthbook/pull/5780/commits/7f08e9202e0a94c21dcc7d179405c83d0a41a49b)

No need for that additional context for now; lets punt until better approval flows.

packages/back-end/src/jobs/updateExperimentStatus.ts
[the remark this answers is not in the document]
lukesonnet: I added some "count up to 5 attempts and warn with experiment.warning event and even more logging" logic here: [7f08e92](https://github.com/growthbook/growthbook/pull/5780/commits/7f08e9202e0a94c21dcc7d179405c83d0a41a49b), but we can remove this warning if it's too noisy (IDK if it is, it's kind of bad to fail to start when it looks like it should). Infinite retry also seems bad, however.

packages/back-end/src/controllers/experiments.ts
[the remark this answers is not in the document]
lukesonnet: Ok, I fully removed it in[9ab62db](https://github.com/growthbook/growthbook/pull/5780/commits/9ab62dbbf6a08d589bc598ffed8085def892f0da), and you have to go through the new `postApproveScheduledExperimentStart` controller created in [7f08e92](https://github.com/growthbook/growthbook/pull/5780/commits/7f08e9202e0a94c21dcc7d179405c83d0a41a49b).

msamper:
@greptileai

msamper:
@greptileai

lukesonnet:
Fixed one greptile concern but not the holdout one since the holdout can accept that schedule via REST API, and it's ok if the start endpoint doesn't apply to it.

lukesonnet (approved):

PR #5782, opened 2026-05-01
Fix React console warnings

jdorn:
### Features and Changes

Fixes 3 React warnings that were clogging up console logs on dev:

1. Invalid prop `preventDefault` on HTML elements
2. `<div>` inside of `<p>`
3. Smooth scroll behavior during navigation (https://nextjs.org/docs/messages/missing-data-scroll-behavior)


ahdriel (approved):
Haven't tested but the changes do make sense to me!

PR #5787, opened 2026-05-01
chore(design-system): Rename Dialog to Modal and restrict imports of old Modal

msamper:
### Features and Changes

Renames the new `Dialog` component and associated files to use `Modal` to avoid confusion. Also adds an eslint rule to restrict new imports of the old `Modal` component


packages/front-end/ui/Modal/Patterns/ModalStandard.tsx
ahdriel, line 1: nit: I am not the biggest fan of `ModalLayout` name because I think it does not represent what the component does.

Some ideas:
- `StandardModal` is the first that comes to mind. Alternatively we can do `ModalStandard` to make it easier to find based on the prefix?
- `ModalTemplate` ?

For the future (more thinking out loud, less of a request):
- Maybe we adopt a convention to have `{Component}Patterns` and use that namespace for pre-built structures?
  - In this case it would be `ModalPatterns.Standard` and we could have `ModalPatterns.MultiPage` for example

ahdriel (approved):

eslint.config.mjs
```diff
@@ -216,6 +216,11 @@ export default defineConfig([
                 "Text",
               ],
             },
+            {
+              name: "@/components/Modal",
+              message:
+                "Use the new Modal from @/ui/Modal instead of the legacy Modal component.",
+            },
```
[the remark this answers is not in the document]
msamper: I don't think we want to block imports for any of these components right now since we don't have an alternative using the new `Modal` yet

packages/front-end/ui/Modal/Patterns/ModalStandard.tsx
msamper: Yeah I'm with you. I like `ModalStandard` as an alternative because it has the Modal prefix and agree it could make it easier to find. 

I like the idea of adopting the convention you're proposing in the future. I think it would be more clear than what I have now

PR #5790, opened 2026-05-02
feat: support {{experimentId}} in fact-table SQL for incremental refresh + docs

nadapzy:
## Use case

We have a shared Fact Table that emits classifier-score rows for **many** experiments simultaneously, keyed by a column like `sample_type = 'experiment:<trackingKey>'`. When GrowthBook joins exposures ⋈ fact-table on the bare identifier, a user enrolled in N experiments gets N× the rows in every experiment's metric — so we need the fact-table SQL itself to filter to the experiment being analyzed:

```sql
WHERE sample_type = CONCAT('experiment:', '{{ experimentId }}')
```

## What this PR does

`{{ experimentId }}` already exists in `compileSqlTemplate` and is already threaded into `getFactMetricCTE` for the **standard** experiment-analysis path (`experiment-fact-metrics-query.ts`, `experiment-metric-query.ts`). This PR closes two small gaps:

1. **Incremental pipeline mode** — `getInsertMetricSourceCovariateDataQuery` and `getInsertMetricSourceDataQuery` were calling `getFactMetricCTE` without `experimentId`, even though `params.settings.experimentId` is in scope. Fact tables that reference `{{ experimentId }}` would silently render as `%` under incremental refresh, returning rows for all experiments. Now passes `params.settings.experimentId`.
2. **Docs** — adds an *Experiment Id Filter* section to the SQL Templates guide. The variable was previously only documented for assignment queries / dimensions, not fact tables.
3. **Test** — adds a unit test on `getFactMetricCTE` asserting both substitution of the tracking key and the `%` fallback.

## Behavior outside experiment context

Unchanged: when there is no experiment in scope (fact-table preview, standalone Metric Analysis, column introspection, product analytics), `{{ experimentId }}` resolves to `%` per the existing fallback in `compileSqlTemplate`. The docs recommend writing the filter as a `LIKE` if the query needs to return rows in both contexts.

## Testing

- `packages/back-end`: `jest test/sqlintegration.test.ts` — 2257 passed (incl. all 2244 snapshots)
- `packages/back-end`: `tsgo --noEmit` — clean

### DIFF
--- a/docs/docs/experimentation-analysis/sql-templates.mdx
+++ b/docs/docs/experimentation-analysis/sql-templates.mdx
@@ -67,6 +67,26 @@ The inserted values do not have surrounding quotes, so you must add those yourse
 
 :::
 
+## Experiment Id Filter
+
+When a Fact Table or Metric query is run as part of an experiment analysis, the `{{ experimentId }}` variable is replaced with the experiment's tracking key. This is useful when a single Fact Table contains rows for many experiments and you need to scope it to the one being analyzed:
+
+```sql
+SELECT
+  user_id,
+  timestamp,
+  score
+FROM classifier_scores
+WHERE
+  sample_type = CONCAT('experiment:', '{{ experimentId }}')
+```
+
+When the same SQL is run outside of an experiment context (for example, when previewing a Fact Table or running a standalone Metric Analysis), `{{ experimentId }}` is replaced with `%`. If you need the query to return rows in both contexts, write the filter as a `LIKE` so that `%` matches everything:
+
+```sql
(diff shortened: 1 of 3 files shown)


lukesonnet:
@cursor review

lukesonnet:
@greptileai

lukesonnet (approved):
Updated the docs and added a couple of other fields that were not added, but this looks good, thanks!

lukesonnet (approved):
Updated the docs and added a couple of other fields that were not added, but this looks good, thanks!

lukesonnet (approved):
Updated the docs and added a couple of other fields that were not added, but this looks good, thanks!

PR #5793, opened 2026-05-04
feat(fact-metrics): add "hll merge" / "kll merge" column aggregations for pre-built sketch columns

johnham-ant:
# feat(fact-metrics): "hll merge" / "kll merge" column aggregations for pre-built sketch columns

## Motivation

We run GrowthBook against pre-aggregated fact tables where the heavy lifting happens upstream: an hourly job rolls raw events up to one row per user-hour, and for distinct-count and percentile inputs it stores **BigQuery sketch columns** (`HLL_COUNT.INIT(...)` / `KLL_QUANTILES.INIT_FLOAT64(...)` → `BYTES`) rather than raw values.

Today `columnAggregation` only supports `sum | max | count distinct`, so there's no way to tell a fact metric "this column is already a sketch — merge it, don't re-init it." That forces us either back onto the raw event table (billions of rows per experiment query) or onto a separate non-GrowthBook stats path.

#5532 added every dialect primitive this needs (`hllReaggregate`, `hllCardinality`, `kllMergePartial`, `kllExtractPoint`, `kllRankApprox`) for the incremental-refresh pipeline. This PR is the small follow-up that wires those same primitives to a `columnAggregation` value so a fact metric can consume a pre-built sketch column directly.

## Changes

- **`columnAggregation` enum** — add `"hll merge"` and `"kll merge"` in `shared/validators/fact-table.ts` and the duplicated API / bulk-import validators; update `.describe()` strings; regenerate `back-end/generated/spec.yaml`.
- **`getAggregationMetadata`** (`back-end/src/integrations/sql/fact-metrics/aggregation-metadata.ts`) — two new branches:
  - `"hll merge"`: identical to `"count distinct"` but `partialAggregationFunction` / `fullAggregationFunction` call `dialect.hllReaggregate` (BigQuery `HLL_COUNT.MERGE_PARTIAL`) instead of `dialect.hllAggregate` (`HLL_COUNT.INIT`). Result is an integer per user, so mean/ratio statistics work unchanged.
  - `"kll merge"`: identical to the existing event-quantile branch but `partialAggregationFunction` calls `dialect.kllMergePartial` (`KLL_QUANTILES.MERGE_PARTIAL`) instead of `dialect.kllInit`. Guarded to `metricType: "quantile"` + `quantileSettings.type: "event"`. The two-pass KLL rank-recovery path from #5532 is reused as-is.
- **`validateAggregationSpecification`** (`back-end/src/api/fact-metrics/postFactMetric.ts`) — reject `hll merge` / `kll merge` on `string` / `number` columns with a clear error.
- **Tests** (`back-end/test/integrations/BigQuery.test.ts`) — 4 new cases:
  - `"hll merge"` emits `HLL_COUNT.MERGE_PARTIAL` + `HLL_COUNT.EXTRACT`, never `HLL_COUNT.INIT`.
  - `"kll merge"` emits `KLL_QUANTILES.MERGE_PARTIAL`, never `INIT_FLOAT64`.
  - `"kll merge"` on a non-event-quantile metric falls through (guard holds).
  - E2E `getInsertMetricSourceDataQuery` with a `"kll merge"` metric emits `MERGE_PARTIAL` and not `INIT`.

**No new dialect primitives, no new capability flags.** The base dialect already throws on `hllReaggregate` / `kllMergePartial`, so data sources without HLL/KLL support fail with the existing message. The front-end metric modal is untouched — these aggregations are API-first for programmatically managed metrics; happy to add UI in a follow-up if wanted.

## Example

```yaml
# p90 latency from a pre-built KLL sketch column
- id: fact__latency_p90
  metricType: quantile
  quantileSettings: { quantile: 0.9, type: event }
  numerator:
    factTableId: ftb__hourly_events
    column: latency_ms_kll       # BYTES, KLL_QUANTILES.INIT_FLOAT64 upstream
    aggregation: "kll merge"

# distinct sessions per user from a pre-built HLL sketch column
- id: fact__sessions_per_user
  metricType: mean
  numerator:
    factTableId: ftb__hourly_events
    column: session_id_hll       # BYTES, HLL_COUNT.INIT upstream
    aggregation: "hll merge"
```

Generated per-user SQL (BigQuery):

```sql
-- kll merge
KLL_QUANTILES.MERGE_PARTIAL(m0_value) AS m0_value           -- BYTES per user

-- hll merge
HLL_COUNT.EXTRACT(HLL_COUNT.MERGE_PARTIAL(m1_value)) AS ... -- INT64 per user
```

## Impact

Lets experiment queries for distinct-count and event-percentile metrics run against pre-aggregated roll-ups instead of raw event tables — in our case that moves a class of metrics from billions of rows scanned per query to the hourly grain.

## Testing

- `pnpm --filter shared run type-check` — clean
- `pnpm --filter back-end run type-check` — clean
- `pnpm --filter shared run test` — 620 passed
- `pnpm --filter back-end run test -- test/integrations/` — 23 passed (4 new)

### DIFF
--- a/packages/back-end/src/api/fact-metrics/postFactMetric.ts
+++ b/packages/back-end/src/api/fact-metrics/postFactMetric.ts
@@ -8,12 +8,9 @@ import {
   DEFAULT_PROPER_PRIOR_STDDEV,
   DEFAULT_WIN_RISK_THRESHOLD,
 } from "shared/constants";
-import { getSelectedColumnDatatype } from "shared/experiments";
 import { postFactMetricValidator } from "shared/validators";
 import {
-  ColumnRef,
   CreateFactMetricProps,
-  FactMetricType,
   FactTableInterface,
 } from "shared/types/fact-table";
 import { OrganizationInterface } from "shared/types/organization";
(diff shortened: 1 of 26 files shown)


packages/back-end/src/api/fact-metrics/postFactMetric.ts
```diff
@@ -39,6 +39,15 @@ export function validateAggregationSpecification({
...
   }
+  if (
+    (column.aggregation === "hll merge" ||
+      column.aggregation === "kll merge") &&
+    (datatype === "string" || datatype === "number")
+  ) {
+    throw new Error(
+      `${errorPrefix}Cannot use '${column.aggregation}' aggregation with the ${datatype} column '${column.column}'. The column must be a pre-built sketch (e.g. BigQuery BYTES).`,
+    );
+  }
```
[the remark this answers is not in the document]
lukesonnet: Now resolved in later data checks.

lukesonnet:
@cursor review

packages/shared/test/product-analytics.test.ts
```diff
@@ -13,20 +13,24 @@ describe("productAnalytics", () => {
...
+    hllCardinality: (col) => `HLL_COUNT(${col})`,
+    kllMergePartial: (col) => `KLL_MERGE(${col})`,
+    kllExtractPoint: (col, quantile) => `KLL_POINT(${col}, ${quantile})`,
     toTimestamp: (d: Date) =>
       // Do not include the timestamp component to make the test deterministic
       `'${d.toISOString().substring(0, 10)} 00:00:00'`,
     formatDialect: "bigquery",
     castToFloat: (col) => `CAST(${col} AS FLOAT)`,
-  };
+  } as Partial<SqlDialect> as SqlDialect;
```
ahdriel: question: why this change?

I would vote to keep the original `: SqlDialect`

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -519,10 +524,15 @@ function getEventValueExpr(
...
 }
 
-function getUnitAggregationExpr(columnRef: ColumnRef, alias: string): string {
-  if (columnRef.column === "$$distintDates") {
+function getUnitAggregationExpr(
+  columnRef: ColumnRef,
+  alias: string,
+  helpers: SqlDialect,
+): string {
+  if (columnRef.column === "$$distinctDates") {
```
ahdriel, line 531: oooh -- this `typo` meant this was broken before?

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -534,24 +544,55 @@ function getUnitAggregationExpr(columnRef: ColumnRef, alias: string): string {
...
+
+  return assertNever(aggregation);
 }
 
 function getRollupAggregationExpr(
   metric: MinimalMetric,
+  columnRef: ColumnRef,
   alias: string,
   helpers: SqlDialect,
+  fromUnitAggregation: boolean,
```
ahdriel, line 575: nit: I feel that boolean flags usually are code smells for patterns that would require a bigger refactor.

I think we can proceed as-is, but raising as something to take a deeper look later

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -876,11 +896,7 @@ export default function FactMetricPage() {
         </div>
         <div className="col-12 col-md-4">
           <div className="appbox p-3">
-            <RightRailSection
-              title="Advanced Settings"
-              open={() => setEditOpen("openWithAdvanced")}
-              canOpen={canEdit}
-            >
+            <RightRailSection title="Advanced Settings">
```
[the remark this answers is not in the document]
Comment (lukesonnet) [reply packages/front-end/pages/fact-metrics/[fmid].tsx]: nice, thanks, resolved.

packages/front-end/pages/fact-metrics/[fmid].tsx
```diff
@@ -560,7 +573,7 @@ export default function FactMetricPage() {
             open={openDropdown}
             onOpenChange={setOpenDropdown}
           >
-            {canEdit && (
+            {canEdit && !editViaApiOnly && (
```
Comment (ahdriel) [inline packages/front-end/pages/fact-metrics/[fmid].tsx]: suggestion: I feel this is just waiting for a refactor so we have a bug and show 2 `Edit Metric` buttons in the UI

wdyt about using `canEdit ? (editViaApiOnly ? disabled : button) : null)`

Or we can always render it and show a `disabled` tooltip saying that either it is editable only via API or that they don't have permission to edit it.

packages/back-end/src/services/factMetricAggregationValidation.ts
```diff
@@ -0,0 +1,123 @@
...
+  // context to tell.
+  if (
+    column.aggregation === "kll merge" &&
+    metricType !== undefined &&
+    (metricType !== "quantile" || quantileType !== "event")
+  ) {
+    throw new Error(
+      `${errorPrefix}'kll merge' aggregation is only valid for event-quantile metrics (metricType=quantile, quantileSettings.type=event).`,
+    );
+  }
```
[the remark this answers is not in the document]
lukesonnet: Nice, thanks.

packages/front-end/components/FactTables/FactMetricList.tsx
```diff
@@ -32,24 +32,49 @@ import {
 } from "@/ui/DropdownMenu";
 import FactMetricModal from "./FactMetricModal";
 
+const REST_API_ONLY_EDIT_MESSAGE =
```
ahdriel: nit: we now have this message in 3 places

I like the colocation of where we are showing them and rendering, but also feel that the message will drift as we update this.

wdyt about a shared const?

ahdriel (approved):

packages/shared/src/enterprise/product-analytics/sql.ts
lukesonnet: yep, probably no one used it.

packages/shared/test/product-analytics.test.ts
lukesonnet: I'm sure the robot didn't like that it was throwing type errors before and muted it while trying to add the KLL methods.

packages/shared/src/enterprise/product-analytics/sql.ts
lukesonnet: Yes, but the alternative here is to basically do the full "aggregation", "reaggregation", "fullaggregation" data model we have in sql-integration.

packages/front-end/pages/fact-metrics/[fmid].tsx
Comment (lukesonnet) [reply packages/front-end/pages/fact-metrics/[fmid].tsx]: This approach was crazy, sorry I missed what happened here. I don't like the combined disabled tooltip, esp. until we align on a general strategy around permissions and disappearing buttons.

LMK if you like the new approach.

packages/front-end/components/FactTables/FactMetricList.tsx
lukesonnet: Moved the const and the logic checking the condition.

packages/front-end/components/FactTables/FactMetricList.tsx
```diff
@@ -424,12 +454,20 @@ export default function FactMetricList({
                     {metric.dateUpdated ? date(metric.dateUpdated) : null}
                   </td>
                   <td>
+                    {/*
```
ahdriel: nit: can we remove this comment? it really isn't adding a lot of context here.

packages/front-end/components/FactTables/FactMetricList.tsx
```diff
@@ -79,7 +95,14 @@ function FactMetricRowMenu({
             Edit
           </DropdownMenuItem>
         )}
-        {canDuplicate && (
+        {canShowDisabledEdit && (
```
ahdriel: need to also remove the double DropdownMenuitem

packages/front-end/pages/fact-metrics/[fmid].tsx
Comment (ahdriel) [reply packages/front-end/pages/fact-metrics/[fmid].tsx]: Nice! much better.

`ui/Tooltip` also has the `enabled` flag, so you can pass `editViaApiOnly` to it and wrap it unconditionally around the text.

One thing to note is that the same fix needs to be applied to `FactMetricList` & also `MetricList` and their Edit & Duplicate buttons.

lukesonnet:
@cursor review

packages/front-end/pages/fact-metrics/[fmid].tsx
Comment (lukesonnet) [reply packages/front-end/pages/fact-metrics/[fmid].tsx]: Fixed in all places and standardized a little bit with the new component, thanks for the heads up.

packages/front-end/components/FactTables/FactMetricList.tsx
lukesonnet: Similarly consolidated into one item as on the other pages, thanks.

PR #5796, opened 2026-05-04
docs: add Command Palette page

csbailey5t:
Document the Cmd+K / Ctrl+K command palette: how to open it, what's searchable across pages, features, experiments, metrics, saved groups, dashboards, docs, and the REST API reference, how strict vs. fuzzy matching works, and the in-palette keyboard controls. Add a top-level sidebar entry below Quick Start and a tip in the Quick Start guide that links to the new page.


jdorn:
I don't think this should be in the sidebar at all, especially not as prominent as it is in this PR.  I imagine the entry point is a little alert banner on the : `Tip: Use Command+K to quickly navigate GrowthBook. <Learn More>`.  Or if we have a Best Practices or Power Users section in the future, it could live under that.

csbailey5t:
@jdorn I'll pull it out of the sidebar, and leave the entry in the quick start page tip.

jdorn:
There's a bug with our docs where if a page isn't in the sidenav, the sidenav disappears when you load the page.  Example: https://docs-hw7mlxc5p-growthbook.vercel.app/command-palette

csbailey5t:
> There's a bug with our docs where if a page isn't in the sidenav, the sidenav disappears when you load the page. Example: https://docs-hw7mlxc5p-growthbook.vercel.app/command-palette

Interesting. In the meantime, I added a sidebar link under Knowledge Base, next to Glossary, for these Docs so we can get them in.

Kevin-Chant (approved):
Thanks for the contribution! I think the current placement is subtle enough to address Jeremy's concerns so I'll go ahead and merge this

PR #5736, opened 2026-04-24
Ty/pr2 major tables

teresayung:
### Features and Changes

This PR migrates high-traffic **list and table views** to the shared **`@/ui/Table`** pattern (list variant, **sticky headers**, **rounded corners** where it fits) and wires sortable columns to **`SortableTableColumnHeader`** from **`useSearch`** / **`useExperimentSearch`** instead of legacy **`SortableTH`** + raw `<table>` on the migrated screens.

The branch is **merged with latest `main`**; conflict resolution kept the new table primitives while taking **`main`** improvements where they mattered (for example **environments** row actions use the Radix **`DropdownMenu`** pattern, **bandits** search state uses **`localStorageKey: "bandits-page"`**, **features** hides usage/stale UI for **archived** features, and the **environment toggle** confirmation uses **`ModalStandard`** from `@/ui/Modal/Patterns/ModalStandard` instead of the legacy modal).

**Front-end pages and components (representative):**

- **Experiments** — `experiments/index` and related list/table wiring; tabbed lists use **`SortableTableColumnHeader`** consistently where search drives the table.
- **Features** — `features/index` and **`FeaturesDraftTable`** (search, sort, project badges, revision/draft flows); long feature keys use **`TruncateMiddleWithTooltip`**; environment toggle confirmation uses **`ModalStandard`**.
- **Bandits, dimensions, environments, fact-tables, attributes** — index list tables moved to **`@/ui/Table`** with aligned sort/search behavior (**`NextLink`** where needed for Radix menus, e.g. dimensions).
- **Metrics** — **`MetricsList`** / **`MetricGroupsList`** (and metrics index wiring); row actions use **`MetricRowMenu`** (dropdown actions aligned with the newer menu pattern).
- **Tags** — **`Tag`** / **`SortedTags`** adjustments for dense table cells (truncation / layout, including **`maxChars`** / **`maxWidth`** where useful).
- **UI** — **`TruncateMiddleWithTooltip`** for long identifiers in tables.
- **Styling** — table-related rules updated (**`global.scss`**, **`global-radix-overrides.scss`**, **`theme.scss`**) so list tables match the design system and Radix layout.
- **Supporting** — **`search.tsx`** exposes **`SortableTableColumnHeader`** for table headers; small **`RadixTheme`** / **`features.ts`** (and related) tweaks as needed for migrated screens and merged **`main`** behavior.

[...]

PR #5808, opened 2026-05-06
Ty/pr2 major tables

teresayung:
Remerging after fixes to feature table 
https://github.com/growthbook/growthbook/pull/5736

Old
image

New 
image

**The columns for environment now use read only icons and Changes doesn't show version number**


packages/front-end/services/features.ts
```diff
@@ -981,11 +981,13 @@ export function getUnreachableRuleIndex(
       continue;
     }
 
-    // Only force rules and 100%-coverage rollouts consume all traffic
-    const isFullCoverage =
-      rule.type === "force" || (rule.type === "rollout" && rule.coverage >= 1);
-    if (!isFullCoverage) continue;
+    // Skip non-force rules (require a non-null hash attribute, so may not match)
```
bryce-fitzsimons: what does this change have to do with table UI?

packages/front-end/services/RadixTheme.tsx
```diff
@@ -8,14 +8,13 @@ interface RadixThemeProps extends PropsWithChildren {
 
 export const RadixTheme: FC<RadixThemeProps> = ({ children, flip }) => {
   const { theme } = useAppearanceUITheme();
-  const computedTheme = flip ? (theme === "dark" ? "light" : "dark") : theme;
+  // When flip is false/undefined, pass no appearance so Theme inherits from parent
```
bryce-fitzsimons: do we need this?

packages/front-end/styles/global-radix-overrides.scss
```diff
@@ -128,7 +128,7 @@
     }
     .rdp-day_button {
```
bryce-fitzsimons, line 129: changes in this file do not look intentional

packages/front-end/styles/global.scss
```diff
@@ -609,25 +609,6 @@ pre {
     }
   }
 }
-.feature-tabs {
```
bryce-fitzsimons, line 612: huh?

packages/front-end/services/features.ts
```diff
@@ -1470,9 +1472,11 @@ export function getExperimentDefinitionFromFeature(
 export function useRealtimeData(
   features: FeatureInterface[] = [],
   mock = false,
+  enabled = true,
 ): { usage: FeatureUsageRecords; usageDomain: [number, number] } {
   const { data, mutate } = useApi<{ usage: FeatureUsageRecords }>(
     `/usage/features`,
+    { shouldRun: () => enabled },
```
bryce-fitzsimons: Intentional?

packages/front-end/pages/features/index.tsx
```diff
@@ -475,28 +587,47 @@ export default function FeaturesPage() {
...
+            setFeatureToDuplicate(null);
           }}
-          featureToDuplicate={featureToDuplicate || undefined}
         />
       )}
-
-      <div className="row my-3">
-        <div className="col">
+      {featureToToggleStaleDetection && (
+        <StaleDetectionModal
```
bryce-fitzsimons: i don't think this is supposed to be in here

packages/front-end/services/features.ts
teresayung: Good catch — reverted in latest commit. That was an unintentional logic change unrelated to the table UI work.

packages/front-end/services/RadixTheme.tsx
teresayung: This is an intentional cleanup — it avoids spreading the `appearance` prop when `flip` is not set, so the Theme inherits appearance from its parent instead of always overriding it. It's related to making the dark/light theme behave correctly in contexts where the Table component renders inside nested Themes.

packages/front-end/styles/global-radix-overrides.scss
teresayung: Reverted in latest commit. The calendar border-radius (rdp-day_button), multiline-item padding, and dropdown-trigger-select-style changes were unintentional. Also restored the .dropdown-trigger-header class since it's still referenced in RevisionDropdown and RampScheduleSection.

packages/front-end/styles/global.scss
teresayung: Reverted in latest commit. The .feature-tabs, .chevron-right, .draft-selector-collapsible-trigger, and .draft-selector-collapsible-content blocks were unintentionally deleted — they're still used in FeaturesHeader, FeaturesOverview, and DraftSelectorForChanges.

packages/front-end/services/features.ts
teresayung: Yes, intentional. Adds an `enabled` flag to `useRealtimeData` so the `/usage/features` API is only called when the realtime graphs are actually visible (gated by the feature flag). Avoids unnecessary polling when graphs are hidden.

packages/front-end/pages/features/index.tsx
teresayung: This is intentional — it's wired to the Stale column in the table. When a user clicks the disable button in the stale cell (`onDisable` callback), it calls `setFeatureToToggleStaleDetection(feature)` which opens this modal. Allows disabling stale detection directly from the feature list.

teresayung: reviewed again, removed to reflect master

packages/front-end/services/features.ts
teresayung: reverted

bryce-fitzsimons (approved):

PR #5810, opened 2026-05-06
fix: attribute archive/unarchive fails with null optional fields in DB

madhuchavva:
## Summary

Archiving or unarchiving a targeting attribute was returning a 400 error for orgs that have attributes with `null` stored for optional fields (`hashAttribute`, `format`, `enum`, etc.) in the DB.

The archive/unarchive handler in `attributes.tsx` was passing raw DB values directly into the PUT request body. Zod's `.optional()` accepts `undefined` but rejects `null`, causing the validation failure. The fix in `[property].tsx` (description save) already guarded against this, but the archive toggle in the list page did not.

## Changes

**`packages/front-end/pages/attributes.tsx`**
Uses `?? undefined` to coerce any `null` DB values to `undefined` before building the PUT payload, matching the existing pattern already present in `[property].tsx`. As a bonus, this self-heals the DB: when `undefined` flows through the controller spread, MongoDB omits the field on the next write, removing the stored `null`.

**`packages/back-end/src/routers/attributes/attributes.router.ts`**
Changes `.optional()` to `.nullish()` on all optional fields in the POST and PUT route validators. Defense-in-depth — any caller (scripts, API clients) that sends `null` for an optional field now gets accepted instead of a 400. Also adds the previously unvalidated `disableEqualityConditions` field to both schemas to match what the controller already handles.

## Root cause

`null` values enter the DB from attributes created before strict Zod validation was added to these endpoints. Newer attributes created via the UI always have `undefined` for unset fields, so this only surfaces for older orgs.


madhuchavva:
@greptileai

madhuchavva:
Closing this PR - as the fix was covered in https://github.com/growthbook/growthbook/pull/5819

PR #5813, opened 2026-05-06
feat(experiment): Add precomputed unit dimension time series

ahdriel:
### Features and Changes

Adds experiment-level `precomputedUnitDimensionIds` so selected unit dimensions are computed automatically on refresh and persisted to `MetricTimeSeries`.

### Screenshots

Screenshot 2026-05-21 at 9 50 12 AM
Screenshot 2026-05-21 at 9 50 01 AM

### Testing

- New unit tests


ahdriel:
cursor review

packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -903,11 +954,34 @@ const AnalysisForm: FC<{
                   lazyRender={true}
                 >
                   <div className="rounded px-3 pt-3 pb-1 bg-highlight">
+                    <div className="form-group mb-2">
+                      <UITooltip
+                        enabled={!hasEligiblePrecomputedUnitDimensions}
+                        content={precomputedUnitDimensionDisabledReason}
+                      >
+                        <div>
+                          <MultiSelectField
```
msamper: Do we want to surface these pre-computed unit dimensions in the Analysis Settings block on the experiment overview or is it enough to have it in the pre-computed section in the dimension dropdown?

msamper (changes requested):
Tested and found what seems like a small bug. After running a new snapshot if I click on the pre-computed unit dimension in the dimension chooser it seems to kick off a new query. Checked with Claude and it seems like we may need to update in `RefreshResultsButton` and `RefreshSnapshotButton` to recognize `experiment.precomputedUnitDimensionIds`. Right now it's only recognizing dimensions with the `precomputed:` prefix and triggering a query for `precomputedUnitDimensionIds`.

lukesonnet (commented):
Based on customer who requested this: we can gate this functionality to only if some pipeline mode is available. We should point the unit dimension queries to use the ephemeral table if it exists as a new exploratory query for that pipeline type, and leave incremental refresh as is. This might require some way to manage that ephemeral table's name that doesn't exist, but I think it's worth it.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1224,6 +1269,17 @@ export async function postExperiments(
   try {
     validateVariationIds(obj.variations);
 
+    if (data.precomputedUnitDimensionIds !== undefined) {
+      await assertExperimentPrecomputedUnitDimensionIdsAreValid({
+        context,
+        datasource,
+        exposureQueryId:
+          data.exposureQueryId ||
+          datasource?.settings.queries?.exposure?.[0]?.id,
```
msamper, line 1303: Why do we fallback to the first datasource exposure query id here but no in the REST api `postExperiment`?

msamper (approved):
Code looks good from what I can tell, but I'm not able to test since I don't have a datasource with pipeline mode configured using the ephemeral strategy. Maybe worth setting that up since we're doing a lot of incremental refresh work, but for now might be worth having someone else with the setup test this PR

lukesonnet (approved):

PR #5819, opened 2026-05-07
Fix nulls when archiving attributes

Kevin-Chant:
### Features and Changes

Since #5251 we've been sending optional fields explicitly as part of the PUT /attributes endpoint, but when not present on the attribute they're stored as `null` in mongo (as it's assumed to be an `$unset`). Then on the next round trip, those null fields hit the zod validator which has `.optional()` but not `.nullish()`, failing validation.

The fix is to only pass the relevant fields in the frontend, and to store only the fields that were part of the request body in the backend. This way an attribute not being changed won't lead to it being nulled, and even if it's already been nulled by this bug, the frontend won't pass it unless it's changing (and will correctly be revalidated).

### Testing

1. On `main`, archive an attribute (works) and then attempt to unarchive it (errors)
2. Use the network tab to inspect the `/organization` response and see that `attributeSchema` now has `null` for most fields on the archived attribute
3. Swap to this branch and unarchive the attribute (now works)
4. Archive a fresh attribute and inspect `/organization` again to verify that it didn't add the nulls to this one

### DIFF
--- a/packages/back-end/src/routers/attributes/attributes.controller.ts
+++ b/packages/back-end/src/routers/attributes/attributes.controller.ts
@@ -84,10 +84,16 @@ export const putAttribute = async (
   }
 
   const existing = attributeSchema[index];
+  // Only pass `projects` when the client actually sent it — passing
+  // `{ projects: undefined }` would be interpreted as a request to scope the
+  // attribute globally and incorrectly deny project-scoped users.
   if (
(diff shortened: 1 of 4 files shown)


Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

ahdriel (approved):

PR #5821, opened 2026-05-07
fix: avoid prompt validation errors in general settings

### LINKED ISSUE #5053: [Bug] Making Changes to the General settings throws an error

Sambhram1:
## Summary
- stop the AI prompt subform from becoming dirty during initial settings load
- skip undefined prompt values when building the \/ai/prompts\ save payload

## Why
Saving non-AI general settings could still trigger the AI prompt save path with undefined prompt values, which caused validation errors like:

\[prompts.0.prompt] Invalid input: expected string, received undefined\

The settings change still persisted, but the UI showed an error during save.

Fixes #5053

### DIFF
--- a/packages/front-end/components/GeneralSettings/AISettings.tsx
+++ b/packages/front-end/components/GeneralSettings/AISettings.tsx
@@ -248,13 +248,13 @@ export default function AISettings({
   useEffect(() => {
     if (data) {
       const prompts = getPrompts(data);
-      prompts.forEach((prompt) => {
-        promptForm.setValue(prompt.promptType, prompt.promptValue);
-        promptForm.setValue(
-          `${prompt.promptType}-model`,
-          prompt.overrideModel || "",
-        );
-      });
+      promptForm.reset(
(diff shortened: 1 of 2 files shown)


ahdriel:
Thanks for your contribution!

I am going to close this as I believe #5070 has fixed the issue linked (our bad on forgeting to close the issue) & I am unable to reproduce the error from the issue on latest main.

If you have repro steps that would be awesome and then we can reopen this PR.

PR #5833, opened 2026-05-08
fix(incremental-refresh): force standard runner on full refresh even with dimension selected

nadapzy:
If the user clicks **Full Refresh** while a dimension is selected, `postSnapshot` sends `?force=true` (→ `useCache=false`) **and** `dimension` in the body. `getSnapshotType` sees the dimension and returns `"exploratory"`, which routes to the exploratory runner — that runner only *reads* existing materialized sources and never rebuilds them, so the force flag is silently discarded and the pipeline isn't actually refreshed.

Gate the `dimension → exploratory` branch on `useCache`. A forced full refresh now always hits the standard runner that rebuilds the incremental source tables.

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1697,13 +1697,20 @@ function getSnapshotType({
   experiment,
   dimension,
   phaseIndex,
+  useCache,
 }: {
   experiment: ExperimentInterface;
   dimension: string | undefined;
   phaseIndex: number;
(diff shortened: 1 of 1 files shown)


lukesonnet:
I would rather disable a full refresh via dimension altogether, or at least make some messaging more clear. Let's discuss this.

lukesonnet:
Closed in favor of #6114 #6120 and a few other related PRs around fall back to full refresh.

PR #5834, opened 2026-05-08
fix(incremental-refresh): chunk metrics by percentile-cap count too

nadapzy:
The `__capValue` CTE in the incremental-refresh statistics query emits one approximate-quantile aggregate per percentile-capped column into a **single output row**. BigQuery limits the intermediate aggregation state for a row to 100MB, so a chunk with many percentile-capped metrics can fail with `Aggregation state exceeds 100 MB` even though it's well under the column-count limit `chunkMetrics` already budgets for.

Adds a second budget to `chunkMetrics`: `maxPercentileCapsPerChunk` (default 25; ratio metrics count double for numerator + denominator caps). When adding a metric would exceed either the column budget, the percentile-cap budget, or `MAX_METRICS_PER_QUERY`, start a new chunk. Reuses the existing chunking machinery instead of splitting the `__capValue` CTE.

Also guards against emitting an empty leading chunk if the very first metric already trips a budget.

Existing `chunkMetrics` tests still pass; new param is optional so callers are unchanged.

### DIFF
--- a/packages/back-end/src/services/experimentQueries/experimentQueries.ts
+++ b/packages/back-end/src/services/experimentQueries/experimentQueries.ts
@@ -174,10 +174,31 @@ export function maxColumnsNeededForMetric({
   }
 }
 
+// Each percentile-capped metric contributes one APPROX_QUANTILES (or
+// equivalent) aggregate per cap column to the single-row __capValue CTE.
+// On BigQuery the intermediate aggregation state for that row is bounded at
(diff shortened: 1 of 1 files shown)


nadapzy:
Closing in favor of #5836, which fixes the same 100MB row-limit issue by reshaping the `__capValue` CTE (UNPIVOT → GROUP BY → PIVOT) rather than splitting into more chunks — cleaner and avoids the extra query round-trips this approach would add.

PR #5836, opened 2026-05-08
fix(bigquery): reshape __capValue CTE to avoid 100MB row limit with many percentile-capped metrics

johnham-ant:
## Problem

The fact-metric stats query's `__capValue` CTE computes one `APPROX_QUANTILES(col, 10000)` per percentile-capped metric in a single ungrouped SELECT. With many capped metrics on one factTable and a large experiment, the intermediate aggregation row exceeds BigQuery's 100MB-per-row limit and the query fails with `Cannot query rows larger than 100MB limit`.

## Why the existing splitter doesn't catch it

`chunkMetrics()` budgets by **output column count**. A percentile-capped mean emits ~6-12 output cols (sum, sum_sq, cap, uncapped_sum, uncapped_sum_sq, + CUPED covariates), all 8-byte floats. The budget doesn't see the ~1.5MB **intermediate** sketch each `APPROX_QUANTILES` call produces before extracting the 8-byte cap value. So 65 capped metrics looks like ~650 cols (< 1000), stays in one query, and fails.

## Repro context

| | |
|---|---|
| Experiment units | ~41M |
| Percentile-capped metric columns on one factTable | 65 |
| Measured break point | 64× `APPROX_QUANTILES(col, 10000)` in one ungrouped SELECT over ~41M rows passes; **65 fails** |
| Derived | ~1.54 MB intermediate state per sketch at ~41M input rows |

## Fix

Reshape `__capValue` for BigQuery: UNPIVOT the value columns → `GROUP BY col_name` so each aggregation row holds **one** sketch (~1.5MB, always safe) → PIVOT the extracted scalar caps back to wide. Downstream `CROSS JOIN __capValue` unchanged.

Implemented as a BigQuery-dialect override of `percentileCapSelectClause` — other dialects keep the wide form. Single-column calls (legacy-metric path, metric-analysis numerator-only) fall back to the wide form to avoid pointless UNPIVOT overhead. Per-column `percentile` is handled by a CASE-driven OFFSET into the per-group sketch; per-column `ignoreZeros` by nulling zeros inside the aggregate argument for opted-in columns.

## Test results (empirical, 40.5M-unit input)

| N capped cols | Wide (current) | Long (this PR) |
|---|---|---|
| 64 | ✅ 560s | — |
| 65 | ❌ 100MB | — |
| 70 | ❌ 100MB | ✅ 17s, 2.15 slot-hrs |
| 100 | — | ✅ 20s, 3.13 slot-hrs |
| 200 | — | ✅ 34s, 4.84 slot-hrs |

Long-form is ~20× faster wall-clock and scales to any N and any unit count. ~2-5× more slot-hrs (parallelized GROUP BY).

## Alternative considered

Bump `maxColumnsNeededForMetric()` by +15 per percentile-cap column (1.5MB ÷ ~100KB/column-unit). Smaller diff (~5 lines), uses existing chunker. Covers up to ~1B units. Offered as a fallback if reshape is too invasive.


lukesonnet:
@cursor review

lukesonnet (approved):

PR #5837, opened 2026-05-09
Beta api for revsion on saved groups

gazzdingo:
### Features and Changes


packages/back-end/src/api/features/features.v2.router.ts
```diff
@@ -37,6 +37,9 @@ export const featureV2Routes: OpenApiRoute[] = [
   // Feature CRUD
   listFeaturesV2,
   postFeatureV2,
+  // MUST precede getFeatureV2: GET /features/revisions would otherwise be
+  // matched by GET /features/:id with `id = "revisions"`.
```
bryce-fitzsimons: this seems problematic. we don't want "revisions" to collide with the features/:id param. Might need separate namespacing here

packages/back-end/src/api/saved-groups/saved-groups.router.ts
```diff
@@ -5,10 +5,61 @@ import { postSavedGroup } from "./postSavedGroup";
...
+import { putSavedGroupRevisionArchive } from "./putSavedGroupRevisionArchive";
+import { postSavedGroupRevisionItemsAdd } from "./postSavedGroupRevisionItemsAdd";
+import { postSavedGroupRevisionItemsRemove } from "./postSavedGroupRevisionItemsRemove";
+
 export const savedGroupsRoutes: OpenApiRoute[] = [
+  // Saved-group CRUD
   listSavedGroups,
   postSavedGroup,
+  // MUST precede getSavedGroup: GET /saved-groups/revisions would otherwise be
+  // matched by GET /saved-groups/:id with `id = "revisions"`.
```
bryce-fitzsimons: same problem here as with features.router.ts. Use a separate namespace/stem

packages/back-end/src/api/features/features.v2.router.ts
bryce-fitzsimons: We could either do a big change on v2 route sonly (which is technically ok since its still beta, although a bit messy):
- `feature/:id/` - everything specific to a feature
- `features/...` - global feature actions
- `GET features/revisions` - the list endpoint

Or we could do something like:
- `feature-revisions/`

packages/back-end/src/api/saved-groups/postSavedGroupRevisionRevert.ts
```diff
@@ -0,0 +1,173 @@
...
+import { getAdapter } from "back-end/src/revisions";
+import {
+  applyPatchToSnapshot,
+  createOrUpdateRevision,
+  ensureLiveRevisionExists,
+} from "back-end/src/revisions/util";
+import { loadRevisionByVersion } from "./validations";
+import { toApiSavedGroupRevision } from "./toApiSavedGroupRevision";
+
+const REVERTABLE_FIELDS = [
```
bryce-fitzsimons: remove this in favor of zod validator + .pick()

packages/back-end/src/api/saved-groups/postSavedGroupRevisionPublish.ts
```diff
@@ -0,0 +1,130 @@
...
+    normalizeProposedChanges(revision.target.proposedChanges),
+  );
+  if (!conflictResult.success) {
+    throw new ConflictError(
+      "Merge conflicts exist — rebase before publishing",
+      conflictResult.conflicts,
+    );
+  }
+
+  const updatableFields = adapter.getUpdatableFields();
```
bryce-fitzsimons, line 99: we probably shouldn't have this either (from the previous PR you merged). `saved-group.adapter.ts` has
```
const UPDATABLE_FIELDS = new Set<string>([
  "groupName",
  "owner",
  "values",
  "condition",
  "attributeKey",
  "description",
  "projects",
  "useEmptyListGroup",
  "archived",
]);
```

But we could use the same sort of zod validator + .pick() strategy that I mentioned on the revert endpoint. The goal is to keep things dynamic so we don't need to update a bunch of consts in different places when extending saved groups in the future.

bryce-fitzsimons:
A few more comments from Claude:

  1. toApiSavedGroupRevisions email batching comment is wrong — diff lines 4933–4934

  The comment says "single email lookup for the page" but each call to toApiSavedGroupRevision runs its own resolveOwnerEmails, so a 25-item page makes 25 DB calls. Either fix the implementation to actually batch,
   or remove the misleading comment.

  ---
  2. postSavedGroupRevisionDiscard uses stale snapshot for permission check — diff line 3493

  The non-author permission check runs against revision.target.snapshot instead of the live savedGroup. If the group was moved to a different project between draft creation and now, the wrong permission is
  evaluated. Every other handler uses the live entity — this one should too.

  ---
  3. postSavedGroupRevision drops the comment field — diff line 3440

  Only req.body.title is passed to createOrUpdateRevision; any comment the caller provides is silently ignored. The comment in the code even mentions title / comment being dropped as the reason for
  forceCreate=true, making this a straightforward omission.

gazzdingo:
@greptileai

gazzdingo:
@greptileai

gazzdingo:
@greptileai

gazzdingo:
@greptileai

gazzdingo:
@greptileai

gazzdingo:
@greptileai

bryce-fitzsimons (approved):

PR #5845, opened 2026-05-11
drag and drop dashboards

mknowlton89:
### Features and Changes

Replaces the dashboard editor's vertical "move up / move down" controls with a true drag-and-drop + resize grid powered by `react-grid-layout`. Users can now arrange blocks on a 24-column grid, place blocks side-by-side, and resize them via a corner handle while editing. View/share mode renders the same layout as a static grid.

- **Layout persistence (shared validators + back-end model)**
  - Adds `BlockLayout` (`x`, `y`, `w`, `h`, optional `static`) to each dashboard block and an optional per-dashboard `grid` config (`cols`, `rowHeight`).
  - Introduces a single source of truth for per-type sizing in `DEFAULT_BLOCK_SIZE_BY_TYPE` plus a hard cap on the canonical column count (`DASHBOARD_GRID_COLS = 24`). `minW/minH` are runtime-only and intentionally not persisted, so tweaks to defaults automatically apply to existing dashboards.
  - Block height is intentionally uncapped — users can grow a block as tall as they need.
  - `DashboardModel` always runs blocks through a new `normalizeLayouts` helper on read, create, update, and migration — clamping `w/x/y` to grid bounds and auto-stacking any block missing a layout so persisted state can never overlap or overflow the grid horizontally.
- **24-column grid**
  - Per-type `minW` defaults are tuned against 24 cols (markdown = 4, explorers = 8, experiment blocks = 12), so the same fractional-width minimums (1/6, 1/3, 1/2) carry over from the original 12-col design but at twice the resolution.
  - `dashboardGridConfig.cols` now caps at `DASHBOARD_GRID_COLS` to stay in sync with the canonical column count.
- **Editor (`DashboardEditor`)**
  - New `DashboardGrid` wraps blocks in `ResponsiveGridLayout`. All breakpoints share the canonical 24-column count to prevent RGL's smaller-breakpoint compactor from silently reordering saved x positions when the container narrows (e.g. when the editing drawer opens).
  - Persistence is bound to `onDragStop` / `onResizeStop` only — `onLayoutChange` is deliberately not a save trigger because RGL fires it on mount after compaction, which would overwrite saved positions on every page open.
  - Drag/resize are disabled while a staged block is being added/edited so blocks can't shuffle mid-edit.
  - When edit interaction is disabled, the resize handle stays visible-but-dimmed and a Radix tooltip (`"Close the edit panel to resize this block"`) is anchored to a transparent overlay in the bottom-right corner. The tooltip renders through a portal so it escapes the grid's stacking context and isn't clipped by the edit drawer.
  - Switches header text to the design-system `@/ui/Heading` and `@/ui/Text` wrappers (drops the lint suppression for Radix imports).
- **Block UI (`DashboardBlock`)**
  - Removes the move-up/move-down dropdown and replaces it with a six-dot drag handle (`.dashboard-block-drag-handle`) in the block toolbar. Click is swallowed so it doesn't toggle title-edit, but mousedown bubbles so RGL can start the drag.
  - Rewrites `scrollToBlock` to use `getBoundingClientRect` math instead of `offsetTop`, because RGL positions blocks via `transform`, which leaves `offsetTop` at 0.
- **Workspace (`DashboardWorkspace`)**
  - Replaces `moveBlock` with `updateLayout`, which diffs incoming RGL layouts against stored blocks and only submits when something actually changed.
  - Duplicating a block now strips its `layout` so the copy stacks below instead of overlapping the source.
  - Opens the editing drawer automatically whenever a block becomes staged (duplicate previously could leave a staged block with no visible way to commit it).
- **Chart resize fix (`ExplorerChart`)**: ECharts only auto-resizes on window resize, so a `ResizeObserver` on the chart wrapper now calls `chart.resize()` whenever the surrounding grid cell or drawer changes width.
- **Misc bug fixes pulled in along the way**
  - `MetricTabContent`: allow adding the first metric to a `bigNumber` chart (previously the "Add metric" button was disabled when `chartType === "bigNumber"`, so a new big-number block had no way to ever get a metric).
  - `ui/Text`: stop emitting an inline `white-space` style when `truncate` is set so Radix's `.rt-truncate` (which sets `white-space: nowrap`) actually wins.
  - `ProductAnalyticsExplorerBlock`: drop the hard-coded `height: 500` and let the block fill its grid cell.
- **Dependencies**: adds `react-grid-layout` + `@types/react-grid-layout`, imports its base stylesheet in `_app.tsx`, and adds dashboard-specific overrides in `global.scss` (violet-tinted placeholder, hidden resize handles outside edit mode, dimmed handle + overlay-driven tooltip when interaction is disabled, full-height block container).
- **OpenAPI**: regenerated `packages/back-end/generated/spec.yaml` to pick up the new `layout` field, `grid` config, and updated `w`/`x` bounds (now 24) with no `h` ceiling.

### Dependencies

None.

### Testing

- [x] Open an existing experiment dashboard — verify all blocks render in their previous order with no visual regressions.
- [x] Enter edit mode and confirm a six-dot drag handle appears in the top-left of each block; a corner resize handle appears on the bottom-right.
- [x] Drag a block to a new position (including side-by-side with another block) — release and confirm the new position persists after a page reload.
- [x] Resize a block both wider/narrower and taller/shorter — confirm width bounds are respected (can't exceed 24 cols, can't go below per-type `minW`), height has no upper bound, and the new size persists after reload.
- [x] Open the editing drawer for a block — confirm the corner resize handle is visually dimmed, doesn't respond to clicks, and shows the tooltip "Close the edit panel to resize this block" above the handle (not clipped by the drawer).
- [x] Duplicate a block — confirm the copy stacks below the source instead of overlapping it, and that the editing drawer auto-opens.
- [x] Narrow the viewport / open the editing side panel — confirm side-by-side blocks keep their relative x positions (no reordering) and just shrink proportionally.
- [x] Resize a block containing an ECharts visualization (e.g. metric-explorer, fact-table-exploration) — confirm the chart redraws to fill the new cell size.
- [x] Create a new `bigNumber` metric-exploration block — confirm you can add exactly one metric and the "Add metric" button disables after that.
- [x] View a dashboard in share mode (logged out / read-only) — confirm the grid renders statically with no drag/resize affordances.
- [x] Hit the dashboards REST API (`GET /api/v1/dashboards/:id`) — confirm `layout` is present on every block and `grid` is returned when set.


packages/front-end/enterprise/components/Dashboards/DashboardEditor/index.tsx
```diff
@@ -147,6 +158,84 @@ export const isBlockTypeAllowed = (
...
+export function getGridKeyForBlock(
+  block: DashboardBlockInterfaceOrData<DashboardBlockInterface>,
+  index: number,
+): string {
+  return dashboardBlockHasIds(block) ? block.id : `__staged_block_${index}__`;
+}
+
+// Build the RGL LayoutItem[] from blocks. Blocks without an existing layout
+// get stacked full-width below the highest occupied y so they don't collide.
+// minW/minH come from the shared per-type defaults; maxW is the absolute grid
```
nhat-growthbook, line 172: Would be a nice feature to let them decide how many grid cols they wanna have. In case they wanna put the dashboard on a massive TV

packages/front-end/enterprise/components/ProductAnalytics/SideBar/MetricTabContent.tsx
```diff
@@ -168,7 +168,13 @@ export default function MetricTabContent() {
       <Button
         size="sm"
         variant="outline"
-        disabled={draftExploreState.chartType === "bigNumber"}
+        // Big-number charts can only display a single metric, so cap the
```
nhat-growthbook, line 171: Ohhh this will conflict with the comparing PR i gotta fix up. I changed it to support multiple. But please let me know if I shouldn't support multiple bigNumber

nhat-growthbook (commented):
LGTM

nhat-growthbook (approved):
Forgot to approve. LGTM, though I guess we don't want to merge till after 4.4?

PR #5846, opened 2026-05-11
better badge ui, confirmation step when setting to enabled + scheduled enable date. fix targeting condition save bug

bryce-fitzsimons:
- Warn on enabling a scheduled-to-enable rule
- Better badge status
- Fix targeting + simple schedule save bug

image

image


bryce-fitzsimons:
@cursor review

jdorn (approved):

PR #5848, opened 2026-05-11
Add Compare logic product analytics explorations

nhat-growthbook:
### Features and Changes
- Add a compare button to analytics charts to compare against same period previous
- 7 days -> against 14->7 days before
- Custom date -> Same period directly prior to select range
- Custom lookback --> same period directly prior to current

### Testing
- Enable Compare and run through all combination of date range and chart types

### Screenshots
Screenshot 2026-05-11 at 10 45 12


nhat-growthbook:
I don't like this approach. Will start a new one

PR #5687, opened 2026-04-17
Fix metric drilldown modal not refreshing with dimension breakdowns

lukesonnet:
### Features and Changes

Fixes a regression where the baseline variation index and difference chooser controls in the metric drilldown modal were not properly triggering new analyses when dimension breakdowns were active.

**Root Cause:**
In `MetricDrilldownModal.tsx`, when `dimensionInfo` was provided (dimension breakdown active), the component was always using `initialResults` (captured when modal opened) instead of getting the updated results from the refreshed `analysis` after baseline/difference changes.

```tsx
// Before (bug):
const results = dimensionInfo
  ? initialResults  // Always stale!
  : (analysis?.results?.[0] ?? initialResults);

// After (fix):
const results = dimensionInfo
  ? (analysis?.results?.[dimensionInfo.index] ?? initialResults)  // Uses fresh data
  : (analysis?.results?.[0] ?? initialResults);
```

This worked correctly on the main page because `BreakDownResults` rebuilds from `analysis.results` directly, but the modal had pinned itself to the initial snapshot.

### Dependencies

None

### Testing

1. Open an experiment with dimension breakdowns enabled
2. Click on a metric row in the breakdown table to open the metric drilldown modal
3. Change the baseline variation using the dropdown in the modal
4. Verify the data in the modal updates to reflect the new baseline
5. Change the difference type (relative/absolute) and verify it updates correctly
6. Confirm that the same controls work properly without dimension breakdowns (non-regression)

### Screenshots

This is a behavior fix for data refresh - the UI elements themselves remain unchanged.

[Slack Thread](https://growthbookapp.slack.com/archives/C07NK4F016Z/p1776459675227379?thread_ts=1776459675.227379&cid=C07NK4F016Z)

[...]

PR #5857, opened 2026-05-13
Fix metric drilldown modal not refreshing with dimension breakdowns in reports

lukesonnet:
### Features and Changes

Fixed the metric drilldown modal for **reports** where the difference type chooser and baseline variation selector would get stuck (appear to do nothing) when clicked.

**Root Cause:**

When the user clicks a different difference type inside the drilldown modal, `DifferenceTypeChooserChangeColumnLabel` does two things:

1. POSTs `/snapshot/${snapshot.id}/analysis` to add the new analysis to the snapshot — this works fine for reports, the backend already supports multiple analyses on a single snapshot.
2. Calls `mutate()` to refresh the local snapshot.

The bug was in step 2. `LocalSnapshotProvider.mutateSnapshot` was fetching `/experiment/${experiment.id}/snapshot/${phase}`, which returns the **experiment's current latest snapshot** — a different snapshot than the one the report is rendering. The local snapshot would be replaced with the wrong one (which doesn't contain the newly-added analysis), causing `getSnapshotAnalysis(localSnapshot, newSettings)` to return `undefined`, and leaving the UI stuck on the previous difference type.

**The Fix:**

Change `LocalSnapshotProvider.mutateSnapshot` to refresh by snapshot ID via `/snapshot/${snapshot.id}` instead of via experiment+phase. This is correct for both contexts:
- **Experiments:** same snapshot comes back, now containing the newly-added analysis.
- **Reports:** the report's specific snapshot comes back, now containing the newly-added analysis — allowing the report snapshot to accumulate multiple analyses over time.

Additionally, simplified the props threading: since the snapshot already carries its own `dimension`, and no drilldown child component reads `experiment` from `useSnapshot()`, only the `snapshot` prop needs to be passed down to `MetricDrilldownProvider`/`MetricDrilldownModal` for the report context (down from three props: `snapshot`, `experiment`, `dimension`).

- Related to: PR #5687 (original experiments drilldown fix)

### Dependencies

None

### Testing

1. Open a report (not an experiment) that has results
2. Open the metric drilldown modal by clicking a metric row
3. Click the difference type chooser (the "% Change" / "Absolute" / "Scaled" dropdown in the column header) and switch to a different type
4. Verify the results update — previously this would appear to do nothing or get stuck
5. Repeat with a report that has a dimension breakdown active and verify the modal still opens correctly

### Screenshots

Behavior fix — no visual UI changes.

[Slack Thread](https://growthbookapp.slack.com/archives/C07NK4F016Z/p1776459675227379?thread_ts=1776459675.227379&cid=C07NK4F016Z)

### DIFF
--- a/packages/front-end/components/Experiment/ExperimentResultTooltipContent/ExperimentResultTooltipContent.tsx
+++ b/packages/front-end/components/Experiment/ExperimentResultTooltipContent/ExperimentResultTooltipContent.tsx
@@ -166,7 +166,8 @@ export default function ExperimentResultTooltipContent({
     if (resultsStatus !== "draw" || minPercentChange === undefined) return null;
     return (
       <Text as="div" size="1" style={{ color: "var(--color-text-mid)" }}>
-        <b>Draw:</b> this occurs when the % Change is smaller than the
+        <b>Draw:</b>
+        {" this occurs when the % Change is smaller than the "}
         metric&apos;s min change ({percentFormatter.format(minPercentChange)})
       </Text>
     );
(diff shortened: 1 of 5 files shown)


ahdriel (approved):
Changes LGTM!

I think we might have some opportunity to create a `ReportSnapshotProvider` and rename `SnapshotProvider` to `ExperimentSnapshotProvider` to be clearer on the expectations.

But this would be a bigger refactor work and not a blocker at all -- so we should merge this as it fixes a real problem.

PR #5866, opened 2026-05-15
Fix athena top values query

jdorn:
### Features and Changes

One more fix from #5720 , this time for Athena/Presto


ahdriel (approved):
Changes LGTM!

Can we add a unit test to prevent regression?

PR #5869, opened 2026-05-15
fix(incremental): harden refresh lock - add fencing + heartbeat

nadapzy:
Builds on #5466, which added the per-experiment incremental-refresh lock. Three remaining gaps allowed concurrent runs to corrupt the shared `gb_units_<experimentId>` / `gb_metrics_<experimentId>_*` warehouse tables.

## Problem

1. **Exploratory reader unlocked.** `createSnapshotFromPlan` only acquires the lock for `runnerKind === "incremental"`. The `incremental-exploratory` runner reads the same per-experiment units / metric-source tables that the incremental runner DROP/CREATE/RENAMEs, so a scheduled auto-refresh overlapping with a dimension-slice analysis can observe a missing or empty units table and return empty results.
2. **Stale check is wall-clock, not liveness.** A lock was considered stale once `dateUpdated` was >1h old. Long-running refreshes legitimately exceed that, so a second trigger (auto-refresh cron + manual Update overlap) could steal the lock mid-run while the first holder was still issuing DDL on the shared tables.
3. **No fence at DDL time.** Even with a correct lock, if it *is* lost mid-run (stale takeover, cancel + restart), the original holder's already-queued DROP/RENAME on the shared tables would still execute and clobber the new holder's state.

## Fix

1. **Lock the exploratory reader.** Acquire the lock for both `incremental` and `incremental-exploratory` runner kinds; release it from the exploratory runner's `updateModel` on terminal status (mirroring the incremental runner). The `results` runner is unaffected — it writes only to per-snapshot ephemeral tables (`growthbook_tmp_units_<snapshotId>`) and shares no warehouse state across runs. This also serves as in-flight snapshot dedup for incremental pipeline mode: a second concurrent `createSnapshot` caller fails fast with `ConcurrentIncrementalRefreshError`, which the auto-refresh job already handles gracefully.
2. **Heartbeat-based staleness.** Add `lockHeartbeatAt` to `IncrementalRefresh`; `acquireLock` now treats a lock as stale only when the heartbeat (falling back to `dateUpdated` for legacy docs) is older than `INCREMENTAL_LOCK_STALE_MS` (10 min). Add `touchLockHeartbeat()` (fenced on `currentExecutionSnapshotId`) and call it from a new `QueryRunner.onHeartbeat()` hook driven by the existing 30s per-query heartbeat interval. Both incremental runner kinds override the hook.
3. **Execute-time fence on shared-table DDL.** Wrap the `run` callbacks for the units DROP/CREATE/RENAME and metric-source CREATE queries in a `fenced()` helper that re-reads `currentExecutionSnapshotId` immediately before executing the SQL and aborts with a clear error if this snapshot no longer holds the lock. Checked at execute time (not enqueue time) because all queries are enqueued up-front but run sequentially via dependencies. `releaseLock` is already fenced on `snapshotId`, so the terminal-status release is a safe no-op once the lock is lost.

### DIFF
--- a/packages/back-end/src/models/IncrementalRefreshModel.ts
+++ b/packages/back-end/src/models/IncrementalRefreshModel.ts
@@ -8,8 +8,11 @@ import { MakeModelClass } from "./BaseModel";
 
 export const COLLECTION_NAME = "incrementalrefresh";
 
-// If a lock hasn't been updated in this long, consider it stale
-const STALE_LOCK_TIMEOUT_MS = 60 * 60 * 1000;
+// A lock is considered stale once its heartbeat is older than this. The
+// runner refreshes the heartbeat every ~30s while queries are executing, so
+// this only needs to cover slow polls / brief stalls — not the full runtime
+// of a refresh.
+export const INCREMENTAL_LOCK_STALE_MS = 10 * 60 * 1000;
 
 const BaseClass = MakeModelClass({
   schema: incrementalRefreshValidator,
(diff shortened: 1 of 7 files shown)


lukesonnet:
@cursor review

ahdriel (approved):

PR #5876, opened 2026-05-16
fix(incremental): degrade gracefully when some metrics are incompatible with the pipeline

nadapzy:
## Problem

`validateIncrementalPipeline` throws on the first metric that fails any per-metric check (non-fact metric, ratio metric whose numerator/denominator span different fact tables, or event-quantile metric on a warehouse without KLL sketch support). `planSnapshotQueryRunner` catches that throw and routes the entire snapshot to the inline `"results"` runner. So a single incompatible metric — even one out of dozens — forfeits incremental refresh for *all* metrics on the experiment, with the reason only visible in an `info`-level server log.

## Fix

Partition instead of throw for per-metric checks:

- `validateIncrementalPipeline` now collects per-metric incompatibilities into a returned `incompatibleMetrics: {id, name, reason}[]` list. Experiment-wide gates (skipPartialData, activation metric, feature/datasource eligibility, settings-hash mismatch) still throw and fall back to inline as before. If *every* selected metric is incompatible it also still throws.
- Both incremental runners (`ExperimentIncrementalRefreshQueryRunner` and `ExperimentIncrementalRefreshExploratoryQueryRunner`) drop the incompatible metrics from `snapshotSettings.metricSettings` before building queries, and persist a non-fatal `warnings[]` entry on the snapshot listing the skipped metric names and reasons.
- Adds an optional `warnings?: string[]` field to `ExperimentSnapshotInterface` / mongoose schema / `SnapshotResult` to carry these messages. This is the same shape proposed in #5832 — whichever lands first, the other rebases trivially.

## Out of scope / follow-up

- **Hybrid mode** (compute the incompatible metrics via the inline path alongside the incremental run, so they still get results) — this PR only skips and warns; hybrid is the obvious next step.
- **UI surfacing** of `snapshot.warnings` — left for the front-end to render once the field exists (also needed by #5832).

### DIFF
--- a/packages/back-end/src/models/ExperimentSnapshotModel.ts
+++ b/packages/back-end/src/models/ExperimentSnapshotModel.ts
@@ -81,6 +81,7 @@ const experimentSnapshotSchema = new mongoose.Schema({
   dimension: String,
   unknownVariations: [String],
   multipleExposures: Number,
+  warnings: [String],
   hasCorrectedStats: Boolean,
   status: String,
   settings: {},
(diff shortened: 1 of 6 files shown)


ahdriel:
Closing this as it is not the highest lever at this moment and we don't plan on pursuing it. We can always reopen later if things change.

Issue #4917, opened 2025-12-01
[Feature] Incremental refresh support


natasha-growthbook:
### Description of Feature

Incremental refresh support

Priority might be the following:
1. Legacy metrics
2. Ratio metrics
3. Activation metrics
4. Numerators
For us to use this I would need first 3 so we would be able to use

Issue with using only fact tables is that we have metrics with loads of different parameters and for same metrics we add new parameters per test basis. That makes it difficult to always manage fact tables. While legacy metric if you created you just left to live with the previous experiment and no need to manage.
We could create event_params so we would be able to filter, but not sure if its the right approach

ahdriel:
We have implemented [Incremental Pipeline mode](https://docs.growthbook.io/app/data-pipeline#enable-incremental-refresh) -- but as mentioned we do not support Legacy Metrics in this flow.

PR #5882, opened 2026-05-17
Support cross-fact-table ratio metrics in incremental refresh

johnham-ant:
### Features and Changes

**Support cross-fact-table ratio metrics in incremental refresh (pipeline mode).**

#### The problem

A ratio metric whose numerator and denominator live in different fact tables (e.g. `revenue` from a `purchases` fact table divided by `$$count` from a `sessions` fact table) currently causes `validateIncrementalPipeline()` to throw:

> Ratio metrics must have the same numerator and denominator fact table with incremental refresh.

Because the validation is experiment-wide, a single cross-fact-table ratio metric disqualifies the **entire experiment** from the incremental query runner and forces it back to the inline path. On the inline path, every refresh re-scans all the raw fact-table data from the start of the experiment. Pipeline mode exists precisely to avoid that: it materializes per-user aggregates into cache tables (`gb_metrics_exp_*`) and only scans new data on each refresh. On a long-running experiment with large fact tables, falling back can make every refresh orders of magnitude more expensive, and it happens silently.

#### Why the limitation existed

In pipeline mode, each metric source group has exactly one cache table, built from exactly one fact table. The cache table schema (`metric-source-table-schema.ts`) stores a ratio metric's `{metric}_value` and `{metric}_denominator_value` columns side by side, and `getInsertMetricSourceDataQuery()` scans a single `__factTable` CTE. That assumes both sides of the ratio are sourced from the same fact table. There are explicit `TODO(incremental-refresh)` markers acknowledging this in `ExperimentIncrementalRefreshQueryRunner.ts` ("handle cross-table metrics") and `SqlIntegration.ts` ("generalize to multiple sources").

#### Why it's solvable without a redesign

The inline path already handles cross-fact-table ratios: `getExperimentFactMetricsQuery()` builds one `__factTable{N}` / `__userMetricAgg{N}` chain per fact table and joins the per-user aggregates on the unit id before computing statistics. The pipeline path already materializes exactly those per-user aggregates into cache tables. The only missing piece is: cache the denominator side in its own table, and join two cache tables in the statistics query instead of reading one.

#### What this PR does

1. **`validateIncrementalPipeline` (dataPipeline.ts)** — narrow the guard. Cross-fact-table ratio metrics no longer disqualify the experiment. The guard is kept for the CUPED (regression adjustment) case — see "Scope limitations" below.

2. **Metric source grouping (`ExperimentIncrementalRefreshQueryRunner.ts`)** — cross-fact-table ratio metrics are grouped by their fact-table pair (`{numeratorFT}_{denominatorFT}`), separate from same-FT metrics. A group that spans two fact tables now carries `denominatorFactTableId` and, once built, persists a `denominatorTableFullName` and `denominatorMaxTimestamp` on the metric source. Each fact table feeding the group is scanned and tracked independently, so one source lagging the other cannot cause dropped or double-counted rows.

3. **Cache table schema + CREATE + INSERT (`metric-source-table-schema.ts`, `SqlIntegration.ts`)** — `getCreateMetricSourceTableQuery()` and `getInsertMetricSourceDataQuery()` take an optional `factTableId` that identifies which side of the ratio a cache table represents. The numerator cache table only carries `{metric}_value` columns, the denominator cache table only `{metric}_denominator_value`, and each insert scans only its own fact table with its own incremental `max_timestamp`.

4. **Statistics query (`SqlIntegration.ts`)** — `getIncrementalRefreshStatisticsQuery()` accepts an optional `metricSourceDenominatorTableFullName`. When present, it adds a `__metricSourceDenominatorData` / `__metricDenominatorDataAggregated` CTE and LEFT JOINs it onto `__experimentUnits` alongside the numerator cache table on the unit id. Both joins COALESCE missing per-user values to 0, which matches the inline path's missing-user semantics (where every unit in `__distinctUsers` appears in both `__userMetricAgg` and `__userMetricAgg1` via LEFT JOINs and then gets COALESCEd in the stats CTE).

5. **Cache invalidation / schema migration** — `getMetricSettingsHashForIncrementalRefresh()` already hashes `factMetric.denominator` and the denominator fact table's SQL/filters, and the new `denominator*` fields on the metric source model are optional. Editing a ratio from same-FT to cross-FT (or vice versa) changes the settings hash and triggers the existing "configuration is outdated, run a Full Refresh" path rather than silently reading a stale cache. No migration of existing incremental refresh documents is needed.

#### Before / After (generated SQL, BigQuery)

Before this change, a cross-FT ratio metric could not reach pipeline mode at all. After this change, a `revenue (purchases) / $$count (sessions)` ratio produces two cache tables and one joined statistics query:

```sql
-- numerator cache
CREATE TABLE proj.ds.gb_metrics_exp_1_ft_purchases_ft_sessions_abc (
  user_id STRING,
  fact_revenue_per_session_value FLOAT64,
  refresh_timestamp TIMESTAMP, max_timestamp TIMESTAMP, metric_date DATE
) ...

-- denominator cache
CREATE TABLE proj.ds.gb_metrics_exp_1_ft_purchases_ft_sessions_abc_denom (
  user_id STRING,
  fact_revenue_per_session_denominator_value INT64,
  refresh_timestamp TIMESTAMP, max_timestamp TIMESTAMP, metric_date DATE
) ...
```

```sql
-- statistics (abridged)
WITH
  __metricSourceData AS (SELECT * FROM proj.ds.gb_metrics_..._abc),
  __metricSourceDenominatorData AS (SELECT * FROM proj.ds.gb_metrics_..._abc_denom),
  __experimentUnits AS (SELECT ... FROM proj.ds.gb_units_exp_1 e),
  __metricDataAggregated AS (
    SELECT user_id, SUM(COALESCE(umj.fact_revenue_per_session_value, 0)) AS fact_revenue_per_session_value
    FROM __metricSourceData umj GROUP BY user_id
  ),
  __metricDenominatorDataAggregated AS (
    SELECT user_id, SUM(COALESCE(umj.fact_revenue_per_session_denominator_value, 0)) AS fact_revenue_per_session_denominator_value
    FROM __metricSourceDenominatorData umj GROUP BY user_id
  ),
  __joinedData AS (
    SELECT u.user_id, u.variation,
      COALESCE(m.fact_revenue_per_session_value, 0) AS m0_value,
      COALESCE(md.fact_revenue_per_session_denominator_value, 0) AS m0_denominator
    FROM __experimentUnits u
    LEFT JOIN __metricDataAggregated m ON u.user_id = m.user_id
    LEFT JOIN __metricDenominatorDataAggregated md ON u.user_id = md.user_id
  )
SELECT m.variation, COUNT(*) AS users,
  SUM(COALESCE(m.m0_value, 0)) AS m0_main_sum,
  SUM(POWER(COALESCE(m.m0_value, 0), 2)) AS m0_main_sum_squares,
  SUM(COALESCE(m.m0_denominator, 0)) AS m0_denominator_sum,
  SUM(POWER(COALESCE(m.m0_denominator, 0), 2)) AS m0_denominator_sum_squares,
  SUM(COALESCE(m.m0_denominator, 0) * COALESCE(m.m0_value, 0)) AS m0_main_denominator_sum_product
FROM __joinedData m GROUP BY m.variation
```

Same-fact-table ratio metrics, non-ratio metrics, and quantile metrics generate the same SQL as before — see the "unchanged behavior" tests.

#### Scope limitations

- **CUPED cross-fact-table ratios remain unsupported in pipeline mode.** The pre-exposure covariate cache (`gb_metrics_*_covariate`) would also need to span two fact tables, which adds another full create/insert/maxTimestamp/join lifecycle. Rather than double the surface area here, the CUPED case keeps the validation guard with a clearer error message (disable CUPED for that metric, or analyze via a full query). This can be a follow-up.
- The denominator fact table is scanned once per group it participates in. If a metric with numerator FT-A also exists alongside a cross-FT ratio with denominator FT-A, FT-A is scanned twice — once for the same-FT group's cache table and once for the cross-FT group's denominator cache table. Both scans are still incremental (only new rows since the last `max_timestamp`), so the cost is bounded by new data, not total data.

Touches the second priority item in #4917 (ratio metric support for incremental refresh).

- Closes the `TODO(incremental-refresh): handle cross-table metrics` and `TODO(incremental-refresh): generalize to multiple sources` markers.

### Dependencies

None. The three new fields on `incrementalRefreshMetricSourceValidator` are optional, so existing incremental refresh documents remain valid.

### Testing

- Added a `describe("BigQuery cross-fact-table ratio metric incremental refresh SQL generation")` block to `packages/back-end/test/integrations/BigQuery.test.ts` covering: cache table schema split by `factTableId`, per-fact-table insert queries, the joined statistics query (and its error path when a denominator cache is missing), and "unchanged behavior" assertions for same-fact-table ratios.
- Added `packages/back-end/test/queryRunners/ExperimentIncrementalRefreshQueryRunner.test.ts` covering `getCrossFtDenominatorFactTableId()` and the cross-FT grouping logic in `getIncrementalRefreshMetricSources()`, including reuse of an existing group from a persisted metric source.
- `cd packages/back-end && pnpm test` — 96 suites / 3995 tests pass, including the 2259-test SQL integration snapshot suite with no snapshot changes.
- `cd packages/shared && pnpm test` — 26 suites / 701 tests pass.
- `pnpm type-check` clean in `back-end` and `shared`; `eslint` and `prettier` clean on changed files.

### Screenshots

No UI changes.

### DIFF
--- a/packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
+++ b/packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
@@ -2,6 +2,7 @@ import { tabulateCovariateImbalance } from "shared/health";
 import {
   ExperimentMetricInterface,
   isFactMetric,
+  isRatioMetric,
   isRegressionAdjusted,
   quantileMetricType,
 } from "shared/experiments";
@@ -88,6 +89,22 @@ export interface MetricSourceGroups {
   metrics: FactMetricInterface[];
 }
 
+// Returns the denominator fact table id for a metric whose numerator and
+// denominator live in different fact tables. Undefined for non-ratio metrics
+// and for ratio metrics where both sides share a fact table.
+export function getCrossFtDenominatorFactTableId(
+  metric: FactMetricInterface,
+): string | undefined {
+  if (
+    isRatioMetric(metric) &&
+    metric.denominator?.factTableId &&
+    metric.denominator.factTableId !== metric.numerator.factTableId
+  ) {
+    return metric.denominator.factTableId;
+  }
+  return undefined;
+}
+
 export function getIncrementalRefreshMetricSources({
   metrics,
   existingMetricSources,
@@ -102,14 +119,19 @@ export function getIncrementalRefreshMetricSources({
   metrics: FactMetricInterface[];
   groupId: string;
   factTableId: string;
+  denominatorFactTableId?: string;
 }[] {
   // TODO(incremental-refresh): skip partial data is currently ignored
   // TODO(incremental-refresh): error if no efficient percentiles
   // shouldn't be possible since we are unlikely to build incremental
   // refresh for mySQL
   const getMetricGroupKey = (metric: FactMetricInterface) => {
+    // Cross-fact-table ratio metrics get their own group keyed by the fact
+    // table pair so they can be cached in two tables (numerator + denominator)
+    // without expanding the schema of other groups' cache tables.
+    const crossFtDenominator = getCrossFtDenominatorFactTableId(metric);
     // Keep quantiles in their own source (similar to experimentQueries grouping)
-    return `${metric.numerator.factTableId}${quantileMetricType(metric) ? "_qtile" : ""}`;
+    return `${metric.numerator.factTableId}${crossFtDenominator ? `_${crossFtDenominator}` : ""}${quantileMetricType(metric) ? "_qtile" : ""}`;
   };
 
   const groups: Record<
@@ -117,6 +139,7 @@ export function getIncrementalRefreshMetricSources({
     {
       alreadyExists: boolean;
       factTableId: string;
+      denominatorFactTableId?: string;
       metrics: FactMetricInterface[];
     }
   > = {};
@@ -130,18 +153,20 @@ export function getIncrementalRefreshMetricSources({
       groups[existingGroup.groupId] = groups[existingGroup.groupId] || {
         alreadyExists: true,
         factTableId: existingGroup.factTableId,
+        denominatorFactTableId: existingGroup.denominatorFactTableId,
         metrics: [],
       };
       groups[existingGroup.groupId].metrics.push(metric);
       return;
     }
 
-    // TODO(incremental-refresh): handle cross-table metrics
     const factTableId = metric.numerator.factTableId;
+    const denominatorFactTableId = getCrossFtDenominatorFactTableId(metric);
     const groupKey = getMetricGroupKey(metric);
     groups[groupKey] = groups[groupKey] || {
       alreadyExists: false,
       factTableId,
+      denominatorFactTableId,
       metrics: [],
     };
     groups[groupKey].metrics.push(metric);
@@ -150,6 +175,7 @@ export function getIncrementalRefreshMetricSources({
   const finalGroups: {
     groupId: string;
     factTableId: string;
+    denominatorFactTableId?: string;
     metrics: FactMetricInterface[];
   }[] = [];
   Object.entries(groups).forEach(([groupId, group]) => {
@@ -157,6 +183,7 @@ export function getIncrementalRefreshMetricSources({
       finalGroups.push({
         groupId,
         factTableId: group.factTableId,
+        denominatorFactTableId: group.denominatorFactTableId,
         metrics: group.metrics,
       });
       return;
@@ -183,6 +210,7 @@ export function getIncrementalRefreshMetricSources({
       finalGroups.push({
         groupId: groupId + "_" + randomId + i,
         factTableId: group.factTableId,
+        denominatorFactTableId: group.denominatorFactTableId,
         metrics: chunk,
       });
     });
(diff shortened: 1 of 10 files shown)


packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -102,21 +119,27 @@ export function getIncrementalRefreshMetricSources({
...
   // shouldn't be possible since we are unlikely to build incremental
   // refresh for mySQL
   const getMetricGroupKey = (metric: FactMetricInterface) => {
+    // Cross-fact-table ratio metrics get their own group keyed by the fact
+    // table pair so they can be cached in two tables (numerator + denominator)
+    // without expanding the schema of other groups' cache tables.
+    const crossFtDenominator = getCrossFtDenominatorFactTableId(metric);
     // Keep quantiles in their own source (similar to experimentQueries grouping)
-    return `${metric.numerator.factTableId}${quantileMetricType(metric) ? "_qtile" : ""}`;
+    return `${metric.numerator.factTableId}${crossFtDenominator ? `_${crossFtDenominator}` : ""}${quantileMetricType(metric) ? "_qtile" : ""}`;
```
lukesonnet, line 134: The main alternative to putting these metrics in their own group and creating 2 metric sources for this group is:

Instead alllow cross-fact-table metrics to store their numerator and denominator data in their respective metric sources (e.g. bundled with non-cross fact table metrics from the numerator and denominator fact tables, respectively), and then just joining them in the stats query.

This would require a greater restructuring, but I think we should consider doing it early next week.

lukesonnet:
Thank you for this, closing in favor of #5908

PR #5883, opened 2026-05-17
OpenAPI validation improvements

jdorn:
### Features and Changes

Fix issues in the OpenAPI spec that prevent SDK/CLI generators from working.

1. Option to generate a version of the OpenAPI spec without deprecated routes
2. Add a description to all 200 responses in the OpenAPI spec
3. Remove `id` from the `putVisualChangeset` route body since it isn't used and conflicts with the URL param
4. Fix missing variable in the `servers` section of the OpenAPI route
5. Rework some of the named schemas to remove conflicts in the resulting spec


Kevin-Chant:
Closing in favor of #6141 (ported over the logic)

PR #5885, opened 2026-05-17
fix(incremental): Prevent deadlock when query finishes before DAG

johnham-ant:
## Problem

Triggering a **Full Refresh** on an experiment with many metrics (~100+) can silently deadlock the refresh and leave the experiment worse off than before:

- The `DROP TABLE` that begins the Full Refresh runs and completes. Nothing after it ever starts — no `CREATE TABLE`, no metric-source inserts, no stats query.
- The snapshot stays `status: "running"` indefinitely. The UI shows a spinner that never resolves.
- No error is raised. No `warn`/`error` log is emitted. The only trace in the warehouse's job history is the lone `DROP TABLE`.
- The user's only recourse is to manually cancel and retry — which hits the same condition again.

This is **deterministic, not intermittent**. Once an experiment crosses the metric-count threshold where SQL generation for the full query DAG takes longer than ~1 second, every Full Refresh attempt hits the same wall, every time.

**Reproduced:** on an experiment with ~120 fact metrics, three consecutive Full Refresh attempts all deadlocked identically after the `DROP TABLE` — including attempts where the metric set and fact table SQL were changed between tries, ruling out config as the cause. The same experiment config rebuilt its cache successfully via a regular auto-refresh that detected the missing cache table — that path's first query is a slow `CREATE TABLE AS SELECT`, which gives `startQueries()` time to finish before the callback timer fires.

And once the cache table has been dropped, the experiment can't self-heal: regular (non-full) incremental updates don't rebuild the cache — they either fail the settings-hash check (if the experiment config changed, which is usually why someone is doing a Full Refresh) or can't find the cache table to append to, and fall back to inline full-rescans. A single failed Full Refresh permanently downgrades the experiment from pipeline mode to inline mode until a Full Refresh succeeds — and because the race is deterministic, that never happens.

## Root cause: a timer race between query execution and DAG persistence

1. `startAnalysis()` calls `startQueries()`, which builds the full query DAG (for an incremental Full Refresh with many metrics this is ~50+ query docs: drop old cache → create cache → per-metric-source create/insert/timestamp queries → stats → traffic).
2. Inside `startQueries()`, `startQuery()` fires `executeQuery()` **immediately and fire-and-forget** for any query with no dependencies (the `readyToRun` branch). For a Full Refresh, the first such query is a `DROP TABLE IF EXISTS` on the old cache table.
3. `startAnalysis()` only persists the full `queries` array to the model **after** `startQueries()` returns.
4. The `DROP TABLE` is essentially instant. When it finishes, `onQueryFinish()` arms a 1-second `setTimeout` to check what runs next.
5. If SQL generation for the remaining queries takes longer than `drop_latency + 1000ms` — which it reliably does for experiments with many metrics — the timer fires *before* the `queries` array is persisted.
6. The timer callback calls `getLatestModel()` (reloads the snapshot — still `queries: []`) and then `refreshQueryStatuses()`, which finds no running or queued queries and returns `new Map()` with only a `logger.debug()`.
7. `startQueries()` eventually returns and `startAnalysis()` persists the queries array. But nothing processes it — the only thing that ever re-drives the DAG is `onQueryFinish()`, the DROP already fired its one shot, and the timer already came up empty.
8. Every downstream query stays `status: "queued"` forever. The snapshot stays `status: "running"` forever.

### Why it only affects Full Refresh

On a regular (incremental) refresh, the first `dependencies: []` query is an `UPDATE` or `CREATE TABLE AS SELECT` that scans raw data — it always takes far longer than SQL generation, so the timer always fires after the DAG is persisted. Full Refresh is the only DAG where the first query is an instant DDL statement, so it's the only DAG that can win this race.

### Why the safety nets don't catch it

Two cleanup mechanisms should in theory detect a stuck snapshot, and both miss this case by design:

- `getStaleQueries()` marks `running` queries with a stale heartbeat as failed. But the orphaned queries here are `queued`, not `running` — they never started, so they have no heartbeat to go stale.
- `reapStalledSnapshots()` (added in #5611) reaps snapshots stuck `running` after their queries finished. But it requires **all** queries to be in a terminal state (`succeeded`/`failed`) before reaping. The orphaned queries are `queued`, which is non-terminal, so the snapshot never qualifies.

There is therefore **no timeout, no error, no reaper, and no log** that surfaces the stuck state.

## Fix

Three small, independent changes. The first closes the race; the other two make the remaining surface area observable and recoverable.

### 1. Re-drive the DAG after it's persisted (`QueryRunner.startAnalysis()`)

After `updateModel()` has persisted the full `queries` array and the runner has been marked `running`, call `onQueryFinish()` one more time. This guarantees at least one pass through `refreshQueryStatuses()` / `startReadyQueries()` happens **after** the DAG is visible in the persisted model.

- If the timer from the early-finishing query is still pending, this is a no-op (the existing `this.timer` guard de-dupes).
- If the timer already fired and gave up, this re-arms it with the real DAG and the refresh proceeds.
- If nothing has finished yet (the normal incremental case), this is a cheap extra status poll that finds nothing to do.

**Timeline, before and after** (for a Full Refresh where SQL generation takes ~3s):

```
Before:
 t=0.0s  DROP fired (fire-and-forget)
 t=0.5s  DROP finishes → onQueryFinish → 1s timer armed
 t=1.5s  timer fires → reloads model → queries:[] → gives up
 t=3.0s  startQueries() returns → queries array persisted
         ...nothing ever looks at it again. Stuck forever.

After:
 t=0.0s  DROP fired (fire-and-forget)
 t=0.5s  DROP finishes → onQueryFinish → 1s timer armed
 t=1.5s  timer fires → reloads model → queries:[] → logs warn → gives up
 t=3.0s  startQueries() returns → queries array persisted
 t=3.0s  onQueryFinish() called → 1s timer armed
 t=4.0s  timer fires → reloads model → full DAG visible → DROP is succeeded
         → startReadyQueries() starts CREATE → refresh proceeds normally.
```

This was chosen over deferring `executeQuery()` for `readyToRun` queries until after persistence because (a) it's a one-line behavioral change that doesn't alter when queries start executing, and (b) the cached-query reuse path in `startQuery()` can also call `onQueryFinish()` before the DAG is persisted (when a cached query is already finished) — a post-persist pass covers that path too, whereas deferring execution would not.

### 2. Make the condition visible (`QueryRunner.refreshQueryStatuses()`)

When `refreshQueryStatuses()` finds no running/queued queries on a runner that is **not** in a terminal state, log at `warn` instead of `debug`. This exact condition is the signature of the race, and the fact that it was only ever a debug log is why this failure mode was invisible in production logs.

### 3. Reap orphaned DAGs (`expireOldQueries.reapStalledSnapshots()`)

Extend the stalled-snapshot reaper (from #5611) to also reap snapshots where **zero queries are running and one or more are queued**. That combination means nothing is executing and nothing ever will — the in-memory timers that drive a DAG forward are process-local, so once they're gone (server restart mid-run, or this race before the fix above), queued queries are stuck permanently. The reaper marks the snapshot `error` with a descriptive message so the UI stops spinning and the user can retry. It reuses the existing 60-minute stall threshold and 10-minute finalize grace period.

This is a safety net for snapshots stuck by older code, by a mid-run restart, or by any future regression in the DAG lifecycle — not just this specific race.

## Impact

- Experiments with enough metrics that SQL generation takes >1s can now reliably Full Refresh, restoring their pipeline-mode caches instead of being permanently downgraded to inline full-rescans.
- Snapshots that do get stuck (server restart mid-run, etc.) are reaped within ~70 minutes instead of showing "Running" forever.
- The previously-silent race condition is now logged at `warn` level when it fires.

## Scope / what this does not change

- The 1-second `onQueryFinish` polling interval is unchanged.
- The fire-and-forget execution of dependency-free queries inside `startQueries()` is unchanged.
- The `reapStalledSnapshots()` thresholds (60 min stall, 10 min grace) are unchanged; the new branch reuses them.
- `getStaleQueries()` (heartbeat-based query reaping) is unchanged — queued queries still have no heartbeat, but the DAG-level reaper now covers that gap.
- The all-terminal reap branch in `reapStalledSnapshots()` is behaviorally unchanged (the new `orphanedDag` branch is additive).

## Testing

- Added two tests to `packages/back-end/test/queryRunners/QueryRunner.test.ts`:
  - `re-arms the follow-up timer after persisting the query DAG` — asserts `onQueryFinish()` is called after `updateModel()` persists the queries array when the runner enters `running`.
  - `does not re-arm the timer when queries are already finished (cached)` — asserts no extra timer is armed for the fully-cached synchronous-finish path.
- Full `packages/back-end` jest suite passes (95 suites / 3986 tests / 2244 snapshots).
- `tsc --noEmit`, `eslint --max-warnings 0`, `prettier --check` all pass on changed files.

Manual repro for reviewers: on an experiment with enough metrics that `startExperimentIncrementalRefreshQueries()` takes >1s to generate SQL, trigger a Full Refresh. Before this change the DROP appears in the warehouse's job history and nothing follows; after it the full DAG proceeds.

## Related

- #5611 added `reapStalledSnapshots()` for the "all queries terminal but snapshot still running" case — this extends it to the "no queries running but some still queued" case.
- #5882 is part of the same series hardening incremental refresh.

### DIFF
--- a/packages/back-end/src/jobs/expireOldQueries.ts
+++ b/packages/back-end/src/jobs/expireOldQueries.ts
@@ -162,31 +162,57 @@ async function reapStalledSnapshots() {
       queryIds,
     );
     if (statuses.length !== queryIds.length) continue;
+
+    const running = statuses.filter((q) => q.status === "running");
+    const queued = statuses.filter((q) => q.status === "queued");
     const allTerminal = statuses.every(
       (q) => q.status === "succeeded" || q.status === "failed",
     );
-    if (!allTerminal) continue;
+
+    // An orphaned DAG: nothing is actually executing, but one or more
+    // queries are still "queued" and will never be started. This happens
+    // when the in-memory timer that drives the DAG forward was lost —
+    // e.g. the process restarted mid-run, or a fast first query finished
+    // and fired its follow-up timer before the full query DAG was
+    // persisted (the Full Refresh race). Queued queries have no heartbeat
+    // and are never "running", so neither getStaleQueries() nor the
+    // all-terminal reap path above will ever touch them. Without this
+    // branch the snapshot would show "Running" forever.
+    const orphanedDag = running.length === 0 && queued.length > 0;
+
+    if (!allTerminal && !orphanedDag) continue;
 
     const latestFinishedAt = Math.max(
+      0,
       ...statuses.map((s) => s.finishedAt?.getTime() ?? 0),
     );
-    if (Date.now() - latestFinishedAt < STALLED_FINALIZE_GRACE_MS) continue;
+    // For an orphaned DAG nothing may have finished yet (latestFinishedAt
+    // stays 0), in which case fall back to the snapshot's age — the
+    // STALLED_SNAPSHOT_THRESHOLD_MS check above has already guaranteed it.
+    const lastActivityAt =
+      latestFinishedAt > 0 ? latestFinishedAt : snapshot.dateCreated.getTime();
+    if (Date.now() - lastActivityAt < STALLED_FINALIZE_GRACE_MS) continue;
 
     const statusById = new Map(statuses.map((s) => [s.id, s.status]));
     snapshot.queries.forEach((q) => {
       q.status = statusById.get(q.query) ?? q.status;
     });
 
+    const error = orphanedDag
+      ? "Snapshot stalled: queries were never started. This can happen when the server restarts mid-refresh. Please try updating results again."
+      : "Snapshot stalled: queries finished but results were never finalized. This usually means the analysis step failed (check server logs) or the process was restarted.";
+
     const context = await getContextForAgendaJobByOrgId(snapshot.organization);
     const reaped = await errorSnapshotIfStillRunning(context, snapshot.id, {
       queries: snapshot.queries,
-      error:
-        "Snapshot stalled: queries finished but results were never finalized. This usually means the analysis step failed (check server logs) or the process was restarted.",
+      error,
     });
     if (!reaped) continue;
 
     logger.info(
-      `Reaped stalled snapshot ${snapshot.id} (experiment ${snapshot.experiment}): all ${queryIds.length} queries terminal but status still running`,
+      orphanedDag
+        ? `Reaped orphaned snapshot ${snapshot.id} (experiment ${snapshot.experiment}): ${queued.length} of ${queryIds.length} queries stuck in "queued" with nothing running`
+        : `Reaped stalled snapshot ${snapshot.id} (experiment ${snapshot.experiment}): all ${queryIds.length} queries terminal but status still running`,
     );
 
     await context.models.incrementalRefresh
--- a/packages/back-end/src/queryRunners/QueryRunner.ts
+++ b/packages/back-end/src/queryRunners/QueryRunner.ts
@@ -311,6 +311,24 @@ export abstract class QueryRunner<
       this.setStatus("finished", error, result);
     } else {
       this.setStatus("running");
+      // Schedule one more pass through refreshQueryStatuses/startReadyQueries
+      // now that the full `queries` array has been persisted to the model.
+      //
+      // This closes a race: queries with no dependencies are executed
+      // fire-and-forget inside startQueries() (see startQuery's readyToRun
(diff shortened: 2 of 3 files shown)


ahdriel (approved):
Thanks for the contribution & the focused fix.

PR #5890, opened 2026-05-18
API: add metric & variations names to results endpoint

lukesonnet:
Enrich results endpoint with variation and metric names to make parsing the results easier for programmatic and agentic workflows.


packages/back-end/src/services/experiments.ts
```diff
@@ -2287,7 +2286,7 @@ function getExperimentMetric(
 
 export async function toExperimentApiInterface(
   context: ReqContext | ApiReqContext,
-  experiment: ExperimentInterfaceExcludingHoldouts,
+  experiment: ExperimentInterface,
```
ahdriel, line 2308: We have validation on other endpoints that throw if the experiment is a holdout.

So now this creates a scenario where `GET api/v1/experiments/:id` throw but `GET api/v1/experiments/:id/results` for a holdout experiment works.

We should do an audit of the other get/list experiments to ensure we are properly handling holdouts now to prevent unexpected behavior.

packages/back-end/src/services/experiments.ts
```diff
@@ -2535,6 +2537,34 @@ export function toSnapshotApiInterface(
...
+  // Resolve display names for every metric that appears in the results, using
+  // the canonical "Parent (col: val, ...)" format for slice metrics so the
+  // payload is self-describing.
+  const baseMetricIds = Array.from(
+    new Set(
+      Array.from(metricIds).map((id) => parseSliceMetricId(id).baseMetricId),
+    ),
+  );
+  const baseMetricsById = new Map(
+    (await getExperimentMetricsByIds(context, baseMetricIds)).map((m) => [
```
ahdriel: Related to the greptile comment, maybe we need to receive `metricMap` as a param for this function (required) and let the parent build it, which will allow it do to in a single pass for all experiment in the example of `listExperimentResults`.

packages/back-end/src/services/experiments.ts
```diff
@@ -2535,6 +2537,34 @@ export function toSnapshotApiInterface(
...
+  );
+  const getMetricName = (id: string): string | undefined => {
+    const { baseMetricId, sliceLevels } = parseSliceMetricId(id);
+    const baseName = baseMetricsById.get(baseMetricId)?.name;
+    if (!baseName) return undefined;
+    if (!sliceLevels.length) return baseName;
+    const sliceContext = sliceLevels
+      .map(
+        (s) =>
+          `${s.column}: ${s.levels.length ? s.levels.join(" OR ") : "other"}`,
```
ahdriel, line 2583: question: Is this the same logic we use for the front-end? If yes, should we build a shared helper?

packages/back-end/src/services/experiments.ts
```diff
@@ -2522,7 +2522,9 @@ export function toSnapshotApiInterface(
   const activationMetric =
     snapshot.settings.activationMetric || experiment.activationMetric;
 
-  const variationIds = getLatestPhaseVariations(experiment).map((v) => v.id);
+  const phaseVariations = getLatestPhaseVariations(experiment);
+  const variationIds = phaseVariations.map((v) => v.id);
+  const variationNames = phaseVariations.map((v) => v.name);
```
ahdriel: question: This is related to variations in phases.

We can pass a snapshot from not the last phase for a given experiment, but the name for variations and ids will be use the latest phase instead.

I don't think we can fix this right now, but flagging.

packages/back-end/src/services/experiments.ts
lukesonnet: No, it isn't. In the front end we split the metric name and slice name into two lines (bold top, grey lower), so this is a different "UI" where it's a single string field.

packages/back-end/src/services/experiments.ts
lukesonnet: Going the metricMap route, thanks.

packages/back-end/src/services/experiments.ts
lukesonnet: I refactored the getLatestPhaseVariations so this wouldn't be hard to do.

packages/back-end/src/services/experiments.ts
lukesonnet: I agree, but we just cast `as ExperimentInterfaceExcludingHoldouts` everywhere without it buying us anything... not sure I am making things worse or partially better?

packages/back-end/test/api/experiments.test.ts
```diff
@@ -1905,6 +1910,116 @@ describe("experiments API", () => {
...
+            prerequisites: [],
+            namespace: { enabled: false },
+          },
+        ],
+      };
+      (getExperimentById as jest.Mock).mockResolvedValue(experimentWithResults);
+      (getMetricMap as jest.Mock).mockResolvedValue(
+        new Map([["met_1", { id: "met_1", name: "Signups" }]]),
+      );
+      (getLatestSnapshot as jest.Mock).mockResolvedValue({
```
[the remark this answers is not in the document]
lukesonnet: fixed, thanks

packages/back-end/src/services/experiments.ts
ahdriel: `getExperiment` does it correctly because we check `.type === 'holdout'` and throw an error if that's the case.

I think we can merge as-is but this feels like a change that we need to address more holistically against our experiment API endpoints.

packages/back-end/src/api/experiments/listExperimentResults.ts
```diff
@@ -33,10 +34,10 @@ export const listExperimentResults = createApiRequestHandler(
     }
   }
 
-  const snapshots = await getLatestSnapshotMultipleExperiments(
-    req.context,
-    experimentPhaseMap,
-  );
+  const [snapshots, metricMap] = await Promise.all([
+    getLatestSnapshotMultipleExperiments(req.context, experimentPhaseMap),
+    getMetricMap(req.context),
```
ahdriel: I think this is a better trade-off, but I do have concerns for orgs with lots of metrics, for which we will load them but use just a subset.

Should we use `getExperimentMetricsByIds`? This function returns an array so we would need to `  return new Map(metrics.map((m) => [m.id, m]));` as well.

And not sure if we want to get the metrics from the requested snapshot or the experiment as a whole (leaning towards the former).

ahdriel (approved):

PR #5899, opened 2026-05-19
Add GrowthBook error tracking plugin to sdk-js

tzjames:
### Features and Changes
Adds changes to the sdk-js package to allow for error-tracking.

### Dependencies
Errors will not be properly handled in clickhouse until the errors table and MV is created (they will go into the main events table).  https://github.com/growthbook/central-license-server/pull/120/changes will need to land, and then we will need to drop and recreate tables in the admin for the orgs that want to use this (or eventually create a script to update the tables and MV for all orgs).

After this lands we can land https://github.com/growthbook/growthbook/pull/5886 which needs @growthbook/growthbook to be published.

### Testing
See the testing in https://github.com/growthbook/growthbook/pull/5886.

### DIFF
--- a/packages/sdk-js/package.json
+++ b/packages/sdk-js/package.json
@@ -9,6 +9,7 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./dist/index.d.ts",
       "import": {
         "types": "./dist/index.d.ts",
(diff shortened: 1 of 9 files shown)


tzjames:
@cursoragent review

PR #5901, opened 2026-05-19
pp-updates

mknowlton89:
### Features and Changes


tzjames (approved):
Looks good.  I tested out an upgrade flow for a new org and it finished successfully.  I didn't try out any cases where it is supposed to actually catch bad actors, but the proof will be on the real deal.

PR #5906, opened 2026-05-19
Bryce/safe rollouts v2

bryce-fitzsimons:
# Monitored Ramp Schedules
Brings monitoring (previously only available on the standalone Safe Rollout rule) directly into ramp schedules, and deprecates the old Safe Rollout rule type.

## Monitoring on ramp steps
- Individual steps can now be marked monitored: true, backed by a live analysis experiment that tracks guardrail and signal metrics
- Monitored steps can automatically hold, roll back, or advance the ramp based on metric health — no manual intervention required
- Configurable monitoring settings per schedule: datasource, exposure query, guardrail metrics, signal metrics, no-traffic grace period, SRM handling, auto-update frequency

## Hold conditions
- Steps now support composite hold conditions: interval timer, manual approval, and minimum sample size — all independently combinable
- Approval gates track who approved, when, and through which surface (UI or API)
- The advance REST action respects approval gates by default (409 if unapproved), with an admin-level force override

## REST API
- New endpoints to configure monitoring and lockdown settings, update steps, trigger manual monitoring refreshes, and introspect schedule health (guardrail statuses, metric deviations, CI, p-value, SRM, exposure counts)
- Actions (advance, approve-step, rollback, etc.) are fully documented with clear guidance on when to use each

## UI
- New monitoring dashboard on the ramp rule page with live metric health, guardrail statuses, and manual refresh controls
- Ramp timeline popover now shows hold conditions, approval status, and monitoring state per step
- Editor supports the full new step configuration inline

## Bug fixes & reliability
- Fixed instant steps (no interval) stranding the schedule indefinitely
- Fixed stale approval data surviving rollbacks
- Fixed the edit modal not pre-populating pending draft changes when re-opened
- Fixed several step-chaining gaps in the advance and approval flows
- Lots of other assorted bug fixes

image


bryce-fitzsimons:
@greptileai

packages/back-end/src/services/rampSchedule.ts
```diff
@@ -431,28 +792,70 @@ export async function advanceStep(
...
+  );
 
-  const newStatus = isApprovalStep
-    ? ("pending-approval" as const)
-    : ("running" as const);
+  let monitoredStepDueAt: Date | null = null;
+  if (isMonitoredStep && step.interval != null) {
+    monitoredStepDueAt = new Date(now.getTime() + step.interval * 1000);
+  }
+
```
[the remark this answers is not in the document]
bryce-fitzsimons: this was already fixed, stale comment may have caused this to re-flag

bryce-fitzsimons:
@greptileai please re-review

packages/back-end/src/services/rampScheduleEvaluator.ts
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  if (
+    step.holdConditions?.requiresApproval &&
+    schedule.stepApproval?.stepIndex !== schedule.currentStepIndex
+  ) {
+    return { action: "hold", reason: "Awaiting approval" };
+  }
+
+  return { action: "advance" };
```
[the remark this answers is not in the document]
bryce-fitzsimons: Non-issue. Sample size is only measured for monitored steps. Adding comments to clarify.

packages/back-end/src/services/rampScheduleEvaluator.ts
```diff
@@ -0,0 +1,416 @@
...
+    return { action: "hold", reason: "Awaiting approval" };
+  }
+
+  // `minSampleSize` is intentionally not checked here. It requires snapshot
+  // data, which is only available on monitored steps (handled above by
+  // evaluateMonitoredStep). A non-monitored step with minSampleSize set has
+  // no data source to evaluate against, so the gate is a no-op. See the
+  // stepHoldConditions schema comment for the full explanation.
+  return { action: "advance" };
+}
```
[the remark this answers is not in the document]
bryce-fitzsimons: This is an incorrect understanding of "locked" mode. Locked prevents users from publishing changes to a feature if a locking ramp schedule is active. It does NOT lock the schedule itself.

Comments had some errors; they have been updated to reflect this.

packages/shared/src/util/environments.ts
```diff
@@ -0,0 +1,12 @@
+// Environments whose IDs match any of these patterns are considered
+// non-production (dev-like). A feature enabled only in these environments
+// is still treated as "draft" for status purposes.
+export const NON_PRODUCTION_ENV_PATTERNS: RegExp[] = [
+  /^dev/i,
+  /^staging/i,
+  /^qa/i,
```
jdorn: These are the most popular environment ids on GrowthBook Cloud:

```
production
dev
staging
development
test
qa
local
stage
uat
prod
develop
demo
testing
stg
sandbox
preview
preprod
beta
pre-prod
integration
homolog
```

Adding `^pre`, `^test`, `^stage`, `^stg`, `^uat`, `^local`, and `^homolog` covers most of these (plus a bunch more long tail env names).

Also, if we're going to have a bunch of these, we could do it in a single regex:

```ts
/^(dev|staging|qa|pre|test|stage|stg|uat|local|homolog)/i
```

packages/back-end/src/models/OrganizationModel.ts
```diff
@@ -249,6 +249,12 @@ export async function createOrganization({
   return toInterface(doc);
 }
 
+/** Every org id — used by Agenda pollers that run org-scoped work per tenant. */
+export async function listAllOrganizationIds(): Promise<string[]> {
```
jdorn: We probably don't need this method at all anymore, but if we still do, prefix the name with `_dangerously` since it is a cross-tenant query.

packages/back-end/src/jobs/updateRampSchedules.ts
```diff
@@ -40,9 +47,23 @@ async function queueRampScheduleAdvance(
 export default async function addRampScheduleJob(agenda: Agenda) {
   agenda.define(QUEUE_RAMP_SCHEDULE_ADVANCES, async () => {
     const now = new Date();
-    const scheduleDocs = await findSchedulesDueForProcessing(now);
-    for (const doc of scheduleDocs) {
-      await queueRampScheduleAdvance(agenda, doc);
+    const orgIds = await listAllOrganizationIds();
+
+    for (const organization of orgIds) {
```
jdorn: This will loop through tens of thousands of orgs on Cloud.  Instead, we need to run 1 single Mongo query that looks for due schedule ids across all orgs and then only loop through orgs that have any.

packages/back-end/src/routers/safe-rollout/safe-rollout.controller.ts
```diff
@@ -86,13 +92,32 @@ export const postSafeRolloutSnapshot = async (
...
+      snapshotId: snapshot.id,
+      hasHealth: !!snapshot.health,
+      totalUsers:
+        snapshot.health?.traffic?.overall?.variationUnits?.reduce(
+          (a: number, b: number) => a + b,
+          0,
+        ) ?? 0,
+    },
+    "POST /safe-rollout/:id/snapshot — snapshot created",
+  );
```
jdorn: Info logs will show up in production. Is that intentional to leave in this PR?

packages/back-end/src/util/features.ts
```diff
@@ -857,17 +859,73 @@ export function getFeatureDefinition({
...
+            if (clampedCoverage < 1) {
+              rule.filters = [
+                {
+                  seed: monitoredSeed,
+                  attribute: r.hashAttribute,
+                  hashVersion: 1,
+                  ranges: [[0, clampedCoverage] as [number, number]],
+                },
+              ];
+            }
```
jdorn: Do we want both `coverage` and a filter? Won't that double apply the coverage using potentially 2 different hashing strategies?

I think the goal is to force monitored and unmonitored steps to use the exact same coverage ranges, but I'm not confident the current code does this properly.

bryce-fitzsimons: you're right, there was a bug in the range calc. Fixing and hardening the behavior with tests.

jdorn (approved):

PR #5908, opened 2026-05-19
feat: support cross-fact-table ratio metrics in incremental refresh

lukesonnet:
### Problem

A ratio metric whose numerator and denominator live in different fact tables (e.g. `revenue` from `purchases` ÷ `$$count` from `sessions`) currently makes `validateIncrementalPipeline()` throw:

> Ratio metrics must have the same numerator and denominator fact table with incremental refresh.

Because the guard is experiment-wide, a single cross-FT ratio metric disqualifies the **entire experiment** from the incremental query runner and forces it back to the inline path, which re-scans all raw fact-table data on every refresh. Pipeline mode exists to avoid exactly that, so the fallback can make every refresh orders of magnitude more expensive — silently.

### Approach

Cross-FT ratio metrics don't get their own dedicated cache pair (as in #5882). Instead, each cross-FT ratio metric's **numerator side lives in its numerator fact table's existing cache**, and its **denominator side lives in its denominator fact table's existing cache**, bundled with the same-FT metrics that already live there. A second pass joins both caches in the stats query.

That gives us:
- one cache per fact table (not one per metric pair), with shared incremental `max_timestamp` and shared CUPED covariate cache;
- same-FT metrics that happen to share a fact table with a cross-FT ratio half ride the same incremental update — no duplicate scans;
- cross-FT CUPED falls out for free, because each fact table's covariate cache already exists.

### Core changes

**`planMetricFanOut(metrics)` — single source of truth for cache layout** (`services/experimentQueries/planMetricFanOut.ts`)

Returns `{ perFt[], crossFtPairs[] }` where:
- `perFt[i] = { factTableId, metrics }` — one entry per fact table that needs a cache. A cross-FT ratio metric appears in **both** of its fact tables' entries.
- `crossFtPairs[i] = { factTableIds: [A,B], metrics }` — one entry per **unordered** fact-table pair with at least one cross-FT ratio; bundles A/B and B/A combinations into a single pair so they share one joined stats query.

The runner, the (factTableId, metricId) cache-invalidation diff, and the cross-FT stats join all derive their state from one call to this function.

**Role is derived, not persisted.** A cache "knowing" whether it owns the numerator or denominator side of a cross-FT metric is computed on the fly from `metric.numerator.factTableId === source.factTableId` (and likewise denominator). The `IncrementalRefreshMetricSourceInterface` validator gains no new fields and no migration is required.

**Two-pass orchestration** (`ExperimentIncrementalRefreshQueryRunner.ts`)
1. **Per-FT pass** — for each `perFt[]` entry: drop/create/insert the cache table, optionally drop/create/insert its covariate cache, and emit a same-FT stats query for the metrics that have both sides on this FT.
2. **Cross-FT pair pass** — for each `crossFtPairs[]` entry: look up both per-FT pipelines from the first pass, build an unordered `{ factTableId → tableFullName }` map (plus a matching covariate map), and emit one joined stats query per pair. `crossFtSubGroups.ts` does the lookup and dedupes A/B + B/A into a single sub-group.

The exploratory runner mirrors the same two-pass shape.

**Cache invalidation by `(factTableId, metricId)` diff** — instead of the persisted schema describing which side of a ratio each cache owns, we compute a bidirectional diff between the stored caches' `(factTableId, metricId)` tuples and `planMetricFanOut`'s desired tuples. Tables with new or orphaned tuples are rebuilt end-to-end (DROP + CREATE + full INSERT); everything else updates incrementally. Orientation flips on the same metric (swapping numerator/denominator fact tables) are still caught earlier by `getMetricSettingsHashForIncrementalRefresh`.

**FT-scoped schema + insert SQL** (`SqlIntegration.ts`, `metric-source-table-schema.ts`)

`getMetricSourceTableSchema`, `getCreateMetricSourceTableQuery`, `getInsertMetricSourceDataQuery`, and the covariate equivalents all take a `factTableId` and project only the side(s) that FT actually hosts. Same-FT metrics → both `{m}_value` and `{m}_denominator_value`; cross-FT ratio metrics → just the side(s) anchored in this FT. Identical helpers, identical SQL output for the same-FT case; the cross-FT case just emits fewer columns per cache.

To make per-FT inserts work when a "hub" FT participates in multiple cross-FT ratios (e.g. `[A/B, A/C]`), `getFactTablesForMetrics` / `parseExperimentFactMetricsParams` take an optional `targetFactTableId` that scopes fact-table discovery to a single FT. In that mode the 2-FT cap is irrelevant by construction. The original (no `targetFactTableId`) mode is unchanged and still enforces the cap for stats queries.

**Multi-source stats SQL** (`getIncrementalRefreshStatisticsQuery`)

Takes `metricSourceTables: Record<factTableId, tableFullName>` and (optionally) `metricSourceCovariateTables: Record<factTableId, tableFullName>` — one entry for single-FT, two for cross-FT pair. The SQL layer derives source ordering and `m{i}` aliases on the fly from each metric's fact-table first-appearance order; callers don't impose an ordering. Each metric's `__joinedData{i}` CTE picks its numerator column from the numerator FT's `m{i}_value` and its denominator column from the denominator FT's `m{i}_denominator_value` — for same-FT metrics those resolve to the same source, for cross-FT metrics to two different sources. CUPED LEFT JOINs work the same way across both covariate caches.

**Validator narrowing** (`dataPipeline.ts`)

`validateIncrementalPipeline` no longer rejects cross-FT ratio metrics, and no longer rejects cross-FT ratio metrics with CUPED. The same hash-based invalidation already handles orientation flips.

### Tests

- `test/services/planMetricFanOut.test.ts` — same-FT, cross-FT, hub (A/B + A/C), and orientation-flip ordering.
- `test/queryRunners/incrementalRefreshMetricSources.test.ts` — group reuse, settings-hash drift, and the (factTableId, metricId) cache-invalidation diff.
- `test/integrations/BigQuery.test.ts` — schema/CREATE/INSERT/stats SQL across single-FT, cross-FT, cross-FT with CUPED, mixed same-FT + cross-FT in one pipeline (per-FT stats reads only same-FT covariates, cross-FT pair stats reads both), and 3-FT hub.

### Differences vs #5882

#5882 took the "dedicated group per fact-table pair" approach: a cross-FT ratio metric gets its own group with two sibling cache tables (numerator + `_denom`), each with its own `denominatorMaxTimestamp` / `denominatorTableFullName` / `denominatorFactTableId` fields on the persisted metric source. That kept the rest of the pipeline mostly unchanged but meant:
- A fact table that appears in both a same-FT group and a cross-FT group gets scanned twice per refresh.
- The validator persists a new "this source has a denominator" shape, with new optional fields and partial-failure edge cases (e.g. stale `denominatorMaxTimestamp` if only the denominator insert fails — flagged in greptile review).
- CUPED on cross-FT ratios stays unsupported (covariate cache would also need to span two FTs).

This PR does the larger restructuring suggested in the #5882 review: cross-FT halves live in their fact tables' existing caches, so:
- one cache per fact table — no duplicate scans even when a FT participates in many cross-FT ratios;
- no new fields on `IncrementalRefreshMetricSourceInterface`, no schema migration;
- cross-FT CUPED works (each FT's existing covariate cache holds its side);
- `planMetricFanOut` is the single source of truth for cache layout, change detection, and stats-query orchestration — reducing the surface area for the orientation/role bookkeeping that #5882 spread across the runner and the validator.

The trade-off is a slightly larger diff in `SqlIntegration.ts` (schema/insert/stats now take a `factTableId` and project FT-scoped columns) and a new two-pass runner shape. The end-state SQL for same-FT metrics is byte-identical to today's (see "unchanged behavior" assertions in the BigQuery snapshot tests).


packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -123,50 +140,48 @@ export function getIncrementalRefreshMetricSources({
...
-  Object.entries(groups).forEach(([groupId, group]) => {
-    if (group.alreadyExists) {
+  const finalGroups: MetricSourceGroups[] = [];
+  Object.entries(buckets).forEach(([bucketKey, bucket]) => {
+    if (bucket.alreadyExists) {
       finalGroups.push({
-        groupId,
-        factTableId: group.factTableId,
-        metrics: group.metrics,
+        groupId: bucketKey.slice("__existing__".length),
```
ahdriel: nit: this feels quite fragile if we change the bucketKey for either `__existing__` or `__new__`

packages/front-end/components/Experiment/ExperimentMetricsSelector.tsx
```diff
@@ -142,18 +124,6 @@ export default function ExperimentMetricsSelector({
       } else {
         // Check if individual metric is a cross fact-table ratio metric
```
ahdriel: nit: can remove this comment too

ahdriel (approved):
Nice work! The logic is a bit more complicated but not that hard to follow.

lukesonnet:
@cursor review

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2599,118 +2833,184 @@ export default abstract class SqlIntegration
...
-      )
-      ${
-        percentileData.length > 0
-          ? `
-        , __capValue AS (
-            ${this.getSqlDialect().percentileCapSelectClause(percentileData, "__joinedData")}
+                    qm.variation = u.variation
+                    ${allDimensionCols.map((c) => `AND qm.${c.alias} = u.${c.alias}`).join("\n")}
+                  )`
+              : "";
```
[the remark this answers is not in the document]
lukesonnet: IMO to be resolved as part of the quantile work with @ahdriel , will leave as a follow-up for now. Decent callout though.

PR #5921, opened 2026-05-21
feat: introduce slim snapshot status polling

lukesonnet:
## Summary

Builds on PR #5920 (which skipped analysis chunk decoding for the `latest` field) by moving snapshot status lookups onto a dedicated endpoint and decoupling the UI's polling loop from the heavy snapshot fetch. The result: poll loops no longer re-fetch and re-decode per-metric analysis chunks every 2–20s, and the safety of "no per-metric results in `latest`" is now enforced at the type level instead of by a runtime flag.

### New `SnapshotStatusSummary` type + endpoint

- **`shared/types/experiment-snapshot.d.ts`** — new `SnapshotStatusSummary = Pick<ExperimentSnapshotInterface, ...>` capturing only the status/metadata fields (`id`, `status`, `error`, `queries`, `runStarted`, `dateCreated`, `multipleExposures`, `health`, `banditResult`, `type`, `triggeredBy`).
- **`ExperimentSnapshotModel.ts`** — new `getLatestSnapshotStatus()` that returns this summary using a Mongo projection + `.lean()` (skips reading inline analysis blobs *and* skips the per-metric chunk decode). The "running-over-scheduled-error" tiebreaker moves here, where it was always meant to live.
- **`controllers/experiments.ts` + `app.ts`** — new `GET /experiment/:id/snapshot-status/:phase` route serving the summary. The existing `getSnapshot`/`getSnapshotWithDimension` controllers no longer compute their own redundant `latest` (which was previously decoding chunks twice per request).
- **`getLatestSnapshot` → `getLatestSuccessfulSnapshot`** — renamed to reflect what it actually does (and always did) for current callers; the now-dead `withResults` / `populateResults` parameters are removed. Five call sites (`presentations`, `reports`, `experimentNotifications`, `getExperimentResults` API, `postReport` API, plus a couple in the controllers) updated in lockstep.

### Front-end provider rework

- **`SnapshotProvider.tsx`**
  - Adds a second `useApi` against the status endpoint, narrows the context `latest` to `SnapshotStatusSummary`, and exposes a new `mutateLatest` for status-only refresh alongside the existing `mutateSnapshot` (which refreshes both fetches).
  - The heavy `useApi` opts out of focus/reconnect revalidation (`autoRevalidate: false`); the cheap status fetch carries that signal instead.
  - Two new module-scope hooks coordinate the two fetches:
    - `useRefetchHeavyOnStatusSuccess` — when the status endpoint reports a newer successful snapshot id, pull the matching heavy fetch exactly once. Lets poll loops use the cheap mutator and still get fresh analyses on completion.
    - `useCoherentLatest` — atomic transition gate. Running/errored/progress updates pass through immediately; the final "success" flip is held back until the heavy fetch agrees on the snapshot id, preventing a flicker of "queries done" alongside stale analyses.
- **`AnalysisSettingsSummary.tsx`** — `RunQueriesButton` polling and post-submit refresh now use `mutateLatest` (provider auto-upgrades the heavy fetch once a new success arrives). `ForceRerunQueriesButton` flipped to `mutateLatest` too — POSTing creates a new snapshot id, so the auto-upgrade fires.
- **`RefreshBanditButton.tsx`**, **`HealthTabOnboardingModal.tsx` / `HealthTab.tsx`** — same flip: their POSTs create new snapshot ids, so they only need to invalidate status.

### `RefreshResultsButton` cleanup

- The `onSubmitSuccess(snapshot)` callback was only used by call sites for the `trackSnapshot(...)` telemetry call. That call is now encapsulated inside the button itself, gated by an experiment-only `experimentSnapshotTrackingProps?: { trackingSource; datasourceType }` prop, with the post-success UI hook split out as a generic `onSuccess?: () => void`. The `apiCall` is also entity-branched so the typed snapshot passed to `trackSnapshot` is unambiguously `ExperimentSnapshotInterface`.

### Type-safety tightening on `getSRMValue`

- **`shared/src/health/health.ts`** — replaced the loose union signature with two overloads:
  - `getSRMValue("multi-armed-bandit", Pick<...>)` — accepts a status summary.
  - `getSRMValue(ExperimentType, ExperimentSnapshotInterface)` — requires the full snapshot for standard/holdout.
- Standard/holdout's fallback path reads `analyses[0].results[0]` and would silently return `undefined` if fed a `SnapshotStatusSummary`. The overloads make that a compile error.
- All three existing call sites resolve cleanly: the two bandit-specific front-end callers hit overload 1, and the back-end `services/experiments.ts` caller (which has a variable `ExperimentType` and a full snapshot) hits overload 2.

### Test updates

- `test/models/ExperimentSnapshotModel.test.ts` — the `describe("getLatestSnapshot")` block split into two: 6 status-summary tests retargeted at `getLatestSnapshotStatus`, plus 1 retargeted at `getLatestSuccessfulSnapshot`. The one `beforeSnapshot`-specific test was dropped (that parameter no longer exists on either function).
- `test/api/experiments.test.ts` and `test/services/experimentNotifications/experimentSignificance.test.ts` — rename `getLatestSnapshot` → `getLatestSuccessfulSnapshot` in imports, mock symbols, and mock setters.

## Expected impact

For an experiment page open in a tab while a snapshot is running (poll cadence 2–20s, plus default SWR focus revalidation):

- **Before:** every poll = full `GET /snapshot/:phase` = inline-snapshot read + double per-metric chunk decode. One open tab + a 10-minute refresh window ≈ ~57 full snapshot GETs.
- **After:** every poll = projected `GET /snapshot-status/:phase` (status fields only, no chunks). One full snapshot GET per *successful* completion. For experiments with many metrics, the chunk-decode cost was previously dominating that endpoint.

## Test plan

- [X] `pnpm --filter back-end test` — green (96 suites / 4027 tests)
- [X] `pnpm --filter front-end test` — green (22 suites / 314 tests)
- [X] `pnpm --filter shared test` — green (26 suites / 697 tests)
- [X] `pnpm type-check` — green across shared, front-end, back-end
- [X] Manually verify on a running experiment: status updates appear immediately during refresh; the "Running…" → "Success" indicator transitions only once analyses are loaded (no flicker); cancelling refresh keeps the previous successful snapshot visible.
- [x] Manually verify the health status does not flicker off on a regular update
- [ ] Manually verify on a bandit: `RefreshBanditButton` updates the queries panel without re-fetching analyses for the intermediate states.
- [x] Manually verify on Force Refresh from `ResultMoreMenu`: new snapshot id appears in status, then analyses upgrade exactly once.
- [ ] Manually verify the Health Tab onboarding "Refresh now" path: new snapshot appears via auto-upgrade, no double-fetch.
- [ ] Spot-check Network tab on a snapshot-running page: `/snapshot-status/:phase` fires every poll; `/snapshot/:phase` fires only on completion or initial load.

## Follow-ups not in this PR

- Safe-rollout symmetry: mirror `getLatestSnapshotStatus` + the provider hooks for `SafeRollout/SnapshotProvider.tsx` so the two flows stay analogous.
- `getSnapshotStatus` controller and `_getSuccessfulSnapshot` both inline the same `getExperimentById` + org check — extract to a `requireExperimentAccess(context, id)` helper once a third snapshot endpoint shows up.
- Watchdog on the `useCoherentLatest` hold: if the heavy fetch is degraded and stays >Ns, surface a sub-state ("Loading results…") instead of leaving "Running…" up indefinitely.
- Dimension URL encoding asymmetry between `/snapshot/:phase/:dimension` (raw path) and `/snapshot-status/:phase?dimension=...` (encoded query). Pre-existing on the heavy URL; worth harmonising.


packages/shared/src/health/health.ts
```diff
@@ -95,9 +95,20 @@ export function getSRMHealthData({
...
   experimentType: ExperimentType,
-  snapshot: ExperimentSnapshotInterface,
+  snapshot: Pick<
+    ExperimentSnapshotInterface,
+    "id" | "banditResult" | "health"
+  > & {
+    type?: ExperimentSnapshotInterface["type"];
+    analyses?: ExperimentSnapshotInterface["analyses"];
+  },
 ): number | undefined {
```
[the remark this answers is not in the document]
lukesonnet: revised approach anyways, but i do think this issue still exists.

packages/front-end/components/Experiment/SnapshotProvider.tsx
```diff
@@ -65,6 +81,58 @@ export function getPrecomputedDimensions(
...
+function useRefetchHeavyOnStatusSuccess(
+  statusLatest: SnapshotStatusSummary | undefined,
+  snapshotId: string | undefined,
+  refetchHeavy: () => Promise<unknown>,
+): void {
+  useEffect(() => {
+    if (statusLatest?.status !== "success") return;
+    if (statusLatest.id !== snapshotId) void refetchHeavy();
+  }, [statusLatest?.id, statusLatest?.status, snapshotId, refetchHeavy]);
+}
```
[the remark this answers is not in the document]
lukesonnet: Thanks, fixed.

lukesonnet:
Before, regularly loading full 374kb snapshot
Screenshot 2026-05-20 at 9 45 50 PM

After, only loading 1-2 kb snapshot that won't grow nearly as quickly as one with full results
Screenshot 2026-05-20 at 9 45 08 PM

packages/back-end/src/models/ExperimentSnapshotModel.ts
```diff
@@ -987,55 +965,132 @@ export async function getLatestSnapshot({
...
+
+  const mostRecent = await ExperimentSnapshotModel.findOne(
+    {
+      ...query,
+      status: { $in: ["success", "running", "error"] },
+    },
+    snapshotStatusProjection,
+    { sort: { dateCreated: -1 } },
+  )
+    .lean<Partial<ExperimentSnapshotInterface>>()
```
ahdriel, line 1061: TIL about `lean()`

packages/front-end/components/Experiment/SnapshotProvider.tsx
```diff
@@ -14,9 +21,18 @@ const snapshotContext = React.createContext<{
   experiment?: ExperimentInterfaceStringDates;
   snapshot?: ExperimentSnapshotInterface;
   analysis?: ExperimentSnapshotAnalysis | undefined;
-  latest?: ExperimentSnapshotInterface;
   dimensionless?: ExperimentSnapshotInterface;
+
+  latest?: SnapshotStatusSummary;
```
ahdriel: suggestion: we should rename this to be `latestSummary` so it also makes it clearer around `mutateLatestSummary` vs `mutateSnapshot`

So we have
`snapshot` = currently loaded (depends on other configs)
`latestSummary` = last generated snapshot summary

and the associated mutate functions.

packages/front-end/components/Experiment/SnapshotProvider.tsx
```diff
@@ -78,16 +159,54 @@ export default function SnapshotProvider({
...
 
+  // `latest` is sourced from a dedicated status endpoint that skips loading
+  // and decoding the per-metric analysis chunks. Keyed by the same
+  // phase/dimension/type tuple as the main snapshot fetch.
+  const statusQuery = [
+    dimension ? `dimension=${encodeURIComponent(dimension)}` : "",
+    snapshotType ? `type=${snapshotType}` : "",
+  ]
+    .filter(Boolean)
+    .join("&");
```
ahdriel: nit: we can use `new URLSearchParams({...}).toString()` to handle the formatting and encoding for us.

packages/back-end/src/app.ts
```diff
@@ -648,6 +648,10 @@ app.get(
   "/experiment/:id/snapshot/:phase/:dimension",
   experimentsController.getSnapshotWithDimension,
 );
+app.get(
+  "/experiment/:id/snapshot-status/:phase",
+  experimentsController.getSnapshotStatus,
```
ahdriel: nit: should these be `snapshot-summary` and `getSnapshotSummary`?

packages/front-end/components/Experiment/RefreshResultsButton.tsx
```diff
@@ -119,11 +131,28 @@ export default function RefreshResultsButton<
...
+                  snapshotEndpoint,
+                  { method: "POST" },
+                );
+              } else {
+                const res = await apiCall<{
+                  snapshot: ExperimentSnapshotInterface;
+                }>(snapshotEndpoint, {
+                  method: "POST",
+                  ...(body && { body }),
+                });
```
ahdriel, line 145: The `apiCall` is the same -- should we merge them in a single call and conditionally call the tracking?

```
const res = await apiCall<{ snapshot: SafeRolloutSnapshotInterface | ExperimentSnapshotInterface }>(snapshotEndpoint, {
  method: "POST",
  ...(body && { body }),
});
if (entityType !== "safe-rollout" && experimentSnapshotTrackingProps) {
  trackSnapshot(
    "create",
    experimentSnapshotTrackingProps.trackingSource,
    experimentSnapshotTrackingProps.datasourceType,
    res.snapshot as ExperimentSnapshotInterface,
  );
}
```

I don't know if this will work for the apiCall types though

packages/front-end/components/Experiment/SnapshotProvider.tsx
```diff
@@ -14,9 +21,18 @@ const snapshotContext = React.createContext<{
...
+  latest?: SnapshotStatusSummary;
+  // Refreshes both fetches. Use when the operation mutates fields on the
+  // current snapshot in place without changing its id (e.g., appending an
+  // analysis). For operations that create a new snapshot, prefer
+  // mutateLatest and let the provider auto-upgrade.
   mutateSnapshot: () => Promise<unknown>;
+  // Refreshes only the cheap status fetch. Use for poll loops; the provider
+  // automatically refreshes the heavy snapshot when status indicates a
+  // newer successful snapshot is available.
+  mutateLatest: () => Promise<unknown>;
```
ahdriel: As a separate suggestion from the comment above, we could also attempt to handle the distinction between summary and snapshot internally and provide a single `mutate` function externally.

So users would always call `mutate` which we would fetch summary and then decide if we need to fetch the full snapshot all in this hook.

And we can have a `mutate({ force: true })` for the scenarios where we want both to be fetched regardless.

ahdriel (approved):
Nice perf improvement! 🚀

packages/back-end/src/models/ExperimentSnapshotModel.ts
lukesonnet: YIAL

packages/front-end/components/Experiment/RefreshResultsButton.tsx
lukesonnet: Yes that works if we're happy with the `as` I'll make the change.

packages/front-end/components/Experiment/SnapshotProvider.tsx
lukesonnet: Yeah I'll look into this, thanks.

packages/front-end/components/Experiment/SnapshotProvider.tsx
lukesonnet: Updated to `latestSummary` and will have a single exposed mutate function as per your other comment.

PR #5922, opened 2026-05-21
Performance improvements for graph lookups, including prereq states

bryce-fitzsimons:
- Optimize prerequisite batch endpoint: Precompute a single dependency graph and reuse it for all cycle checks, instead of running a separate graph traversal per candidate feature. Cache prerequisite state evaluations across the batch to avoid redundant work.
- Optimize stale/dependents endpoints: Build a reverse dependency index once per request so finding a feature's dependents is an O(1) lookup instead of scanning all features.
- Yield the event loop: Add periodic setImmediate yields in long-running loops to avoid blocking the Node process.
- Simplify batch API surface: Remove checkRulePrerequisites and checkPrerequisite options; derive rule-level cycle status client-side from the existing wouldBeCyclic flags.
- Reduce redundant frontend API calls: Set the default prerequisite condition inline on selection instead of going through an intermediate empty state that triggered a double fetch.
- Fix cyclic warning display: Show wouldBeCyclic warnings regardless of environment count (was incorrectly hidden in multi-env orgs without a "production" env).


packages/back-end/src/controllers/features.ts
```diff
@@ -5232,106 +5222,132 @@ export async function postBatchPrerequisiteStates(
...
+      }
+    }
+    if (deps.size > 0) {
+      adj.set(f.id, deps);
+    }
+  }
+  return adj;
+}
+
+// Returns the set of features that transitively depend on targetId.
```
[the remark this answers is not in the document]
bryce-fitzsimons: fixed

anna-yn (commented):
This addresses everything we measured. For context on scale: with a few thousand features x ~5 environments we were seeing this endpoint at ~11s average (max 13s) per request, with the event loop blocked for the full duration -- and the per-candidate clone + traversal and the un-memoized re-evaluation were exactly where the cost model pointed. The single adjacency build + ancestor-set lookup, the cross-batch state cache, and the `yieldEventLoop` in every batch loop are each the fix we'd have proposed, and the same pattern applied to `features/stale` / `dependents` covers the other two endpoints we saw blocking. We'll verify against our baseline numbers once this ships.

Two small notes inline on the cycle-check semantics -- neither is blocking.

packages/back-end/src/controllers/features.ts
```diff
@@ -5232,106 +5222,141 @@ export async function postBatchPrerequisiteStates(
...
+  );
+
+  const stateCache = new Map<string, PrerequisiteStateResult>();
+
+  let ancestorsOfBase: Set<string> | null = null;
+  if (baseFeature) {
+    const graphFeatures = [...allFeatures];
+    if (!featuresMap.has(baseFeature.id)) {
+      graphFeatures.push(baseFeature);
+    }
```
anna-yn, line 5388: I think the draft-rules overlay ends up inert here: `featuresMap` is built from `getAllFeatures()`, which already contains the persisted base feature, so this branch never fires and the overlaid copy (from `postBatchPrerequisiteStates`, where `revision.rules` is spread onto `baseFeature`) never enters `graphFeatures` / the adjacency.

I don't think it can affect `wouldBeCyclic` either way -- adding `base -> candidate` only closes a cycle if a path `candidate -> ... -> base` already exists, and the base feature's own *outgoing* edges (the only thing the draft rules change) can't be part of that path. So this is dead code rather than a bug, but it currently reads as if the draft rules are meant to influence the check. Either drop the overlay + this branch, or substitute the overlaid copy into the graph for real and comment why.

packages/back-end/src/controllers/features.ts
```diff
@@ -5351,248 +5378,25 @@ async function evaluateBatchPrerequisiteStates({
...
-          visited.add(feature.id);
-          return false;
-        };
-
-        return visit(f, 0);
-      };
-
-      wouldBeCyclic = await checkCyclicAsync(testFeature);
-    }
+    const wouldBeCyclic = ancestorsOfBase
```
anna-yn, line 5420: Two behavior differences from the old per-candidate DFS, both probably fine but worth being deliberate about:

1. If `featureIds` ever includes `baseFeatureId` itself, the old code reported it as cyclic (the DFS immediately revisited the base); the new code reports `false` unless the base already participates in a cycle, since the BFS never adds the start node to its own ancestor set. Fine if the picker always filters the base feature out of the candidate list.

2. The old DFS started *from* the base feature, so a pre-existing cycle anywhere in the base's existing prerequisite chain marked **every** candidate as `wouldBeCyclic`. The new check only flags candidates that would close a new cycle through the proposed edge. We think the new behavior is more correct, but it is a change someone could notice.

(Related: the old check also treated `depth >= PREREQUISITE_MAX_DEPTH` as cyclic; the BFS has no depth cap, so a very deep but acyclic chain is no longer flagged. Also seems like an improvement.)

packages/shared/src/util/features.ts
```diff
@@ -431,6 +431,7 @@ interface IsFeatureStaleInterface {
   environments?: string[];
```
ahdriel: question: a lot of these params are optional but they seem to be very important.

should we remove the optionality of some/all of them?

packages/shared/src/util/features.ts
```diff
@@ -1671,22 +1678,69 @@ export function evalDeterministicPrereqValue(
...
     );
-    return prerequisites.some((p) => p.id === feature.id) || rules.length > 0;
-  });
-  return dependentFeatures.map((f) => f.id);
+  };
+
+  if (reverseDependencyIndex) {
+    const candidates = reverseDependencyIndex.get(feature.id);
+    if (!candidates || candidates.size === 0) return [];
+    const featuresById = new Map(features.map((f) => [f.id, f]));
```
ahdriel: nit: should we get featureMap as a parameter as well so we don't recreate it?

packages/front-end/components/Features/PrerequisiteInput.tsx
```diff
@@ -58,6 +64,11 @@ import {
   withOperatorCaseInsensitivity,
 } from "./ConditionInput";
 
+export interface RuleCyclicResult {
```
ahdriel, line 67: nit: should this type be in a `cyclic` or `rules` helper/service file?

packages/front-end/components/Features/RuleModal/index.tsx
```diff
@@ -1590,6 +1580,7 @@ export default function RuleModal({
             pendingDetach={hasPendingDetach}
             envScope={envScopeProps!}
             isLiveRule={isLiveRule}
+            onRuleCyclicChange={onRuleCyclicChange}
```
ahdriel, line 1598: This is a good candidate for a context so we don't need to pass the props to components that just forward it to PrerequisiteInput.

no need to change right now, but I think it could simplify a lot of these props.

packages/back-end/src/util/yield.ts
```diff
@@ -0,0 +1,8 @@
+const YIELD_INTERVAL = 100;
```
ahdriel, line 1: question: I have no idea how long the functions wrapped on this take, so I don't know if 100 is a good/reasonable number.

One alternative would be to change the yield to be time-based and yield after X milliseconds have passed, so that if we have 100 features that will take 50ms each, we are not blocked for 5 seconds.

what do you think?

packages/back-end/src/api/features/getFeatureStale.ts
```diff
@@ -36,10 +37,20 @@ export async function computeFeatureStale(
 
   const features = allFeatures.filter((f) => idSet.has(f.id));
 
+  const featuresMap = new Map(allFeatures.map((f) => [f.id, f]));
```
ahdriel: This is similar to getFeaturesStaleStates and getFeaturesDependents (not 1:1 match).

Should we move the logic of building featureMap + experimentMap + reverseDependency to a shared helper?

packages/back-end/src/controllers/features.ts
```diff
@@ -5232,106 +5222,129 @@ export async function postBatchPrerequisiteStates(
...
     baseFeature,
-    revision,
-    checkPrerequisite,
-    checkRulePrerequisites,
     res,
   });
 }
 
-// Shared logic for evaluating batch prerequisite states
+function buildPrerequisiteAdjacency(
```
ahdriel, line 5301: suggestion: should we move these function to a different file? maybe prereq focused?

ahdriel (approved):
Changes make sense!

I am not sure how to test it though, so this is just a code review.

packages/shared/src/util/features.ts
bryce-fitzsimons: agree, refactoring this

packages/front-end/components/Features/RuleModal/index.tsx
bryce-fitzsimons: probably not worth the scope creep for now

packages/front-end/components/Features/PrerequisiteInput.tsx
bryce-fitzsimons: this interface is FE only, not derived from a service/endpoint. leaving as is

packages/back-end/src/controllers/features.ts
bryce-fitzsimons: probably, but might want to take a fresh pass at this separately

packages/back-end/src/util/yield.ts
bryce-fitzsimons: at the scale of 10k features with a 10s avg compute time, 100 iterations is pretty reasonable. It works about to about 100 yields across the batch, each about 100ms or so. The goal should be no more than 100-200ms per yield batch, plus a healthcheck endpoint that times out after 5-10s is well over this threshold, so this checks out with room to spare.

I don't think we need to get too fancy with dynamic intervals unless needed.

PR #5934, opened 2026-05-22
feat: Snowflake incremental mode

lukesonnet:
**Add Snowflake support for incremental refresh pipeline**

- Enables Snowflake as a supported data source for incremental pipeline mode
- Implements KLL quantile sketches via Snowflake's t-digest functions (`APPROX_PERCENTILE_ACCUMULATE/COMBINE/ESTIMATE`); stores sketch state as `OBJECT` (not `BINARY`)
- Works around Snowflake's correlated subquery restriction in `kllRankApprox` by unrolling the CDF count into an inline sum of `CASE WHEN` expressions <- this results in large SQL statements we may need to contend with later
- Fixes `ALTER TABLE ... RENAME TO` resolving against the session's default schema instead of the write schema by fully qualifying the rename target on all non-BigQuery dialects

**Test plan**

- [X] Incremental pipeline runs with variety of metric types
- [x] Incremental pipeline matches results with CUPED off with non-incremental mode, both full refresh and then with an incremental refresh
- [X] Pre-computed dimensions are materialized


packages/back-end/src/integrations/sql/queries/alter-new-incremental-units-query.ts
```diff
@@ -6,9 +6,19 @@ export function getAlterNewIncrementalUnitsQuery(
   dialect: SqlDialect,
   params: AlterNewIncrementalUnitsQueryParams,
 ): string {
+  // BigQuery requires the rename target to be a bare table name and resolves
+  // it within the source table's dataset. Most other engines (Snowflake,
+  // Postgres, etc.) resolve a bare target against the session's current
+  // schema instead — which silently breaks when the connection's default
+  // schema is not the pipeline's write schema. Use the fully-qualified target
+  // everywhere except BigQuery to remove that footgun.
```
ahdriel: We also need to validate this against Presto/Trino.

I think for this scenario, it depends on the connector -- but I remember having to remove the full name and use just the table name for Presto/Hive setup.

I would play it safe and make this change effective only for Snowflake for now and keep BQ & Trino with the previous format.

packages/back-end/src/integrations/dialects/snowflake.ts
```diff
@@ -18,6 +18,62 @@ export const snowflakeDialect: SqlDialect = {
   hllAggregate: (col: string) => `HLL_ACCUMULATE(${col})`,
   hllReaggregate: (col: string) => `HLL_COMBINE(${col})`,
   hllCardinality: (col: string) => `HLL_ESTIMATE(${col})`,
+  // Snowflake has no native KLL sketch; we approximate via the t-digest
+  // family (APPROX_PERCENTILE_ACCUMULATE/COMBINE/ESTIMATE). t-digest state
+  // is returned as an OBJECT (JSON), so quantileSketch storage columns use
+  // OBJECT (not BINARY like HLL — see getDataType below). Quantile results
+  // will differ slightly from BigQuery's KLL_QUANTILES for the same input.
```
ahdriel: question: do we need to callout BQ here explicitly?

ahdriel (approved):
Awesome addition! On my small dataset I saw the units table data read, when refreshing 1 experiment, go from 1.4MB to 500kb (~65% reduction).

It will certainly be even more if the units table has more concurrent experiments than my test data set.

packages/back-end/src/integrations/dialects/snowflake.ts
```diff
@@ -18,6 +18,62 @@ export const snowflakeDialect: SqlDialect = {
...
+  // only consumer is quantileSketchRankApprox, and the obvious implementation
+  // (ARRAY_CONSTRUCT of per-grid-point APPROX_PERCENTILE_ESTIMATE calls)
+  // can't be used inside a scalar subquery whose FROM is TABLE(FLATTEN(...))
+  // when the array depends on outer-row columns — Snowflake's optimizer
+  // refuses to decorrelate ("Unsupported subquery type cannot be
+  // evaluated"). We work around that in quantileSketchRankApprox below;
+  // falling through to baseDialect.quantileSketchExtractQuantiles (which
+  // throws) signals clearly that the primitive isn't safely usable outside
+  // that path.
+  quantileSketchInit: (col: string) => `APPROX_PERCENTILE_ACCUMULATE(${col})`,
```
ahdriel, line 25: suggestion: can we add unit tests for these new functions to ensure they are being built as we expect?

packages/back-end/src/integrations/dialects/snowflake.ts
```diff
@@ -18,6 +18,62 @@ export const snowflakeDialect: SqlDialect = {
...
+    // batch call), but row-level cost is still linear in numQuantiles.
+    //
+    // NULL-handling: if the sketch is NULL, every APPROX_PERCENTILE_ESTIMATE
+    // returns NULL, so each CASE WHEN takes the ELSE 0 branch (NULL < x is
+    // unknown/false). The sum is 0, and COALESCE is a no-op. The COALESCE
+    // only fires when nEventsCol is NULL (0 * NULL = NULL).
+    const sumBelow = Array.from(
+      { length: numQuantiles + 1 },
+      (_, i) =>
+        `CASE WHEN APPROX_PERCENTILE_ESTIMATE(${sketchCol}, ${(
```
ahdriel, line 58: question: does this mean we are doing `numQuantiles + 1` `APPROX_PERCENTILE_ESTIMATE` calls? Are we concerned about the potential performance impact here? The comment says `but row-level cost is still linear in numQuantiles` so I think this is known, but I wanted to flag to make sure we validated this won't make the query super expensive.

lukesonnet: It's not cheap, but it should use the same t-digest so it just has to access that multiple times; the query planner is smart enough not to build the digest multiple times.

packages/back-end/src/integrations/sql/queries/alter-new-incremental-units-query.ts
lukesonnet: moved it to just make this change for snowflake.

lukesonnet (commented):
tested and values match, will re-trigger greptile and merge.

packages/back-end/src/integrations/dialects/snowflake.ts
ahdriel: Nice! yeah -- I was just looking into some query planning for BQ around APPROX_QUANTILES and the compiler does share the sketch in this case (but not for slices given that each slice is a new metric, so we need to calculate it separately).

PR #5941, opened 2026-05-25
Remove old code, cleanup org scoping

jdorn:
### Features and Changes

- Removed "GithubIntegration" - This was the original, experimental implementation for Code References.  The code was pushed behind a feature flag, but it was never actually made live. We ended up implementing Code References in a completely different way, but never cleaned up this old code.
- Cleaned up how org scoping was applied to mongoose model queries to be more consistent and secure. Temporary improvement before doing a full migration to BaseModel


tzjames (approved):
Besides the one failing test, this looks good to go.

PR #5945, opened 2026-05-26
Lj/session replay playground

lillialjackson:
### Features and Changes


packages/back-end/src/services/clickhouse.ts
```diff
@@ -143,3 +145,132 @@ export async function updateMaterializedColumns({
...
+  const limit = Math.max(1, Math.min(100, Math.floor(options?.limit ?? 100)));
+  const offset = Math.max(0, Math.floor(options?.offset ?? 0));
+  // Always exclude soft-deleted from the list; callers that explicitly want
+  // them must use a future bulk-list-by-id endpoint that doesn't go through
+  // this function. Add this BEFORE the user-supplied conditions so it can't
+  // be overridden by an empty filter.
+  const allConditions = ["deleted_at IS NULL", ...conditions];
+  const where = `WHERE ${allConditions.join(" AND ")}`;
+
+  const { rows } = await integration.runQuery(`
```
lillialjackson, line 234: Does this Select also need to go through the CLS as opposed to this backend service? This is for the Replay view to list sessions from CH.

packages/back-end/src/util/secrets.ts
```diff
@@ -76,6 +76,9 @@ export const S3_BUCKET = process.env.S3_BUCKET || "";
 export const S3_REGION = process.env.S3_REGION || "us-east-1";
 export const S3_DOMAIN =
   process.env.S3_DOMAIN || `https://${S3_BUCKET}.s3.amazonaws.com/`;
+// Optional override for S3-compatible endpoints (MinIO, R2, etc.).
+// Leave empty to use AWS S3.
+export const S3_ENDPOINT = process.env.S3_ENDPOINT || "";
```
lillialjackson, line 81: dev env related can remove if necessary

packages/back-end/src/util/secrets.ts
```diff
@@ -97,6 +100,17 @@ if ((prod || !IS_LOCALHOST) && !IS_CLOUD && JWT_SECRET === "dev") {
...
 
+// Optional override for the session-replay S3 bucket — replay payload chunks
+// are stored in their own bucket so that the back-end's read role can be
+// scoped separately from the general uploads bucket (`S3_BUCKET`). Leave
+// empty to disable signed-URL session-replay reads.
+export const S3_SESSION_REPLAY_BUCKET =
+  process.env.S3_SESSION_REPLAY_BUCKET || "";
+// Optional override for the role used to read the session-replay bucket. Falls
+// back to AWS_ASSUME_ROLE when unset, so single-role setups need no extra env.
+export const S3_SESSION_REPLAY_ASSUME_ROLE =
```
lillialjackson, line 111: same with these two, I am not sure if we need the same role settings I needed to get things working locally once this is in prod

packages/front-end/components/SessionReplay/RrwebPlayer.tsx
```diff
@@ -0,0 +1,159 @@
...
+ * parent renders this with `key={sessionId}` so React physically
+ * unmounts/remounts the host DOM on session switch — that's what makes
+ * teardown bulletproof in rrweb-player 2.0.0-alpha.20, where the inner
+ * Replayer's destroy() doesn't fully tear down its Timer and mirror.
+ * Reusing the same Player instance across sessions leaves zombie state
+ * that logs "Looks like your replayer has been destroyed" + "Node with
+ * id N not found" against the new session's events.
+ *
+ * Also silences rrweb's chatty warnings three ways:
+ *   1. showWarning:false prop (the documented knob, but the prop
```
lillialjackson, line 59: right now I need to make some adjustments to how we process events, since we allow users to maintain a session even through refreshes this means sometimes there is more than one snapshot or some of our sessions are stitched together. rrweb doesn't like this and in dev will dump the warnings and the events into logs which is very noisy, need to revisit this, but for now turned log warnings from rrweb off

packages/sdk-js/src/plugins/auto-attributes.ts
```diff
@@ -9,6 +9,26 @@ export type AutoAttributeSettings = {
   uuidKey?: string;
   uuid?: string;
   uuidAutoPersist?: boolean;
+  /**
+   * Attribute key under which the per-browser-session id is exposed.
+   */
+  sessionIdKey?: string;
```
lillialjackson, line 15: Verify that this is what we want for sessionId. I wanted to be able to leverage sessionStorage so that we could tie the session replays to other session data like errors, I did not see where we were actually setting up a shared sessionID currently, so I set it up to check if there is a session id and if not set one. LMK if I am missing something there

packages/sdk-js/src/plugins/session-replay.ts
```diff
@@ -0,0 +1,885 @@
...
+ * 401 (bad clientKey) or 403 (org/key forbidden), retrying produces the
+ * same failure forever — and because flushes are also triggered by the
+ * pre-push size cap, a misconfigured key turns into a hot loop: emit →
+ * over cap → flush → 401 → revert + re-buffer → over cap again → flush
+ * again, multiple requests per second. After this many consecutive
+ * auth failures we stop the recorder entirely, on the assumption that
+ * the configuration is wrong and won't fix itself within a page load.
+ * Tab refresh resets the counter (fresh plugin instance, fresh roll).
+ */
+const AUTH_FAILURE_LIMIT = 3;
```
lillialjackson, line 130: I noticed an issue due to a misconfigured instance of the plugin that caused a bunch of 401s repeatedly if the plugin had a bad clientKey. Obviously the real issue was that the plugin was set up incorrectly but seemed like bad behavior to keep getting 401s. What do y'all think? is this worth checking here?

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -238,6 +238,7 @@ export type Options = {
   /** @deprecated */
   subscribeToChanges?: boolean;
   enableDevMode?: boolean;
+  captureLogs?: boolean;
```
lillialjackson, line 241: currently do not have access to feature flag and experiment evaluations except through the logs, this feels like a hack but we would need to put some effort into exposing those evaluations another way. Thoughts on this?

packages/sdk-js/rollup.config.mjs
```diff
@@ -24,6 +24,34 @@ const terserSettings = terser({
 });
 
 export default [
+  {
+    // Plugins bundle — rrweb is inlined so consumers don't need it as a dep
```
lillialjackson, line 28: feedback on this pls

packages/back-end/src/util/secrets.ts
gazzdingo: this might be nice for self hosted to be able to use this. but also we might want to limit this to only cloud users 
also graham is doing some s3 work so you might want to see what his is thinking

packages/sdk-js/src/plugins/auto-attributes.ts
gazzdingo: yeah im not sure if there is a session id here i can dive a little deeper into it but bryce or jermey might know from the top of there head

packages/sdk-js/src/types/growthbook.ts
gazzdingo: im worried that this is going to cause leaking with the devtools but i need to look deeper into it. i wonder if we can have another log that is used for the session replays then we can more tightly control stuff in the future

packages/sdk-js/rollup.config.mjs
gazzdingo: how much does it add to the bundle size since we want to keep that super small for load time

PR #5954, opened 2026-05-27
Added backend changes required for the new visual editor

Auz:


Auz:
@cursor review

Auz:
@greptileai review

packages/back-end/src/api/visual-editor-ai/getBootstrap.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+  >();
+  for (const cs of changesets) {
+    if (!changesetByExperiment.has(cs.experiment)) {
+      changesetByExperiment.set(cs.experiment, {
+        id: cs.id,
+        urlPatterns: cs.urlPatterns ?? [],
+      });
+    }
+  }
```
jdorn: This does not support experiments with multiple changesets

packages/back-end/src/api/visual-editor-ai/getLibraryImages.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  const orgId = req.organization.id;
+  const limit = Math.min(req.query.limit ?? 1000, 1000);
+  // Sentinel log so dev users can confirm the request actually
+  // reached the route handler vs hanging upstream (a stuck event
+  // loop, a misconfigured proxy, the wrong API host etc.).
+  logger.info(
+    { orgId, limit },
+    "[visual-editor-ai/library] listing image library",
+  );
```
jdorn, line 42: Change to debug log instead of info

packages/back-end/src/api/visual-editor-ai/postCreateExperiment.ts
```diff
@@ -0,0 +1,232 @@
...
+  // Build the experiment shell. Status "draft" so it doesn't fire on
+  // anyone until the user finishes setup in GrowthBook. Two-variant
+  // (Control + Variant 1) is the right default for a visual experiment;
+  // they can add more variants from the side panel later via add-variant.
+  const controlId = uniqid("var_");
+  const variantId = uniqid("var_");
+  const experimentToCreate: Partial<ExperimentInterface> = {
+    name,
+    status: "draft",
+    hashAttribute: "id",
```
jdorn: Cannot hardcode `id` here.  Get the first identifier attribute from the org settings and use that instead.

```ts
org.settings.attributeSchema?.find(a => a.hashAttribute) || "id"
```

Or better yet, instead of guessing, prompt the user for how they want to split traffic and pass an explicit hash attribute into this endpoint.

packages/front-end/components/OpenVisualEditorLink.tsx
```diff
@@ -59,7 +59,15 @@ export async function openVisualEditor({
...
   });
 
-  if (!bypassChecks) {
+  // Skip the extension-detection probe in local dev. The probe pings the
+  // production Chrome Web Store extension ID, which won't match a locally
+  // unpacked dev build (Chrome assigns those a random ID).
+  const isLocalDev =
+    typeof window !== "undefined" &&
+    (window.location.hostname === "localhost" ||
+      window.location.hostname === "127.0.0.1");
```
jdorn: I don't think this change is right. We don't care about where the GrowthBook app is hosted (localhost vs prod).  We care about if the user has the published browser extension or is using a local dev build of it.  This is going to be extremely rare - only devs actively working on the visual editor extension.  Lots of people are running GrowthBook locally, but still want to use the published browser extension.

packages/back-end/src/api/visual-editor-ai/getBootstrap.ts
Auz: Addressed

packages/back-end/src/api/visual-editor-ai/postCreateExperiment.ts
Auz: fixed

packages/back-end/src/api/visual-editor-ai/getLibraryImages.ts
Auz: warn okay?

jdorn: This is a GET request.  Will this run frequently (like every time someone uses the visual editor)?  If so, anything higher than debug will clog up the logs. If this is a relatively rare event, then it's fine.

jdorn (approved):

PR #5958, opened 2026-05-27
Improve query cancellation (dependency management + BQ location bug)

lukesonnet:
### Problem

Cancelling a snapshot left some jobs running, esp. in BigQuery. Two root causes:

1. **Missing region.** `BigQuery.cancelQuery` reconstructed jobs without `location`. BQ's job-control API requires it for any non-US/EU multi-region dataset and returns 404 otherwise. The 404 was swallowed at `logger.debug` so it was invisible.
2. **Stale pointer status drove the cancel path.** `cancelQueries` filtered the BQ-cancel block on `snapshot.queries[i].status === "running"`. That field is the snapshot pointer, written by the *original* runner's polling loop. The cancel controller spins up a fresh `ExperimentResultsQueryRunner` that just reads the snapshot doc and never polls, so pointers often still read `"queued"` while the Mongo `Query` docs were `"running"` with `externalId` set. Result: 0 (or 1 racy) cancels sent to BQ even when 11 queries were in flight.

### Changes

- **`externalIdMetadata` on `QueryInterface`** (shared types + Mongo schema). Optional `Record<string, string>` for integration-managed job side-data. `ExternalIdCallback` and `SourceIntegrationInterface.cancelQuery` both gained an optional `metadata` parameter.
- **BigQuery** now persists `job.location` via `setExternalId` and threads it back through `client.job(id, { location })` on cancel. Cancel/error logs upgraded from `debug` to `info`/`warn` so failures are visible.
- **`QueryRunner.cancelQueries`** rewritten:
  - Drives off `pendingIds` (both `running` and `queued` pointers), not pointer-running.
  - Marks docs failed in Mongo *before* issuing warehouse cancels — pairs with a new conditional `updateQueryIfQueued` promote in `executeQuery` so pending timers can't resurrect a cancelled query and fire a fresh external job during the cancel window. Also reflects the cancel in the UI immediately.
  - Dedupes external jobs by id (cached copies share their upstream's `externalId` via `cachedQueryUsed`).
- **New `QueryModel` helpers**: `updateQueryIfQueued` (conditional queued→running promote) and `markPendingQueriesAsFailed` (bulk mark with `status ∈ {running, queued}` filter).

### Notes

- `executeQuery`'s conditional promote applies to all integrations, not just BigQuery. Behavior change: if a cancel lands between scheduling and execution, the runner now bails instead of starting the external job.
- `cancelDataSourceQuery` (force-cancel UI endpoint) passes metadata through but keeps its existing `debug`-level error log; intentionally out of scope here.
- No tests added — per workspace policy on routers/controllers/models. The new `QueryModel` helpers are thin Mongo wrappers; the race ordering in `cancelQueries` isn't unit-testable without mocking Mongo + integration.

### Test plan

- [X] Cancel an in-flight non-US/EU BigQuery snapshot; verify all running jobs flip to `CANCELLED` in the BQ console (not `DONE`).
- [X] Verify backend logs show `Marked queries as cancelled in Mongo` (attempted == affected), `Cancelling N external jobs` (N > 0), and one `BigQuery cancel request accepted` per job.
- [X] Cancel during the queued→running window; verify no new BQ jobs spawn afterwards and `Skipping execution — query no longer queued` is logged for the affected queries.


packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -580,82 +582,119 @@ export abstract class QueryRunner<
...
+    // executeQuery, this stops a queued query from being promoted (and
+    // firing a fresh external job) while parallel cancel calls are in
+    // flight. Also reflects the cancel in the queries-log UI immediately.
+    if (pendingIds.length) {
+      const affected = await markPendingQueriesAsFailed(
+        this.context,
+        pendingIds,
+        "Query cancelled by user",
+      );
+      logger.info(
```
jdorn: This should be debug, not info

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -580,82 +582,119 @@ export abstract class QueryRunner<
...
+        {
+          datasourceId: this.integration.datasource.id,
+          modelId: this.model.id,
+          externalJobs: externalJobs.map((j) => ({
+            id: j.id,
+            metadataKeys: j.metadata ? Object.keys(j.metadata) : [],
+          })),
+        },
+        `Cancelling ${externalJobs.length} external jobs`,
+      );
```
jdorn, line 666: Same, debug instead of info

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -735,16 +774,32 @@ export abstract class QueryRunner<
...
         startedAt: new Date(),
         status: "running",
         heartbeat: new Date(),
       });
+      if (!promoted) {
+        clearInterval(timer);
+        logger.info(
+          { queryId: doc.id, modelId: this.model.id },
+          "Skipping execution — query no longer queued (likely cancelled)",
+        );
```
jdorn: Same, debug instead of info

jdorn (approved):
Approved with minor notes about log levels

lukesonnet:
Choosing to ignore Greptile's complaint about "Success path can overwrite cancel-triggered "failed" status". In this case, I think writing the query status as succeeded is fine even if the snapshot model thinks it's cancelled. That is technically correct, the query succeeded, it may have returned data or mutated the warehouse, so we kind of want to behave as such. It's a bit odd, but because the other queries will have been set to failed, it shouldn't trigger any dependent queries.

PR #5960, opened 2026-05-28
implement a variety of feature "has" search filters

bryce-fitzsimons:
New lazyloaded search filters for a variety of both general "has some" and specific "has this exact thing" filters. Generally searches the live revision + unpublished drafts only.

image


gazzdingo (approved):
LGTM only one small comment 

 Duplicate searchValue source of truth (pages/features/index.tsx)
  The set-state-during-render block plus searchValueRef duplicates
  searchInputProps.value into local state purely to feed contentSearchParams. You can
  drop the useState/useRef/render-time setState entirely and derive directly:
  ```const contentSearchParams = useMemo(
    () => extractContentSearchParams(searchInputProps.value),
    [searchInputProps.value],
  );
```
  This removes a fragile pattern and a redundant state copy.

PR #5964, opened 2026-05-28
Add incremental refresh opt-in from ephemeral mode

lukesonnet:
Add opt-in to Incremental Mode for some experiments while in Ephemeral mode, to aid in testing incremental refresh.

image


packages/shared/src/enterprise/pipeline.ts
```diff
@@ -5,6 +5,59 @@ import type {
...
+  return true;
+}
+
+/**
+ * True when the data source has any experiments configured to use incremental
+ * refresh — either by setting `mode === "incremental"` or by opting individual
+ * experiments in. Used by the UI to decide whether to run incremental
+ * permission validation before saving.
+ */
+export function pipelineRequiresIncrementalConfig(
```
ahdriel: nit: looks like this function is not being used?

packages/front-end/components/Settings/EditDataSource/DataSourcePipeline/EditDataSourcePipeline.tsx
```diff
@@ -80,27 +83,45 @@ export const EditDataSourcePipeline = ({
...
     const formValues = form.getValues();
     if (formValues.mode === "disabled" || !validateBeforeSaving) {
       return true;
     }
 
+    // If any experiment opts into incremental refresh, validate against the
+    // incremental config (create + insert + drop probes). Otherwise validate
+    // using the selected default mode.
+    const optInCount = formValues.incrementalOptInExperimentIds.length;
+    const modeToValidate: DataSourcePipelineMode =
```
ahdriel, line 106: I think this is where the unused function could be applied to; but happy to keep the logic here as-is too.

packages/back-end/src/enterprise/services/data-pipeline.ts
```diff
@@ -62,30 +63,22 @@ export async function validateIncrementalPipeline({
     );
   }
 
-  // Check if pipeline mode is set to incremental
+  // Check if this experiment is enabled for incremental refresh on the
```
ahdriel: nit: should we remove this comment?

now `isExperimentIncrementalEnabled` provides enough context IMO

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -402,6 +431,7 @@ export async function updateDataSource(
       datasource,
       updates.settings,
     );
+    validatePipelineSettingsInvariants(updates.settings);
```
ahdriel: question: we have this only on `update`, should we add it to `create` as well?

packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -387,6 +387,35 @@ export function hasActualChanges(
...
+// existing customers updating an unrelated field on a data source with
+// pre-existing (potentially non-strict) pipeline settings aren't affected.
+function validatePipelineSettingsInvariants(
+  settings: Partial<DataSourceSettings>,
+) {
+  const pipelineSettings = settings.pipelineSettings;
+  if (!pipelineSettings) return;
+
+  const optInCount =
+    pipelineSettings.incrementalOptInExperimentIds?.length ?? 0;
```
ahdriel: question: does the UI set `incrementalOptInExperimentIds` to empty when we select `None` for pipeline mode?

I am thinking we might have a conflict of some sorts if we don't do that when disabling pipeline mode but this field is still set from the previous configuration.

ahdriel (approved):

packages/back-end/src/models/DataSourceModel.ts
lukesonnet: Changed the logic in the form to

```
       incrementalOptInExperimentIds: supportsIncremental &&
          optInIds.length > 0 &&
          formValues.mode === "ephemeral"
            ? optInIds
            : undefined,
```

And I'll also add this guard here.

packages/back-end/src/models/DataSourceModel.ts
lukesonnet: It isn't in the creation form, so it's not relevant? IDK what the drawbacks are though so I'll add it.

packages/front-end/components/Settings/EditDataSource/DataSourcePipeline/EditDataSourcePipeline.tsx
lukesonnet: removed the unused fn ty.

packages/shared/src/enterprise/pipeline.ts
lukesonnet: removed ty.

packages/front-end/components/Settings/EditDataSource/DataSourcePipeline/EditDataSourcePipeline.tsx
```diff
@@ -125,6 +147,11 @@ export const EditDataSourcePipeline = ({
         includedExperimentIds: formValues.applyToAllExperiments
           ? undefined
           : formValues.includedExperimentIds,
+        // Only persist when the data source supports incremental and the
+        // user has actually opted experiments in. Storing `undefined`
+        // matches the type's "absent => no overrides" semantics.
+        incrementalOptInExperimentIds:
+          supportsIncremental && optInIds.length > 0 ? optInIds : undefined,
```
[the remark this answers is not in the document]
lukesonnet: Fixed ty.

PR #5972, opened 2026-05-29
perf(view-queries): Use preview for sql + virtualize results

ahdriel:
### Features and Changes

In the Experiment > View Queries modal we have 2 performance bottlenecks that are client-side.

1. For each query, we instantiate a full Prism Highlighter
	This is specially impactful if we have 20 queries and each has more than 4k lines. This would freeze the UI completely.
	
2. We would render the full results for each query using a native `<table>` component
	Meaning if we have many queries with many results, this would also generate a lot of DOM nodes and the browser would freeze.

We fixed this performance issue by:
- Rendering a cheaper SQL Highlighter when it is not expanded
- Virtualizing the results table

In a test scenario for a single query block with 4200 lines, 1000 columns and 1000 rows, we went from `10s` to `30ms` of render time. Increasing to 60 queries with the same configuration would crash the tab with the previous implementation and now it finishes in ~500ms.

### Testing

- Unit tests
- Perf page to compare before and after (removed from this PR)

### DIFF
--- a/packages/front-end/components/SyntaxHighlighting/Code.tsx
+++ b/packages/front-end/components/SyntaxHighlighting/Code.tsx
@@ -4,7 +4,6 @@ import React, {
   ReactElement,
   Suspense,
   lazy,
-  useEffect,
   useMemo,
   useState,
 } from "react";
@@ -86,11 +85,11 @@ export default function Code({
 
   const [isExpanded, setIsExpanded] = useState(!expandable);
   const [hasExpanded, setHasExpanded] = useState(!expandable);
-  useEffect(() => {
-    if (isExpanded) {
-      setHasExpanded(true);
-    }
-  }, [isExpanded]);
+
+  const expand = () => {
+    setIsExpanded(true);
+    setHasExpanded(true);
+  };
 
   const { theme } = useAppearanceUITheme();
 
@@ -188,7 +187,7 @@ export default function Code({
             style={{ cursor: "pointer" }}
             onClick={async (e) => {
               e.preventDefault();
-              setIsExpanded(!isExpanded);
+              isExpanded ? setIsExpanded(false) : expand();
             }}
           >
             {isExpanded ? <FaCompressAlt /> : <FaExpandAlt />}
@@ -206,7 +205,7 @@ export default function Code({
           showLineNumbers={showLineNumbers}
           startingLineNumber={startingLineNumber ?? 1}
           previewOnly
-          onPreviewExpand={() => setIsExpanded(true)}
+          onPreviewExpand={expand}
         />
       ) : (
         <Suspense
--- a/packages/front-end/components/SyntaxHighlighting/Code.tsx
+++ b/packages/front-end/components/SyntaxHighlighting/Code.tsx
@@ -4,6 +4,8 @@ import React, {
   ReactElement,
   Suspense,
   lazy,
+  useEffect,
+  useMemo,
   useState,
 } from "react";
 import { FaCompressAlt, FaExpandAlt } from "react-icons/fa";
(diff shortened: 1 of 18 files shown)


packages/front-end/components/SyntaxHighlighting/Code.tsx
```diff
@@ -81,27 +84,44 @@ export default function Code({
...
   if (language === "sh") language = "bash";
 
-  const [expanded, setExpanded] = useState(!expandable);
+  const [isExpanded, setIsExpanded] = useState(!expandable);
+  const [hasExpanded, setHasExpanded] = useState(!expandable);
+  useEffect(() => {
+    if (isExpanded) {
+      setHasExpanded(true);
+    }
+  }, [isExpanded]);
```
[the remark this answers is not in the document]
msamper: +1

packages/front-end/components/SyntaxHighlighting/Code.tsx
```diff
@@ -81,27 +84,44 @@ export default function Code({
...
 
   const { theme } = useAppearanceUITheme();
 
   const enoughLines = code.split("\n").length > 8;
```
msamper, line 97: It seems like right now this collapses a query when it's longer than 8 lines regardless of the total number of queries which is sort of a regression because now folks without performance issues have an extra click to view the whole query. I wonder if we should be smarter about when we collapse the queries. Maybe using total number of queries & lines per query as the threshold. Also I think queries will almost always be more than 8 lines.

packages/front-end/components/SyntaxHighlighting/PrismFallback.tsx
```diff
@@ -9,26 +11,196 @@ export interface Props {
   };
   language: Language;
   className?: string;
+  showLineNumbers?: boolean;
+  startingLineNumber?: number;
+  previewOnly?: boolean;
+  onPreviewExpand?: () => void;
+}
+
+const SQL_PREVIEW_LINES = 16;
```
msamper, line 20: Should `enoughLines` match this number? Wondering if the query is somewhere between 9-16 lines if we'd still render the `View Full Query` button even though theres nothing else to see

msamper (approved):
Looks good overall. Just a few nit comments. Curious to see what the virtualized version of the the query result table looks like compared to the regular one. Wasn't able to test since I don't have an example with a large amount of queries

packages/front-end/components/SyntaxHighlighting/Code.tsx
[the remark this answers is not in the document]
ahdriel: Fixed

lukesonnet (approved):
QA'd briefly and I don't mind the UI, and the copy button still works :). Lots of lines here and a new dependency, so I leave that to your best judgement.

I think it would also be fine for it to just be like 3 lines with 0 syntax highlighting and then "View Full Query" as you currently have it, but with only 3 lines the "View Full Query" button it would always appear so people would just be able to one click into seeing the full query (and it might even be fine to have a new UI that just shows "View Full Query" instead of any lines...)

But that's just a comment, listen to your heart, the performance improvements will be worth it.

Another (pre-existing) issue is that the data in this modal doesn't mutate/refresh as queries finish, I don't think.

packages/front-end/components/SyntaxHighlighting/Code.tsx
lukesonnet: Yeah I think something like this is reasonable, but I also think most folks are clicking "copy" or "expand" anyways, or we should coach them to click "expand" since it's way better.

ahdriel: For a 3k lines SQL snippet it takes ~500ms to render with the full code. So for now I am keeping the whole `View queries` modal virtualized for the queries given the impact.

Talking to Kari there's also a design where we might render only 1 query at a time, and having a picker for each query, which would change how this calculation works.

So I am postponing the 'overall decision if we should virtualize sql queries or not' as we might change the strategy soon. I'll follow-up on this either this week or next.

lukesonnet: Talking to Kari there's also a design where we might render only 1 query at a time, and having a picker for each query, which would change how this calculation works.

I worry this makes finding queries hard. Sometimes I ctrl-f for something and it appears in a query. I understand we're already losing that power with this PR, but I feel a selector will make it even more problematic.

PR #5974, opened 2026-05-29
Add webhook events for saved groups and their revisions

gazzdingo:
Adds webhook/notification events for **saved groups** and their **revisions**, mirroring how features do it.

### Events

**Lifecycle:** `savedGroup.created` · `savedGroup.updated` · `savedGroup.deleted`

**Revisions:** `savedGroup.revision.*` — `created`, `updated` (with a `change` of `values`/`condition`/`metadata`/`archive`), `reviewRequested`, `approved`, `changesRequested`, `commented`, `published`, `discarded`, `rebased`, `reverted`, `reopened`

### How it works

- **Lifecycle events** fire from `SavedGroupModel`'s `afterCreate/afterUpdate/afterDelete` hooks, so they cover every write path (internal API, public REST, and revision publish).
- **Revision events** are dispatched from the controllers/handlers (same call-site pattern features use). A small handler-layer **webhook-adapter registry** (`events/revisionWebhookAdapters.ts`) maps entity type → dispatcher, so the generic `/revision` controller stays entity-agnostic and **adding a new approval type just needs one registry entry**. (Kept separate from `revisions/index` because that adapter loads during model-init and can't import the event pipeline without a circular import.)
- Registers the `savedGroup` resource + payload schemas in `notificationEvents`, and adds Slack/Discord message builders.

### Testing

Verified end-to-end against a live server: all 13 event types fire with correct payloads across **both** the front-end (`/revision` controller) and the public REST API. Unit tests added for the pure helpers (`applyPatchToSnapshot` date-preservation, `deriveChange`).


gazzdingo:
@greptileai

gazzdingo:
@greptileai

bryce-fitzsimons (approved):

PR #5976, opened 2026-05-29
perf(bigquery): Pack quantile in a single array column 

ahdriel:
### Features and Changes

For quantile metrics we compute a confidence-interval grid — a lower/upper bound at each
of the 20 sample-size thresholds in `N_STAR_VALUES` — and previously emitted every bound as
its own SQL column (1 point estimate + `N_STAR_VALUES.length * 2` = 41 columns per metric).
On BigQuery we now pack the whole grid into a single `ARRAY` column instead, dropping a
quantile metric's column cost ~5×.

Since we chunk metrics into queries under a max-columns-per-query budget, this lets
many more quantile metrics fit per query, reducing
the number of queries and repeated fact-table scans per snapshot.

- **BigQuery-only, opt-in via dialect:** new `hasArrayQuantileGrid()` capability (default off);
  all other warehouses keep the existing scalar columns unchanged.
- **Statistically identical:** Just changes how the query is created and read

### DIFF
--- a/packages/back-end/src/integrations/dialects/bigquery.ts
+++ b/packages/back-end/src/integrations/dialects/bigquery.ts
@@ -147,6 +147,8 @@ export const bigQueryDialect: SqlDialect = {
     const countBelow = `(SELECT COUNT(*) FROM UNNEST(${cdfArray}) AS p WHERE p < ${thresholdCol})`;
     return `COALESCE(${countBelow} * ${nEventsCol} / ${numQuantiles}.0, 0)`;
   },
+  hasArrayQuantileGrid: () => true,
+  arrayLiteral: (elements: string[]) => `[${elements.join(", ")}]`,
   percentileApprox: (value: string, quantile: string | number) => {
     const multiplier = APPROX_QUANTILES_MULTIPLIER;
     const quantileVal = Number(quantile)
--- a/packages/back-end/src/integrations/dialects/bigquery.ts
+++ b/packages/back-end/src/integrations/dialects/bigquery.ts
@@ -148,7 +148,13 @@ export const bigQueryDialect: SqlDialect = {
     return `COALESCE(${countBelow} * ${nEventsCol} / ${numQuantiles}.0, 0)`;
   },
   hasArrayQuantileGrid: () => true,
-  arrayLiteral: (elements: string[]) => `[${elements.join(", ")}]`,
+  // BigQuery rejects NULL containing arrays in a query result, so collapse the
+  // whole grid to NULL when an array would contain NULLs.
+  // The quantile-grid elements are all-or-nothing, so we can test the first element.
+  quantileGridArrayLiteral: (elements: string[]) =>
+    elements.length > 0
+      ? `IF(${elements[0]} IS NULL, NULL, [${elements.join(", ")}])`
+      : `[${elements.join(", ")}]`,
   percentileApprox: (value: string, quantile: string | number) => {
     const multiplier = APPROX_QUANTILES_MULTIPLIER;
     const quantileVal = Number(quantile)
--- a/packages/back-end/src/integrations/sql/columns/quantile-bounds-from-query-response.ts
+++ b/packages/back-end/src/integrations/sql/columns/quantile-bounds-from-query-response.ts
@@ -18,12 +18,29 @@ export function getQuantileBoundsFromQueryResponse(
 
     const smallestNStar = Math.min(...N_STAR_VALUES);
 
-    // process grid for quantile data
-    N_STAR_VALUES.forEach((n) => {
-      const lowerColumn = `${prefix}quantile_lower_${n}`;
-      const upperColumn = `${prefix}quantile_upper_${n}`;
-      if (row[lowerColumn] === undefined || row[upperColumn] === undefined)
-        return;
+    // For datasources that we can pack the values in a single array column
+    const gridArray = row[`${prefix}quantile_grid`];
+    const isArrayGrid = Array.isArray(gridArray);
+    const expectedGridLength = N_STAR_VALUES.length * 2;
+    if (isArrayGrid && gridArray.length !== expectedGridLength) {
+      throw new Error(
+        `Expected ${prefix}quantile_grid array of length ${expectedGridLength}, got ${gridArray.length}. ` +
+          `This indicates a mismatch between SQL generation (getQuantileGridColumns / getQuantileSketchGridColumns) ` +
+          `and N_STAR_VALUES on the read side.`,
+      );
+    }
+
+    const getLower = isArrayGrid
+      ? (k: number) => gridArray[2 * k]
+      : (k: number) => row[`${prefix}quantile_lower_${N_STAR_VALUES[k]}`];
+    const getUpper = isArrayGrid
+      ? (k: number) => gridArray[2 * k + 1]
+      : (k: number) => row[`${prefix}quantile_upper_${N_STAR_VALUES[k]}`];
+
+    N_STAR_VALUES.forEach((n, k) => {
+      const lowerVal = getLower(k);
+      const upperVal = getUpper(k);
+      if (lowerVal === undefined || upperVal === undefined) return;
 
       if (
         // if nstar is smaller, or if it's the smallest nstar, proceed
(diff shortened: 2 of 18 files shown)


packages/back-end/src/integrations/sql/columns/quantile-bounds-from-query-response.ts
```diff
@@ -18,12 +18,29 @@ export function getQuantileBoundsFromQueryResponse(
...
+      ? (k: number) => gridArray[2 * k]
+      : (k: number) => row[`${prefix}quantile_lower_${N_STAR_VALUES[k]}`];
+    const getUpper = isArrayGrid
+      ? (k: number) => gridArray[2 * k + 1]
+      : (k: number) => row[`${prefix}quantile_upper_${N_STAR_VALUES[k]}`];
+
+    N_STAR_VALUES.forEach((n, k) => {
+      const lowerVal = getLower(k);
+      const upperVal = getUpper(k);
+      if (lowerVal === undefined || upperVal === undefined) return;
```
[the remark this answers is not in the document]
lukesonnet: I think this fix makes sense?

packages/back-end/src/integrations/dialects/bigquery.ts
```diff
@@ -147,6 +147,8 @@ export const bigQueryDialect: SqlDialect = {
     const countBelow = `(SELECT COUNT(*) FROM UNNEST(${cdfArray}) AS p WHERE p < ${thresholdCol})`;
     return `COALESCE(${countBelow} * ${nEventsCol} / ${numQuantiles}.0, 0)`;
   },
+  hasArrayQuantileGrid: () => true,
+  arrayLiteral: (elements: string[]) => `[${elements.join(", ")}]`,
```
lukesonnet: This construction can fail because it can return an array with `NULL` which is not allowed in a final return statement by BigQuery: https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/data-types#array_type

We instead need to somehow error out the whole array if any element is NULL (which is possible if there's only NULL input data) or somehow filter out NULL elements when only some elements are NULL (I genuinely don't know if this will ever happen).

So it might be easiest just to say "if `m0_quantile` NULL then return NULL for the whole grid instead of trying to filter down the array and keep a record of which elements (i.e. which quantiles in the grid) still remain. I'm not sure if the latter case is ever meaningful, so it probably isn't worse the added complexity.

Should check in with the roboto.

lukesonnet (changes requested):

packages/back-end/src/integrations/dialects/bigquery.ts
lukesonnet: The robot says:

## Confirming the problem

You're right. `arrayLiteral` just emits `[e0, e1, …]`, and each element comes from `getQuantileGridArrayColumn`:

```7:18:packages/back-end/src/integrations/sql/columns/quantile-grid-array-column.ts
export function getQuantileGridArrayColumn(
  dialect: SqlDialect,
  quantile: number,
  prefix: string,
  boundExpr: (bound: number) => string,
): string {
  const elements = N_STAR_VALUES.flatMap((nstar) => {
    const { lower, upper } = getQuantileBoundValues(quantile, 0.05, nstar);
    return [boundExpr(lower), boundExpr(upper)];
  });
  return `, ${dialect.arrayLiteral(elements)} AS ${prefix}quantile_grid`;
}
```

Each `boundExpr` is either `percentileApprox` (`APPROX_QUANTILES(value, 10000 IGNORE NULLS)[OFFSET(...)]`) or `quantileSketchExtractPoint` (`KLL_QUANTILES.EXTRACT_POINT_FLOAT64(...)`). Both yield `NULL` when the group has no non-null input, and BigQuery rejects a `NULL`-containing array in a query result.

## On the "partial NULL" case — it can't happen here

Your instinct is correct, and it's actually provable for both element generators:

- `APPROX_QUANTILES(value, 10000 IGNORE NULLS)` returns `NULL` (the whole aggregate) only when there are **zero** non-null values. If there's ≥1 non-null value, every `[OFFSET(k)]` (for any `k` in `[0, 10000]`) is non-null. So all grid elements are null or none are.
- `KLL_QUANTILES.EXTRACT_POINT_FLOAT64(sketch, q)` is null only for an empty/null sketch, which is again all-or-nothing regardless of `q`.

So it's strictly all-or-nothing, keyed on whether the group has data — exactly the same condition that makes the central `${prefix}quantile` column null. The "filter out some elements" branch would never fire, so it's not worth the complexity (and it would actively break the read side — see below).

## The whole-array-NULL approach is already safe downstream

The reader degrades gracefully when the grid is `NULL`:

```22:43:packages/back-end/src/integrations/sql/columns/quantile-bounds-from-query-response.ts
    const gridArray = row[`${prefix}quantile_grid`];
    const isArrayGrid = Array.isArray(gridArray);
    const expectedGridLength = N_STAR_VALUES.length * 2;
    if (isArrayGrid && gridArray.length !== expectedGridLength) {
      throw new Error(...);
    }

    const getLower = isArrayGrid
      ? (k: number) => gridArray[2 * k]
      : (k: number) => row[`${prefix}quantile_lower_${N_STAR_VALUES[k]}`];
    ...
      if (lowerVal === undefined || upperVal === undefined) return;
```

If `gridArray` is `NULL`, `isArrayGrid` is `false`, so it falls back to the scalar `quantile_lower_*` columns — which don't exist for the array dialect, yielding `undefined`, which is skipped. Net result: no bounds populated, identical to the no-data case for scalar dialects. 

Crucially, this also confirms you must **not** return a shortened array: the `expectedGridLength` check throws on any length other than `N_STAR_VALUES.length * 2`. So it has to be "full array or `NULL`," never a filtered subset.

## Recommended fix

Guard the array so it collapses to `NULL` when empty. Since it's all-or-nothing, testing the first element is sufficient and avoids duplicating the whole array (each duplicated element is another `APPROX_QUANTILES` call):

```ts
// in getQuantileGridArrayColumn, instead of dialect.arrayLiteral(elements):
const grid =
  dialect.hasArrayQuantileGrid() && elements.length > 0
    ? dialect.arrayLiteralOrNullIfEmpty(elements)
    : dialect.arrayLiteral(elements);
return `, ${grid} AS ${prefix}quantile_grid`;
```

with BigQuery implementing:

```ts
arrayLiteralOrNullIfEmpty: (elements: string[]) =>
  `IF(${elements[0]} IS NULL, NULL, [${elements.join(", ")}])`,
```

### Where to put it — two options

1. **New dialect method (preferred).** Keep `arrayLiteral` meaning "a plain array literal" so the existing contract/test stays intact:

```1493:1493:packages/back-end/test/sqlintegration.test.ts
      expect(bigQueryDialect.arrayLiteral(["1", "2"])).toBe("[1, 2]");
```

   Adding a separate `arrayLiteralOrNullIfEmpty` (or `nullableArrayLiteral`) keeps that test valid and makes the intent explicit at the call site in `getQuantileGridArrayColumn` (the only place the grid is built, used by both `getQuantileGridColumns` and `getQuantileSketchGridColumns`).

2. **Change `arrayLiteral` semantics in-place.** Simpler diff but semantically muddier (a generic "array literal" helper that secretly null-guards), and it forces an update to the test above. I'd avoid this.

### One implementation note

Don't try to reference the `${prefix}quantile` column alias in the `IF` — BigQuery doesn't allow referencing a `SELECT`-list alias from another expression in the same `SELECT`. Referencing the first element expression inline (as shown) sidesteps that; BigQuery will typically common-subexpression-eliminate the repeated `APPROX_QUANTILES`/`EXTRACT_POINT` call, so the cost is negligible.

So: go with "if empty → `NULL` for the whole grid," implemented as a dedicated nullable-array dialect method called from `getQuantileGridArrayColumn`. Skip partial filtering entirely — it can't occur and would violate the length invariant on the read side.

lukesonnet: It's not perfect, because it relies on the "any element is null means they all are null by construction" but I think if we name the methods as about quantiles and add a small comment about this, we should be safe.

ahdriel: Nice find! Thanks for the detailed report.

I have implemented your suggestion with `quantileGridArrayLiteral` and removed `arrayLiteral` as it was not being used anymore. We can always re-add it easily later when needed.

lukesonnet (approved):

PR #5982, opened 2026-05-30
feat(dashboards): add dimension time series to exp dashboards

lukesonnet:
### Summary
- Adds an optional, precomputed-only **Dimension + Dimension Levels** selector to the experiment Time-Series block, rendering one time series per dimension level (mirrors the existing Dimension block, but scoped to precomputed dimensions).
- Makes dimension selection **config-driven** via a new `BLOCK_DIMENSION_CONFIG` map + `getBlockDimensionConfig` helper, and extracts a reusable `BlockDimensionFields` component so the editor has no per-block-type branching.
- Dimension selection is optional for the Time-Series block (empty = the dimensionless series, shown as main results), clearable via an explicit "None" option, and disables metric slices/expansion while a dimension is active.

### Changes
- `shared/validators/dashboard-block.ts`: optional `dimensionId` / `dimensionValues` on the time-series block (back-compatible under `.strict()`).
- `shared/dashboards/utils.ts`: `BlockDimensionConfig` + `BLOCK_DIMENSION_CONFIG` + `getBlockDimensionConfig`; default empty dimension fields in the create factory; fix `snapshotSatisfiesBlock` to treat precomputed **unit** dimensions (`precomputedUnitDimensionIds`) as satisfied on the dimensionless snapshot (also benefits the existing Dimension block).
- `DashboardEditorSidebar/BlockDimensionFields.tsx` (new): config-driven Dimension/Levels selectors (scope filter, "None" option, required/visibility gating).
- `DashboardEditorSidebar/EditSingleBlock.tsx`: replace inline selectors with `BlockDimensionFields`.
- `DashboardBlock/index.tsx`: only flag a missing dimension as "needs configuration" when `config.required`.
- `DashboardBlock/ExperimentTimeSeriesBlock.tsx`: render per-level graphs with aligned, human-readable dimension labels; share a single `renderGraph` helper across the dimensionless and per-level layouts.

### Test plan
- [X] Time-Series block with **no** dimension shows the normal dimensionless results (no "requires configuration" message).
- [X] Selecting a precomputed experiment dimension and a precomputed unit dimension each render one series per level with correct `name=value` labels aligned to the metric column.
- [X] Selecting specific levels filters the rendered series; clearing via "None" resets levels and returns to main results.
- [X] `pnpm --filter front-end type-check`, `pnpm --filter shared type-check`, and lint pass.

https://www.loom.com/share/685c26ffd28241bc8fa823d0d94a13da


packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentTimeSeriesBlock.tsx
```diff
@@ -141,9 +177,6 @@ export default function ExperimentTimeSeriesBlock({
       {Object.entries(rowGroups).map(([resultGroup, rows]) =>
         !rows.length ? null : (
           <div key={resultGroup} className="mb-4">
-            <h4 className="mb-3">
-              {`${resultGroup.charAt(0).toUpperCase() + resultGroup.slice(1)} Metrics`}
-            </h4>
             {rows.map((row) => {
               // Only render parent rows (not slice rows) for time series
```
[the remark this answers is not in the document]
lukesonnet: Intentional, thanks for checkin bud.

packages/shared/src/enterprise/dashboards/utils.ts
```diff
@@ -221,6 +271,8 @@ export const CREATE_BLOCK_TYPE: {
     snapshotId: experiment.analysisSummary?.snapshotId || "",
     variationIds: [],
     differenceType: "relative",
+    dimensionId: "",
+    dimensionValues: [],
```
ahdriel, line 267: question: should we add this to `migrate` as well if this should be the new default value?

I think some checks validate if this field exists and is empty; so I am thinking we might have different experiences for new blocks vs existing blocks without it. (which might be okay, but wanted to flag)

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/BlockDimensionFields.tsx
```diff
@@ -0,0 +1,99 @@
...
+        labelClassName="font-weight-bold"
+        placeholder={
+          config.required
+            ? "Choose which dimension to use"
+            : "No dimension breakdown"
+        }
+        value={block.dimensionId}
+        containerClassName="mb-0"
+        onChange={(value) => {
+          // Clearing the dimension also clears any selected levels.
```
ahdriel, line 71: issue (part of what greptile raised): If we pick Dimension A, then select a specific level on the Dimension Values, when we switch to Dimension B the dimensionValues is not cleared up because we are doing ...block in the setBlock.

Should we remove the !value check and just do if hasFieldOfType dimensionValues then set dimensionValues: []?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentTimeSeriesBlock.tsx
```diff
@@ -141,9 +177,6 @@ export default function ExperimentTimeSeriesBlock({
       {Object.entries(rowGroups).map(([resultGroup, rows]) =>
         !rows.length ? null : (
           <div key={resultGroup} className="mb-4">
-            <h4 className="mb-3">
-              {`${resultGroup.charAt(0).toUpperCase() + resultGroup.slice(1)} Metrics`}
```
ahdriel, line 145: intentional that we are not differentiating between metric groups anymore?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/BlockDimensionFields.tsx
lukesonnet: I changed it to check if the value changes, which I think is reasonable.

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardBlock/ExperimentTimeSeriesBlock.tsx
lukesonnet: Yes, it was really ugly.

packages/shared/src/enterprise/dashboards/utils.ts
lukesonnet: added migration to `migrateBlock`, thanks.

lukesonnet:
@cursor review

lukesonnet:
@cursor review

PR #5983, opened 2026-05-30
fix(time-series): keep time series data if at least one variation has real data

lukesonnet:
## Keep partial metric time-series points instead of dropping them

`isValidDataPoint` now drops a data point only when **all** treatment variations are degenerate (absolute CI of `[0, 0]`), rather than as soon as any single variation is degenerate. This stops early-phase points (where one variation lacks data but others have real estimates) from being silently discarded on write and render.

- `packages/shared/src/util/metric-time-series.ts`: invert the per-variation check to require all treatments degenerate before invalidating a point.
- `packages/shared/test/metric-time-series.test.ts`: add `isValidDataPoint` and `filterInvalidMetricTimeSeries` coverage for all-degenerate, mixed-validity, and missing-CI cases.

### DIFF
--- a/packages/shared/src/util/metric-time-series.ts
+++ b/packages/shared/src/util/metric-time-series.ts
@@ -24,17 +24,17 @@ export function isValidDataPoint(
     return false;
   }
 
-  // Check variations from index 1 onwards (skip control)
-  for (let i = 1; i < dataPoint.variations.length; i++) {
-    const variation = dataPoint.variations[i];
-    if (
-      variation.absolute?.ci &&
-      variation.absolute.ci[0] === 0 &&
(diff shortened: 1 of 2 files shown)


jdorn (approved):

ahdriel (approved):
I think the changes LGTM!

But we likely need to fix the front-end renderer of time series as well.

Given the assumption that we are not saving invalid points is now changing, and we are going to have certain variations with point [0, 0] for CI for example, we need to ensure we are not rendering those as 0, and instead we skip them altogether and render just the valid data point for the day.

You can look into the `mapTimeSeriesPointToVariationCells` function where we likely need to map and filter out the variations that have invalid CI.

PR #5986, opened 2026-06-01
Insight finder & saved learnings

Auz:
When experimentation programs scale into the hundreds and thousands, finding patterns across these experiments is important for leveraging these learnings for future projects. Typically, if this is done at all, it is exists in the teams brain only. This project addresses this in two ways. 

1. Adds the concept of saved learnings (for the lack of a better term), that can be saved to materialize these learnings. Each saved learning can also be linked to experiments that support and contradict these findings. 
2. Adds a way to let AI search for patterns across these experiments.


Auz:
@cursor review

Auz:
@cursor review

Auz:
@greptileai review again

packages/back-end/src/models/InsightModel.ts
```diff
@@ -0,0 +1,299 @@
...
+            })),
+          };
+        },
+      }),
+    ],
+  },
+});
+
+export class InsightModel extends BaseClass {
+  protected migrate(doc: unknown): InsightInterface {
```
lukesonnet: AI migration slop from an old model that never existed. Consider a full pass that checks the assumption that insights never existed before so any code that handles migrations can be removed.

packages/back-end/src/models/InsightModel.ts
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  protected canRead(doc: InsightInterface): boolean {
+    return this.context.permissions.canReadMultiProjectResource(doc.projects);
+  }
+
+  protected canCreate(doc: InsightInterface): boolean {
+    const projects = doc.projects && doc.projects.length ? doc.projects : [""];
+    return projects.some((project) =>
+      this.context.permissions.canCreateExperiment({ project }),
```
lukesonnet: Probably worthy of its own permission class rather than using create experiment.

packages/shared/src/validators/insights.ts
```diff
@@ -0,0 +1,192 @@
...
+    // subsequent editors in the order they first edited.
+    authors: z.array(z.string()),
+    title: z.string(),
+    text: z.string(),
+    tags: z.array(z.string()),
+    supportingExperimentIds: z.array(z.string()),
+    // Experiments whose outcomes seem to run counter to this insight. Can be
+    // suggested by AI at generation time, surfaced by later contrary scans, or
+    // curated manually by the team.
+    contraryEvidence: z.array(z.string()),
```
lukesonnet: nit: throughout, should we call these contradictingExperimentIds for symmetry and in the product should we call it "Contrary Experiments" again for symmetry?

packages/front-end/components/Insights/FindInsightsModal.tsx
```diff
@@ -0,0 +1,380 @@
...
+                      {s.suggestion.confidence && (
+                        <Badge
+                          label={`${s.suggestion.confidence} confidence`}
+                          color={CONFIDENCE_COLORS[s.suggestion.confidence]}
+                          variant="soft"
+                          size="sm"
+                        />
+                      )}
+                    </Flex>
+                    {s.saved ? (
```
lukesonnet: Instead of an un-revertable "save" button this should just be a checkbox to "save" so it can be turned on and off easily and match more of our other patterns, and makes it clearer the saving only happens when you submit the modal, which seems right.

packages/front-end/components/Insights/SavedInsightsList.tsx
```diff
@@ -0,0 +1,699 @@
...
+
+  if (insights.length === 0) {
+    return (
+      <>
+        <EmptyState
+          title="No saved insights yet"
+          description="Use the Experiment Results tab to find common patterns across your experiments, save what you want to keep, or write one from scratch."
+          rightButton={null}
+          leftButton={
+            <Button onClick={() => setShowNew(true)}>New saved learning</Button>
```
lukesonnet: This should be above in the header, like we have for all our other assets (e.g. create experiment on the experiments page), and consider having it visible on both tabs to drive adoption?

packages/front-end/components/Insights/SavedInsightsList.tsx
```diff
@@ -0,0 +1,699 @@
...
+
+  return (
+    <>
+      {error && (
+        <Box mb="3">
+          <Callout status="error">{error}</Callout>
+        </Box>
+      )}
+      <Flex justify="end" mb="3">
+        <Button onClick={() => setShowNew(true)}>New saved learning</Button>
```
lukesonnet: Same for the non-empty state here, just move it to the top.

packages/front-end/components/Insights/SavedInsightsList.tsx
```diff
@@ -0,0 +1,699 @@
...
+                precision="date"
+                containerClassName=""
+              />
+            </Flex>
+          </Flex>
+        </Flex>
+        <Flex gap="3" wrap="wrap" align="end">
+          {projectOptions.length > 0 && (
+            <Box style={{ minWidth: 200, flexGrow: 1, maxWidth: 360 }}>
+              <MultiSelectField
```
lukesonnet: Project should be inline just like it is on the experiment search/filter component.

packages/front-end/components/Insights/EditInsightModal.tsx
```diff
@@ -0,0 +1,210 @@
...
+      </Box>
+      <Box mb="4">
+        <SelectField
+          label="Status"
+          value={status}
+          options={statusOptions}
+          onChange={(v) => setStatus(v)}
+          sort={false}
+        />
+        {selectedStatusObject && (
```
lukesonnet: Why show this badge? not our usual pattern

packages/front-end/components/Insights/SavedInsightsList.tsx
```diff
@@ -0,0 +1,699 @@
...
+                          title={
+                            s
+                              ? undefined
+                              : "This status no longer exists in settings"
+                          }
+                        />
+                      );
+                    })()}
+                </Flex>
+                {allowManage && (
```
lukesonnet: I think we just use three button context menus with edit and delete as our established pattern here (and also for the individual `[lid].tsx` page.

packages/front-end/components/GeneralSettings/ExperimentSettings/index.tsx
```diff
@@ -449,6 +450,11 @@ export default function ExperimentSettings({
             <Box mb="4" width="100%">
               <DecisionFrameworkSettings />
             </Box>
+
+            {/* Saved Learning Statuses */}
+            <Box mb="4" width="100%">
+              <LearningStatusSettings />
```
lukesonnet: Can we move these to their own section below bandit settings

"Learning Settings" and then also probably remove the pencil and delete to make them use our three dot menus?

lukesonnet (commented):
I left a ton of nits, two more:

1. We should either rename the UI to be `Insights` or rename all the models and internal stuff to be Learnings to match the UI. Consider renaming "Experiment Results" to "Experiment Library" on the Learnings page.
2. The insightsfindcachemodel seems not that important and could be removed for simplicity.

---

Main feedback that I want to make sure we do or have a clear plan for keeping these insights fresh as new experiment lands.

Step 1 (I think in this PR): A new API endpoint and a "refresh insights" button that scans experiments that stopped since either the latest experiment in the insight OR some last refresh try, suggests updates, and lets you review and add them. Also give them an "update all" button or something like that.
Step 2: Automate this process with a cron job (but we can skip this for now and just let people use the above API endpoint).

Auz:
@greptileai review

Auz:
@greptileai review

Auz:
@greptileai review

Auz:
@greptileai review

Auz:
@greptileai review

packages/back-end/src/routers/learnings/learnings.controller.ts
```diff
@@ -0,0 +1,901 @@
...
+  );
+
+  // Stamp what we actually checked so the next run starts from here
+  await Promise.all(
+    refreshedIds.map(async (id) => {
+      const doc = learnings.find((l) => l.id === id);
+      if (!doc) return;
+      try {
+        await context.models.learnings.update(doc, {
+          lastRefreshedAt: new Date(),
```
lukesonnet, line 855: If you cancel, there's no way to get back to what you lost in the refresh because we advance the `lastRefreshedAt`. But i think that's ok for now.

packages/back-end/src/routers/learnings/learnings.controller.ts
```diff
@@ -0,0 +1,901 @@
...
+    ...learning.supportingExperimentIds,
+    ...learning.contradictingExperimentIds,
+  ].reduce((max, id) => {
+    const exp = experimentsById.get(id);
+    return exp ? Math.max(max, experimentRecency(exp)) : max;
+  }, 0);
+  return Math.max(citedRecency, learning.lastRefreshedAt?.getTime() || 0);
+}
+
+export const postRefreshLearnings = async (
```
lukesonnet, line 700: I was thinking you should just be able to refresh one at a time from preventing too large of a request. I think there may be concerns with the unbounded nature of this request.

packages/front-end/pages/learnings/index.tsx
```diff
@@ -91,13 +133,41 @@ const LearningsPage = (): React.ReactElement => {
...
+                    <Button
+                      variant="outline"
+                      disabled={!aiEnabled}
+                      onClick={() => setRefreshOpen(true)}
+                    >
+                      <PiArrowsClockwise /> Refresh Learnings
+                    </Button>
+                  </Tooltip>
+                )}
+                <Button onClick={() => setShowNewLearning(true)}>
```
lukesonnet: Not permission gated?

packages/back-end/src/controllers/discussions.ts
```diff
@@ -196,6 +197,61 @@ export async function getDiscussion(
   }
 }
 
+export async function getDiscussionCounts(
```
lukesonnet, line 205: The robot is concerned about this endpoint:

GET /discussions/counts/:parentType is unbounded and can crash. discussions.ts:211-213

* ids has no cap, and each id triggers its own getProjectsByParentId DB read inside a single Promise.all — a client can request 10k ids and get 10k concurrent queries.
* ((req.query?.ids as string) || "").split(",") sits outside the try. ?ids=a&ids=b makes req.query.ids an array, .split throws, and since the route is registered bare on app.get (no wrapController) it's an unhandled rejection rather than a 400.
* The route has no validateRequestMiddleware, so parentType is an unvalidated cast; unknown values fall through getProjectsByParentId's switch with no default and return undefined.

lukesonnet (approved):

Auz:
@greptileai review

Auz:
@greptileai review

Auz:
@greptileai review

Auz:
@greptileai review

PR #5987, opened 2026-06-01
Materialize assignment query into stable table for inactive pipeline

### LINKED ISSUE #5925: [Feature] Optimization of fetching experiment metadata during importing existing experiments
### Description of Feature

Separate execution of assignment queries in a data source when users are importing pre-existing queries.

gouthamx67:
This update optimizes data processing when the full pipeline is inactive but table writing is enabled. Instead of inlining assignment/exposure queries as CTEs, queries are now materialized into stable per-experiment tables (gb_units_assignment_*), significantly reducing redundant computation during refreshes.

Key changes:
* Refactored units table logic into shouldCreateUnitsTable and canWriteTable
* Added ASSIGNMENT_UNITS_TABLE_PREFIX constant for consistent table naming
* Extended drop-units-table validation to support both prefix types
* Precomputation of unit dimensions now occurs only in full pipeline mode

Closes: https://github.com/growthbook/growthbook/issues/5925

### DIFF
--- a/packages/back-end/src/integrations/sql/queries/drop-units-table-query.ts
+++ b/packages/back-end/src/integrations/sql/queries/drop-units-table-query.ts
@@ -1,10 +1,16 @@
 import { DropTableQueryParams } from "shared/types/integrations";
-import { UNITS_TABLE_PREFIX } from "back-end/src/queryRunners/ExperimentResultsQueryRunner";
+import {
+  UNITS_TABLE_PREFIX,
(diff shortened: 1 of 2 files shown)


lukesonnet:
I believe this is already covered under "Ephemeral" pipeline mode: https://docs.growthbook.io/app/data-pipeline.

It also is not the issue called out in #5925, which refers to importing existing expeirments.

PR #5992, opened 2026-06-01
Contextual bandits MVP

lukebrawleysmith:
### Summary

Introduces a **Contextual Bandit Query** (`cbq_` prefix) — a dedicated query
model that replaces the overloaded Experiment Assignment Query (EAQ) for
contextual bandits. CB queries live in their own `contextualbanditqueries`
collection with a purpose-built SQL contract, so the EAQ is no longer
burdened with CB-specific columns (`leaf_id`, `bandit_version`,
`variation_weights`) and is free to remain experiment-scoped.

Also ships correctness fixes to the SRM query (degrees-of-freedom) and
replaces the `snapshotUpdateCount` column name with `banditVersion` to
match the updated SQL contract.

### What changed

**New `ContextualBanditQueryModel`** (`packages/back-end/src/enterprise/models/ContextualBanditQueryModel.ts`)
- New Mongo collection `contextualbanditqueries`, ID prefix `cbq_`
- Enterprise-gated (`contextual-bandits` premium feature)
- Write permission delegated to the owning datasource's `editDatasourceSettings`
- `customValidation` enforces: ≥1 targeting attribute column, every column is a
  safe SQL identifier, and each column maps to a non-archived org attribute
- `handleApiList` accepts an optional `?datasourceId=` filter (used by the
  CB create-form query picker)

**New `contextualBanditQueryValidator` schema** (`packages/shared/src/validators/contextual-bandit-query.ts`)
- Source-of-truth Zod schema; infers `ContextualBanditQueryInterface`
- Stores `query`, `userIdType`, `targetingAttributeColumns`, `dimensions`,
  `hasNameCol` alongside `datasourceId` / `name` / `description`

**REST API** (`GET|POST|PATCH|DELETE /api/v1/contextual-bandit-queries`)
- Wired through `contextualBanditQueryApiSpec` in `api.router.ts`
- Standard CRUD via `includeDefaultCrud: true`; list supports datasource scoping

**`AddEditContextualBanditQueryModal`** (`packages/front-end/components/ContextualBandit/`)
- New modal for authoring CB queries (the CB-specific counterpart to the EAQ modal)
- SQL editor + targeting attribute column picker with inline validation warnings
- Targeting columns are always required (no "is contextual" toggle — the whole
  collection is CB queries)

**`useContextualBanditQueries` hook** (`packages/front-end/hooks/useContextualBanditQueries.ts`)
- SWR-backed hook; returns list + map; only fetches when `datasourceId` is set

**`ContextualBanditResultsQueryRunner`** query runner now reads from the CB query doc
- Fetches the `cbq_` query from `context.models.contextualBanditQueries`
- `getUnitsQueryOverride()` returns `{ query, userIdType }` from the CB query doc
- Two sub-queries: metric rows (fact-metric query) + optional SRM query

**SRM query correctness fixes** (`packages/back-end/src/integrations/sql/queries/contextual-bandit-srm-query.ts`)
- Degrees of freedom computed in SQL as `(usable cells) − (kept groups)` rather
  than `(variations − 1)`, giving a correct χ² test when cells are dropped
- `banditVersion` column replaces `snapshotUpdateCount` everywhere in the SQL

**Deleted** `packages/back-end/src/enterprise/services/contextualBanditQueries.ts`
- Logic migrated into `ContextualBanditQueryModel`; no behaviour change

### Test plan
- [ ] Create a CB query via the new modal; verify it appears in the list for
  that datasource and is rejected when `targetingAttributeColumns` is empty
- [ ] Edit / delete a CB query; confirm datasource-settings permission gate fires
  for a non-admin user
- [ ] Create/run a contextual bandit that references a `cbq_` query; confirm
  the fact-metric SQL and SRM query both execute
- [ ] Verify SRM p-value is sensible on a known synthetic dataset
- [x] Run back-end tests: `pnpm --filter back-end test`
- [x] Run shared tests: `pnpm --filter shared test`


packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
```diff
@@ -0,0 +1,195 @@
...
+  settings: ExperimentSnapshotSettings,
+): ContextualBanditUnitsSqlConfig | null {
+  const bs = settings.banditSettings;
+  if (!bs?.banditIsContextual || !bs.targetingAttributeColumns?.length) {
+    return null;
+  }
+  const aliases = getSafeTargetingSqlIdentifiers(bs.targetingAttributeColumns);
+  if (!aliases.length) {
+    return null;
+  }
```
lukesonnet, line 68: If you hit this use case and there are no attributes, does the contextual bandit just silently fall back to a kind of simple MAB? Seems like it should, but we should test this use case and consider logging somewhere (logger.warn at some altitude) that we found no attributes but we should have given this is a contextual bandit.

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
```diff
@@ -0,0 +1,195 @@
...
+  const k = Math.max(1, settings.variations?.length ?? 1);
+  const maxRankedContexts = Math.max(1, Math.floor(3000 / k) - 1);
```
lukesonnet: Let's use a hard-coded constant here `SQL_ROW_LIMIT` instead of 3k. Consider overriding it for BQ to go up to like 5k, since we solved the mongo limit with result chunking.

So maybe it needs to be a `dialect` property.

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
```diff
@@ -0,0 +1,195 @@
...
+    const col = contextualBanditAttrCol(a);
+    dimensionCols.push({
+      alias: col,
+      value: col,
+    });
+  }
+}
+
+export function getSafeTargetingSqlIdentifiers(columns: string[]): string[] {
+  return columns.filter((c) => SAFE_SQL_IDENT.test(c));
```
lukesonnet: Is this the right place to do this? What happens if this fails? Will a CB silently not be optimizing on some attribute?

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
lukesonnet: I know some of the logic later on in the query relies on `aliases` having some length, so maybe that's fine, but does the fact that it does none of the extra bandit query also work with the final processing? It should in theory, we don't need any contexts or anything, but we should test it.

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
```diff
@@ -0,0 +1,195 @@
...
+        ${baseColumnRefs}
+        ${aliases
+          .map(
+            (a) =>
+              `, m.${contextualBanditAttrCol(a)} AS ${contextualBanditAttrCol(a)}`,
+          )
+          .join("")}
+      FROM
+        __experimentUnitsBase b
+      INNER JOIN __attr_cb_userContextMapped m ON (
```
lukesonnet: This join is unit level and seems unnecessary, why can't `__attr_cb_userContextMapped` just carry the user baseColumnRefs throughout and then all you need is that LEFT JOIN and don't need to re-inner join?

packages/shared/types/experiment-snapshot.d.ts
```diff
@@ -183,6 +184,20 @@ export interface SnapshotBanditSettings {
   }[];
   useFirstExposure?: boolean;
   windowSettings?: MetricWindowSettings;
+  /**
```
lukesonnet: Let's just use a separate contextual bandit snapshot settings.

lukesonnet (commented):
Some comments inline, plus:

Lets block on the decision metric being a fact metric (need to add validation to creation forms as well). This will simplify some SQL and make it easier to do incremental refresh contextual bandit in the future.

packages/shared/types/experiment-snapshot.d.ts
lukebrawleysmith: Teresa is working on this, it should be in her changes.

teresayung: ah yes, im in the process of removing this

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
lukebrawleysmith: made it a `dialect` property

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
lukebrawleysmith: The FE now checks for proper formatting when setting attributes.  Also, it now no longer silently fails if the formatting is wrong; an error is thrown here.

lukebrawleysmith:
> Some comments inline, plus:
> 
> Lets block on the decision metric being a fact metric (need to add validation to creation forms as well). This will simplify some SQL and make it easier to do incremental refresh contextual bandit in the future.

Confirmed that now the decision metric must be Fact Metric

packages/back-end/src/integrations/sql/ctes/contextual-bandit-experiment-units-cte.ts
lukebrawleysmith: Good callout, removed the unneeded INNER JOIN

packages/shared/src/validators/exposure-query-targeting-attribute-columns.ts
```diff
@@ -0,0 +1,99 @@
...
+  "Settings → Attributes";
+
+/** Closing punctuation after the settings phrase (parenthesis + sentence period). */
+export const TARGETING_ATTRIBUTE_COLUMN_HELP_AFTER_SETTINGS_LINK = ").";
+
+/** Shared suffix for UI and API validation errors (plain text). */
+export const TARGETING_ATTRIBUTE_COLUMN_HELP_SENTENCE =
+  TARGETING_ATTRIBUTE_COLUMN_HELP_BEFORE_SETTINGS_LINK +
+  TARGETING_ATTRIBUTE_COLUMN_SETTINGS_LINK_LABEL +
+  TARGETING_ATTRIBUTE_COLUMN_HELP_AFTER_SETTINGS_LINK;
```
jdorn: Don't use constants, just put this in the inline error messages

packages/shared/src/validators/features.ts
```diff
@@ -155,6 +172,12 @@ const experimentRule = baseRule
     banditConversionWindowUnit: z.enum(["hours", "days"]).optional().nullable(),
     templateId: z.string().optional(),
     customFields: z.record(z.string(), z.any()).optional(),
+    /** Per-context arm-weight entries for contextual-bandit experiments. */
+    contexts: z.array(contextsEntryValidator).optional(),
+    /** True when this rule drives a contextual-bandit experiment. */
+    isContextualBandit: z.boolean().optional(),
+    /** Attribute names that must be present for the rule to evaluate. */
+    attributesRequired: z.array(z.string()).optional(),
```
jdorn: Delete this. Old legacy rule type

packages/shared/src/validators/features.ts
```diff
@@ -115,9 +115,26 @@ const experimentValue = z
...
+  "multi-armed-bandit",
+  "contextual-bandit",
+] as const;
 const banditStageType = ["explore", "exploit", "paused"] as const;
 
+/**
+ * One context entry for a contextual-bandit experiment rule.
+ * Each entry maps a contextId (derived from attribute values) to per-arm weights.
+ */
+export const contextsEntryValidator = z
```
teresayung: can remove - already have reference rule

packages/shared/src/validators/reports.ts
```diff
@@ -127,7 +127,12 @@ export const apiReportValidator = namedSchema(
     experimentMetadata: z
       .object({
         type: z
-          .enum(["standard", "multi-armed-bandit", "holdout"])
+          .enum([
+            "standard",
+            "multi-armed-bandit",
+            "contextual-bandit",
```
jdorn: Revert this. Don't need contextual support in custom reports

packages/shared/src/validators/contextual-bandit.ts
```diff
@@ -0,0 +1,750 @@
...
+  statusUpdateScheduleValidator,
+  variation,
+} from "./experiments";
+import { priorSettingsValidator } from "./fact-table";
+import { namedSchema } from "./openapi-helpers";
+import { ownerEmailField, ownerField, ownerInputField } from "./owner-field";
+
+/** Per-leaf arm weights stored on a CB phase. */
+export const leafWeightValidator = z.object({
+  contextId: z.string(),
```
jdorn: Add condition here instead of needing to join to events collection

packages/shared/src/validators/contextual-bandit.ts
```diff
@@ -0,0 +1,750 @@
...
+  /**
+   * Current variation weights for this period — analogous to
+   * `experimentPhase.variationWeights`. Used by the SDK rule emitter
+   * to set arm allocation when a user has no context match.
+   */
+  variationWeights: z.array(z.number()).optional(),
+  /** Per-leaf arm weights for this CB period. */
+  currentLeafWeights: z.array(leafWeightValidator),
+});
+export type CbPhase = z.infer<typeof cbPhaseValidator>;
```
jdorn: Get rid of phases array.  Move these fields to the top-level of the bandit object

packages/shared/src/validators/contextual-bandit.ts
```diff
@@ -0,0 +1,750 @@
...
+  /** JSON-string MongoDB-style targeting condition, mirrors experimentPhase. */
+  condition: z.string().optional(),
+  /** Seed for the hashing function. Optional; defaults to the CB's trackingKey. */
+  seed: z.string().optional(),
+  /**
+   * Current variation weights for this period — analogous to
+   * `experimentPhase.variationWeights`. Used by the SDK rule emitter
+   * to set arm allocation when a user has no context match.
+   */
+  variationWeights: z.array(z.number()).optional(),
```
jdorn: Rename to `fallbackLeaf` or something like that

packages/back-end/src/services/features.ts
```diff
@@ -172,6 +173,8 @@ export function generateFeaturesPayload({
   savedGroupsMap?: Record<string, SavedGroupInterface>;
   includeRuleIds?: boolean;
   includeExperimentNames?: boolean;
+  /** Optional map of experimentId → CB doc for contextual-bandit payload injection. */
+  cbMap?: Map<string, import("shared/validators").ContextualBanditInterface>;
```
jdorn: No inline import

packages/back-end/src/util/features.ts
```diff
@@ -819,6 +830,120 @@ export function getFeatureDefinition({
...
+
+          // Running CB: emit the SDK experiment-rule.
+          rule.variations = cb.variations.map((v) => {
+            const variation = r.variations?.find(
+              (rv) => rv.variationId === v.id,
+            );
+            return variation
+              ? getJSONValue(feature.valueType, variation.value)
+              : null;
+          });
```
jdorn, line 947: Having `variations` will make old SDK versions think this is a normal experiment. We instead want them to skip this rule

packages/back-end/src/util/features.ts
```diff
@@ -819,6 +830,120 @@ export function getFeatureDefinition({
...
+        // sourced from the CB doc directly, with no parent experiment. SDK
+        // wire format is identical to a CB-driven experiment-rule — the
+        // payload builder produces an `experiment-rule` shape decorated
+        // with `isContextualBandit`, `attributesRequired`, and (eventually)
+        // `contexts`. SDKs that lack the `contextualBandits` capability get
+        // a byte-equivalent payload without the CB-specific fields and fall
+        // back to MAB on `phase.variationWeights`, matching the legacy
+        // experiment-ref + CB-typed-experiment path.
+        if (r.type === "contextual-bandit-ref") {
+          const cb = cbMap?.get(r.contextualBanditId);
```
jdorn, line 916: For now, just return `null` from this branch and do the SDK payload logic in a separate PR

packages/back-end/src/enterprise/models/ContextualBanditModel.ts
```diff
@@ -0,0 +1,707 @@
...
+   * validator post-Commit-3, so this reads the raw Mongo collection to
+   * keep the cascade-delete in `onExperimentDelete` and the legacy
+   * `/experiments/:id/contextual-bandit/*` routes working through
+   * Commit 6. After Commit 6 the routes go away; the cascade path
+   * keeps using this helper.
+   */
+  public async getByExperimentId(
+    experiment: string,
+  ): Promise<ContextualBanditInterface | null> {
+    const raw = await this._dangerousGetCollection().findOne({
```
jdorn: Don't use `dangerous` methods

jdorn: Actually, delete this whole method

packages/back-end/src/enterprise/models/ContextualBanditModel.ts
```diff
@@ -0,0 +1,707 @@
...
+      >[1],
+    );
+  }
+
+  public async deleteForExperiment(experiment: string): Promise<void> {
+    await this._dangerousGetCollection().deleteMany({
+      organization: this.context.org.id,
+      experiment,
+    });
+  }
```
jdorn, line 470: Don't need this method

packages/front-end/components/ContextualBandit/StartContextualBanditModal.tsx
```diff
@@ -0,0 +1,419 @@
...
+            </>
+          ),
+        });
+      }
+    });
+
+  return { hardBlockerItems, softBlockerItems };
+}
+
+function SubmitButton({ cta, disabled }: { cta: string; disabled: boolean }) {
```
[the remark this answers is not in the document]
teresayung: these two modals will eventually diverge, so keeping it as two distinct modals

lukesonnet:
@greptileai

packages/front-end/components/ContextualBandit/AddEditContextualBanditQueryModal.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const accessibleAttributes = useAttributeSchema(false);
+
+  const userIdTypeOptions = dataSource?.settings?.userIdTypes?.map(
+    ({ userIdType }) => ({ display: userIdType, value: userIdType }),
+  );
+  const defaultUserId = userIdTypeOptions
+    ? userIdTypeOptions[0]?.value
+    : "user_id";
+
+  const defaultQuery = `SELECT\n  ${defaultUserId} as ${defaultUserId},\n  timestamp as timestamp,\n  experiment_id as experiment_id,\n  variation_id as variation_id,\n  leaf_id as leaf_id,\n  snapshot_update_count as snapshot_update_count,\n  variation_weights as variation_weights\nFROM my_contextual_bandit_assignments`;
```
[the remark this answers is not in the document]
lukesonnet: done

lukesonnet:
@cursor review

packages/back-end/src/enterprise/queryRunners/ContextualBanditResultsQueryRunner.ts
```diff
@@ -0,0 +1,378 @@
...
+
+    const srm = this.extractSrmResult(queryMap);
+
+    const cb = await this.loadCbDoc();
+
+    const statsSettings = getContextualBanditSettingsForStatsEngine(
+      cb,
+      this.snapshotSettings.variations.map((v) => v.id),
+    );
+
```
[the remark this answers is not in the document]
lukesonnet: We should always honor the `cb.contextualAttributes` array at the query and analysis stages. We should also validate that these attributes live within the `cbQuery` at update time and error if there is a mismatch, and that is the only way the `cbQuery` should influence the attribute set for a given contextual bandit update.

lukesonnet:
@cursor review

lukesonnet (approved):

PR #5995, opened 2026-06-02
Address security vulnerabilities (2026-06-02)

Kevin-Chant:
Addresses the current Vanta/ECR `growthbook` image findings (20 total). Triaged top-to-bottom; easy in-major wins shipped here, cross-major / runtime-risky items left for follow-up with reasons below.

## Fixed

| Package | Change | CVE | Notes |
| ------- | ------ | --- | ----- |
| `mongoose` | `^6.13.8` → `^6.13.9` (back-end) | CVE-2026-42334 | Sanitization-bypass fix was **backported to the 6.x line** (6.13.9), so no cross-major bump. `mongodb` stays `4.17.2`, matching the direct `^4.17.2` spec — no lagging duplicate. |
| `ip-address` | override `socks@2` → `^2.8.9` (root) | CVE-2026-42338 | `ip-address@9.0.5` only entered the tree via `socks@2.8.4`. `socks@2.8.9` declares `ip-address@^10.1.1`, so bumping the parent pulls the fixed `ip-address` (resolves to **10.2.0**) instead of forcing a cross-major override directly onto the old `socks`. `ip-address@10` ships CJS — no ESM-only cliff. |
| `@tootallnate/once` | override `@tootallnate/once@2` → `^2.0.1` (root) | CVE-2026-3449 | **In-major backport.** AWS Inspector reported `fixed: 3.0.1` (flagging it cross-major), but the advisory (GHSA-vpq2-c234-7xj6) lists **two** patched branches — `2.0.1` and `3.0.1`. `2.0.1` stays CommonJS, so no ESM-only cliff and no need to touch the `http-proxy-agent@5` consumer. Resolves the prod `2.0.0` copy to `2.0.1`. |
| `dulwich` | uninstall in Dockerfile poetry-cleanup | CVE-2026-42305 / GHSA-897w-fcg9-f6xj | `dulwich` is a **build-time-only dependency of `poetry==1.8.5`** (poetry pins `dulwich<0.22`); gbstats does not use it at runtime. It was being left behind in `/opt/venv` after poetry was uninstalled. Added `dulwich` to the existing `pip uninstall -y` line so the vulnerable copy never ships — clears the finding rather than leaving stale build cruft. (The advisory is a Windows-only clone-time RCE; this image is Linux, so it was unreachable anyway — but removing it is the correct fix.) |

## Auto-clears on next image rebuild + rescan (no code change)

These are OS / base-image packages where upstream `node:24-slim` (Debian 12) already ships the patched version; the next image build + Vanta rescan clears them. Verified upstream caught up via `bundled-status.py`.

- `gnutls28` 3.7.9 → 3.7.9-2+deb12u7 — **10 findings** (CVE-2026-33845, CVE-2026-42010 [both CRITICAL]; CVE-2026-33846, CVE-2026-42009, CVE-2026-42011, CVE-2026-3833, CVE-2026-42013, CVE-2026-5260, CVE-2026-42012 [HIGH]; CVE-2026-42015 [MEDIUM])
- `libgcrypt20` 1.10.1 → 1.10.1-3+deb12u1 — CVE-2026-41989 (MEDIUM)

## Skipped (with reasons)

- **`thrift` 0.16.0 → 0.23.0 — CVE-2026-41636 (HIGH, uncontrolled-recursion DoS), due 2026-06-07.** No backport exists below 0.23.0 (all versions `<0.23.0` are vulnerable). The only consumer is `@databricks/sql`, which **even at its latest (1.15.0) still pins `thrift@^0.16.0`** — so forcing 0.23.0 via override would fight the parent's resolution and risks breaking the Databricks SQL data-source transport at runtime (type-check can't catch this; thrift is loosely-typed JS). Needs a human-reviewed `@databricks/sql` compatibility check or an upstream bump. ⚠️ **Due 2026-06-07 — flagged for escalation.**
- **`fast-xml-parser` 4.5.5 → 5.7.0 — CVE-2026-41650 (MEDIUM), due 2026-07-26.** Cross-major; advisory states all versions `<5.7.0` are affected (no 4.x backport — `fast-xml-parser@4` is already pinned to the latest 4.x in `overrides`). The vulnerable `4.5.5` copies come from `snowflake-sdk` + `@google-cloud/storage`, which declare `^4.x`; a patched `5.8.0` already coexists in the tree via `@aws-sdk`, but forcing v5 onto the v4 consumers risks parser API breakage. Cross-major bump needs human review.
- **OpenTelemetry cohort (`@opentelemetry/auto-instrumentations-node`, `@opentelemetry/sdk-node`, `@opentelemetry/exporter-prometheus`) — CVE-2026-44902 (HIGH ×3), due 2026-06-27.** All three share one CVE: a malformed HTTP request crashes the **Prometheus exporter** (`:9464`). This back-end uses the **OTLP proto/http exporters, not the Prometheus exporter** — `exporter-prometheus` is only present transitively (via `sdk-node`) and is never instantiated, so the vulnerable code path isn't reachable in production. Clearing the on-disk version requires a **full cohort bump to the 0.217/0.75 line, which crosses `@opentelemetry/core` 1.x → 2.x** and needs ≥3 code patches in `tracing.opentelemetry.ts` (`core.getEnv()` removed; `new Resource()` → `resourceFromAttributes()`; `SemanticResourceAttributes` → `ATTR_*`) plus a smoke-test boot. That's a deliberate cross-major migration deserving its own PR with runtime validation, not a same-day security bump. Tracked as follow-up.

## Post-deploy projection

| Severity | Total findings | Addressed or auto-clears | Remaining |
| -------- | -------------- | ------------------------ | --------- |
| CRITICAL | 2              | 2                        | 0         |
| HIGH     | 13             | 9                        | 4         |
| MEDIUM   | 4              | 3                        | 1         |
| LOW      | 1              | 1                        | 0         |

Remaining: 4 HIGH (`thrift` + 3× OTel cohort), 1 MEDIUM (`fast-xml-parser`).


Dockerfile
```diff
@@ -21,8 +21,11 @@ RUN \
   && poetry export -f requirements.txt --output requirements.txt \
   && pip install --no-cache-dir -r requirements.txt \
   && pip install --no-cache-dir dist/*.whl ddtrace==4.3.2 "cryptography>=46.0.6,<47" \
-  && pip uninstall -y poetry poetry-core poetry-plugin-export keyring jaraco.classes setuptools wheel
+  && pip uninstall -y poetry poetry-core poetry-plugin-export keyring jaraco.classes setuptools wheel dulwich
```
ahdriel: question: I am not familiar with python setup, but I am curious if there's a difference between `dev dependencies` and `runtime dependencies` like we have in `package.json` that we can leverage here to uninstall all `devDependencies` and not have to keep this list up to date.

Would be less maintenance burden and also a smaller image (in theory).

no action required, I was just wondering.

Dockerfile
```diff
@@ -21,8 +21,11 @@ RUN \
...
 # cryptography version is specified above to override transitive dependency and fix vulnerability
+# dulwich is a build-time-only poetry dependency (poetry pins dulwich<0.22); it is not used by
+# gbstats at runtime, so we uninstall it alongside poetry to keep it out of the production venv
+# (clears CVE-2026-42305 / GHSA-897w-fcg9-f6xj rather than leaving stale build cruft in the image)
```
ahdriel: nit: should we remove this update and make a comment above the `uninstall` line on how we are removing non runtime dependencies?

ahdriel (approved):

Dockerfile
Kevin-Chant: https://github.com/growthbook/growthbook/blob/9fac722355067c5f64e41dbcc8c6a2c068864255/packages/stats/pyproject.toml#L25-L34 we do have that split in poetry as well, but these are poetry's own dev dependencies and poetry doesn't have an easy command to uninstall itself. We should probably add an extra stage to the dockerfile so poetry isn't installed in the final image and we just copy over the artifacts, but that's definitely a separate task

PR #6000, opened 2026-06-02
Fixes bug with initial plan selector not loading valid plan data

mknowlton89:
### Features and Changes

Fixes a bug where existing Cloud organizations could be incorrectly shown the `SelectInitialPlan` signup flow (and in the worst case have their effective plan downgraded) when navigating to `app.growthbook.io/?plan=pro` from the marketing site.

**Root cause:** The `showSelectPlanFlow` guard in `ProtectedPage` was calling `useUser()` outside of the `UserContextProvider` it renders in its own return value. Since there is no outer `UserContextProvider` in the main app render path, `effectiveAccountPlan` always resolved to `undefined` from the default context — making `hasExistingPaidPlan` permanently `false` and the guard ineffective.

**Additional issue:** For orgs using the legacy `enterprise: true` flag (rather than a license key), the plan selection flow could write a new `licenseKey` to the org. Once a `licenseKey` is present, `getEffectiveAccountPlan` uses the license plan rather than the legacy flag, silently downgrading those orgs from enterprise.

**Changes:**

- Fixes initial plan selection so users with multiple organizations can choose the correct organization before confirming a plan.
- Adds a skip option for the initial plan selection flow, allowing users to dismiss the prompt and continue into the app.
- Updates protected page gating so the plan selection flow only appears once an organization is selected and only for eligible cloud users without an existing paid plan.
- Restores enterprise plan detection for cloud organizations that still rely on the legacy `enterprise` flag before license migration is complete.
- Refreshes onboarding welcome panels with GrowthBook branding.

### Dependencies

None.

### Testing

- [ ] Verify a new cloud signup with an initial plan selection sees the plan selection screen.
- [x] Verify users with multiple organizations can switch organizations before selecting a plan.
- [x] Verify selecting Starter completes the flow and clears the initial plan selection.
- [x] Verify the skip link dismisses plan selection and routes the user into the app.
- [x] Verify existing paid plan organizations do not see the initial plan selection gate.
- [x] Verify cloud organizations with the legacy `enterprise` flag resolve to the enterprise plan.

### Screenshots
Before
Screenshot 2026-06-17 at 7 29 06 AM
After
Screenshot 2026-06-17 at 7 30 27 AM


mknowlton89:
> ### Confidence Score: 3/5
> The primary React context fix is correct and safe. The `licenseUtil.ts` change, however, incompletely addresses the 'already-affected enterprise orgs' scenario described in the PR, leaving those orgs in a silently downgraded state on the plan they were accidentally written into.
> 
> The `ProtectedPage` architectural change correctly prevents the plan-selection flow from firing for paid orgs going forward. But the `licenseUtil.ts` guard in `getAccountPlan` does not propagate through `getEffectiveAccountPlan` — the API endpoint computes `effectiveAccountPlan` using `getEffectiveAccountPlan`, which bypasses `getAccountPlan` entirely for any Cloud org that already holds a `licenseKey`. Enterprise orgs that had a pro key accidentally written during the bug window will continue to have their effective plan returned as 'pro' by the API, meaning they may still be missing enterprise features. The fix prevents the bug from recurring but does not restore the plan for orgs already affected.
> 
> `packages/back-end/src/enterprise/licenseUtil.ts` — `getEffectiveAccountPlan` needs the same `org.enterprise` guard that was added to `getAccountPlan`.
> 
> ### Comments Outside Diff (1)
> 1. `packages/back-end/src/enterprise/licenseUtil.ts`, line 959-981 ([link](https://github.com/growthbook/growthbook/blob/494ec9335cd1f80d22abc94d08293fc6faf5b351/packages/back-end/src/enterprise/licenseUtil.ts#L959-L981))
>    [P1](#) **`getEffectiveAccountPlan` not guarded — affected orgs remain downgraded**
>    The `if (org.enterprise) return "enterprise"` guard was added to `getAccountPlan`, but the API's `effectiveAccountPlan` field (line 942 in `organizations.controller.ts`) is computed by `getEffectiveAccountPlan`, which never calls `getAccountPlan` when a `licenseKey` is present on a Cloud org. For a Cloud org that has both `enterprise: true` and an accidentally-written pro `licenseKey`, `getEffectiveAccountPlan` falls straight into the license-validation path (line 966+), reads the pro license, and returns `"pro"` — the enterprise downgrade described in the PR description is still live for these orgs. Adding the same early-return guard at the top of `getEffectiveAccountPlan` (before the `IS_CLOUD` branch) would close this gap.
>    Prompt To Fix With AI
>    ```
>    This is a comment left during a code review.
>    Path: packages/back-end/src/enterprise/licenseUtil.ts
>    Line: 959-981
>    
>    Comment:
>    **`getEffectiveAccountPlan` not guarded — affected orgs remain downgraded**
>    
>    The `if (org.enterprise) return "enterprise"` guard was added to `getAccountPlan`, but the API's `effectiveAccountPlan` field (line 942 in `organizations.controller.ts`) is computed by `getEffectiveAccountPlan`, which never calls `getAccountPlan` when a `licenseKey` is present on a Cloud org. For a Cloud org that has both `enterprise: true` and an accidentally-written pro `licenseKey`, `getEffectiveAccountPlan` falls straight into the license-validation path (line 966+), reads the pro license, and returns `"pro"` — the enterprise downgrade described in the PR description is still live for these orgs. Adding the same early-return guard at the top of `getEffectiveAccountPlan` (before the `IS_CLOUD` branch) would close this gap.
>    
>    How can I resolve this? If you propose a fix, please make it concise.
>    ```
> 
> Prompt To Fix All With AI
> ```
> Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.
> 
> ---
> 
> ### Issue 1 of 2
> packages/back-end/src/enterprise/licenseUtil.ts:959-981
> **`getEffectiveAccountPlan` not guarded — affected orgs remain downgraded**
> 
> The `if (org.enterprise) return "enterprise"` guard was added to `getAccountPlan`, but the API's `effectiveAccountPlan` field (line 942 in `organizations.controller.ts`) is computed by `getEffectiveAccountPlan`, which never calls `getAccountPlan` when a `licenseKey` is present on a Cloud org. For a Cloud org that has both `enterprise: true` and an accidentally-written pro `licenseKey`, `getEffectiveAccountPlan` falls straight into the license-validation path (line 966+), reads the pro license, and returns `"pro"` — the enterprise downgrade described in the PR description is still live for these orgs. Adding the same early-return guard at the top of `getEffectiveAccountPlan` (before the `IS_CLOUD` branch) would close this gap.
> 
> ### Issue 2 of 2
> packages/back-end/src/enterprise/licenseUtil.ts:132-135
> The original `if (org.enterprise) return "enterprise"` inside the Cloud block at line 134 is now unreachable — the new early return on line 126 exits first for any enterprise-flagged org. This is now dead code.
> 
> ```suggestion
>     // Vercel starter orgs have the `restrictLoginMethod` set, but they're not pro_sso
>     if (org.isVercelIntegration) return "starter";
>     if (org.restrictAuthSubPrefix || org.restrictLoginMethod) return "pro_sso";
> ```
> ```
> 
> Reviews (1): Last reviewed commit: ["Merge branch 'main' into mk/downgrade-bu..."](https://github.com/growthbook/growthbook/commit/494ec9335cd1f80d22abc94d08293fc6faf5b351) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35136473)

I don't think this is an actual issue - technically Greptile is correct, but AFAIK no orgs are affected.

jdorn (approved):

PR #6002, opened 2026-06-02
Enable fact table optimization for managed warehouse

jdorn:
### Features and Changes

- Always enable fact table optimization for Managed Warehouse data sources, even for non-enterprise orgs
- Remove the setting to disable fact table optimization completely - it was originally built as a failsafe in case we broke something with the initial implementation. It is now fully battle tested and safe to enable for everyone.
- Fix the front-end `Optimized` badge to only show for queries that actually include 2+ metrics in them.  Currently, it also shows when there's only 1 metric included when the query isn't really "optimized".

### DIFF
--- a/packages/front-end/components/Queries/QueryStatsRow.tsx
+++ b/packages/front-end/components/Queries/QueryStatsRow.tsx
@@ -1,12 +1,14 @@
 import { QueryInterface, QueryStatistics } from "shared/types/query";
 import { ReactElement } from "react";
+import { isManagedWarehouse } from "shared/util";
 import PremiumTooltip from "@/components/Marketing/PremiumTooltip";
 import { GBInfo } from "@/components/Icons";
 import { useUser } from "@/services/UserContext";
+import { useDefinitions } from "@/services/DefinitionsContext";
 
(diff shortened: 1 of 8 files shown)


lukesonnet (approved):

packages/front-end/components/Queries/QueryStatsRow.tsx
```diff
@@ -40,12 +47,14 @@ export default function QueryStatsRow({
...
-  const factTableOptimizedMetrics = !hasOptimizedQueries
-    ? 0
-    : queries
-        .filter((q) => q.queryType === "experimentMultiMetric")
-        .map((q) => getNumberOfMetricsInQuery(q))
-        .reduce((sum, n) => sum + n, 0);
+  // A fact-metric query is only "optimized" when it actually combines more than
+  // one metric. Single-metric queries share the same query type/builder but
+  // weren't combined, so exclude them from the optimized count.
+  const factTableOptimizedMetrics = queries
```
lukesonnet, line 53: This can be kind of misleading, sometimes there's only one metric but it's still "optimal".

PR #6004, opened 2026-06-02
Updates SSO flow to respect autoApprove toggle

mknowlton89:
### Features and Changes

Previously, `addMemberFromSSOConnection` unconditionally added SSO users as full members, ignoring the organization's `autoApproveMembers` setting. This meant orgs that required manual member approval had no way to enforce that gate for SSO-authenticated users.

This change makes SSO respect the `autoApproveMembers` flag:

- If `autoApproveMembers === false` (explicitly disabled), the SSO user is added to the pending queue instead of as a full member, and the org owner receives a `sendPendingMemberEmail` notification with a link to the team settings page to review the request.
- Orgs that have never set the flag (`undefined`) continue to auto-approve, preserving the existing behavior.

### Dependencies

None

### Testing

- [ ] On an org with `autoApproveMembers` **explicitly set to `false`**: log in via SSO as a new user — confirm they are added as a **pending** member and that the org owner receives a pending member notification email
- [ ] On an org with `autoApproveMembers` **unset (`undefined`)**: log in via SSO as a new user — confirm they are added as a full member immediately (no regression)
- [ ] On an org with `autoApproveMembers` **set to `true`**: log in via SSO as a new user — confirm they are added as a full member immediately
- [ ] Approve a pending SSO member and confirm they gain normal access

### DIFF
--- a/packages/back-end/src/services/organizations.ts
+++ b/packages/back-end/src/services/organizations.ts
@@ -1189,6 +1189,18 @@ export async function addMemberFromSSOConnection(
   }
   if (!organization) return null;
 
+  // If the org doesn't have autoApproveMembers enabled, add the user as a pending member
+  if (!organization.autoApproveMembers) {
+    await addPendingMemberToOrg({
+      organization,
+      name: req.name || "",
+      email: req.email || "",
+      userId: req.userId,
+      ...getDefaultRole(organization),
+    });
+    return null;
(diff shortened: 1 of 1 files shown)


Kevin-Chant:
cursor review

Kevin-Chant (approved):

PR #6011, opened 2026-06-03
fix: keep create CTAs enabled when sample-data is the only project

gazzdingo:
The sample-data (demo) project is a real project that the front-end forces read-only by wrapping create permission checks to return false for it. On list pages with "All Projects" selected, the create CTA was gated by projects.some(p => canCreate(p)), so when the sample-data project was the only project in the org the button was disabled for everyone — admins included.

Check the global (no-project) create permission first so a user who can create at the org level isn't blocked by a non-creatable project. Fixes Add Feature, Add Holdout, and Add Fact Table; adds a regression test.

### Features and Changes

### DIFF
--- a/packages/front-end/pages/features/index.tsx
+++ b/packages/front-end/pages/features/index.tsx
@@ -613,19 +613,21 @@ export default function FeaturesPage() {
     if (project) {
       return permissionsUtil.canManageFeatureDrafts({ project });
     }
-    if (projects?.length) {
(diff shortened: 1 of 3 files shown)


Kevin-Chant (approved):
Repro'd the issue & confirmed this fixes features. Waiting on review bots to pass then I'll merge (or address their complaints myself @gazzdingo )

Kevin-Chant:
cursor review

PR #6017, opened 2026-06-03
session-replay SDK plugin 

lillialjackson:
### Features and Changes
Session Replay SDK, for now only available via script tag. Includes rrweb wrapped privacy settings, exponential back off error handling, and plugin evals work.
### Dependencies

Need to merge Central License Serve repo and Ingestor service repo ahead of this, will also need to update terraform with the necessary env var for prod. Clickpipe set up to be finished when events are being sent via kafka. 

### Testing
 
Manual, some unit


packages/sdk-js/src/plugins/session-replay.ts
```diff
@@ -0,0 +1,755 @@
...
+ */
+const IDLE_TIMEOUT_MS = 15 * 60 * 1000;
+const MAX_DURATION_MS = 30 * 60 * 1000;
+const FLUSH_INTERVAL_MS = 60_000;
+const FLUSH_BYTE_SIZE = 256 * 1024;
+const MAX_BUFFERED_EVENTS = 500;
+const COMPRESS_REQUESTS = true;
+
+export function sessionReplayPlugin({
+  trackingHost = "",
```
lillialjackson, line 174: validate what is the best way to set env var in the sdk build

packages/sdk-js/src/plugins/auto-attributes.ts
lillialjackson, line 1: Apparently we weren't storing the sessionID where it could be used between plugins so this is just to set a sessionID if one isn't manually set and to store it for the duration of the session

packages/sdk-js/src/plugins/retry-manager.ts
lillialjackson, line 1: tried to make the exponential backoff and retries somewhat reusable between plugins, I think there are some other places where we could potentially leverage this.

packages/sdk-js/src/auto-wrapper-plus.ts
lillialjackson, line 1: basically a copy of autowrapper witht he additions of session replay

packages/sdk-js/src/core.ts
```diff
@@ -141,6 +141,16 @@ function onFeatureUsage(
         logType: "feature",
       });
     }
+
```
lillialjackson, line 149: validate growthbookiness of this, wanted to create new subs for evals and saw a similar non-blocking error handling elsewhere, is there somewhere else we might want to send error logs to?

PR #6018, opened 2026-06-03
feat #5095: add optional documentationUrl to attributes

### LINKED ISSUE #5095: [Feature] Attributes Overhaul
Submitted by a large Enterprise customer.

1. When using attributes in the context of creating a feature rule, show additional information beyond just the attribute name, like the full description and/or tags. 
2. On the attributes page, add filtering/search/sort options to make it easier to navigate.
3. Add tags to attributes as a way of grouping them logically.
4. Ability to embed links in the attribute description and/or have a first class property for documentation links.

Monteir016:
## Description
Targeting attributes had no way to link to external documentation. Teams
managing large sets of attributes had no in-product mechanism to point
engineers or analysts to a source of truth (e.g. a data dictionary or
engineering spec) for each attribute.

## Implementation
Adds an optional `documentationUrl` field to attributes:

- Accepts only `http`/`https` URLs, validated with Zod in the shared
  package so both front-end and back-end enforce the same rule
- Empty string or `undefined` is treated as "no URL" and the field is
  omitted from storage, preserving full backward compatibility
- Shown as a clickable external link on the attribute detail page
- Editable via the existing attribute modal
- Exposed through the internal API and the public v1 API
  (`POST /api/v1/attributes` and `PUT /api/v1/attributes`)
- OpenAPI spec regenerated (`generated/spec.yaml`)

## Tests
- `pnpm --filter shared test` — unit tests for `documentationUrlSchema`
  covering valid URLs, invalid schemes, empty string normalization, and
  missing field
- `pnpm --filter back-end test test/api/attributes.test.ts` — API tests
  for create/update with a `documentationUrl`, invalid URL rejection, and
  field clearing
- All existing attribute tests continue to pass

## Dependencies
None.

## Testing
```bash
pnpm --filter shared test
pnpm --filter back-end test test/api/attributes.test.ts
```

### Screenshots

documentationUrl

## Closes
Closes #5095

Co-authored-by: Joao Gomes <joao.s.gomes@tecnico.ulisboa.pt>


joaomsgomes:
I've addressed all three issues. In [property].tsx I added rel="noreferrer" to the documentation link so it matches the attributes list page, and applied the same fix to the tooltip's Docs link in AttributeOptionTooltip.tsx. For the validation gap in AttributeModal.tsx, the form now reuses the shared documentationUrlSchema directly instead of the weaker startsWith check, so client-side validation matches the API exactly (including z.string().url()). Let me know if anything else needs changing!

PR #6019, opened 2026-06-03
General site-wide AI Assistant

oelshaikh:
# Site-wide GrowthBook AI assistant (general agent)

## Summary

Adds a general-purpose, site-wide AI assistant that can read and modify a user's GrowthBook data by calling the REST API in-process, scoped to the logged-in user's own permissions. It's launched from a sparkle button in the `TopNav` and opens as a slide-in chat panel available on every page.

### Backend

- **`/agent` router** (`routers/agent/`) with chat endpoints (`POST /chat`, list/get/cancel/feedback/delete), reusing shared `chat-controllers` and the existing `agent-handler` streaming infrastructure.
- **General agent** (`agent/general-agent.ts`): a system prompt + three tools — `callApi` (issue any REST request), `loadSkill` (load a skill's workflow on demand), and `askUser` (multiple-choice prompt that ends the turn).
- **In-process API dispatcher** (`agent/dispatcher.ts`): routes the agent's `callApi` requests through the same Express handlers customers hit, so validation/response/error shapes match the public REST API — with no separate API key.
- **Skill system** (`agent/skills/`): markdown skills loaded from disk as domain routers (`feature-flags`, `experiments`, `product-analytics`, `growthbook-docs`) plus on-demand leaves. Only routers appear in the system prompt; full workflows load via `loadSkill`.
- **Deterministic mutation-confirmation gate**: any non-GET call (minus a read-only allowlist for snapshots/explorations/column-values) is parked, surfaced to the UI as a confirm/cancel prompt, and replayed verbatim on confirm — the model never bypasses the gate.

### Supporting changes

- `ai-chat-to-model`: inject a `[Page context: <url>]` prefix into the user message; never compact `loadSkill` results.
- `agent-handler`: page-context plumbing, pending-action resolve/replay, and persist the final turn *before* closing the SSE stream to fix a stale-message race on reload.
- `conversation-buffer`/validators: persist `pendingAction` and `currentPage`.

### Frontend

- `Agent/` components: `AgentPanel`, `AgentLauncher`, `AgentChatHistory`, and `AgentPanelContext` (open/expanded state persisted per-tab in `sessionStorage`).
- `TopNav` sparkle trigger, gated on the `ai-suggestions` commercial feature + AI being enabled.
- Shared `ChatInputBar`/`AIChatPrimitives` refactors so the new panel and the existing product-analytics chat reuse the same primitives.

## Tests

Added unit tests for the dispatcher, general agent (body coercion, mutation gate), and skill loading.

## Screenshots

image


oelshaikh:
@greptileai

packages/back-end/src/agent/dispatcher.ts
```diff
@@ -0,0 +1,399 @@
...
+      // is mounted at /api in app.ts, so the public URL is /api/<version>/...
+      // We bake that prefix in here so internal dispatch uses the exact same
+      // path string a customer would.
+      const fullPath = `/api/${version}${r.path}`;
+      const paramNames: string[] = [];
+      const reSource = fullPath
+        .split("/")
+        .map((seg) => {
+          if (!seg) return "";
+          if (seg.startsWith(":")) {
```
jdorn: Double check - I think `r` might already have params schema

packages/back-end/src/agent/dispatcher.ts
```diff
@@ -0,0 +1,399 @@
...
+  if (noQuery.startsWith("/api/")) return noQuery;
+  if (/^\/v\d+\//.test(noQuery)) return `/api${noQuery}`;
+  if (noQuery.startsWith("/")) return `/api/v1${noQuery}`;
+  return noQuery;
+}
+
+class MalformedPathError extends Error {}
+
+function matchRoute(
+  method: string,
```
jdorn: Might want to use OpenAPI operationIds instead

packages/back-end/src/agent/dispatcher.ts
```diff
@@ -0,0 +1,399 @@
...
+};
+
+let compiled: CompiledRoute[] | null = null;
+let routesOverride: OpenApiRoute[] | null = null;
+
+function compileRoutes(): CompiledRoute[] {
+  if (compiled) return compiled;
+  const source = routesOverride ?? allRoutes;
+  compiled = source
+    .filter((r) => !!r.method)
```
jdorn: Filter out deprecated routes

packages/back-end/src/agent/dispatcher.ts
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * Authorization note: every field that gates access is derived from the
+ * caller's `ctx`, so a handler can only do what the caller's org role allows.
+ * `req.context.permissions` (the path all `/api/*` handlers use today) is the
+ * caller's role-scoped `Permissions` instance, and `checkPermissions` (the
+ * legacy `ApiRequestLocals` gate, unused by current `/api/*` handlers but part
+ * of the contract) delegates to the same role check. There is no privilege
+ * elevation here: dispatch is exactly as capable as the caller's role.
+ */
+function buildFakeReq(
```
jdorn, line 234: See if we can call the route handlers directly without using express

oelshaikh: Route handlers for our REST APIs are built using `createApiRequestHandler` which makes each handler expect an express request object as input, so it doesn't seem like we can get around creating the fake request object. 

I did restructure things so we're no longer going through express and instead call the rawHandler directly. This makes dispatchers job a bit easier (no faking response, no manual `next()` handling).

packages/back-end/src/agent/dispatcher.ts
oelshaikh: We still need to create regex for each path which requires positional knowledge of the params which the schema doesn't have.

packages/back-end/src/agent/dispatcher.ts
oelshaikh: I brainstormed this with opus a bit and it had a good point about LLMs being naturally good at constructing HTTP requests with method + path since that's part of their training data vs operationId which would be a unique approach that the LLM needs to be prompted into doing.

Also as long as we have to create fake request objects, we'll need to re-create the path anyway so we don't get away from the regex compilation.

packages/back-end/src/agent/dispatcher.ts
```diff
@@ -0,0 +1,389 @@
...
+/**
+ * Build a minimal Express-shaped request with the fields our API handlers read.
+ * Every access-gating field derives from the caller's `ctx`, so a handler can
+ * only do what the caller's org role allows — no privilege elevation.
+ */
+function buildFakeReq(
+  ctx: ReqContext,
+  input: DispatchInput,
+  params: Record<string, string>,
+): Record<string, unknown> {
```
jdorn: See if we can strongly type this as an Api Request

packages/back-end/src/util/secrets.ts
```diff
@@ -13,6 +13,13 @@ export const LOG_LEVEL = process.env.LOG_LEVEL;
 export const IS_CLOUD = stringToBoolean(process.env.IS_CLOUD);
 export const IS_MULTI_ORG = stringToBoolean(process.env.IS_MULTI_ORG);
 
+// Logs the assembled general agent system prompt at startup. Defaults to false;
+// set to "true" to enable (useful in development).
+export const LOG_AGENT_SYSTEM_PROMPT = stringToBoolean(
+  process.env.LOG_AGENT_SYSTEM_PROMPT,
+  false,
+);
```
jdorn: Remove

packages/back-end/src/server.ts
```diff
@@ -4,13 +4,19 @@ import "./instrumentation";
...
 const server = app.listen(app.get("port"), () => {
   logger.info(
     `Back-end is running at http://localhost:${app.get("port")} in ${app.get(
       "env",
     )} mode. Press CTRL-C to stop`,
   );
+  // Warm the skill cache at startup so the first agent request doesn't pay the
+  // synchronous-read cost (and block concurrent requests) on a cold cache.
+  getAllSkills();
+  logGeneralAgentSystemPrompt();
```
jdorn: Can probably remove this

packages/back-end/package.json
```diff
@@ -7,11 +7,12 @@
   },
   "scripts": {
     "build:emails": "mkdir -p dist/templates && cp -r src/templates/* dist/templates",
+    "build:skills": "mkdir -p dist/agent/skills && find src/agent/skills -name '*.md' -exec sh -c 'dest=\"dist/agent/skills/${1#src/agent/skills/}\"; mkdir -p \"$(dirname \"$dest\")\"; cp \"$1\" \"$dest\"' _ {} \\;",
```
jdorn: Change directory structure so we can just directly copy a single directory to dist

jdorn (approved):

packages/back-end/src/agent/dispatcher.ts
oelshaikh: The full `ApiRequest` type has all the express fields and we're not populated all of them so I had to create a `FakeApiRequest` type but still better than `Record<string, unknown>`

PR #6020, opened 2026-06-03
fix inverted baseline analyses for safe rollouts v2

bryce-fitzsimons:
Analysis was swapped for v2 safe rollouts (control versus variation arm).
Fix, update API docs, add tests.


gazzdingo (approved):
lgtm

PR #6021, opened 2026-06-03
Enable fact table optimization when skipPartialData is true

jdorn:
### Features and Changes

When `skipPartialData` is true for an experiment, we were skipping Fact Table Optimization entirely.  Every metric would be run in a separate query.

We did this because with this setting, the experiment end date in the SQL is dynamically changed based on the metric's conversion window, so it wasn't safe to combine arbitrary metrics together which may have different conversion window settings.

In reality, there is a very small set of distinct conversion windows an organization typically uses, so there is still a big opportunity for performance improvements.

This PR enables Fact Table Optimization even when this setting is true and simply adds the conversion window to the group key.  So two metrics with the same conversion window will be grouped together.  This is fully safe and avoids needing to do any SQL refactor work.

### DIFF
--- a/packages/back-end/src/services/experimentQueries/experimentQueries.ts
+++ b/packages/back-end/src/services/experimentQueries/experimentQueries.ts
@@ -17,6 +17,7 @@ import { isManagedWarehouse } from "shared/util";
 import { SourceIntegrationInterface } from "back-end/src/types/Integration";
 import { orgHasPremiumFeature } from "back-end/src/enterprise";
 import { applyMetricOverrides } from "back-end/src/util/integration";
+import { getMaxHoursToConvert } from "back-end/src/integrations/sql/dates/max-hours-to-convert";
 import {
   BANDIT_CUPED_FLOAT_COLS,
   BASE_METRIC_CUPED_FLOAT_COLS,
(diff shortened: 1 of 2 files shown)


packages/back-end/src/services/experimentQueries/experimentQueries.ts
```diff
@@ -228,7 +229,17 @@ export function chunkMetrics({
...
-export function getFactMetricGroup(metric: FactMetricInterface) {
+export function getFactMetricGroup(
+  metric: FactMetricInterface,
+  { skipPartialData }: { skipPartialData: boolean },
+) {
+  // When `skipPartialData` is enabled, the experiment end date is pulled back to
+  // exclude users who haven't had a full conversion window to convert.
+  // Add the conversion window to the group key to keep same-window metrics grouped together
+  const conversionWindowKey = skipPartialData
+    ? `_cw${getMaxHoursToConvert(false, [metric], null)}`
```
lukesonnet, line 240: We might want to explicitly use `isFunnelMetric` so that if that logic changes, this method is aware.

lukesonnet (approved):
looks right, happy to QA if you can't.

PR #6023, opened 2026-06-03
warn when a rule targets a disabled environment

gazzdingo:
Shows a callout in the rule modal when the selected environment(s) are not enabled for the feature, helping users avoid silently saving rules that will have no effect.

### Features and Changes

### Screenshots
image
image


bryce-fitzsimons (approved):
conditionally approved...

check: make sure we show a warning if no environments are on.

PR #6036, opened 2026-06-04
feat(performance): pre-aggregate fact table data nightly for use in incremental refresh CUPED

lukesonnet:
### Summary
Adds an opt-in nightly pipeline that materializes per-(fact table, id type) daily
aggregate tables in the customer's warehouse, and teaches the incremental experiment
refresh to read CUPED covariates from those tables instead of re-scanning raw events.

- New `aggregatedFactTables` registry model (watermark, coverage, schema hashes,
  execution lock, in-flight marker) + `aggregatedFactTableRuns` history model.
- Nightly Agenda poller (every 24h) enqueues one worker per (org, fact table, id type);
  worker runs a dedicated `AggregatedFactTableQueryRunner` DAG: optional drop/create,
  append insert, then a coverage query that advances the event-time watermark.
- Mode is resolved by the driver: full `restate` on first build, forced refresh,
  schema drift, or an incomplete prior write; otherwise `incremental` append of events
  after the watermark.
- Covariate read path (`resolveCovariateInsertPath`) picks aggregated vs. legacy scan
  per metric group behind conservative coverage/freshness gates, and records the
  decision + structured reason to the experiment update execution logger.
- Gated on the `pipeline-mode` premium feature and a configured pipeline write dataset.

### Reliability design
- Single-writer **execution lock** (`currentExecutionId` + heartbeat, 30-min stale
  takeover); all durable writes are compare-and-set on the lock.
- **Reaper** (`expireOldQueries`) finalizes orphaned/stalled runs and releases locks.
- **In-flight write marker** forces a restate on ambiguous failure so an append that
  may have committed without a durable watermark advance never double-counts.

### Recent reliability hardening (this PR's latest commits)
1. In-flight marker is now set only immediately before the first warehouse-committing
   query, so SQL-build/pre-insert failures no longer trigger a spurious nightly restate.
2. Schema-drift detection tolerates metric/slice **removals** (harmless orphan columns)
   while still restating on additions, type changes, and re-adds — cutting avoidable
   full rebuilds.
3. Removed the coverage merge-guard in favor of the authoritative coverage query
   (`MAX(max_timestamp)` is non-null/monotonic over a non-empty table), which also
   fixes a stale `firstEventDate` after a window-trimming restate.

### Test plan
- [x] `pnpm --filter back-end type-check`
- [x] `pnpm --filter back-end test aggregatedFactTable` (33 tests)
- [x] `pnpm --filter back-end test aggregatedCovariateReadPath experiment-update-log`
- [x] Manual: enable on a fact table, verify nightly materialization, force-restate,
      and that an experiment refresh reads the aggregated covariate table (logged reason).


packages/back-end/src/jobs/updateAggregatedFactTables.ts
```diff
@@ -0,0 +1,341 @@
...
+    UPDATE_SINGLE_AGGREGATED_FACT_TABLE,
+    updateSingleAggregatedFactTable,
+  );
+
+  await startUpdateJob();
+
+  async function startUpdateJob() {
+    const updateJob = agenda.create(QUEUE_AGGREGATED_FACT_TABLE_UPDATES, {});
+    updateJob.unique({});
+    updateJob.repeatEvery("24 hours");
```
lukesonnet: Repeat job every 1-2 minutes or whatever and have `queue` use data set on the model to determine update time.

packages/back-end/generated/spec.yaml
```diff
@@ -16490,6 +16506,60 @@ paths:
...
+                idType:
+                  description: >-
+                    Limit the refresh to a single id type. If omitted, all of
+                    the fact table's aggregatedFactTableIdTypes are refreshed.
+                  type: string
+                fullRestate:
+                  description: >-
+                    Drop and recreate the table, re-scanning the retained
+                    window. This is significantly more expensive than the
+                    default incremental append (it scans ~2-3 months of
```
ahdriel, line 16625: nit: update to include the setting name that drive the amount of data here when we can

packages/back-end/src/jobs/updateAggregatedFactTables.ts
```diff
@@ -0,0 +1,341 @@
...
+  forceRestate?: boolean;
+}>;
+
+export default async function (agenda: Agenda) {
+  agenda.define(QUEUE_AGGREGATED_FACT_TABLE_UPDATES, async () => {
+    const factTables = await getAllFactTablesWithAggregatedTablesEnabled();
+
+    for (const factTable of factTables) {
+      for (const idType of factTable.aggregatedFactTableIdTypes ?? []) {
+        await queueAggregatedFactTableUpdate({
```
ahdriel: here we queue all the update as the same time... but maybe that's okay for BigQuery?

packages/back-end/src/jobs/updateAggregatedFactTables.ts
```diff
@@ -0,0 +1,341 @@
...
+    return;
+  }
+
+  await runAggregatedFactTableUpdate(context, factTable, idType, {
+    forceRestate: !!forceRestate,
+  });
+};
+
+// Shared materialization driver for both the nightly worker and the force refresh/restate back door.
+export async function runAggregatedFactTableUpdate(
```
ahdriel: This should be in a service file or somewhere else.

packages/back-end/src/jobs/updateAggregatedFactTables.ts
```diff
@@ -0,0 +1,341 @@
...
+  context: ReqContext,
+  factTable: FactTableInterface,
+  idType: string,
+  {
+    forceRestate,
+    // true (nightly worker): block until queries finish. false (manual UI trigger): return after creating the run and finish in the background.
+    awaitResults = true,
```
ahdriel: default should be `false` so we don't get stuck waiting for the updates in scenario we didn't mean to

packages/back-end/src/integrations/sql/dates/covariate-day-bounds.ts
```diff
@@ -0,0 +1,14 @@
...
+export function snapToUtcDayStart(date: Date): Date {
+  const snapped = new Date(date);
+  snapped.setUTCHours(0, 0, 0, 0);
+  return snapped;
+}
+
+export function precedingUtcDayStart(date: Date): Date {
+  const dayStart = snapToUtcDayStart(date);
+  return new Date(dayStart.getTime() - 24 * 60 * 60 * 1000);
+}
```
ahdriel: nit: these are just date functions and should be in the shared/dates instead

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -420,6 +463,234 @@ export const deleteFactTable = async (
...
+  const factMetrics = await context.models.factMetrics.getAll();
+  const metrics = getAggregatedFactTableMetrics({ factMetrics, factTable });
+  const { factTableSettingsHash, metricState } =
+    buildAggregatedFactTableSchemaState({ factTable, metrics });
+
+  // One row per configured id type, even if it has never been materialized yet.
+  const aggregatedFactTables: AggregatedFactTableStatus[] = idTypes.map(
+    (idType) => {
+      const doc = byIdType.get(idType);
+      const status: AggregatedFactTableMaterializationStatus = !doc
```
ahdriel: nit: should we have a helper for this?

packages/back-end/src/routers/fact-table/fact-table.controller.ts
```diff
@@ -420,6 +463,234 @@ export const deleteFactTable = async (
...
+  status: QueryStatus;
+  runStarted: Date | null;
+  dateCreated: Date;
+  finishedAt: Date | null;
+  error: string | null;
+  queryIds: string[];
+};
+
+// Mirrors QueryRunner.getOverallQueryStatus so the history UI shows the same
+// status the runner used.
```
ahdriel: nit: should we merge them? or would we consider any failed query a complete fail in the context of the aggregated covariate given that it is not recoverable-ish?

packages/back-end/src/jobs/updateAggregatedFactTables.ts
lukesonnet: Now it is per fact table they would all get queued up at the same time, with the new `aggregatedFactTableSettings`.

packages/back-end/src/jobs/updateAggregatedFactTables.ts
lukesonnet: yes, ty.

packages/back-end/src/jobs/updateAggregatedFactTables.ts
lukesonnet: yes, ty, some of the methods here should move.

packages/back-end/src/integrations/sql/dates/covariate-day-bounds.ts
lukesonnet: yes, ty.

packages/back-end/src/routers/fact-table/fact-table.controller.ts
lukesonnet: I'm going to leave this here but clean it up as it behaves differently, thanks.

ahdriel (approved):

PR #6039, opened 2026-06-04
more fixes for a variety of small bugs and UI annoyances for schedules/ramps

bryce-fitzsimons:
Fix a broad variety of minor bugs and UI annoyances:

- Fix "resume after jump to last step" skipping the step's interval and immediately completing the ramp
- Add "complete ramp and disable rule" as a distinct action from "complete ramp" for schedules with a cutoff date
- Show ramp/schedule status and disable date prominently in badge (e.g. "Ramp complete · Disables Jun 15")
- Disable and grey out rule editor fields for terminal simple schedules, with a "Remove schedule" option
- Disable "Restart" when cutoff date has already passed


gazzdingo (approved):
Good to go 🚢

PR #6046, opened 2026-06-06
feat(incremental): Do not force inline mode or full refresh if metric settings changed

ahdriel:
### Features and Changes

When a metric definition change, we consider it incompatible with Incremental Refresh, because they invalidate the previously computed results. This would also fallback to the 'inline' query runner automatically, forcing the experiment out of incremental refresh until a Full Refresh was manually executed.

Now we are smarter about this and we conditionally rebuild just the metric source that has changed, and we recover the user automatically so they can keep benefitting from incremental updates.

### Testing

- Added unit tests
- Manual testing
  - Enable Incremental Refresh
  - Force a Full Refresh
  - Change a metric setting (like conversion window)
  - Click Update
  - Before: observe how you are automatically added to the inline results with no UI
  - After: observe how incremental still runs and only modified metric source is rebuilt


jdorn (approved):
Did not test, but code looks reasonable and fairly straight forward to me.

PR #6047, opened 2026-06-06
Custom Hook Soft Warnings

jdorn:
### Features and Changes

Custom Hooks can now emit "soft" warnings that are able to be skipped.  This interaction works both in the UI and via the REST API so both humans and agents can make informed choices about how to proceed.

image

On the front-end, when submitting, a confirmation modal opens. This hooks into low-level `apiCall` code so it works globally without any per-modal special handling (and also works for actions that are triggered outside of modals).

image

When making a REST API call, the request fails with a 422 response and instructions for how to resolve it.

```sh
curl -X POST http://localhost:3100/api/v2/features \
  -H 'Content-Type: application/json' \
  -H 'Authorization: Bearer secret_user_xxxxxxxxxxx' \
  -d '{"id":"my-feature-3","valueType":"boolean","defaultValue":"false"}'
{
  "message": "You should strongly consider adding a description \
    Either address the warnings or append '?ignoreWarnings=true' to the URL to proceed.",
  "warnings": [
    "You should strongly consider adding a description"
  ]
}
```

Re-submitting the request with `?ignoreWarnings=true` appended to the query string will skip this check and continue:

```sh
curl -X POST http://localhost:3100/api/v2/features?ignoreWarnings=true \
  -H 'Content-Type: application/json' \
  -H 'Authorization: Bearer secret_user_xxxxxxxxxxx' \
  -d '{"id":"my-feature-3","valueType":"boolean","defaultValue":"false"}'
{
  "feature": {
    "id": "my-feature-3",
    "description": "",
    "archived": false,
    "dateCreated": "2026-06-06T02:56:11.081Z",
    ...
```


jdorn:
@cursor review

packages/back-end/src/services/context.ts
```diff
@@ -303,6 +303,15 @@ export class ReqContextClass {
     this.initModels();
   }
 
+  // Whether soft warnings should be ignored for this request
+  // Background jobs have no req and always ignore - users have no way to respond to the feedback.
+  public get ignoreWarnings(): boolean {
+    if (!this.req) return true;
+    const v = this.req.query?.ignoreWarnings;
+    if (Array.isArray(v)) return v.some((x) => x === "true" || x === "1");
```
bryce-fitzsimons: fyi we have a shared `stringToBoolean` method for this

bryce-fitzsimons (approved):

PR #6049, opened 2026-06-06
Lj/session replay app

lillialjackson:
### Features and Changes

This change set includes ui and api additions for the session replays work

### Dependencies

All deps have been deployed prior including terraform, ingest service, CLS. Feature currently sits behind a feature flag and should be only available internally initially.

### Testing

unit, manual e2e

### Screenshots
Screenshot 2026-06-24 at 3 23 40 PM
Screenshot 2026-06-24 at 3 23 55 PM
Screenshot 2026-06-24 at 9 04 51 PM
Screenshot 2026-06-24 at 9 36 58 PM

Screenshot 2026-06-25 at 11 38 46 AM

Project without the sdk associated:
Screenshot 2026-06-25 at 11 29 06 AM


packages/back-end/src/routers/session-replay/session-replay.router.ts
```diff
@@ -0,0 +1,52 @@
...
+import { validateRequestMiddleware } from "back-end/src/routers/utils/validateRequestMiddleware";
+import * as sessionReplayControllerRaw from "./session-replay.controller";
+
+const router = express.Router();
+
+const sessionReplayController = wrapController(sessionReplayControllerRaw);
+const MAX_FILTER_STRING_LENGTH = 255;
+const MAX_SHORT_FILTER_STRING_LENGTH = 64;
+const MAX_URL_FILTER_LENGTH = 2048;
+const MAX_DURATION_SECS = 30 * 24 * 60 * 60;
```
lillialjackson, line 13: some limits on filters, not sure if we need maximums on duration and event count here when we already implement maximums on the recording end for how long a session can be etc

packages/back-end/src/services/clickhouse.ts
```diff
@@ -143,3 +145,202 @@ export async function updateMaterializedColumns({
...
+    /** Filter to sessions where this experiment was exposed */
+    experimentKey?: string;
+    limit?: number;
+    offset?: number;
+  },
+): Promise<SessionReplayRow[]> {
+  const datasource = await getGrowthbookDatasource(context);
+  if (!datasource) return [];
+
+  const integration = getSourceIntegrationObject(
```
lillialjackson, line 242: validate this is the growthbookish way of building our filtered query. Also I believe right now we filter after a single filter is added, thoughts about this vs. adding all filters then building the query and making the call?

packages/back-end/src/services/files.ts
```diff
@@ -13,32 +15,124 @@ import {
...
+  bucket: string,
+  prefix: string,
+): Promise<string[]> {
+  const response = await client.send(
+    new ListObjectsV2Command({ Bucket: bucket, Prefix: prefix }),
+  );
+  return (response.Contents ?? []).map((obj) => obj.Key ?? "").filter(Boolean);
+}
+
+function s3GetSignedReadUrl(
```
lillialjackson: Some of this was a refactor when originally the get/list session logic used a signed url the same way images works, so that there was less code duplication. Not sure if its necessarily worth undoing.

packages/front-end/components/Layout/sidebarNav.ts
```diff
@@ -85,6 +85,13 @@ export const navlinks: SidebarLinkProps[] = [
         href: "/product-analytics/dashboards",
         path: /^product-analytics\/dashboards/,
       },
+      {
+        name: "Session Replay",
+        href: "/session-replay",
+        path: /^session-replay/,
+        beta: true,
```
lillialjackson, line 92: is there an alpha tag if we want to release in an earlier stage?

packages/front-end/components/SessionReplay/player.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }),
+    [],
+  );
+
+  useEffect(() => {
+    if (!containerRef.current) return;
+
+    const measured = measurePlayerDims(containerRef.current);
+
+    // TODO Remove: No-op logger drops warn/log; errors still go to the real console.
```
lillialjackson: this was due to some edge cases around sessions that are missing certain chunks, non blocking but noisy.

packages/front-end/services/AppearanceUIThemeProvider.tsx
```diff
@@ -77,7 +77,7 @@ export const AppearanceUIThemeProvider: FC<PropsWithChildren> = ({
     useState<AppearanceUITheme>(initialSystemTheme);
 
   const initialPreferredTheme =
-    typeof localStorage !== "undefined"
+    typeof window !== "undefined"
```
lillialjackson, line 80: claude decided to clean this up because this matched patterns elsewhere apparently

packages/front-end/styles/global.scss
```diff
@@ -3808,3 +3808,120 @@ input:disabled {
     left: 0 !important;
   }
 }
+
+.rr-player {
```
lillialjackson, line 3812: custom styling for the rrweb player

packages/shared/src/permissions/permissions.constants.ts
lillialjackson, line 1: claude took the lead on following the permissions structure, please double check this as I don't think I know it well enough to catch everything

packages/shared/src/permissions/permissionsClass.ts
lillialjackson, line 1: same here

packages/back-end/src/services/clickhouse.ts
```diff
@@ -164,6 +166,170 @@ export async function updateMaterializedColumns({
...
+  }
+
+  const limit = Math.max(1, Math.min(100, Math.floor(options?.limit ?? 100)));
+  const offset = Math.max(0, Math.floor(options?.offset ?? 0));
+  // Always exclude soft-deleted from the list; callers that explicitly want
+  // them must use a future bulk-list-by-id endpoint that doesn't go through
+  // this function. Add this BEFORE the user-supplied conditions so it can't
+  // be overridden by an empty filter.
+  const allConditions = ["deleted_at IS NULL", ...conditions];
+  const where = `WHERE ${allConditions.join(" AND ")}`;
```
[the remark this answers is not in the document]
lillialjackson: There shouldn't be a org filter here, orgs will only have access to their own orgs session metadata mv in clickhouse. When connecting to the datasource it is scoped to org and  there is no shared tenancy.

packages/back-end/src/services/files.ts
```diff
@@ -74,23 +115,67 @@ const s3Clients = new Map<string, S3Client>();
...
+  return Buffer.concat(chunks);
+}
+
+async function s3ListByPrefix(
+  client: S3Client,
+  bucket: string,
+  prefix: string,
+): Promise<string[]> {
+  const response = await client.send(
+    new ListObjectsV2Command({ Bucket: bucket, Prefix: prefix }),
```
mknowlton89, line 174: I'm learning as I review this PR, so this might not be accurate, but from what I can tell, this is fetching chunks of a particular session recording, correct? Looking at the ListObjectsV2Command, it looks like this only returns a max of 1,000 records. I'm guessing 1000 chunks of a single session is way outside of the length of session we'd support, but wanted to call it out incase I'm incorrect.

packages/front-end/pages/session-replay/index.tsx
```diff
@@ -0,0 +1,1083 @@
...
+          )}
+          {/* Player fills the full container — dimensions measured from container */}
+          {events && (
+            <RrwebPlayer
+              key={selectedSessionId}
+              events={events}
+              ref={playerHandle}
+            />
+          )}
+        </Box>
```
mknowlton89: Should we make this a ternary? Would there be issues with rendering the `RrwebPlayer` if any of the above checks (`!selectedSessionId`, `playerLoading` or `playerError`) are valid?

E.G.
```ts
{!selectedSessionId || playerLoading || playerError ? (
  <Flex>{/* empty/loading/error state */}</Flex>
) : events ? (
  <RrwebPlayer
    key={selectedSessionId}
    events={events}
    ref={playerHandle}
  />
) : null}
```

packages/back-end/src/routers/session-replay/session-replay.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+      eventCountMax?: string;
+      featureKey?: string;
+      experimentKey?: string;
+      page?: string;
+    }
+  >,
+  res: Response<SessionsResponse>,
+) {
+  const context = getContextFromReq(req);
+  const page = parseIntWithDefaultCapped(req.query.page, 1, 1_000_000);
```
mknowlton89: Nit: Looking at other uses for `parseIntWithDefaultCapped` it looks like we use much smaller max values (e.g. `100`). Do we need it to be this high?

packages/front-end/pages/session-replay/index.tsx
```diff
@@ -0,0 +1,1083 @@
...
+                  </Box>
+                );
+              })}
+          </Box>
+
+          {/* Pagination */}
+          <Flex justify="between" align="center" mt="3">
+            <Text color="text-low" size="small">
+              {sessions.length} session{sessions.length === 1 ? "" : "s"}{" "}
+              matched
```
mknowlton89: Nit: Could this be misleading? I don't have my local set up to actually render this, but should we instead say something like "Sessions Shown" - it looks like sessions will always be a max of 100 - so the query might match more than the sessions returned.

packages/back-end/src/services/files.ts
lillialjackson: Sessions will not hit this limit with our current max duration and flush settings, good to maybe revisit for the other paths that hit this

mknowlton89 (approved):

PR #6050, opened 2026-06-06
Invalidate cached experiment data when referenced definitions change

olu-an:
## What this fixes

Several caching layers compare definition **ids** but not the **definitions behind them**. Editing an exposure query, segment, fact-table filter, or metric leaves the stored id unchanged, so data computed from the old definition is silently reused alongside data computed from the new one. Depending on the layer, that means wrong experiment results, wrong power estimates, or missing "settings changed" markers — all with no signal to the user.

This PR closes those gaps in three places, ordered by severity:

### 1. Incremental refresh: settings-hash gate misses the definitions behind the ids (wrong experiment results)

The incremental pipeline appends new rows to a persistent units table on every refresh. A settings hash (`getExperimentSettingsHashForIncrementalRefresh`) gates this: if the experiment configuration changed since the table was built, the refresh must not append.

The hash covered `exposureQueryId` and the segment **id** — but not the exposure query's SQL/userIdType, the segment's definition, or the `phase`/`customFields` SQL-template variables. Editing any of those passes the gate, so rows produced by the new definition get appended to a table built with the old one, and the mixed table is analyzed as if it were consistent.

**Fix:**
- Hash the exposure query SQL + userIdType, the segment definition (including resolved fact-table filter SQL for FACT segments), `phase`, `customFields`, and `lookbackOverride` alongside the existing fields. On mismatch, planning falls back to the standard non-incremental runner (results stay correct); a Full Refresh re-materializes the units table and resumes incremental updates.
- **Version the hash** (`v2:` prefix) and keep the legacy computation for comparing hashes stored by older builds. Without this, every existing pipeline would read as "configuration outdated" on deploy and silently drop out of incremental refresh permanently (the stored hash is only rewritten by the incremental runner itself). With it, existing pipelines keep validating against the legacy inputs and pick up the v2 hash on their next successful incremental update.
- Add compile-time exhaustiveness guards over `ExperimentSnapshotSettings` and over the `ExposureQuery`/`SegmentInterface` fields the hash reads, mirroring the existing guard for metric computed settings: adding a field to any of these types now fails compilation until the field is classified as hashed or intentionally ignored. This is the structural fix for the bug class — the gate can no longer silently fall behind the types it depends on.

`validateIncrementalPipeline` now takes the request context instead of just the org so the gate can load the segment definition; the three call sites are updated.

### 2. Population data: 7-day cache with no definition validation (wrong power estimates)

`postPopulationData` reuses cached population documents for up to 7 days, keyed only on `sourceId` + `userIdType` (there was a `TODO hash metric and datasource to validate cache` at the lookup). Editing a metric's definition, the segment/fact table SQL, or datasource settings within that window silently reused the old means/standard deviations, skewing power calculations.

**Fix:** store a source settings hash and per-metric settings hashes on the document — dateUpdated stamps of the datasource, segment/fact table, each metric, its fact table(s), and (for classic ratio metrics) the denominator metric. Over-invalidation just costs a re-query, so stamps are safer than field enumeration here. The cached document is only returned when the source hash and **every** requested metric's hash match; otherwise all requested metrics are re-queried into a fresh, complete document.

Re-querying everything (rather than just the stale metrics) is deliberate: the new document replaces the cached one as the most recent cache entry and the front-end reads a single document, so a partial document leaves the omitted, still-valid metrics looking like they have no data (`setMetricDataFromPopulationData` zeroes metrics missing from the response). That partial-document hazard predates this PR (it was reachable by adding a metric to a recent power calculation); this PR removes the partial path entirely. The `TODO: incrementally do an update` remains the eventual proper fix.

### 3. Time series: denominator fact-table filters not hashed (missing "settings changed" markers)

`getMetricSettingsHash` (experiment time series) and `getSafeRolloutMetricSettingsHash` both had a `TODO: also include denominator filters?`. Editing a denominator filter changed results without producing the settings-changed marker on the time series. Both TODOs are resolved by hashing the resolved filter SQL, the same way the numerator side already does.

## How to reproduce

**Incremental refresh (before this PR):**
1. On a datasource with pipeline mode set to `incremental`, start an experiment and run an update so the `gb_units_*` table is materialized.
2. Edit the datasource's exposure query SQL (same query id) — e.g. change the join or point it at a different events table. Editing the experiment's segment SQL works too.
3. Trigger a regular (non-full) refresh.
4. The gate passes (only the id was hashed), the update appends rows from the new SQL onto the table built with the old SQL, and results are computed from the mixed table with no warning. After this PR, the same refresh falls back to the standard runner and results are computed consistently; the UI's Full Refresh restores incremental operation.

**Population data (before this PR):**
1. Open the power calculator, select a population (segment or fact table) and metrics, and run it — this caches a population document for 7 days.
2. Edit one of the metrics (SQL, column, capping, …), the segment definition, or — for a classic ratio metric — its denominator metric.
3. Re-run the power calculation with the same population and metrics.
4. The cached values computed from the old definition are returned. After this PR, the cache is invalidated and the metrics are re-queried.

**Time series (before this PR):**
1. Create a ratio fact metric whose denominator applies a fact-table filter, use it in an experiment that has time series data points.
2. Edit the filter's SQL and refresh.
3. No settings-changed marker appears even though the metric's data changed. After this PR, the next data point is marked.

## Rollout notes

- Existing incremental pipelines are **not** invalidated on deploy (legacy hash comparison). The new inputs are enforced from each pipeline's next successful update onward. Corollary: a definition drift that happened entirely *before* this ships can't be detected retroactively — the old inputs were never captured.
- Cached population documents are re-queried once (they carry no hashes).
- Ratio metrics whose denominator references a fact table will show a one-time settings-changed marker on their next time-series data point after deploy, because the hash inputs grew (`getFiltersForHash` returns `undefined` when there's no denominator, so other metrics' hashes are unchanged). This seemed acceptable versus carrying versioned hashes in four more places — happy to add the same legacy-comparison treatment there if preferred.
- The gate loads the segment through the permission-scoped model. If the hash was written by a context that can read the segment and validated by one that can't (or vice versa), the comparison fails conservatively: the refresh falls back to the standard runner (correct results, not incremental) until a Full Refresh under a consistent context.

## Test plan

- New unit tests for the settings hash and matcher: stability, version prefix, endDate exclusion, exposure-SQL / userIdType / segment-SQL / fact-filter sensitivity, and the legacy-hash compatibility path (the legacy algorithm is pinned verbatim in the test so drift fails).
- Existing `snapshot-planning`, `snapshot-lifecycle`, `experimentTimeSeries`, and query-runner suites pass unchanged.
- `tsgo --noEmit` passes for `shared` and `back-end` (the new exhaustiveness guards compile, i.e. all fields of the hashed types are classified).

### DIFF
--- a/packages/back-end/src/enterprise/services/data-pipeline.ts
+++ b/packages/back-end/src/enterprise/services/data-pipeline.ts
@@ -10,8 +10,9 @@ import {
   ExperimentSnapshotSettings,
   MetricForSnapshot,
 } from "shared/types/experiment-snapshot";
-import { OrganizationInterface } from "shared/types/organization";
 import { ExperimentInterface } from "shared/types/experiment";
+import { ExposureQuery } from "shared/types/datasource";
+import { SegmentInterface } from "shared/types/segment";
 import {
   FactMetricInterface,
   FactTableInterface,
@@ -20,11 +21,13 @@ import { FactTableMap } from "back-end/src/models/FactTableModel";
 import { orgHasPremiumFeature } from "back-end/src/enterprise";
 import { SourceIntegrationInterface } from "back-end/src/types/Integration";
 import { getFiltersForHash } from "back-end/src/services/experimentTimeSeries";
+import { ReqContext } from "back-end/types/request";
+import { ApiReqContext } from "back-end/types/api";
 
 // If the given settings / experiment is not compatible with incremental refresh, throw an error.
 // Otherwise, return void.
 export async function validateIncrementalPipeline({
-  org,
+  context,
   integration,
   snapshotSettings,
   metricMap,
@@ -33,7 +36,7 @@ export async function validateIncrementalPipeline({
   incrementalRefreshModel,
   analysisType,
 }: {
-  org: OrganizationInterface;
+  context: ReqContext | ApiReqContext;
   integration: SourceIntegrationInterface;
   snapshotSettings: ExperimentSnapshotSettings;
   metricMap: Map<string, ExperimentMetricInterface>;
@@ -54,7 +57,7 @@ export async function validateIncrementalPipeline({
 
   // Check if organization has the incremental refresh feature
   const hasIncrementalRefreshFeature = orgHasPremiumFeature(
-    org,
+    context.org,
     "incremental-refresh",
   );
   if (!hasIncrementalRefreshFeature) {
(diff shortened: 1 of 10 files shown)


olu-an:
Superseded by #6209, which narrows scope to the two cache sites that return wrong numbers (incremental-refresh units-table gate + product-analytics exploration cache) and uses a read-side freshness check for the latter instead of a hash-format change.

PR #6058, opened 2026-06-08
Add extra hardening for proxy URL resolution

Kevin-Chant:
### Features and Changes

Replaces the url string interpolation with a helper that delegates to `new URL` to ensure that control characters like `@` are sanitized and enforces the origin hasn't changed.

### Testing

Added unit test covering the following cases
**Statsig** (base `https://statsigapi.net/console/v1/`)

| Input | Result |
|-------|--------|
| `gates` | ✅ `https://statsigapi.net/console/v1/gates` |
| `metrics/list` | ✅ `https://statsigapi.net/console/v1/metrics/list` |
| `metrics/metric_source/list` | ✅ `https://statsigapi.net/console/v1/metrics/metric_source/list` |
| `experiments?page=2` | ✅ `https://statsigapi.net/console/v1/experiments?page=2` |
| `/foo` | ✅ `https://statsigapi.net/console/v1/foo` (leading slash stripped) |
| `//evil.com` | ✅ `https://statsigapi.net/console/v1/evil.com` (authority injection neutralized to same-host path) |
| `@evil.com` | ✅ `https://statsigapi.net/console/v1/@evil.com` (stays on host as path segment) |
| `../foo` | 🚫 rejected — escapes base path |
| `https://evil.com` | 🚫 rejected — host mismatch |

**LaunchDarkly** (base `https://app.launchdarkly.com`)

| Input | Result |
|-------|--------|
| `/api/v2/projects?limit=300` | ✅ `https://app.launchdarkly.com/api/v2/projects?limit=300` |
| `/api/v2/flags/x?summary=true` | ✅ `https://app.launchdarkly.com/api/v2/flags/x?summary=true` |
| `api/v2/x` | ✅ `https://app.launchdarkly.com/api/v2/x` |
| `//evil.com` | ✅ `https://app.launchdarkly.com/evil.com` (neutralized to same-host path) |
| `@evil.com/x` | ✅ `https://app.launchdarkly.com/@evil.com/x` (stays on host as path segment) |
| `../etc` | ✅ `https://app.launchdarkly.com/etc` (base path is `/`, so no escape) |

### DIFF
--- a/packages/back-end/src/routers/importing/importing.controller.ts
+++ b/packages/back-end/src/routers/importing/importing.controller.ts
@@ -16,17 +16,29 @@ function validateHttpMethod(method: string): AllowedHttpMethod {
 }
 
 // Resolve a user-supplied path/segment against a fixed base URL and ensure it
-// stays on the expected host. Using the URL constructor neutralizes authority
-// injection (e.g. a leading "@" or "//") that string concatenation would allow,
-// preventing SSRF to arbitrary hosts.
+// stays on the expected host and within the base path. Using the URL
+// constructor neutralizes authority injection (e.g. a leading "@" or "//") that
+// string concatenation would allow, preventing SSRF to arbitrary hosts.
 function resolveProxyUrl(pathOrSegment: string, baseUrl: string): string {
+  const base = new URL(baseUrl);
+
+  // Strip leading slashes so the input always resolves relative to the base
+  // path. Without this, "/foo" would resolve from the host root and drop the
+  // base path entirely (e.g. ".../console/v1/" + "/foo" -> "/foo").
+  const relative = pathOrSegment.replace(/^\/+/, "");
+
   let resolved: URL;
   try {
-    resolved = new URL(pathOrSegment, baseUrl);
+    resolved = new URL(relative, baseUrl);
   } catch {
     throw new UnrecoverableApiError("Invalid request URL.");
   }
-  if (resolved.origin !== new URL(baseUrl).origin) {
+  // Reject anything that changes the host (authority injection) or escapes the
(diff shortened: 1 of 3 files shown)


Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

packages/back-end/src/routers/importing/importing.controller.ts
```diff
@@ -48,9 +65,11 @@ export const proxyStatsigRequest = async (
   // Validate HTTP method
   const validatedMethod = validateHttpMethod(method);
 
-  try {
-    const url = `https://statsigapi.net/console/v1/${endpoint}`;
+  // Resolve the URL before the try block so an invalid/origin-violating URL
+  // surfaces as a 4xx (via the global error handler) rather than a 500.
+  const url = resolveProxyUrl(endpoint, "https://statsigapi.net/console/v1/");
```
jdorn, line 54: Quick thing to sanity check - if any endpoints have a leading slash, this will change the behavior.  Previously it would have double slashes in the URL which are usually just ignored.  However, with the `URL` constructor, it replaces the entire path which would break things completely.

With an endpoint of `/foo`, before it would be `https://statsigapi.net/console/v1//foo`, but now it would be `https://statsigapi.net/foo`.

Kevin-Chant:
cursor review

packages/back-end/src/routers/importing/importing.controller.ts
Kevin-Chant: Good catch. Added some extra logic and a unit test to verify the relevant cases. (See the PR comment for a table)

tzjames (approved):

PR #6060, opened 2026-06-08
Avoid cloneDeep in reduceFeaturesWithPrerequisites

tzjames:
## Summary
- Replaces unconditional `cloneDeep(feature)` in `reduceFeaturesWithPrerequisites` with copy-on-write, since the function only mutates top-level prerequisites and per-rule prerequisites for the target environment.
- Features with no prerequisite changes return the original reference; removed features skip cloning entirely; partial changes use shallow copies of the feature shell and only the rules that actually change.

This reduces lodash CPU on SDK payload rebuilds, especially for orgs with many features where most have no active prerequisites.

## Test plan
- [x] `pnpm --filter back-end test prerequisites.test.ts sdk-payload-generation.test.ts`
- [ ] Confirm SDK payload output is unchanged for features with and without prerequisites after deploy


tzjames:
@cursoragent review

bryce-fitzsimons (approved):

PR #6067, opened 2026-06-09
Visual Editor Improvements

Auz:
Minor improvements to display and functionality of the new visual editor


Auz:
@greptileai review

packages/back-end/src/services/imageOptimization.ts
```diff
@@ -1,71 +1,29 @@
...
-
-// Downscale AI-generated images to a sane longest-edge cap and
-// re-encode as WebP. Gemini returns ~1 MB lossless PNGs; this yields
-// ~80–150 KB WebP (~10× reduction) with no perceptible quality loss.
+// AI-generated images are optimized via the Kraken.io API (resize + WebP),
+// NOT in-process. We deliberately do no CPU-bound image work on the API
+// server: the native `sharp` module's platform binary went missing in
+// production (`pnpm install --no-optional`), and the `wasm-vips` fallback ran
+// synchronously on the main event loop — blocking it long enough to starve
+// the /healthcheck endpoint and get pods killed by the liveness probe.
```
tzjames: This is a bit of an AI slop comment.  Alternate approaches are good to mention in PR comments, but IMO don't belong in code comments.

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
```diff
@@ -282,11 +282,20 @@ export default function ExperimentHeader({
...
+  // trapping the Back button so it could never return to the experiments
+  // list. Once `tab` is already "overview" this is a no-op, so Back exits
+  // cleanly.
   useEffect(() => {
-    if (shouldHideTabs) {
+    if (shouldHideTabs && tab !== "overview") {
       setTab("overview");
     }
-  }, [shouldHideTabs, setTab]);
+  }, [shouldHideTabs, tab, setTab]);
```
tzjames: while I was able to reproduce the tab issue once, I can't do it again. Deleting the `experiments-list-tab` cookie going from one page, to experiments page, then clicking drafts adds "#drafts" not "#overview" now.  Clicking back takes you expectedly back to the previous page.

packages/front-end/components/Experiment/TabbedPage/index.tsx
```diff
@@ -205,20 +205,39 @@ export default function TabbedPage({
...
+          // Non-draft experiments: re-add the remembered tab from state to the
+          // URL so navigating back to the experiment restores the last tab.
+          const newUrl =
+            window.location.href.replace(/#.*/, "") + "#" + tabRef.current;
+          router.replace(newUrl, undefined, { shallow: true }).catch((e) => {
+            if (!e.cancelled) {
+              throw e;
+            }
+          });
+        }
```
tzjames: Like I said above I can't reproduce this now.  Also it is not relevant to the visual editor and should be in a different PR anyway.  I'd suggest removing from this PR and making a new one with a clear test plan that is reproducible. Also I don't quite understand why default was different and this code necessary.

tzjames (approved):
Approving to not block, but please move the default tab fix to a different PR.

packages/front-end/components/Experiment/TabbedPage/ExperimentHeader.tsx
Auz: this change is critical to fix the bug. - the bug is from going from the list page to a draft exp details page.

PR #6069, opened 2026-06-09
Avoid cloneDeep in getFeatureDefinitionsResponse

tzjames:
## Summary
- Skip unconditional `cloneDeep` in `getFeatureDefinitionsResponse` when building SDK cache payloads
- Deep-clone features only when saved-group inlining or secure-attribute hashing will mutate rule conditions; clone experiments only when hashing runs
- Draft experiment filtering uses a shallow `.filter()` instead of cloning the full experiments array up front

Follow-up to #6060 — on the typical refresh path (saved group references off after upstream expansion, no secure-attribute hashing), this removes another full-payload lodash clone per SDK connection rebuild.

## Test plan
- [ ] `pnpm --filter back-end test sdk-payload-generation.test.ts`
- [ ] Verify SDK payload refresh CPU drops for orgs without secure-attribute hashing (Datadog profiler)
- [ ] Confirm encrypted payloads and secure-attribute hashing still work for connections with `hashSecureAttributes` enabled

### DIFF
--- a/packages/back-end/src/services/features.ts
+++ b/packages/back-end/src/services/features.ts
@@ -875,31 +875,44 @@ export async function getFeatureDefinitionsResponse({
   savedGroups?: SavedGroupsValues;
   encryptedSavedGroups?: string;
 }> {
-  features = cloneDeep(features);
-  let processedExperiments: AutoExperiment[] =
-    experiments !== undefined ? cloneDeep(experiments) : [];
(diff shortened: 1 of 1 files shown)


tzjames:
@cursoragent review

Kevin-Chant:
Closing this in favor of a separate PR that fully removes cloneDeep from this path

PR #6073, opened 2026-06-09
Surface incremental-refresh fallback to inline mode (coded reasons, API runner info, UI callout + intercept)

ahdriel:
When pipeline mode cannot be honored, snapshot updates silently fall back to the standard full re-query runner. The reason is a free-text string that dies in a telemetry log. Users keep paying for full scans without knowing; API consumers cannot tell which runner produced a snapshot.

### What changes

**For experimenters.** The results page warns when the next update will run as a full query, with copy naming the cause and the fix. For the one case recoverable by a Full Refresh (experiment settings hash drift), clicking Update opens a modal offering Run Full Refresh (primary), Update anyway, or Cancel. Healthy incremental, pipeline-off, and first-run states render nothing.

**For API consumers.** `POST /api/v1/experiments/:id/snapshot` (and the internal snapshot endpoint) now returns a required `runnerInfo` sibling: `runner: "inline" | "incremental" | "incremental-exploratory"` plus `fallback: {code, message} | null`. Requests are never blocked by a fallback. `GET /experiment/:id/incremental-refresh` additionally returns `nextUpdatePlan`, the prospective answer for a standard update.

### Design notes

- Fallback reasons are a closed Zod enum (`shared/src/validators/incremental-refresh.ts`); a discriminated union makes `incremental` + non-null fallback unrepresentable. `fallback: null` on the `inline` arm means inline was the expected mode (pipeline off/excluded), distinguishing "never expected" from "fell back".
- `assertIncrementalRefreshPrerequisites` is now a thin throwing wrapper over a pure `checkIncrementalRefreshEligibility`; the planner and the prospective check consume the result form, the two incremental query runners keep the throw contract. Single eligibility path; messages unchanged.
- `nextUpdatePlan` is computed by reusing the real `planExperimentSnapshot` (standard, no dimension, latest phase, useCache), so the preview cannot diverge from what Update actually does. Planning failures degrade to `null` rather than breaking the GET.
- Nothing new is persisted on the snapshot document. Runner info is plan-time metadata and rides only creation responses.
- The FE copy map is `Record<IncrementalFallbackCode, FallbackCopy>`, so an unclassified new code fails to compile. `RunQueriesButton` (28 call sites) and `ResultMoreMenu` are untouched; the intercept lives in `RefreshResultsButton`.

### Verification

Type-checks pass on shared, back-end, front-end. 43 backend tests pass across the four affected suites, including a new unit suite covering every eligibility code. OpenAPI spec regenerated and committed. Not yet verified against a live warehouse with pipeline mode enabled; the callout/modal surfaces are exercised by types and lint only, hence the draft state.

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -11604,8 +11604,59 @@ paths:
                 properties:
                   snapshot:
                     $ref: '#/components/schemas/snapshot'
+                  runnerInfo:
+                    anyOf:
+                      - type: object
+                        properties:
+                          runner:
+                            type: string
+                            enum:
+                              - incremental
+                              - incremental-exploratory
+                          fallback:
+                            type: 'null'
+                        required:
+                          - runner
+                          - fallback
+                        additionalProperties: false
+                      - type: object
+                        properties:
+                          runner:
+                            type: string
+                            const: inline
+                          fallback:
+                            anyOf:
+                              - type: object
+                                properties:
+                                  code:
+                                    type: string
+                                    enum:
+                                      - settings-outdated
+                                      - non-fact-metrics
+                                      - no-metrics-selected
+                                      - event-quantile-metric
+                                      - activation-metric
+                                      - skip-partial-data
+                                      - experiment-type
+                                      - no-materialized-units-table
+                                      - datasource-unsupported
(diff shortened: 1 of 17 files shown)


ahdriel:
Closing as I have split this work in different PRs.

PR #6076, opened 2026-06-10
feat(incremental): Auto-promote scheduled incremental update to full refresh when required

ahdriel:
### Features and Changes

When the background experiment update job runs an incremental refresh, but the experiment configuration has changed and we can't do an incremental update anymore, we automatically run a **full refresh** now instead of silently falling back to the non-incremental query runner.

Before, the scheduled refresh would see that we cannot honor an incremental update and fallback to the non-incremental update, which provides the same answers but never rebuilds the incremental pipeline tables.

This meant that the experiment would be 'stuck' in the non-incremental mode until manual action was taken to run a Full Refresh and that would fix it.

Given that both full refresh and non-incremental query the same amount of data, but the former stores intermediate data to optimize future updates, it is better to automatically recover in the background if we can.

#### How

- `planSnapshotQueryRunner()` now takes `triggeredBy`. When the incremental prerequisites fail specifically because the config is outdated **AND** `triggeredBy === "schedule"` **AND** the snapshot is `"standard"`, it re-validates as a full refresh and keeps the `"incremental"` runner with `fullRefresh = true`, rebuilding the pipeline tables from scratch.
- No changes for UI or API triggered updates at this moment

### Testing

- Unit tests


packages/back-end/src/services/experiments.ts
```diff
@@ -1327,6 +1331,105 @@ function resolveFullRefresh(
...
+        runnerKind: "results",
+        incrementalFallbackReason: fullRefreshValidationError,
+        fullRefresh,
+        fullRefreshReason,
+      };
+    }
+  }
+
+  logger.info(
+    `Experiment ${experiment.id} does not support incremental refresh: ${validationError}`,
```
msamper, line 1421: Would we want to note that we're falling back to non incremental here?

msamper: or is that the purpose of `runnerKind`

msamper (approved):

packages/back-end/src/services/experiments.ts
ahdriel: Yes -- it would be captured by the final log in the `runnerKind` property.

PR #6080, opened 2026-06-10
Warn on prerequisite dependents in edit/publish flows and REST API

gazzdingo:
### Features and Changes

When a feature is used as a prerequisite by other features or experiments, changing it can silently affect them. This PR surfaces that at the point of change:

- **UI**: non-blocking warning callout (with the list of dependents) in the publish draft, edit default value, and rule modals.
- **REST API (v1 + v2)**:
  - Single-feature responses now include an optional `dependents` field.
  - Updates return a non-blocking `warnings` array when dependents exist.
  - Archiving or deleting a feature with dependents is blocked unless `bypassDependentsCheck` is passed (body field on update, query param on delete).

### Dependencies

None

### Testing

- New API tests in `test/api/features.test.ts` covering the update warning, archive/delete blocks, and the bypass flag.
- To test the UI: give a feature a prerequisite, then open the publish draft / edit default value / rule modals on the prerequisite feature — a warning callout lists the dependents.
- `pnpm type-check` + lint clean, OpenAPI spec regenerated.

### Screenshots
image
image
image


gazzdingo:
@greptileai

bryce-fitzsimons (changes requested):
I think the UI is pretty heavy, especially since it has an expanded "dependent features" list that can push everything down the modal pretty far.

It also doesn't show up for changing env kill switches or top level prerequisites (which are the riskiest change for dependents).

Also not loving the full screen banners on the Review and Publish tab. Something more inline with the submit review or publish buttons might be better, as a pre-publish check.

It's also worth thinking about how to surface this information in the REST API.

PR #6085, opened 2026-06-11
Fix various bugs from production error logs

jdorn:
### Features and Changes

Fixes 5 bugs discovered in Sentry:
1. Mongo unique index collisions for audit logs and snapshots.  Switched from `uniqid` to `generateId`
2. Bandits with no metrics would attempt to refresh every 10 minutes and immediately fail (just to be requeued 10 minutes later). Fixed by differentiating unrecoverable errors from other errors and disabling automatic updates.
3. Some fact metrics were persisted with `numerator.aggregation === null`, which fails Zod validation (the field is optional, but not nullable).  Added a JIT migration to fix this.
4. Some legacy metrics were persisted with `runStarted` as a string instead of a date.  Added a JIT migration to fix this (plus some tests to verify the fix).
5. Some old audit log events in Mongo are missing a user which crashed the audit history table on the front-end. Added a fallback on the front-end to make this more lenient.

### DIFF
--- a/packages/back-end/src/controllers/reports.ts
+++ b/packages/back-end/src/controllers/reports.ts
@@ -3,7 +3,6 @@ import { DEFAULT_STATS_ENGINE } from "shared/constants";
 import { getValidDate } from "shared/dates";
 import { getSnapshotAnalysis } from "shared/util";
 import { pick, omit } from "lodash";
-import uniqid from "uniqid";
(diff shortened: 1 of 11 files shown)


Kevin-Chant (approved):
I moved the audit user fallback to the model's `toInterface` since there's a few other frontend components that would hit the same error. Other than that LGTM

Kevin-Chant:
cursor review

PR #6087, opened 2026-06-11
Fix undefined being written as null in BaseModel

Kevin-Chant:
### Features and Changes

BaseModel guarantees that the document a write returns equals what a subsequent read returns — `newDoc = { ...doc, ...updates }` is what `create`/`update` return, audit-log, and hand to hooks. The raw MongoDB driver silently broke that invariant: it serializes `undefined` (top-level or nested) as BSON `null`, so `update(doc, { field: undefined })` returned a `newDoc` with the field cleared while the DB stored `null`. That `null` then read back and failed the field's Zod `.optional()` validation and strict-equality checks, which forced repeated symptom patches: call-site undefined filtering (#6085, #5997), per-model `?? undefined` coercions in `migrate()`, and per-operation `ignoreUndefined` in `MetricTimeSeriesModel`. This fixes the root cause in BaseModel so the stored document matches the returned one (legacy mongoose models are untouched).

- **All BaseModel writes pass `ignoreUndefined: true`** — `insertOne` in `_createOne`, `updateOne` in `_updateOne`, and both `dbSafeBulkWrite` paths. Nested `undefined` is now dropped instead of stored as `null`. Per-operation rather than collection-level on purpose: collection options are inherited by query *filters*, and `MetricTimeSeriesModel.getMetricTimeSeriesByMetricIdsQuery` deliberately relies on an `undefined` filter value serializing to `null` (matches null-or-missing).
- **`update(doc, { field: undefined })` resolves against the field's schema.** For a field whose schema permits absence (`.optional()`/`.nullish()`), `undefined` clears it via `$unset` so the DB matches the returned `newDoc` — `undefined` is the typed absence signal, since `.omit().partial().strict()` surfaces `undefined` (never `null`) as the empty value for such a field. For a field that *can't* be undefined (required, or `.nullable()` without `.optional()`), absence is impossible, so an explicit `undefined` is treated as "no change" and dropped — not written, not an error — matching how ORMs ignore `undefined` (see the call-site audit for why this matters). To set a `.nullable()` field to `null`, pass `null` (written as-is, not `$unset`). The undefined is dropped *before* the diff, so the returned doc still equals a subsequent read; `{ field: undefined }` for an already-absent field remains a no-op; empty `$set`/`$unset` operators are guarded. `bulkWrite` `updateOne` ops run the same normalization, so they behave identically to `update()` — non-clearable `undefined` dropped, clearable `undefined` translated to `$unset` (pipeline-style array updates pass through untouched). :warning: Previously *any* `undefined` (top-level or nested) was written as BSON `null`.
- **Write filters are guarded against `undefined`.** With `ignoreUndefined`, an `undefined` value in a write filter would be dropped from the BSON entirely — broadening the filter to match an arbitrary document, where the old `null` serialization matched nothing. `getPrimaryKeyFilter` now throws on a missing/undefined primary key value (covering update/delete/audit ids, and closing a pre-existing hole where lodash `pick` silently dropped absent keys), and `bulkWrite` rejects `updateOne` filters containing top-level `undefined` values. Read filters (`_find`/`_findOne`) intentionally keep default serialization (`undefined` → `null`).
- **Reads strip legacy nulls.** After `migrate()`, top-level keys holding `null` are deleted when the field's Zod schema rejects `null` but accepts `undefined` (computed once per schema, cached in a module-level `WeakMap`). `.nullable()`/`.nullish()` fields keep their nulls — a deliberately-nullable field's `null` is a real value, so the absence/value split tracks the schema exactly. Years of already-written nulls become invisible without a data migration; the existing `migrate()` coercions stay (redundant but harmless).

Note on the mongoose-style "ignore `undefined`" behavior reviewers asked for: it now applies to every field that *can't* be undefined — those undefined-passes are treated as "no change", which is the safe, expected outcome and can't break the partial-update call-sites that spread possibly-undefined values. `undefined` still clears `.optional()`/`.nullish()` fields, because there absence is the field's valid empty state and `undefined` is the only empty value the schema-mirrored `UpdateProps` surfaces for it (there is no `null`). Making `undefined` a no-op *everywhere* would require migrating those fields to `.nullish()` and clearing via `null` — a self-contained follow-up kept out of this PR. Either way the read/write invariant holds: dropped keys never reach `newDoc`, so the returned doc equals a subsequent read.

### Call-site audit

Reviewed every call-site of BaseModel update entry points (`update`, `updateById`, `dangerousUpdate*`, internal `_updateOne`, and the newer `updateWithCas`) across `routers/`, `controllers/`, `api/`, `services/`, `jobs/`, `queryRunners/`, `models/`, and `enterprise/`. JSON request bodies cannot encode `undefined`, so zod-parsed body spreads are unaffected by construction.

**Fields that get cleared** (schema permits absence → `undefined` becomes `$unset`, or `null` for a nullable field). All were previously written as BSON `null`:

| Call-site | Field (schema) | Intent | Verdict |
|---|---|---|---|
| `holdout.controller.ts` | `holdout.analysisStartDate` (`.optional()`) | Clear when reverting analysis → running/draft | `$unset` — Improved |
| `api/fact-metrics/updateFactMetric.ts` | `factMetric.denominator` (`.nullable()`) | Clear when metric type no longer expects a denominator | now assigns `null` (was `undefined`) — a nullable field is cleared with `null`, matching the model's own `upgradeFactMetricDoc` |
| `RevisionModel.ts` (`reopen`) | `revision.resolution` (`.optional()`) | Clear resolution when reopening a discarded revision | `$unset` — Improved |
| `ai.controller.ts` | `aiPrompt.overrideModel` (`.optional()`) | Pass-through of possibly-undefined body value | `$unset` — Improved |
| `RampScheduleModel.ts` | `steps[].holdConditions` (`.optional()`), `steps[].approvalNotes` (`.nullish()`) | Nested undefined inside `$set` values | key dropped — Improved |
| `dashboards.controller.ts`, `ramp-schedule.controller.ts` | `experimentId`, `startDate` (create path) | `x \|\| undefined` literals | key absent — Improved |

**Partial-update sites that spread possibly-`undefined` values for fields that can't be undefined** (required or `.nullable()`-only) — flagged by @ahdriel. Under the "drop undefined for non-clearable fields" rule these are now safe no-ops (the field keeps its value) instead of erroring:

| Call-site | Field (schema) | Now |
|---|---|---|
| `PopulationDataQueryRunner` (`runStarted`) | `runStarted` (`.nullable()`) | no-op when not passed |
| `updateFactMetric` + `postBulkImportFacts` | `winRisk`, `loseRisk` (`z.number()`) | no-op when the deprecated `riskThreshold*` body fields are omitted |
| `ProductAnalyticsExplorationQueryRunner` (`result`) | `result` (required) | no-op on run-start updates that don't carry a result |

One partial-update site did clear an *optional* field when the caller omitted it, so a code fix was still warranted: `saved-queries.controller.ts` now omits `dateLastRan` (required — would no-op) and `dataVizConfig` (`.optional()` — would have `$unset`) when they aren't in the body, rather than passing `undefined`.

Queries that could distinguish `null` from missing (the only genuinely risky class) were checked individually:

| Query | Assessment |
|---|---|
| `MetricTimeSeriesModel` — `{ sourcePhase: { $exists: true, $ne: null } }` | `null` fails `$ne: null`, missing fails `$exists: true` — both excluded, results identical |
| MetricTimeSeries upsert filter `sourcePhase = undefined` | Read filters still serialize `undefined` → `null` (reads intentionally keep default behavior), matching null **or** missing — covers docs written before and after this change |
| `DashboardModel.getDashboardsToUpdate` — `{ nextUpdate: { $exists: false } }` | The only writer always sets a real Date: `determineNextDate` returns `null` only for `type: "never"`, which the dashboard `updateSchedule` union doesn't include. If it ever fired, "absent" matches the query's explicit needs-update branch |
| `AggregatedFactTableModel` (`$exists: false` on `lastScheduledRunAt`), `RampScheduleModel` sparse index on `nextProcessAt` | Fields are `.nullable()`/`.nullish()` — explicit nulls are never touched by this PR |

The feature-flag Constants feature and `updateWithCas` paths merged from `main` were re-audited under these rules and are clean: the Constants update path guards every field with `!== undefined`, and `updateWithCas`'s guard converts an `undefined` guard value to `{ $exists: false }` before it reaches the filter.

Read-normalization risk: no `=== null` checks on top-level BaseModel-read fields exist (the TS types already say `T | undefined`), and the models with `migrate()` null-coercions (`RampScheduleModel`, `DashboardModel`, `FactMetricModel`) had already established null→undefined as the intended per-model semantics.

### Testing

- Unit tests in `BaseModel.test.ts` (50 in the suite, all passing) cover the change, including: the invariant directly (after clearing an optional field, `update()`'s return value equals a subsequent read); `$unset` translation for `.optional()` fields; `undefined` on a non-clearable field is a no-op (not an error, not a write) — for both `update()` and `bulkWrite`; a `null` on a nullable field is written as-is rather than `$unset`; unset-only updates emit no empty `$set`; undefined-for-absent-field no-op; undefined-primary-key rejection; legacy-null stripping on `.optional()` fields vs preservation on `.nullable()`; `_find` stripping; and the `bulkWrite` option threading / `$set`→`$unset` translation / undefined-filter rejection.
- `pnpm --filter back-end type-check` is clean after merging `main`.

Note: the call-site undefined-filtering pattern from #6085 becomes unnecessary and can be removed in a follow-up.


Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

ahdriel:
@Kevin-Chant changes look good and I agree with this direction!

But the current implementation has far reaching consequences IMO that we need to address.

`saved-queries.controller` has the following condition: `dateLastRan: req.body.dateLastRan ? getValidDate(req.body.dateLastRan) : undefined`. Meaning that if there's no `dateLastRan` in the payload, we will set it to `undefined` which in this PR is not compatible with `null` so now we throw instead of coercing to `null`.

Here's all the places the robot flagged this behavior change (undefined was accepted and converted to null before, now it throws):
- PopulationDataQueryRunner: runStarted
- updateFactMetric + postBulkImportFacts: winRisk & loseRisk
- (shared above) saved-queries.controller: dateLastRan
- ProductAnalyticsExplorationQueryRunner: AnalyticsExploration.result

Note: these are ordered in importance (most important at top) as they change behavior without any other code change, while the bottom ones everything still works because it is for internal API that we always send the full payload for.

So it feels that these paths in the past have the potential of cleaning up data when they shouldn't and now we are throwing. Just need to check if this is intentional or if we should clean it up to not break API behavior.

Separately... should TypeScript be complaining here for a field that is `zod.nullable()` and we try to pass `field: undefined` (explicitly naming the key with a invalid type?).

Kevin-Chant:
@ahdriel great catch, and agreed these shouldn't throw. I ended up changing BaseModel to ignore an explicit undefined for any field whose schema can't be undefined (required, or `.nullable()` without `.optional()`). It's kind of a fork in behavior where `undefined` might mean "clear the field" or "do nothing" depending on the schema, but it seems like the best way to get consistency and safety. Now `runStarted`, `winRisk`/`loseRisk`, and `result` are  safe no-ops when omitted, `undefined` still clears `.optional()`/`.nullish()` fields (where absence is the valid empty state), and `null` clears a `.nullable()`. I did fix saved-queries to omit `dateLastRan`/`dataVizConfig` when not sent, since `dataVizConfig` is optional and would otherwise be cleared.

On your TS question, it's allowed because of UpdateProps's `.partial()` usage. Without changing `exactOptionalPropertyTypes` it's not possible to catch these (which was one of the factors for my change above). Now that it's safe though I wouldn't worry too much about it.

ahdriel (approved):
Thanks for making the updates! Awesome addition

As a separate work we can try to remove the `.null()` or `.nullish()` and consolidate on `undefined` for fields that we do not care about & keep `null` only if we want to differentiate between `not set` and `explicitly null`.

packages/back-end/test/models/BaseModel.test.ts
```diff
@@ -17,6 +17,7 @@ const BaseModel = MakeModelClass({
       name: z.string(),
       readonlyField: z.string().optional(),
       testDefaultField: z.string().optional(),
+      nullableField: z.string().nullable().optional(),
```
tzjames, line 20: I think it might be worthwhile to add a test to ensure passing {field: undefined} for a `nullableRequiredField` is a no-op.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -1093,17 +1165,29 @@ export abstract class BaseModel<
...
+            );
+          }
           return {
             updateOne: {
               ...op.updateOne,
-              filter: this.applyBaseQuery(op.updateOne.filter),
+              filter,
+              update: Array.isArray(op.updateOne.update)
+                ? op.updateOne.update
+                : translateUndefinedSetToUnset(op.updateOne.update),
```
tzjames: `_updateOne` drops undefined for non-clearable fields before building the write, but `bulkWrite` only runs `translateUndefinedSetToUnset`.  I think we should probably bring the implementations in-line with each other.

tzjames (changes requested):

tzjames (approved):

PR #6092, opened 2026-06-11
Use Docker Hardened Image (dev variant) for the runtime base

Kevin-Chant:
### Features and Changes

Swaps the final-stage runtime base from `node:24-slim` to the Docker Hardened Image dev variant (`dhi.io/node:24-debian12-dev`) — a continuously-patched, near-zero-CVE base. This eliminates the recurring class of OS-package findings (e.g. `gnutls28`, `libgcrypt20`) that never clear on a rebuild, because the stock base freezes those packages and our Dockerfile never runs `apt-get upgrade`.

The `dev` variant keeps `apt`/shell/root, so the existing `apt-get install` + `pm2` + `pnpm` flow is unchanged. It already bundles `pnpm`, so the `npm install -g pnpm` and npm-removal lines are dropped (they'd collide / no-op otherwise). Build stages stay on the official slim images since they don't ship.

Pulling `dhi.io/*` requires authentication, so this also wires up Docker auth across every build path that compiles the Dockerfile from source:

- **`deploy.yml` (production) and `deploy-branch.yml` (manual).** Add a `docker login dhi.io` step before the Depot build. Depot forwards the runner's registry credentials to its remote builder, so `FROM dhi.io/...` resolves. Both reuse the existing `DOCKER_HUB_*` secrets.
- **`preview.yml` (Fly PR previews).** Fly's remote builder can't authenticate to a private base registry, so previews no longer build on Fly. Instead CI builds `preview/Dockerfile.generated` via Depot (which can auth to `dhi.io`), pushes to `registry.fly.io`, and hands the prebuilt image to `fly-pr-review-apps` via its `image:` input. The build steps are gated on `github.event.action != 'closed'` so teardown is unaffected, and the job gains `id-token: write` for Depot's OIDC auth.
- **`release.yml` / `rollback.yml` are untouched** — they only retag an already-built image (`imagetools create`), so they never pull the base.

Note: only building the image **from source** now requires `docker login dhi.io`. Pulling the published `growthbook/growthbook` image (including existing self-host `docker compose` setups) needs no login — they get the hardened base automatically on their next pull.

:warning: This changes the production runtime base image and the preview build pipeline. Building from source now needs `docker login dhi.io` (DHI Community tier — free, no subscription).

Scan delta vs the current `main` build (`docker scout quickview`):
```
            stock     this branch
Critical      1    ->      0
High         21    ->      8
Medium       13    ->      9
Low          43    ->     54
```
Low rises because the dev variant bundles build tooling (`curl`/`git`/`gcc`/`perl`) carrying won't-fix low CVEs, which aren't removable on the free tier. The distroless runtime variant reaches `0C/6H/5M/11L` but needs a substantially more complex Dockerfile (multi-stage copy-assembly + a custom process supervisor); deferred in favor of this minimal change.

### Testing

- [x] `docker build` completes on the hardened base; container boots (`pm2` starts back-end + front-end), serves on `:3000`/`:3100`; Mongo connects (SCRAM) and an experiment results refresh runs the `gbstats` engine
- [x] `docker scout quickview` shows 0 Critical / 8 High
- [x] **Preview pipeline, end-to-end green** ([run 27664206015](https://github.com/growthbook/growthbook/actions/runs/27664206015)) — Depot pulls the `dhi.io` base, pushes to `registry.fly.io`, `fly-pr-review-apps` deploys by image, healthcheck passes
- [x] **Preview server functionality verified** - login, navigation, and refreshing experiment results works
- [x] **Production build path proven via `deploy-branch`** ([run 27664951467](https://github.com/growthbook/growthbook/actions/runs/27664951467)) — Depot pulls the `dhi.io` base and pushes `growthbook/growthbook:branch-kc_harden-dockerfile-base` (same Depot + `dhi.io`-login pattern as `deploy.yml`)

### DIFF
--- a/.github/workflows/deploy-branch.yml
+++ b/.github/workflows/deploy-branch.yml
@@ -30,7 +30,7 @@ jobs:
       # Required to pull the hardened base image (FROM dhi.io/...) in the Dockerfile.
       # Depot forwards these credentials to its remote builder.
       - name: Login to Docker Hardened Images
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: dhi.io
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
--- a/.github/workflows/deploy-branch.yml
+++ b/.github/workflows/deploy-branch.yml
@@ -22,8 +22,17 @@ jobs:
           ref: ${{ inputs.branch }}
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      # Required to pull the hardened base image (FROM dhi.io/...) in the Dockerfile.
+      # Depot forwards these credentials to its remote builder.
+      - name: Login to Docker Hardened Images
+        uses: docker/login-action@v3
+        with:
+          registry: dhi.io
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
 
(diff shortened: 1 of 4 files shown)


tzjames (approved):
I like the incremental approach here.  I think this is an "easy" win without too much risk.  Also building our own image and pushing to fly registry is a worthy change in its own right as the fly build has had issues in the past.

PR #6093, opened 2026-06-11
redesign(experiments): Display Traffic Allocation as a Diagram

msamper:
### Features and Changes

Introduces a redesigned **Traffic Allocation** experience for the experiment Implementation tab, gated behind the `simple-experiment-flow` feature flag. The old `TrafficAndTargeting` block is replaced with a visual **Traffic Allocation Funnel** that walks top-to-bottom through Namespace → Targeting → Traffic (coverage) → per-variation split, with inline editing and a redesigned variations table.
All user-facing changes are behind the `simple-experiment-flow` flag; with the flag off, the legacy UI and flows are unchanged.

> Note: The simple flow deliberately uses `postExperiment` (partial update) rather than `postExperimentTargeting` (full-payload, phase-rewriting) because it only edits a small subset of fields and shouldn't risk clobbering targeting fields the modal never showed. The new backend guard ensures it can't mutate a live phase.

### Dependencies

#6148 

### Testing

> Enable the `simple-experiment-flow` feature flag before testing the new UI. Spot-check with the flag OFF to confirm the legacy UI is unchanged.
#### Funnel rendering
- [x] Open a **draft** experiment → Implementation tab shows the Traffic Allocation Funnel (Targeting → Traffic → variation cards).
- [x] Experiment **with no targeting** shows the Targeting card as "Everyone".
- [x] Experiment **with** attribute / saved-group / prerequisite targeting shows each section correctly.
- [x] Assignment attribute (and fallback, if set) renders; "Sticky bucketing: Disabled" appears only when disabled.
- [x] Coverage bar width matches the configured coverage %.
- [x] **Namespace** card appears only when a namespace is enabled, with the correct "% traffic included" label; "Add Namespace" button appears only when no namespace is set and namespaces exist in the org.
- [x] Variation fork arrows align with the variation cards across viewport widths (resize from wide → narrow). With 4+ variations, cards stack and only up to 3 arrow columns render (expected).
#### Variation cards / table
- [x] Variation cards show name, description (hidden when empty), and screenshots; carousel arrows (caret icons) navigate images.
- [x] Add-variation button appears for editable, non-running experiments and opens the traffic modal and adds a new variation with a focus on the name field.
- [x] Edit pencil on a variation card opens the metadata editor and auto-focuses that variation's Name field.
- [x] Variation cards collapse on running experiments when none of them have screenshots
- [x] Variation cards are always the same height across rows
#### Editing — draft / not live in payload (safeToEdit = true)
- [x] Edit **Targeting** → save → funnel updates.
- [x] Edit **Traffic** (coverage + weights) → save → coverage bar and splits update.
- [x] Edit **Namespace** → save → namespace card updates.
- [x] Add a variation and adjust weights → save persists.
#### Editing — running & live in SDK payload (safeToEdit = false)
- [x] Pencil / "Add Namespace" edit affordances are hidden on funnel cards.
- [x] Editing traffic/targeting/namespace routes through the "Make Changes" flow (new phase), not an in-place edit.
- [x] Backend guard: attempting to change coverage, variation order/keys, or weights via the POST experiment endpoint returns a 400 ("Cannot change: [<fields>] while the experiment is running and live in the SDK payload"). This should not be doable via the UI.
#### Experiment types
- [x] **Multi-armed bandit**: funnel renders, splits hidden where appropriate.
#### Hash Version Fallback
- [x] Fall back to V1 hashing during experiment creation when an SDK in experiment project doesn't support V2
- [x] Render callout on new create experiment modal when falling back to V1
- [x] Hide hash version selector in **Edit Targeting** modal unless hash version is V1 and outdated SDKs have been updated
#### Flag OFF
- [x] Legacy Implementation tab (TrafficAndTargeting) renders and edits as before.
- [x] Legacy Edit Traffic modal still works.
- [x] Variations render as they do today in production 

### Screenshots

**Better warning states in new creation modal**
Screenshot 2026-07-02 at 11 36 45 AM

**Collapsed description when not added**
Screenshot 2026-07-02 at 11 34 45 AM

**New Traffic Allocation diagram UI**
Screenshot 2026-07-02 at 11 32 29 AM

**New Targeting Modal**
Screenshot 2026-07-02 at 11 40 27 AM

**New Traffic Modal**
Screenshot 2026-07-02 at 11 40 52 AM

**New Pre-launch checklist steps**
Screenshot 2026-07-02 at 11 39 42 AM

**New Drag & Drop image UI**
Screenshot 2026-07-02 at 12 43 32 PM


packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,424 @@
...
+                disabled={isRunning}
+              />
+              <FunnelConnector label={includedLabel} />
+            </>
+          )}
+
+          <FunnelCard
+            title="Targeting"
+            titleColor={!hasConfiguredTargeting ? "text-disabled" : undefined}
+            onEdit={isRunning ? null : editTargeting}
```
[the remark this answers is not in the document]
msamper: This is intentional. Users are expected to use the Make Changes button in the experiment header in this UI when the experiment is running rather than the funnel cards. We want to disable the edit icon on funnel cards when the experiment is running

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,424 @@
...
+              ) : null}
+              <AssignmentAttribute experiment={experiment} />
+            </Flex>
+          </FunnelCard>
+
+          <FunnelConnector />
+
+          <FunnelCard
+            title="Traffic"
+            onEdit={isRunning ? null : editTraffic}
```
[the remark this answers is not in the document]
msamper: This is intentional. See my comment in the other comment about this

packages/front-end/components/Experiment/VariationsTable.tsx
```diff
@@ -205,99 +266,106 @@ export function VariationBox({
...
-                      </ScreenshotUpload>
-                    </>
-                  ) : (
-                    <NoImageBox height={height} canEdit={false} />
-                  )}
-                </>
-              )}
-            </Box>
-          )}
+            {canEdit && onEditMetadata && onEditTraffic ? (
```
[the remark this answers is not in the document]
msamper: That's intentional. We don't need that modal for the legacy table

packages/front-end/components/Experiment/EditTrafficModal.tsx
```diff
@@ -23,22 +31,179 @@ export interface Props {
...
+    ExperimentInterfaceStringDates & {
+      variationWeights: number[];
+      coverage: number;
+    }
+  >({
+    defaultValues: {
+      variations: experiment.variations,
+      variationWeights:
+        latestPhase?.variationWeights ??
+        getEqualWeights(experiment.variations.length, 4),
```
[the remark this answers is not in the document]
msamper: This new version of the EditTrafficForm mirrors `EditVariationsForm` and there is only an `active` status today. Once we support `inactive` phase variations we can revisit this

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -102,6 +106,9 @@ export default function Implementation({
   );
 
   const isHoldout = experiment.type === "holdout";
+  const simpleExperimentFlow = useFeatureIsOn("simple-experiment-flow");
+
+  const showTrafficFunnel = !isHoldout && simpleExperimentFlow;
```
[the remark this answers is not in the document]
msamper: We're keeping the UI for holdouts the same for now

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1625,6 +1627,34 @@ export async function postExperiment(
...
+  const coverageChanged = data.coverage !== undefined;
+  if (
+    experiment.status === "running" &&
+    (variationCountChanged || coverageChanged)
+  ) {
+    const linkedFeaturesForPayload = await getFeaturesByIds(
+      context,
+      experiment.linkedFeatures || [],
+    );
+    const inPayload = includeExperimentInPayload(
```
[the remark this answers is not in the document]
msamper: There's no UI flow that can get us into this state. Wont fix

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,421 @@
...
+  title: string;
+  titleColor?: "text-disabled" | "text-high";
+  inlineSummary?: ReactNode;
+  onEdit?: (() => void) | null;
+  children?: ReactNode;
+  disabled?: boolean;
+}) {
+  return (
+    <Box
+      className="appbox"
```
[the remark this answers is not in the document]
msamper: Design sometimes uses `20px` for padding which is off the Radix spacing scale. Frame is using Radix spacing so I won't use it here. As far as the `appbox` class, that's still what `Frame` uses under the hood. Leaving the `FunnelCard` component as is

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,421 @@
...
+          >
+            {!isHoldout ? (
+              <Box>
+                <Text weight="semibold" color="text-high">
+                  Included in this experiment:{" "}
+                  <Text color="text-high" weight="regular">
+                    {Math.round(phase.coverage * 100)}%
+                  </Text>
+                </Text>
+                <Box
```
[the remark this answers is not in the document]
msamper: For now, no. We aren't using this anywhere else, but will extract in the future if we want to reuse

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
```diff
@@ -324,6 +335,12 @@ const SimpleNewExperimentForm: FC<SimpleNewExperimentFormProps> = ({
       throw new Error("You must select an assignment attribute");
     }
 
+    const hasSDKWithNoBucketingV2 = !allConnectionsSupportBucketingV2(
+      sdkConnectionsData?.connections,
+      project,
```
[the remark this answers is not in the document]
msamper: Fixed!

packages/front-end/components/Experiment/EditTrafficModal.tsx
```diff
@@ -23,22 +31,180 @@ export interface Props {
...
```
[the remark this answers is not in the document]
msamper: Fixed! Seems to be an issue with the existing edit variations form but I solved it here by using latest phase variations

packages/front-end/components/Experiment/EditTrafficModal.tsx
```diff
@@ -23,22 +31,186 @@ export interface Props {
...
+                  screenshots: [],
+                  ...newData,
+                  key: value,
+                };
+              }),
+            );
+            form.setValue(
+              `variationWeights`,
+              v.map((v) => v.weight),
+            );
```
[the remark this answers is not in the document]
msamper: This is a false positive. Verified with the UI and with the robot. This only happens when there are no custom keys

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
```diff
@@ -397,19 +416,39 @@ const SimpleNewExperimentForm: FC<SimpleNewExperimentFormProps> = ({
       header="Create Experiment"
       cta="Create"
       size="lg"
-      ctaEnabled={hasProjectPermission}
+      ctaEnabled={hasProjectPermission && !sdkConnectionsLoading}
```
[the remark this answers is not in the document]
msamper: This is fine. This is rare and even if we do save with hash version v2 the outdated sdks will fall back to v1. I don't want to block creation on this

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -398,6 +330,94 @@ export default function SetupTabOverview({
...
+                        <Button
+                          variant="ghost"
+                          stopPropagation
+                          onClick={() => setShowDescriptionModal(true)}
+                        >
+                          Edit
+                        </Button>
+                      ) : null}
+                      {experiment.description ? (
+                        <FaAngleRight className="chevron" />
```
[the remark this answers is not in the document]
msamper: Don't want to do this in this PR without consulting design

packages/front-end/components/Experiment/EditTrafficModal.tsx
```diff
@@ -23,22 +31,186 @@ export interface Props {
...
+    } else {
+      const latestVariationWeights = latestPhase?.variationWeights ?? [];
+      if (
+        data.variations.length !== data.variationWeights.length ||
+        data.variations.length !== latestVariationWeights.length
+      ) {
+        // only recompute weights if original weights are the wrong size
+        data.variationWeights = getEqualWeights(data.variations.length || 2, 4);
+      } else {
+        data.variationWeights = [...latestVariationWeights];
```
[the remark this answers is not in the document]
msamper: This is an existing issue that is live today via `EditVariationsForm` I don't want to address this here

packages/front-end/components/Experiment/VariationsTable.tsx
```diff
@@ -3,22 +3,61 @@ import {
...
+export const MAX_VARIATION_WIDTH = 336;
+
+// Floor height for the "no image" placeholder when no variation in the row has
+// a screenshot; otherwise it grows to match the row height.
+const NO_IMAGE_MIN_HEIGHT = 72;
+const MAX_IMAGE_HEIGHT = 150;
+
+// Radix Themes breakpoints (px), mirroring `@radix-ui/themes` `--xs`/`--sm`.
+const XS_BREAKPOINT = 520;
+const SM_BREAKPOINT = 768;
```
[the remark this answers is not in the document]
msamper: No shared helpers exist and radix doesn't expose the breakpoint pixel values so I'm keeping these hardcoded values for now

packages/front-end/components/Experiment/CompletedExperimentList.tsx
```diff
@@ -215,7 +215,6 @@ const CompletedExperimentList = ({
                       i={variantImageToShow || 0}
                       v={variations[variantImageToShow || 0]}
                       experiment={e}
-                      showDescription={false}
```
ahdriel, line 218: question: Is this intentional? It changes the behavior on `main` even for those with the flag off.

packages/front-end/components/Experiment/VariationsTable.tsx
```diff
@@ -373,10 +451,42 @@ const VariationsTable: FC<Props> = ({
               shareUid={shareUid}
               shareType={shareType}
               onEditMetadata={onEditMetadata}
+              onEditTraffic={onEditTraffic}
+              showNoImage={experiment.status === "draft"}
```
ahdriel: question: This applies to the previous layout as well, is this intentional?
Also impacts the public experiment page via PublicExperimentOverview

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1625,6 +1627,37 @@ export async function postExperiment(
     validateVariationIds(data.variations);
   }
 
+  const variationCountChanged =
+    !!data.variations &&
+    data.variations.length !== experiment.variations.length;
+  const coverageChanged = data.coverage !== undefined;
```
ahdriel: issue: this is wrong, based on the line above.

we should check if coverage is different than the current coverage to be able to say that it has changed.

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1625,6 +1627,37 @@ export async function postExperiment(
...
+  const variationWeightsChanged = data.variationWeights !== undefined;
```
ahdriel: same here -- should compare against the current value, no?

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1625,6 +1627,37 @@ export async function postExperiment(
...
+    const inPayload = includeExperimentInPayload(
+      experiment,
+      linkedFeaturesForPayload,
+    );
+    if (inPayload) {
+      const field = variationCountChanged
+        ? "the number of variations"
+        : coverageChanged
+          ? "coverage"
+          : "traffic split";
```
ahdriel: nit: should we make this handle more than 1 field? `Cannot change field(s): [list]`

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1939,6 +1972,16 @@ export async function postExperiment(
     changes.phases = phases;
   }
 
+  if (data.coverage !== undefined) {
```
ahdriel, line 1998: nit: should we reuse the variable `coverageChanged`?

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -37,7 +40,8 @@ export interface Props {
   urlRedirects: URLRedirectInterface[];
   mutate: () => void;
   editTargeting?: (() => void) | null;
-  editTraffic?: (() => void) | null;
+  editTraffic?: ((variationId?: string) => void) | null;
```
ahdriel, line 44: `variationId` is not optional I believe -- so we should remove the `?:`

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
```diff
@@ -102,6 +107,13 @@ export default function Implementation({
...
 
   const isHoldout = experiment.type === "holdout";
+  const simpleExperimentFlow = useFeatureIsOn("simple-experiment-flow");
+
+  const safeToEdit =
+    experiment.status !== "running" ||
+    !experimentHasLiveLinkedChanges(experiment, linkedFeatures);
+
+  const showTrafficFunnel = !isHoldout && simpleExperimentFlow;
   const canEditHoldoutDefaultState =
```
ahdriel, line 122: nit: can we add a comment saying that we want to support `holdout` and this is a temporary check?

packages/front-end/components/Experiment/TabbedPage/SetupTabOverview.tsx
```diff
@@ -156,35 +158,48 @@ export default function SetupTabOverview({
...
-                      new Date(experiment.statusUpdateSchedule.startAt),
-                      "MMM d, yyyy 'at' h:mm a (z)",
-                    )
-                  : ""}{" "}
-                <PiPencilSimpleFill className="ml-1" />
-              </Button>
-            </Tooltip>
-          ) : null}
+          <Flex align="center" gap="4">
+            {canEditExperiment && !experiment.description ? (
```
ahdriel: question: these changes are also not gated by the feature-flag, is it intentional?

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,420 @@
...
+  const { namespaces } = useOrgSettings();
+
+  const phase = experiment.phases?.[phaseIndex ?? experiment.phases.length - 1];
+  const hasNamespace = phase?.namespace && phase.namespace.enabled;
+
+  // Fraction of traffic let through by the namespace (1 if none), for connector labels.
+  const namespaceRange =
+    hasNamespace && phase?.namespace
+      ? calculateNamespaceCoverage(phase.namespace)
+      : 1;
```
ahdriel: question: the comment says `1 if none` and the default is `1` when a namespace does not exist.

What does `1` mean here? that all traffic is in the namespace or that none is?

Also should we rename it to `threshold` or something as `range` implies a 2 number pair but we only have one here?

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+  const namespaceName = hasNamespace
```
ahdriel: nit: should we merge the finding the name and finding the range in the same function for a namespace?

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+  const isBandit = experiment.type === "multi-armed-bandit";
+  const isHoldout = experiment.type === "holdout";
+  const isRunning = experiment.status === "running";
+
+  const hasConfiguredTargeting =
+    phase &&
+    ((phase.condition && phase.condition !== "{}") ||
+      (phase.savedGroups && phase.savedGroups.length > 0) ||
+      (phase.prerequisites && phase.prerequisites.length > 0));
```
ahdriel: suggestion: should this be a helper somewhere? feels like a logic we should keep in sync

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
```diff
@@ -0,0 +1,420 @@
...
+              hasConfiguredTargeting ? undefined : (
+                <Text size="large">
+                  <em>Everyone</em>
+                </Text>
+              )
+            }
+            disabled={!safeToEdit}
+          >
+            <Flex direction="column" gap="3">
+              {hasConfiguredTargeting ? (
```
ahdriel, line 273: not sure how to address it but we have a strong link between `hasConfiguredTargeting` and what is rendered inside of it -- meaning that if we change it they can get out of sync.

what do you think about breaking it down into 3 separate consts and then `hasConfiguredTargeting = hasCondition || hasSavedGroup || hasHoldout`

packages/front-end/pages/sdks/[sdkid].tsx
```diff
@@ -65,6 +73,19 @@ export default function SDKConnectionPage() {
     );
   }
 
+  const supportsBucketingV2 =
+    getConnectionSDKCapabilities(connection).includes("bucketingV2");
```
Comment (ahdriel) [inline packages/front-end/pages/sdks/[sdkid].tsx:77]: I think this will return `false` even for the user who has no SDK Connections, which is technically correct, but would render a confusing UI.

packages/front-end/pages/experiment/[eid].tsx
```diff
@@ -134,7 +139,18 @@ const ExperimentPage = (): ReactElement => {
   const editTargeting = canRunExperiment
     ? () => setTargetingModalOpen(true)
     : null;
-  const editTraffic = canRunExperiment ? () => setTrafficModalOpen(true) : null;
+  const editTraffic = canRunExperiment
+    ? (variationId?: string) => {
+        // Callers may pass a DOM event, so only string ids focus a variation.
```
Comment (ahdriel) [inline packages/front-end/pages/experiment/[eid].tsx]: issue (minor): this feels like a code smell to me. We control the callers, so we should be able to address this.

ahdriel (approved):

packages/front-end/pages/sdks/[sdkid].tsx
Comment (msamper) [reply packages/front-end/pages/sdks/[sdkid].tsx:77]: This is on the SDK detail page so they will only see this if they have a connection

packages/front-end/pages/experiment/[eid].tsx
Comment (msamper) [reply packages/front-end/pages/experiment/[eid].tsx]: Fixed!

packages/front-end/components/Experiment/TabbedPage/Implementation.tsx
msamper: It is optional. The Traffic card doesn't pass in a `variationId`

packages/front-end/components/Experiment/CompletedExperimentList.tsx
msamper: This was unintentional, reverted to retain this prop

packages/front-end/components/Experiment/VariationsTable.tsx
msamper: It was but the cards look odd collapsed in the current UI so going to always `showNoImage` when the flag is off

packages/back-end/src/controllers/experiments.ts
msamper: Yup, fixed!

packages/back-end/src/controllers/experiments.ts
msamper: good idea!

packages/back-end/src/controllers/experiments.ts
msamper: Fixed

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
msamper: This isn't really needed anymore. This is from when we always rendered the namespace block to show 100% of traffic was still flowing through. Tweaked the const name and the else case to make more sense

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
msamper: I actually found an existing helper that does exactly that so I moved it to a util file and using that instead now!

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
msamper: great idea! this logic was duplicated in so many places

packages/front-end/components/Experiment/TabbedPage/TrafficAllocationFunnel.tsx
msamper: Went the helper approach but still broke out the 3 into consts

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
```diff
@@ -493,39 +539,45 @@ const SimpleNewExperimentForm: FC<SimpleNewExperimentFormProps> = ({
...
-        )}
-      />
-      {!!holdoutHashAttribute &&
-        form.watch("hashAttribute") !== holdoutHashAttribute && (
-          <HelperText status="warning" size="sm" mb="2">
+      <Box mb="4">
+        <Text as="label" weight="semibold" mb="1">
+          Assignment Attribute
+        </Text>
+        <Text as="div" color="text-mid" mb="2">
```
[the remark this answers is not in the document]
msamper: using those props again

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
```diff
@@ -493,39 +539,45 @@ const SimpleNewExperimentForm: FC<SimpleNewExperimentFormProps> = ({
...
+          Will be hashed together with the Tracking Key to determine which
+          variation to assign variation to assign
```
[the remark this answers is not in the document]
msamper: fixed

packages/front-end/components/Experiment/VariationsTable.tsx
```diff
@@ -331,37 +408,50 @@ const VariationsTable: FC<Props> = ({
...
+            })}
       >
-        {variations.map((v, i) =>
-          variationsList && !variationsList.includes(v.id) ? null : (
+        {variations.map((v, i) => {
+          if (variationsList && !variationsList.includes(v.id)) return null;
+          const box = (
             <VariationBox
-              key={v.id}
               i={v.index}
```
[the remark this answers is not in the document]
msamper: RIght now this should never happen because we reorder phase variations based on the top level variations. I'm fine keeping this for now and we can revisit when we allow deleting variations on a phase

packages/front-end/components/Experiment/TabbedPage/HoldoutEnvironments.tsx
```diff
@@ -9,19 +12,21 @@ export default function HoldoutEnvironments({
   editEnvironments: () => void;
 }) {
   return (
-    <div className="box p-4 my-4">
+    <Frame>
       <div className="d-flex flex-row align-items-center justify-content-between text-dark mb-2">
```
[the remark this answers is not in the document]
msamper: Will do this in another PR. This PR is already way too large

PR #6104, opened 2026-06-12
Validate timestamp column type in exposure queries on save

gouthamx67:
## Summary

Adds timestamp type validation to experiment assignment query validation.

Previously, GrowthBook validated that required columns existed but did not validate their data types. This could allow queries with incorrectly typed timestamp columns (e.g. string/varchar) to be saved successfully and only fail later when generated analysis SQL executed.

This change extends `testQueryValidity()` to:

* Build a type map using datasource metadata when available
* Fall back to `determineColumnTypes()` inference when metadata is unavailable
* Validate that the required `timestamp` column resolves to the `date` type
* Return a clear validation error when the type is invalid

Example:

```text
Column "timestamp" must be date, but is string in experiment assignment query "Logged In Users"
```

## Tests

Added coverage for:

* Invalid timestamp type via row inference
* Invalid timestamp type via metadata
* Valid timestamp via row inference
* Valid timestamp via metadata

## Notes

Some datasources (e.g. Snowflake LIMIT 0 validation path) currently do not return column type metadata, so validation may be skipped in those cases. This is an existing limitation and can be addressed separately.

### DIFF
--- a/packages/back-end/src/services/datasource.ts
+++ b/packages/back-end/src/services/datasource.ts
@@ -417,6 +417,20 @@ export async function testQueryValidity(
       )}`;
     }
 
+    // Validate column data types
+    const typeMap = new Map<string, FactTableColumnType>();
+    results.columns?.forEach((col) => {
+      if (col.dataType) typeMap.set(col.name, col.dataType);
+    });
+    determineColumnTypes(results.results, typeMap).forEach((col) => {
+      if (col.datatype) typeMap.set(col.column, col.datatype);
+    });
+
+    const timestampType = typeMap.get("timestamp");
+    if (timestampType && timestampType !== "date") {
+      return `Column "timestamp" must be date, but is ${timestampType} in experiment assignment query${query.name ? ` "${query.name}"` : ""}`;
+    }
+
     return undefined;
   } catch (e) {
     return e.message;
--- a/packages/back-end/test/services/datasource.test.ts
+++ b/packages/back-end/test/services/datasource.test.ts
@@ -262,4 +262,124 @@ describe("testQueryValidity", () => {
       "testQuery",
     );
   });
+
+  describe("column type validation", () => {
+    it("should return error when timestamp column value is a non-date string — row-based inference", async () => {
+      const query = {
+        id: "u",
+        name: "Test Query",
+        userIdType: "user_id",
+        dimensions: [],
+        hasNameCol: false,
+        query: "SELECT * FROM t",
+      };
+
+      mockDataSourceIntegration.getTestValidityQuery = jest
+        .fn()
+        .mockReturnValue("SELECT * FROM t LIMIT 1");
+      mockDataSourceIntegration.runTestQuery = jest.fn().mockResolvedValue({
+        results: [
+          {
+            user_id: "1",
+            experiment_id: "e1",
+            variation_id: "v1",
+            timestamp: "not-a-date",
+          },
+        ],
+      });
+
+      const result = await testQueryValidity(mockDataSourceIntegration, query);
+
+      expect(result).toBe(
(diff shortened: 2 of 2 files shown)


packages/back-end/src/services/datasource.ts
```diff
@@ -417,6 +417,20 @@ export async function testQueryValidity(
...
+    const typeMap = new Map<string, FactTableColumnType>();
+    results.columns?.forEach((col) => {
+      if (col.dataType) typeMap.set(col.name, col.dataType);
+    });
+    determineColumnTypes(results.results, typeMap).forEach((col) => {
+      if (col.datatype) typeMap.set(col.column, col.datatype);
+    });
+
+    const timestampType = typeMap.get("timestamp");
+    if (timestampType && timestampType !== "date") {
```
najito, line 430: `determineColumnTypes` is only checking for type based on shape and leaning on it to determine the date type will cause for it to fail silently. It's better to only look at the `column.datatype` then otherwise skip the check.

packages/back-end/src/services/datasource.ts
```diff
@@ -417,6 +417,20 @@ export async function testQueryValidity(
...
     }
 
+    // Validate column data types
+    const typeMap = new Map<string, FactTableColumnType>();
+    results.columns?.forEach((col) => {
+      if (col.dataType) typeMap.set(col.name, col.dataType);
+    });
+    determineColumnTypes(results.results, typeMap).forEach((col) => {
+      if (col.datatype) typeMap.set(col.column, col.datatype);
+    });
```
najito, line 427: This is already done before earlier in the file. It's probably worth looking into that and trying to stay dry if possible.

packages/back-end/src/services/datasource.ts
```diff
@@ -417,6 +417,20 @@ export async function testQueryValidity(
...
+
+    const timestampType = typeMap.get("timestamp");
+    if (timestampType && timestampType !== "date") {
+      return `Column "timestamp" must be date, but is ${timestampType} in experiment assignment query${query.name ? ` "${query.name}"` : ""}`;
+    }
```
najito, line 432: Assignment queries and feature-usage queries already share the same “are the columns present?” logic. This PR only patches assignment queries, so the two functions drift further apart.

najito: This also does not cover Snowflake warehouses.

najito: The string should also say that the type should the type for SQL. `Say “must be a timestamp or datetime type.”`

packages/back-end/test/services/datasource.test.ts
```diff
@@ -262,4 +262,124 @@ describe("testQueryValidity", () => {
...
+  describe("column type validation", () => {
+    it("should return error when timestamp column value is a non-date string — row-based inference", async () => {
+      const query = {
+        id: "u",
+        name: "Test Query",
+        userIdType: "user_id",
+        dimensions: [],
+        hasNameCol: false,
+        query: "SELECT * FROM t",
+      };
```
najito, line 275: Could be good to unify the query as a top level variable so that it can be reused across the tests.

packages/back-end/test/services/datasource.test.ts
```diff
@@ -262,4 +262,124 @@ describe("testQueryValidity", () => {
...
+          { name: "variation_id", dataType: "string" },
+          { name: "timestamp", dataType: "date" },
+        ],
+      });
+
+      const result = await testQueryValidity(mockLimitZeroIntegration, query);
+
+      expect(result).toBeUndefined();
+    });
+  });
```
najito, line 384: Adding a test for no type metadata and a string column with date-looking values could be valuable given the topic of the PR as well.

najito (changes requested):

PR #6105, opened 2026-06-12
fix(saved-groups): land on the created draft when editing a group with no revision history

gazzdingo:
### Features and Changes

Editing a saved group that had no revision history (created before revisions existed) bounced you back to the live view instead of the draft it created. The server backfills a merged baseline revision alongside the draft, but the front-end updated the URL before the revisions refetch returned it, so the URL-correction effect couldn't resolve the draft version and reset to live.

- `onRevisionCreated` now awaits the revisions cache refresh before pointing the URL at the new draft
- The URL-correction effect skips correcting while the live version is unknown (refetch in flight)
- Removed a redundant `selectFlow` call in the revert flow that raced the same refetch

### Testing

- Edited saved groups with no revision history (condition + list) and confirmed saving lands on the new draft (`?v=2`)
- Confirmed groups with existing history, auto-publish (still lands on live), and revert flows are unchanged


gazzdingo:
@greptileai

jdorn (approved):

PR #6109, opened 2026-06-14
Feature rule conflict detection

jdorn:
### Features and Changes

Better detection of unreachable feature rules and potential targeting conflicts.  Includes 50+ unit tests for various scenarios.

### Examples

Completely unreachable rule:

image

Hard conflict - there is a direct overlap with an earlier rule.  We're able to identify the exact overlap and which targeted users exactly are being excluded.

image

Soft conflict - there is very likely an overlap with an earlier rule, but we don't know exactly which of the targeted users are impacted.

image

Supports number attributes and ranges (mutually exclusive ranges do not warn, but overlapping ones do):

image

Mutually exclusive targeting on the same attribute is always fine and does not warn:

image

Common beta user rollout pattern does not warn, even though technically those beta users are missing from the 2nd rule. This is to prevent warning fatigue.

image


jdorn:
cursor review

madhuchavva (approved):

bryce-fitzsimons (approved):
approved. fixed a few minor issues (status completeness / tests, UI)

PR #6121, opened 2026-06-15
Fix tests failure

nhat-growthbook:
### Features and Changes
- Remove some redundant event forwarder tests.
- Attempt to spin up worker to avoid test randomly failed


nhat-growthbook:
Close in favour of https://github.com/growthbook/growthbook/pull/6125

PR #6127, opened 2026-06-16
Fix search tokenization

### LINKED ISSUE #6117: [Bug] Feature Search Returns Partial Token Matches for Multi-Token Queries
## Summary

Feature search appears to tokenize feature keys on underscores as intended, but it may be matching individual tokens instead of the full query when multiple tokens are provided.

## Expected Behavior

* A feature with key `foo_bar_baz` should be returned when searching for `bar`.
* Searching for `foo_bar` should only match features containing both tokens (or the full phrase), such as `foo_bar_baz`.

## Actual Behavior

* Searching for `foo_bar` can return features that only contain `foo`, even when `bar` is not present.

## Reproduction

1. Create a feature with key `foo`.
2. Create a feature with key `foo_bar_baz`.
3. Search for `foo_bar`.
4. Observe that results may include the feature named `foo`.

## Notes

Tokenizing on underscores is expected behavior. The issue appears to be that multi-token searches are being treated too permissively, matching partial subsets of the query rather than requiring all query tokens (or the full phrase) to match.

Kevin-Chant:
### Features and Changes

Fixes feature search treating multi-token queries too permissively. MiniSearch's default tokenizer splits on punctuation — including underscores and hyphens — and that same tokenizer was applied to the search query, so `foo_bar` was split into `foo` OR `bar` and matched a feature named just `foo`.

The index still splits keys into parts (so `bar` finds `foo_bar_baz`), but now also indexes the whole key, and the query is split on whitespace only. So `foo_bar` stays a single term that prefix-matches the full key rather than OR-matching its parts.

- Closes #6117

### Testing

Verified against the reported cases (`foo` = a feature named `foo`):

```
"bar"            => foo_bar_baz                  (inner-token search still works)
"foo_bar"        => foo_bar_baz                  (no longer matches "foo")
"multi_account"  => multi_account                (not foo_account_bar)
"comms_foo"     => comms_foo_bar             (not content_foo)
"comms"          => comms_foo_bar             (prefix still works)
"checkout-new"   => checkout-new-flow            (hyphens treated like underscores)
```

Added regression tests in `search.test.ts`; existing search tests still pass.

### DIFF
--- a/packages/front-end/services/search.tsx
+++ b/packages/front-end/services/search.tsx
@@ -80,6 +80,10 @@ export function tagFilterOnClick(
   };
 }
 
+// Whitespace + punctuation (incl. underscores), matching MiniSearch's default
+// tokenizer. Used to split indexed fields into searchable parts.
+const wordSeparators = /[\n\r\p{Z}\p{P}]+/u;
+
 const searchTermOperators = [">", "<", "^", "=", "~", ""] as const;
 
 export type SyntaxFilter = {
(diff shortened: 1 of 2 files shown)


Kevin-Chant:
cursor review

tzjames (approved):

PR #6128, opened 2026-06-16
Fix cross-tenant path traversal in local upload endpoint

Kevin-Chant:
### Features and Changes

The image read handlers validated the requesting org by inspecting only the first path segment (`path.split("/")[0]`). Because `path.join()` collapses `..` afterwards, a path like `org_A/../org_B/2025-06/img.jpeg` passed the org check (first segment `org_A`) but resolved to another org's directory — letting any authenticated user read other orgs' uploaded images on local storage (`UPLOAD_METHOD="local"`).

- Adds `getOrgScopedPath()`, which normalizes the path (collapsing `..`) **before** the org check and requires the result to stay within the org's own folder. Applied to `getImage`, `getSignedImageToken`, and `getSignedPublicImageToken`. The `startsWith(`${orgId}/`)` form also closes a prefix-collision (an org `org_abc_evil` can no longer pass as `org_abc`).
- Hardens `files.ts`: routes all local-storage path joins through a new `resolveUploadPath()` helper that uses a separator-aware boundary check, so a sibling directory like `uploads-evil` can no longer satisfy the previous `indexOf(root) !== 0` prefix match. Replaces four copies of the old check.

:warning: Requests whose normalized path contains `..` (or otherwise leaves the org folder) now return `400`/`403` instead of being served. Legitimate paths (`org_xxx/YYYY-MM/img_*.jpeg`) are unaffected — only S3/GCS-keyed cloud storage was never vulnerable (keys are literal).

### Testing

- [x] Unit tests for `getOrgScopedPath` (`upload.util.test.ts`) — accepts legit/leading-slash/`.` paths; rejects the `../` cross-tenant attack, dir-escape, wrong-org, and prefix-collision cases
- [x] `pnpm --filter back-end type-check` and `eslint` clean
- [ ] Manual (local storage): `GET /upload/<yourOrg>/<month>/` still serves your image
- [ ] Manual (local storage): `curl --path-as-is /upload/<yourOrg>/../<otherOrg>/<month>/` now returns `400` instead of the other org's file

### DIFF
--- a/packages/back-end/src/routers/upload/upload.controller.ts
+++ b/packages/back-end/src/routers/upload/upload.controller.ts
@@ -23,6 +23,7 @@ import {
   getReportByUid,
   getReportsByExperimentId,
 } from "back-end/src/models/ReportModel";
+import { getOrgScopedPath } from "back-end/src/routers/upload/upload.util";
 
 const SIGNED_IMAGE_EXPIRY_MINUTES = 15;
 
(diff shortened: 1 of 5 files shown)


Kevin-Chant:
cursor review

tzjames (approved):

PR #6130, opened 2026-06-16
Add optional shared-secret auth to PYTHON_SERVER_MODE /stats endpoint

Kevin-Chant:
### Features and Changes

When `PYTHON_SERVER_MODE` is enabled, the back-end registers `POST /stats` before all auth middleware (JWT, cookie parsing, CORS), so the endpoint is unauthenticated and accepts payloads up to 50MB. This is fine when the "Jobs Server" is correctly network-isolated, but provides no application-layer protection if it's ever exposed.

This adds an opt-in shared secret:

- If `PYTHON_SERVER_AUTH_TOKEN` is set, the `/stats` endpoint requires it as a `Bearer` token — checked with a constant-time compare **before** the body is parsed, so unauthorized callers can't even push a 50MB payload through `bodyParser`.
- The main back-end sends the same token as an `Authorization` header when proxying to `EXTERNAL_PYTHON_SERVER_URL`.
- Opt-in: when the env var is unset, behavior is unchanged, so existing network-isolated deployments are unaffected.

Documented the new env var in the self-host production guide.

Note: this is the necessary core of an old security-scan finding. Per-IP rate limiting and Zod body validation were also suggested but deliberately skipped — rate limiting has little value for a single trusted internal caller, and a valid-shaped 50MB payload would still pass a Zod check (the Python engine validates its own input). The auth gate is what closes the "unauthenticated endpoint" vector.

### Testing

- [ ] With `PYTHON_SERVER_MODE=true` and no `PYTHON_SERVER_AUTH_TOKEN`: `POST /stats` works as before.
- [ ] With `PYTHON_SERVER_AUTH_TOKEN=secret` set: `POST /stats` with no/incorrect `Authorization` returns `401`; with `Authorization: Bearer secret` it succeeds.
- [ ] With `EXTERNAL_PYTHON_SERVER_URL` + `PYTHON_SERVER_AUTH_TOKEN` set on the main instance: proxied requests carry the bearer token and the jobs server accepts them.

### DIFF
--- a/docs/docs/self-host/production.mdx
+++ b/docs/docs/self-host/production.mdx
@@ -109,6 +109,8 @@ GrowthBook container instances are stateless and can be horizontally scaled behi
 
 The most intensive part of GrowthBook is the stats engine and background jobs. If you are running a high volume of experiments, processing those experiments may impact other requests to growthbook. If you notice this, you can separate concerns by starting one "Jobs Server" with increased resourced as needed and the env vars `PYTHON_SERVER_MODE=true`, `GB_STATS_ENGINE_POOL_SIZE=<number of vCPUs>` on the "Jobs Server" and the others with `GB_STATS_ENGINE_MIN_POOL_SIZE=0`, `EXTERNAL_PYTHON_SERVER_URL=<url of jobs server>`, and `CRON_DISABLED=true`.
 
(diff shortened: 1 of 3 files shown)


Kevin-Chant:
cursor review

tzjames (approved):

PR #6132, opened 2026-06-16
Fix search to match multi-token substrings within keys

Kevin-Chant:
### Features and Changes

Follow-up to #6127. That fix kept a typed query as one whole token and prefix-matched the full key, so an *infix* query like `test_key` found nothing in `search_test_key` (it's a substring in the middle, not a prefix).

MiniSearch has no substring/contains query — it only matches terms by prefix/fuzzy. So instead we index every **word-boundary suffix** of each key (`search_test_key` → `["search_test_key", "test_key", "key"]`), which turns a prefix query into a substring-at-word-boundary match. Punctuation is normalized so `-` and `_` are interchangeable, and matches stay contiguous.

### Testing

Given features `other_search_key`, `search_test_key`, `search-test-key`, `test-my-key`, `foo`, `foo_bar_baz`, `multi_account`:

```
"test_key" / "test-key" => search_test_key, search-test-key   (excludes test-my-key, other_search_key)
"bar"                    => foo_bar_baz
"foo_bar"               => foo_bar_baz                          (not "foo")
"multi_account"         => multi_account                        (not ...sub_account... infix)
"account"               => multi_account, ...sub_account...     (single word still broad)
```

Covered by new hook tests in `search.test.ts` (substring-in-the-middle + hyphen/underscore interchangeability); all 40 search tests pass and `search.tsx` type-checks clean.

### DIFF
--- a/packages/front-end/services/search.tsx
+++ b/packages/front-end/services/search.tsx
@@ -80,9 +80,28 @@ export function tagFilterOnClick(
   };
 }
 
-// Whitespace + punctuation (incl. underscores), matching MiniSearch's default
-// tokenizer. Used to split indexed fields into searchable parts.
-const wordSeparators = /[\n\r\p{Z}\p{P}]+/u;
+// Whitespace separates distinct words; punctuation (incl. _ and -) separates
+// sub-tokens within a word.
+const wordBoundary = /[\n\r\p{Z}]+/u;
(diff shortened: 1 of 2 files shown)


Kevin-Chant:
cursor review

tzjames (approved):

PR #6135, opened 2026-06-16
perf(back-end): speed up /features/stale and /features/dependents on large orgs

anna-yn:
### Features and Changes

`GET /features/stale`, `GET /features/dependents`, and the REST `GET /api/v{1,2}/features/stale` endpoints currently call `getAllFeatures(context, {})`, which loads, hydrates, and JIT-migrates **every** feature document in the org before filtering down to the handful of `?ids=` the request actually asked about. The full fetch is necessary for the reverse-dependency check ("does any other feature/experiment list me as a prerequisite?"), but it doesn't need fully hydrated Mongoose documents — only the fields the dependency graph reads.

On orgs with a few thousand features this dominates request latency and, because Mongoose hydration + `toJSON()` + migration are synchronous, blocks the Node event loop for the duration.

This PR makes two targeted changes with no semantic difference in output:

1. **Lean feature fetch for the dependency graph.** Adds `getAllFeaturesForStaleGraph()` as a sibling of `getAllFeatures()` that uses `.lean()` (skip Mongoose document hydration) and an exclusion projection that drops the four heaviest fields the graph never reads (`description`, `jsonSchema`, `customFields`, legacy `draft`). Same migration step and same permission filter as `getAllFeatures()`, so results are interchangeable for graph callers. The three stale/dependents handlers now use this helper; `getAllFeatures()` and its other call sites are untouched.

2. **Prebuilt experiment→feature prerequisite index.** `isFeatureStale` previously called `getDependentExperiments(feature, allExperiments)` per feature, which is an `O(experiments)` scan each time. `buildFeatureLookups()` already precomputes the analogous feature→feature reverse index; this adds an `experimentDependencyIndex` alongside it (one pass over all experiments' last-phase prerequisites) and threads it through `isFeatureStale` / `getDependentExperiments` as an optional fast path. Falls back to the existing scan when no index is supplied, so existing callers are unaffected.

Includes a small equivalence test that asserts the indexed and scan paths of `getDependentExperiments` return identical results.

### Testing

- `pnpm --filter shared test` — 1006 pass (incl. 69 in `isFeatureStale.test.ts`, 2 new)
- `pnpm --filter back-end test -- features` — 132 pass
- `pnpm --filter shared --filter back-end --filter front-end type-check` — clean


madhuchavva:
Thanks for the contribution @anna-yn. we've some ongoing effort for this usecase. we'll take a look and merge the changes in.

bryce-fitzsimons (approved):
Tested, works as intended. No mongo-compat issues with .lean()

PR #6136, opened 2026-06-16
feat(incremental): warn before falling back from Incremental Pipeline mode

ahdriel:
### Features and Changes

For Incremental Pipeline we have 2 concepts:
- Is this experiment covered by it? Meaning is `pipeline mode` enabled in the datasource and is this experiment not excluded
- Is this experiment supported? Meaning, is there any setting/configuration on this experiment that we cannot honor in Pipeline Mode at the moment? (note: some gaps we already closed, and we will close more, and some we will not)

When the answer to the second question is `no` is where we are making the changes:

**Before**: we would silently fallback to the `non-incremental` refresh and scan the whole experiment data without any information in the UI on how the user can fix it

**Now**: We have 2 interventions, 1 Callout explaining why it is not supported and 1 Confirmation when `Refresh` is clicked, to ensure users understand that the cost of this update will potentially be orders of magnitude different.

This also provides an opportunity for the user to fix the problem if they want, which will make the Callout go away and they can update using the Incremental Pipeline again.

### Screenshots

Screenshot 2026-06-16 at 4 42 53 PM

Screenshot 2026-06-16 at 4 42 36 PM

### Testing

- Unit tests
- Manual testing

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -1237,9 +1237,13 @@ function isIncrementalRefreshEnabledForSnapshot({
   datasource: DataSourceInterface;
   experiment: ExperimentInterface;
 }): boolean {
+  // Pass undefined for the type so the type gate is skipped here:
+  // resolveSnapshotRunner checks it separately to surface a descriptive
+  // fallback reason for unsupported types.
   return isExperimentIncrementalEnabled(
     datasource.settings.pipelineSettings,
     experiment.id,
+    undefined,
   );
 }
 
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -130,7 +130,10 @@ import { ExperimentRefRule, FeatureRule } from "shared/types/feature";
 import { ProjectInterface } from "shared/types/project";
 import { MetricGroupInterface } from "shared/types/metric-groups";
 import { ExperimentQueryMetadata } from "shared/types/query";
-import { isExperimentIncrementalEnabled } from "shared/enterprise";
+import {
+  isExperimentCoveredByIncrementalPipeline,
+  getUnsupportedIncrementalExperimentTypeReason,
+} from "shared/enterprise";
 import { generateId } from "back-end/src/util/uuid";
 import { orgHasPremiumFeature } from "back-end/src/enterprise";
 import { updateExperiment } from "back-end/src/models/ExperimentModel";
@@ -1230,23 +1233,6 @@ export type SnapshotQueryRunnerKind =
   | "incremental"
   | "incremental-exploratory";
 
-function isIncrementalRefreshEnabledForSnapshot({
-  datasource,
-  experiment,
-}: {
-  datasource: DataSourceInterface;
-  experiment: ExperimentInterface;
-}): boolean {
-  // Pass undefined for the type so the type gate is skipped here:
-  // resolveSnapshotRunner checks it separately to surface a descriptive
-  // fallback reason for unsupported types.
-  return isExperimentIncrementalEnabled(
-    datasource.settings.pipelineSettings,
-    experiment.id,
-    undefined,
-  );
-}
-
 export function resolveSnapshotRunner({
   datasource,
   experiment,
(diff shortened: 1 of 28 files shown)


ahdriel:
cursor review

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -1,47 +1,129 @@
...
+  }
+
+  if (
+    !isExperimentIncrementalEnabled(
+      params.pipelineSettings,
+      params.experimentId,
+      params.experimentType,
+    )
+  ) {
+    return "Incremental Pipeline mode is not enabled for this experiment.";
```
lukesonnet, line 109: nit: This isn't really "unsupported" in all cases. Maybe this is just semantics, maybe it speaks to a slightly more important separation of concerns we aren't following here.

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -1,47 +1,129 @@
...
+  if (params.skipPartialData) {
+    return "'Exclude In-Progress Conversions' is not supported with Incremental Pipeline mode while in beta. Please select 'Include' in the Analysis Settings for Metric Conversion Windows.";
+  }
+
+  if (params.activationMetric) {
+    return "Activation metrics are not supported with Incremental Pipeline mode while in beta. Please remove the Activation Metric in the Analysis Settings.";
+  }
+
+  if (params.metrics.length === 0) {
+    return "Experiment must have at least 1 metric.";
```
lukesonnet, line 121: minor: This also doesn't seem like it belongs here, and we could plausibly allow folks to start caching unit exposures before adding metrics.

ahdriel:
cursor review

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -1,47 +1,129 @@
...
+  }
+
+  if (params.metrics.some((m) => !isFactMetric(m))) {
+    return "Legacy metrics aren't supported with Incremental Pipeline mode. Convert them or remove non-Fact Metrics.";
```
lukesonnet, line 125: possible follow-up: might be nice to callout which metrics these are somehow.... probably would need these reasons to be consts that have separate front-end rendering logic dictated by the reason.

packages/front-end/components/Experiment/IncrementalPipelineFallbackDialog.tsx
```diff
@@ -0,0 +1,32 @@
...
+}: {
+  reason: string;
+  onConfirm: () => void | Promise<void>;
+  onCancel: () => void;
+}) {
+  return (
+    <ModalStandard
+      open={true}
+      header="Update without Incremental Pipeline"
+      subheader="This re-scans all experiment data from the start instead of only new data, so it may take longer and cost more."
```
lukesonnet, line 17: I wonder if we should say something like "Please contact your organization's GrowthBook administrators if this is unexpected."

But maybe that's overkill.

packages/back-end/src/services/experiments.ts
```diff
@@ -1237,9 +1237,13 @@ function isIncrementalRefreshEnabledForSnapshot({
   datasource: DataSourceInterface;
   experiment: ExperimentInterface;
 }): boolean {
+  // Pass undefined for the type so the type gate is skipped here:
```
lukesonnet: nit: seems... unnecessary?

packages/front-end/hooks/useExperimentSnapshotUpdate.ts
```diff
@@ -0,0 +1,166 @@
...
+  phase,
+  dimension,
+  mutate,
+  mutateAdditional,
+  setRefreshError,
+  onSuccess,
+  customValidation,
+  experimentSnapshotTrackingProps,
+}: {
+  // Optional so a host component that also serves non-experiment entities (e.g.
```
lukesonnet, line 46: I don't love this, but I think this must be work for later. I guess returning "void" methods is reasonable enough to warn consumers of this method about the no-op behavior when this is missing.

packages/front-end/hooks/useIncrementalPipelineUnsupportedReason.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!datasource) return undefined;
+
+    if (
+      !getIsExperimentIncludedInIncrementalRefresh(
+        datasource,
+        experiment.id,
+        experiment.type,
+      )
+    ) {
+      return undefined;
```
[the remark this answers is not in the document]
ahdriel: This is intentional for now as we work on expanding Incremental support to holdouts and bandits.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -894,14 +914,32 @@ export default function AnalysisSettingsSummary({
         </Box>
       )}
 
+      {latest?.status !== "running" && incrementalUpdatesUnavailable && (
```
lukesonnet: Why have this flicker based on `running` status? I feel like it should just always be on... no reason to hide it when running.

packages/shared/src/enterprise/pipeline.ts
```diff
@@ -1,49 +1,148 @@
...
+      }
+    | undefined;
+  pipelineSettings: DataSourcePipelineSettings | undefined;
+  experimentId: string;
+  orgHasIncrementalPipelineFeature: boolean;
+  skipPartialData: boolean;
+  activationMetric: string | null | undefined;
+  metrics: ExperimentMetricInterface[];
+  experimentType: ExperimentInterface["type"];
+}): string | null {
```
lukesonnet, line 93: nit: I feel like this should return an array of strings so we can better help users triage and problem solve, instead of solving one-by-one.

lukesonnet (approved):

PR #6141, opened 2026-06-17
Prepare OpenAPI spec for Speakeasy

Kevin-Chant:
### Features and Changes

Prepares the generated OpenAPI spec to drive Speakeasy CLI generation (consumed in a separate repo). Two groups of changes: runtime/versioning wiring for the CLI, and a set of spec-correctness fixes that block SDK/CLI codegen (ported from #5883).

**Runtime & versioning**

- **Fix invalid self-hosted server block.** The `https://{domain}/api` server declared a `{domain}` template variable with no matching `variables` entry — invalid OpenAPI 3.1 that strict parsers/generators reject. Adds a `variables.domain` block defaulting to `localhost:3100`.
- **Add `GET /v1/version`.** Returns `{ version, commit, date }` from `getBuild()`, tagged `meta` (not `version`, to avoid colliding with the generated CLI's built-in `version` command). This is the runtime source of the connected server's build for skew checks.
- **Track the app version in the spec.** `info.version` changes from a static `1.0.0` to the app version (e.g. `4.4.0`), giving the CLI a deterministic "generated-against" version straight from the standard OpenAPI field.

**Spec-correctness fixes for codegen (ported from #5883)**

- **Add a description to every 200 response.** OpenAPI generators/linters expect a `description` on each response. Defaults one by HTTP method (`Resource created`/`updated`/`deleted`, else `Successful response`) when a route doesn't supply its own, and makes the field required on the internal `Response` type.
- **Drop `id` from the `putVisualChange` PUT body.** The field is stripped server-side (a client can't rename a visual change via PUT) and keeping it collides with the path `id` param during SDK code generation. Omitted from the validator.
- **Suffix the v1 `Feature*` schemas with `V1`.** `Feature`, `FeatureRule`, `FeatureRevision`, `FeatureEnvironment`, and `FeatureWithRevisions` collide with their V2 counterparts once generators strip the `V2` suffix. Names them explicitly with a `V1` suffix, and extracts the inline `FeatureV2` revision into a named `FeatureRevisionSummary` so it isn't auto-named `FeatureRevision`.

Note: the committed spec deliberately carries **no** git commit/date. Baking those in made `spec.yaml` a moving target — every commit changed the file, so the `generate-openapi` CI check could never pass (writing the commit hash changes the hash). Generation now depends only on `package.json`, so it's deterministic. Live commit/date are available at runtime via `GET /v1/version`; the CLI sources its own generated-against date from the commit date of `spec.yaml` at build time.

Note: the three ported fixes supersede the relevant parts of #5883 (which is stale and conflicting). The deprecated-route-stripping and broader schema rework from that PR are intentionally **not** included here — the CLI keeps deprecated routes as `-vN` commands.

### Testing

- [x] `pnpm generate-openapi` regenerates `packages/back-end/generated/spec.yaml` cleanly, and re-running it produces no diff regardless of the checked-out commit (CI check passes).
- [x] Spec top shows `info.version` = app version and no `x-growthbook-build`; the self-hosted server entry has a `variables.domain` block.
- [x] `curl -H "Authorization: Bearer <key>" http://localhost:3100/api/v1/version` -> `{"version":"4.4.0","commit":"...","date":"..."}` (commit/date empty in dev without `buildinfo`).
- [x] After the `V1` renames, no dangling `$ref`s to the old un-suffixed schema names and no orphaned definitions remain in the spec.
- [x] Every 200 response in the generated spec has a `description`; `putVisualChange` PUT body no longer carries `id`.
- [x] (Optional) Run an OpenAPI validator to confirm the server block is now valid.

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -262,8 +262,10 @@ tags:
     x-displayName: Usage
     description: Usage information for metrics in experiments.
   - name: meta
-    x-displayName: meta
-    description: ''
+    x-displayName: Meta
+    description: >-
+      Server metadata, including the running build's version and commit for
+      version-skew checks.
   - name: Dashboards
     x-displayName: Dashboards
     description: ''
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -370,18 +370,15 @@ tags:
   - name: FactTableFilter_model
     x-displayName: Fact Table Filter
     description: <SchemaDefinition schemaRef="#/components/schemas/FactTableFilter" />
-  - name: Feature_model
-    x-displayName: Feature
-    description: <SchemaDefinition schemaRef="#/components/schemas/Feature" />
   - name: FeatureBaseRule_model
     x-displayName: Feature Base Rule
     description: <SchemaDefinition schemaRef="#/components/schemas/FeatureBaseRule" />
   - name: FeatureDefinition_model
     x-displayName: Feature Definition
     description: <SchemaDefinition schemaRef="#/components/schemas/FeatureDefinition" />
-  - name: FeatureEnvironment_model
-    x-displayName: Feature Environment
-    description: <SchemaDefinition schemaRef="#/components/schemas/FeatureEnvironment" />
+  - name: FeatureEnvironmentV1_model
+    x-displayName: Feature Environment V1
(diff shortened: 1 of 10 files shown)


tzjames (approved):
Looks good. I'm a bit uncertain exactly how the info.version, the version endpoint, and the spec version will interact with Speakeasy.  My main concern is that it would be too noisy about needing to upgrade, or too noisy about it being ahead of self-hosted versions. But I think this is a path forward and we can work out the kinks and any changes needed before we push the cli out to everyone.

PR #6144, opened 2026-06-17
feat(incremental): prompt for a Full Refresh when an incremental update requires one

ahdriel:
### Features and Changes

When an experiment uses Incremental Pipeline mode, some refreshes can't run incrementally — either settings changed in a way that requires rebuilding the pipeline (a Full Refresh), or Overall Results need to be updated before Dimension Results can be. Previously these cases silently fell back to a slower non-incremental update (or read stale data) with no signal to the user.

This PR makes those cases explicit and gives users a way to recover.

**Back-end**

- Manual, user-triggered refreshes now throw structured errors (`requires_full_refresh`, `requires_overall_update`) instead of silently falling back, so the client can prompt the user. Background/scheduled jobs keep their existing automatic behavior (promote to a Full Refresh or downgrade to a non-incremental update).
- Added a `force` flag to `POST /experiments/:id/snapshot` to run a Full Refresh and rebuild the Incremental Pipeline cache. `force` is rejected on Dimension snapshots when Incremental mode is enabled.

**Front-end**

- New confirmation dialogs when a Full Refresh or an Overall Results update is required 
- Some updates to be proactive about what we can check

### Testing

- Unit tests
- Manual tests


ahdriel:
cursor review

ahdriel:
cursor review

ahdriel:
cursor review

ahdriel:
cursor review

packages/shared/src/validators/experiments.ts
```diff
@@ -1731,6 +1731,12 @@ export const postExperimentSnapshotValidator = {
           "Zero-based phase index to snapshot, where 0 is the first experiment phase. Defaults to the latest phase.",
         )
         .optional(),
+      force: z
+        .boolean()
+        .describe(
+          "Set to true to ignore cached query results. For experiments using Incremental Pipeline mode, this runs a Full Refresh for Overall Results and rebuilds the Incremental Pipeline cache. Defaults to false.",
```
lukesonnet: This feels like encoding the wrong behavior. I think using `force` to mean `fullRefresh` was something we wanted to go away from, not enshrine in a public API.

packages/shared/src/validators/api-errors.ts
```diff
@@ -51,6 +51,14 @@ export const apiErrorRegistry = {
       expectedStatuses: z.array(z.string()),
     }),
   },
+  requires_full_refresh: {
+    status: 409,
+    description:
+      "The Incremental Pipeline data is outdated and must be rebuilt before an Incremental Update can proceed",
```
lukesonnet: should you tell me what options I can use to override this? or no.

packages/front-end/hooks/useExperimentSnapshotUpdate.ts
```diff
@@ -9,6 +13,83 @@ import { useUser } from "@/services/UserContext";
 import { getHonoredPrecomputedUnitDimensionIds } from "@/services/experiments";
 import { trackSnapshot } from "@/services/track";
 
+export type SnapshotRefreshBlocker = {
+  kind: "requires-full-refresh";
+  reason: string;
+};
+
+type SubmitUpdateOptions = {
+  force?: boolean;
```
lukesonnet, line 18: It does therefore feel like now is the time to make the `force` change, to split it and name it explicitly for full refresh, given that you added it to the public API in this PR (my other comment)

packages/back-end/src/util/errors.ts
```diff
@@ -173,10 +173,14 @@ export class ConcurrentIncrementalRefreshError extends Error {
...
 
-export class IncrementalUpdateRequiresFullRefreshError extends Error {
-  constructor(message: string) {
-    super(message);
-    this.name = "IncrementalUpdateRequiresFullRefreshError";
+export class ExperimentIncrementalPipelineRequiresFullRefreshError extends ApiError<"requires_full_refresh"> {
+  constructor(reason: string) {
+    super(
+      "requires_full_refresh",
+      `${reason}\n\nRe-issue this request with {"force": true, "dimension": ""} in the request body to run a full refresh.`,
```
lukesonnet: oh i see, i guess you can ignore my other comment.

packages/back-end/src/services/experiments.ts
```diff
@@ -1512,6 +1551,28 @@ export type PlannedExperimentSnapshot = {
...
+  triggeredBy: SnapshotTriggeredBy,
+): boolean {
+  if (!useCache) {
+    return false;
+  }
+
+  switch (triggeredBy) {
+    case "manual":
+      return true;
+    case "manual-dashboard":
```
lukesonnet, line 1565: This one is kind of interesting. In the future we should be more careful about the orchestration of dashboard updates with incremental refresh (it's a mess right now that I'm kind of tackling).

packages/shared/src/validators/experiments.ts
ahdriel: I think we can either add this parameter so the user knows what is happening, or automatically fallback to a Full Refresh in this scenario and drop this parameter. The latter means that sometimes queries will be more expensive/take longer and we are not surfacing the reason why via the API (we do have it in the UI).

What do you think? Or was your suggestion more towards this should be called something else?

packages/back-end/src/services/experiments.ts
```diff
@@ -1268,28 +1272,25 @@ export function resolveSnapshotRunner({
...
-    return { runnerKind: "incremental", incrementalFallbackReason: null };
-  }
-
-  return {
-    runnerKind:
-      snapshotType === "exploratory"
+    return {
+      runnerKind: hasSnapshotDimensions
         ? "incremental-exploratory"
         : "incremental",
```
lukesonnet, line 1288: Pre-existing issue.... but....

Feels like this shouldn't really happen (snapshotType === "exploratory" but has no dimensions).

In the future there maybe other reasons to run an exploratory query, but for now we don't have any so this is probably "safe" but it should be dead code.

ahdriel: What about previous phase? Don't they use exploratory for non-incremental?

packages/front-end/hooks/useExperimentSnapshotUpdate.ts
ahdriel: Keeping it as-is for the internal endpoint as we are having separate conversations for how to handle it better in the public api

ahdriel:
cursor review

packages/back-end/src/services/experiments.ts
lukesonnet: Great question, but then shouldn't this return `incremental-exploratory` for that case? I think the right behavior is that old phases should be more or less locked in place and a main update of an old phase should probably be `incremental`, but only if it is missing data from when it was last updated until the end of the phase.

This would also require the incremental model to begin being created per-phase, which it currently isn't, instead resetting once a phase changes.

packages/back-end/src/util/errors.ts
```diff
@@ -173,10 +173,18 @@ export class ConcurrentIncrementalRefreshError extends Error {
...
 }
 
-export class IncrementalUpdateRequiresFullRefreshError extends Error {
-  constructor(message: string) {
-    super(message);
-    this.name = "IncrementalUpdateRequiresFullRefreshError";
+// Thrown by the Incremental Pipeline when an update can't proceed without a
+// full rebuild. Only the internal snapshot endpoint surfaces this (the public
+// API auto-promotes), so it carries its own code + details for the front-end
+// instead of joining the public REST API error registry.
```
lukesonnet: robot comment explaining how code changed mid pr, one of my favorite flavors of robot comment.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -250,20 +266,138 @@ export default function AnalysisSettingsSummary({
...
+    ],
+  );
+  const hasOverallResults = !!dimensionless && !dimensionless.dimension;
+  // Overall-first rules apply only while incremental updates are active.
+  const dimensionResultsUseOverallResults = isIncremental && !!dimension;
+
+  const fullRefreshReasons: string[] = useMemo(() => {
+    if (!isIncremental || !dimensionless || !hasOverallResults) return [];
+    const phaseStart = experiment.phases?.[phase]?.dateStarted;
+    const currentComparable: IncrementalFullRefreshComparable = {
```
lukesonnet, line 299: question: why rely on a hash at all if we should really just be storing partials so we can diff them? this is a broader question about how incremental refresh works and is prompted by having to do so much to build this partial inside this front-end component.

packages/front-end/components/Experiment/TabbedPage/AnalysisSettingsSummary.tsx
```diff
@@ -250,20 +266,138 @@ export default function AnalysisSettingsSummary({
...
+    dimensionResultsUseOverallResults &&
+    (!incrementalRefresh?.unitsTableFullName || !hasOverallResults);
+  const dimensionIsPrecomputed = isDimensionPrecomputed(
+    dimension,
+    honoredPrecomputedUnitDimensionIds,
+  );
+  const viewingOnDemandDimension = !!dimension && !dimensionIsPrecomputed;
+  const viewingDimensionThatRequiresOverallFirst =
+    dimensionResultsUseOverallResults && viewingOnDemandDimension;
+  const overallResultsRequiredBeforeDimensionRefresh =
```
lukesonnet, line 365: Fun to read all of these combined logical statements.... but seems right to me!

lukesonnet (approved):

Issue #4037, opened 2025-05-07
[Feature] Set minimum/fixed traffic percentage for a variation in a multi-armed bandit


lukesonnet:
### Description of Feature

In some cases, customers may want a bandit to optimize among a set of variations but retain a control (or other variation) with a fixed percentage for the sake of comparison. While it may be best to just use an experiment in these cases, it is a reasonable use case.

Alternatively, we could let the current default of a 1% minimum per arm to be customizable, which may have broader use cases and could yield a similar outcome for the customer who requested the fixed variation percentage.

Internal discussion: https://growthbookapp.slack.com/archives/C06M01XRQF7/p1746452756312229

Whatsonyourmind:
The additive-floor version (your second option) is the cleaner primitive, and it keeps the Thompson guarantee intact. Put the floor on the probability simplex instead of clipping posterior samples: draw the Thompson allocation as usual, then set p_i = f_i + (1 − Σf_j)·w_i, where f_i is arm i's minimum and w_i is its Thompson weight (the posterior probability it's best). Every arm then gets at least f_i, the residual mass is still allocated by the posterior, and Σp_i = 1 falls out automatically, so adaptivity is preserved and you only pay a bounded exploration tax that scales with Σf times the gap.

What to avoid is enforcing the floor by truncating or re-normalizing the sampled probabilities — that biases the posterior and you lose the regret bound. Pinning one variation at a fixed share is then just this with f_control fixed and the others sharing the remainder, so the customizable per-arm floor subsumes that use case and is the more general thing to build.

Whatsonyourmind:
Opened #6145 implementing the additive-simplex floor from above — `p_i = f + (1 − Σf)·w_i` — as a replacement for the current clip-and-renormalize in `Bandits.compute_result()`, plus a `min_variation_weight` plumb-through so the 1% default is configurable at the stats-engine boundary. Scoped to `gbstats` (the allocation method + a `TestBanditMinVariationWeightFloor` class; full suite 32/32 green). I left the TS/DB/UI control for you to drive, since per-arm-vs-uniform and where the control lives are product calls. Default behaviour (0.01) is unchanged.

PR #6145, opened 2026-06-17
fix(gbstats): apply bandit min variation weight as an additive simplex floor

Whatsonyourmind:
## What

The multi-armed bandit's per-variation minimum weight (`min_variation_weight`, default 1%) is enforced by **clip-and-renormalize** in `Bandits.compute_result()`:

```python
p[p < self.config.min_variation_weight] = self.config.min_variation_weight
p /= sum(p)
```

This biases the Thompson allocation. Clipping raises every below-floor arm up to `f` and then renormalizing shrinks **all** surviving arms proportionally — so the floor doesn't just lift the small arms, it also distorts the weights of arms that were already above the floor (and penalizes the leader by more than the floor itself costs). The result is no longer the posterior allocation Thompson sampling intends.

## Change

Apply the floor as an **additive floor on the probability simplex** instead:

```
p_i = f + (1 - n*f) * w_i
```

where `w_i` are the normalized Thompson weights, `f = min_variation_weight`, `n = num_variations`. Properties:

- every arm gets at least `f`;
- `Σ p_i = 1` automatically (no renormalization that re-distorts);
- the relative spacing of all arms **above** the floor is preserved exactly;
- the total exploration "tax" is bounded by `Σf` and allocated by the posterior;
- `f = 0` reduces to pure Thompson; `n*f ≥ 1` (floors can't fit on the simplex) falls back to uniform.

This is the primitive requested in #4037 ("let the … minimum per arm be customizable"). I also plumb `min_variation_weight` through `BanditSettingsForStatsEngine → BanditConfig`, so the value is configurable from the stats-engine settings boundary. **Default behavior is unchanged** (still 0.01).

## Scope

Deliberately scoped to the **stats engine**: it fixes the allocation method (a real bug affecting the shipped 1% default for every bandit) and makes the value configurable at the engine boundary. The end-to-end user-facing control (TS/DB types + UI input) is intentionally left for maintainers, since that's a product decision (uniform vs. per-arm floor, where the control lives, default value). The engine change is behavior-preserving by default and independently valuable.

## Tests

`packages/stats/tests/test_gbstats.py::TestBanditMinVariationWeightFloor` covers:
- the additive identity `p_i = f + (1 − n·f)·w_i`;
- the floor + sum-to-one invariants;
- `f = 0` → pure-Thompson passthrough;
- the `n·f ≥ 1` → uniform fallback.

`python -m unittest tests.test_gbstats` → **32 passed** (29 existing + 3 new), no regressions.

Refs #4037

### DIFF
--- a/packages/stats/gbstats/bayesian/bandits.py
+++ b/packages/stats/gbstats/bayesian/bandits.py
@@ -174,8 +174,21 @@ class Bandits(ABC):
         else:
             p = best_arm_probabilities.copy()
         update_message = "successfully updated"
-        p[p < self.config.min_variation_weight] = self.config.min_variation_weight
-        p /= sum(p)
+        # Apply the per-variation minimum weight as an additive floor on the
+        # probability simplex rather than clipping and renormalizing. Clipping
+        # (p[p < f] = f; p /= sum(p)) raises every below-floor arm up to f and
+        # then shrinks all survivors proportionally, which biases the Thompson
+        # allocation away from the posterior. The additive form
+        #     p_i = f + (1 - n * f) * w_i        (w = normalized Thompson weights)
+        # guarantees p_i >= f and sum(p_i) == 1 while preserving the posterior
+        # ordering and the relative spacing of the arms above the floor.
+        f = self.config.min_variation_weight
+        total_floor = f * self.num_variations
+        if total_floor >= 1.0:
+            # floors cannot all fit on the simplex; fall back to uniform weights
+            p = np.full(self.num_variations, 1.0 / self.num_variations)
+        else:
+            p = f + (1.0 - total_floor) * (p / np.sum(p))
         credible_intervals: List[ResponseCI] = [
             gaussian_credible_interval(mn, s, self.config.alpha)
             for mn, s in zip(self.variation_means, np.sqrt(self.posterior_variance))
--- a/packages/stats/gbstats/bayesian/bandits.py
+++ b/packages/stats/gbstats/bayesian/bandits.py
@@ -182,7 +182,9 @@ class Bandits(ABC):
         #     p_i = f + (1 - n * f) * w_i        (w = normalized Thompson weights)
         # guarantees p_i >= f and sum(p_i) == 1 while preserving the posterior
         # ordering and the relative spacing of the arms above the floor.
-        f = self.config.min_variation_weight
+        # Clamp negatives to 0: a negative floor would make (1 - n*f) > 1 and
+        # push weights outside [0, 1]; treat it as no floor (pure Thompson).
+        f = max(0.0, self.config.min_variation_weight)
         total_floor = f * self.num_variations
         if total_floor >= 1.0:
             # floors cannot all fit on the simplex; fall back to uniform weights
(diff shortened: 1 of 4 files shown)


Whatsonyourmind:
Addressed the one issue from the automated review: a negative `min_variation_weight` would have made `(1 − n·f) > 1` and produced weights outside `[0,1]`. Pushed a guard clamping `f` to ≥ 0 (a negative floor → no floor / pure Thompson), with `test_negative_floor_is_clamped_to_zero`. Full stats suite now 33/33.

packages/stats/gbstats/bayesian/bandits.py
```diff
@@ -174,8 +174,23 @@ def compute_result(self) -> BanditResponse:
...
+        # ordering and the relative spacing of the arms above the floor.
+        # Clamp negatives to 0: a negative floor would make (1 - n*f) > 1 and
+        # push weights outside [0, 1]; treat it as no floor (pure Thompson).
+        f = max(0.0, self.config.min_variation_weight)
+        total_floor = f * self.num_variations
+        if total_floor >= 1.0:
+            # floors cannot all fit on the simplex; fall back to uniform weights
+            p = np.full(self.num_variations, 1.0 / self.num_variations)
+        else:
+            p = f + (1.0 - total_floor) * (p / np.sum(p))
```
lukebrawleysmith, line 193: do we need to divide by np.sum(p) here?

lukebrawleysmith (commented):
Minor comment, otherwise looks good to me.  Thanks for this improvement!

packages/stats/gbstats/bayesian/bandits.py
Whatsonyourmind: Yes — it's load-bearing, mostly because of the `top_two` branch. When `config.top_two` is set, `p = top_two_weights(...)`, which flags both the largest and second-largest column in each row, so it sums to ~2, not 1. The `p / np.sum(p)` turns whatever `p` is into proper Thompson weights `w` with `sum(w) == 1`, which is what makes the floor identity hold: `sum(p_i) = n·f + (1 − n·f)·sum(w) = 1`. Drop it and the top-two case lands around `n·f + (1 − n·f)·2` and breaks the simplex constraint.

It also matters in the non-top-two branch, just more subtly: `best_arm_probabilities` sums to exactly 1 only when every posterior draw has a unique argmax. Exact ties (measure-zero with continuous draws, but possible with identical mean/variance) put two 1s in a row and push the sum above 1, so normalizing keeps `sum == 1` exact in every case. The original clip-and-renormalize relied on the same `p /= sum(p)` for this.

Happy to add a short inline comment making the "normalize to weights first" intent explicit if that reads more clearly.

lukebrawleysmith: Normalizing by the sum is safer, thanks for explaining.

lukebrawleysmith (approved):

Whatsonyourmind:
Gentle nudge on this one — it's been approved since 10 July and I'd rather ask than let it quietly age out.

Everything that can pass does: `ci` and Snyk are both green. The only red is `preview`, and it fails at `flyctl status` with `no access token available` — the Fly deploy job doesn't get repo secrets on a fork PR, so it isn't something I can fix from this side and it isn't telling you anything about the change itself.

Anything you need from me before this can go in?

PR #6146, opened 2026-06-17
Wraps event track call in try/catch

mknowlton89:
### Features and Changes

- Makes license server event tracking best-effort by catching and logging failures inside `notifyLicenseServerEvent`.
- Prevents license server/network errors from blocking experiment status updates when an experiment is started.
- Simplifies the experiment tracking call site now that the notification helper owns its fire-and-forget error handling.

### Dependencies

None.

### Testing

- [x] Verify that you can still send an experiment tracking event to the license server
- [x] Verify if the license server returns an error when tracking an experiment start, it doesn't block starting the experiment

### Screenshots

N/A

### DIFF
--- a/packages/back-end/src/enterprise/licenseUtil.ts
+++ b/packages/back-end/src/enterprise/licenseUtil.ts
@@ -607,7 +607,7 @@ export function shouldNotifyLicenseServer(
 
 /**
  * Notifies the license server of a billable product event, forwarded to Orb
- * for usage-based billing.
+ * for usage-based billing. Errors are swallowed and logged. Fire and forget.
  *
  * @param eventName         - Event name (must be on the license server's allowlist).
  * @param uniqueId          - Natural identifier for the entity this event is about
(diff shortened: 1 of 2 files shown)


packages/back-end/src/enterprise/licenseUtil.ts
```diff
@@ -634,18 +634,26 @@ export async function notifyLicenseServerEvent({
   uniqueId: string;
   metadata: Record<string, unknown>;
   timestampOverride?: string;
-}) {
+}): Promise<void> {
```
jdorn: Having the function as async indicates that callers should await the result. Might be better to make it immediately return instead

packages/back-end/src/enterprise/licenseUtil.ts
```diff
@@ -634,9 +634,10 @@ export async function notifyLicenseServerEvent({
   uniqueId: string;
   metadata: Record<string, unknown>;
   timestampOverride?: string;
-}) {
+}): void {
   const url = `${LICENSE_SERVER_URL}events/track`;
```
mknowlton89, line 638: @jdorn This what you had in mind?

jdorn (approved):

PR #6149, opened 2026-06-17
perf(back-end): use readline for gbstats stdout to avoid event-loop stalls on large analyses

anna-yn:
### Features and Changes

The in-process Python stats server (`services/python.ts`) writes each analysis result as a single JSON line on stdout. For large experiments (many variations × metrics × dimension slices) this line can reach tens of megabytes.

The current stdout handler does:

```js
buffer += rawData.toString();
const lines = buffer.split("\n");
```

on every `data` event. Node delivers stdout in ~64KB chunks, so for a 20MB line this re-scans the growing buffer ~300 times — quadratic in the total payload size — and then parses with `JSON5.parse`, which is substantially slower than the native parser. All of this runs synchronously on the main event loop, so while a large result is being assembled the process cannot serve HTTP requests (including health-check probes).

This change:

- Replaces the manual buffer/split with `readline.createInterface`, which assembles lines incrementally in linear time.
- Parses with native `JSON.parse` as the fast path, falling back to `JSON5.parse` only when the line is not strict JSON. `stats_server.py` writes via `json.dumps(allow_nan=True)`, so the only case that needs the fallback is a result containing `NaN` / `Infinity` literals; everything else takes the fast path.

The per-line handling logic (id lookup, resolve/reject, error logging) is unchanged.

### Dependencies

None.

### Testing

Local benchmark with an ~18MB single-line JSON payload streamed through a `PassThrough` in 64KB chunks:

| Path | Time |
|---|---|
| current (`buffer += / split` + `JSON5.parse`) | ~3.4s |
| `readline` + `JSON5.parse` | ~2.0s |
| `readline` + `JSON.parse` (this PR, fast path) | ~0.14s |

Also ran `pnpm --filter back-end type-check` and `eslint` on the changed file.

To reproduce manually, run an analysis on an experiment with a large number of metric/variation/dimension combinations (no `EXTERNAL_PYTHON_SERVER_URL` set) and observe backend responsiveness while the result is being read.

### Screenshots

N/A — back-end only.

### DIFF
--- a/packages/back-end/src/services/python.ts
+++ b/packages/back-end/src/services/python.ts
@@ -64,6 +64,7 @@ if (IS_CLOUD) {
 
 class PythonStatsServer<Input, Output> {
   private python: ChildProcess;
+  private rl?: readline.Interface;
   private pid = -1;
   private promises: Map<
     string,
@@ -87,7 +88,7 @@ class PythonStatsServer<Input, Output> {
     this.promises = new Map();
 
     if (this.python.stdout) {
-      readline
+      this.rl = readline
         .createInterface({ input: this.python.stdout, crlfDelay: Infinity })
         .on("line", (line) => {
           const output = line.trim();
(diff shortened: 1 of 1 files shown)


ahdriel (approved):
Nice perf improvement!

PR #6155, opened 2026-06-18
fix(data-pipeline): filter NULL idType rows from aggregated fact table INSERT

johnham-ant:
## Summary

Adds `WHERE <idType> IS NOT NULL` to the `__dailyValues` GROUP BY in the aggregated-fact-table INSERT query.

## Why

The aggregated table's grain is `(idType, event_date)`. When the source fact table contains rows with a NULL idType, every one of them lands in a single GROUP BY partition per event_date. On warehouses that hash-partition the aggregation (BigQuery in particular), this produces extreme key skew — one worker gets the entire NULL slice — and on large telemetry-style fact tables the INSERT can stall or livelock on speculative re-execution.

This is a real-world pattern: high-volume telemetry/event tables often carry a meaningful share (10%+) of rows with NULL idType from headless SDK calls, unauthenticated clients, or pre-auth events.

## Why this is safe

- NULL-idType rows can **never join to experiment exposures** (the exposure table is keyed on idType), so they contribute nothing to any analysis read path that consumes the aggregated table. Dropping them is zero data loss for CUPED / covariate purposes.
- The filter is applied at the `__dailyValues` CTE, **not** at `__factTable`, so `__maxTimestamp` still scans the unfiltered slice. The watermark advances past NULL-idType events and they are never re-scanned on the next incremental run.
- No schema change, no read-path change — strictly fewer rows written.

## Testing

Added a unit test in `aggregatedFactTable.test.ts` asserting the generated SQL contains the filter for the configured idType.

https://claude.ai/code/session_01QB3YeVKoESRoMBwttDT1cC

### DIFF
--- a/packages/back-end/src/integrations/sql/queries/insert-aggregated-fact-table-data-query.ts
+++ b/packages/back-end/src/integrations/sql/queries/insert-aggregated-fact-table-data-query.ts
@@ -129,6 +129,13 @@ export function getInsertAggregatedFactTableDataQuery(
           , ${dialect.castToDate("timestamp")} AS event_date
           ${dailyAggregations}
         FROM __factTable
+        -- NULL idType rows can never join to experiment exposures, so they
+        -- contribute nothing to any analysis read path. Dropping them here is
+        -- zero data loss and avoids a single mega-partition in the GROUP BY
+        -- when the source has high-volume unauthenticated/anonymous traffic.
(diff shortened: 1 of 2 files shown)


johnham-ant:
Superseded by #6159, which includes the `WHERE ${idType} IS NOT NULL` filter as part of the salted two-level aggregation rewrite (applied on both the salted and single-level fallback paths).

PR #6159, opened 2026-06-18
fix(agg-fact-table): chunk fullRestate by date to fit BigQuery per-stage write limits

johnham-ant:
## Problem

A wide aggregated fact table — hundreds of metric columns × autoSlices — whose source SQL contains a JOIN can stall the warehouse on the `fullRestate` INSERT. On BigQuery the signature is: `completed_units` freezes mid-stage, `active_units` sawtooths, zero spill, slot-hours climb — the stage is crash-looping. The daily incremental append (single-day slice) completes fine; only the multi-week restate hits the wall.

## Why chunking works (plain language)

The OOM is `rows × columns` per merge-worker memory partition. Column count is fixed (e.g., 530 metric columns); chunking shrinks the **rows** each merge stage handles:

| | Full-window single INSERT | 2-day chunk |
|---|---|---|
| Rows into the GROUP BY merge | ~70B | ~10B |
| Rows per partition at merge | ~3.5M | ~500K |
| Bytes per partition (× wide row) | ~28 GB → OOM crash-loop | ~4 GB → fits |

`event_date` is part of the GROUP BY key, so the number of distinct `(idType, event_date)` groups a merge worker must hold resident scales with the window length. Bounding each INSERT to a 2-day slice cuts the distinct-group count (and therefore per-worker memory) ~7× for a 14-day window — same row width, ~1/7th the groups → fits. Each chunk's merge completes; the table accumulates disjoint `event_date` rows from sequential INSERTs.

**Why sequential (not parallel chunks):** chunks write disjoint date rows so parallel would be correct, but sequential keeps peak slot usage bounded (one ~3K-slot-hr job at a time vs N× concurrently) and reuses the existing QueryRunner dependency graph. Parallel is a possible follow-up if wall-clock matters more than peak slot pressure.

## What we tried (and why salt isn't in this PR)

Two salted two-level GROUP BY approaches were investigated and **rejected** — neither ships in this change:

- **Salted two-level GROUP BY via CTEs.** BigQuery's optimizer rewrites `SUM(SUM(x)) GROUP BY a,b,c` + `SUM(...) GROUP BY a,b` → `SUM(x) GROUP BY a,b`, proves `__salt` is dead, eliminates it, and emits the original single-level plan. Verified via 3-column execution-graph hash (identical to the unsalted plan) and observed 37,700× per-worker skew at the final-merge stage with the CTE form — the salt did nothing.
- **Salted two-level GROUP BY via `CREATE TEMP TABLE` barrier.** The barrier defeats the fold (level-1 runs as written), but at 14-day scale the job *still* stalls — this time on the `S1C: Output` write stage, not the merge. The barrier moved the bottleneck, it didn't remove it: a single statement writing 14 days × hundreds of wide columns × tens of millions of ids exceeds BigQuery's per-stage write budget regardless of how the GROUP BY is shaped.

Date chunking addresses **both** the merge OOM and the output-write ceiling, so the optional salt path has been dropped entirely to keep the change focused — including the `saltBuckets` setting, the `intHash` dialect hook, and the `mergePartialsFunction` aggregation metadata.

## Fix: optional date-chunked restate

`AggregatedFactTableQueryRunner` can slice the restate window into sequential `restateChunkDays`-wide INSERTs. **Chunking is opt-in:** when `aggregatedFactTableSettings.restateChunkDays` is unset, restate runs a single full-window INSERT (the original behavior, unchanged); set it (1–7) to enable chunking on the wide tables that need it.

When enabled:

- DROP + CREATE once, before chunk 0.
- One INSERT per chunk, each scanning a half-open `[chunkStart, chunkEnd)` slice. Chunk i depends on chunk i-1, so they run strictly sequentially and a chunk failure short-circuits the rest via the existing dependency cascade.
- **Chunk seams snap to UTC midnight** (`snapToUtcDayStart`), so a calendar day (`event_date = DATE(timestamp)`, UTC) never straddles two chunks. Without this, a mid-day seam emits two partial rows for the same `(idType, event_date)`; the read path tolerates that (it re-aggregates partials by id, exactly as it already does for incremental appends), but snapping avoids the needless row bloat. The first chunk keeps the exact window lower bound (a short partial leading day, in one chunk); the final chunk has no upper bound (open to "now") so events arriving between planning and execution aren't dropped.
- The coverage query depends on the final chunk; `lastMaxTimestamp` is set from the materialized table after all chunks land — same durable-completion semantics as before.

Each chunk's output fits inside BigQuery's per-stage write limits — the daily incremental already proves a single-day INSERT works, and 2-day chunks were validated end-to-end (see Validation). Incremental mode is unchanged (single INSERT, single-day-ish tail slice). When only one INSERT runs (unchunked restate or incremental) it keeps the original query name/title; multi-chunk restates are labeled `chunk i/N`.

`getInsertAggregatedFactTableDataQuery` gains an optional `windowEndDate` (exclusive) and always emits a single-statement, single-level GROUP BY on all dialects. The `WHERE ${idType} IS NOT NULL` filter and the inline `__max_ts` watermark fix (single source scan, no separate `__maxTimestamp` CTE) are kept.

The OpenAPI spec (`generated/spec.yaml`) is regenerated to expose `restateChunkDays` (1–7) on `aggregatedFactTableSettings`.

## Validation

Anonymized 530-metric-column fact table with a per-row JOIN over ~80B input rows:

| Approach | 14-day restate | Notes |
|---|---|---|
| Single-level GROUP BY, full window (baseline / default) | stalls | final-merge OOM, crash-loop |
| Two-level CTE salt | stalls | optimizer folds salt away, identical plan |
| Temp-table barrier salt | stalls | merge OK, `S1C: Output` write stage stalls |
| **2-day chunks, single-level** | **progresses** | each chunk completes; sequential |

## Testing

- `getRestateChunkBounds`: 14-day window + `chunkDays=2` → 7 contiguous half-open chunks (`chunk[i].end === chunk[i+1].start`), final chunk open-ended; non-even windows; degenerate window → 1 open chunk; mid-day window → internal seams snap to UTC midnight so no calendar day spans two chunks (first chunk keeps the exact start, interior chunks exactly `chunkDays` wide).
- `getInsertAggregatedFactTableDataQuery`: single-level GROUP BY with a single `__factTable` scan, no temp table, no `__salt`; watermark carried as a per-group `MAX` lifted via window (no `__maxTimestamp` CTE); `WHERE ... IS NOT NULL` retained; `windowEndDate` set → exclusive `<` upper bound on `m.timestamp`, unset → open-ended.


lukesonnet:
@greptileai

lukesonnet (approved):

PR #6161, opened 2026-06-18
Gracefully handle background refresh errors in UI

jdorn:
### Features and Changes

We do background data refreshes in the front-end frequently - when a tab is refocused, after a delay, when an edit is performed.  If there's an error in the background refresh (e.g. temporary network error), it currently wipes the entire page and replaces it with a full screen error message.

This PR changes it to instead continue showing stale data while it tries to refresh automatically in the background.  If an error is still occurring after a retry, show a Toast notification in the bottom right corner.

As part of this PR, added a new UI Toast component to the design system - positioned above the in-app help in the bottom right corner.

Current behavior on main when a background refresh has an error:

image

New behavior (after a retry continues to fail):

Screenshot 2026-06-17 at 11 37 41 PM

This PR also handles the case when an auth token expires.  This can often not be recovered automatically and requires a redirect to the IdP server to sign in again.  Previously, when this was encountered from a background refresh, it would redirect immediately and lose all current page state.  Now, it shows a Toast notification and only redirects on user click.


packages/front-end/hooks/useApi.ts
```diff
@@ -37,9 +41,51 @@ export default function useApi<Response = unknown>(
...
+  useEffect(() => {
+    if (!backgroundRefreshError || !activeKey) return;
+    const err = refreshErrorRef.current;
+    if (hasRefreshError && err) {
+      backgroundRefreshError.report(activeKey, err);
+    } else {
+      backgroundRefreshError.clear(activeKey);
+    }
+    // Drop this key on unmount/key change so a navigated-away component can't keep the toast alive.
+    return () => backgroundRefreshError.clear(activeKey);
```
najito, line 74: If this is tied to the usage of `useApi` does that mean that a user could be in a state where they clear the key before being out of the state where the toast is relevant?

packages/front-end/hooks/usePrerequisiteStates.ts
```diff
@@ -133,6 +140,32 @@ export function useBatchPrerequisiteStates({
...
+
+  useEffect(() => {
+    if (!backgroundRefreshError || !key) return;
+    const err = refreshErrorRef.current;
+    if (hasRefreshError && err) {
+      backgroundRefreshError.report(key, err);
+    } else {
+      backgroundRefreshError.clear(key);
+    }
+    return () => backgroundRefreshError.clear(key);
```
najito, line 166: Does this clear also possibly clear when this hook is reused while staying on the same page? I'm not sure about the overlap of where these hooks are used.

PR #6162, opened 2026-06-18
fix(snapshot): prevent query-DAG persistence race causing phantom 'queued' queries

johnham-ant:
## Problem

Snapshots can stick at "running" forever with a handful of queries showing "queued" but nothing actually executing. The orphan reaper (added in #5885) eventually marks them errored after an hour, but does not retry.

Observed on a self-hosted deployment (Agenda split into a dedicated jobrunner pod):

```
23:20:38 backend  WARN: No running or queued queries for snp_2Cc9... but runner status is "pending".
                        Likely a query finished before the full query DAG was persisted; a follow-up refresh is scheduled.
00:20:37 jobrunner INFO: Reaped orphaned snapshot snp_2Cc9... 6 of 122 queries stuck in "queued" with nothing running
```

Hit 6 experiments in one night.

## Root cause

`startQuery()` dispatches dependency-free queries fire-and-forget *before* `startAnalysis()` has written the full `queries` array to the snapshot doc (`QueryRunner.ts` `startAnalysis()` → `updateModel()`). A fast query (e.g. an Incremental Pipeline `DROP TABLE`) can finish — and call `onQueryFinish()` — while `startQueries()` is still generating SQL for the remaining 100+ queries.

#5885 added a post-persist `onQueryFinish()` in `startAnalysis()` to re-drive the DAG after the write commits. But `onQueryFinish()` is debounced (`if (!this.timer)`), so when an early-finish timer is still pending the post-persist call is a **no-op**. The pending timer's `getLatestModel()` then races the `updateModel()` write — there is no read-your-writes guarantee across pooled Mongo connections / replica reads — and can return the pre-write `{queries: []}` doc. If that stale read lands and it was the last ready-to-run query to finish before persist, the refresh chain dies and every dependent query stays `"queued"` forever.

A second, independent stall: `startReadyQueries()` `return`s (instead of `continue`s) when a `runAtEnd` query is iterated while non-runAtEnd queries are still pending. If Map iteration order puts the `runAtEnd` query before a *ready* non-runAtEnd queued query, the two deadlock.

## Fix

**`QueryRunner.ts`**
- New `private dagPersisted = false`, set `true` immediately after `updateModel()` in `startAnalysis()`.
- `onQueryFinish()` returns early while `!dagPersisted`. Early finishes are deferred; `this.timer` stays null; the post-persist `onQueryFinish()` is guaranteed to actually arm a timer whose `getLatestModel()` observes the committed DAG.
- `startReadyQueries()`: `runAtEnd` "still pending" branch now `continue`s instead of `return`ing.
- WARN message updated (the original cause is now prevented; the warn stays for visibility on any other path).

**`expireOldQueries.ts` — `reapStalledSnapshots()` (defense-in-depth)**
- When reaping an orphaned-DAG snapshot, also fail the underlying queued `Query` docs (they have no heartbeat so `getStaleQueries()` never touches them; without this they linger forever and can be picked up by the query cache as eternally-running).
- Pull `nextSnapshotAttempt` forward and re-enable `autoSnapshots` so the experiment is retried on the next 10-minute `queueExperimentUpdates` poll instead of waiting `EXPERIMENT_REFRESH_FREQUENCY` hours. An orphaned DAG is almost always a lost in-memory driver (process restart / event-loop kill), not a deterministic data problem, so a fresh run should succeed.

## Tests

- `gates onQueryFinish until the query DAG is persisted, then drives once` — early `onQueryFinish()` does not arm `this.timer`; post-persist call does.
- `processes ready non-runAtEnd queries even when a runAtEnd query is iterated first` — regression for the `return`→`continue` fix.
- Existing `startAnalysis` / `startReadyQueries` / heartbeat tests unchanged.

## Related-race audit (multi-pod deployments)

Surveyed for other state writes that assume the QueryRunner is the only writer / single-process:

| Area | Finding |
|---|---|
| `runCallbacks` (in-memory closures) | Architectural: the DAG can only be driven by the process that called `startAnalysis()`. Any process death strands queued queries. The reaper auto-retry in this PR is the mitigation; making the DAG resumable from another process would need `run`/`process` callbacks to be reconstructible from persisted state. |
| `incrementalRefresh` lock | Mongo-backed, fenced on snapshot id, heartbeated; reaper releases. Multi-pod safe. |
| Agenda job locks | Mongo `lockedAt`/`lockLifetime`; multi-worker safe. Default 10-min `lockLifetime` is shorter than long snapshot runs — with a single jobrunner replica that's harmless, but worth a `lockLifetime` bump on `updateSingleExperiment` if jobrunner scales out. |
| Query cache reuse of a `running` query | `createNewQueryFromCached` copies `status: "running"` and the 3s poll only triggers `onQueryFinish()` (a refresh), never updates the *copy's* doc. The copy's heartbeat goes stale and `getStaleQueries()` fails it within ~2 min — so it self-heals to `error`, but the snapshot is wasted. Separate from this PR; noting for follow-up. |
| `expireOldQueries` vs live runner | `errorSnapshotIfStillRunning` is guarded on `status: "running"`; a live runner that just wrote `success` wins. Multi-pod safe. |

### DIFF
--- a/packages/back-end/src/jobs/expireOldQueries.ts
+++ b/packages/back-end/src/jobs/expireOldQueries.ts
@@ -22,7 +22,12 @@ import {
 import {
   getQueryStatusesByIds,
   getStaleQueries,
+  markPendingQueriesAsFailed,
 } from "back-end/src/models/QueryModel";
+import {
+  getExperimentById,
+  updateExperiment,
+} from "back-end/src/models/ExperimentModel";
 import {
   findReportsByQueryId,
   updateReport,
@@ -251,7 +256,7 @@ async function reapStalledSnapshots() {
     });
 
     const error = orphanedDag
-      ? "Snapshot stalled: queries were never started. This can happen when the server restarts mid-refresh. Please try updating results again."
+      ? "Snapshot stalled: queries were never started. This can happen when the server restarts mid-refresh. A retry has been scheduled."
       : "Snapshot stalled: queries finished but results were never finalized. This usually means the analysis step failed (check server logs) or the process was restarted.";
 
     const context = await getContextForAgendaJobByOrgId(snapshot.organization);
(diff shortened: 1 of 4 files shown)


ahdriel:
@johnham-ant Thanks for the contribution!

I agree with both fixes and I this is a known fragile area that we are working on, so the goal is to clean some of this up soon as we build a stronger foundation. No ETA on this work as it requires some untangling, but just sharing FYI.

Good fix + retry logic for orphaned QueryRunner looks good.

FYI for the latter I made it more restrictive for Scheduled Updates only and not ones triggered via UI given they use the scheduled mechanism.

ahdriel (approved):

PR #6166, opened 2026-06-18
fix(aggregated-tables): only include metrics in active experiments

lukesonnet:
## Summary

Aggregated fact tables previously materialized **every** non-archived fact metric referencing a fact table. This PR scopes materialization to metrics that matter for analysis and keeps the materialized table stable, reducing warehouse work and table width without churning on restates.

A base fact metric is materialized when it is:

1. **not archived**, AND
2. **either** referenced by a running experiment (including running holdouts) **or** already present in that idType's pre-aggregated table.

Auto-slice metrics of an eligible base metric are retained.

If you restate for any reason, we slim down the set of eligible metrics to those that are active, ignoring ones previously materialized.

### Refresh / restate behavior ([aggregatedFactTables.ts](packages/back-end/src/services/aggregatedFactTables.ts))

- New helper `getRunningExperimentMetricIds(context)` collects the metric IDs referenced by `status: "running"` experiments and holdouts, expanding metric groups via the shared `getAllMetricIdsFromExperiment`.
- New helper `getMaterializedMetricIds(registry)` reads the metric IDs already persisted in a table's `metricState`.
- `getMetricsForAggregatedFactTable` / `getAggregatedFactTableMetrics` now take `activeMetricIds` and `materializedMetricIds` and apply the rule above.
- A table is refreshed only while at least one **active** (non-archived) metric references it. If nothing is active, the run is skipped (stops refreshing) even when inactive metrics are still materialized; runs resume cleanly once an experiment references the table again.
- **Incremental** runs materialize active ∪ already-materialized so existing columns keep getting fresh data and we avoid the full restate that re-adding a dropped metric would force.
- **Full restate** runs (forced, schema drift, incomplete write, or first run) rebuild from scratch, so they slim the table down to the currently active metrics. The mode is decided against the kept set so a newly added active metric still forces a restate while an inactive-but-materialized metric never reads as a drift-triggering removal.
- `materializedMetricIds`/`keptMetrics` are derived from the **lock-held** registry read so the desired set and the drift comparison share one consistent snapshot (no pre-lock stale read).

### Status endpoints stay consistent

The filter is applied at the nightly/background update job plus both the REST ([getAggregatedFactTables.ts](packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts)) and internal ([fact-table.controller.ts](packages/back-end/src/routers/fact-table/fact-table.controller.ts)) status endpoints, computed **per idType** from each registry doc's `metricState`, so the UI's `pendingRestate`/`metricState` matches what the update job will actually do.

To keep the computed metric state identical for project-scoped user requests and the org-wide nightly job, `getRunningExperimentMetricIds` uses a new org-wide, permission-bypassing query `getRunningExperimentsAcrossProjects` ([ExperimentModel.ts](packages/back-end/src/models/ExperimentModel.ts)).

### `getAllExperiments` refactor

To fetch standard, bandit, and holdout running experiments in a **single** query:

- Replaced the singular `type?: ExperimentType` filter with an optional `types?: ExperimentType[]`.
- Default behavior is unchanged: when `types` is omitted, holdouts are excluded (`{ \$ne: "holdout" }`). When provided, it builds a `\$in` filter, still mapping `"standard"` to `["standard", null]` for legacy docs.
- Updated existing callers: [holdout.controller.ts](packages/back-end/src/routers/holdout/holdout.controller.ts) (`types: ["holdout"]`) and [controllers/experiments.ts](packages/back-end/src/controllers/experiments.ts) (`types: type ? [type] : undefined`).

### Tests

Unit tests cover the eligibility rule (active vs. already-materialized vs. archived), auto-slice retention ([aggregatedFactTable.test.ts](packages/back-end/test/aggregatedFactTable.test.ts)), and the `getAllExperiments` `types` -> Mongo query mapping ([ExperimentModel.test.ts](packages/back-end/test/models/ExperimentModel.test.ts)).

## Test plan
- [X] \`pnpm --filter back-end type-check\` passes (verified)
- [X] \`pnpm --filter back-end test\` for the two suites passes (verified: 63 tests)
- [x] With a running experiment using a fact metric, trigger an update and confirm only eligible metrics (and their auto-slices) are materialized
- [x] Stop the experiment; confirm an incremental run keeps the now-inactive metric (column stays fresh, no restate)
- [x] Trigger a restate (e.g. add a new active metric) and confirm the table slims to active-only metrics
- [x] Remove all running experiments referencing a table and confirm the update is skipped (stops refreshing), then resumes after a new experiment starts
- [x] Confirm the REST and internal status endpoints report the same per-idType \`pendingRestate\`/\`metricState\` as the update job
- [x] Confirm running holdouts contribute their metrics
- [x] Regression: holdouts list (\`getHoldouts\`) and the experiments list endpoint still return expected results


packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -506,14 +506,15 @@ export async function getAllExperiments(
...
 
-  if (type === "multi-armed-bandit") {
-    query.type = "multi-armed-bandit";
-  } else if (type === "standard") {
-    query.type = { $in: ["standard", null] };
-  } else if (type === "holdout") {
-    query.type = "holdout";
-  } else if (!type) {
+  if (!types) {
+    // Default: exclude holdouts (unchanged behavior)
```
lukesonnet: (unchanged behavior) is such an AI comment come on man

lukesonnet:
@greptileai where are you?

packages/back-end/src/services/aggregatedFactTables.ts
```diff
@@ -120,29 +126,56 @@ export type AggregatedFactTableStatus = {
...
 };
 
+export async function getRunningExperimentMetricIds(
+  context: ReqContext | ApiReqContext,
+): Promise<Set<string>> {
+  const [running, metricGroups] = await Promise.all([
+    getAllExperiments(context, {
+      status: "running",
+      types: ["standard", "multi-armed-bandit", "holdout"],
+    }),
```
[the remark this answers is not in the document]
lukesonnet: nice catch. will update.

packages/back-end/src/services/aggregatedFactTables.ts
```diff
@@ -120,29 +126,56 @@ export type AggregatedFactTableStatus = {
...
+    context.models.metricGroups.getAll(),
+  ]);
+  const ids = new Set<string>();
+  for (const experiment of running) {
+    for (const id of getAllMetricIdsFromExperiment(
+      experiment,
+      true,
+      metricGroups,
```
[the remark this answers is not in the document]
lukesonnet: That's ok, I think I'm ok being a little more expansive with adding active metrics, even if it's only used currently as an activation metric. This PR is more a way to remove unused metrics than keep only very actively used metrics.

lukesonnet:
@cursor review

lukesonnet:
@cursor review

lukesonnet:
@cursoragent review

packages/back-end/src/services/aggregatedFactTables.ts
```diff
@@ -120,37 +127,118 @@ export type AggregatedFactTableStatus = {
...
+// Eligibility for the pre-aggregated table is org-wide: it depends on every
+// running experiment (across all projects) and the fact metrics they reference,
+// either of which can live in projects the caller can't read.
+export async function getAggregatedFactTableEligibilityInputs(
+  context: ReqContext | ApiReqContext,
+): Promise<{
+  factMetrics: FactMetricInterface[];
+  activeMetricIds: Set<string>;
+}> {
+  const backgroundContext = getContextForAgendaJobByOrgObject(context.org);
```
ahdriel: nit: this is not technically true anymore as we are getting the company context but not for agenda job only

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
```diff
@@ -24,18 +27,33 @@ export const getAggregatedFactTables = createApiRequestHandler(
   const byIdType = new Map(registryDocs.map((doc) => [doc.idType, doc]));
```
ahdriel: issue: the core logic here is duplicated between both APIs.

we should extract it to a service layer or something. The API boundary for internal or external API should be responsible for getting the params, calling the main logic that is shared, and then formatting the response as they need.

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
```diff
@@ -24,18 +27,33 @@ export const getAggregatedFactTables = createApiRequestHandler(
...
 
   // Recompute the schema state the nightly driver would so callers can see when
-  // the next run will be forced to restate. Read-only; no warehouse query.
-  const factMetrics = await req.context.models.factMetrics.getAll();
-  const metrics = getAggregatedFactTableMetrics({ factMetrics, factTable });
-  const { factTableSettingsHash, metricState } =
-    buildAggregatedFactTableSchemaState({ factTable, metrics });
+  // the next run will be forced to restate.
+  const { factMetrics, activeMetricIds } =
```
ahdriel: question: why do we return just factMetrics but we return all metricIds?

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
```diff
@@ -24,18 +27,33 @@ export const getAggregatedFactTables = createApiRequestHandler(
...
+    });
+    const { factTableSettingsHash, metricState } =
+      buildAggregatedFactTableSchemaState({ factTable, metrics });
     const status = buildAggregatedFactTableStatus({
       idType,
-      doc: byIdType.get(idType),
+      doc,
       factTableSettingsHash,
       metricState,
+      hasActiveMetrics,
```
ahdriel: question: does this overall change now mean that Updating results for experiments that are stopped for Incremental Pipeline, will query 14 days of data for CUPED? or how are we handling that?

That would be because in theory we can have a metric that is not in the CUPED agg table but is still in the experiment.

Similar thing regarding looking at past phases results.

packages/back-end/src/services/aggregatedFactTables.ts
```diff
@@ -120,37 +127,118 @@ export type AggregatedFactTableStatus = {
...
+}
+
+// Expects an org-wide context (e.g. the background admin context built in
+// getAggregatedFactTableEligibilityInputs) so metric group references on
+// running experiments expand fully, regardless of the original caller's scope.
+export async function getRunningExperimentMetricIds(
+  context: ReqContext | ApiReqContext,
+): Promise<Set<string>> {
+  const [running, metricGroups] = await Promise.all([
+    getAllExperiments(context, {
```
ahdriel: potential-issue: For a org with many running experiments, we are are loading all of them in memory but using just the metrics fields.

biggest concern would be around `phases`/`variations` but I am not convinced it would be a big perf problem.

additionally we could create better patterns to use mongodb projection to fetch the minimal amount of data required in these scenarios.

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
```diff
@@ -24,18 +27,33 @@ export const getAggregatedFactTables = createApiRequestHandler(
...
+      activeMetricIds,
+    }).length > 0;
 
   const aggregatedFactTables = idTypes.map((idType) => {
+    const doc = byIdType.get(idType);
+    const metrics = getAggregatedFactTableMetrics({
+      factMetrics,
+      factTable,
+      activeMetricIds,
+      materializedMetricIds: getMaterializedMetricIds(doc),
```
ahdriel: nit: a lot of this function (`getAggregatedFactTableMetrics`) is the same, except for some filtering based on the `doc` -- i am curious if there would be a way to avoid doing the same work repeatedly

with that said, I think the work is quite cheap and we don't have many idTypes per fact table either way, so it is not like this will be called 100x

packages/back-end/src/services/aggregatedFactTables.ts
```diff
@@ -120,37 +127,118 @@ export type AggregatedFactTableStatus = {
...
+      ids.add(id);
+    }
+  }
+  return ids;
+}
+
+// The set of metric IDs already materialized into a fact table's pre-aggregated
+// table, including slice metrics
+export function getMaterializedMetricIds(
+  registry?: Pick<AggregatedFactTableInterface, "metricState"> | null,
```
ahdriel: nit: should we just call this `aggregatedFactTable`?

ahdriel (approved):

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
lukesonnet: It should fall back because the metric won't be in the metric state for the aggregated table, so yes it will work that way, but only if the table was re-stated between the experiment stopping and now, because we have logic to keep updating metrics in the registry so long as nothing else about the hash has drifted.

packages/back-end/src/services/aggregatedFactTables.ts
lukesonnet: Renamed it to `getContextForOrgAdminByOrgObject`

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
lukesonnet: Extracted to services, thanks.

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
lukesonnet: It didn't seem like it mattered to me, but we can definitely narrow down the size of the set since we only use this for fact metrics. Will do.

packages/back-end/src/api/fact-tables/getAggregatedFactTables.ts
lukesonnet: I refactored this.

PR #6172, opened 2026-06-19
fix(sdk-js): resolve nested attribute keys as experiment hash attributes

### LINKED ISSUE #4432: [Bug] [sdk] Nested attribute keys cannot be used as experiment hash attributes
### Summary

The SDKs support nested attributes:

```ts
gb.updateAttributes({
  user: {
    id: "1234",
    email: "test@example.com"
  }
})
```

You can reference these nested attributes in targeting conditions using dot notation:

```json
{
  "user.email": { "$eq": "test@example.com" }
}
```

But, there is no way to use a nested field as the hash attribute.  `user.id` will not resolve properly.

#### Expected Behavior

Specifying a hash attribute using dot notation (e.g. `user.id`) will properly resolve to a nested attribute.

ryanda9910:
## What

Lets a nested attribute be used as the experiment hash attribute via dot notation, e.g. `hashAttribute: "user.id"` against `{ user: { id: "..." } }`.

Closes #4432

## Why

The SDK already supports nested attributes and you can target them with dot notation in conditions:

```json
{ "user.email": { "$eq": "test@example.com" } }
```

But `getHashAttribute` in `packages/sdk-js/src/core.ts` did a flat `attributes[hashAttribute]` lookup, so a dotted key like `user.id` never resolved when the attributes were nested. The value fell through to the empty-hash path and the experiment was skipped.

## How

The repo already has the right primitive: `getPath()` in `packages/sdk-js/src/mongrule.ts` resolves dot-separated paths, and targeting conditions use it. Rather than write a second resolver, I exported `getPath` and reused it.

`getHashAttribute` now resolves the hash attribute (and the fallback attribute) with a small `resolveAttribute` helper that:

1. tries the literal key first, so an attribute key that literally contains a dot keeps working (backward compatible),
2. then falls back to dot notation to resolve a nested attribute,
3. still yields an empty hash value when the attribute is missing (unchanged behavior, so missing-attribute experiments are still skipped).

No new dependency, and the zero-dependency SDK boundary is respected.

## How tested

Added regression tests in `packages/sdk-js/test/experiments.test.ts`:

- a nested attribute (`user.id`) resolves and produces the same assignment as the equivalent flat `"user.id"` key,
- a literal dotted key still takes precedence over the nested path,
- a missing nested hash attribute still skips the experiment.

The nested-resolution test fails on `main` and passes with this change. The existing fallback-attribute coverage in `sticky-buckets.test.ts` exercises the same `resolveAttribute` path and still passes.

```
Test Suites: 15 passed, 15 total
Tests:       679 passed, 679 total
```

Also ran type-check (`tsgo`), ESLint, and Prettier on the changed files, all clean. Verified on Node v24.16.0 (repo requires Node >=24).

### DIFF
--- a/packages/sdk-js/src/core.ts
+++ b/packages/sdk-js/src/core.ts
@@ -941,7 +941,10 @@ function resolveAttribute(
   attributes: Record<string, any>,
   key: string,
 ) {
-  if (attributes[key]) return attributes[key];
+  // Prefer a literal key (so a key that itself contains a dot still works), checking presence
+  // rather than truthiness so a literal falsy value is honored and not overridden by a nested path.
+  if (Object.prototype.hasOwnProperty.call(attributes, key))
+    return attributes[key];
   return getPath(attributes, key) ?? "";
 }
 
--- a/packages/sdk-js/src/core.ts
+++ b/packages/sdk-js/src/core.ts
@@ -16,7 +16,7 @@ import {
   Options,
   ClientOptions,
 } from "./types/growthbook";
-import { evalCondition } from "./mongrule";
+import { evalCondition, getPath } from "./mongrule";
 import { ConditionInterface } from "./types/mongrule";
 import {
   chooseVariation,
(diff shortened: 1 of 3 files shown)


ryanda9910:
Good catch on the falsy edge case. I switched resolveAttribute to check key presence with hasOwnProperty instead of truthiness, so a literal key that is present but falsy (like 0) is honored as a no-value rather than falling through to a same-named nested path. Added a regression test for it. Full sdk-js suite passes (680 tests) on Node 24.

PR #6174, opened 2026-06-19
Sparse mode for JSON flag rules

bryce-fitzsimons:
### What
Adds a **sparse patch** option to JSON feature rules. When a rule is marked sparse,
its value is treated as a *partial* object merged onto the feature's default value at
serve time, instead of replacing it — only the keys you specify override the default,
everything else falls through.

This makes large JSON config flags much easier to maintain: a rule can change a single
field without re-stating (and risking clobbering) the entire object.

Also adds an organization setting (Features) that defaults to editing in sparse mode when available.

### Where you can use it
Available wherever you edit JSON rule values, on JSON flags whose default is a plain object:
- Force and rollout rules
- Experiment and bandit rules — both referencing an existing experiment and creating a new one
- Ramp schedule steps
- Editing linked-feature values from the experiment page
- The REST API (feature create/update and revision rule add/update)

### How it reads
- Value editors gain an **Edit / Preview** tab pair; the preview shows the fully merged
  value with the patched keys in **bold**.
- Rule summaries and the experiment page's linked-feature view show the merged result,
  with patched keys emphasized so it's clear what each rule actually changes.
- Merging happens when the SDK payload is built, so SDKs receive the already-merged
  value — no SDK changes required.

### Notes / out of scope
- Applies only to JSON flags whose default is a plain key/value object (not arrays,
  primitives, or null).
- Safe-rollout rules and holdout values are not included.
- When a flag uses a validated JSON *schema*, the schema editor takes precedence over
  sparse mode (unchanged behavior).

image

image

image


gazzdingo (approved):
LGTM

PR #6175, opened 2026-06-19
Migrate legacy managed warehouses to native JSON columns

Kevin-Chant:
### Features and Changes

Migrates legacy (materialized-column) managed warehouses to native ClickHouse JSON columns. The JSON foundation is already merged — #5994 (new warehouses default to JSON) and central-license-server #139 (license-server JSON code path) — so this PR is the cutover for existing warehouses. Managed warehouses are **cloud-only**, so the goal is to drive every legacy warehouse to JSON and then retire the materialized-column code paths.

- **Two-pronged rollout, cloud-only.** A proactive background **sweep** (`sweepManagedWarehouseMigrations`, gated by `MANAGED_WAREHOUSE_MIGRATION_SWEEP_ENABLED`, default off) drains legacy warehouses in small batches; a **lazy on-read trigger** in `ClickHouse.runQuery` is the safety net for anything the sweep hasn't reached. Both enqueue the same deduped, idempotent per-org job (`migrateManagedWarehouse`, concurrency-capped). The sweep self-quiesces when none remain; its count of not-fully-migrated warehouses (legacy, partially-migrated, or stuck mid-migration) is the **cutover gate** — once it holds at 0, the legacy paths can be removed in a follow-up.

- **Migration mechanic** (`migrateManagedWarehouseToJson`). Enter a transient `migrating` state → `dangerousRecreateClickhouseTables` (LS recreates the per-org `events`/`experiment_views` tables as JSON and repopulates from `enriched_events`) → regenerate the `ch_events` fact table + datasource `userIdTypes`/exposure queries → clear `materializedColumns` → leave `migrating`. No license-server changes: it already branches on `settings.useJsonColumns`.

- **Backward compatibility via column aliasing — no stored SQL is rewritten.** After migration the former materialized columns no longer exist as physical columns (their data lives in the `attributes` JSON column). Rather than textually rewrite any stored SQL, every former column — **identifiers and dimensions alike** — is re-exposed as a top-level `SELECT` alias out of the JSON (`attributes.<sourceField>::<cast> AS <columnName>`; numeric coerces via `toFloat64OrNull`, everything else `Nullable(String)`). So a bare reference to a former column keeps resolving anywhere that routes through the fact table, the exposure queries, or the per-org tables — with zero user intervention. Preserved columns are persisted as `migratedIdentifiers` / `migratedColumns` so the attribute-change sync re-emits the aliases.

- **Identifier parity — nothing blocks the migration.** Legacy `userIdType`s that aren't current `hashAttribute`s (e.g. a custom `company_id`) are preserved as aliased top-level join columns and kept as `userIdType`s, so the join keys experiments/metrics depend on survive. There is no skip/tombstone path; every warehouse migrates.

- **Transient `migrating` state.** A provisioned warehouse mid-recreate sets `settings.migrating` (distinct from never-provisioned). Query surfaces gate on a new `isManagedWarehouseUnavailable` (awaiting-provisioning **or** migrating) helper and the UI shows an honest "being upgraded" message. The settle logic is idempotent and resumable: a crash mid-migration leaves a re-runnable state that the next sweep tick or query recovers.

Note: a migrated dimension that is a live attribute is exposed **both** as a top-level alias column and as an `attributes.<field>` JSON field. This duplicate listing is data-driven (the value genuinely exists in the JSON column, so the column-refresh job surfaces it regardless) and harmless — both resolve to the same data. It is not a SQL conflict; `SELECT *` no longer emits the column physically.

### SQL surface compatibility

**Continues working automatically (no user action):**

- Fact-metric `numerator`/`denominator` column refs, aggregate-filter columns, and row-filter columns
- Raw-SQL fact filters (`factTable.filters[].value`) and `sql_expr` row filters
- Custom exposure breakdown dimensions — both the data and the breakdown listing
- Auto-slice metrics and fact-table column auto-slices
- Built-in and custom identifiers (including preserved non-`hashAttribute` ones)
- Product Analytics explorations — **all** dataset types (`metric` / `fact_table` route through the fact-table SQL; `data_source` gets the same aliases injected into its per-org-table stub)

:warning: **At risk — user-authored raw SQL that queries the physical tables directly.** These bypass the fact table, so no alias is in scope; a reference to a former column fails loudly (`Unknown identifier`) and is fixed by editing the SQL to use `attributes.<field>`:

- Saved SQL queries (SQL explorer)
- SQL segments (`type: "SQL"`)
- SQL dimensions (custom-SQL dimensions)

Production impact (cloud audit): **4 orgs, 7 queries total** across these surfaces (assuming the aliases above resolve as verified). Per-surface breakdown:

| At-risk surface | Orgs affected | Notes |
| --- | --- | --- |
| Saved SQL queries | 2 | Bare references to dropped identifier/dimension columns; one of the orgs already uses `attributes.<field>` in most of its queries, so only a subset is affected |
| SQL segments | 0 | No `type: "SQL"` segments reference a dropped column |
| SQL dimensions | 2 | Bare references to dropped custom dimension columns |

Counts are exact — each candidate was filtered to SQL that actually reads the per-org `events`/`experiment_views` tables, then every match was inspected by hand (matches inside `attributes.<field>` / `JSONExtract(attributes, ...)`, string literals, and CTE aliases were excluded as false positives).

Because the affected set is so small (7 documents across 4 orgs), we'll hand-fix these stored queries directly — rewriting the bare column references to `attributes.<field>` — coordinated with the merge, so affected users shouldn't notice the break.

We will also break a number of queries on our own dogfooding org, which is excluded from the counts above — that's an acceptable risk for our own use.

### Dependencies

- #5994 (merged) — new warehouses default to JSON columns
- central-license-server #139 (merged) — license-server JSON code path

### Testing

**Unit tests:**
```
pnpm --filter shared test test/util/managedWarehouse.test.ts        # predicates, alias clauses, fact-table/exposure builders
pnpm --filter back-end test test/util/migrateManagedWarehouseColumns.test.ts  # migrated-dimension filter + settle-state logic
```

**End-to-end (local managed warehouse + license server):** verified against a live dev warehouse that a migrated warehouse's `ch_events` SQL aliases the preserved identifier + string/numeric dimensions, that they become top-level fact-table columns and exposure breakdowns, and that a real query referencing the former columns (`SELECT plan, revenue, company_id FROM (<ch_events sql>)`) **resolves with no "Unknown identifier."**

**Manual migration flow** (with `CRON_ENABLED` + license server): revert a dev warehouse to a legacy state in `mongosh`:
```js
db.datasources.updateOne(
  { type: "growthbook_clickhouse" },
  { $set: {
      "settings.useJsonColumns": false,
      "settings.materializedColumns": [
        { columnName: "device_id", sourceField: "device_id", datatype: "string", type: "identifier" },
        { columnName: "company_id", sourceField: "company_id", datatype: "string", type: "identifier" },
        { columnName: "plan",       sourceField: "plan",       datatype: "string", type: "dimension" }
      ]
  } }
)
```
Then either set `MANAGED_WAREHOUSE_MIGRATION_SWEEP_ENABLED=true` and restart (sweep enqueues within a minute), or run any query against the warehouse (lazy trigger). Once the job runs, verify:
```js
db.datasources.findOne(
  { type: "growthbook_clickhouse" },
  { "settings.useJsonColumns": 1, "settings.materializedColumns": 1, "settings.migratedIdentifiers": 1, "settings.migratedColumns": 1, "settings.userIdTypes": 1 }
)
// => useJsonColumns: true, materializedColumns: absent, migratedIdentifiers: [company_id],
//    migratedColumns: [plan], userIdTypes includes company_id
```
The `ch_events` fact table SQL should now alias `attributes.company_id AS company_id` and `attributes.plan AS plan`, and a query referencing bare `plan` / `company_id` should resolve.


Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

Kevin-Chant:
@cursoragent review

Kevin-Chant:
@cursoragent review

tzjames (approved):

PR #6176, opened 2026-06-19
Switch the runtime image to a distroless Docker Hardened Image

Kevin-Chant:
### Features and Changes

Swaps the shipped runtime base from `node:24-slim` to the distroless DHI `node:24-debian12` (no shell, no package manager, non-root uid 1000). This structurally removes the recurring class of OS-library CVEs (`gnutls`/`gcrypt`/`perl`/`expat`/`pam`/`util-linux`/`curl`) that a frozen `apt-get install`-only base leaves open until someone hand-bumps them — the packages aren't shipped at all, and DHI's continuous rebuilds keep what remains patched. Build stages stay on `node:24-slim`/`debian:12-slim` (they don't ship, so their CVEs don't reach prod).

:warning: **The chart and image are meant to upgrade together — but the image no longer hard-crashes under a lagging launch config.** The new chart adds the non-root + read-only security context and the writable mounts the distroless image needs. Critically, the image stays **backward-compatible with the previous launch command** (`node_modules/.bin/pm2-runtime …`): that path was a `#!/bin/sh` shim that can't exec in a shell-less runtime, so the image repoints it at pm2's `#!/usr/bin/env node` entry. This matters because the same `growthbook/growthbook:latest` tag is consumed by paths that aren't upgraded in lockstep with the chart — a self-hoster pinning a floating `image.tag`, or an ECS task def that hardcodes the old command — and those would otherwise CrashLoop the moment this image lands. With the shim repointed they **boot and serve**; they only degrade where the old config lacks `fsGroup`/writable mounts (e.g. local uploads fail until the chart is upgraded). A normal `helm upgrade` pulls the matching image and applies the full hardening automatically. Detailed in the release note.

- **Process manager stays pm2, invoked via `node`.** The image still runs `pm2-runtime` from `ecosystem.config.js`, but the CMD and chart now call `node node_modules/pm2/bin/pm2-runtime …` directly — the `node_modules/.bin/pm2-runtime` shim is a `#!/bin/sh` script and can't exec in a shell-less runtime, while pm2's real entry is plain Node. That `.bin` shim is *also* repointed (symlinked) at pm2's real Node entry rather than left dead, so the previous `node_modules/.bin/pm2-runtime …` launch command still boots on this image (see the coupling note above) — `/usr/bin/env` is copied in to satisfy its `#!/usr/bin/env node` shebang. pm2's only shell-out (pidusage's `getconf` for CPU metrics) falls back to a default on failure, so it's a single non-fatal log line at startup. `PM2_HOME=/tmp/.pm2` lets pm2 write under a read-only rootfs.
- **Polyglot assembly by COPY.** The runtime runs Node *and* spawns Python `gbstats`. The hardened Python interpreter + venv are built against the DHI python `-dev` image and copied in, along with the system shared-lib closure they dynamically link (`/opt/pydeps`) and the Kerberos GSSAPI closure the `kerberos` addon `dlopen`s at runtime (`/opt/krb5deps`). `kerberos` is force-built from source (the prebuilt's runtime libc detection fails on distroless).
- **Helm chart (`charts/`).** Both pods' `command` → `[/usr/local/bin/node, node_modules/pm2/bin/pm2-runtime, start, ecosystem.config.js, --only, <app>]`; add `securityContext: {readOnlyRootFilesystem: true, runAsNonRoot: true, runAsUser: 1000}` plus `podSecurityContext: {fsGroup: 1000}` (so the non-root user can write the uploads PVC and emptyDirs — without it a freshly provisioned PVC is root-owned and local uploads fail at runtime); and add `emptyDir` mounts for the runtime's writable paths (`/tmp`, front-end `.next/cache`). Back-end uploads stay on the existing PVC. Probes were already `httpGet`. Note: on upgrades with a large pre-existing local-uploads PVC, `fsGroup` triggers a recursive ownership change at mount time, which can lengthen the first pod's `ContainerCreating` phase. This happens *before* the container starts, so liveness/readiness (and any downstream LB health check) don't begin until it's done — it slows the rollout but doesn't cause a pod to be killed as unhealthy. Deployments on `s3`/`google-cloud` uploads are unaffected.
- **Multi-arch.** `deploy` builds `linux/amd64,linux/arm64`. The krb5 lib path is derived from `dpkg` (not a hardcoded `x86_64-linux-gnu`) and `LD_LIBRARY_PATH` lists both multiarch dirs (the loader skips the absent one), so both architectures build and run.
- **Preview pipeline.** `preview/idle-monitor.sh` → `idle-monitor.js` (shell-free: reads `/proc/net/tcp{,6}` instead of `ss`, and on idle signals `pm2-runtime` via `ppid` instead of `pkill`). pm2 launches it (`ecosystem.config.js`) as a fork-mode app only when `PREVIEW_IDLE_TIMEOUT_SECONDS` is set; `preview/Dockerfile.append` drops its `apt-get`/`chmod` (no shell/apt in distroless) down to a single `COPY`.
- **Encryption-key migration.** The distroless image has no `pnpm`/`yarn`/bash wrapper, so the self-host `migrate-encryption-key` instructions now invoke `node dist/scripts/migrate-encryption-key.js` directly. The cron/stats-pool toggles the one-off run needs are set inside the script (a preamble module imported first, before `secrets.ts`/`python.ts` read them at load time) rather than passed on the command line — which also fixes a previously ineffective `CRON_ENABLED=false` flag (the code reads `CRON_DISABLED`).

Note: the OS slice is what this change moves. npm `node_modules` (14) and Node-bundled OpenSSL (15) findings are unaffected and tracked separately. GSSAPI auth against a real KDC is the one path not exercised locally (needs MongoDB Enterprise + a KDC) — the addon loads and `initializeClient(...)` returns OK, and SCRAM works.

### Testing

Built locally from this branch (`docker build -t gb:dhi .`) and exercised against a local Mongo. The k8s-vs-Docker distinction at the container layer is cosmetic here: `docker run --read-only` enforces the same kernel read-only mount as `readOnlyRootFilesystem: true`, and `docker kill --signal` grace matches pod termination. The remaining authoritative check is the first deploy — worth watching pod-restart rate + logs for 24–48h, where a volume-mount path typo would surface (not at boot).

**1. Boot smoke (all-in-one, under `--read-only`)** — pm2-runtime launches both apps online; back-end `/healthcheck` → `HTTP 200 {healthy:true}`, front-end `/` → `HTTP 200` (`✓ Ready`), Mongo SCRAM connected. `gbstats`/`scipy`/`numpy`/`pandas` import OK, and `docker top` showed `python3 stats_server.py` spawned as a live back-end child (the Node→Python stdio path runs, not just an importable venv).

**2. Read-only rootfs proxy for k8s**
```
docker run --read-only --tmpfs /tmp \
  --tmpfs /usr/local/src/app/packages/front-end/.next/cache \
  --tmpfs /usr/local/src/app/packages/back-end/uploads ...
```
Both endpoints `HTTP 200`, no EROFS errors (pm2 writes its pid/socket/logs to `PM2_HOME=/tmp/.pm2`). Note `docker --tmpfs` is mode `1777`, so it does **not** model k8s volume ownership under a non-root user — that's why the chart sets `fsGroup: 1000`. I confirmed the failure mode directly: a fresh root-owned volume (how a PVC/emptyDir provisions without `fsGroup`) → uid-1000 write fails `EACCES`; group-writable (what `fsGroup` produces) → OK. Full `fsGroup` validation is the first-deploy step on a real cluster.

**3. Per-pod `--only` mode** (mirrors the chart pods)
- `… --only back-end` → only back-end starts, `/healthcheck` `200`
- `… --only front-end` → only front-end starts, `/` `200`

**4. pm2 in distroless + shutdown** — pm2-runtime runs cleanly when launched via `node` (the `.bin` shim is shell-only). Its single shell-out — pidusage's `getconf CLK_TCK`/`PAGESIZE` — logs one error at startup and falls back to a default (verified in the source and at runtime; not on any request path). Graceful `SIGTERM` stops all apps and exits **0**. Crash handling matches the stock image (autorestart off; a dead app is restarted by the k8s liveness probe).

**5. Idle-monitor** (preview variant, `cat Dockerfile preview/Dockerfile.append`)
```
PREVIEW_IDLE_TIMEOUT_SECONDS=15 -> [idle-monitor] Connections: 0, Idle: 15s
                                -> Shutting down -> container exit 0
```
pm2 launched all three apps (gated on the env var); at the threshold the monitor signaled `pm2-runtime` (its parent), which stopped the apps and exited **0** (so a Fly machine won't auto-restart — matching the old bash behavior).

**6. Inspector OS forecast** (authoritative `aws inspector-scan scan-sbom`, the data Vanta integrates — not `docker scout`)

OS + Python slice (115 components):
```
vulnerability_count: { critical: 0, high: 0, medium: 0, low: 0 }
```
OS slice is 16 deb packages; all 22-finding offenders (`perl`/`expat`/`python3.11`/`pam`/`util-linux`/`curl`/`gnutls`/`gcrypt`) are absent. The candidate-vulnerable packages that remain (`glibc`, `gcc-12`, `openssl`, `libssl3`) each scanned to no known vulnerabilities. **OS findings: 22 → 0.**

**7. Tracing + kerberos** — booted with `TRACING_PROVIDER=datadog`: pm2 starts the back-end with `--require ./packages/back-end/dist/tracing.datadog.js` (from `ecosystem.config.js`'s `node_args`), logs carry `"dd":{"service":"growthbook-app"}`, health `200`. The `kerberos` addon loads and `initializeClient(...)` returns a client handle (native `gss_*` calls execute) under the dual-multiarch `LD_LIBRARY_PATH` — confirming `libcrypto` still resolves after the arch-portability change.

**8. Experiment results refresh (end-to-end Python/gbstats)** — ran the built image against a Mongo with real data, logged in, and triggered an experiment results refresh. Results updated successfully — exercising the full Node→Python stdio path (`python.ts` → `gbstats` venv) on the hardened image, not just an import smoke test.

**9. Encryption-key migration** — the documented distroless command runs end-to-end:
```
docker compose run --workdir /usr/local/src/app/packages/back-end \
  growthbook node dist/scripts/migrate-encryption-key.js OLD_KEY
```
(`Done!`, exit 0 against a test DB.) The cron/stats-pool toggles the run needs are now set by a preamble module inside the script rather than on the command line; confirmed against the tsc-emitted `dist` that its `require` runs before `secrets.ts`/`python.ts`, so both are set before load-time reads.

**10. Multi-arch image build (amd64 + arm64)** — local testing above ran on amd64. The production build matrix (`deploy.yml`) builds `linux/amd64,linux/arm64`, but that only runs on push to `main`, so the arm64 leg is validated ahead of merge via a `deploy-branch.yml` dispatch on this branch: [run 27929931166](https://github.com/growthbook/growthbook/actions/runs/27929931166). It builds both arches on the same Depot builder production uses and pushes a throwaway `branch-*` tag (nothing in prod consumes it). This exercises the arch-portable pieces on arm64 — the distroless final stage, the `dpkg`-derived krb5 lib path, and the dual-multiarch `LD_LIBRARY_PATH` — and confirms the build-time entrypoint asserts (`python3`/`ddtrace-run` in the python stage, `pm2-runtime` in the node stage) pass on both architectures.

**11. Backward-compat with the previous launch command** — to confirm an un-upgraded deployment (an older Helm chart, or any task def that pins the previous command) survives this image rather than CrashLooping, ran the built image with the *old* invocation — `--entrypoint node_modules/.bin/pm2-runtime … start ecosystem.config.js --only back-end` on a **writable** rootfs (the old chart set no `readOnlyRootFilesystem`). Result: pm2 launched the back-end (`App [back-end:0] online`), `/healthcheck` → `200`. Ran the same old command all-in-one (both apps) and exercised a **real login flow**, not just the healthcheck: the front-end `/` login page renders `200` (GrowthBook app shell + `__NEXT_DATA__`, no error markers), its login-UI JS chunk serves `200` from the container, `/api/init` returns real config, and a `POST /auth/register` against a fresh Mongo returns a valid **JWT** — so front-end serving + the back-end auth controller + password hashing + the Mongo write + JWT signing all work on the distroless image through the old launch path. The `.bin/pm2-runtime` shim now symlinks to pm2's `#!/usr/bin/env node` entry (`.bin/pm2-runtime -> ../pm2/bin/pm2-runtime`), resolved via the copied-in `/usr/bin/env`, so the prior command boots instead of dying on the missing `/bin/sh`. The pidusage `getconf` shell-out does log `spawn /bin/sh ENOENT` once — non-fatal, *after* the back-end is already online (matches test #4). Note: a front-end pod under an old chart still can't write its on-disk `.next/cache` (the image's `.next` dir is root-owned `0755` and the old chart adds no writable mount / `fsGroup`) — Next.js degrades gracefully to uncached serving, so it boots and serves; the new chart's emptyDir + `fsGroup` restores caching. Local uploads similarly degrade under an old chart unless `UPLOAD_METHOD` is `s3`/`google-cloud`.

**12. Datadog Python profiling (`GB_ENABLE_PYTHON_DD_PROFILING`)** — this exposed flag switches the gbstats spawn from `python3` to `ddtrace-run python3`, so `ddtrace-run` (a venv console-script) must work in the shell-less runtime. Confirmed both ways: directly, `ddtrace-run python3 -c "import ddtrace, gbstats, numpy, scipy, pandas"` bootstraps ddtrace `4.3.2` and imports the full stack; and in the real code path, booting the back-end with the flag set spawns a live worker — process tree `pm2-runtime → node …/server.js → /opt/venv/bin/python3 -u …/stats_server.py` (the `ddtrace-run` wrapper having exec'd into the interpreter), `/healthcheck` `200`. A failed `ddtrace-run` lookup would have produced no worker.

**13. Non-root local uploads (`UPLOAD_METHOD=local`)** — the runtime now runs as uid 1000 (was root), so a local-uploads volume must be writable by 1000. The image ships `packages/back-end/uploads` owned `1000:1000`, so a **freshly provisioned** docker named volume seeds writable — verified a uid-1000 write to a fresh volume succeeds. A volume **pre-seeded root-owned** (modeled by seeding it from a root base image, i.e. what the previous root image left behind) stays root-owned and a uid-1000 write correctly returns `EACCES` until it's `chown`ed to `1000:1000` — matching the release-note guidance (chown / `user: "0"` / S3-GCS / pin the prior image). k8s/chart users are unaffected (the chart's `fsGroup: 1000` handles the PVC).

**14. Custom Hooks sandbox (`isolated-vm`) on the DHI base** — Custom Hooks (self-hosted Enterprise) run user code in an `isolated-vm` V8 isolate inside a forked worker. The addon is a native C++ binary linking `libstdc++` and, unlike `kerberos`, gets no special Dockerfile handling — so it's a likely-to-break-*silently* surface: a worker that can't load it trips the sandbox pool's crash-loop breaker and hooks fail with "sandbox temporarily unavailable" while the server stays up. Verified against the real `dhi.io/node:24-debian12` base: built `isolated-vm@6.0.2` in the `node:24-slim` build stage (both the source-compiled and the prebuilt delivery paths), then in the DHI base ran an isolate end-to-end — addon loads, isolate executes, host `Reference`/`ExternalCopy` round-trip a result — including the real architecture (`fork()` → worker loads `isolated-vm` → runs isolate → IPC result). The base ships `libstdc++.so.6` (Node itself links it) with a `GLIBCXX` new enough for the addon (source build needs ≤ `3.4.30`, prebuilt ≤ `3.4.22`). `isolated-vm` is a regular `dependency` (survives `--no-optional`) and resolves its addon by fixed path, so it needs none of the from-source handling `kerberos` does.

### Release Note

_Proposed as the first **Release Highlight** (kept to one entry; slightly longer than usual because it flags a breaking change)._

**:warning: Hardened distroless runtime image (breaking — action may be required)** — The published image now uses a distroless [Docker Hardened Image](https://www.docker.com/products/hardened-images) base: no shell or package manager, **runs as non-root (uid 1000)** instead of root, and drops a recurring class of OS-package CVEs. Most deployments need no changes. **Helm/Kubernetes:** upgrade the chart and image together so the new launch command and non-root/read-only security context land in lockstep — a lagging chart or a pinned floating `image.tag` no longer CrashLoops, but local file uploads degrade until you upgrade (S3/GCS unaffected). **Direct image / Compose (`:latest` or pinned hashes):** the app still boots, but because it now runs as uid 1000 an **existing** local-uploads volume (previously root-owned) must be `chown`ed to `1000:1000` — or set `user: "0"`, switch to S3/GCS, or pin `image` to the prior version; **freshly created volumes work out of the box.** ([exact commands](https://docs.growthbook.io/self-host#hardened-non-root-shell-less-image)) **Note:** there's no shell in the container, so `docker exec … sh` no longer works (debug by attaching a namespace-sharing container — [docs](https://docs.growthbook.io/self-host#debugging-without-a-shell)) and `migrate-encryption-key` now runs via `node` directly (see the [self-hosting docs](https://docs.growthbook.io/self-host/env)).


Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

Kevin-Chant:
cursor review

docs/docs/self-host/production.mdx
```diff
@@ -27,7 +27,9 @@ If you change the `ENCRYPTION_KEY`, you will need to migrate any existing data s
 
 ```bash
 # If you didn't have an ENCRYPTION_KEY before, leave OLD_KEY blank below
-docker compose run growthbook pnpm migrate-encryption-key OLD_KEY
+docker compose run --workdir /usr/local/src/app/packages/back-end \
+  -e GB_STATS_ENGINE_MIN_POOL_SIZE=0 -e CRON_ENABLED=false \
```
tzjames: I think we can just set these env vars at the top of the script itself, to make the command a little cleaner.

docs/docs/self-host/env.mdx
```diff
@@ -83,7 +83,9 @@ If you change the `ENCRYPTION_KEY`, you will need to migrate any existing data s
 
 ```bash
 # If you didn't have an ENCRYPTION_KEY before, leave OLD_KEY blank below
-docker compose run growthbook yarn migrate-encryption-key OLD_KEY
+docker compose run --workdir /usr/local/src/app/packages/back-end \
+  -e GB_STATS_ENGINE_MIN_POOL_SIZE=0 -e CRON_ENABLED=false \
```
tzjames: I think we can just set these env vars at the top of the script itself, to make the command a little cleaner.

tzjames (approved):
Seems to handle all the edge cases I can think of.

PR #6180, opened 2026-06-19
fix(sql): stop misflagging backslash string literals as multi-statement queries

ahdriel:
### Features and Changes

Before sending any query to a warehouse, GrowthBook checks it for multiple statements and blocks anything that looks like more than one. That check assumed every warehouse treats a backslash inside a string literal the same way. On warehouses where a backslash is an ordinary character (Trino/Presto, Athena, Redshift, Postgres, SQL Server, Vertica), a valid single-statement query containing a backslash literal such as `ESCAPE '\'` was misread and wrongly rejected with "Multi-statement queries are not supported." It showed up most often on experiment queries that break down by a dimension, because those include a generated comment that contains a semicolon.

The check now interprets backslashes using each warehouse's own string rules instead of one shared assumption, so these queries run again. The guard still blocks genuine multi-statement input, and on these warehouses it is now more accurate, because it no longer lets a second statement hide behind a backslash.

### Testing

Unit tests cover multi-statement detection across each dialect's string-escaping rules, including the reported `ESCAPE '\'` case and real multi-statement input that must stay blocked.

### DIFF
--- a/packages/back-end/src/integrations/SqlIntegration.ts
+++ b/packages/back-end/src/integrations/SqlIntegration.ts
@@ -21,6 +21,7 @@ import {
   format,
   isMultiStatementSQL,
   SQL_ROW_LIMIT,
+  usesBackslashStringEscapes,
 } from "shared/sql";
 import { TemplateVariables, SqlDialect } from "shared/types/sql";
 import {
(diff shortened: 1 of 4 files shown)


ahdriel:
cursor review

jdorn (approved):

PR #6184, opened 2026-06-20
feat(nps): in-app NPS survey with internal Slack forwarding

gazzdingo:
## What

Adds a  flag-gated in-app NPS survey. A user picks a 0–10 score and an optional comment (both are optional but you can submit without the comment). responses are stored in the DB and also sent to slack

## Flags

- `nps-survey` — the rollout. Normal show path: flag on + is is the slack hook set in env + not in 90-day cooldown + x days after signup + % of users can see it that day (so we are not blasting all on one day)
## Slack forwarding

- `POST /user/nps-response` forwards to Slack **only when `NPS_SLACK_WEBHOOK` is set** — a private webhook only GrowthBook Cloud configures. Self-hosted never fires it; the post is server-side so Cloud users never see it.
- Fire-and-forget: a Slack failure never affects the user's request.
- Message is a colored attachment (green/yellow/red by score) with the comment and submitter email. Feedback is escaped (`& < >`) so it can't trigger Slack mentions, score is range-validated (0–10), and feedback is capped at 1500 chars.

## Manual setup

- Create the `nps-survey` flags in GrowthBook
- Set `NPS_SLACK_WEBHOOK` on the Cloud back-end (incoming webhook bound to the target channel).

## Test plan

- `?show-nps` with `nps-survey-preview` on → survey appears immediately; submit a score → message lands in the Slack channel.
- Score then close (no submit) → still logged/forwarded; close without a score → silent.
- Unset the webhook → no Slack post, survey otherwise unaffected.
- `pnpm type-check` + `pnpm lint` pass.
image

image


packages/back-end/src/util/secrets.ts
```diff
@@ -139,6 +139,12 @@ export const SITE_MANAGER_EMAIL = process.env.SITE_MANAGER_EMAIL;
 
 export const SLACK_SIGNING_SECRET = process.env.SLACK_SIGNING_SECRET || "";
 
+// Internal-only: incoming webhook URL for forwarding NPS survey responses to
+// GrowthBook's own Slack. Only set on GrowthBook Cloud — self-hosted and Cloud
+// users never see this, and nothing is sent unless it's configured.
+export const INTERNAL_NPS_SLACK_WEBHOOK =
```
gazzdingo: call this  NPS_SLACK_WEBHOOK

packages/back-end/src/models/UserModel.ts
```diff
@@ -27,6 +27,8 @@ const userSchema = new mongoose.Schema({
   superAdmin: Boolean,
   verified: Boolean,
   agreedToTerms: Boolean,
+  npsSurveyStatus: String,
+  npsSurveyAt: Date,
```
ahdriel, line 31: question: should we have this in a different collection?

packages/back-end/src/routers/users/users.controller.ts
```diff
@@ -136,6 +141,145 @@ export async function putUserName(
...
+    // Gated on a private webhook env var that only GrowthBook Cloud sets, so
+    // self-hosted and Cloud users never trigger or see this. Fire-and-forget —
+    // a Slack failure must never affect the user's request.
+    if (
+      status === "responded" &&
+      typeof score === "number" &&
+      Number.isInteger(score) &&
+      score >= 0 &&
+      score <= 10 &&
+      NPS_SLACK_WEBHOOK
```
ahdriel: nit: should we also check for `IS_CLOUD`?

packages/back-end/src/routers/users/users.controller.ts
```diff
@@ -136,6 +141,145 @@ export async function putUserName(
...
+
+    res.status(200).json({
+      status: 200,
+    });
+  } catch (e) {
+    res.status(400).json({
+      status: 400,
+      message: e.message || "An error occurred",
+    });
+  }
```
ahdriel: suggestion: should we remove this and let the proper error bubble up instead of coalescing everything into a 400?

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
```diff
@@ -0,0 +1,429 @@
...
+  const { npsSurveyAt, updateUser } = useUser();
+  const suppressed = withinCooldown(npsSurveyAt);
+
+  const [mounted, setMounted] = useState(false);
+  const [visible, setVisible] = useState(false);
+  const [forceShow, setForceShow] = useState(false);
+  const [closing, setClosing] = useState(false);
+  const [panel, setPanel] = useState<Panel>("question");
+  const [score, setScore] = useState<number | null>(null);
+  const [feedback, setFeedback] = useState("");
```
ahdriel: question: why are we not using react-hook-form?

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
```diff
@@ -0,0 +1,429 @@
...
+  useEffect(() => {
+    scoreRef.current = score;
+  }, [score]);
+  useEffect(() => {
+    feedbackRef.current = feedback;
+  }, [feedback]);
+  useEffect(() => {
+    apiCallRef.current = apiCall;
+    updateUserRef.current = updateUser;
+  }, [apiCall, updateUser]);
```
ahdriel: question: why do we need so many different refs?

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
```diff
@@ -0,0 +1,429 @@
...
+            >
+              Quick question · ~15 sec
+            </Text>
+            <Heading as="h2" size="small" mb="4" mr="3">
+              How likely are you to recommend GrowthBook to a friend or
+              colleague?
+            </Heading>
+            {/* Hand-rolled radiogroup: the 11-cell scale needs per-cell
+                detractor/passive/promoter color grading, which RadioGroup /
+                RadioCards from the design system can't express. */}
```
ahdriel, line 330: question: should we check with the design team if we should add this ability to ui/RadioGroup / radioCards?

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
```diff
@@ -0,0 +1,429 @@
...
+            </Text>
+          </div>
+        )}
+      </div>
+    </div>
+  );
+
+  // Portal into #portal-root (rendered inside <RadixTheme>) rather than document.body,
+  // so the Radix theme CSS variables (panel background, shadow, colors) resolve.
+  const portalTarget = document.getElementById("portal-root") ?? document.body;
```
ahdriel: we have a `Portal` element, we should use that instead.

packages/back-end/src/models/UserModel.ts
gazzdingo: we could but wasnt sure if it was worth it for this

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
gazzdingo: i don't think it is worth it for the design system since this is not really used anywhere if we do decided to we could update the design system

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
gazzdingo: its used for abandonment flushing but i think we can also propely not so such strict flushing and just flush the values which will be easier then it make more sense to use react-hook-form your other comment

packages/front-end/components/NPSSurvey/NPSSurvey.tsx
```diff
@@ -0,0 +1,570 @@
...
+                  tabIndex={score === s || (score === null && s === 0) ? 0 : -1}
+                  className={`${styles.cell} ${
+                    CATEGORY_UI[npsCategoryOf(s)].className
+                  }`}
+                  onClick={() => {
+                    setValue("score", s);
+                    confirmedRef.current = true;
+                    setPanel("feedback");
+                  }}
+                  onKeyDown={(e) => {
```
bryce-fitzsimons, line 422: seems over engineered

bryce-fitzsimons (approved):
no major concerns. Just some UI nits that we discussed offline

gazzdingo:
@greptileai

PR #6188, opened 2026-06-23
perf(back-end): use a projected experiment fetch for the feature stale/dependents graph

anna-yn:
### Features and Changes

`GET /features/stale`, `GET /features/dependents`, and `GET /features/experiment-states` each load every experiment in the org via `getAllExperiments` to build the feature/experiment dependency graph. Each doc goes through `cloneDeep` + `upgradeExperimentDoc`, and full experiment docs carry `phases` (including `banditEvents`), `variations`, metric arrays, and analysis settings that the graph never reads. On large orgs this dominates request time and — because the deserialize/migrate work is synchronous — blocks other requests on the same process.

This adds `getAllExperimentsForStaleGraph`, mirroring `getAllFeaturesForStaleGraph` (#6135) for the experiment side. It projects only the fields the graph consumers read (traced through `buildExperimentDependencyIndex`, `getDependentExperiments`, `includeExperimentInPayload` / `experimentHasLinkedChanges`, `isFeatureStale` / `buildEnvResults`, and the temp-rollout scan in `getFeatureExperimentStates`), and skips `upgradeExperimentDoc`.

Of the projected fields, only `releasedVariationId` is derived by `upgradeExperimentDoc` (from `results` / `winner` / `variations[].id` on legacy stopped experiments). That backfill is mirrored inline so behavior is unchanged for pre-`releasedVariationId` docs; `results`, `winner`, and `variations.id` are projected to feed it. The inline backfill first normalizes missing `variations[].id` to their index (as `upgradeExperimentDoc` does), so a legacy doc whose variation ids aren't stored still derives a truthy `releasedVariationId` and isn't dropped by `includeExperimentInPayload`. All three handlers in `controllers/features.ts`, plus the external REST `GET /api/v1/features/stale` endpoint, are switched to the new helper.

This uses an inclusion projection (vs #6135's exclusion) because experiment docs carry more heavy nested array fields than features and the consumer field set is small and stable. Running `upgradeExperimentDoc` on a sparse inclusion-projected doc is not safe (it dereferences `variations.forEach`, `metricOverrides`, etc.), hence the targeted backfill instead.

The query filter (`organization`, `type: {$ne: "holdout"}`, `archived: {$ne: true}` unless `includeArchived`) and the project-read permission filter match `getAllExperiments` for these call sites.

### Dependencies

None.

### Testing

- ESLint on the changed files
- Manual: started a local back-end against a fresh MongoDB and seeded:
  - two features (`feat_a` with an experiment-ref rule, `feat_b` plain)
  - a running experiment with `feat_a` as a phase prerequisite
  - a stopped experiment with `linkedFeatures` and `excludeFromPayload: false`
  - a `type: "holdout"` experiment with `feat_b` as a prerequisite
  - a legacy stopped experiment with `results: "won"`, `winner: 1`, **no** `releasedVariationId` key, and `feat_b` as a prerequisite

  Verified `GET /features/stale?ids=feat_a,feat_b`, `/features/dependents?ids=feat_a,feat_b`, `/features/experiment-states?ids=feat_a,feat_b`, plus no-`ids`, empty-`ids`, and unknown-id probes:
  - `feat_a` → `stale: false, reason: "has-dependents"`; dependents lists the prerequisite experiment with `id` and `name`
  - `feat_b` → `stale: false, reason: "has-dependents"` after the legacy experiment is added (the inline backfill derives `releasedVariationId`, so `includeExperimentInPayload` returns true as before); the holdout experiment is correctly excluded
  - `feat_a` → `hasTempRollout: true` from the stopped/linked experiment
  - No errors in the back-end log; response shapes unchanged

  Also confirmed at the Mongo layer that `{"phases.prerequisites": 1}` preserves `phases` array length and ordering with only the `prerequisites` sub-field on each element.


Kevin-Chant:
cursor review

Kevin-Chant (approved):
LGTM. Thanks for the contribution!

PR #6200, opened 2026-06-23
Feature flag constants

bryce-fitzsimons:
Adds **Constants** — reusable named values you define once and reference from feature flag values, so you can change a value in one place and have every flag that uses it pick up the change automatically.

## What you can do
- Define constants of type string or JSON, with an optional per-environment value (e.g. a prod vs. staging URL behind one reference).
- Reference them in feature values:
  - String constants interpolate anywhere in a string: `{{ @const:api-server }}`
  - JSON constants drop in as a whole value, or spread into a larger object alongside other keys: `{ "$extends":["@const:default-config"], "timeout": 30 }`
  - Constants can reference other constants; references resolve recursively.
- References are resolved when the SDK payload is built, so SDKs receive fully-resolved values — no SDK changes or upgrades required. Unresolvable/cyclic references fail gracefully (left as-is), and references to archived constants are stripped from the payload.
- Versioning & approvals: constants go through the same draft → review → publish flow as features, governed by your existing require-reviews settings (with revert, compare, and audit history).
- Manage everywhere: a dedicated list page (search, filters, drafts tab) and detail page, plus a full REST API (CRUD, archive, references, and the revision/approval flow).
- References view: see which features (and other constants) use a given constant before changing or archiving it.

Notes
- Scoped by a single project
- Uses a new Manage Constants permission.
- Added a new Constants doc section

Screenshot 2026-06-23 at 11 41 10 PM

Screenshot 2026-06-23 at 4 17 37 PM


bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@cursor review

bryce-fitzsimons:
@cursor review

packages/back-end/src/services/constants.ts
```diff
@@ -0,0 +1,146 @@
...
+  return out;
+}
+
+// The set of constant keys referenced anywhere in a feature's values.
+function featureConstantKeys(feature: FeatureInterface): Set<string> {
+  const keys = new Set<string>();
+  for (const value of featureValueStrings(feature)) {
+    for (const key of getConstantReferenceKeys(value, undefined)) keys.add(key);
+  }
+  return keys;
```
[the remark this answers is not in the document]
bryce-fitzsimons: false positive

bryce-fitzsimons:
@cursor review

packages/back-end/src/models/BaseModel.ts
```diff
@@ -908,6 +908,13 @@ export abstract class BaseModel<
 
     await this._dangerousGetCollection().insertOne(doc);
 
+    // insertOne mutates `doc` in place to add Mongo's `_id`. Strip it (and the
+    // mongoose version key) so the created doc matches the schema/interface and
+    // these internals don't leak into the return value, audit log details, or
+    // hooks — reads already strip them via `_removeMongooseFields`.
+    delete (doc as Record<string, unknown>)._id;
```
gazzdingo: would there be a case where people want the mongo id im guessing not but i feel like we already get it in the shape we need it

bryce-fitzsimons: this shouldn't be there. I removed it

gazzdingo (approved):
LGTM

PR #6202, opened 2026-06-24
feat(clickhouse): cancel in-flight queries via KILL QUERY

ahdriel:
### Features and Changes

For ClickHouse canceling a query would ignore it on our side but would not stop the execution on the warehouse, as it inherited the no-op `cancelQuery` from `SqlIntegration` and never tagged its queries with an id. So "Cancel" only marked the query failed in Mongo while the query kept running on the warehouse.

This wires ClickHouse into the existing cancellation machinery.

- `runQuery` now generates a `query_id`, using the existing `setExternalId` callback, and passes it to the ClickHouse client.
- Implements `cancelQuery`, which issues `KILL QUERY WHERE query_id = '<id>'`. `KILL` is async by default, so it returns once ClickHouse accepts the request, not once the target query has stopped.
- Pulled client construction into a small `getClient()` helper, shared by `runQuery` and `cancelQuery`.

### Testing

- Manual testing


lukesonnet (approved):

packages/back-end/src/integrations/ClickHouse.ts
```diff
@@ -82,6 +101,21 @@ export default class ClickHouse extends SqlIntegration {
...
+  async cancelQuery(externalId: string): Promise<void> {
+    if (isManagedWarehouseAwaitingProvisioning(this.datasource)) {
+      throw new ManagedWarehousePendingError();
+    }
+    const client = this.getClient();
+
+    // KILL QUERY is async by default — this returns once ClickHouse accepts
+    // the request, not once the target query has actually stopped.
+    await client.command({
+      query: "KILL QUERY WHERE query_id = {qid:String}",
```
ahdriel: blocker: For a clustered deployment we need to emit ON CLUSTER {clusterName} so it propagates, as there's no guarantee that the pod receiving the cancel query is the one that is running the other query.

But we cannot blanket emit ON CLUSTER ... for all queries because this fails on a deployment that there's no cluster.

We also need to know the name of the cluster, which by default is named default, but it might not be.

lukesonnet (changes requested):
updating status to better match blocker comment adriel shared

PR #6205, opened 2026-06-24
feat(agg-fact-table): defer first aggregation to next scheduled slot on creation

ahdriel:
### Features and Changes

When a fact table is created with aggregation already enabled, the scheduling poller would pick it up on its next pass and immediately run a full restate if a metric existed, even though the table was created moments ago. That first run is redundant and expensive.

This seeds the schedule on creation so the first aggregation waits for the next scheduled `updateTime` slot instead of firing right away. 

### Testing

- Unit tests


lukesonnet (approved):

ahdriel:
cursor review

PR #6209, opened 2026-06-25
fix(incremental,product-analytics): hash segment/exposure-query definitions; reject stale exploration cache

olu-an:
## The bug

Two cache keys reference definitions by ID, not content, so editing the underlying SQL serves stale data with no error:

- **Incremental refresh** — `getExperimentSettingsHashForIncrementalRefresh` hashes `snapshotSettings.segment` and `exposureQueryId` (ID strings only). Edit a segment's SQL or a FACT segment's filters → hash unchanged → Update reuses a units table built against the old population → wrong experiment results.
- **Product-analytics explorations** — `getConfigHashes` keys on `factTableId` / `metricId`. Edit a fact table's SQL → `findLatestByConfig` returns the prior result and the runner is never reached. No TTL.

## The fix

- **Incremental:** `getExperimentSettingsHashForIncrementalRefresh` now takes a `ResolvedSettingsRefs` carrying the segment's `{sql, factTableId, filters, userIdType}` and the exposure-query SQL, with a `v:2` version byte. `assertIncrementalRefreshPrerequisites` takes `context` (not `org`) and resolves the refs; the runner write site builds them from the `segmentObj`/`exposureQuery` already in scope so write and read stay in lockstep.
- **Explorations:** `findLatestByConfig` rejects a `configHash` match if any referenced fact table or fact metric has `dateUpdated > cachedDoc.dateCreated`. No hash-format change, so existing caches aren't mass-invalidated; refs are only fetched when a candidate exists.

## Cost

Every pipeline-mode experiment Full-Refreshes once on its first post-deploy Update (`v:2` never matches a stored hash). Scheduled runs auto-promote via `resolveIncrementalPrerequisiteFailure` (#6076); manual clicks see the existing "configuration is outdated" error once. Exploration caches are not mass-invalidated.

## Tests

New cases in `test/enterprise/data-pipeline.test.ts`: segment SQL change with unchanged ID throws; unchanged segment SQL still passes (write/read parity); exposure-query SQL change and FACT-segment filter change each produce a different hash. Golden-hash re-pinned. No existing tests cover `findLatestByConfig`.

### DIFF
--- a/packages/back-end/src/enterprise/services/data-pipeline.ts
+++ b/packages/back-end/src/enterprise/services/data-pipeline.ts
@@ -18,13 +18,15 @@ import {
   ExperimentSnapshotSettings,
   MetricForSnapshot,
 } from "shared/types/experiment-snapshot";
-import { OrganizationInterface } from "shared/types/organization";
 import { ExperimentInterface } from "shared/types/experiment";
 import {
   FactMetricInterface,
   FactTableInterface,
 } from "shared/types/fact-table";
+import { SegmentInterface } from "shared/types/segment";
 import { orgHasPremiumFeature } from "back-end/src/enterprise";
+import { ApiReqContext } from "back-end/types/api";
+import { ReqContext } from "back-end/types/request";
 import { ExperimentIncrementalPipelineRequiresFullRefreshError } from "back-end/src/util/errors";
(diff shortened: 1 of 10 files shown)


olu-an:
Closing for now but please feel free to reopen if you still think we need this

PR #6212, opened 2026-06-25
Show effective (user + team) project role in member tables

Kevin-Chant:
### Features and Changes

The member tables on the **Members** page ("View roles and permissions for [PROJECT]") and the **Project** page (Project Members) only showed each member's *user-level* role for the selected project. They didn't reflect the role that actually applies once team memberships are combined in — so a member who is, say, `No Access` directly but `Engineer` via a team appeared as `No Access`.

- Adds a shared util `getEffectiveRolesForProject(member, project, teams)` that resolves the roles actually in effect for a member on a project, combining their own role with any teams they're on. It uses the same precedence as the back-end permission merge (#6179): an explicit project-scoped role — from the member or any team — takes precedence over global roles for that project, and global roles only contribute when no explicit project role applies. The result is the set of contributing roles (a union, which can be more than one). Pass `project = null` for the global/"All Projects" view.
- Adds an **Effective Role** column to `MemberList` (shared by both views). It renders the resulting role(s), and when teams contribute (or more than one role applies) shows a tooltip naming each source (`Direct` / `Team: <name>`). The existing user-level role column is kept for comparison, per the request.

Both views pick this up automatically since they share `MemberList`. No back-end changes — all the needed data (member `teams`, team `projectRoles`) is already on the client via `useUser()`.

Note: the per-environment access columns still reflect the user-level role; extending those to the effective role would be a reasonable follow-up but is out of scope here.

### Testing

- [x] Unit tests for `getEffectiveRolesForProject` in `packages/shared/test/permissions.test.ts` (own-project-role over global; team global not leaking past an explicit project role; union of explicit user + team project roles; global fallback when no project role applies; `project = null` global resolution)
- [x] Visual check of both the Members page and Project page member tables

### Screenshots

image

image

### DIFF
--- a/packages/front-end/components/Settings/Team/MemberList.tsx
+++ b/packages/front-end/components/Settings/Team/MemberList.tsx
@@ -5,7 +5,10 @@ import { date, datetime } from "shared/dates";
 import { RxIdCard } from "react-icons/rx";
 import router from "next/router";
 import { Flex } from "@radix-ui/themes";
-import { getRoleDisplayName } from "shared/permissions";
+import {
+  getEffectiveRolesForProject,
+  getRoleDisplayName,
+} from "shared/permissions";
 import { roleHasAccessToEnv, useAuth } from "@/services/auth";
 import { useUser } from "@/services/UserContext";
 import ProjectBadges from "@/components/ProjectBadges";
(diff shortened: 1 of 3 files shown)


Kevin-Chant:
cursor review

lukesonnet (approved):

lukesonnet:
I think this change is good. It might be further nice to:
1. call out which role is the one that *actually* applies, but I know that can be tricky if the roles are non-overlapping, so maybe this isn't possible.
2. let you sort this column or filter it otherwise to see "who has any access" or "who has at least this role" but that's also a nice to have.

Thanks @Kevin-Chant !

PR #6214, opened 2026-06-25
feat(back-end): use native approximate top-k aggregates for fact table column top values

ahdriel:
### Features and Changes

The fact-table column "top values" scan previously unpivoted every string column and grouped/counted exact values across the lookback window. On large tables this is an expensive full scan with a wide GROUP BY.

Most of our supported warehouses ship a native single-pass approximate top-k aggregate. This PR uses them when available, falling back to the existing exact path on dialects that don't.

- Per-dialect implementations:
  - **BigQuery** — `APPROX_TOP_COUNT`
  - **Snowflake** — `APPROX_TOP_K`
  - **Databricks** — `approx_top_k`
  - **Presto / Athena** — `approx_most_frequent`
  - **ClickHouse** — `topK`
- Dropped the `count` field from `ColumnTopValuesResponseRow`. It was never used by consumers, and the engines disagree on whether/how they expose counts (ClickHouse's `topK` returns none, so we synthesize an ordering-only sort key). Rows are still ordered most-frequent-first per column.

### Testing

- New unit tests in `columns-top-values-query.test.ts`
- Manually validated against live BigQuery, Snowflake, Databricks, ClickHouse, and Presto warehouses

### DIFF
--- a/packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
+++ b/packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
@@ -1,0 +1,78 @@
+import type { ApproxTopValuesParams, SqlDialect } from "shared/types/sql";
+
+/**
+ * Map NULL or over-length values to NULL so the approximate aggregates (which
+ * ignore NULLs) exclude them, mirroring the exact path's
+ * `value IS NOT NULL AND LENGTH(value) <= maxValueLength` filter. With no
+ * `maxValueLength`, the expression is returned unchanged.
+ */
+export function eligibleTopValueExpr(
+  dialect: SqlDialect,
+  valueSql: string,
+  maxValueLength?: number,
+): string {
+  if (maxValueLength === undefined) return valueSql;
+  return dialect.ifElse(
+    `${valueSql} IS NOT NULL AND ${dialect.stringLength(
+      valueSql,
+    )} <= ${maxValueLength}`,
+    valueSql,
+    "NULL",
+  );
+}
+
+/**
+ * Tracking capacity for the Space-Saving top-k family, sized well above `k` so
+ * counts stay accurate for the low-cardinality columns this runs on. Maps to
+ * Snowflake `counters`, Databricks `maxItemsTracked`, and Trino/Presto/Athena
+ * `capacity`; BigQuery's APPROX_TOP_COUNT and ClickHouse's topK take no such
+ * argument.
+ */
+export function approxTopKCapacity(limit: number): number {
+  return Math.min(100000, Math.max(limit * 10, 1000));
+}
+
+/**
+ * Trino / Presto / Athena single-pass approximate top-k using
+ * `approx_most_frequent(buckets, value, capacity)`, which returns a
+ * `MAP(value, count)` per column. One aggregation row holds parallel arrays of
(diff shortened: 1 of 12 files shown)


ahdriel:
cursor review

packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
```diff
@@ -0,0 +1,78 @@
...
+  return Math.min(100000, Math.max(limit * 10, 1000));
+}
+
+/**
+ * Trino / Presto / Athena single-pass approximate top-k using
+ * `approx_most_frequent(buckets, value, capacity)`, which returns a
+ * `MAP(value, count)` per column. One aggregation row holds parallel arrays of
+ * the column names and their maps; we expand both with UNNEST.
+ */
+export function approxMostFrequentTopValuesCTEBody(
```
lukesonnet: question: does this method, which feeds the dialects, belong here? Probably, but wondering if a slightly different organization makes sense.

packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
```diff
@@ -0,0 +1,78 @@
...
+    SELECT
+      ARRAY[${names}] AS col_names,
+      ARRAY[
+        ${maps}
+      ] AS col_maps
+    FROM ${fromTable}
+    WHERE ${whereClause}
+  ) __agg
+  CROSS JOIN UNNEST(__agg.col_names, __agg.col_maps) AS __col (column_name, items)
+  CROSS JOIN UNNEST(__col.items) AS __item (value, count)
```
lukesonnet: This seems right to me! but double cross join unnest seems like a thing worth QAing :)

lukesonnet (approved):
I didn't dig too much into the top-k methods that you added, but they seem reasonable as does this general cross join unnest pattern.

no testing on my end.

packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
ahdriel: I think it is fine here -- because it is shared by presto & athena, so we want a neutral spot, and `approx-top-values` makes sense to hold that query as well.

maybe not under `clauses` I guess?

good question on how we would like to organize this folder more broadly -- I personally do like the concept of, here are all queries for getting `approx-top-values` and you can see how they differ per dialect/engine. so in my world we could expand on this and also add `bigquery` and other engines queries here to this file so we can see in a single place all the queries that satisfy the same contract as they need to emit the same rows in the end so the whole pipeline works.

packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
ahdriel: Indeed! this is a bounded join because it happens after the aggregation, so it is columns x k top-values. finite and expected.

but thanks for the callout

packages/back-end/src/integrations/sql/clauses/approx-top-values.ts
lukesonnet: yeah I feel like they belong here in that way, but then it's weird when they are only used by one dialect if they are created as global functions here. anyways it obv doesn't block this PR.

PR #6220, opened 2026-06-25
feat(settings): deep-link to a section within a settings tab

ahdriel:
### Features and Changes

The settings page tabs persisted the active tab in the URL hash, but there was no way to link directly to a specific section *within* a tab. This adds support for a composite `tab/section` hash so other parts of the app can deep-link users to the exact setting they need.

For example, `/settings#metrics/data-source-settings` switches to the "Metrics & Data" tab and smooth-scrolls to the Data Source Settings section.

- `settingsSections.ts` is a small, typed registry mapping section ids to their owning tab, plus a `parseSettingsHash` helper that resolves a hash into `{ tab, section }` with safe fallbacks for unknown values.
- The settings page now drives the active tab from the parsed hash and runs the scroll-into-view on the targeted section.
- The Data Source Settings `Frame` gets an `id` (and `scrollMarginTop`) so it can be targeted.

### Testing

Unit tests for `parseSettingsHash` and `isSettingsSectionId` cover composite hashes, tab-only hashes, empty/unknown hashes, and unknown section segments.

### DIFF
--- a/packages/front-end/components/GeneralSettings/settingsSections.ts
+++ b/packages/front-end/components/GeneralSettings/settingsSections.ts
@@ -32,11 +32,16 @@ export function parseSettingsHash(hash: string | undefined): {
   section: SettingsSectionId | null;
 } {
   const [tabSegment, sectionSegment] = (hash ?? "").split("/");
+  const section =
+    sectionSegment && isSettingsSectionId(sectionSegment)
+      ? sectionSegment
+      : null;
   return {
-    tab: isSettingsTab(tabSegment) ? tabSegment : DEFAULT_SETTINGS_TAB,
-    section:
-      sectionSegment && isSettingsSectionId(sectionSegment)
-        ? sectionSegment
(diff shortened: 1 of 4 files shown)


packages/front-end/components/GeneralSettings/settingsSections.ts
```diff
@@ -0,0 +1,47 @@
+export const SETTINGS_TAB = {
```
msamper, line 1: Does this file maybe belong in a different directory?

msamper (approved):

packages/front-end/components/GeneralSettings/settingsSections.ts
ahdriel: Maybe it would be more related to `pages/settings` but we cannot add it there (we could if we moved to `app directory`) so I am keeping it here for now. Open to moving it though if you have a better suggestion.

PR #6225, opened 2026-06-26
fix(fact-tables): only refresh columns when SQL or event name actually changes

ahdriel:
### Features and Changes

`needsColumnRefresh` decided whether to run column detection based only on whether `sql` or `eventName` were *present* in the update payload. But clients (the UI and API) often resend those fields unchanged on every edit, so renaming a fact table or tweaking unrelated settings would still trigger a full column detection scan against the warehouse.

Now `needsColumnRefresh` takes the existing fact table and compares values, so column detection only runs when `sql` or `eventName` actually changed. This avoids unnecessary (and potentially expensive) warehouse scans on no-op edits.

### Testing

Unit tests

### DIFF
--- a/packages/back-end/src/api/fact-tables/updateFactTable.ts
+++ b/packages/back-end/src/api/fact-tables/updateFactTable.ts
@@ -1,6 +1,9 @@
 import { omit } from "lodash";
 import { updateFactTableValidator } from "shared/validators";
-import { UpdateFactTableProps } from "shared/types/fact-table";
+import {
+  FactTableInterface,
+  UpdateFactTableProps,
+} from "shared/types/fact-table";
 import { queueFactTableColumnsRefresh } from "back-end/src/jobs/refreshFactTableColumns";
 import { getDataSourceById } from "back-end/src/models/DataSourceModel";
 import {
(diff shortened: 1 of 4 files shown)


packages/back-end/src/api/fact-tables/updateFactTable.ts
```diff
@@ -156,6 +159,12 @@ export const updateFactTable = createApiRequestHandler(
   };
 });
 
-export function needsColumnRefresh(changes: UpdateFactTableProps): boolean {
-  return !!(changes.sql || changes.eventName);
+export function needsColumnRefresh(
+  existing: Pick<FactTableInterface, "sql" | "eventName">,
+  changes: UpdateFactTableProps,
+): boolean {
+  const sqlChanged = changes.sql !== undefined && changes.sql !== existing.sql;
```
lukesonnet, line 166: nit/future-question/question/thing: is this `sql` like white space trimmed or formatted in a way to reduce irrelevant changes? Definitely not useful for here, but something that it might be nice to have a service for at some point if we don't already.

msamper (approved):

PR #6226, opened 2026-06-26
Quick fix to handle NoOutputGeneratedError

Auz:
parsePrompt only caught NoObjectGeneratedError. When the model ends a tool-using run WITHOUT emitting the output object (stops on a tool call or returns prose), the AI SDK throws the distinct NoOutputGeneratedError, which fell straight through to an opaque "No output generated." 400 — no retry, no friendly message, no diagnostics. Observed in prod on a visual-editor move the model couldn't ground.

- parsePrompt: treat both error classes the same — retry once, then the friendly "couldn't format / too large" message, and log diagnostics (errorType, finishReason/textSample where present, cause, logContext). NoOutputGeneratedError carries no finishReason/text/usage, so those reads are guarded.
- Visual-editor edit prompt: strengthen the output contract — the model MUST always end its turn with the JSON object, even to decline (return mutations: [] + an explanation); never stop on a tool call or reply in prose. Prevents the failure at the source.


bryce-fitzsimons (approved):

Auz:
@greptileai review

PR #6230, opened 2026-06-29
Address security vulnerabilities (2026-06-29)

Kevin-Chant:
## Fixed

- **`cryptography` `46.0.7` → `>=48.0.1,<49` (HIGH, GHSA-537c-gmf6-5ccf)** — the pybuild stage already force-pins cryptography (the `<47` cap was added in a prior remediation). That cap was holding the venv at 46.x; the advisory is patched only in `48.0.1`. Bumped the existing pin to `>=48.0.1,<49`. cryptography `48.0.1` publishes a CPython 3.11 wheel (`requires_python >=3.9`), so it installs without a build change.

- **`nodemailer` `8.0.11` → `9.0.1` (HIGH, GHSA-p6gq-j5cr-w38f)** — direct runtime dep of `back-end` used to send transactional email. Verified the major bump against our usage before taking it: the only `9.0` breaking change is TLS certificate validation on **remote content fetching** (attachment `href`/`path` URLs, OAuth2 token endpoints, HTTP/HTTPS proxy `CONNECT`). [`back-end/src/services/email.ts`](packages/back-end/src/services/email.ts) does none of those — it's plain SMTP `createTransport` + `sendMail` with inline `html`/`text`/`headers` — so the change is a no-op for us. Also bumped `@types/nodemailer` `7.0.11` → `8.0.1` (latest, matches our TS 5.7).

- **`@babel/core` → `^7.29.7` via `pnpm.overrides` (LOW, CVE-2026-49356)** — the earlier "purely a leaking devDependency" read didn't hold up: the Dockerfile's `pnpm install --prod --no-optional` already strips front-end's `optionalDependencies` entry, and `@babel/core` lands in the prod image via real production transitive deps (`@sentry/nextjs → @sentry/webpack-plugin → @sentry/bundler-plugin-core`, `react-select → @emotion/react → @emotion/babel-plugin`, `next → styled-jsx`). All of those accept `^7`, so an in-major override pin clears the CVE without the `8.x` major-version risk.

## Skipped

- **`thrift` `0.16.0` → `0.23.0` (HIGH, CVE-2026-41636)** — transitive via `@databricks/sql`. Not self-fixable: the thrift bump is blocked on upstream [databricks/databricks-sql-nodejs#390](https://github.com/databricks/databricks-sql-nodejs/pull/390), still open/unmerged as of this PR (last updated 2026-05-30). An override would break the Databricks SQL connector. Waiting on upstream.

## Post-deploy projection

| Severity | Total findings | Addressed or auto-clears | Remaining | — of which pending upstream |
| -------- | -------------- | ------------------------ | --------- | --------------------------- |
| HIGH     | 3              | 2                        | 1         | 1                           |
| LOW      | 1              | 1                        | 0         | 0                           |


Dockerfile
tzjames, line 31: Our python code does not need cryptography and is unlikely to need to.  We could instead just get rid of cryptography altogether here I believe.

tzjames (approved):

PR #6233, opened 2026-06-29
refactor(ui): migrate alert-class callouts to the design system Callout

ahdriel:
### Features and Changes

Retires the last Bootstrap `alert`-class callouts across the front-end and replaces them with the design-system `@/ui/Callout`. This clears the entire `local/no-alert-classname` suppression ledger (126 entries across 90+ files) so the lint rule now holds with zero exceptions.

The migration also tightens the `Callout` API so status and children are the main customization path:

- Drops `color`, `variant`, and `contentsAs`. Status now drives the color, the variant is always `soft`, and children always render inside a `div` (no more invalid `<div>`-inside-`<p>` nesting).
- Adds `status="attention"` (orange) for banners that should read louder than a warning without being a hard error. It replaces ad-hoc `color="orange"` overrides and keeps the error octagon icon plus `role="alert"`.
  - This is a temporary change, we should either make this official or remove it. Keeping it for now as a product of the migration.
- Adds an `action` slot for a button aligned to the callout's first text line, and `color="inherit"` on `Button` so it picks up the surrounding callout accent.

### Testing

- Visual QAd a bunch of pages that covers the different patterns (just message, message with link, message with button, loading indicator, etc)
compare-gallery


packages/front-end/components/Experiment/TabbedPage/BanditUpdateStatus.tsx
```diff
@@ -208,29 +210,31 @@ export default function BanditUpdateStatus({
...
-                    status={status}
-                    display={null}
-                    color="link link-purple p-0 pb-1"
-                    condensed={true}
-                    hideQueryCount={true}
-                  />
+            <Callout status="error" size="sm" mx="2">
+              <Flex align="start" justify="between" gap="2">
+                <div>
+                  <FaExclamationTriangle className="mr-1" />
```
ahdriel: question: are we rendering the icon twice?

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
```diff
@@ -463,9 +463,9 @@ export default function SDKConnectionForm({
         <div className="form-group">
           <label>SDK Language</label>
           {languageError ? (
-            <span className="ml-3 alert px-1 py-0 mb-0 alert-danger">
+            <Callout status="error" ml="3" mb="0" className="px-1 py-0">
```
ahdriel: Remove className?

packages/front-end/components/Features/CodeSnippetModal.tsx
```diff
@@ -559,14 +558,18 @@ myAttribute = sha256(salt + myAttribute);
...
-                          exclusively on hashing as a means of securing highly
-                          sensitive data. Hashing is an obfuscation technique
-                          that makes it very difficult, but not impossible, to
-                          extract sensitive data.
-                        </div>
+                        <Callout
+                          status="warning"
+                          mt="3"
+                          mb="0"
+                          icon={<FaExclamationCircle />}
```
ahdriel: do we need custom icon here?

packages/front-end/components/Revision/RevisionDetail.tsx
```diff
@@ -459,28 +463,29 @@ function RevisionDetail<T>({
       )}
       {revision.status === "merged" && (
         <Callout status="success" mb="4">
-          <Flex justify="between" align="center">
-            <Text size="medium">
-              This revision has been merged and published.
-            </Text>
-          </Flex>
+          <Text size="medium">
```
ahdriel: question: do we need Text inside Callout? should we?

packages/front-end/ui/Callout.module.scss
```diff
@@ -1,3 +1,14 @@
 .callout p:last-of-type {
   margin-bottom: 0;
 }
+
+// The icon, action, and dismiss button are each sized to the text's first
+// line and centered within it. They align on that line and never grow the
+// row, so multi-line content stays top-aligned and tall actions (buttons)
+// overflow into the padding instead of pushing the content down.
+.firstLineBox {
```
ahdriel: find a better name

packages/front-end/ui/Callout.tsx
```diff
@@ -39,25 +41,27 @@ export default forwardRef<
...
     children: ReactNode;
     status: Status;
-    color?: RadixCallout.RootProps["color"]; // Use status instead of color whenever possible
     size?: "sm" | "md";
     icon?: ReactNode | null;
-    contentsAs?: "text" | "div";
-    variant?: "soft" | "surface" | "outline";
+    action?: ReactNode;
+    className?: string;
+    style?: React.CSSProperties;
```
ahdriel: why are we adding these 2?

packages/front-end/ui/Callout.tsx
```diff
@@ -81,48 +85,58 @@ export default forwardRef<
...
     <RadixCallout.Root
       ref={ref}
-      className={styles.callout}
-      color={color || getRadixColor(status)}
-      role={status === "error" ? "alert" : undefined}
+      className={clsx(styles.callout, className)}
+      color={getRadixColor(status)}
+      role={
+        role ??
+        (status === "error" || status === "attention" ? "alert" : undefined)
```
ahdriel, line 93: get rid of the role overrides if they are not needed

packages/front-end/components/Experiment/ReleaseChangesForm.tsx
```diff
@@ -215,8 +216,7 @@ export default function ReleaseChangesForm({
                 : "Sticky Bucketed users will be excluded from the experiment."
             : "No sticky bucketing."}
           {form.watch("newPhase") && isBandit && (
-            <div className="alert alert-warning text-danger mt-2">
-              <FaExclamationCircle className="mr-2" />
+            <Callout status="warning" mt="2" className="text-danger">
```
msamper: Do we need this classname? seems like maybe this should be a `Text` color prop if we want to support `text-danger`

packages/front-end/components/Features/SDKConnections/SDKConnectionForm.tsx
msamper: investigate if we can just add padding via radix `px` and `py` props

packages/front-end/components/Features/CodeSnippetModal.tsx
```diff
@@ -537,15 +532,19 @@ export default function CodeSnippetModal({
                           Example, using your organization&apos;s secure
                           attribute salt:
                           {secureAttributeSalt === "" && (
-                            <div className="alert alert-warning mt-2 px-2 py-1">
-                              <FaExclamationTriangle /> Your organization has an
-                              empty salt string. Add a salt string in your{" "}
+                            <Callout
+                              status="warning"
+                              mt="2"
+                              className="px-2 py-1"
```
msamper: same here

packages/front-end/components/Features/CodeSnippetModal.tsx
msamper: I think we don't need this icon

packages/front-end/components/HealthTab/BanditSRMCard.tsx
```diff
@@ -127,10 +128,10 @@ export default function BanditSRMCard({
                 />
               </>
             ) : (
-              <div className="alert alert-info font-weight-bold">
+              <Callout status="info" className="font-weight-bold">
```
msamper: Should we do this via a `Text` wrapper with a the appropriate weight set via a prop?

packages/front-end/components/HomePage/ExperimentImpact/ExperimentImpactTab.tsx
```diff
@@ -156,10 +156,9 @@ export default function ExperimentImpactTab({
         </td>
         {e.error ? (
           <td colSpan={3}>
-            <div className="alert alert-danger px-2 py-1 mb-1 ml-1">
-              <FaExclamationTriangle className="mr-1" />
+            <Callout status="error" mb="1" ml="1" className="px-2 py-1">
```
msamper: same here

packages/front-end/components/HomePage/ExperimentsGetStarted.tsx
```diff
@@ -264,15 +265,19 @@ const ExperimentsGetStarted = (): React.ReactElement => {
                 </div>
               </div>
             </div>
-            <div className="alert alert-info text-center">
+            <Callout status="info" className="text-center">
```
msamper: `<Text align="center">` instead of the className?

packages/front-end/components/SafeRollout/Results/ResultsTable.tsx
```diff
@@ -428,10 +428,14 @@ export default function ResultsTable({
                                   })}
                                 </div>
                               ) : null}
-                              <div className="alert alert-danger px-2 py-1 mb-1 ml-1">
-                                <FaExclamationTriangle className="mr-1" />
+                              <Callout
+                                status="error"
+                                mb="1"
+                                ml="1"
+                                className="px-2 py-1"
```
msamper: same here

packages/front-end/components/Settings/DisplayTestQueryResults.tsx
```diff
@@ -384,15 +383,16 @@ export default function DisplayTestQueryResults({
...
+                </Callout>
               )
             ) : (
               showNoRowsWarning &&
               !results.length && (
-                <div className="alert alert-warning mr-auto">
-                  <FaExclamationTriangle /> No rows returned, could not verify
-                  result
-                </div>
+                <Callout status="warning" className="mr-auto">
```
msamper: Checked and it seems like radix does support `mr` `auto` so maybe we should use that

packages/front-end/components/Settings/EditLicenseModal.tsx
```diff
@@ -86,12 +87,12 @@ const EditLicenseModal: FC<{
...
             </button>
             {successMessage ? (
-              <div className="alert alert-success">{successMessage}</div>
+              <Callout status="success">{successMessage}</Callout>
             ) : (
-              <div className="alert alert-warning">
-                <FaExclamationTriangle /> You already have an active license.
-                Click &quot;Change&quot; to enter a new license key.
-              </div>
+              <Callout status="warning" icon={<FaExclamationTriangle />}>
```
msamper: do we need this icon?

packages/front-end/components/Settings/EnvironmentModal.tsx
```diff
@@ -207,12 +208,12 @@ export default function EnvironmentModal({
           closeMenuOnSelect={true}
         />
         {hasMoreSpecificProjectFilter && sdkConnections.length > 0 && (
-          <div className="alert alert-warning">
-            <FaExclamationTriangle /> You have made the projects filter more
-            restrictive than before. {sdkConnections.length} SDK Connection
+          <Callout status="warning" icon={<FaExclamationTriangle />}>
```
msamper: same here. do we need this icon?

packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -141,52 +141,46 @@ export default function SubscriptionInfo() {
               {subscription?.nextBillDate}
             </div>
             {subscription?.hasPaymentMethod === true ? (
-              <div
-                className="mt-3 px-3 py-2 alert alert-success row"
+              <Callout
+                status="success"
                 style={{ maxWidth: 650 }}
```
msamper: can we do `maxWidth` via a `Box` wrapping the callout rather than accepting `style`

packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -141,52 +141,46 @@ export default function SubscriptionInfo() {
...
+                  Click <strong>View Plan Details</strong> below to add a
+                  payment method.
+                </p>
+              </Callout>
             ) : null}
           </div>
         )}
       {subscription?.pendingCancelation && subscription?.dateToBeCanceled && (
-        <div className="col-md-12 mb-3 alert alert-danger">
+        <Callout status="error" mb="3" className="col-md-12">
```
msamper: can we wrap this in a `Box` and give it the `col-md-12` className?

packages/front-end/components/Settings/WebhooksModal.tsx
```diff
@@ -407,9 +408,9 @@ export function CreateSDKWebhookModal({
               helpText={
                 <>
                   {!validHeaders ? (
-                    <div className="alert alert-danger mr-auto">
+                    <Callout status="error" className="mr-auto">
```
msamper: `mr="auto"`?

packages/front-end/components/TempMessage.tsx
msamper: I think you should be able to achieve this styling with a wrapper around the `Callout`

packages/front-end/ui/Callout.tsx
msamper: Would be ideal with we could avoid adding these props here. From what I can tell looking through the PR, we should be able to achieve the styling we're using these props for with wrappers that use the classnames/styles and Radix margin props

packages/front-end/components/TempMessage.tsx
```diff
@@ -48,11 +49,9 @@ const TempMessage: FC<TempMessageProps> = ({
   }, [closing]);
 
   return (
```
msamper, line 48: This component actually doesn't seem to be rendering right now for some reason. Tested on the custom markdown settings page where it should appear when you save. When I apply a wrapper with the className and style then it renders

packages/front-end/ui/HelperText.tsx
```diff
@@ -26,6 +37,8 @@ export function getRadixColor(status: Status): TextProps["color"] {
       return "red";
     case "success":
       return "green";
+    case "attention":
```
msamper, line 40: Might be worth checking with Kari if we want to add another color to this component. I'm not sure users will identify a difference in severity between amber and orange

msamper (changes requested):
Reviewed offline with Adriel and did a pass myself. Ideally I would love it if we can avoid introducing `style` and `className` as props to the `Callout` component. It seems like we may be able to use wrappers and radix margin props to avoid introducing those props. I found a blocking bug with `TempMessage` not rendering on the custom markdown settings when saving so we'll want to fix that before we merge. Otherwise everything else looks good to me. Most of the changes are simple drop in replacements.

ahdriel:
@msamper Thank you so much for the detailed review! I have addressed all the comments (including removing `classname` and `style` and fixing `TempMessage`).

msamper (approved):

PR #6235, opened 2026-06-29
chore(front-end): default useRadix and useRadixButton to true

ahdriel:
### Features and Changes

The `useRadix` / `useRadixButton` props gate whether a component renders with the new Radix design system or the legacy styling. They defaulted to `false`, so every new usage of the components had to opt in manually.

This flips those defaults to `true` so new usages get the Radix styling by default. To keep behavior identical to what is currently deployed, every existing call site is pinned to its current rendering with an explicit `useRadix={false}` / `useRadixButton={false}`.


alyssamichelle (approved):
Audit: default flip looks complete

I pulled the branch and audited all 7 components whose defaults flipped, you did a thorough job, this looks complete.

I do wonder about all the places that have `useRadixButton={true}` still set. Should it be a follow up PR or this one where those redundant defaults are cleaned up?

ahdriel:
thanks for the review @alyssamichelle -- I have pushed a commit here to remove the `useRadix={true}` as you suggested. good callout!

PR #6238, opened 2026-06-29
fix(experiments): preserve URL casing in pre-launch checklist tasks (#6237)

### LINKED ISSUE #6237: [Bug] URLs added to custom Experiment Pre-Launch Checklist tasks are converted to all lowercase
### Summary

In an Experiment Pre-Launch Checklist, when attempting to create a custom task that includes a URL with some uppercase letters, like a Google Drive link, the custom task and URL can be saved, but after saving, the URL no longer works, because the URL has been converted to all lowercase. 

#### Expected Behavior

After the custom task is saved, the URL should be preserved exactly as the user entered it, so that when the user clicks on the task, it takes them to the same location that the URL they pasted in took them to. 

#### Current Behavior

When I paste my URL into a web browser, it takes me to my Google Doc, but when I click the hyperlink of the new custom task I've just created using it, it takes me to Google Drive's "Sorry, the file you have requested does not exist. Make sure that you have the correct URL and the file exists." page. 

In this example, the URL "https://www.google.com/search?q=GrowthBook" that I've saved should point to a Google search page with the search term "GrowthBook". 
Image

It instead points to a Google search page at "https://www.google.com/search?q=growthbook" with the search term "growthbook". Because Google Drive uses mixed case for URL paths, this is a breaking string modification. 
Image

CedricConday:
## What

In an **Experiment Pre-Launch Checklist**, the "Add URL" field of a custom task lower-cased the entire URL:

```ts
const url = e.target.value.toLowerCase();
```

Paths and query strings are case-sensitive, so a pasted link like `https://www.google.com/search?q=GrowthBook` was saved as `...?q=growthbook`, and Google Drive links (mixed-case file IDs) broke entirely — the task hyperlink no longer pointed where the user intended.

## Fix

Extract the normalization into `normalizeChecklistTaskUrl`:
- prefix `https://` when no scheme is present,
- detect an existing scheme **case-insensitively**,
- but **preserve the URL's original casing**.

## Test

Adds `test/components/Settings/NewExperimentChecklistItem.test.ts` covering: casing preserved (incl. a Drive-style link), `https://` prefixed when missing, and case-insensitive scheme detection. Verified the helper produces the original casing across all four cases (the previous logic lower-cased every one).

Closes #6237


msamper (approved):
Thank you for the contribution! Made a tweak to lowercase the URL scheme when included while preserving casing for the rest of the URL, but otherwise lgtm

PR #6242, opened 2026-06-30
Skip inlining cloneDeep on saved-group-references-off refresh path

Kevin-Chant:
### Features and Changes

Builds on this branch's `cloneDeep` avoidance in `getFeatureDefinitionsResponse`. The current diff still takes a full-payload `cloneDeep` (plus a `recursiveWalk`) whenever `needsSavedGroupInlining` is true — but on the `buildSDKPayloadForConnection` path that inlining is already done upstream by `generateFeaturesPayload`, so the walk is a no-op and the clone is wasted. This covers the "saved group references off" connections the original PR description names but didn't actually optimize.

- **Skip the redundant inlining clone.** `buildSDKPayloadForConnection` hoists the resolved `savedGroupReferencesEnabled` into one local that drives *both* `generateFeaturesPayload`'s expansion and a new `featuresHaveInlinedSavedGroups` flag passed to `getFeatureDefinitionsResponse`. When features arrive already inlined, the downstream inlining pass and its `cloneDeep` are skipped. Net: on the refresh path, the features clone now fires only when secure-attribute hashing runs. Direct callers (e.g. tests) don't pass the flag, so their inlining behavior is unchanged.
- **Tie the two expansion conditions to one variable.** The upstream (`shouldExpandSavedGroups`) and downstream (`needsSavedGroupInlining`) checks weren't textually identical — they even handled `undefined` differently (`=== false` vs `!x`) — and only agreed because both were fed a pre-resolved boolean computed inline three times. Hoisting it removes that divergence structurally.
- **Consistency.** The experiments return spreads now key off `processedExperiments` to match the hashing block.
- **Mutation-safety test.** Added a test that builds a hashing connection and a non-hashing connection from one shared `rawData` (mirroring `refreshSDKPayloadCache`) and asserts they don't corrupt each other's saved groups. This guards the `clone -> cloneDeep` change in `applySavedGroupHashing`, which is the one load-bearing protection now that the upfront clones are gone — verified it fails against a shallow saved-group clone.

Note: `featuresHaveInlinedSavedGroups` still encodes the assumption that `generateFeaturesPayload` inlines exactly when references are off. Making that fully structural would mean changing the generator's return shape — heavier than this optimization warrants.

### Testing

- [x] `pnpm --filter back-end test sdk-payload-generation.test.ts features.test.ts` — 131 pass
- [x] `pnpm --filter back-end type-check` clean (pre-existing `sharp` errors only)
- [x] New mutation-safety test fails against a shallow `applySavedGroupHashing` clone, passes with `cloneDeep`

### DIFF
--- a/packages/back-end/src/services/features.ts
+++ b/packages/back-end/src/services/features.ts
@@ -1167,6 +1167,10 @@ export async function buildSDKPayloadForConnection(
     projectsMap = new Map(allProjects.map((p) => [p.id, p]));
   }
 
+  // Expands $inGroup -> $in whenever resolvedSavedGroupReferencesEnabled is
+  // false. getFeatureDefinitionsResponse relies on this via
+  // featuresHaveInlinedSavedGroups below; keep the two in lockstep if this
+  // expansion condition ever changes.
   const featureDefinitions = generateFeaturesPayload({
     features: filteredFeatures,
     environment,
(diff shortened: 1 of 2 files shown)


Kevin-Chant:
Closing this in favor of a separate PR that fully removes cloneDeep from this path

Issue #6258, opened 2026-07-01
[Bug] API Nulling of Experiment Targeting & Splits


pentium10:
REST API Bug Report: Silent Overwrite / Nulling of Experiment Targeting & Splits

## Overview & Bug Summary
This document outlines a critical issue discovered in the GrowthBook REST API (`v1.0.0`) where sending a standard `POST /api/v1/experiments/{id}` request to update analysis metrics silently overrides and wipes out existing user targeting rules and traffic splits, resetting them to system defaults (`{}` and `50/50` split). 

This happens because the API parser accepts the `phases` parameter in the request body but silently filters out `targetingCondition` and `trafficSplit` (which are ignored by the backend), resulting in a partial overwrite that nulls out these fields in the database.

---

### 1. 09:23 AM — The Initial API Run
* **Action:** An automated metrics attachment script was executed to sync goals, secondary metrics, and guardrails to the newly created hybrid onboarding experiments.
* **Mechanism:** The script fetched the current experiment schema using `GET /experiments/{id}`, updated the metrics arrays, and sent the entire experiment object back via `POST /experiments/{id}` (including the `phases` array verbatim).
* **Result:** The API returned `200 OK`, and the metrics were successfully attached.

### 2. 10:30 AM — Visual Check & Dashboard Anomaly
* **Observation:** During a routine check of the GrowthBook UI, we observed that user targeting had stopped working and the traffic allocation split had shifted.
* **Audit Log Analysis:** Checking the GrowthBook **Experiment Audit History** (captured in `image.png`) revealed that the API call at **9:23 AM** had deleted the path targeting rules (`path is any of (...) -> None`) and changed the traffic allocation from `90/10 split` to `50/50 split`. The targeting fields had been completely nulled out.

### 3. 11:21 AM — Attempted Programmatic Fix
* **Action:** We attempted to resolve the issue programmatically by sending a target fix script to restore the phase details.
* **Payload Configured:** We explicitly provided `trafficSplit` (weights `0.9` and `0.1`) and `targetingCondition` (path targeting MongoDB filters) inside the `phases` payload.
* **API Response:** The endpoint returned `200 OK`.
* **Result:** The verified response showed that the API **completely ignored** our targeting values. The splits remained `50/50` and the targeting condition remained `{}` (empty).

### 4. 11:39 AM — Discovery of GitHub Pull Request #5930
[...]

PR #6259, opened 2026-07-01
Accept experiment phases array from GET endpoint in POST body

Kevin-Chant:
### Features and Changes

Fixes the silent data-loss bug reported in #6258: fetching an experiment, editing one field, and POSTing it back would wipe phase targeting and traffic splits. The root cause is that the GET response and the create/update request use different field names for the same data, so echoing the response back dropped those fields and the phase was rebuilt with defaults.

Rather than change the endpoint's long-standing "a `phases` array is a full replacement" semantics (which merge-patch, RFC 7396, treats as correct for arrays), this makes the round-trip lossless by teaching the request to accept the GET-response field names as aliases:

- **Phase fields:** `targetingCondition` → `condition`, `trafficSplit` → `variationWeights` (reshaped from `[{ variationId, weight }]` to a weight array by variation index), `reasonForStopping` → `reason` (already accepted in the schema, now actually honored).
- **Variation field:** `variationId` → `id`. Without this, echoing GET variations back (which carry `variationId`, not `id`) regenerated variation ids, desyncing the phase `variations` envelope and sticky bucketing. Resolved before `validateVariationIds` runs so it doesn't assign a fresh id first.
- **The POST-only fields (`condition`/`variationWeights`/`reason`/`id`) take precedence** when both a field and its alias are set, and the phase POST-only fields are marked `deprecated` in the spec so the API converges on the response vocabulary.
- **Enabled namespaces now round-trip.** The GET response emitted a phase `namespace` only when enabled but omitted the `enabled` flag; posting it back through `toPhaseNamespaceValue` (`enabled ?? false`) then disabled it. The response now emits `enabled: true`. (thanks @ahdriel)

Applies to both `POST /api/v1/experiments` and `POST /api/v1/experiments/{id}`. This is purely additive — no existing caller's behavior changes; previously-ignored fields simply start working. Regenerated the OpenAPI spec.

:warning: This does not unify the top-level request/response shapes (the GET response nests analysis config under `settings` and results under `resultSummary`, and `winner` is a variation-ID string in the response but a numeric index in the request). Those mismatches are non-destructive — omitted top-level fields are left unchanged — so they're out of scope here.

### Testing

- [x] Round-trip test: an update sent with `targetingCondition`/`trafficSplit`/`reasonForStopping` applies them to the phase.
- [x] Variation round-trip test: an update sent with `variationId` (no `id`) keeps the ids instead of regenerating them.
- [x] Precedence test: when both a POST-only field and its alias are set, the POST-only field wins.
- [x] Serialization test: an enabled namespace serializes with `enabled: true`.
- [x] Existing experiments API tests (68 total) pass; back-end + shared type-check and lint clean.

### DIFF
--- a/packages/back-end/test/api/experiments.test.ts
+++ b/packages/back-end/test/api/experiments.test.ts
@@ -230,6 +230,39 @@ describe("experiments API", () => {
       });
     });
 
+    it("serializes an enabled namespace with enabled: true so it round-trips", async () => {
+      const experimentWithNamespace = {
+        ...experiment,
+        phases: [
+          {
+            name: "Main",
+            dateStarted: new Date("2024-01-01"),
+            dateEnded: null,
+            reason: "",
+            seed: "test-seed",
+            coverage: 1,
+            variationWeights: [0.5, 0.5],
+            condition: "",
+            savedGroups: [],
+            prerequisites: [],
+            namespace: { enabled: true, name: "ns_1", range: [0, 0.5] },
+          },
+        ],
+      };
+      (getExperimentById as jest.Mock).mockResolvedValue(
+        experimentWithNamespace,
+      );
+      const res = await request(app)
+        .get("/api/v1/experiments/exp_123")
+        .set("Authorization", "Bearer foo");
+
+      expect(res.status).toBe(200);
+      expect(res.body.experiment.phases[0].namespace).toMatchObject({
+        namespaceId: "ns_1",
+        enabled: true,
+      });
+    });
+
     it("returns experiment with draft status", async () => {
       const draftExperiment = { ...experiment, status: "draft" };
       (getExperimentById as jest.Mock).mockResolvedValue(draftExperiment);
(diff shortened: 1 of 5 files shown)


packages/back-end/test/api/experiments.test.ts
```diff
@@ -1534,6 +1534,74 @@ describe("experiments API", () => {
...
+
+      expect(res.status).toBe(200);
+      const updateCall = (updateExperiment as jest.Mock).mock.calls[0][0];
+      const phase = updateCall.changes.phases[0];
+      expect(phase.condition).toBe('{"path":{"$in":["/checkout"]}}');
+      expect(phase.variationWeights).toEqual([0.9, 0.1]);
+      expect(phase.coverage).toBe(0.8);
+      expect(phase.savedGroups).toEqual([{ match: "all", ids: ["grp_1"] }]);
+      expect(phase.seed).toBe("seed_123");
+      expect(phase.namespace).toEqual({
```
ahdriel: There's a bug with how we are serializing namespace that we should address as well:

When there's a enabled namespace from the GET response, toExperimentApiInterface emits:
```
namespace: {
  namespaceId,
  range,
  ranges,
}
```

with no enabled: true. Then toPhaseNamespaceValue() does:
```
const enabled = apiNs.enabled ?? false;
```

So if a client GETs an experiment and POSTs the phase back with that namespace object, it gets converted to enabled: false.

ahdriel (dismissed):
Changes look good to me! One additional fix for a roundtrip GET -> modify -> POST and we are good to go.

tzjames (approved):

PR #6264, opened 2026-07-01
fix(ramp-schedules): run scheduled snapshots for monitored ramps so they auto-advance

ahdriel:
### Features and Changes

Monitored ramps (ramp-up schedules with monitoring) never auto-advanced. A monitored ramp relies on its linked SafeRollout snapshot being refreshed on the configured cadence, and the ramp advance job only advances once a fresh snapshot exists. The scheduled snapshot job silently skipped monitored ramps, so no fresh snapshot ever arrived and the ramp held forever. Manual and API refresh worked, because they create the snapshot directly and bypass the job.

The cause is a guard in the shared SafeRollout snapshot job (`addSafeRolloutSnapshotJob`). It bailed for any SafeRollout without a matching `safe-rollout` feature rule. A monitored ramp uses a shadow SafeRollout that has no such rule, since its tracking key lives on the SafeRollout entity itself, so every scheduled refresh was dropped. The guard was correct for classic Safe Rollouts and was never updated when monitored ramps started reusing this job.

The fix scopes the feature-rule guard to classic Safe Rollouts. Ramp-linked SafeRollouts (identified by `rampScheduleId`) are no longer skipped. Their refresh eligibility is already governed by the `getAllSafeRolloutsToUpdate` query filter (`autoSnapshots` plus running status), kept in sync by `syncLinkedSafeRolloutForRampState`. The decision is extracted into a small pure predicate, `shouldSkipScheduledSafeRolloutSnapshot`.

### Testing

Unit tests for `shouldSkipScheduledSafeRolloutSnapshot` and `getSafeRolloutRuleFromFeature`, covering the ramp-linked path and the classic paths (enabled rule, disabled rule, missing rule, rule in a disabled environment).

### DIFF
--- a/packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
+++ b/packages/back-end/src/jobs/addSafeRolloutSnapshotJob.ts
@@ -4,7 +4,7 @@ import { getContextForAgendaJobByOrgId } from "back-end/src/services/organizatio
 import { logger } from "back-end/src/util/logger";
 import { getCollection } from "back-end/src/util/mongo.util";
 import { getFeature } from "back-end/src/models/FeatureModel";
-import { getSafeRolloutRuleFromFeature } from "back-end/src/routers/safe-rollout/safe-rollout.helper";
+import { shouldSkipScheduledSafeRolloutSnapshot } from "back-end/src/routers/safe-rollout/safe-rollout.helper";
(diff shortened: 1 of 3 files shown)


msamper (approved):
This looks correct to me, but I'd wait for someone more familiar with ramp ups to review

bryce-fitzsimons (approved):

PR #6270, opened 2026-07-02
Add Growthbook Client to the back-end

tzjames:
### Features and Changes

We now can use GrowthBook feature flags in our back-end code! This is based off of https://github.com/growthbook/growthbook/pull/3594.

Details:
- This is available only for logged-in routes, which means it applies mostly for our Cloud App.
  - This is by design to avoid creating a circular dependency for features/SDK initialization.
- Uses the same SDK Clients as the front-end (including the ID)
- Explicitly check for IS_CLOUD for initialization 
- Sets the url to be the referrer for front-end calls.  Api Calls will be the api url.
- Forwards the page_id, session_id, and device_id from the front-end and sets those.
- Also Forwards the url since browsers strip the path from referrers for cross-origin requests
- Adds the ip, url, and ua from request headers.

### Testing

Manual testing
- Call `req.gb?.isOn('feature-name')` and ensure it returns the appropriate value
  - You can add a Force Rule to ensure true/false based on your User Id on app.growthbook.io
- Log `req.gb` and observe how all features and attributes are properly set
- Have a clickhouse set up like in https://app.notion.com/p/growthbook/Dev-Managed-Data-Warehouse-251a857e74a080eda450d70045cb3770?assetsVersion=23.13.20260630.0707&clientBuildTarget=client
```

/** Feature keys used to smoke-test backend SDK tracking via /organization/definitions */
const TRACKING_SMOKE_TEST_FEATURE = "ff_2" satisfies keyof AppFeatures;
const TRACKING_SMOKE_TEST_EXPERIMENT_FEATURE =
  "gb-ax5-bandit" satisfies keyof AppFeatures;
const TRACKING_SMOKE_TEST_INLINE_EXPERIMENT_KEY =
  "backend-definitions-tracking-test";

export type GrowthBookTrackingSmokeTest = {
  feature: {
    key: string;
    on: boolean;
    value: unknown;
    source: string;
  };
  experimentFeature: {
    key: string;
    on: boolean;
    value: unknown;
    source: string;
    variationId: string | null;
  };
  inlineExperiment: {
    key: string;
    variationId: string;
    inExperiment: boolean;
  };
};

/**
 * Evaluate a feature and experiment to emit SDK tracking events to the ingestor.
 * Intended for manual verification while developing backend GrowthBook integration.
 */
export function runGrowthBookTrackingSmokeTest(
  gb: UserScopedGrowthBook<AppFeatures> | undefined,
): GrowthBookTrackingSmokeTest | null {
  if (!gb) return null;

  const feature = gb.evalFeature(TRACKING_SMOKE_TEST_FEATURE);
  const experimentFeature = gb.evalFeature(
    TRACKING_SMOKE_TEST_EXPERIMENT_FEATURE,
  );
  const inlineExperiment = gb.runInlineExperiment({
    key: TRACKING_SMOKE_TEST_INLINE_EXPERIMENT_KEY,
    variations: [false, true],
    weights: [0.5, 0.5],
  });

  gb.logEvent("manual-event-test-backend", { test: "test" });

  return {
    feature: {
      key: TRACKING_SMOKE_TEST_FEATURE,
      on: feature.on,
      value: feature.value,
      source: feature.source,
    },
    experimentFeature: {
      key: TRACKING_SMOKE_TEST_EXPERIMENT_FEATURE,
      on: experimentFeature.on,
      value: experimentFeature.value,
      source: experimentFeature.source,
      variationId: experimentFeature.experimentResult?.key ?? null,
    },
    inlineExperiment: {
      key: TRACKING_SMOKE_TEST_INLINE_EXPERIMENT_KEY,
      variationId: inlineExperiment.key,
      inExperiment: inlineExperiment.inExperiment,
    },
  };
}
```
- Then add `runGrowthBookTrackingSmokeTest` to the organizations/definitions endpoint.
- On http://localhost:3000/sql-explorer/ run the queries `SELECT * FROM experiment_views ORDER BY timestamp DESC;` , `SELECT * FROM feature_usage ORDER BY timestamp DESC;`, `SELECT * FROM events ORDER BY timestamp DESC;` and see full data.


tzjames:
@cursoragent review

tzjames:
@cursoragent review

tzjames:
@cursoragent review

packages/front-end/services/utils.tsx
```diff
@@ -289,6 +289,33 @@ function getOrGenerateSessionId() {
...
+): Record<string, string> {
+  if (typeof window === "undefined" || (!isCloud() && !isLocalhost())) {
+    return {};
+  }
+
+  const headers: Record<string, string> = {
+    "X-GB-Session-Id": getOrGenerateSessionId(),
+    "X-GB-Device-Id": getOrGenerateDeviceId(),
+    "X-GB-Page-Id": getOrGeneratePageId(),
+    "X-GB-Page-Url": window.location.href,
```
Kevin-Chant: How do you want to handle sensitive query params? The only one that comes to mind is `/invite/accept?key=...`, ~and granted that should only be logged after the key's been used, but is it worth stripping params before logging?~
I looked at it further and realized it would be logged before the user accepts: https://github.com/growthbook/growthbook/blob/ji/gb-backend/packages/front-end/pages/invitation.tsx#L24-L27
I'm not sure if there's anything relevant we'd be losing, but it seems safer to drop query params from the logging

packages/back-end/src/services/auth/index.ts
```diff
@@ -312,6 +322,79 @@ export async function processJWT(
         dateCreated: new Date(),
       });
     };
+
+    if (IS_CLOUD || IS_LOCALHOST) {
```
Kevin-Chant: Does IS_LOCALHOST only apply to dev boxes or will any self-hosted customers run into this? (I see it's defined by using `localhost` for the app host, but I don't know if there's any usage patterns that match that in the wild)

packages/front-end/services/utils.tsx
tzjames: good call.  That is what we were doing on the front-end tracking.  I have reused the same logic here.

packages/back-end/src/services/auth/index.ts
tzjames: I've gone ahead and removed `IS_LOCALHOST`.  I think devs now have no reason not to set `IS_CLOUD=true`, `IS_LOCALHOST=true` in general since with that they no longer need an SSO_CONFIG.

Kevin-Chant (approved):
LGTM, thanks for those fixes

PR #6274, opened 2026-07-02
Ty/cb sdk payload v2

teresayung:
### Features and Changes

Builds the contextual bandit (CB) **SDK payload**: back-end payload assembly, shared payload/versioning/validators, and supporting UI. The `sdk-js` runtime that *consumes* this payload (leaf selection, exposure result fields, tracking) landed separately in #6479 — this PR produces the payload that runtime reads.

**Payload shape (back-end)**

- CB leaf contexts are deduped into a top-level `contextualBandits` map in the payload (savedGroups-style), keyed by CB id. Rules only carry a pointer (`contextualBanditRef`), so a CB linked to F features costs one copy instead of F (`filterUsedContextualBandits`).
- CB rules omit `variations` and instead carry their variations under a new `contextualVariations` key. Older SDKs detect a running experiment by the presence of `variations`, so they skip the rule and fall through rather than mis-bucketing users into an even split; CB-capable SDKs read `contextualVariations` back into the experiment. (This replaces the earlier `isContextualBandit` boolean / `type` discriminator approach — both are gone; a CB rule is now identified by the presence of `contextualBanditRef`.)
- Each map entry (`ContextualBanditDefinition`) carries its `contexts` (`leafId` / `condition` / `weights`) and the `banditVersion` (training epoch).
- The map, `contextualBanditRef`, and `contextualVariations` are only emitted for SDKs declaring the `contextualBandits` capability. Encryption is supported via `encryptedContextualBandits`.
- Weight updates from bandit events refresh linked feature payloads (`refreshLinkedFeaturePayloads`). Affected features are derived by scanning all features for rules pointing at the CB (`contextualBanditRef`) rather than trusting the CB's `linkedFeatures` array, and the refresh now covers `start` / `stop` / `refresh` audit events.

**SDK versioning**

- The `contextualBandits` capability is declared at **v1.7.0** for the JavaScript, Node, and React SDK manifests, gating the new payload keys behind capability detection. Older SDKs (or any without the capability) receive neither the map nor the CB rule keys and fall back to marginal/aggregate weights (MAB behavior).

**Guardrails / instrumentation**

- `maxLeaves` capped at `MAX_CONTEXTUAL_BANDIT_LEAVES = 12` on the create and update API bodies (request validators only, so stored docs are never rejected on read); OpenAPI spec regenerated.
- `apiUpdateContextualBanditBody` is now strict (`z.strictObject`) — unsupported fields (e.g. `namespace`, which CBs intentionally don't support) fail loudly instead of being silently stripped.
- CB payload-size metrics (`sdk_payload.cb_bytes`, `sdk_payload.cb_fraction`) with a warn log above 512KB or >50% of the payload (`measureContextualBanditPayload` / `recordContextualBanditPayloadMetrics`).

**UI**

- Contextual bandit detail page: warning callout when a rule's attribute targeting overlaps with the bandit's contextual attributes (overlap can create unreachable variations).
- Start contextual bandit modal: hard blocker requiring at least one linked feature flag before a CB can start.

### Testing

```
pnpm --filter back-end test test/services/contextualBanditPayload.test.ts
pnpm --filter back-end test test/services/contextualBanditChanges.test.ts
pnpm --filter back-end test test/services/contextualBandits.test.ts
pnpm --filter shared test
```

- `contextualBanditPayload.test.ts` covers map dedup, capability gating, encryption, and size instrumentation.
- `contextualBanditChanges.test.ts` covers the weight-update refresh path (features found via `contextualBanditRef`, not the CB's `linkedFeatures`).
- `contextual-bandit-validators.test.ts` + `sdk-payload.test.ts` (shared) cover the `maxLeaves` cap, the strict update body, and the payload allow-list keys.
- Manual: link a running CB to a feature, hit the features endpoint with a CB-capable connection (SDK ≥ 1.7.0), and confirm the top-level `contextualBandits` map + `contextualBanditRef` + `contextualVariations` on the rule; non-capable connections should see none of these.

### Screenshots

Contextual bandit detail page conflicting-attributes warning + start-modal linked-feature blocker (only UI changes; attach if useful).


packages/back-end/src/services/features.ts
```diff
@@ -1672,7 +1775,7 @@ export async function evaluateAllFeatures({
       allFeatures[f.id] = {
         ...f,
         project: f.project,
-      } as FeatureDefinition;
```
teresayung, line 1701: teresa - look into this - why was it cast as unknown?

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -48,8 +48,22 @@ export type FeatureRule<T = any> = {
     experiment: Experiment<T>;
     result: Result<T>;
   }>;
+  type?: "standard" | "multi-armed-bandit" | "contextual-bandit";
```
teresayung: extract this to ExperimentTypes

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -48,8 +48,22 @@ export type FeatureRule<T = any> = {
...
+  contextualBanditRef?: string;
 };
 
+export type ContextualBanditData = {
```
teresayung: misnomer - not clear this is the definition used for cb

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -152,13 +166,23 @@ export interface Result<T> {
   hashValue: string;
   featureId: string | null;
   stickyBucketUsed?: boolean;
+  leafId?: number;
+  variationWeights?: number[];
```
teresayung, line 173: Look into changing this so we don't get confused on what variation maps to correct weights. Record<variationId, weight> <-- change format to this, or {1: 20%}

teresayung: essentially changing ContextualBanditInfo

teresayung: nevermind

packages/sdk-js/src/core.ts
```diff
@@ -100,7 +102,7 @@ function onExperimentViewed(
   }
   if (ctx.user.trackingCallback) {
     const cb = ctx.user.trackingCallback;
-    calls.push(safeCall(() => cb(experiment, result)));
+    calls.push(safeCall(() => cb(experiment, result, getAttributes(ctx))));
```
teresayung: use this instead
calls.push(safeCall(() => cb(experiment, result, ctx.user)));

teresayung: keep it consistent

packages/sdk-js/src/core.ts
```diff
@@ -307,6 +309,7 @@ export function evalFeature<V = unknown>(
               ctx.global.saveDeferredTrack({
                 experiment: t.experiment,
                 result: t.result,
+                attributes: getAttributes(ctx),
```
teresayung: saveDeferredTrack needs to be serialized - so maybe a leaner UserContext might make sense that only picks up the attributes

user: {attributes: ctx.user.attributes} - make a helper function

packages/sdk-js/src/core.ts
```diff
@@ -350,8 +353,38 @@ export function evalFeature<V = unknown>(
       if (rule.filters) exp.filters = rule.filters;
       if (rule.condition) exp.condition = rule.condition;
 
+      let cbInfo: ContextualBanditInfo | undefined;
+      if (rule.type === "contextual-bandit" && rule.contextualBanditRef) {
+        const cbData = ctx.global.contextualBandits?.[rule.contextualBanditRef];
```
teresayung: cbData -> rename cbDefinition

packages/sdk-js/src/core.ts
```diff
@@ -350,8 +353,38 @@ export function evalFeature<V = unknown>(
...
+        if (!cbData) {
```
teresayung: runExperiment should handle this logic rather than evalFeature

runExperiment - all experiment specific logic
evalFeature is everything else that's not experiment

teresayung: create a buildContextualBanditExperiment instead

packages/sdk-js/src/core.ts
```diff
@@ -350,8 +353,38 @@ export function evalFeature<V = unknown>(
...
+            leafId: leaf.leafId,
+            variationWeights: leaf.weights,
+            banditVersion: cbData.banditVersion,
+          };
+        }
+      }
+
       // Only return a value if the user is part of the experiment
-      const { result } = runExperiment(exp, id, ctx);
+      const { result } = runExperiment(exp, id, ctx, cbInfo);
```
teresayung: put it in experiment instead of this top level

packages/sdk-js/src/core.ts
```diff
@@ -818,6 +854,28 @@ function getAttributes(ctx: EvalContext) {
   };
 }
 
+function getContextualBanditLeaf(
```
teresayung, line 850: We need to make sure that we are sending down experiment.disableStickyBucketing = true - make sure it's set in the payload

packages/sdk-js/src/types/growthbook.ts
teresayung: Remove type - use `isContextualBandit` instead so it gets stripped properly

packages/sdk-js/src/core.ts
teresayung: make sure this doesn't crash (try/catch)

packages/sdk-js/src/core.ts
```diff
@@ -875,6 +933,7 @@ export function getExperimentResult<T>(
   featureId: string | null,
   bucket?: number,
   stickyBucketUsed?: boolean,
+  cbInfo?: ContextualBanditInfo,
```
teresayung: likely remove

packages/sdk-js/src/core.ts
```diff
@@ -910,6 +969,12 @@ export function getExperimentResult<T>(
   if (meta.name) res.name = meta.name;
   if (bucket !== undefined) res.bucket = bucket;
   if (meta.passthrough) res.passthrough = meta.passthrough;
+  if (cbInfo && inExperiment) {
```
teresayung: get from exp instead of cbinfo

teresayung:
@greptileai review this

packages/back-end/src/services/features.ts
lukesonnet: did we resolve this?

packages/back-end/src/util/features.ts
```diff
@@ -1067,12 +1067,22 @@ export function getFeatureDefinition({
             rule.seed = cb.seed;
           }
           rule.hashVersion = 2;
+          // Contextual bandit weights (leaf and aggregate) are retrained each
+          // epoch, so a sticky-bucket assignment would lock users to stale
+          // weights. Disable it for all consumers, not just CB-capable ones —
+          // the aggregate-weight (MAB) fallback reweights over time too.
```
lukesonnet, line 1073: This comment isn't quite right. Sticky bucketing doesn't lock you into stale weights, it just locks you into your existing variation. We could remove it or replace it with "contextual bandits do not currently use sticky bucketing"

lukesonnet (approved):
Approved with two nits!

teresayung:
I'll update the nits in my following PR!

PR #6275, opened 2026-07-02
chore(experiments): Add tracking around adding variations and adding targeting

msamper:
### Features and Changes

Add tracking events for:

- Adding variations from the experiment overview page
- Adding targeting from the experiment overview page

### DIFF
--- a/packages/front-end/components/Experiment/EditTargetingModal.tsx
+++ b/packages/front-end/components/Experiment/EditTargetingModal.tsx
@@ -1,4 +1,5 @@
 import { ExperimentInterfaceStringDates } from "shared/types/experiment";
+import { hasAttributeCondition } from "shared/experiments";
 import { Box } from "@radix-ui/themes";
 import { useFeatureIsOn } from "@growthbook/growthbook-react";
 import { useAttributeSchema, useEnvironments } from "@/services/features";
@@ -13,6 +14,7 @@ import Switch from "@/ui/Switch";
 import useOrgSettings from "@/hooks/useOrgSettings";
 import ModalStandard from "@/ui/Modal/Patterns/ModalStandard";
 import Text from "@/ui/Text";
+import track from "@/services/track";
 import useSDKConnections from "@/hooks/useSDKConnections";
 import SDKCapabilityWarning from "@/components/Features/SDKCapabilityWarning";
 import HashVersionSelector, {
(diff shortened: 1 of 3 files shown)


lukesonnet (approved):

packages/front-end/components/Experiment/EditTargetingModal.tsx
```diff
@@ -92,12 +118,15 @@ export default function EditTargetingModal({
...
       <ModalStandard
-        trackingEventModalType=""
+        trackingEventModalType="edit-targeting-modal"
         open={true}
         close={close}
         header="Edit Targeting"
         ctaEnabled={canSubmit}
-        submit={onSubmit(mutate, "targeting")}
+        submit={async () => {
+          await onSubmit(mutate, "targeting")();
```
lukesonnet, line 127: Does this await make this submit feel slower?

packages/front-end/components/Experiment/EditVariationsForm.tsx
```diff
@@ -95,6 +95,16 @@ const EditVariationsForm: FC<{
         });
         mutate();
         track("edited-variations");
+
+        const numVariationsAdded =
+          data.variations.length - experiment.variations.length;
```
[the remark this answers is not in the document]
lukesonnet: small nit, but might as well

packages/front-end/components/Experiment/EditTargetingModal.tsx
msamper: No the Modal component already awaits the submit prop under the hood (so it was being awaited before) and if we don't await here, errors won't propagate up to the modal and we could close the modal when an error was actually returned

msamper:
> ### Issue 1 of 2
> packages/front-end/components/Experiment/EditTargetingModal.tsx:234-245
> **Tracking gap in MakeChangesFlow path**
> 
> `trackAddedTargeting` is defined and called only in the `safeToEdit` branch. When `safeToEdit` is `false`, the component renders `MakeChangesFlow`, which calls `onSubmit(mutate, scope)()` directly — there is no call to `trackAddedTargeting`. If users add targeting conditions through the `MakeChangesFlow` path (e.g., on running experiments), those additions will not produce an "Added targeting" event.

This is fine. I'm mainly interested in tracking for draft experiments not running experiments

PR #6276, opened 2026-07-02
feat(experiments): auto-select EAQ for managed warehouse in simple create flow

msamper:
### Features and Changes
In the simple new-experiment flow (`SimpleNewExperimentForm`), the Experiment Assignment Query (EAQ) is auto-selected via `getAutoExposureQueryId`. That helper relied on `settings.userIdTypes[].attributes`, which **managed warehouses** (`growthbook_clickhouse`) never populate — so the EAQ was never auto-selected, even though managed warehouses always have exactly one exposure query per identifier.
This resolves the EAQ from the attribute → identifier-column mapping stored on the datasource instead.

### Dependencies
None

### Testing
- `pnpm --filter shared test managedWarehouse` — 33 passing
- `pnpm --filter front-end test SimpleNewExperimentForm` — 20 passing
- type-check (shared + front-end) and lint clean


packages/shared/src/util/managedWarehouse.ts
```diff
@@ -369,6 +369,61 @@ export function getManagedWarehouseUserIdTypeSettings(
...
+// parseAttributes in the growthbook-tracking SDK plugin: the `user_id` attribute
+// maps to the `user_id` column; `device_id`, `anonymous_id`, and `id` all map to
+// the `device_id` column. Any other attribute is a custom identifier whose column
+// name equals the attribute property.
+const BUILTIN_ATTRIBUTE_TO_IDENTIFIER: Record<string, string> = {
+  user_id: "user_id",
+  device_id: "device_id",
+  anonymous_id: "device_id",
+  id: "device_id",
+};
```
msamper, line 382: Ideally this should pull from a shared const but I don't want to tweak sdk plugin code in this PR

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
```diff
@@ -107,6 +108,17 @@ export function getAutoExposureQueryId({
   }
 
   if (exposureQueries.length === 1) return exposureQueries[0].id;
+
+  // Managed warehouses don't populate userIdType.attributes links, so the generic
+  // lookup below can't resolve the assignment query. Map the hash attribute to its
+  // exposure query directly instead.
+  if (datasource?.type === "growthbook_clickhouse") {
+    return getManagedWarehouseExposureQueryIdForAttribute({
```
ahdriel, line 116: nit: does not need to be in this PR, but ideally we would have this `get exposure query id for attribute` abstracted away in a helper that know how to handle the different datasources

basically meaning that we should move this function away from this component into a more reusable spot

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
```diff
@@ -356,10 +368,11 @@ const SimpleNewExperimentForm: FC<SimpleNewExperimentFormProps> = ({
       project,
       templateDatasource: data.datasource || "",
     });
+    const selectedDatasource = datasource
+      ? getDatasourceById(datasource)
```
ahdriel: suggestion: let's rename `datasource` to `datasourceId`

otherwise `selectedDatasource` being the full object but `datasource` being just the id will be confusing in the future

ahdriel (approved):

packages/front-end/components/Experiment/SimpleNewExperimentForm.tsx
msamper: agreed

PR #6280, opened 2026-07-04
Add computed column support

nhat-growthbook:
### Features and Changes

Adds support for **computed (derived) columns** in Fact Table Metrics. Users can now define metric-scoped columns built from arithmetic/concatenation of existing columns and use them as the metric value, in row filters, and in the user filter — without having to add extra physical columns to the fact table.

- New **Computed Columns** section inside "Participation Event" (Create/Edit Fact Table Metric), gated behind a toggle under the Fact Table selector. Toggling it off with existing definitions prompts a confirmation before discarding them.
- **Numeric computed columns** are a sum-of-products: terms joined by `+`/`-`, each term a group of operands joined by `*`/`/`. Operands can be a column **or a numeric literal** (e.g. `amount / 100`, `price * 1.08`). Options include:
  - **Treat nulls as 0** → wraps column operands in `COALESCE(col, 0)`
  - **Rounding** → round / floor / ceiling with optional decimal places
  - Division is guarded with `NULLIF(divisor, 0)`
- **String computed columns** concatenate columns and text literals (usable in row filters / slices only, not as a numeric value).
- Computed columns are stored on the metric's `numerator`/`denominator` `ColumnRef` (new `computedColumns` field) and referenced elsewhere via the `$$computed:<id>` marker — the same pseudo-column convention as `$$count`.
- The **Live SQL Preview** expands computed columns (e.g. `SUM((price * quantity))`).
- SQL generation is dialect-aware: new `round` and `concat` methods on `SqlDialect` with correct overrides for Postgres (`ROUND` numeric cast), MSSQL (`CEILING`), ClickHouse (lowercase fns), and Redshift (multi-arg concat via `||`).

**External API usage:** `POST`/`PUT /api/v1/fact-metrics` now accept an optional `computedColumns` array on `numerator`/`denominator`, and the field is returned on the `FactMetric` response. Reference a computed column from `column`, `aggregateFilterColumn`, or a row filter's `column` using `$$computed:<id>`.

### Endpoint

- `POST /api/v1/fact-metrics` — `numerator`/`denominator` now accept optional `computedColumns`
- `PUT /api/v1/fact-metrics/:id` — same
- `GET /api/v1/fact-metrics` and `GET /api/v1/fact-metrics/:id` — responses now include `computedColumns` on `numerator`/`denominator`

(OpenAPI spec regenerated — `generated/spec.yaml`.)

### Dependencies

None.

### Testing

1. **Unit tests** — `pnpm --filter shared test experiments.test.ts` (new `buildComputedColumnSQL` cases cover sum-of-products precedence, `NULLIF` division guard, numeric literals, `COALESCE`, rounding modes, string concat, and computed-column resolution inside a row filter).
2. **UI** — Open **Create Fact Table Metric** → *Participation Event*:
   - Toggle **Computed Columns** on; add a numeric column (e.g. `price * quantity`), confirm the **Value** dropdown, **Row Filter**, and **User Filter** all list it and the **Live SQL Preview** expands it.
   - Add a string (concat) computed column and confirm it appears only in the Row Filter.
   - Toggle the switch **off** with definitions present → confirm the warning dialog appears and clears definitions (and resets any references) on confirm.
3. **API** — `POST /api/v1/fact-metrics` with a `numerator.computedColumns` entry and a `numerator.column` of `$$computed:<id>`; confirm it saves and is returned by `GET`.
4. `pnpm type-check` and `pnpm lint` pass across shared / back-end / front-end.

### DIFF
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -1,0 +1,11 @@
+{
+  "version": "0.0.1",
+  "configurations": [
+    {
+      "name": "front-end",
+      "runtimeExecutable": "pnpm",
+      "runtimeArgs": ["--filter", "front-end", "dev"],
+      "port": 3000
+    }
+  ]
+}
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -20514,6 +20514,73 @@ paths:
                                         - type
                                         - value
                                       additionalProperties: false
+                              operations:
+                                type: array
+                                items:
+                                  anyOf:
+                                    - type: object
+                                      properties:
+                                        type:
+                                          type: string
+                                          const: replace
+                                        find:
+                                          type: string
+                                        replaceWith:
+                                          type: string
+                                      required:
+                                        - type
+                                        - find
+                                        - replaceWith
+                                      additionalProperties: false
+                                    - type: object
+                                      properties:
+                                        type:
+                                          type: string
+                                          const: regexpReplace
+                                        pattern:
+                                          type: string
+                                        replaceWith:
+                                          type: string
+                                      required:
+                                        - type
+                                        - pattern
+                                        - replaceWith
+                                      additionalProperties: false
+                                    - type: object
+                                      properties:
+                                        type:
+                                          type: string
+                                          const: regexpExtract
+                                        pattern:
+                                          type: string
+                                      required:
+                                        - type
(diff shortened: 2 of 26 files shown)


nhat-growthbook:
Change direction to this https://github.com/growthbook/growthbook/pull/6364

PR #6281, opened 2026-07-05
fix: sanitize APP_ORIGIN to resolve SSO trailing slash bug

### LINKED ISSUE #6261: [Bug] SSO login fails when the configured base URL contains a trailing slash.
### Summary

SSO login fails when the configured base URL contains a trailing slash.

In this case, the issue was caused by a trailing / in the base URL. Removing the trailing slash resolved the SSO flow, but the presence of a trailing slash should not cause the entire SSO connection/login flow to fail.

#### Expected Behavior

The SSO flow should handle base URLs with or without a trailing slash.

For example, both of the following should be treated safely/equivalently:

https://example.com
https://example.com/

If a trailing slash is not supported, the UI should validate the value when saving the SSO configuration and show a clear error before users attempt to log in.

#### Current Behavior

When the base URL includes a trailing /, the SSO flow fails with an “Unknown SSO connection” error.

The error does not clearly indicate that the base URL formatting is the issue, making this difficult to debug. Removing the trailing slash resolves the problem.

Impact

This can block users from logging in via SSO due to a small formatting issue in the base URL.

It also creates a poor debugging experience because the error message points to an unknown SSO connection rather than highlighting the invalid or unsupported URL format.

AnkittChauhan:
### Features and Changes

Added defensive sanitization to strip trailing slashes from the `APP_ORIGIN` and `API_HOST` environment variables globally upon initialization in both the backend (`secrets.ts`, `licenseUtil.ts`) and frontend (`init.ts`). 

Previously, if a user accidentally configured `APP_ORIGIN` with a trailing slash, the SSO flow would construct a malformed callback URL (`//oauth/callback`). This double-slash path caused browsers to drop the `SSO_CONNECTION_ID` cookie, resulting in an "Unknown SSO connection" error when processing the OAuth callback. Sanitizing the origin URLs at the root level ensures a normalized URL is used for all downstream SSO flows, emails, and integrations, completely preventing this bug.

- Closes https://github.com/growthbook/growthbook/issues/6261

### Testing

1. In local `.env.local` files, intentionally configure `APP_ORIGIN` with a trailing slash (e.g. `APP_ORIGIN="http://localhost:3000/"`).
2. Ensure `IS_CLOUD=true` to enable SSO routes.
3. Navigate to `/oauth/lookup` and initiate a dummy SSO login flow.
4. Verify that the generated callback URL does not contain double slashes and that you no longer receive the "No authorization token was found" / "Unknown SSO connection" error upon submission.


AnkittChauhan:
hi @Kevin-Chant , @lukesonnet , @tzjames, @bryce-fitzsimons : can you review this PR.

jdorn (approved):

PR #6285, opened 2026-07-06
Perf: Fetch only most recent published revision and batch experiment fetches when finding linked experiments

tzjames:
## Summary
- The fire-and-forget linkage sync that runs after any feature revision rule change (`syncFeatureExperimentLinkages` / `syncFeatureContextualBanditLinkages`) looped over every referenced experiment/contextual-bandit ID one at a time, awaiting each DB round-trip sequentially. Batched both loops with the existing `promiseAllChunks` helper (concurrency of 10) instead of a plain `for...of` with `await` inside — same reads/writes, same logic per ID, just bounded-concurrency execution.
- The revision data feeding that sync was also being over-fetched: `getNonDiscardedRevisionSummaries` (now `getLinkageSyncRevisionSummaries`) pulled *every* non-discarded revision for a feature, but the sync functions only ever use open drafts plus the single latest published revision. On a feature with a long publish history this meant fetching and holding thousands of superseded published revisions in memory on every rule change, for no behavioral benefit — one production feature currently has ~2,074 non-discarded revisions (nearly all old published history) despite having zero open drafts. Rewrote the query to fetch only `ACTIVE_DRAFT_STATUSES` revisions plus a single `findOne` for the latest published revision.
- While making that change, found 3 more call sites (`updateRevision`, `reopenRevision`, `discardRevision`) duplicating the same over-fetching query inline — consolidated all of them (plus one external caller in `experiments.ts`) onto the shared `getLinkageSyncRevisionSummaries` helper.
- `getLinkageSyncRevisionSummaries` now returns `{ openDrafts, liveRevision }` pre-split instead of a flattened array, since both sync functions immediately re-derived that same split via filter+sort. Passing it pre-split removes the redundant work and closes off a footgun: a future caller can no longer silently reintroduce the over-fetch bug by passing an unfiltered revision list, since there's no re-filtering step left to mask it.
- `OPEN_DRAFT_STATUSES` (a hand-duplicated literal status list in both sync files) is now gone entirely — status classification happens once, upstream, in `getLinkageSyncRevisionSummaries`, sourced from the canonical `ACTIVE_DRAFT_STATUSES`.

## Why this is safe
Both sync functions previously filtered their input down to open-draft statuses plus the single latest published revision before doing anything with it. Pre-filtering (and now pre-splitting) at the query level produces the exact same effective input set — it just avoids transferring/holding/re-deriving the dead historical revisions used to compute it. No behavior change, just less wasted work.

## Test plan
- [x] `eslint` passes for the changed files with no errors
- [x] Manually verify on a feature with many draft revisions that experiment/contextual-bandit linkage fields (`linkedFeatures`, `pendingFeatureDrafts`) still reconcile correctly after adding/removing/reopening/discarding a rule

### DIFF
--- a/packages/back-end/src/util/featureContextualBanditSync.ts
+++ b/packages/back-end/src/util/featureContextualBanditSync.ts
@@ -1,15 +1,9 @@
 import { FeatureRevisionInterface } from "shared/types/feature-revision";
-import {
-  ACTIVE_DRAFT_STATUSES,
-  ContextualBanditRefRule,
-  RevisionStatus,
-} from "shared/validators";
+import { ContextualBanditRefRule } from "shared/validators";
 import { ReqContext } from "back-end/types/request";
 import { ApiReqContext } from "back-end/types/api";
 import { logger } from "back-end/src/util/logger";
 import { promiseAllChunks } from "back-end/src/util/promise";
-
-const OPEN_DRAFT_STATUSES: Set<RevisionStatus> = new Set(ACTIVE_DRAFT_STATUSES);
 
 function getContextualBanditIdsFromRules(
   rules: FeatureRevisionInterface["rules"] | unknown,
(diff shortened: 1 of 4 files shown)


packages/back-end/src/util/featureContextualBanditSync.ts
lukesonnet: nit: Might be nice to have live vs. open drafts passed in separately since we just package them together in the new `getLinkageSyncRevisionSummaries` and then unpackage them here?

lukesonnet (approved):
Some of the new comments are pretty "we used to do it this way but it was wasteful" which is pretty claude-y, but eh it doesn't really matter.

Kevin-Chant (approved):

PR #6290, opened 2026-07-06
feat(back-end): cloud-only GrowthBook SDK client (streaming, fail-open)

gazzdingo:
### Features and Changes

Dedicated server-side GrowthBook SDK client so the back-end can read GrowthBook's own feature flags. Re-implements the pattern from #3594 on `main`, with the fixes from the back-end SDK security audit. First consumer (Pricing Phase 1 limits) comes in a follow-up PR.

- `services/growthbook.ts` — cloud-only `GrowthBookClient` singleton: SSE streaming, 3s init timeout, fail-soft init at boot + cleanup on shutdown (`server.ts`). Disabled unless `IS_CLOUD` and a client key are set, so consumers always fall back to in-app defaults (fail-open).
- New env vars `GROWTHBOOK_SDK_CLIENT_KEY` / `GROWTHBOOK_SDK_API_HOST` — a dedicated server connection (deliberately not the front-end's key; no dev mode).
- `getTrustedOrgAttributes(org)` — org-level targeting attributes derived from server-side data only, mirroring the front-end's attribute names (hashed `organizationId`, effective `accountPlan`, `orgDateCreated`, …) so one flag config can target both apps.
- `types/app-features.d.ts` — typed `AppFeatures` for the back-end.
- `eventsource@4` polyfill for SSE (named export; Node 24 ships no global `EventSource`).

Security constraints from the audit, documented in the code: enforcement-gating flags are evaluated **only** against trusted server-derived attributes — never request context, because the SDK honors `?flag=<n>` query-string variation overrides whenever the eval context carries a URL, independent of dev mode — and flags are never read on the SDK-payload-generation path (keeps a circular dependency impossible).

Out of scope on purpose: per-request `req.gb` / per-user flags with sticky buckets — that remains #3594's scope, and enforcement reads must not go through it.

- Closes — N/A (infrastructure for the Pricing Phase 1 rollout)

### Dependencies

None to deploy — the SDK stays off until the env vars are set. Activating on cloud requires creating a dedicated SDK connection in GrowthBook's own account and setting `GROWTHBOOK_SDK_CLIENT_KEY`.

### Testing

- Without a key: boot logs `Back-end GrowthBook client disabled…`, `getGrowthBookClient()` returns `null` (fail-open path).
- With a key + `IS_CLOUD`: boot logs `Back-end GrowthBook client initialized` and flag changes propagate live via SSE; shutdown logs the client destroyed.
- `type-check` + ESLint clean. No runtime consumers yet — the first flag read lands with the pricing PR.

### Screenshots

N/A (back-end only)


Kevin-Chant:
@gazzdingo see #6270 which I just approved

PR #6293, opened 2026-07-07
Apply enum-constrained list attributes via the REST API

Kevin-Chant:
### Features and Changes

Extends #6252 (enum-constrained list attributes) to the external attributes REST API. The `enum` (Allowed Values) field already flowed through the API, but it wasn't documented as the allowed-values concept and the API didn't mirror the attribute editor's normalization.

- Normalize `enum` in the shared `POST`/`PUT` attribute validation: keep it only for the `enum` datatype and array datatypes (`string[]`/`number[]`/`secureString[]`), and clear it for any other datatype — matching the UI. On update this also clears a stale allowed-values list when converting to an unsupported datatype.
- Document the `enum` field on the OpenAPI validators and regenerate the spec.

### Testing

- [x] `POST /api/v1/attributes` with `datatype: "string[]"` and `enum: "admin,editor,viewer"` -> allowed values are kept
- [x] `POST` with `datatype: "boolean"` and an `enum` -> `enum` is cleared to `""`
- [x] `PUT` an existing `enum` attribute to `datatype: "string"` -> stale allowed values are cleared
- [x] `pnpm --filter back-end test test/api/attributes.test.ts` (19 passing, incl. the 3 new cases)

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -15804,6 +15804,11 @@ paths:
                   description: Shall the attribute be hashed
                   type: boolean
                 enum:
+                  description: >-
+                    Comma-separated list of allowed values. Required for the
+                    'enum' datatype. For array datatypes (string[], number[],
+                    secureString[]) it optionally restricts the list to these
+                    values. Ignored for all other datatypes.
                   type: string
                 format:
                   description: The attribute's format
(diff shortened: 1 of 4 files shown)


Kevin-Chant:
@cursoragent review

tzjames (approved):

PR #6297, opened 2026-07-07
Add environment displayName to support renaming environments

tzjames:
## Summary
- Adds an optional `displayName` field to environments. The `id` stays immutable (it's used as a reference key throughout features, SDK connections, permissions, Slack integrations, etc.), but `displayName` is a purely cosmetic user-facing label shown in the UI wherever an environment's name is displayed, falling back to `id` when unset.
- Exposed via both the internal app (Settings → Environments) and the external REST API (`POST`/`PUT`/`GET /api/v1/environments`).

## Test plan
- [x] `pnpm type-check` clean across `shared`, `back-end`, `front-end`
- [x] `pnpm lint` clean on all changed files
- [x] `pnpm --filter back-end generate-openapi` regenerated and committed `generated/spec.yaml`
- [x] Manually verified in the browser: set a display name on an environment and confirmed it renders correctly (with `id` falling back when unset) in:
  - Environments settings list + edit modal
  - Feature list column headers
  - Feature Overview "Enabled Environments" row
  - Feature Rules tab bar and rule environment-scope badges
  - Team Members permission table
- [x] Verified round-trip through the external API validators (create/update/list all correctly handle `displayName`)

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -15615,6 +15615,11 @@ paths:
                 description:
                   description: The description of the new environment
                   type: string
+                displayName:
+                  description: >-
+                    A user-facing name for the environment. Falls back to the id
+                    when not set.
+                  type: string
                 toggleOnList:
                   description: Show on feature list page
                   type: boolean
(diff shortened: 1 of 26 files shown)


tzjames:
Going to put this on hold. We might try the approach of being able to archive environments instead.

PR #6300, opened 2026-07-07
docs: Update Helm Chart URL

ahdriel:
### Features and Changes

We have been managing and publishing our own Helm Chart for a bit now, but we have a outdated link.
So this fixes the link to point to the latest helm chart in our GitHub repository.


jdorn:
Instead of pointing to our github / ghcr, I think we want to point to our official AritfactHub listing instead: https://artifacthub.io/packages/helm/growthbook/growthbook

PR #6302, opened 2026-07-07
update Contextual Bandit model

lukebrawleysmith:
### Features and Changes

This PR replaces our CB model.  The current CB model uses one-hot encoding and has the ability to create leaves only from singleton sets (e.g., country = {US} vs country not in {US}).  The new approach permits  multiple levels of an attribute for leaf   (e.g., country = {US, CA} vs country not in {US, CA}).   If there are 15 or fewer levels of the attribute (e.g., countries), then we can find the exact optimal split that minimizes Sum of Squared Error.  If there are greater than 15 levels, then we resort to K-Means, which finds a local minimum for SSE.


packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -15,10 +16,25 @@ import {
...
+import { SampleMeanStatistic, ProportionStatistic } from "./statistics";
 
 const COMBINED_CONTEXT_ATTRIBUTE_VALUE = "Combined";
 
+/**
+ * Tree-growth split strategy.
+ *  - `"kmeans"` (default): each split groups an attribute's categories into two
+ *    sets via weighted k-means (`country in (US, CA)` vs not), porting gbstats
+ *    `UpdateWeightsContextualTreeKMeans`.
+ *  - `"onehot"`: greedy one-hot splits (`country == US` vs not), porting gbstats
```
lukesonnet: A lot of these comments are AI comments about migrating code that never existed. Let's clean them up.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -28,6 +44,8 @@ export type ContextualBanditWeightsInput = {
   metricSettings: MetricSettingsForStatsEngine;
   analysisWeights: number[];
   rows: ExperimentMetricQueryResponseRows;
+  /** Defaults to `"kmeans"`. */
+  splitStrategy?: ContextualBanditSplitStrategy;
```
lukesonnet: Lets make it required.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -107,123 +125,62 @@ function addArms(a: ArmColumns, b: ArmColumns): ArmColumns {
...
-    case "mean":
-    case "mean_ra": {
-      if (metric.main_metric_type === "binomial") {
-        const stat = forBandit
-          ? sampleMeanStat(arm.main_sum, arm.main_sum, n)
-          : proportionStat(arm.main_sum, n);
-        return {
+  let stat: SampleMeanStatistic | ProportionStatistic;
+  if (metric.main_metric_type === "binomial") {
+    stat = forBandit
```
lukesonnet: Let's just assume all of this code is only for bandits and remove checks like this.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -442,8 +441,177 @@ function buildTree(
...
+
+/**
+ * Greedy SSE regression tree up to `maxLeaves` where each split groups one
+ * attribute's categories into two sets via weighted k-means (porting gbstats
+ * `UpdateWeightsContextualTreeKMeans`). This admits multi-category splits like
+ * `country in (US, CA)` vs not, rather than only `country == US` vs not.
+ *
+ * A split is taken only when the best available (non-degenerate) binary
+ * category partition strictly reduces total SSE; there is no minimum-users-
+ * per-leaf guard. For attributes with at most `MAX_EXHAUSTIVE_CATEGORIES`
```
lukesonnet: Won't this always split to max n of leaves, then? Do we still have a min users count on the setting of the weights via `MIN_UNITS_PER_VARIATION`? What happens if we create a tiny leaf but that `MIN_UNITS_PER_VARIATION` is not met?

packages/stats-ts/src/multivariateKMeans.ts
```diff
@@ -0,0 +1,298 @@
...
+        let bestCluster = from;
+        let bestDelta = 0;
+        let bestToWith: number[] = [];
+        let bestToWithSse = 0;
+        for (let to = 0; to < k; to++) {
+          if (to === from) continue;
+          const toWith = [...clusters[to], i];
+          const toWithSse = sumOfSquaresError(groupOf(toWith));
+          const delta = removeDelta + (toWithSse - clusterSse[to]);
+          if (delta < bestDelta - 1e-12) {
```
lukesonnet: Should the convergence criteria be a function of the total SSE be safer?

packages/shared/src/validators/exposure-query-targeting-attribute-columns.ts
```diff
@@ -18,16 +18,34 @@ export const CONTEXTUAL_BANDIT_EAQ_VARIATION_WEIGHTS_COLUMN =
...
+ * rather than failing the contextual bandit run.
+ */
+export function queryHasContextualBanditSrmColumns(
+  query: string | undefined,
+): boolean {
+  if (!query) {
+    return false;
+  }
+  return CONTEXTUAL_BANDIT_SRM_REQUIRED_COLUMNS.every((col) =>
+    new RegExp(`\\b${col}\\b`).test(query),
```
lukesonnet, line 38: I don't love us doing this instead of validating on CBAQ save, but that's ok, we can change this later.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -1,3 +1,4 @@
+/** TypeScript port of the gbstats contextual-bandit weight pipeline. */
```
lukesonnet: This comment is incorrect.

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: done

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: done

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: Yes, it will always split to the maximum number of leaves.  There is not a minimum users count for creating a new leaf.  If we create a tiny leaf, then constant variation weights are returned.  I don't think this is bad behavior.  

We could miss out on a bit of reward, but it is simple, and given that the sample size is small in the leaf, leaning towards exploring seems reasonable to me.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -479,19 +651,28 @@ export function computeContextualBanditWeights(
...
+          maxLeaves,
+          minUsersPerLeaf,
+        )
+      : buildTreeKMeans(
+          contexts,
+          attributes,
+          metricSettings,
+          numVariations,
+          maxLeaves,
+        );
```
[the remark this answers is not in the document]
lukebrawleysmith: I think that having small leaves and returning constant weights for these leaves is reasonable.  The bandit will error on the side of exploring in these cases.

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: Done, and tried to make others less verbose.  If still too verbose in general, I am happy to take another pass.

packages/stats-ts/src/multivariateKMeans.ts
lukebrawleysmith: The total SSE and cluster-level SSE convergence criteria is identical (cluster-level SSE sums to total SSE).  I think that using total SSE potentially introduces rounding error, so I don't think it's safer.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -442,8 +441,177 @@ function buildTree(
...
+        const km =
+          categories.length <= MAX_EXHAUSTIVE_CATEGORIES
+            ? new ExhaustiveBinaryKMeans<ArmColumns[]>().fit(
+                categoryArms,
+                groupSse,
+              )
+            : new MultivariateKMeans<ArmColumns[]>(2, 100).fit(
+                categoryArms,
+                groupSse,
+              );
```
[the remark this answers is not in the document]
lukebrawleysmith: Non-converged K-means splits can still result in the largest reduction in SSE.  Even if not converged, we should keep the best candidate split.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -479,14 +975,12 @@ export function computeContextualBanditWeights(
     return { attributes, responses: [], leaf_map: [] };
   }
 
-  const features = buildFeatures(contexts, attrColumns);
-  const { leafByContext, sseTrajectory } = buildTree(
+  const { leafByContext, sseTrajectory, pathAttrsByLeaf } = buildTreeKMeans(
```
lukesonnet: Shouldn't this be called `buildTree` because it can either be kmeans or it can be exhaustive?

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -298,68 +267,510 @@ function partitionByContext(
...
+          pathAttributes,
+          attrMaps.filter((_, c) => !ownSet.has(c)),
+        ),
+      };
+    });
+}
+
 type BuildTreeResult = {
+  /** Per-context leaf assignment (parallel to `contexts`): `leafByContext[c]`. */
   leafByContext: number[];
```
lukesonnet: Seems like you can combine `leafByContext` and `pathAttrsByLeaf` into:

```
type LeafInfo = Map<number, { memberContexts: number[]; pathAttributes: Set<number> }>;
```

lukesonnet (approved):

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: removec

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: done

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: done

PR #6308, opened 2026-07-08
Fix for inserting elements correctly with the visual editor. 

Auz:
The visual editor sometimes gets confused when asked to insert elements. It can replace large contents of the page using innerHTML which causes problems. This change allows the AI to better inject elements using JS.


Auz:
@greptileai review

jdorn (approved):

PR #6309, opened 2026-07-08
Introduces new chart settings controls

mknowlton89:
### Features and Changes

Product Analytics charts often need axis titles that are not the default “Total” / “Per unit” (or a blank category axis). This adds optional X/Y axis labels in Chart Settings so explorations and dashboard blocks can show the labels people actually mean, without changing the query.

- Optional `chartSettings.axes` (`categoryAxisLabel`, `valueAxisLabel`) on the exploration config. Existing saved explorations, dashboard blocks, URLs, and API clients keep working because the field is optional.
- Collapsible Chart Settings in the explorer sidebar for line, area, and bar chart types. Labels are stored on the data axis vs category axis, so horizontal bars still map X/Y correctly.
- Empty labels fall back to the existing defaults. Whitespace-only labels are stripped on submit so we do not persist empty settings.
- Server display config now also applies requested `showAs` and `chartSettings` so dashboard/block renders match the explorer.
- `@/ui/Button` forwards `aria-expanded` for the expand/collapse control.

### Dependencies

None

### Follow-ups
This section will likely expand - to allow users to specify chart colors, chart scales (for both x/y axis), and handle things like anchoring a chart to 0 or not. 

### Testing

- [x] Open an exploration with a line, area, or vertical bar chart. Expand Chart Settings, set X-axis and Y-axis labels, run/update the chart, and confirm ECharts shows those titles.
- [x] Clear both fields (or whitespace only) and confirm the chart returns to default labels (including “Total” / “Per unit” when Show as applies) and that `chartSettings` is not kept as empty objects.
- [x] Switch to a horizontal / stacked horizontal bar and confirm X/Y labels follow the on-screen axes, not the stored category/value keys.
- [x] Confirm the Chart Settings section is hidden for chart types without axes (e.g. big number, table, pie).
- [x] Changing only axis labels should not trigger a new warehouse fetch.
- [x] Save the exploration / add it to a dashboard and reload: custom labels still appear.


packages/front-end/enterprise/components/ProductAnalytics/SideBar/ChartSettingsSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+        border: "1px solid var(--gray-a3)",
+        borderRadius: "var(--radius-4)",
+        backgroundColor: "var(--color-panel-translucent)",
+      }}
+    >
+      <Text weight="medium">Chart Settings</Text>
+      <TextField
+        label="X-axis label"
+        placeholder="Default"
+        value={draftExploreState.chartSettings?.[xAxisKey] ?? ""}
```
nhat-growthbook: I think the initialValue of these 2 fields should be the Current values in the axis. Currently they are empty when I first open up. 
This will also tie to my other suggestion where users can directly click on the axis label on the chart and converts it into a Text Field --> then they can directly rename the axis.
That would require a 2 way tie to this text field.

mknowlton89: Updated the logic so the fields' initial values are the current values in [7531792](https://github.com/growthbook/growthbook/pull/6309/commits/7531792bee9ac50e529f73579612d6a800913d70).

Gonna hold off on the new feature of allowing the user to click on the axis labels directly and edit in-line - we don't have that pattern on the chart today, so want to do that work in a separate branch

packages/front-end/enterprise/components/ProductAnalytics/util.ts
```diff
@@ -932,6 +949,54 @@ export function removeIncompleteInputs(
...
+  chartSettings: ProductAnalyticsChartSettings | undefined,
+): ProductAnalyticsChartSettings | undefined {
+  if (!chartSettings) return undefined;
+  const next: ProductAnalyticsChartSettings = {};
+  // Preserve empty strings: they mean "hide this label", not "use the default".
+  if (chartSettings.categoryAxisLabel !== undefined) {
+    next.categoryAxisLabel = chartSettings.categoryAxisLabel.trim();
+  }
+  if (chartSettings.valueAxisLabel !== undefined) {
+    next.valueAxisLabel = chartSettings.valueAxisLabel.trim();
```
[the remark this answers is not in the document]
mknowlton89: This is intentional - we want a user to be able to override the default value and remove it.

packages/back-end/src/enterprise/services/product-analytics.ts
```diff
@@ -60,6 +60,8 @@ function withRequestedDisplayConfig(
     config: {
       ...existing.config,
       chartType: requested.chartType,
+      showAs: requested.showAs,
```
nhat-growthbook, line 63: Just 1 last note, I got confused by `showAs` name. Is it `showAsIs`?

nhat-growthbook (approved):
Approved with a small confused note.

packages/back-end/src/enterprise/services/product-analytics.ts
mknowlton89: Yea, I can see how the `showAs` isn't very contextual.

That was introduced [here](https://github.com/growthbook/growthbook/issues/5730) - we introduced the concept of letting a user change between showing the chart as the `Totals` or `Per Unit`. E.g. for an orders you could show total orders, or total orders per user_id.

PR #6313, opened 2026-07-09
Perf: slim heavy metric fields out of the definitions endpoint

Kevin-Chant:
### Features and Changes

`GET /organization/definitions` returned full legacy metric documents, including `sql`, `templateVariables`, `conditions`, `queries`, `analysis`, and `analysisError` — none of which the definitions consumers read. For orgs with many metrics this is a large share of the endpoint's payload (and of its Mongo read + serialization time). Follows the same pattern as saved groups (`getAllWithoutValues`) and constants.

- **Back-end**: new `getMetricsForDefinitions` in `MetricModel` excludes the fields via a DB projection (`lodash/omit` for the config.yml branch, where projections can't apply). `getMetricsByOrganization`/`getMetricMap` are untouched, so experiment analysis still sees full metrics. New shared type `MetricDefinitionInterface` (`Omit` of the dropped fields).
- **Front-end types**: `DefinitionsContext` metrics are now `MetricDefinitionInterface`, and `getExperimentMetricById` returns the new `ExperimentMetricDefinition` union. Shared utils and component props that never read the dropped fields were loosened to the slim types — no runtime changes, and a full metric remains assignable everywhere.
- **Edit/duplicate guard**: `MetricModal` now fetches the full metric via `GET /metric/:id` before seeding `MetricForm`. Previously it seeded from the context metric, so without this the slimmed payload would have silently wiped `sql` on save.
- **config.yml export/import**: `BackupConfigYamlButton`/`RestoreConfigYamlButton` fetch full metrics from the existing (previously FE-unused) `GET /metrics` endpoint so the exported YAML keeps metric SQL.

### Testing

- [x] Jest tests for `getMetricsForDefinitions`: heavy fields absent from DB and config.yml reads, org scoping, doc migration still applied
- [x] Type-check passes across shared/back-end/front-end (the compiler enumerates every consumer of the dropped fields)
- [ ] Edit + duplicate a metric from the `/metrics` list and from a datasource page → SQL/conditions preserved after save
- [ ] Export config.yml → metric definitions still include `sql`

### DIFF
--- a/packages/back-end/src/models/MetricModel.ts
+++ b/packages/back-end/src/models/MetricModel.ts
@@ -325,7 +325,7 @@ async function findMetrics(
   const metrics: MetricInterface[] = [];
   const metricIds = new Set<string>();
 
-  // If using config.yml, first check there (projection can't apply here)
+  // If using config.yml, first check there
   if (usingFileConfig()) {
     getConfigMetrics(context)
       .filter((m) => !additionalQuery || evalCondition(m, additionalQuery))
@@ -376,11 +376,15 @@ export async function getMetricsByOrganization(
   options?: {
     datasourceId?: string;
     projectId?: string;
+    includeArchived?: boolean;
   },
 ) {
   const query: FilterQuery<LegacyMetricInterface> = {
     ...(options?.datasourceId && { datasource: options.datasourceId }),
     ...(options?.projectId && projectFilterQuery(options.projectId)),
+    ...(options?.includeArchived === false && {
+      status: { $ne: "archived" },
+    }),
   };
 
   return findMetrics(context, query);
--- a/packages/back-end/src/models/MetricModel.ts
+++ b/packages/back-end/src/models/MetricModel.ts
@@ -1,9 +1,11 @@
 import mongoose, { FilterQuery } from "mongoose";
 import { evalCondition } from "@growthbook/growthbook";
 import { ExperimentMetricInterface } from "shared/experiments";
+import omit from "lodash/omit";
 import {
   InsertMetricProps,
   LegacyMetricInterface,
+  MetricDefinitionInterface,
   MetricInterface,
 } from "shared/types/metric";
 import { getConfigMetrics, usingFileConfig } from "back-end/src/init/config";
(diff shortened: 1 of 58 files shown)


packages/back-end/src/models/MetricModel.ts
```diff
@@ -318,16 +320,19 @@ export async function getMetricMap(
...
   context: ReqContext | ApiReqContext,
   additionalQuery?: FilterQuery<LegacyMetricInterface>,
-) {
+  excludeFields?: readonly (keyof MetricInterface)[],
+): Promise<MetricInterface[]> {
   const metrics: MetricInterface[] = [];
   const metricIds = new Set<string>();
 
-  // If using config.yml, first check there
+  // If using config.yml, first check there (projection can't apply here)
```
tzjames: Kind of a useless add-on comment.
```suggestion
  // If using config.yml, first check there
```

packages/back-end/src/models/MetricModel.ts
```diff
@@ -378,6 +386,23 @@ export async function getMetricsByOrganization(
   return findMetrics(context, query);
 }
 
+const METRIC_DEFINITION_EXCLUDED_FIELDS = [
+  "sql",
+  "templateVariables",
+  "conditions",
+  "queries",
+  "analysis",
+  "analysisError",
```
tzjames, line 399: FWIW sql is 30%, analysis is 10% (though that was already filtered before this), templateVariables are 3%, the others are less than 2%.  Besides the ones listed here windowSettings would be the next biggest at 7.6% but that is used in too many places .

packages/front-end/components/Settings/RestoreConfigYamlButton.tsx
```diff
@@ -49,9 +52,17 @@ export default function RestoreConfigYamlButton({
...
+  const { datasources, dimensions, mutateDefinitions, segments } =
     useDefinitions();
 
+  // Definitions only contain slimmed metrics (no sql, etc.), so fetch the
+  // full versions for the import diff. Skip archived metrics to match the old
+  // definitions-based behavior (/metrics includes them, definitions didn't).
+  const { data: metricsData } = useApi<{ metrics: MetricInterface[] }>(
+    "/metrics",
+  );
+  const metrics = metricsData?.metrics.filter((m) => m.status !== "archived");
```
tzjames: We seem to remove archived metrics in a few places.  What about adding that as a query parameter to the /metrics endpoint and not even bother fetching it from the database?

packages/shared/types/metric.d.ts
```diff
@@ -96,6 +96,21 @@ export interface MetricInterface {
   queryFormat?: "sql" | "builder";
 }
 
+// Slimmed metric returned by the definitions endpoint. Heavy fields are
+// excluded; fetch the full metric by id when they're needed.
+// NOTE: omitting the required `queries` field is what makes this type
+// non-assignable to MetricInterface (the other omitted fields are optional).
+// If `queries` ever becomes optional, that compile-time guard disappears.
```
tzjames: queries are on average just 1.6% of the total document size.  Though I think having this extra type would still make sense to make it clear that those optional properties are never set in case a future dev tries to access it. So I think the NOTE: is unnecessary.

tzjames (approved):
Approved.  A few nits about comments. We could consider adding ?includeArchived as a parameter for the /metrics endpoint defaulting to false as the general case might be to want to exclude them.

PR #6314, opened 2026-07-09
Perf: exclude fact table SQL from the definitions endpoint

Kevin-Chant:
### Features and Changes

`GET /organization/definitions` returned full fact table documents, but `sql` and per-column `jsonFields` are only needed on a few detail/editor screens. `getAllFactTablesForDefinitions` excludes both at the DB layer with a `{ sql: 0, "columns.jsonFields": 0 }` projection, and a new `FactTableDefinition` shared type (omits `sql`; each column omits `jsonFields`) flows through `DefinitionsContext`. Same pattern as #6313. The column skeleton (name, datatype) and saved filters stay in the payload — filter values feed SQL previews via `getRowFilterSQL` in several components.

Components that need the full document fetch it by id from a new internal `GET /fact-tables/:id` endpoint via a `useFullFactTable` hook, which falls back to the slim copy while the request is in flight. The fact table detail page revalidates that fetch whenever the definitions copy changes, so child modals (SQL editor, filters, columns) that only call `mutateDefinitions` never leave the page stale. `ImportFromStatsig` fetches full fact tables from `GET /fact-tables` (wiring up the previously unrouted `getFactTables` controller) since its import diff compares `sql`. Remaining changes are type-only loosening of props/utils that never read `sql` or `jsonFields`; a full `FactTableInterface` stays assignable to the slim type everywhere.

Note: the slim guard is fully compiler-enforced for `sql` (a required field, so its absence blocks assignment) but only structural for `jsonFields` (optional) — see the comment on `FactTableColumnDefinition`.

### Testing

- [x] Jest tests for `getAllFactTablesForDefinitions`: `sql` and column `jsonFields` absent, column skeleton / filter values preserved, org scoping
- [x] Type-check passes across shared/back-end/front-end (the compiler enumerates every `sql` consumer)
- [x] Fact table detail page renders SQL; edits via the SQL/filter/column modals refresh the page
- [x] Fact metric modal SQL preview shows saved-filter WHERE clauses and JSON sub-field column options
- [ ] Statsig import diff still detects fact table SQL changes


packages/front-end/pages/fact-tables/[ftid].tsx
```diff
@@ -89,11 +90,29 @@ export default function FactTablePage() {
     _factMetricsIncludingArchived: factMetrics,
     getDatasourceById,
   } = useDefinitions();
-  const factTable = getFactTableById(ftid as string);
+  // Definitions only contain a slimmed fact table (no sql or filter values),
+  // so fetch the full version for this page
```
Comment (tzjames) [inline packages/front-end/pages/fact-tables/[ftid].tsx]:
```suggestion
  // Definitions exclude sql, so fetch the full version for this page
```

Comment (tzjames) [reply packages/front-end/pages/fact-tables/[ftid].tsx]: This comment seems out of date. We only strip sql, not filters.

tzjames (dismissed):
Just one comment seemed off.  The rest looks good to go.  Though sql consists of only 19% of the total document size on average.  Columns would be a bigger win.

tzjames (approved):

PR #6319, opened 2026-07-09
SQL Explorer V2

mknowlton89:
### Features and Changes

- Introduces a new SQL Explorer within Product Analytics, including query previews, schema browsing, autocomplete, AI-assisted SQL generation, formatting, and visualization configuration.
- Supports SQL explorations without a timestamp column while restricting date-dependent charts, comparisons, and filters appropriately.
- Adds a dedicated workflow for creating and editing SQL Dashboard Blocks, including external editing sessions and dashboard date controls.
- Deprecates legacy SQL Reports and SQL Explorer Dashboard Blocks while preserving access to existing reports. Organizations with saved reports see a dismissible migration callout. (existing blocks still work, users just won't see the option to create new ones in the Dashboard)
- Updates Product Analytics APIs, validators, caching, and generated OpenAPI specifications to support SQL explorations.

### Dependencies

None.

### Fast Follows
- Restrict “Sum” value-column options to columns typed as number for SQL and Data Source explorations, matching Fact Table behavior. Add backend validation to reject non-numeric sum columns so malformed or stale configurations cannot generate invalid SUM(varchar) queries.

### Testing

- [x] Create and run SQL explorations with and without timestamp columns.
- [x] Verify query previews, autocomplete, formatting, schema browsing, and AI-assisted SQL generation.
- [x] Configure count and sum values and verify supported chart types render correctly.
- [x] Save a SQL exploration to a dashboard and edit it from the Dashboard Editor.
- [x] Verify updating and exiting an external Dashboard Block editing session.
- [x] Confirm organizations with saved legacy SQL Reports see the dismissible migration callout and link.
- [x] Confirm organizations without saved legacy SQL Reports do not see the callout.
- [x] Verify existing SQL Reports and legacy Dashboard Blocks remain accessible.
- [x] Verify an organization can create a Sql Exploration via the REST API
- [x] Verify an org without a data source sees the empty state if they land on the sql explorer v2
- [x] Validate that a user must have the `canRunSqlExplorerQueries` permission to actually run queries via the SQL Explorer
- [x] Validate a user can write a query that doesn't contain a timestamp column and it works ok

### Screenshots / Walk-Thru Videos

https://www.loom.com/share/343f1b789e0149febd8a01d032e82694


packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -39,8 +40,36 @@ import {
 // Internal Type definitions
 type MinimalFactTable = Pick<
   FactTableInterface,
-  "sql" | "columns" | "filters" | "userIdTypes" | "timestampColumn"
->;
+  "sql" | "columns" | "filters" | "userIdTypes"
+> & {
+  timestampColumn: string | null;
+  quoteTimestampColumn: boolean;
```
jdorn, line 46: I don't really understand the name of this setting and what it's supposed to be doing.

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -182,6 +214,19 @@ function getFactTableGroups({
...
+    case "sql":
+      return [
+        {
+          index: 0,
+          factTable: createStubFactTable(
+            config.dataset.sql.replace(/;\s*$/, "").trim(),
+            config.dataset.timestampColumn,
+            config.dataset.columnTypes,
+            datasourceSettings,
+          ),
```
jdorn, line 226: I believe Redshift and MS SQL do not support nested `WITH` clauses.  So if a user writes SQL with CTEs (fairly common in BI reports) it would have worked in the old SQL Explorer, but would fail in this one.

I think this is a perfectly fine restriction to have. We might just want to document it somewhere for those data sources.

jdorn (commented):
Currently, we're defaulting to have 1 value (`COUNT`) pre-selected for all SQL explorations.  Clicking "Run" fires off 2 queries - a `LIMIT 100` query to populate the results table, plus the exploration query (`COUNT(*)`) to populate the visualization.

However, I think it will be common to first iterate a bunch just using the SQL result preview BEFORE adding your first visualization so we'd be running a lot of unnecessary (and expensive) SQL.

I think it makes better sense to instead default to 0 values and only run the `LIMIT 100` query.  Wait until the user explicitly adds values before we run the exploration query.

Also, I'm not fully sold on the UX right now for entering the exploration mode. I think the "table" chart will be popular for SQL Explorer, and people don't really think of that as a "visualization" so it's not obvious how to get there.  In my mind, it's more of a 2-step thing where first I build out the dataset, and then I explore it.

packages/shared/src/enterprise/product-analytics/sql.ts
mknowlton89: @jdorn Are you referring to the `quoteTimestampColumn` or the `MinimalFactTable` type and `toMinimalFactTable` helper?

The `quoteTimestampColumn` was added to address a potential issue with quoting the timestamp column in the sql if the column was something like `Event Time` (with a space) or it was a reserved work.

It distinguishes between

- SQL dataset stub: quote the timestamp identifier.
- Real Fact Table: preserve the timestamp expression unchanged.
- SQL dataset with null: treat it as timeless rather than defaulting to timestamp.

I could see something like `timestampColumnIsIdentifier` being a better property.

The concrete example is, a user enters a SQL query with an aliased timestamp column

```
SELECT created_at AS "Event Time"
FROM events
```

Without quoting, the GrowthBook generated sql would be invalid:

```
WHERE Event Time >= TIMESTAMP '2026-08-01'
```

mknowlton89:
> Currently, we're defaulting to have 1 value (`COUNT`) pre-selected for all SQL explorations. Clicking "Run" fires off 2 queries - a `LIMIT 100` query to populate the results table, plus the exploration query (`COUNT(*)`) to populate the visualization.
> 
> However, I think it will be common to first iterate a bunch just using the SQL result preview BEFORE adding your first visualization so we'd be running a lot of unnecessary (and expensive) SQL.
> 
> I think it makes better sense to instead default to 0 values and only run the `LIMIT 100` query. Wait until the user explicitly adds values before we run the exploration query.
> 
> Also, I'm not fully sold on the UX right now for entering the exploration mode. I think the "table" chart will be popular for SQL Explorer, and people don't really think of that as a "visualization" so it's not obvious how to get there. In my mind, it's more of a 2-step thing where first I build out the dataset, and then I explore it.

@jdorn Yea, I can see what you mean. I like the idea of not defaulting to a `COUNT` value and having the user explicity add a value. I've made that change, and added documentation.

RE: The UX, I can see what you mean - you mentioned

```
In my mind, it's more of a 2-step thing where first I build out the dataset, and then I explore it.
```

The part that I feel like we haven't quite nailed yet is when you build out the dataset, its really just the raw data set, and then the exploration wraps that base query, so the return value of the exploration is in a way different format. 

When you are building out the dataset, do you feel like you'd build it in a manner so the shape of the data returned is exactly what would be displayed in the visualization, or would you build the raw data and then expect the explore tool to do all of the reshaping?

packages/front-end/components/SchemaBrowser/AiSqlGenerator.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const generateSql = async () => {
+    if (disabled) return;
+    if (!aiAgreedTo) {
+      openAgreementModal();
+      return;
+    }
+    if (!aiEnabled) {
+      setError("AI is disabled for your organization. Adjust in settings.");
+      return;
+    }
```
nhat-growthbook, line 71: Should these 2 conditions flipped? Do we want to move the `setError` to above `if (!aiAgreedTo)` so it catches the error condition first?

packages/front-end/components/SchemaBrowser/AiSqlGenerator.tsx
```diff
@@ -0,0 +1,212 @@
...
+      .catch(() => {
+        // Error handling is done by the apiCall errorHandler
+      })
+      .finally(() => {
+        updateLoading(false);
+      });
+  };
+
+  const handleTriggerClick = () => {
+    if (!hasCommercialFeature("ai-suggestions")) {
```
nhat-growthbook, line 115: Not sure if it's better UX, but instead of enabling the button and throw error onClick, should we disable the buttons and show these texts in tooltips? Take it or leave it.

packages/front-end/components/Settings/DisplayTestQueryResults.tsx
```diff
@@ -213,197 +297,235 @@ export default function DisplayTestQueryResults({
...
+                        {totalPages > 1 ? (
+                          <Flex align="center">
+                            <div className="mr-1">
+                              Showing {page * pageSize - pageSize + 1} -{" "}
+                              {Math.min(page * pageSize, results.length)} of{" "}
+                              <Tooltip
+                                body={
+                                  "GrowthBook limits the result to 1,000 rows max"
+                                }
+                                shouldDisplay={results.length >= 1000}
```
nhat-growthbook: Just nit, should `1000` be a shared constant between BE and FE?

packages/front-end/enterprise/components/ProductAnalytics/SideBar/ValueCard.tsx
```diff
@@ -124,10 +125,17 @@ export default function ValueCard({
 
   let supportsUnitSelection = false;
 
-  if (dataset.type === "fact_table" || dataset.type === "data_source") {
-    supportsUnitSelection = dataset.values[index].valueType === "unit_count";
-  } else if (dataset.type === "metric") {
-    const factMetric = getFactMetricById(dataset.values[index].metricId ?? "");
+  if (
+    draftExploreState.dataset.type === "fact_table" ||
```
nhat-growthbook: Just nit, `["fact_table", "data_source", "sql"].includes(draftExploreState.dataset.type)` would clean up and more extendable.

packages/front-end/enterprise/components/ProductAnalytics/util.ts
```diff
@@ -977,6 +1060,17 @@ export function isSubmittableConfig(
       return false;
   }
 
+  if (
+    cleanedConfig.dataset.type === "sql" &&
```
nhat-growthbook: wow this condition is hard to follow

nhat-growthbook:
Screenshot 2026-08-11 at 10 51 55
2 things from the screenshots:
1. The vertical scrollbar should be inside the dialog only.
2. The horizontal scrollbar shouldn't exist on the main body. I'm guessing the dialog width isn't scaling properly with screenwidth

nhat-growthbook:
Screenshot 2026-08-11 at 11 12 57
Screenshot 2026-08-11 at 11 09 29
I tried to follow the suggestions in the AI box. It generated the SQL and it returns value, but the amount is empty. Probably because I didn't specify the timeframe "last month" or "current".
Maybe we need to add a Coalesce or something to numeric column?

packages/front-end/components/SchemaBrowser/AiSqlGenerator.tsx
mknowlton89: Yea, absolutely. great catch

packages/front-end/components/SchemaBrowser/AiSqlGenerator.tsx
mknowlton89: Oof - yea, I missed that case with the `disabled` prop. Added it now.

packages/front-end/components/Settings/DisplayTestQueryResults.tsx
mknowlton89: 100% - it actually is already a shared const, I just didn't use it lol. Updated.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/ValueCard.tsx
mknowlton89: Yea, this isn't very easy to read. The downside about using `.includes` (AFAIK) is that t doesn't narrow the discriminated union, so it requires some casting.

I've updated to use a switch case to try and make it a bit easier to read.

```ts
switch (draftExploreState.dataset.type) {
    case "fact_table":
    case "data_source":
    case "sql":
      supportsUnitSelection =
        draftExploreState.dataset.values[index].valueType === "unit_count";
      break;
    case "metric": {
      const factMetric = getFactMetricById(
        draftExploreState.dataset.values[index].metricId ?? "",
      );
      if (
        factMetric?.metricType === "mean" ||
        factMetric?.metricType === "proportion" ||
        factMetric?.metricType === "retention" ||
        factMetric?.metricType === "dailyParticipation"
      ) {
        supportsUnitSelection = true;
      } else if (
        factMetric?.metricType === "ratio" &&
        factMetric.numerator.column === "$$distinctUsers"
      ) {
        supportsUnitSelection = true;
        // TODO: handle separate denominator unit selector
      }
      break;
    }
    case "funnel":
      break;
  }
  ```

File: packages/front-end/enterprise/components/ProductAnalytics/util.ts (hunk 7 above)
Comment (mknowlton89) [reply packages/front-end/enterprise/components/ProductAnalytics/util.ts]: Yea...

I've updated to:

```ts
  if (cleanedConfig.dataset.type === "sql") {
    const { sql, timestampColumn, columnTypes } = cleanedConfig.dataset;
    const hasSql = sql.trim().length > 0;
    const hasColumnTypes = Object.keys(columnTypes).length > 0;
    const timestampIsDate =
      timestampColumn === null || columnTypes[timestampColumn] === "date";
      // Config is not submittable without sql, columns, or if timestamp column is not a date
    if (!hasSql || !hasColumnTypes || !timestampIsDate) return false;
  }
```

mknowlton89:
@greptile review

nhat-growthbook:
Screenshot 2026-08-24 at 09 34 06
Overall it LGTM, I don't see any obvious issue. I still see the scrollbars issue when it's in a modal.
The AI generated SQL seems to pick the "incomplete" table path.
It generated `growthbook.orders` but the table is `growthbook.public.orders`.

nhat-growthbook:
Also I ran it through `ponytail-review` and it claims to be able to clean up like 400 lines in case you wanna run an audit there.

mknowlton89:
> Also I ran it through `ponytail-review` and it claims to be able to clean up like 400 lines in case you wanna run an audit there.

Great idea - went through and cleaned up 7 of the 8 claims from ponytail

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -39,14 +40,45 @@ import {
...
+    timestampColumn: getFactTableTimestampColumn(factTable),
+    quoteTimestampColumn: false,
+  };
+}
+
+function getTimestampColumnExpression(
+  factTable: MinimalFactTable,
+  helpers: SqlDialect,
+): string | null {
+  if (!factTable.timestampColumn) {
```
najito: Should an empty string in `timestampColumn` be treated the same as `null`? The check `if (!factTable.timestampColumn)` also matches `""`, but `dataSourceDatasetValidator` doesn't normalize an empty string to `null` the way the SQL dataset validator does. If a `data_source` exploration is submitted before the user picks a column, the date range filter is silently dropped instead of raising an error. Is that the intended fallback? If not, should this validate or normalize the value like the sibling dataset type does?

packages/front-end/enterprise/components/ProductAnalytics/ExplorerPageActions.tsx
```diff
@@ -0,0 +1,131 @@
...
+        : loading || isStale || needsFetch
+          ? "Run the updated exploration before saving to a dashboard."
+          : undefined;
+
+  return (
+    <>
+      {showSaveToDashboardModal && (
+        <SaveToDashboardModal
+          close={() => setShowSaveToDashboardModal(false)}
+          config={stripExplorerDraftFields(draftExploreState)}
```
najito, line 59: When a user clears a SQL exploration's timestamp column, should the chart type and date dimension reset too? `updateTimestampColumn` and `applyColumnMetadata` handle the transition from `null` to a value by adding a date dimension and switching to a line chart, but they don't handle the reverse. `stripExplorerDraftFields` doesn't clean this up before saving either. Can this let a dashboard block save with `chartType: "line"`, `timestampColumn: null`, and a leftover date dimension?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/ProductAnalyticsExplorerSettings.tsx
```diff
@@ -186,6 +205,7 @@ export default function ProductAnalyticsExplorerSettings({
           | MetricExplorationBlockInterface
```
najito, line 217: Now that `SqlExplorationBlockInterface` is included in this union, does that also route SQL exploration blocks through the `chartType: block.config?.chartType || exploration.config?.chartType` merge a few lines below (line 202)? `normalizeTimelessSqlConfig` downgrades a config that pairs a line chart with no date dimension, but that merge picks the old `block.config?.chartType` first because it's still truthy. If a user removes the timestamp column and reruns the query, does this reintroduce the same invalid combination?

packages/front-end/enterprise/components/ProductAnalytics/MainSection/useSqlQueryPreview.ts
```diff
@@ -0,0 +1,188 @@
...
+
+        const columns = response.columns ?? [];
+        const columnTypes = Object.fromEntries(
+          columns.map((column) => [column.name, column.dataType ?? "other"]),
+        ) as SqlDataset["columnTypes"];
+        const timestampColumn =
+          columns.find((column) => column.dataType === "date")?.name ?? null;
+
+        lastPreviewedSqlRef.current = sql;
+        applyColumnMetadata(sql, columnTypes, timestampColumn);
```
najito, line 122: Is it intentional that every "Run Query" click re-derives the timestamp column from the first detected date column, even after the user picks a different one in `TimestampColumnSelector`? The Run button stays enabled after a successful run, so if the user reruns the query, does this silently overwrite a manual selection, for example replacing `event_at` with `created_at`, with no warning?

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -191,6 +226,19 @@ function getFactTableGroups({
         },
       ];
     }
+    case "sql":
+      return [
+        {
+          index: 0,
+          factTable: createStubFactTable(
+            config.dataset.sql.replace(/;\s*$/, "").trim(),
```
najito: Should the trailing semicolon strip, `/;\s*$/`, also match a semicolon followed by a trailing comment, for example `SELECT ...;\n-- note`? A query that passes `isReadOnlySQL` as a valid single statement can still fail downstream with a "multi statement" error in this case.

packages/back-end/src/enterprise/services/product-analytics.ts
```diff
@@ -138,6 +139,42 @@ export async function runProductAnalyticsExploration(
...
+        throw new BadRequestError(
+          "Timestamp column must exist in query results",
+        );
+      }
+      if (dataset.columnTypes[dataset.timestampColumn] !== "date") {
+        throw new BadRequestError(
+          "Timestamp column must be a date or timestamp",
+        );
+      }
+    } else {
```
najito, line 163: Should SQL dataset validation also check `values[].valueColumn` for the `"$$distinctDates"` sentinel against a timeless dataset, the same way it already checks dimensions and chart type? Right now that combination passes validation, then throws inside `getEventValueExpr`, and the surrounding `catch` swallows the error. Does that leave the exploration stuck at `status: "running"` instead of returning a clean 400?

packages/front-end/enterprise/components/ProductAnalytics/ExplorerContext.tsx
```diff
@@ -920,22 +897,57 @@ export function ExplorerProvider({
...
+                    column: "date",
+                    dateGranularity: "auto",
+                  },
+                  ...prev.dimensions,
+                ]
+              : prev.dimensions,
           dataset: { ...prev.dataset, timestampColumn: column },
         } as ExplorationConfig;
       });
+      if (column === null) {
```
najito: Should `updateTimestampColumn`'s null column branch also reset `comparisonError`, to match `clearAllDatasets` and the `doSubmit` finalize and poll paths? As written, it resets `comparisonExploration`, `comparisonQuery`, and `comparisonComputed`, but leaves a stale error rendered wherever `compareEnabled && comparisonError` is checked.

packages/front-end/components/Settings/DisplayTestQueryResults.tsx
```diff
@@ -77,13 +82,19 @@ export default function DisplayTestQueryResults({
...
   const [downloadError, setDownloadError] = useState<string | null>(null);
+  const [showQueryModal, setShowQueryModal] = useState(false);
   const cols = orderedColumnKeys ?? Object.keys(results?.[0] || {});
   const labels = columnLabels ?? cols;
   const useTwoRowHeader = headerStructure != null && orderedColumnKeys != null;
+  const durationStatus = error ? "Query failed" : "Query succeeded";
+  const showDurationStatus = showDuration && duration > 0;
 
-  const forceShowSql = error || !results.length;
+  const showRenderedSqlContent =
```
najito, line 95: Was narrowing the fallback to rendered SQL on empty results behind a new `showNoRowsWarning` prop intended to affect chart mode? `ExplorerDataTable.tsx` always passes `showNoRowsWarning={... && !hasChart}`, so if a chart's filters return zero rows, the table now renders empty with no SQL shown. Is that the intended behavior, or should chart mode also show the query when results are empty?

packages/front-end/enterprise/components/ProductAnalytics/SideBar/TimestampColumnSelector.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </DropdownMenuItem>
+      ))}
+    </DropdownMenu>
+  );
+
+  return (
+    <Flex direction="column" gap="2" width="100%">
+      <Text weight="medium">Timestamp column</Text>
+      <Flex justify="between" align="center">
+        <Text color="text-low">{timestampColumn ?? "Not set"}</Text>
```
najito: Should this use `||` instead of `??` for the "Not set" fallback? `timestampColumn ?? "Not set"` doesn't catch an empty string, and `DatasourceConfigurator.tsx` passes `databaseDataset?.timestampColumn ?? ""` as the default and reset value. Does that make new or retargeted explorations show a blank label instead of "Not set"?

packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerMainSection.tsx
```diff
@@ -144,56 +181,10 @@ export default function ExplorerMainSection() {
...
-                  Configure your explorer to visualize data
-                </Text>
-              </>
-            )}
-          </Flex>
+          emptyState
         )}
 
-        {(isStale || loading) && !suppressStaleFloatingCallout && (
+        {(needsFetch || loading) && !suppressStaleFloatingCallout && (
```
najito: Should the "stale, click refresh" callout key off `needsFetch`, or does it need a signal closer to the old `isStale`, which only fired after a confirmed cache miss? `needsFetch` becomes true synchronously on any draft edit, including cache hit paths where `loading` never becomes true. Can this cause the "Latest changes not applied" message to flash even when the request resolves from cache almost immediately?

najito (commented):
SQL support was added as a mode inside the existing Explorer instead of as its own surface that uses Explorer for visualization. That decision is what drives most of the other findings.

`timestampColumn` absence is encoded three different ways (an empty string on the data source dataset, `null` on the SQL dataset, and the separate `quoteTimestampColumn` flag in the shared SQL generator), and `type === "sql"` checks leak through shared code instead of staying inside the dataset-specific tab content. I checked the cited code against this commit and every line reference holds up.

That root cause explains several of the correctness bugs, not just a style preference:

- `packages/shared/src/enterprise/product-analytics/sql.ts:70`: `getTimestampColumnExpression` treats an empty string the same as `null` because nothing normalizes the two absences to one value, so a `data_source` exploration with an unset timestamp silently drops its date filter instead of erroring.
- `packages/front-end/enterprise/components/ProductAnalytics/SideBar/TimestampColumnSelector.tsx:74`: `timestampColumn ?? "Not set"` doesn't catch the empty-string absence used by `data_source` datasets, so the label renders blank instead of "Not set". Same root cause as above, different surface.
- The `chartType: block.config?.chartType || exploration.config?.chartType` merge in `ProductAnalyticsExplorerSettings.tsx` and the timestamp-clear handling in `ExplorerContext.tsx:934` and `useSqlQueryPreview.ts:159` are all separate, slightly different copies of the same "what happens when the timestamp column appears or disappears" policy. Each copy handles a different subset of the state (chart type, comparison state, dimension), which is exactly the kind of drift you'd expect from duplicating one policy instead of centralizing it.

Consolidating this into `hasTimeAxis(dataset)` plus one `applyTimestampColumn` policy function looks like it would delete most of these bugs at the root and seems worth doing before merge. (Build/Explore tabs, `SqlEditorContext`) out of `Explorer.tsx` entirely; that's a real structural concern too, but it reads as a separate, larger follow-up rather than a blocker for this PR, and reasonable people could land on either side of that call depending on timeline.

Not blocking on this comment; leaving it as context alongside my earlier line comments.

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -676,10 +733,11 @@ function getEventValueExpr(
   } else if (columnRef.column === "$$count") {
     rawValue = "1";
   } else if (columnRef.column === "$$distinctDates") {
-    rawValue = helpers.dateTrunc(
-      factTable.timestampColumn || "timestamp",
-      "day",
-    );
+    const timestampColumn = getTimestampColumnExpression(factTable, helpers);
+    if (!timestampColumn) {
```
najito: Is this condition okay with guarding falsey and not null explicitly or nullish? Are there any instances where the string is empty but still valid?

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -552,14 +608,15 @@ export function generateDimensionExpression(
   const factTable = factTableGroup.factTable;
   switch (dimension.dimensionType) {
     case "date": {
+      const timestampColumn = getTimestampColumnExpression(factTable, helpers);
+      if (!timestampColumn) {
```
najito: Is this condition okay with guarding falsey and not null explicitly or nullish? Are there any instances where the string is empty but still valid?

jdorn (approved):

PR #6323, opened 2026-07-09
fix(sdk-js): match falsy array elements in $elemMatch

CedricConday:
### Problem
`$elemMatch` guarded each array element with `actual[i] && check(actual[i])`, which skips **falsy** members (`0`, `""`, `false`) before ever testing them. So `{ nums: { $elemMatch: { $eq: 0 } } }` never matched `[0]`, and likewise for `false` / `""`.

### Fix
Skip only *nullish* elements (`actual[i] != null`) so falsy-but-present values are evaluated against the condition like any other element.

### Test
Added a scoped `mongrule.test.ts` block (not the shared `cases.json`, to avoid cross-SDK impact): falsy elements now match; a non-matching array still returns `false`. Verified both directions and formatted with prettier 3.6.2.

### DIFF
--- a/packages/sdk-js/src/mongrule.ts
+++ b/packages/sdk-js/src/mongrule.ts
@@ -142,7 +142,9 @@ function elemMatch(actual: any, expected: any, savedGroups: SavedGroupsValues) {
     ? (v: any) => evalConditionValue(expected, v, savedGroups)
     : (v: any) => evalCondition(v, expected, savedGroups);
   for (let i = 0; i < actual.length; i++) {
-    if (actual[i] && check(actual[i])) {
+    // Only skip nullish elements; falsy values like 0, "" and false are valid
+    // array members and must still be tested against the condition.
+    if (actual[i] != null && check(actual[i])) {
       return true;
     }
   }
--- a/packages/sdk-js/src/mongrule.ts
+++ b/packages/sdk-js/src/mongrule.ts
@@ -144,7 +144,7 @@ function elemMatch(actual: any, expected: any, savedGroups: SavedGroupsValues) {
   for (let i = 0; i < actual.length; i++) {
     // Only skip nullish elements; falsy values like 0, "" and false are valid
     // array members and must still be tested against the condition.
-    if (actual[i] != null && check(actual[i])) {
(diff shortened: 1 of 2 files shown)


packages/sdk-js/src/mongrule.ts
```diff
@@ -142,7 +142,9 @@ function elemMatch(actual: any, expected: any, savedGroups: SavedGroupsValues) {
     ? (v: any) => evalConditionValue(expected, v, savedGroups)
     : (v: any) => evalCondition(v, expected, savedGroups);
   for (let i = 0; i < actual.length; i++) {
-    if (actual[i] && check(actual[i])) {
+    // Only skip nullish elements; falsy values like 0, "" and false are valid
+    // array members and must still be tested against the condition.
+    if ((actual[i] ?? null) !== null && check(actual[i])) {
```
najito: Could this be solved with `actual[i] != null` instead? Rather than doing a nullish coercion to null to check for null.

CedricConday: Done — pushed in 1dc80e9, back to `actual[i] != null`.

Context on why it was the other form: the automated review on this PR flagged the original `!= null` as conflicting with a strict-comparison convention, so I switched it to `(actual[i] ?? null) !== null` in e0b7bf2. Checking properly, there is no `eqeqeq` rule in `eslint.config.mjs` and `== null` is already used elsewhere in the package (e.g. `auto-wrapper.ts:134`), so the loose form lints clean. Happy to leave it as you suggested.

najito: Oh I see. I didn't read that. Sorry to make you flip flop. Might be better to maintain the style that's outlined. Approved contingent on that change. Thanks.

najito (approved):

CedricConday:
Thanks for the review, and no problem on the flip-flop — you were right that following the documented style is the better call regardless of what the linter does or doesn't enforce.

The strict form is in as of 7322e1e7, which is what you approved.

Anything else needed from me before this can go in?

najito:
No go ahead and merge thank you.

CedricConday:
Thanks, @najito. I don't have push access on this repository, so I'm not able to merge it myself — would you be able to merge it when you get a chance?

PR #6325, opened 2026-07-09
Gl/pricing phase 1 limits

gazzdingo:
### Features and Changes

Free-org usage limits, stamped per-org and tunable via a GrowthBook flag. Builds on #6289 (contains it via merge; rebases to the delta if it lands first).

**Architecture**

- New orgs get a `limits` snapshot at creation (`{maxProjects, customEnvironments, roleManagement}`). Orgs created before this ships have no snapshot and are never limited (implicit grandfathering). Paid plans read `license.limits` instead (empty today → unrestricted).
- **Cloud** stamps from the `pricing-phase-1-limits` flag (evaluated with empty attributes, so targeting rules can't affect stamps); **self-hosted** always uses the hardcoded defaults in `shared/enterprise` and never reads the CDN.
- **Kill switch:** the flag's `enabled: false` — as the base value (global) or a targeting rule on org attributes (per-customer exemption) — lifts enforcement instantly without a deploy. Enforcement-time evals send trusted server-derived attributes (`accountPlan`, hashed org id, etc.), so future per-plan values can be served with targeting rules.
- Front-end mirrors via `useOrgLimits`; limit surfaces use `PremiumCallout` and route to the Pro upgrade.

**Enforcement policy (deliberate decisions)**

- **Free orgs are all-admin for human role assignment.** Role limits gate only the endpoints where a human picks a role (invites, role changes, orphaned-user add, API keys, org default-role setting) — and only on an actual **change**, via `assertRoleChangeAllowed`.
- **Existing assignments are never auto-changed.** A lapsed Pro org's `readonly` member stays `readonly`; grandfathered keys, invites, and team configs keep working. Automated joins (SSO auto-join, verified-domain self-join, member approval, accept-invite) keep the org's configured default role and never throw. The one exception: Vercel SSO members clamp to admin on role-restricted orgs (those orgs are always freshly-stamped free with no configured default to honor).
- **Team role edits are intentionally ungated** — gating them would break trial-lapsed orgs managing existing teams (team *creation* is already premium-gated).
- **SCIM role paths are intentionally ungated for now** — unreachable until licenses ship restrictive limits; revisit when `license.limits` is populated.
- Project/environment limits enforce in the model/util layer; the Vercel integration pre-checks the project limit (after the billing block) with a message that's actionable inside Vercel's UI, and Vercel request paths warm the license cache so paid installs resolve their real plan.

### Dependencies

- The `pricing-phase-1-limits` flag should exist in our GrowthBook account **before deploy** if you want to launch dark (`enabled: false`), since with no flag the stamp falls back to the hardcoded free defaults and enforcement is on for new orgs. Setup doc: `.context/pricing-phase-1/flag-setup.md` (internal).

### Testing

- `packages/shared/test/pricing-limits.test.ts` + `entitlements.test.ts` (35 tests) cover the flag-value resolver, per-field fallbacks, kill-switch semantics, and plan resolution.
- Manual: create a fresh org (gets stamped) → project limit blocks the 2nd project with the Pro upgrade modal; environment modal restricts to default ids; invite modal defaults to admin; editing a grandfathered non-admin member/key leaves the role untouched.
- Grandfathering: pre-existing orgs (no `limits` field) behave exactly as before.


packages/back-end/src/services/organizations.ts
```diff
@@ -781,6 +821,7 @@ export async function inviteUser({
   ) {
     throw new Error("Invalid role");
   }
+  assertRoleAssignmentAllowed(organization, role);
```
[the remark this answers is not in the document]
gazzdingo: this should not be a problem since the account was created before the cutoff date and if its a downgraded account we want to keep them at the same level to reduce security issues if we auto update them

packages/back-end/src/models/ProjectModel.ts
```diff
@@ -63,6 +64,33 @@ export class ProjectModel extends BaseClass {
...
+    if (maxProjects && maxProjects > 0 && !isDemo) {
+      const existingProjects = await this.context.getProjects();
+      const nonDemoProjectCount = existingProjects.filter(
+        (p) =>
+          !isDemoDatasourceProject({
+            projectId: p.id,
+            organizationId: this.context.org.id,
+          }),
+      ).length;
+      if (nonDemoProjectCount >= maxProjects) {
```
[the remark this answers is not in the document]
gazzdingo: in real life not sure if this is going to happen

packages/back-end/src/routers/vercel-native-integration/vercel-native-integration.controller.ts
mknowlton89, line 65: @gazzdingo So with the Vercel integration - each Vercel project that you connect via the GrowthBook integration creates a corresponding project in GB, so we may want to add a bullet point here like:

```
{ label: "Projects": value: "Limit 1" }
```

Image

mknowlton89 (commented):
Checked out the Vercel specific stuff and overall I don't see any issues - left one small comment that you can take/leave.

packages/back-end/src/services/organizations.ts
```diff
@@ -495,6 +496,31 @@ export function getInviteUrl(key: string) {
...
+// Free (role-restricted) plans can only assign the admin global role. Only the
+// global role is checked here.
+export function assertRoleAssignmentAllowed(
+  organization: OrganizationInterface,
+  role: string,
+) {
+  if (getEffectiveOrgLimits(organization).orgSupportsRoles()) return;
+  if (role === "admin") return;
+
+  throw new Error(
```
bryce-fitzsimons: 402 error?

packages/front-end/components/Settings/EnvironmentModal.tsx
```diff
@@ -103,6 +131,11 @@ export default function EnvironmentModal({
           if (newEnvs.find((e) => e.id === value.id)) {
             throw new Error("Environment id is already in use");
           }
+          if (!isEnvironmentIdAllowed(value.id)) {
+            throw new Error(
```
bryce-fitzsimons: 402 error?

bryce-fitzsimons (approved):
Everything looks good here. Some minor nits about status codes (agents know how to read these), otherwise lgtm.

packages/back-end/src/api/environments/postEnvironment.ts
```diff
@@ -20,6 +20,12 @@ export const postEnvironment = createApiRequestHandler(
     req.context.permissions.throwPermissionError();
   }
 
+  if (!req.context.limits.isEnvironmentIdAllowed(environment.id)) {
+    throw new Error(
```
bryce-fitzsimons: Arguably should be a 402 error

PR #6330, opened 2026-07-10
Ramp schedules: collapse catch-up into one publish + serialize advances

bryce-fitzsimons:
## What this does

Fixes a production incident with feature ramp schedules (staged rollouts). Resuming a rollout with a large backlog of overdue steps replayed every step at once inside one request, pinning the server and taking pods down. A second bug let two processes advance the same rollout at once, leaving the UI and the served coverage disagreeing.

## Changes

**Catch up in one jump instead of step-by-step**
An overdue rollout now advances straight to where it should be in a single publish instead of replaying each step. Same end state, ~60× cheaper — the behavior that crashed pods can't happen anymore.

**One advance at a time per rollout**
A per-schedule lock ensures only one process advances a given rollout at once. Every path that changes a rollout — resume, advance, approve, jump, edit, delete, add/remove targets, monitoring toggles, the scheduler, and the publish hooks — goes through it and re-reads fresh state, so nothing collides mid-flight.

**Correctness hardening**
- Rollouts no longer finish with the rule left enabled past its end date, or disabled when it should be live.
- An advance can't slip past a freshly-added approval gate, and approvals can't land on the wrong step.
- "Start now," edits, and approvals are never silently lost to a timing collision — they apply or return a clear conflict error.
- The lock recovers correctly from crashes, slow operations, and network blips.

**Observability**
- Webhooks and the audit trail now correctly show when a catch-up folded multiple steps together (via `previousStepIndex`), so external consumers stay in sync.
- Docs updated to describe the catch-up behavior.

## Testing

Full back-end suite green (163 suites / 5,051 tests), including new regression tests for the catch-up collapse, the lock protocol, and the cutoff/approval edge cases.

### DIFF
--- a/packages/back-end/src/models/FeatureModel.ts
+++ b/packages/back-end/src/models/FeatureModel.ts
@@ -2067,21 +2067,32 @@ async function createRampSchedulesForRevision(
     // enabled immediately when this revision publishes rather than waiting
     // for the next poller tick.
     //
-    // The pre-built auditEvent ("config-edited") is passed via contentUpdates
-    // so startReadyScheduleNow appends "started" on top of it, preserving
-    // full audit history parity with the direct-edit path.
+    // The "config-edited" audit event is passed so startReadyScheduleNow
+    // appends "started" on top of it, preserving full audit history parity
+    // with the direct-edit path.
+    let currentSchedule = existingSchedule;
+    let startDeferredToScheduler = false;
     if (
       updateAction.startDate === null &&
       existingSchedule?.status === "ready"
     ) {
-      await startReadyScheduleNow(context, existingSchedule, {
+      const configEditedEvent =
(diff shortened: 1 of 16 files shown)


packages/back-end/src/models/FeatureModel.ts
```diff
@@ -2067,86 +2066,188 @@ async function createRampSchedulesForRevision(
...
+              const ensured = await ensureSafeRolloutForMonitoredRamp(
+                context,
+                updatedRunning,
+              );
+              await syncLinkedSafeRolloutForRampState(context, ensured);
+            }
+            return;
+          }
+
+          const freshContentUpdates = {
```
gazzdingo: lol this const is a mess we might want to try clean this up but its not a blocker

gazzdingo (approved):

PR #6339, opened 2026-07-13
fix: prevent long variation names from overlapping Actual Units column in balance check table

### LINKED ISSUE #6304: [Bug] Experiment Balance Check table truncates Actual Units due to overlapping variation name
### Summary

The **Experiment Balance Check** table on the **Health** tab has a UI layout issue where long variation names overlap with the **Actual Units** column. This makes the unit counts unreadable.

#### Expected Behavior

- Variation names should not overlap adjacent columns.
- The full Actual Units value should always remain visible.
- Long variation names should either:
  - truncate with an ellipsis and show the full name on hover, or
  - wrap to the next line, or
  - have a wider column with responsive layout.

#### Current Behavior

Long variation names extend into the **Actual Units** column, obscuring the numeric values. As a result, users cannot accurately read the actual unit counts.

hiro-nikaitou:
## Summary

The **Experiment Balance Check** table on the **Health** tab has a UI layout issue where long variation names overlap with the **Actual Units** column, making unit counts unreadable.

## Changes

- Added Bootstrap's `text-truncate` class to the variation name cell in `VariationUsersTable`
- Long names now truncate with an ellipsis instead of overflowing into adjacent columns
- The full name remains visible via the existing tooltip on `VariationLabel`

Closes #6304

### DIFF
--- a/packages/front-end/components/Experiment/TabbedPage/VariationUsersTable.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/VariationUsersTable.tsx
@@ -75,7 +75,7 @@ export default function VariationUsersTable({
           {variations.map((v, i) => {
             return (
               <tr key={v.id}>
-                <td className={"border-right"}>
+                <td className="border-right text-truncate">
                   {hideVariationIndex ? (
                     <Text color="text-mid" weight="medium">
                       {v.name}


najito (approved):
It looks like the Greptile comment is talking about work that is slightly outside of the scope of the PR. I'll make note of this to address but this should be fine as is.

hiro-nikaitou:
Hi @ahdriel, this PR fixes long variation names overlapping the Actual Units column in the balance check table. It already has an approval and all checks are green, but the merge is still blocked. Could you take a look and merge it when you have a chance? Thanks!

PR #6340, opened 2026-07-13
Allow removing preserved legacy identifiers from a managed warehouse

Kevin-Chant:
### Features and Changes

The JSON-columns migration (#6175) preserves legacy join keys in `settings.migratedIdentifiers` so historical experiments don't break, but that list is write-once and never reconciled — so an identifier that has since gone dead (e.g. an attribute renamed from `org_id` to `organizationId`) stays selectable as an Experiment Assignment identifier with no backing attribute, and the managed-warehouse Identifiers panel was read-only with no way to remove it.

Adds a way to remove a preserved-legacy identifier:

- `removeManagedWarehouseLegacyIdentifier` drops the entry from `settings.migratedIdentifiers` and re-runs `syncManagedWarehouseIdentifiers`, which rebuilds `userIdTypes` + exposure queries and marks the orphaned `ch_events` column deleted. Only entries in `migratedIdentifiers` are removable — builtins and current hashAttribute identifiers stay managed through attributes.
- New `POST /datasource/:id/managed-warehouse/remove-legacy-identifier` (gated on `canUpdateDataSourceSettings`).
- The Identifiers panel now renders preserved-legacy identifiers in gray with a remove action (behind a confirm noting any experiment still using it will need a different identifier); current identifiers stay non-removable. Corrected the panel copy, which previously claimed all identifiers come from attributes.

Note: removing an identifier that an experiment still references intentionally invalidates that experiment's assignment query — that's the point of retiring a dead identifier, and the confirm dialog says so.

`settings.migratedColumns` (preserved legacy *dimensions*) has the same write-once staleness but no surface in this panel; left as a follow-up.

### Testing

`back-end` + `front-end` type-check and lint pass. Not exercised in a live UI — the removable set is `migratedIdentifiers` membership (builtins/current-attribute identifiers are excluded by construction), and the existing sync handles all downstream cleanup.

### DIFF
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -1102,6 +1102,10 @@ app.post(
   "/datasource/:datasourceId/recreate-managed-warehouse",
   datasourcesController.postRecreateManagedWarehouse,
 );
+app.post(
+  "/datasource/:datasourceId/managed-warehouse/remove-legacy-identifier",
+  datasourcesController.postRemoveManagedWarehouseLegacyIdentifier,
+);
 
 if (IS_CLOUD) {
   app.post(
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -2086,7 +2086,7 @@ export async function postRemoveManagedWarehouseLegacyIdentifier(
     context.permissions.throwPermissionError();
   }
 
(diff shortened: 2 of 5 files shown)


tzjames (approved):
Looks good.  Though the cursorbot's suggestion here: https://github.com/growthbook/growthbook/pull/6340/changes#r3567832321 is a good one.  I'd implement that before landing.

PR #6341, opened 2026-07-13
Add bump-on-write ETag for /organization/definitions

Kevin-Chant:
### Features and Changes

`/organization/definitions` has a ~94% ETag hit rate, but computing the ETag today requires running the full endpoint (14 collection reads + transforms), so a 304 costs as much as a 200. This returns the 304 from a single indexed point-read before those reads, gated behind the `definitions-etag-304` flag. Builds on #6334.

The ETag is `"v<version>-<orgId>-<permissionsFingerprint>-b<buildFingerprint>[-p<projectsHash>][-<configHash>]"` (the build fingerprint — git SHA, or a random per-process id on SHA-less builds — covers code-only response changes, so every deploy invalidates all cached copies):

- **Bump-on-write version.** A per-org `definitionsversions` doc holds a monotonic `$inc` counter that every definitions-relevant write increments (a counter, not a timestamp, so same-millisecond writes can't collide, and it moves on deletes without relying on `dateUpdated`). BaseModels set `affectsDefinitionsVersion`; the legacy models and raw writes call `touchDefinitionsVersion` after the DB write commits. A read-coverage guard test asserts every collection the endpoint reads — including the nested event-forwarder configs inside each datasource — is covered.
- **Per-project scoping.** The response is permission-filtered by project, so the counter is split into a global counter plus a per-project map. A write bumps only the affected project(s) (`definitionsVersionProjectField` on BaseModels, `definitionsScope(...)` in legacy models; the union of old+new projects on a reassignment), and a reader's ETag folds in only the versions of the projects they can read — so a write in project A no longer invalidates a project-B-only reader. Org-wide resources, all-projects writes, and any unknown scope bump globally, a safe superset that never serves stale.
- **Skip response-invisible writes.** `definitionsVersionExcludedFields` skips the bump when the only change is a field the response projects out — saved-group `values`/`condition` (the latter dropped from the response by #6457), constant `value`/`environmentValues`, config `value`, webhook-secret `value` — without suppressing `dateUpdated`. Accepted tradeoff: since `dateUpdated` still moves and is in the response, a 304 after a values-only write serves a stale `dateUpdated` for that row (e.g. the saved-groups list's "Date Updated" column) until the next bumping write.

The ETag must include the org id (the URL is shared across orgs via `X-Organization`, so bare versions and permission fingerprints false-match cross-org) and the permission fingerprint (the response is permission-filtered). Under `usingFileConfig()`, file-managed resources bypass the Mongo writes that bump the version, so a hash of the parsed config joins the ETag. The endpoint sends `Cache-Control: private, no-cache` + `Vary: X-Organization`, and logs a `definitions-endpoint` event (latency + status) so the short-circuit can be evaluated as an experiment.

A failed bump never fails the user's write (logged, retried once) and self-heals on the next successful write; the whole fast path sits behind the flag as a kill switch.

Note: the flag name `definitions-etag-304` is a placeholder pending the flag/experiment plan.

### Testing

- `pnpm --filter back-end type-check`, lint, and unit tests (ETag composition incl. the per-project readable-subset, `definitionsScope`, `If-None-Match` parsing): clean
- Read-coverage guard test against an in-memory Mongo; verified it fails when a covered collection is removed from the list
- Live against dev with the flag on:
  - 200 returns our strong `"v…"` ETag (not overwritten by Express), with `Cache-Control`/`Vary` set; conditional GET → 304 with an empty body
  - a project-scoped create bumped only that project's counter; an all-projects / org-wide write bumped the global counter
  - a values-only saved-group update did not bump; a name change did
  - a project-B-only reader: a project-A write left their ETag unchanged (304), a project-B write invalidated it (200)


packages/back-end/src/models/DefinitionsVersionModel.ts
```diff
@@ -0,0 +1,67 @@
...
+      { organization },
+      { $inc: { version: 1 }, $set: { dateUpdated: new Date() } },
+      { upsert: true },
+    );
+  } catch (e) {
+    logger.error(
+      e,
+      `Failed to bump definitions version for organization ${organization}`,
+    );
+  }
```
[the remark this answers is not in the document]
Kevin-Chant: Intentional, the definitions write has already committed by the time this runs, so failing here would surface a 500 on a write that actually succeeded. A missed bump self-heals on the next successful write, and the whole fast path is behind the definitions-etag-304 flag as a kill switch. Accepting the rare, self-correcting staleness is the intended tradeoff.

packages/back-end/src/models/BaseModel.ts
```diff
@@ -1176,13 +1195,21 @@ export abstract class BaseModel<
...
     operations: AnyBulkWriteOperation[],
   ) {
+    if (this.config.affectsDefinitionsVersion) {
+      // Tripwire: ops span orgs, so there is no single org whose definitions
+      // version we can bump. Call touchDefinitionsVersion for each affected
+      // org manually if a definitions model ever needs this.
+      throw new Error(
+        "_dangerousBulkWriteCrossOrganization is not supported on models with affectsDefinitionsVersion",
+      );
+    }
```
tzjames, line 1270: I take it there is nowhere in the code now that calls this function from a model with affectsDefinitionsVersion?

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -166,6 +171,31 @@ export async function getDefinitions(req: AuthRequest, res: Response) {
     throw new Error("Must be part of an organization");
   }
 
+  // Short-circuit with a cheap 304 before the expensive reads below when the
+  // client's cached copy is still current. The ETag combines the org's
+  // definitions version (bumped by every relevant write via
+  // touchDefinitionsVersion) with the user's permission fingerprint (the
+  // response is permission-filtered). Skipped under config.yml, whose
```
tzjames: I'm not sure how common it is for people to be using the usingFileConfig. Theoretically, if it is true, we could just add the file hash as part of the etag?

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -166,6 +171,31 @@ export async function getDefinitions(req: AuthRequest, res: Response) {
...
+  // metrics/dimensions/datasources bypass the Mongo writes that bump the
+  // version, and gated behind a feature flag.
+  if (req.gb?.isOn("definitions-etag-304") && !usingFileConfig()) {
```
tzjames: Gating behind a feature flag is good, we could theoretically turn it on for awhile then turn it off and then do an analysis in datadog/cloudwatch.  However, it might be easier and a good chance to dogfood to make this into a full experiment.  We can have:
```
export async function getDefinitions(req: AuthRequest, res: Response) {
  const startTime = Date.now();
  const orgId = context.org.id;
  if (!orgId) {
    throw new Error("Must be part of an organization");
  }

  res.once("finish", () => {
    req.gb?.logEvent("definitions-endpoint", {
      latencyMs: Date.now() - startTime,
      statusCode: res.statusCode,
    });
  });
```
At the top of the definitions file, and then make fact metrics for etag hit rate, 200 response time, 304 response time, and avg response time (hit rate*304 response time + (1-hit rate) *200 response time).  It might then be easier to evaluate whether this change is worth it with real world data.

tzjames (changes requested):
This is looking good and is well thought out. I'm glad you caught that both the org Id and permissions affect the response.  I did leave a few comments.

I think the biggest danger to correctness is if a model forgets to bump on a write.
We should add something to the AI .md files about if the organization/defintions endpoint changes to be dependent on a new model make sure the models bump the version, or if a model adds a new write function to make sure to bump the version from the new function.

I had AI do another pass at finding other things the endpoint is dependent on and it found:
Confirmed gap: EventForwarderConfigModel

Each datasource in the definitions response is enriched with event-forwarder config:
```
// datasourceResponse.ts:33-46
export async function getDataSourceWithParams(context, integration) {
  const eventForwarderConfig = getEventForwarderSinkTypeForDatasource(integration.datasource)
    ? await getEventForwarderMetadataForDatasource(context, integration.datasource)
    : null;
  return buildDataSourceWithParams(integration, eventForwarderConfig);
}
```
EventForwarderConfigModel is a normal BaseModel but has no affectsDefinitionsVersion: true, and doesn't appear anywhere in the touchDefinitionsVersion call sites. So editing a datasource's event-forwarder settings changes what /organization/definitions should return, but won't bump the version — a real stale-304 risk, same shape as the tags bug from earlier, just in a different model.

packages/back-end/src/models/BaseModel.ts
Kevin-Chant: Yep, it's a simple guard to prevent future usage (so a dev wanting to use that bulkwrite will need to figure out the right approach)

tzjames (changes requested):
These are cases where the etag might change even though the output doesn't change, making the hit rate drop compared to the status quo:

Biggest one, structural: the version is per-org, not per-project. Nothing scopes it to projects, but the response itself is permission-filtered by project (canReadMultiProjectResource). So in any org with multiple active projects, a write to a metric/segment/fact-table in Project A bumps the same counter used by a user who only has access to Project B — invalidating their cache even though their own permission-filtered response is byte-identical. In a large multi-team org where different teams edit different projects constantly, this is probably the single biggest hit-rate tax, and it's inherent to the org-wide-counter design, not a bug in any one model.

Field-level cases — writes to a field the definitions payload strips out, with no skipDateUpdatedFields gate to catch it:

SavedGroupModel.getAllWithoutValues() projects out values entirely ({values: 0}), but nothing in the model config excludes values from the dateUpdated/version-bump decision. Saved groups with large ID lists that get re-synced periodically (CSV re-upload, "refresh from warehouse" jobs — a very normal saved-group workflow) will bump the version on every sync even though the definitions response for that saved group never changes.
ConstantModel.getAllWithoutValues() projects out value and environmentValues the same way. Constants are explicitly designed to hold values that change over time, so this is plausibly one of the more frequent offenders.
WebhookSecretModel.getAllForFrontEnd() omits value — same shape, but secret rotation is probably rare enough in practice that this matters less.

Kevin-Chant:
@cursoragent review

tzjames (commented):
This change looks good to go, if we want it. I don't think it is missing anything, nor can I think of a better way of doing it.  However I feel like it is a whole lot of lines of code and complexity and risk for some indeterminate amount of performance gain.  I feel like with AI vibe coding these days, the complexity and risk is mitigated enough by the updates to the .md files. However the risks remain that either we don't bump the version when the output of  organization/definitions would change and hence we have inaccurate data, or we do bump the version too often when something inconsequential gets changed and hence users get 200s and download data where before this change they would see 304s, making the situation worse.

By wrapping this in a back-end test, we should be able to see in the real world, for cloud orgs at least, whether latency and cache hit rate are improved with this PR or not.  So the danger of bumping the version to often should be caught.  That leaves just the danger that we don't bump it when we should.

tzjames (approved):

Kevin-Chant:
@cursoragent review

PR #6342, opened 2026-07-13
fix: handle duplicate attributes in rule conditions

### LINKED ISSUE #6279: [Feature] Prevent duplicate attribute selection in rule conditions.
Currently, it's possible to add multiple conditions using the same attribute (e.g. multiple "is not equal to" conditions on billingPlanCategory), but only the last condition is preserved in the generated JSON because conditions are stored as key–value pairs. This can lead to unexpected behavior and confusion.

It would improve the UX if the rule builder prevented selecting an attribute that's already in use (or at least warned the user), unless the chosen operator supports multiple values (e.g. "is none of" → $nin).

This would make the UI better reflect how conditions are actually serialized and help users avoid creating rules that don't behave as expected.

akramcodez:
Closes #6279

### Features and Changes

This PR improves the rule builder UX by warning users when multiple conditions on the same attribute will overwrite each other during serialization.

Previously, users could create multiple conditions for the same attribute (for example, multiple `is` or `is not` conditions), but only the last condition would be preserved in the generated JSON because of how `condToJson` serializes conditions. This could lead to confusing and unexpected behavior.

### What's changed

* Added conflict detection based on the serialization behavior of `condToJson`.
* Displays an inline warning when two conditions on the same attribute serialize to the same JSON key and one would overwrite the other.
* Preserves valid combinations (such as numeric ranges using different operators) by warning only when an actual serialization conflict exists.
* Keeps existing rule-building behavior intact while providing users with clear feedback before saving a conflicting rule.

### Testing

* Verified the project compiles successfully with:

  ```bash
  pnpm --filter front-end type-check
  ```

* Added/updated tests for the warning behavior in `ConditionInput.test.tsx`.

* Manually verified:

  * Duplicate conflicting conditions display an inline warning.
  * Valid combinations (for example, range conditions using different operators) do not display a warning.
  * Existing configurations continue to render correctly.

### DIFF
--- a/packages/front-end/components/Features/ConditionInput.tsx
+++ b/packages/front-end/components/Features/ConditionInput.tsx
@@ -98,8 +98,35 @@ const BASE_OPERATOR: Record<string, string> = {
   $ini: "$in",
   $nini: "$nin",
   $regexi: "$regex",
-  $notRegexi: "$notRegex",
+  $notRegexi: "$notRegexi",
 };
+
+/**
+ * Returns the internal object keys used by `condToJson` for a given operator.
+ * This is used to detect if two conditions on the same attribute will overwrite each other.
+ */
+function getSerializationKeys(operator: string): string[] {
+  if (
+    operator === "$notRegex" ||
+    operator === "$notRegexi" ||
+    operator === "$notIncludes"
+  ) {
+    return ["$not"];
+  }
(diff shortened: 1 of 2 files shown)


akramcodez:
PTAL @august-growthbook thanks!

najito:
The Greptile asks are important so they should be addressed.

PR #6346, opened 2026-07-13
Fix: enforce datasource project access on external dimension read APIs

Kevin-Chant:
### Features and Changes

Dimensions were scoped by organization only across their read and write surfaces. A dimension has no `projects` field of its own — its project access is inherited from the parent datasource, and dimension permissions (`createDimensions`) are global — so a caller restricted to a subset of projects could read dimension SQL from, and mutate, dimensions tied to datasources in projects they were excluded from. This affected the external REST API (reachable by a project-restricted API key) and the internal endpoints.

Access is now enforced from the parent datasource everywhere, mirroring how metrics and segments already gate project access:

- **Read** — `hasDimensionDatasourceAccess` (new helper) gates `GET /api/v1/dimensions/:id`, and the list endpoints (`GET /api/v1/dimensions`, internal `GET /dimensions`) and the internal `/definitions` endpoint filter dimensions to accessible datasources. In `/definitions`, dimensions were previously the only definition type returned unfiltered alongside already-scoped metrics/datasources/segments.
- **Write** — update and delete (external API + internal controller) now verify access to the dimension's existing datasource before mutating. Previously they ran only the global `createDimensions` permission, letting a restricted caller edit or delete dimensions in projects they couldn't access. Create paths already validate the target datasource.
- **Orphaned dimensions** — a dimension whose datasource is inaccessible *or no longer exists* is treated as not found across all these endpoints. Deleting a datasource soft-deletes its dimensions, so an orphaned dimension is legacy-only; rather than special-case it, access is gated purely on datasource readability (`getDataSourceById`, which returns null for both an access-denied and a deleted datasource). Orphans therefore become invisible and unmanageable through these endpoints for every caller — the intended behavior per review.

### Testing

- [x] Caller restricted to Project A, dimension's datasource in Project B → `GET /api/v1/dimensions/:id` returns "Could not find dimension with that id"; `PUT`/`DELETE` likewise rejected
- [x] Same caller → `GET /api/v1/dimensions`, internal `GET /dimensions`, and `/definitions` all omit Project B dimensions
- [x] Unrestricted caller → all read endpoints return every dimension as before
- [x] Dimension whose datasource was deleted → omitted from the list endpoints and GET-by-id (treated as not found)

Verified on a local dev server using a PAT for a member restricted to Project A (`noaccess` global + project role on A) against a dimension whose datasource sat in Project B, plus a control run on `main` confirming the leak this closes: on `main`, the restricted PAT's `GET /api/v1/dimensions/:id` returned the Project B dimension including its SQL; on this branch the same call returns "Could not find dimension with that id", and the dimension disappears from `GET /api/v1/dimensions`, internal `GET /dimensions`, and `/definitions` for that caller. Update/delete as the restricted caller were rejected (403 from the global `createDimensions` check, which runs ahead of the new datasource gate). An unrestricted PAT saw all dimensions on every read endpoint and could update the Project B dimension.

### DIFF
--- a/packages/back-end/src/api/dimensions/deleteDimension.ts
+++ b/packages/back-end/src/api/dimensions/deleteDimension.ts
@@ -16,11 +16,11 @@ export const deleteDimension = createApiRequestHandler(
   const organization = req.organization.id;
   const dimension = await findDimensionById(req.params.id, organization);
 
-  if (
-    !dimension ||
-    !(await hasDimensionDatasourceAccess(req.context, dimension))
-  ) {
+  if (!dimension) {
     throw new Error("Could not find dimension with that id");
+  }
+  if (!(await hasDimensionDatasourceAccess(req.context, dimension))) {
+    throw new Error("You don't have access to this dimension");
   }
 
   await deleteDimensionById(req.context, dimension);
--- a/packages/back-end/src/api/dimensions/deleteDimension.ts
+++ b/packages/back-end/src/api/dimensions/deleteDimension.ts
@@ -2,6 +2,7 @@ import { deleteDimensionValidator } from "shared/validators";
 import { createApiRequestHandler } from "back-end/src/util/handler";
 import {
   findDimensionById,
+  hasDimensionDatasourceAccess,
   deleteDimensionById,
 } from "back-end/src/models/DimensionModel";
 
@@ -15,7 +16,10 @@ export const deleteDimension = createApiRequestHandler(
   const organization = req.organization.id;
   const dimension = await findDimensionById(req.params.id, organization);
 
-  if (!dimension) {
+  if (
+    !dimension ||
+    !(await hasDimensionDatasourceAccess(req.context, dimension))
+  ) {
     throw new Error("Could not find dimension with that id");
   }
 
--- a/packages/back-end/src/api/dimensions/getDimension.ts
+++ b/packages/back-end/src/api/dimensions/getDimension.ts
@@ -3,6 +3,7 @@ import {
   findDimensionById,
   toDimensionApiInterface,
 } from "back-end/src/models/DimensionModel";
+import { getDataSourceById } from "back-end/src/models/DataSourceModel";
 import { resolveOwnerEmail } from "back-end/src/services/owner";
 import { createApiRequestHandler } from "back-end/src/util/handler";
 
@@ -12,7 +13,12 @@ export const getDimension = createApiRequestHandler(getDimensionValidator)(
       req.params.id,
       req.organization.id,
     );
-    if (!dimension) {
+    // A dimension's project access is inherited from its datasource, which
+    // returns null when the caller lacks access to its projects.
+    if (
+      !dimension ||
+      !(await getDataSourceById(req.context, dimension.datasource))
+    ) {
       throw new Error("Could not find dimension with that id");
     }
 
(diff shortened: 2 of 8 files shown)


packages/back-end/src/models/DataSourceModel.ts
```diff
@@ -125,6 +125,23 @@ export async function getDataSourcesByOrganization(
   );
 }
 
+// Returns the ids of every datasource in the org, ignoring project access.
+// Used to distinguish an inaccessible datasource from a deleted one.
+export async function getAllDatasourceIdsByOrganization(
```
tzjames: Since this bypasses permissions, I think it would be good to call it
`dangerouslyGetAllDatasourceIdsByOrganizationBypassPermissions`

tzjames (changes requested):
We have a soft check when deleting datasources to first delete all dimensions tied to that datasource.  Hence orphaned datasources shouldn't happen. If they do exist, it would only be legacy.  I think it is fine to pretend they don't exist for everyone rather than have special code to check for this edge case, and have dimensions appear in the ui with no datasource attached.

packages/back-end/src/api/dimensions/deleteDimension.ts
```diff
@@ -15,7 +16,10 @@ export const deleteDimension = createApiRequestHandler(
   const organization = req.organization.id;
   const dimension = await findDimensionById(req.params.id, organization);
 
-  if (!dimension) {
+  if (
+    !dimension ||
+    !(await hasDimensionDatasourceAccess(req.context, dimension))
```
tzjames: You could have a different message "You don't have access to this dimension" for this case.

packages/back-end/src/api/dimensions/getDimension.ts
```diff
@@ -12,7 +13,10 @@ export const getDimension = createApiRequestHandler(getDimensionValidator)(
       req.params.id,
       req.organization.id,
     );
-    if (!dimension) {
+    if (
+      !dimension ||
+      !(await hasDimensionDatasourceAccess(req.context, dimension))
```
tzjames, line 18: You could have a different message "You don't have access to this dimension" for this case.

packages/back-end/src/api/dimensions/updateDimension.ts
```diff
@@ -22,7 +23,10 @@ export const updateDimension = createApiRequestHandler(
   const organization = req.organization.id;
   const dimension = await findDimensionById(req.params.id, organization);
 
-  if (!dimension) {
+  if (
+    !dimension ||
+    !(await hasDimensionDatasourceAccess(req.context, dimension))
```
tzjames: You could have a different message "You don't have access to this dimension" for this case.

packages/back-end/src/routers/dimension/dimension.controller.ts
```diff
@@ -139,7 +152,7 @@ export const putDimension = async (
   const { id } = req.params;
   const dimension = await findDimensionById(id, org.id);
 
-  if (!dimension) {
+  if (!dimension || !(await hasDimensionDatasourceAccess(context, dimension))) {
```
tzjames: You could have a different message "You don't have access to this dimension" for this case.

packages/back-end/src/routers/dimension/dimension.controller.ts
```diff
@@ -194,7 +207,7 @@ export const deleteDimension = async (
   const { org } = context;
   const dimension = await findDimensionById(id, org.id);
 
-  if (!dimension) {
+  if (!dimension || !(await hasDimensionDatasourceAccess(context, dimension))) {
```
tzjames: You could have a different message "You don't have access to this dimension" for this case.

tzjames (approved):

PR #6349, opened 2026-07-13
Add EF Form Validation

nhat-growthbook:
### Features and Changes

Reworked the Event Forwarder creation/edit modal so the **Confirm** button no longer hides *why* a configuration is invalid.

Previously the button was disabled whenever the config was incomplete, and a single generic tooltip ("Enter ... required connection fields") lumped every failure together — so reasons like *auth method must be key-pair* or an *invalid table prefix* were never surfaced to the user.

Now:
- **Confirm** is enabled once the US-data-flow consent checkbox is acknowledged; its tooltip is reduced to just the consent reminder.
- Configuration is validated **on submit**. The new `getEventForwarderValidationErrors()` helper (replacing the boolean `getCanConfirmEventForwarder`) returns one specific message per failing condition — per-field for BigQuery and Snowflake, plus the table-prefix normalizer's own error — and the modal throws the combined message so it renders in the error area.
- Surfaces two reasons that were previously invisible: auth method must be key-pair, and invalid table prefix.

Behaviour preserved: private-key value is still not required (the prior logic only required key-pair auth and never validated the key string), avoiding a regression in the edit flow where the secret isn't re-sent.

### Dependencies

None

### Testing

1. Open a BigQuery or Snowflake data source in **Settings → Data Sources** and launch **Set Up Event Forwarder**.
2. Leave the consent checkbox unchecked → **Confirm** is disabled and its tooltip explains the consent requirement.
3. Check consent but leave required fields blank / use password auth / enter an invalid table prefix → click **Confirm** → a combined error message lists each specific problem, and no request is sent.
4. Fill in all required fields with key-pair auth and a valid prefix → **Confirm** succeeds and the forwarder is created.
5. Edit an existing forwarder (consent pre-checked) and confirm it still saves without re-entering the private key.

### DIFF
--- a/packages/front-end/components/Settings/EditDataSource/EventForwarder/EventForwarder.tsx
+++ b/packages/front-end/components/Settings/EditDataSource/EventForwarder/EventForwarder.tsx
@@ -179,6 +179,12 @@ function getEventForwarderDraft(
   return null;
 }
 
+// Only validates what the browser's native `required` validation can't cover:
+// read-only/derived fields (Snowflake access URL), datasource connection params
+// that aren't inputs in this modal (account/username/auth method), and table
+// prefix *format*. Empty visible required fields (BigQuery project/dataset,
+// Snowflake database/schema) are handled by native `required` on the inputs, so
+// don't re-check them here — that would duplicate the per-field tooltip.
 function getEventForwarderValidationErrors(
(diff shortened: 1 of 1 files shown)


mknowlton89 (approved):
Looks good - agree with the P2 from greptile, but approving

PR #6351, opened 2026-07-13
product analytics funnels

lillialjackson:
### Features and Changes

Includes work to allow users to create funnels in the product analytics explorations and save them to dashboards

### Dependencies

Create product-analytics-funnels flag

### Testing

Unit and manual
For manual testing, I tested:
- Multiple steps with 1+ filters
- optional steps
- step with a conversion window
- multiple steps with a groupby filter
- steps with multiple filters
- steps from different fact tables
* There was not great data available for testing multiple fact tables especially on snowflake but queries did run without error.
The manual testing was done for clickhouse, postgres, snowflake, and bigquery dialects

### Screenshots

Screenshot 2026-07-13 at 10 59 19 AM

Screenshot 2026-07-13 at 10 59 04 AM

Screenshot 2026-07-13 at 10 58 09 AM
Screenshot 2026-07-13 at 10 58 09 AM


mknowlton89 (commented):
I'm still going through the code, but wanted to share a few some comments on the UX.

1. Do we want to support the `Compare` feature here? In my testing it doesn't seem like it's actually working, so we may just want to omit it for now?
Image


2. I noticed there seemed to be some bug with the date granularity not auto-populating?
Image

3. I would love to update the sidebar with some drag and drop functionality - not necessary for the initial release, but it would be nice to be able to move the items around more easily.
Image

4. When viewing the chart, we're not always rendering every column's title, which makes it had to read/understand the graph - you have to hover on some columns to know what the column represents.
Image

packages/back-end/src/enterprise/services/product-analytics.ts
```diff
@@ -172,16 +212,33 @@ export async function runProductAnalyticsExploration(
...
+      // rejection after the timeout wins; errors are persisted to the model
+      // and surfaced to the client via polling.
+      queryRunner.waitForResults().catch(() => {}),
+      timeout,
+    ]);
   } catch (e) {
     // Ignore errors here, still return the model
+  } finally {
+    // Clear the timer so the fast path doesn't leave a dangling timeout.
+    if (syncTimer) clearTimeout(syncTimer);
```
mknowlton89, line 259: It looks like this code is implementing the comment of adding a 5 second timeout, but from what I can see it doesn't look like we've built the polling on the front-end. Do we need to include this change in this PR, or can we just go back to the `await queryRunner.waitForResults()`?

If we do want this, we should build the front-end polling

packages/back-end/src/enterprise/services/product-analytics.ts
```diff
@@ -131,6 +138,39 @@ export async function runProductAnalyticsExploration(
...
+    // Load every fact table referenced by any funnel step. We don't fold
+    // ids into a Set first because the order in `dataset.steps` is
+    // significant for SQL generation; getFactTablesByIds dedupes for us.
+    const factTableIds = Array.from(
+      new Set(dataset.steps.map((s) => s.factTable).filter(Boolean)),
+    );
+    if (factTableIds.length === 0) {
+      throw new BadRequestError("Funnel steps require fact tables");
+    }
+    const factTables = await getFactTablesByIds(context, factTableIds);
```
mknowlton89, line 172: Invalid funnel configs can leave explorations permanently stuck in running.

funnelDatasetValidator permits unit: null and any number of steps. The stricter checks occur later in buildFunnelSql()—after runProductAnalyticsExploration() has created a model with status: "running".

If SQL generation throws (for example, fewer than two steps, a missing unit, or a unit unavailable on a step’s fact table), startAnalysis() exits before creating a query. The service then swallows the exception and returns the unchanged model. Since there is no query or runner callback to update it, polling will always receive running.

Could we validate these semantic requirements before creating the exploration?

```ts
if (dataset.steps.length < 2) {
  throw new BadRequestError("Funnels require at least two steps");
}
if (!dataset.unit) {
  throw new BadRequestError("Funnel unit is required");
}
// After loading the fact tables:
for (const factTable of factTables) {
  if (!factTable.userIdTypes.includes(dataset.unit)) {
    throw new BadRequestError(
      `Funnel unit "${dataset.unit}" must exist on every step's fact table`,
    );
  }
}
```

mknowlton89: I think in most cases the front-end protects against this, but I think it could be possible to get into this situation via the API maybe?

packages/front-end/enterprise/components/ProductAnalytics/ExplorerContext.tsx
```diff
@@ -472,64 +495,124 @@ export function ExplorerProvider({
...
+        pollTimerRef.current = null;
       }
 
-      // Set staleness to false and update submitted state when there is a result
-      if (fetchResult) {
+      // If the backend returned a still-running exploration (query exceeded the
+      // ~5s sync budget), don't treat the empty in-progress model as final —
+      // keep the loading state and poll by id until it reaches a terminal
+      // state. The warehouse query keeps running server-side regardless.
+      if (!fetchError && fetchResult?.status === "running" && fetchResult.id) {
```
mknowlton89: Should we also poll comparison.exploration when it is returned with status: "running"?

Although the backend includes both explorations in the same response, each independently completes or hits the five-second timeout, so the primary can be terminal while the comparison is still running. This branch only polls the primary ID, leaving the comparison state at its initial running/empty value.

I think the dashboard polling mechanism handles this already - it splits the primary query and the comparison separately and share the polling logic.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/FunnelStepCard.tsx
```diff
@@ -0,0 +1,533 @@
...
+                setNameDraft(step.name);
+                setIsEditingName(true);
+              }}
+            >
+              <Text
+                weight="medium"
+                truncate
+                as="div"
+                whiteSpace="nowrap"
+                title="Double-click to rename"
```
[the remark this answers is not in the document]
mknowlton89: Probably a good idea to implement for consistency

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -1702,6 +1711,14 @@ export default function EditSingleBlock({
                 onSaveAndClose={submit}
               />
             )}
+            {block.type === "funnel-exploration" && (
+              <ProductAnalyticsExplorerSettings
+                block={block}
+                setBlock={setBlock}
+                saveAndCloseTrigger={saveAndCloseTrigger}
+                onSaveAndClose={submit}
+              />
```
mknowlton89, line 1768: Given that funnels are included in the list of blocks that support global dashboard filters, we need to add the `dashboardGlobalControls={dashboardGlobalControls}` prop

```suggestion
              <ProductAnalyticsExplorerSettings
                block={block}
                setBlock={setBlock}
                saveAndCloseTrigger={saveAndCloseTrigger}
                dashboardGlobalControls={dashboardGlobalControls}
                onSaveAndClose={submit}
              />
```

mknowlton89 (approved):

PR #6354, opened 2026-07-14
Add "start on approval" option for ramp schedules

bryce-fitzsimons:
Holds a ramp at step -1 with the rule off until a human approves the start. Also fixes the re-entrant lock on confirm/approve start and softens the "behind live" draft warnings.

image

image


gazzdingo (approved):

gazzdingo:
edge cases we might want to fix  or think about 

- Unchecking on approval on a held ready schedule never persists the change before start, so the start-approval assertion throws and the publish fails. The gate can’t be turned off this way.

- The poller’s early return skips pending schedules, so if the publish hook defers or crashes the schedule is stuck forever, and approving it is rejected because it’s still pending.

- An approval-gated guardrail rollback doesn’t set the rollback reason, so the status just shows “Awaiting approval” with no indication a guardrail tripped — re-approving relaunches into the same failure.

- The awaiting-approval webhook isn’t gated by the emit flag, so a restart fires it twice.

- The “publish disabled” helper is true whenever start-approval is set (no status check) and stays true after start, so editing a running 0-step approval schedule can disable a live rollout. Only reachable via API-created 0-step schedules.

PR #6357, opened 2026-07-14
New Slack App & Integration

Auz:
A whole new Slack app, with visual cards shared to any channel. Also features a new UI from within GrowthBook to easily control what is shown to which channel.


packages/back-end/src/routers/slack-actions/slack-actions.router.ts
```diff
@@ -0,0 +1,352 @@
...
+    if (!webhook || !experimentId) {
+      return res.json({ text: "Unable to snooze this notification." });
+    }
+
+    await snoozeSlackExperimentNotifications({
+      organizationId: webhook.organizationId,
+      eventWebHookId: webhook.id,
+      experimentId,
+      snoozedUntil: new Date(Date.now() + 24 * 60 * 60 * 1000),
+    });
```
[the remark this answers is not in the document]
Auz: @greptileai review this again please

Auz:
@greptileai review

Auz:
@greptileai review

Auz:
@greptileai review

packages/back-end/src/jobs/eventWebhookDailyDigest.ts
```diff
@@ -0,0 +1,85 @@
...
+    .limit(200)
+    .lean<
+      {
+        id: string;
+        organization: string;
+        project?: string;
+        tags?: string[];
+        dateUpdated?: Date;
+      }[]
+    >();
```
jdorn: This doesn't look right. On Cloud, this query matches 10,000s of feature flags across all orgs and limits to the ~same random 200 every time.  Then, it creates `feature.stale.candidate` events for each of these the first time. Each subsequent run will do the 200 `EventModel.findOne()` calls and will see they were all already created and skip.  So it's basically a really expensive no-op every hour.

Not sure what the intention is of this job, but it almost certainly does not do what you want it to do.

packages/shared/src/validators/event-webhook.ts
```diff
@@ -15,6 +15,265 @@ export const eventWebHookMethods = ["POST", "PUT", "PATCH"] as const;
...
+
+// Whether a resolved digest should fire at `now`. Jobs run hourly, so we match
+// on the hour plus the frequency's day rule. `custom` is anchored to the Unix
+// epoch so "every N days" is deterministic across runs.
+export const isSlackDigestDue = (
+  r: ResolvedSlackDigest,
+  now: Date,
+): boolean => {
+  if (!SLACK_DIGEST_LIVE_FREQUENCIES.has(r.frequency)) return false;
+  if (now.getUTCHours() !== r.hourUtc) return false;
```
jdorn: Agenda does not have exact-once semantics for hourly jobs like this.  So this will cause double digests to get sent to users (or missed digests) occasionally.  Would be much more robust if we stored a timestamp in Mongo.  Best would be storing `nextDigestAt` timestamp and then we can restrict which ones are eligible by looking for when this date is in the past.  We do the same thing with experiment updates.

packages/back-end/src/models/EventWebHookCoalesceBucketModel.ts
```diff
@@ -0,0 +1,188 @@
...
+  id: string;
+  organizationId: string;
+  eventWebHookId: string;
+  objectType: NotificationEventResource;
+  objectId: string;
+  eventIds: string[];
+  firstSeenAt: Date;
+  lastSeenAt: Date;
+  flushAt: Date;
+}
```
jdorn, line 22: Don't use Mongoose for new models. Use BaseClass instead.

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -468,6 +468,18 @@ export async function getExperimentById(
...
+  organizationId: string,
+  id: string,
+): Promise<ExperimentInterface | null> {
+  const doc = await getCollection(COLLECTION).findOne({
+    organization: organizationId,
+    id,
+  });
+
+  return doc ? toInterface(doc) : null;
+}
```
jdorn: We already have `getExperimentById` that takes context (which has the org id).  I'd prefer not adding another almost identical method

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -691,6 +691,37 @@ export async function getFeaturesByIds(
...
+export async function getFeatureLinkedExperimentIds(
+  organizationId: string,
+  featureId: string,
+): Promise<string[]> {
+  const feature = await FeatureModel.findOne(
+    { organization: organizationId, id: featureId },
+    { linkedExperiments: 1, _id: 0 },
+  ).lean<{ linkedExperiments?: string[] }>();
+  return feature?.linkedExperiments || [];
+}
```
jdorn, line 721: For both of these, take `context` instead of `organizationId` as an argument.

packages/back-end/src/models/SlackNotificationSnoozeModel.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+const slackNotificationSnoozeSchema =
+  new mongoose.Schema<SlackNotificationSnoozeInterface>({
+    organizationId: { type: String, required: true },
+    eventWebHookId: { type: String, required: true },
+    experimentId: { type: String, required: true },
+    snoozedUntil: { type: Date, required: true },
+    dateCreated: { type: Date, required: true },
+  });
```
jdorn: Same here - do not use Mongoose for new models.

packages/back-end/src/models/SlackUserLinkModel.ts
```diff
@@ -0,0 +1,67 @@
...
+  slackTeamId: string;
+  slackUserId: string;
+  // The org the link was created from, for reference/audit only. The acting org
+  // is resolved per-mention by channel/team routing, and membership re-checked
+  // at use time.
+  organizationId: string;
+  growthbookUserId: string;
+  dateCreated: Date;
+  dateUpdated: Date;
+};
```
jdorn, line 18: Same - don't use Mongoose

Auz:
All addressed — two I'd like your call on at the end.

- **`eventWebhookDailyDigest`** — deleted. Its body was already gated behind a `= false` constant, so it scheduled hourly and did nothing. No behavior change.
- **Digest scheduling** — now driven by stored `nextExperimentDigestAt` / `nextFeatureDigestAt` (indexed, queried `$lte: now`) and **claimed atomically** before sending, so a duplicated run finds nothing to claim. Next run is computed from the *due* time, not `now`, so a late job doesn't drift the cadence; pre-existing connections get seeded with a future first run rather than firing on deploy. +11 unit tests for the cadence math.
- **`getExperimentByIdForOrganization`** — removed; both call sites use `getExperimentById(context, id)`.
- **`FeatureModel` helpers** — take `context` now (as does `getFeatureMetricIds`); the handler builds one lazily and shares it.
- **`SlackNotificationSnooze`** — converted to BaseModel. One note: the old schema had a TTL index auto-expiring snoozes, and `additionalIndexes` doesn't support TTL, so expired rows now linger. Reads already filter `snoozedUntil > now`, so no behavior change.

Two Mongoose models I'd like your opinion on before converting:

**`EventWebHookCoalesceBucket`** — it's a concurrency primitive rather than a data model. Both ops are atomic: `findOneAndUpdate` with `upsert` + `$setOnInsert` + `$addToSet` + `rawResult` (concurrent events merge into one bucket, and only the *inserting* caller schedules the flush), and `findOneAndDelete` to claim-and-drain so a duplicated flush can't double-deliver. I can subclass BaseModel, but both ops would go through `_dangerousGetCollection()` — bypassing create/update, validation, and audit — so it'd be BaseModel in name only. Convert anyway, or leave it with a comment documenting the atomicity requirement?

**`SlackUserLink`** — the primary read is `getSlackUserLink({ slackTeamId, slackUserId })`, which runs when a Slack message arrives, *before* we know the org — resolving the org is what the lookup is for. BaseModel's constructor requires a context and scopes queries to `context.org.id`, so it'd need a context we can only build after the query. It's also intentionally a cross-org identity map, and authorization happens immediately after via `getContextForUserIdInOrg` (re-checking membership). Is there a system/unscoped-context pattern you'd want here, or is Mongoose + a comment on the bootstrap constraint OK?

docs/snippets/partials/event-webhook/_event-webhook-list.mdx
```diff
@@ -3111,6 +3174,12 @@ Triggered when a goal or guardrail metric reaches significance in an experiment
             statsEngine: string;
             criticalValue: number;
             winning: boolean;
+            metricRole?: ("goal" | "secondary" | "guardrail") | undefined;
+            uplift?: number | undefined;
+            ci?: [
+                number,
+                number
+            ] | undefined;
```
ahdriel, line 3182: are these really optional?

EXPERIMENT_IDEAS_PROJECT_PLAN.md
```diff
@@ -0,0 +1,779 @@
+# GrowthBook Experiment Ideas — AI Agent Project Plan
```
ahdriel: we should remove this

packages/back-end/src/app.ts
```diff
@@ -1175,6 +1183,9 @@ app.use(eventWebHooksRouter);
 // Slack integration
 app.use("/integrations/slack", slackIntegrationRouter);
 
+// Slack card/webhook test endpoints (admin-only)
+app.use("/admin/slack-test", slackTestRouter);
```
ahdriel, line 1187: question: do we want to keep this one?

packages/shared/types/slack-integration.d.ts
```diff
@@ -15,3 +15,37 @@ export interface SlackIntegrationInterface {
   linkedByUserId: string;
   slackSigningKey: string;
 }
+
+export interface SlackOAuthIntegrationInterface {
```
ahdriel, line 19: do we plan on removing `SlackIntegrationInterface` now that we have this one? or are we going to keep both?

packages/shared/types/slack-integration.d.ts
```diff
@@ -15,3 +15,37 @@ export interface SlackIntegrationInterface {
...
+  experiments: string[];
+  metrics: string[];
+  features?: string[];
+  environments: string[];
+  tags: string[];
+  coalesceWindowMs?: number;
+  slackOptions?: import("../src/validators/event-webhook").SlackEventWebHookOptions;
+  lastRunAt: Date | null;
+  lastState: "none" | "success" | "error";
+  slack?: {
```
ahdriel, line 37: should this really be optional?

packages/shared/types/events/event.d.ts
```diff
@@ -16,6 +16,8 @@ export interface BaseEventInterface<T, V> {
   dateCreated: Date;
   data: T;
   organizationId: string;
+  object?: NotificationEventResource;
+  objectId?: string;
```
ahdriel, line 20: why are we moving this outside of `data` into a high-level `object`?

packages/back-end/src/events/handlers/slack/slack-event-handler-utils.ts
```diff
@@ -282,6 +436,104 @@ export const getSlackMessageForNotificationEvent = async (
...
+// switch: the event union is huge and only these few map to a card.
+const compactEventForNotification = (
+  event: NotificationEvent,
+): CompactEvent | undefined => {
+  // A stop is emitted as shipped/rolledback, but the real outcome is in
+  // `results`: a non-ship stop can be inconclusive/dnf, i.e. a neutral
+  // "stopped" rather than a "rolled back / no lift" loss.
+  if (
+    event.event === "experiment.stopped.shipped" ||
+    event.event === "experiment.stopped.rolledback"
```
ahdriel, line 454: should we move the card generation stuff outside of the slack handler to make it more generic?

PR #6363, opened 2026-07-14
fix(safe-rollouts): Ensure we drop `null` customFields

ahdriel:
### Features and Changes

SafeRollouts, if they have been saved with `null` before we landed fixes to `BaseModel`, now throw a validation error.

This fixes it by coalescing from `null` to `(empty)` which is the correct API.

Note: this is a targeted fix. The proper fix is on the `BaseModel` but is riskier, so we are merging this to unblock users while we keep working on the `BaseModel` fix.


gazzdingo (approved):
lgtm

PR #6371, opened 2026-07-15
Add OAuth 2.1 authorization server for MCP/CLI login

oelshaikh:
## Summary

OAuth 2.1 authorization server for MCP (and future CLI) login — off by default (`OAUTH_AS_ENABLED`).

- Discovery, DCR, auth-code + PKCE, refresh rotation, revoke
- Opaque access tokens (`gbo_…`) stored hashed in `apikeys` (PAT-equivalent)
- Consent page at `/oauth/authorize`
- Connected Apps UI to list/revoke grants

## Data model

| Piece | Storage | Notes |
| --- | --- | --- |
| Auth codes | `OAuthAuthCodeModel` | BaseModel on `ReqContext` |
| Refresh tokens | `OAuthRefreshTokenModel` | BaseModel; rotation sets `consumedAt` (reuse detection) |
| Grants | `OAuthGrantModel` | Durable per-(org, client, user) revoke target |
| DCR clients | Mongoose helpers | Global (no org); idle TTL via `expiresAt` |

TTL indexes live on `BaseModel.additionalIndexes` (`expireAfterSeconds`). Unauthenticated token lookups use static `dangerous*` helpers.

## Lifecycle hardening

- **Membership** — mint/refresh requires a member context; removed users can't keep refreshing
- **Revoke** — client ownership required; marks grant revoked, then tears down refresh + access tokens
- **Revoke vs refresh race** — issuance re-reads the grant after writing and self-cleans if revoked
- **Reuse detection** — replaying a consumed refresh token tears down the whole grant (RFC 9700 §4.14.2)


packages/back-end/src/models/OAuthModels.ts
```diff
@@ -0,0 +1,153 @@
...
+  redirectUris: { type: [String], required: true },
+  tokenEndpointAuthMethod: { type: String, default: "none" },
+  grantTypes: [String],
+  responseTypes: [String],
+  scope: String,
+  clientUri: String,
+  dateCreated: { type: Date, default: Date.now },
+});
+
+export const OAuthClientModel = mongoose.model<OAuthClientInterface>(
```
jdorn: Don't export Mongoose models directly.  Doesn't look like they are being used anywhere

packages/back-end/src/models/OAuthModels.ts
```diff
@@ -0,0 +1,153 @@
...
+  resource: String,
+  used: { type: Boolean, default: false },
+  expiresAt: {
+    type: Date,
+    required: true,
+    // Mongo TTL: delete shortly after expiry (auth codes are ~10 min)
+    expires: 0,
+  },
+  dateCreated: { type: Date, default: Date.now },
+});
```
jdorn: Ideally, we should be using the BaseModel for all of these.  Or at least the ones with an `organization` field.  Provides audit logging, multi-tenant protections, and other guardrails out-of-the-box.

packages/back-end/src/services/oauth/index.ts
```diff
@@ -0,0 +1,426 @@
...
+import { ApiKeyInterface } from "shared/types/apikey";
+import { OAuthDcrRequest } from "shared/validators";
+import {
+  APP_ORIGIN,
+  OAUTH_ACCESS_TOKEN_TTL_SECONDS,
+  OAUTH_ISSUER,
+  OAUTH_REFRESH_TOKEN_TTL_SECONDS,
+} from "back-end/src/util/secrets";
+import { getCollection } from "back-end/src/util/mongo.util";
+import { COLLECTION_NAME as API_KEY_COLLECTION } from "back-end/src/models/ApiKeyModel";
```
jdorn: I don't feel great about accessing the collection directly from this service file.  We try to keep all db queries within the associated model files directly.

packages/back-end/src/services/oauth/index.ts
```diff
@@ -0,0 +1,529 @@
...
+  // new pair.
+  const doc = await context.models.oauthRefreshTokens.getByTokenHash(tokenHash);
+  if (doc) await context.models.oauthRefreshTokens.delete(doc);
+
+  return issueTokenPair(context, {
+    clientId: existing.clientId,
+    userId: existing.userId,
+    scope: existing.scope,
+    resource: existing.resource,
+  });
```
jdorn, line 425: Reuse detection (RFC 9700 §4.14.2 / OAuth 2.1) requires that presenting an already-consumed token invalidates the entire token family — every descendant refresh token and, ideally, the outstanding access tokens for that grant. The point is to defend the case where the attacker gets there first:

1. Attacker steals RT1, redeems it before the legit client → gets RT2, RT1 deleted.
2. Legit client later redeems RT1 → invalid_grant.
3. The attacker's RT2 stays valid, and the grant is never torn down.

With proper reuse detection, step 2 would fire the alarm and revoke RT2 + the grant. The current design cannot do this, because it can't tell "reused consumed token" from "unknown token" — the consumed row is gone. The only case rotation-with-delete handles is the legit-client-first race, which locks the attacker out but gives no compromise signal.

packages/back-end/src/models/OAuthRefreshTokenModel.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Org-scoped OAuth refresh tokens.
+ *
+ * Revoke/refresh look up by hash before the org is known
+ * ({@link dangerousFindByHash}); subsequent lifecycle ops use a ReqContext
+ * instance so deletes are org-scoped and audited.
+ */
+export class OAuthRefreshTokenModel extends BaseClass {
+  constructor(...args: ConstructorParameters<typeof BaseClass>) {
+    super(...args);
+    ensureExpiresAtTtlIndex();
```
jdorn: This is used multiple times in this PR, might be better to just implement in BaseModel.  Or at the very least, abstracting out the TTL index creation to a shared helper function.

jdorn (approved):

PR #6375, opened 2026-07-15
Emit revision.published event when reverting a feature

anna-yn:
### Features and Changes

Reverting a feature publishes a new revision immediately, but webhook consumers couldn't tell:

- The one-click revert (`postFeatureRevert`) dispatched `feature.revision.reverted` using the in-memory revision object, which still had `status: "draft"` — `createRevision` starts revisions as drafts and `markRevisionAsPublished` only updates the stored document, never the in-memory one. So the webhook payload described a just-published revert as a draft.
- No revert path emitted `feature.revision.published`, so consumers that track live-state changes via publish lifecycle events never saw reverts. The REST revert endpoint (`revertFeatureCore`) emitted no revision events at all.

Reverts usually happen when something is wrong, so they're exactly the changes downstream consumers most need to see promptly.

This PR:

- Re-reads the revision after `publishRevision` (the same pattern the publish endpoints already use, with the same in-memory fallback if the re-read returns nothing) so `revision.reverted` carries the real published status.
- Dispatches `revision.published` from all three revert paths: the app's one-click revert, `POST /api/v2/features/:id/revert`, and `POST /api/v2/features/:id/revisions/:version/revert` with `strategy: "publish"` (the draft strategy is unchanged — nothing is published there).

Note on behavior: a revert now emits both `revision.reverted` and `revision.published` for the same revision. Consumers subscribed to both event types will receive two events per revert; `revision.reverted` remains the one carrying `revertedToVersion`. A possible future refactor is to dispatch publish events from `publishRevision` itself so every publish-shaped flow is covered uniformly, but that's a larger behavioral change than this fix needs.

- Closes **(no linked issue)**

### Dependencies

None

### Testing

- Extended `test/api/revertFeature.test.ts` with three cases: events dispatched with the re-read published revision; fallback to the in-memory revision when the re-read returns null; no events when there is nothing to revert.
- Verified manually against a local server (Mongo + back-end): before the change, a revert emitted `revision.reverted` with `status: "draft"` and no publish event; after, all three revert paths emit `feature.updated` + `revision.reverted` (status `published`, `revertedToVersion` set) + `revision.published`, confirmed in the stored events.

### Screenshots

N/A (back-end only)

### DIFF
--- a/packages/back-end/src/api/features/postFeatureRevisionPublish.ts
+++ b/packages/back-end/src/api/features/postFeatureRevisionPublish.ts
@@ -21,7 +21,10 @@ import {
   getMergeResultPublishEnvs,
   toApiRevision,
 } from "back-end/src/services/features";
-import { dispatchFeatureRevisionEvent } from "back-end/src/services/featureRevisionEvents";
+import {
+  dispatchFeatureRevisionEvent,
+  getPublishedRevisionForEvents,
+} from "back-end/src/services/featureRevisionEvents";
 import { getEnvironments } from "back-end/src/util/organization.util";
 import {
   BadRequestError,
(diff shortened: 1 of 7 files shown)


anna-yn:
Pushed a follow-up commit hardening the failure mode of the post-publish re-read. The null case already fell back to the in-memory revision, but a *thrown* read error would have failed a request whose publish had already committed — and a retry of a revert would then confusingly hit "Nothing to revert". The re-read is now extracted into `getPublishedRevisionForEvents`, which falls back to the in-memory revision on a null result or a read error, and it's applied consistently at every publish-shaped dispatch site (both publish endpoints and all three revert paths — the publish endpoints previously had the same unguarded read). Direct unit tests cover the success / null / error cases.

bryce-fitzsimons (approved):

PR #6380, opened 2026-07-16
feat(visual-editor): duplicate a variant (copy css/js/domMutations)

Auz:
## What

Extends the internal add-variant endpoint (`POST /visual-editor/add-variant`, [postAddVariant.ts](packages/back-end/src/api/visual-editor-ai/postAddVariant.ts)) so the Visual Editor can **duplicate** an existing variant, not just create a blank one.

## Why

Users asked for a way to "duplicate this variant so I can save the one I'm working on while making edits to a new one." Doing the copy atomically on the server (rather than add-then-PUT from the extension) avoids a partial-state window and keeps it one round-trip.

## How

- New optional body field `sourceVariationId`.
- When present: the appended visual change copies the source variation's `css`, `js`, and `domMutations` (each mutation deep-copied so the two variations don't share object references), and the default name becomes `"<source> (copy)"`.
- An unknown `sourceVariationId` returns a 400 instead of silently creating a blank variant.
- Omitting the field preserves the existing blank-variant behavior exactly.

## Scope

Single file, additive. Back-end `type-check` passes. The matching "Duplicate this variant" affordance in the extension is a separate change that will ship after this lands.

### DIFF
--- a/packages/back-end/src/api/visual-editor-ai/postAddVariant.ts
+++ b/packages/back-end/src/api/visual-editor-ai/postAddVariant.ts
@@ -22,6 +22,10 @@ const bodySchema = z
   .object({
     visualChangesetId: z.string(),
     name: z.string().min(1).max(120).optional(),
+    // When set, the new variant is a duplicate: its css / js / domMutations
+    // are copied from this existing variation (matched on the internal
+    // `variation` id). Omit for a blank variant.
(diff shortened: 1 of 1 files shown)


gazzdingo (approved):
LGTM didnt test but code is all good

PR #6383, opened 2026-07-16
Mk/date picker bug

mknowlton89:
### Features and Changes

- Fixes manual date entry so partial values are preserved while typing instead of reverting after the debounce.
- Supports typing complete date ranges, including preserving the separator before the end date is entered.
- Aligns Product Analytics date-range input colors with adjacent Select controls in dark mode without changing DatePicker styling app-wide.

### Dependencies

None.

### Testing

- [x] Enter a start date slowly and confirm partial input is not reset.
- [x] Edit an existing date using backspace and confirm the partial value remains.
- [x] Type a complete range such as `2022-01-01 - 2022-01-31`.
- [x] Confirm calendar-based date selection still works.
- [x] Verify Product Analytics date inputs match adjacent Select backgrounds in light and dark modes.
- [x] Verify the date picker elsewhere in the app still works without regressions (tested the scheduled rule picker and the insights timeline picker)

### Screenshots

https://www.loom.com/share/20bfea0628804669b8a1fc9630a995a1

### DIFF
--- a/packages/front-end/components/DatePicker/index.tsx
+++ b/packages/front-end/components/DatePicker/index.tsx
@@ -42,6 +42,7 @@ type Props = {
   wrapRangeInputs?: boolean;
   compact?: boolean;
   disabled?: boolean;
+  inputStyle?: React.CSSProperties;
   fixedSpanMode?: {
     phase: "committed" | "choosing";
     anchorDate?: Date;
@@ -116,6 +117,7 @@ export default function DatePicker({
   wrapRangeInputs = false,
   compact = false,
   disabled,
+  inputStyle,
   fixedSpanMode,
 }: Props) {
   const inputHeight = compact ? 32 : 38;
@@ -404,6 +406,7 @@ export default function DatePicker({
                       minHeight: inputHeight,
                       cursor: "pointer",
                       ...compactFieldStyle,
+                      ...inputStyle,
                     }}
                     className={clsx("date-picker-field", {
                       "text-muted": isRange ? !date || !date2 : !date,
--- a/packages/front-end/components/DatePicker/index.tsx
+++ b/packages/front-end/components/DatePicker/index.tsx
@@ -159,6 +159,11 @@ export default function DatePicker({
   const fieldClickedTime = useRef(new Date());
 
   useEffect(() => {
+    // While editing a range, the buffered values are the source of truth.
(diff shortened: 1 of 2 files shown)


packages/front-end/components/DatePicker/index.tsx
```diff
@@ -156,20 +162,22 @@ export default function DatePicker({
...
-    } else {
-      setBufferedDate("");
-    }
-    if (date2) {
-      setBufferedDate2(format(parseDateInput(date2), dateFormat));
-    } else {
-      setBufferedDate2("");
-    }
-  }, [date, date2, dateFormat, parseDateInput]);
+  const dateRef = useRef(date);
```
nhat-growthbook, line 167: Holy refs

packages/front-end/components/DatePicker/index.tsx
```diff
@@ -275,51 +285,92 @@ export default function DatePicker({
...
         setBufferedDate("");
       }
 
       if (endTrim) {
         const parsedDate2 = parseDateInput(endTrim);
-        const finalDate2 = clampParsedDate(parsedDate2);
-        setDate2?.(finalDate2);
-        setBufferedDate2(format(finalDate2, dateFormat));
-        anchor = finalDate2;
+        if (format(parsedDate2, dateFormat) === endTrim) {
```
nhat-growthbook, line 331: Instead of string comparison, can we use `isEqual` in date-fns? Would that be more accurate?

packages/front-end/components/DatePicker/index.tsx
mknowlton89: Yea - this is the big code smell that makes me want to just go to two independent date fields for `startDate` and `endDate` lol

packages/front-end/components/DatePicker/index.tsx
mknowlton89: I don’t think isEqual would be more accurate here. This check verifies that the input is valid and round-trips to the picker’s required format (yyyy-MM-dd, or the datetime equivalent). If we parsed both "07/02/2026" and "2026-07-02" and used isEqual, they could represent the same date, but we would lose the format validation.

mknowlton89:
Closing this - going to move to the two date setup instead.

PR #6389, opened 2026-07-16
feat(visual-editor): rename + delete variant endpoints

Auz:
## What

Two new internal Visual Editor endpoints that back the variant actions menu in the extension:

- **`POST /visual-editor/rename-variant`** — updates a variation's display name on the experiment. The variation key/id is left untouched so SDK bucketing and analytics stay stable; a no-op write is skipped to avoid a spurious `dateUpdated` bump.
- **`POST /visual-editor/delete-variant`** — removes a variation from the experiment (variations, every phase's variations list, equal-redistributed weights) plus its matching visual change. Rejects deleting control (index 0) or the last remaining variant. Returns the refreshed changeset + experiment so the side panel re-renders in one round-trip (mirrors `add-variant`).

Both registered in `visualEditorAi.router.ts`.

## Not included (no backend change needed)

- **Duplicate** — reuses the existing `add-variant` + `sourceVariationId` (shipped in #6380).
- **Clear all changes** — empties the variation's visual change via the existing `PUT /visual-changesets/:id`.

## Notes

- Internal extension endpoints (`excludeFromSpec: true`), same auth/permission pattern as `add-variant`: `canUpdateExperiment` + `canUpdateVisualChange`, and `validateExperimentChange` before writing.
- Back-end `type-check` passes.

## Deploy dependency

The extension's rename/delete actions call these endpoints, so they must be deployed before that extension build is relied on in cloud (duplicate/clear work without it).

### DIFF
--- a/packages/back-end/src/api/visual-editor-ai/postDeleteVariant.ts
+++ b/packages/back-end/src/api/visual-editor-ai/postDeleteVariant.ts
@@ -1,0 +1,163 @@
+import { z } from "zod";
+import type { Changeset } from "shared/types/experiment";
+import type { ExperimentInterfaceExcludingHoldouts } from "shared/validators";
+import {
+  findVisualChangesetById,
+  toVisualChangesetApiInterface,
+  updateVisualChangeset,
+} from "back-end/src/models/VisualChangesetModel";
+import {
+  getExperimentById,
+  updateExperiment,
+} from "back-end/src/models/ExperimentModel";
+import { validateExperimentChange } from "back-end/src/services/experimentChanges/changeExperimentStatus";
+import { toExperimentApiInterface } from "back-end/src/services/experiments";
+import { resolveOwnerEmail } from "back-end/src/services/owner";
+import { createApiRequestHandler } from "back-end/src/util/handler";
+import { requireUserAuth } from "./requireUserAuth";
+
+const bodySchema = z
+  .object({
+    visualChangesetId: z.string(),
+    // Internal variation `id` — matches experiment.variations[].id and
+    // visualChange.variation.
+    variationId: z.string(),
+  })
+  .strict();
+
+const validation = {
+  bodySchema,
+  querySchema: z.never(),
+  paramsSchema: z.never(),
+  responseSchema: z.any(),
+  method: "post" as const,
+  path: "/visual-editor/delete-variant",
+  operationId: "postVisualEditorDeleteVariant",
+  // Internal Visual Editor extension endpoint — not part of the
+  // public OpenAPI spec.
+  excludeFromSpec: true,
+};
+
+// Removes a variation from the experiment (variations, every phase's
+// variations list, and weights) plus its matching visual change. The inverse
+// of postAddVariant. Returns the refreshed changeset + experiment so the side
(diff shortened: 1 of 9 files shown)


Auz:
@greptileai review again

Auz:
@greptileai review again

Auz:
@greptileai review

packages/back-end/src/api/visual-editor-ai/postDeleteVariant.ts
```diff
@@ -0,0 +1,219 @@
...
+    if (next.variations !== undefined) {
+      next.variations = next.variations.filter((pv) => pv.id !== variationId);
+    }
+    if (next.variationWeights !== undefined && next.variationWeights.length) {
+      next.variationWeights =
+        next.variationWeights.length === originalCount
+          ? renormalizeWeights(
+              next.variationWeights.filter((_, i) => i !== idx),
+            )
+          : renormalizeWeights(new Array(nextVariations.length).fill(1));
```
Kevin-Chant, line 122: This editing should only apply to the latest phase instead of rewriting history, right? That's the behavior across our other endpoints

packages/back-end/src/api/visual-editor-ai/postDeleteVariant.ts
```diff
@@ -0,0 +1,219 @@
...
+  const experiment = await getExperimentById(context, changeset.experiment);
+  if (!experiment) return context.throwNotFoundError("Experiment not found");
+
+  // Mutates the experiment AND the changeset — both gates required.
+  if (!context.permissions.canUpdateExperiment(experiment, {})) {
+    context.permissions.throwPermissionError();
+  }
+  if (!context.permissions.canUpdateVisualChange(experiment)) {
+    context.permissions.throwPermissionError();
+  }
```
Kevin-Chant, line 70: Should there also be a guard about the experiment status not being running, or is it intentional that this allows for clobbering an ongoing experiment?

packages/back-end/src/api/visual-editor-ai/postRenameVariant.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const nextVariations = experiment.variations.map((v, i) =>
+    i === idx ? { ...v, name: trimmed } : v,
+  );
+  const changes: Changeset = { variations: nextVariations };
+  await validateExperimentChange({ context, experiment, changes });
+  await updateExperiment({ context, experiment, changes });
+
+  return { name: trimmed };
```
Kevin-Chant: Should this be a richer return shape or is this sufficient?

Kevin-Chant (changes requested):
A few questions, the only one that seems blocking is not deleting the variation from old phases.
Also, it's intentional that this isn't included in our rest api documentation right? I assume so but wanted to check

Auz:
Yes, all the internal Visual Editor endpoints use excludeFromSpec: true (add-variant, create-changeset, etc.) as I figured they only really make sense from the extension. We could discuss this you think this is incorrect.

Kevin-Chant (approved):
LGTM

PR #6391, opened 2026-07-17
Visual Editor: server-side experiment name search in bootstrap

Auz:
## What

Adds an optional `?search=` query to the internal `getBootstrap` endpoint so the Visual Editor's experiment picker can find experiments by name — including ones outside the recent-changeset window.

## Why

The picker's list is capped: it ranks the newest **200** changesets down to **30**. In an org with many visual experiments, a target whose changeset is older than the newest 200 can't appear at all. Server-side name search sidesteps both caps.

## How

- `getBootstrap` gains `querySchema: { search?: string (max 100) }`.
- **Search mode:** `findExperimentsByName` (new, in `ExperimentModel`) — case-insensitive, **regex-escaped**, org-scoped, permission-filtered, newest-first, capped at 100 — then `findVisualChangesetsByExperimentIds` (new, in `VisualChangesetModel`) fetches those experiments' changesets.
- **Default mode** (no search) is unchanged: newest-200 changesets → rank → 30.
- Either way results are ranked **draft-first + recency** and trimmed to `MAX_RECENT`.

## Notes

- Regex metacharacters in the search are escaped so user text can't inject a pattern.
- Internal extension endpoint (`excludeFromSpec: true`); back-end `type-check` passes.
- **Stacked on `gm/ve-variant-actions`** (base branch) so it carries the draft-first bootstrap sort. Retarget to `main` once that merges.
- Extension side (a search box in the switcher) ships separately.


packages/back-end/src/api/visual-editor-ai/getBootstrap.ts
```diff
@@ -47,15 +66,32 @@ export const getBootstrap = createApiRequestHandler(validation)(async (req) => {
...
+  // reachable. Default mode keeps the fast "newest changesets" path.
+  const search = (req.query.search ?? "").trim();
+  let changesets: VisualChangesetInterface[];
+  let experiments: ExperimentInterface[];
+  if (search) {
+    experiments = await findExperimentsByName(
+      context,
+      search,
+      SEARCH_EXPERIMENT_CAP,
+    );
```
jdorn, line 76: This doesn't filter on `hasVisualChangesets: true`, so it will include non-visual editor experiments as well. Is that intentional?

Auz: Addressed with FindVisualExperimentsByName()

Auz:
@greptileai review

jdorn (approved):
I would clean up the overly verbose AI comments, but code itself looks fine.

PR #6414, opened 2026-07-20
feat(datasources): add OAuth M2M authentication for databricks

ahdriel:
### Features and Changes

The recommended way to authenticate with Databricks is OAuth. We only supported Personal Authentication Tokens (PAT); so this expands the Databricks auth mechanism and adds OAuth support.

### Testing

- Unit tests for `buildDatabricksConnectionOptions`
- Manual: create or edit a Databricks data source with OAuth credentials and run a test query.


packages/front-end/components/Settings/DatabricksForm.tsx
```diff
@@ -1,14 +1,21 @@
...
 
 const DatabricksForm: FC<{
   params: Partial<DatabricksConnectionParams>;
   existing: boolean;
-  onParamChange: ChangeEventHandler<HTMLInputElement | HTMLSelectElement>;
+  onParamChange: ChangeEventHandler<HTMLInputElement>;
   setParams: (params: { [key: string]: string | boolean }) => void;
 }> = ({ params, existing, onParamChange, setParams }) => {
+  const [originalAuthType] = useState(params.authType);
+  const authType = params.authType ?? "oauth-m2m";
```
lukesonnet: Shouldn't this fall back to pat?

packages/back-end/src/integrations/Databricks.ts
```diff
@@ -12,8 +12,11 @@ export default class Databricks extends SqlIntegration {
   requiresDatabase = true;
   requiresSchema = false;
   setParams(encryptedParams: string) {
-    this.params =
+    const params =
       decryptDataSourceParams<DatabricksConnectionParams>(encryptedParams);
+    // Legacy rows predate authType and are always PAT
```
lukesonnet, line 17: legacy connections?

ahdriel: as in I should update the text to be `legacy connections`?

I would like to avoid it because the connection is not necessarily legacy as we still support it, but the row that is saved is legacy as it does not conform with our latest declared type where `authType` exists.

Or if you are asking if this exists for the already existing connections, then yes, this just ensures `authType` is always set in our code, by automatically converting from undefined to `pat` at read time.

packages/front-end/components/Settings/DatabricksForm.tsx
ahdriel: Oooh this took me down a rabbit hole-ish.

Effectively, here, `authType` is always defined (either read-time migration, or defaultValues from the Form). But for the form we have to use `Partial` to follow the convention we have (bigger work to change this), so we still need to handle it here, even though it is effectively dead code as it won't happen.

I will update to `pat` to be on the conservative side, but the original intent was that New datasources would get the `oauth` instead of `pat` by default, which is accomplished by the `defaultValues` for the form.

So... this code won't really affect anything, will use `pat` to be more conservative.

lukesonnet: I tried to figure this out myself and couldn't quickly because I saw it was required in the form. I think being safer and using a different part of the code to set the actual default to oauth makes sense, but is minor.

lukesonnet (approved):

packages/back-end/src/integrations/Databricks.ts
lukesonnet: I see, this is just used to migrate legacy rows and not legacy datasource objects?

ahdriel: It is to migrate the datasource objects from the db at read time (my bad on using rows, not sql here).

PR #6416, opened 2026-07-20
Fix ramp schedules re-disabling their rule from the startActions snapshot

anna-yn:
### Features and Changes

Ramp schedules created via the REST API (`POST /api/v1/ramp-schedules`) or a deferred revision publish get a server-attached `startActions` snapshot of the rule's pre-ramp state — including `enabled` (`getStartPatchForRule`). Because `computeEffectivePatch` seeded every forward publish from that snapshot, a schedule created over a **disabled** rule re-applied `enabled: false` on later publishes.

Most visible failure: a zero-step "enable on date" schedule enables its rule (`applyRampStartActions` injects `enabled: true`) and then, in the same scheduler tick, `advanceUntilBlocked` auto-completes it via `completeRollout` — which republished the snapshot's `enabled: false`. The rule flips ON→OFF within milliseconds, the schedule reports "completed", and no error surfaces anywhere. We hit this on a self-hosted deployment: a launch flag's history showed "Schedule started" and "Schedule ended" revisions ~250ms apart with the rule left disabled.

The same seeding also meant a multi-step ramp materialized over a disabled rule would re-disable its rule on every advance past step 0 (the `enabled: true` injection only fires on the -1 crossing).

**Fix:** drop `enabled` from the `startActions` seed layer in `computeEffectivePatch`:

- The engine owns the rule's enabled state on forward publishes (injected on the -1 crossing, forced off by cutoff/disable paths).
- Steps and `endActions` can still set `enabled` explicitly — only the seed layer is stripped.
- Full rollback (-1) applies the raw `startActions` (it doesn't go through this function), so a captured enabled state is still honored for restore-to-pre-ramp semantics.
- This matches UI-built snapshots (`buildRampStartActionsFromRule`), which already omit `enabled` for exactly this reason.

Two deliberate semantics worth calling out for review:

1. A rule manually disabled mid-ramp now **stays** disabled through later advances/completion (previously, an API-created snapshot with `enabled: true` would silently re-enable it). We think respecting the manual kill is the safer behavior, and it's what UI-created schedules already did.
2. An explicit `startState: { enabled: ... }` passed to the REST API now only takes effect on full rollback, consistent with the OpenAPI description of `startActions` ("Actions that restore controlled rules to their pre-ramp state").

- Closes **N/A** (no issue filed; happy to open one if preferred)

### Dependencies

None.

### Testing

Unit tests added in `packages/back-end/test/services/rampSchedule.test.ts`:
- `drops enabled from the startActions seed`
- `steps and endActions can still set enabled explicitly`

End-to-end repro against a locally-run back-end (before/after):
1. Create a feature with a disabled force rule.
2. `POST /api/v1/ramp-schedules` with that rule, no steps, no `startActions`, `startDate` in the near future — the server attaches a snapshot with `"enabled": false`.
3. Wait for the schedule to fire.
   - **Before:** revision history shows "Schedule started" then "Schedule ended" ~35ms apart; the rule ends disabled; the schedule reports completed.
   - **After:** same schedule shape; the rule ends enabled and stays enabled; schedule completes normally.

`pnpm --filter back-end test test/services/rampSchedule.test.ts` (195 passing) and `pnpm --filter back-end type-check` are green.

### Screenshots

N/A (back-end only).

### DIFF
--- a/packages/back-end/src/services/rampSchedule.ts
+++ b/packages/back-end/src/services/rampSchedule.ts
@@ -423,7 +423,15 @@ export function computeEffectivePatch(
 
   // Seed with startActions — the base-layer rule state (condition, savedGroups,
   // coverage, etc.) that all steps inherit from unless explicitly overridden.
-  for (const a of schedule.startActions ?? []) merge(a);
+  // `enabled` is deliberately dropped from the seed: the engine owns the rule's
+  // enabled state while a ramp is live, and a snapshot captured from a disabled
+  // rule would otherwise re-disable the rule on every forward publish. Full
+  // rollback (-1) applies the raw startActions instead of this function, so a
+  // captured enabled state is still honored there.
+  for (const a of schedule.startActions ?? []) {
+    const { enabled: _enabled, ...patch } = a.patch;
+    merge({ ...a, patch });
+  }
 
   const lastStepIdx = Math.min(stepIndex, schedule.steps.length - 1);
   for (let i = 0; i <= lastStepIdx; i++) {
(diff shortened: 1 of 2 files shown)


bryce-fitzsimons (approved):
thanks for the fix, looks correct

PR #6418, opened 2026-07-21
Added filtering to get experiments endpoints

Auz:
Also updated skills so that agents can use the new endpoint options.

### DIFF
--- a/packages/back-end/src/models/ExperimentModel.ts
+++ b/packages/back-end/src/models/ExperimentModel.ts
@@ -386,6 +386,9 @@ const experimentSchema = new mongoose.Schema({
 experimentSchema.index({ organization: 1, datasource: 1 });
 experimentSchema.index({ organization: 1, project: 1 });
 experimentSchema.index({ organization: 1, trackingKey: 1 });
+experimentSchema.index({ organization: 1, dateCreated: 1 });
+experimentSchema.index({ organization: 1, dateUpdated: 1 });
+experimentSchema.index({ organization: 1, name: 1 });
 experimentSchema.index(
   { "nextScheduledStatusUpdate.date": 1 },
   { sparse: true },
@@ -483,6 +486,7 @@ export async function getAllExperiments(
   {
     project,
     includeArchived = false,
+    archived,
     type,
     datasourceId,
     trackingKey,
@@ -492,6 +496,9 @@ export async function getAllExperiments(
   }: {
     project?: string;
     includeArchived?: boolean;
+    // Tri-state archived filter: true = archived only, false = exclude
+    // archived, undefined = fall back to `includeArchived`.
+    archived?: boolean;
     type?: ExperimentType;
     datasourceId?: string;
     trackingKey?: string;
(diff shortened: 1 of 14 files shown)


Auz:
@greptileai review

packages/shared/src/validators/experiments.ts
```diff
@@ -1534,6 +1554,49 @@ export const listExperimentsValidator = {
...
+        .optional(),
+      owner: ownerInputField
+        .describe("Filter by comma-separated owner ids, names, or emails")
+        .optional(),
+      result: csvQueryField(
+        experimentResultsType,
+        "Filter by comma-separated results (won, lost, inconclusive, dnf). Only stopped experiments carry a result",
+      ),
+      tag: z.string().describe("Filter by comma-separated tags").optional(),
+      type: csvQueryField(
```
lukesonnet: Can we rename this `implementationType` to mach the UI and also update `StructuredExperimentFilters`, `normalizeExperimentFilters`, `experimentHasType` and so on to match, since there *is* already an internal top-level `experiment.type` field?

packages/shared/src/validators/experiments.ts
```diff
@@ -1534,6 +1554,49 @@ export const listExperimentsValidator = {
...
+        .describe(
+          "Raw experiment search/filter string (same syntax as the app's experiment list filters, e.g. `status:running tag:checkout`). Negation (`!`) and operators (`~`, `^`, `>`, `<`, `=`) are not supported and return a 400",
+        )
+        .optional(),
+      owner: ownerInputField
+        .describe("Filter by comma-separated owner ids, names, or emails")
+        .optional(),
+      result: csvQueryField(
+        experimentResultsType,
+        "Filter by comma-separated results (won, lost, inconclusive, dnf). Only stopped experiments carry a result",
```
lukesonnet: It looks like we don't actually validate that it is only stopped experiments. If you switch a status back to running manually a result can still exist on the model. So we either need to add that check or clean up this language here and in `filterExperiments` to make sure it's totally clear.

Auz: This is feature compatible with the front end. I'm not sure we should have different experiments returned by the same query.

Auz:
@greptileai review again

packages/back-end/src/models/ExperimentModel.ts
```diff
@@ -386,6 +386,9 @@ const experimentSchema = new mongoose.Schema({
 experimentSchema.index({ organization: 1, datasource: 1 });
 experimentSchema.index({ organization: 1, project: 1 });
 experimentSchema.index({ organization: 1, trackingKey: 1 });
+experimentSchema.index({ organization: 1, dateCreated: 1 });
+experimentSchema.index({ organization: 1, dateUpdated: 1 });
+experimentSchema.index({ organization: 1, name: 1 });
```
jdorn: Same comment as the features PR.  Probably not worth the index overhead to add these.

jdorn (approved):

packages/back-end/src/models/ExperimentModel.ts
Auz: pagination is in-memory, so the indexes were optimizing the cheapest stage. Dropped them; sorting now happens in Node after the permission filter

packages/shared/src/validators/experiments.ts
lukesonnet: Updating the comment is fine.

lukesonnet (approved):

PR #6419, opened 2026-07-21
Added filter and sort endpoints to features

Auz:
And updated skills to tell agents how to use it


Auz:
@greptileai review

packages/shared/src/validators/shared.ts
```diff
@@ -79,6 +79,48 @@ export const paginationQueryFields = {
...
+      // dropped from the generated OpenAPI spec
+      .describe(description)
+      .meta({
+        type: "array",
+        items: { type: "string", enum: [...allowed] },
+        explode: false,
+      })
+      .optional()
+  );
+};
```
jdorn, line 122: I don't love the `refine` approach here. It feels a little too magical.  Probably fine though.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -178,6 +179,11 @@ const featureSchema = new mongoose.Schema({
 
 featureSchema.index({ id: 1, organization: 1 }, { unique: true });
 featureSchema.index({ organization: 1, project: 1 });
+// Compound indexes for API list sorting (org equality + sort field) — keep in
+// sync with sortableFeatureFields in shared/validators/features-v2.ts
+featureSchema.index({ organization: 1, id: 1 });
+featureSchema.index({ organization: 1, dateCreated: 1 });
+featureSchema.index({ organization: 1, dateUpdated: 1 });
```
jdorn: I don't know how often these indexes will actually be used.  Any filtering whatsoever will make the query skip this index. Might not be worth the index overhead.

jdorn (approved):

packages/back-end/src/models/FeatureModel.ts
Auz: removed

packages/back-end/generated/spec.yaml
```diff
@@ -13062,9 +13068,19 @@ paths:
               - running
               - stopped
         - $ref: '#/components/parameters/q'
-        - $ref: '#/components/parameters/owner'
+        - name: owner
+          in: query
+          description: Filter by comma-separated owner ids, names, or emails
```
ahdriel, line 13073: question: are we making `names` unique?

I think we should allow filtering only for fields that are unique, which we should enforce for `id` and `email`, but unsure about name.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -543,6 +572,16 @@ export async function getAllFeaturesWithoutEditorFields(
...
 }
 
+// Optional field filters shared by the list/count/page query builders. Values
+// within a filter are ORed ($in); separate filters AND together.
+export type FeatureListFilters = {
+  tags?: string[];
+  owners?: string[];
+  valueTypes?: FeatureValueType[];
+  // Exact config key — matches Config-mode flags backed by that config
+  baseConfig?: string;
```
ahdriel, line 582: question: is this and `id` or `key`? can we add a qualifier to it?

packages/shared/src/validators/features-v2.ts
```diff
@@ -738,6 +740,16 @@ export const updateFeatureBodyV2 = z
 
 // ---- Route validators ----
 
+// None of these are index-backed — Mongo does a bounded blocking sort per
+// page (see the index note in FeatureModel). Keep the list short: each entry
+// is a sort the database has to do the hard way.
+export const sortableFeatureFields = [
```
ahdriel, line 746: I am concerned about the performance implications of this. should we add indexes? or what's the best way to handle generic sorting for the features collection?

Because now we are opening the door for unoptimized mongodb queries that we don't control.

ahdriel (approved):

packages/shared/src/validators/features-v2.ts
Auz: We had indexes in this PR, but Jeremy thought they are unnecessary.

PR #6423, opened 2026-07-21
feat(fact-tables)!: handle columns consistently across the API

ahdriel:
### Features and Changes

> [!WARNING]
> This is a potentially breaking change!
> While we published our OpenAPI for `bulk-import/facts` with `additionalProperties: false`, this was **NOT** being enforced. This PR enforces it, so unknown keys that previously were silently ignored now will throw an error.
>
> `additionalProperties: false` was already properly enforced on `POST` & `PUT /fact-tables`

The fact table API handled the `columns` property inconsistently.

`POST /fact-tables` ignored columns entirely (but didn't error out), `PUT /fact-tables/{id}` could only patch existing columns (it errored on any column name it didn't recognize and required a `datatype`), and each path normalized stored columns differently.

This makes the create, update, and bulk-import paths treat `columns` the same way:

- Create and bulk import now accept a `columns` array and store it.
- Update upserts by `column`: existing columns are modified, new columns are created, and columns not included are left unchanged.
- `datatype` is now optional. Omit it (or send `""`) to have the type auto-detected from the SQL (before this was required for creation, but a later update would allow you to override the detected `datatype`), and a column refresh is queued whenever any column still needs detection.
- We now normalize incompatible settings. For example `alwaysInlineFilter` requires `dataType: 'string'`. Before if you updated `datatype` to something else, we would keep `alwaysInlineFilter` as it was before. Now we remove it to ensure the column always has valid settings.

### Testing

- New unit tests
- Robot tests via curl
- Manual tests via curl


packages/shared/src/validators/bulk-import.ts
```diff
@@ -8,39 +9,47 @@ const postBulkImportFactsBody = z
...
+                )
+                .optional(),
+              managedBy: z
+                .enum(["", "api", "admin"])
+                .describe(
+                  'Set this to "api" to disable editing in the GrowthBook UI',
+                )
+                .optional(),
+            })
+            .strict(),
```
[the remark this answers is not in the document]
ahdriel: This is intentional as this reflects what was already in the OpenAPI docs. We want to fail and let the user know instead of silently accepting but ignoring the fields and misleading the user.

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -101,13 +114,27 @@ export const postBulkImportFacts = createApiRequestHandler(
...
+        if (
+          needsColumnRefresh(existing, data) ||
+          columnsNeedDetection(existing.columns)
+        ) {
           await queueFactTableColumnsRefresh(existing);
         }
         factTableMap.set(existing.id, {
           ...existing,
           ...data,
+          columns: existing.columns,
```
lukesonnet, line 137: Is the upsertion our usual pattern? Makes it a little hard to read, shouldn't we use a method that returns the updated fact table instead of `upsertColumns` that modifies in place?

packages/back-end/src/util/factTable.ts
```diff
@@ -3,9 +3,94 @@ import { utcToZonedTime, zonedTimeToUtc } from "date-fns-tz";
...
+  CreateColumnProps,
+  JSONColumnFields,
 } from "shared/types/fact-table";
 import { DataSourceInterface } from "shared/types/datasource";
 
+/**
+ * Clears datatype-incompatible props. Empty datatype skips checks while
+ * auto-detection is pending.
+ */
+export function stripIncompatibleFields(
```
lukesonnet, line 15: I don't love that this all happens silently, but I think it's more backwards compatible, so probably fine.

packages/shared/src/validators/bulk-import.ts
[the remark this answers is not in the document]
teresayung: got it, i think this is the only breaking change i see, otherwise everything else looks fine!

teresayung (approved):

packages/back-end/src/util/factTable.ts
ahdriel: The current experience is that the old configuration is left in place so it is broken (like `datatype: boolean` with `alwaysInlineFilter: true` which is not possible).

Agreed that is not ideal, but prioritizing backwards compatibility for now.

We can talk more tomorrow on our standup; one thing to note is that you can pass `datatype: ""` (so it is auto-detected) and other configs like slices and etc. So the failure is async. I guess we can always fully adopt an async pattern and just return `Accepted` for the requests, return and id they can query the status of, and then fail the whole fact table creation if any of the configs are invalid.

On the sync side of things, when sending an `update` with `datatype: "boolean"` for a string with a property that requires string. Right now we accept and automatically fix it by removing the `alwaysInlineFilter`, but we could return an error saying `Column ${name} cannot be updated to datatype ${newDatatype} because of the current `alwaysInlineFilter` setting. Retry the request with `alwaysInlineFilter: false` to use the new datatype`.

Just some rambling -- as you pointed out, I am being more consistent in the backend than what we have today, but the changes are done silently instead of leaving the model in an inconsistent state.

lukesonnet: I think the sync failures are more consistent with our broader patterns.

I'm not sure I understand exactly what happens in the async case.

PR #6424, opened 2026-07-21
Add ESLint guardrails against raw <a> tags and next/link imports

oelshaikh:
## Summary
Adds ESLint guardrails so new raw `<a>` tags and `next/link` imports can't land in the front-end.

- `react/forbid-elements` blocks raw `<a>` → use `@/ui/Link` / `@/ui/LinkButton` / `@/ui/Button`
- `no-restricted-imports` blocks `next/link` → use `@/ui/Link` / `@/ui/LinkButton`
- Both are **errors**. The existing backlog is baselined in `eslint-suppressions.json` so CI stays green today, while any *new* violation fails.
- Each migration auto-prunes its suppression via the existing `--prune-suppressions` pre-commit step, so the count only ratchets down.

## Notes
- Baseline was generated cacheless (`rm -f .eslintcache` first) to capture the full backlog — don't regenerate it with `--cache`.
- Genuine exceptions (e.g. `download` anchors, Markdown renderer) can use an `eslint-disable-next-line` with a reason when those files are migrated.

## Test plan
- [x] New raw `<a>` / `next/link` in a front-end file → ESLint errors
- [x] `pnpm lint:ci` passes on the current tree (backlog suppressed)

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -1,4 +1,9 @@
 {
+  "packages/front-end/components/ActivityList.tsx": {
+    "no-restricted-imports": {
+      "count": 1
+    }
+  },
   "packages/front-end/components/Admin/EditOrganization.tsx": {
     "no-restricted-imports": {
       "count": 1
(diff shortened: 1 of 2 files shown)


ahdriel (approved):
Nice! Thanks for adding it.

PR #6425, opened 2026-07-21
Add "while you're here" link-migration pre-commit nudge

oelshaikh:
## Summary
Adds a non-blocking "while you're here" pre-commit nudge for the link migration.
- For staged front-end files that still contain raw `<a>` or `next/link`, it prints a friendly reminder to migrate to `@/ui/Link` / `@/ui/LinkButton`.
- Only looks at files you're already touching; **never fails the commit**.
- Independent of the ESLint rules — the hard gate against new violations lives in the guardrails PR.
## Test plan
- [x] Stage a file with a baselined `<a>` → nudge prints, commit still succeeds
- [x] `node scripts/link-nudge.js` exits 0 even with offenders present
### Screenshots

image

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -26,6 +26,10 @@ fi
 echo "#### Linting typescript, scss, and markdown files"
 pnpm lint-staged
 
+# Non-blocking "while you're here" reminder for staged files that still use
+# raw <a> tags or next/link. Never fails the commit.
+node scripts/link-nudge.js || true
+
 # Add eslint-suppressions.json if it was modified during linting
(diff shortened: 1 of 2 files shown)


.husky/pre-commit
```diff
@@ -26,6 +26,10 @@ fi
 echo "#### Linting typescript, scss, and markdown files"
 pnpm lint-staged
 
+# Non-blocking "while you're here" reminder for staged files that still use
+# raw <a> tags or next/link. Never fails the commit.
+node scripts/link-nudge.js || true
```
ahdriel, line 31: question: should we make this more generic? like for all `eslint` violations that are in the `eslint-suppressions` for the files they are modifying?

PR #6426, opened 2026-07-21
design(VariationLabel): Replace remaining usages of legacy variation labels

msamper:
### Features and Changes

Replaces hand-rolled variation labels (using `variation … with-variation-label` classname) across the front-end with the shared design-system components `VariationLabel` and `VariationNumber` and deletes the `with-variation-label` classname to prevent future usages.

`VariationLabel` inside a `flipTheme` tooltip rendered unreadable text because the custom `--color-text-*` tokens are keyed to the app-level `.light-theme`/`.dark-theme` `<html>` class and don't follow a nested `<Theme appearance>` flip. Added scoped rules in `radix-config.css` that re-map `--color-text-*` to the local Radix appearance class (`.radix-themes.radix-themes.light` / `.dark`), so flipped tooltips (baseline/variation chooser column labels) now show correct text. 

### Dependencies

N/A

### Testing

- `pnpm --filter front-end type-check` passes; ESLint clean on all touched files (new rule verified to fire on the old pattern).
- Manual UI checks (light + dark mode)
- Baseline/variation chooser column-label tooltips render readable variation names (the theme-flip bug).

### Screenshots

Screenshot 2026-07-29 at 4 51 16 PM
Screenshot 2026-07-29 at 4 51 32 PM
Screenshot 2026-07-29 at 4 51 44 PM


packages/front-end/components/Experiment/VariationChooser.tsx
```diff
@@ -1,7 +1,9 @@
 import clsx from "clsx";
```
msamper: This entire component is unused. Wondering if we can just get rid of it

packages/front-end/enterprise/components/Insights/MetricCorrelations/MetricCorrelationsExperimentTable.tsx
```diff
@@ -290,27 +292,25 @@ const ExperimentWithMetricsTable: FC<Props> = ({
...
+            <Box style={{ maxWidth: 200 }} minWidth="0">
+              <VariationLabel
+                number={e.variationIndex}
+                name={e.variationName}
+                size="medium"
+              />
+            </Box>
+            {e.shipped ? (
+              <Tooltip body={"Variation marked as the winner"}>
+                <FaShippingFast />
```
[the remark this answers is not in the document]
msamper: keeping this for now. out of scope for this PR

packages/front-end/components/Experiment/FieldAlignedVariationNumber.tsx
```diff
@@ -0,0 +1,29 @@
+import { Box, Flex } from "@radix-ui/themes";
+import VariationNumber from "@/ui/VariationNumber";
+
+// Height of a standard `Field` text input (Bootstrap `.form-control`). The badge
+// is vertically centered within this height so it lines up with the input rather
+// than the label rendered above it.
+const FIELD_INPUT_HEIGHT = "35px";
```
ahdriel: issue: It is technically 36px I think

packages/front-end/components/Experiment/UrlRedirectModal.tsx
```diff
@@ -118,7 +121,6 @@ const UrlRedirectModal: FC<{
       );
       mutate();
     }
-    close();
```
ahdriel, line 121: question: when does `close` get called now?

If it autoCloses -- what happens if this`onSubmit` throws/fails?

The concern I have is that an error is now closing the modal even when it fails. Maybe we need to update the code to properly throw if `ModalStandard` expects either `void` or `throw` to decide if it closes or not

packages/front-end/components/Experiment/VariationChooser.tsx
```diff
@@ -44,23 +45,15 @@ export default function VariationChooser({
     );
   }
   if (filteredVariations.length <= 1) {
+    const lastVariation = filteredVariations[filteredVariations.length - 1];
```
ahdriel: nit: it is correct, but also it is the only variation no?

should we check if `== 1`? how does this get rendered if `filteredVariations` is an empty array?

packages/front-end/components/Holdout/LinkedExperimentsTable.tsx
```diff
@@ -103,25 +104,12 @@ const LinkedExperimentsTable = ({ holdout, experiments }: Props) => {
...
-                        }}
-                      >
-                        {variation?.name}
-                      </span>
-                    </div>
+                    <VariationLabel
+                      number={variationIndex}
+                      name={variation.name}
+                      size="medium"
+                      maxWidth={150}
```
ahdriel: nit: this should be `170` as well right? (20 from number + 150 from name)

packages/front-end/components/HomePage/ExperimentImpact/ExperimentImpactTab.tsx
```diff
@@ -45,22 +47,14 @@ export default function ExperimentImpactTab({
...
-            >
-              {v.name}
-            </span>
-          </div>,
+          <Box key={`var-experiment${ei}-variation${i}`} className="my-1">
+            <VariationLabel
+              number={i}
+              name={v.name}
+              size="medium"
+              maxWidth={200}
```
ahdriel: In some of the places we migrated `maxWidth` of just the name, while in others we added `+20` for the number.

Should we consolidate in one approach? or is this better handled as-is on a per-file basis?

packages/front-end/enterprise/components/Dashboards/DashboardEditor/DashboardEditorSidebar/EditSingleBlock.tsx
```diff
@@ -1425,25 +1414,13 @@ export default function EditSingleBlock({
...
-                          className="d-inline-block"
-                          style={{
-                            width: 150,
-                            lineHeight: "14px",
-                          }}
-                        >
-                          {label}
-                        </span>
-                      </div>
+                      <Box style={{ maxWidth: 150 }} minWidth="0">
```
ahdriel: nit: can we use `maxWidth=""` as a prop instead?

packages/front-end/ui/VariationLabel.tsx
```diff
@@ -74,14 +77,14 @@ export default function VariationLabel({
 
   if (disableTooltip) {
     return (
-      <Box ref={rootRef} minWidth="0">
+      <Box ref={rootRef} minWidth="0" style={{ maxWidth }}>
```
ahdriel: question: why not `maxWidth=""` as a prop?

packages/front-end/styles/radix-config.css
```diff
@@ -42,3 +42,23 @@
...
 }
+
+/* The custom --color-text-* tokens above are keyed off the app-level
+   .light-theme/.dark-theme <html> class. When a nested Radix <Theme> overrides
+   the local appearance (e.g. flipTheme tooltips), Radix adds a .light/.dark
+   class on that element but the <html> class is unchanged, so the text tokens
+   would keep the outer theme's values (dark text on a dark tooltip). Re-map the
+   text tokens to the local Radix appearance class. The doubled .radix-themes
+   raises specificity above the ancestor rules so the local appearance wins. */
+.radix-themes.radix-themes.light {
```
ahdriel: suggestion: We should attempt to fix this at the root, so we don't need to redeclare the same variables twice.

Would having this rule at the same declaration work?

```
.light-theme :root,
.light-theme .radix-themes,
.radix-themes.radix-themes.light { }

.dark-theme :root,
.dark-theme .radix-themes,
.radix-themes.radix-themes.dark { }
```

ahdriel (changes requested):

packages/front-end/styles/radix-config.css
msamper: Yes! Good call, thank you

packages/front-end/components/Experiment/FieldAlignedVariationNumber.tsx
msamper: fixed!

packages/front-end/components/Experiment/UrlRedirectModal.tsx
msamper: `ModalForm` has a submit handler that catches errors thrown by the `onSubmit` and the `apiCall` throws on non okay responses so it'll be caught and won't close

packages/front-end/components/Experiment/VariationChooser.tsx
msamper: This file is unused so I may just delete it unless you think we should keep it for some reason. Otherwise yeah we should be checking `=== 1`

packages/front-end/components/Holdout/LinkedExperimentsTable.tsx
msamper: yes!

packages/front-end/components/HomePage/ExperimentImpact/ExperimentImpactTab.tsx
msamper: I think this was just a mistake on my part. Consolidating to using a `maxWidth` thats the `maxWidth` of the name plus `20` for the number

packages/front-end/ui/VariationLabel.tsx
msamper: fixed!

packages/front-end/components/Experiment/VariationChooser.tsx
msamper: deleted

packages/front-end/components/Experiment/VariationChooser.tsx
msamper: deleted

packages/front-end/components/Experiment/BaselineChooserColumnLabel.tsx
```diff
@@ -138,9 +138,11 @@ export default function BaselineChooserColumnLabel({
             setDropdownOpen(false);
           }}
         >
-          <Box style={{ maxWidth: 200, minWidth: 0 }}>
-            <VariationLabel number={variation.index} name={variation.name} />
-          </Box>
+          <VariationLabel
+            number={variation.index}
+            name={variation.name}
+            maxWidth={200}
```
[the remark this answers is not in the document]
msamper: fixed

ahdriel (approved):
nice! thanks for the cleanup 🚀

Issue #6331, opened 2026-07-10
[Bug] Ratio metric percentile capping with zeros included (ignore-zeros off): denominator is still capped over non-zero users only, numerator over all users


roger-casas:
## Summary

Scope: this affects ratio metrics capped with zeros included (percentile capping with the default "ignore zeros" off).

For such a metric, the generated SQL computes the winsorization cap for the **numerator** over *all* exposed users (zeros included) but the cap for the **denominator** over only the users with a denominator event (an `INNER JOIN`, zeros excluded). The two sides of the same ratio are capped on different populations, which biases the metric's level and can flip the sign of small effects. In effect the denominator does not honor the "ignore zeros" setting (it always excludes zeros), so with zeros included the two sides only agree when every user has the denominator event. The distortion grows as the share of users with no denominator event grows.

With "ignore zeros" on, the numerator also excludes zeros, so both sides cap over the same non-zero population and the bug does not occur. Mean metrics are unaffected (single aggregation path).

## Minimal example

100 people; 20 of them buy; one is a big spender. Every buyer's Revenue/Order is exactly €20, so the metric should return **€20**. (Money is the numerator, orders the denominator.)

| Who | How many | Orders each | Spent each |
|---|---|---|---|
| Normal buyers | 19 | 1 | €20 |
| One big spender | 1 | 20 | €400 |
| Bought nothing | 80 | 0 | €0 |

Capping trims the big spender down to normal. Where the trim lands depends on how many users are in the list: the 80 zero-value users drag the percentile down, so the same 98% mark gives a completely different cap depending on whether they are included.

| Per-user value | Cap over all 100 (zeros in) | Cap over 20 buyers (zeros out) |
|---|---|---|
| Revenue per user | €20 (big spender €400 clipped to €20) | €400 (untouched) |
| Orders per user | 1 (big spender 20 clipped to 1) | 20 (untouched) |

Build the ratio three ways. What matters is whether both sides cap over the same population:

| How the ratio is capped | Numerator = capped revenue | Denominator = capped orders | Revenue/Order |
|---|---|---|---|
| Consistent, all 100 users | €400 (over all 100) | 20 (over all 100) | 400 ÷ 20 = **€20** ✓ |
| Consistent, 20 buyers only | €780 (over 20 buyers) | 39 (over 20 buyers) | 780 ÷ 39 = **€20** ✓ |
[...]

PR #6428, opened 2026-07-21
Align legacy ratio percentile cap populations

AliFozooni:
## Summary

- Build the legacy ratio denominator's include-zero percentile population from the same cap-population skeleton as the numerator.
- Preserve the existing denominator event aggregation, conversion-window filtering, and custom aggregation semantics.
- Limit the new denominator cap-population CTE to percentile caps that include zeros; other cap-population paths remain unchanged.
- Prevent denominator-only capped bandit queries from cross-joining a numerator cap CTE that was never generated.

Addresses the remaining legacy-query analogue of #6331.

## Scope

#6331 was reported against a v4.0.0 fact-ratio query. On current `main`, fact metrics are routed through the newer fact-metric query pipeline, where numerator and denominator values already share a full-unit skeleton. The same population mismatch is still present in the legacy `MetricInterface` query path, which this PR fixes.

The affected legacy configuration is deliberately narrow: a ratio metric whose denominator uses percentile capping with `ignoreZeros` false or omitted, with at least one unit missing from—or contributing `NULL` to—the previous denominator cap input. Mean metrics, numerator-only caps, absolute caps, uncapped ratios, and denominator caps with `ignoreZeros: true` do not use the new CTE.

## Root cause and fix

The [legacy metric documentation](https://github.com/growthbook/growthbook/blob/7946ef2144967246040320f02df45e3f9e560b8f/docs/docs/metrics/legacy-metrics.mdx#L238-L240) defines the percentile population as the unit-level aggregate for all units; zeros are removed only when `ignoreZeros` is enabled. The legacy numerator cap follows that contract because `__userMetricAgg` contains an aggregate row for every unit in the numerator cap population. The denominator cap instead reads `__userDenominatorAgg`, whose inner event join contains only units with an in-window denominator row.

That inner join is correct for the existing ratio arithmetic: the final join already maps a missing denominator aggregate to zero. It is only the cap distribution that is inconsistent. Changing the event join itself would also change valid custom aggregations such as `AVG(COALESCE(value, 1))` by injecting synthetic rows.

This PR therefore leaves `__userDenominatorAgg` intact and adds a cap-only population when zeros are included:

```sql
__userDenominatorCapPopulation AS (
  SELECT COALESCE(d.value, 0) AS value
  FROM __userMetricAgg m
  LEFT JOIN __userDenominatorAgg d ON d.user_id = m.user_id
)
```

`__capValueDenominator` reads this population only when the denominator includes zeros. When `ignoreZeros` is enabled, it continues reading the original denominator aggregates and filtering zero values.

## Statistical validation

Let `pi` be the fraction of units contributing a non-null value to the previous denominator cap input, `p` the configured percentile, and `G` that value distribution. For ordinary nonnegative denominator metrics, the intended include-zero cap is zero when `p <= 1 - pi`; otherwise it is `G^-1(1 - (1 - p) / pi)`. The previous query always used `G^-1(p)`. At `p = 0.98` and `pi = 0.20`, the previous query selects the P98 among event-bearing units instead of the intended P90.

A P95 fixture chosen to avoid interpolation ambiguity makes the resulting estimand mismatch explicit:

| Units | Numerator `X` | Denominator `Y` |
| ---: | ---: | ---: |
| 80 | 0 | 0 |
| 16 | 20 | 1 |
| 4 | 400 | 20 |

Here `X = 20Y` for every unit. Consistent all-unit caps `(20, 1)` return 20, and consistent nonzero-only caps `(400, 20)` also return 20. The previous mixed caps `(20, 20)` return `400 / 96 = 4.1667`, violating proportionality.

The mismatch can also reverse a decision. With pooled P95 caps over two 100-unit arms:

- Control: 80 zeros and 20 units at `(20, 1)`.
- Treatment: 80 zeros, 16 units at `(22, 1)`, and 4 units at `(44, 2)`.

The consistent caps produce control 20, treatment 22, and a **+10%** lift. Computing only the denominator cap over the 40 units with an in-window denominator row—all nonzero in this fixture—produces control 20, treatment 18.333, and a **-8.33%** lift. The cap mismatch also changes denominator squares and numerator-denominator cross-products, so the impact is not limited to the ratio point estimate.

This does not imply a universal direction of treatment-effect bias; under exchangeable arm distributions, level distortion may cancel in the contrast. It does show that the generated query can target a different winsorized estimand than configured and can reverse effects when arm tail structure differs.

## Adjacent bandit correction

The cap-mode regression matrix exposed an independent generated-SQL error for denominator-only capped bandits. `getBanditStatisticsCTE` treated any cap as evidence that the numerator `__capValue` CTE existed, producing a dangling cross join. The renamed `hasNumeratorCapping` flag now controls that join specifically; the denominator cap remains controlled by `denominatorIsPercentileCapped`.

## Testing

```bash
pnpm test
pnpm type-check
pnpm lint:ci
pnpm --filter back-end test snapshot-metric-query.test.ts --runInBand
pnpm --filter back-end test sqlintegration.test.ts --runInBand
pnpm --filter back-end type-check
pnpm exec eslint \
  packages/back-end/src/integrations/sql/ctes/bandit-statistics-cte.ts \
  packages/back-end/src/integrations/sql/queries/snapshot-metric-query.ts \
  packages/back-end/test/integrations/sql/queries/snapshot-metric-query.test.ts \
  --max-warnings 0
pnpm exec prettier --check \
  packages/back-end/src/integrations/sql/ctes/bandit-statistics-cte.ts \
  packages/back-end/src/integrations/sql/queries/snapshot-metric-query.ts \
  packages/back-end/test/integrations/sql/queries/snapshot-metric-query.test.ts
git diff --check origin/main...HEAD
```

- 32 focused query-generation tests covering all 11 SQL dialect generators, both zero policies, omitted and asymmetric settings, custom aggregation preservation, no/absolute/numerator-only/denominator-only/dual cap modes, and the denominator-only bandit path.
- Full backend SQL integration suite: 2,289 tests and 2,248 snapshots passed with no snapshot changes.
- Full monorepo test, type-check, and lint suites passed. The backend portion completed 191 suites, 5,473 tests, and 2,251 snapshots.
- Targeted ESLint, Prettier, and `git diff --check` passed.

### DIFF
--- a/packages/back-end/src/integrations/sql/ctes/bandit-statistics-cte.ts
+++ b/packages/back-end/src/integrations/sql/ctes/bandit-statistics-cte.ts
@@ -24,7 +24,7 @@ export function getBanditStatisticsCTE(
     metricData,
     dimensionCols,
     hasRegressionAdjustment,
-    hasCapping,
+    hasNumeratorCapping,
     ignoreNulls,
     denominatorIsPercentileCapped,
   }: {
@@ -32,7 +32,7 @@ export function getBanditStatisticsCTE(
     metricData: BanditMetricData[];
     dimensionCols: DimensionColumnData[];
     hasRegressionAdjustment: boolean;
-    hasCapping: boolean;
+    hasNumeratorCapping: boolean;
     ignoreNulls?: boolean;
     denominatorIsPercentileCapped?: boolean;
   },
@@ -118,7 +118,7 @@ export function getBanditStatisticsCTE(
         `
         : ""
     }
-    ${hasCapping ? `CROSS JOIN __capValue cap` : ""}
+    ${hasNumeratorCapping ? `CROSS JOIN __capValue cap` : ""}
     ${ignoreNulls ? `WHERE m.value != 0` : ""}
     GROUP BY
       m.variation
--- a/packages/back-end/src/integrations/sql/queries/snapshot-metric-query.ts
+++ b/packages/back-end/src/integrations/sql/queries/snapshot-metric-query.ts
@@ -435,6 +435,18 @@ WITH
         ${
           denominator && denominatorIsPercentileCapped
             ? `
+          ${
+            denominator.cappingSettings.ignoreZeros
+              ? ""
+              : `, __userDenominatorCapPopulation AS (
+                  -- Materialize implicit zeros without changing denominator aggregation semantics
+                  SELECT COALESCE(d.value, 0) AS value
+                  FROM __userMetricAgg m
+                  LEFT JOIN __userDenominatorAgg d ON (
+                    d.${baseIdType} = m.${baseIdType}
+                  )
+                )`
+          }
           , __capValueDenominator AS (
             ${dialect.percentileCapSelectClause(
               [
@@ -446,7 +458,9 @@ WITH
                   sourceIndex: 0,
                 },
               ],
-              "__userDenominatorAgg",
+              denominator.cappingSettings.ignoreZeros
+                ? "__userDenominatorAgg"
+                : "__userDenominatorCapPopulation",
               `WHERE value IS NOT NULL${
                 denominator.cappingSettings.ignoreZeros ? " AND value != 0" : ""
               }`,
@@ -504,7 +518,7 @@ WITH
           ],
           dimensionCols,
           hasRegressionAdjustment: regressionAdjusted,
-          hasCapping: isPercentileCapped || denominatorIsPercentileCapped,
+          hasNumeratorCapping: isPercentileCapped,
           ignoreNulls: "ignoreNulls" in metric && metric.ignoreNulls,
           denominatorIsPercentileCapped,
         })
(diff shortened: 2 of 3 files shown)


AliFozooni:
Hi @lukebrawleysmith — could you confirm whether a fix for the remaining legacy MetricInterface analogue of #6331 is still desired? The production change here is 19 additions and 5 deletions; the other 422 added lines are regression coverage across cap modes and SQL dialects. It preserves the existing aggregation semantics and only aligns the denominator include-zero cap population with the numerator. The test matrix also exposed a separate three-line bandit cross-join guard. I am happy to split that into its own PR if that would make review easier. If you are not the right reviewer, could you point me to the owner for this SQL path? Thanks!

lukebrawleysmith:
Hi @AliFozooni , I'm not seeing the issue in prod that you describe.  I created 3 legacy ratio metrics (uncapped, capped using zeros, capped ignoring zeros) and their 3 fact metric counterparts.  I get identical results for both.  Do you have a reproducable example in prod?  If so, could you please share?  Feel free to slack me (lukesmith@GrowthBook).  Thanks!
Screenshot 2026-07-30 at 9 14 10 AM

PR #6430, opened 2026-07-21
feat(sdk): add optional user arg to trackingCallback

teresayung:
Extend the trackingCallback signature with an optional third argument: a lean UserContext carrying only the user's merged attributes. This lets consumers access the resolved attributes at track time without changing existing (2-arg) callbacks, which keep working unchanged.

- types: TrackingCallback gains `user?: UserContext`; TrackingData gains an optional `user` field so deferred calls carry the same context.
- core: new getTrackingUserContext(ctx) helper builds the lean context from merged attributes; passed to the user trackingCallback and stored on deferred tracks (onExperimentViewed + runExperiment).
- GrowthBook.fireDeferredTrackingCalls forwards the stored user arg.
- tests: experiments/features updated for the 3rd arg (undefined when a deferred call was set manually without a user field).

Split out of #6274 (contextual bandit SDK payload) so the callback shape change can be reviewed and land independently.

### DIFF
--- a/packages/sdk-js/src/core.ts
+++ b/packages/sdk-js/src/core.ts
@@ -15,6 +15,7 @@ import {
   FeatureApiResponse,
   Options,
   ClientOptions,
+  UserContext,
 } from "./types/growthbook";
 import { evalCondition } from "./mongrule";
 import { ConditionInterface } from "./types/mongrule";
@@ -100,7 +101,9 @@ function onExperimentViewed(
   }
   if (ctx.user.trackingCallback) {
     const cb = ctx.user.trackingCallback;
-    calls.push(safeCall(() => cb(experiment, result)));
+    calls.push(
+      safeCall(() => cb(experiment, result, getTrackingUserContext(ctx))),
+    );
   }
   if (ctx.global.eventLogger) {
     const cb = ctx.global.eventLogger;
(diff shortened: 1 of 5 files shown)


packages/sdk-js/src/core.ts
```diff
@@ -818,6 +823,13 @@ function getAttributes(ctx: EvalContext) {
   };
 }
 
+function getTrackingUserContext(ctx: EvalContext, deep = false): UserContext {
```
bryce-fitzsimons: We shouldn't need to do the whole deep serialize/unserialze thing here. It should be safe to just say `attributes: { ...attributes }`, even for deferred tracking. As long as they aren't manually mutating the underlying attributes then there should be no problem doing this; and we already assume that behavior throughout the SDK.

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -159,6 +159,9 @@ export type Attributes = Record<string, any>;
 export interface TrackingData {
   experiment: Experiment<any>;
   result: Result<any>;
+  // A lean UserContext carrying only the user's attributes, so deferred
+  // tracking calls stay serializable and match the trackingCallback shape.
+  user?: UserContext;
```
bryce-fitzsimons: rather than comment the intent, might be better to have a `UserContextAttributes = Pick<UserContext, "attributes">` and make that the interface instead.

teresayung: 💡 that's a great idea, thank you!

bryce-fitzsimons (approved):

PR #6436, opened 2026-07-22
Fix stats-engine readiness race causing prod-api-python OOM at startup

tzjames:
### Features and Changes

`prod-api-python` tasks were intermittently getting OOM-killed shortly after boot (CloudWatch showed individual tasks spiking to ~99.8% of the 4096 MB task memory limit in that window, while the service-wide average stayed ~40%). The cause was a readiness race: `/healthcheck` (`app.ts`) returned 200 unconditionally the instant Express started listening, well before the Python stats-engine pool (`packages/back-end/src/services/python.ts`) was actually warm — only `GB_STATS_ENGINE_MIN_POOL_SIZE` (default 1) process is eagerly spawned at boot. The ALB marks a target healthy after 3 checks at a 10s interval (~20-30s post-boot) and immediately starts routing real traffic. If several concurrent stats-engine requests land right at that moment, the pool has to cold-start multiple additional Python interpreters simultaneously to keep up — each a genuinely heavy boot (numpy/pandas/scipy import measured at ~1.7-2s, vs. ~1ms for the process spawn itself) — stacking directly on top of Next.js/Node's own still-elevated boot-time memory. That combined spike is what tipped tasks over the memory limit.

Fix:
- `stats_server.py` now emits a `{"ready": true}` signal once its import completes, and `python.ts`'s pool `create()` factory waits for that signal before considering a resource created — so `statsServerPool.available`/`size` reflect actual readiness instead of the instant a process is spawned.
- `/healthcheck` uses that to hold off returning 200 until the local pool has reached its minimum warm size, for any instance that serves stats requests from its own pool (`!EXTERNAL_PYTHON_SERVER_URL`, matching the condition `python.ts` already uses elsewhere to decide if the local pool is in play) — so the ALB doesn't route traffic until the pool can actually absorb a burst.

Also included: `server.ts` now drains and clears `statsServerPool` on graceful shutdown (SIGTERM/SIGINT). Nothing previously killed pooled Python children explicitly on shutdown — Node already holds their `ChildProcess` handles, so this is a straightforward gap-fill, unrelated to the OOM issue above.

### Testing

- [x] Spawned `stats_server.py` directly and confirmed the ready signal fires only after the import completes (~1.7-2s), with the real response coming back near-instantly once ready
- [ ] Deploy to prod and confirm ALB health checks don't flap during boot and memory stays flatter under a traffic burst right after deploy

### DIFF
--- a/packages/back-end/scripts/stats_server.py
+++ b/packages/back-end/scripts/stats_server.py
@@ -5,6 +5,12 @@ import sys
 import traceback
 from gbstats.gbstats import process_multiple_experiment_results
 
+# Signal readiness now that the slow numpy/pandas/scipy import chain above is
+# done, so the Node side doesn't treat this process as usable the instant
+# it's spawned (well before it can actually read stdin).
+sys.stdout.write(json.dumps({"ready": True}) + "\n")
+sys.stdout.flush()
+
 for line in sys.stdin:
     start = time.time()
 
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -32,6 +32,7 @@ import {
   getExperimentConfig,
   getExperimentsScript,
 } from "./controllers/config";
+import { statsServerPool } from "./services/python";
 import { getAuthConnection, processJWT, usingOpenId } from "./services/auth";
 import { wrapController } from "./routers/wrapController";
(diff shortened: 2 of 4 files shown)


tzjames:
Something else is killing it. This health check isn't necessarily needed.

PR #6439, opened 2026-07-22
added eval subs for events and features

lillialjackson:
### Features and Changes

Adds session experiment, flags, and event eval subs to the sdk.
### Dependencies

none

### Testing

manual and unit

### Screenshots

n/a


madhuchavva (commented):
Review comments focused on callback semantics, existing-callback compatibility, session-replay behavior, and a consistent contract for future SDK implementations.

packages/sdk-js/src/core.ts
```diff
@@ -141,6 +141,16 @@ function onFeatureUsage(
         logType: "feature",
       });
     }
+
+    if (ctx.user.featureEvalSubs?.size) {
```
madhuchavva: Could we clarify whether this is intended to represent every feature evaluation or the existing feature-usage/value-change signal? Because it sits inside `trackedFeatureUsage`, repeated evaluations are suppressed based only on the serialized value. A rule, source, or experiment can change without notifying subscribers if the resulting value stays the same, and override results never reach this path.

This distinction will matter if other SDKs implement the same hook. If the current behavior is intentional, could we document it explicitly and add coverage for same-value metadata changes and overrides? Otherwise, a true evaluation subscription would need to fire outside the existing usage deduplication and override exclusion.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -977,6 +998,15 @@ export class GrowthBook<
         logType: "event",
       });
     }
+    if (this._eventEvalSubs.size) {
```
madhuchavva: Could we explicitly define this as observing custom events passed through `logEvent`? `eventLogger` also receives SDK-generated events such as feature evaluations and experiment views, while these subscribers only receive explicit `logEvent` calls.

Since other SDKs are adding custom-event support, defining that boundary now should help prevent some implementations from subscribing to custom events while others subscribe to the full internal event stream.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -519,6 +525,18 @@ export class GrowthBook<
     };
   }
 
+  /** @internal — for plugin use only (e.g. sessionReplayPlugin) */
+  public _onFeatureEval(cb: FeatureEvalCallback): () => void {
```
madhuchavva: Are these subscriptions intentionally limited to the singleton `GrowthBook` class? The general `Plugin` type supports `GrowthBook`, `GrowthBookClient`, and `UserScopedGrowthBook`, but the new methods are only available on the singleton.

If these are later added to multi-user clients, I think the callback contract needs to include the exact `UserContext` used for the evaluation or `logEvent` call, similar to `FeatureUsageCallbackWithUser` and `EventLogger`. Looking up the current context later is unsafe for buffered plugins because the active user or attributes may have changed before the data is flushed.

We should also clarify whether subscribers receive the mutable internal `UserContext`, a read-only view, or a snapshot. Passing the full mutable context lets a plugin accidentally change attributes, callbacks, deduplication sets, or sticky-bucket state. Could we document the singleton-only scope now, or define the intended scoped/global callback signatures before this becomes the pattern for other SDKs?

packages/sdk-js/test/eval-subs.test.ts
```diff
@@ -0,0 +1,193 @@
+import { GrowthBook } from "../src";
+
+describe("eval subscriptions", () => {
```
madhuchavva, line 3: Could we add regression coverage showing that these subscriptions coexist with the existing `onFeatureUsage` and `eventLogger` callbacks? In particular, it would be useful to verify that both observers receive the expected call, a throwing subscriber does not prevent the existing callback from running, and unsubscribing removes only the subscriber.

Preserving existing callback behavior is an important part of making these hooks safe for plugins to install transparently.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -519,6 +525,18 @@ export class GrowthBook<
...
+    this._featureEvalSubs.add(cb);
```
madhuchavva: Since session replay will use the existing `subscribe()` API alongside these new hooks, could we document that it observes experiment assignment/result changes rather than experiment exposure tracking? That is meaningfully different from `trackingCallback`: a subscription can observe changes to `inExperiment` or `variationId`, while the tracking callback represents an exposure intended for analytics delivery.

Defining the three plugin streams together—feature usage, experiment evaluations, and custom events—would give other SDKs a consistent contract to implement.

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -189,6 +189,17 @@ export type FeatureUsageCallbackWithUser = (
   user: UserContext,
 ) => void;
 
+// Callback types for internal plugin subscriptions (e.g. session replay).
+export type FeatureEvalCallback = (
```
madhuchavva: Could we document that subscription callbacks are synchronous and that observers which buffer data must snapshot the values they need during the callback? The current `try/catch` isolates synchronous exceptions, but it will not handle a rejected promise from an async callback.

The feature result, its value, event properties, and potentially `UserContext` are also passed by reference. If a session-replay plugin stores those references and serializes them later, application mutations can change what is eventually recorded or associate it with a later user context. Either the SDK or the plugin can own the cloning, but the contract should make that responsibility explicit.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -519,6 +525,18 @@ export class GrowthBook<
...
+    return () => this._featureEvalSubs.delete(cb);
```
madhuchavva: Nit: `_subscribeFeatureUsage` and `_subscribeCustomEvents` seem to communicate the actual behavior more clearly than `_onFeatureEval` and `_onEvent`. The current names suggest every feature evaluation and potentially every SDK event, while these hooks expose the deduped feature-usage stream and explicit custom events. The `subscribe` prefix also makes the returned unsubscribe function more intuitive.

packages/sdk-js/src/types/growthbook.ts
```diff
@@ -189,6 +189,17 @@ export type FeatureUsageCallbackWithUser = (
...
+  key: string,
+  result: FeatureResult<any>,
+) => void;
+
+export type EventEvalCallback = (
```
madhuchavva: Nit: could we align the callback type names with the behavior as well? `FeatureEvalCallback` and especially `EventEvalCallback` are a little misleading because custom events are not evaluated. Something like `FeatureUsageSubscriptionCallback` and `CustomEventSubscriptionCallback` would make the intended plugin contract more self-describing.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -977,6 +998,15 @@ export class GrowthBook<
         logType: "event",
       });
     }
+    if (this._eventEvalSubs.size) {
+      this._eventEvalSubs.forEach((cb) => {
+        try {
+          cb(eventName, properties);
```
madhuchavva: Nit: subscribers currently receive `undefined` when `logEvent` has no properties, while `eventLogger` receives `{}`. Could we normalize both paths to `{}` and reuse `EventProperties` in the callback type? A single shape will be easier to carry consistently into other SDK implementations.

packages/sdk-js/src/GrowthBook.ts
```diff
@@ -519,6 +525,18 @@ export class GrowthBook<
     };
   }
 
+  /** @internal — for plugin use only (e.g. sessionReplayPlugin) */
```
madhuchavva: Nit/API hygiene: these methods are marked `@internal`, but they are public class methods and appear in the generated declarations. Should plugin authors treat them as a supported compatibility surface or as implementation details that may change without notice? A short note about stability would help, especially if session replay and eventually other plugins will depend on them.

madhuchavva (approved):

PR #6440, opened 2026-07-22
create and store gb session id

lillialjackson:
### Features and Changes

Adds gb specific session replay id and tracks for duration of session. note: wanted to call sessionReplayID since it is closely associated with session replays and wanted to maintain users being able to pass a session_id as part of attributes without issue.

### Dependencies

none
### Testing

unit and manual

### Screenshots

n/a


packages/sdk-js/src/plugins/session-replay-id.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  if (!stored || now - stored.lastTouchedAt >= SESSION_REPLAY_IDLE_TIMEOUT_MS) {
+    return;
+  }
+
+  persistSessionReplayState({ ...stored, lastTouchedAt: now });
+}
+
+// Use the browser's crypto.randomUUID if set to generate a UUID.
+export function genUUID(crypto?: Crypto) {
```
bryce-fitzsimons: can we move this to a separate util? We have the same pattern in auto-attributes.ts, and the new plugins will likely also use this.

packages/sdk-js/src/plugins/gb-session.ts
```diff
@@ -0,0 +1,67 @@
...
+  } catch {
+    inMemoryFallback = state;
+  }
+}
+
+export function getOrCreateGbSessionId(options?: {
+  forceNew?: boolean;
+  maxDuration?: number;
+}): string {
+  const forceNew = options?.forceNew ?? false;
```
bryce-fitzsimons:
```suggestion
  const forceNew = !!options?.forceNew;
```

packages/sdk-js/src/plugins/gb-session.ts
```diff
@@ -0,0 +1,67 @@
+import { genUUID } from "../util";
+
+type StoredGbSessionState = {
+  gbSessionId: string;
+  createdAt: number;
+};
+
+const SESSION_STORAGE_KEY = "gb_session";
+export const DEFAULT_MAX_DURATION_MS = 30 * 60 * 1000;
```
bryce-fitzsimons: Is 30 minutes the correct default? Maybe more like 5-10?

packages/sdk-js/src/plugins/gb-session.ts
```diff
@@ -0,0 +1,67 @@
...
+  const maxDuration = options?.maxDuration ?? DEFAULT_MAX_DURATION_MS;
+  const now = Date.now();
+  const stored = forceNew ? null : readStoredGbSessionState();
+
+  if (stored && now - stored.createdAt < maxDuration) {
+    return stored.gbSessionId;
+  }
+
+  const fresh: StoredGbSessionState = {
+    gbSessionId: genUUID(window.crypto),
```
bryce-fitzsimons: do we always have a window object here? I'd think not for serverside usage

bryce-fitzsimons: maybe `globalThis?.crypto`

bryce-fitzsimons: do we need a way to seed in alternate crypto libs?

packages/sdk-js/src/plugins/gb-session.ts
```diff
@@ -0,0 +1,67 @@
...
+  ) {
+    return null;
+  }
+
+  return { gbSessionId: stored.gbSessionId, createdAt: stored.createdAt };
+}
+
+function readStoredGbSessionState(): StoredGbSessionState | null {
+  try {
+    const raw = sessionStorage.getItem(SESSION_STORAGE_KEY);
```
bryce-fitzsimons: `sessionStorage` needs to be polyfillable for node environments

PR #6446, opened 2026-07-22
migrate docs from Docusaurus to Mintlify

alyssamichelle:
### Features and Changes

Migrates the GrowthBook docs site from Docusaurus to Mintlify.

- Adds a Mintlify docs site under `docs/` with `docs.json`, theme/branding, navigation, and migrated MDX content
- Preserves the previous Docusaurus site under `docs-archive/` for reference during the transition
- Switches docs CI from Vercel/Docusaurus typecheck to `mint validate` and `mint broken-links`

### Dependencies

- Mintlify project/hosting must be configured to point at the new `docs/` directory before this is considered live


csbailey5t:
@alyssamichelle this looks good. Do we need to be ready to do the domain switch before merging to make sure we don't interrupt the live Docs?

alyssamichelle:
@csbailey5t Yes. I assume so because Jeremy mentioned needed to switch over the DNS AND we are going to need to point to mintlify UI also to the main branch and not my branch.

PR #6448, opened 2026-07-22
Ty/cb add remove variations

teresayung:
### Features and Changes

Adds and removes variations (arms) on a running Contextual Bandit, keeping linked feature rules, weights, and analysis in sync.

**New endpoint.** `POST /api/v1/contextual-bandits/:id/variations` takes the full desired variation list plus optional per-feature values for new arms. `variations` and `variationWeights` are removed from the generic `PUT /contextual-bandits/:id` body — arm changes have side effects on linked features and weights, so they no longer ride along with metadata edits.

**Arm lifecycle.** Variations gain a `status`:

- `active` (default when absent) — served, weighted, analyzed
- `pending` — added, but its value isn't live yet on every linked feature; holds no weight and receives no traffic
- `deactivated` — removed; tombstoned so the id is never reused or resurrected

Tombstones are stripped from API responses; `pending` surfaces so the UI can warn.

**Linked features.** Adding an arm writes a value into every linked feature's CB rule (provided value → control value → feature default, type-validated). Removing one drops it from those rules. When the bandit is running the rule change auto-publishes; if that fails it's staged as a draft and returned in `featureDraftPublishFailures` with a reason (`needs-approval`, `merge-conflict`, `publish-error`). Pending arms activate automatically from the `publishRevision` hook, and re-saving the variations retries any staged drafts.

**Weights.** Reconciled over active arms only — uniform during `explore`, proportional redistribution during `exploit`/`paused`. Leaf weights reconcile the same way; `banditVersion` bumps whenever the arm set or ordering changes.

**Analysis correctness.** Snapshots record the `banditVersion` they ran against. If the arm set changes mid-run, the returned weights are discarded rather than applied to a different arm set. Engine results are now keyed to the run's frozen variation list instead of today's `cb.variations`, and snapshot settings count active arms only.

**UI.** The Edit Variations modal sets values for new arms on each linked feature, explains what will happen to weights for the current stage, and surfaces publish failures. The detail page warns when arms are pending. Editing variations is disabled once the bandit is stopped.

### Dependencies

None. No migration required — `status` is optional and absent means `active`, so existing Contextual Bandits keep working unchanged.

### Testing

~1,700 lines of new unit tests covering the weight/diff helpers, the variation-change service, add-variation values, payload generation, and the SDK.

```
yarn workspace shared test contextual-bandit-variation-changes
yarn workspace back-end test contextualBanditVariationChange
yarn workspace back-end test contextualBanditAddVariationValues
yarn workspace back-end test contextualBanditPayload
yarn workspace back-end test contextualBandits
yarn workspace @growthbook/growthbook test contextual-bandit
```

Manual, on a running Contextual Bandit with a linked feature:

1. **Add an arm** — set its value in the modal, save. Confirm the feature rule gains the arm, weights re-balance for the current stage, and the arm serves traffic.
2. **Add an arm on a feature requiring approval** — save. Confirm the modal reports the publish failure, the detail page shows the pending warning, the arm gets no traffic, and it activates on its own once the revision is approved and published.
3. **Remove an arm** — confirm it's dropped from the feature rule, its weight redistributes across survivors, and re-adding the same variation is rejected.
4. **Add an arm mid-refresh** — confirm the in-flight snapshot's weights are discarded (`banditVersion` mismatch logged) rather than applied to the new arm set.
5. **Stopped bandit** — confirm the edit-variations affordance is gone.


teresayung:
@greptileai

PR #6457, opened 2026-07-24
perf(saved-groups): drop condition from /organization/definitions

ahdriel:
### Features and Changes

To optimize `/organization/definitions` performance, we previously removed `values` from Saved Groups as it is not used. This now removes `condition` to further reduce the payload size.

The pages that actually need the heavy fields fetch them on demand instead:

- The Saved Groups admin page reads a new internal `GET /saved-groups` that keeps `condition` (values still stripped to match previous behavior)
- Single group page needs all the information, it already had the logic but now it is extracted in a `useFullSavedGroup` hook

### Testing

- Unit tests
- Local stress testing


packages/front-end/hooks/useFullSavedGroup.ts
```diff
@@ -0,0 +1,14 @@
+import { SavedGroupInterface } from "shared/types/saved-group";
+import useApi from "@/hooks/useApi";
+
+/**
+ * Fetches one full saved group, including the heavy `condition` and `values`
+ * fields that `/organization/definitions` omits. Use for edit/detail/diff;
+ * for lists, prefer the light `savedGroups` from `useDefinitions()`.
+ */
+export function useFullSavedGroup(id: string | undefined | null) {
```
bryce-fitzsimons: Do we need this?

bryce-fitzsimons (approved):
some minor cleanup needed, otherwise tested and lgtm

packages/front-end/hooks/useFullSavedGroup.ts
ahdriel: good catch -- we were using it before but ended up refactoring and forgot to clean up

Kevin-Chant (approved):

PR #6459, opened 2026-07-24
agentic-bugfix/Surface a clear error when legacy JSONExtract calls hit native JSON columns

dlin-growthbook:
### Features and Changes

Post-migration to native JSON attributes/properties columns, legacy JSONExtractString(attributes, 'key')-style queries throw ClickHouse's raw illegal type: JSON exception in the SQL Explorer, with no indication of the actual fix. Catch this specific error in ClickHouse.runQuery and surface a stable code + docs-linked callout instead, mirroring the existing ManagedWarehousePendingError pattern:

ManagedWarehouseJsonSyntaxError (shared/util) — thrown only for growthbook_clickhouse datasources when the ClickHouse message matches illegal type: JSON + a JSONExtract/has( call
formatQueryExecutionErrorForApi now maps it to a stable managed_warehouse_json_syntax_error code
New ManagedWarehouseJsonSyntaxCallout renders guidance + a link to the dot-notation docs section

Not in scope: a backward-compatible shim that makes old queries "just work." Unlike the identifier/dimension aliasing already done for named migrated columns, SQL Explorer allows arbitrary JSONExtract paths — there's no generic way to alias those, so the fix is a clear error, not silent compatibility.

TODO for reviewer: wire ManagedWarehouseJsonSyntaxCallout in wherever the query-result panel currently checks isManagedWarehousePendingQueryError — didn't have visibility into that exact call site.

### Testing

Run a Managed Warehouse SQL Explorer query using old-style JSONExtractString(attributes, 'key') syntax and confirm it now shows the JSON-syntax callout with a link to the docs, instead of a raw ClickHouse exception. Confirm unrelated ClickHouse errors (bad table name, syntax error, etc.) still pass through unchanged, and that BYO-ClickHouse datasources are unaffected.

### DIFF
--- a/packages/back-end/src/integrations/ClickHouse.ts
+++ b/packages/back-end/src/integrations/ClickHouse.ts
@@ -11,7 +11,9 @@ import {
   isManagedWarehouse,
   isManagedWarehouseAwaitingJsonMigration,
   isManagedWarehouseAwaitingProvisioning,
+  isManagedWarehouseJsonSyntaxError,
   isManagedWarehouseMigrating,
+  ManagedWarehouseJsonSyntaxError,
   ManagedWarehousePendingError,
 } from "shared/util";
 import { SqlDialect } from "shared/types/sql";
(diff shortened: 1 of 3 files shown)


Kevin-Chant:
Closing this since a Clickhouse update to 26.4 also fixes it

PR #6460, opened 2026-07-24
Fix rollout rules re-bucketing when a legacy flag is saved

bryce-fitzsimons:
## The bug

Rollout rules created before we started persisting a `seed` were stored without one, and every SDK buckets a seedless rollout by the **feature key** (`rule.seed || featureId`). When such a flag was next saved or published, the server backfilled the missing seed with the **rule id** — silently re-drawing which users were in the rollout (same coverage %, different people). This hit three production rollouts.

## The fix

- **New rollout rules seed off their own rule id**, so stacked rollouts hash independently — a 50% + 50% stack serves ~75% of traffic rather than 50%. This preserves the independence Safe Rollouts have always had via their random seed.
- **Pre-existing seedless rollout rules are pinned to the feature key at read time** — the SDK's own fallback — so saving or publishing a legacy flag never re-draws its cohort. They settle onto that seed permanently on their next save.
- **Every path that creates or edits a rollout now seeds it consistently**: the bulk feature update (v1 + v2), the per-rule add and edit endpoints, the UI rule editor, and rule/feature duplication. Several of these previously left a rollout seedless — including a force rule promoted to a rollout by dropping coverage below 100% — which meant a second stacked rollout silently served no traffic.
- **`seed` and `hashVersion` survive the v1 and v2 REST round-trips.** v2's rule input schema didn't accept `seed`, so the documented GET → edit → PUT flow dropped it and re-bucketed the rollout; `hashVersion` wasn't returned at all. A bulk update that omits either now inherits it from the stored rule, so a client that rebuilds the rules array can't reset it.
- **Duplicating a rollout rule or a feature no longer copies the seed**, so a copy buckets independently of its original instead of serving the same users.

## Rules the earlier backfill already re-seeded

Those rules keep their current seed here — we can't distinguish an intentional seed from a backfilled one, and rewriting would move their cohort a second time. Restoring one to its original cohort is a manual seed edit: set it back to the feature key (hashing is deterministic, so that reproduces the original membership exactly).

One caveat: those rules can still be rewritten by the existing feature/revision drift repair, which runs on a feature page load and treats the revision as authoritative. That is pre-existing behavior rather than something introduced here — and this PR stabilizes it, because once the feature and revision agree on the feature-key seed they stop drifting. Tracking separately.

## Testing

- Unit tests for the read-time pin, stored-seed inheritance, both REST converters, the force→rollout promotion, and the publish path.
- Verified against the running API: new rollouts get independent seeds, v1 and v2 round-trips preserve them, and a seedless rule resolves to the feature key in both the API response and the SDK payload without moving anyone.
- Measured the stacking fix end-to-end with the real SDK over 20k users: two stacked 50% rollouts enrolled **49.6%** before and **75.0%** after.

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -7153,6 +7153,11 @@ paths:
                             type: number
                           hashAttribute:
                             type: string
+                          seed:
+                            description: >-
+                              Optional seed for the hash function; defaults to
+                              the rule id
+                            type: string
                           hashVersion:
                             anyOf:
                               - type: number
@@ -7852,6 +7857,11 @@ paths:
                           coverage:
                             type: number
                           hashAttribute:
+                            type: string
+                          seed:
+                            description: >-
+                              Optional seed for the hash function; defaults to
+                              the rule id
                             type: string
                           hashVersion:
                             anyOf:
--- a/packages/back-end/src/api/features/postFeatureRevisionRuleAdd.ts
+++ b/packages/back-end/src/api/features/postFeatureRevisionRuleAdd.ts
@@ -253,9 +253,7 @@ export const postFeatureRevisionRuleAdd = createApiRequestHandler(
     const rule = buildRuleFromInput(ruleInput, uuidv4());
 
     // Seed a new rollout off its own rule id so stacked rollouts hash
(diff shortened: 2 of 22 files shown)


gazzdingo (approved):
LGTM after you fix the few things

PR #6471, opened 2026-07-27
Add EU (eu-west-1) region support to managed warehouse and Event Forwarder

tzjames:
## Summary
- Adds a self-serve "Data Region" (`us-east-1` / `eu-west-1`) selector when creating a Managed Warehouse datasource (both the normal-user and admin flows) and when configuring an Event Forwarder — region is stored on the datasource/forwarder doc for central-license-server and the ingestor to route on downstream.
- Event Forwarder gets its own independent region field (rather than inheriting from a managed-warehouse datasource), since a forwarder doesn't require an org to have a managed warehouse at all.
- Updates the `managed-warehouse.mdx` and `event-forwarder.mdx` docs to cover both regional ingestion hostnames and note that the SDK tracking config must match the region chosen for the datasource/forwarder.

## Dependencies
License Server PR needs to land first to be able to create the DB in the eu.
https://github.com/growthbook/central-license-server/pull/169
Ingestor needs to land for the eu ingestor service to pick it up.
https://github.com/growthbook/growthbook-ingestor/pull/90

## Test plan
- [x] `pnpm --filter shared type-check`, `pnpm --filter back-end type-check`, `pnpm --filter front-end type-check` all pass
- [ ] Manually create a managed warehouse as a normal user and as super-admin (via `/admin`), confirming the chosen region round-trips into `settings.region`
- [ ] Manually create/edit an Event Forwarder, confirming region is stored and immutable after creation

### DIFF
--- a/docs/docs/experimentation-analysis/event-forwarder.mdx
+++ b/docs/docs/experimentation-analysis/event-forwarder.mdx
@@ -195,6 +195,21 @@ req.growthbook.logEvent("Sign Up", {
 });
 ```
 
+#### Sending events to the right region
+
+The tracking plugin sends events to `https://us1.gb-ingest.com` by default. If you selected **eu-west-1** as your Event Forwarder's Data Region, override the ingestor host so events reach the right Kafka/Confluent resources instead of being dropped:
+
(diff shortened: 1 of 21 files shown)


lillialjackson (approved):
LGtm!

PR #6473, opened 2026-07-27
chore(lint): enforce exhaustive switches and split prettier

ahdriel:
### Features and Changes

A really good strategy to ensure that, as we expand with new capabilities, is using an exhaustive switch to ensure that previous behavior that handled 3 values now also needs to be updated to handle the new 4th value.

It helps the engineer to realize that 'oh true, we also need to handle this scenario' and it might be as simple as adding it to an already existing `case`, or creating a custom behavior if needed.

This PR does that via linting with the `@typescript-eslint/switch-exhaustiveness-check` rule.

**Note:** Currently all violations are recorded in `eslint-suppressions` but we will work on removing them soon.

### Prettier for mdx and other files

We used prettier to format our docs for example, but we never enforced it in CI. Now we are enforcing it as well.

### Engineering notes

Before we used `prettier` with `eslint`, but that made some processes more expensive due to how these 2 tools interact. To enable the `switch-exhaustiveness` rule we also needs to make ESLint TypeScript aware, which is more expensive.

So we split `prettier` from `eslint` and updated some rules (following this [performance guide](https://typescript-eslint.io/troubleshooting/typed-linting/performance/)) to attempt to keep the linting runtime the same.

Additionally, we had to remove eslint cache because otherwise the typescript rules would be stale because of how eslint manages its cache.

### Testing

- `pnpm lint:ci`
- `pnpm pretty:check`
- `pnpm type-check`


najito (approved):
Nice. Makes sense. Appreciate the documentation update as well in the contribution guide.

PR #6475, opened 2026-07-27
Add static dimension support to Group By editor in dashboard explorer

najito:
### Features and Changes

Adds a "pin specific values" workflow to the Group By editor in the dashboard/product-analytics explorer, built on top of the existing `static` dimension type — which already existed in the schema and SQL layer but had no authoring path (the editor only ever created `dynamic` dimensions, and the AI agent silently downgraded any `static` dimension it received to `dynamic`).

- **`GroupBySection.tsx`**: dimension rows for both `dynamic` (top-N) and `static` (pinned values) now render together. A new "Values" multi-select under Advanced Options lets you pin specific column values — doing so switches the dimension to `static` (column + up to 20 values); clearing all pinned values switches it back to `dynamic`, restoring the last max-values limit that was set. Changing the dimension's column always resets to a fresh dynamic dimension, since pinned values (and any remembered max-values limit) are scoped to the previous column.
- **`util.ts`**: added `getDimensionColumnTopValues` to source real column values for the Values multi-select from cached fact-table `topValues`, resolved per dataset type (fact_table / metric / funnel). Extended `validateDimensions` so stale `static` dimensions (whose column no longer exists) get cleaned up the same way `dynamic` ones already were.
- **`shared/enterprise/product-analytics/sql.ts`**: rows outside a static dimension's pinned values are now dropped entirely via a `column IN (...)` filter — no "other" bucket — for both the standard query builder and the funnel query builder. For funnels, the filter is applied post-aggregation on the first-touch `dimension_1` column rather than on raw per-event rows, since funnel dimensions are per-user first-touch values, not per-row ones; this also keeps it from affecting other steps that happen to share the same underlying fact table. The filter is resolved once against the same fact table the dimension's SELECT expression itself uses, so it stays consistent even when a ratio metric's numerator and denominator live on different fact tables.
- **`shared/validators/product-analytics.ts`**: `staticDimensionValidator.values` now requires 1–20 entries (previously unconstrained).
- **`product-analytics-agent.ts`** (AI chat): the agent can now author `static` dimensions directly instead of always downgrading them to `dynamic`.
- Regenerated `spec.yaml` from the updated validator, and removed the now-unused `GroupBySection.tsx` eslint suppression by fixing the underlying `size="legacy"` lint errors instead of continuing to suppress them.

This supersedes #6467, which took the same idea in a different, ad-hoc direction (bolting a `values` field onto `dynamic` instead of using the existing `static` dimension type).

### Dependencies

None

### Testing

- Added/updated unit tests:
  - `shared/test/product-analytics-dimension.test.ts`: static dimension SQL expression now returns the raw column (no CASE/'other'), plus validator min/max constraint coverage.
  - `shared/test/product-analytics.test.ts`: static-dimension row filtering for the standard query builder, including a regression test for a ratio metric whose denominator lives on a different fact table than the pinned dimension's column.
  - `shared/test/sql.funnel.test.ts`: static dimension filter applied post-aggregation on `dimension_1`.
  - `front-end/test/components/ProductAnalytics/util.test.ts`: `getDimensionColumnTopValues` coverage across dataset types.
- Manually verify in the dashboard/explorer UI: add a Group By dimension, expand Advanced Options, pin specific values, and confirm the chart only includes those values (no "other" bucket) and Max Values becomes disabled while pinned.

### Screenshots
Screenshot 2026-07-30 at 9 38 56 AM
Screenshot 2026-08-04 at 1 20 09 PM
Screenshot 2026-08-04 at 1 20 16 PM
Screenshot 2026-07-27 at 4 05 35 PM

### DIFF
--- a/packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
+++ b/packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
@@ -384,6 +384,7 @@ export default function GroupBySection() {
                       }
                       options={BREAKDOWN_TYPE_OPTIONS}
                       sort={false}
+                      containerStyle={{ marginBottom: 0 }}
                     />
                   </Flex>
                 </Tooltip>
@@ -450,6 +451,7 @@ export default function GroupBySection() {
                         : undefined
                     }
                     errorLevel="warning"
+                    showCopyButton={false}
                   />
                 )}
               </Flex>
--- a/packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
+++ b/packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
@@ -2,17 +2,64 @@ import { Flex } from "@radix-ui/themes";
 import { PiCaretDown, PiCaretRight, PiPlus, PiX } from "react-icons/pi";
 import { useEffect, useMemo, useRef, useState } from "react";
 import Collapsible from "react-collapsible";
+import type {
+  ProductAnalyticsDimension,
+  ProductAnalyticsDynamicDimension,
+  ProductAnalyticsStaticDimension,
+} from "shared/validators";
 import Button from "@/ui/Button";
-import { getMaxDimensions } from "@/enterprise/components/ProductAnalytics/util";
+import {
+  getMaxDimensions,
+  getColumnTopValues,
+} from "@/enterprise/components/ProductAnalytics/util";
 import { useExplorerContext } from "@/enterprise/components/ProductAnalytics/ExplorerContext";
+import { useDefinitions } from "@/services/DefinitionsContext";
 import Text from "@/ui/Text";
 import Tooltip from "@/ui/Tooltip";
 import SelectField from "@/components/Forms/SelectField";
 import Field from "@/components/Forms/Field";
+import MultiSelectField from "@/ui/MultiSelectField";
+
+const DEFAULT_MAX_VALUES = 5;
+const MAX_PINNED_DIMENSION_VALUES = 20;
+
+type GroupByDimension =
+  | ProductAnalyticsDynamicDimension
+  | ProductAnalyticsStaticDimension;
+
+function isGroupByDimension(
+  dim: ProductAnalyticsDimension,
+): dim is GroupByDimension {
+  return dim.dimensionType === "dynamic" || dim.dimensionType === "static";
+}
+
+// Explicit choice, not inferred from pinned values — one input shows at a
+// time, and it scales to future dimension types.
+const BREAKDOWN_TYPE_OPTIONS: { label: string; value: "dynamic" | "static" }[] =
+  [
+    { label: "Top values", value: "dynamic" },
+    { label: "Pinned values", value: "static" },
+  ];
+
+// Shifts keys above `removedIndex` down by one and drops it, so index-keyed
+// maps stay aligned after a splice-out instead of losing other dimensions' state.
+function reindexAfterRemoval<T>(
+  record: Record<number, T>,
+  removedIndex: number,
+): Record<number, T> {
+  const next: Record<number, T> = {};
+  Object.entries(record).forEach(([key, value]) => {
+    const idx = Number(key);
+    if (idx === removedIndex) return;
+    next[idx > removedIndex ? idx - 1 : idx] = value;
+  });
+  return next;
+}
 
 export default function GroupBySection() {
   const { draftExploreState, setDraftExploreState, commonColumns } =
     useExplorerContext();
+  const { getFactTableById, getFactMetricById } = useDefinitions();
   const [advancedSettingsOpen, setAdvancedSettingsOpen] = useState(
     Array(draftExploreState.dimensions.length).fill(false),
   );
@@ -21,15 +68,38 @@ export default function GroupBySection() {
   >({});
   const latestMaxValuesRef = useRef<Record<number, string>>({});
   const skipBlurCommitRef = useRef(false);
+  // True when the last Values selection/paste exceeded the cap, so the
+  // field can surface that extras were dropped.
+  const [capExceeded, setCapExceeded] = useState<Record<number, boolean>>({});
 
-  const prevDimensionsLengthRef = useRef(draftExploreState.dimensions.length);
+  // Tracks the length these maps were last synced to; Add/Remove update it
+  // directly. A mismatch means something outside those handlers reshaped
+  // `dimensions` (e.g. changeChartType) — full reset only then.
+  const expectedDimensionsLengthRef = useRef(
+    draftExploreState.dimensions.length,
+  );
   useEffect(() => {
-    if (draftExploreState.dimensions.length < prevDimensionsLengthRef.current) {
+    if (
+      draftExploreState.dimensions.length !==
+      expectedDimensionsLengthRef.current
+    ) {
       setLocalMaxValues({});
       latestMaxValuesRef.current = {};
+      setCapExceeded({});
+      expectedDimensionsLengthRef.current = draftExploreState.dimensions.length;
     }
-    prevDimensionsLengthRef.current = draftExploreState.dimensions.length;
   }, [draftExploreState.dimensions.length]);
+
+  // handleAdd/RemoveDimension already keep this in sync for their own
+  // changes. A mismatch means something else reshaped `dimensions` (e.g.
+  // changeChartType) — resync then, not on every ordinary Add/Remove.
+  useEffect(() => {
+    if (advancedSettingsOpen.length !== draftExploreState.dimensions.length) {
+      setAdvancedSettingsOpen(
+        Array(draftExploreState.dimensions.length).fill(false),
+      );
+    }
+  }, [draftExploreState.dimensions.length, advancedSettingsOpen.length]);
 
   const availableColumns = useMemo(() => {
     // Filter out columns already used in dimensions
@@ -43,8 +113,7 @@ export default function GroupBySection() {
 
   const getColumnOptionsForDimension = (index: number) => {
     const dim = draftExploreState.dimensions[index];
-    if (!dim || dim.dimensionType !== "dynamic" || !("column" in dim))
-      return [];
+    if (!dim || !isGroupByDimension(dim)) return [];
     const usedByOthers = new Set(
       draftExploreState.dimensions
         .map((d, i) => (i !== index && "column" in d ? d.column : null))
@@ -57,6 +126,9 @@ export default function GroupBySection() {
 
   const handleAddDimension = () => {
     setAdvancedSettingsOpen((prev) => [...prev, false]); // New dimension defaults to collapsed
+    // Appending doesn't shift existing indices — just mark this length
+    // change as already accounted for.
+    expectedDimensionsLengthRef.current += 1;
     setDraftExploreState((prev) => ({
       ...prev,
       dimensions: [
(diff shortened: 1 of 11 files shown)


packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
```diff
@@ -155,8 +298,16 @@ export default function GroupBySection() {
       </Flex>
       {/* Display existing dimensions */}
       {draftExploreState.dimensions.map((dim, i) => {
-        if (dim.dimensionType === "date") return null; // Skip date dimension as it's usually handled separately or fixed
-        if (dim.dimensionType !== "dynamic") return null; // Skip static and slice dimensions for now
+        if (!isGroupByDimension(dim)) return null; // Skip date and slice dimensions for now
+        const isStatic = dim.dimensionType === "static";
+        const pinnedValues = isStatic ? dim.values : [];
+        const valueOptions = getDimensionColumnTopValues(
```
lillialjackson: Do we want this called every render or do we want to add to useMemo so that this is only called on change and initial render?

packages/front-end/enterprise/components/ProductAnalytics/util.ts
```diff
@@ -598,7 +639,8 @@ export function validateDimensions(
   const maxDims = getMaxDimensions(config.dataset);
 
   let validDimensions = config.dimensions.filter((d) => {
-    if (d.dimensionType !== "dynamic") return true;
+    if (d.dimensionType !== "dynamic" && d.dimensionType !== "static")
+      return true;
     if (columns.length === 0) return true;
     return columns.some((c) => c.column === d.column || d.column === null);
```
lillialjackson, line 660: Robot flagged a potential prexisting bug here that would extend to your work here:  validateDimensions doesn't validate static dimensions with JSON (dotted) column paths correctly.                                                                                                                
                                                                                                                                                                                                                     
  In util.ts:645:                                                                                                                                                                                                    
  return columns.some((c) => c.column === d.column || d.column === null);                                                                                                                                            
                                                                                                                                                                                                                   
  columns contains top-level column names (e.g. "props"), but a static dimension on a JSON field has d.column = "props.plan". The strict equality c.column === d.column would fail — "props" !== "props.plan" — so   
  the dimension would be filtered out as invalid, even though it's a perfectly valid JSON column reference. Dynamic dimensions have the same issue, but it was pre-existing. This branch extends the bug to static   
  dimensions by adding them to this filter.                                                                                                                                                                          
                                                                                                                                                                                                                     
  This means: if a user pins values on a JSON column, then the dataset changes and validateDimensions runs, the static dimension gets incorrectly pruned.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
najito: Getting the top values is relatively cheap here. Also you’d still invalidate whenever dimensions / dataset / definitions change, which is exactly when this section already re-renders.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
```diff
@@ -119,6 +228,40 @@ export default function GroupBySection() {
     delete latestMaxValuesRef.current[index];
   };
 
+  // Pinning/clearing values toggles the dimension between "static" (pinned
+  // list, no "other" bucket) and "dynamic" (top-N) — the two are mutually
+  // exclusive, so there's no combined shape to represent both at once.
+  const handleValuesChange = (index: number, values: string[]) => {
```
lillialjackson, line 235: Another raised by ai review for bugs, though this one seems like its not as likely to have much impact: 
 handleValuesChange reads dim from draftExploreState but writes column into the functional updater.                                                                                                              
                                                                                                                                                                                                                   
  In GroupBySection.tsx:234-262:                                                                                                                                                                                     
  const dim = draftExploreState.dimensions[index];                                                                                                                                                                 
  // ...                                                                                                                                                                                                             
  const column = dim.column;       // captured from render-time state                                                                                                                                                
  // ...                                                             
  if (dim.dimensionType === "dynamic") {                                                                                                                                                                             
    lastMaxValuesRef.current[index] = dim.maxValues;  // also from render-time                                                                                                                                     
  }                                                                                                                                                                                                                  
                                                                                                                                                                                                                     
  setDraftExploreState((prev) => ({                                                                                                                                                                                  
    dimensions: prev.dimensions.map((d, i) => {                                                                                                                                                                      
      if (i !== index || !isGroupByDimension(d)) return d;                                                                                                                                                         
      if (capped.length === 0) {                                                                                                                                                                                     
        return { dimensionType: "dynamic", column, maxValues: ... };                                                                                                                                                 
        //                                 ^^^^^^ stale closure value                                                                                                                                                
      }                                                                                                                                                                                                              
      return { dimensionType: "static", column, values: capped };                                                                                                                                                    
      //                                ^^^^^^ stale closure value                                                                                                                                                 
    }),                                                                                                                                                                                                              
  }));                                                                                                                                                                                                             
                                                                                                                                                                                                                     
  The column inside the functional updater comes from the closure, not from prev.dimensions[index].column. If a queued handleColumnChange runs first (changing the column), the functional updater in              
  handleValuesChange would overwrite it with the old column. In practice this is low-risk since both handlers fire from different UI controls and React batches synchronous state updates, but the pattern is        
  inconsistent with using a functional updater — the whole point of the (prev) => form is to read from prev, not from a stale closure. The fix would be const column = isGroupByDimension(d) ? d.column : column;
  inside the .map().

packages/shared/src/validators/product-analytics.ts
```diff
@@ -155,6 +155,13 @@ export const dynamicDimensionValidator = z.object({
 export const staticDimensionValidator = z.object({
   dimensionType: z.literal("static"),
   column: z.string(),
+  // Intentionally unbounded (matches dynamicDimensionValidator.maxValues,
+  // which is also unconstrained) — this schema is used to parse
+  // already-persisted/URL-encoded explorations, not just fresh writes from
+  // the editor, so a strict min/max here would reject previously valid
+  // saved configs. The editor caps pins at MAX_PINNED_DIMENSION_VALUES (20)
+  // and never produces an empty pinned list; the SQL layer skips the filter
+  // entirely for an empty values array rather than emitting invalid SQL.
```
jdorn: Shorter comments

packages/shared/src/enterprise/product-analytics/sql.ts
```diff
@@ -1648,11 +1704,31 @@ export function generateProductAnalyticsSQL(
...
+    // so each group only gets the filter for dimensions its own fact table
+    // can actually resolve, and the filter expression is built against that
+    // same table's column metadata — otherwise the group's rows are left
+    // unfiltered, or filtered using another table's incompatible column
+    // expression.
+    const applicableDimensions = config.dimensions.filter(
+      (d) =>
+        d.dimensionType !== "static" ||
+        factTableHasResolvableColumn(factTableGroup.factTable, d.column),
+    );
```
jdorn: If a fact table doesn't include the dimension column, don't just omit the filter. That will return all rows in the fact table.  Instead, don't include any data from the fact table.  If you are splitting by `browser` and said you want `chrome` and `safari`, it doesn't make sense to include rows that don't have a browser field. They should be excluded

packages/front-end/enterprise/components/ProductAnalytics/util.ts
```diff
@@ -537,6 +540,44 @@ export function getCommonColumns(
...
+/** Cached top values for a dimension's selected column, resolved against
+ *  whichever fact table(s) the dataset draws from. Seeds the "pin specific
+ *  values" multi-select with real column values; empty for data_source
+ *  datasets, which carry no cached column metadata. */
+export function getDimensionColumnTopValues(
+  dataset: ExplorationDataset | null,
+  column: string | null,
+  getFactTableById: (id: string) => FactTableDefinition | null,
+  getFactMetricById: (id: string) => FactMetricInterface | null,
+): string[] {
```
jdorn, line 567: This is not specific to dimensions and can also be used for the filtering UI.  Lets rename this to be more generic `getColumnTopValues`.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
```diff
@@ -2,17 +2,57 @@ import { Flex } from "@radix-ui/themes";
...
+  removedIndex: number,
+): Record<number, T> {
+  const next: Record<number, T> = {};
+  Object.entries(record).forEach(([key, value]) => {
+    const idx = Number(key);
+    if (idx === removedIndex) return;
+    next[idx > removedIndex ? idx - 1 : idx] = value;
+  });
+  return next;
+}
```
jdorn, line 57: The changes in this file are pretty hacky. I think it would be a lot better if we made the dimension type an explicit choice in the UI. Then we wouldn't need to infer the dimension type and we would only show one value input at a time. Would also scale better to additional dimension types.

najito: Changed to explicit dimension selection before showing either pinned values or max values.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
najito: There was a refactor that should have made this irrelevant.

packages/front-end/enterprise/components/ProductAnalytics/util.ts
najito: I can't seem to find this issue?

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
```diff
@@ -57,38 +126,83 @@ export default function GroupBySection() {
...
+        if (d.dimensionType === "static") {
+          return { dimensionType: "static", column, values: [] };
+        }
+        return {
+          dimensionType: "dynamic",
+          column,
+          maxValues: DEFAULT_MAX_VALUES,
+        };
+      }),
+    }));
```
jdorn, line 170: There's a bug today with this - select "browser" as the dimension and add pinned values.  Then collapse "Advanced Options" and switch the column to "country" - nothing happens and the change isn't reflected in the graph and there's no indication of what the user has to do to get it to update.

What I would recommend is to simplify this logic - when changing the column, always switch static dimensions to dynamic instead of just wiping the array of values.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
```diff
@@ -21,16 +68,39 @@ export default function GroupBySection() {
   >({});
   const latestMaxValuesRef = useRef<Record<number, string>>({});
   const skipBlurCommitRef = useRef(false);
+  // True when the last Values selection/paste exceeded the cap, so the
+  // field can surface that extras were dropped.
+  const [capExceeded, setCapExceeded] = useState<Record<number, boolean>>({});
```
jdorn: Don't need separate state for this. We can just slice the array in the update handler and have help text on the MultiSelect field saying "Select up to X values".

packages/front-end/enterprise/components/ProductAnalytics/util.ts
```diff
@@ -485,6 +488,22 @@ export function getCommonColumns(
         const ft = getFactTableById(factMetric.numerator.factTableId);
         valueColumns = ft?.columns || [];
         ft?.userIdTypes?.forEach((u) => userIdTypes.add(u));
+
+        // A ratio metric's denominator can live on a different fact table —
+        // only offer columns both sides can resolve, so a dimension can
+        // never be picked that a group-by query can't evaluate.
```
mknowlton89, line 494: My robots flagged a potential issue:

Static JSON dimensions can break cross-table ratio metrics
getCommonColumns checks denominator compatibility using only the base column name (props), not nested JSON fields (props.plan). The UI can therefore offer props.plan even when the denominator lacks that field. SQL then adds WHERE 1 = 0 to the denominator CTE, producing a null/zero denominator rather than rejecting the dimension.
packages/front-end/enterprise/components/ProductAnalytics/util.ts:492-504
packages/shared/src/enterprise/product-analytics/sql.ts:433-450

mknowlton89 (commented):
Looks really good! Left one comment about a potential bug.

Aside from that, my comments are only minor.

1) When a user selects `Pinned values` - I think we should require at least 1 value added to the values before we allow the user to refresh. I think the ExplorerContext has a helper that outlines if a config is submittable that we can lean on.

2) Idk if there's much value in the `copy` icon in the pinned values input field. My vote would be to remove unless we have an explicit idea for it

Image

3) I think the padding between the `Breakdown Type` and `Values` is a bit large. Can we clean that up so it's more consistent with the other sections on the configuration tab

Image

packages/front-end/enterprise/components/ProductAnalytics/util.ts
najito: This issue appears to be something that predates this PR in the dynamic dimensions. Having the static dimensions accessible by users with this opens another way for this problem to surface and requires some engineering effort to fix.

packages/front-end/enterprise/components/ProductAnalytics/SideBar/GroupBySection.tsx
najito: Resets to dynamic dimension now when dimension is changed.

mknowlton89 (approved):
Tested and it looks good to me - I know we're doing the json column work in a separate PR.

Also calling out there's a small merge conflict, but approving so you can merge when that's resolved (no pressure on merging today, I'm good with waiting until Monday)

PR #6479, opened 2026-07-28
feat(sdk-js): consume contextual bandit payload

teresayung:
### Features and Changes

Adds SDK-side support for consuming contextual bandit (CB) payloads to `@growthbook/growthbook` (`packages/sdk-js`). This is the SDK-only slice split out of the larger contextual-bandit branch; the back-end payload generation, versioning capability, and app/UI changes ship separately.

The SDK can now read CB definitions from the payload and select per-context ("leaf") variation weights at evaluation time, while remaining fully backwards compatible with existing payloads and older SDK behavior.

**What changed:**

- **New payload fields (all optional):** `contextualBandits` / `encryptedContextualBandits` on the feature API response, plumbed through `GrowthBook`, `GrowthBookClient`, and the global eval context. `decryptPayload` now decrypts `encryptedContextualBandits`.
- **CB rule handling in `core.ts`:** experiment rules may carry a `contextualBanditRef` (pointing into the top-level `contextualBandits` map) and their variations under `contextualVariations` instead of `variations`. At eval time the SDK resolves the matching leaf by condition, applies that leaf's `weights`, and records `leafId` / `variationWeights` / `banditVersion` on the result.
- **Graceful fallback:** if the ref is missing, no leaf matches, or leaf selection throws, the SDK falls back to the experiment's existing/equal weights (fallback `leafId = -1`) rather than skipping or erroring.
- **New types:** `ContextualBanditDefinition(s)` and `CBContext`, plus optional `contextualBandit` on `Experiment` and `leafId` / `variationWeights` / `banditVersion` on `Result`.
- **Sticky bucketing:** CB rules now register their hash/fallback attributes (they were previously missed because they key off `variations`).

**Backwards compatibility:**

- All new fields on existing types are optional — existing payloads evaluate exactly as before.
- CB rules deliberately store variations under `contextualVariations` (not `variations`), so **older SDKs skip these rules** instead of mis-bucketing users into an even split.
- Full `cases.json` spec suite passes unchanged.

### Dependencies

None. This PR is standalone and can merge on its own — the back-end does not import any new sdk-js exports (it uses its own `ContextualBanditDefinitions` type from `shared/types/sdk`), and nothing consumes the CB payload until the back-end/versioning PR ships the `1.7.0 → contextualBandits` capability. No merge-order requirement.

### Testing

- `packages/sdk-js`: `npx jest` — **752 tests pass** across all 17 suites.
  - New `test/contextual-bandit.test.ts` covers leaf matching, fallback weights, ref-not-found, encrypted payloads, and sticky bucketing for CB rules.
  - New `contextualBandit` cases added to `test/cases.json` and run via the shared `test/json.test.ts` spec runner; all pre-existing `run` / `feature` / `evalCondition` cases still pass.
- `npx tsc --noEmit` compiles clean.

### Screenshots

N/A — SDK library changes only, no UI.

### DIFF
--- a/packages/sdk-js/test/cases.json
+++ b/packages/sdk-js/test/cases.json
@@ -1,5 +1,5 @@
 {
-  "specVersion": "0.7.1",
+  "specVersion": "0.8.0",
   "evalCondition": [
     [
       "$not - pass",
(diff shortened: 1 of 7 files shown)


packages/sdk-js/test/cases.json
```diff
@@ -6288,979 +6823,4003 @@
...
         "features": {
-          "exp1": {
-            "defaultValue": "control",
+          "promo": {
+            "defaultValue": "default",
             "rules": [
               {
-                "key": "feature-exp",
-                "seed": "feature-exp",
+                "key": "promo_bandit",
```
bryce-fitzsimons: nit: suggest generalizing the key/seed to bit a bit more general/permanent (not "promo...") since this config will get shipped around to all SDKs to implement.

Not a blocker, take it or leave it

bryce-fitzsimons (approved):
Please double check the Greptile finding and see whether the reporting side (trackingCallback) reports the wrong experiment weights. Otherwise, looks good.

PR #6484, opened 2026-07-28
Fall back to attributes JSON for built-in identifier columns in managed warehouse SQL

Kevin-Chant:
### Features and Changes

Events from integrations that predate the tracking plugin carry their identifiers only inside the `attributes` JSON — the plugin-fed `user_id`/`device_id` columns are empty, so exposure queries match nothing. Legacy materialized columns used to extract these values, but the JSON migration intentionally dropped builtin-named identifier mappings, silently breaking experiment results for such orgs (one affected customer org has ~9.8M exposure rows with `attributes.id` populated and both identifier columns empty).

Generated SQL (`ch_events` fact table, exposure queries, Product Analytics `data_source` stubs) now emits `SELECT * REPLACE (...)`, coalescing each built-in identifier column with its folded attribute paths in the SDK plugin's own precedence: `user_id <- attributes.user_id`; `device_id <- attributes.device_id, anonymous_id, id`. Plugin-era rows are unaffected — populated columns win, and the plugin strips folded keys from the JSON, so rows can't double-resolve. Query-time only: no DDL, no backfill, no license-server change.

Bumps `MANAGED_WAREHOUSE_JSON_ERGONOMICS_VERSION` to 2 so the ergonomics sweep re-persists the regenerated SQL fleet-wide (requires the `managed-warehouse-json-ergonomics-sweep` flag, same as v1).

### Testing

Verified against the dev managed warehouse (ClickHouse 26.4) with the real builder output:

- [x] synthetic customer-shaped row (empty columns, `{"id":"cust-9007199254740993"}`) -> `device_id` resolves from the JSON with exact digits
- [x] plugin-era row (columns populated) -> column values win, JSON ignored
- [x] `anonymous_id` and `id` both in JSON -> `anonymous_id` wins (plugin precedence)
- [x] empty-string JSON values fall through (`{"device_id":"","id":"x"}` -> `x`), matching the plugin's falsy `||` chain; an all-empty row resolves NULL, not `""`
- [x] `user_id` stays NULL when neither the column nor `attributes.user_id` has a value (no cross-leak from `id`)
- [x] fact-table SQL, all generated exposure queries, and the PA `data_source` stub execute unmodified

Affected orgs' experiments hashing on `id` should use the `device_id` assignment query (the canonical post-migration mapping for that attribute); with this change that query returns their full history.

### DIFF
--- a/packages/shared/src/util/managedWarehouse.ts
+++ b/packages/shared/src/util/managedWarehouse.ts
@@ -552,10 +552,12 @@ function migratedColumnSelectExpr(col: MaterializedColumn): string {
 // keys in the same precedence, so both eras resolve (folded rows can't
 // double-resolve: their JSON no longer contains the keys).
 function builtinIdentifierReplaceClause(): string {
+  // nullIf-wrapped so empty strings fall through, matching the plugin's falsy
+  // `||` chain — a JSON `device_id: ""` must not shadow a populated `id`.
   const path = (key: string) =>
-    `${MANAGED_WAREHOUSE_ATTRIBUTES_COLUMN}.${chIdentifier(
+    `nullIf(${MANAGED_WAREHOUSE_ATTRIBUTES_COLUMN}.${chIdentifier(
       key,
-    )}::Nullable(String)`;
+    )}::Nullable(String), '')`;
   return ` REPLACE (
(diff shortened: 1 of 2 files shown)


Kevin-Chant:
@cursoragent review

tzjames (approved):

PR #6488, opened 2026-07-29
feat(experiments): persist snapshot runner kind

ahdriel:
### Features and Changes

- Add `runnerKind` the snapshot document in mongo so we record how each snapshot was actually generated
- Split the concept in two: an internal `SnapshotQueryRunnerFamily` (used to pick the query runner) and a persisted `SnapshotQueryRunnerKind` which details how the update happened

### Testing

- Unit tests


ahdriel:
cursor review

lukesonnet (approved):

PR #6489, opened 2026-07-29
fix(features): Relax diagnostics query permissions check & surface errors

msamper:
## Summary

Customers hitting query failures on the Feature Diagnostics tab only saw a generic **"Error: There was an error"** instead of the actual SQL/query failure reason.

**Root cause:** `POST /query/feature-eval-diagnostic` returned HTTP 400 with `{ error: e.message }`, but front-end `apiCall` only reads `responseData.message` and falls back to `"There was an error"`.

**Fix:** Align with the exposures debugger path:
- `runFeatureEvalDiagnosticsQuery` now returns `{ error, sql }` on query failure (via `formatQueryExecutionErrorForApi`) instead of throwing `SQLExecutionError`
- Controller returns HTTP 200 with `error` / `sql` so the Diagnostics tab can display the real message
- Frontend short-circuits on `results.error` so failures don’t also show “No feature evaluations found”

We're also relaxing the permissions check for running diagnostics query so that users with `manageFeature` permissions can now run the query. Previously running the query required `runQueries` permissions which is used for experiment queries.

## Test plan

- [ ] Open a feature → Diagnostics tab with a datasource that has a feature usage query
- [ ] Force a query failure (e.g. invalid feature usage SQL / missing table)
- [ ] Confirm the Callout shows the underlying warehouse/SQL error text, not a generic “There was an error”
- [ ] Confirm a successful empty result still shows “No feature evaluations found”
- [ ] Confirm a failed refresh after a prior success keeps previous rows and shows the error

**Validated:** `pnpm --filter back-end type-check`, `pnpm --filter front-end type-check`, ESLint on changed files

## Screenshots

Screenshot 2026-07-28 at 10 29 25 PM


packages/back-end/src/services/datasource.ts
```diff
@@ -308,7 +309,10 @@ export async function runFeatureEvalDiagnosticsQuery(
       sql,
     };
   } catch (e) {
-    throw new SQLExecutionError(e.message, sql);
+    return {
```
bryce-fitzsimons: I'd undo this change

packages/front-end/components/Features/FeatureDiagnostics.tsx
```diff
@@ -195,6 +195,10 @@ export default function FeatureDiagnostics({
           }),
         },
       );
+      if (results.error) {
+        setError(results.error);
```
bryce-fitzsimons: looks like some errors will also send back things like "sql" which might be useful to wire to some sort of drilldown component

bryce-fitzsimons (approved):

PR #6491, opened 2026-07-29
Fix event forwarder tracking for GrowthBookClient

bryce-fitzsimons:
Feature evaluations and experiment exposures from `GrowthBookClient` were never sent to the event forwarder. Surfaced by a customer, who saw feature usage for flags evaluated by `GrowthBook` in the browser but nothing for the flag behind an experiment evaluated by `GrowthBookClient`.

`GrowthBookClient` never handed the tracking plugin's logger to its evaluation context, so Experiment Viewed and Feature Evaluated never fired. Explicit `logEvent()` calls still worked, which hid it. Broken since automatic events shipped.

Fixing that exposed a second issue: one `GrowthBookClient` is shared by every user, and Feature Evaluated carries no identity, so the first user to evaluate a flag suppressed everyone else's identical evaluation. Feature Evaluated now de-dupes per user. Exposures still de-dupe per bucketing unit, so an account-level experiment stays at one exposure per account.

Also adds the `setFeatureUsageCallback` method `UserScopedGrowthBook` was missing.

`GrowthBook` behavior is unchanged throughout.

Known gap: `growthbookTrackingPlugin` is a no-op when passed only to `createScopedInstance()`, since `UserScopedGrowthBook` has no `setEventLogger`. Separately, a plugin passed to the `GrowthBookClient` constructor is re-applied to every scoped instance, so one that registers a tracking or feature usage callback fires twice per event. Both need their own PR.


madhuchavva (dismissed):

packages/sdk-js/src/plugins/growthbook-tracking.ts
```diff
@@ -266,6 +266,15 @@ export function growthbookTrackingPlugin({
...
             properties,
           };
+          // Feature Evaluated has no unit identity, unlike Experiment Viewed
+          if (eventName === EVENT_FEATURE_EVALUATED) {
+            dedupeKeyData.identity = [
+              data.attributes.user_id,
+              data.attributes.device_id,
+              data.attributes.anonymous_id,
+              data.attributes.id,
+            ];
```
jdorn: Hardcoding identifier attribute keys does not feel like the right solution.

bryce-fitzsimons: good point. however there are no obviously good fallbacks when dedupeKeyAttributes is missing, so taking a naive approach instead...

madhuchavva (approved):

PR #6496, opened 2026-07-29
Adds signup check

mknowlton89:
### Features and Changes


tzjames (changes requested):
Let's have the regex itself come from an env var, so we can change it on the fly.  We will need to add the env var regex to terraform and land it first, for this to take effect.

tzjames (approved):

Auz (approved):

PR #6498, opened 2026-07-29
feat(fact-tables): preserve warehouse-reported column types

ahdriel:
### Features and Changes

- We now have `dataTypeFromWarehouse` for fact table columns
- Users (and our json string handling) are still allowed to override `dataType`

### Testing

- Unit tests

### DIFF
--- a/packages/shared/src/validators/fact-table.ts
+++ b/packages/shared/src/validators/fact-table.ts
@@ -350,16 +350,13 @@ export const apiFactTableColumnValidator = namedSchema(
       column: z
         .string()
         .describe("The actual column name in the database/SQL query"),
-      datatype: z.enum([
-        "number",
-        "string",
-        "date",
-        "boolean",
-        "json",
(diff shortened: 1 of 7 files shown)


packages/shared/src/validators/fact-table.ts
```diff
@@ -350,16 +350,13 @@ export const apiFactTableColumnValidator = namedSchema(
...
-        "binary",
-        "other",
-        "",
-      ]),
+      datatype: factTableColumnTypeValidator.describe(
+        "The column's data type (can be an override of the warehouse-reported datatype, for JSON string columns for example).",
+      ),
+      dataTypeFromWarehouse: factTableColumnTypeValidator
+        .describe("The warehouse-reported datatype.")
+        .readonly()
```
lukesonnet, line 358: TIL, nice

lukesonnet (approved):

PR #6500, opened 2026-07-29
feat(event-forwarder): Include Properties in Event Forwarder Feature Usage Query

msamper:
### Features and Changes

Surfaces Feature Evaluated event metadata in the **Feature Diagnostics** tab for event forwarder users. The managed feature usage query now selects `environment` (a top-level column) plus `value`, `source`, `rule_id`, and `variation_id` (extracted from the Avro `properties` map). These columns flow through automatically to the diagnostics table, so no front-end changes were needed.

### Dependencies

N/A

### Testing

- [x] Unit tests for the new columns and `buildEventForwarderPropertyValueSql` 
- [x] Manually run the generated query against a real Snowflake `*_feature_usage` table and confirm `value`/`source`/`rule_id`/`variation_id`/`environment` populate
- [x] Verify the columns render in the Feature Diagnostics tab

### Screenshots

N/A

### DIFF
--- a/packages/back-end/src/services/eventForwarder/bigquery.ts
+++ b/packages/back-end/src/services/eventForwarder/bigquery.ts
@@ -53,6 +53,7 @@ const EXPERIMENT_VIEWED_SCHEMA: bq.TableField[] = [
 const FEATURE_USAGE_SCHEMA: bq.TableField[] = [
   ...BASE_FIELDS,
   { name: "feature_key", type: "STRING", mode: "NULLABLE" },
+  { name: "properties", type: "JSON", mode: "NULLABLE" },
   { name: "attributes", type: "JSON", mode: "NULLABLE" },
 ];
 
--- a/packages/shared/src/util/event-forwarder-warehouse-queries.ts
+++ b/packages/shared/src/util/event-forwarder-warehouse-queries.ts
@@ -13,6 +13,7 @@ import {
 import {
   buildBigQueryEventForwarderTableReference,
   buildEventForwarderNestedAttributeValueSql,
+  buildEventForwarderPropertyValueSql,
   buildSnowflakeEventForwarderTableReference,
   EVENT_FORWARDER_AVRO_PARTITION_FIELD,
   quoteBigQueryIdentifier,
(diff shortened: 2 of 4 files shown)


packages/back-end/src/services/eventForwarder/bigquery.ts
```diff
@@ -53,6 +53,7 @@ const EXPERIMENT_VIEWED_SCHEMA: bq.TableField[] = [
 const FEATURE_USAGE_SCHEMA: bq.TableField[] = [
   ...BASE_FIELDS,
   { name: "feature_key", type: "STRING", mode: "NULLABLE" },
+  { name: "properties", type: "JSON", mode: "NULLABLE" },
```
msamper, line 56: Can remove this if it does not belong here

packages/shared/src/util/event-forwarder-warehouse-queries.ts
```diff
@@ -330,14 +331,24 @@ export function buildEventForwarderFeatureUsageQuerySql({
   if (sinkType === "bigquery") {
     return `SELECT
   timestamp AS timestamp,
-  feature_key AS feature_key
+  feature_key AS feature_key,
+  environment AS environment,
+  ${buildEventForwarderPropertyValueSql({ sinkType, propertyKey: "value" })} AS value,
+  ${buildEventForwarderPropertyValueSql({ sinkType, propertyKey: "source" })} AS source,
+  ${buildEventForwarderPropertyValueSql({ sinkType, propertyKey: "ruleId" })} AS rule_id,
+  ${buildEventForwarderPropertyValueSql({ sinkType, propertyKey: "variationId" })} AS variation_id
```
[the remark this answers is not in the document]
msamper: There are no Big Query event forwarder data sources provisioned right now for any customers so this should not be an issue

packages/back-end/src/services/eventForwarder/bigquery.ts
nhat-growthbook: This is right

nhat-growthbook (approved):

packages/back-end/src/services/eventForwarder/bigquery.ts
msamper: Should we add to `BASE_FIELDS` instead since all the tables should have `properties`?

PR #6507, opened 2026-07-29
fix(holdouts): derive experiment linkage from published rules

bryce-fitzsimons:
Adding an experiment rule to a draft immediately enrolled the experiment in the feature's holdout, on both the experiment and the holdout. Editing the rule away or discarding the draft left that enrollment behind, so a holdout kept an experiment — and stayed in the SDK payload — for a rule that was never published and no longer exists. Publishing is now what changes linkage, so an abandoned draft leaves nothing behind.

## What changes

- **Linkage is derived from published rules.** It is reconciled on every publish, including a rules-only change, which is how a rule added to an already-held feature gets enrolled.
- **Unlinking is bounded by what the feature itself contributed** and skips anything another feature in the holdout still references. A holdout's experiment list is not only a projection of feature rules — experiments can be added to a holdout directly — so publishing one feature can never withdraw someone else's membership.
- **Leaving a holdout drops the experiments the feature brought.** Previously the feature was unlinked but its experiments stayed enrolled, because the guard that should force detaching them first reads post-publish rules and a draft can remove the holdout and the rule together.
- **Re-adding an experiment rule after discarding a draft works again.** It was refused with "the experiment already has linked Feature Flags — unlink them first", naming a link to the feature being edited, which the UI gives no way to remove.
- **One owner for experiment linkage.** The holdout transition keeps the feature side; the publish reconcile handles joining, leaving, and rule changes alike. Bulk publish plans and compensates it the same way.
- **A failed publish rewinds further.** The steps that decide what is live — the feature document and revision status — no longer get abandoned because a satellite like ramp schedules or holdout linkage could not be reversed.

Also fixes a crash this surfaced: `statusUpdateSchedule` is a nested Mongoose path, so an experiment with no schedule hydrates it as `{}` rather than undefined, and serializing the experiment for its update event threw on the missing `startAt`. That failed any publish which wrote an experiment.

## Testing

Verified against a real dataset — draft add, discard, re-add, publish, rule removal, and revert — checking the holdout and experiment documents at each step, with membership counts confirming nothing collateral moved. Unit tests cover the linkage delta, including the multi-feature and added-directly cases.

### DIFF
--- a/packages/back-end/src/models/FeatureModel.ts
+++ b/packages/back-end/src/models/FeatureModel.ts
@@ -2169,6 +2169,8 @@ async function planLinkageForHoldout(
   if (candidates.length) {
     // Only other features' LIVE rules count: an unpublished draft elsewhere has
     // no linkage of its own to protect, which is what keeps this decidable.
+    // TODO: consider querying with a rules[] projection if performance becomes
+    // an issue — only `rules` and `holdout` are read here.
     const otherFeatures = await getFeaturesByIds(
       context,
       Object.keys(holdout.linkedFeatures).filter((id) => id !== feature.id),
--- a/packages/back-end/src/models/FeatureModel.ts
+++ b/packages/back-end/src/models/FeatureModel.ts
@@ -2080,59 +2080,39 @@ export async function applyHoldoutSideEffects(
     );
   }
 
-  // Feature side only. The experiments in its rules are reconciled from
-  // published state by planHoldoutExperimentLinkage, which owns that half for
-  // every direction (join, leave, and a rules-only change).
+  // Feature side only; planHoldoutExperimentLinkage owns the experiment half.
   if (newHoldoutId) {
     await context.models.holdout.addFeatureToHoldout(newHoldoutId, feature.id);
   }
 }
 
-/**
- * The `linkedExperiments` writes a publish is about to make for one feature,
- * plus the experiment `holdoutId`s they overwrite. Read-only to compute, so the
- * rewind can be registered before any of it is applied.
- */
 export type HoldoutExperimentLinkagePlan = {
   holdoutId: string;
   toLink: string[];
   toUnlink: string[];
-  // Pre-publish values, so the reversal converges to them instead of guessing.
   // "" is the clear sentinel `updateExperiment` expects.
   prevExperimentHoldoutIds: Record<string, string>;
 };
 
-/**
- * Work out how a holdout's experiment linkage should change given what this
- * feature is publishing.
- *
- * Linkage is derived from published rules rather than written when a rule is
- * added to a draft, so a draft that is edited or discarded leaves nothing to
- * unwind. This runs for a rules-only publish too — that is how an
- * experiment-ref rule added to an already-held feature gets enrolled.
- */
+// Derived from published rules, so an edited or discarded draft leaves nothing
+// to unwind.
 export async function planHoldoutExperimentLinkage(
   context: ReqContext | ApiReqContext,
   feature: FeatureInterface,
-  // Post-publish holdout and rules for this feature.
   holdoutId: string | null,
   publishedRules: FeatureRule[],
 ): Promise<HoldoutExperimentLinkagePlan[]> {
   const plans: HoldoutExperimentLinkagePlan[] = [];
 
-  // Leaving a holdout (or moving to another) has to drop what this feature
-  // contributed to the old one. Nothing else does: the transition path unlinks
-  // only the feature, and the guard that would force detaching experiments first
-  // reads post-publish rules — so a draft that removes the holdout and the rule
-  // together sails past it.
+  // Nothing else drops what this feature contributed to the holdout it is
+  // leaving: the transition unlinks only the feature, and the guard that would
+  // force detaching experiments first reads post-publish rules.
   const prevHoldoutId = feature.holdout?.id ?? null;
   if (prevHoldoutId && prevHoldoutId !== holdoutId) {
     const leaving = await planLinkageForHoldout(
       context,
       feature,
       prevHoldoutId,
-      // No rules under the old holdout any more, so everything this feature
-      // brought is a candidate to unlink.
       [],
       feature.rules ?? [],
     );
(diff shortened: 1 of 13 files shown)


packages/back-end/src/models/FeatureModel.ts
```diff
@@ -2075,37 +2080,192 @@ export async function applyHoldoutSideEffects(
...
+  // split the rewind exists to prevent — so it is logged rather than thrown, and
+  // the rest of the rewind proceeds.
+  try {
+    await setExperimentHoldoutIds(context, plan.prevExperimentHoldoutIds);
+  } catch (e) {
+    logger.error(
+      e,
+      `Failed to restore experiment holdoutIds while rewinding publish for holdout ${plan.holdoutId}`,
+    );
   }
```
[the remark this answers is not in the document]
bryce-fitzsimons: Valid — fixed in fdc0233161. The internal catch is gone, so reverseHoldoutExperimentLinkage now throws on a failed experiment write and each caller applies its own policy: bulk compensation records it in reversalFailures and reports the item stuck (never a clean rollback), while publishRevision classifies holdout linkage as a satellite and carries on to restore the feature document rather than stranding it published on a draft revision.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -3343,15 +3448,23 @@ export async function publishRevision({
...
         await undo();
       } catch (rewindErr) {
+        if (critical) criticalFailed = true;
         logger.error(
           rewindErr,
-          `Failed to rewind ${what} for feature ${feature.id} revision ${revision.version} after a failed publish — stopping rewind, everything not yet rewound stays at the published state`,
+          `Failed to rewind ${what} for feature ${feature.id} revision ${revision.version} after a failed publish${critical ? " — the feature stays at the published state" : " (satellite; continuing)"}`,
         );
-        break;
       }
```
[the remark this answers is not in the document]
bryce-fitzsimons: Partly addressed in 412b665776 — the visibility half. I kept the policy, deliberately: stopping instead of continuing gives you the stale linkage AND the feature published on a revision still marked draft, so the stranded state is added rather than avoided. Continuing at least leaves the feature and its revision mutually consistent at the pre-publish state.

What was missing is what you point at: the caller got only the original error. The rewind now collects the steps it could not reverse and appends them to the thrown message, so the response names the residue instead of it living only in the logs — the single-publish counterpart to bulk's reversalFailures. Note this is the reverse trade-off from the other thread (3678837494): there the fix was to stop swallowing so bulk reports the item stuck; the shared helper throws, and each caller applies its own policy.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -2075,38 +2080,158 @@ export async function applyHoldoutSideEffects(
...
+): Promise<HoldoutExperimentLinkagePlan | null> {
+  const holdout = await context.models.holdout.getByIdForLinkage(holdoutId);
+  if (!holdout) return null;
+
+  // Only other features' LIVE rules count: an unpublished draft elsewhere has no
+  // linkage of its own to protect, which is what keeps this decidable.
+  const otherFeatureIds = Object.keys(holdout.linkedFeatures).filter(
+    (id) => id !== feature.id,
+  );
+  const otherFeatures = await getFeaturesByIds(context, otherFeatureIds);
```
lukesonnet: This could grow large (hundreds, thousands?), do we care?

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -2075,38 +2080,158 @@ export async function applyHoldoutSideEffects(
...
+  const experimentIdsReferencedElsewhere = otherFeatures
+    .filter((f) => f.holdout?.id === holdoutId)
+    .flatMap((f) => getExperimentIdsFromRules(f.rules ?? []));
+
+  const { toLink, toUnlink } = computeHoldoutExperimentLinkageDelta({
+    publishedRules,
+    previousRules,
```
lukesonnet: nit; This felt weird when reading this method being called in `planHoldoutExperimentLinkage`. It feels like it should be a "remove" method and an "add" method, where you don't have to pass in an empty array.

Something like this?

```
async function planLinkageForHoldout(
  context,
  feature,
  holdoutId,
  args:
    | { mode: "reconcile"; publishedRules: FeatureRule[]; previousRules: FeatureRule[] }
    | { mode: "withdraw"; previousRules: FeatureRule[] },
)
```

But not a big deal.

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -2075,38 +2080,158 @@ export async function applyHoldoutSideEffects(
...
+    plan.toUnlink,
+  );
+  await context.models.holdout.removeExperimentsFromHoldout(
+    plan.holdoutId,
+    plan.toLink,
+  );
+  // Throws on failure: publishRevision treats this as a satellite and carries on
+  // to the feature document, while bulk compensation records it and reports the
+  // item stuck rather than cleanly rolled back.
+  await setExperimentHoldoutIds(context, plan.prevExperimentHoldoutIds);
```
lukesonnet, line 2253: domo arigato mr roboto

packages/back-end/src/models/FeatureModel.ts
```diff
@@ -2075,38 +2080,158 @@ export async function applyHoldoutSideEffects(
...
+  );
+  await setExperimentHoldoutIds(context, {
+    ...Object.fromEntries(plan.toLink.map((id) => [id, plan.holdoutId])),
+    ...Object.fromEntries(plan.toUnlink.map((id) => [id, ""])),
+  });
+}
+
+// Converges to the pre-image rather than inverting each write, so a forward pass
+// that failed partway still lands on the pre-publish state.
+export async function reverseHoldoutExperimentLinkage(
```
lukesonnet, line 2238: Both this approach *and* an approach where we just store the whole pre-image and restore it (instead of un-unlinking and un-linking) both suffer from the potential for race conditions, but I think this approach is a little bit safer since it won't clobber irrelevant changes.

Nice!

lukesonnet (approved):
Left some nits.

Overall seems reasonable, tested the happy path in the UI and it works as expected.

Thanks Bryce!

packages/back-end/src/models/FeatureModel.ts
bryce-fitzsimons: yes, good catch. doing some optimizations to only do this fetch when it matters, and based on a subset of candidate IDs.

PR #6511, opened 2026-07-30
Skip definitions-version bump on no-op fact table / metric updates

tzjames:
### Features and Changes

Follow-up to #6341 (bump-on-write ETag for `/organization/definitions`).

`updateFactTable` and `updateFactFilter` bumped the org's definitions version unconditionally on every update, and `updateMetric` decided whether to bump based on which keys were *submitted* rather than whether their values actually *changed*. Since the front-end resubmits the whole form on every save (and some API clients re-PUT the same fact table definition on a schedule), a no-op edit still invalidated the `/organization/definitions` ETag for every reader in the org — tanking the cache hit rate for reasons unrelated to any real content change.

Confirmed via production analytics + audit log: a `factTable.update` audit entry with byte-for-byte identical `pre`/`post` content landed inside an observed spurious-cache-miss window (control-arm request confirmed nothing had actually changed org-wide; the treatment-arm request right before it still got a needless cache miss).

- `updateFactTable`: skip `touchDefinitionsVersion` when none of the submitted `changes` differ from the stored fact table.
- `updateFactFilter`: same guard, comparing against the filter being replaced.
- `updateMetric`: replaced the presence-based check (`"dateUpdated" in updates`) with a value-based `isEqual` comparison against the stored metric.

All three mirror the existing `isEqual`-based guard already used by `updateFactTableColumns` for the same reason (that one was guarded because it runs from a background cron; these run from the interactive/API save path, which turns out to matter just as much).

### Testing

- [x] `pnpm --filter back-end test -- test/models/factTableModel.test.ts test/models/MetricModel.test.ts` — 31/31 existing tests pass unmodified
- [x] Lint clean
- [ ] No new tests added for the no-op-update case itself — happy to add if wanted

### DIFF
--- a/packages/back-end/src/models/FactTableModel.ts
+++ b/packages/back-end/src/models/FactTableModel.ts
@@ -408,6 +408,15 @@ export async function updateFactTable(
   );
 
   await audit.logUpdate(context, factTable, { ...factTable, ...changes });
+
+  // Skip the bump when nothing actually changed (e.g. a no-op save, or a
+  // client re-PUTting the same definition) — an unconditional touch here
+  // tanks the ETag hit rate for no reason. Mirrors updateFactTableColumns.
+  const changedDefinitionFields = Object.entries(changes).some(
+    ([k, v]) => !isEqual(factTable[k as keyof FactTableInterface], v),
+  );
+  if (!changedDefinitionFields) return;
+
   await touchDefinitionsVersion(
     factTable.organization,
     definitionsScope(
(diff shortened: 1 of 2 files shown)


tzjames:
Both issues confirmed and fixed in 3b04921:

**Issue 1** — `updateFactTable`/`updateFactFilter` now diff against a pre-write snapshot captured atomically via `findOneAndUpdate({ new: false })` instead of the (potentially stale) `factTable` argument, closing the race.

**Issue 2** — confirmed via `LegacyMetricAnalysisQueryRunner.updateModel`, which calls `updateMetric` directly with `{queries, runStarted, analysis, analysisError}` on every analysis run. Restored the `FIELDS_NOT_REQUIRING_DATE_UPDATED` exclusion alongside the value comparison so routine analysis completion no longer bumps the definitions version.

Type-check, lint, and the existing `factTableModel`/`MetricModel` test suites (31/31) all still pass.

tzjames:
Good catch — the metric path was still diffing against the caller-supplied \`metric\` argument. Fixed in 51d93d5: \`updateMetric\` now uses \`findOneAndUpdate({ new: false })\` for both the config.yml/upsert and normal write paths, diffing against the atomically-captured pre-write state instead, matching the fact-table fix. Type-check, lint, and the existing test suites (31/31) still pass.

Kevin-Chant:
@greptile Real mechanism, but intentional. The guard only skips when the submitted value
equals what that caller itself read — so the dropped request is echoing a value,
not making an edit — and the concurrent write that won went through the same
guard, so it bumped the version and readers stay on a correct ETag.

The alternative we tried first (always write, diff an atomic pre-image) fails the
other way: `updateFactFilter` `$set`s the whole `filters` array from the caller's
snapshot, so a concurrent edit to a different filter got reverted with no bump,
stranding the whole org on a stale ETag. We'd rather drop a no-op than miss an
invalidation.

Kevin-Chant (approved):
Approving, but this version is mostly my code so you should also confirm you agree @tzjames. I would also approve your initial approach but it had some heavier queries and my review bots (rightly imo) concluded that this approach would be better

PR #6513, opened 2026-07-30
Update docs to reflect current tracking policy

Kevin-Chant:
### Features and Changes

We realized that due to dogfooding our own SDK tracking plugin our tracking-related docs had fallen behind and were now stale. Updating to clarify the current state of logging:
* Hashed org ID (raw for cloud)
* user ID
* IP & derived geolocation, user agent
* App version

(All of which is still disabled by `DISABLE_TELEMETRY` for self-hosted)


jdorn (approved):

najito (approved):
Stamped

PR #6515, opened 2026-07-30
Per-org id attribute mapping for managed warehouse identifiers

Kevin-Chant:
### Features and Changes

Follow-up to #6484. Orgs that send a logged-in user ID under the `id` attribute key (Ingestion API integrations) see it resolve as `device_id` — mechanically correct, but their metrics are keyed on `user_id` and the label misleads experimenters.

Adds `settings.idAttributeIdentifier` (`"user_id" | "device_id"`, default `device_id`) as an explicit input to the generated-SQL derivation. It moves the `attributes.id` coalesce between the two built-in identifier chains in the `REPLACE` fallback (each key feeds exactly one identifier, never both), and hash-attribute &rarr; assignment-query folding follows it. A new `PUT /datasource/:id/managed-warehouse/id-attribute-identifier` endpoint updates the setting and re-syncs the org's exposure queries and fact-table SQL; the datasource page gets a select with a confirm dialog. The ergonomics sweep treats the flag as a freshness-guard input. Query-time only — no DDL — so flipping back fully reverts.

Note: the setting only helps Ingestion API senders. The SDK tracking plugin folds `id` into the physical `device_id` column client-side and strips it from the JSON, out of reach of query-time remapping — the UI warns on both the control and the confirm dialog.

### Testing

- [x] Shared tests (61 passing, 7 new): flagged SQL moves `attributes.id` to the `user_id` chain and out of the `device_id` chain, explicit `device_id` matches the default output byte-for-byte, fold + exposure-query resolution remap only `id`, legacy materialized-column mapping unaffected
- [x] Type-check + lint across shared / back-end / front-end
- [x] Dev e2e (managed warehouse, ClickHouse 26.4): `PUT` flip to `user_id` -> persisted EAQs and `ch_events` fact-table SQL regenerate with the remapped chains; customer-shaped row (empty identifier columns, `{"id":"cust-9007199254740993"}`) resolves through the persisted `user_id` EAQ verbatim with exact digits and `device_id` NULL (no double-resolve); flip back to `device_id` -> exposure queries, fact SQL, and userIdTypes byte-identical to the pre-flip baseline

### DIFF
--- a/packages/back-end/src/app.ts
+++ b/packages/back-end/src/app.ts
@@ -1123,6 +1123,10 @@ app.post(
   "/datasource/:datasourceId/managed-warehouse/remove-legacy-identifier",
   datasourcesController.postRemoveManagedWarehouseLegacyIdentifier,
 );
+app.put(
+  "/datasource/:datasourceId/managed-warehouse/id-attribute-identifier",
+  datasourcesController.putManagedWarehouseIdAttributeIdentifier,
+);
 
 if (IS_CLOUD) {
   app.post(
--- a/packages/back-end/src/controllers/datasources.ts
+++ b/packages/back-end/src/controllers/datasources.ts
@@ -101,7 +101,10 @@ import {
 import { DimensionSlicesQueryRunner } from "back-end/src/queryRunners/DimensionSlicesQueryRunner";
 import { logger } from "back-end/src/util/logger";
 import { IS_CLOUD } from "back-end/src/util/secrets";
-import { removeManagedWarehouseLegacyIdentifier } from "back-end/src/services/clickhouse";
+import {
+  removeManagedWarehouseLegacyIdentifier,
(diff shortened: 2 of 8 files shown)


Kevin-Chant:
@greptile The remaining concern is the full-settings write inside syncManagedWarehouseIdentifiers, shared by every existing caller on main (attribute-change sync, legacy-identifier removal, the ergonomics sweep) — not something this PR adds or worsens. This PR's own writes are race-free: the flag is a single-key $set, and the sync re-fetches its inputs. The shared fix is optimistic locking in updateDataSource (CAS on dateUpdated), tracked as a follow-up — doing it here would couple an org-wide model change to a settings toggle.

tzjames (approved):
Besides gating the front-end on whether they even have `id` as an sdk-attribute, I think this is fine.

Issue #4031, opened 2025-05-06
Feature flag updates not reflected unless app is restarted (Streaming not working)


NawafSwe:
## Overview

We're using the GrowthBook SDK (`@growthbook/growthbook`) in a Node.js application to manage feature flags. However, updates to feature flags made in the GrowthBook portal are **not** reflected in the running application **unless the app is restarted** or we **explicitly call `refreshFeatures()`**.

We expected the GrowthBook client to pick up feature updates automatically using **server-sent events (SSE)** when `streaming: true` is enabled, but that doesn't seem to be working as intended.

---

## Environment

- **Node.js version**: `20.11.0`
- **@growthbook/growthbook version**: `1.4.1`

---

## Expected Behavior

When initializing the GrowthBook client with streaming enabled, feature flag updates should be delivered to the application in real-time without requiring an app restart or manual refresh.

---

## Actual Behavior

- Updating a flag in the GrowthBook portal does **not** reflect in the app.
- Feature updates are only applied after:
  - Restarting the application, **or**
  - Manually calling `growthbookClient.refreshFeatures()`, which is expensive and not suitable for frequent polling.

---

## Setup and Reproduction

We followed GrowthBook's recommended setup for enabling streaming with Node.js using the `eventsource` polyfill:

### Install dependencies

```bash
npm install --save @growthbook/growthbook eventsource
```

# Initialize GrowthBook client
```js
const { setPolyfills, GrowthBookClient } = require("@growthbook/growthbook");

// Set polyfill for Node.js to support EventSource
setPolyfills({
  EventSource: require("eventsource"),
});

const gbClient = new GrowthBookClient({
  clientKey: "sdk-abc123", // Replace with your actual SDK key
});

// Enable streaming
await gbClient.init({ streaming: true });
```

## Observation
- Start the app.

- Update a feature flag value in the GrowthBook dashboard.

- No change is reflected in the app until either:
    - The app is restarted, or

     - gbClient.refreshFeatures() is called manually.

## Troubleshooting Attempts
-  Confirmed that eventsource is installed and imported correctly.

- No errors appear in the logs about EventSource or streaming connections.

## Questions / Request
- Is streaming supposed to work in this configuration (Node.js + eventsource)?
- Are there any additional headers, network configurations, or prerequisites we need to fulfill for SSE to function correctly?
- Are there known limitations or open issues in version 1.4.1 related to streaming?

0xnikter:
Same issue on our side:
[...]

PR #6516, opened 2026-07-30
Fix: don't silently swallow SDK streaming setup failures

Kevin-Chant:
### Features and Changes

`streaming: true` had three ways to silently do nothing, leaving the SDK serving the payload it fetched during `init()` indefinitely while reporting success (#4031).

An `EventSource` that isn't a constructor threw inside `fetchFeatures`' promise chain, so the `.catch` returned `data: null` and `init()` overwrote the payload it had just downloaded successfully — every flag fell back to its default. `enableChannel` now contains stream setup failures rather than propagating them. Separately, `startStreaming` warns once per distinct reason when streaming is enabled but no stream started: either no `EventSource` is available, or the host didn't return `x-sse-support`.

A contained failure still has to leave the `streams` map in a state that allows a retry, since the existing-key guard in `startAutoRefresh` treats any registered channel as live. That cleanup lives in `enableChannel`'s own catch rather than at the call site, because `enableChannel` has three entry points — initial setup, `onSSEError`'s backoff retry, and `onVisible` — and only the first had any cleanup.

Both sides of that retry are now identity-checked against the map, which fixes a pre-existing leak the above made visible. `onSSEError` only gated its backoff callback on `channel.state`, and a channel dropped by `clearAutoRefresh` mid-backoff is left `"disabled"` — which passes that check. So a retry that outlived a `destroy()` or `clearCache()` would reconnect successfully and open an `EventSource` that nothing tracks and nothing can ever close, since `clearAutoRefresh` only iterates `streams` and that now holds the replacement channel. The orphan leaked for the process lifetime while its callback kept writing to the shared cache and pushing payloads to the replacement's subscribers.

The docs compounded this by recommending `setPolyfills({ EventSource: require("eventsource") })`, which works only on `eventsource` v2 and below — v3+ uses a named export, so that snippet passes the module object and hits the case above. Also corrects the documented `maxAge` cache default (4 hours, not 24).

Note: Node still has no unflagged global `EventSource` — it is `Stability: 1 - Experimental` behind `--experimental-eventsource` on `main` (27.0.0-dev), so the polyfill path stays load-bearing.

### Testing

- [x] `streaming: true` with `EventSource: require("eventsource")` on v4 -> `init` succeeds, features intact, warning names the constructor error
- [x] `EventSource` whose constructor succeeds but `addEventListener` throws -> the half-built connection is closed, not just dereferenced (verified the test fails without the fix)
- [x] failed setup, then a corrected polyfill -> a later `init` establishes a stream instead of being blocked by the dead channel
- [x] established stream parked via `onHidden`, polyfill broken, then `onVisible` -> the failed re-connect is evicted, so a later `init` still establishes a stream (verified the test fails without the fix)
- [x] stream errors into backoff, `clearCache()` drops it, a new instance takes the key over -> the stale re-connect never fires, so only 2 `EventSource`s are ever constructed (verified the test fails without the fix, opening a 3rd untracked connection)
- [x] `streaming: true` with no `EventSource` -> features intact, warning points at `setPolyfills`
- [x] `streaming: true` against a host that omits `x-sse-support` -> warning names the missing header
- [x] `streaming: true` with a valid `EventSource` -> stream still connects and updates features
- [x] `pnpm --filter @growthbook/growthbook test` -> 782 pass


tzjames (approved):

packages/sdk-js/src/feature-repository.ts
```diff
@@ -190,6 +192,15 @@ export function onVisible() {
 
 // Private functions
 
+// Streaming stays silent when it can't start, so the SDK looks healthy while serving a frozen payload
+function warnStreamingUnavailable(reason: string): void {
+  if (streamingWarnings.has(reason)) return;
+  streamingWarnings.add(reason);
+  console.warn(
```
bryce-fitzsimons, line 199: this is a non-standard way of logging. We should follow something like
```
            instance.log(reason, contextObj)
```

packages/sdk-js/src/feature-repository.ts
```diff
@@ -532,18 +548,47 @@ function disableChannel(channel: ScopedChannel) {
...
+        channel.src.close();
+        channel.src.onerror = null;
+        channel.src.onopen = null;
+      }
+    } catch (closeError) {
+      // Ignore cleanup errors from incompatible implementations
+    }
+    channel.src = null;
+    channel.state = "disabled";
+    // A channel with no EventSource must not sit in the map, where the existing-key
```
bryce-fitzsimons, line 580: comments should be tightened to match the style of the sdk codebase

packages/sdk-js/src/feature-repository.ts
Kevin-Chant: It's non-standard because it's a possible production misconfiguration issue, so the `console.warn` is warranted (otherwise the users who need the warning won't get it)

PR #6517, opened 2026-07-30
Add built-in polling to the JS SDK via pollingInterval

Kevin-Chant:
### Features and Changes

Polling is already the documented alternative to streaming for long-running Node processes, but the implementation is left to the caller. This makes it an init option, matching what the Go SDK exposes as `WithPollDatasource`.

```js
await gbClient.init({ pollingInterval: 60 * 1000 });
```

It does four things a user-land `setInterval(() => gbClient.refreshFeatures(), ms)` can't. It dedupes per key, so many instances sharing a `clientKey` poll once rather than once each. It bypasses the cache, so the configured interval is the interval used — `refreshFeatures()` is silently capped by `staleTTL`, which is why a sub-60s loop currently issues no requests at all. It backs off exponentially with jitter on errors, capped at 5 minutes, sharing the `backoffDelay` helper `onSSEError` now also uses. And the timer is `unref`'d, so it never holds a serverless or CLI process open.

The interval is validated rather than passed straight to `setTimeout`, which collapses every out-of-range delay to 1ms — so both a negative typo and one past `setTimeout`'s signed 32-bit maximum would have produced a continuous request loop instead of the configured interval. Out-of-range values are rejected with a warning rather than clamped, so a typo is reported instead of silently reinterpreted.

`startStreaming` becomes `startBackgroundSync` since it now sets up either mechanism. Internal only — it was never exported from `index`.

Both backoffs go through one `backoffDelay(base, factor, attempt)`, since `onSSEError`'s delay factors exactly to that shape: `3^(e-3) * (1000 + random*1000)` is `1000 * 3^(e-3) * (1 + random)`. `onSSEError` is therefore unchanged, just factored. Polling moves from flat `+0-1s` jitter to proportional, which widens its spread — a 60s interval can reach 120s after an error — in exchange for one jitter model across the file and better behavior when many clients retry the same host at once. Only the new polling path changes.

Polling is opt-in. It runs until streaming is *confirmed* live — the `onopen` handler retires the poller — rather than until an `EventSource` is merely constructed. That distinction matters: a constructor returning says nothing about whether the connection succeeds, and SSE blocked by a firewall or proxy is exactly why someone chooses polling, so retiring the poller at construction would leave those users with no refresh at all. For the same reason polling isn't gated on an absent stream, since `startAutoRefresh` opens one opportunistically whenever the host advertises SSE.

#6516 makes an unusable stream warn; this PR deliberately does not make it auto-poll — see below.

### Known issues (open question for review)

I originally had streaming fall back to polling at `staleTTL` automatically, so that `streaming: true` without a working `EventSource` degraded instead of going stale. Two existing tests failed, both on the legacy `subscribeToChanges` / `autoRefresh` path — which means the blast radius is every existing user whose streaming is silently broken, not an edge case. They'd all start issuing a request per `staleTTL` (60s default) on upgrade.

That's arguably correct, since those users already believe they have live updates. But it's a traffic change that shouldn't ride along in a feature PR, so polling is opt-in here. Happy to add the fallback if we'd rather take it — it's a three-line change plus those two test expectations.

Also not handled: a stream that connects and then dies permanently does not resurrect polling. That case self-heals, since `onSSEError` retries with backoff capped at 5 minutes, so staleness is bounded rather than permanent. Resurrecting pollers on stream death would mean a second state machine for the better-behaved failure.

### Dependencies

Stacked on #6516 (base is `kc/sdk-streaming-silent-failures`, not `main`).

### Testing

- [x] `pollingInterval: 50` -> features update without any `refreshFeatures()` call
- [x] `pollingInterval: 20` with `staleTTL: 100` -> still refreshes, confirming the cache isn't capping the interval
- [x] `streaming: true` with a live stream + `pollingInterval` set -> no polling requests
- [x] `streaming: true` with no SSE and no `pollingInterval` -> no polling requests, warning names `pollingInterval`
- [x] instance destroyed -> polling stops
- [x] `GrowthBookClient` that started a shared poller is destroyed while a sibling remains -> polls keep going to the configured host, never `cdn.growthbook.io` with a blank key
- [x] a later response starts advertising `x-sse-support` -> stream opens and polling stops
- [x] SSE advertised but the stream never opens (no handler for `/sub/`) -> polling continues and features keep updating
- [x] `pollingInterval: -5` -> ignored with a warning instead of looping at the 1ms `setTimeout` clamp
- [x] `pollingInterval: 2147483648` -> ignored with a warning; verified the test fails without the bound, issuing 35 fetches in 80ms
- [x] `pollingInterval: 0` -> warns rather than silently disabling refreshes
- [x] two clients sharing a `clientKey`, `pollingInterval: 300` -> 3 fetches over 800ms (1 init + 2 polls), not 6
- [x] script with an active poller and no `process.exit()` -> node exits on its own in 0.8s
- [x] `pnpm --filter @growthbook/growthbook test` -> 793 pass, no existing tests modified


Kevin-Chant:
@greptile The first half is fixed — poller retirement is gated on onopen, not on channel.src being truthy. The second half is deliberate and documented under Known issues: onSSEError retries with backoff capped at 5 minutes, so a stream that dies permanently leaves staleness bounded rather than permanent. Resurrecting pollers on stream death would mean a second state machine for the better-behaved failure mode.

packages/sdk-js/src/feature-repository.ts
```diff
@@ -197,7 +208,7 @@ function warnStreamingUnavailable(reason: string): void {
   if (streamingWarnings.has(reason)) return;
   streamingWarnings.add(reason);
   console.warn(
-    `[GrowthBook] Streaming is enabled, but not active: ${reason}. Features will not be updated in the background. See https://docs.growthbook.io/lib/node#refreshing-features`,
+    `[GrowthBook] Streaming is enabled, but not active: ${reason}. Features will not be updated in the background. Set \`pollingInterval\` to refresh on a timer instead, or see https://docs.growthbook.io/lib/node#refreshing-features`,
```
bryce-fitzsimons, line 211: please follow the standard env-gated instance.log()

packages/sdk-js/src/feature-repository.ts
```diff
@@ -560,6 +571,12 @@ function enableChannel(channel: ScopedChannel) {
     channel.src.onerror = () => onSSEError(channel);
     channel.src.onopen = () => {
       channel.errors = 0;
+      // Only now is streaming confirmed working, so it can take over from polling.
```
bryce-fitzsimons, line 574: tighten up comments

packages/sdk-js/src/feature-repository.ts
```diff
@@ -596,6 +613,64 @@ function destroyChannel(channel: ScopedChannel, key: string) {
...
+  const poller: Poller = { timer: null, interval, errors: 0 };
+  pollers.set(key, poller);
+  scheduleNextPoll(key, poller);
+}
+
+function scheduleNextPoll(key: string, poller: Poller): void {
+  // Back off on repeated errors, with jitter, capped at 5 minutes
+  const delay =
+    poller.errors > 0
+      ? Math.min(
```
bryce-fitzsimons, line 637: very similar calc in `onSSEError()`. Maybe just generalize that one and import it?

tzjames (approved):

PR #6527, opened 2026-07-31
Contextual Bandit Visual Nits and Cancel API

teresayung:
### Features and Changes
A set of visual/UX polish fixes for the Contextual Bandit feature (detail page + datasource CBAQ section):
- **Detail page variations use the shared `VariationBox`** — replaced the bespoke `ContextualBanditVariations` component with the reusable `VariationBox` from `Experiment/VariationsTable` (screenshots disabled, split % preserved), and deleted the now-unused `ContextualBanditVariations.tsx`. `VariationBox` now takes an explicit `showSplit` prop (defaulting to the previous `!isBandit` behavior) so the CB split display no longer depends on spoofing `type: "standard"`.
- **Running bandits open on the Results tab** — the detail page now defaults to `results` when the bandit is running; draft/stopped still open on `Overview`. Tab choice still persists in the URL.
- **Results notice no longer flashes "no results" while loading** — the empty-snapshot notice is now gated on the hook's request state (`loading || isValidating`), so it isn't shown during the initial fetch, a manual refresh, or the polling loop where SWR retains a stale (possibly empty) response.
- **Single results-tab notice** — the empty-snapshot callout and the exploratory-stage callout no longer stack. They're merged into one adaptive info callout that covers each state and their overlap.
- **Decision metric name on the Results tab** — the Comparison section now shows the decision metric name next to the heading (interim placement pending a results-layout redesign).
- **Refresh button parity with experiments** — the Results-tab refresh/run control reuses the experiment `RunQueriesButton` behavior (fixes the lightning-bolt auto-update).
- **CBAQ heading cleanup** — removed the `[1]` count badge next to "Contextual Bandit Assignment Queries" on the datasource page (kept the standard `BETA` badge).
- **CBAQ edit/delete flyout migrated to Radix** — the per-query menu now uses `@/ui/DropdownMenu` (`DropdownMenu` + `DropdownMenuItem`), matching the Edit/Delete pattern used elsewhere (e.g. `FactFilterList`). Fixes the misaligned Edit/Delete text and moves Delete to the shared confirmation modal.
- **CBAQ expand/collapse caret uses the design system** — replaced the raw Bootstrap `<button className="btn ml-3 text-dark">` caret with the shared Radix `IconButton`, so it matches the sizing/hover of the adjacent more-menu trigger.
- **Overview modal** — removed a duplicate owner field and dropped the leftover legacy `form-group` wrapper around `SelectOwner`.

### Dependencies
None.

### Testing
1. Open a **running** contextual bandit → it should land on the **Results** tab. Open a **draft**/**stopped** one → lands on **Overview**. Switch tabs and confirm the choice persists in the URL.
2. On the Results tab, verify only **one** info notice shows in each state: no snapshot yet, exploratory stage, and exploratory + no snapshot (previously two stacked).
3. On the Results tab, refresh/run results on a bandit that already has data → confirm the "no results" notice does **not** flash during the fetch or while polling.
4. On the Results tab, confirm the **decision metric name** appears next to the **Comparison** heading.
5. On the Overview tab, confirm variation cards render via `VariationBox` (name, ID, split %) with no screenshot/upload affordances.
6. On the **datasource page → Contextual Bandit Assignment Queries**: confirm the `[1]` badge is gone and `BETA` remains. Open a query's `⋯` menu → **Edit** and **Delete** text are aligned; Edit opens the modal; Delete shows the confirmation modal and removes the query. Confirm the expand/collapse caret matches the more-menu button's sizing/hover.


teresayung:
@greptileai

teresayung:
@greptileai

packages/front-end/components/ContextualBandit/ContextualBanditDetailPage.tsx
```diff
@@ -391,7 +428,10 @@ export default function ContextualBanditDetailPage({
         </div>
       </div>
 
-      <Tabs defaultValue="overview" persistInURL={true}>
+      <Tabs
+        defaultValue={cb.status === "running" ? "results" : "overview"}
```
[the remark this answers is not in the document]
teresayung: fixed

packages/back-end/generated/spec.yaml
```diff
@@ -41189,6 +41189,46 @@ paths:
...
+        required: false
+        content:
+          application/json:
+            schema:
+              type: object
+              properties: {}
+              additionalProperties: false
+      responses:
+        '200':
+          description: Resource created
```
ahdriel: nit: is this description correct? what do we use for other endpoints? and what should this be?

`Request canceled` maybe?

packages/back-end/generated/spec.yaml
```diff
@@ -41189,6 +41189,46 @@ paths:
...
+    post:
+      operationId: cancelContextualBandit
+      summary: Cancel a running Contextual Bandit snapshot refresh
+      tags:
+        - ContextualBandits
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: ''
```
ahdriel: suggestion: can we add that this is the `id` of the contextual-bandit?

I got confused thinking this is the `id` of the snapshot/update, but this seems to not be the case.

Alternatively, should we instead take a `snapshotId` to cancel? or are we enforcing that there can only be 1 contextual bandit snapshot update at a time for a given contextual bandit, therefore cancel can only point to a single resource?

packages/back-end/src/enterprise/services/contextualBandits.ts
```diff
@@ -905,6 +905,30 @@ export async function runContextualBanditSnapshot(
   };
 }
 
+export async function cancelContextualBanditSnapshot(
```
ahdriel: suggestion: should this also be named `latest`? `cancelContextualBanditLatestRunningSnapshot`? or we might drop the `running`

packages/front-end/components/ContextualBandit/ContextualBanditDetailPage.tsx
```diff
@@ -204,6 +208,37 @@ export default function ContextualBanditDetailPage({
...
+    );
+    return (match?.weight ?? fallback) * 100;
+  };
+
+  const experimentForVariations = useMemo(
+    () =>
+      ({
+        id: cb.id,
+        status: cb.status,
+      }) as unknown as ExperimentInterfaceStringDates,
```
ahdriel: issue: we should remove this cast and fix our types if needed

meaning if we just need `experiment.id` and `experiment.status` in VariationBox then we should not have `experiment` typed as `ExperimentInterfaceStringDates` and instead we can do `Pick<ExperimentInterfaceStringDates, 'id' | 'status'>`

packages/front-end/components/ContextualBandit/ContextualBanditDetailPage.tsx
```diff
@@ -204,6 +208,37 @@ export default function ContextualBanditDetailPage({
...
+      })),
+    [cb.variations],
+  );
+  const variationPercent = (i: number): number => {
+    const fallback = numVariations > 0 ? 1 / numVariations : 0;
+    const variationId = cb.variations[i]?.id;
+    const match = cb.variationWeights?.find(
+      (w) => w.variationId === variationId,
+    );
+    return (match?.weight ?? fallback) * 100;
```
ahdriel: question: when do we realistically expect this fallback to happen?

should we render an error instead of something if this happens? or is this fallback safe and matches what the payload serves in this scenario?

packages/front-end/hooks/useContextualBandits.ts
```diff
@@ -83,12 +83,13 @@ export function useContextualBanditResults(cbId: string | undefined) {
...
-    cbId
-      ? `/api/v1/contextual-bandits/${cbId}/results`
-      : "/api/v1/contextual-bandits/__missing__/results",
-    { shouldRun: () => !!cbId },
-  );
+  const { data, error, mutate, isValidating } =
+    useApi<ContextualBanditResultsResponse>(
+      cbId
+        ? `/api/v1/contextual-bandits/${cbId}/results`
+        : "/api/v1/contextual-bandits/__missing__/results",
```
ahdriel: question: why are we using `__missing__` as the fallback? I don't see it being used anywhere.

why can't we drop the `__missing__` branch here and in the other places we have it?

packages/shared/src/validators/contextual-bandit.ts
```diff
@@ -320,6 +322,16 @@ export const apiContextualBanditRefreshReturn = z.object({
   cbeId: z.string().optional(),
 });
 
+export const apiContextualBanditCancelValidator = {
+  paramsSchema: z.strictObject({ id: z.string() }),
+  bodySchema: z.strictObject({}).optional(),
```
ahdriel: suggestion: should we use `z.never()` here as we do not accept anything in the body at the moment?

packages/front-end/components/ContextualBandit/ContextualBanditDetailPage.tsx
teresayung: The even split fallback is expected and not an error path. In draft mode and exploratory phase, the CB starts with equal weights for all variations.  This matches what the sdk does as well - so it shouldn't be an error.

packages/front-end/hooks/useContextualBandits.ts
teresayung: Ah yes youre right, dropped it

teresayung:
@greptileai

packages/shared/src/validators/contextual-bandit.ts
teresayung: done

packages/front-end/components/ContextualBandit/ContextualBanditDetailPage.tsx
teresayung: good point, done

packages/back-end/src/enterprise/services/contextualBandits.ts
teresayung: done

packages/back-end/generated/spec.yaml
teresayung: done

packages/back-end/generated/spec.yaml
teresayung: done

packages/front-end/components/ContextualBandit/ContextualBanditResultsTable.tsx
```diff
@@ -302,11 +264,21 @@ export default function ContextualBanditResultsTable({
   const showQueries =
     !!queryLatest && (status === "failed" || status === "partially-succeeded");
 
+  const isExploring = cb.stage === "explore";
+  const resultsNotice =
+    (loading || isValidating) && !hasTableData
+      ? null
+      : getResultsNotice(hasTableData, isExploring);
```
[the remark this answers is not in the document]
teresayung: fixed

teresayung:
@greptileai

ahdriel (approved):

PR #6528, opened 2026-08-03
docs: deprecate Segments, document multi-dimension workarounds, clari…

dlin-growthbook:
## Doc Updates:
deprecate Segments, document multi-dimension workarounds, clarify Edit Settings permissions

Segments have been deprecated and are being phased out. Update all references to reflect this: FAQ, Segmentation section, and glossary entry now direct users to Dimensions or a custom SQL filter in the experiment's analysis settings instead. Also document the two supported workarounds for breaking down results by more than one dimension, since there is no native multi-dimension breakdown.

Separately, clarify that the Analysis Settings section is opened via the "Edit Settings" button, and that this button is only visible to users with edit permissions on the experiment (Admin, Experimenter, Project Admin, or an equivalent custom role).

- faq.mdx: update Dimension vs Segment entry to reflect full deprecation; add new FAQ for multi-dimension breakdowns
- using/experimenting.mdx: replace Segmentation section with deprecation notice and custom SQL filter guidance
- experimentation-analysis/dimensions.mdx: add "Breaking down by more than one dimension" section (custom SQL filter vs. custom dimension SQL intersection)
- kb/glossary.mdx: mark Segment glossary entry as deprecated
- experimentation-analysis/experiment-configuration.mdx: name the "Edit Settings" button and add a permissions note linking to User Permissions
- kb/experiments/troubleshooting-experiments.mdx: reference the Edit Settings button by name with the same visibility caveat

### DIFF
--- a/docs/docs/experimentation-analysis/experiment-configuration.mdx
+++ b/docs/docs/experimentation-analysis/experiment-configuration.mdx
@@ -75,7 +75,13 @@ You can also configure targeting and traffic to your experiment's linked feature
 
 ### Analysis Settings
 
-Here is where much of your experiment analysis is configured. Many of the fields here have some text explaining how they affect your analysis. Here are a few of them in more detail:
+Here is where much of your experiment analysis is configured. On the experiment's Overview tab, this section is accessed via the **Edit Settings** button — the label shown on the button that opens the Analysis Settings form. Many of the fields here have some text explaining how they affect your analysis. Here are a few of them in more detail:
+
+:::note Permissions required
+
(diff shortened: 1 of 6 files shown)


docs/docs/experimentation-analysis/experiment-configuration.mdx
```diff
@@ -75,7 +75,13 @@ You can also configure targeting and traffic to your experiment's linked feature
 
 ### Analysis Settings
 
-Here is where much of your experiment analysis is configured. Many of the fields here have some text explaining how they affect your analysis. Here are a few of them in more detail:
+Here is where much of your experiment analysis is configured. On the experiment's Overview tab, this section is accessed via the **Edit Settings** button — the label shown on the button that opens the Analysis Settings form. Many of the fields here have some text explaining how they affect your analysis. Here are a few of them in more detail:
```
lukesonnet, line 78: It's actually via the "Edit" button on the Analysis Settings box on Overview tab, but on the "Results" tab it is the Edit Settings button at the top of the tab.

lukesonnet (commented):
In general, I would mention custom SQL filter as a sort of last result, and suggest Dimensions (on EAQ or standalone) as the primary path forward here.

lukesonnet (approved):

PR #6530, opened 2026-08-03
Prevent stale license after upgrade

mknowlton89:
### Features and Changes

- Prevent stale Starter organization state after a successful Pro subscription by forcing an authoritative license refresh.
- Add an optional forced refresh mode to `refreshOrganization`, bypassing process-local license caches while preserving existing behavior for all other callers.
- Restrict forced license refreshes to users with billing permissions.
- Apply the forced refresh to both existing-organization upgrades and initial Pro plan selection.

### Dependencies

None.

### Testing

- [x] Start a Pro subscription from an existing Starter organization and verify the Pro experience appears after closing the success modal.
- [x] Select Pro during initial organization setup and verify the organization reflects the active subscription.
- [x] Verify normal organization refreshes continue using cached license data.
- [x] Verify users without billing permissions cannot request a forced license refresh.
- [x] Run front-end and back-end type checks.

### Screenshots

None — no visual changes.

### DIFF
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -873,10 +873,8 @@ export async function getOrganization(
 
   const context = getContextFromReq(req);
   const { org, userId } = context;
-  const forceLicenseRefresh = req.query.forceLicenseRefresh === "true";
-  if (forceLicenseRefresh && !context.permissions.canManageBilling()) {
-    context.permissions.throwPermissionError();
-  }
+  const forceLicenseRefresh = req.query.forceLicenseRefresh !== undefined;
+
   const {
     invites,
     members,
--- a/packages/back-end/src/routers/organizations/organizations.controller.ts
+++ b/packages/back-end/src/routers/organizations/organizations.controller.ts
@@ -861,7 +861,7 @@ export async function putInviteRole(
 }
 
 export async function getOrganization(
-  req: AuthRequest,
+  req: AuthRequest<unknown, unknown, { forceLicenseRefresh?: string }>,
   res: Response<GetOrganizationResponse | { status: 200; organization: null }>,
 ) {
   if (!req.organization) {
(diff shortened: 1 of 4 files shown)


packages/front-end/components/Auth/SelectInitialPlan.tsx
```diff
@@ -484,7 +484,7 @@ const ProPaymentFormInner: FC<ProPaymentFormProps> = ({
               : undefined,
         }),
       });
-      await refreshOrganization();
+      await refreshOrganization({ forceLicenseRefresh: true });
```
ahdriel: suggestion: we should do the same `refreshOrganizationAfterUpgrade` approach from CloudProUpgradeModal here

so we retry refreshing the organization and don't bundle form failure with refresh failure

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -873,6 +873,10 @@ export async function getOrganization(
 
   const context = getContextFromReq(req);
   const { org, userId } = context;
+  const forceLicenseRefresh = req.query.forceLicenseRefresh === "true";
```
ahdriel: nit: can we instead use this as a truthy value?

I think that's the pattern we use in other places so we would also accept a request with `?forceLicenseRefresh` and not just `?forceLicenseRefresh=true`.

packages/back-end/src/routers/organizations/organizations.controller.ts
```diff
@@ -873,6 +873,10 @@ export async function getOrganization(
...
+  if (forceLicenseRefresh && !context.permissions.canManageBilling()) {
```
ahdriel: question: is this the right permission check?

shouldn't we allow this to proceed regardless as this refresh will happen eventually without any user intervention too?

ahdriel (approved):

packages/back-end/src/routers/organizations/organizations.controller.ts
mknowlton89: I honestly went back and forth on this. It was originally GTP 5.6's suggestion that I just ran with.

I'm going to remove the permission check entirely. We don't gate any license refreshes anywhere else in the app, so now that I've sat on it for a day, I don't think it makes sense at all.

PR #6540, opened 2026-08-03
fix(api): attribute API-created reports to the authenticated user

itsgrimetime:
## Problem

`POST /api/v1/reports` never stamps the created report's `userId`, while the internal controller does (`controllers/reports.ts` sets `userId: req.userId`). Reports created through the REST API therefore:

- render with no owner in the UI, and
- can only be edited, refreshed-from-the-UI, or deleted by admins — **including by their own creator** — because report permissions are ownership-gated (`report.userId === req.userId`, otherwise the admin-only `canSuperDeleteReport` path).

This bit us with automation that creates reports on users' behalf using user-bound credentials: the reports come out orphaned, and the only recourse is asking an admin to delete them.

Other resources already handle this — experiments fall back to `context.userId` in `postExperiment`, features use `resolveOwnerForCreate`, and BaseModel-backed resources got a blanket fallback in #5412. Reports are a legacy non-BaseModel resource that missed that sweep.

## Fix

Follows the existing owner-field convention:

- **`POST /v1/reports`** accepts an optional `owner` (`optionalOwnerInputField` — userId or email, resolved strictly via `resolveOwnerToUserId`) and otherwise falls back to the authenticated user (`req.context.userId`, populated for Personal Access Tokens). The resolved user is stamped as the report's `userId`.
- Organization secret API keys with no body `owner` still create owner-less reports (there is no user to attribute to) — behavior unchanged.
- **All report API responses** (get / list / create / refresh / put metadata / put settings) now expose `owner` + resolved `ownerEmail` via `resolveOwnerEmail(s)`, per the same convention used by experiments, dimensions, archetypes, etc.
- OpenAPI spec regenerated (`generate-openapi`).

## Verification

Manually verified end-to-end against a local dev server:

- explicit `owner` email in the POST body → resolved to the member's userId, stamped on the Mongo doc, returned as `owner`/`ownerEmail` on POST and GET responses; the creator can subsequently edit/delete the report in the UI
- unknown `owner` email/userId → 400 `"Unable to find user: …"`
- org secret key with no `owner` → owner-less report, same as before

`pnpm --filter shared type-check`, `pnpm --filter back-end type-check`, lint, and the owner-service test suite (20/20) all pass.

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -31873,6 +31873,15 @@ paths:
                 description:
                   description: Report description
                   type: string
+                owner:
+                  description: >-
+                    The userId or email address of the owner. If an email
+                    address is provided, it will be used to look up the userId
+                    of the matching organization member. If an ID is provided,
+                    it will be validated as existing in the organization. When
+                    omitted, it defaults to the user associated with the
+                    request's Personal Access Token (PAT), if one is being used.
+                  type: string
                 statsEngine:
                   description: Stats engine override
                   type: string
@@ -57566,6 +57575,16 @@ components:
             `public`.
           type: string
         experimentId:
+          type: string
+        owner:
+          description: >-
+            The userId of the report owner. Absent for reports created before
+            owner attribution existed.
+          type: string
+        ownerEmail:
+          description: >-
+            The email address of the owner, when the owner can be resolved to a
+            known user.
           type: string
         snapshotId:
           description: Snapshot ID (experiment-snapshot type only)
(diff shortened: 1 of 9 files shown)


ahdriel (approved):
Thanks for this fix! Tested and everything works exactly as described.

One potential breaking change for the future is to make `owner` required (either provided or inferred from PAT), but forbid owner-less reports from being created. Do you think this aligns with your expectation @itsgrimetime or do you still see a use for owner-less Reports? (if we go this way we should do it in a separate PR as it is a breaking change).

itsgrimetime:
> Thanks for this fix! Tested and everything works exactly as described.
> 
> One potential breaking change for the future is to make `owner` required (either provided or inferred from PAT), but forbid owner-less reports from being created. Do you think this aligns with your expectation @itsgrimetime or do you still see a use for owner-less Reports? (if we go this way we should do it in a separate PR as it is a breaking change).

nope I don't think ownerless reports make sense (for us at least)

PR #6544, opened 2026-08-04
fix(ui,experiments): prevent stray results menu dividers

ahdriel:
### Features and Changes

- Render a divider only between visible Results menu items, so a menu with just Add / remove metrics has no leading divider.

### Testing

- `pnpm --filter front-end test test/components/experiment/ResultMoreMenu.test.tsx`


tzjames (approved):

ahdriel:
cursor review

PR #6553, opened 2026-08-05
feat(experiments): Track Experiment Start & Stop events from the backend

msamper:
### Features and Changes

Anytime a user, an agent, or an agenda job starts or stops an experiment, we now emit GrowthBook telemetry events. Previously we were not capturing experiment starts or stops that were triggered outside of the UI.

Adds `trackEventForOrganization()` in the back-end GrowthBook service so background jobs can log events with org-scoped attributes (so the `accountPlan === "loading"` filter does not drop them) as well as `trackEventForContext` to include user attributes whenever they're available.

Events include `auditUser.type` as way to distinguish the `source`

### Dependencies

None.

### Testing

- [x] `pnpm --filter back-end type-check`
- [x] With `DISABLE_TELEMETRY=enable-with-debug`, let a scheduled start fire and confirm a `Experiment Started` event is logged with `source: "agenda-job"`
- [x] Same for a scheduled stop that actually stops (not notify-only), expecting `Experiment Stopped`


packages/back-end/src/jobs/updateExperimentStatus.ts
```diff
@@ -138,6 +139,11 @@ const updateSingleExperimentStatus = async (
           },
           details: auditDetailsUpdate(experimentBefore, updated),
         });
+        trackEventForOrganization(context.org, "Start experiment", {
```
lukesonnet: should we instead be tracking in the core `executeExperimentStart`? Does that handle scheduling? These change are good, but it might be nice to have a like core service emit a telemetry event like "experiment_started_at_service" that will emit on every single time an experiment moves to "running" status?

msamper: Yes! good call. This will help with tracking these kinds of events as they happen via rest API calls as well

lukesonnet (approved):
nice! Thanks for re-organizing some of the tracking logic as well!

PR #6555, opened 2026-08-05
Override cache after subscription cancellation

mknowlton89:
### Features and Changes

- Force-refresh organization license data after a subscription is scheduled for cancellation so the updated plan status appears immediately.
- Handle refresh failures separately from cancellation failures, allowing users to retry the refresh without resubmitting the cancellation.
- Report refresh failures to Sentry for visibility.
- Fast follow to this PR: **(https://github.com/growthbook/growthbook/pull/6530)**

### Dependencies

None.

### Testing

- [x] Cancel an active subscription and confirm the pending cancellation state appears.
- [x] Confirm cancellation API errors display in the cancellation modal.
- [x] Confirm organization refresh failures display a warning without retrying cancellation.
- [x] Retry a failed refresh and confirm the warning disappears after success.

### Screenshots
N/A


packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -42,6 +47,21 @@ export default function SubscriptionInfo() {
     subscription?.nextBillDate &&
     !subscription?.pendingCancelation;
 
+  const refreshOrganizationAfterCancellation = async () => {
+    try {
+      await refreshOrganization({ forceLicenseRefresh: true });
+      setOrganizationRefreshStatus("idle");
+    } catch (error) {
+      sentryCaptureException(error);
```
ahdriel: nit: Ideally we have a shared `error` handler that conditionally calls sentry or whatever other system we have setup.

not sure if we have it at the moment but would be nice to add so we don't start adding `sentryCaptureException` to lots of different places

packages/front-end/components/Settings/SubscriptionInfo.tsx
```diff
@@ -42,6 +47,21 @@ export default function SubscriptionInfo() {
     subscription?.nextBillDate &&
     !subscription?.pendingCancelation;
 
+  const refreshOrganizationAfterCancellation = async () => {
```
ahdriel: suggestion: should we make this into a `hook`? and the hook would also handle `organizationRefreshStatus` and return that to the consumer?

So the attempt to refresh organization + return error if needed is centralized in a single hook.

ahdriel (approved):

packages/front-end/components/Settings/SubscriptionInfo.tsx
mknowlton89: Agreed - I think that makes sense to do in a standalone PR.

packages/front-end/components/Settings/SubscriptionInfo.tsx
mknowlton89: Yea, good call. I've implemented the `useForceLicenseRefresh` hook

PR #6560, opened 2026-08-05
update reo tracker to enable third party

alyssamichelle:
### Features and Changes

Updates the Reo tracker script on the docs site (`docs/static/scripts/reo.js`) to match Reo’s current snippet:

- Enables `enableThirdPartyTracking: true` in `Reo.init`
- Loads the remote script with `defer` instead of `async`
- Keeps the previous snippet commented out for reference

- Closes **(n/a)**

### Dependencies

None

### Testing

- Confirm docs still load ` /scripts/reo.js` from `docusaurus.config.js`
- Load a docs page and verify the Reo script request to `https://static.reo.dev/123409b1f7b05ac/reo.js` succeeds
- In the console, confirm `Reo` initializes without errors

### Screenshots

N/A


docs/static/scripts/reo.js
```diff
@@ -1,2 +1,6 @@
+// Old Reo script (before enableThirdPartyTracking / defer)
 // eslint-disable-next-line
-!function(){var e,t,n;e="123409b1f7b05ac",t=function(){Reo.init({clientID:"123409b1f7b05ac"})},(n=document.createElement("script")).src="https://static.reo.dev/"+e+"/reo.js",n.async=!0,n.onload=t,document.head.appendChild(n)}();
\ No newline at end of file
+// !function(){var e,t,n;e="123409b1f7b05ac",t=function(){Reo.init({clientID:"123409b1f7b05ac"})},(n=document.createElement("script")).src="https://static.reo.dev/"+e+"/reo.js",n.async=!0,n.onload=t,document.head.appendChild(n)}();
```
ahdriel: question: do we need to keep the old version here? why not have only the new script?

ahdriel (approved):

docs/static/scripts/reo.js
alyssamichelle: I was keeping incase I needed to revert, but you right. It isn't needed.

PR #6562, opened 2026-08-05
ui(pagination): Update design and make it stable

ahdriel:
### Features and Changes

- Update design of the Pagination component to match the latest figma file
- Ensure the width of it does not change as user moves through pages (before it would jump around)

### Screenshots

| Before | After |
|--------|--------|
| pagination-old | pagination-new |


packages/front-end/ui/Pagination.tsx
```diff
@@ -1,9 +1,99 @@
...
+
+export type PaginationModel = {
+  pageCount: number;
+  currentPage: number;
+  items: PaginationItem[];
+};
+
+const BOUNDARY_COUNT = 1; // pages pinned at each end
+const SIBLING_COUNT = 1; // pages shown on each side of the current page
+const SLOT_COUNT = 2 * BOUNDARY_COUNT + 2 * SIBLING_COUNT + 3;
```
msamper, line 19: niceee

packages/front-end/ui/Pagination.tsx
```diff
@@ -1,9 +1,99 @@
...
+  return Array.from({ length: Math.max(end - start + 1, 0) }, (_, index) => ({
+    type: "page",
+    page: start + index,
+  }));
+}
+
+/**
+ * A 1-page gap renders that page so the slot count stays constant. A wider gap
+ * becomes an ellipsis that jumps to the midpoint of the pages it hides, so
+ * repeated clicks bisect toward that end of the range.
```
msamper: It looks like we don't want the ellipsis to ever be interactive per the design system. I did find it kind of unexpected that when I clicked on the ellipsis it jumped to another page.

packages/front-end/ui/Pagination.tsx
```diff
@@ -19,41 +109,88 @@ const Pagination: FC<PaginationProps> = ({
...
+    <nav aria-label="Pagination" className={clsx(styles.root, className)}>
+      <ul className={styles.container}>
+        <li>
+          <button
+            type="button"
+            className={clsx(styles.button, styles.arrow)}
+            aria-label="Go to previous page"
+            disabled={model.currentPage <= 1}
+            onClick={() => onPageChange(model.currentPage - 1)}
+          >
```
msamper: Are we not using our `ui/Button` to be able to style with an scss module?

msamper (changes requested):
Nice job! Tested locally and the behavior is mostly as expected. Only one callout around ellipsis being interactive that we should fix.

packages/front-end/ui/Pagination.tsx
ahdriel: Removed this functionality

msamper (approved):

PR #6563, opened 2026-08-05
feat(experiments): Add funnel metric

lillialjackson:
### Features and Changes

We are adding a new Metric type of `funnel` for pro customers. Funnel metrics are a list of ordered steps where each step is an event with its own configuration. For each steps it answer the question: 'did the user do this complete sequence?' with breakdown per-step.

Currently functionality is limited; more settings + ability to use it with dimensions, incremental refresh, and more are planned follow-ups.

### Testing

Unit Testing
Manual Testing
- Create a new Metric of Type Funnel
- Can add multiple steps to it
- I can add that metric to my experiment
- I can see the results for the metric and each individual step in the Results page

### Screenshots

Screenshot 2026-08-12 at 8 23 16 AM

Screenshot 2026-08-12 at 8 13 25 AM


packages/back-end/src/services/experiments.ts
```diff
@@ -1332,6 +1336,16 @@ export function resolveSnapshotRunner({
     return { runnerFamily: "results", incrementalFallbackReason: null };
   }
 
+  // Funnel resolution needs the full per-unit event timestamps, which the
+  // materialized caches don't retain, so there is no incremental equivalent.
+  if (hasFunnelMetric) {
```
lukesonnet: Feels almost like we should throw for stuff added after incremental was created, esp. since we're going to fast follow. WDYT @ahdriel ?

ahdriel: Yeah -- it should be handled the same way activation metrics are handled. (throw and not supported).

ahdriel: Addressed

packages/front-end/components/FactTables/FactMetricModal.tsx
```diff
@@ -1808,7 +2080,26 @@ export default function FactMetricModal({
                     form.setValue("windowSettings.delayUnit", "hours");
                   }
 
-                  form.setValue("metricType", type as FactMetricType);
+                  if (type === "funnel") {
+                    if (!funnelSettings) {
+                      const factTableId =
+                        form.getValues("numerator")?.factTableId || "";
+                      setFunnelSettings({
+                        steps: [
```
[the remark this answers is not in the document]
lukesonnet: Authors will learn to add a step, or we will get feedback otherwise.

packages/shared/src/experiments/experiments.ts
```diff
@@ -1360,6 +1410,87 @@ export function parseSliceMetricId(
...
+  metricId: string,
+): FunnelStepMetricInfo {
+  const match = metricId.match(/^(.+)\?step=(\d+)$/);
+  if (!match) {
+    return {
+      isFunnelStepMetric: false,
+      baseMetricId: metricId,
+      stepIndex: null,
+    };
+  }
```
ahdriel, line 1438: suggestion: should we just return null if it is not a step instead? then we can make `stepIndex` non-nullable

packages/back-end/src/api/fact-metrics/postFactMetric.ts
```diff
@@ -28,13 +28,22 @@ export async function getCreateMetricPropsFromBody(
     organization,
   });
 
-  const factTable = await getFactTable(body.numerator.factTableId);
+  const primaryFactTableId =
```
ahdriel, line 31: should we use the getPrimaryFactTable something something function that we have?

packages/shared/src/funnels.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const CONVERSION_WINDOW_UNIT_TO_SECONDS: Record<ConversionWindowUnit, number> =
+  {
+    minutes: 60,
+    hours: 3600,
+    days: 86400,
+    weeks: 86400 * 7,
+  };
+
+export function conversionWindowToSeconds(window: {
```
ahdriel, line 15: nit: this is not specific to funnels, should it be in a more generic place?

packages/shared/src/validators/fact-table.ts
```diff
@@ -373,9 +433,52 @@ export const factMetricValidator = z
     metricAutoSlices: z.array(z.string()).optional(),
 
     quantileSettings: quantileSettingsValidator.nullable(),
+
+    funnelSettings: funnelSettingsValidator.nullable(),
   })
   .strict();
 
+type FactMetricFields = z.infer<typeof factMetricObjectValidator>;
```
ahdriel, line 441: suggestion: should these `FactMetric` validators/types be in `fact-metrics`?

packages/shared/src/validators/fact-metrics.ts
```diff
@@ -145,6 +146,74 @@ const apiQuantileSettings = z
...
+    ordering: z
+      .enum(["sequential", "strict", "unordered"])
+      .describe("Step ordering mode. Only 'sequential' is supported in v1.")
+      .optional(),
+    concurrencyWindowSeconds: z
+      .number()
+      .int()
+      .min(0)
+      .describe(
+        "Out-of-order tolerance between adjacent steps in seconds. Defaults to 0.",
```
ahdriel, line 189: question: is this 'defaults to 0' true? if it is optional doesn't it default to `undefined`?

packages/shared/src/validators/fact-metrics.ts
```diff
@@ -529,7 +602,7 @@ const postFactMetricBody = z
     projects: z.array(z.string()).optional(),
     tags: z.array(z.string()).optional(),
     metricType: apiMetricTypeEnum,
-    numerator: postNumeratorRef,
+    numerator: postNumeratorRef.nullable().optional(),
```
ahdriel, line 605: question: do we need both? should we consolidate only on `null` or `undefined` instead?

packages/shared/src/validators/fact-metrics.ts
```diff
@@ -608,7 +722,7 @@ const updateFactMetricBody = z
     projects: z.array(z.string()).optional(),
     tags: z.array(z.string()).optional(),
     metricType: apiMetricTypeEnum.optional(),
-    numerator: postNumeratorRef.optional(),
+    numerator: postNumeratorRef.nullable().optional(),
```
ahdriel, line 725: question: same here -- do we need to add `nullable`? why is `optional` not enough?

packages/front-end/services/metrics.tsx
```diff
@@ -54,6 +55,24 @@ export function getInitialInlineFilters(
...
+// Form-state shape backing the shared metric modal. Widens metricType to include
+// "funnel" so one form can author every metric type. funnelSettings stays null
+// here: its deeply nested step/filter shape breaks react-hook-form's typed
+// field-path resolution, so the modal tracks it in dedicated state instead.
+export type CreateFactMetricFormProps = Omit<
+  CreateStandardFactMetricProps,
+  "metricType"
+> & {
+  metricType: FactMetricInterface["metricType"];
+};
```
ahdriel, line 74: I wonder if we can fix this react-hook-form issue and use `FactMetricInterface` instead.

ahdriel (approved):

lukesonnet (approved):

PR #6570, opened 2026-08-05
refactor(holdouts): extract holdout logic into reusable services

msamper:
### Features and Changes

Extracts the holdout create, schedule-normalization, stage-transition, and targeting logic out of `holdout.controller.ts` into reusable functions in `services/holdouts.ts`: `createHoldoutWithExperiment`, `normalizeHoldoutScheduleUpdates`, `setHoldoutStage`, and `applyHoldoutTargetingChanges`. The controller now delegates to them, with no behavior change. Adds the shared `holdoutStage` enum and `getHoldoutStage` helper (a Holdout's stage is derived from the companion experiment's status plus `analysisStartDate`, never stored), used by the services and by `ScheduleStatusChangeInputs`. Also refactors `updateHoldoutStatus` to use `setHoldoutStage` rather than duplicating the logic.

### Dependencies

Stacked on #6569

### Testing

- [x] Create a holdout successfully
- [x] Edit holdout metadata and ensure it persists
- [x] Link experiments and features to a holdout
- [x] Add a schedule to a holdout
- [x] Make sure schedule fires as expected
- [x] Clear a schedule from a holdout
- [x] Delete a holdout

`pnpm --filter shared test holdout-stage` (getHoldoutStage) and `pnpm --filter back-end test holdoutSchedule` (normalizeHoldoutScheduleUpdates).

### Screenshots

N/A — no user-facing change.


packages/back-end/src/services/holdouts.ts
```diff
@@ -129,10 +152,392 @@ export async function resolveHoldoutExperimentToLink({
...
+  switch (stage) {
+    case "stopped": {
+      if (phases[0]) {
+        phases[0].dateEnded = new Date();
+      }
+      if (phases[1]) {
+        phases[1].dateEnded = new Date();
+      }
+      Object.assign(changes, { phases, status: "stopped" });
+      // await validateExperimentChange({ context, experiment, changes });
```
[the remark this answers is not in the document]
msamper: Holdouts should really have their own validation hooks since they're composed of an experiment object and a holdout object so I'm likely going to remove calls to `validateExperimentChange` for holdouts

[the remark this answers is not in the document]
msamper: Yes please

packages/back-end/src/jobs/updateHoldoutStatus.ts
```diff
@@ -1,25 +1,60 @@
...
+      return "stopped";
+    default:
+      scheduledType satisfies never;
+      throw new Error(
+        `Unhandled scheduled holdout update type: ${scheduledType}`,
+      );
+  }
+}
+
+export function isScheduledTransitionApplicable(
```
lukesonnet, line 40: nit: This and the above function both seem like somewhat restatements of the new `getHoldoutStage` function. Is that right? Could we at least in this method just use the holdout stage and then check if schedule transition applicable? But it just seems like you need to compare desired stage to current stage, and have stage be the only thing holding the business logic.

packages/back-end/src/services/holdouts.ts
```diff
@@ -129,10 +152,382 @@ export async function resolveHoldoutExperimentToLink({
...
+      });
+      refreshPayload("Status changed to stopped");
+      return;
+    }
+
+    case "draft": {
+      if (!phases[0]) {
+        throw new BadRequestError("Holdout does not have a phase");
+      }
+      phases[0].dateEnded = undefined;
```
lukesonnet, line 484: Since you could get here from stopped, do we also need to unset phase 1 date ranges if they exist? Or should we keep them around in case you're flicking back and forth for some reason and only unset them when you enter running or analysis period, which require setting the right dates?

packages/back-end/src/services/holdouts.ts
```diff
@@ -129,10 +152,382 @@ export async function resolveHoldoutExperimentToLink({
...
+          nextScheduledStatusUpdate: getNextScheduledStatusUpdateForStage(
+            holdout.statusUpdateSchedule,
+            "running",
+          ),
+        });
+        refreshPayload("Status changed to running");
+        return;
+      }
+
+      if (!phases[0]) {
```
lukesonnet: Feels like this should come before the other change so we throw before updating the experiment?

packages/back-end/src/services/holdouts.ts
```diff
@@ -129,10 +152,382 @@ export async function resolveHoldoutExperimentToLink({
...
+  if (statusUpdateSchedule.startAt && experiment.status === "draft") {
+    potentialUpdates.push({
+      date: statusUpdateSchedule.startAt,
+      type: "start",
+    });
+  }
+  if (
+    statusUpdateSchedule.startAnalysisPeriodAt &&
+    experiment.status === "running" &&
+    !holdout.analysisStartDate
```
lukesonnet: Again isn't this kind of just the stage? Maybe the logic has to diverge in some places, but feels like we should enshrine and use the stage in more places, but maybe I'm missing something.

lukesonnet (approved):
Nice, left some nits about the business logic and re-using it more, but this looks great. I didn't test it.

packages/back-end/src/services/holdouts.ts
msamper: Phase 1 is already removed below. We replace the phases array with just the first phase

PR #6571, opened 2026-08-05
feat(api): add REST API for creating, reading, and updating Holdouts

msamper:
### Features and Changes

Adds a REST API for Holdouts:
- **`GET /holdouts/{id}`** — read a single Holdout
- **`GET /holdouts`** — list Holdouts, with optional `projectId`, `datasourceId`, `stage`, and `archived` filters
- **`POST /holdouts`** — create a Holdout (starts in the `draft` stage)
- **`PUT /holdouts/{id}`** — update a Holdout
- **`POST /holdouts/{id}/start`**, **`/start-analysis`**, and **`/stop`** — dedicated lifecycle endpoints for moving through the Holdout's stages

The public API merges each Holdout with its required companion experiment, hiding the internal two-document implementation from consumers. Writes are coordinated across both documents with permission checks, lifecycle validation, rollback protection for partial failures (the two documents share no transaction), and DK payload refreshes.
Permission and validation logic is factored into shared util/service functions so the REST API and the internal app API can't drift apart.

**Not included:** Holdout deletion. That endpoint is more involved given the cleanup required on linked Features and Experiments, so I'll tackle it in a follow-up PR.

**Internal API Changes**
- The internal `updateHoldout` handler now whitelists client-settable fields instead of spreading `req.body`, so server-managed fields (`experimentId`, linkage maps, `analysisStartDate`, computed schedule pointer) can no longer be written from the request body.
- The internal create endpoint now takes `CreateHoldoutInput` rather than a partial experiment/holdout body; `NewHoldoutForm` is updated to match.
- Experiment custom hooks are now skipped for `type === "holdout"`. We should add separate custom hooks for holdouts given that they're composed of two documents.

### Dependencies

Stacked on #6570 (and #6569).

### Testing

- Unit tests for the new shared Holdout utilities (stage derivation, allowed transitions, size/coverage conversion, enabled-environment resolution).
- Unit tests for the Holdout service functions covering create/update coordination, schedule normalization, stage transitions, and experiment rollback on partial failure.
- Manual Tests
- [Claude API Test Suite](https://claude.ai/code/artifact/858bc2a0-3ade-4f9f-904f-7d5aa5c75caf)

### Screenshots

N/A

### DIFF
--- a/packages/back-end/src/api/ApiModel.ts
+++ b/packages/back-end/src/api/ApiModel.ts
@@ -95,6 +95,7 @@ export type OpenApiEndpointSpec = {
   validator: RequestSchemas<z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>;
   zodReturnObject: z.ZodTypeAny;
   summary: string;
+  description?: string;
   /** Error codes this endpoint may throw, used to generate OpenAPI error response schemas. */
   possibleErrors?: readonly ApiErrorCode[];
 };
@@ -297,6 +298,7 @@ export function getOpenApiRoutesForApiConfig(
       verb,
       operationId,
       summary,
+      description,
       zodReturnObject,
       possibleErrors,
     }) => {
@@ -306,6 +308,7 @@ export function getOpenApiRoutesForApiConfig(
         path: getFullPath(apiConfig.openApiSpec.pathBase, pathFragment),
         operationId,
         summary,
+        description,
         tags: [tag],
         responseSchema: zodReturnObject,
         possibleErrors,
--- a/packages/back-end/src/api/specs/holdout.spec.ts
+++ b/packages/back-end/src/api/specs/holdout.spec.ts
@@ -28,8 +28,6 @@ export const holdoutApiSpec = {
     createBody: apiCreateHoldoutBody,
     updateBody: apiUpdateHoldoutBody,
   },
-  // Deleting a Holdout cascades into its companion experiment and unlinks every
-  // held-out Feature Flag and experiment, so it stays UI-only for now.
   crudActions: ["get", "create", "list", "update"],
   crudValidatorOverrides: {
     list: apiListHoldoutsValidator,
@@ -37,9 +35,9 @@ export const holdoutApiSpec = {
   customEndpoints: [holdoutStageEndpoint],
   crudDescriptions: {
     create:
-      "Creates a Holdout along with its companion experiment. The Holdout starts in the `draft` stage — use POST /holdouts/{id}/stage to start it.",
+      "Creates a Holdout. The Holdout starts in the `draft` stage — use POST /holdouts/{id}/stage to start it.",
     update:
-      "Updates a Holdout. Fields are written to the Holdout and its companion experiment as needed, so callers do not need to know where each field is stored. Use POST /holdouts/{id}/stage to change the stage.",
+      "Updates a Holdout. Use POST /holdouts/{id}/stage to change the stage.",
   },
   navDescription:
     "Hold a share of traffic out of all experiments to measure their combined effect.",
--- a/packages/back-end/src/api/specs/holdout.spec.ts
+++ b/packages/back-end/src/api/specs/holdout.spec.ts
@@ -1,7 +1,7 @@
 import {
   apiCreateHoldoutBody,
-  apiHoldoutStageReturn,
-  apiHoldoutStageValidator,
+  apiHoldoutActionReturn,
+  apiHoldoutActionValidator,
   apiHoldoutValidator,
   apiListHoldoutsValidator,
   apiUpdateHoldoutBody,
@@ -10,13 +10,34 @@ import { OpenApiModelSpec } from "back-end/src/api/ApiModel";
 
 /** REST API surface for Holdouts under `/api/v1/holdouts/*`. */
 
-export const holdoutStageEndpoint = {
-  pathFragment: "/:id/stage",
+export const holdoutStartEndpoint = {
+  pathFragment: "/:id/start",
   verb: "post" as const,
-  operationId: "setHoldoutStage",
-  validator: apiHoldoutStageValidator,
-  zodReturnObject: apiHoldoutStageReturn,
-  summary: "Move a Holdout to a different stage",
+  operationId: "startHoldout",
+  validator: apiHoldoutActionValidator,
+  zodReturnObject: apiHoldoutActionReturn,
+  summary: "Start a Holdout",
+  possibleErrors: ["invalid_status"] as const,
+};
+
+export const holdoutStartAnalysisEndpoint = {
+  pathFragment: "/:id/start-analysis",
+  verb: "post" as const,
+  operationId: "startHoldoutAnalysis",
+  validator: apiHoldoutActionValidator,
+  zodReturnObject: apiHoldoutActionReturn,
+  summary: "Start a Holdout analysis period",
+  possibleErrors: ["invalid_status"] as const,
+};
+
+export const holdoutStopEndpoint = {
+  pathFragment: "/:id/stop",
+  verb: "post" as const,
+  operationId: "stopHoldout",
+  validator: apiHoldoutActionValidator,
+  zodReturnObject: apiHoldoutActionReturn,
+  summary: "Stop a Holdout",
+  possibleErrors: ["invalid_status"] as const,
 };
 
 export const holdoutApiSpec = {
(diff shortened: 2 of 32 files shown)


msamper:
@greptileai

packages/back-end/src/services/holdouts.ts
```diff
@@ -295,11 +303,220 @@ export async function createHoldoutWithExperiment(
...
+    return { holdout: updatedHoldout, experiment };
+  } catch (e) {
+    if (experimentChanged) {
+      await rollbackExperimentAfterHoldoutFailure(
+        context,
+        experiment,
+        originalExperimentValues,
+        "after holdout update failed",
+      );
+    }
```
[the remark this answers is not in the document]
msamper: Did a best effort to reduce the likelihood of this happening. However, the possibility remains and I think if we want to address it we should do it in a separate PR. The likelihood is pretty low so I'm not too concerned about it right now.

packages/back-end/src/services/holdouts.ts
```diff
@@ -457,6 +680,37 @@ export async function setHoldoutStage(
...
+    } catch (e) {
+      const reverted = await rollbackExperimentAfterHoldoutFailure(
+        context,
+        updatedExperiment,
+        { status: originalStatus, phases: originalPhases },
+        `after holdout update failed during "${event}"`,
+      );
+      if (reverted) {
+        refreshPayload(`Reverted: ${event}`);
+      }
```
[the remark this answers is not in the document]
msamper: Again did a best effort to reduce the likelihood of this happening. If we want to try and guard against this even more I think I'd prefer to tackle this in a different PR to avoid ballooning this PR. Ideally we would solve this with transactions but we avoid using them to maintain compatibility with other non mongo data stores.

packages/back-end/src/api/ApiModel.ts
```diff
@@ -95,6 +95,7 @@ export type OpenApiEndpointSpec = {
   validator: RequestSchemas<z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>;
   zodReturnObject: z.ZodTypeAny;
   summary: string;
+  description?: string;
```
lukesonnet, line 98: Why is this necessary? Seems fine/good, but wondering why it didn't exist before.

packages/back-end/src/api/specs/holdout.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const holdoutStartAnalysisEndpoint = {
+  pathFragment: "/:id/start-analysis",
+  verb: "post" as const,
+  operationId: "startHoldoutAnalysis",
+  validator: apiHoldoutActionValidator,
+  zodReturnObject: apiHoldoutActionReturn,
+  summary: "Start the Holdout's Analysis Period",
+  description:
+    "New Feature Flags and experiments can no longer be added, but existing traffic splits remain active. Results exclude data from before the analysis period so you can measure the cumulative impact after changes are frozen.",
```
lukesonnet:
```suggestion
    "Move the holdout into an analysis phase. New Feature Flags and experiments can no longer be added, but existing traffic splits remain active for existing and new traffic. Results exclude data from before the analysis period so you can measure the cumulative impact after changes are frozen.",
```

packages/shared/src/validators/holdout.ts
```diff
@@ -121,3 +128,302 @@ export const createHoldoutInputValidator = z.object({
...
+  .describe(
+    "Automatic stage transitions for the Holdout. Dates must be consecutive: `startAt` before `startAnalysisPeriodAt` before `stopAt`. Send `null` to delete the schedule.",
+  );
+
+/**
+ * Read shape for a Holdout. A Holdout is stored as two documents — the holdout
+ * itself and a companion experiment of type `holdout` — and this schema merges
+ * both so callers never have to know which half a field lives on.
+ */
+export const apiHoldoutValidator = namedSchema(
```
lukesonnet, line 198: q: Should we return an experiment ID to get results from? Should we block all changes to an experiment that is a holdout (maybe we already do that)? Or should we expose a holdout / results endpoint that lets you get the results just as you would for an experiment?

I think maybe just the latter.... We may need to then add at a later point "get dimension breakdown" endpoints like we have for experiments to allow that functionality programmatically for holdouts... which is why it could be good to just use the experiment endpoints with the holdout linked experiment ID, but then we need to make sure to block all mutating experiment endpoints from affecting holdout type experiments (which I think they already do).

packages/back-end/src/services/holdouts.ts
```diff
@@ -295,11 +412,290 @@ export async function createHoldoutWithExperiment(
...
+}
+
+/**
+ * Roll an experiment back to a prior snapshot after its companion holdout write
+ * failed (the two documents share no transaction). Returns whether the restore
+ * succeeded so callers can gate follow-up work.
+ *
+ * The restore is guarded on the exact field values this operation wrote, not the
+ * whole-document `dateUpdated`. An unrelated edit to some other experiment field
+ * therefore no longer blocks compensation, while a concurrent write to one of the
```
lukesonnet: Seems a little risky. Is it worth handling this case when we should just try and rollback and fail if it can't? IDK it seems unlikely the managed holdout experiment will have unrelated changes unless we aren't properly collecting the set of changes caused by the holdout change (which we should remedy anyways).

packages/back-end/src/services/holdouts.ts
```diff
@@ -295,11 +412,290 @@ export async function createHoldoutWithExperiment(
...
+    if (body.targetingCondition !== undefined) {
+      const conditionResult = validateCondition(body.targetingCondition);
+      if (!conditionResult.success) {
+        throw new Error(
+          `Invalid targeting condition: ${conditionResult.error}`,
+        );
+      }
+    }
+
+    const current = phases[phases.length - 1];
```
lukesonnet: We use phases to basically be stand-ins for "running" and "analysis-period" in holdout experiments. Does this work the way we expect? I feel that we should update *all* phases? This might require more investigation and thought as to the best way to do it.

packages/back-end/src/api/ApiModel.ts
msamper: Not sure why it didn't exist, but it allows us to provide a description for custom api handlers which is helpful here to explain the holdout lifecycle endpoints

packages/shared/src/validators/holdout.ts
msamper: My initial thought was we should do the latter to continue to hide the holdout implementation details from the user. And we already have guards in experiment endpoints that block changes when the experiment type is `holdout`. I don't have a super strong opinion though. Agree that it would be easier to just reuse the existing results endpoints to avoid restating it. I'm okay with surfacing an `experimentId` for that purpose and making it clear in the documentation that its used for fetching results

packages/back-end/src/services/holdouts.ts
msamper: Yes you're right that we should update all phases. Partially fixing this in #6749 but I'll update the code here to mirror targeting changes across phases.

packages/back-end/src/services/holdouts.ts
msamper: Probably not. The likelihood of the write failing is already low and the likelihood of a concurrent write to the same field is even lower. I simplified the rollback function and got rid of the CAS logic.

packages/back-end/src/services/holdouts.ts
```diff
@@ -295,11 +302,203 @@ export async function createHoldoutWithExperiment(
...
+ * reconciliation rather than masking the caller's original error.
+ */
+async function rollbackExperimentAfterHoldoutFailure(
+  context: ReqContext | ApiReqContext,
+  experiment: ExperimentInterface,
+  revertChanges: Partial<ExperimentInterface>,
+  logContext: string,
+): Promise<boolean> {
+  try {
+    await updateExperiment({ context, experiment, changes: revertChanges });
```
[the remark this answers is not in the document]
msamper: The likelihood of this happening is very low given that the write to the holdout would need to fail which is already unlikely and then a concurrent update would also need to be happening. This logic for the rollback is easier to reason about and is fine given how unlikely this is.

msamper:
> ### Confidence Score: 4/5
> The PR is not yet safe to merge because lifecycle compensation can leave SDK caches serving a failed transition when asynchronous forward and rollback refreshes complete out of order.
> 
> The database compensation path is present, but its forward, rollback, and explicit SDK refreshes are launched independently without an ordering barrier, so a stale forward rebuild may finish after the restored-state rebuild.
> 
> **Files Needing Attention:** packages/back-end/src/services/holdouts.ts; packages/back-end/src/services/features.ts
> 
> Reviews (15): Last reviewed commit: ["handle payload refresh edge case"](https://github.com/growthbook/growthbook/commit/75cf0900b2f30c346eefb9d04cffcb84aeaea49d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54231355)

The forward and rollback refreshes are mutually exclusive. Since the catch in `commitTransition` rethrows, the forward `refreshPayload` is unreachable after a failure. `updateExperiment`'s own refresh is the only other possible refresh, and it never triggers for holdout experiments.

msamper:
@greptileai

lukesonnet (approved):

PR #6572, opened 2026-08-05
Fix retry on timeout and copy for error banner in explorer

najito:
### Features and Changes

In the Explorer, when a query times out, users previously saw a generic "Latest changes not applied" banner with a "Refresh" button that looked identical to the normal stale-config state — there was no clear indication that the query had actually failed, and no dedicated way to retry it.

This PR:

- Adds a distinct timeout error state to the floating callout in `ExplorerMainSection`: it now renders with `status="error"`, an error status icon, the text "Query timed out", and a "Retry" button (instead of the info-styled "Latest changes not applied" / "Refresh").
- Makes the timeout error copy dynamic based on dataset type via `getQueryTimeoutErrorMessage()` — funnel queries mention "or fewer steps" as an additional mitigation, non-funnel queries don't.
- Extracts `isQueryTimeoutError()` in `util.ts` to detect the timeout error by matching a shared prefix constant, so `ExplorerContext` (where the error is raised) and `ExplorerMainSection` (where it's matched to offer the retry) stay in sync instead of duplicating the message string.
- Repositions the callout when in the error state (offset further down) so it doesn't overlap other UI when both a timeout error and stale state could otherwise compete for the same spot.

### Dependencies

None

### Testing
Set timeout to 1s on query and use explorer.

### Screenshots
Screenshot 2026-08-05 at 4 11 43 PM
Screenshot 2026-08-05 at 4 11 49 PM
Screenshot 2026-08-05 at 4 12 01 PM


jdorn (commented):
I think the fix for this can be much simpler.

```diff
diff --git a/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerDataTable.tsx b/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerDataTable.tsx
index 8525de575..7eb62eb6f 100644
--- a/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerDataTable.tsx
+++ b/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerDataTable.tsx
@@ -126,7 +126,7 @@ export default function ExplorerDataTable({
       csvColumnKeys={csvColumnKeys}
       csvColumnLabels={csvColumnLabels}
       renderCell={renderCell}
-      paddingTop={(isStale || loading) && !hasChart ? 35 : 0}
+      paddingTop={(isStale || loading || !!error) && !hasChart ? 35 : 0}
     />
   );
 }
diff --git a/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerMainSection.tsx b/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerMainSection.tsx
index 9be1419b3..87e9779ae 100644
--- a/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerMainSection.tsx
+++ b/packages/front-end/enterprise/components/ProductAnalytics/MainSection/ExplorerMainSection.tsx
@@ -193,7 +193,7 @@ export default function ExplorerMainSection() {
           </Flex>
         )}

-        {(isStale || loading) && !suppressStaleFloatingCallout && (
+        {(isStale || loading || error) && !suppressStaleFloatingCallout && (
           <Box
             style={{
               position: "absolute",
@@ -227,7 +227,7 @@ export default function ExplorerMainSection() {
                   >
                     <Flex align="center" gap="2">
                       <PiArrowsClockwise />
-                      Refresh
+                      {isStale ? "Refresh" : "Retry"}
                     </Flex>
                   </Button>
                 )
@@ -235,10 +235,12 @@ export default function ExplorerMainSection() {
             >
               {loading ? (
                 "Loading..."
-              ) : (
+              ) : isStale ? (
                 <Text title="Some configuration changes require running a new SQL query against your data source">
                   Latest changes not applied
                 </Text>
+              ) : (
+                "Query failed"
               )}
             </Callout>
           </Box>
```

najito:
For non-timeout errors (bad SQL, a missing column, a permission issue), does `Retry` resubmit the same broken query, or does it fix anything? If the query needs a config change first, would the `Retry` button on it is misleading.

`DisplayTestQueryResults` already renders the full error text in the table. If the floating banner shows "Query failed" for every error type too, should it show the actual error message instead of a generic label? If it does, that brings back the wrapping and sizing work that scoping to timeout avoided.

If `Retry` is correct across every error type, generalizing this makes sense. If it isn't, this keeps the fix scoped to the one error class where retry is correct, and the inline error display stays the single source of truth for the rest.

PR #6578, opened 2026-08-06
Updates Vercel Integration to specify installation-level billing

mknowlton89:
### Features and Changes

- Explicitly identify GrowthBook’s Vercel plans as installation-level billing plans.
- Specify whether each plan requires a payment method.
- Require these fields in the billing plan validator to prevent future contract regressions.

This fix is necessary because GrowthBook billing applies to the entire Vercel installation, not individual resources. Without an explicit `scope`, Vercel can interpret the plans as resource-level and send plan changes to the resource endpoint, which does not create a GrowthBook license or Orb subscription.

I think Vercel must've made some changes 

### Dependencies

Installation-level Billing Plans must be enabled in the Vercel integration configuration.

### Testing

- [x] Open the GrowthBook integration settings in Vercel and confirm Starter and Pro plans load.
- [x] Upgrade an existing installation to Pro and confirm Vercel calls the installation update endpoint.
- [x] Confirm the organization receives a Pro license and Orb subscription.
- [x] Confirm Starter does not require a payment method.
- [x] Confirm Pro requires a payment method.
- [x] Confirm if a user tries to change the plan via the Resource page, there is a link to take them to the installation page

### Screenshots
Before (could attempt to change plan at resource level)
Screenshot 2026-08-06 at 9 26 39 AM

After (redirects customer to installation level)
Screenshot 2026-08-06 at 9 27 00 AM

The Integration we have has the toggle enabled for `Installation-Level Billing Plans` - this hasn't changed since we launched.
Screenshot 2026-08-06 at 9 30 19 AM


ahdriel (approved):
I have not tested but the changes seem straightforward!

PR #6588, opened 2026-08-06
fix(funnels): use one shared WITHIN GROUP ordering for Redshift step arrays

lukesonnet:
## Fix Redshift funnel error: `within group ORDER BY clauses for aggregate functions must be the same`

Funnels with 3+ steps failed on Redshift because each step's sorted timestamp array used its own column in `WITHIN GROUP (ORDER BY ...)`, and Redshift requires every such clause in a SELECT to be identical.

`arrayAggSorted` now accepts an optional shared `orderByColAlias ` expression, and the funnel builder passes the raw event timestamp (`ts`) for every step array. This is functionally equivalent, since the `stepX_ts` items are always just the `timestamp` from the previous step anyways. This argument to `arrayAggSorted` is a bit risky since now if `orderByColAlias` is not the same as `col` then this logic somewhat breaks down, but for funnels this isn't a problem and I note it in the argument name and in the docs for that argument.

**Before**

```sql
-- __funnel_ft0_events
SELECT
  user_id,
  timestamp AS ts,
  CASE WHEN (event_name = 'add_to_cart') THEN timestamp END AS step2_ts,
  CASE WHEN (event_name = 'purchase')    THEN timestamp END AS step3_ts
FROM ...

-- __funnel_user_aggregates
SELECT
  user_id,
  MIN(step1_ts) AS step1_resolved_ts,
  LISTAGG(CAST(step2_ts AS VARCHAR), '||~gb~||') WITHIN GROUP (ORDER BY step2_ts) ...,
  LISTAGG(CAST(step3_ts AS VARCHAR), '||~gb~||') WITHIN GROUP (ORDER BY step3_ts) ... --
FROM __funnel_ft0_events
GROUP BY user_id
```

This errors because the ORDER BY statement is different in this GROUP BY.

**After**

```sql
-- __funnel_ft0_events (same as above)
SELECT
  user_id,
  timestamp AS ts,
  CASE WHEN (event_name = 'add_to_cart') THEN timestamp END AS step2_ts,
  CASE WHEN (event_name = 'purchase')    THEN timestamp END AS step3_ts
FROM ...

-- __funnel_user_aggregates
SELECT
  user_id,
  MIN(step1_ts) AS step1_resolved_ts,
  LISTAGG(CAST(step2_ts AS VARCHAR), '||~gb~||') WITHIN GROUP (ORDER BY ts) ...,
  LISTAGG(CAST(step3_ts AS VARCHAR), '||~gb~||') WITHIN GROUP (ORDER BY ts) ...
FROM __funnel_ft0_events
GROUP BY user_id
```
This works because you can see how `step2_ts` and `step3_ts` are just `timestamp` or NULL, so when we use `ts` in `__funnel_user_aggregates`, the order still works.

Other dialects are unaffected (only Redshift uses the parameter). 

## Testing

Manual testing, PA redshift funnel with 3 steps now works, counts seem correct.

### DIFF
--- a/packages/back-end/src/integrations/dialects/redshift.ts
+++ b/packages/back-end/src/integrations/dialects/redshift.ts
@@ -40,9 +40,10 @@ export const redshiftDialect: SqlDialect = {
   },
   // Redshift rejects a SELECT whose WITHIN GROUP (ORDER BY) clauses differ
   // across aggregates ("within group ORDER BY clauses for aggregate functions
-  // must be the same"), so honor the shared `orderBy` when provided.
-  arrayAggSorted: (col: string, orderBy?: string) =>
-    `SPLIT_TO_ARRAY(LISTAGG(CAST(${col} AS VARCHAR), '||~gb~||') WITHIN GROUP (ORDER BY ${orderBy ?? col}), '||~gb~||')`,
+  // must be the same"), so honor the shared alias when provided. See the
+  // SqlDialect docs: the alias must equal `col` wherever `col` is non-null.
+  arrayAggSorted: (col: string, orderByColAlias?: string) =>
+    `SPLIT_TO_ARRAY(LISTAGG(CAST(${col} AS VARCHAR), '||~gb~||') WITHIN GROUP (ORDER BY ${orderByColAlias ?? col}), '||~gb~||')`,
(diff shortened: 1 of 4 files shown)


ahdriel (approved):
Nice catch and fix!

PR #6591, opened 2026-08-07
chore(test): limit workers on non-CI to avoid OOM

ahdriel:
### Features and Changes

I noticed in my local env (and also Cursor cloud agents) that sometimes running `pnpm test` would take up a lot of my memory, making things crash sometimes if we had 2 `test` processes running at the same time with dev server, etc.

So we are keeping the `CI` configuration as-is but limiting the workers for dev machines, and also recycling the `jest` workers sooner.

From my testing this should improve this situation without a lot of additional drawback.

### DIFF
--- a/packages/back-end/jest.config.js
+++ b/packages/back-end/jest.config.js
@@ -12,12 +12,10 @@ module.exports = {
       "<rootDir>/../../node_modules/.pnpm/@typespec+ts-http-runtime@0.3.1/node_modules/@typespec/ts-http-runtime/dist/commonjs/$1/internal.js",
   },
   setupFilesAfterEnv: ["<rootDir>/test/jest.setup.ts"],
-  // Cap parallelism only where a `pnpm dev` stack competes for the same box
-  // (local laptop or cloud dev agent). CI has a dedicated runner and runs the
-  // workspaces sequentially, so it keeps Jest's default worker count. Jest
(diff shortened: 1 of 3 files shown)


packages/back-end/jest.config.js
```diff
@@ -12,11 +12,10 @@ module.exports = {
...
-  // crashes with "JavaScript heap out of memory" (intermittently, depending on
-  // how files get distributed). Recycle a worker once its RSS passes 2GB so
-  // heap usage stays well under the ceiling regardless of file distribution.
-  workerIdleMemoryLimit: "2GB",
+  // Off CI, tests share the box with a running `pnpm dev` stack, so cap
+  // workers.
+  ...(process.env.CI ? {} : { maxWorkers: 2 }),
+  // Each file's module graph stays resident (~140MB/file); recycle workers
+  // before the heap fills. Tighter off CI, where memory is shared.
+  workerIdleMemoryLimit: process.env.CI ? "2GB" : "768MB",
```
bryce-fitzsimons: 768MB feels a little low for some of our bulkier single tests. Also wondering if maxWorkers could be relative like "50%" instead of 2

ahdriel: I tested and it worked ... but wdyt about 1GB?

bryce-fitzsimons (approved):

PR #6592, opened 2026-08-07
fix(custom-fields): Custom boolean fields display correctly on experiment overview

lukesonnet:
### Features and Changes

**Bug fixes**

- Boolean custom fields display correctly. `"false"` was being read as truthy, so an unchecked toggle rendered as "yes" on the experiment/feature overview.
- Boolean custom fields can actually save `false`. Untouched toggles are now sent explicitly instead of being omitted.

**Refactor**

- `CustomFieldInput` is now a controlled presentational component. It previously mutated its parent's form from two `useEffect`s.
- A new `useReconciledCustomFields` hook is the single owner of seeding defaults, pruning fields that don't apply to the selected project, and the `custom-metadata` license gate.
- Custom field value logic moved from `hooks/` to `services/customFields.ts`, removing a hook that imported from a component directory.

### Testing

Unit Tests

- [x] A boolean field left untouched saves as `false` and displays "no".
- [x] Toggling a boolean on and off saves and displays correctly.
- [x] A field with a configured default seeds on open, and clearing it sticks — both before saving and after reopening.
- [x] Switching projects mid-form swaps in the new project's fields.
- [x] Duplicating a feature carries its custom field values over.
- [x] Custom fields section should be hidden when the commercial feature is not available


packages/front-end/components/CustomFields/CustomFieldDisplay.tsx
```diff
@@ -156,25 +181,29 @@ const CustomFieldDisplay: FC<{
...
-      new Date(cValue).toLocaleString()
-    ) : cValue ? (
-      cValue
+      <>{isCustomFieldBooleanTrue(cValue) ? "yes" : "no"}</>
+    ) : v.type === "date" && stringValue ? (
+      new Date(stringValue).toLocaleDateString()
+    ) : v.type === "datetime" && stringValue ? (
+      new Date(stringValue).toLocaleString()
+    ) : stringValue ? (
+      stringValue
```
msamper: Kind of out of the scope of this PR but ideally this should be a switch. This is a bit hard to read as is. I'll make the change

packages/front-end/components/CustomFields/CustomFieldDisplay.tsx
```diff
@@ -83,30 +87,51 @@ const CustomFieldDisplay: FC<{
...
   const form = useForm<
     Partial<ExperimentInterfaceStringDates | FeatureInterface>
   >({
     defaultValues: {
-      customFields: target?.customFields || defaultFields,
+      customFields: normalizeCustomFieldValues(target?.customFields),
     },
   });
+
+  const openEditModal = () => {
```
msamper: This is a side effect of not having the edit custom fields modal be its own standalone component. We wouldn't need to do this if we broke it out into a separate component. Just calling it out, but no need to address here

packages/front-end/components/CustomFields/CustomFieldInput.tsx
```diff
@@ -30,84 +59,69 @@ const CustomFieldInput: FC<{
...
+        ([key, value]) => normalizedCustomFields[key] !== value,
+      );
+    if (!changed) return;
 
-    // Only update when we actually need to remove disallowed keys.
-    if (filteredEntries.length !== currentEntries.length) {
-      setCustomFields(Object.fromEntries(filteredEntries));
-    }
+    // Defer so parent react-hook-form setValue applies after mount
+    queueMicrotask(() => setCustomFields(reconciled));
```
msamper: This feels like a code smell to me. We don't use `queueMicrotask` anywhere else in our codebase. It makes me think that this logic doesn't belong here and maybe we need a hook to derive the reconciled values for forms with `CustomFieldInput`

msamper (approved):
Confirmed that this fixes the bug. There's a lot of other unrelated issues I'm seeing with this code that makes me think it could use a dedicated pass to clean up. I think it's fine to merge this fix for now and we can try and tackle cleaning up these components separately.

packages/front-end/components/CustomFields/CustomFieldDisplay.tsx
msamper: Swapped with a Switch

teresayung (approved):

PR #6593, opened 2026-08-07
fix(ui): truncate long breadcrumbs and wrap long page titles

ahdriel:
### Features and Changes

During the refactor of Breadcrumbs #5742 I broke the truncating functionality, which meant that the layout for resources with long text was not rendering as expected.

So this PR:
- Fixes the Breadcrumb truncation
- Fixes page titles to also break in case of single words with no spaces (use `anywhere` for `overflowWrap`)

### Screenshots

| Before | After |
|--------|--------|
| 7-report | 7-report | 

### Testing

- Visual tests


lukesonnet (approved):

lukesonnet:
Should it not use `...` at the truncation? IDK just wondering

PR #6595, opened 2026-08-08
Refresh stale SDK connections via a per-org Agenda job (REST API writes)

tzjames:
### Features and Changes

Alternative approach to reducing SDK payload cache refresh load from REST-API-triggered writes (see also https://github.com/growthbook/growthbook/pull/6059 ` for another approach to the same problem.

Behind a new opt-in flag, `SDK_PAYLOAD_REFRESH_STALE_TRACKING_ENABLED` (default off):

- A REST API write marks the specific SDK connections it affects as `staleSince` (reusing the existing `isSDKConnectionAffectedByPayloadKey` matching against the org's connection list — orgs typically have only a handful of connections, so this lookup is negligible next to the payload-rebuild work it lets us defer).
- It then enqueues a single unique per-org Agenda job to rebuild those connections, instead of rebuilding inline on whichever web server happened to receive the request. Concurrent writes for the same org collapse onto that one job via Agenda's `unique()`.
- When the job finishes, it checks whether more staleness has arrived since it started and re-enqueues itself if so — so at most one refresh per org ever runs at a time, and a lone write is never delayed by a fixed debounce window.
- UI-triggered writes are unaffected and always refresh immediately, inline, regardless of this setting.

**Two correctness issues found and fixed while building this, both verified against a real (not mocked) Agenda instance:**

1. **Reschedule-from-inside-the-handler is silently clobbered.** Agenda's `job.run()` freezes `nextRunAt` (to `null` — this job has no `repeatInterval`) *before* calling the handler, then unconditionally re-persists that frozen snapshot when the handler finishes, success or failure. A reschedule attempted from inside the handler gets silently erased by that save. Fixed by rescheduling from an `agenda.on('complete:...')` listener instead, which fires only after that save has already landed.
2. **`job.unique()` alone isn't atomic.** `findOneAndUpdate(..., {upsert:true})` can still insert duplicate job documents under truly concurrent first-time enqueues for the same org without a real index backing the matched fields — verified with 20 concurrent upserts producing up to 5 duplicates in testing. Fixed with a partial unique index on Agenda's own collection, scoped to this job's name.

### Dependencies

None.

### Testing

- `packages/back-end/test/models/SdkConnectionStaleness.test.ts` — mongodb-memory-server-backed tests for marking/finding/clearing stale connections.
- `packages/back-end/test/services/sdkPayloadStaleRefresh.test.ts` — `queueSDKPayloadRefresh` and `refreshStaleSdkConnectionsForOrg` behavior with tracking enabled (mocked dependencies).
- `packages/back-end/test/services/sdk-payload-lifecycle.test.ts` — confirms default (tracking disabled) behavior is unchanged.
- `packages/back-end/test/jobs/refreshStaleSdkConnections.test.ts` — runs a **real** Agenda instance against mongodb-memory-server to prove: the job runs once and doesn't reschedule when nothing new arrived; it correctly reschedules via the `complete` event when more staleness exists (a write landing mid-run isn't dropped); and concurrent enqueues for the same org collapse onto a single job document.
- `tsc --noEmit` and `eslint` clean.


packages/back-end/src/services/features.ts
```diff
@@ -771,8 +841,43 @@ export function queueSDKPayloadRefresh(data: {
...
+    });
+  };
+
+  if (
+    !data.context.isApiRequest ||
+    !SDK_PAYLOAD_REFRESH_STALE_TRACKING_ENABLED
+  ) {
+    runRefresh();
+    return;
+  }
```
tzjames, line 891: I don't think this is really necessary to distinguish between api requests or ui ones.  As long as the job queue has enough capacity that no job is delayed, I don't think there will be any wait here.  It is just whether it is run on this server or a jobs server, and a jobs server is preferred.

docs/docs/self-host/env.mdx
```diff
@@ -216,3 +216,4 @@ Logs collection is supported but disabled by default. You can set the `GROWTHBOO
...
 - **ROBOTS_TXT_PATH** - Define an absolute path to a robots.txt file that will be served on the `/robots.txt` path of your backend domain. If this env var is not defined, the app will look for a robots.txt file in the parent folder of the built `server.js` (which should be the `packages/back-end` directory in the default configuration)
 - **DISABLE_API_ROOT_PATH** - We include a default path in the API server to expose certain settings. Set to "true" to withhold that information. The Visual Editor however relies on the information in this path, so disabling it will prevent some of its features from working correctly.
+- **SDK_PAYLOAD_REFRESH_STALE_TRACKING_ENABLED** - When "true", REST API writes mark affected SDK connections stale and a background job rebuilds them on the job server instead of refreshing inline. Default off. UI-triggered refreshes always run immediately. Useful for orgs with very frequent API updates to reduce MongoDB load on request-serving instances.
```
tzjames: I don't see why we wouldn't want to default this to on.  The job will run immediately either way (unless another is currently running, in which case it will only be delayed until the former job finishes), it is just a question of whether it is on a request server or a job server.

The only reason I can think of that we want the env var at all would be if CRON_DISABLED=true and they don't have another job server somewhere.  I think in actuality that may only happen in a dev environment?

packages/back-end/src/jobs/refreshStaleSdkConnections.ts
```diff
@@ -0,0 +1,106 @@
...
+            logger.error(
+              e2,
+              `Failed to schedule follow-up refresh for org ${organization}; refreshing directly instead`,
+            );
+            return refreshOrgDirectly(organization).catch((e3) => {
+              logger.error(
+                e3,
+                `Direct fallback refresh also failed for org ${organization}`,
+              );
+            });
```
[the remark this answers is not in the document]
tzjames: @greptile-app This is turtles all the way down.  I think this is already one try/catch too far.  If there is a DB outage, running the heavier refresh will also error.  This could go on recursively forever.  Logging the issue and then not constantly rehammering the database seems right to me.

tzjames:
@greptile-app This is turtles all the way down. I backed one try/catch out because it went too far. If there is a DB outage, running the heavier refresh will also error. This could go on recursively forever. Logging the issue and then not constantly rehammering the database seems right to me.

docs/docs/self-host/env.mdx
Kevin-Chant: I was also surprised this was an option instead of just always on, but CRON_DISABLED is a good point.

Kevin-Chant:
I'm "approving" this PR but I pushed a number of changes via #6687 and James had wanted more reviewers than just me, so I'm not actually hitting the :white_check_mark: button

Issue #673, opened 2022-10-27
[Feature] Option to return experiment and variation names from SDK endpoints


jdorn:
### Description of Feature

Currently, the only way to identify an experiment in the SDK trackingCallback is via the `trackingKey` and variation array index:

```ts
function(experiment, result) {
  // trackingKey and array index
  console.log(experiment.key, result.variationId);
}
```

The improve logs and data usability, we should allow human-readable names as well for both the experiment and variations.  Ideally, you would be able to do:

```ts
function(experiment, result) {
  console.log(
    experiment.key,
    experiment.name,
    result.variationId,
    result.variationName
  )
}
```

This involves updating the SDKs to support these additional properties and changing our SDK endpoint to add them to the response.  We may want to make it optional on a per-endpoint basis.  Some users may not want to expose that info to client-side code for example.

bttf:
Customer asked about this today
https://growthbookusers.slack.com/archives/C01T6Q0TEG3/p1694093853210739

jdorn:
This is supported now as an SDK Connection setting

PR #6597, opened 2026-08-09
Expose the real experiment id to the SDK when you opt in

### LINKED ISSUE #6584: [Bug] non-unique experiment.key possible
### Summary

The sdk's `experiment.key` is the experiment's `trackingKey`, which isn't a strictly unique identifier per experiment. In the UI, a unique tracking key is generated when you create a new experiment, but users can assign a _pre-existing_ tracking key to a new experiment, so tracking keys are not guaranteed unique (we hit this in practice with an internal user reusing tracking keys across experiments).

The REST API's experiments/ endpoint _does_ expose a strictly unique `id` per experiment (`exp_abc123...`). It would be beneficial to have this `id` available in the sdks alongside the tracking key.

Today, the only experiment identifier available client-side is `experiment.key`, which can have one-to-many relationships with experiments.

#### Expected Behavior

Expose the experiment's immutable `id` in addition to `key`, e.g.:

```ts
function (experiment, result) {
  console.log(
    experiment.id,        // "exp_abc123..."
    experiment.key,       // "experiment-name-reformatted-2"
    result.variationId,
  );
}
```

#673 raised a privacy concern --- this could be opt-in per sdk connection and default to off.

#### Current Behavior

`experiment.key` (the tracking key) is the only experiment identifier available in the sdk / trackingCallback. Because tracking keys can be reused across experiments, this identifier is not guaranteed unique.

EmmS21:
### Features and Changes

**Product**

Today, when someone is assigned to an experiment, the SDK / `trackingCallback` only gets the experiment’s **tracking key** (`experiment.key`). That’s the label meant for analytics — not GrowthBook’s unique experiment id (`exp_...`), which already exists in the DB and REST API.

The problem is that tracking keys are not guaranteed unique. People can reuse an existing tracking key on a new experiment. When that happens, two different experiments show up in analytics under the same key, so client-side events can’t reliably say which GrowthBook experiment they came from. That makes it hard to join those events back to the right experiment and compare results cleanly. See [#6584](https://github.com/growthbook/growthbook/issues/6584).

**Changes**
- New SDK Connection setting: `includeExperimentIds` (checkbox: “Include experiment IDs in payload”)
- Default **off** (opt-in), same pattern as other “include …” payload extras like rule IDs
- When on, send `exp_...` in:
  - feature `experiment-ref` rules as `experimentId`
  - visual / URL redirect auto-experiments as `id`
- sdk-js maps `experimentId` → `experiment.id` so `trackingCallback` can see it
- Leave feature-rule `id` alone (that’s already used for rule IDs)

**Rationale**

An experiment can show up in the SDK payload as part of a feature flag, or as a visual / URL-redirect experiment. There is already a pattern for that with `includeExperimentNames` — one SDK Connection setting covers both. I followed that here so turning on experiment IDs fixes both branches.

Visual / redirect experiments also have a `changeId`, but that’s a different identifier — it doesn’t replace the unique `exp_...` id for joining back to the API — so we still include experiment IDs on that branch when the setting is on.

**Tradeoffs**

- Default off / opt-in: [#6584](https://github.com/growthbook/growthbook/issues/6584) calls out the privacy concern from [#673](https://github.com/growthbook/growthbook/issues/673) — extra experiment metadata on the client shouldn’t ship unless a team asks for it. Same idea as other payload extras like rule IDs (default off). Safer default, but `trackingCallback` only gets the unique id after someone turns the setting on.
- One setting updates both feature-flag experiments and visual / redirect experiments — same pattern as `includeExperimentNames` ([#1161](https://github.com/growthbook/growthbook/pull/1161)). A bit more surface area than only fixing feature flags, but otherwise the other branch would still be missing the id.

**Implementation**

- Added `includeExperimentIds` across shared types/validators, the SDK Connection model/API, payload builders (`getFeatureDefinition` + `generateAutoExperimentsPayload`), and the connection form checkbox
- After payload capability `pick` cleanup, put `experimentId` back on the rule when the setting is on (same idea as rule ids) so it doesn’t get stripped
- Added payload + sdk-js tests for the handoff

- Fixes https://github.com/growthbook/growthbook/issues/6584

### Testing

1. SDK Connections → edit a connection → under Observability and QA, confirm **Include experiment IDs in payload** starts unchecked
2. Turn it on, save, fetch that connection’s features payload
3. Confirm feature `experiment-ref` rules have `experimentId: "exp_..."`, and visual/redirect experiments have `id: "exp_..."` (when those are enabled)
4. Turn it off, refetch — id fields gone, tracking `key` still there
5. With it on, `trackingCallback` gets `experiment.id` and `experiment.key`

### Automated checks

- Payload includes/omits experiment ids:  
  `pnpm --filter back-end test test/services/sdk-payload-generation.test.ts -t includeExperimentIds`
- sdk-js maps `experimentId` → `experiment.id`:  
  `pnpm --filter @growthbook/growthbook test test/features.test.ts -t experimentId`


alyssamichelle:
@bryce-fitzsimons Want to ship an opt-in SDK Connection setting that puts the unique exp_... id on trackingCallback, so analytics can tell experiments apart when someone reuses a tracking key?

Issue #6386, opened 2026-07-16
SDK: redirect experiments cause SRM in Managed Warehouse (cookie scope + exposure loss on navigation)


madhuchavva:
## Objective

A redirect experiment (feature flag `0/1/2`, weights `10/45/45`, arm 0 = no redirect) tripped the Sample Ratio Mismatch check in GrowthBook's Managed Warehouse (observed split ≈ `10.9 / 46.7 / 42.4`) while Mixpanel showed a clean `10/45/45` — and GrowthBook counted ~36% more unique units than Mixpanel.

Goal is to find answers to 3 questions with a controlled, reproducible experiment rather than reasoning from production data alone:

1. What actually causes the SRM — duplicated identifiers, lost events, or both?
2. Does the proposed `uuidCookieDomain` cookie fix resolve it?
3. What explains the gap between the two equal-weight arms (arm1 > arm2)?

## Stage 1 — isolate the mechanisms

Six configs vary one factor at a time: cookie scope (host-only vs parent domain), whether the redirect fires before or after the SDK's `100ms` event flush, and whether the destination page fires exposures.

| Finding | Config | Evidence |
|---|---|---|
| The bug reproduces | A | SRM p = 4e-29, arm-0 share 17.4% vs 10% weight |
| Duplicate ids alone do NOT cause SRM | B | 1.90x over-count, split clean (p = 0.33) — re-randomized duplicates land uniformly, preserving the ratio |
| Lost events are what skew the split | A vs B | With an instant redirect, 93% of origin exposures on arms 1/2 never arrive (the browser cancels in-flight `fetch` on navigation); arm 0 never navigates, so it never loses any |
| The cookie fix works — with a caveat | C | Units == N exactly, clean split. But it stays clean only because the destination page re-fires the exposure for the *same* id, rescuing the lost origin event |
| The caveat is real | E | Same cookie fix, destination not firing exposures: arms 1/2 keep only 3–5% of their units, SRM p = 3e-296 |
| Keepalive delivery eliminates the loss | D, F | With requests that survive navigation, loss is 0% and every scenario is clean |

**Conclusion → next fix**: the cookie fix (already written) solves the duplicate-identity half. The event-loss half needed its own SDK change: deliver exposure events so they survive the page being torn down.

## Stage 2 — explain arm1 > arm2

[...]

PR #1696, opened 2023-09-29
Draft: Clarification of wording about trackingCallback

yaiir-a:
### Features and Changes

I found the documentation around experimentation confusing and difficult to get my head around. Part of the issue was the wording here which threw me off. 
As far as I can see, there is no other mechanism for GrowthBook to figure out who was actually put into a given experiment.

Issue #1646, opened 2023-09-05
[Feature] Add indicator that will show experiment assignment is finished, regardless of trackingCallback is called


tusmenko:
### Description of Feature

Currently, to check if growthbook is fully initialized, we are supposed to read the "`ready`" flag: 
https://docs.growthbook.io/lib/react#waiting-for-features-to-load

I'm using it to ensure that all the experiment data is ready before sending any events to my CDP, which should include experiment and version IDs for further A\B comparison.  

Unfortunately, the `ready` flag shows only Feature initialization readiness but tells you nothing about the Experiment assignment status.

This leads to the events sent without experiment details in them, in case I'm sending events ASAP, like "Page view event".
This corrupts data because I can't later compare the "Page view event" with the following events, like clicks and so on.

The current workaround is to postpone event sending by 1-2 seconds, allowing `trackingCallback` to be fired.

The more proper solution might be saving some flag manually in `trackingCallback`, and checking it later, but unfortunately, `trackingCallback` calls only when an experiment assignment actually happens, not when the experiment assignment is finished with whatever the result is. 

So nothing allows you to track that "Fine, everything is settled up, feels free to fire event".
Which leads to the following sequence:

> isGrowthBookConfigured =>  true
> growthbook.ready => true
> Sending OPENED_PAGE ExperimentId: undefined
> growthbook trackingCallback  => EXPERIMENT_DATA

Proposed solutions:

1) Add an `experimentsReady` flag to the GrowthBook root, indicating that the experiment targeting is finished regardless if the assignment to experiment actually happened

or

2) Include experiment assignment status in the `ready` value meaning.

alyssamichelle:
**Our audit of stale tickets suggests this may already be resolved, or no longer applicable, on the current version — but we couldn't prove it from the code alone.** @tusmenko could you confirm whether you still see this on the latest release? @ahdriel assigning to you to make the final call once they reply (or after a couple of weeks of silence).

[...]

PR #6598, opened 2026-08-09
Clarify that evaluating a feature enrolls users in feature flag experiments

### LINKED ISSUE #3121: [Docs] Better explain how an experiment is started and what triggers the trackingCallback
Submitting on behalf of a customer.

The current documentation doesn't make clear that a `get` method like `getFeatureValue` is actually a "start experiment" method and also triggers the `trackingCallback`. 

Since most of the methods eventually call `evalFeature` this could be better explained in the SDK documentation.

It would be nice to add a "getting started" section specifically for engineers to better build/interact with an experiment, as opposed to what seems like a mixture of different topics (current state of the docs).

Even a separate API section would be helpful also.

EmmS21:
Hi all, I genuinely think this worth looking into. I see a pattern of Slack issues where people are comparing the wrong people at the wrong time because of documentation creating confusion around when a person is randomized.

I made a doc update on this MR, but I believe this needs to be expanded to include the pages those threads actually read.

This MR originally only rewrote the SDK docs. The confusion requires expanding this to the dashboard copy people follow when they click Start, the feature-flag guide that says publishing will start randomizing users, and the script-tag docs that make it sound like dropping the snippet on the site is enough.

### How it shows up

From Slack, over the last 12 weeks there have been 3 threads that relate to confusion regarding when a person enters an experiment and when the dashboard says the test has started:

- Sparsh asked whether copying the script into their Shopify with data tracking set to growthbook meant the enrollment process would start automatically even if they never called isOn in Shopify
- Ajay assumed setting a test to 100% meant every visitor would be enrolled in the test, only a third showed up in his GA, causing confusion about the delta
- Dev Meena believed retrieving the flag meant a user was already included in an experiment. They treated the tracking callback function running as confirmation that this person was assigned (the exposure). They assumed the SDK failed when they noticed inconsistent results.

This issue bubbles up in the repo multiple times (#6386, #3121, #1646)

### What this extension adds

- **Start Experiment** (get-started + Start modal): Start publishes the test. Feature Flag experiments enroll when the app evaluates the flag. Visual Editor experiments enroll when the page loads.
- **Feature flag guide** (Experiment Results): save + publish does not randomize users. Evaluate + qualify does. 100% is 100% of those evaluations, not all visitors. Results only count people who got an exposure.
- **Script tag**: snippet is enough for Visual Editor experiments, not Feature Flags. `data-tracking="growthbook"` forwards an exposure that already fired; it does not enroll anyone. Same sentence on Event Forwarder / Managed Warehouse.
- **SDK pages** (already in this PR): evaluate is when this user enters. Added: if they are not included, you still get a value and `trackingCallback` does not run.

---

### Original SDK-only notes

[#3121](https://github.com/growthbook/growthbook/issues/3121) asked us to better explain how an experiment is started for a user and what triggers `trackingCallback`. It was filed from a customer conversation (August on behalf of the customer; [simonsmith](https://github.com/simonsmith) later confirmed he was the source and offered to walk through the misunderstandings).

Along the way I also looked at:
- [#1696](https://github.com/growthbook/growthbook/pull/1696) — Earlier docs PR tried to say `trackingCallback` is needed for measurement, not just something you “can” use. Approved, never merged. Same confusion neighborhood as #3121.
- [#1646](https://github.com/growthbook/growthbook/issues/1646) — A customer treated `ready` and `trackingCallback` as proof that assignment was done. `ready` only means features loaded, and the callback only fires when the user is actually enrolled. Same timing confusion as #3121, even though they were asking for a product flag, not docs.
- Comments on [#3121](https://github.com/growthbook/growthbook/issues/3121) — hard to trigger `trackingCallback` in tests when it’s unclear what starts enrollment

**Core confusion**

The docs already showed how to **set up** `trackingCallback`, and later said feature experiments “just evaluate the flag” with tracking “in the background.” What was easy to miss: for feature flag experiments, **evaluating the feature** (`getFeatureValue`, `isOn`, hooks, etc.) is when **this user** enters the experiment — and that is also when `trackingCallback` runs.

`get…` reads like a pure UI read. People look for a separate “start experiment” call. There isn’t one on this path. `trackingCallback` does not start enrollment; it **logs** the exposure after evaluation enrolls the user (when they qualify).

- Fixes https://github.com/growthbook/growthbook/issues/3121


alyssamichelle (changes requested):
Got some helpful feedback internally. There does seem to be a real education gap here, specifically around the fact that feature evaluation is what triggers experiment enrollment.

The consensus is that the app changes in this PR probably aren’t the right solution, since the current behavior is intentional. Instead, we should make this much clearer in our docs/onboarding and address the underlying confusion there rather than adding handholding across individual pages.

EmmS21:
> Got some helpful feedback internally. There does seem to be a real education gap here, specifically around the fact that feature evaluation is what triggers experiment enrollment.
> 
> The consensus is that the app changes in this PR probably aren’t the right solution, since the current behavior is intentional. Instead, we should make this much clearer in our docs/onboarding and address the underlying confusion there rather than adding handholding across individual pages.

@alyssamichelle 

Thanks. I removed the product copy. The PR is now docs only.

Out of curiosity;
1. Where do you feel most worried about similar confusions 
2. Why is that area a priority

EmmS21:
@alyssamichelle 

I couldn't stop thinking about this.

I think workflows can be explained more intuitively with less text and something more intuitive.

I built a simulation of a universe explaining the flow of information. I think this can also predictively catch issues when changes are being proposed:

emms21.github.io/growthbook-universe/

alyssamichelle:
Pushed a follow-up commit onto this branch: https://github.com/growthbook/growthbook/commit/a820f952e

We kept the facts from the Slack threads. Publish, fetch, and `data-tracking` are not this visitor entering. Evaluate + qualify is. `trackingCallback` logs that exposure.

We ran two internal review passes on the original paste (a maintainability pass and an over-engineering pass). Both said the facts are right and the structure is not: one invariant was stickered onto every nearby paragraph. This commit is those edits.

## Why each change

### `docs/feature-flag-experiments.mdx` — this page owns the model

**Intro of Running Experiments.** The opening still said the experiment rule “will randomly assign variations to your users” and “When a user is placed in an experiment via an experiment override rule, the assignment will be tracked.” That is the Slack/GitHub confusion, on the one page whose job is this model. Publish makes the rule available. A user is enrolled when the app evaluates the feature (`isOn` / `getFeatureValue`) and they qualify. That is when `trackingCallback` runs. If they are not included, they still get a value and the callback does not run (Dev Meena treated a returned value as proof of exposure).

**Exposure, Variations, and Weights.** “100% traffic means 100% of those evaluations, not 100% of all visitors” belongs next to the traffic control, not under Results. That is Ajay’s thread.

**Experiment Results.** The whole model had been jammed under the wrong H2. Results now only says results only include people who got an exposure, plus the existing results link.

### SDK docs — one sentence + a link, not seven copies

**Deleted the Step 2 enrollment sentences** on js / node / react / react-native. Quickstart is the wrong layer. Experimentation already exists ~300 lines later on those pages. The JS npm README never got that extra sentence, so docs and the package README had already forked.

**Replaced the three-sentence paste** on js / node / react / react-native / roku with:

> Evaluating the feature is what enrolls a user and fires `trackingCallback`. See [Feature Flag Experiments](/feature-flag-experiments).

Method names already sit in the code sample under each heading. A shared snippet would still be seven call sites. A link deletes them, so the next wording tweak cannot drift.

**npm READMEs** cannot rely on Mintlify routes, so they keep one short local sentence (including “if they are not included…”) instead of the longer paste.

### `docs/lib/script-tag.mdx` — say it once, and do not lie about Visual Editor

**Intro.** Distinguishes VE (runs on page load, no code) vs feature flags (`isOn`). Kept the fourth sentence because dropping the snippet is Sparsh’s confusion. Dropped the em-dash; house style.

**`data-tracking` bullet.** Still says it forwards experiment view events that already fired and does not enroll users. Dropped “does not evaluate flags”: this attribute never evaluated flags, and saying so is extra.

**Tracking Experiment Views.** Removed “Loading the script does not fire this by itself.” That is true for feature flags and **false for Visual Editor**, which does enroll on page load when targeting matches. Tracking is a shared path; a feature-flag-only claim does not belong there. Kept “included” rather than “views.”

**Using Feature Flags.** Restored “Here is an example.” The intro already taught enrollment. Repeating it here was the fourth restatement on one page.

### Warehouse pages — forwarding vs enrolling, then stop

**`event-forwarder.mdx` and `managed-warehouse.mdx`.** `data-tracking="growthbook"` forwards experiment view events that already fired. It does not enroll users. Dropped “Simply,” “to your script tag to enable,” and “or evaluate Feature Flags.” Warehouse setup should not re-teach feature-flag enrollment.

## Why not the universe sim in this PR

The [universe simulation](https://emms21.github.io/growthbook-universe/) is a clear walkthrough of create → store → fetch → evaluate. Same sequence this commit puts in the docs. It is a separate teaching artifact. This PR does not need to ship a simulation to fix the contradicting concept page.

alyssamichelle (approved):

Issue #4889, opened 2025-11-25
[Bug] Node SDK does not auto-refresh feature definitions unless refreshFeatures() is manually called


Rakesh-46-VR:
### Summary

When using the GrowthBook Node SDK (@growthbook/growthbook), updated feature values on the GrowthBook dashboard are not reflected in new incoming requests unless:

* the server is restarted or
* await client.refreshFeatures({ skipCache: true }) is manually called inside the request middleware

cacheSettings.backgroundSync inside init() does not appear to refresh features automatically, and the feature cache remains stale.

A similar issue was previously reported by another user in the GrowthBook Slack community for Next.js:
https://growthbookusers.slack.com/archives/C01T6Q1SVFV/p1757507869836329

#### Expected Behavior

* After calling client.init({ cacheSettings: { backgroundSync: true, … } }), the SDK should automatically refresh feature definitions in the background.
* New requests should receive updated feature values without requiring a server restart.
* Calling client.createScopedInstance() should use the freshest feature definitions.

#### Current Behavior

* The SDK loads feature definitions once during initialization and keeps serving stale values.
* Changes made in the GrowthBook dashboard (toggling defaults, modifying rules, etc.) are not reflected in Express route handlers.
* The only way to get updated values is to manually call:
   * `await client.refreshFeatures({ skipCache: true });`
inside each request’s middleware — which defeats the purpose of backgroundSync and caching.

#### Reproduction Steps

* Run the below node js server with you env's configured and create a `feature1` feature flag in your dashboard that returns boolean.

```
import express from "express";
import { GrowthBookClient } from "@growthbook/growthbook";
import "dotenv/config";

const app = express();

// GrowthBookClient instance
const client = new GrowthBookClient({
  apiHost: process.env.GROWTHBOOK_HOST,
  clientKey: process.env.GROWTHBOOK_CLIENT_KEY,
});

// Initialize it
client.init({ cacheSettings: { backgroundSync: true } }).then((status) => {
  console.log("GrowthBook initialized", status);
});

app.use(async (req, res, next) => {
  // Attributes about the current user/request
  const userContext = {
    attributes: {
      // In a real app this would come from a cookie or session
      // We would also add more attributes like country, etc
      id: "123",
    },
  };

  req.growthbook = client.createScopedInstance(userContext);
  next();
});

app.get("/", (req, res) => {
  let message = "Hello World!";

  if (req.growthbook.isOn("feature1")) {
    message = "Hola Mundo!";
  }

[...]

Issue #1965, opened 2023-12-19
[Bug] Streaming Updates doesn't work properly with streamingHost property


davidtaboas:
### Summary

Setting streamingHost property is not working properly with the streaming updates because of this line

https://github.com/growthbook/growthbook/blob/e1f2314ad397645df955182fb43c1ef2765c75b5/packages/sdk-js/src/feature-repository.ts#L407

#### Expected Behavior

The sse support should be checked against streamingHost, not apiHost

#### Current Behavior

The sse support is using the apiHost although streamingHost is set

bryce-fitzsimons:
Can you please provide more information about your setup, both from an SDK and networking standpoint? We intended that the `apiHost` would be responsible for setting the `"x-sse-support"` header. If the header is present, it means: SSE is supported, go ahead and try to establish a streaming connection.

dlemaigl:
We understand that if we specify a _streamingHost_ the check if it is possible to use SSE should be done against that URL and not against _apiHost_.

In our case the _apiHost_ URL does not support _SSE_, we want to use it only to serve the static JSON of the SDKs, and the _streamingHost_ URL is the one in charge of serving those SSE.

If we send the "x-sse-support" header from _apiHost_ we would be lying because in that URL it is not supported.

bryce-fitzsimons:
I understand the confusion, but we didn't design it to work that way. Basically, _apiHost_ is the main delegator or decider about whether or not streaming is allowed (either on itself or on _streamingHost_). When the SDK first handshakes with GrowthBook, it's on the `/api/features/sdk-xxx` endpoint (apiHost). It will hydrate down the features as well as inform the SDK that "everything is ok for streaming". If we did the `x-sse-support` check on the _streamingHost_ instead, then we'd need an additional hop before starting streaming.

bryce-fitzsimons:
Closing this issue, but feel free to reach out if you'd like to discuss more.

bryce-fitzsimons:
https://github.com/growthbook/growthbook-flutter/commits/main/

Issue #1457, opened 2023-07-13
Published features taking more than 1 min to take effect on the client side W/ Proxy server


vivian940425:
Hi,

I am trying out Growthbook in our internal server. Originally I did not enable proxy server and noticed the published features took very long time ~2 mins to get to the clients. I thought it was because I did not use Proxy server so I set up the proxy server later. It did help a bit now the latency is about ~1 min.

Compared to other solutions we tried before like launchdarkly or flagsmith, the latency seems to be pretty high. These products took somewhere between 1 to 10 seconds to be effective. I am not sure if this is kinda expected or I am doing something wrong. 

Thank you so much for the work!

jdorn:
Hello.  Which SDK are you using?  Some of them have their own caching layers that you need to configure.  Currently, our Javascript-based SDKs are the only ones which support realtime streaming updates (features released in under a second when using the Proxy server).  The other SDKs mostly use polling and fetch new features every ~60 seconds by default, although this is configurable.  We are working to enable realtime streaming updates for all of our SDKs.

vivian940425:
> Hello. Which SDK are you using? Some of them have their own caching layers that you need to configure. Currently, our Javascript-based SDKs are the only ones which support realtime streaming updates (features released in under a second when using the Proxy server). The other SDKs mostly use polling and fetch new features every ~60 seconds by default, although this is configurable. We are working to enable realtime streaming updates for all of our SDKs.

Hi Jeremy, 

Thanks for the reply. I am indeed using the JS based SDK in a react environment. I noticed this is somewhat inconsistent, at times I can see the changes taking effect immediately but most of the time it took a bit longer(~ a minute). Is there anything I can capture to better analyze the delay? 

Thanks!

alyssamichelle:
**Our audit of stale tickets suggests closing this as no longer applicable.** @teresayung assigning to you for a decision — agree?

**Why:** This was a support/configuration question, not a confirmed defect. Realtime SSE streaming is now first-class in the JS/React SDK via `streaming: true` / `streamingHost` (docs/lib/js.mdx 'Streaming Updates', packages/sdk-js/src) and has since been added to Go, Swift, Kotlin, C#, Flutter, React Native and others (docs/lib/*.mdx). Polling interval and SSE behaviour are documented; the reporter never followed up with reproduction details after jdorn's 2023 reply.

[...]

PR #6599, opened 2026-08-09
docs: clarify Node backgroundSync vs streaming vs cache refresh

EmmS21:
## Summary
- Clarifies on the Node SDK docs what **streaming**, **cache refresh**, and **`backgroundSync`** each actually mean, so people stop treating `cacheSettings.backgroundSync: true` as auto-refresh.
- Callout under Refreshing Features, plus short notes in Streaming and Caching (and document `backgroundSync` in the cache settings example).
- Fixes a small typo (`defintiions` → `definitions`).

Closes / relates to: https://github.com/growthbook/growthbook/issues/4889

## Investigation

I dug into #4889. In a long-running Node process, feature values stay on whatever loaded at startup until you restart or call refreshFeatures() by hand. People were using init({ cacheSettings: { backgroundSync: true } }) thinking that meant auto-refresh. The original reporter hit this, irsooti confirmed it later, and Slack shows up in the issue too.

### What I tested
I reproduced the reported setup locally with a mock `/api/features/...` server and `GrowthBookClient` using the same init as the issue:

1. Init → feature off  
2. Flip the API payload and wait → still off  
3. `refreshFeatures({ skipCache: true })` → updates  

The observed issue is real with that config.

### What that showed
The assumption in the issue is that `backgroundSync: true` means “keep refreshing in the background.”

From the SDK code (`startStreaming` only runs when `streaming: true`; `backgroundSync` only *allows* SSE; Node has no `EventSource` unless you polyfill) and the existing Node docs, that’s not what it does. Auto-updates need **streaming** (`streaming: true` + `eventsource`) or **polling** (`refreshFeatures` on a timer). Cache / `staleTTL` only matter when a fetch already runs — they don’t wake a sitting process.

### Other bugs around the same topic
Same class of pain shows up under other titles, not only #4889:

- https://github.com/growthbook/growthbook/issues/4031 — flags don’t update unless restart / streaming not working  
- https://github.com/growthbook/growthbook/issues/960 — autoRefresh not working as expected (Node)  
- Related streaming/proxy timing: https://github.com/growthbook/growthbook/issues/1457, https://github.com/growthbook/growthbook/issues/1965  

### How I learned what `backgroundSync` does
- Code path: `startStreaming` / `startAutoRefresh` / `refreshFeatures`  
- Changelog / README: `backgroundSync: false` disables opening a streaming connection; it isn’t “start syncing”  
- Node docs already say default is fetch-once, then streaming or polling — but Caching / the name `backgroundSync` make it easy to grab the wrong knob  

**Streaming** = get told features changed (push), which often leads to a refetch.  
**Cache refresh** = refetch the payload when something asks (`refreshFeatures`, poller, etc.).  
Related end state, different jobs.

### Why docs only
With the config in #4889, staying stale is expected behavior, not a proven broken refresh engine. I didn’t show that correctly configured streaming still fails. The failure mode is confusion about which settings turn auto-update on — so clarifying the Node docs is the smallest fix that matches what I found.

### DIFF
--- a/docs/docs/lib/node.mdx
+++ b/docs/docs/lib/node.mdx
@@ -218,10 +218,20 @@ const gbClient = new GrowthBookClient().initSync({
 
 By default, the GrowthBookClient will only fetch features once during initialization. This works great for short-running processes (e.g. serverless functions).
 
-For long-running Node.js processes, there are 2 main approaches to keeping feature defintiions up-to-date.
+For long-running Node.js processes, there are 2 main approaches to keeping feature definitions up-to-date.
 
 1. Streaming
 2. Polling
+
+:::note Streaming vs cache refresh vs `backgroundSync`
+These are easy to mix up:
+
+- **Streaming** — GrowthBook (Cloud or Proxy) pushes a notification when features change. On Node you must opt in with `init({ streaming: true })` and an `EventSource` polyfill (below).
+- **Cache refresh** — the SDK refetches the features payload (for example by calling `refreshFeatures()`). Cache settings like `staleTTL` only apply when a fetch already runs; they do **not** start fetches on their own.
+- **`backgroundSync`** — only controls whether a streaming connection is **allowed** (default `true`). Setting `cacheSettings: { backgroundSync: true }` alone does **not** enable auto-refresh on Node.
+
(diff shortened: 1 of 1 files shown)


EmmS21:
@greptileai

PR #6600, opened 2026-08-10
feat(agent): add a learnings skill and route experiment skills to it

Auz:
Agents could analyze an experiment but had no idea the Learnings corpus existed — they couldn't check what the team already concluded, and nothing was ever recorded back into it.

Adds a flat `learnings` skill (matching the product-analytics precedent; the surface is small enough that a domain router with leaves would be scaffolding for its own sake) covering search, list, read, and record.

Cross-references where they actually matter: experiment-design consults prior Learnings before designing so a settled question isn't re-tested, and experiment-analyze offers to record a result that generalizes.

Documents only what /api/v1 actually exposes — CRUD plus semantic search. The AI find and refresh flows are internal routes, so the skill tells the agent to point users at the app rather than attempt a call that would 404.


Auz:
@greptileai review

PR #6602, opened 2026-08-10
fix(sdk-connections): Update instructions to send Managed Warehouse events to the correct region's ingestor

tzjames:
### Features and Changes

Two related fixes to the SDK Connections setup instructions for the "GrowthBook Managed Warehouse" event tracker (JS/React):

**1. Wrong ingestor region.** The generated snippet always called `growthbookTrackingPlugin()` with no arguments (and the no-code `<script>` tag had no `data-event-ingestor-host`), so it silently defaulted to the `us-east-1` ingestor (`https://us1.gb-ingest.com`) - even for orgs whose Managed Warehouse is provisioned in `eu-west-1`. Per the docs added when EU region support shipped (#6471), that means events copy-pasted from the generated snippet never reach an EU-region warehouse.

Now looks up the org's `growthbook_clickhouse` datasource and, when its `settings.region` is `eu-west-1`, passes the correct `ingestorHost` (`https://eu-west-1.gb-ingest.com`) into the generated JS/React snippet and the correct `data-event-ingestor-host` attribute into the no-code script tag. `us-east-1` (the default) is unaffected - output is byte-identical to before.

- `services/dataRegions.ts`: added `getEventIngestorHost(region)` mapping region -> ingestor host.
- `GrowthBookSetupCodeSnippet.tsx`: new `managedWarehouseRegion` prop, threaded into the JS code generator.
- `InstallationCodeSnippet.tsx`: same, for the no-code script tag.
- `VerifyConnectionPage.tsx` / `CodeSnippetModal.tsx`: look up the org's Managed Warehouse datasource region via `useDefinitions()` and pass it down.

**2. Silent no-op on old SDK versions.** `growthbookTrackingPlugin` requires SDK version >= 1.4.0 (that's when the plugin system shipped). When a connection's recorded SDK version is older than that, the snippet generator fell back to a generic `trackingCallback` with a `console.log` placeholder - i.e. it silently never sent events anywhere, with no indication anything was wrong.

Now shows a warning callout instead, telling the user to update the SDK version or send events directly via the Ingestion API, with a link to the docs.

- `GrowthBookSetupCodeSnippet.tsx`: added `supportsGrowthbookTrackingPlugin(version)` and a warning `Callout` in the JS/React branches when the tracker is `"growthbook"` but the version doesn't support the plugin.

### Dependencies

None.

### Testing

1. On an org with a Managed Warehouse datasource provisioned in `eu-west-1`, open SDK Connections -> a JS or React connection -> select "GrowthBook Managed Warehouse" as the event tracker.
   - Confirm the generated snippet includes `growthbookTrackingPlugin({ ingestorHost: "https://eu-west-1.gb-ingest.com" })`.
   - Confirm the no-code script tag includes `data-event-ingestor-host="https://eu-west-1.gb-ingest.com"`.
2. For `us-east-1` orgs (or orgs without a Managed Warehouse), confirm the snippet is unchanged.
3. On a connection with an SDK version below 1.4.0, select "GrowthBook Managed Warehouse" and confirm a warning callout appears (JS and React) linking to the Ingestion API docs, instead of a silent no-op `console.log` placeholder.
4. On a connection with SDK version >= 1.4.0, confirm the existing `logEvent` examples still render as before.

Verified with `pnpm --filter front-end type-check` and `eslint` on the touched files.

### Screenshots

N/A (generated code snippet text and a warning callout only, no broader UI change).


packages/front-end/components/SyntaxHighlighting/Snippets/GrowthBookSetupCodeSnippet.tsx
```diff
@@ -146,21 +155,23 @@ window.growthbook_config.trackingCallback = (experiment, result) => {
...
-            If you want to use GrowthBook for experiments (and metrics), you
-            will need to log events you care about. Read more about our{" "}
-            <DocLink useRadix={false} docSection="managedWarehouseTracking">
-              managed warehouse tracking
-            </DocLink>
-            . Here are some examples:
-            <Code
-              language="javascript"
-              code={`
+        {eventTracker === "growthbook" &&
```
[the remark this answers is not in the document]
tzjames: We stopped allowing people to set up new multi language sdk connections a long time ago Oct 2023, before the managed warehouse.  So I don't think this is a realistic issue.  People could have set up a new managed warehouse data source and be on a super old sdk mutli language sdk connection, but in practice I don't think that has happened.  There is no good way to know their version for multi-language connections so I think it is fair to assume they are on one >1.4.0.

packages/front-end/components/SyntaxHighlighting/Snippets/GrowthBookSetupCodeSnippet.tsx
```diff
@@ -9,12 +9,19 @@ import EventTrackerSelector, {
...
 
 function indentLines(code: string, indent: number | string = 2) {
   const spaces = typeof indent === "string" ? indent : " ".repeat(indent);
   return code.split("\n").join("\n" + spaces);
 }
 
+// growthbookTrackingPlugin (and the plugin system generally) requires this version or later
+function supportsGrowthbookTrackingPlugin(version?: string): boolean {
+  return paddedVersionString(version) >= paddedVersionString("1.4.0");
+}
```
jdorn, line 29: This version is only for JavaScript based SDKs, but this file is for all SDKs. We should probably be using our SDK capabilities system instead of hard coding versions here.

tzjames: Good point.  Done.

packages/front-end/components/SyntaxHighlighting/Snippets/GrowthBookSetupCodeSnippet.tsx
```diff
@@ -9,12 +10,21 @@ import EventTrackerSelector, {
...
 function indentLines(code: string, indent: number | string = 2) {
   const spaces = typeof indent === "string" ? indent : " ".repeat(indent);
   return code.split("\n").join("\n" + spaces);
 }
 
+function supportsGrowthbookTrackingPlugin(
+  language: SDKLanguage,
+  version?: string,
+): boolean {
+  return getSDKCapabilities(language, version).includes("trackingPlugin");
```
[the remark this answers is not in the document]
tzjames: I'm handling node in https://github.com/growthbook/growthbook/pull/6610#pullrequestreview-4906442170

packages/front-end/services/dataRegions.ts
```diff
@@ -20,3 +24,32 @@ export function useDataRegionOptions(): { label: string; value: DataRegion }[] {
...
+  const managedWarehouse = datasources.find(
+    (d): d is GrowthbookClickhouseDataSourceWithParams =>
+      d.type === "growthbook_clickhouse",
+  );
+  if (managedWarehouse) {
+    return managedWarehouse.settings?.region;
+  }
+
+  return datasources.find((d) => d.eventForwarderConfig)?.eventForwarderConfig
+    ?.region;
```
Kevin-Chant: This will lead to an inaccurate region for eu users without permission to read their org's datasources. We'll need to decouple the source of truth for ingestor region for the org from the `datasources` part of useDefinitions

packages/front-end/components/SyntaxHighlighting/Snippets/GrowthBookSetupCodeSnippet.tsx
```diff
@@ -1471,7 +1511,9 @@ const gb = new GrowthBook({
   enableDevMode: true,${!useInit ? `\n  subscribeToChanges: true,` : ""}
   plugins: [
     autoAttributesPlugin(),
-    growthbookTrackingPlugin()
+    growthbookTrackingPlugin(${
+      ingestorHost ? `{ ingestorHost: ${JSON.stringify(ingestorHost)} }` : ""
+    })
```
Kevin-Chant, line 1525: This will render us-east-1 until the definitions endpoint returns. We should probably gate this component on `ready` from `useDefinitions`

packages/front-end/services/dataRegions.ts
```diff
@@ -20,3 +24,32 @@ export function useDataRegionOptions(): { label: string; value: DataRegion }[] {
...
+const EVENT_INGESTOR_HOSTS: Record<DataRegion, string> = {
+  "us-east-1": "https://us1.gb-ingest.com",
+  "eu-west-1": "https://eu-west-1.gb-ingest.com",
+};
+
+// The SDK's growthbookTrackingPlugin defaults to the us-east-1 host, so
+// callers only need to pass this along explicitly for non-default regions.
+export function getEventIngestorHost(region: DataRegion): string {
+  return EVENT_INGESTOR_HOSTS[region];
+}
```
Kevin-Chant, line 34: There's an env var `INGESTOR_HOST` that's meant to be used for overrides (I assume for dev envs?) and a `getIngestorHost()` to read it. We should clean that up if it's unused or merge it into this logic if it's still relevant

Kevin-Chant (changes requested):
A few issues that could lead to the wrong region rendered in the snippet but otherwise looks pretty good

packages/front-end/components/SyntaxHighlighting/Snippets/GrowthBookSetupCodeSnippet.tsx
```diff
@@ -169,9 +183,18 @@ gb.logEvent("Button Click", {
...
 });
               `}
-            />
-          </>
-        )}
+              />
+            </>
+          ) : (
+            <Callout status="warning" mt="3">
+              Upgrade this SDK Connection to SDK version 1.4.0 or later to send
```
Kevin-Chant: Nit: could be programmatic instead of hardcoding 1.4.0 (`getSDKCapabilityVersion(language, "trackingPlugin")`)

Kevin-Chant: Actually that's an extra catch - need to run `pnpm generate-sdk-report` to regenerate `CAPABILITIES.md` for the tracking plugin row to populate

packages/front-end/services/dataRegions.ts
tzjames: That is for where our dogfooding Growthbook instance sends data.   This is for where customers SDK should send events, which is separate.

Kevin-Chant: Ah gotcha, then disregard

Kevin-Chant (approved):

PR #6608, opened 2026-08-11
Prefer us-east-1 alias

tzjames:
### Features and Changes
Swaps `us1.gb-ingest.com` with  `us-east-1.gb-ingest.com` as the default ingestor url. Now that we have eu-west-1.gb-ingest.com, it standardizes the url format.  It also makes it clear the exact region.  If we add other ingestors it will make it clearer where the ingestor is located.

### Dependencies
https://github.com/growthbook/terraform/pull/218 needs to land first!


Kevin-Chant (approved):
I applied the TF changes and checked the new URL resolves. LGTM

PR #6613, opened 2026-08-11
fix: Hide Start Time field in Analysis Settings for draft experiments

lukesonnet:
### Features and Changes

Hides the Start Time field in the Analysis Settings modal when an experiment is in draft status. This field was incorrectly displayed for draft experiments even though it has no effect until the experiment starts running, it also I think wasn't saving on back-end validation.

### Testing

Manually confirmed.

**Before in draft exp**
Screenshot 2026-08-11 at 11 27 26 AM

**After in same draft exp**
Screenshot 2026-08-11 at 11 27 52 AM

### DIFF
--- a/packages/front-end/components/Experiment/AnalysisForm.tsx
+++ b/packages/front-end/components/Experiment/AnalysisForm.tsx
@@ -761,36 +761,40 @@ const AnalysisForm: FC<{
             </small>
           </div>
         )}
-        {!!phaseObj && editDates && !isBandit && !isHoldout && (
-          <div className="row">
-            <div className="col">
-              <DatePicker
-                label="Start Time (UTC)"
-                helpText="Only include users who entered the experiment on or after this date"
-                date={form.watch("dateStarted")}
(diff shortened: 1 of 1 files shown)


packages/front-end/components/Experiment/AnalysisForm.tsx
```diff
@@ -761,36 +761,40 @@ const AnalysisForm: FC<{
...
-                setDate={(v) => {
-                  form.setValue("dateStarted", v ? datetime(v) : "");
-                }}
-                scheduleEndDate={form.watch("dateEnded")}
-                disableAfter={form.watch("dateEnded") || undefined}
-              />
-            </div>
-            {experiment.status === "stopped" && (
+        {!!phaseObj &&
+          editDates &&
```
lukesonnet, line 765: We could just push the logic upstream into `editDates` but it feels more correct to just block it here when in draft, it makes no sense to me.

msamper (approved):

PR #6616, opened 2026-08-11
Fix: dashboard metric filter showing all metrics in a group when selecting individual metric

lukesonnet:
### Features and Changes

When a user selected an individual metric that was part of a metric group in the dashboard metrics filter, the dashboard was incorrectly showing all metrics in that group instead of just the selected metric.

**Fix:**
Ensure only the specifically selected metrics (or all metrics from explicitly selected groups) are shown. This handles the following cases correctly:
1. User selects an individual metric → shows only that metric  
2. User selects a metric group → shows all metrics in the group
3. User selects both → shows all selected metrics plus all metrics from selected groups

**Testing:**
Now I can pick a child metric that only exists in a group without showing the other group metrics.

Screenshot 2026-08-24 at 1 51 01 PM
Screenshot 2026-08-24 at 1 50 54 PM


ahdriel (approved):
Fix makes sense to me!

There's a potential opportunity here to split some of the logic that is duplicated between these 2 hooks into a 3rd one that they both use or something, so we don't need to always remember to implement the fix on both views.

PR #6626, opened 2026-08-12
Automate sdk-versions: dispatch receiver + capabilities + verify

madhuchavva:
## Summary

Fallback to dispatcher event model to automate version sync. 

  -  Sender: Each SDK (**growthbook/sdk-version-sync** action) on a successful release.
  - Receiver: upon event dispatch, opens a PR for that SDK, regenerates sdk info and docs.

## Notes
- **`SDKInfo.ts` is regenerated** — via the pre-commit hook. `CAPABILITIES.md` is intentionally untouched (not in the hook; already drifted, so gating on it would false-fail every PR).
- **Security** (comment-command has write): `authorize` job requires the commenter be OWNER/MEMBER/COLLABORATOR (`author_association`) and the PR be same-repo, so fork code never runs with the write token. Least-privilege token; only the write job gets `contents: write`.
- **Capabilities can't be derived** automatically — Comment `/capabilities streaming, redirects` on the PR to manage them.

## What's included
- **`.github/workflows/sync-sdk-version.yml`** — receiver. On `repository_dispatch` (`update-sdk-version`) or manual `workflow_dispatch`, prepends a version-only entry to `sdk-versions/<sdk>.json` (idempotent; no-op if not newer), regenerates `SDKInfo.ts` via `gen-sdk-resources-for-docs` (the same command the pre-commit hook runs), and opens a `sdk-version/<sdk>-<x.y.z>` PR.
- **`.github/workflows/sdk-version-capabilities.yml`** — ChatOps. Comment `/capabilities streaming, redirects` → validates each token against the `SDKCapability` union in `types.ts` (unknown → rejected, no change), sets them on the entry, regenerates `SDKInfo.ts`, commits back. Empty → version-only.
- **`.github/workflows/verify-sdk-versions.yml`** — required check: every capability in every `sdk-versions/*.json` is a real `SDKCapability`.
- **`.github/scripts/apply-sdk-capabilities.sh`** — the validate-and-edit logic (unit-tested).

## Validation
The whole flow — receiver, all `/capabilities` cases (empty / unknown / valid), the authz gate, and idempotency — was proven end-to-end in **growthbook/examples** with a stubbed generator. The only piece not exercisable there is the real `gen-sdk-resources-for-docs` build, whose generators were verified locally.

### DIFF
--- a/.github/scripts/apply-sdk-capabilities.sh
+++ b/.github/scripts/apply-sdk-capabilities.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# Validate a comma-separated capability list against the SDKCapability union in
+# packages/shared/src/sdk-versioning/types.ts and set it on a specific version
+# entry in that SDK's sdk-versions JSON.
+#
+# Usage: apply-sdk-capabilities.sh <repo_root> <sdk> <version> <capabilities_csv>
+#
+# Exit codes:
+#   0  capabilities applied (edited the JSON)
+#   2  empty capability list -> nothing to do (version-only)
+#   3  one or more capabilities are not in types.ts (no changes made)
+#   1  usage / structural error
+#
+# On success (0) and no-op (2) it prints "applied=<csv>" so the caller can echo
+# it back. On rejection (3) it prints "invalid=<csv>".
+set -euo pipefail
+
+ROOT="${1:?repo_root}"
+SDK="${2:?sdk}"
+VERSION="${3:?version}"
+CSV="${4-}"
+
+TYPES="$ROOT/packages/shared/src/sdk-versioning/types.ts"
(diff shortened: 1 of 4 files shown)


madhuchavva:
@greptileai

madhuchavva:
Closing in favor of a pull-based replacement:

- A scheduled agent now polls every SDK's registry/tags daily, diffs against `sdk-versions/*.json`, and opens a sync PR — including evidence-backed capability proposals for human review, which this dispatch model couldn't derive (first output: #6641).
- The deterministic guard from this PR survives as its own PR: #6639 (`verify-sdk-versions.yml`).

The pull model removes the need for per-SDK-repo sender adoption (`growthbook/sdk-version-sync`) and the `/capabilities` ChatOps flow. Worst-case sync lag is the polling cadence (24h), which matches review latency in practice.

PR #6628, opened 2026-08-12
Lj/exp funnels docs v1

lillialjackson:
### Features and Changes

Adds docs for the experiment funnels v1 release
adds the funnel type to metrics list view filters

see:
/app/metrics#funnel-metrics
/app/metrics/funnel
/app/experiment-results#funnel-metric-results

### Dependencies

none

### Testing

initial edit

### Screenshots

Screenshot 2026-08-12 at 2 35 30 PM
Screenshot 2026-08-12 at 2 35 47 PM
Screenshot 2026-08-12 at 2 36 01 PM

Screenshot 2026-08-12 at 2 38 47 PM

### DIFF
--- a/docs/docs/metrics/funnel-metrics.mdx
+++ b/docs/docs/metrics/funnel-metrics.mdx
@@ -76,7 +76,7 @@ Take a three-step funnel where a unit is exposed at 10:00:
 | 2. Added to cart  | 2 hours           | Step 1 at 10:05   | Between 10:05 and 12:05. Say it lands at 11:30 |
 | 3. Purchased      | 1 day             | Step 2 at 11:30   | Between 11:30 and 11:30 the next day           |
 
-A step with no conversion window has no upper bound. It counts as long as it happens after its anchor and inside the metric window.
+A step with no conversion window has no upper bound. It counts as long as it happens after the previous non-optional step and inside the metric window.
 
 ### Optional steps
 
--- a/docs/docs/metrics/funnel-metrics.mdx
+++ b/docs/docs/metrics/funnel-metrics.mdx
@@ -87,7 +87,7 @@ In the example above, if **Added to cart** were optional and the unit never adde
 Use optional steps for parts of a journey you want to measure but that are not required to continue.
 
 :::note Partial data
-When the metric uses a conversion window and every step has its own conversion window, GrowthBook bounds the time a unit needs to complete the funnel at the sum of the step windows, capped by the metric window. If any step has no conversion window, the metric window alone is used. This only matters if the experiment has **Skip Partial Data** enabled, where it controls which recent exposures are excluded as not having had time to convert.
+Metric level conversion windows are a hard cap on the completion of all steps. Data that lands after a conversion window has closed are not considered. If the experiment has **Skip Partial Data** enabled, where it controls which recent exposures are excluded as not having had time to convert, time to convert is measured as the sum of each steps conversion windows, capped by the metric window.
 :::
 
 ## Where funnel metrics can be used
--- a/docs/docs/metrics/funnel-metrics.mdx
+++ b/docs/docs/metrics/funnel-metrics.mdx
@@ -15,7 +15,7 @@ Funnel Metrics measure the share of exposed units who complete a series of steps
 A funnel metric is a [Fact Table](/app/metrics) metric like any other. Instead of a numerator, you define an ordered list of steps, and GrowthBook turns each step into its own conversion rate.
 
 :::note Two things named "funnel"
-This page covers the **funnel metric type** built on Fact Tables. It is unrelated to the [legacy denominator behavior](/app/metrics/legacy#funnel-metrics), where a binomial denominator metric acted as a filter on the numerator. If you built that pattern with legacy metrics, a funnel metric is the direct replacement.
+This page covers the **funnel metric type** built on Fact Tables. It is unrelated to the [legacy denominator behavior](/app/metrics/legacy#funnel-metrics), where a binomial denominator metric acted as a filter on the numerator. Those legacy metrics should be replaced by these funnel metrics, although there are some slight differences in behavior.
 
 The [Funnel explorer](/app/product-analytics/explorer#funnels) in Product Analytics answers similar questions outside of an experiment. Funnel metrics are for experiment analysis; the explorer is for ad-hoc exploration.
 :::
--- a/docs/docs/metrics/funnel-metrics.mdx
+++ b/docs/docs/metrics/funnel-metrics.mdx
@@ -12,7 +12,7 @@ import CommercialFeature from '@site/src/components/CommercialFeature';
 
 Funnel Metrics measure the share of exposed units who complete a series of steps in order. One metric gives you both the end-to-end completion rate and a per-step breakdown of where units drop off, so a single row in your experiment results expands into the whole journey.
 
-A funnel metric is a [Fact Table](/app/metrics) metric like any other. Instead of a numerator, you define an ordered list of steps, and GrowthBook turns each step into its own conversion rate.
+A funnel metric is a [Fact Table](/app/metrics) metric like any other. You define an ordered list of steps and GrowthBook turns each step into its own conversion rate.
 
 :::note Two things named "funnel"
 This page covers the **funnel metric type** built on Fact Tables. It is unrelated to the [legacy denominator behavior](/app/metrics/legacy#funnel-metrics), where a binomial denominator metric acted as a filter on the numerator. Those legacy metrics should be replaced by these funnel metrics, although there are some slight differences in behavior.
--- a/docs/docs/metrics/funnel-metrics.mdx
+++ b/docs/docs/metrics/funnel-metrics.mdx
@@ -80,7 +80,7 @@ A step with no conversion window has no upper bound. It counts as long as it hap
 
 ### Optional steps
 
-An optional step is measured like any other step: you still get its conversion rate, its lift, and its own row in the results. What changes is that it does not anchor the steps after it.
+An optional step is measured like any other step: you still get its conversion rate, its lift, and its own row in the results. What changes is that it does not anchor or filter the steps after it.
 
 In the example above, if **Added to cart** were optional and the unit never added to cart, **Purchased** would be measured within 1 day of **Viewed product** instead. Units who did add to cart are still anchored on their view, not their add to cart.
 
--- a/docs/docs/metrics/funnel-metrics.mdx
+++ b/docs/docs/metrics/funnel-metrics.mdx
@@ -94,17 +94,17 @@ When the metric uses a conversion window and every step has its own conversion w
 
 Funnel metrics work as **goal**, **secondary**, and **guardrail** metrics on experiments and reports, including with dimensions, date results, and [Experiment Dashboards](/app/experiment-dashboards).
 
-The following are not supported:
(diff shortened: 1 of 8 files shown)


docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+import CommercialFeature from '@site/src/components/CommercialFeature';
+
+# Funnel Metrics
+
+<CommercialFeature feature="funnel-metrics" />
+
+Funnel Metrics measure the share of exposed units who complete a series of steps in order. One metric gives you both the end-to-end completion rate and a per-step breakdown of where units drop off, so a single row in your experiment results expands into the whole journey.
+
+A funnel metric is a [Fact Table](/app/metrics) metric like any other. Instead of a numerator, you define an ordered list of steps, and GrowthBook turns each step into its own conversion rate.
```
lukesonnet:
```suggestion
A funnel metric is a [Fact Table](/app/metrics) metric like any other. You define an ordered list of steps and GrowthBook turns each step into its own conversion rate.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+:::note Two things named "funnel"
+This page covers the **funnel metric type** built on Fact Tables. It is unrelated to the [legacy denominator behavior](/app/metrics/legacy#funnel-metrics), where a binomial denominator metric acted as a filter on the numerator. If you built that pattern with legacy metrics, a funnel metric is the direct replacement.
```
lukesonnet:
```suggestion
This page covers the **funnel metric type** built on Fact Tables. It is unrelated to the [legacy denominator behavior](/app/metrics/legacy#funnel-metrics), where a binomial denominator metric acted as a filter on the numerator. Those legacy metrics should be replaced by these funnel metrics, although there are some slight differences in behavior.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+The [Funnel explorer](/app/product-analytics/explorer#funnels) in Product Analytics answers similar questions outside of an experiment. Funnel metrics are for experiment analysis; the explorer is for ad-hoc exploration.
+:::
+
+## Every step shares one denominator
+
+This is the most important thing to know about funnel metrics: **each step is measured against all exposed units, not just the units who completed the previous step.**
+
+If 1,000 units are exposed to a variation and 400 view a product, 120 add to cart, and 60 purchase, the three steps report 40%, 12%, and 6%. They do not report 40%, 30%, and 50%.
+
+That choice keeps every step statistically comparable to every other metric in your experiment. Each step is an ordinary proportion metric over the same denominator, so its lift, confidence interval, and chance to win mean exactly what they mean everywhere else. Step-to-step conversion rates are still available in the [funnel chart](/app/experiment-results#funnel-metric-results) in the metric drilldown.
```
lukesonnet:
```suggestion
That choice is critical for deriving the correct interpretation from your results. Using the conversion total from the previous step as the denominator for your next step will mislead most experimenters, because it carries forward the indirect effect of the last step into the current step's analysis.

**Example: why step-to-step conversion rates can mislead**

Suppose you run an experiment on a two-step funnel: Add to Cart -> Purchase, with 1,000 users in each variation. Imagine we know the true effect: the treatment brings in many new, lower-intent cart-adders, but checkout rates are unchanged by this experiment (low-intent users still checkout at the same rates as before and high-intent users still checkout at the same rates as before.


| Metric                        | Control     | Treatment   |
| ----------------------------- | ----------- | ----------- |
| Users assigned                | 1,000       | 1,000       |
| Added to cart                 | 200 (20%)   | 400 (40%)   |
| Purchased                     | 100         | 140         |
| Cart → Purchase rate (conditional)          | 50%         | 35%         |
| Assignment → Purchase rate (unconditional) | 10%     | 14%     |

The step-to-step, conditional conversion rate (Cart → Purchase) fell from 50% to 35% so it looks like the treatment hurt checkout. But nothing about checkout changed. The control arm's carts were 200 high-intent shoppers; the treatment arm added 200 casual browsers who convert at ~20%, dragging the average down. Reading the 50% -> 35% drop as "checkout got worse" conflates the treatment's effect on who reaches the cart with its effect on checkout itself.

The last row is what GrowthBook would represent as the second step conversion rate and it tells the true story in this case: purchases per assigned user rose from 10% to 14%. That comparison is safe because randomization guarantees the two groups of 1,000 are comparable, comparing the groups of cart-adders are not. This is an example of the well known "post-treatment bias" problem, and is akin to what happens when sample ratio mismatch is a sign of assignment bias.

Step-to-step conversion rates are still available in the [funnel chart](/app/experiment-results#funnel-metric-results) in the metric drilldown, but we do not currently provide them as a top-level metric with the confidence intervals and statistics used for decision making. Those are reserved for the safer, unconditional conversion rates.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+- **Row filters**: What counts as completing this step, such as `event_name = add_to_cart`. These work the same as [Row Filters](/app/metrics#row-filters) on any other metric, including saved filters and custom SQL expressions.
+- **Optional step**: Units who skip this step can still complete later steps. See [Optional steps](#optional-steps).
+- **Conversion window**: How long the unit has to reach this step. Enter a number and a unit of `minutes`, `hours`, `days`, or `weeks`.
+
+### Metric-level settings
+
+Funnel metrics support [Metric Windows](/app/metrics#metric-windows), a [metric delay](/app/metrics#metric-delay), a target MDE, and [Bayesian priors](/app/metrics#bayesian-priors). The metric window bounds which events are eligible at all; step conversion windows then apply inside that envelope.
+
+Capping, quantile settings, and [Metric Slices](/app/metrics#metric-slices) are not supported for funnel metrics, so those sections are hidden when you choose the funnel type. The **Metric Goal** setting is also hidden, so funnel metrics created in the UI treat an increase as an improvement.
```
lukesonnet:
```suggestion
[Metric Slices](/app/metrics#metric-slices) are not supported for funnel metrics.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+A funnel metric is a [Fact Table](/app/metrics) metric like any other. Instead of a numerator, you define an ordered list of steps, and GrowthBook turns each step into its own conversion rate.
+
+:::note Two things named "funnel"
+This page covers the **funnel metric type** built on Fact Tables. It is unrelated to the [legacy denominator behavior](/app/metrics/legacy#funnel-metrics), where a binomial denominator metric acted as a filter on the numerator. If you built that pattern with legacy metrics, a funnel metric is the direct replacement.
+
+The [Funnel explorer](/app/product-analytics/explorer#funnels) in Product Analytics answers similar questions outside of an experiment. Funnel metrics are for experiment analysis; the explorer is for ad-hoc exploration.
+:::
+
+## Every step shares one denominator
```
lukesonnet: I would move this section below creating one and How Steps Are Resolved

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+Take a three-step funnel where a unit is exposed at 10:00:
+
+| Step              | Conversion window | Anchored on       | Counts if the event happens                    |
+| ----------------- | ----------------- | ----------------- | ---------------------------------------------- |
+| 1. Viewed product | 1 hour            | Exposure at 10:00 | Between 10:00 and 11:00. Say it lands at 10:05 |
+| 2. Added to cart  | 2 hours           | Step 1 at 10:05   | Between 10:05 and 12:05. Say it lands at 11:30 |
+| 3. Purchased      | 1 day             | Step 2 at 11:30   | Between 11:30 and 11:30 the next day           |
+
+A step with no conversion window has no upper bound. It counts as long as it happens after its anchor and inside the metric window.
```
lukesonnet:
```suggestion
A step with no conversion window has no upper bound. It counts as long as it happens after the previous non-optional step and inside the metric window.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+### Optional steps
+
+An optional step is measured like any other step: you still get its conversion rate, its lift, and its own row in the results. What changes is that it does not anchor the steps after it.
```
lukesonnet:
```suggestion
An optional step is measured like any other step: you still get its conversion rate, its lift, and its own row in the results. What changes is that it does not anchor or filter the steps after it.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+In the example above, if **Added to cart** were optional and the unit never added to cart, **Purchased** would be measured within 1 day of **Viewed product** instead. Units who did add to cart are still anchored on their view, not their add to cart.
+
+Use optional steps for parts of a journey you want to measure but that are not required to continue.
+
+:::note Partial data
+When the metric uses a conversion window and every step has its own conversion window, GrowthBook bounds the time a unit needs to complete the funnel at the sum of the step windows, capped by the metric window. If any step has no conversion window, the metric window alone is used. This only matters if the experiment has **Skip Partial Data** enabled, where it controls which recent exposures are excluded as not having had time to convert.
```
lukesonnet:
```suggestion
Metric level conversion windows are a hard cap on the completion of all steps. Data that lands after a conversion window has closed are not considered. If the experiment has **Skip Partial Data** enabled, where it controls which recent exposures are excluded as not having had time to convert, time to convert is measured as the sum of each steps conversion windows, capped by the metric window.
```

docs/docs/metrics/funnel-metrics.mdx
```diff
@@ -0,0 +1,185 @@
...
+MySQL, MSSQL, and Vertica cannot run funnel queries. Adding a funnel metric to an experiment on one of these data sources fails the query with an unsupported data source error naming the funnel metrics involved. Mixpanel and Google Analytics do not support Fact Tables at all.
+:::
+
+## Funnel metrics in experiment results
+
+In the results table, a funnel metric is one row you can expand into a row per step. Each step row shows that step's conversion rate, lift, and significance against the shared denominator, and steps always stay in funnel order no matter how you sort. Clicking the metric opens the drilldown with a **Funnel** tab that charts each variation's funnel side by side.
+
+See [Funnel Metric Results](/app/experiment-results#funnel-metric-results) for the full walkthrough.
+
+## Managing funnel metrics with the API
```
lukesonnet, line 141: I think this is not really our pattern to include API docs here. I dont feel strongly though.

lukesonnet (approved):

docs/docs/metrics/funnel-metrics.mdx
lillialjackson: I thought it was a nice add since we are trying to move toward agent readable docs, but I can remove if we prefer to surface using our api in another way

lukesonnet: It's just buried in our API main documentation, so I agree we should have a more user-friendly way to share this information and more and more of these instructions are in our docs, but it does require more updating if anything changes.

PR #6629, opened 2026-08-12
Fix for trailing characters on invite email addresses

Auz:
Inviting users via outlook in particular, can be semicolon separated. This additional character will correctly send an email, but cause the invite to be not join-able as it will never match the email. This fixes this edge case.


jdorn (approved):
Clean up AI comments. Otherwise looks good.

PR #6633, opened 2026-08-13
chore(lint): enforce stricter React key checks

ahdriel:
### Features and Changes

- Enforce more strict `react/jsx-key` checks
- Fix the existing violations

### Testing

- `pnpm lint`

### DIFF
--- a/packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
+++ b/packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
@@ -1,3 +1,4 @@
+import { Fragment } from "react";
 import Collapsible from "react-collapsible";
 import { Box, Flex, Separator, Text, TextField } from "@radix-ui/themes";
 import { PiNetwork, PiTrash } from "react-icons/pi";
@@ -98,10 +99,9 @@ export default function DataVizDimensionPanel({
                 <>
                   {dimensions.map((dimension, index) => {
                     return (
-                      <>
+                      <Fragment key={index}>
                         {index > 0 && <Separator size="4" mt="2" />}
                         <Select
-                          key={index}
                           label={
                             <Flex justify="between" align="center">
                               <Text as="label">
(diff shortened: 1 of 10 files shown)


packages/front-end/components/Layout/Layout.tsx
```diff
@@ -244,8 +244,8 @@ const Layout = (): React.ReactElement => {
                   </button>
                 </li>
                 <ProjectSelector />
-                {navlinks.map((v, i) => (
-                  <SidebarLink {...v} key={i} />
+                {navlinks.map((v) => (
+                  <SidebarLink key={v.href} {...v} />
```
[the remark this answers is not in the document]
ahdriel: I don't think this is true. The duplicated links are only for sub-items, which are rendered in a different loop so they are not siblings. Keys just needs to be unique between siblings.

packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
```diff
@@ -98,10 +99,9 @@ export default function DataVizDimensionPanel({
                 <>
                   {dimensions.map((dimension, index) => {
                     return (
-                      <>
+                      <Fragment key={index}>
```
nhat-growthbook, line 102: Should we use `index` as key here? Can we use `dimension.fieldName` instead or something less mutable?

nhat-growthbook (approved):

packages/front-end/components/DataViz/DataVizDimensionPanel.tsx
ahdriel: `dimension.fieldName` is editable by the user, and therefore is less stable than `index`. using it would mean re-rendering this section for each keystroke, creating a bad UX.

given that at the moment we only render 1 dimension, and we use a fixed array and don't allow adding more, `index` is the identifier for the dimension configuration

in the future, when we make this a list that can be reordered, with more than 1 item, we should find (and probably create/mint) a more stable identifier for it, not tied to its position.

PR #6636, opened 2026-08-13
feat(datasource): add support to adobe experience platform query service

ahdriel:
### Features and Changes

We now support **Adobe Experience Platform Query Service** as a data source.

### Testing

- Unit tests
- Manual test

### Screenshots

Screenshot 2026-08-13 at 11 54 33 AM


msamper (approved):
Code looks good to me. Just unable to test beyond the UI since I don't have an Adobe account to test on. 

Minor but one thing I noticed in the datasource form is that under advanced settings the help text for Max Concurrent Queries mistakingly says that an empty field will result in no limits to the number of queries, but under the hood we actually set the max for this datasource to 1.

Image

ahdriel:
cursor review

PR #6640, opened 2026-08-13
Add metric mentions to AI input

nhat-growthbook:
### Features and Changes
- Add Tiptap Mentions to track Metrics in chat messages
- Add Tiptap commands to pull up a list of available skills 
- Track the stale metrics when users switch data source and handle it on submit correctly
- Add a tooltip to describe the metrics description and skills description on hover

- Closes **[AI Chat with mentions and commands](https://app.notion.com/p/growthbook/Tag-metrics-in-AI-chat-revenue-3a4a857e74a08020bf30cc0718d4b93f?source=copy_link)**

### Dependencies
N/A

### Testing
- [x] Mention `@` a metric in chat and see if it can generate a graph
- [x] In global ai panel, use `/` to tag a specific skill to to things.
- [x] Try to chain multiple Metrics in a same chat turn
- [x] Chain multiple skills in the same turn
- [x] Chain both metrics and skill in the global AI Chat turn
- [x] Mention a metric and then switch data source to the one that doesn't have it. Make sure it's flagged correctly

### Screenshots

Screenshot 2026-08-14 at 14 55 52
Screenshot 2026-08-14 at 14 55 59
Screenshot 2026-08-14 at 14 56 09
Screenshot 2026-08-17 at 10 40 11
Screenshot 2026-08-17 at 10 41 23
Screenshot 2026-08-14 at 14 56 15
Screenshot 2026-08-14 at 14 56 28
Screenshot 2026-08-14 at 14 56 38


packages/front-end/components/Agent/AgentPanel.tsx
```diff
@@ -170,6 +171,11 @@ export default function AgentPanel({
   // Holds the decision to attach to the next outgoing message. Consumed (and
   // cleared) by buildRequestBody so it only rides along with one request.
   const pendingDecisionRef = useRef<ConfirmDecisionBody | null>(null);
+  // Same one-shot pattern as the decision above: set just before sendMessage,
+  // consumed by buildRequestBody so it rides along with exactly one request.
+  const pendingMentionsRef = useRef<AIChatMention[]>([]);
+
+  const mentionItems = useMetricMentionItems();
```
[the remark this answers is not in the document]
nhat-growthbook: Leaving this one unscoped — the general agent is deliberately cross-datasource, so scoping the picker would hide metrics it can genuinely use.

The `datasourceId` here isn't a constraint the way it is in PA Explorer. It rides along as a soft hint (the system prompt calls it "a soft hint about the datasource the user currently has selected", and it's injected as `[Active product-analytics datasource: …]` rather than as a filter). The agent reaches metrics through the REST API across the whole org, and its work isn't datasource-bound at all — most of it is feature flags and experiments.

The scoping in `ExplorerAIChat` exists for a narrower reason: PA runs queries against one datasource, so an out-of-datasource metric would build a chart that can't execute. That failure mode doesn't exist here.

Metric groups are the clearest case: scoping drops them entirely (they have no datasource of their own), but the general agent can use them fine.

packages/front-end/enterprise/components/AIChat/Composer/ChatComposer.tsx
```diff
@@ -116,24 +196,38 @@ function ChatComposer(
...
   const action = isLocalStream
     ? { onClick: onCancel, label: "Cancel generation", disabled: false }
-    : { onClick: onSend, label: "Send message", disabled: !canSend };
+    : { onClick: submit, label: "Send message", disabled: !canSend };
 
   // Each variant keeps its own control so the button dimensions stay exactly as
-  // designed: a 30x30 chip in the narrow panel, a `@/ui/Button` in the wide bar.
+  // designed: a 30x30 chip in the narrow panel, a `@/ui/Button` elsewhere.
   const sendButton = isCompact ? (
     <button
```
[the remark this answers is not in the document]
nhat-growthbook: Leaving as-is. This came up on #6631 and the call there was to keep the existing button dimensions — swapping the compact chip for `@/ui/Button` (or `IconButton`) changes its width, which we don't want. Details in [that thread](https://github.com/growthbook/growthbook/pull/6631#discussion_r3772169422).

Your read is right that the 30x30 is intentional. If a sized icon-only `@/ui/Button` lands later it'd be a clean follow-up, but it's a design-system change we didn't want to bundle in here.

packages/front-end/enterprise/components/ProductAnalytics/AIChat/ExplorerAIChat.tsx
```diff
@@ -44,6 +44,9 @@ export default function ExplorerAIChat() {
   const permissionsUtil = usePermissionsUtil();
   const hasAISuggestions = hasCommercialFeature("ai-suggestions");
   const { draftExploreState } = useExplorerContext();
+  // Scoped to the active datasource — PA queries run against it, so offering a
+  // metric from another datasource would build a chart that can't run.
+  const mentionItems = useMetricMentionItems(draftExploreState.datasource);
```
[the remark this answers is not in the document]
nhat-growthbook: Fix, we'll mark the stale metric in the editor directly so users can tell that it doesn't exist in the current datasource.
At the same time, we'll filter it out from the mentions list once submit since it wouldn't do anything and might cause issue.
The AI would also flag the stale metrics.

packages/front-end/enterprise/components/AIChat/Composer/ChatComposer.tsx
```diff
@@ -116,24 +386,43 @@ function ChatComposer(
...
+    editor?.setEditable(!loading && !disabled);
+  }, [editor, loading, disabled]);
+
+  // Extension options are frozen at creation but the metric list arrives later
+  // from `useDefinitions`, so it lives in the extension's storage — which the
+  // suggestion `items` callback reads through the editor it is handed.
+  useEffect(() => {
+    if (!editor) return;
+    editor.storage[METRIC_MENTION_NAME].items = mentionItems ?? [];
+  }, [editor, mentionItems]);
```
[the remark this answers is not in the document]
nhat-growthbook: Accurate observation, and it's intentional — the mention is kept on purpose, but it is no longer silently authoritative. Handled in three places:

**1. The user sees it.** `MetricMention` runs a ProseMirror plugin that decorates any mention no longer offered by the `@` menu with `data-stale` (`extensions/metricMention.ts`). The chip drops its violet accent for gray and picks up a red `!`, and its hover card says "Not available in the selected Data Source". Staleness is a decoration, not an attribute, so it's a live view of the current metric list rather than something baked into the document — switching back un-marks it. Readiness comes from `useDefinitions().ready`, so a Data Source with genuinely zero metrics correctly strands every mention instead of marking none.

**2. The server decides, not the client.** `AgentConfig.resolveMentions` (`agent-handler.ts`) resolves each mention's own datasource before the message is persisted; PA implements it in `resolveProductAnalyticsMentions`. An id that no longer resolves at all counts as stale too. The wire validator has no `stale` field, so a client can't assert that a metric is usable here. The verdict is persisted on the message, so a reloaded transcript keeps what was true when it was sent.

**3. The agent is told.** `toModelMessages` renders it inline — `Revenue (metric: met_abc, STALE — not in this datasource)` — and appends an instruction to name it, say it's unavailable in the current datasource, and ask the user to re-pick it, explicitly *not* to query it or substitute a similar metric. The PA system prompt describes the same marker.

So the failure mode in the comment — "the agent receives an authoritative metric ID that cannot run" — no longer holds: the id arrives explicitly marked unusable, and the agent's own search tool is datasource-scoped, so it can't re-resolve it either.

On the two alternatives:

- **Clearing the mention on switch** destroys text the user wrote, and the composer is a draft they may be mid-sentence in. Switching Data Sources shouldn't silently edit their message.
- **Filtering at send** is worse than either: the user still sees `@Revenue` in their sent message, but the agent answers as though they never mentioned it. The mismatch is invisible and unexplained.

Keeping the reference and labelling it means the user gets told which metric is the problem, in their own words, by both the UI and the answer.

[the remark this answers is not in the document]
nhat-growthbook: Already handled — this is the same finding as the thread above, and the behaviour is intentional. Short version: the mention is deliberately kept, but it is no longer silently authoritative.

- **The editor marks it.** A ProseMirror decoration flags any mention the `@` menu no longer offers (`extensions/metricMention.ts`); the chip goes gray with a red `!`, carries an `aria-label` saying it isn't available in the selected Data Source, and its card repeats that in words.
- **The server decides it, not the client.** `AgentConfig.resolveMentions` resolves each mention's own datasource before the message is persisted (`resolveProductAnalyticsMentions` for PA). The wire validator has no `stale` field, so a client can't claim a metric is usable here.
- **The agent is told.** `toModelMessages` renders it as `Revenue (metric: met_abc, STALE — not in this datasource)` plus an instruction to name it, say it's unavailable, and ask the user to re-pick it — explicitly not to query it or substitute a similar metric.

So "the server injects it as authoritative context" no longer holds: it arrives explicitly marked unusable, and the PA search tool is datasource-scoped, so the id can't be re-resolved either.

Clearing the node on switch would silently edit a draft the user is mid-sentence in, and dropping it at send would leave `@Revenue` visible in their sent message while the agent answers as though they never mentioned it. Marking it and saying so is the deliberate choice.

packages/back-end/src/enterprise/services/agent-handler.ts
```diff
@@ -415,10 +475,74 @@ function appendUserMessage(
...
+    {
+      role: "tool",
+      id: randomUUID(),
+      ts: Date.now(),
+      content: [
+        {
+          type: "tool-result",
+          toolCallId,
+          toolName: "loadSkill",
+          result: stringifyToolResultForStorage(body),
```
najito: The entire result is stringified? Is there a reason for that? I think I saw that in other places it's more rich with `status`, `name`, `description` and `body`.

packages/front-end/enterprise/components/AIChat/Composer/ChatComposer.tsx
```diff
@@ -116,24 +437,51 @@ function ChatComposer(
...
+    const storage = editor.storage[METRIC_MENTION_NAME] as MentionStorage;
+    storage.items = mentionItems ?? [];
+    storage.ready = mentionItemsReady;
+    // Storage is a plain object, so writing to it changes nothing on screen.
+    // The stale-mention decorations are computed from it, and only a new state
+    // makes ProseMirror ask for them again — hence an empty transaction. It
+    // touches neither the document nor the selection, so `onUpdate` (and the
+    // caret) are untouched, and it is kept out of the undo history.
+    editor.view.dispatch(editor.state.tr.setMeta("addToHistory", false));
+  }, [editor, mentionItems, mentionItemsReady]);
```
najito, line 362: Using `mentionItems` here as a dependency can lead to a lot of firing of this useEffect. Mostly a performance thing than anything since the dispatch isn't super heavy or cause any issues. Just a heads up.

packages/front-end/enterprise/components/AIChat/Composer/TokenHoverCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * give the same result but puts a component inside the contenteditable, which
+ * changes how editing and selection behave for a much smaller payoff.
+ */
+export function readHoveredToken(
+  target: EventTarget | null,
+  box: HTMLElement | null,
+): HoveredToken | null {
+  if (!(target instanceof HTMLElement) || !box) return null;
+
+  const el = target.closest<HTMLElement>("span[data-type]");
```
najito, line 28: `data-type` comes from tiptap itself so this could change in version updates in the future.

packages/front-end/enterprise/components/AIChat/Composer/useSkillCommandItems.ts
```diff
@@ -0,0 +1,48 @@
+import { useMemo } from "react";
+import useApi from "@/hooks/useApi";
+import type { SkillItem } from "./extensions/skillCommand";
+
+interface SkillSummary {
+  name: string;
+  description: string;
+  kind: "domain" | "leaf";
+  group?: string;
+}
```
najito: Is this the same type as the one at `agent.controller.ts:20`? Should the typing be unified in a shared file somewhere?

packages/front-end/enterprise/components/AIChat/Composer/ChatComposer.tsx
nhat-growthbook: Checked — it's stable, so this doesn't fire per render. useMetricMentionItems memoizes the array on [metrics, factMetrics, metricGroups, datasourceId], and those come from DefinitionsContext, each useMemo'd on the SWR data reference. All three callers (ExplorerAIChat, AgentPanel, PA EmptyState) go through that hook, so the prop identity holds.

So it fires on mount, on a real definitions refetch, and on a datasource switch — and each of those it needs to, since the empty transaction is what makes ProseMirror recompute the stale-mention decorations. Leaving as-is.

packages/front-end/enterprise/components/AIChat/Composer/TokenHoverCard.tsx
nhat-growthbook: Yeah the data-type (plus data-id / data-label) comes from @tiptap/extension-mention's renderHTML, not from us. Leaving the read as-is though:

Tiptap round-trips through those same attributes in its own parseHTML (tag: span[data-type="${this.name}"]), so they're load-bearing for serialization rather than incidental markup — changing them would be a breaking change upstream, not patch drift. Both extensions build on the stock Mention node specifically to avoid reimplementing its rendering, and overriding renderHTML to own the attributes means reproducing the suggestion-char/label output for less than it buys. The coupling is also confined to readHoveredToken, so a version bump that did change it is a one-place fix.

packages/front-end/enterprise/components/AIChat/Composer/useSkillCommandItems.ts
nhat-growthbook: Fixed. Thanks

packages/back-end/src/enterprise/services/agent-handler.ts
nhat-growthbook: Yeah good catch, I addressed this, thanks.

packages/back-end/src/enterprise/services/agent-handler.ts
```diff
@@ -92,6 +92,28 @@ export interface AgentConfig<TParams = unknown> {
     emit?: AgentEmit,
   ) => ToolSet;
 
+  /**
```
oelshaikh: nit: remove excessive comments

oelshaikh (approved):
I tested the PR and it works well locally, the high level structure looks good. 

nits on code: lots of AI generated comments that can sometimes make it hard to follow, I'd say reduce those.

packages/front-end/enterprise/components/AIChat/Composer/serialize.ts
```diff
@@ -20,6 +29,98 @@ export function textToContent(text: string): JSONContent {
...
+  return (
+    text
+      .replace(
+        /(^|[ \t])@([ \t]|$)/g,
+        (_match, before: string, after: string) => before || after,
+      )
+      // Leading "/", plus the whitespace it left behind.
+      .replace(/^\/(?=[ \t]|$)[ \t]*/, "")
+      // Trailing "/", plus the whitespace before it.
+      .replace(/[ \t]+\/$/, "")
```
najito: This allows `@` to be put in literally if you type it, press space and then continue. Just FYI

packages/shared/src/validators/ai-conversation.ts
```diff
@@ -84,6 +114,10 @@ const aiChatUserMessageValidator = z
     // Optional soft datasource hint — see AIChatUserMessage in
     // shared/ai-chat.ts. Cap matches the agent router's datasourceId.
     datasourceHint: z.string().max(256).optional(),
+    // Entities the user @-mentioned — see AIChatMention in shared/ai-chat.ts.
+    mentions: aiChatStoredMentionValidator.array().max(20).optional(),
```
najito: Is there a client-side notification for this limit?

najito:
Just a couple of small things but I also think it would be ready to go

packages/shared/src/validators/ai-conversation.ts
nhat-growthbook: Yeah it was arbitrary number I added early on. Removed it now

packages/front-end/enterprise/components/AIChat/Composer/serialize.ts
nhat-growthbook: Oh it gets filtered out when we submit now

nhat-growthbook:
> I tested the PR and it works well locally, the high level structure looks good.
> 
> nits on code: lots of AI generated comments that can sometimes make it hard to follow, I'd say reduce those.

Thanks. I reduced like 500 lines of comments in this PR now.

najito (approved):

PR #6643, opened 2026-08-14
Address security vulnerabilities (2026-08-14)

tzjames:
- Bump next (16.2.6 -> 16.2.11) and js-yaml (4.3.0 -> 4.3.1) direct deps.
- Bump mongoose (6.13.9 -> 6.13.10) using the OSV in-major backport instead of AWS's cross-major "fixed" target (9.7.2); mongodb stays pinned at 4.17.2 in both versions, so no transitive/direct drift.
- Bump pnpm.overrides for fast-uri, brace-expansion, postcss, and js-yaml to their latest in-major patched releases (OSV backports, not AWS's cross-major "fixed" targets).
- Add pnpm.overrides for @opentelemetry/propagator-jaeger, ip-address, and nanoid (transitives whose parents are already at their latest release but still ship the vulnerable nested version).


packages/front-end/package.json
```diff
@@ -80,13 +80,13 @@
...
     "js-cookie": "^3.0.7",
-    "js-yaml": "^4.3.0",
+    "js-yaml": "^4.3.1",
     "json-stringify-pretty-compact": "^3.0.0",
     "json2csv": "^5.0.7",
     "lodash": "^4.17.23",
     "md5": "^2.3.0",
     "minisearch": "^7.2.0",
-    "next": "^16.2.6",
+    "next": "^16.2.11",
```
Kevin-Chant, line 89: The lockfile is shipping `16.3.1`. Semver means it should be safe, but `16.3.0` has a [huge changelog](https://github.com/vercel/next.js/releases/tag/v16.3.0), would it be better to lock to 16.2 explicitly?

Kevin-Chant (approved):
See my note about `next`'s resolved version

packages/front-end/package.json
tzjames: Good find.  We will have to upgrade at some point, though.  I'm going to click around the preview environment a bunch.  If I don't come across anything I'll land it, and if necessary we can roll back.

PR #6656, opened 2026-08-16
Adds a derived, always-present revision.tags to GrowthBook's validate…

dlin-growthbook:
### Features and Changes
The validateFeatureRevision custom hook received the raw revision document alongside the pre-merge live feature, so a hook could not cleanly read the tags a revision would publish. Tags were reachable at revision.metadata.tags, but that envelope is sparse — an absent key means "unchanged", indistinguishable from "cleared" without manually comparing against feature.tags — and it was absent from both the docs and the in-app Test fixture. Add a derived revision.tags holding the effective tag list, mirroring the enrichment config hooks already get via ConfigHookInput. feature.tags still carries the live value, so hooks can diff the two to detect additions:
* tags staged by this revision → revision.tags is the staged value
* tags untouched → revision.tags falls back to the feature's current tags
* tags cleared → revision.tags is [], distinct from untouched
Derived on the original args too, so incrementalChangesOnly suppression keeps matching.

### Testing
Create a validateFeatureRevision hook that throws when revision.tags includes a restricted tag. Confirm it blocks a draft save that adds the tag, allows saves that leave tags untouched, and that the Test panel's prefilled revision now exercises the tag branch. New unit suite covers the derivation, the original-args path, and the publish-gate collector.

### DIFF
--- a/docs/docs/features/custom-hooks.mdx
+++ b/docs/docs/features/custom-hooks.mdx
@@ -140,7 +140,34 @@ if (feature.valueType === "json" && feature.defaultValue === "{}") {
 
 ### validateFeatureRevision
 
-Called whenever a feature revision is about to be created or updated. Receives the full feature and the revision as inputs.
+Called whenever a feature revision is about to be created or updated. Receives:
+
+- `feature` — the **live** feature, i.e. the state before this revision publishes. Use it as the "before" side when checking what a revision changes.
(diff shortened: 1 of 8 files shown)


Kevin-Chant:
Closing based on slack discussions. For posterity, the desired behavior of "diff what will go live" is already part of `validateFeature` so this is more a documentation/discoverability gap

PR #6658, opened 2026-08-17
fix(stats): keep the sign of relative lift when the baseline mean is negative

anna-yn:
### Features and Changes

`gbstats.utils.frequentist_diff` computes the relative difference as `(mean_b - mean_a) / mean_a`. When the control mean is negative, the division flips the sign of the result:

| control → variation | absolute | relative (before) | relative (after) |
|---|---|---|---|
| −10 → −5 (up) | +5 | −50% | +50% |
| −10 → −15 (down) | −5 | +50% | −50% |
| −2 → +3 (crosses zero) | +5 | −250% | +250% |

The relative point estimate is what `EffectMoments` hands to both the frequentist and Bayesian tests, so it isn't just the displayed number: the CI, p-value / chance to win, and the win/lose colouring all invert. Driving `analyze_metric_df` with a control mean of −10 and a variation mean of −5 gives `expected = −0.5, chanceToWin = 0.000` in relative mode on both engines, while absolute mode correctly reports `+5, 1.000`.

This PR divides by `abs(mean_a_unadjusted)` instead, so the sign of the relative lift always matches the sign of the absolute difference. That's the convention already used elsewhere in the engine — `frequentist_variance_relative_cuped_ratio` scales by `np.abs(stat_a.unadjusted_mean)`, and the Bayesian prior conversion between relative and absolute uses `abs(stat_a.unadjusted_mean)` — so this brings the point estimate in line with them. The relative variance is a squared quantity and was already sign-agnostic. Absolute mode is unchanged.

Behavior change to be aware of: any existing metric with a negative control mean will show its relative lift (and win/lose colouring) with the opposite sign after upgrading — that's the corrected sign, but a re-refresh will look like the result flipped. Absolute-mode results are unaffected.

Not addressed here: whether "relative lift" is a meaningful summary at all when the baseline is near or crosses zero (the −250% row). That's a presentation question rather than a sign bug; absolute mode remains the better view for such metrics.

- Related: #401

### Dependencies

None

### Testing

- `tests/test_utils.py`: `frequentist_diff` unit tests — absolute unchanged, positive baseline unchanged, negative baseline keeps direction (incl. crossing zero), and the CUPED `mean_a_unadjusted` path.
- `tests/test_shared_models.py`: `EffectMoments` end-to-end with `SampleMeanStatistic`s — for each of five baseline/variation pairs, the sign of the relative point estimate equals the sign of the absolute one; magnitude check for −10 → −5.
- The four new tests fail on `main` and pass here; the full gbstats suite passes (110). black / flake8 / pyright clean.

### Screenshots

N/A

### DIFF
--- a/packages/front-end/components/Experiment/DateResults.tsx
+++ b/packages/front-end/components/Experiment/DateResults.tsx
@@ -212,7 +212,7 @@ const DateResults: FC<{
               if (differenceType === "absolute") {
                 up = crB - crA;
               } else {
-                up = crA ? (crB - crA) / crA : 0;
+                up = crA ? (crB - crA) / Math.abs(crA) : 0;
               }
             }
 
--- a/packages/front-end/components/Experiment/TabbedPage/CovariateImbalanceTable.tsx
+++ b/packages/front-end/components/Experiment/TabbedPage/CovariateImbalanceTable.tsx
@@ -257,7 +257,7 @@ function CovariateImbalanceTableSection({
                     row.baselineMean !== 0 ? (
                       percentageFormatter.format(
                         (row.variationMean - row.baselineMean) /
-                          row.baselineMean,
+                          Math.abs(row.baselineMean),
                       )
                     ) : (
                       "-"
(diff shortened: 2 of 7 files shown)


lukebrawleysmith (approved):
removed a verbose comment, otherwise looks good to me.  Thanks for your help!

PR #6662, opened 2026-08-17
Keep product analytics AI streams alive

oelshaikh:
## Summary
- Fixes a bug where product analytics agents would see long running queries return after 5 seconds and tell the user there are 0 rows. The agent should instead wait for the query to properly finish and then report back to the user. 
- Keep Product Analytics AI streams active during long warehouse queries with SSE keepalives and persisted conversation heartbeats, including across page reloads and backend instances.
- Poll persisted exploration status using the same backoff as the frontend, stopping on completion, cancellation, or the ten-minute cutoff.

## Test plan
- Manually verified the polling path with a slow, uncached warehouse query.


nhat-growthbook (approved):
LGTM

PR #6664, opened 2026-08-17
fix(timeseries): Use stable id for variation

### LINKED ISSUE #6645: [Bug] Renaming a variation causes metric time-series charts to disappear
### Summary

Changing only a variation’s display name in a running experiment causes the historical time-series charts for every metric to disappear.

For example, renaming 1 to 1X triggers the issue. Reverting to the original name restores the charts.

### Replication steps:

1. Open a running experiment with existing time-series data.
2. Rename a variation without changing its ID, traffic split, or configuration.
3. Save and return to the experiment results.
4. View any metric’s time-series chart.

### Expected Behavior

Renaming a variation should only update its displayed label. Existing time-series data should remain associated with the variation and continue to render.

#### Current Behavior

Historical time-series data is no longer displayed for any metric after the variation is renamed.

### Technical finding

The frontend currently matches stored time-series variations using the mutable variation name:

point.variations.find((v) => v.name === gv.name)

Since stored points also contain a stable variation ID, matching should use that ID instead, with a fallback for older records if required.

Source: https://github.com/growthbook/growthbook/blob/main/packages/front-end/components/Experiment/ExperimentMetricTimeSeriesGraphWrapper.tsx

### Workaround

Change the variation label back to its original value.

ahdriel:
### Features and Changes

Experiment variations used to not have a stable identifier in the past, so we matched them for graphs based on the name. This, obviously, is not the best as it breaks if a rename happens.

Now that we have stable ids for variations, let's use that instead of using the name.

**Note:** we cannot reuse `id` field as this one gets stored in mongo, so to avoid a migration we are adding an explicit `experimentVariationId` field (which is in-memory only, so we can rename this one easily in the future if we need).

### Testing

Unit tests


packages/front-end/services/timeSeriesVariations.ts
```diff
@@ -0,0 +1,26 @@
+import type { MetricTimeSeriesVariation } from "shared/validators";
```
msamper, line 2: I think this file might be misplaced. This should be a services file right?

msamper (approved):

PR #6665, opened 2026-08-17
Stop silently dropping saved-group targeting on API writes

bryce-fitzsimons:
The bulk feature and experiment write endpoints now accept `savedGroups` for rule targeting. They previously ignored it, so a rule sent that way was published with no targeting at all.

`savedGroupTargeting` still works but is no longer documented. These endpoints also return a `warnings` list naming any request fields that were ignored.


packages/back-end/src/util/handler.ts
```diff
@@ -146,8 +240,10 @@ export async function runApiHandler(
     query?: ZodType;
   },
   handler: (req: never) => Promise<unknown>,
+  options?: { surfaceInputWarnings?: boolean },
```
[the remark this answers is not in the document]
bryce-fitzsimons: wired these in. not important though, agent shouldn't be using invalid fields to begin with.

packages/back-end/src/services/experiments.ts
```diff
@@ -4239,10 +4239,12 @@ export function postExperimentApiPayloadToInterface(
...
-      savedGroups: (p.savedGroupTargeting || []).map((s) => ({
-        match: s.matchType,
-        ids: s.savedGroups,
-      })),
+      savedGroups:
+        p.savedGroups ??
+        (p.savedGroupTargeting || []).map((s) => ({
+          match: s.matchType,
+          ids: s.savedGroups,
+        })),
```
jdorn: This exact logic is copy/pasted many times. Abstract this to a helper function

packages/back-end/src/util/handler.ts
```diff
@@ -124,6 +158,102 @@ export type RawApiRequestHandler<
...
+  );
+  return [
+    ...stripped.map((key) => `Unrecognized field \`${key}\` was ignored.`),
+    ...collectAlternateKeys(raw).map(
+      ({ path, use }) =>
+        `\`${path}\` is accepted for compatibility but is not part of the documented contract. Use \`${use}\` instead.`,
+    ),
+  ];
+}
+
```
jdorn: All the changes in this file feel unnecessary and not worth the complexity/line count.

bryce-fitzsimons: Yeah agree. Was trying to be clever with warnings to help consumers find the right payload shape, but not worth the complexity

Kevin-Chant (approved):

PR #6666, opened 2026-08-17
Assemble in-app agent skills at build time

oelshaikh:
The in-app assistant used to keep its own copy of the skills from [growthbook/skills](https://github.com/growthbook/skills). That copy drifted, which made it hard to maintain.

We no longer check those skills into this repo. At build time we assemble them from a checkout of growthbook/skills (`assemble-skills`, also run by `dev` and `build`). An allowlist in `assemble-agent-skills.mjs` copies only the skills that belong in the app: `feature-flags`, `experiments`, `analytics`. `gb-setup` is left out on purpose. Skills that only exist for the in-app assistant, like `growthbook-docs`, live in `src/agent/skills-local/` and get merged in after.

Local checkout, the lock file, and how CI picks up a new pin are in [`src/agent/README.md`](https://github.com/growthbook/growthbook/blob/oe/skills-refactor/packages/back-end/src/agent/README.md).

The canonical skills from `growthbook/skills` were written for shell agents (`gb-call`), so the prompt tells the model to translate them into `callApi` instead of us rewriting the markdown on the way in. This seems to work well in testing.


packages/back-end/scripts/assemble-agent-skills.mjs
```diff
@@ -0,0 +1,217 @@
...
+  return root;
+}
+
+const packageRoot = resolvePackageRoot();
+const repoRoot = resolve(packageRoot, "..", "..");
+const outputDir = join(packageRoot, "generated", "agent-skills");
+const localSkillsDir = join(packageRoot, "src", "agent", "skills-local");
+
+const SKILLS_REPOSITORY = "https://github.com/growthbook/skills";
+const CANONICAL_SKILLS = ["feature-flags", "experiments", "analytics"];
```
bryce-fitzsimons, line 52: Do skill authors need to remember to update this whitelist, and is this desired?

bryce-fitzsimons (approved):

packages/back-end/scripts/assemble-agent-skills.mjs
oelshaikh: Yes, but only if entirely new domains are introduced. The alternative is to have a blacklist to block any skills that are specific to coding / CLI agents (like `gb-setup`), but then if a new domain is added that's coding / CLI specific, it could leak to our in-app assistant.

PR #6667, opened 2026-08-17
Add Workload Identity Federation auth method for Snowflake data sources

jmcginnis-times:
> **Status:** both findings from the automated review are addressed (`8344a3f` — server-side provider validation; credentials cleared on auth-method switch) and unit tests cover the new auth branches (`02a9ed5`, 3/3 passing). The bot's "Confidence Score: 2/5" comment predates those fixes — see the comment thread for the finding→fix mapping.

### Motivation

Snowflake's [Workload Identity Federation](https://docs.snowflake.com/en/user-guide/workload-identity-federation) went GA in August 2025, and the `snowflake-sdk` GrowthBook already bundles (`^2.4.3`) has supported it since 2.2.0 — but the Snowflake integration only exposes password and key-pair auth, so the capability is unreachable. WIF lets a self-hosted GrowthBook authenticate as the cloud identity it already runs under (an AWS IAM role, Azure managed identity, or GCP service account): no stored credential to leak, rotate, or expire.

We (LA Times data engineering) run self-hosted GrowthBook on ECS and want to retire its stored Snowflake credential — this PR is how.

### Changes

- **shared types**: `authMethod` gains `"workload-identity"`; new optional `workloadIdentityProvider?: "AWS" | "AZURE" | "GCP"`, classified `"public"` in `datasource-params`
- **back-end**: `buildSnowflakeConnection` passes `authenticator: "WORKLOAD_IDENTITY"` + `workloadIdentityProvider` — both already in sdk 2.4.3's `ConnectionOptions`, so **no SDK bump needed**
- **front-end**: third option in the Authentication Method select — hidden on GrowthBook Cloud (no customer-controlled identity to bind there) — plus a required cloud-provider select with help text
- **docs**: new "Authentication Methods" section on the Snowflake page, including the Snowflake-side `CREATE USER … WORKLOAD_IDENTITY` setup example and two caveats (a shared server role means everything under it can authenticate as the Snowflake user; the Event Forwarder still requires key-pair, and its existing error covers that)

Deliberately out of scope: the `OIDC` provider (needs token plumbing rather than ambient credentials) — happy to follow up if there's interest.

### Testing

- `type-check`, `prettier`, and `eslint` clean on all touched packages (the shared package's classification map correctly forced the new field to be classified — nice mechanism)
- Connection parameters verified against snowflake-sdk 2.4.3's typings and the [Node driver auth docs](https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-authenticate)
- We'll be dogfooding immediately on our ECS deployment against a `WORKLOAD_IDENTITY`-bound service user — happy to iterate on review feedback


jmcginnis-times:
Both review findings addressed in 8344a3f:

1. **Server-side provider validation** — `buildSnowflakeConnection` now throws a clear error when `authMethod` is `workload-identity` without a `workloadIdentityProvider`, so API/config callers can't create a connection the SDK would reject opaquely.
2. **Stale credentials on method switch** — the auth-method select now clears the superseded method's params in one atomic `setParams` call (following PrestoForm's precedent of receiving `setParams`): switching to workload-identity blanks `password`/`privateKey`/`privateKeyPassword` — secretless should mean nothing stored — and switching away blanks `workloadIdentityProvider`.

jmcginnis-times:
Added unit tests in 02a9ed5 — the three auth-method contracts of `buildSnowflakeConnection`: workload-identity passes `authenticator=WORKLOAD_IDENTITY` + the provider with **no stored-credential fields** on the connection; a missing provider **throws before any connection is created**; and the password default is unchanged. `buildSnowflakeConnection` is exported for tests per the existing repo idiom, and `SqlIntegration` is mocked (the module only needs `TEST_QUERY_SQL` from it; the real import drags a circular chain under jest's module order). 3/3 passing locally.

jmcginnis-times:
Two updates:

1. **Merged `main`** (clean, no conflicts) so the branch is current.
2. **The self-hosted-only restriction is now enforced server-side** (85da704), closing the loop on the earlier review's point that the UI gating alone could be bypassed by a direct API request: `buildSnowflakeConnection` rejects `workload-identity` when `IS_CLOUD`, before any connection is created — on GrowthBook Cloud the ambient identity would be GrowthBook's own infrastructure rather than the customer's, so failing loudly is the only right behavior. Covered by a fourth unit test (4/4 passing).

jmcginnis-times:
Both new review findings addressed in 61cae72:

1. **Unsupported providers bypass validation** — confirmed valid: the TS union constrains the UI but not direct API requests. `buildSnowflakeConnection` now validates against the AWS/AZURE/GCP allowlist (covers both the missing and the invalid case), with a unit test for a lowercase `aws` rejection.
2. **Empty updates retain old secrets** — confirmed valid against `mergeRecord`'s semantics (blank secret updates mean keep-existing by design, since the UI never round-trips secrets), which made the earlier client-side clearing insufficient exactly as flagged. The fix is now server-side where it belongs: `mergeDataSourceParams` strips `password`/`privateKey`/`privateKeyPassword` when a snowflake datasource's **merged** `authMethod` is `workload-identity` — so a direct API update can't leave stored credentials behind either. Covered by shared tests for both the strip and the unchanged keep-existing behavior for other auth methods; the client-side clearing stays as UX belt-and-suspenders.

The commit also repairs the back-end test harness: eslint's `no-var` auto-fix had rewritten the mock's hoisted-`var` backing store to `let` during a pre-commit hook run, breaking the suite at import (TDZ) — the store now lives on `globalThis` with a comment explaining why. All suites green: 5/5 back-end snowflake, 15/15 shared datasource-params.

jmcginnis-times:
Rebased on `main` (Sep 9) — the conflict with the new async-query-status work is resolved, and the new `getSnowflakeQueryStatus` path routes through the WIF-aware `buildSnowflakeConnection`, so workload-identity connections pick up query-status support for free. Type-check and all three snowflake test suites green (30 tests), including the auth-method contract tests added earlier.

Status recap for a human reviewer: all four rounds of greptile findings are addressed (server-side provider allowlist, stored-credential clearing on auth-method switch, self-hosted-only enforced server-side, unit tests for the three auth-method contracts). The change is additive — hidden on Cloud, no behavior change for existing password/key-pair connections. Happy to make any changes review turns up.

PR #6669, opened 2026-08-18
Custom hook Test panel: simulate incremental suppression and prefill from a Feature Flag

Kevin-Chant:
### Features and Changes

Two gaps that made Custom Hooks hard to author, both surfaced by a support escalation where the correct hook looked broken in the Test panel.

The panel could not exercise **Incremental Changes Only**. It ran the hook once with no prior state, so a hook that correctly throws in the panel and correctly does not block on save looked wrong. `/custom-hooks/test` now accepts an optional `originalFunctionArgs` and applies the real suppression rule via `applyIncrementalSuppression`, extracted from `_runCustomHook` so the panel cannot drift from runtime. The response reports what suppression hid — "nothing happened" is the confusing outcome, so the panel names it.

Fixtures created from Settings were always generic samples with no way to point at real data. The panel now offers a picker that seeds both the proposed and prior state from an existing Feature Flag.

`_runCustomHook` keeps its call pattern: the second sandbox run still only happens when there is an error or warning to suppress, so successful hooks don't pay for an extra run.

Also corrects the `FeatureRevisionModel` schema comment, which claimed every revision envelope is present only when explicitly changed. True of its neighbours, but not of `metadata`, which `prepareFeatureRevision` always writes as a complete snapshot. Reading it as a sparse diff is what sent #6656 down the wrong path.

### Testing

Endpoint, against a local stack:

- [x] Same error before and after -> `success: true`, error reported under `suppressed`
- [x] Error only in the proposed state -> `success: false`, nothing suppressed
- [x] Same warning before and after -> dropped from `warnings`, reported under `suppressed`
- [x] No `originalFunctionArgs` -> unchanged response, no `suppressed` key

Runtime path, to confirm the extraction didn't change behavior:

- [x] Adding a blocked tag with Incremental Changes Only on -> still blocked
- [x] Unrelated edit on a Feature Flag that already has it -> still passes
- [x] `experimentCustomHooks` and `publishImmediateCustomHooks` suites pass

### Screenshots

With incremental changes checked (default)
hookmodal

With incremental changes unchecked
hookmodal-no-incremental

Prefilled from a feature
hookmodal-prefilled

### DIFF
--- a/packages/back-end/generated/spec.yaml
+++ b/packages/back-end/generated/spec.yaml
@@ -26182,6 +26182,15 @@ paths:
                   propertyNames:
                     type: string
                   additionalProperties: {}
+                originalFunctionArgs:
+                  description: >-
+                    State before the change. When supplied, the hook also runs
+                    against it and Incremental Changes Only suppression is
+                    applied to the result.
+                  type: object
+                  propertyNames:
+                    type: string
+                  additionalProperties: {}
                 entityType:
                   type: string
                   enum:
@@ -26215,6 +26224,8 @@ paths:
                   log:
                     description: Captured console output
                     type: string
+                  suppressed:
+                    $ref: '#/components/schemas/suppressed'
                 required:
                   - success
                 additionalProperties: false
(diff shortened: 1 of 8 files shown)


Kevin-Chant:
@cursoragent review

Kevin-Chant:
@cursoragent  review

bryce-fitzsimons (approved):

PR #6670, opened 2026-08-18
Fix for experiment analysis prompt size

Auz:
As we pass a lot of information to the experiment results generator, it's easy to go over the token limits. This fix summarizes the experiments more efficiently to not only reduce the chance of hitting the token limits, but also increase the max number of metrics we can analyze.

### DIFF
--- a/packages/back-end/src/controllers/experiments.ts
+++ b/packages/back-end/src/controllers/experiments.ts
@@ -45,6 +45,7 @@ import { EventUserForResponseLocals } from "shared/types/events/event-types";
 import { CreateURLRedirectProps } from "shared/types/url-redirect";
 import isEqual from "lodash/isEqual";
 import { getMetricMap } from "back-end/src/models/MetricModel";
+import { summarizeExperimentForAI } from "back-end/src/util/ai-prompt.util";
 import {
   AuthRequest,
   ResponseWithStatusAndError,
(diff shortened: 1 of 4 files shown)


Auz:
@greptileai review

packages/back-end/src/util/ai-prompt.util.ts
```diff
@@ -0,0 +1,369 @@
...
+  }
+  if (covariateImbalance) {
+    health.covariateImbalance = covariateImbalance.isImbalanced;
+  }
+  if (traffic?.error) health.trafficError = traffic.error;
+
+  return Object.keys(health).length ? health : undefined;
+}
+
+export function summarizeExperimentForAI({
```
tzjames: nit: There is a function of the same name in learnings.controller.ts.  Maybe rename this to: `summarizeExperimentForAI`

packages/back-end/src/controllers/experiments.ts
```diff
@@ -286,117 +287,87 @@ export async function postAIExperimentAnalysis(
...
-    "\n- Ratio Metrics: Represent the ratio of two numeric values among experiment users." +
-    // Statistical results
-    "\n- Statistical results for metrics include: Conversion Rate (CR)" +
-    "\n- Statistical results for metrics include: Value: Represents the total value of the metric across all users who saw the variation." +
-    "\n- Statistical results for metrics include: Confidence Interval (CI): Represents the range within which the true metric value is likely to fall." +
-    "\n- Statistical results for metrics include: Uplift: Represents the difference in metric performance between variations (e.g., the increase in average revenue per user for a variation compared to the baseline)." +
-    "\n- Statistical results for metrics include: Chance to Win: Represents the probability that a variation is better than others for this metric." +
+    "\n  - metricType: proportion (also called binomial), mean, count, duration, revenue, ratio, quantile, or retention." +
```
tzjames: There is a new dailyParticipation metricType that we can add to give guidance.

tzjames (approved):
Ok looks good.  I added a small commit that:
1. Uses the orgs actual srm value.
2. renames `summarizeExperimentForAI` to `summarizeExperimentAnalysisForAI` to not have duplicate function names in the code base.
3. Adds info on the new dailyParticipation metricType.
You could maybe have a quick look over that before landing.  

The AI thinks that the current implementation might be worse if:
```
1. Only the first/default analysis is used (getSnapshotAnalysis(snapshot) → analyses[0]). If an org runs both Bayesian and frequentist analyses, or has a secondary analysis with different settings (e.g. a different differenceType or a CUPED-adjusted variant), the AI now only ever sees one. The old raw dump exposed every analysis in snapshot.analyses, however unstructured.
2. Dimension-sliced results are dropped (analysis.results?.[0] — the "All" row only). If a metric moved for mobile but not desktop, or the experiment was primarily evaluated by a dimension breakdown, that's invisible to the new summary. This is consistent with existing codebase convention (I checked — every other consumer does the same .results[0] access), but it is a real reduction versus the old raw dump, which included every dimension's results.
```
I'll leave it to you to decide whether the trade off there is worth it.

Removing variations[].value is worth doing, at the MAX_METRICS_FOR_AI limit removing it would save up to 8% theoretically ... according to a rough analysis by AI.  
```
variations[].value (the raw summed value per metric per variation) — the strongest candidate. Once cr (per-user rate), lift, ci, pValue/chanceToWin are present, value is just cr × users restated as an absolute sum. It doesn't change any interpretation of significance or direction, and the model's job here is "is this metric a statistically meaningful win/loss," not "what was the raw total." I'd cut it — that's one number per metric per variation, so it scales with the same metrics × variations multiplier the MAX_METRICS_FOR_AI comment already worries about.
```

PR #6673, opened 2026-08-18
Enforce cookies only contain string values

Auz:
Enforce that cookie value are always strings to harden against potential security vulnerabilities.

### DIFF
--- a/packages/back-end/src/controllers/auth.ts
+++ b/packages/back-end/src/controllers/auth.ts
@@ -374,10 +374,17 @@ export async function getResetPassword(
   });
 }
 
-export async function getSSOConnectionFromDomain(req: Request, res: Response) {
+export async function getSSOConnectionFromDomain(
+  // eslint-disable-next-line
+  req: Request<any, any, { domain: unknown }>,
+  res: Response,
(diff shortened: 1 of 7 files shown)


Auz:
Thanks — flagging that the `Request<any, any, { domain: unknown }>` signature here is intentional: it matches the convention every other pre-auth handler in this file already uses (`postLogin`, `postRegister`, `postForgotPassword`, `postResetPassword` — all `Request<any, any, {…: unknown}>` with the same eslint-disable).

Worth noting the change actually moved *toward* the strict-TS intent, not away from it: the original signature was plain `req: Request`, which typed `req.body.domain` as `any`. Typing the body as `{ domain: unknown }` is what forces the new `typeof domain !== "string"` guard. The only `any`s are in the Express `Params`/`ResBody` generic slots, which are mechanical and shared by all the siblings.

Going fully strict (`Request<Record<string, never>, unknown, { domain: unknown }>`, no suppression) would make this one handler diverge from the five around it for no security or correctness gain, so leaving it consistent with the file.

jdorn (approved):

PR #6675, opened 2026-08-18
Added more margin on the auth error card

Auz:
Log Out and Reload sat flush against each other with no gap. Replaces the Bootstrap with Radix Flex plus a real gap, and swaps the legacy Button/h3 for the design-system equivalents.


Auz:
@greptileai review

PR #6679, opened 2026-08-19
feat(results): Handle analysis failures per metric-analysis

ahdriel:
### Features and Changes

Currently, during an Experiment Update, if all queries succeed, we send the data to `gbstats` to be analyzed. But if any metric analysis fails, we throw away the whole analysis.

This PR changes so the failure is per metric analysis

- Each metric can have error per analysis (so relative can succeed while absolute fails)
- UI renders the error message per metric analysis

### Testing

- Unit tests

### Screenshots

Screenshot 2026-08-19 at 10 56 25 AM


ahdriel:
cursor review

packages/stats/gbstats/gbstats.py
```diff
@@ -1413,13 +1413,23 @@ def process_experiment_results(
...
-                                analyses=d.analyses,
+                        try:
+                            results.append(
+                                process_single_metric(
+                                    rows=rows,
+                                    metric=this_metric,
+                                    analyses=d.analyses,
+                                )
+                            )
+                        except Exception as e:
```
lukesonnet: I think the error should be per analysis, e.g. inside the loop over analyses in `process_single_metric`.

packages/back-end/src/services/stats.ts
```diff
@@ -512,15 +512,48 @@ const getFormattedCI = (
   return [ci[0] ?? -Infinity, ci[1] ?? Infinity];
 };
 
-function parseStatsEngineResult({
+export function addMetricErrorsToDimensions(
```
lukesonnet: Does this go away if the error is at the analysis level? I feel it does...

lukesonnet (changes requested):

packages/stats/gbstats/gbstats.py
ahdriel: done!

packages/back-end/src/services/stats.ts
ahdriel: it would not go away -- because we fill in 'unexpected empty spaces' with `NO_DATA`, we need to also copy the `errorMessage` to those so we properly understand that the failure at the analysis level is passed down to the metric level.

Agree with the concept though, that each layer should have its own error tracking + handling, instead of just delegating it down by copying it -- but it is a bigger refactor to reduce the amount of `NO_DATA` records we create

ahdriel:
cursor review

packages/stats/gbstats/gbstats.py
```diff
@@ -1169,25 +1169,41 @@ def process_single_metric(
             attempted_quantile_dimension_reaggregation
             or attempted_quantile_overall_reaggregation
         ):
+            analysis_results.append(
+                ExperimentMetricAnalysisResult(
```
lukesonnet, line 1173: minor q: Is this added empty return object needed?

packages/stats/gbstats/gbstats.py
```diff
@@ -1169,25 +1169,41 @@ def process_single_metric(
...
+                )
+            )
+        except Exception as e:
+            analysis_results.append(
+                ExperimentMetricAnalysisResult(
+                    unknownVariations=list(unknown_var_ids),
+                    dimensions=[],
+                    multipleExposures=0,
+                    error=str(e),
+                    traceback=traceback.format_exc(),
```
lukesonnet, line 1201: maybe a dumb q but we tested this actually works in our set-up, yes?

packages/back-end/src/services/experimentTimeSeries.ts
```diff
@@ -233,6 +233,11 @@ export async function updateExperimentAnalysisTimeSeries({
     );
 
     for (const metricId of allMetricIds) {
+      const metricHasError = baseResult.variations.some(
+        (variation) => variation.metrics[metricId]?.errorMessage !== undefined,
+      );
+      if (metricHasError) continue;
```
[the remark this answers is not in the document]
lukesonnet: +1

lukesonnet (approved):

packages/stats/gbstats/gbstats.py
ahdriel: yes because we need our array to be aligned by the index (analysissettings in -> analysis results out)

ahdriel:
cursor review

packages/stats/gbstats/gbstats.py
ahdriel: yes! validated both in the same process and via external python service.

packages/stats/gbstats/gbstats.py
```diff
@@ -1169,25 +1169,41 @@ def process_single_metric(
...
+                    multipleExposures=0,
+                )
+            )
+        except Exception as e:
+            analysis_results.append(
+                ExperimentMetricAnalysisResult(
+                    unknownVariations=list(unknown_var_ids),
+                    dimensions=[],
+                    multipleExposures=0,
+                    error=str(e),
```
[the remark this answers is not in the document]
ahdriel: Given this is org-scoped, and the infra is owned by them and configured, having this information verbatim is super helpful to debug. And given the scoping, this is not a new information they cannot access. So we are keeping as is, as we cannot control all the potential errors from all warehouses, etc, and we want to surface those to the user.

ahdriel:
cursor review

ahdriel:
cursor review

ahdriel:
cursor review

packages/back-end/src/models/SafeRolloutSnapshotModel.ts
```diff
@@ -107,7 +111,12 @@ export class SafeRolloutSnapshotModel extends BaseClass {
       latestSafeRolloutSnapshot === null ||
       latestSafeRolloutSnapshot?.id === updatedDoc.id;
 
-    if (isLatestSnapshot && updatedDoc.status === "success") {
+    if (
+      isLatestSnapshot &&
+      updatedDoc.status === "success" &&
+      findAnalysisComputeFailure(getSafeRolloutSnapshotAnalysis(updatedDoc)) ===
+        null
+    ) {
```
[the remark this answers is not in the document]
ahdriel: This is intentional for this PR -- on `main` any stats failures in analysis would invalidate the whole snapshot, so summary stays on previous successful snapshot and side-effects do not run.

Now that we have partially successful snapshots, we still don't want to change the previous behavior as it requires more product decisions to ensure the right experience for the users: should surviving metrics trigger notifications? what about Experiment Decision Framework evaluation (and should it be different for guardrail vs secondary vs goal metrics?)

So this is not a new behavior, so the plan is to address this in a follow-up PR

PR #6688, opened 2026-08-19
Include referenced experiments in project-scoped SDK payloads

bryce-fitzsimons:
Previously `experiment-ref` rules were dropped from the SDK Payload if the underlying experiment was in a different project than the feature (features in All Projects were unaffected). Fixes by always including referenced experiments irrespective of project alignment. Especially important now that we have targeting projects. Note: this fix also works for `contextual-bandit-ref` rules.

### DIFF
--- a/packages/back-end/src/util/features.ts
+++ b/packages/back-end/src/util/features.ts
@@ -592,6 +592,36 @@ export function getSDKPayloadKeysByDiff(
   return getSDKPayloadKeys(environments, projects);
 }
 
+export function referencedExperimentIds(
+  features: FeatureInterface[],
+): string[] {
+  const ids = new Set<string>();
+  features.forEach((feature) => {
+    (feature.rules ?? []).forEach((rule) => {
+      if (rule.type === "experiment-ref" && rule.enabled !== false) {
+        ids.add(rule.experimentId);
+      }
+    });
+  });
+  return [...ids];
+}
+
+// An experiment a delivered feature references belongs in that feature's payload
+// even when the experiment itself lives in another project.
+export function experimentMapForFeatures(
+  experimentMap: Map<string, ExperimentInterface>,
+  features: FeatureInterface[],
+  projects: string[],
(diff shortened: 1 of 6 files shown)


packages/back-end/src/util/features.ts
```diff
@@ -592,6 +592,36 @@ export function getSDKPayloadKeysByDiff(
   return getSDKPayloadKeys(environments, projects);
 }
 
+export function referencedExperimentIds(
```
lukesonnet: Do we not have existing canonical definitions for this? Seems like there should already be helpers for this with business logic in one place.

packages/back-end/src/util/features.ts
```diff
@@ -592,6 +592,36 @@ export function getSDKPayloadKeysByDiff(
...
+      if (rule.type === "experiment-ref" && rule.enabled !== false) {
+        ids.add(rule.experimentId);
+      }
+    });
+  });
+  return [...ids];
+}
+
+// An experiment a delivered feature references belongs in that feature's payload
+// even when the experiment itself lives in another project.
```
lukesonnet, line 633: I feel like we may want to warn on the experiment page about this mis-match in the Linked Features section. I wonder if we should allow it at all if the experiment's project is not within the set of feature projects.

packages/back-end/src/util/features.ts
bryce-fitzsimons: good catch, we had a similar method that we can tap into.

bryce-fitzsimons: Actually we didn't, we just repeated this same sort of logic in multiple places. Exporting this as a shared method and using it in other locations too.

lukesonnet (approved):

packages/back-end/src/util/features.ts
bryce-fitzsimons: agree, something we should take on soon (not this PR though)

PR #6691, opened 2026-08-19
funnel explorer v1

lillialjackson:
### Features and Changes

Allows users to select existing funnel metrics and analyze them, also allows them to build a new funnel and then save that funnel as a new funnel fact metric

### Dependencies

none

### Testing

manual and unit

### Screenshots

Screenshot 2026-08-19 at 8 47 23 AM
Screenshot 2026-08-19 at 8 47 12 AM
Screenshot 2026-08-19 at 8 46 59 AM
Screenshot 2026-08-19 at 8 46 53 AM
Screenshot 2026-08-18 at 3 17 37 PM

Currently don't support updating a metric from explorer:
Screenshot 2026-08-19 at 9 03 50 AM


packages/front-end/enterprise/components/ProductAnalytics/SideBar/FunnelTabContent.tsx
```diff
@@ -195,6 +324,51 @@ export default function FunnelTabContent() {
...
+          {funnelMetricOptions.map((o) => (
+            <SelectItem key={o.value} value={o.value}>
+              <Flex align="center" gap="2">
+                <PiFunnelBold size={14} />
+                {o.label}
+              </Flex>
+            </SelectItem>
+          ))}
+          {funnelMetricOptions.length > 0 && <SelectSeparator />}
+          <SelectItem value={NEW_FUNNEL_VALUE}>
```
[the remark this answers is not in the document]
lillialjackson: may not fix based on what was discussed with design

packages/front-end/enterprise/components/ProductAnalytics/SideBar/FunnelTabContent.tsx
```diff
@@ -195,6 +324,51 @@ export default function FunnelTabContent() {
...
+            <Flex align="center" gap="2">
+              <PiPlus size={14} />
+              Build a New Funnel
+            </Flex>
+          </SelectItem>
+        </Select>
+        <Callout status="info" size="sm" mt="1">
```
mknowlton89: I like that we're calling this out, but I think we may be able to simplify this UX?

If there are no saved funnel metrics for a data source, we should just default to the `+ Build a New Funnel` option so the user doesn't have to click it. We can probably remove this case from the callout, or minimize the treatment as it takes up a good chunk of vertical space in a pretty space-starved spot.

Image

packages/front-end/enterprise/components/ProductAnalytics/Explorer.tsx
```diff
@@ -218,12 +254,18 @@ function ExplorerInner({ type }: { type: DatasetType }) {
...
+  );
+
   const getQueryParam = (value: string | string[] | undefined) =>
     Array.isArray(value) ? value[0] : value;
   const rawParam = getQueryParam(router.query.config);
   const metricId = getQueryParam(router.query.metricId);
   const factTableId = getQueryParam(router.query.factTableId);
   const datasourceId = getQueryParam(router.query.datasourceId);
+  const funnelMetricId = getQueryParam(router.query.funnelMetricId);
   const seedId =
```
mknowlton89, line 282: Came across a small bug - The funnel type seeds the `datasourceId` - so if you're viewing a funnel metric via the funnel metric explorer, and then click the `Funnels` sidebar link, the URL will update and remove the seed, but the actual page doesn't re-render.

Not a big thing, but wanted to call it out.

mknowlton89 (commented):
PR is looking great. 

Came across a few things, mostly small nits.

1) I ran into a slightly weird user experience on the dashboard. I had a dashboard that had a funnel block built from a funnel metric. If I changed the dropdown from the funnel metric to the `+ Build a new funnel`, the actual block was still rendering the original funnel, which felt a bit weird.

Image

2) For this, issue, I created a funnel exploration via the funnel explorer and created it from an existing funnel metric. I then added a `Group by` clause before adding the funnel to a dashboard. 

When I opened that funnel via the dashboard editor, it wasn't linked to the original funnel (because of the groupby). And the dropdown that allows the user to select a funnel metric or create a new one was just empty. I feel like if there is no metric found, should we default to the `+ Build a New Funnel` or have a fallback for something like `Custom Funnel`, or should the matching logic ignore any groupby fields?

mknowlton89 (approved):

PR #6692, opened 2026-08-19
fix(fact-tables): update factTable public API

ahdriel:
### Features and Changes

When a fact table is created, we have 2 processes that run:
- Column detection (if not provided)
- Top Values detection

This information was already being returned in the API because we were spreading the object, so now we are:
- Updating spec to match what was already being returned
- Updating some fields to be `readonly`
- Making `columnRefreshPending` behave as expected and give the proper signal to the API consumer
- And fix the error message when column type is not known

### Testing

- Unit tests


ahdriel:
cursor review

packages/back-end/src/util/factTable.ts
```diff
@@ -164,6 +164,12 @@ export function columnsHaveAutoSlices(
   return (columns ?? []).some((c) => !!c.isAutoSliceColumn || !!c.autoSlices);
 }
 
+export function columnsNeedDetection(
+  columns?: Array<{ datatype?: string }>,
+): boolean {
+  return (columns ?? []).some((c) => !c.datatype);
```
lukesonnet: Don't we sometimes keep "deleted" columns around instead of actually deleting them?

Should this check if `!c.deleted`?

packages/back-end/src/services/factMetricAggregationValidation.ts
```diff
@@ -27,6 +27,28 @@ export function validateAggregationSpecification({
...
+
+  const isNamedColumnReference =
+    !!column.column &&
+    column.column !== "$$count" &&
+    column.column !== "$$distinctUsers" &&
+    column.column !== "$$distinctDates";
+  const typeUnknown = datatype === undefined || datatype === "";
+
+  if (isNamedColumnReference && typeUnknown) {
+    if (factTable.columnRefreshPending) {
```
lukesonnet, line 39: Eh, I'm not sure we should be blocking on this? I feel somewhat like we should just let them pick what they want and fix it later if it's in this intermediate state, or instead warn that column detection is still pending and how to resolve it loudly on the fact table page itself, or if you enter the fact metric modal from the main Metrics page, then warn loudly when you pick the fact table, but let people proceed.

lukesonnet: Same with below `columnsError`, especially if resolving may not be trivial (and it's a bit unclear they have to close the modal -> click "refresh" on columns).

packages/back-end/src/services/demo-datasource.ts
```diff
@@ -310,7 +310,9 @@ async function ensureDemoFactTables(
           datatype: "string",
         },
       ],
-      columnRefreshPending: true,
+      // Datatypes are supplied above, so column detection is not pending. The
+      // refresh queued below only enriches topValues for filter autocomplete.
```
lukesonnet: 🤖

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -220,7 +250,50 @@ export function populateAutoSlices(
...
+
+    if (currentColumn.isAutoSliceColumn) {
+      updatedColumn.autoSlices = refreshedColumn.autoSlices;
+    }
+
+    return updatedColumn;
+  });
+}
+
+export async function runColumnDetectionQuery(
```
lukesonnet: nit: This belongs in a different file.

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -383,6 +456,17 @@ export async function runRefreshColumnsQuery(
     Object.assign(col, normalizePersistedColumn(col));
   }
 
+  return columns;
+}
+
+/** Mutates the supplied columns and returns those enriched successfully. */
+export async function populateColumnTopValues(
```
lukesonnet: nice, good to split these! Should this also hint in it's name that it is running queries?

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -220,7 +250,50 @@ export function populateAutoSlices(
...
+      return currentColumn;
+    }
+
+    const updatedColumn = {
+      ...currentColumn,
+      topValues: refreshedColumn.topValues,
+      topValuesDate: refreshedColumn.topValuesDate,
+    };
+
+    if (currentColumn.isAutoSliceColumn) {
```
lukesonnet: q: What if refreshedColumns has `isAutoSliceColumn`? Is that not possible? Shouldn't it check the refreshed column incase future consumers pass new slices here at the same time as an updated column?

packages/back-end/src/jobs/refreshFactTableColumns.ts
```diff
@@ -116,30 +116,60 @@ const refreshFactTableColumns = async (job: RefreshFactTableColumnsJob) => {
...
   } catch (e) {
-    updates.columnsError = e.message;
+    columnDetectionChanges.columnsError = e.message;
   }
 
-  // Always set columnRefreshPending to false - job is done (even if it failed)
-  updates.columnRefreshPending = false;
-  await updateFactTableColumns(factTable, updates, context);
+  // Column types are now known (or detection failed). Clear pending status before the slow top-values
+  // scan so metric creation unblocks as soon as types are available.
```
lukesonnet, line 68: nice

packages/back-end/src/models/FactTableModel.ts
```diff
@@ -1274,23 +1274,24 @@ export function toFactTableApiInterface(
   factTable: FactTableInterface,
 ): ApiFactTable {
```
lukesonnet, line 1275: q: is there a way to make optional fields mandatory, so that you have to pass them as `undefined` instead of just ommitting them below? Just in general, could be easy to add a new optional field and not see you need to add it in this passthrough below, especially now that we aren't just spreading the whole thing.

packages/back-end/src/api/fact-tables/updateFactTable.ts
```diff
@@ -173,11 +173,15 @@ export const updateFactTable = createApiRequestHandler(
...
-  ) {
+    (columnsUpserted && columnsNeedDetection(factTable.columns));
+  if (willRefresh) {
+    data.columnRefreshPending = true;
+  }
+
+  await updateFactTableInDb(req.context, factTable, data);
+  if (willRefresh) {
     await queueFactTableColumnsRefresh(factTable);
   }
```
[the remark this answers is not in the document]
lukesonnet: In general, this kind of thing is why I'm worried about hard blocking user actions on `columnRefreshPending`.

packages/back-end/src/services/factMetricAggregationValidation.ts
ahdriel: The current experience is that we throw an error because the validation fails, so this is not a new behavior (metric creation erroring out if column type is unknown).

For `columnsError` they might need to fix the fact table sql, no? but this should not happen very often.

packages/back-end/src/util/factTable.ts
ahdriel: This is previous code, but agreed that we might be queuing a `column refresh` for previously deleted columns that we don't need. will fix.

lukesonnet: leave the trail cleaner than you found it <3

packages/back-end/src/jobs/refreshFactTableColumns.ts
ahdriel: added `refresh` to the name. we unfortunately already have a `runColumnTopValueQuery` so the refactor would be a bit bigger (touching more files)

I do plan on cleaning up more of the top values later in a separate PR

packages/back-end/src/jobs/refreshFactTableColumns.ts
ahdriel: it is possible, but the `topValues` is always older/staler than `currentColumn` so we trust the latest status.

this is for the scenario where more modifications were done to the column while the topValues was running -- so if you disable the autoSliceColumn in between, we don't need to do these updates.

lukesonnet (approved):

PR #6695, opened 2026-08-19
feat(query-runner): recover stalled updates

johnham-ant:
### Features and Changes

When all warehouse queries for a snapshot finish, the finalize/publish step runs once from an in-memory callback with no retry. If that hand-off is dropped — a lost timer, a thrown concurrency check, a stale read — the snapshot stays `running` until the cleanup job errors it out about an hour later, and the whole refresh is wasted.

This adds a safety net around that hand-off and closes the gaps that let a refresh silently stall or publish bad data:

- **Refresh watchdog.** A timer re-arms a dropped refresh pass while the runner is active. Passes are serialized, a pass stuck too long is fenced off and  replaced, and after a bounded number of failed passes the run fails loudly  instead of hanging forever.
- **Completeness gate on finalize.** Finalize when every persisted query pointer is terminal, even when a stale read reports no pointer change. Refuse to run the analysis when query docs or results are missing from the read, instead of  publishing a zero-filled snapshot.
- **Safer shutdown.** Error writes are conditional on the snapshot still running, and the runner stands down (no error, no auto-refresh disable) when the snapshot was cancelled or concluded elsewhere.
- **Re-queue stranded queries** that a thrown concurrency check left `queued`.

### Testing

- New unit tests

### Future work

Out of scope, proposed as a follow-up: having the cleanup job (`expireOldQueries`) re-finalize a stalled snapshot from persisted query results instead of erroring it, for the case where the process that owned the runner is gone entirely.

### DIFF
--- a/packages/back-end/src/queryRunners/QueryRunner.ts
+++ b/packages/back-end/src/queryRunners/QueryRunner.ts
@@ -80,6 +80,40 @@ const FINISH_EVENT = "finish";
 const INITIAL_CONCURRENCY_TIMEOUT = 250;
 const MAX_CONCURRENCY_TIMEOUT = 4000;
 
+// How often the refresh watchdog re-arms the debounced refresh while the
+// runner is active. Query completions normally drive the refresh, but every
+// trigger is a single-shot in-memory hand-off: a transient error, a stale
+// status read, or a lost concurrency-backoff timer can silently strand the
+// model in "running" until the stalled-snapshot reaper errors it an hour
+// later. The watchdog turns any such dropped trigger into a bounded delay.
+// It exists for lost-trigger recovery, not latency, so the interval is
+// deliberately long to keep the steady-state read load negligible.
+const REFRESH_WATCHDOG_INTERVAL_MS = 5 * 60 * 1000;
+// If a single refresh pass has been in flight for this long, assume one of
+// its awaits will never settle: fence the stuck pass off (it abandons at its
+// next checkpoint instead of publishing stale state) and start a fresh one.
+// The clock restarts at every pass checkpoint, so this bounds one SEGMENT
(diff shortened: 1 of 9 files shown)


packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -80,6 +80,40 @@ const FINISH_EVENT = "finish";
...
+// It exists for lost-trigger recovery, not latency, so the interval is
+// deliberately long to keep the steady-state read load negligible.
+const REFRESH_WATCHDOG_INTERVAL_MS = 5 * 60 * 1000;
+// If a single refresh pass has been in flight for this long, assume one of
+// its awaits will never settle: fence the stuck pass off (it abandons at its
+// next checkpoint instead of publishing stale state) and start a fresh one.
+// The clock restarts at every pass checkpoint, so this bounds one SEGMENT
+// between checkpoints (model read / analysis / publish), not the whole pass.
+// The fence exists because Mongo awaits are unbounded (no socketTimeout is
+// configured), so a wedged connection would otherwise hang a pass forever.
```
ahdriel: @johnham-ant have you observed this on your end?

I am leaning towards removing this logic and adding a configurable mongodb timeout for all queries, which would simplify this PR quite a lot.

And it would avoid the double analysis if this is considered stuck just because `gbstats` is doing expensive calculations.

johnham-ant: confirmed it was not observed! agreed with removing

ahdriel:
cursor review

ahdriel:
cursor review

ahdriel:
cursor review

ahdriel:
cursor review

ahdriel (approved):

PR #6704, opened 2026-08-20
Bulk fact metric API improvements

jdorn:
### Features and Changes

While working on an agent skill to migrate legacy metrics to fact tables, several gaps/improvements were identified in our current API endpoints.

`POST /bulk-import/facts`

- Funnel metrics are now supported (plus a Zod refactor to avoid schema drift like this in the future)
- New `defaultManagedBy` option so it doesn't have to be specified on every resource individually
- New `dryRun` option
- Improved error messages when a failure happens mid-run

`GET /metrics`

- New `?includeArchived` option

### Testing

Manually verified against API:

- [x] 28 tests ran against bulk endpoint testing dryRun, defaultManagedBy, error handling, funnel metrics, and normal usage
- [x] Tested legacy metrics endpoint `includeArchived`


packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -76,253 +126,400 @@ export const postBulkImportFacts = createApiRequestHandler(
...
+
+          let counted = false;
+          if (!dryRun && updateData.columns) {
+            await upsertColumns({
+              context: req.context,
+              factTable: existing,
+              columns: updateData.columns,
+            });
+            delete updateData.columns;
+            numUpdated.factTables++;
```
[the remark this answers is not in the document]
jdorn: Should not happen very often and the impact is pretty low. Not worth fixing.

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -76,253 +126,400 @@ export const postBulkImportFacts = createApiRequestHandler(
...
-      // Update existing fact table
-      if (existing) {
-        if (!req.context.permissions.canUpdateFactTable(existing, data)) {
-          req.context.permissions.throwPermissionError();
-        }
-        if (data.userIdTypes) {
-          validateUserIdTypes(existing.datasource, data.userIdTypes);
-        }
+        // Update existing fact table
+        if (existing) {
```
najito: Does updating a fact table require `requireOfficialResourcesPlan`?

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -76,253 +126,400 @@ export const postBulkImportFacts = createApiRequestHandler(
...
-        );
-        createProps.id = id;
+          if (!req.context.permissions.canUpdateFactMetric(existing, changes)) {
+            req.context.permissions.throwPermissionError();
+          }
 
-        const newFactMetric =
-          await req.context.models.factMetrics.create(createProps);
-        factMetricMap.set(newFactMetric.id, newFactMetric);
+          if (dryRun) {
```
najito, line 467: `factMetricMap` isn't updated in `dryRun` branches, so `dryRun` can't detect duplicate-id creates that live mode would hit.

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -38,6 +51,45 @@ export const postBulkImportFacts = createApiRequestHandler(
...
+      await req.context.registerTags([...tagsToAdd]);
+    }
+  };
+
+  const onItemError = async (
+    resourceType: BulkImportError["resourceType"],
+    id: string,
+    err: unknown,
+  ) => {
+    if (err instanceof PermissionError) throw err;
```
najito: Are permission errors meant to be rethrown raw instead of wrapped?

This discards partial-success counts after real writes already happened.

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -76,253 +126,400 @@ export const postBulkImportFacts = createApiRequestHandler(
...
-      data.tags?.forEach((t) => tagsToAdd.add(t));
-      if (data.projects) validateProjectIds(data.projects);
-
-      // This bulk endpoint is mostly used to sync from version control
-      // So default these resources to only be managed by API and not the UI
-      if (data.managedBy === undefined) {
-        data.managedBy = "api";
-      }
+      try {
+        data.tags?.forEach((t) => tagsToAdd.add(t));
```
najito: Can tagging first here lead to orphaned tags if the data fails validation?

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
najito, line 232: Should the `existing` branch also check for `requireOfficialResourcesPlan`

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
```diff
@@ -76,253 +126,400 @@ export const postBulkImportFacts = createApiRequestHandler(
...
+              existingFactFilter.id,
+              filterPayload,
+            );
+          }
+          Object.assign(existingFactFilter, filterPayload);
+          numUpdated.factTableFilters++;
+        }
+        // Create new filter
+        else {
+          if (!factTable.managedBy && filterPayload.managedBy) {
```
najito, line 394: `!factTable.managedBy`, does not catch "admin" (truthy), so admin-managed tables can get api-managed filters attached.

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
jdorn: We typically only enforce enterprise features on create, not update, to avoid breaking things too badly when an org downgrades.

packages/back-end/src/api/bulk-import/postBulkImportFacts.ts
jdorn: I think this is probably ok for now

Auz (approved):

PR #6705, opened 2026-08-20
incremental-pipeline: Add support to "exclude in-progress conversions"

johnham-ant:
### Features and Changes

**Problem**

Today, **Exclude in-progress conversions** is not supported by Incremental Pipeline mode, automatically moving the experiment to ExperimentResultsQueryRunner which is more expensive to run on large data sets.

We now close this gap and properly support this setting in the Incremental Pipeline.

**What this PR does**

- **Read-time units filter.** The statistics query filters the units table to `first_exposure_timestamp <= min(phase end, asOf - longest window in the slice)`. The units cache keeps collecting every exposure up to the phase end, so users are picked up automatically once their window passes.
- **Partition metrics by conversion window.** When the setting is on, `partitionMetricsByConversionWindow` splits a slice into one stats query per conversion window (`_cw<minutes>m` suffix, e.g. `_cw90m`), so every metric in a query shares one cutoff and one user count. This intentionally keys on the same `getMaxHoursToConvert` call the stats CTE uses, rather than reusing `getFactMetricGroup`, so the partition can never disagree with the read-side cutoff.
- **Cross-FT ratio metrics** are partitioned the same way: `buildCrossFtSubGroups` takes a `getWindowKey`, so mixed-window ratios over the same two caches split per window. The window is computed *after* metric overrides (`getOverriddenMetricConversionWindowHours`), since cross-FT fans out from raw metrics.
- **Exploratory (Dimension) refreshes** anchor the cutoff to `asOf` = the last Overall snapshot's `dateCreated` (`sourceSnapshotDateCreated`), so a stale cache never admits a unit whose window extends past the cached data. 
- **Sub-hour window fix.** `getExperimentEndDate` now subtracts elapsed milliseconds instead of `setHours`, so 1-minute / 30-minute windows aren't integer-truncated.

### Testing

- Unit tests
- Manual tests


ahdriel:
cursor review

lukesonnet (approved):

PR #6711, opened 2026-08-21
Apply segment and query filter in the incremental units table query

johnham-ant:
### Features and Changes

**Problem**

In Incremental Pipeline mode, the units table query (`getUpdateExperimentIncrementalUnitsQuery`) defines a `__segment` CTE when the experiment has a segment, but never joins it: `__filteredNewExposures` selects straight from `__newExposures` and the rows are merged into the units table. The experiment's SQL query filter is never applied on this path either. Both settings are honored by the standard (non-incremental) units query, so an experiment with a segment or query filter that is switched to incremental refresh silently publishes full-population results under the segment label. We hit this in production with a segmented experiment.

**Fix**

`__filteredNewExposures` now filters the new exposures the same way the standard units query does:

- inner join `__segment s` on the base id type, with `s.date <= e.timestamp`;
- add the query filter as an extra `AND (...)` clause on the exposure rows.

The exposure CTE is aliased as `e` so the segment join and the query filter can reference it. When neither a segment nor a query filter is set, the generated query is unchanged apart from the alias.

Only new exposures need filtering: units already in the table were filtered when they were first inserted, and both `segment` and `queryFilter` are in `INCREMENTAL_FULL_REFRESH_SETTINGS_FIELDS`, so changing either already forces a full rebuild of the units table.

We do need to invalidate existing caches that silently hid this functionality as it would change before and after this PR lands.

Also included, since it came up while exercising the fix through the API: `updateExperiment` mapped `segmentId` with a truthiness check (`segmentId ? { segment: segmentId } : {}`), so `segmentId: ""` was dropped and there was no way to clear a segment via the public API. It now checks `!== undefined`, matching how `queryFilter` and the other clearable fields are handled. One-line change in `services/experiments.ts`.

Back-end only. No schema or front-end changes.

### Dependencies

None.

### Testing

New `packages/back-end/test/integrations/incrementalRefreshUnitsFilters.test.ts` builds the incremental units query through the BigQuery integration and checks the `__filteredNewExposures` CTE:

- with a segment: `JOIN __segment s ON (s.user_id = e.user_id)` and `AND s.date <= e.timestamp` are present;
- with a query filter: the filter appears as an `AND (...)` clause and no `__segment` CTE is emitted;
- with neither: no join, no segment predicate, and only the experiment id and timestamp bounds remain in the `WHERE`.

Two cases added to `updateExperimentApiPayloadToInterface` in `test/services/experiments.test.ts`: `segmentId: ""` clears the segment, and an omitted `segmentId` leaves it untouched.

`tsc` (back-end, shared), eslint and prettier are clean on the touched files; the full `packages/back-end` jest suite passes.

### Notes for reviewers

- The segment join uses the same shape as the standard units query (`getExperimentUnitsQuery`), so a segment that is valid there behaves identically here.
- The exposure columns (id, variation, timestamp, experiment dimensions) are now qualified with `e.` because the joined segment CTE carries the same id column, which would otherwise be ambiguous.


lukesonnet:
@cursor review

packages/back-end/src/enterprise/services/data-pipeline.ts
```diff
@@ -115,6 +115,12 @@ export function getExperimentSettingsHashForIncrementalRefresh(
     settingsForHash[field] = snapshotSettings[field];
   }
 
+  // Incremental units SQL used to ignore segment and queryFilter before #6711.
+  // Salt only those hashes so pre-fix tables refresh; unfiltered tables are still valid.
+  if (snapshotSettings.segment || snapshotSettings.queryFilter) {
+    settingsForHash.unitsFiltersApplied = true;
```
ahdriel, line 121: question: don't we need to add `segment` and `queryFilter` to the settings so the hash gets modified if the segment or queryFilter gets updated?

I am pro making sure to invalidate just the hash of the ones that use segments/queryFilters -- meaning that yay for the conditional expansion of the settings object.

lukesonnet: They're already there.

```
export const INCREMENTAL_FULL_REFRESH_SETTINGS_FIELDS = [
  "activationMetric",
  "attributionModel",
  "queryFilter",
  "segment",
  "skipPartialData",
  "datasourceId",
  "exposureQueryId",
  "startDate",
  "regressionAdjustmentEnabled",
  "experimentId",
] as const satisfies readonly (keyof ExperimentSnapshotSettings)[];
```

ahdriel: oooh got it!

nice

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -1720,39 +1719,45 @@ export default abstract class SqlIntegration
...
+              , e.variation
+              , e.timestamp
+              , MAX(e.timestamp) OVER () AS max_timestamp
+              ${activationMetric ? `, e.activation_timestamp` : ""}
+              ${experimentDimensions.map((d) => `, e.dim_exp_${d.id}`).join("\n")}
+            FROM __filteredNewExposures e
+            ${
+              segment
+                ? `JOIN __segment s ON (s.${baseIdType} = e.${baseIdType})
+            WHERE s.date <= ${this.getSqlDialect().castUserDateCol("e.timestamp")}`
```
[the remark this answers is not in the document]
lukesonnet: In the interest of simplicity, I think we're ok with slight re-scanning in order to prevent having to scan even more data for the max timestamp.

lukesonnet (approved):

ahdriel (approved):

PR #6712, opened 2026-08-21
Prep the attributes table for column customization

Kevin-Chant:
### Features and Changes

Groundwork for #6713 and #6716, split out so those read cleanly.

`AttributeRowMenu` moves out of the page component so re-renders stop remounting it and losing its open state, and the references modal is keyed by property rather than an array index. The internal `POST`/`PUT /attribute` now require `property` to be at least one character, as the REST API already does.

### Testing

- [x] Open a row's kebab menu, let references load -> menu stays open
- [x] Click "N references" on several rows -> modal shows that row's attribute
- [x] `PUT /attribute` with `property: ""` -> 400

### DIFF
--- a/packages/back-end/src/routers/attributes/attributes.router.ts
+++ b/packages/back-end/src/routers/attributes/attributes.router.ts
@@ -21,7 +21,7 @@ router.post(
   "/",
   validateRequestMiddleware({
     body: z.strictObject({
-      property: z.string(),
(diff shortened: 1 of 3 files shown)


Kevin-Chant:
@greptile There's no @/ui/IconButton wrapper to switch to, and this import is also pre-existing: the PR moves it verbatim out of attributes.tsx rather than introducing it. Leaving as-is.

PR #6722, opened 2026-08-21
fix(back-end): respect case-sensitivity for warehouses properly when validating query

ahdriel:
### Features and Changes

Each warehouse handles column names differently `ClickHouse` is case sensitive, `SnowFlake` is case insensitive but it uppercases the column names before returning, `BigQuery` is case insensitive but returns the column the same way we sent it, for example.

So we are fixing a bug where:
- Validation was failing for BigQuery for columns with uppercase names (we were searching for `userid` instead of `userId` and complaining it was missing)
- Handle column names more consistently between warehouses in general
- Handle case sensitive checks (enabled only for Clickhouse)

### Testing

- Unit tests


ahdriel:
Addressed the Greptile concern in `3b20be8f2f`. Metadata case folding is now an explicit integration capability. BigQuery and Snowflake opt in because their adapters lowercase `TestQueryResult.columns`; Presto and integrations without the capability retain exact matching. Added tests for opt-in matching, absent and false defaults, and exact row fallback.

ahdriel:
cursor review

ahdriel:
cursor review

jdorn (approved):

ahdriel:
cursor review

PR #6723, opened 2026-08-21
fix(ui): improve callout action wrapping

ahdriel:
### Features and Changes

Fix a scenario where a Callout with short text + short action was wrapping into a new line when it shouldn't.

### Screenshots

| Before | After |
|--------|--------|
| callout-final-before | callout-final-after |


jdorn (approved):
Tested on my branch and it fixed the issues

PR #6724, opened 2026-08-21
Load the Reo docs tracker as a Mintlify custom script.

alyssamichelle:
The Docusaurus scripts entry was dropped in the migration, so the snippet under static/scripts never ran.

For Mintlify, any .js in the docs folder is included automatically, which is how docs/reo.js, docs/kapa-widget.js, and docs/docs-ui.js all run.


alyssamichelle:
Closing to replace with a clean rename-only PR from a fresh branch.

PR #6727, opened 2026-08-21
Load the Reo docs tracker as a Mintlify custom script.

alyssamichelle:
## Summary
- Move `docs/static/scripts/reo.js` to `docs/reo.js` so Mintlify auto-loads it as a custom script (same snippet, no rewrite).
- Point `.prettierignore` at the new path so the minified IIFE stays untouched.
- Replaces closed #6724, which rewrote the snippet instead if just renaming the file.


Kevin-Chant (approved):
The mintlify deploy for the PR is loading the script, lgtm

PR #6729, opened 2026-08-21
ci: fail docs CI on broken heading anchors

alyssamichelle:
Mintlify slugs differ from the old Docusaurus ones, so in-page jumps were silently landing at the top of the right page. Enable --check-anchors and fix the true-positive mismatches.


lukesonnet (approved):
this seems to also fix a bunch of broken links!!

PR #6733, opened 2026-08-22
Add support for factTable.timestampColumn in generated SQL

jdorn:
### Features and Changes

We want `select * from my_table` to be a valid fact table definition.  No aliases just to satisfy GrowthBook.

First step of this is `timestamp`.  Now, fact tables can specify a `timestampColumn`. When set, it will be used instead of assuming the column is called `timestamp`.

The real column is aliased to `timestamp` in the very first CTE, so the vast majority of SQL generation code remains untouched.  Only a handful of places needed updating.

**Note**: This PR only adds support in the data model / SQL generation.  The new field is NOT exposed in the UI or API yet, so this change should be a complete no-op on merge.

There will be a follow-up PR in this stack doing the same thing for identifier columns.

### DIFF
--- a/packages/back-end/src/integrations/sql/ctes/fact-metric-cte.ts
+++ b/packages/back-end/src/integrations/sql/ctes/fact-metric-cte.ts
@@ -1,6 +1,7 @@
 import {
   getColumnRefWhereClause,
   getFactTableTemplateVariables,
+  getFactTableTimestampColumn,
   isFactFunnelMetric,
   isRatioMetric,
   parseSliceMetricId,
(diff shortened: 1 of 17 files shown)


packages/back-end/src/integrations/sql/ctes/fact-metric-cte.ts
```diff
@@ -69,8 +70,12 @@ export function getFactMetricCTE(
     }
   }
 
+  // The fact table's timestamp column, aliased to `timestamp` on the way out so
+  // downstream SQL never has to know its real name.
+  const timestampColumn = `m.${getFactTableTimestampColumn(factTable)}`;
+
   // BQ datetime cast for SELECT statements (do not use for where)
-  const timestampDateTimeColumn = dialect.castUserDateCol("m.timestamp");
+  const timestampDateTimeColumn = dialect.castUserDateCol(timestampColumn);
```
lukesonnet, line 78: technically I think *this* is the one that gets aliased, not the one above, which only gets used for the pushed up where clause.

lukesonnet (approved):

PR #6740, opened 2026-08-23
New Fact Table Creation Flow

jdorn:
### Features and Changes

A brand new streamlined flow for creating fact tables.

Step 1: Write SQL with ability to test / view sample rows.

image

Step 2: Configure the fact table / column mappings (auto-detected when possible)

image


packages/shared/src/validators/fact-table.ts
```diff
@@ -41,6 +41,12 @@ export const numberFormatValidator = z.enum([
   "memory:kilobytes",
 ]);
 
+export const factTableTypeValidator = z
+  .enum(["event", "model", "rollup"])
```
mknowlton89, line 46: I think we need to add "other" here to support the 4th table type option shown in the UI. It's currently not supported and throws a 400 error.

Image

packages/front-end/components/FactTables/NewFactTableSqlStep.tsx
```diff
@@ -0,0 +1,514 @@
...
+              Your warehouse reported no output columns for this query.
+              Double-check the SQL, then run it again.
+            </Callout>
+          )}
+
+          <Flex align="center" justify="between" gap="3">
+            {detected?.length ? (
+              <Callout status="success" size="sm">
+                Query ran successfully
+              </Callout>
```
mknowlton89, line 452: Small nit: but I think we may want to key this off the `hasFreshResults` - otherwise a user could run a successful query, change it, and the `Query ran successfully` persists.

mknowlton89 (commented):
Found one small tweak to make to the fact table validator around the `tableType` property.

I top-hatted it and aside from the `other` fact table type, it all worked well - I like the inclusion of the `tableType` property - I think that'll be really nice moving forward.

A few small comments/questions:
1) Can you outline the reasoning for the new `View sample rows` link button - I found it a bit odd that clicking that ran the query, given that it had the downward caret to the left, I was just expecting it to open a drawer. I was also a bit confused about why the tooltip was rendering on hover - the tooltip made sense once I clicked the link button.

2) I'm also curious about the new pattern for the results - `DisplayTestQueryResults` does already have a success indicator, so we're doubling that up when the `DisplayTestQueryResults` is rendered.

I do have an update to the `DisplayTestQueryResults` component in the sql-explorer-v2 branch that adds a green check to make this slightly more prominent. Screenshots below.

Current experience in this branch:
Image

And the new (not yet approved experience in the sql-explorer-v2 branch). Low fidelity screenshot, sorry. The copy for `Sample Results` and others is configurable.
Image

3) I did feel a little overwhelmed on the second page of the modal and it wasn't super clear on what I was supposed to do. I think some of that was because on the first page (and in many other places in the app) when there are two columns, the left is the column where the action is taken, and the right is metadata that is helpful, and the second page has the opposite. I understand why, but it felt a little confusing/jarring.

PR #6743, opened 2026-08-24
Clean up Sentry error logging

jdorn:
### Features and Changes

As an extra layer of security, instead of relying on app settings within Sentry to redact sensitive info, make this redaction explicit in the code.  Also, cleans up some other inconsistencies with error logging code.


ahdriel:
cursor review

ahdriel (approved):

PR #6745, opened 2026-08-24
Fix saved-group conflict resolution follow-ups

bryce-fitzsimons:
Fixes two bugs from the concurrent-edit guard: clicking "keep mine" on a saved-group list conflict no longer re-warns when you save again, choosing "create a new draft" now clears the conflict and saves, and huge ID lists are no longer shipped back in conflict responses (you get a Reload button instead).

### DIFF
--- a/packages/back-end/src/routers/saved-group/saved-group.controller.ts
+++ b/packages/back-end/src/routers/saved-group/saved-group.controller.ts
@@ -499,6 +499,9 @@ type PutSavedGroupRequest = AuthRequest<
   }
 >;
 
+// Cap on a single conflict value shipped back for display in a 409.
+const MAX_CONFLICT_VALUE_BYTES = 20_000;
+
 type PutSavedGroupResponse =
   | {
       status: 200;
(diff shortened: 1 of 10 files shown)


oelshaikh:
Still seeing one issue regarding saved groups:

1. Change the name in tab A without saving.
2. Change and save a different name in tab B.
3. Tab A revalidates and its unsaved name is replaced by tab B’s value.
4. Saving tab A succeeds without a conflict because it now submits tab B’s value.
5. Tab A changes are ignored silently

AI hint on fix: 

The useEffect around SavedGroupForm.tsx:254-297 resets the entire form whenever currentRevision or liveVersion changes. Consequently, the request never reaches the conflict path.

Could we pin the form state/baseline while the modal is open and only reset when the user explicitly selects a different revision? Updates to the same revision from SWR should preserve dirty fields so the stale baseline reaches the server and produces the expected 409.

oelshaikh (approved):
LGTM, tested and previous issue seems to be fixed

PR #6749, opened 2026-08-24
fix(holdouts): Include Holdouts targeting in payload

msamper:
### Features and Changes

This PR fixes a bug where targeting on a holdout was not being included when generating the holdouts payload. This PR also ensures that when we write any targeting or traffic changes to a running holdout they're written to both the phases if they exist so that when we show results we're using the correct coverage for scaled impact since snapshots are based off of the latest phase for experiments. A consequence of this fix is that orgs that have a running holdout thats being included in the payload will have targeting conditions applied for their holdout the next time an event triggers an SDK payload refresh. 

Also added some util functions that encapsulate logic for getting the active phase for an experiment and a small fix for editing environments when new environments have been added to the org since holdout creation. 
Only applying the new util function to the changes in this PR to keep the scope here tight, but we can use them in the future to avoid branching logic around getting phase 0 for holdouts and getting the latest phase for other types of experiments.

### Testing

- Unit Tests
- Verified saved groups and conditions now show up in the payload for holdouts
- Verified targeting changes are now mirrored across phases
- Verified runtime error when trying to edit environments after new environments have been added to the org is fixed


msamper:
@greptileai

packages/back-end/src/controllers/experiments.ts
```diff
@@ -1682,25 +1682,31 @@ export async function postExperiment(
     validateVariationIds(data.variations);
   }
 
-  const latestPhase = experiment.phases?.[experiment.phases.length - 1];
+  const payloadPhase =
+    experiment.type === "holdout"
```
lukesonnet: seems like it would be safer to just divert to a different helper if it is experiment type holdout and duplicate some code instead of all the branching in line below.

msamper: Added a helper and cleaned up some branching logic around latest/active phase with a new util function

packages/back-end/src/controllers/experiments.ts
```diff
@@ -2052,35 +2065,43 @@ export async function postExperiment(
...
 
-  if (data.variationWeights) {
-    changes.phases = applyVariationWeightsToLatestPhase(
-      experiment,
-      data.variationWeights,
-    );
-  }
+  if (experiment.type === "holdout") {
+    // Holdouts have fixed variations/weights; coverage is the only
+    // payload-affecting field an update can change, and it applies to every phase.
```
lukesonnet: Not prerequisites or targeting?

msamper: Holdouts don't support pre-reqs and updating targeting is done via the targeting endpoint internally. I can clarify in the comment

lukesonnet (approved):

PR #6753, opened 2026-08-25
Fix OAuth login failures with many open tabs and return to the original page

Kevin-Chant:
### Features and Changes

Fixes the `OAuth Error: Error Signing In` that shows up when several GrowthBook tabs re-authenticate at once, and the redirect to the home page after pressing Retry. About **10%** of prod `/auth/callback` requests currently fail this way.

The root cause is that `AUTH_CHECKS` held a single pending `state`/PKCE verifier per browser, so concurrent logins overwrote each other. We now derive those values per flow from one long-lived `AUTH_SECRET` cookie instead of storing them, so there's nothing for tabs to race on. Since `state` is no longer single-use, each one embeds its mint time and callbacks older than an hour are rejected.

Design note: we considered the more common per-flow transaction cookie (single-use `state`, the Auth.js/Auth0 SDK pattern) and opted for the smaller stateless surface instead, since expiry plus PKCE and the single-use authorization code cover replay. Happy to swap to per-flow cookies if reviewers think strict single-use matters here.

Enterprise SSO visits also change visibly: main computed the "Sign In Required" confirm *after* writing the checks cookie, so the modal appeared on every unauthenticated visit and dropped the deep link. It now only appears when a login started in the last 10 minutes was left unfinished; first visits redirect straight to the IdP.

On the front-end, a failed callback now checks whether another tab already logged in and continues instead of erroring. Retry and the Vercel SSO return page go back to the saved page instead of home; the saved path is consumed on use and any stale `org` param in it is replaced.

### Known issues

A brand-new browser (no `AUTH_SECRET` yet) starting several flows in the same instant keeps only the last secret; those tabs fail once and recover via the refresh fallback.

### Testing

Verified on a dev stack pointed at a real Auth0/Google `SSO_CONFIG`:

- [x] Sign in end to end -> the derived state/PKCE values round-trip a real IdP
- [x] Two tabs with expired sessions -> both succeed, each lands on its own experiment
- [x] Deep link while logged out -> returns there after sign-in, not home

Still needs Cloud:

- [ ] Retry on the error page -> returns to the page the session expired on
- [ ] Vercel SSO re-auth from an experiment page -> returns to that experiment
- [ ] Enterprise SSO: abandon a login, revisit within 10 minutes -> confirm modal; complete it -> lands on the saved page
- [ ] Post-deploy: "Error signing in" in the prod API logs drops from ~2.9k/week

### DIFF
--- a/packages/back-end/src/services/auth/OpenIdAuthConnection.ts
+++ b/packages/back-end/src/services/auth/OpenIdAuthConnection.ts
@@ -20,7 +20,9 @@ import {
 import { AuthRequest } from "back-end/src/types/AuthRequest";
 import { MemoryCache } from "back-end/src/services/cache";
 import {
-  AuthChecksCookie,
+  clearPendingAuthChecks,
+  getPendingAuthChecks,
+  setPendingAuthChecks,
   SSOConnectionIdCookie,
 } from "back-end/src/util/cookie";
 import { APP_ORIGIN, IS_CLOUD, USE_PROXY } from "back-end/src/util/secrets";
@@ -41,9 +43,6 @@ type AuthChecks = {
   state: string;
   code_verifier: string;
 };
-
-// The cookie holds one entry per in-flight login so concurrent tabs don't clobber each other
-const MAX_PENDING_AUTH_CHECKS = 10;
 
 if (USE_PROXY) {
   custom.setHttpOptionsDefaults(getHttpOptions());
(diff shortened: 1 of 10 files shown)


packages/back-end/src/services/auth/OpenIdAuthConnection.ts
```diff
@@ -127,26 +130,26 @@ export class OpenIdAuthConnection implements AuthConnection {
...
-    // Get rid of temporary codeVerifier cookie
-    const checks = this.getAuthChecks(req);
-    AuthChecksCookie.setValue("", req, res);
+    const params = client.callbackParams(req.originalUrl);
 
-    if (!checks) {
-      throw new Error("Missing auth checks in session");
+    const secret = AuthSecretCookie.getValue(req);
+    if (!secret) {
+      throw new Error("Missing auth secret cookie");
```
ahdriel, line 137: question: does this mean that everyone will hit this error after deployment? meaning that the race condition between tabs will happen for any user that has multiple tabs open after this gets merged, right?

anything we can do to avoid it?

ahdriel (approved):

packages/back-end/src/services/auth/OpenIdAuthConnection.ts
Kevin-Chant: Yes it'll be a race for any multi-tab cases (which are already pretty bad today), but the improved Retry should help smooth that out (dropping them back in the right place instead of home page). There's a way to avoid it but it's a bit hacky and I don't think it's worth the extra code complexity

PR #6754, opened 2026-08-25
Perf: cut CI test time with per-entity matrix split, sharding, and parallel jobs

Kevin-Chant:
### Features and Changes

Our unit tests have gotten very slow lately, about 5x duration in the past 5 months. That growth is mostly from new test files, so we might want to consider trimming some of the AI-added tests if they're not pulling their weight.

For now, there's 4 wins for speeding up the test suite:
* Split the entity permission matrix test file so Jest can shard the tests. Same 1088 tests, but **168.6s -> 94.6s.**
* Shard the back-end suite into 4 parallel tracks.  Back-end's limit becomes the slowest shard instead of the total, or **248.4s -> 79s**
* Swap CI from sequential jobs to a parallel matrix. Shared setup happens first, and we drop **~225s off the serial path.**
*  Enable Jest's `cacheDirectory` to stop transpiling the whole TypeScript graph every time. **241s -> 205s.**

Total CI speedup approx **15m -> 5m**

### Testing

Measured locally, back-end at `--maxWorkers=3` to match the cloud runner. Confirmed test counts are identical before/after:
- [x] Permission matrix: 1088 tests before and after the split
- [x] Back-end: 8187 tests pass, splitting 4084/1588/1397/1118 across the four shards
- [x] `pnpm type-check`, `pnpm lint:ci` and `pnpm pretty:check` clean

Back-end suite, 3 workers:

```
before   248.4s
shards    79s / 57s / 51s / 55s
```


.github/workflows/ci-nochanges.yml
Kevin-Chant, line 1: This was conflicting with the main `ci` if both `paths-ignore` (from here) and `paths` (from ci) matched, so it'd be better as a filter in the main workflow

.github/workflows/ci.yml
```diff
@@ -2,55 +2,95 @@ name: CI
...
+            exit 0
+          fi
+
+          echo "Changed files:"
+          echo "$files"
+          if echo "$files" | grep -qE '^(packages/|eslint-rules/|\.github/(workflows/ci\.yml|actions/)|pnpm-lock\.yaml$|package\.json$|eslint\.config\.mjs$|eslint-suppressions\.json$)'; then
+            echo "code=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "code=false" >> "$GITHUB_OUTPUT"
+          fi
```
najito: Folding `ci-nochanges` into this workflow was the right fix for paths vs paths-ignore double-firing.

This job then reintroduces path policy as a shell script: `gh` api, pagination, fail-open, and a regex that has to stay in sync with every job's `if:` and with the aggregator treating `skipped` as success.

Now that there is only one workflow, is a detector still necessary?

packages/back-end/test/api/permission-matrix-revision-entities.fixture.ts
```diff
@@ -105,6 +108,8 @@ const ENTITIES: Entity[] = [
   },
 ];
 
+export const [CONSTANTS, CONFIGS, SAVED_GROUPS] = ENTITIES;
```
najito: The per-file split and `describeEntityMatrix()` are the right Jest workaround.

This binds CONSTANTS/CONFIGS/SAVED_GROUPS to array order. Reordering ENTITIES would silently retarget the Constants-only cases in `permission-matrix-constants.test.ts`. Just FYI

packages/front-end/vitest.config.js
```diff
@@ -4,7 +4,8 @@ import { defineConfig } from "vite";
 export default defineConfig({
   test: {
     globals: true,
-    environment: "jsdom",
+    // Files that need one opt in with a `// @vitest-environment jsdom` docblock.
+    environment: "node",
```
najito: This is ~6s (82.9 → 77.1) while collect stays the bottleneck. In exchange, every new render/renderHook spec must remember `// @vitest-environment jsdom` or fail oddly.

Setting the environments in one place (`environmentMatchGlobs` / directory rules) with separate setup files could be good here if there's a need in the future to toggle back and forth.

.github/workflows/ci.yml
```diff
@@ -77,39 +116,135 @@ jobs:
...
+        uses: actions/cache@v5
+        with:
+          path: packages/shared/.jest-cache
+          key: ${{ runner.os }}-jest-shared-${{ hashFiles('pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-jest-shared-
+
+      # stats (pytest) runs in its own job; back-end is sharded above.
+      - name: Test
+        run: pnpm --filter '!back-end' --filter '!stats' -r --workspace-concurrency=1 run test
```
najito, line 177: Front-end and shared are still serialized after we parallelized everything else. A 2-wide matrix (or concurrency > 1) would match the rest of the design.

Also: format (`setup-workspace` + full `pnpm install`) is heavier than old ci-nochanges (`--ignore-scripts`). If we keep a docs-only fast path, install should stay as dumb as Prettier needs; if we always run full CI, this goes away with the `changes` job.

.github/actions/setup-workspace/action.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    # Only the build produces tsbuildinfo, so a job that skips it would other-
+    # wise race to save an empty cache under the shared key.
+    - name: Cache tsbuildinfo
+      if: inputs.build-deps == 'true'
+      uses: actions/cache@v5
+      with:
+        path: |
+          **/tsconfig.tsbuildinfo
+        key: ${{ runner.os }}-tsbuildinfo-${{ hashFiles('pnpm-lock.yaml') }}
```
najito: All `build-deps: true` jobs (lint, openapi, type-check, 4 test-back-end shards, test-packages) share one `tsbuildinfo` cache key, but only some of them (e.g. type-check) actually produce the full-repo buildinfo; on a cold cache these jobs race to save under the identical key.

.github/workflows/ci.yml
najito: This also hand-rolls changed-file detection, adding a serial network round-trip and a new single point of failure in front of every gated job on every run.

.github/workflows/ci.yml
```diff
@@ -2,55 +2,95 @@ name: CI
...
         with:
-          node-version: "24"
-          cache: "pnpm"
+          build-deps: "false"
+
+      - name: Check formatting
+        run: pnpm pretty:check
+
+  lint:
+    needs: changes
```
najito: The code-change gate is implemented as three independently-maintained lists that must move together: the regex in `changes`, the per-job `needs: changes` + `if: needs.changes.outputs.code == 'true'` pair copy-pasted across 6 jobs, and the aggregator `ci` job's `needs: [...]` array.

.github/actions/setup-workspace/action.yml
```diff
@@ -0,0 +1,42 @@
...
+        restore-keys: |
+          ${{ runner.os }}-tsbuildinfo-
+
+    - name: Install pnpm dependencies
+      run: pnpm install
+      shell: bash
+
+    - name: Build dependencies
+      if: inputs.build-deps == 'true'
+      run: pnpm build:deps
```
najito, line 27: The composite action's `pnpm install` step has no `--ignore-scripts` option, unlike the deleted `ci-nochanges.yml` which used it; the `format` job (build-deps: false) now always pays for full postinstall lifecycle scripts.

.github/workflows/ci.yml
```diff
@@ -77,39 +116,135 @@ jobs:
...
+          echo "Upstream results: $RESULTS"
+          for r in $RESULTS; do
+            case "$r" in
+              success | skipped) ;;
+              *)
+                echo "A required job did not succeed."
+                exit 1
+                ;;
+            esac
+          done
```
najito: The `ci` aggregator job hand-rolls a bash for-loop/case-statement over `join(needs.*.result, ' ')` to detect any non-success/non-skipped dependency, reimplementing what a single GitHub Actions expression (e.g. `contains(needs.*.result, 'failure')`) already provides natively.

najito (changes requested):

.github/workflows/ci.yml
Kevin-Chant: Fair. I had left the detector logic because of the old paradigm where CI is expensive, but dropping it will take our non-code PRs (about 10% of recent merges) from 80s -> 270s which is probably worth it for the simplification

.github/workflows/ci.yml
Kevin-Chant: True, but test-packages runs faster than the current critical path, so splitting it wouldn't drop CI time. Leaving it serial for now.

.github/actions/setup-workspace/action.yml
Kevin-Chant: Checked this but the only lifecycle scripts are the root postinstall (no-ops without poetry) and prepare: husky install. So the practical cost should be 0

packages/front-end/vitest.config.js
Kevin-Chant: Yeah collect is the bottleneck and it looks like there's a good solution there but it won't move CI (local testing win only) & is a bit larger than I want to fold into this PR. 
~For this 6s win, `environmentMatchGlobs` is deprecated, and directory rules doesn't cleanly handle our usage, so I'm gonna leave it with the docblock & just add an AI-facing note. If it still gets forgotten, it's a loud failure that'll be an easy fix.~
Actually I just hit a CI failure because of it and it's probably not worth the effort. I'm cutting this out, and the larger change is still a separate PR to make local front-end runs faster

.github/workflows/ci.yml
```diff
@@ -77,39 +74,119 @@ jobs:
...
+  ci:
+    # `always()` matters: without it this job is skipped when a dependency
+    # fails, and branch protection counts a skipped required check as passing.
+    if: always()
+    needs:
+      [format, lint, stats, openapi, type-check, test-back-end, test-packages]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if any required job did not succeed
+        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
```
najito, line 191: Nothing in this workflow skips today, so this is consistent.
If a later PR puts `if:` on one of the `needs` jobs, a skipped required check can go green again — same class of hole this branch already hit once. Adding `contains(needs.*.result, 'skipped')` here would close that. Not blocking.

najito (approved):
Looks good. Just the single comment about a missed case but it's a bit of an edge. I'll approve and you can address if you want or it can be done by someone who makes that case more relevant.

PR #6756, opened 2026-08-25
feat(experiments): funnels in incremental refresh

lillialjackson:
### Features and Changes

Adds incremental pipeline compatibility for funnel metrics

### Dependencies

multifact table additions need to be merged and then added

### Testing

manual and unit

### Screenshots

First run:
Screenshot 2026-08-26 at 11 02 05 AM
Screenshot 2026-08-26 at 11 02 38 AM
Screenshot 2026-08-26 at 11 02 44 AM

Next run after new data added:

Screenshot 2026-08-26 at 11 05 57 AM
Screenshot 2026-08-26 at 11 07 03 AM


packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts
```diff
@@ -649,7 +650,7 @@ const startExperimentIncrementalRefreshQueries = async (
     // half-populated columns.
     const sameFtMetrics = group.metrics.filter(
       (m) =>
-        m.numerator?.factTableId === group.factTableId &&
+        getFactMetricPrimaryFactTableId(m) === group.factTableId &&
```
lukesonnet: I think this entire filter statement can be simplified to get all tables, and make sure that the metric only has one table and it is the `group.factTableID`.

packages/back-end/src/integrations/dialects/base.ts
```diff
@@ -77,6 +77,12 @@ export const baseDialect: Omit<SqlDialect, "unpivotLabeledPairs"> = {
     );
   },
 
+  arrayConcatAgg: () => {
+    throw new Error(
+      "Merging arrays across rows is not supported by this data source.",
```
lukesonnet, line 82: Nit: I assume some of the other engines do support this merging, it's just not important for us to figure it out and support it for this PR as it is only used in incremental refresh. Probably no action to take here, but we could (1) try our best to implement it in the other integrations or (2) think about whether there's a better way to make it so future extenders of incremental refresh (IR) realize they need to implement this method for the engine they're extending IR to if they want to support funnels.

packages/shared/types/integrations.d.ts
```diff
@@ -47,8 +47,14 @@ export type DataType =
   | "boolean"
   | "date"
   | "timestamp"
+  // Fact-table event-timestamp type (what `castUserDateCol` produces): DATETIME
+  // on BigQuery, TIMESTAMP elsewhere. Distinct from `timestamp` (used for
+  // units/refresh columns, genuinely TIMESTAMP) — funnel step caches store
+  // event timestamps and must match the resolver's DATETIME arithmetic on BQ.
```
lukesonnet, line 53: Should we cast BQ events to timesetamp to keep this simple and harmonize across DBs? I think this is only used for step 0, since we use arrayTimestamp for other steps. So I'm not sure there's an advantage to keeping this just for the step 0 on BQ if timestamp is free for us to use with just a CAST.

packages/back-end/src/services/experimentQueries/planMetricFanOut.ts
```diff
@@ -101,13 +119,23 @@ export function planMetricFanOut(metrics: FactMetricInterface[]): MetricFanOut {
   };
 
   metrics.forEach((metric) => {
-    const numeratorFactTableId = metric.numerator?.factTableId;
+    const numeratorFactTableId = getFactMetricPrimaryFactTableId(metric);
```
lukesonnet: nit: The error and function name don't match now that we get either the first step OR numerator fact table with this method.

packages/back-end/src/services/experimentQueries/planMetricFanOut.ts
```diff
@@ -33,16 +46,20 @@ export interface MetricFanOut {
     factTableIds: [string, string];
     metrics: CrossFtRatioMetric[];
   }>;
+  // Funnel metrics whose steps span 2+ fact tables. Each needs a multi-source
+  // stats query that joins all of its caches. Funnels sharing the same FT set
+  // can be grouped into a single stats query by the runner.
+  multiFtFunnels: MultiFtFunnelGroup[];
```
lukesonnet: I believe this choice means that funnels are always in a separate stats query to the cross fact table ratio metrics that require the same set of sub-tables. It could reduce some user's query loads if we were able to group ratio metrics on {ft_A, ft_B} with funnels on the same 2 fact metrics.

A separate design could be something like:

```
multiFtMetrics: Array<{
    factTableIds: string[]; // sorted array (or otherwise easy way to look up this id set)
    metrics: FactMetricInterface[]
}>
```

Then in `buildCrossFtSubGroups` you can properly build the right statistics sub-groups from single object for both ratio and funnel instead of doing ratio metrics in `buildCrossFtSubGroups` and funnel metrics in line in ExperimentIncrementalRefreshQueryRunner here (and duplicated in the Exploratory runner): https://github.com/growthbook/growthbook/blob/cd34926405e5fc56c1b3761ab684730ccdd2cae8/packages/back-end/src/queryRunners/ExperimentIncrementalRefreshQueryRunner.ts#L1152-L1195.

packages/back-end/src/integrations/sql/ctes/funnel-resolution-cte.ts
```diff
@@ -31,6 +31,39 @@ export function funnelStep0NeedsExposureWindow(
   return !!metric.funnelSettings.steps[0]?.conversionWindow;
 }
```
lukesonnet: We can delete this method now, I think it isn't needed.

packages/back-end/src/enterprise/services/data-pipeline.ts
```diff
@@ -221,6 +222,9 @@ export function getMetricSettingsHashForIncrementalRefresh({
     denominator: factMetric.denominator,
     cappingSettings: factMetric.cappingSettings,
     quantileSettings: factMetric.quantileSettings,
+    ...(isFactFunnelMetric(factMetric)
+      ? { funnelSettings: factMetric.funnelSettings }
```
lukesonnet, line 225: Technically, everything in here needs to be hashed for reliability EXCEPT `funnelSettings.steps.[].name`... so we could build a helper to remove that field and pass through the rest (defaulting to passing through any new fields an engineer might add), or we could ignore it.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2244,6 +2311,27 @@ export default abstract class SqlIntegration
             , ${this.getSqlDialect().castToDate("timestamp")} AS metric_date
             ${metricData
               .map((m) => {
+                if (isFactFunnelMetric(m.metric)) {
```
lukesonnet, line 2414: question: Do we do this like this because extending the `aggFunction` to be one per step was too complicated? I like the structure of that approach (e.g. each step has its own "Agg" and "ReAgg" methods), but maybe it's too clunky to make it work here.

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2786,8 +2923,44 @@ export default abstract class SqlIntegration
                   )`
               : "";
 
-          return `
-        , __joinedData${sourceSuffix(i)} AS (
+          // Single-FT funnels resolve per-source; multifact funnels are
```
lukesonnet, line 3084: Just a comment: everything below here is really hard to read, but I don't think it's this PRs job to solve the problem...

packages/back-end/src/integrations/SqlIntegration.ts
```diff
@@ -2796,15 +2969,181 @@ export default abstract class SqlIntegration
...
+            if (ftIds.length <= 1) return [];
+            return [{ metric: data.metric, alias: data.alias }];
+          });
+        if (multiFtFunnelMetrics.length === 0) return "";
+
+        // Collect output columns from sources >0 for the flattened join.
+        const sourceColumnsByIndex = new Map<number, string[]>();
+        for (const data of metricData) {
+          if (!isFactFunnelMetric(data.metric)) {
+            // Non-funnel metrics: value/denominator columns
```
lukesonnet, line 3194: note: so like I think all this code is dead UNTIL we resolve the grouping that forces cross fact table metrics to always be in their own stats query (my comment here: https://github.com/growthbook/growthbook/pull/6756#discussion_r3914998366). So we want to keep this ability to pass through cross-ft ratio columns if we do put them in the same query OR we can intentionally keep them separate and remove some of this complexity.

I argue we should just try and do the former (put them all in one query, keep this complexity, but fix the grouping in `planMetricFanOut` as per my other comment).

packages/back-end/src/services/experimentQueries/planMetricFanOut.ts
lukesonnet: Confirmed, they end up in different stats queries, even though we have code in SqlIntegration to handle this case (I think)

lukesonnet (commented):
nit: let's rename the PR to something like `feat(experiments): funnels in incremental refresh` or something like that, please

packages/back-end/src/integrations/dialects/base.ts
lillialjackson: This will throw in our existing pattern, which I think is fairly easy to find in our codebase. I hesitate to surface the issue in more detail through the error message as it does get surfaced to the user. Do you have thoughts on how to make this gap more apparent?

packages/shared/types/integrations.d.ts
lillialjackson: There are some methods pre-existing to this branch that use datetime that would also require refactor and retesting to ensure those paths didn't break, do you feel strongly that this should be done or done now?

packages/back-end/src/enterprise/services/data-pipeline.ts
lillialjackson: Do you think renaming steps will be a frequent occurrence and worth the savings of adding this helper?

packages/back-end/src/integrations/SqlIntegration.ts
lillialjackson: yeah, it required fewer changes this way and fewer impacts to working metric paths.

packages/back-end/src/integrations/SqlIntegration.ts
lillialjackson: attempted to make this a little better but not sure its all that much more readable

packages/back-end/src/integrations/dialects/base.ts
lukesonnet: No, not really :/ would be nice to have a base of methods required for incremental refresh and then not allow a dev to add a dialect to incremental refresh without all of those methods implemented (without showing anything to the user). Fine to punt for now.

packages/shared/types/integrations.d.ts
lukesonnet: We actually had to land a fix recently because we support DATETIME or TIMESTAMP as the datatype for BQ `timestamp` columns in fact tables (https://github.com/growthbook/growthbook/pull/6873).

For this PR: Are we certain that both data types in the Fact Table work with this caching of results as a DATETIME?

I'm also not certain that this would cause a huge refactor if all we have to do is CAST AS TIMESTAMP in the funnel resolver, seems pretty safe, but we can skip this whole thread if the answer to my question above is yes.

packages/back-end/src/enterprise/services/data-pipeline.ts
lukesonnet: I don't think it will be frequent, but I like the idea of our hash not including irrelevant stuff. No strong opinion if the helper is gross.

lukesonnet (approved):

PR #6761, opened 2026-08-25
Skip no-op migration drift in the concurrent-edit guard

bryce-fitzsimons:
Fields added by migrations (like targeting projects) no longer trigger false "modified while you were editing" warnings, and conflict rows now show project names instead of ids.

Previous spurious bug:
image


oelshaikh (approved):
LGTM

PR #6764, opened 2026-08-26
fix(get-started): ensure we render a row per feature version

ahdriel:
### Features and Changes

In the "Feature flags requiring attention" table on the home page, we use `featureId` as the key but a single feature can have several revisions that need attention.

- Using the same `key` leads to duplicated rows as React does not clean them up
- We collide versions into a single row with no differentiator

While we are not changing the UI, we are fixing the `key` and also deep linking to the revision that needs attention, instead of dropping you in the live version.

### Testing

https://github.com/user-attachments/assets/445bdd2a-2f79-4403-bcd9-f14a0d28c68c


gazzdingo (approved):
LGTM

PR #6767, opened 2026-08-26
Silence pm2's getconf spawn errors on the hardened image

Kevin-Chant:
### Features and Changes

pm2's `pidusage` shells out to `getconf` for CPU metrics, which logs an error twice on boot in the shell-less image. It's only cosmetic as the defaults are correct, but we need to set `PIDUSAGE_SILENT` to stop the logging.

### Testing

- [x] Boot the image; no `getconf` errors in the startup logs.


ahdriel (approved):
🚀

PR #6768, opened 2026-08-26
Adds try/catch to ensureDefinitionsVersionIndex

mknowlton89:
### Features and Changes

v5.0 startup awaits a unique index on `definitionsversions` and calls `process.exit(1)` if `createIndex` throws. Cosmos DB for MongoDB cannot add unique indexes to an existing collection (and with continuous backup, unique indexes must be created only at collection creation), so self-hosted Cosmos upgrades fail to boot and look like a login outage.

This change matches `BaseModel.updateIndexes`: log the unique-index failure and continue so the API stays up. On MongoDB/Atlas, `createIndex` remains a no-op success when the index already exists.

Also increases the back-end `watch-dev` delay from 1s to 5s so `pnpm run dev` does not start nodemon before SWC has written `dist/server.js`.

### Dependencies

None

### Testing

- [ ] Boot against MongoDB/Atlas: `definitionsversions` still gets `{ organization: 1 }` unique, no error in logs
- [ ] Boot against Cosmos (or force `createIndex` to throw): process stays up, login works, logs `Error creating unique index for definitionsversions`
- [ ] `GET /organization/definitions` still returns 200 after a definitions write (metric/datasource/etc.)

### Notes

Cosmos still will not have this unique index. Concurrent first-touch upserts for the same org can create two `definitionsversions` docs, which can split ETag reads/writes and 304 stale `/organization/definitions` in the UI (SDK evaluation is unaffected). If they need uniqueness, pre-create the empty collection with Cosmos `customAction: "CreateCollection"` including the unique index — do not drop collections that have data.


packages/back-end/package.json
```diff
@@ -24,7 +24,7 @@
     "generate-dummy-data": "ts-node ./test/data-generator/new-generator.ts",
     "test-integration-queries": "./test/integrations/integration-query-test.sh",
     "watch-compile": "swc src -w --out-dir dist",
-    "watch-dev": "delay 1 && nodemon --exec 'node --enable-source-maps --no-node-snapshot' --watch dist --watch '../shared/dist' -e js ./dist/server.js",
+    "watch-dev": "delay 5 && nodemon --exec 'node --enable-source-maps --no-node-snapshot' --watch dist --watch '../shared/dist' -e js ./dist/server.js",
```
Kevin-Chant, line 27: Is this why I was having trouble booting dev today? I thought it was something I broke in my environment

Kevin-Chant (approved):
Approving, but technically this makes things less safe for non-CosmoDB cases (the index stops being guaranteed for Mongo). Probably not an issue, just worth confirming that we want to go ahead with a targeted fix

jdorn (approved):

PR #6772, opened 2026-08-27
Surface approval requirements earlier and clearer

bryce-fitzsimons:
Squashing some approval flow + team permissions papercuts:

- Drafts now say up front which environments triggered the approval requirement and which teams must approve, before the first review lands. Previously the required-team warning only appeared after an approval, so it was surprising and frustrating.
- Improve clarity about project overrides for approval settings.
- Improving the actual approval required messaging on the Review & Publish tab (dedupe redundant teams, only show the ones on the critical path)
- Refactored the way we display publish blocks and approval requirements using a new easy-to-understand status banner.
- Insufficient approvals now render outlined in the reviewer list and timeline — including an approval from someone who isn't in a required team.
- Only tell someone they can't approve their own draft when that's actually what's blocking them.

image

image

Screenshot 2026-08-26 at 10 20 27 PM


packages/front-end/components/GeneralSettings/ApprovalFlowSettings.tsx
```diff
@@ -143,6 +144,11 @@ export default function ApprovalFlowSettings() {
     (p) => !allTabs.some((t) => scopeProjects(t.scope).includes(p.id)),
   );
 
+  // Overrides are wholesale rules the base tab's settings never reach.
+  const overriddenProjects = [
+    ...new Set(tabs.flatMap((t) => scopeProjects(t.scope))),
+  ].map((id) => ({ id, name: projects.find((p) => p.id === id)?.name ?? id }));
```
Kevin-Chant: Should this be memoized?

Kevin-Chant (approved):
Approving now as it's basically there, see my comment in Slack for a fix to some misleading copy

PR #6773, opened 2026-08-27
Perf: speed up local test runs

Kevin-Chant:
### Features and Changes

#6754 sped up CI by ~3x, but didn't affect local runs of the test suite. Almost all of back-end's time turns out to be module init being re-run for all 309 test files, rather than the tests themselves.

Four wins:
* Stub the packages no test exercises. `googleapis` is the worst, eagerly requiring 472 modules covering every Google API. **-42s**
* Reuse one `mongod` for the whole run instead of the 53 that `api.setup.ts` was spawning per file. **-4s**
* Raise `workerIdleMemoryLimit` to 2GB to prevent excessive worker restarts **-4s**
* Let Vitest's `maxWorkers` scale to machine size **74s -> 32s** (for my 8 core/16 thread laptop)

Back-end **158s -> 106s**, front-end **74s -> 32s**

### Testing

Two runs per step, and test counts are identical before and after so nothing got silently skipped:

- [x] Back-end: 309 suites, 8249 tests, 2254 snapshots pass
- [x] Front-end: 69 files, 1110 tests pass
- [x] Single-file runs unaffected by the shared mongod, 3.47s -> 3.50s
- [x] `pnpm lint:ci` and `pnpm pretty:check` clean

### DIFF
--- a/packages/back-end/jest.config.js
+++ b/packages/back-end/jest.config.js
@@ -1,3 +1,46 @@
+// Packages no test exercises, stubbed to keep their init cost out of 300+ fresh
+// module registries. The googleapis barrel alone eagerly requires 472 modules.
+const STUBBED_DRIVERS = [
+  "googleapis",
+  "@sentry/node",
+  "stripe",
+  "openid-client",
+  "jwks-rsa",
+  "ai",
+  "openai",
+  "@ai-sdk/anthropic",
+  "@ai-sdk/google",
+  "@ai-sdk/mistral",
+  "@ai-sdk/openai",
+  "@ai-sdk/xai",
+  "@dqbd/tiktoken",
+  "tiktoken",
+  "proxy-agent",
+  "kerberos",
+  "@google-cloud/bigquery",
+  "@google-cloud/storage",
+  "@aws-sdk/client-athena",
+  "@aws-sdk/client-sts",
+  "@aws-sdk/client-s3",
+  "@aws-sdk/client-cloudwatch",
+  "@aws-sdk/credential-providers",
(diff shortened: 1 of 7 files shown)


jdorn (approved):

packages/back-end/jest.config.js
```diff
@@ -9,14 +52,16 @@ module.exports = {
...
+  globalSetup: "<rootDir>/test/globalSetup.ts",
+  globalTeardown: "<rootDir>/test/globalTeardown.ts",
   setupFilesAfterEnv: ["<rootDir>/test/jest.setup.ts"],
   // For non-CI, lets make sure to cap the workers
   ...(process.env.CI ? {} : { maxWorkers: "50%" }),
-  // Each file's module graph stays resident (~140MB/file); recycle workers
-  // before the heap fills.
-  workerIdleMemoryLimit: process.env.CI ? "2GB" : "1GB",
+  // Recycle workers before the heap fills; 1GB cost more in restarts than it saved.
+  workerIdleMemoryLimit: "2GB",
```
ahdriel, line 66: I recently reduced this to 1GB because some Cursor cloud agents were dying when running dev server + unit tests.

Have we tested that 2gb with the other changes will not kill the cloud machine?

I did also change to `50%` maxWorkers on the same PR, so maybe 2gb + 50% works just fine.

Kevin-Chant: No I hadn't realized that was an issue. If it happens again then maybe we can swap it based on an env var for the cloud agents but default 2GB otherwise

PR #6774, opened 2026-08-27
Fix Edit Metric Group modal to filter metrics by project scoping

lukesonnet:
### Features and Changes

* Scope metric groups to Projects the same way as metrics: createMetricGroups is project-scoped, and new members must cover every group Project
* Leave existing incompatible groups usable and editable. Rename, reorder, archive, and partial cleanup do not force leftover members out; only new members and Project changes are revalidated.
* Block narrowing a metric’s Projects when that would reduce coverage in a group it belongs to, including groups the editor cannot read.
* Filter the groups list by the current Project, default new groups to that Project, and keep grandfathered members visible but not re-addable in the editor.

### Testing

* Adding project scope to metrics list filters out metrics from other projects.
* Metrics not in selected projects not shown in metrics list


packages/front-end/components/Metrics/MetricGroupsList.tsx
```diff
@@ -147,6 +164,13 @@ const MetricGroupsList: FC = () => {
                   {mg.description}
                 </TableCell>
                 <TableCell>{dsName}</TableCell>
+                <TableCell>
+                  {mg.projects.length === 0
```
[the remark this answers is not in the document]
lukesonnet: I don't think metrics use this.

packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -276,12 +270,40 @@ const MetricsSelector: FC<{
...
+          id,
+          name: metric?.name || id,
+          description: metric?.description || "",
+          datasource: metric?.datasource || "",
+          tags: metric?.tags || [],
+          projects: metric?.projects || [],
+          factTables: [],
+          joinable: true,
+          isGroup: false,
+          disabled: true,
```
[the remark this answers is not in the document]
lukesonnet: This is wrong. `disabled: true` lets you remove it.

packages/back-end/src/models/FactMetricModel.ts
```diff
@@ -467,6 +467,13 @@ export class FactMetricModel extends BaseClass {
     data: FactMetricInterface,
     previousData?: FactMetricInterface,
   ): Promise<void> {
+    if (previousData) {
+      await this.context.models.metricGroups.validateMetricProjectChange(
```
ahdriel: Should we move this into `validateFactMetric` too instead of calling it conditionally here?

It is part of the validation of a fact metric so it makes sense to me.

ahdriel: Also `bulk import` runs `validateFactMetric` directly and not this function, which is a gap the current PR has. So it would also fix that to validate bulk import properly.

packages/front-end/test/footprintParity.test.ts
```diff
@@ -346,3 +347,41 @@ describe("the cancel footprint is the entity's own scope, unwidened", () => {
     expect(constantPublishEnvironments()).toEqual(NO_ENVIRONMENT_BINDING);
   });
 });
+
+describe("metric group edits have no environment footprint", () => {
```
ahdriel: question: why are these tests here? I don't even understand what `footprintParity` is...

should we move it to a `permissions` test file?

packages/shared/src/util/index.ts
```diff
@@ -546,6 +548,74 @@ export function isProjectListValidForProject(
   return projects.includes(project);
 }
 
+export function isProjectListValidForProjects(
```
ahdriel: I find a bit hard to understand the meaning of the function based just on the same, and also confusing how the plural version of it changes a lot on how the checks are performed.

wdyt about

`isAvailableInProject(documentProjects, project)` 
`isAvailableInEveryRequiredProject(documentProjects, requiredProjects)`

packages/back-end/src/routers/metric-group/metric-group.controller.ts
```diff
@@ -102,17 +104,16 @@ export const deleteMetricGroup = async (
 ) => {
```
ahdriel, line 104: (not related to this file, just needed a place)

We also need to expand `cleanupProjectReferences` to clean up metricGroups now that we are being more strict around editing. Otherwise deleting a project might soft-lock a metric group.

packages/shared/src/util/index.ts
```diff
@@ -546,6 +548,74 @@ export function isProjectListValidForProject(
...
+    )
+  ) {
+    return group.metrics;
+  }
+
+  const previousMetrics = new Set(previousGroup.metrics);
+  return group.metrics.filter((id) => !previousMetrics.has(id));
+}
+
+export function doesMetricProjectChangeReduceGroupAvailability(
```
ahdriel, line 603: question: a lot of these functions are named for `metricGroups` specifically, but I think they are answering questions around permissions and how projects relate to it.

any thoughts on how (and if we should) make them more generic?

it might be more specific to metric groups as the only resource with its own permission that holds other resources that have individual permissions as well, so I can see an argument to leave it as is

packages/shared/src/permissions/permissions.constants.ts
```diff
@@ -989,6 +989,7 @@ export const PROJECT_SCOPED_PERMISSIONS = [
   "createSegments",
   "createIdeas",
   "createMetrics",
+  "createMetricGroups",
```
ahdriel, line 992: note that this PR is technically implementing a breaking change for how permissions are handled.

I agree with the direction, but we should tell support when this lands as some users might be impacted and need to update their permissions configs.

packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -98,12 +95,14 @@ export const MetricsSelectorTooltip = ({
...
   datasource?: string;
   project?: string;
+  projects?: string[];
   exposureQueryId?: string;
   selected: string[];
   onChange: (metrics: string[]) => void;
   autoFocus?: boolean;
   includeFacts?: boolean;
   includeGroups?: boolean;
+  preserveSelectedMetrics?: boolean;
```
ahdriel: nit: what about `preserveUnavailableMetrics` ? (very nit, feel free to ignore)

packages/front-end/components/Metrics/MetricGroupDetails.tsx
```diff
@@ -272,25 +278,27 @@ function MetricRow({
...
-                method: "DELETE",
-              },
-            ).then(async () => {
-              if (mutate) {
-                await mutate();
-              }
-            });
-          }}
-        />
+        {canEdit && (
```
ahdriel, line 281: nit: should we show the button in a disabled state instead of hiding it?

packages/shared/src/util/index.ts
ahdriel: Theoretically this also applies to `fact metrics -> fact tables`, `experiments -> metrics`, `config -> other configs` -- not saying we should address in this PR, but conceptually I think it would be worth to have a generic handling on permissions and what model of project ownership is applied to them and their members.

packages/front-end/components/Experiment/MetricsSelector.tsx
```diff
@@ -98,12 +95,14 @@ export const MetricsSelectorTooltip = ({
 const MetricsSelector: FC<{
   datasource?: string;
   project?: string;
+  projects?: string[];
```
ahdriel: ohh I don't quite like this -- very arcane to know the difference between `project` and `projects`

what about `projectFilter` and `requiredProjects` ?

packages/back-end/src/models/FactMetricModel.ts
lukesonnet: done ty

packages/front-end/components/Experiment/MetricsSelector.tsx
lukesonnet: used `contextProject` and `requiredProjects`, thanks.

packages/shared/src/permissions/permissions.constants.ts
lukesonnet: Yes, that's right, thanks.

packages/front-end/test/footprintParity.test.ts
lukesonnet: It's testing no interaction with environments. I'll remove it, it's not important.

packages/shared/src/util/index.ts
lukesonnet: Renamed them, it's still not great.

packages/front-end/components/Metrics/MetricGroupDetails.tsx
lukesonnet: skipping, this should be a dropdown to be in-sync with our design system anyways....

packages/shared/src/util/index.ts
lukesonnet: I think you're right, but I don't know if I want to build them generically and then consider if they actually cover all the edge cases for other callers....

It could be `parent` `children` and generic but that seems overkill. IDK how to indicate to a future developer they could generalize if they wanted, but would then need to make decisions about any vs all type membership.

packages/back-end/src/routers/metric-group/metric-group.controller.ts
lukesonnet: Done, also applied to fact metrics which also had this issue.

packages/front-end/components/Metrics/MetricGroupModal.tsx
```diff
@@ -150,11 +159,13 @@ const MetricGroupModal: FC<{
           </p>
           <MetricsSelector
             datasource={form.watch("datasource")}
+            projects={form.watch("projects")}
```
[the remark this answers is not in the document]
ahdriel: @lukesonnet I believe CI will fail because of this (which is a good thing!); so will need to be addressed before merge.

ahdriel (approved):
🚀

PR #6776, opened 2026-08-27
Fix approval requirements leaking from switched-off rules

bryce-fitzsimons:
A disabled approval rule's stored approver teams were still being inherited by active project overrides, enforcing a requirement that no settings screen showed (reported by a customer). Switched-off rules now take no part in field inheritance, and dormant team associations are scraped off disabled rules on save.

Also adds a permissions summary card to the team page, including which projects/environments the team is a required approver for.

image


gazzdingo (approved):
LGTM

PR #6779, opened 2026-08-27
feat(query-runner): add queryRunnerRun model

ahdriel:
QueryRunner executions now have a durable `QueryRunnerRun` record. Startup persists the complete query roster and target query graph before submitting warehouse work. This removes the race where an execution timer could fire while SQL generation or graph persistence was still in progress.

### Execution and startup

- `QueryRunnerRun.id` identifies an execution, `lockToken` identifies its worker lease, and `targetType` / `targetId` identify the product document receiving results. Several run records may reference the same target.
- Create the run and start its heartbeat before building queries. `startQuery()` creates queued documents and registers callbacks without scheduling warehouse execution.
- Persist the run's query IDs and the target graph, renew the worker lease, then run an immediate serialized readiness pass. Concurrency backoff applies when work needs to wait. Cached queries, dependencies, and `runAtEnd` retain their behavior.
- Keep the atomic queued-to-running claim so competing dispatchers cannot submit the same query twice. Start the query heartbeat only after winning that claim.
- On handled startup failure or confirmed lease loss, fail registered fresh queries without dispatching them. Clean up startup queries before releasing a still-owned worker lease. The existing pending state replaces the separate `dagPersisted` flag.

### Worker lease and warehouse completion

`renewRunLease()` describes the check it actually performs. Successful renewal allows the worker to continue; a confirmed mismatch stops it; an exception leaves ownership unknown and does not authorize dispatch or a target write. Heartbeat, queued execution, and refresh failures use their existing retry paths. Startup errors propagate through startup cleanup. An uncertain shutdown check keeps the refresh watchdog available to retry.

The aggregate-table coverage callback and the incremental pipeline's units, metric-source, and covariate callbacks are now named `onWarehouseSuccess`. They reconcile completed warehouse effects before attempting to accept the query result, including when the query was cancelled or the worker has stopped. Their existing resource execution predicates still decide whether coverage, table pointers, in-flight markers, and watermarks may change. Reconciliation does not restore a cancelled query's status or publish analysis results, and stale resource executions cannot update the newer execution's metadata.

### Guarantees and follow-up work

The stale-run reaper still fails forward. Runs become eligible after five minutes without a worker heartbeat and are handled by a subsequent sweep. The reaper claims the stale run, fails its recorded queued queries and running queries with stale query heartbeats, and applies the existing guarded target-error writes. It leaves running queries with fresh heartbeats alone. Live run records continue to protect their targets from the legacy stale-query sweep. This does not resume warehouse work or reconstruct callbacks. Recovery depends on the persisted roster; a hard process exit during graph construction can still leave query documents that were never recorded in that roster.

A successful worker-lease renewal and a later target write are separate operations. Renewal does not grant exclusive ownership of the target or replace a target-level conditional publication predicate. Existing conditional writes are preserved. Atomic cancellation/replacement checks for late query and analysis publication, staging result chunks before conditionally publishing their pointers, and target-wide run exclusivity remain follow-up work.

Warehouse resource locks remain distinct from the worker lease. Lease loss or an accepted cancel request does not establish that an external warehouse statement stopped. Existing cancel/reaper paths can release resource locks while warehouse work is still in flight. Coordinating that release with confirmed warehouse completion remains follow-up work. This PR preserves execution-guarded reconciliation without claiming to solve warehouse cancellation or lock lifetime.

No execution resumption, workflow engine, exactly-once target publication, additional execution identity, or lock fields across target schemas are added.

### Validation

- 146 tests passed across 12 suites: all query-runner tests, stale-run reaping, snapshot lifecycle, and the existing query/run model suites.
- Fake-timer tests hold graph construction, roster persistence, and target persistence beyond the former execution delay; no warehouse work starts until persistence completes. They also cover startup cleanup, lease loss, transient lease failures and retries, cached reuse, dependencies, and `runAtEnd`.
- MongoDB-backed runner tests prove that competing dispatchers submit one job, rejected late query results still allow valid warehouse reconciliation, and replaced resource executions cannot change aggregate coverage or incremental watermarks.
- Back-end type check and a separate type check of the changed runner tests passed.
- ESLint, Prettier checks on all changed files, and `git diff --check` passed.


ahdriel:
cursor review

packages/back-end/src/jobs/expireOldQueries.ts
```diff
@@ -49,94 +46,395 @@ const STALLED_SNAPSHOT_REAP_LIMIT = 50;
...
+          );
+          continue;
+        }
+
+        const wrote = await TARGET_ERROR_WRITERS[run.targetType]({
+          context,
+          targetDoc,
+          failedQueryIds: run.queryIds,
+          error: LEASE_REAP_ERROR,
+        });
```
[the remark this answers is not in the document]
ahdriel: The process that updates the runner heartbeat is the same one that updates the queries heartbeat, so this cannot happen.

packages/back-end/src/jobs/expireOldQueries.ts
```diff
@@ -49,94 +46,395 @@ const STALLED_SNAPSHOT_REAP_LIMIT = 50;
...
     await context.models.incrementalRefresh
-      .releaseLock(snapshot.experiment, snapshot.id)
+      .releaseLock(targetDoc.experiment ?? "", targetDoc.id)
       .catch((e) =>
         logger.warn(
           e,
           "Failed to release incremental lock for expired snapshot",
         ),
       );
+    return wrote;
```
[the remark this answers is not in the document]
ahdriel: This writer only runs from the reaper, where we do check if there is an active live run before releasing the lock (see `getActiveByTarget`) -- so no need to change anything.

ahdriel:
cursor review

packages/back-end/src/jobs/expireOldQueries.ts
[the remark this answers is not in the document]
ahdriel: Out of scope

ahdriel:
cursor review

packages/back-end/src/models/QueryRunnerRunModel.ts
```diff
@@ -0,0 +1,169 @@
...
+} from "shared/validators";
+import { getCollection } from "back-end/src/util/mongo.util";
+import { MakeModelClass } from "./BaseModel";
+
+export const COLLECTION_NAME = "queryrunnerruns";
+
+/**
+ * A held lock is considered stale when its heartbeat is older than this threshold.
+ */
+export const QUERY_RUNNER_LOCK_STALE_MS = 5 * 60 * 1000;
```
lukesonnet, line 15: ?
```suggestion
export const QUERY_RUNNER_RUN_LOCK_STALE_MS = 5 * 60 * 1000;
```

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -300,17 +308,86 @@ export abstract class QueryRunner<
...
+        this.queryRunnerRun.id,
+        this.queryRunnerRunLockToken,
+      );
+    } catch (e) {
+      logger.warn(
+        { err: e, queryRunnerRunId: this.queryRunnerRun.id },
+        "QueryRunner failed to touch the run lock heartbeat; will retry",
+      );
+      // A failed heartbeat does not prove that this runner lost its lease.
+      return true;
```
lukesonnet, line 324: question: hard for me to see immediately, but do we have a way to make sure a query runner doesn't get stuck in a loop of trying to touch a heartbeat, failing, but then just continuously re-trying? Maybe this is very minor.

packages/back-end/src/models/QueryModel.ts
```diff
@@ -224,9 +221,11 @@ export async function updateQueryIfPending(
     {
       organization: context.org.id,
       id: query.id,
-      status: { $in: ["queued", "running"] },
+      status: "queued",
+    },
+    {
+      $set: { status: "running", startedAt: new Date(), heartbeat: new Date() },
```
lukesonnet, line 227: thank you for this change.

packages/back-end/src/queryRunners/QueryRunner.ts
```diff
@@ -1282,14 +1381,9 @@ export abstract class QueryRunner<
         | ((rows: RowsType) => void | Promise<void>)
         | undefined,
     };
-    if (readyToRun) {
-      this.executeQuery(doc, { run, process, onFailure, onSuccess });
-    } else if (dependenciesComplete && !runAtEnd) {
-      this.runCallbacks[doc.id] = runCallbacksEntry;
+    this.runCallbacks[doc.id] = runCallbacksEntry;
+    if (dependenciesComplete && !runAtEnd) {
       this.queueQueryExecution(doc);
```
lukesonnet, line 1386: Now that all queries run this way, doesn't `queueQueryExecution` do a jittered delay and so aren't ALL queries going to wait now? I think this bug pre-existed in that child queries would wait even if there wasn't a concurrency limit, but now that bug is moving to even the first queries in a queue.

PR #6783, opened 2026-08-28
contextual model selection

lukebrawleysmith:
### Features and Changes

GrowthBook's contextual bandits model uses a tree to group users with the same attribute values with other collections of users.   These groups of users are called "leaves".  Creating leaves helps avoid overfitting and makes manageable the size of the SDK payload.  

Currently this number of leaves is fixed at 12 (if there at least 12 user collections to begin with).  In many instances 12 is too big.  For example, if two different user groups have the same variation dominating, then they can be grouped in the same leaf with no loss of reward.  

This PR uses Bayesian Information Criterion (BIC) to determine how many leaves should be used.

### DIFF
--- a/packages/shared/src/validators/contextual-bandit-event.ts
+++ b/packages/shared/src/validators/contextual-bandit-event.ts
@@ -33,9 +33,22 @@ export type ContextualLeafStatsEntryInterface = z.infer<
 export const contextualSseTrajectoryEntryValidator = z.object({
   numSplits: z.number().int().nonnegative(),
   totalSse: z.number(),
+  /** Per-variation SSE at this stage; sums to `totalSse`. */
+  ssePerVariation: z.array(z.number()),
 });
 export type ContextualSseTrajectoryEntryInterface = z.infer<
   typeof contextualSseTrajectoryEntryValidator
+>;
+
+/** BIC model-selection statistic for one greedy tree split (observability only). */
+export const contextualBicTrajectoryEntryValidator = z.object({
+  numSplits: z.number().int().nonnegative(),
+  logLikelihoodRatio: z.number(),
+  penalty: z.number(),
+  deltaBic: z.number(),
+});
+export type ContextualBicTrajectoryEntryInterface = z.infer<
+  typeof contextualBicTrajectoryEntryValidator
 >;
 
 /** Mirrors gbstats `ContextualBanditResponse`. */
@@ -67,6 +80,7 @@ export const contextualBanditEventValidator = baseSchema
     leaf_map: z.array(contextualLeafMapEntryValidator).optional(),
     leaf_stats: z.array(contextualLeafStatsEntryValidator).optional(),
     sse_trajectory: z.array(contextualSseTrajectoryEntryValidator).optional(),
+    bic_trajectory: z.array(contextualBicTrajectoryEntryValidator).optional(),
     weightsWereUpdated: z.boolean(),
     degreesOfFreedom: z.number().int().nonnegative().optional(),
     // The seed that was applied to the CB when this event was persisted.
--- a/packages/shared/src/validators/contextual-bandit-event.ts
+++ b/packages/shared/src/validators/contextual-bandit-event.ts
@@ -33,8 +33,7 @@ export type ContextualLeafStatsEntryInterface = z.infer<
 export const contextualSseTrajectoryEntryValidator = z.object({
   numSplits: z.number().int().nonnegative(),
   totalSse: z.number(),
-  /** Per-variation SSE at this stage; sums to `totalSse`. */
-  ssePerVariation: z.array(z.number()),
+  ssePerVariation: z.array(z.number()).optional(),
 });
 export type ContextualSseTrajectoryEntryInterface = z.infer<
   typeof contextualSseTrajectoryEntryValidator
--- a/packages/stats-ts/src/banditWeights.ts
+++ b/packages/stats-ts/src/banditWeights.ts
@@ -5,6 +5,8 @@ const BANDIT_PRIOR_VARIANCE = 1e4;
 const BANDIT_PRIOR_PRECISION = 1 / BANDIT_PRIOR_VARIANCE;
 const MIN_VARIATION_WEIGHT = 0.01;
 const MIN_UNITS_PER_VARIATION = 100;
+// Minimum variance for a bandit variation.
+const BANDIT_MIN_VARIANCE = 1e-9;
 
 export type BanditArmStatistic = {
   n: number;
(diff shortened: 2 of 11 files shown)


packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -918,6 +987,63 @@ function buildTree(
...
+  armTotalSampleSizes: number[],
+  penalty: number,
+): { logLikelihoodRatio: number; deltaBic: number } {
+  let logLikelihoodRatio = 0;
+  for (let v = 0; v < armTotalSampleSizes.length; v++) {
+    if (beforePerVariation[v] > 0 && afterPerVariation[v] > 0) {
+      logLikelihoodRatio +=
+        armTotalSampleSizes[v] *
+        Math.log(beforePerVariation[v] / afterPerVariation[v]);
+    }
```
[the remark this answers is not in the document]
lukebrawleysmith: Good catch!  Addressed by using a small positive number instead of 0.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -918,6 +980,64 @@ function buildTree(
...
+  }
+  return { logLikelihoodRatio, deltaBic: penalty - logLikelihoodRatio };
+}
+
+/**
+ * BIC model-selection statistic for each greedy split, derived from the
+ * per-(split, variation) SSE trajectory.
+ *
+ */
+function computeBicTrajectory(
```
lukesonnet: why do we have to re-compute this instead of storing it along the way like we do for the SSE trajectory?

packages/stats-ts/src/banditWeights.ts
```diff
@@ -198,8 +200,8 @@ export function updateVariationWeights(
     };
   }
 
-  const dataPrecision = stats.map((s) =>
-    s.variance > 0 ? s.n / s.variance : 0,
+  const dataPrecision = stats.map(
+    (s) => s.n / Math.max(s.variance, BANDIT_MIN_VARIANCE),
```
lukesonnet, line 204: I don't love the idea of faking the variance here at this high level. Shouldn't this happen closer to wherever the precision is being consumed? Feels like this has side-effects for thompson sampling, not just the issue being addressed in this PR.

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -875,7 +892,52 @@ function buildTree(
...
+    for (let c = 0; c < contexts.length; c++) {
+      if (currentLeaf[c] !== bestLeaf) continue;
+      parentEntries.push(contexts[c]);
+      if (bestGroup.has(contexts[c].tuple[bestAttr])) {
+        movingEntries.push(contexts[c]);
+      } else {
+        stayingEntries.push(contexts[c]);
+      }
+    }
+    const parentSse = sumOfSquaredErrorsPerVariation(
```
lukesonnet: Do we really have to re-compute all of these here? There's a ton of looping around stuff in this PR that feels redundant, but it's hard to follow.

Like can't `LeafSplit` carry most of this information?

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -897,7 +959,7 @@ function buildTree(
         currentLeaf[c] = newLeaf;
       }
     }
-    sseTrajectory.push(totalSse());
+    sseTrajectory.push(totalSsePerVariation());
```
lukesonnet: Isn't this re-computing what we just computed above as well?

packages/stats-ts/src/banditWeights.ts
lukebrawleysmith: It does have side-effects for TS, and I think it improves it.  For example, if suppose we have 2 arms with 1% and 100% conversion rates respectively, and there is n=1MM for each arm.  If the variance is 0 in the second arm, we currently return constant weights, which seems wrong given large n.  Happy to hear further concerns, but I think this will address a potential gap in TS.

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: No, this sums the SSE across leaves and variations, and is the SSE for the whole tree at this stage.

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: Good callout, and LeafSplit now carries parent and child SSE.

packages/stats-ts/src/contextualBanditWeights.ts
lukebrawleysmith: refactored so it is stored along the way

packages/stats-ts/src/contextualBanditWeights.ts
```diff
@@ -918,6 +980,64 @@ function buildTree(
...
+  penalty: number,
+): { logLikelihoodRatio: number; deltaBic: number } {
+  let logLikelihoodRatio = 0;
+  for (let v = 0; v < armTotalSampleSizes.length; v++) {
+    const before = beforePerVariation[v];
+    // No baseline variance => no improvement to measure (ln(0/0) is undefined).
+    if (before <= 0) continue;
+    const after = Math.max(afterPerVariation[v], before * SSE_RATIO_FLOOR);
+    logLikelihoodRatio +=
+      armTotalSampleSizes[v] * (Math.log(before) - Math.log(after));
```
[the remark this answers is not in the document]
lukebrawleysmith: Not a defect — this is the BIC penalty working as intended in a degenerate case, so we're leaving the floor as-is.
If there are only 2 observations (or only a few) in one variation while another variation has 1MM observations, this is not a split we should chase.  We want to split only if there are moderate sample sizes per variation.

packages/stats-ts/src/contextualBanditWeights.ts
lukesonnet: Isn't that `afterPerVariation`?

lukebrawleysmith: yes - pushed fix

packages/shared/src/validators/contextual-bandit-event.ts
```diff
@@ -33,11 +33,24 @@ export type ContextualLeafStatsEntryInterface = z.infer<
 export const contextualSseTrajectoryEntryValidator = z.object({
   numSplits: z.number().int().nonnegative(),
   totalSse: z.number(),
+  /** Per-variation SSE at this stage; sums to `totalSse`. */
+  ssePerVariation: z.array(z.number()),
```
lukesonnet: This may not exist on old objects. Should this be optional? Or do we need to add a `migrate` method?

lukebrawleysmith: good callout, pushed fix

lukesonnet (approved):

PR #6793, opened 2026-08-31
Pre-download and cache the mongod binary in CI

Kevin-Chant:
### Features and Changes

Nothing puts a mongod on a CI box before Jest starts, so a back-end shard's three workers race to download the same 70MB binary and whichever one loses the fight over mongodb-memory-server's lockfile takes its suite down with it. That's what failed a shard on #6789. Now we fetch the binary once up front and cache it, so the workers just find it.

`MONGOMS_VERSION` pins the version the cache key names, so it needs bumping whenever mongodb-memory-server's default moves.

### Testing

- [x] `test-back-end` shards no longer log `Downloading MongoDB` during the `Test` step

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,9 @@ jobs:
     runs-on: depot-ubuntu-24.04-4
     # A silently killed Jest worker leaves the parent waiting on the 6h default.
     timeout-minutes: 15
+    env:
+      # mongodb-memory-server's own default, pinned so the cache key is exact.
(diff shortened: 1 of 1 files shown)


jdorn:
Alternatively, now that we have Depot, we could skip the shards entirely and just use a beefier machine.  Claude estimates using `depot-ubuntu-24.04-16` would end up finishing faster AND costing less, while also getting rid of any weirdness caused by sharding.

ahdriel (approved):

PR #6795, opened 2026-08-31
Stop contact autofill on experiment name fields

alyssamichelle:
# Features and Changes

Browsers treated **Experiment Name** as a person's name because the input used `name="name"` from `form.register("name")`, so autofill offered saved contacts.

Override the HTML `name` to `experimentTitle` (and `banditTitle` for bandits) and set `autoComplete="off"`. React Hook Form still tracks and submits the `name` field.

Touched the simple create flow, the legacy create form, create-from-feature, and the edit-info modals.

# Screenshots

Before (live site): browsers offer saved contacts in the Experiment Name field.

Create Experiment name field offering saved contacts


msamper (changes requested):
I think the fix here makes sense, but the pattern is somewhat fragile in that it's opt-in and we have several forms across the app that have `name` fields. I wonder if we could have the `Field` component handle this. It looks like there are some fields where we do want to trigger `autoComplete` (i.e. in `Welcome.tsx`) and we pass it in as a prop so maybe that can be our signal for whether or not to disable auto complete.

PR #6796, opened 2026-08-31
Sql browser ux improvements

mknowlton89:
### Features and Changes

Schema Browser used a hidden double-click to paste a table path into SQL. That was easy to miss and I don't think many users know about it.

This PR replaces it with hover actions on tables and columns: copy, and insert SQL. Clicking a table still only opens the column preview.

And upon hovering on a table, the user has two buttons, both include a tooltip to give the user context. The first button allows the user to copy the table path (e.g. the table may be `events`, but the sql query needs the full page. e.g. `datasource.myschema.events`. 

The second button updates the query in the text area with a simple select statement e.g. `SELECT * FROM <full.table.path>`

When it comes to the new functionality for the actual table columns, we're got two buttons. The first is another copy button, it just copies the name of the table's column so the user can paste it in. This is handy if the org doesn't have autocomplete on, or if the column name is long. 

The second button inserts the column into the SELECT statement of the query, and handles trailing commas correctly.

### Dependencies

None

### Follow-ups
I'll put up a PR that updates the code text area toolbar and adds an easy `undo` button.

### Testing

- [x] Open SQL Explorer (or Edit SQL with a schema browser) against a warehouse that has an information schema
- [x] Click a table name: preview opens; the editor does not change
- [x] Double-click a table name: path is not pasted
- [x] Hover a table: copy and insert icons appear on the right; selected row is purple, not gray; icons stay readable in light and dark mode
- [x] Copy table: tooltip “Copy full table path”, then “Copied”; clipboard is the full path
- [x] Table insert on an empty editor: `SELECT * FROM <path>`
- [x] Table insert on `SELECT * FROM other` or a query with `WHERE`: editor is overwritten with `SELECT * FROM <path>`
- [x] Hover a column: copy and insert sit next to the name; type stays on the right
- [x] On `SELECT * FROM this_table`, Add to SELECT replaces `*` with the column; a second column appends; the same column again is a no-op
- [x] On a JOIN / `WHERE` query, Add to SELECT overwrites with `SELECT col FROM this_table`
- [x] CTE, `UNION`, or multiple `SELECT`s: Add to SELECT is disabled; tooltip “This query is too complex. Copy the column name instead.”
- [x] Insert table A, then click table B to view its columns: Add to SELECT is disabled; tooltip “This query selects from a different table. Copy the column name instead.”; copy still works
- [x] JSON sub-fields (`attributes.foo`): copy only, no insert icon
- [x] After clicking an icon, move the pointer off the row: icons hide

### Screenshots / Video Walkthrough

https://www.loom.com/share/912f47d70e65474d8217b84a6c141315

Updated insert tooltip copy:
Screenshot 2026-08-31 at 11 35 51 AM


packages/front-end/components/SchemaBrowser/SchemaBrowser.tsx
```diff
@@ -381,25 +351,47 @@ export default function SchemaBrowser({
...
+                                          value={tablePath}
+                                          idleTooltip="Copy full table path"
+                                        />
+                                        {updateSqlInput ? (
+                                          <SchemaSqlInsertButton
+                                            tooltip={`Insert SELECT * FROM ${tablePath} query into editor`}
+                                            onClick={() => {
+                                              selectTable(table.id, tablePath);
+                                              updateSqlInput(
+                                                `SELECT * FROM ${tablePath}`,
```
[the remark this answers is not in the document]
mknowlton89: This is not what I want.

packages/front-end/components/SchemaBrowser/SchemaBrowserSqlActions.module.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.columnTable tbody tr:hover td {
+  background-color: var(--accent-a3);
+}
+
+.actions {
+  display: flex;
+  align-items: center;
+  flex-shrink: 0;
+  opacity: 0;
```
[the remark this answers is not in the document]
mknowlton89: I don't like this

packages/front-end/components/SchemaBrowser/DatasourceTableData.tsx
```diff
@@ -256,24 +268,48 @@ export default function DatasourceSchema({
       }
     >
       <div style={{ overflow: "auto", height: "100%" }}>
-        <table className="table table-sm">
+        <table className={`table table-sm ${actionStyles.columnTable}`}>
```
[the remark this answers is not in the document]
mknowlton89: This is intentionally out of scope for this PR

packages/front-end/components/SchemaBrowser/DatasourceTableData.tsx
```diff
@@ -256,24 +271,50 @@ export default function DatasourceSchema({
...
+                            {updateSqlInput && !column.jsonField ? (
+                              <SchemaSqlInsertButton
+                                tooltip="Add to SELECT"
+                                disabled={!!columnInsertDisabledReasonText}
+                                disabledTooltip={
+                                  columnInsertDisabledReasonText ?? undefined
+                                }
+                                onClick={() => {
+                                  updateSqlInput(
+                                    insertColumnIntoSelect(
```
nhat-growthbook, line 310: Maybe we want to disable this button if the column is already in Select, with the tooltip to avoid confusion. The select list could get long and people might not know that the column is already added in the list.

packages/front-end/components/SchemaBrowser/DatasourceTableData.tsx
```diff
@@ -256,24 +271,50 @@ export default function DatasourceSchema({
       }
     >
       <div style={{ overflow: "auto", height: "100%" }}>
-        <table className="table table-sm">
+        <table className={`table table-sm ${actionStyles.columnTable}`}>
```
nhat-growthbook: Oh maybe you can use `clsx` here instead.

packages/front-end/services/schemaBrowserSql.ts
```diff
@@ -0,0 +1,86 @@
...
+    .replace(/;+\s*$/, "")
+    .trim();
+  if (!stripped) return false;
+  if (stripped.includes(";")) return true;
+  if (/^\s*WITH\b/i.test(stripped)) return true;
+  if (/\bUNION\b/i.test(stripped)) return true;
+  return (stripped.match(/\bSELECT\b/gi)?.length ?? 0) > 1;
+}
+
+export function columnInsertDisabledReason(
```
nhat-growthbook, line 53: To combine with my comment above, this should check if the column already in the SELECT list and return with a tooltip message.

packages/front-end/components/SchemaBrowser/DatasourceTableData.tsx
```diff
@@ -256,24 +271,50 @@ export default function DatasourceSchema({
...
                 {filteredColumns?.map((column) => {
                   return (
                     <tr key={`${table.tableName}:${column.columnName}`}>
                       <td className="pl-3">
-                        {column.jsonField ? (
+                        <div className={actionStyles.row}>
                           <span
-                            className="text-muted"
-                            style={{ paddingLeft: 16 }}
+                            className={`${actionStyles.label}${
```
nhat-growthbook: Ditto here `clsx` will help clean this up I think.

packages/front-end/components/SchemaBrowser/SchemaBrowser.tsx
```diff
@@ -381,25 +352,59 @@ export default function SchemaBrowser({
...
-                                        )
+                                      onClick={() =>
+                                        selectTable(table.id, tablePath)
                                       }
+                                      onKeyDown={(e) => {
+                                        if (e.target !== e.currentTarget) {
+                                          return;
+                                        }
+                                        if (
+                                          e.key === "Enter" ||
```
nhat-growthbook: Just a nit but I'm assuming these keys comparison would be some useful helper utils.

packages/front-end/components/SchemaBrowser/SchemaBrowserSqlActions.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+function ActionTooltip({
+  content,
+  children,
+}: {
+  content: string;
+  children: ReactNode;
+}) {
+  return (
+    <Tooltip content={content} disableHoverableContent>
```
nhat-growthbook, line 15: Not in scope for this PR but I think we should add Tooltip support to the primitive Buttons, especially in Disabled state. That would be very useful.

nhat-growthbook (commented):
Nothing blocking so I'll just approve. Just a few comments and nitpicks.
I think we should add a bit of right padding between the background block and the Insert SQL button
Image
Can we make the buttons square? It will save some horizontal space.
Image

nhat-growthbook (approved):

PR #6800, opened 2026-08-31
fix(experiments): ping Slack when a scheduled results update fails

### LINKED ISSUE #5924: [Feature] Better error alerting on data source failures.
### Description of Feature

Better error alerting on data source failures.- it would be useful to have some kind of monitoring or alerting for failed queries in the future.

The customer was facing timeout errors on Trino and was not alerted of this

EmmS21:
## Summary
- Fixes #5924

Scheduled refresh failed and Slack never fired. I was passing `success: true`. That is the ticket. I fixed that.

`autoSnapshots` is what keeps running the Results update loop. When a scheduled Results update does not finish, this is set to `false` to stop the loop. I inferred that this is done to avoid running the same failing query again from code comments.

After that save, `notifyAutoUpdate` was called with `success: true` because the save worked, and Slack does not ping on `true`. I change that to `success: false` so Slack pings.

To replicate this, I started with local Postgres to trigger a timeout and noticed `success` remained set to `true`.

I then ran Trino (435 - running on an older machine) with the same client GrowthBook users. Trino’s time limit came back as Query exceeded maximum time limit of 2.00s. GrowthBook’s wait cutoff came back as execution error:query timed out. Both land in `e.message` on the query.

Greptile was watching Slack. Either it did not fire when Results failed, or it fired again for a fail I already warned about.

I kept treating a row I wrote as proof Slack already went out, or that Results already worked. Those writes can miss, land too early, or mean something else, so the next run guessed wrong.

So I stopped adding a note on the experiment. Slack already writes a fail Event when it pings. A working refresh already writes a successful snapshot. First fail Slacks. If that Event is there and there is no newer snapshot, I stay quiet. If a snapshot lands after the Event, the next fail Slacks again.

I also stopped treating an experiment edit as a new start, and I stopped writing a success Event. A bandit can finish the Results queries and still fail when it saves the bandit settings. I still Slack on that fail. I do not count this run’s snapshot as Results working. If I save the Event and never queue Slack, I delete the Event so the next fail does not look already pinged.

## Test plan
- [ ] Scheduled Results update fails (timeout or query error) → Slack: `Automatic snapshot creation for <experiment name> failed!` plus experiment link
- [ ] Slack does not include `e.message` / the warehouse timeout string
- [ ] `autoSnapshots` still turns off for a normal experiment so the failing query is not retried on a loop
- [ ] Overlapping refresh (lock) still does not turn `autoSnapshots` off and still does not Slack
- [ ] Same fail again, Results have not worked since → no second Slack
- [ ] Results work, then a later fail → Slack again
- [ ] Bandit: Results queries finish, bandit save fails → Slack. Retryable bandits stay on the schedule
- [ ] Experiments that never failed still do not Slack on a working refresh


EmmS21:
@greptileai

EmmS21:
@greptileai

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@greptile

EmmS21:
@alyssamichelle Please review. For added context

Greptile was watching Slack. Either it did not fire when Results failed, or it fired again for a fail we already warned about.

I kept treating a row I wrote as proof Slack already went out, or that Results already worked. Those writes can miss, land too early, or mean something else, so the next run guessed wrong.

I stopped adding a note on the experiment. Slack already writes a fail Event when it pings. A working refresh already writes a successful snapshot. First fail Slacks. If that Event is there and there is no newer snapshot, nothing happens. If a snapshot lands after the Event, the next fail Slacks again.

I stopped treating an experiment edit as a new start, and stopped writing a success Event. A bandit can finish the Results queries and still fail when it saves the bandit settings. I still Slack on that fail. I don’t count this run’s snapshot as Results working. If I save the Event and never queue Slack, we delete the Event so the next fail does not look already pinged.

alyssamichelle:
@EmmS21 I ran ponytail and thermonuclear on this. The Slack bug is real (`success: false` on fail, and stop tying that ping to the `autoSnapshots` save). The rest is extra.

Please drop:
- deleting the Event in `createEventWithPayload` if notify throws
- `getLatestAutoUpdateFailEvent` on EventModel
- optional `phase` on `getLatestSuccessfulSnapshot`
- `ignoreSnapshot` / hoisted `currentSnapshot`
- the leftover `success: true` no-op

`pastNotifications` can’t re-alert after Results recover, so don’t go back to that. If we keep the streak (first fail Slacks; quiet until a successful snapshot; then Slack again), do it with `startedAt` on the job plus `experiment.lastAutoUpdateFailNotifiedAt` written only after notify succeeds. Pass the current phase. Don’t change the shared Event or snapshot helpers.

PR #6805, opened 2026-09-01
Pricing phase 1: Pro tier limits

gazzdingo:
### Features and Changes

Builds the **Pro** tier on the same stamp + `pricing-phase-1-limits` flag system that shipped for Free in #6325.

| | Free | Pro | Enterprise |
|---|---|---|---|
| Projects | 1 | **3** | Unlimited |
| Custom environments | ✗ | **✗** | ✓ |
| Role management | ✗ | **✓** | ✓ |

Sample data projects still don't count toward the project cap, and grandfathering is unchanged: orgs created before limits shipped carry no stamp and stay unlimited on every plan.

**Keeping Enterprise clear**, via three new commercial features:

- `role-management` — Pro and above
- `unlimited-projects` — Enterprise only
- `custom-environments` — Enterprise only

These short-circuit the limit checks, so a plan that owns the feature can never be restricted by a stamp or a license snapshot. They also give the upsell UI the correct plan badge and CTA for free — the environment upsells now point at `custom-environments`, so they route to Enterprise rather than Pro.

Two drive-by fixes that were wrong under the new pricing: the Vercel Pro billing plan advertised unlimited projects (now derived from the constants), and the plan-availability notes in the docs.

### Dependencies

The `pricing-phase-1-limits` flag needs a targeting rule serving Pro's values to `accountPlan` in `[pro, pro_sso]`. Without it Pro orgs fall back to the hardcoded `PRO_ORG_LIMITS`, which already match the spec, so this can ship in either order.

### Testing

- `packages/shared/test/enterprise/entitlements.test.ts` — rewritten around plan tiers: Pro upgrades a stamped org off the frozen free values, license snapshots vs. tier defaults, Pro keeps roles, Enterprise ignores both stamp and license snapshot, grandfathering on every plan.
- `packages/shared/test/pricing-limits.test.ts` — per-tier fallback on the flag resolver.
- Full `shared` (2750), `back-end` (8324) and `front-end` (1115) suites pass. Two back-end suites failed only under parallel workers (SIGSEGV / socket hang up) and pass in isolation.

Three assertions in the old entitlements test changed on purpose: Pro now keeps role management even if a license snapshot revokes it, and Enterprise is no longer limited by a license snapshot.

### DIFF
--- a/packages/back-end/src/services/plan-limits.ts
+++ b/packages/back-end/src/services/plan-limits.ts
@@ -19,9 +19,8 @@ import {
 import { IS_CLOUD } from "back-end/src/util/secrets";
 
 // Limits stamped onto a newly created org. Cloud reads the flag; self-hosted
-// always uses the hardcoded defaults. New orgs always start on the free tier,
-// so the stamp holds free values — its presence is also what marks the org as
-// subject to plan limits at all (missing = grandfathered, on every plan).
+// always uses the hardcoded defaults. Every org is created free, so the stamp
+// holds free values; its presence is what marks the org as non-grandfathered.
 export async function getStampedOrgLimits(): Promise<OrgLimits> {
   if (!IS_CLOUD) return { ...FREE_ORG_LIMITS };
 
@@ -34,9 +33,8 @@ export async function getStampedOrgLimits(): Promise<OrgLimits> {
   return resolveOrgLimitsConfig(raw);
 }
 
-// The flag evaluated for this org, with trusted server-derived attributes only
-// (no request context, so the SDK's query-string override cannot influence
-// enforcement). Self-hosted never reads the CDN for limits.
+// Trusted server-derived attributes only — no request context, so the SDK's
+// query-string override cannot influence enforcement.
 function evalLimitsFlagForOrg(org: OrganizationInterface): unknown {
   if (!IS_CLOUD) return undefined;
   return getGrowthBookClient()?.evalFeature(PRICING_PHASE_1_FLAG_KEY, {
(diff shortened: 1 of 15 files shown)


packages/back-end/src/services/plan-limits.ts
```diff
@@ -30,22 +32,30 @@ export async function getStampedOrgLimits(): Promise<OrgLimits> {
   return resolveOrgLimitsConfig(raw);
 }
 
-function isPricingLimitsDisabledForOrg(org: OrganizationInterface): boolean {
-  if (!IS_CLOUD) return false;
-
-  const raw = getGrowthBookClient()?.evalFeature(PRICING_PHASE_1_FLAG_KEY, {
+function evalLimitsFlagForOrg(org: OrganizationInterface): unknown {
```
andywhite37, line 35: ❓ curious if `unknown` is deliberate here? Do the things that use this value expect to have to parse unknown values? I see it being used down below - just curious why it's not a known type. No need to change anything as it seems like that's how this has worked previously - mostly just wanting to understand the history.

packages/shared/src/enterprise/pricing-limits.ts
andywhite37, line 31: Ok, I see how it gets parsed here

andywhite37 (commented):
I reviewed the changes to get some exposure to this, but I think someone with more background should probably approve for safety.

andywhite37:
Ok, I did some manual testing to check some of the changes in the web app, and I had AI do some testing of the APIs with different configurations. I think everything looked good

# AI results

Note: the one behavior it called out that seemed odd was the `400` 

```
Results — 41/42 API assertions + 45/45 unit tests passed

All four orgs, all three enforcement gates, plus four scenario tests. Everything ran against the live back-end (ProjectModel.beforeCreate, util/environments.ts, assertRoleAssignmentAllowed, ApiKeyModel), not the UI.

Per-plan gates — expected and actual matched on every row except one:

┌─────────────────────────┬──────────────┬─────────────┬────────────┬───────────────┐
│                         │ Free starter │ Pro pro_sso │ Enterprise │ Grandfathered │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ Stamp on org            │ free         │ free        │ free       │ none          │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ Projects                │ 1 → 402      │ 3 → 4th 402 │ 5+ all OK  │ 5+ all OK     │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ Env, default id (test)  │ allowed      │ allowed     │ allowed    │ allowed       │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ Env, custom id (qa-1)   │ 402          │ 402         │ allowed    │ allowed       │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ Invite non-admin role   │ 402          │ allowed     │ allowed    │ allowed       │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ Invite admin role       │ allowed      │ allowed     │ allowed    │ allowed       │
├─────────────────────────┼──────────────┼─────────────┼────────────┼───────────────┤
│ API key, non-admin role │ blocked 400  │ allowed     │ allowed    │ allowed       │
└─────────────────────────┴──────────────┴─────────────┴────────────┴───────────────┘
```

The one finding

The org API key role gate returns 400 Bad Request, while every other plan-limit rejection returns 402 Payment Required — same message, different status. It's in ApiKeyModel.ts:150 (throwBadRequestError), and git log -S traces it to #6325, so it's pre-existing rather than introduced here. It matters because this PR extends that gate's reach to Pro orgs, and anything keying off 402 to trigger an upsell will miss this path. Worth a one-line fix in this PR or a follow-up.

andywhite37 (approved):

PR #6823, opened 2026-09-01
Add metric form type translation and reset-rule logic

najito:
Map stored metric definitions to 12 editor types and back. Preserve retention window types and fractional intervals, and flag lookback windows that the editor cannot represent.

Base of the metric-editor stack.

Validation: the updated stack passes front-end and back-end type checks and 438 targeted tests. Hosted code checks pass. Preview deployment fails on a build-service keepalive timeout. UI walkthrough not performed.

Review note: funnel settings remain cached during editing; save and preview payloads remove them for non-funnel metrics.

Readiness: ready for review.


packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+
+// columnsFor(shape, factTable) from the spec. [] means omit the Column field.
+// "distinct" also needs datasource.properties.hasCountDistinctHLL — where
+// that's false, the spec says to hide the shape entirely; returning no
+// columns here is what makes that rule enforceable ("everything else follows
+// from columnsFor(shape).length > 0", per the spec's Gates section).
+export function columnsForShape(
+  shape: RatioShape,
+  factTable: MinimalFactTable,
+  hasCountDistinctHLL = false,
```
jdorn: Accept a `SqlDialect` here instead and make it required.  Can use `Pick<SqlDialect, "hasCountDistinctHLL">` to make it easier to mock for testing.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  ) {
+    return [];
+  }
+  const timestampColumn = getFactTableTimestampColumn(factTable);
+  const userIdTypes = factTable.userIdTypes ?? [];
+  const columns = factTable.columns.filter(
+    (c) =>
+      !c.deleted &&
+      c.column !== timestampColumn &&
+      !userIdTypes.includes(c.column),
```
jdorn: This is not correct anymore. Now, fact tables can define a `userIdColumns` mapping.  So `userIdTypes` might be `["user_id"]`, but the `c.column` might actually be `uid` instead.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  currentColumn: string,
+  hasCountDistinctHLL = false,
+): string {
+  if (shape === "count") return "$$count";
+  if (shape === "days") return "$$distinctDates";
+  if (shape === "users") return "$$distinctUsers";
+  const candidates = columnsForShape(shape, factTable, hasCountDistinctHLL);
+  return candidates.includes(currentColumn)
+    ? currentColumn
+    : (candidates[0] ?? "");
```
jdorn: If someone picked a column that was valid at the time, but is no longer a candidate, will this break their metric? Should we keep it as-is and warn instead in the UI?

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+// deterministically (shapeFromColumnRef is the reverse mapping), so a
+// read-only display can derive "Per-User Aggregation" from shape alone
+// rather than needing the stored ColumnRef's own aggregation field.
+export function aggregationForShape(
+  shape: RatioShape,
+): ColumnAggregation | undefined {
+  if (shape === "sum") return "sum";
+  if (shape === "max") return "max";
+  if (shape === "distinct") return "count distinct";
+  return undefined; // count / days / users carry no aggregation
```
jdorn: Many of these functions can be made more future-proof by using exhaustive switch statements (or similar).  Right now, if we add a new "shape", it's really easy to forget to update all of these methods and it will just silently fall back to the default behavior.  The idea is to make it a compile-time error if you forget to explicitly handle a new shape everywhere.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  newFactTableId: string,
+  factTable: MinimalFactTable,
+  hasCountDistinctHLL = false,
+): ColumnRef {
+  const shape = shapeFromColumnRef(current) ?? "sum";
+  return {
+    ...current,
+    factTableId: newFactTableId,
+    column: fitColumn(shape, factTable, current.column, hasCountDistinctHLL),
+    rowFilters: [],
```
jdorn: Use `getInitialInlineFilters` instead of hard-coding empty row filters.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+    aggregation: aggregationForShape(newShape),
+  };
+}
+
+// Per-part refit for a fact-table change. Which parts to call this on for a
+// given type (one vs. every part) is a MetricEditor (PR 3) orchestration call.
+export function onFactTableChange(
+  current: ColumnRef,
+  newFactTableId: string,
+  factTable: MinimalFactTable,
```
jdorn, line 187: Feels weird to pass a fact table id separate from the fact table interface

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  hasCountDistinctHLL = false,
+): ColumnRef {
+  if (newScope === "event") {
+    return {
+      ...current,
+      column: fitColumn("sum", factTable, current.column),
+      aggregation: undefined,
+    };
+  }
+  const shape = shapeFromColumnRef(current) ?? "sum";
```
jdorn: I think we'd probably want to fall back to `count`, not `sum`.  Count is always possible, but sum requires specific column types to exist first.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  >,
+): number {
+  return (
+    windowSettings.delayValue +
+    convertDuration(
+      windowSettings.windowValue,
+      windowSettings.windowUnit,
+      windowSettings.delayUnit,
+    )
+  );
```
jdorn: Feels like this should standardize to the smaller of the time units.

If `delayUnit` is days, but `windowUnit` is hours, right now 1 day + 1 hour will result in `1.0416666667 days`.  Instead, we could have it result in `25 hours`

Also address below in `onRetentionDelayOrModeChange`

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+// Single source of truth with SHAPE_FORM_TYPES/storedTypeAndNumeratorFor
+// below - reuses both rather than re-encoding the five types a third time.
+export function shapeForValueType(type: FormMetricType): Shape | undefined {
+  if (!typeHasShape(type)) return undefined;
+  const spec = storedTypeAndNumeratorFor(type).numerator;
+  return spec.kind === "shape" ? spec.shape : undefined;
+}
+
+export function cappingOk(type: FormMetricType): boolean {
+  return type === "ratio" || typeHasShape(type);
```
jdorn: Slight nuance - only percentile capping is allowed for ratio.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  windowSettings?: MetricWindowSettings;
+};
+
+const DEFAULT_QUANTILE = 0.5;
+
+function defaultFunnelStep(name: string, factTableId: string): FunnelStep {
+  return {
+    name,
+    factTableId,
+    rowFilters: [],
```
jdorn: This might not be the right place, but when someone adds a new step to a funnel, it should use `getInitialInlineFilters(factTable)`` to seed the `rowFilters` and not just always default to an empty array.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+// already hides the field it would need to be edited back to "".
+function resetCappingForTypeSwitch(
+  cappingSettings: MetricCappingSettings | undefined,
+  newFormType: FormMetricType,
+): MetricCappingSettings | undefined {
+  if (!cappingSettings) return cappingSettings;
+  const clear =
+    !cappingOk(newFormType) ||
+    (newFormType === "ratio" && cappingSettings.type === "absolute");
+  return clear ? { ...cappingSettings, type: "" } : cappingSettings;
```
jdorn: Minor: If someone switches to ratio and they had absolute capping before, they probably still want capping to be enabled.  I don't know if it's better to wipe it out (like we do here) or switch them to percentile automatically.  Also need to figure out how/where to tell them this changed.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+  }
+  return windowSettings;
+}
+
+export function applyFormType<T extends MetricTypeSwitchState>(
+  current: T,
+  newFormType: FormMetricType,
+  factTable: MinimalFactTable,
+  hasCountDistinctHLL = false,
+): T {
```
jdorn, line 661: Is it worth an early return here for when the new type is identical to the current type? Most of the transformations below are safe, so it should be no-op today, but there's risk of data loss if we aren't really careful.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+            ],
+          };
+    return {
+      ...current,
+      metricType,
+      numerator: null,
+      denominator: null,
+      quantileSettings: null,
+      funnelSettings,
+      ...(cappingSettings !== undefined && { cappingSettings }),
```
jdorn, line 696: I don't think funnels support cappingSettings

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+      ...(windowSettings !== undefined && { windowSettings }),
+    };
+  }
+
+  const base: ColumnRef = current.numerator ?? {
+    factTableId: "",
```
jdorn: Why not use the passed in `factTable` if available?

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+    column: "",
+    rowFilters: [],
```
jdorn: Similar here - if they pass in a `factTable`, we can call `getInitialInlineFilters(factTable)` to get the initial row filters.

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
```diff
@@ -0,0 +1,744 @@
...
+    // Only ratio/quantile carry a denominator/quantileSettings - clear the
+    // other's field on every switch, not just spread the stale value
+    // forward, so it doesn't reach save on a type that never showed a
+    // control to edit it back.
+    denominator:
+      newFormType === "ratio"
+        ? (current.denominator ?? {
+            factTableId: numerator.factTableId,
+            column: "$$count",
+            rowFilters: [],
```
jdorn: Same here about `getInitialInlineFilters`

packages/front-end/components/FactTables/MetricEditor/metricFormTranslation.ts
najito: They don't. I'm just always sending in `cappingSettings` with the `type = ""` and handling that.

PR #6827, opened 2026-09-02
fix(api): don't 400 the experiments list on a phase with no start date

### LINKED ISSUE #3841: [Bug] Growthbook API producing 400 request on certain offset/limit pairs
### Summary

I'm trying to query the `/experiments` endpoint of the Growthbook API to get the list of all our experiments. Certain experiments (e.g offset=233, limit=1) produce a 400 Bad Request response with the following body:

```
{
    "message": "Cannot read properties of null (reading 'toISOString')"
}
```

The message suggests a failure in server-side logic, and should ideally return a 500 code. 

I am setting only the `limit` and `offset` parameters. 

I have also raised this issue in [Growthbook Slack](https://growthbookusers.slack.com/archives/C01T6Q1SVFV/p1738752964795079).

#### Expected Behavior

A typical API response for the `experiments` endpoint, as an example:

```json
{
    "experiments": [
        {
            "id": "exp_19g61ml5jdk1xt",
            "name": "Gringotts | Security Framework | Email Login Notification Variants v1",
....
    ],
    "limit": 1,
    "offset": 100,
    "count": 1,
    "total": 1310,
    "hasMore": true,
    "nextOffset": 101
}
```

#### Current Behavior

API returns 400 Bad Request with body:

```
{
    "message": "Cannot read properties of null (reading 'toISOString')"
}
```

avionicharshit-byte:
### Features and Changes

`toExperimentApiInterface` calls `p.dateStarted.toISOString()` on every phase with no null check, while the `dateEnded` line directly below it already guards. an experiment whose phase has no stored `dateStarted` throws a `TypeError` there, and `app.ts` turns any error without a status into a 400 carrying the raw message, so the caller gets `{"message": "Cannot read properties of null (reading 'toISOString')"}` and it reads as a client error.

serialization runs per page, so one such experiment breaks whichever page it lands on and there is no way to page past it. that is why #3841 saw it on specific offset/limit pairs like `offset=233, limit=1`.

this guards `dateStarted` the same way `dateEnded` already is, so a missing date serializes as `""`. `controllers/metrics.ts` builds the same phase shape for the experiments-using-this-metric views and had the same unguarded call twice, so those are guarded too.

not touching the 400 vs 500 part of the report. that default lives in the shared error handler and applies to every endpoint, so it belongs in its own PR.

- Closes #3841

### Dependencies

none

### Testing

added `packages/back-end/test/experimentApiPhaseDates.test.ts`, one phase with a start date and one with a null start date. the null case fails on main with `TypeError: Cannot read properties of null (reading 'toISOString')` and passes with the guard.

also ran the experiment and metric back-end suites, 32 suites and 545 tests, all green.

### Screenshots

none, no UI change.

### DIFF
--- a/packages/back-end/src/controllers/metrics.ts
+++ b/packages/back-end/src/controllers/metrics.ts
@@ -607,7 +607,7 @@ export const getMetricExperimentResults = async (
     dateUpdated: e.dateUpdated.toISOString(),
     phases: e.phases.map((p) => ({
       ...p,
-      dateStarted: p.dateStarted ? p.dateStarted.toISOString() : "",
+      dateStarted: p.dateStarted?.toISOString(),
       dateEnded: p.dateEnded?.toISOString(),
     })),
     nextScheduledStatusUpdate: e.nextScheduledStatusUpdate
@@ -707,7 +707,7 @@ export const getMetricNorthstarData = async (
     dateUpdated: e.dateUpdated.toISOString(),
     phases: e.phases.map((p) => ({
       ...p,
-      dateStarted: p.dateStarted ? p.dateStarted.toISOString() : "",
+      dateStarted: p.dateStarted?.toISOString(),
       dateEnded: p.dateEnded?.toISOString(),
     })),
     nextScheduledStatusUpdate: e.nextScheduledStatusUpdate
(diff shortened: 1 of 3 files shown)


jdorn (approved):

ahdriel:
Thank you for your contribution @avionicharshit-byte 🚀

PR #6829, opened 2026-09-02
Allow custom values on enum and multiselect custom fields

bryce-fitzsimons:
Adds an "Allow custom values" option to enum and multiselect custom fields, letting users type in values beyond the predefined list - in the UI and via the REST API.

image


oelshaikh (approved):
LGTM, small UX idea: add a warning if updating a custom field from createable -> not creatable that says that any current custom values will persist.

PR #6840, opened 2026-09-02
feat(queries): confirm warehouse cancel and log when unconfirmed

ahdriel:
### Features and Changes

When we cancel a query on the warehouse, we now verify the cancellation actually happened and log a `warn` when it did not. The query is still marked cancelled on our side right away; the check is fire-and-forget and only produces logs.

To do that we also added a new ability to get the current status of the external query for Integrations that support it.

### Engineering notes

- The `SqlIntegration` no-op default for `cancelQuery` is removed. Absence of the method is the "unsupported" signal and both callers already guard on it.
- We add a patch to `snowflake-sdk` to make `isStillRunning`/`isAnError` accept `string`. This just changes types to reflect the current code behavior (we already had casts on our code that we could clean up by patching the sdk too).

### Testing

- Unit tests
- Manual tests
  - Cancelling against real bigquery still works, and force failure to cancel logs the warning 30s later


ahdriel:
cursor review

packages/back-end/src/integrations/Presto.ts
```diff
@@ -110,8 +139,30 @@ export default class Presto extends SqlIntegration {
...
+    return new Promise<ExternalQueryStatus>((resolve) => {
+      client.query(externalId, (error, data) => {
+        if (error) {
+          // Trino returns 404/410 once a query id has aged out of the
+          // coordinator's memory.
+          const code = error.code;
+          resolve(
+            code === 404 || code === 410
+              ? { state: "unknown", reason: "expired" }
+              : { state: "unknown", reason: "unreachable" },
```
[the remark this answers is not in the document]
ahdriel: `error.code` is set here -- Client.prototype.query wraps the error into `{ message, error, code }`

packages/back-end/src/integrations/Presto.ts
```diff
@@ -27,6 +29,31 @@ const PRESTO_QUERY_TAG_MAX_LENGTH = 2000;
...
+  state: z.string().min(1),
+  failureInfo: z.object({ message: z.string().nullish() }).nullish(),
+  errorCode: z.object({ name: z.string().nullish() }).nullish(),
+});
+
+// FINISHED, FAILED, and FAILING are terminal (FAILING is a failed query whose
+// tasks are still being torn down); every other reported state is still
+// executing. An unparseable payload means we can't tell, so return unknown
+// rather than assuming the query is alive.
+export function prestoStateToStatus(info: unknown): ExternalQueryStatus {
```
msamper: Why is info `unknown`?

packages/back-end/src/integrations/Presto.ts
```diff
@@ -27,6 +29,31 @@ const PRESTO_QUERY_TAG_MAX_LENGTH = 2000;
...
+  const parsed = prestoQueryInfoSchema.safeParse(info);
+  if (!parsed.success) return { state: "unknown", reason: "unrecognized" };
+  const { state, failureInfo, errorCode } = parsed.data;
+  if (state === "FINISHED") return { state: "succeeded" };
+  if (state === "FAILED" || state === "FAILING") {
```
msamper: nit: I don't think `FAILING` is a real state for [Presto](https://raw.githubusercontent.com/prestodb/presto/master/presto-main-base/src/main/java/com/facebook/presto/execution/QueryState.java) or [Trino](https://raw.githubusercontent.com/trinodb/trino/master/core/trino-main/src/main/java/io/trino/execution/QueryState.java)

msamper (commented):
lgtm! Just one small nit I found with a claude review, but not blocking

msamper (approved):

packages/back-end/src/integrations/Presto.ts
ahdriel: Nice catch! Likely a mix up with `TaskState` (not `QueryState`). Cleaned up.

PR #6849, opened 2026-09-03
fix: multiple exposures with pre-computed dimensions

lukesonnet:
### Features and Changes

`multipleExposures` with dimension rows was wrong on snapshots. We only used this for customers in the UI on the main snapshot which traditionally did not have dimensions, but with precomputed dimensions it is possible. So users with pre-computed dimensions had incorrect counts of multiple exposures.

This fixes that by correcting both the metric data derived multiple exposures as well as by introducing multiple exposures counts from the health traffic query, preferring the latter when it exists. We now sum across all variation === "__multiple__" rows when relevant.

### Testing

Tested manually with pre-computed dimensions, replicating the bug and seeing that it is resolved now.


msamper (approved):
Thanks for fixing! lgtm

PR #6851, opened 2026-09-03
Replace AGPL dependencies and add a license check in CI

Kevin-Chant:
### Features and Changes

Replaces two AGPL-3.0 license packages - `dirty-json` and `pm2`

`dirty-json` -> `jsonrepair`: It returns a repaired string rather than a value, so the call sites go through a `parseLooseJSON` helper. It handles everything `dirty-json` did (that matters, `{"a": +1}` now rejects) plus comments and truncated JSON, and it's fast enough to drop `MEDIUM_FILE_SIZE`, which only existed to keep `dirty-json` off anything over 1KB. A 35KB malformed value goes **842ms -> 1.25ms**.

`pm2` -> `scripts/run-apps.js`: We only used a fraction of `pm2`'s functionality, so the replacement is lightweight. The image keeps both `pm2-runtime` paths as shims onto it since published charts and ECS task defs pin them, and the `PM2_*` env vars keep their names.

A new `licenses` CI job checks the installed tree against an allowlist, so a new AGPL dependency fails the build.

Behavior changes:
* The back-end was running in pm2 cluster mode, as `instances: 1` with no `exec_mode` selects it. It's a plain child process now.
* PM2 config we weren't using is now unsupported, and logs on startup if caught
```
[run-apps] clustered: instances=4 unsupported, starting one
[run-apps] clustered: ignoring exec_mode
```
* Invoking PM2 against the source checkout (outside a docker image) is no longer supported
* Supervisor lines go to stderr, so stdout stays purely the apps' own output (and matches PM2)

### Testing

- [x] `docker stop` on the built image -> graceful shutdown, exit 0
- [x] Old pinned command `node node_modules/pm2/bin/pm2-runtime start ecosystem.config.js --only front-end` -> serves 200
- [x] `TRACING_PROVIDER=datadog` -> `dd` fields in back-end logs, so `--require` still resolves
- [x] Kill the app inside the container -> supervisor exits 2
- [x] Over `max_memory_restart` while ignoring SIGTERM -> SIGKILL, then restart
- [x] Preview topology, idle-monitor signals its ppid -> all three apps stop, exit 0
- [x] `POST /api/v2/features` with `{"a": 1 // note}` -> stored as `{"a": 1}`; `{"a" = = =}` -> 400


scripts/run-apps.js
```diff
@@ -0,0 +1,192 @@
...
+
+const toArray = (value) =>
+  Array.isArray(value) ? value : value ? String(value).split(/\s+/) : [];
+
+function parseMemory(value) {
+  if (typeof value === "number") return value;
+  const match = String(value ?? "").match(/^(\d+(?:\.\d+)?)\s*([KMG])?B?$/i);
+  if (!match) return null;
+  const scale = { K: 1024, M: 1024 ** 2, G: 1024 ** 3 };
+  return Math.round(+match[1] * (match[2] ? scale[match[2].toUpperCase()] : 1));
```
jdorn, line 44: Worth a sanity check here to avoid ending up with like 5 bytes of memory. Clamp to reasonable min/max values. Possibly throw if way off

scripts/run-apps.js
```diff
@@ -0,0 +1,192 @@
...
+  const child = spawn(
+    process.execPath,
+    [...toArray(app.node_args), app.script, ...toArray(app.args)],
+    {
+      cwd: app.cwd ? path.resolve(rootDir, app.cwd) : rootDir,
+      stdio: "inherit",
+    },
+  );
+
+  running.set(app.name, child);
```
jdorn, line 166: Any chance of this already being set and an orphan child process being left around?

scripts/run-apps.js
```diff
@@ -0,0 +1,192 @@
...
+    "usage: run-apps.js start <ecosystem.config.js> [--only <app>]",
+  );
+  process.exit(1);
+}
+
+const resolvedConfig = path.resolve(configPath);
+const rootDir = path.dirname(resolvedConfig);
+const apps = (require(resolvedConfig).apps || []).filter(
+  (app) => !only || app.name === only,
+);
```
jdorn, line 73: Is it worth a sanity check to make sure all apps have a name and they are all unique?

jdorn (commented):
Code looks good. Will definitely want to test this thoroughly, especially looking for backwards compatibility breaks.

tzjames (approved):

PR #6856, opened 2026-09-03
fixing undefined start date

lukebrawleysmith:
Fix spurious "Analysis start date changed" and guard analysis when a phase has no start date.  Behavior stays unchanged for normally-configured experiments; the guard only trips for legacy/corrupt data with no resolvable phase start.

Problem
When an experiment phase has no start date, the incremental full-refresh comparison normalizes absent dates via getValidDate(), which returns a fresh new Date() on each call. Two absent dates therefore resolve to different timestamps, deterministically (and incorrectly) reporting Analysis start date changed.

Changes
shared/enterprise/pipeline.ts: Normalize an absent startDate to a stable null sentinel instead of a fabricated timestamp; make startDate optional on IncrementalFullRefreshComparable to reflect that legacy snapshots may lack it.
front-end/.../AnalysisSettingsSummary.tsx: Stop fabricating new Date() for the current-side start when there's no started phase, so it stays symmetric with an absent baseline.
back-end/util/migrations.ts: In upgradeExperimentDoc, backfill a missing phase dateStarted from the previous phase's dateEnded (or the experiment's dateCreated for the first phase). Runs on every read.
back-end/services/experiments.ts: In getSnapshotSettings, throw a clear, user-facing error when a phase still has no resolvable start date (first phase with no dateCreated) — blocking analysis rather than producing a fabricated window. Surfaces on-screen when the user clicks Update Results.
Tests: Add coverage for the absent/one-sided startDate comparison and the migration backfill (both-absent, first-phase fallback, prior-phase fallback, and no-op when already set).

Example error state when phase start date cannot be resolved. 

Screenshot 2026-09-03 at 1 49 51 PM

### DIFF
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -615,6 +615,13 @@ export function getSnapshotSettings({
   if (!phase) {
     throw new Error("Invalid snapshot phase");
   }
+  // upgradeExperimentDoc backfills a missing phase start date.
+  // If the backfill fails, throw an error.
+  if (!phase.dateStarted) {
+    throw new Error(
+      "This experiment phase has no start date, so results can't be analyzed. Set a start date for the phase and try again.",
+    );
+  }
 
   const defaultPriorSettings = orgPriorSettings ?? {
     override: false,
--- a/packages/back-end/src/services/experiments.ts
+++ b/packages/back-end/src/services/experiments.ts
@@ -619,7 +619,7 @@ export function getSnapshotSettings({
   // If the backfill fails, throw an error.
   if (!phase.dateStarted) {
     throw new Error(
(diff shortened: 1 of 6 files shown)


lukesonnet (changes requested):
Why would dateStarted be unset for a running experiment? I would think we should fix that instead.

lukebrawleysmith:
@lukesonnet 
phase.dateStarted is not enforced to be present at storage:
https://github.com/growthbook/growthbook/blob/8899dd8d755a8f4053191093c7cdd7df8fb5ec45/packages/back-end/src/services/experiments.ts#L3062

If an experiment is loaded with no phase.dateStarted, a read-time migration now backfills it
1. use the end date for previous phase (if exists) as start date for current 
2. otherwise prevent analysis and ask user to modify phase

lukebrawleysmith:
unneeded for now, as all current experiments have start date

PR #6864, opened 2026-09-04
Introduces new raw table for sql-explorer

mknowlton89:
### Features and Changes

SQL explorer currently only offers an aggregated visualization (Count, Group By, etc.) This PR extends the SQL explorer v2 by adding a **Raw table** view so users can add an unaggregated chart to a dashboard, like they can for the original sql explorer, while retaining the benefits of the exploration tools date filtering. 

- New `rawTable` chart type (SQL-only): wraps the user’s query in `SELECT *` with the existing row limit, optional timestamp date filter, and no grouping or aggregates.
- Results come back as `rawRows` (plus `truncated` when the warehouse returned more than the limit). Hidden columns are a display-only preference and do not refetch.
- After a successful test query, users can **Customize table** (raw results + column checkboxes) or **Create visualization** (seed Count / chart as before). Raw table is the default SQL explore state.
- Comparisons are disabled for raw tables. Switching chart types keeps draft values/dimensions so going back to a visualization is reversible; those fields are stripped before persist/run.
- Empty SQL canvas uses a **Load table** / **Render chart** submit instead of a duplicate Refresh callout. Dashboards and AI chat bubbles render raw tables with the same data table.

### Dependencies

None

### Testing

- [x] Run a SQL test query, then **Customize table**: sidebar shows column checkboxes (last visible column cannot be hidden), main pane loads unaggregated rows, CSV download respects hidden columns
- [x] From the same preview, **Create visualization**: Count is seeded and charts/tables work as before; switch back to Raw table without losing the draft
- [x] With a timestamp column: date range filters raw rows; compare is hidden. Without a timestamp: date controls stay disabled with the raw-table tooltip
- [x] Change SQL so a previously visible column disappears: hidden-column list is pruned; if every remaining column would be hidden, visibility resets
- [x] Truncated results show “the first N rows”; Refresh still works after SQL or date-range edits
- [x] Save a raw-table exploration to a dashboard and confirm the block renders the table (no comparison)
- [x] Confirm metric / dataset / funnel explorers are unchanged (no Raw table option)

### Screenshots
Screenshot 2026-09-09 at 9 31 28 AM


jdorn (approved):
Did not test, but read through the code and it looks low risk.

PR #6865, opened 2026-09-04
fix(fact-tables): avoid mutating columns during detection

itsgrimetime:
### Features and Changes

- Reconcile detected columns against a copy so the input Fact Table remains unchanged.

### What this fixes

`runColumnDetectionQuery` mutated `factTable.columns` in place and returned the same array. Both callers (`refreshFactTableColumns` job and `putFactTable`) then used the same `factTable` in `updateFactTableColumns` as the "before" state, so every before/after comparison saw no difference.

The Mongo write itself was correct; only the change-driven side effects were lost.

### Testing

- Added regression coverage for input mutation.

### DIFF
--- a/packages/back-end/src/services/factTableColumns.ts
+++ b/packages/back-end/src/services/factTableColumns.ts
@@ -1,4 +1,5 @@
 import chunk from "lodash/chunk";
+import cloneDeep from "lodash/cloneDeep";
 import {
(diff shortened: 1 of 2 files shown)


ahdriel (approved):

ahdriel:
Thanks for the fix!

PR #6866, opened 2026-09-04
fix(models): check permissions for no-op updates

itsgrimetime:
### Features and Changes

- Run update permission checks before returning from no-op updates.

### Testing

- Added API coverage for empty and identical update payloads.

### DIFF
--- a/packages/back-end/src/services/rampScheduleEvaluator.ts
+++ b/packages/back-end/src/services/rampScheduleEvaluator.ts
@@ -31,6 +31,7 @@ import {
   RampAdvanceLockBusyError,
 } from "back-end/src/util/errors";
 import { logger } from "back-end/src/util/logger";
+import { getContextForAgendaJobByOrgObject } from "back-end/src/services/organizations";
 
 export type EvalDecision =
   | { action: "advance" }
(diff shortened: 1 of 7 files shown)


packages/back-end/src/services/rampScheduleEvaluator.ts
```diff
@@ -534,9 +534,12 @@ export async function applyRampEvaluationDecision(
       nextSnapshotAt: schedule.nextSnapshotAt,
       cutoffDate: schedule.cutoffDate,
     });
-    return ctx.models.rampSchedules.updateById(schedule.id, {
-      nextProcessAt,
-    });
+    // Scheduler bookkeeping, not a user edit: the snapshot that triggered this
+    // may run under a member who cannot edit or publish the feature.
+    return ctx.models.rampSchedules.dangerousUpdateByIdBypassPermission(
```
ahdriel: question: do we need to apply the same fixes to `rampSchedule.ts`?

This file calls a lot of functions like `advanceStep` which are defined on that file and internally use `rampschedules.updateById`.

If the answer is yes, should we remove this change from here and open a separate PR?

ahdriel (approved):

packages/back-end/src/services/rampScheduleEvaluator.ts
Kevin-Chant: I went with a different direction that I think is more holistically correct. Instead of adding bypassPermission calls to a bunch of random places, those knock-on effects will get funneled through their own job-level context so it completely detaches from user permissions.
https://github.com/growthbook/growthbook/pull/6866/changes/83339430ca9ff1475cb0bba3f3280ebebf219083

PR #6867, opened 2026-09-04
fix(fact-tables): preflight and guard auto-slice cleanup

itsgrimetime:
### Features and Changes

- Preflight all auto-slice cleanup updates before writing.
- Use compare-and-swap updates to preserve concurrent Fact Metric changes.

### Testing

- Added API coverage for permission failures and concurrent updates.


itsgrimetime:
@greptile

Kevin-Chant:
closing in favor of #6891. There's a different direction that I think works better here

PR #6868, opened 2026-09-04
fix(metrics): await metric creation before registering tags

itsgrimetime:
### Features and Changes

- Wait for metric creation to succeed before registering its tags.

### Testing

- Added an API regression test for rejected metric creation.


itsgrimetime:
@greptile

MattCassar (approved):

ahdriel (approved):

